PR #1, opened 2021-11-30
Add isEmpty() to Check for Empty Value

3KeyPradeep:
Issue:

During the creation of a new client, when uploading a certificate to create a new client, the client is not created

Steps to Reproduce:
1. Login to CZERTAINLY
2. Navigate to Client Page
3. Create a new client
4. Upload a new certificate
5. Click on Create

Root Cause:
When creating a new client, the code checks for the input to determine either to use the existing certificate or upload a new certificate. During this process, the code checks if the existing check UUID is null instead of isEmpty(). Hence the code tries to fetch the certificate with empty uuid and fails

Fix:
- Updated the check from null checking to isEmpty check
- Updated to upload the certificate and validate it
- Updated edit client operation as it has same issue

### DIFF
--- a/src/main/java/com/czertainly/core/service/impl/ClientServiceImpl.java
+++ b/src/main/java/com/czertainly/core/service/impl/ClientServiceImpl.java
@@ -7,6 +7,7 @@ import java.util.stream.Collectors;
 
 import javax.transaction.Transactional;
 
+import com.czertainly.api.core.modal.*;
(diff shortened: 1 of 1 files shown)


3keyroman (changes requested):
Make sure that you forked the `develop` branch and not committing master changes related to the releases.

3keyroman (approved):

PR #155, opened 2022-09-22
User management flow updates and fixes

3KeyPradeep:


src/main/java/com/czertainly/core/service/impl/UserManagementServiceImpl.java
```diff
@@ -126,4 +118,22 @@ public List<RoleDto> getUserRoles(String userUuid) {
...
+
+        Certificate certificate;
+        if (StringUtils.isNotBlank(certificateUuid)) {
+            certificate = certificateService.getCertificateEntity(SecuredUUID.fromString(certificateUuid));
+        } else {
+            X509Certificate x509Cert = CertificateUtil.parseCertificate(certificateData);
+            try {
+                certificate = certificateService.getCertificateEntityBySerial(x509Cert.getSerialNumber().toString(16));
+            } catch (NotFoundException e) {
+                logger.debug("New Certificate uploaded for admin");
```
lubomirw, line 132: Should be user in the message, not admin

3KeyPradeep: It should be. I will update it

src/main/java/com/czertainly/core/service/impl/AuthServiceImpl.java
```diff
@@ -28,8 +33,11 @@ public class AuthServiceImpl implements AuthService {
 
     @Override
     @AuditLogged(originator = ObjectType.FE, affected = ObjectType.ACCESS, operation = OperationType.REQUEST)
-    public UserDto getAuthProfile() throws NotFoundException {
-        return userManagementApiClient.getUserProfile().getData().getUser();
+    public UserDto getAuthProfile() throws NotFoundException, JsonProcessingException {
+        CzertainlyUserDetails userDetails = (CzertainlyUserDetails) SecurityContextHolder.getContext().getAuthentication().getPrincipal();
```
lubomirw, line 37: In case of support of anonymous user, handle case of not authenticated user.
If no authentication or principal is present, throw exception with message of Unknown or anonymous user

lubomirw (changes requested):

src/main/java/com/czertainly/core/service/impl/AuthServiceImpl.java
3KeyPradeep: log and Message updated

lubomirw (approved):

PR #497, opened 2023-07-13
Add sending timestamp and target object identification to notification

### LINKED ISSUE #496: Add sending timestamp and target object identification to notification
- add following columns to `Notification` entity and its corresponding DTO(s)
  - sent_at - `timestamp`
  - target_object_type - `Resource` enum
  - target_object_identification - `text`
- sort listed notifications automatically by sent_at timestamp from newest to oldest

Target object (type + identification) contains data identifying object that notification is dealing with. That data can be used to get all info necessary to identify object. Use the same principle as in `ScheduledJobHistory` entity.

dmaixner-moro:
close #496

### DIFF
--- a/src/main/java/com/czertainly/core/service/impl/NotificationServiceImpl.java
+++ b/src/main/java/com/czertainly/core/service/impl/NotificationServiceImpl.java
@@ -5,6 +5,7 @@ import com.czertainly.api.exception.ValidationException;
 import com.czertainly.api.model.client.notification.NotificationDto;
 import com.czertainly.api.model.client.notification.NotificationRequestDto;
 import com.czertainly.api.model.client.notification.NotificationResponseDto;
+import com.czertainly.api.model.core.auth.Resource;
 import com.czertainly.api.model.core.auth.UserDto;
 import com.czertainly.core.dao.entity.Notification;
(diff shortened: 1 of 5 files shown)


lubomirw (approved):

src/main/java/com/czertainly/core/service/impl/NotificationServiceImpl.java
```diff
@@ -44,16 +45,18 @@ public class NotificationServiceImpl implements NotificationService {
...
     }
 
     @Override
-    public NotificationDto createNotificationForUsers(String message, String detail, List<String> userUuids) throws ValidationException {
+    public NotificationDto createNotificationForUsers(String message, String detail, List<String> userUuids, Resource target, String targetUuids) throws ValidationException {
         Notification notification = new Notification();
         notification.setUuid(UUID.randomUUID());
         notification.setMessage(message);
         notification.setDetail(detail);
+        notification.setTargetObjectType(target);
```
lubomirw, line 58: We need to also set sentAt timestamp to current timestamp

lubomirw (changes requested):

lubomirw (approved):

Issue #490, opened 2023-06-29
Implement approvals and approval flow based on approval profile configuration


lubomirw:
### Approvals

- Implement paginated listing of user's approvals with query parameter to filter only pending approvals. List approval when:
  - if authenticated user can act on it - he is requested in current step or member of group/role 
  - if authenticated user already approved to view it but not to change his decision
- Implement paginated listing of approvals with `SecurityFilter` to list approvals based on Access Control
- detail page of approvals with status of each step with users recipients
- approve/reject user actions that control approval flow

### Approval flow

Implement approval flow with following characteristics
- user cannot approve action which he requested
- user can approve action only once across all steps
- reject by eligible user makes whole approval rejected
- if there is no more eligible users to approve because of incorrect configuration (only one user in group that also requested action, etc...), reject whole approval

PR #505, opened 2023-07-18
CZERTAINLY-Core #490 - updated ApprovalController by listing User Approvals + small refactoring of naming convention

moro-lukasrejha:


moro-lukasrejha:
wrong branches

Issue #514, opened 2023-07-31
Integrate notification sending with external notification providers


lubomirw:
Notifications are currently handled in `NotificationListener` by sending internal notifications only. Integrating it with external notification providers gives to user flexibility how to notify in case of different events.
- based on notifications settings, if there is configured notification instance for some notification type, used that notification instance to send notification by notification provider
- if notification recipient is not existing, skip him
- if no recipients to notify externally, do not send any request to notification provider
- integrate with `/notify` endpoint of notification provider

PR #521, opened 2023-08-04
External notifications integration

klaraf755:
Implements #514


src/main/java/com/czertainly/core/messaging/listeners/NotificationListener.java
```diff
@@ -37,6 +83,12 @@ public void processMessage(NotificationMessage notificationMessage) {
                     case TEXT -> {
                         NotificationDataText data = (NotificationDataText) notificationMessage.getData();
                         sendInternalNotifications(data.getText(), data.getDetail(), notificationMessage);
+                        if (notificationTypeStringMap != null) {
```
3keyroman, line 86: Those ifs are duplicates, it is same for every case.

src/main/java/com/czertainly/core/messaging/listeners/NotificationListener.java
```diff
@@ -69,12 +139,82 @@ public void processMessage(NotificationMessage notificationMessage) {
...
                     }
                 }
             }
         }
     }
 
+
+    private void sendExternalNotifications(String notificationInstanceUUID, NotificationMessage notificationMessage, NotificationType notificationType) {
+        Optional<NotificationInstanceReference> notificationInstanceReference = notificationInstanceReferenceRepository.findByUuid(UUID.fromString(notificationInstanceUUID));
+        if (notificationInstanceReference.isPresent()) {
```
3keyroman, line 157: This if checking the notification instance reference will only skip if it is not available, but requested, see the warn message as part of the else.

What about something like:
```java
Optional<NotificationInstanceReference> notificationInstanceReference = notificationInstanceReferenceRepository.findByUuid(UUID.fromString(notificationInstanceUUID))
.orElseThrow(() -> new NotFoundException(NotificationInstanceReference.class, uuid));
```
?

src/main/java/com/czertainly/core/messaging/listeners/NotificationListener.java
```diff
@@ -69,12 +139,82 @@ public void processMessage(NotificationMessage notificationMessage) {
...
+                            }
+                        }
                     }
                 }
             }
         }
     }
 
+
+    private void sendExternalNotifications(String notificationInstanceUUID, NotificationMessage notificationMessage, NotificationType notificationType) {
```
3keyroman, line 155: I would suggest to use `UUID` instead of `String` for `notificationInstanceUuid`

src/main/java/com/czertainly/core/messaging/listeners/NotificationListener.java
```diff
@@ -69,12 +139,82 @@ public void processMessage(NotificationMessage notificationMessage) {
...
+                if (recipient.getRecipientType().equals(RecipientTypeEnum.USER)) {
+                    UUID recipientUuid = recipient.getRecipientUuid();
+                    try {
+                        UserDetailDto userDetailDto = userManagementApiClient.getUserDetail(recipientUuid.toString());
+                        NotificationRecipientDto user = new NotificationRecipientDto();
+                        user.setEmail(userDetailDto.getEmail());
+                        user.setName(userDetailDto.getUsername());
+                        recipientsDto.add(user);
+                    } catch (Exception e) {
+                        logger.warn(String.format("User with UUID %s was not reached, notification was not sent for this user.", recipientUuid));
```
3keyroman, line 170: Although there is an exception, the notification is sent

src/main/java/com/czertainly/core/messaging/listeners/NotificationListener.java
```diff
@@ -69,12 +139,82 @@ public void processMessage(NotificationMessage notificationMessage) {
...
+                    try {
+                        List<UserDto> userDtos = roleManagementApiClient.getRoleUsers(roleUuid.toString());
+                        for (UserDto userDto : userDtos) {
+                            NotificationRecipientDto user = new NotificationRecipientDto();
+                            user.setEmail(userDto.getEmail());
+                            user.setName(userDto.getUsername());
+                            recipientsDto.add(user);
+                        }
+                    } catch (Exception e) {
+                        logger.warn(String.format("Role with UUID %s was not found, notification was not sent for this role."), roleUuid);
```
3keyroman, line 184: Although there is an exception, the notification is sent

src/main/java/com/czertainly/core/messaging/listeners/NotificationListener.java
```diff
@@ -69,12 +139,82 @@ public void processMessage(NotificationMessage notificationMessage) {
...
+                if (recipient.getRecipientType().equals(RecipientTypeEnum.GROUP)) {
+                    UUID groupUuid = recipient.getRecipientUuid();
+                    Optional<Group> group = groupRepository.findByUuid(groupUuid);
+                    if (group.isPresent()) {
+                        NotificationRecipientDto groupDto = new NotificationRecipientDto();
+                        groupDto.setName(group.get().getName());
+                        groupDto.setEmail(group.get().getEmail());
+                        recipientsDto.add(groupDto);
+                    } else {
+                        logger.warn(String.format("Group with UUID %s was not reached, notification was not sent for this group."), groupUuid);
```
3keyroman, line 196: Although there is an exception, the notification is sent

src/main/java/com/czertainly/core/messaging/listeners/NotificationListener.java
klaraf755: I tried to change it, but it can't resolve uuid in the exception

src/main/java/com/czertainly/core/messaging/listeners/NotificationListener.java
klaraf755: I think the best way to fix it is to make the getNotificationSettings method return empty hash set instead of null when there are no settings, can I change it in this pull request?

src/main/java/com/czertainly/core/messaging/listeners/NotificationListener.java
klaraf755: changed it

3keyroman (approved):

PR #589, opened 2023-10-31
Change how certificate DN strings are stored, represented and compared

### LINKED ISSUE #583: Change how certificate DN strings are stored, represented and compared
In actual situation, DN strings (issuer or subject DN) are stored as-is in string format, where value is provided by BouncyCastle or Java classes itself. That could cause issues in case when we are comparing or searching certificates by DN string because of different formatting of individual DN name parts (RDNs). There are following caveats when comparing two DN strings from different sources
- ordering of RDNs
- white spaces
- different string formats of RDNs values
- delimiter (`,` or `;`)
- use of different alias - e.g.: Email RDN can be represented as `E=email@example.com` or `EMAIL=email@example.com`
- use of `OID` instead of name alias - e.g.: `CN=testcert` or `2.5.4.3=testcert`

Hence it is necessary to define form (format) in which DN string will be stored by `CZERTAINLY` to be represented consistently. In that way, it will be possible to compare two DN names strings with certainty of correct result of comparison.

> **Note**
> We can implement custom `X500NameStyle` of `BouncyCastle` package or extending existing implementations. Custom style can than be used to format DNs and store it in same style.
> Worth to consider is to use different approach to comparison and transform all DN parts to OID representation to be always represented by same identificator and compare afterwards.

### Tasks
- [x] **Define enum `RdnType` of common RDNs used in DNs.** It should contain following properties:
  - Code - default alias of RDN type, e.g.: **CN**
  - Label - label of RDN, e.g.: **CommonName**
  - OID - corresponding OID of RDN, e.g. for CN is **2.5.4.3**
  - Aliases - array of alternative aliases, e.g. for Email (EMAIL) it is [E, EMAIL, EMAILADDRESS], for StateOrProvinceName (ST) it is [S, ST]
  - Add following RDN types with their default alias:
  1. Email (EMAIL)
  2. Common Name (CN)
  3. Organization Unit (OU)
  4. Organization (O)
  5. Locality (L)
  6. State (ST)
  7. Domain Component (DC)
  8. Country (C)
- [x] **X500NameStyle requirements**
- settings for ordering
  - `Last-to-first` (LDAP or forward) ordering. Example: `CN=Common Name, O=Organization, C=Country`
  - `First-to-last` (reverse) ordering. Example: `C=Country, O=Organization, CN=Common name`
- utility static method to create X500Principal from String to be able to style it
- default style to format DN - use to store certificate's issuer and subject DN
  - use `Last-to-first` ordering
  - algorithm of styling to String:
    1. cycle through RDNs from last
    2. if OID of RDN is found in enum `RdnType`, use its default alias
    3. if not found, use `defaultSymbols` inherited from `BCStyle` parent
    4. if no default alias, style should use OID to represent RDN
- normalization style to format DN - use to store normalized form of certificate's issuer and subject DN for searching and comparing purposes
  - convert RDN to its OID
  - sort RDNs by OID alphabetically
  - sort multiple AVAs of RDN alphabetically by their value
- [x] **DB migration**
- format `issuer_dn` and `subject_dn` of certificate table and `subject_dn` of certificate_request table with default style
- add new columns `issuer_dn_normalized` and `subject_dn_normalized` and set them to formatted `issuer_dn` and `subject_dn` with normalized style
- add new column `subject_dn_normalized` to certificate request table and set them to formatted `subject_dn` with normalized style
- [x] **Usage of X500NameStyle**
- when creating certificate from `X509Certificate` object, format DNs with default style and normalization style to store it in DB
- when trying to identify certificate by DN string, first normalize it and search by normalized column (case of `UpdateIntuneRevocationRequestsTask#processRevocationRequests`)

klaraf755:
Close #583

### DIFF
--- a/src/main/java/com/czertainly/core/api/web/SettingControllerImpl.java
+++ b/src/main/java/com/czertainly/core/api/web/SettingControllerImpl.java
@@ -4,17 +4,33 @@ import com.czertainly.api.interfaces.core.web.SettingController;
 import com.czertainly.api.model.core.settings.NotificationSettingsDto;
 import com.czertainly.api.model.core.settings.PlatformSettingsDto;
 import com.czertainly.api.model.core.settings.SettingsSection;
+import com.czertainly.core.dao.entity.Certificate;
+import com.czertainly.core.dao.repository.CertificateRepository;
 import com.czertainly.core.service.SettingService;
+import com.czertainly.core.util.CertificateUtil;
+import com.czertainly.core.util.X500NameStyleCustom;
 import com.czertainly.core.util.converter.SettingsSectionCodeConverter;
+import org.bouncycastle.asn1.x500.X500Name;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.web.bind.WebDataBinder;
 import org.springframework.web.bind.annotation.InitBinder;
 import org.springframework.web.bind.annotation.RestController;
 
+import javax.security.auth.x500.X500Principal;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+import java.util.List;
+import java.util.Objects;
+
 @RestController
 public class SettingControllerImpl implements SettingController {
 
     private SettingService settingService;
+    @Autowired
+    private CertificateRepository certificateRepository;
+    private static final Logger logger = LoggerFactory.getLogger(SettingControllerImpl.class);
 
     @Autowired
     public void setSettingService(SettingService settingService) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/czertainly/core/util/X500NameStyleCustom.java
3keyroman, line 1: Shouldn’t this class be named something like `CzertainlyStyle`? The custom name is invoking that user can customize the name, however, it will be handled by CZERTAINLY for internal purposes to make DNs consistent, right?

src/main/java/com/czertainly/core/api/web/SettingControllerImpl.java
```diff
@@ -45,4 +62,31 @@ public NotificationSettingsDto getNotificationsSettings() {
     public void updateNotificationsSettings(NotificationSettingsDto notificationSettingsDto) {
         settingService.updateNotificationSettings(notificationSettingsDto);
     }
+
+    private void testCustomDnTest(){
```
3keyroman, line 66: Why we have a test method in main codebase?

lubomirw (changes requested):

lubomirw (approved):

PR #600, opened 2023-11-16
Add an option to mark CA certificate as trusted

### LINKED ISSUE #592: Add an option to mark CA certificate as trusted
Currently, `CZERTAINLY` has mechanism how to get and update certificate chain. But from validation viewpoint, root CA certificates (if their signature is valid) are automatically trusted and certificate is marked as `Valid`.

For the sake of security, it is more suitable to give that control to the user. User should have an option to mark certificate as trusted, if certificate is CA certificate. In addition, having list of trusted certificates allows to make other integration and security operations (e.g. to ensure that trusted certificates are also accepted by infrastructure). 

Requirements:
- update `CertificateDto` to include new property
- add new column to `certificate` table - `trusted_ca` which is nullable boolean. (Null is for non CA certificates, otherwise `false` as a start). For sake of migration, it could be based on column `basic_constraints`.
- add check if CA certificate is trusted to `X509CertificateValidator`
- add search filter based on `trusted_ca` property
- add option to set if certificate is trusted to endpoint `PATCH /v1/certificates/{uuid}` as new property to request DTO and handle it in service

Based on that, validation in addition should check if CA certificate in chain is also trusted before marking it as `Valid` for certificate chain check.

klaraf755:
Closes #592

### DIFF
--- a/src/main/java/com/czertainly/core/dao/entity/Certificate.java
+++ b/src/main/java/com/czertainly/core/dao/entity/Certificate.java
@@ -14,7 +14,6 @@ import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.lang.Nullable;
 
 import java.io.Serializable;
 import java.time.LocalDateTime;
--- a/src/main/java/com/czertainly/core/dao/entity/Certificate.java
+++ b/src/main/java/com/czertainly/core/dao/entity/Certificate.java
@@ -51,8 +51,14 @@ public class Certificate extends UniquelyIdentifiedAndAudited implements Seriali
     @Column(name = "issuer_dn")
     private String issuerDn;
 
+    @Column(name = "issuer_dn_normalized")
+    private String issuerDnNormalized;
+
     @Column(name = "subject_dn")
     private String subjectDn;
+
+    @Column(name = "subject_dn_normalized")
+    private String subjectDnNormalized;
 
     @Column(name = "not_before")
     private Date notBefore;
@@ -428,6 +434,22 @@ public class Certificate extends UniquelyIdentifiedAndAudited implements Seriali
 
     public void setSubjectDn(String subjectDn) {
         this.subjectDn = subjectDn;
+    }
+
+    public String getIssuerDnNormalized() {
+        return issuerDnNormalized;
+    }
+
+    public void setIssuerDnNormalized(String issuerDnNormalized) {
+        this.issuerDnNormalized = issuerDnNormalized;
+    }
+
+    public String getSubjectDnNormalized() {
+        return subjectDnNormalized;
+    }
+
+    public void setSubjectDnNormalized(String subjectDnNormalized) {
+        this.subjectDnNormalized = subjectDnNormalized;
     }
 
     public CertificateContent getCertificateContent() {
--- a/src/main/java/com/czertainly/core/dao/entity/Certificate.java
+++ b/src/main/java/com/czertainly/core/dao/entity/Certificate.java
@@ -184,8 +184,7 @@ public class Certificate extends UniquelyIdentifiedAndAudited implements Seriali
     private String revokeAttributes;
 
     @Column(name = "trusted_ca")
-    @Nullable
-    private boolean trustedCa;
+    private Boolean trustedCa;
 
     @Override
     public CertificateDetailDto mapToDto() {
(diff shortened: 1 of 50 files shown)


src/main/java/com/czertainly/core/dao/entity/Certificate.java
```diff
@@ -210,6 +213,7 @@ public CertificateDetailDto mapToDto() {
         dto.setValidationStatus(validationStatus);
         dto.setCertificateType(certificateType);
         dto.setOwner(owner);
+        if (trustedCa != null) dto.setTrustedCa(trustedCa);
```
lubomirw, line 216: You don't need this IF when applied requested change in Interfaces

src/main/java/com/czertainly/core/dao/entity/Certificate.java
```diff
@@ -293,6 +297,7 @@ public CertificateDto mapToListDto() {
         dto.setValidationStatus(validationStatus);
         dto.setFingerprint(fingerprint);
         dto.setOwner(owner);
+        if (trustedCa != null) dto.setTrustedCa(trustedCa);
```
lubomirw, line 300: same thing here as above on line 216

src/main/java/com/czertainly/core/dao/entity/Certificate.java
```diff
@@ -744,4 +749,16 @@ private String getIssuerCommonNameToDto() {
...
+
+    public boolean isTrustedCa() {
+        return trustedCa;
+    }
+
+    public void setTrustedCa(boolean trustedCa) {
+        this.trustedCa = trustedCa;
+    }
+
+    public Boolean getTrustedCa() {
```
lubomirw, line 761: Why do we need 2 getters `isTrustedCa` and `getTrustedCa`?

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -316,6 +316,15 @@ public void updateCertificateObjects(SecuredUUID uuid, CertificateUpdateObjectsD
...
         }
+
+        if (request.getTrustedCa() != null) {
+            updateTrustedCaMark(uuid, request.getTrustedCa());
+        }
+    }
+
+    private void updateTrustedCaMark(SecuredUUID uuid, Boolean trustedCa) throws NotFoundException {
+        Certificate certificate = getCertificateEntity(uuid);
+        if (certificate.getTrustedCa() != null) certificate.setTrustedCa(trustedCa);
```
lubomirw, line 327: Throw here ValidationException when trying to set certificate as trusted and it is not CA certificate. It will need to add it to throws clause in Interfaces.

src/main/java/com/czertainly/core/validation/certificate/X509CertificateValidator.java
```diff
@@ -97,14 +105,18 @@ private Map<CertificateValidationCheck, CertificateValidationCheckDto> validateP
...
+    private CertificateValidationCheckDto checkCertificateChain(X509Certificate certificate, X509Certificate issuerCertificate, boolean trustedCa, CertificateValidationStatus issuerCertificateStatus, boolean isCompleteChain) {
         if (issuerCertificate == null) {
             // should be trust anchor (Root CA certificate)
-            if (isCompleteChain) {
+            if (isCompleteChain && trustedCa) {
                 return new CertificateValidationCheckDto(CertificateValidationCheck.CERTIFICATE_CHAIN, CertificateValidationStatus.VALID, "Certificate chain is complete. Certificate is Root CA certificate (trusted anchor).");
-            } else {
+            } else if (isCompleteChain) {
+                return new CertificateValidationCheckDto(CertificateValidationCheck.CERTIFICATE_CHAIN, CertificateValidationStatus.INVALID, "Certificate chain is complete, but the issuer certificate is not marked as trusted CA.");
+            }
```
lubomirw, line 115: missing else

src/main/java/com/czertainly/core/validation/certificate/X509CertificateValidator.java
```diff
@@ -97,14 +105,18 @@ private Map<CertificateValidationCheck, CertificateValidationCheckDto> validateP
         return validationOutput;
     }
 
-    private CertificateValidationCheckDto checkCertificateChain(X509Certificate certificate, X509Certificate issuerCertificate, CertificateValidationStatus issuerCertificateStatus, boolean isCompleteChain) {
+    private CertificateValidationCheckDto checkCertificateChain(X509Certificate certificate, X509Certificate issuerCertificate, boolean trustedCa, CertificateValidationStatus issuerCertificateStatus, boolean isCompleteChain) {
         if (issuerCertificate == null) {
             // should be trust anchor (Root CA certificate)
-            if (isCompleteChain) {
+            if (isCompleteChain && trustedCa) {
```
lubomirw, line 111: Rewrite if to be more readable, first should be if `isCompleteChain` and inside if `trustedCa` or not.

src/main/resources/db/migration/V202311131230__trusted_ca_mark.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE certificate ADD COLUMN trusted_ca BOOLEAN NULL;
+UPDATE certificate
+SET trusted_ca = CASE WHEN basic_constraints LIKE '%Subject Type=CA%' AND trusted_ca is NULL THEN false ELSE NULL END;
```
lubomirw, line 3: It is not necessary to use CASE WHEN construct. Also expression `trusted_ca is NULL` will be always true since it is new column.

You just need to set it to false and add `WHERE` condition with `LIKE` as you have it.

lubomirw (changes requested):

lubomirw (approved):

PR #609, opened 2023-11-23
Fix ACME finalize order to reflect async issuance process

lubomirw:
Based on [ACME RFC8555](https://datatracker.ietf.org/doc/html/rfc8555#section-7.1).

At the end of the referred section is table that illustrates a typical sequence of requests required to establish a new account with the server, prove control of an identifier, issue a certificate, and fetch an updated certificate some time after issuance.

### DIFF
--- a/src/main/java/com/czertainly/core/dao/entity/acme/AcmeOrder.java
+++ b/src/main/java/com/czertainly/core/dao/entity/acme/AcmeOrder.java
@@ -71,7 +71,7 @@ public class AcmeOrder extends UniquelyIdentifiedAndAudited implements Serializa
         order.setAuthorizations(authorizations.stream().map(AcmeAuthorization::getUrl).collect(Collectors.toList()));
         order.setFinalize(getFinalizeUrl());
         if(certificateId != null){
-            order.setCertificate(getCertificateUrl());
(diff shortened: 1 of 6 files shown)


src/main/java/com/czertainly/core/dao/entity/acme/AcmeOrder.java
```diff
@@ -71,7 +71,7 @@ public Order mapToDto(){
         order.setAuthorizations(authorizations.stream().map(AcmeAuthorization::getUrl).collect(Collectors.toList()));
         order.setFinalize(getFinalizeUrl());
         if(certificateId != null){
-            order.setCertificate(getCertificateUrl());
+            order.setCertificate(getCertificateUrl()); // should we include URL of certificate if order is still in PROCESSING state?
```
3keyroman, line 74: The certificate is optional according to the https://datatracker.ietf.org/doc/html/rfc8555#section-7.1.3 and should be included once the certificate has been issued. From my perspective, it does not make sense to have certificate URL in other state than VALID. Once VALID, it should have the URL where to download the certificate.

lubomirw: I agree, I will update PR accordingly

3keyroman (approved):

PR #697, opened 2024-04-17
Implement discovery rules and their matching on discovered certificates

### LINKED ISSUE #322: Implement discovery rules and their matching on discovered certificates
After certificates are found by discovery, rules assigned to discovery should be applied to each certificate. Rules consist of defined criteria on properties of certificate and action that needs to be performed when certificate matches criteria.

Criteria that specifies which certificates will be affected by rule can be represented by search fields DTOs used for filters.

Action should be of different types like
- assign certificate to RA profile, group and owner
- set custom attributes to discovered certificate
- skip and do not add it to the inventory

Example:
When certificate is of type X.509, assign it to RA profile 'X.509 RA Profile' and set its owner to 'Discovery'.

Add support for managing rules of discovery and their matching and execution when discovery is completed.

For this following needs to be implemented:
- add list of events for individual resource (mapping through enum `ResourceEvent`) as PlatformEnum, mapping to resource and produced resource (optional)
- add endpoint to `ResourceService` to list events by resource
- add `Ignore` to `RuleActionType` enum - does not perform any action and prevents any processing of resource object further (in case of discovery, do not add to inventory)
- add optional list of selected triggers to `DiscoveryDto` which then will be stored in M:N table with its order as provided in request (`trigger_order` column)
- split process of discovery completion to first create discovered certificates and raise event for discovery completed
- handle triggers for discovery completed in event listener and process creation of certificates based on triggers results
  - in case of multiple triggers, actions are performed by order of triggers.
  - if `Ignore` is action type of trigger, it should be the only action of trigger and has to be first one to execute
- make necessary changes to `EventMessage` to handle trigger events

klaraf755:
Closes #322

### DIFF
--- a/src/main/java/com/czertainly/core/service/impl/DiscoveryServiceImpl.java
+++ b/src/main/java/com/czertainly/core/service/impl/DiscoveryServiceImpl.java
@@ -22,7 +22,7 @@ import com.czertainly.api.model.core.certificate.CertificateEvent;
 import com.czertainly.api.model.core.certificate.CertificateEventStatus;
 import com.czertainly.api.model.core.connector.FunctionGroupCode;
 import com.czertainly.api.model.core.discovery.DiscoveryStatus;
-import com.czertainly.api.model.core.rules.RuleActionDto;
+import com.czertainly.api.model.core.rules.ResourceEvent;
 import com.czertainly.api.model.core.rules.RuleActionType;
 import com.czertainly.api.model.core.search.FilterFieldSource;
 import com.czertainly.api.model.core.search.SearchFieldDataByGroupDto;
@@ -57,7 +57,6 @@ import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
-import java.security.NoSuchAlgorithmException;
 import java.security.cert.X509Certificate;
 import java.util.*;
 import java.util.function.BiFunction;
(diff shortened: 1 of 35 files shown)


src/main/java/com/czertainly/core/service/impl/DiscoveryServiceImpl.java
```diff
@@ -456,4 +484,101 @@ public List<SearchFieldDataByGroupDto> getSearchableFieldInformationByGroup() {
...
     }
+
+    @Override
+    public void evaluateDiscoveryTriggers(UUID discoveryUuid) {
+
+        // Get discovered certificates
+        List<DiscoveryCertificate> discoveredCertificates = discoveryCertificateRepository.findByDiscoveryUuid(discoveryUuid);
+        DiscoveryHistory discoveryHistory = discoveryRepository.findByUuid(discoveryUuid).orElse(null);
+        // Get triggers for the discovery, separately for triggers with ignore action, the rest of triggers are in given order
+        List<RuleTrigger2Object> ruleTrigger2Objects = object2TriggerRepository.findAllByResourceAndObjectUuid(Resource.DISCOVERY, discoveryUuid);
```
lubomirw, line 495: select from DB could be already ordered so there would not be a need for sorting afterwards

src/main/java/com/czertainly/core/service/impl/DiscoveryServiceImpl.java
```diff
@@ -456,4 +484,101 @@ public List<SearchFieldDataByGroupDto> getSearchableFieldInformationByGroup() {
         logger.debug("Searchable Fields by Groups: {}", searchFieldDataByGroupDtos);
         return searchFieldDataByGroupDtos;
     }
+
+    @Override
+    public void evaluateDiscoveryTriggers(UUID discoveryUuid) {
+
+        // Get discovered certificates
+        List<DiscoveryCertificate> discoveredCertificates = discoveryCertificateRepository.findByDiscoveryUuid(discoveryUuid);
```
lubomirw, line 492: Why  not select only newlyDiscovered certificate directly?

src/main/java/com/czertainly/core/service/impl/DiscoveryServiceImpl.java
```diff
@@ -456,4 +484,101 @@ public List<SearchFieldDataByGroupDto> getSearchableFieldInformationByGroup() {
...
+                    continue;
+                }
+                Certificate entry = certificateService.createCertificateEntity(x509Cert);
+
+                // First, check the triggers that have action with action type set to ignore
+                boolean ignored = false;
+                for (RuleTrigger trigger : ignoreTriggers) {
+                    if (satisfiesRules(entry, trigger, discoveryCertificate.getDiscoveryUuid())) {
+                        ignored = true;
+                        break;
```
lubomirw, line 531: Here can be straight continue

lubomirw (changes requested):

lubomirw (approved):

PR #872, opened 2024-10-18
Implement authentication using OAuth2 Protocol

### LINKED ISSUE #873: Implement OAuth2 protocol with support to add custom OpenID providers to settings
Implement OAuth2 protocol and allow authorization of CZERTAINLY user with custom external OpenID providers (e.g. KeyCloak).
Order in security filter chain defining order of auth method should be following:
1. protocol access as anonymous user - `/protocols/{protocol_name}/**` paths
2. certificate in header
3. JWT token in header
4. OAuth2 login with redirection to selected custom OpenID provider

Settings should store all necessary information about provider necessary to utilize OAuth2 protocol.

klaraf755:
Closes #873


src/main/resources/application.properties
```diff
@@ -102,3 +102,12 @@ otel.exporter.otlp.traces.protocol=${OTEL_EXPORTER_OTLP_TRACES_PROTOCOL:grpc}
...
 cmp.verbose=false
 cmp.protocol.poll.feature.timeout=20
+
+# Spring Session
+spring.session.store-type=jdbc
+spring.session.jdbc.initialize-schema=never
+spring.session.jdbc.table-name=${DB_SCHEMA:core}.spring_session
+
+spring.session.jdbc.max-inactive-interval=1000
+server.servlet.session.timeout=30m
```
3keyroman, line 112: This should go to the settings and should not be in application.properties

PR #884, opened 2024-11-19
Implement authentication using OAuth2 Protocol

### LINKED ISSUE #873: Implement OAuth2 protocol with support to add custom OpenID providers to settings

klaraf755:
Closes #873


src/main/java/com/czertainly/core/auth/oauth2/CzertainlyOAuth2FailureHandler.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class CzertainlyOAuth2FailureHandler implements AuthenticationFailureHandler {
+
+    private static final Logger logger = LoggerFactory.getLogger(CzertainlyOAuth2FailureHandler.class);
+
+    @Override
+    public void onAuthenticationFailure(HttpServletRequest request, HttpServletResponse response, AuthenticationException exception) {
+        logger.debug("Error occurred when trying to authenticate using OAuth2 protocol: {}", exception.getMessage());
+        try {
+            response.sendRedirect("/api/login?error=" + URLEncoder.encode(exception.getMessage(), StandardCharsets.UTF_8));
```
3keyroman, line 22: Context should not be hardcoded, when changed, it will stop working.

src/main/java/com/czertainly/core/auth/oauth2/OAuth2LoginFilter.java
```diff
@@ -0,0 +1,167 @@
...
+
+import java.io.IOException;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.*;
+
+@Component
+public class OAuth2LoginFilter extends OncePerRequestFilter {
+
+    @Value("${auth.token.header-name}")
```
3keyroman, line 43: `X-USERINFO` should not resolve as authenticated user. Only valid tokens that are properly signed should be allowed.

src/main/java/com/czertainly/core/security/authn/client/CzertainlyAuthenticationClient.java
```diff
@@ -41,12 +46,11 @@ public class CzertainlyAuthenticationClient extends CzertainlyBaseAuthentication
     private String authTokenHeaderName;
```
3keyroman, line 46: We do not use unsigned `X-USERINFO`, the proper and valid JWT must be used.

src/main/java/com/czertainly/core/service/CertificateService.java
```diff
@@ -78,7 +78,7 @@ public interface CertificateService extends ResourceExtensionService  {
 
     CertificateContent checkAddCertificateContent(String fingerprint, String content);
 
-    CertificateDetailDto upload(UploadCertificateRequestDto request) throws AlreadyExistException, CertificateException, NoSuchAlgorithmException, NotFoundException, AttributeException;
+    CertificateDetailDto upload(UploadCertificateRequestDto request, boolean ignoreCustomAttributes) throws AlreadyExistException, CertificateException, NoSuchAlgorithmException, NotFoundException, AttributeException;
 
     // TODO AUTH - unable to check access based on certificate serial number. Make private? Special permission? Call opa in method?
```
3keyroman, line 83: Resolve TODO

src/main/java/com/czertainly/core/service/CertificateService.java
```diff
@@ -78,7 +78,7 @@ public interface CertificateService extends ResourceExtensionService  {
 
     CertificateContent checkAddCertificateContent(String fingerprint, String content);
 
-    CertificateDetailDto upload(UploadCertificateRequestDto request) throws AlreadyExistException, CertificateException, NoSuchAlgorithmException, NotFoundException, AttributeException;
+    CertificateDetailDto upload(UploadCertificateRequestDto request, boolean ignoreCustomAttributes) throws AlreadyExistException, CertificateException, NoSuchAlgorithmException, NotFoundException, AttributeException;
```
3keyroman, line 81: Why there is `boolean ignoreCustomAttributes` in this PR? Is it related to OAuth2 implementation?

src/main/java/com/czertainly/core/service/impl/LocalAdminServiceImpl.java
```diff
@@ -1,169 +1,49 @@
...
-            }
-        }
-        return certificate;
+        return userDetailDto;
     }
 
-    private String getSuperAdminRoleUuid(){
-        return roleManagementApiClient.getRoles().getData().stream().filter(e -> e.getSystemRole().equals(true) && e.getName().equals("superadmin")).collect(Collectors.toList()).get(0).getUuid();
+    private String getSuperadminRoleUuid() {
+        return roleManagementApiClient.getRoles().getData().stream().filter(e -> e.getSystemRole().equals(true) && e.getName().equals(AuthHelper.SUPERADMIN_USERNAME)).toList().getFirst().getUuid();
```
3keyroman, line 47: @lubomirw Why we are looking for `SUPERADMIN_USERNAME` that is not mandatory, when looking for the `superadmin` role?

src/main/java/com/czertainly/core/auth/oauth2/CzertainlyOAuth2FailureHandler.java
```diff
@@ -4,24 +4,31 @@
...
 
+    @Value("${server.servlet.context-path}")
+    private String contextPath;
+
     @Override
     public void onAuthenticationFailure(HttpServletRequest request, HttpServletResponse response, AuthenticationException exception) {
         logger.debug("Error occurred when trying to authenticate using OAuth2 protocol: {}", exception.getMessage());
         try {
-            response.sendRedirect("/api/login?error=" + URLEncoder.encode(exception.getMessage(), StandardCharsets.UTF_8));
+            response.sendRedirect(contextPath + "/login?error=" + URLEncoder.encode(exception.getMessage(), StandardCharsets.UTF_8));
```
3keyroman, line 29: Is this going to work when the application is running behind proxy or gateway? What about something like `ServletUriComponentsBuilder.fromCurrentContextPath().build().toUriString()`?

src/main/java/com/czertainly/core/service/CertificateService.java
lubomirw: It is part of localhost user authentication implementation which was part of this PR due to changes in authentication flow.

Ignoring custom attributes when uploading certificate is necessary when registering new local admin with new uploaded certificate. In case users are locked out of system (lost access or other reasons) don't know which custom attributes are required and that would prevent uploading new certificate to system by localhost user.

src/main/java/com/czertainly/core/service/impl/LocalAdminServiceImpl.java
lubomirw: System roles/users share the same name. Name of constant is misleading but name refers to role name and also to user username. Constants have _USERNAME suffix since they are almost always used to refer to and retrieve users. Should it be renamed to avoid confusion?

lubomirw (approved):

PR #1035, opened 2025-05-19
Add implementation of hybrid certificates

klaraf755:


src/main/java/com/czertainly/core/attribute/engine/AttributeOperation.java
```diff
@@ -5,6 +5,10 @@ public class AttributeOperation {
     public static final String CERTIFICATE_ISSUE = "issue";
     public static final String CERTIFICATE_REVOKE = "revoke";
     public static final String CERTIFICATE_REQUEST_SIGN = "sign";
+    public static final String CERTIFICATE_REQUEST_ALT_SIGN = "altSign";
```
3keyroman, line 8: Why we need to create here `altSign` attribute operation?

src/main/java/com/czertainly/core/config/TokenContentSigner.java
```diff
@@ -80,7 +83,7 @@ public OutputStream getOutputStream() {
 
     @Override
     public byte[] getSignature() {
-        byte[] dataToSign = outputStream.toByteArray();
+        byte[] dataToSign = (signPublicKey && outputStream.size() == 0) ? publicKey.getBytes() : outputStream.toByteArray();
```
3keyroman, line 86: What is intended logic of this?

src/main/java/com/czertainly/core/model/request/Pkcs10CertificateRequest.java
```diff
@@ -53,6 +60,21 @@ public PublicKey getPublicKey() throws NoSuchAlgorithmException, CertificateRequ
         }
     }
 
+    @Override
+    public PublicKey getAltPublicKey() throws NoSuchAlgorithmException, CertificateRequestException {
+        Extensions requestedExtensions = jcaObject.getRequestedExtensions();
```
3keyroman, line 65: Seems to me like duplicated code of the CrmfCertificateRequest.

src/main/java/com/czertainly/core/model/request/Pkcs10CertificateRequest.java
```diff
@@ -63,4 +85,12 @@ public AlgorithmIdentifier getSignatureAlgorithm() {
         return jcaObject.getSignatureAlgorithm();
     }
 
+    @Override
+    public AlgorithmIdentifier getAltSignatureAlgorithm() {
+        Extensions requestedExtensions = jcaObject.getRequestedExtensions();
```
3keyroman, line 90: Seems to me like duplicated code of the CrmfCertificateRequest.

src/main/java/com/czertainly/core/service/impl/CryptographicOperationServiceImpl.java
```diff
@@ -409,6 +411,25 @@ public String generateCsr(UUID keyUuid, UUID tokenProfileUuid, X500Principal pri
             );
         }
 
+        Map<KeyType, CryptographicKeyItem> defaultKeyPair = getPublicAndPrivateKey(tokenProfileUuid, keyUuid);
+        Map<KeyType, CryptographicKeyItem> altKeyPair = new EnumMap<>(KeyType.class);
+        if (altKeyUUid != null && altTokenProfileUuid != null) altKeyPair = getPublicAndPrivateKey(altTokenProfileUuid, altKeyUUid);
+
+        // Generate the CSR
+        return generateCsr(
```
3keyroman, line 419: What if the key and alternative key are the same?

src/main/java/com/czertainly/core/validation/certificate/X509CertificateValidator.java
```diff
@@ -382,6 +408,37 @@ private boolean verifySignature(X509Certificate subjectCertificate, X509Certific
         }
     }
 
+    private static byte[] getDERWithoutSignatureAndAltSignature(X509Certificate subjectCertificate) throws IOException, CertificateEncodingException {
+        X509CertificateHolder holder = new JcaX509CertificateHolder(subjectCertificate);
+        TBSCertificate tbsCert = holder.toASN1Structure().getTBSCertificate();
+
+        Extensions extensions = tbsCert.getExtensions();
```
3keyroman, line 415: Do we need to generate extension? Isn't possible to just remove what is not needed?

src/main/java/com/czertainly/core/config/TokenContentSigner.java
3keyroman: It is a different use case which should not be mixed.

src/main/java/com/czertainly/core/validation/certificate/X509CertificateValidator.java
klaraf755: The `Extensions` class is not modifiable, so we need to create a new instance

lubomirw (approved):

PR #1106, opened 2025-07-31
Implement support for archived certificates and prevent using them in any process or action

### LINKED ISSUE #1098: Implement support for archived certificates and prevent using them in any process or action
- implement archive/unarchive endpoints for certificate
- add boolean `archived` property to certificate entity and table representing that certificate is archived
- use new search DTO with `includeArchived` property. If is set to true, select also archived certificates, otherwise automatically add predicate to where conditions `archived` equals `False`
- add boolean `includeArchived` query parameter for retrieving statistics and use it add predicate as in point above
- do not allow any operations on archived certificates and add check for certificate state where it is necessary
  - client operations - renew, rekey, revoke
  - no validation or compliance check
  - cannot select them as SCEP CA certificate or in other protocols
  - cannot use it as user certificate
  - cannot update certificate
  - cannot push certificate to location or renew in location 
- the only operations allowed for archived certificates are get, download and delete
- add record to certificate history with `Archive` / `Unarchive` action when archiving/unarchiving certificate
- add `Archive` action in `ExternalAuthorization` to authorize for archive/unarchive operations

klaraf755:
Closes #1098


src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -271,14 +272,17 @@ public void setApplicationEventPublisher(ApplicationEventPublisher applicationEv
...
     @ExternalAuthorization(resource = Resource.CERTIFICATE, action = ResourceAction.LIST, parentResource = Resource.RA_PROFILE, parentAction = ResourceAction.MEMBERS)
-    public CertificateResponseDto listCertificates(SecurityFilter filter, SearchRequestDto request) {
+    public CertificateResponseDto listCertificates(SecurityFilter filter, CertificateSearchRequestDto request) {
         setupSecurityFilter(filter);
         RequestValidatorHelper.revalidateSearchRequestDto(request);
         final Pageable p = PageRequest.of(request.getPageNumber() - 1, request.getItemsPerPage());
-
-        final TriFunction<Root<Certificate>, CriteriaBuilder, CriteriaQuery, Predicate> additionalWhereClause = (root, cb, cr) -> FilterPredicatesBuilder.getFiltersPredicate(cb, cr, root, request.getFilters());
+        List<SearchFilterRequestDto> filters = new ArrayList<>(request.getFilters());
+        if (!request.isIncludeArchived()) {
```
lubomirw, line 280: This should be handled by adding predicate for that field to additionalWhereClause through `TriFunction`.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -271,14 +272,17 @@ public void setApplicationEventPublisher(ApplicationEventPublisher applicationEv
...
+            filters.add(new SearchFilterRequestDto(FilterFieldSource.PROPERTY, FilterField.ARCHIVED.name(), FilterConditionOperator.EQUALS, false));
+        }
+        final TriFunction<Root<Certificate>, CriteriaBuilder, CriteriaQuery, Predicate> additionalWhereClause = (root, cb, cr) -> FilterPredicatesBuilder.getFiltersPredicate(cb, cr, root, filters);
         final List<CertificateDto> listedKeyDTOs = certificateRepository.findUsingSecurityFilter(filter, List.of(), additionalWhereClause, p, (root, cb) -> cb.desc(root.get("created"))).stream().map(Certificate::mapToListDto).toList();
-        final Long maxItems = certificateRepository.countUsingSecurityFilter(filter, additionalWhereClause);
+        final Long maxItems = certificateRepository.countUsingSecurityFilter(filter, additionalWhereClause, request.isIncludeArchived());
```
lubomirw, line 285: This should be handled by adding predicate for that field to additionalWhereClause.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -800,6 +808,7 @@ public void validate(Certificate certificate) {
     @ExternalAuthorization(resource = Resource.CERTIFICATE, action = ResourceAction.DETAIL)
     public CertificateValidationResultDto getCertificateValidationResult(SecuredUUID uuid) throws NotFoundException {
         Certificate certificate = getCertificateEntityWithAssociations(uuid);
+        if (certificate.isArchived()) throw new ValidationException("Cannot validate archived certificate with UUID %s.".formatted(uuid));
```
lubomirw, line 811: Exception should not be thrown, it should behave like with validation disabled below. For example, user navigating to detail page of archived certificate would get exception when refreshing validation status on `Validation` tab.

It already handled by method `CertificateUtil.isValidationEnabled`.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1175,6 +1185,10 @@ public int updateCertificatesStatusScheduled() {
     // Internal Use Only
     public void updateCertificateUser(UUID certificateUuid, String userUuid) throws NotFoundException {
         Certificate certificate = certificateRepository.findByUuid(certificateUuid).orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
+        if (certificate.isArchived()) {
+            logger.info("Certificate with UUID {} is archived and the user cannot be set.", certificateUuid);
```
lubomirw, line 1189: Here is more suitable to throw Exception since it is user triggered operation that is not allowed. It should say it is not allowed to set archived certificate for user as auth method.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1197,49 +1211,50 @@ public void removeCertificateUser(UUID userUuid) {
 
     @Override
     @ExternalAuthorization(resource = Resource.CERTIFICATE, action = ResourceAction.LIST, parentResource = Resource.RA_PROFILE, parentAction = ResourceAction.LIST)
-    public Long statisticsCertificateCount(SecurityFilter filter) {
+    public Long statisticsCertificateCount(SecurityFilter filter, boolean includeArchived) {
         setupSecurityFilter(filter);
-        return certificateRepository.countUsingSecurityFilter(filter);
+        return certificateRepository.countUsingSecurityFilter(filter, includeArchived);
```
lubomirw, line 1216: This should be handled by adding predicate for that field to additionalWhereClause.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1197,49 +1211,50 @@ public void removeCertificateUser(UUID userUuid) {
...
-    public StatisticsDto addCertificateStatistics(SecurityFilter filter, StatisticsDto dto) {
+    public StatisticsDto addCertificateStatistics(SecurityFilter filter, StatisticsDto dto, boolean includeArchived) {
         setupSecurityFilter(filter);
 
         long start = System.nanoTime();
         try (ExecutorService executor = Executors.newVirtualThreadPerTaskExecutor()) {
             executor.invokeAll(List.of(
                     () -> {
-                        dto.setCertificateStatByKeySize(certificateRepository.countGroupedUsingSecurityFilter(filter, null, Certificate_.keySize, null));
+                        dto.setCertificateStatByKeySize(certificateRepository.countGroupedUsingSecurityFilter(filter, null, Certificate_.keySize, null, includeArchived));
```
lubomirw, line 1228: This should be handled by adding predicate for that field to additionalWhereClause. In all calls of `countGroupedUsingSecurityFilter` in this method.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1325,8 +1340,10 @@ public void evaluatePermissionChain(SecuredUUID uuid) throws NotFoundException {
     @Override
     public void updateCertificateKeys(UUID keyUuid, String publicKeyFingerprint) {
         for (Certificate certificate : certificateRepository.findByPublicKeyFingerprint(publicKeyFingerprint)) {
-            certificate.setKeyUuid(keyUuid);
-            certificateRepository.save(certificate);
+            if (!certificate.isArchived()) {
```
lubomirw, line 1343: Relations to archived certificates can be maintained to keep inventory items relations actual. Certificate could be also unarchived and then missing correct existing relation to key already in inventory.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1745,6 +1798,10 @@ private String downloadChain(String chainUrl) {
 
     public void switchRaProfile(SecuredUUID uuid, SecuredUUID raProfileUuid) throws NotFoundException, CertificateOperationException, AttributeException {
         Certificate certificate = getCertificateEntity(uuid);
+        if (certificate.isArchived()) {
+            logger.info("Certificate with UUID {} is archived and its RA Profile will not be updated.", uuid);
```
lubomirw, line 1802: Here is more suitable to throw Exception since it is user triggered operation that is not allowed. It should say it is not allowed to change RA profile for archived certificate.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1803,6 +1860,11 @@ public void switchRaProfile(SecuredUUID uuid, SecuredUUID raProfileUuid) throws
     public void updateCertificateGroups(SecuredUUID uuid, Set<UUID> groupUuids) throws NotFoundException {
         Certificate certificate = getCertificateEntityWithAssociations(uuid);
 
+        if (certificate.isArchived()) {
+            logger.info("Certificate with UUID {} is archived and its groups will not be updated.", uuid);
```
lubomirw, line 1864: Here is more suitable to throw Exception since it is user triggered operation that is not allowed. It should say it is not allowed to change groups for archived certificate.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1824,6 +1886,11 @@ public void updateCertificateGroups(SecuredUUID uuid, Set<UUID> groupUuids) thro
     public void updateOwner(SecuredUUID uuid, String ownerUuid) throws NotFoundException {
         Certificate certificate = getCertificateEntityWithAssociations(uuid);
 
+        if (certificate.isArchived()) {
+            logger.info("Certificate with UUID {} is archived and owner will not be updated.", uuid);
```
lubomirw, line 1890: Here is more suitable to throw Exception since it is user triggered operation that is not allowed. It should say it is not allowed to change owner for archived certificate.

lubomirw (changes requested):

PR #1133, opened 2025-09-05
Invalidate existing session when user is deleted or disabled

### LINKED ISSUE #1132: Invalidate existing session when user is deleted or disabled
When user is logged in with OAuth2 provider and in the meantime user is deleted or disabled while session is active, user can continue to use application.

When user is deleted, its session should be invalidated.

klaraf755:
Closes #1132


3keyroman:
@copilot Implement test for OAuth2Util

PR #1212, opened 2026-01-02
Change login to provide list of providers in JSON

3keyroman:
Provide list of providers instead of `html` template. The reason for this change is to own the login page by the frontend application.

This also removed dependency on `thymeleaf`.


3keyroman:
@copilot update `LoginControllerTest.java` to pass all the tests.

lubomirw (approved):

PR #1222, opened 2026-01-11
Performance improvement: RA profile names

ivosh:
With several hundreds of RA profiles, `/api/v1/certificates/search` API endpoint takes easily 10 seconds. With the proposed optimization, we are down to 200 ms total, with `RaProfileReposity.findAllNames()` taking just 10 ms.
The culprit is a costly SQL SELECT on the `core.ra_profile` table.


ivosh:
Hmm, I forgot to push my test updates.

Anyway, it is not as simple as that. This whole API endpoint is full of security vulnerabilities because it gives out all RA profile names (along with other information) without any authorization. I'll file a dedicated bug for that.

PR #1281, opened 2026-02-13
Fix owner and group association clearing for transient entities

ivosh:
Test case CryptographicKeyServiceTest#testDeleteKey() was consistently failing with the following message:
```
org.hibernate.TransientObjectException: persistent instance references an unsaved transient instance of 'com.czertainly.core.dao.entity.OwnerAssociation' (save the transient instance before flushing)
```

Explicitly clearing the references to owner association and groups, Hibernate does not see any transient references when the transaction is committed and the session flushed.

Snímek obrazovky 2026-02-13 v 23 00 22


ivosh:
@lubomirw Please could you have a look?

lubomirw (approved):

PR #1283, opened 2026-02-16
Improve performance of certificate listing

ivosh:


ivosh:
@lubomirw Please could you have a look.

PR #1289, opened 2026-02-19
feat: SaaS migration - proxy support, JMS messaging, AMQP support

pajikos:
Squash-merge of main-saas-rebase branch containing:
- AMQP 1.0 / Azure Service Bus + RabbitMQ dual support via JMS
- Proxy CRUD and proxy-based connector communication
- ConnectorApiFactory replacing direct API clients
- Trusted certificate management and provisioning API
- Oauth2 authentication for messaging
- Redis/Valkey pub/sub for proxy response distribution
- Username claim support for OAuth2


src/main/java/com/czertainly/core/provisioning/ProvisioningApiConfig.java
```diff
@@ -0,0 +1,83 @@
...
+            .build();
+        return httpServiceProxyFactory.createClient(ProxyProvisioningApiClient.class);
+    }
+
+    @Bean
+    public TrustedCertificateProvisioningApiClient trustedCertificateProvisioningApiClient(RestClient provisioningRestClient) {
+        HttpServiceProxyFactory httpServiceProxyFactory = HttpServiceProxyFactory.builderFor(RestClientAdapter.create(provisioningRestClient))
+            .build();
+        return httpServiceProxyFactory.createClient(TrustedCertificateProvisioningApiClient.class);
+    }
```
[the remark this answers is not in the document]
pajikos: TESTED!

False Positive: @Retryable on HttpServiceProxyFactory clients

  All three concerns are unfounded after code investigation.

  1. HttpServiceProxyFactory duplication — Cosmetic, not a bug

  Both @Bean methods (ProvisioningApiConfig.java:71-74, 79-81) create a separate HttpServiceProxyFactory instance, but HttpServiceProxyFactory is a
  lightweight stateless builder. Creating two instances sharing the same provisioningRestClient has zero functional or performance impact.

  2. @Retryable "silently ignored" — Incorrect

  The claim that @Retryable annotations are silently ignored is wrong. The reasoning "HttpServiceProxyFactory creates its own proxy → not a Spring AOP
  proxy → @Retryable ignored" conflates "not created by Spring AOP" with "not processable by Spring AOP."

  Here's why @Retryable works:

  1. @EnableRetry on RetryConfig.java:17 registers RetryAnnotationBeanPostProcessor — a BeanPostProcessor that applies to all beans in the context.
  2. Beans returned from @Bean methods go through the full BeanPostProcessor pipeline, regardless of how they were constructed internally.
  3. RetryAnnotationBeanPostProcessor.postProcessAfterInitialization() calls AopUtils.canApply() on the bean. For JDK proxies (which
  HttpServiceProxyFactory.createClient() produces), this checks all implemented interfaces — it finds
  ProxyProvisioningApiClient/TrustedCertificateProvisioningApiClient with @Retryable on their methods.
  4. The post-processor wraps the original HttpServiceProxyFactory proxy in a new Spring AOP proxy with retry advice.

  The resulting call chain is: Caller → [Spring Retry AOP Proxy] → [HttpExchange JDK Proxy] → HTTP call.

  3. @EnableRetry "not active" in context — Incorrect

  @EnableRetry on RetryConfig.java:17 is application-wide. It registers RetryAnnotationBeanPostProcessor globally, not scoped to the JMS package. The
  provisioning API client beans are in the same Spring application context and are processed identically to any other bean.

  4. Retry properties — Properly configured

  The retry properties referenced by @Retryable SpEL expressions are defined in application.yml:235-238 with sensible defaults:
  provisioning.api.retry:
    max-attempts: ${PROVISIONING_RETRY_MAX_ATTEMPTS:3}
    delay: ${PROVISIONING_RETRY_DELAY:1000}
    max-delay: ${PROVISIONING_RETRY_MAX_DELAY:5000}
    multiplier: ${PROVISIONING_RETRY_MULTIPLIER:2}

src/main/java/com/czertainly/core/api/web/TrustedCertificateControllerImpl.java
```diff
@@ -0,0 +1,67 @@
...
+        UuidDto dto = new UuidDto();
+        dto.setUuid(trustedCertificateDto.getUuid());
+        return ResponseEntity.created(location).body(dto);
+    }
+
+    @Override
+    @AuditLogged(module = Module.CORE, resource = Resource.TRUSTED_CERTIFICATE, operation = Operation.DELETE)
+    public void deleteTrustedCertificate(@LogResource(uuid = true) @PathVariable String uuid) throws NotFoundException {
+        trustedCertificateService.deleteTrustedCertificate(SecuredUUID.fromString(uuid));
+    }
```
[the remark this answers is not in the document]
pajikos: this is standard codebase pattern, false positive

src/main/java/com/czertainly/core/messaging/jms/configuration/JmsConfig.java
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    @Bean
+    public MessageConverter messageConverter(ObjectMapper jacksonObjectMapper) {
+        MappingJackson2MessageConverter converter = new MappingJackson2MessageConverter();
+        converter.setObjectMapper(jacksonObjectMapper);
+        converter.setTargetType(MessageType.TEXT);
+
+        // Configure ObjectMapper with Java 8 date/time support
+        ObjectMapper objectMapper = new ObjectMapper();
```
lubomirw: The injected `jacksonObjectMapper` bean is set on the converter (line 171), but then immediately overwritten by a brand-new `ObjectMapper` (lines 175-178). This discards all application-wide Jackson configuration (custom serializers, modules, feature flags) and replaces it with a bare mapper that only has `JavaTimeModule`.

Either register the module on the injected mapper:
```java
jacksonObjectMapper.registerModule(new JavaTimeModule());
jacksonObjectMapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
converter.setObjectMapper(jacksonObjectMapper);
```
Or, if you don't want to mutate the shared bean, use `.copy()`:
```java
ObjectMapper objectMapper = jacksonObjectMapper.copy();
objectMapper.registerModule(new JavaTimeModule());
objectMapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
converter.setObjectMapper(objectMapper);
```

Which ObjectMapper is intended here — the application-wide one or the bare one? Could the loss of custom serializers/modules from the application ObjectMapper cause serialization issues for JMS messages (e.g. attribute types, enums, or custom model classes that rely on app-level Jackson config)?

lubomirw (commented):
Code review: 1 issue found.

src/main/java/com/czertainly/core/messaging/jms/configuration/JmsConfig.java
davidsvamberk: You're right that the original code was broken — the injected mapper was
  discarded immediately. However, I'd like to propose a different fix than the `.copy()` approach.

  **Proposed fix: use `Jackson2ObjectMapperBuilder`**

  ```java
  @Bean
  public MessageConverter messageConverter(Jackson2ObjectMapperBuilder objectMapperBuilder) {
      ObjectMapper objectMapper = objectMapperBuilder.createXmlMapper(false)
              .featuresToDisable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
              .build()
              .findAndRegisterModules();

      MappingJackson2MessageConverter converter = new MappingJackson2MessageConverter();
      converter.setObjectMapper(objectMapper);
      converter.setTargetType(MessageType.TEXT);

      return converter;
  }
```
  
Why this is better than `.copy()`

Spring Boot itself creates the primary ObjectMapper bean by using Jackson2ObjectMapperBuilder
(via JacksonAutoConfiguration). Injecting the builder and calling .build() goes through the
same configuration pipeline — all Jackson2ObjectMapperBuilderCustomizer beans are applied
automatically, so application-wide Jackson configuration is preserved without touching the shared
bean.

`.copy()` works, but it copies the mapper's state at a point in time and the semantics can be
surprising (e.g. mixins, registered modules). Using the builder is more explicit and idiomatic.

Why WRITE_DATES_AS_TIMESTAMPS is disabled explicitly

I checked — there is no global spring.jackson.serialization.write-dates-as-timestamps=false in
application.yml, and WebAppConfig sets this feature locally on a static
Jackson2ObjectMapperBuilder.json() instance rather than via a shared customizer bean. The
explicit .featuresToDisable(...) in JmsConfig is therefore intentional and necessary.

Do you like this approach?

lubomirw: Yes, that seems as good approach.

davidsvamberk: Fixed by [218fb37](https://github.com/OmniTrustILM/CZERTAINLY-Core/pull/1289/commits/218fb3795406ae8c66fb218579d732b11bdc6d75)

PR #1401, opened 2026-04-09
Ensure updating secrets in sync vault profiles is always performed per sync profile

### LINKED ISSUE #1398: Ensure updating secrets in sync vault profiles is always performed per sync profile
Currently, updating, adding, or deleting a secret in a sync vault profile makes only one call to the associated vault instance—even if there are multiple sync profiles associated with that vault instance. 
Delete secret and remove sync vault profile should be modified to accept new optional query parameter for deleting secrets in vault on provider side.

Modify the behaviour so that the update/add/delete operation on a vault instance is performed for each sync vault profile, regardless of the vault instance association. Each sync vault profile update should trigger its own operation, ensuring all relevant profiles are consistently managed.

klaraf755:
Closes #1398


src/main/java/com/czertainly/core/service/impl/SecretServiceImpl.java
```diff
@@ -553,13 +542,9 @@ public void removeVaultProfileFromSecret(UUID uuid, UUID vaultProfileUuid) throw
                 () -> new NotFoundException(VaultProfile.class, vaultProfileUuid)
         ).getVaultProfile();
         Secret2SyncVaultProfile secret2SyncVaultProfile = secret2SyncVaultProfileRepository.getReferenceById(new Secret2SyncVaultProfileId(secret.getUuid(), removedVaultProfile.getUuid()));
-        // Check if there are any vault profiles related to the secret with the same vault instance as the one being removed
-        // If there are no such profiles, remove the secret from the vault
-        if (!Objects.equals(removedVaultProfile.getVaultInstanceUuid(), secret.getSourceVaultProfile().getVaultInstanceUuid()) && secret.getSyncVaultProfiles()
-                .stream()
-                .noneMatch(profile -> profile.getVaultProfile().getVaultInstanceUuid().equals(removedVaultProfile.getVaultInstanceUuid()))) {
+        // Remove the secret from the vault
```
lubomirw, line 554: `removeVaultProfileFromSecret` now always deletes the physical secret from the vault, which is a silent behavior change with data-loss potential.

The previous guard ensured deletion only happened when no other profile (source or sync) shared the same vault instance, preventing removal of a secret that was still in use by another profile. The new code removes that guard and unconditionally deletes. A caller who simply wants to detach the sync profile association from Core — without touching the vault — has no way to do that now.

Consider introducing a `deleteFromVault` boolean parameter (defaulting to `false`) on this method and its corresponding API endpoint. This would make the behavior explicit: by default, removing a sync profile only removes the Core association; callers who want to also clean up the vault opt in with `deleteFromVault=true`. This design eliminates the need for the old vault-instance guard entirely and gives the caller control over whether vault state is mutated.

Same approach with query param can be applied for delete secret and remove sync profile from secret.

src/main/java/com/czertainly/core/service/impl/SecretServiceImpl.java
```diff
@@ -385,19 +385,14 @@ private void updateSecretAction(UUID secretUuid, SecretUpdateRequestDto secretRe
...
             for (Secret2SyncVaultProfile profile : secret.getSyncVaultProfiles()) {
                 VaultProfile syncVaultProfile = profile.getVaultProfile();
-                if (!processedVaultInstanceUuids.contains(syncVaultProfile.getVaultInstance().getUuid())) {
-                    SecretResponseDto syncResponse = updateSecretInVault(secret, syncVaultProfile, secretRequest, profile.getSecretAttributes());
-                    attributeEngine.updateMetadataAttributes(syncResponse.getMetadata(), new ObjectAttributeContentInfo(syncVaultProfile.getVaultInstance().getConnectorUuid(), Resource.SECRET, secret.getUuid(), Resource.VAULT_PROFILE, syncVaultProfile.getUuid(), syncVaultProfile.getName()));
-                    processedVaultInstanceUuids.add(syncVaultProfile.getVaultInstance().getUuid());
-                }
+                SecretResponseDto syncResponse = updateSecretInVault(secret, syncVaultProfile, secretRequest, profile.getSecretAttributes());
```
lubomirw: Partial failure mid-loop leaves source and already-processed sync profiles updated while remaining sync profiles are not — no rollback.

Before this PR, deduplication reduced the number of vault calls per loop iteration, so the failure window was smaller. With each sync profile now making its own vault call, a failure on profile N leaves profiles 0..N-1 fully updated and profiles N+1..end untouched. The exception message `"Failed to update secret"` gives no indication of how far the update progressed. This is pre-existing behaviour, but the expanded call surface makes it more likely to manifest in practice. Consider either logging which profile failed (to aid diagnostics) or tracking partial progress so a retry can resume from where it left off.

src/main/java/com/czertainly/core/service/impl/SecretServiceImpl.java
lubomirw, line 554: Leftover extra indentation on `deleteSecretFromVault` call.

After removing the wrapping `if` block, the `deleteSecretFromVault` call retains one extra level of indentation compared to the surrounding statements. Should be aligned with the rest of the method body.

lubomirw (changes requested):

src/main/java/com/czertainly/core/service/impl/SecretServiceImpl.java
klaraf755: Done

klaraf755: Done

src/main/java/com/czertainly/core/service/impl/SecretServiceImpl.java
klaraf755: Added try-catch for each individual call

src/main/java/com/czertainly/core/service/impl/SecretServiceImpl.java
```diff
@@ -553,13 +542,9 @@ public void removeVaultProfileFromSecret(UUID uuid, UUID vaultProfileUuid) throw
                 () -> new NotFoundException(VaultProfile.class, vaultProfileUuid)
         ).getVaultProfile();
         Secret2SyncVaultProfile secret2SyncVaultProfile = secret2SyncVaultProfileRepository.getReferenceById(new Secret2SyncVaultProfileId(secret.getUuid(), removedVaultProfile.getUuid()));
-        // Check if there are any vault profiles related to the secret with the same vault instance as the one being removed
-        // If there are no such profiles, remove the secret from the vault
-        if (!Objects.equals(removedVaultProfile.getVaultInstanceUuid(), secret.getSourceVaultProfile().getVaultInstanceUuid()) && secret.getSyncVaultProfiles()
-                .stream()
-                .noneMatch(profile -> profile.getVaultProfile().getVaultInstanceUuid().equals(removedVaultProfile.getVaultInstanceUuid()))) {
+        // Remove the secret from the vault
             deleteSecretFromVault(removedVaultProfile, secret, secret2SyncVaultProfile.getSecretAttributes());
```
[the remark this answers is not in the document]
klaraf755: done

src/main/java/com/czertainly/core/service/impl/SecretServiceImpl.java
```diff
@@ -384,24 +384,24 @@ private void updateSecretAction(UUID secretUuid, SecretUpdateRequestDto secretRe
...
-            attributeEngine.updateMetadataAttributes(sourceVaultProfileResponse.getMetadata(), new ObjectAttributeContentInfo(currentSourceVaultProfile.getVaultInstance().getConnectorUuid(), Resource.SECRET, secret.getUuid(), Resource.VAULT_PROFILE, currentSourceVaultProfile.getUuid(), currentSourceVaultProfile.getName()));
+        } catch (Exception e) {
+            throw new SecretOperationException(String.format("Failed to update secret %s for vault profile %s: %s", secret.getName(), secret.getSourceVaultProfile().getName(), e.getMessage()));
+        }
+        attributeEngine.updateMetadataAttributes(sourceVaultProfileResponse.getMetadata(), new ObjectAttributeContentInfo(currentSourceVaultProfile.getVaultInstance().getConnectorUuid(), Resource.SECRET, secret.getUuid(), Resource.VAULT_PROFILE, currentSourceVaultProfile.getUuid(), currentSourceVaultProfile.getName()));
             for (Secret2SyncVaultProfile profile : secret.getSyncVaultProfiles()) {
                 VaultProfile syncVaultProfile = profile.getVaultProfile();
-                if (!processedVaultInstanceUuids.contains(syncVaultProfile.getVaultInstance().getUuid())) {
```
lubomirw, line 396: **! Major** | Correctness

Wrong vault profile name in sync profile error message

The error message uses `secret.getSourceVaultProfile().getName()` but this code is inside the loop iterating over **sync** vault profiles. When a sync vault profile update fails, the error will misleadingly report the source vault profile's name instead of the actual failing sync vault profile. Should be `syncVaultProfile.getName()` and the message should read `"Failed to update secret %s for sync vault profile %s: %s"` to distinguish it from the source vault profile error on line 391.

src/main/java/com/czertainly/core/service/impl/SecretServiceImpl.java
```diff
@@ -384,24 +384,24 @@ private void updateSecretAction(UUID secretUuid, SecretUpdateRequestDto secretRe
...
         newVersion.setFingerprint(newFingerprint);
         newVersion.setVaultInstance(currentSourceVaultProfile.getVaultInstance());
-        secretVersionRepository.save(newVersion);
-        Set<UUID> processedVaultInstanceUuids = new HashSet<>();
-        processedVaultInstanceUuids.add(currentSourceVaultProfile.getVaultInstance().getUuid());
-        SecretResponseDto sourceVaultProfileResponse = null;
+        SecretResponseDto sourceVaultProfileResponse;
         try {
             sourceVaultProfileResponse = updateSecretInVault(secret, secret.getSourceVaultProfile(), secretRequest, secretRequest.getAttributes());
-            attributeEngine.updateMetadataAttributes(sourceVaultProfileResponse.getMetadata(), new ObjectAttributeContentInfo(currentSourceVaultProfile.getVaultInstance().getConnectorUuid(), Resource.SECRET, secret.getUuid(), Resource.VAULT_PROFILE, currentSourceVaultProfile.getUuid(), currentSourceVaultProfile.getName()));
```
lubomirw, line 393: **~ Minor** | Maintainability

Stale indentation on for loop

The `for` loop and its body have an extra level of indentation left over from when this code was inside the previous `try` block. The loop is now at the same nesting level as the `attributeEngine.updateMetadataAttributes` call above it (line 393) but is indented one level deeper. Should be dedented to align with the surrounding code.

src/main/java/com/czertainly/core/service/impl/SecretServiceImpl.java
```diff
@@ -410,11 +410,14 @@ private void updateSecretAction(UUID secretUuid, SecretUpdateRequestDto secretRe
```
lubomirw, line 410: **~ Minor** | Maintainability

Extra indentation on for loop in deleteSecretAction

The `for` loop and its body are indented one level too deep relative to the enclosing `if` block. The `vaultProfiles.add(...)` line above is at the correct level, but the `for` loop is shifted right by one extra level.

lubomirw (changes requested):

src/main/java/com/czertainly/core/service/impl/SecretServiceImpl.java
klaraf755: Fixed

src/main/java/com/czertainly/core/service/impl/SecretServiceImpl.java
klaraf755: fixed

src/main/java/com/czertainly/core/service/impl/SecretServiceImpl.java
klaraf755: fixed

src/main/java/com/czertainly/core/service/impl/SecretServiceImpl.java
```diff
@@ -385,19 +385,14 @@ private void updateSecretAction(UUID secretUuid, SecretUpdateRequestDto secretRe
...
                 VaultProfile syncVaultProfile = profile.getVaultProfile();
-                if (!processedVaultInstanceUuids.contains(syncVaultProfile.getVaultInstance().getUuid())) {
-                    SecretResponseDto syncResponse = updateSecretInVault(secret, syncVaultProfile, secretRequest, profile.getSecretAttributes());
-                    attributeEngine.updateMetadataAttributes(syncResponse.getMetadata(), new ObjectAttributeContentInfo(syncVaultProfile.getVaultInstance().getConnectorUuid(), Resource.SECRET, secret.getUuid(), Resource.VAULT_PROFILE, syncVaultProfile.getUuid(), syncVaultProfile.getName()));
-                    processedVaultInstanceUuids.add(syncVaultProfile.getVaultInstance().getUuid());
-                }
+                SecretResponseDto syncResponse = updateSecretInVault(secret, syncVaultProfile, secretRequest, profile.getSecretAttributes());
+                attributeEngine.updateMetadataAttributes(syncResponse.getMetadata(), new ObjectAttributeContentInfo(syncVaultProfile.getVaultInstance().getConnectorUuid(), Resource.SECRET, secret.getUuid(), Resource.VAULT_PROFILE, syncVaultProfile.getUuid(), syncVaultProfile.getName()));
             }
```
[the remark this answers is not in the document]
klaraf755: Added check for number of requests

lubomirw (approved):

PR #1422, opened 2026-04-23
Fix wrong token instance UUID sent to connector

### LINKED ISSUE #167: Cryptographic Operations: Wrong token instance UUID sent to connector
## Description

In `CryptographicOperationServiceImpl`, the methods `encryptData`, `decryptData`, `signData`, `verifyData`, and `generateCsr` all passed the wrong identifier to the external cryptographic connector.

The code used `tokenInstanceReferenceUuid` — the database primary key of the internal `TokenInstanceReference` entity — as the token instance identifier in connector API calls. The connector expects `tokenInstanceUuid`, a separate string field on the entity that represents the token instance identifier as known to the external provider. These two values are different in practice, causing all cryptographic operations routed through the connector to fail with a 404 or similar error.

Affected connector call pattern:
```
POST /v1/cryptographyProvider/tokens/{tokenInstanceUuid}/keys/{keyReferenceUuid}/encrypt|decrypt|sign|verify
```

The same issue affected `generateCsr` when constructing `TokenContentSigner` for both the primary and alternative (dual-key) signing paths.

## Steps to Reproduce

1. Configure a cryptography token instance with an external software cryptography provider connector
2. Create a cryptographic key associated with the token instance and a token profile
3. Attempt any cryptographic operation: encrypt data, decrypt data, sign data, verify signature, or generate a CSR
4. Observe that the connector call fails — the wrong token instance UUID is sent in the URL path

## Expected Result

- Connector receives the correct `tokenInstanceUuid` (the provider-side identifier) in the URL path for all cryptographic operations
- All operations — encrypt, decrypt, sign, verify, generateCsr — complete successfully

## Actual Result

- Connector receives `tokenInstanceReferenceUuid` (the internal DB primary key) instead of `tokenInstanceUuid`
- The connector cannot locate the token instance, causing all cryptographic operations to fail

## Environment

ILM Core:
Commit:
Branch:
Build time:

Client:
- OS:
- Browser:

ivosh:


src/test/java/com/czertainly/core/service/CryptographicOperationServiceTest.java
```diff
@@ -110,7 +110,7 @@ public void setUp() {
         tokenInstanceReference.setConnector(connector);
         tokenInstanceReference.setConnectorUuid(connector.getUuid());
         tokenInstanceReference.setKind("sample");
-        tokenInstanceReference.setTokenInstanceUuid("1l");
+        tokenInstanceReference.setTokenInstanceUuid("11111111-1111-1111-1111-111111111111");
         tokenInstanceReferenceRepository.save(tokenInstanceReference);
```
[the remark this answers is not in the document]
ivosh: This is test setup code. The value is intentionally deterministic and human-readable — it makes test failures easy to diagnose. UUID.randomUUID().toString() would make tests non-deterministic. A named constant would add noise with no benefit since it's only used once. The fix from "1l" to a valid UUID is the real improvement here; style nit is not worth the change.

src/main/java/com/czertainly/core/service/impl/CryptographicOperationServiceImpl.java
```diff
@@ -537,7 +537,7 @@ private String generateCsr(X500Principal principal, String key, CryptographicKey
         ContentSigner signer = new TokenContentSigner(
                 cryptographicOperationsApiClient,
                 privateKeyItem.getKey().getTokenInstanceReference().getConnector().mapToApiClientDtoV1(),
-                privateKeyItem.getKey().getTokenInstanceReferenceUuid(),
+                UUID.fromString(privateKeyItem.getKey().getTokenInstanceReference().getTokenInstanceUuid()),
                 privateKeyItem.getKeyReferenceUuid(),
                 publicKeyItem.getKeyReferenceUuid(),
```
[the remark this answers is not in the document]
ivosh: A check has been added to TokenInstanceService.createTokenInstance.

src/main/java/com/czertainly/core/service/impl/TokenInstanceServiceImpl.java
```diff
@@ -182,6 +182,13 @@ public TokenInstanceDetailDto createTokenInstance(TokenInstanceRequestDto reques
                 connectorApiFactory.getTokenInstanceApiClient(connector).createTokenInstance(connector, tokenInstanceRequestDto);
         TokenInstanceStatusDto status = connectorApiFactory.getTokenInstanceApiClient(connector).getTokenInstanceStatus(connector, response.getUuid());
         logger.debug("Token Instance Response from the connector: {}", response);
+
+        try {
+            UUID.fromString(response.getUuid());
```
lubomirw, line 184: **! Major** | Correctness

Validation runs too late and does not cover null UUIDs.

`response.getUuid()` is already passed to `getTokenInstanceStatus(...)` at line 183 before any validation, so a malformed UUID produces a confusing connector error (likely a 404) rather than the intended `ValidationException`. Additionally, `UUID.fromString(null)` throws `NullPointerException`, which is not caught here — a connector returning `{ "uuid": null }` would surface a raw NPE instead of the controlled `ValidationException`. Move the validation block above the `getTokenInstanceStatus` call and either null-check `response.getUuid()` first or broaden the catch to include `NullPointerException`.

src/main/java/com/czertainly/core/service/impl/TokenInstanceServiceImpl.java
```diff
@@ -182,6 +182,13 @@ public TokenInstanceDetailDto createTokenInstance(TokenInstanceRequestDto reques
...
+        } catch (IllegalArgumentException e) {
+            throw new ValidationException(ValidationError.create(
+                    "Connector returned invalid token instance UUID: " + response.getUuid()));
```
lubomirw: **~ Minor** | Maintainability

Error message names neither the connector nor the requested token instance, so a triager seeing this in logs has to cross-reference other context to find which connector misbehaved. Include `connector.getName()` (or its UUID) and the requested token instance name from `request.getName()`. Example: `"Connector '%s' returned invalid token instance UUID '%s' for token instance '%s'".formatted(connector.getName(), response.getUuid(), request.getName())`.

lubomirw (changes requested):

src/main/java/com/czertainly/core/service/impl/TokenInstanceServiceImpl.java
ivosh: Fixed.

src/main/java/com/czertainly/core/service/impl/TokenInstanceServiceImpl.java
ivosh: Fixed.

lubomirw (approved):

PR #1437, opened 2026-05-04
Implement event history and per-object event history endpoints

### LINKED ISSUE #1423: Implement retrieving events associated with an object
### Description

Implement endpoint defined in https://github.com/OmniTrustILM/interfaces/issues/654 in new `EventHistoryService`. 

Data for the response can already be obtained from already existing tables for `TriggerAssociation`, `TriggerHistory` and `TriggerHistoryRecord`, so no additional migration is needed in this case.

### User Story

_No response_

### Use Case

Use Case 2: Event firing history log for specific object

### Acceptance Criteria

- Event history for object detail page
- Each record should contain information about trigger and its evaluation details (which rule and action was checked and performed)

### LINKED ISSUE #1425: Implement retrieving history of an event
### Description

Implement the endpoint defined in https://github.com/OmniTrustILM/interfaces/issues/656.

To properly track individual firing of an event, a new table `EventHistory` needs to be created, with columns:
- `uuid`
- `EventResource`
- `eventResourceUuid`
- `started_at`
- `finished_at`
- `status`

Existing `TriggerHistory` table should then be extended with column `event_history_uuid` referencing an instance of an event that fired that trigger.

Migration of existing `TriggerHistory`: 
For existing trigger histories, an event history is assigned as follows:
1. retrieve all trigger history for an event (identified by event type and resource firing the event), ordered by time of execution
2. when there are histories for an object "close by", but with different trigger associations, assign them to the same event history, with time of start equal to triggered at of the first trigger

Cascade behaviour of removing histories associated with trigger should also be changed from removing histories when trigger is removed to keeping the histories available.

### User Story

_No response_

### Use Case

Use Case 1: Event firing history log
Use case 3: Provide information if notification were sent

### Acceptance Criteria

- Existing trigger history is preserved and migrated
- Event history under Events in Settings displays correct log of individual firing of events
- Each event history record should contain information about trigger and its evaluation details (which rule and action was checked and performed) similar to current trigger history for discovery

klaraf755:
Closes #1423
Closes #1425


src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
```diff
@@ -0,0 +1,169 @@
...
+    @Override
+    public PaginationResponseDto<ObjectEventHistoryDto> getEventHistory(Resource resource, UUID uuid, PaginationRequestDto pagination) throws NotFoundException {
+        resourceService.evaluateDetailsPermission(resource, uuid);
+        Page<TriggerHistory> triggerHistoryPage = triggerHistoryRepository.findByObjectUuidAndTriggerResourceOrderByTriggeredAtDesc(uuid, resource, Pageable.ofSize(pagination.getItemsPerPage()).withPage(pagination.getPageNumber()));
+        PaginationResponseDto<ObjectEventHistoryDto> response = new PaginationResponseDto<>();
+        response.setItemsPerPage(pagination.getItemsPerPage());
+        response.setPageNumber(pagination.getPageNumber());
+        response.setTotalItems(triggerHistoryPage.getTotalElements());
+        response.setTotalPages((int) Math.ceil((double) triggerHistoryPage.getTotalElements() / pagination.getItemsPerPage()));
+        response.setItems(
```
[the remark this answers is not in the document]
klaraf755: Changed to PageRequest

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
```diff
@@ -0,0 +1,169 @@
...
+                        triggerHistory -> {
+                            ObjectEventHistoryDto dto = new ObjectEventHistoryDto();
+                            dto.setEvent(triggerHistory.getEvent());
+                            if (triggerHistory.getTriggerAssociation().getResource() != null) {
+                                try {
+                                    dto.setOrigin(resourceService.getResourceObject(triggerHistory.getTriggerAssociation().getResource(), triggerHistory.getTriggerAssociation().getUuid()));
+                                } catch (NotFoundException e) {
+                                    dto.setOrigin(new ResourceObjectDto(triggerHistory.getTriggerAssociation().getResource(), null, null));
+                                }
+                            }
```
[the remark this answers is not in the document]
klaraf755: Fixed

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
```diff
@@ -0,0 +1,169 @@
...
+                }).toList());
+                triggerHistoriesInEvent.add(triggerHistoryObjectSummaryDto);
+            }
+
+            PaginationResponseDto<TriggerHistoryObjectSummaryDto> triggerHistoriesPaginated = new PaginationResponseDto<>();
+            triggerHistoriesPaginated.setItemsPerPage(request.getObjectsPagination().getItemsPerPage());
+            triggerHistoriesPaginated.setPageNumber(request.getObjectsPagination().getPageNumber());
+            triggerHistoriesPaginated.setItems(triggerHistoriesInEvent);
+            triggerHistoriesPaginated.setTotalItems(dto.getObjectsEvaluated());
+            dto.setObjectHistories(triggerHistoriesPaginated);
```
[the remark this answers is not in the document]
klaraf755: Fixed

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
```diff
@@ -0,0 +1,169 @@
...
+    private static Boolean notificationsSent(TriggerHistory triggerHistory) {
+        // If there was any action sending notifications and conditions were met and there is no trigger history record for failed execution with send notification type
+        boolean notificationsInTrigger = triggerHistory.getTrigger().getActions().stream()
+                .anyMatch(action -> action.getExecutions().stream().anyMatch(e -> e.getType() == ExecutionType.SEND_NOTIFICATION));
+        if (notificationsInTrigger) {
+            return
+                    triggerHistory.isConditionsMatched() &&
+                            triggerHistory.getRecords().stream().noneMatch(r -> r.getExecution() != null && r.getExecution().getType() == ExecutionType.SEND_NOTIFICATION);
+        }
+        return null;
```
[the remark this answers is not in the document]
klaraf755: Added check for null

src/main/java/com/czertainly/core/events/EventHandler.java
```diff
@@ -153,39 +173,50 @@ protected void processAllTriggers(EventContext<T> context) throws EventException
...
-        }
 
-        // If some trigger ignored this object, processing is stopped
-        if (isIgnored) {
-            return;
-        }
+            // If some trigger ignored this object, processing is stopped
+            if (isIgnored) {
+                return;
+            }
```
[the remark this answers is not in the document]
klaraf755: Added finished status

src/main/resources/db/migration/V202604281200__event_history.sql
```diff
@@ -0,0 +1,82 @@
...
+    SELECT
+        uuid,
+        event,
+        object_uuid,
+        trigger_association_uuid,
+        triggered_at,
+        CASE
+            WHEN LAG(triggered_at) OVER (PARTITION BY event, object_uuid ORDER BY triggered_at) IS NULL
+                OR triggered_at - LAG(triggered_at) OVER (PARTITION BY event, object_uuid ORDER BY triggered_at) > INTERVAL '2 seconds'
+            THEN 1
```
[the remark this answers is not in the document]
klaraf755: Aligned

src/main/resources/db/migration/V202604281200__event_history.sql
```diff
@@ -0,0 +1,82 @@
...
+-- Add event_history_uuid to trigger_history
+ALTER TABLE trigger_history ADD COLUMN event_history_uuid UUID;
+ALTER TABLE trigger_history ADD CONSTRAINT trigger_history_to_event_history_fk
+    FOREIGN KEY (event_history_uuid) REFERENCES event_history(uuid) ON UPDATE CASCADE ON DELETE SET NULL;
+
+-- Change trigger FK: keep trigger_history when trigger is deleted (SET NULL instead of CASCADE)
+ALTER TABLE trigger_history DROP CONSTRAINT trigger_history_trigger_uuid_fkey;
+ALTER TABLE trigger_history ALTER COLUMN trigger_uuid DROP NOT NULL;
+ALTER TABLE trigger_history ADD CONSTRAINT trigger_history_trigger_uuid_fkey
+    FOREIGN KEY (trigger_uuid) REFERENCES trigger(uuid) ON UPDATE CASCADE ON DELETE SET NULL;
```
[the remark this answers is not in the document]
klaraf755: Fixed

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
```diff
@@ -0,0 +1,169 @@
...
+            dto.setObjectsMatched(triggerHistoryRepository.countDistinctObjectUuidByEventHistoryUuidAndConditionsMatchedTrue(eventHistory.getUuid()));
+            dto.setObjectsIgnored(triggerHistoryRepository.countDistinctObjectUuidByEventHistoryUuidAndConditionsMatchedTrueAndTriggerIgnoreTriggerTrue(eventHistory.getUuid()));
+            Page<UUID> objectUuids = triggerHistoryRepository.findDistinctObjectUuidsByEventHistoryUuid(eventHistory.getUuid(), Pageable.ofSize(request.getObjectsPagination().getItemsPerPage()).withPage(request.getObjectsPagination().getPageNumber() - 1));
+            List<TriggerHistoryObjectSummaryDto> triggerHistoriesInEvent = new ArrayList<>();
+            for (UUID objectUuid : objectUuids) {
+                List<TriggerHistory> triggerHistories = triggerHistoryRepository.findByEventHistoryUuidAndObjectUuidOrderByObjectUuidAscTriggeredAtDesc(eventHistory.getUuid(), objectUuid);
+                TriggerHistoryObjectSummaryDto triggerHistoryObjectSummaryDto = new TriggerHistoryObjectSummaryDto();
+                triggerHistoryObjectSummaryDto.setObjectUuid(objectUuid);
```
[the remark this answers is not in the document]
klaraf755: Changed to fetching trigger histories in one query

src/main/java/com/czertainly/core/dao/repository/workflows/TriggerHistoryRepository.java
```diff
@@ -19,4 +25,19 @@ public interface TriggerHistoryRepository extends SecurityFilterRepository<Trigg
     List<TriggerHistory> findByTriggerAssociationObjectUuidOrderByTriggerUuidAscTriggeredAtAsc(UUID triggerAssociationObjectUuid);
 
     Long deleteByTriggerAssociationObjectUuid(UUID triggerAssociationObjectUuid);
+
+    @EntityGraph(attributePaths = {"records", "triggerAssociation", "records.execution", "records.execution.items", "records.execution.items.notificationProfile"})
+    Page<TriggerHistory> findByObjectUuidAndTriggerResourceOrderByTriggeredAtDesc(UUID objectUuid, Resource resource, Pageable pageable);
```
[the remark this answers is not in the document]
klaraf755: I think the best solution for this is to add a resource column to TriggerHistory, since both event and trigger are nullable

src/main/java/com/czertainly/core/dao/repository/workflows/TriggerHistoryRepository.java
```diff
@@ -19,4 +25,23 @@ public interface TriggerHistoryRepository extends SecurityFilterRepository<Trigg
...
```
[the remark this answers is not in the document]
klaraf755: Changed to object resource

src/main/java/com/czertainly/core/dao/repository/workflows/TriggerHistoryRepository.java
```diff
@@ -19,4 +25,23 @@ public interface TriggerHistoryRepository extends SecurityFilterRepository<Trigg
...
+
+    @Query(value = "SELECT DISTINCT t.objectUuid FROM TriggerHistory t WHERE t.eventHistoryUuid = :uuid",
+           countQuery = "SELECT COUNT(DISTINCT t.objectUuid) FROM TriggerHistory t WHERE t.eventHistoryUuid = :uuid")
+    Page<UUID> findDistinctObjectUuidsByEventHistoryUuid(@Param("uuid") UUID uuid, Pageable pageable);
+
```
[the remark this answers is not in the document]
klaraf755: Added

src/main/java/com/czertainly/core/service/impl/DiscoveryServiceImpl.java
```diff
@@ -728,6 +728,12 @@ public void evaluatePermissionChain(SecuredUUID uuid) throws NotFoundException {
         getDiscoveryEntity(uuid);
     }
 
+    @Override
+    @ExternalAuthorization(resource = Resource.DISCOVERY, action = ResourceAction.DELETE)
```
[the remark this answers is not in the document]
klaraf755: Fixed

src/main/java/com/czertainly/core/events/EventHandler.java
```diff
@@ -153,39 +173,53 @@ protected void processAllTriggers(EventContext<T> context) throws EventException
...
+                try {
+                    context.getTriggerEvaluator().evaluateTrigger(trigger, triggerAssociation, resourceObject, null, eventData, eventHistory);
+                    logger.debug("Trigger '{}' on {} object {} processed successfully", trigger.getName(), context.getResource().getLabel(), resourceObject.getUuid());
+                } catch (Exception e) {
+                    logger.error("Unable to process trigger '{}' on {} object {}. Message: {}", trigger.getName(), context.getResource().getLabel(), resourceObject.getUuid(), e.getMessage());
+                }
             }
+        } catch (Exception e) {
+            logger.error("Unable to process triggers for {} object {}. Message: {}", context.getResource().getLabel(), resourceObject.getUuid(), e.getMessage());
+            eventHistory.setStatus(EventStatus.FAILED);
```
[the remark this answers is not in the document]
klaraf755: Added

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+                triggerHistoryPage.get().map(
+                        triggerHistory -> {
+                            ObjectEventHistoryDto dto = new ObjectEventHistoryDto();
+                            dto.setEvent(triggerHistory.getEvent());
+                            if (triggerHistory.getTriggerAssociation() != null && triggerHistory.getTriggerAssociation().getResource() != null) {
+                                try {
+                                    dto.setOrigin(resourceService.getResourceObject(triggerHistory.getTriggerAssociation().getResource(), triggerHistory.getTriggerAssociation().getObjectUuid()));
+                                } catch (NotFoundException e) {
+                                    dto.setOrigin(new ResourceObjectDto(triggerHistory.getTriggerAssociation().getResource(), null, null));
+                                }
```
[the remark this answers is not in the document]
klaraf755: I think it can stay as null, since the not found exception represents the fact that the object does not exist, so null UUID is correct

src/main/resources/db/migration/V202604281200__event_history.sql
```diff
@@ -0,0 +1,91 @@
...
+        END AS new_session
+    FROM trigger_history
+    WHERE event IS NOT NULL AND object_uuid IS NOT NULL
+) t;
+
+-- One event_history row per session; derive resource from trigger_association or trigger
+CREATE TEMP TABLE temp_event_history AS
+SELECT DISTINCT ON (event, object_uuid, session_num)
+    gen_random_uuid() AS uuid,
+    tsm.event,
```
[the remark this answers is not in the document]
klaraf755: used hash function

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    @Autowired
+    public void setEventHistoryRepository(EventHistoryRepository eventHistoryRepository) {
+        this.eventHistoryRepository = eventHistoryRepository;
+    }
+
+
+    @Override
+    public PaginationResponseDto<ObjectEventHistoryDto> getEventHistory(Resource resource, UUID uuid, PaginationRequestDto pagination) throws NotFoundException {
```
ivosh, line 56: There is no `@ExternalAuthorization`. Why?

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.PageRequest;
+import org.springframework.stereotype.Service;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+@Service
+@Transactional
```
ivosh: Use method-level `@Transactional` annotation so the intent is clear. Class-level annotations are a relict from history and we should not repeat the same mistake over and over.

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+                                try {
+                                    dto.setOrigin(resourceService.getResourceObject(triggerHistory.getTriggerAssociation().getResource(), triggerHistory.getTriggerAssociation().getObjectUuid()));
+                                } catch (NotFoundException e) {
+                                    dto.setOrigin(new ResourceObjectDto(triggerHistory.getTriggerAssociation().getResource(), null, null));
+                                }
+                            }
+                            if (triggerHistory.getTrigger() != null) {
+                                dto.setTrigger(new NameAndUuidDto(triggerHistory.getTriggerUuid(), triggerHistory.getTrigger().getName()));
+                            }
+                            dto.setConditionsMatched(triggerHistory.isConditionsMatched());
```
ivosh: Use a dedicated mapper or at least a helper function, to reduce the cognitivity level of this method.

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+    public PaginationResponseDto<EventHistoryDto> getEventHistory(ResourceEvent event, Resource resource, UUID uuid, EventHistoryRequestDto request) throws NotFoundException {
+        if (uuid == null && resource != null || uuid != null && resource == null) {
+            throw new ValidationException("Missing UUID or Resource");
+        }
+        if (uuid != null) {
+            resourceService.evaluateDetailsPermission(resource, uuid);
+        }
+
+        Page<EventHistory> eventHistories = eventHistoryRepository.findByEventAndResourceAndResourceUuidOrderByStartedAtDesc(event, resource, uuid, PageRequest.of(request.getPagination().getPageNumber() - 1, request.getPagination().getItemsPerPage()));
+        List<EventHistoryDto> eventHistoriesResponse = new ArrayList<>();
```
ivosh: Use a dedicated mapper or at least a helper function, to reduce the cognitive level of this method.

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+            dto.setObjectsEvaluated(triggerHistoryRepository.countDistinctObjectUuidByEventHistoryUuid(eventHistory.getUuid()));
+            dto.setObjectsMatched(triggerHistoryRepository.countDistinctObjectUuidByEventHistoryUuidAndConditionsMatchedTrue(eventHistory.getUuid()));
+            dto.setObjectsIgnored(triggerHistoryRepository.countDistinctObjectUuidByEventHistoryUuidAndConditionsMatchedTrueAndTriggerIgnoreTriggerTrue(eventHistory.getUuid()));
+            Page<UUID> objectUuids = triggerHistoryRepository.findDistinctObjectUuidsByEventHistoryUuid(eventHistory.getUuid(), PageRequest.of(request.getObjectsPagination().getPageNumber() - 1, request.getObjectsPagination().getItemsPerPage()));
+            Map<UUID, List<TriggerHistory>> allTriggerHistories =
+                    triggerHistoryRepository.findByEventHistoryUuidAndObjectUuidInOrderByObjectUuidAscTriggeredAtDesc(eventHistory.getUuid(), objectUuids.toList())
+                            .stream().collect(Collectors.groupingBy(TriggerHistory::getObjectUuid, LinkedHashMap::new, Collectors.toList()));
+            List<TriggerHistoryObjectSummaryDto> triggerHistoriesInEvent = new ArrayList<>();
```
ivosh: Use a dedicated mapper or at least a helper function, to reduce the cognitive level of this method.

src/main/resources/db/migration/V202604281200__event_history.sql
```diff
@@ -0,0 +1,95 @@
...
+            THEN 1
+            ELSE 0
+        END AS new_session
+    FROM trigger_history
+    WHERE event IS NOT NULL AND object_uuid IS NOT NULL
+) t;
+
+-- One event_history row per session; derive resource from trigger_association or trigger
+CREATE TEMP TABLE temp_event_history AS
+SELECT DISTINCT ON (event, object_uuid, session_num)
```
ivosh, line 63: There is a subtle problem deriving a UUID this way - the UUID may not be valid: bits 6 (version nibble) and 8 (variant) carry random MD5 entropy rather than the required 0100 / 10xx patterns. Stricter validators may choke on these UUIDs.

Why not just use something like `gen_random_uuid`?

src/main/resources/db/migration/V202604281200__event_history.sql
```diff
@@ -0,0 +1,95 @@
...
+WHERE th.trigger_uuid = t.uuid;
+
+-- Change trigger FK: keep trigger_history when trigger is deleted (SET NULL instead of CASCADE)
+ALTER TABLE trigger_history DROP CONSTRAINT trigger_history_trigger_uuid_fkey;
+ALTER TABLE trigger_history ALTER COLUMN trigger_uuid DROP NOT NULL;
+ALTER TABLE trigger_history ADD CONSTRAINT trigger_history_trigger_uuid_fkey
+    FOREIGN KEY (trigger_uuid) REFERENCES trigger(uuid) ON UPDATE CASCADE ON DELETE SET NULL;
+
+-- Migrate existing data: group trigger_history records into event_history sessions.
+-- Records sharing the same (event, object_uuid) and triggered within 2 seconds of each other
```
ivosh, line 33: Why 2 seconds was chosen as the "right" number here? Perhaps you conducted a small research with data collected from various customers and decided that `2` is the right trade-off in this heuristic? Could it happen that some existing customers would need a different value for the migration?

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+        if (notificationsInTrigger) {
+            return
+                    triggerHistory.isConditionsMatched() &&
+                            triggerHistory.getRecords().stream().noneMatch(r -> r.getExecution() != null && r.getExecution().getType() == ExecutionType.SEND_NOTIFICATION);
+        }
+        return null;
+    }
+
+    @Override
+    @ExternalAuthorization(resource = Resource.RESOURCE_EVENT, action = ResourceAction.DETAIL)
```
ivosh, line 88: Check how many SQL queries this method actually uses and if you can combine them.

src/main/java/com/czertainly/core/dao/entity/workflows/TriggerHistory.java
ivosh: Check if `CascadeType.ALL` still makes sense here after the changes introduced in this PR on the TriggerHistoryRecord side, now that `createTriggerHistoryRecord` sets only uuid and never the reference.

src/main/java/com/czertainly/core/service/impl/TriggerServiceImpl.java
```diff
@@ -361,6 +361,11 @@ public TriggerHistory createTriggerHistory(UUID triggerUuid, TriggerAssociation
         triggerHistory.setObjectUuid(objectUuid);
         triggerHistory.setReferenceObjectUuid(referenceObjectUuid);
         triggerHistory.setTriggeredAt(OffsetDateTime.now());
+        if (eventHistory != null) {
+            triggerHistory.setEventHistoryUuid(eventHistory.getUuid());
+            triggerHistory.setEventHistory(eventHistory);
```
ivosh: TriggerHistory changed the semantics - only UUID matters to be set, not the reference (updatable=false). If you insist on doing it, consider creating a setter on TriggerHistory: setEventHistory(EventHistory) and fill both fields there.

ivosh (changes requested):
Thank you for the PR.

In addition to comments posted in-line, I have the following general ones:
* Please, pretty please, re-consider the current approach with `evaluatePermissionChainOnDetails`. I hope you agree with me it creates a ripple effect and it just does not look right (design smell). Consider what would happen if in future another action context is needed later? A single `evaluatePermissionChain(SecuredUUID, ResourceAction)` with the annotation moved to the caller (or a strategy lookup) would scale better. 
* There is a pre-existing problem with the transaction boundary in EventHandler but this PR changes the expectations and amplifies the risk, so it needs to be addressed. I recommend doing it in a follow-up PR, but it needs to be done so the user story is actually completed. See below.
* I'd like you to add a new test which would cover the following case (not present atm). I think there is a problem with how triggers are processed and EventHistory created. The intent of the PR is (I hope I am correct here): "one EventHistory per handleEvent call". However a single handleEvent call where `getResourceObjects().size() == N` and `getOverridingTriggers()` returns M groups, produces N × (M+1) EventHistory rows. Refer below.

## TriggerHistory transaction boundary problem
TriggerEvaluator is `@Transactional` at the class level. Combined with EventHandler also being `@Transactional` (EventHandler.java:32) and the default `Propagation.REQUIRED`, the entire `handleEvent → processAllTriggers → processTriggers → evaluateTrigger → createTriggerHistory(save)` chain runs in one transaction, started at the public handleEvent method.

So pre-PR: any uncaught exception escaping handleEvent already rolled back every TriggerHistory and TriggerHistoryRecord row. That's a pre-existing observability gap.

What this PR does:

1. Adds EventHistory rows on top of the same transaction — same rollback fate.
2. Materially increases the cost: now there are 1 INSERT + 1 UPDATE per processTriggers call queued, all of which get reverted on rollback.
3. Changes user expectations. EventHistory.status includes FAILED ([EventStatus enum]) — the API contract advertises that failures will be visible. Pre-PR, no one expected to see failure rows because none existed; post-PR, users will reasonably expect to see them, and the transaction behavior contradicts that expectation.

So: pre-existing for TriggerHistory, new and more visible for EventHistory.

Concrete flow that triggers the rollback

Using DiscoveryFinishedEventHandler.sendFollowUpEventsNotifications, an existing override (DiscoveryFinishedEventHandler.java:66-77):
```
66    @Override
67    protected void sendFollowUpEventsNotifications(EventContext<DiscoveryHistory> eventContext) {
68        DiscoveryHistory discovery = eventContext.getResourceObjects().getFirst();
69        Object eventData = eventContext.getResourceObjectsEventData().getFirst();
70        NotificationMessage notificationMessage = new NotificationMessage(...);
71        notificationProducer.produceMessage(notificationMessage);   // ← can throw JmsException
72
73        if (eventContext.getScheduledJobInfo() != null) {
74            ScheduledTaskResult scheduledTaskResult = new ScheduledTaskResult(...);
75            applicationEventPublisher.publishEvent(new ScheduledJobFinishedEvent(...));  // ← can throw if a sync listener fails
76        }
77    }
```

Sequence on the unhappy path:

1. JMS broker has a transient hiccup (or the message fails JSON serialization,
   or a synchronous Spring listener registered for ScheduledJobFinishedEvent throws).
2. EventHandler.handleEvent enters         [TRANSACTION T1 begins]
3. processAllTriggers runs, calls processTriggers once
4.   createEventHistory(...)  →  INSERT EventHistory(IN_PROGRESS)         [queued in T1]
5.   evaluateTrigger(...)     →  INSERT TriggerHistory(...)               [queued in T1]
6.                                INSERT TriggerHistoryRecord(...)        [queued in T1]
7.   eventHistoryRepository.save(...) with FINISHED                       [queued in T1]
8. processAllTriggers returns successfully
9. sendFollowUpEventsNotifications(eventContext) is called
10.   notificationProducer.produceMessage(...)  →  throws JmsException    [unchecked]
11. Exception propagates out of handleEvent
12. Spring's transactional aspect sees an uncaught RuntimeException        [TRANSACTION T1 rolls back]
13. All queued writes from steps 4-7 are reverted.
User opens Event History page in Settings → empty. No IN_PROGRESS row. No FAILED row.
No TriggerHistory row. No TriggerHistoryRecord row. The event is invisible.

## EventHistory multiplier test case
Adapt as needed:
```
@Test
void testEventHistoryNotInflatedByOverridingScopes()
        throws EventException, NotFoundException, AlreadyExistException {

    // 1. Two groups + one RA profile — these force CertificateEventsHandler.getOverridingTriggers()
    //    to return three EventContextTriggers entries, on top of the platform group.
    Group g1 = new Group(); g1.setName("g1"); groupRepository.save(g1);
    Group g2 = new Group(); g2.setName("g2"); groupRepository.save(g2);

    RaProfile raProfile = new RaProfile();
    raProfile.setName("ra-profile-test");
    // …minimal RA profile setup; mirror what testCertificateStatusChangedAndApprovalEvents does
    raProfileRepository.save(raProfile);

    // 2. Build a certificate associated with both groups AND the RA profile.
    Certificate certificate = new Certificate();
    // …minimal cert + content setup
    certificate.setRaProfile(raProfile);
    certificate.setGroups(Set.of(g1, g2));
    certificateRepository.save(certificate);

    // 3. Stand up the auth WireMock the existing tests rely on (copy from line 306-343).
    mockServer = new WireMockServer(10001);
    mockServer.start();
    // …(existing /auth/users + /auth stubs)

    // 4. NOTE: we deliberately do NOT register any trigger associations.
    //    The bug fires from the existence of the overriding scopes alone — fetchEventTriggers
    //    returns a (possibly-empty) EventContextTriggers per group / RA-profile, and
    //    processTriggers is called for each of them.

    // 5. One handleEvent call.
    certificateStatusChangedEventHandler.handleEvent(
        CertificateStatusChangedEventHandler.constructEventMessage(
            certificate.getUuid(),
            CertificateValidationStatus.INACTIVE,
            CertificateValidationStatus.VALID));

    // 6. The contract under test.
    List<EventHistory> rows = eventHistoryRepository.findAll();
    Assertions.assertEquals(
        1, rows.size(),
        "Exactly one EventHistory record per handleEvent call, regardless of overriding scopes");

    mockServer.stop();
}
```

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
klaraf755: I did not add this annotation here, because I permissions for detail (which I think are the permissions you need for the history of object) are evaluated individually based on the resource object by calling the resource service.

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
klaraf755: Moved annotation to methods.

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
klaraf755: I created a mapper class for these DTOs, and also a mapper for `PaginationResponseDto` which can be also reused by other classes that return paginated responses.

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
klaraf755: I created a mapper class for these DTOs, and also a mapper for `PaginationResponseDto` which can be also reused by other classes that return paginated responses.

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
klaraf755: I created a mapper class for these DTOs, and also a mapper for `PaginationResponseDto` which can be also reused by other classes that return paginated responses.

src/main/resources/db/migration/V202604281200__event_history.sql
klaraf755: I originally did use `gen_random_uuid`, but then I received this comment from Copilot: https://github.com/OmniTrustILM/core/pull/1437#discussion_r3194151463. I do not think it is correct to add plugin in migration, but I was considering leaving that function there anyway since in newer Postgres versions the function is there without additional plugin. However I did not find requirements in ILM for Postgres version, so I am not sure if I can just assume it will work.

src/main/resources/db/migration/V202604281200__event_history.sql
klaraf755: This number is quite arbitrary- I just checked trigger history in my DB and DB on develop server. Since there is no way to deterministically determine whether the trigger was really a part of the same event, this estimate should work for at least partial reconstruction of history.

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
klaraf755: Reduced the queries to three.

src/main/java/com/czertainly/core/service/impl/TriggerServiceImpl.java
klaraf755: Created a setter

src/main/java/com/czertainly/core/dao/entity/workflows/TriggerHistory.java
klaraf755: Removed cascade, was considering leaving `CascadeType.REMOVE`, but since currently trigger histories are not deleted at all, it is not needed. (But I think there could be some cleanup for trigger histories which have no link to existing object/event to be deleted.)

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
ivosh: I see the pattern now and a comment is inside the method. Thanks.

src/main/resources/db/migration/V202604281200__event_history.sql
ivosh: Copilot is just an AI and often has stale information.
See here: https://pgpedia.info/g/gen_random_uuid-function.html
For tests, we use testcontainers with PostgreSQL 13 which is the minimum baseline. I bet customers are using also higher versions with no problems, for better performance.

TL;DR; `gen_random_uuid` is usable. Try it.

src/main/resources/db/migration/V202604281200__event_history.sql
ivosh: Ok.

src/main/resources/db/migration/V202604281200__event_history.sql
klaraf755: Ok, I'll revert here to `gen_random_uuid`, in the worst case the plugin can be added outside of the migration.

klaraf755:
To https://github.com/OmniTrustILM/core/pull/1437#pullrequestreview-4235468993:
- first issue has been resolved by removing the extra method for evaluating permissions for details, since same can be achieved with existing method for retrieving resource objet
- the issue with transaction will be resolved in follow-up PR
- the intent in the PR was to create one event history per overriding resource, so creating an event history like this is correct

ivosh (approved):

PR #1438, opened 2026-05-05
Implement Snowflake-based serial number generation utility.

### LINKED ISSUE #198: Snowflake serial generator
Refer to https://github.com/OmniTrustILM/ilm/issues/198#issuecomment-4379501951

LukasNajman:


src/main/java/com/czertainly/core/util/serialnumber/InstanceIdResolver.java
```diff
@@ -0,0 +1,95 @@
...
+    private InstanceIdResolver() {
+    }
+
+    static int resolve() {
+        return resolve(System.getenv(ENV_VAR), InstanceIdResolver::findLocalAddress);
+    }
+
+    private static InetAddress findLocalAddress() {
+        try {
+            return InetAddress.getLocalHost();
```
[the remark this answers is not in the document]
LukasNajman: Implemented. The `findLocalAddress` now guards againts loopback, link local, any local and multicast addresses

src/main/java/com/czertainly/core/util/serialnumber/SerialNumberConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Bean
+    SerialNumberGenerator serialNumberGenerator(ClockSource clockSource) {
+        String envValue = System.getenv("TSA_INSTANCE_ID");
+        int instanceId = InstanceIdResolver.resolve();
+        if (envValue != null && !envValue.isBlank()) {
+            log.info("Instance ID resolved from TSA_INSTANCE_ID environment variable: {}", instanceId);
+        } else {
+            log.info("Instance ID resolved from IP address: {}", instanceId);
+        }
```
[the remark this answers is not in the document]
LukasNajman: The env is not queried only within resolve method

src/main/java/com/czertainly/core/util/serialnumber/SnowflakeSerialNumberGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+                if (sequence > 255) {
+                    currentTick = waitForNextTick(currentTick);
+                    sequence = 0;
+                }
+            } else {
+                sequence = 0;
+            }
+
+            lastTick = currentTick;
+            return BigInteger.valueOf((currentTick << 24) | ((long) instanceId << 8) | sequence);
```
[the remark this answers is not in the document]
LukasNajman: The computation is now done using the BigInteger arithmetic.

src/test/java/com/czertainly/core/util/serialnumber/SnowflakeSerialNumberGeneratorTest.java
```diff
@@ -0,0 +1,209 @@
...
+        var generator = new SnowflakeSerialNumberGenerator(clock, INSTANCE_ID);
+        generator.generate(); // establishes lastTick = 20
+
+        // when — clock jumps backward by 50ms (within tolerance of 100ms), then recovers
+        clock.wallTimeMillis(startTime - 50);
+        Thread.startVirtualThread(() -> {
+            try { Thread.sleep(100); } catch (InterruptedException e) { Thread.currentThread().interrupt(); }
+            clock.wallTimeMillis(startTime);
+        });
+        BigInteger id = generator.generate();
```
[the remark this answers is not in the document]
LukasNajman: Added @Timeout(2). The suggestion about deterministic latch coordination doesn't apply here - the point of this test is that the generator actually waits for real time (the virtual thread sleeps 100ms to simulate a clock
  recovering), so real time must pass. The @Timeout guard is sufficient to prevent an infinite hang.

src/test/java/com/czertainly/core/util/serialnumber/InstanceIdResolverTest.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    @Test
+    void shouldProduceValidIdFromNoArgResolve() {
+        // given — no TSA_INSTANCE_ID env var is expected to be set in test environment
+        // when — exercises the fallback chain (getLocalHost → NetworkInterface)
+        int id = InstanceIdResolver.resolve();
+
+        // then
+        assertThat(id).isBetween(0, 65535);
```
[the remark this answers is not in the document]
LukasNajman: Implemented

src/main/java/com/czertainly/core/util/serialnumber/SnowflakeSerialNumberGenerator.java
```diff
@@ -0,0 +1,93 @@
...
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    private long computeTick() {
+        return (clockSource.wallTimeMillis() - EPOCH_MILLIS) / TICK_MS;
```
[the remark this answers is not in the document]
LukasNajman: We will guard against the time before the epoch.

src/test/java/com/czertainly/core/util/clocksource/TestClockSource.java
```diff
@@ -0,0 +1,67 @@
+package com.czertainly.core.util.clocksource;
+
+
+
+import java.time.Instant;
+
+public final class TestClockSource implements ClockSource {
+
+    long wallTimeMillis;
+    long monotonicNanos;
```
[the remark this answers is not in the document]
LukasNajman: Made volatile

src/main/java/com/czertainly/core/util/serialnumber/InstanceIdResolver.java
```diff
@@ -0,0 +1,103 @@
+package com.czertainly.core.util.serialnumber;
+
+import java.net.*;
+import java.util.ArrayList;
+import java.util.function.Supplier;
+
+final class InstanceIdResolver {
+
+    static final String ENV_VAR = "ILM_INSTANCE_ID";
```
ivosh: Where is this environment variable populated from?
I searched in Core and Helm Charts but could not find anything.

src/main/java/com/czertainly/core/util/serialnumber/ClockDriftException.java
```diff
@@ -0,0 +1,8 @@
+package com.czertainly.core.util.serialnumber;
+
+public class ClockDriftException extends RuntimeException {
```
ivosh: Make `ClockDriftException` extends `SerialNumberGenerationException`. Right now they are both `RuntimeException` siblings, so a caller that wants to broadly handle "serial generation failed" has to catch both. With the proposed hierarchy, `catch (SerialNumberGenerationException)` covers everything, and code that cares about drift specifically can still `catch (ClockDriftException)` first. Same checked/unchecked status, better taxonomy.

src/main/java/com/czertainly/core/util/serialnumber/InstanceIdResolver.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    static int resolve(String envValue, Supplier<InetAddress> addressSupplier) {
+        if (envValue != null && !envValue.isBlank()) {
+            int id;
+            try {
+                id = Integer.parseInt(envValue.strip());
+            } catch (NumberFormatException e) {
+                throw new IllegalArgumentException(
+                        "ILM_INSTANCE_ID must be a valid integer, got: '" + envValue.strip() + "'");
```
ivosh: Use `ENV_VAR` constant here.

src/main/java/com/czertainly/core/util/serialnumber/InstanceIdResolver.java
```diff
@@ -0,0 +1,103 @@
...
+            }
+            if (id < 0 || id > 65535) {
+                throw new IllegalArgumentException(
+                        "ILM_INSTANCE_ID must be between 0 and 65535, got: " + id);
```
ivosh: Use `ENV_VAR` constant here.

src/main/java/com/czertainly/core/util/serialnumber/SnowflakeSerialNumberGenerator.java
```diff
@@ -0,0 +1,100 @@
+package com.czertainly.core.util.serialnumber;
+
```
ivosh, line 2: Superfluous lines. Please remove.

src/main/java/com/czertainly/core/util/serialnumber/SnowflakeSerialNumberGenerator.java
```diff
@@ -0,0 +1,100 @@
...
+import com.czertainly.core.util.clocksource.ClockSource;
+
+import java.math.BigInteger;
+import java.time.Instant;
+import java.util.concurrent.locks.ReentrantLock;
+
+final class SnowflakeSerialNumberGenerator implements SerialNumberGenerator {
+
+    static final long EPOCH_MILLIS = Instant.parse("2026-02-01T00:00:00Z").toEpochMilli();
```
ivosh, line 48: Is this hard-coded start of epoch good enough? Why not 2020? Or 1970?

src/test/java/com/czertainly/core/util/clocksource/TestClockSource.java
```diff
@@ -0,0 +1,67 @@
+package com.czertainly.core.util.clocksource;
+
```
ivosh, line 2: Superfluous lines. Please remove.

src/main/java/com/czertainly/core/util/serialnumber/InstanceIdResolver.java
```diff
@@ -0,0 +1,103 @@
...
+            // XOR-fold the last 4 bytes into 16 bits. Unlike simple truncation, XOR-folding
+            // preserves entropy from both halves of the input. We use only the last 4 bytes
+            // because the IPv6 prefix (first 8-12 bytes) is shared across hosts on the same
+            // network and adds no distinguishing value.
+            int hi = ((address[12] & 0xFF) << 8) | (address[13] & 0xFF);
+            int lo = ((address[14] & 0xFF) << 8) | (address[15] & 0xFF);
+            return (hi ^ lo) & 0xFFFF;
+        }
+        // IPv4 address
+        return ((address[address.length - 2] & 0xFF) << 8) | (address[address.length - 1] & 0xFF);
```
ivosh, line 130: Using only 16 bits from IPv4, there is a high chance of collision across pods. If we intend to use this Snowflake serial number generator for CA serial number, we get into troubles with RFC 5280 uniqueness.

Investigate the situation with `ILM_INSTANCE_ID` and issue a warning if fallback to IPv4 is happening.

ivosh (changes requested):
Good job with the Snowflake serial number generator and its tests!

In addition to the review comments applied at individual places in the source code, I have the following general comments:
* I am missing a small documentation section (such as on `serialnumber` package level or `SnowflakeSerialNumberGenerator` class) which summarizes (roughly in the same level of detail as https://github.com/OmniTrustILM/ilm/issues/198#issuecomment-4379501951)  what it actually provides, and more importantly, what are the limitations (such as fallbacks, throughput etc.).
* Bit-layout numbers (0xFFFF, 24, 8, 255) are scattered in the code (src and test) with no named constants.
* No logging on clock drift, fallbacks or other problems.
* Some files are missing a new line at the end of the file.
* Check there is a test which covers millis == EPOCH_MILLIS case. I was not able to find any but unsure if my search was exhaustive enough.
* There is one SonarQube issue unaddressed.

src/main/java/com/czertainly/core/util/serialnumber/SnowflakeSerialNumberGenerator.java
```diff
@@ -0,0 +1,100 @@
...
+        this.clockSource = clockSource;
+        this.instanceId = instanceId;
+        this.maxClockDriftMs = 100;
+    }
+
+    @Override
+    public BigInteger generate() {
+        lock.lock();
+        try {
+            long currentTick = computeTick();
```
ivosh, line 92: ## Contract Mismatch 
SerialNumberGenerator.generate() returns BigInteger — a type with no upper bound. The implementation here uses values designed to fit into 64 bits. The type signature advertises one contract (unbounded); the implementation honors a stricter, undocumented one. Nothing enforces the stricter contract, so neither callers nor the compiler can detect a violation.

## Why it matters
1. Consumers cannot see the bound. A reader of `BigInteger generate()` reasonably assumes the value can be arbitrarily large and sizes their persistence (PostgreSQL `numeric`, `BigInteger` field) accordingly. A consumer who instead assumes "Snowflake → 64 bits" and uses a `BIGINT/long` column is correct today but has no signal from the API that they're correct.
2. The implementation cannot fail on overflow. Because `BigInteger` has no ceiling, `BigInteger.valueOf(currentTick).shiftLeft(24)...` will silently produce a 65-, 80-, 160-bit result if currentTick ever exceeds 2⁴⁰. There is no controlled `SerialNumberGenerationException` — the failure surfaces downstream as a JDBC range error, a truncated certificate serial, or just a wider value than schemas were designed for.
3. The bound is one configuration change away from being violated. With EPOCH_MILLIS = 2026-02-01 and TICK_MS = 10, the tick stays under 2³⁹ for ~174 years. But if `EPOCH_MILLIS` becomes configurable  and is set to Unix epoch, or if `TICK_MS` drops to 1 for higher resolution, the headroom shrinks dramatically — with no compile-time or runtime check to flag it.
3. Tests reinforce the mismatch instead of catching it. `SnowflakeSerialNumberGeneratorTest` extracts fields using `.shiftRight(24)` and `.and(0xFFFF)`, mirroring the implementation's shifts. A future change that pushes the result past 64 bits (e.g., `shiftLeft(24)` becomes `shiftLeft(25)`) would still pass these tests — none of them assert result.bitLength() <= 64.

## Recommendation
Pick one contract and make the type system enforce it.

If the contract is "fits in 64 bits": change the return type to `long`. Arithmetic overflow then becomes detectable, no allocation per call, and consumers see the bound in the signature. Callers who need `BigInteger` for a specific column type can convert at their boundary.

If the contract is genuinely "arbitrary precision": keep `BigInteger`, but document why (e.g., a planned 128-bit layout, or X.509 serials up to 20 bytes per RFC 5280) — and either (a) add an explicit `MAX_TICK` assertion in `computeTick()` so misconfiguration fails loudly, or (b) widen the layout to actually use the precision.

The current state — `BigInteger` return with a 64-bit-shaped layout and no bound check — does not commit to either contract and gives consumers no way to tell which one they're getting.

src/main/java/com/czertainly/core/util/serialnumber/InstanceIdResolver.java
LukasNajman: I have created PR to helm charts: https://github.com/OmniTrustILM/helm-charts/pull/292

src/main/java/com/czertainly/core/util/serialnumber/ClockDriftException.java
LukasNajman: Implemented

src/main/java/com/czertainly/core/util/serialnumber/InstanceIdResolver.java
LukasNajman: Implemented

src/main/java/com/czertainly/core/util/serialnumber/SnowflakeSerialNumberGenerator.java
LukasNajman: Removed

src/main/java/com/czertainly/core/util/serialnumber/SnowflakeSerialNumberGenerator.java
LukasNajman: This is as intended - we want the serial numbers to start as small as possible. So we have picked a date in a recent past.

src/test/java/com/czertainly/core/util/clocksource/TestClockSource.java
LukasNajman: removed

src/main/java/com/czertainly/core/util/serialnumber/SnowflakeSerialNumberGenerator.java
LukasNajman: As discussed, we are keeping the BigInteger as the serial number may increase above 64 bits in the future. For now, we have guard in place, that would throw an error if the serial number generated by snowflake generator exceeds 64 bits. We also have tests that verify the length is exactly 64 bits.

src/main/java/com/czertainly/core/util/serialnumber/InstanceIdResolver.java
LukasNajman: We have documented this in code and helm chart. Also we have task to update the documentation where we will suggest using explicitly set ILM_INSTANCE_ID variable to prevent any issue. That said, the 16 lower bits should be fine for most of the deployments.

src/main/java/com/czertainly/core/util/serialnumber/InstanceIdResolver.java
LukasNajman: Done

ivosh (approved):

PR #1439, opened 2026-05-05
Asynchronous authority provider operations

### LINKED ISSUE #1216: Implement offline RA profiles with its operations
Update certificate operations to check if authority is not external when performing some action
- check response of authority when issuing certificate, when no content or not supported exception
  - update certificate state correctly to leave in `Issue Pending` and `Revoke Pending`
- add new endpoints implementations for finalizing issue and revoke operation manually
  - allow only if state of certificate is Pending
  - should use identify endpoint to verify if posted certificate content is issued by CA?
- extract code for handling certificate after successfully issued/revoked by authority to be reused for manual issue and revoke of certificate in case of external authority

3keyroman:
## Summary

This PR activates **asynchronous (HTTP 202) authority provider operations** end-to-end across the certificate lifecycle, plus the three operator-driven client endpoints needed to finalize, confirm, or cancel a pending operation. CMP and SCEP protocol layers are wired up; pre-existing synchronous flows are unchanged.

### What changes

| Area | Behaviour |
| --- | --- |
| Authority connector contract | A `202 Accepted` response on issue / revoke is now an **asynchronous-acceptance signal**. The cert moves to `PENDING_ISSUE` / `PENDING_REVOKE`; optional metadata from the 202 body is persisted on issue / renew / rekey. Connectors that always return `200` are unaffected. |
| New operator endpoints | `manuallyIssueCertificate` (finalize asynchronous issuance — operator uploads the issued cert), `manuallyConfirmRevoke` (confirm asynchronous revocation), `cancelPendingCertificateOperation` (state-aware: `PENDING_ISSUE` → `FAILED`, `PENDING_REVOKE` → `ISSUED`). |
| Cancel contract | Four-bucket connector-cancel handling: `204` aborted upstream → proceed locally; `404` / `5xx` / network → soft failure, proceed locally; `4xx` other than `404` → **hard failure, state preserved**; `422` → hard refusal, state preserved. |
| Authorization | The three new endpoints invoke `assertCertificateBelongsToRaProfile(cert, authorityUuid, raProfileUuid, op)` as a second-layer guard before any operation, ensuring the URL's RA profile + authority match the certificate's actual binding. Prevents a stale URL from operating on a cert that has since moved RA profiles. |
| CMP | New `pollReq` handler reconstructs `ip` / `cp` / `kup` based on the original request body type stored on the transaction; `pollRep` emitted when polling reaches `PENDING_*`. `BodyValidator` now routes `TYPE_POLL_REQ` / `TYPE_POLL_REP` to a dedicated `BodyPollValidator` (previously fell into the "not implemented" branch). All five previously-disabled CMP message-handler tests are re-enabled and fixed. |
| SCEP | Returns `pkiPending` for non-terminal cert states; renewal blocked on certs with a pending operation. Pure state-mapping helper extracted and unit-tested. |
| Pending-state guards | `issue` / `renew` / `rekey` / `revoke`, location push (for `FAILED` / `PENDING_REVOKE`), and RA-profile switch are all blocked on certs in `PENDING_*` with clear error messages. |
| Logging | Aligned with the [contributor logging guide](https://docs.otilm.com/docs/contributors/logging) — proper levels, full exception objects, no PII. `LoggerWrapper.logEvent` (structured event log) wired for system-level decisions, distinct from `@AuditLogged` user actions. |
| Proxy | `ProxyClientImpl.sendRequestForEntity` overrides the new default method on `ProxyClient` to **propagate the actual upstream HTTP status** (`200` vs `202`) instead of collapsing to `200` — required for the MQ-proxied path to detect asynchronous completion. |

### Notable internals

- `transitionToPendingIssue` records a `FAILED` cert-event-history entry when 202-body metadata fails to persist, rather than silently proceeding.
- `cancelPendingCertificateOperation` runs in a single transaction (dropped `@Transactional(propagation = NOT_SUPPORTED)`) so predecessor-relation cleanup and state save happen atomically.

## Backwards compatibility

- `PENDING_ISSUE` / `PENDING_REVOKE` have been in `CertificateState` since PR #322; this work activates already-defined states. DB columns, DTOs, and filter UI already support them.
- Two new Flyway migrations are pure `ADD COLUMN` (nullable, no `DEFAULT`). No existing migrations modified.
- Old connector binaries continue to work: cancel returns 404 → soft fail → core proceeds locally. The status endpoint is not required.

## Test plan

- [x] Full suite passes under JDK 21 (`mvn -B clean package -Dmaven.compiler.proc=full`).
- [x] CMP message-handler tests re-enabled and passing.
- [x] SCEP state-mapping unit tests + new `ScepServiceImplPollUnitTest` for the `pollCertificate` / `getExistingTransaction` kernels (via `ReflectionTestUtils` to avoid the heavy PKCS#7-envelope construction needed for `handlePost`).
- [x] End-to-end smoke tests for the full async lifecycle with stub connector.
- [x] `ProxyClientImpl` status-preservation + exception-unwrapping tests (57 cases).
- [x] Authority + RA-profile binding tests for the three new endpoints (success + RA-profile mismatch + authority mismatch + `NotFoundException`).
- [x] `BodyPollValidator` dispatcher tests — regression guard for the previously-broken `BodyValidator` routing.
- [x] `RevocationMessageHandlerITest` async-acceptance: when `PollFeature` returns `null`, the response carries `PKIStatus.rejection` rather than a successful `revocationNotification`.
- [x] **SonarCloud Quality Gate: OK** — coverage ≥ 80% on new code, duplications < 3%, 0 bugs / vulnerabilities / open code smells.

## Depends on

- OmniTrustILM/interfaces#664 — **merged**. The interfaces version this branch consumes is reachable via Maven Central as a snapshot fallback while the GitHub Packages publish URL is corrected by OmniTrustILM/interfaces#660.

Closes #1216


src/main/java/com/czertainly/core/service/v2/ClientOperationService.java
```diff
@@ -103,4 +105,24 @@ boolean validateRevokeCertificateAttributes(
...
+    ) throws NotFoundException, CertificateException, AlreadyExistException, ConnectorException, AttributeException;
+
+    void manuallyConfirmRevoke(
+            SecuredParentUUID authorityUuid,
+            SecuredUUID raProfileUuid,
+            String certificateUuid
+    ) throws NotFoundException;
+
+    CertificateDetailDto cancelPendingCertificateOperation(
+            SecuredParentUUID authorityUuid,
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341 — the three new endpoints now invoke `assertCertificateBelongsToRaProfile(cert, authorityUuid, raProfileUuid, ...)` as a second-layer authz binding check before any operation, so both `authorityUuid` and `raProfileUuid` are now used. Coverage added in `ClientOperationServiceV2Test` (auth-mismatch + RA-profile-mismatch tests for all three endpoints).

src/main/java/com/czertainly/core/service/v2/ClientOperationService.java
```diff
@@ -103,4 +105,24 @@ boolean validateRevokeCertificateAttributes(
...
+            SecuredUUID raProfileUuid,
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341 — the three new endpoints now invoke `assertCertificateBelongsToRaProfile(cert, authorityUuid, raProfileUuid, ...)` as a second-layer authz binding check before any operation, so both `authorityUuid` and `raProfileUuid` are now used. Coverage added in `ClientOperationServiceV2Test` (auth-mismatch + RA-profile-mismatch tests for all three endpoints).

src/main/java/com/czertainly/core/service/v2/ClientOperationService.java
```diff
@@ -103,4 +105,24 @@ boolean validateRevokeCertificateAttributes(
...
+
+    CertificateDetailDto manuallyIssueCertificate(
+            SecuredParentUUID authorityUuid,
+            SecuredUUID raProfileUuid,
+            String certificateUuid,
+            UploadCertificateRequestDto request
+    ) throws NotFoundException, CertificateException, AlreadyExistException, ConnectorException, AttributeException;
+
+    void manuallyConfirmRevoke(
+            SecuredParentUUID authorityUuid,
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341 — the three new endpoints now invoke `assertCertificateBelongsToRaProfile(cert, authorityUuid, raProfileUuid, ...)` as a second-layer authz binding check before any operation, so both `authorityUuid` and `raProfileUuid` are now used. Coverage added in `ClientOperationServiceV2Test` (auth-mismatch + RA-profile-mismatch tests for all three endpoints).

src/main/java/com/czertainly/core/service/v2/ClientOperationService.java
```diff
@@ -103,4 +105,24 @@ boolean validateRevokeCertificateAttributes(
...
+            SecuredUUID raProfileUuid,
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341 — the three new endpoints now invoke `assertCertificateBelongsToRaProfile(cert, authorityUuid, raProfileUuid, ...)` as a second-layer authz binding check before any operation, so both `authorityUuid` and `raProfileUuid` are now used. Coverage added in `ClientOperationServiceV2Test` (auth-mismatch + RA-profile-mismatch tests for all three endpoints).

src/main/java/com/czertainly/core/service/v2/ClientOperationService.java
```diff
@@ -103,4 +105,24 @@ boolean validateRevokeCertificateAttributes(
             SecuredUUID raProfileUuid,
             List<RequestAttribute> attributes
     ) throws ConnectorException, ValidationException, NotFoundException;
+
+    CertificateDetailDto manuallyIssueCertificate(
+            SecuredParentUUID authorityUuid,
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341 — the three new endpoints now invoke `assertCertificateBelongsToRaProfile(cert, authorityUuid, raProfileUuid, ...)` as a second-layer authz binding check before any operation, so both `authorityUuid` and `raProfileUuid` are now used. Coverage added in `ClientOperationServiceV2Test` (auth-mismatch + RA-profile-mismatch tests for all three endpoints).

src/main/java/com/czertainly/core/service/v2/ClientOperationService.java
```diff
@@ -103,4 +105,24 @@ boolean validateRevokeCertificateAttributes(
...
+            SecuredUUID raProfileUuid,
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341 — the three new endpoints now invoke `assertCertificateBelongsToRaProfile(cert, authorityUuid, raProfileUuid, ...)` as a second-layer authz binding check before any operation, so both `authorityUuid` and `raProfileUuid` are now used. Coverage added in `ClientOperationServiceV2Test` (auth-mismatch + RA-profile-mismatch tests for all three endpoints).

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1323,273 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
+            SecuredParentUUID authorityUuid, SecuredUUID raProfileUuid, String certificateUuid,
+            UploadCertificateRequestDto request) throws NotFoundException, java.security.cert.CertificateException, AlreadyExistException, ConnectorException, AttributeException {
+        // Two-layer authz: RA_PROFILE/DETAIL gates use of the RA profile (above);
+        // CERTIFICATE/ISSUE gates the actual issuance authority. Mirrors the synchronous
+        // issueCertificateAction flow which calls checkIssuePermissions() before persisting.
+        certificateService.checkIssuePermissions();
+
+        Certificate certificate = certificateRepository.findWithAssociationsByUuid(UUID.fromString(certificateUuid))
+                .orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
+
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341. The three endpoints now invoke a private `assertCertificateBelongsToRaProfile(certificate, authorityUuid, raProfileUuid, operation)` helper as the first thing they do (right after `certificateService.checkXxxPermissions()`). The helper compares the URL's RA profile + authority against `certificate.getRaProfileUuid()` and `certificate.getRaProfile().getAuthorityInstanceReferenceUuid()` and throws `ValidationException` on mismatch — preventing a stale URL from operating on a cert that has since moved RA profiles. Tests in `ClientOperationServiceV2Test` cover both the RA-profile-mismatch and authority-mismatch paths for all three endpoints (commit 94d6b64f).

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1323,273 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
+    @ExternalAuthorization(resource = Resource.RA_PROFILE, action = ResourceAction.DETAIL, parentResource = Resource.AUTHORITY, parentAction = ResourceAction.DETAIL)
+    public void manuallyConfirmRevoke(
+            SecuredParentUUID authorityUuid, SecuredUUID raProfileUuid, String certificateUuid) throws NotFoundException {
+        // Two-layer authz: RA_PROFILE/DETAIL gates use of the RA profile (above);
+        // CERTIFICATE/REVOKE gates the actual revocation authority.
+        certificateService.checkRevokePermissions();
+
+        Certificate cert = certificateRepository.findWithAssociationsByUuid(UUID.fromString(certificateUuid))
+                .orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
+
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341. The three endpoints now invoke a private `assertCertificateBelongsToRaProfile(certificate, authorityUuid, raProfileUuid, operation)` helper as the first thing they do (right after `certificateService.checkXxxPermissions()`). The helper compares the URL's RA profile + authority against `certificate.getRaProfileUuid()` and `certificate.getRaProfile().getAuthorityInstanceReferenceUuid()` and throws `ValidationException` on mismatch — preventing a stale URL from operating on a cert that has since moved RA profiles. Tests in `ClientOperationServiceV2Test` cover both the RA-profile-mismatch and authority-mismatch paths for all three endpoints (commit 94d6b64f).

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1323,273 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
+    @Override
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    @ExternalAuthorization(resource = Resource.RA_PROFILE, action = ResourceAction.DETAIL, parentResource = Resource.AUTHORITY, parentAction = ResourceAction.DETAIL)
+    public CertificateDetailDto cancelPendingCertificateOperation(
+            SecuredParentUUID authorityUuid, SecuredUUID raProfileUuid, String certificateUuid,
+            CancelPendingCertificateRequestDto request) throws NotFoundException {
+        Certificate cert = certificateRepository.findWithAssociationsByUuid(UUID.fromString(certificateUuid))
+                .orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
+
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341. The three endpoints now invoke a private `assertCertificateBelongsToRaProfile(certificate, authorityUuid, raProfileUuid, operation)` helper as the first thing they do (right after `certificateService.checkXxxPermissions()`). The helper compares the URL's RA profile + authority against `certificate.getRaProfileUuid()` and `certificate.getRaProfile().getAuthorityInstanceReferenceUuid()` and throws `ValidationException` on mismatch — preventing a stale URL from operating on a cert that has since moved RA profiles. Tests in `ClientOperationServiceV2Test` cover both the RA-profile-mismatch and authority-mismatch paths for all three endpoints (commit 94d6b64f).

src/main/java/com/czertainly/core/service/cmp/message/handler/PollFeature.java
```diff
@@ -69,17 +78,33 @@ public Certificate pollCertificate(ASN1OctetString tid, String serialNumber, Str
...
+                // Asynchronous-completion signal: if the certificate has reached a PENDING_*
+                // state, the authority provider connector returned 202 Accepted and the
+                // operation will complete asynchronously. Stop polling and signal the caller to
+                // return a CMP pollRep so the client knows to retry later. Captured here (inside
+                // the loop) to also catch the realistic case where the transition happens during
+                // the wait window, not only when the cert is already PENDING_* at the start.
+                if (polledCert.getState() == CertificateState.PENDING_ISSUE
+                        || polledCert.getState() == CertificateState.PENDING_REVOKE) {
+                    isPendingAsynchronous = true;
+                    break;
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit dac58bf0 + 426a8341. `RevocationMessageHandler` now consumes the null return from `PollFeature.pollCertificate(..., REVOKED)` explicitly: when null is returned (cert is in `PENDING_REVOKE`), the handler throws `CmpProcessingException` with a clear message — "revocation accepted asynchronously by authority (certificate is in PENDING_REVOKE); CMP does not support pending revocation. Use the platform API to confirm or cancel." The outer per-cert catch turns this into `PKIStatus.rejection` rather than the previous `revocationNotification`. RFC 4210 §5.2.6 limits poll-request/poll-response to ip/cp/kup contexts, so this rejection is the cleanest possible CMP-level response. Pinned by `RevocationMessageHandlerITest#test_handle_revocation_rejectsCleanly_whenAuthorityAcceptsAsynchronously` (commit 94d6b64f).

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -326,6 +351,69 @@ public void issueCertificateAction(final UUID certificateUuid, boolean isApprove
...
+    private void transitionToPendingIssue(Certificate certificate, CertificateDataResponseDto acceptedBody, ResourceAction originatingAction) {
+        if (acceptedBody != null && acceptedBody.getMeta() != null && !acceptedBody.getMeta().isEmpty()) {
+            try {
+                attributeEngine.updateMetadataAttributes(acceptedBody.getMeta(),
+                        ObjectAttributeContentInfo.builder(Resource.CERTIFICATE, certificate.getUuid())
+                                .connector(certificate.getRaProfile().getAuthorityInstanceReference().getConnectorUuid())
+                                .build());
+            } catch (Exception metaEx) {
+                logger.warn("Failed to persist metadata from 202 response for cert {}: {}",
+                        certificate.getUuid(), metaEx.getMessage(), metaEx);
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341. The catch block now records a `FAILED` cert-event-history entry (with the `CertificateEvent.ISSUE` event type) describing that connector metadata returned with HTTP 202 could not be persisted, and warns that cancellation of this pending operation may be limited if the connector requires the original metadata. The certificate still moves to PENDING_ISSUE because the connector has already accepted the operation upstream — leaving it in REQUESTED would orphan the upstream operation. The cancel path falls back to whatever metadata is queryable via the attribute engine (possibly empty).

src/main/java/com/czertainly/core/service/cmp/message/handler/PollReqMessageHandler.java
```diff
@@ -0,0 +1,181 @@
...
+
+        try {
+            CMPCertificate cmpCert = CMPCertificate.getInstance(x509.getEncoded());
+            CertResponse certResponse = new CertResponse(
+                    certReqId,
+                    new PKIStatusInfo(PKIStatus.granted),
+                    new CertifiedKeyPair(new CertOrEncCert(cmpCert)),
+                    null);
+            CertRepMessage repMessage = new CertRepMessage(null, new CertResponse[]{certResponse});
+            PKIBody body = new PKIBody(responseBodyType, repMessage);
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341. `PollReqMessageHandler` now rejects polls for revocation transactions explicitly: at the top of `handle()`, if `transaction.getOriginalRequestBodyType() == TYPE_REVOCATION_REQ`, the handler throws `CmpProcessingException` ("pollReq not supported for revocation transactions"). Defensively, it also rejects polls landing on a cert in `PENDING_REVOKE` even when `originalRequestBodyType` is null (legacy transactions). Pinned by `PollReqMessageHandlerTest#rejectsRevocationPoll_whenTransactionOriginatedFromRevocationRequest` and `rejectsPoll_whenCertificateInPendingRevoke_andNoOriginalBodyType` (commit eabb9b02).

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1323,273 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
+                        .connector(raProfile.getAuthorityInstanceReference().getConnectorUuid()).build()));
+        try {
+            connectorApiFactory.getCertificateApiClientV2(connectorDto)
+                    .identifyCertificate(connectorDto,
+                            raProfile.getAuthorityInstanceReference().getAuthorityInstanceUuid(), idReq);
+        } catch (ConnectorException | ValidationException e) {
+            certificateEventHistoryService.addEventHistory(certificate.getUuid(), CertificateEvent.ISSUE,
+                    CertificateEventStatus.FAILED,
+                    "Manual upload rejected by connector identify: " + e.getMessage(), "");
+            throw new ValidationException("Manual upload rejected by connector identify: " + e.getMessage());
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341. The catch around `certificateService.identifyCertificate(...)` now narrows to `ValidationException` only — connector outages, 5xx responses, and auth failures (`ConnectorException` and subtypes) propagate to the caller as the appropriate connector exception type rather than being collapsed into a user-input `ValidationException`. Only genuine input-shape validation failures from the identify step are wrapped with the "Manual upload" context.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1323,273 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
+        }
+
+        // Persist cert content + transition to ISSUED via the standard issuance pipeline.
+        try {
+            certificateService.issueRequestedCertificate(certificate.getUuid(), request.getCertificate(), List.of());
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341. `manuallyIssueCertificate` now extracts `response.getMeta()` from the `CertificateIdentificationResponseDto` returned by `identifyCertificate` and passes it through to `certificateService.issueRequestedCertificate(uuid, certData, identifyMeta)`. Connector-emitted metadata is no longer dropped — matches the behaviour of the RA-profile-switch flow.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1323,273 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
+                    OperationResult.FAILURE,
+                    null,
+                    List.of(new ResourceObjectIdentity(cert.getSerialNumber(), cert.getUuid())),
+                    "Authority refused to cancel pending "
+                            + (isCancelIssue ? "issuance" : "revocation")
+                            + ": " + upstreamRefused.getMessage());
+            throw new ValidationException(
+                    "Authority refused to cancel the operation: " + upstreamRefused.getMessage());
+        } catch (Exception soft) {
+            // 404 / 5xx / network: soft failure, log and continue with local transition.
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341. The cancel exception handling is now refined into four buckets: (1) `ValidationException` (mapped from upstream 422) → hard refusal, state preserved; (2) `ConnectorEntityNotFoundException` (404) → soft failure, proceed locally — the connector simply doesn't track this operation; (3) `ConnectorClientException` (4xx other than 404, e.g. 400/401/403) → **hard failure**, state preserved — these indicate auth or malformed-request issues that should not allow silent cleanup; (4) `ConnectorServerException` / `ConnectorCommunicationException` (5xx / network) → soft failure, proceed locally. Pinned by `cancelPendingCertificateOperation_isRejectedAndStateUnchanged_when4xxClientErrorFromConnector` in `ClientOperationServiceV2Test`.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1323,273 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
+                    null,
+                    List.of(new ResourceObjectIdentity(cert.getSerialNumber(), cert.getUuid())),
+                    "Connector cancel call failed ("
+                            + soft.getClass().getSimpleName()
+                            + "); proceeding with local cancel");
+        }
+
+        // Local transition (only reached when 422 was NOT thrown above).
+        cert.setState(targetState);
+        if (!isCancelIssue) {
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341. The cancel-pending-issue path now mirrors the renew/rekey failure paths: it calls `certificateRelationRepository.deleteAll(cert.getPredecessorRelations())` followed by `cert.getPredecessorRelations().clear()` before the state save. Dropped `@Transactional(NOT_SUPPORTED)` from `cancelPendingCertificateOperation` so the relation cleanup and the state save happen in a single transaction — avoiding the Hibernate proxy-load failure on save when relations are gone. Pinned by `cancelPendingCertificateOperation_clearsPredecessorRelations_whenCancellingPendingIssue` in `ClientOperationServiceV2Test`.

src/main/java/com/czertainly/core/service/cmp/impl/CmpServiceImpl.java
```diff
@@ -235,12 +242,15 @@ public ResponseEntity<byte[]> handlePost(String profileName, byte[] request) thr
                     pkiResponse = certConfirmMessageHandler.handle(pkiRequest, configuration);
                     LoggingHelper.putAuditLogOperation(Operation.CMP_CONFIRM);
                     break;
+                case PKIBody.TYPE_POLL_REQ:                        // (25) pollReq, Polling Request (RFC 4210 §5.2.6)
+                    pkiResponse = pollReqMessageHandler.handle(pkiRequest, configuration);
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341. `BodyValidator` now routes `TYPE_POLL_REQ` to `BodyPollValidator.validateIn` and `TYPE_POLL_REP` to `BodyPollValidator.validateOut` instead of the previous "is not implemented" path. The new `BodyPollValidator` enforces RFC 4210 §5.3.22 / §5.2.6 invariants: at least one `certReqId` for both, non-negative `checkAfter` for pollRep. Pinned by `BodyPollValidatorTest#bodyValidator_routesPollReq_toValidatorInsteadOfNotImplementedError` and `bodyValidator_routesPollRep_toValidatorInsteadOfNotImplementedError` (regression guards against the previously-broken behaviour).

src/main/java/com/czertainly/core/service/cmp/message/handler/CrmfMessageHandler.java
```diff
@@ -190,6 +190,43 @@ public PKIMessage handle(PKIMessage request, ConfigurationContext configuration)
...
+                        .addBody(PkiMessageBuilder.createPollRepBody(
+                                crmf.getCertReqId(),
+                                60L /* checkAfter, seconds */,
+                                "Awaiting asynchronous completion"))
+                        .addExtraCerts(null)
+                        .build();
+            } catch (Exception e) {
+                LOG.error("TID={} | CRMF pollRep message cannot be built (type={})", tid, msgBodyType, e);
+                throw new CmpCrmfValidationException(tid, request.getBody().getType(), PKIFailureInfo.systemFailure,
+                        "CRMF pollRep cannot be built, type=" + PkiMessageDumper.msgTypeAsString(request.getBody().getType()));
```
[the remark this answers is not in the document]
3keyroman: Resolved in commit 426a8341. `BodyValidator` now routes `TYPE_POLL_REQ` to `BodyPollValidator.validateIn` and `TYPE_POLL_REP` to `BodyPollValidator.validateOut` instead of the previous "is not implemented" path. The new `BodyPollValidator` enforces RFC 4210 §5.3.22 / §5.2.6 invariants: at least one `certReqId` for both, non-negative `checkAfter` for pollRep. Pinned by `BodyPollValidatorTest#bodyValidator_routesPollReq_toValidatorInsteadOfNotImplementedError` and `bodyValidator_routesPollRep_toValidatorInsteadOfNotImplementedError` (regression guards against the previously-broken behaviour).

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1381,369 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
+            logger.warn("Connector cancel call failed for cert {} ({}: {}) — proceeding with local cancel",
+                    cert.getUuid(), infra.getClass().getSimpleName(), infra.getMessage(), infra);
+            eventLogger.logEvent(
+                    Operation.CANCEL,
+                    OperationResult.SUCCESS,
+                    null,
+                    List.of(new ResourceObjectIdentity(cert.getSerialNumber(), cert.getUuid())),
+                    "Connector cancel call failed (" + infra.getClass().getSimpleName()
+                            + "); proceeded with local cancel");
+        } catch (ConnectorException unexpected) {
```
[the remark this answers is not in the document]
3keyroman: Not unreachable. `ConnectorException` has more subtypes than `ConnectorClientException`, `ConnectorCommunicationException`, `ConnectorEntityNotFoundException`, and `ConnectorServerException` — notably `ConnectorProblemException` (and any future subtype). The defensive catch handles those cases as soft failures, mirroring the 5xx/network treatment, rather than letting them propagate as opaque errors that would leave the cert in PENDING_*.

src/main/java/com/czertainly/core/service/cmp/message/validator/impl/BodyPollValidator.java
```diff
@@ -0,0 +1,89 @@
...
+                    "pollReq body validator: at least one certReqId is required");
+        }
+        return null;
+    }
+
+    /**
+     * Validate an outbound {@code pollRep} message: assert the body type and the
+     * presence of exactly one {@code certReqId} with a non-negative {@code checkAfter}.
+     */
+    public Void validateOut(PKIMessage response, ConfigurationContext configuration) throws CmpBaseException {
```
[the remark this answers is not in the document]
3keyroman: The `configuration` parameter is part of the `Validator<PKIMessage, Void>` interface signature; `BodyPollValidator` implements that contract along with all other validators in this package (most also don't read it). The current poll-message validation is purely structural per RFC 4210 §5.3.22 / §5.2.6 — no profile-driven rules apply yet — but the parameter remains for signature compatibility and future extensibility.

src/main/java/com/czertainly/core/service/cmp/message/validator/impl/BodyPollValidator.java
```diff
@@ -0,0 +1,89 @@
...
+ *       reason       PKIFreeText OPTIONAL }
+ * </pre>
+ */
+public class BodyPollValidator extends BaseValidator implements Validator<PKIMessage, Void> {
+
+    /**
+     * Validate an inbound {@code pollReq} message: assert the body type and that at
+     * least one {@code certReqId} is present.
+     */
+    public Void validateIn(PKIMessage request, ConfigurationContext configuration) throws CmpBaseException {
```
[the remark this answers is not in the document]
3keyroman: Same as above: `configuration` is part of the `Validator<PKIMessage, Void>` interface signature. Removing it would break interface conformance with all other validators in this package.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1381,380 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
```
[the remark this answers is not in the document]
3keyroman: Same as the previous unreachable-catch finding (resolved earlier in this PR). Not unreachable: `ConnectorException` has more subtypes than the four explicitly caught (`ConnectorClientException`, `ConnectorCommunicationException`, `ConnectorEntityNotFoundException`, `ConnectorServerException`) — notably `ConnectorProblemException` (and any future subtype). The defensive catch handles those cases as soft failures, mirroring the 5xx/network treatment, rather than letting them propagate as opaque errors that would leave the cert in PENDING_*. Marking won't-fix on SonarCloud accordingly.

src/main/java/com/czertainly/core/messaging/proxy/ProxyClientImpl.java
```diff
@@ -86,6 +87,81 @@ public <T> T sendRequest(
...
+            ApiClientConnectorInfo connector,
+            String path,
+            String method,
+            Object body,
+            Class<T> responseType) throws ConnectorException {
+        Duration timeout = proxyProperties.requestTimeout();
+        try {
+            CompletableFuture<ResponseEntity<T>> future =
+                    sendRequestForEntityAsync(connector, path, method, body, responseType, timeout);
+            return future.get(timeout.toMillis() + 5000, TimeUnit.MILLISECONDS);
```
ivosh: At least define a constant for this hard-coded value so it has a clear intent. Ideally configurable via application properties.

src/main/java/com/czertainly/core/messaging/proxy/ProxyClientImpl.java
```diff
@@ -86,6 +87,81 @@ public <T> T sendRequest(
...
+        } catch (TimeoutException e) {
+            throw new ConnectorCommunicationException(
+                    "Proxy request timed out after " + timeout, e, connector);
+        } catch (ExecutionException e) {
+            Throwable cause = e.getCause();
+            if (cause instanceof java.util.concurrent.CompletionException ce && ce.getCause() != null) {
```
ivosh: Use import for `CompletionException`.

src/main/java/com/czertainly/core/service/cmp/message/handler/CrmfMessageHandler.java
```diff
@@ -190,6 +190,43 @@ public PKIMessage handle(PKIMessage request, ConfigurationContext configuration)
...
+                    request.getBody().getType()));
+
+            try {
+                LOG.info("TID={} | CRMF {} accepted asynchronously (cert {}); returning pollRep",
+                        tid, msgBodyType, requestedCert.getUuid());
+                return new PkiMessageBuilder(configuration)
+                        .addHeader(PkiMessageBuilder.buildBasicHeaderTemplate(request))
+                        .addBody(PkiMessageBuilder.createPollRepBody(
+                                crmf.getCertReqId(),
+                                60L /* checkAfter, seconds */,
```
ivosh: Do you agree this parameter should be configurable? At least application properties, ideally exposed via UI.

src/main/java/com/czertainly/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -549,24 +549,41 @@ private ScepResponse generateCsr(ScepRequest scepRequest, IntuneScepServiceClien
         return scepResponse;
     }
 
+    /**
+     * Maps a {@link CertificateState} to the {@link PkiStatus} that SCEP should report per
+     * RFC 8894 §3.3.2. Pure function — has no side effects and does not depend on instance
```
ivosh: I don't think we need the sentence "Pure function..." as it is obvious.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -67,6 +83,13 @@
 @Transactional
 public class ClientOperationServiceImpl implements ClientOperationService {
     private static final Logger logger = LoggerFactory.getLogger(ClientOperationServiceImpl.class);
+
+    // Structured event-log surface (per CZERTAINLY contributors logging guide).
```
ivosh: There is a lot of comments around this eventLogger. Centralize them in one place only - the intent is clear also by reading code only  - no commentary is necessary.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -326,6 +357,121 @@ public void issueCertificateAction(final UUID certificateUuid, boolean isApprove
...
+                List.of(new ResourceObjectIdentity(certificate.getSerialNumber(), certificate.getUuid())),
+                "Connector accepted asynchronously (HTTP 202); certificate transitioned to PENDING_ISSUE");
+
+        logger.info("Certificate {} transitioned to PENDING_ISSUE (originating action: {})",
+                certificate.getUuid(), originatingAction.getCode());
+    }
+
+    /**
+     * Reject the operation when the URL's authority / RA profile do not match the
+     * certificate's authority / RA profile. The class-level
```
ivosh: I don't think we need all this reasoning here - the intent is clear just from the method name.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1381,380 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
+    /**
+     * Two-layer authorization. The annotation gates use of the RA profile (RA_PROFILE
+     * with action DETAIL); the call to checkIssuePermissions below gates the actual
+     * issuance authority on CERTIFICATE. Mirrors what issueCertificateAction does
+     * before persisting in the synchronous flow.
+     */
+    @ExternalAuthorization(resource = Resource.RA_PROFILE, action = ResourceAction.DETAIL, parentResource = Resource.AUTHORITY, parentAction = ResourceAction.DETAIL)
+    public CertificateDetailDto manuallyIssueCertificate(
+            SecuredParentUUID authorityUuid, SecuredUUID raProfileUuid, String certificateUuid,
+            UploadCertificateRequestDto request) throws NotFoundException, java.security.cert.CertificateException, AlreadyExistException, ConnectorException, AttributeException {
```
ivosh: Use import.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1381,380 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
+
+        eventProducer.produceMessage(
+                CertificateActionPerformedEventHandler.constructEventMessage(cert.getUuid(), ResourceAction.REVOKE));
+
+        logger.info("Certificate {} revocation confirmed manually", cert.getUuid());
+    }
+
+    @Override
+    @ExternalAuthorization(resource = Resource.RA_PROFILE, action = ResourceAction.DETAIL, parentResource = Resource.AUTHORITY, parentAction = ResourceAction.DETAIL)
+    public CertificateDetailDto cancelPendingCertificateOperation(
```
ivosh, line 1594: I think `@Transactional(propagation = Propagation.NOT_SUPPORTED)` is missing on this method. Refer to the other existing methods and also new methods manuallyIssueCertificate and manuallyConfirmRevoke which all call a connector.

You probably want to grab a transaction manually via `transactionManager.getTransaction()`, like it is done in `renewCertificate`.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1381,380 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
+        return refreshed.mapToDto();
+    }
+
+    @Override
+    /**
+     * Two-layer authorization, same shape as manuallyIssueCertificate above: the
+     * annotation gates RA profile use; checkRevokePermissions gates the actual
+     * revocation authority on CERTIFICATE.
+     */
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
```
ivosh, line 1507: Why `Propagation.NOT_SUPPORTED` is used here? This method does not call a connector.

src/main/java/com/czertainly/core/service/cmp/message/handler/PollReqMessageHandler.java
```diff
@@ -0,0 +1,199 @@
...
+ *       {@code originalRequestBodyType} stored on the transaction (so a client that sent ir
+ *       gets ip back, cr → cp, kur → kup);</li>
+ *   <li><b>error response</b> — when the operation failed or was rejected, or when no
+ *       transaction exists for this {@code transactionID}.</li>
+ * </ul>
+ */
+@Component
+public class PollReqMessageHandler implements MessageHandler<PKIMessage> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(PollReqMessageHandler.class);
```
ivosh: Use `@Slf4j` instead.

src/main/java/com/czertainly/core/service/cmp/message/handler/PollReqMessageHandler.java
```diff
@@ -0,0 +1,199 @@
...
+                                              ASN1OctetString tid, ASN1Integer certReqId,
+                                              CmpTransaction transaction, Certificate certificate)
+            throws CmpBaseException {
+        Integer originalBodyType = transaction.getOriginalRequestBodyType();
+        // Default to cp (TYPE_CERT_REP) when the transaction was created before the body-type
+        // column existed (NULL on legacy rows). Most cmp v2 clients accept cp as the response
+        // for either ir or cr.
+        int responseBodyType = (originalBodyType == null
+                ? PKIBody.TYPE_CERT_REP
+                : (originalBodyType + 1));
```
ivosh: What is this magic good for? Could you use an explicit switch on TYPE_INIT_REQ / TYPE_CERT_REQ / TYPE_KEY_UPDATE_REQ instead?

src/main/java/com/czertainly/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -549,24 +549,41 @@ private ScepResponse generateCsr(ScepRequest scepRequest, IntuneScepServiceClien
...
+     * state — so the mapping itself is unit-testable independent of message-flow setup.
+     *
+     * @return {@code SUCCESS} for ISSUED; {@code FAILURE} for REJECTED/FAILED;
+     *         {@code PENDING} for everything else (REQUESTED, PENDING_APPROVAL, PENDING_ISSUE,
+     *         PENDING_REVOKE, REVOKED).
```
ivosh: Why is `REVOKED` mapped to `PENDING`? I think the intent of this method is to map certificate state into PKI status categories:
* certificate final states (ISSUED, REJECTED, FAILED) -> SUCCESS/FAILURE
* certificate transitional states (REQUESTED, PENDING_*) -> PENDING

However I see REVOKED as a final state so it should be mapped to FAILURE. Do you agree?

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1381,380 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
+    @Override
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    /**
+     * Two-layer authorization. The annotation gates use of the RA profile (RA_PROFILE
+     * with action DETAIL); the call to checkIssuePermissions below gates the actual
+     * issuance authority on CERTIFICATE. Mirrors what issueCertificateAction does
+     * before persisting in the synchronous flow.
+     */
+    @ExternalAuthorization(resource = Resource.RA_PROFILE, action = ResourceAction.DETAIL, parentResource = Resource.AUTHORITY, parentAction = ResourceAction.DETAIL)
+    public CertificateDetailDto manuallyIssueCertificate(
```
ivosh, line 1401: The new methods manuallyIssueCertificate, manuallyConfirmRevoke, cancelPendingCertificateOperation are each 150+ lines long, putting a high cognitive pressure on everyone reading the code. It is apparent they have been written by AI with a context window going through the roof. Try to refactor them. Delete superfluous comments - well-written code is self-describing.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -777,11 +945,20 @@ public void revokeCertificateAction(final UUID certificateUuid, ClientCertificat
             caRequest.setCertificate(certificate.getCertificateContent().getContent());
 
             var connectorDto = raProfile.getAuthorityInstanceReference().getConnector().mapToApiClientDtoV1();
-            connectorApiFactory.getCertificateApiClientV2(connectorDto).revokeCertificate(
+            ResponseEntity<Void> revokeResponse = connectorApiFactory.getCertificateApiClientV2(connectorDto).revokeCertificate(
                     connectorDto,
                     raProfile.getAuthorityInstanceReference().getAuthorityInstanceUuid(),
                     caRequest);
 
+            if (revokeResponse.getStatusCode().value() == 202) {
```
ivosh, line 952: If the connector invocation `revokeCertificate` throws a `ConnectorException`, it is caught at line 967 via generic `catch (Exception)` and certificate state is ISSUED even though it could have been `PENDING_APPROVAL` as per the guard at line 929. Check the intent here.

src/main/java/com/czertainly/core/service/impl/LocationServiceImpl.java
```diff
@@ -426,8 +426,18 @@ public LocationDto pushCertificateToLocation(SecuredParentUUID entityUuid, Secur
             throw new ValidationException(ValidationError.create("Cannot push archived certificate %s to location %s".formatted(certificate, location.getName())));
 
 
-        if (certificate.getState().equals(CertificateState.REJECTED)) {
-            throw new ValidationException(ValidationError.create("Cannot push rejected certificate %s to location %s".formatted(certificate, location.getName())));
+        if (certificate.getState() == CertificateState.REJECTED
```
ivosh, line 429: Check if a similar guard is needed also in removeCertificateFromLocation method

src/main/java/com/czertainly/core/service/cmp/message/handler/PollFeature.java
```diff
@@ -41,9 +41,17 @@ public void setCertificateService(CertificateService certificateService) {
      * Convert asynchronous behaviour (manipulation with certificate, e.g. issuing/re-keying/revoking) to synchronous
      * (cmp client ask for certificate) using polling certificate until certificate.
      *
+     * <p>If the certificate is in an asynchronous {@code PENDING_ISSUE} or {@code PENDING_REVOKE}
```
ivosh: Is there a possibility for race condition here? Imagine thread entering the method with expectedState PENDING_ISSUE. Then it goes on polling. Somewhere in the middle an asynchronous cancelPendingCertificateOperation sets the state to FAILED. Now the method does not end cleanly but with a systemFailure.

ivosh (changes requested):
Thank you for the PR. It is a large one. I was not able to review all aspects of the CMP/SCEP protocol handling as I am not familiar with them - you should involve another reviewer as well.

In addition to the comments distributed in the files, I have also the following general comments:
1. How does the existing `CertificateServiceImpl.revokeCertificate()` operate with the new async model? Is it possible that a PENDING_ISSUE/PENDING_REVOKE is silently overwritten by REVOKED here?
2. Similarly, how does the existing `CertificateServiceImpl.issueRequestedCertificate()` operate with the new async model? Because prepareIssuedCertificate sets certificate state to ISSUED.
3. Check the comments generated by AI. Sometimes they are really distracting for the reader - well-written code is self-documenting and does not need to be heavily commented. If one aspect needs to be documented, then document it at one place only - not at every call site.
4. I am missing a central place for Certificate state transitions: a state machine which would have a method such as `CertificateStateService.transition(cert, to, reason, actor)` with all the preconditions and centralized addEventHistory. Please create a follow-up issue - we don't want to scope-screep this PR anymore.

src/main/java/com/czertainly/core/service/impl/LocationServiceImpl.java
3keyroman: No guard needed for removeCertificateFromLocation. Removing a cert from a location only invalidates the destination-side copy (the deployed cert at e.g. a server keystore) — it does not propagate any state back to the cert. During PENDING_REVOKE, removing from a location is actually desirable in many cases: the operator wants the cert pulled from production keystores ASAP, before the CA confirms revocation. During PENDING_ISSUE the cert isn't actually deployed yet, so removal is harmless (the push guard prevents pushes). The asymmetry with push is intentional: push could deploy a cert that's about to be revoked, which is wrong; remove cannot create that kind of inconsistency.

src/main/java/com/czertainly/core/messaging/proxy/ProxyClientImpl.java
3keyroman: Extracted as `MESSAGE_ROUND_TRIP_BUFFER_MS = 5_000L` with a doc comment explaining what it covers (broker dispatch + proxy worker round-trip + response correlation overhead) — commit bc924725. Making it configurable via `application.properties` is a fair follow-up but feels separable: the buffer is intrinsic to the MQ-proxied transport, not a tunable per-deployment knob, and exposing it would require thinking about whether 5 s is the right default across all environments. Happy to do it in a follow-up if you feel strongly — let me know.

src/main/java/com/czertainly/core/messaging/proxy/ProxyClientImpl.java
3keyroman: Done in commit bc924725 — added `import java.util.concurrent.CompletionException` and dropped the FQN at the instanceof site.

src/main/java/com/czertainly/core/service/cmp/message/handler/CrmfMessageHandler.java
3keyroman: Extracted as `POLL_REP_CHECK_AFTER_SECONDS = 60L` with a Javadoc citing RFC 4210 §5.3.22 — commit bc924725. Same trade-off as the proxy buffer: 60 s balances "client doesn't hammer the server" vs "operator doesn't wait too long for state to propagate", and exposing it via `application.properties` (or per-CMP-profile in the UI) is reasonable but feels like a follow-up. Happy to wire it up if you want it in this PR — let me know.

src/main/java/com/czertainly/core/service/scep/impl/ScepServiceImpl.java
3keyroman: Removed in commit bc924725.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
3keyroman: Done in commit bc924725 — collapsed the per-call-site comments into a single canonical Javadoc on the `eventLogger` field declaration. The Operation + OperationResult + description arguments at each `eventLogger.logEvent` site are self-describing.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
3keyroman: Removed in commit bc924725 — the method name `assertCertificateBelongsToRaProfile` carries the intent.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
3keyroman: Done in commit bc924725 — dropped the `java.security.cert.CertificateException` FQN on the throws clause; the import was already present at file scope.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
3keyroman: Done in commit bc924725. `cancelPendingCertificateOperation` now has `@Transactional(propagation = NOT_SUPPORTED)` on the method, with the cleanup writes (`deleteAll(predecessorRelations)` + state save) bracketed in an explicit `transactionManager.getTransaction(new DefaultTransactionDefinition())` block after the connector call — same shape as `renewCertificate`. Then refactored further in 08e36371 into a named helper `commitLocalCancel(cert, target)`. The `JpaObjectRetrievalFailureException` I had hit earlier was a symptom of running `deleteAll` outside any transaction, not a reason to span the connector call with one.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
3keyroman: Right — `manuallyConfirmRevoke` does not call a connector (the connector revoke already happened upstream; this just finalizes local state), so suspending the transaction was a paste from `manuallyIssueCertificate`, not an architectural intent. Dropped in commit bc924725.

src/main/java/com/czertainly/core/service/cmp/message/handler/PollReqMessageHandler.java
3keyroman: Done in commit bc924725 — switched to `@Slf4j` and dropped the now-unused `Logger` / `LoggerFactory` imports. Same change applied to `PollFeature` while I was there.

src/main/java/com/czertainly/core/service/cmp/message/handler/PollReqMessageHandler.java
3keyroman: Done in commit bc924725 — replaced with `mapRequestBodyTypeToResponseBodyType(originalBodyType)` that switches over `TYPE_INIT_REQ` → `TYPE_INIT_REP`, `TYPE_CERT_REQ` → `TYPE_CERT_REP`, `TYPE_KEY_UPDATE_REQ` → `TYPE_KEY_UPDATE_REP`, with `TYPE_CERT_REP` as the legacy-row fallback. The old `+1` arithmetic relied on BouncyCastle's PKIBody enum values being arranged so `req+1 == rep` — true today but obscure.

src/main/java/com/czertainly/core/service/scep/impl/ScepServiceImpl.java
3keyroman: Agreed and fixed in commit bc924725. The new mapping uses an explicit switch: `ISSUED → SUCCESS`; `REJECTED, FAILED, REVOKED → FAILURE`; everything else (REQUESTED, PENDING_APPROVAL, PENDING_ISSUE, PENDING_REVOKE) → PENDING. The pre-refactor behaviour preserved a fallthrough that lumped REVOKED with transitional states; the explicit switch makes the intent legible. Test renamed from `revoked_maps_to_pending` to `revoked_maps_to_failure` to match.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
3keyroman: Refactored in commit 08e36371. The three methods now read as orchestration:

- `manuallyIssueCertificate` extracts `assertCertificateInPendingIssueWithCsr`, `parseStoredCsr`, `identifyUploadedCertificateOrReject`, `pushFinalizedCertificateToAllLocations`.
- `manuallyConfirmRevoke` extracts `applyPreservedRevokeAttributes`, `destroyKeyBestEffort`.
- `cancelPendingCertificateOperation` extracts a `CancelTarget` record (per-state choices) plus `determineCancelTarget`, `assembleCancelRequest`, `invokeConnectorCancelOrThrow` (the four-bucket cancel contract — Javadoc carries the per-status table that was inline as a comment), `recordCancelFailure` / `recordCancelSoftFailure` (deduplicate the cert-event-history + log + structured-event-log triple), and `commitLocalCancel`.

Several inline comments that restated what the next line obviously did are dropped; the helper names carry the intent.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
3keyroman: You're right — real bug, predates this PR. Fixed in commit bc924725: the entry state is now captured as `final CertificateState entryState = certificate.getState();` right after the entry guard, and the catch block restores it with `certificate.setState(entryState)` instead of unconditionally setting `ISSUED`. A `PENDING_APPROVAL` cert hitting a connector failure during revoke now stays in `PENDING_APPROVAL` rather than being silently flipped.

src/main/java/com/czertainly/core/service/cmp/message/handler/PollFeature.java
3keyroman: Real race, fixed in commit 22f8fcc6 — refactored the (Cert \

src/main/java/com/czertainly/core/service/cmp/message/handler/CrmfMessageHandler.java
```diff
@@ -250,6 +265,49 @@ public PKIMessage handle(PKIMessage request, ConfigurationContext configuration)
...
+     * emit a CMP {@code pollRep} so the client knows to retry later (RFC 4210 §5.2.6).
+     */
+    private PKIMessage handleAsynchronousAcceptance(ASN1OctetString tid, PKIMessage request,
+                                                    ConfigurationContext configuration, String msgBodyType,
+                                                    ClientCertificateDataResponseDto requestedCert, CertRequest crmf) throws CmpBaseException {
+        CmpTransactionState trxState = transactionStateForRequestBodyType(tid, msgBodyType, request.getBody().getType());
+        cmpTransactionService.save(cmpTransactionService.createTransactionEntity(
+                tid.toString(),
+                configuration.getCmpProfile(),
+                requestedCert.getUuid().toString(),
```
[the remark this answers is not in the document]
3keyroman: Fixed in commit 183dd020 — `requestedCert` is a `ClientCertificateDataResponseDto` whose `getUuid()` already returns a `String`; the redundant `.toString()` was a leftover from the refactor that extracted `handleAsynchronousAcceptance`.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1375,372 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
+                    "Connector cancel rejected with HTTP " + status + "; certificate stays in " + cert.getState(),
+                    badRequest);
+            throw new ValidationException("Connector cancel rejected (" + status + "): " + badRequest.getMessage());
+        } catch (ConnectorServerException | ConnectorCommunicationException infra) {
+            recordCancelSoftFailure(cert, target,
+                    "Connector cancel call failed with infrastructure error (proceeding with local cancel): " + infra.getMessage(),
+                    "Connector cancel call failed (" + infra.getClass().getSimpleName() + "); proceeded with local cancel",
+                    "Connector cancel call failed for cert {} ({}: {}) — proceeding with local cancel",
+                    infra.getClass().getSimpleName() + ": " + infra.getMessage(), infra);
+        } catch (ConnectorException unexpected) {
```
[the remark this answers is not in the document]
3keyroman: Same recurring false positive as the previous unreachable-catch findings on this PR (now in `invokeConnectorCancelOrThrow` after the refactor in 08e36371). Not unreachable: `ConnectorException` has more subtypes than the four explicitly caught (`ConnectorClientException`, `ConnectorCommunicationException`, `ConnectorEntityNotFoundException`, `ConnectorServerException`) — notably `ConnectorProblemException`, plus any future subtype. The defensive catch handles those as soft failures rather than letting them propagate as opaque errors that would leave the cert in PENDING_*. Marking won't-fix on SonarCloud accordingly (mirroring the rationale on the prior identical findings).

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1166,4 +1387,372 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
```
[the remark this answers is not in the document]
3keyroman: Permanently silenced in commit c4db70bd by widening the defensive trailing catch from `catch (ConnectorException unexpected)` to `catch (Exception unexpected)`. The original intent (handle ConnectorException subtypes not explicitly listed above + any unchecked exception leaking from the connector client stack as a soft failure for the cancel local-state operation) is preserved; `Exception` is a supertype that covers RuntimeExceptions not handled by the four prior catches, so the static-analysis class-hierarchy reasoning that flagged this catch as redundant no longer applies. This avoids the same finding being re-emitted on every push.

ivosh (approved):

PR #1443, opened 2026-05-08
Fix crash on optional attribute with null content

3keyroman:
## Summary

`AttributeEngine.createObjectAttributeContent` calls `attributeContentItems.size()` without a null guard, but the immediately-preceding `validateAttributeContent` accepts `null` and empty content equivalently for non-required attributes (`boolean noContent = attributeContent == null || attributeContent.isEmpty();`). A `RequestAttribute` with `content == null` for an optional attribute therefore passes validation and then triggers an NPE on the next line, which surfaces as HTTP 500 with the framework-internal message `Cannot invoke \"java.util.List.size()\" because \"attributeContentItems\" is null` via the global handler.

## What changed

After `validateAttributeContent` passes, return cleanly when the content is null or empty. The two halves of `createObjectAttributeContent` now agree on what null means for non-required attributes.

## Test plan

- [x] New regression test `updateObjectDataAttributesContentAcceptsNullContentForNonRequiredAttribute` in `AttributeEngineTest` — registers a non-required v3 data attribute, sends a `RequestAttribute` with `content=null`, asserts no exception.
- [x] Full `AttributeEngineTest` passes (27 tests).

### DIFF
--- a/src/main/java/com/czertainly/core/attribute/engine/AttributeEngine.java
+++ b/src/main/java/com/czertainly/core/attribute/engine/AttributeEngine.java
@@ -1291,6 +1291,14 @@ public class AttributeEngine {
 
         validateAttributeContent(attributeDefinition, attributeContentItems);
 
+        // validateAttributeContent treats null and empty content equivalently for non-required
+        // attributes; mirror that contract here. Without this guard the iteration below NPEs on
+        // attributeContentItems.size() and surfaces as a 500 with framework-internal message,
+        // instead of a clean no-op for an optional attribute the connector left unset.
+        if (attributeContentItems == null || attributeContentItems.isEmpty()) {
+            return;
+        }
+
         for (int i = 0; i < attributeContentItems.size(); i++) {
             AttributeContent attributeContentItem = attributeContentItems.get(i);
             AttributeContentItem contentItemEntity = null;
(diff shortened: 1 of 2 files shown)


src/main/java/com/czertainly/core/attribute/engine/AttributeEngine.java
```diff
@@ -1291,6 +1291,14 @@ private void createObjectAttributeContent(AttributeDefinition attributeDefinitio
 
         validateAttributeContent(attributeDefinition, attributeContentItems);
 
+        // validateAttributeContent treats null and empty content equivalently for non-required
```
ivosh, line 1294: It is really not necessary to explain what happens without this guard.

Ideally, validateAttributeContent returns `noContent` flag from its processing, stating what it has already inferred - the attribute is optional and has no content. Based on this flag, `createObjectAttributeContent` simply returns. No need to duplicate the same logic in two methods.

src/test/java/com/czertainly/core/attribute/AttributeEngineTest.java
```diff
@@ -330,6 +330,35 @@ void testExtensibleListAttributeContentValidation() throws AttributeException {
 
     }
 
+    @Test
+    void updateObjectDataAttributesContentAcceptsNullContentForNonRequiredAttribute() throws AttributeException {
+        // Regression: a connector legitimately returning content=null for an optional attribute
```
ivosh, line 335: I don't see any reason stating why we have another test method. The method name is self-describing and the causalities are available in the linked PR.

PR #1445, opened 2026-05-11
loadResourceObjectContentData handles multi-select list resource object attribute

ivosh:
Note that this is a preliminary workaround because proper fix requires updating AttributeDefinitionUtils in interfaces first.


ivosh:
This is handled properly by #1454.

PR #1454, opened 2026-05-12
Load all values from multiselect Resource attributes

### LINKED ISSUE #1446: Attributes - Data Attributes: Resource content type multiselect saves only one value
## Description
When a data attribute is defined with a **Resource** content type and `multiselect: true` in its definition, submitting multiple selected resource values results in only a single content item being persisted. The remaining selections are silently discarded with no error or warning surfaced to the caller.

## Steps to Reproduce
1. Define a data attribute with content type **Resource** and `multiselect: true`
2. When filling in the attribute (e.g. during object creation or update), select **two or more** resource items
3. Submit the form / call the API
4. Retrieve the saved attribute values

## Expected Result
- All selected resource items are persisted as individual content entries for the attribute
- Retrieving the object returns all previously submitted values

## Actual Result
- Only one of the selected resource items is saved
- The remaining selected values are silently discarded
- Retrieving the object returns a single content entry

## Environment

ILM Core:
Commit:
Branch:
Build time:

Client:
- OS:
- Browser:

## Attachments

klaraf755:
Closes #1446 

Depends on https://github.com/OmniTrustILM/interfaces/pull/671

### DIFF
--- a/src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
+++ b/src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
@@ -257,11 +257,11 @@ public class ResourceServiceImpl implements ResourceService {
         }
 
         for (DataAttribute attribute : attributes) {
-            if (!AttributeContentType.RESOURCE.equals(attribute.getContentType())) {
+            if (!AttributeContentType.RESOURCE.equals(attribute.getContentType()) || attribute.getContent() == null || ((List<?>) attribute.getContent()).isEmpty()) {
                 continue;
             }
             List<NameAndUuidDto> resourceIds = AttributeDefinitionUtils.getNameAndUuidDataList(attribute.getName(), AttributeDefinitionUtils.getClientAttributes(attributes));
-            if (resourceIds == null || resourceIds.isEmpty()) {
+            if (resourceIds.isEmpty()) {
                 throw new AttributeException("No Resource Object UUIDs found for attribute: " + attribute.getName(), attribute.getUuid(), attribute.getName(), AttributeType.DATA, "");
             }
             List<ResourceObjectContent> contents = new ArrayList<>();
--- a/src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
+++ b/src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
@@ -261,6 +261,9 @@ public class ResourceServiceImpl implements ResourceService {
                 continue;
             }
             List<NameAndUuidDto> resourceIds = AttributeDefinitionUtils.getNameAndUuidDataList(attribute.getName(), AttributeDefinitionUtils.getClientAttributes(attributes));
+            if (resourceIds == null || resourceIds.isEmpty()) {
+                throw new AttributeException("No Resource Object UUIDs found for attribute: " + attribute.getName(), attribute.getUuid(), attribute.getName(), AttributeType.DATA, "");
+            }
             List<ResourceObjectContent> contents = new ArrayList<>();
             for (NameAndUuidDto resourceId : resourceIds) {
                 if (resourceId == null || resourceId.getUuid() == null)
--- a/src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
+++ b/src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
@@ -260,11 +260,15 @@ public class ResourceServiceImpl implements ResourceService {
             if (!AttributeContentType.RESOURCE.equals(attribute.getContentType())) {
                 continue;
             }
-            NameAndUuidDto resourceId = AttributeDefinitionUtils.getNameAndUuidData(attribute.getName(), AttributeDefinitionUtils.getClientAttributes(attributes));
-            if (resourceId == null || resourceId.getUuid() == null)
-                throw new AttributeException("UUID of Resource Object is missing.", attribute.getUuid(), attribute.getName(), AttributeType.DATA, "");
-            ResourceObjectContentData data = getResourceObjectContentData(attribute.getProperties().getResource(), UUID.fromString(resourceId.getUuid()), resourceId.getName());
-            attribute.setContent(List.of(new ResourceObjectContent(resourceId.getName(), data)));
+            List<NameAndUuidDto> resourceIds = AttributeDefinitionUtils.getNameAndUuidDataList(attribute.getName(), AttributeDefinitionUtils.getClientAttributes(attributes));
+            List<ResourceObjectContent> contents = new ArrayList<>();
+            for (NameAndUuidDto resourceId : resourceIds) {
+                if (resourceId == null || resourceId.getUuid() == null)
+                    throw new AttributeException("UUID of Resource Object is missing.", attribute.getUuid(), attribute.getName(), AttributeType.DATA, "");
+                ResourceObjectContentData data = getResourceObjectContentData(attribute.getProperties().getResource(), UUID.fromString(resourceId.getUuid()), resourceId.getName());
+                contents.add(new ResourceObjectContent(resourceId.getName(), data));
+            }
+            attribute.setContent(contents);
         }
     }
 
--- a/src/test/java/com/czertainly/core/service/ResourceServiceTest.java
+++ b/src/test/java/com/czertainly/core/service/ResourceServiceTest.java
@@ -324,14 +324,21 @@ class ResourceServiceTest extends BaseSpringBootTest {
         resourceAttribute.setName("resource");
         ResourceCertificateContentData data = new ResourceCertificateContentData();
         data.setUuid(certificate.getUuid().toString());
-        resourceAttribute.setContent(List.of(new ResourceObjectContent("ref", data)));
+        resourceAttribute.setContent(List.of(new ResourceObjectContent("ref", data), new ResourceObjectContent("ref2", data)));
         DataAttributeProperties properties = new DataAttributeProperties();
         resourceAttribute.setProperties(properties);
         properties.setResource(AttributeResource.CERTIFICATE);
 
         resourceService.loadResourceObjectContentData(List.of(nonResourceAttribute, resourceAttribute));
         Assertions.assertNull(nonResourceAttribute.getContent());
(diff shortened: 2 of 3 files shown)


src/test/java/com/czertainly/core/service/ResourceServiceTest.java
```diff
@@ -324,18 +324,25 @@ void testLoadResourceObjectContentDataFromDataAttributes() throws NotFoundExcept
         resourceAttribute.setName("resource");
         ResourceCertificateContentData data = new ResourceCertificateContentData();
         data.setUuid(certificate.getUuid().toString());
-        resourceAttribute.setContent(List.of(new ResourceObjectContent("ref", data)));
+        resourceAttribute.setContent(List.of(new ResourceObjectContent("ref", data), new ResourceObjectContent("ref2", data)));
         DataAttributeProperties properties = new DataAttributeProperties();
```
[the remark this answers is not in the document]
klaraf755: Fixed

src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
```diff
@@ -260,11 +260,15 @@ public void loadResourceObjectContentData(List<DataAttribute> attributes) throws
...
-            NameAndUuidDto resourceId = AttributeDefinitionUtils.getNameAndUuidData(attribute.getName(), AttributeDefinitionUtils.getClientAttributes(attributes));
-            if (resourceId == null || resourceId.getUuid() == null)
-                throw new AttributeException("UUID of Resource Object is missing.", attribute.getUuid(), attribute.getName(), AttributeType.DATA, "");
-            ResourceObjectContentData data = getResourceObjectContentData(attribute.getProperties().getResource(), UUID.fromString(resourceId.getUuid()), resourceId.getName());
-            attribute.setContent(List.of(new ResourceObjectContent(resourceId.getName(), data)));
+            List<NameAndUuidDto> resourceIds = AttributeDefinitionUtils.getNameAndUuidDataList(attribute.getName(), AttributeDefinitionUtils.getClientAttributes(attributes));
+            List<ResourceObjectContent> contents = new ArrayList<>();
+            for (NameAndUuidDto resourceId : resourceIds) {
+                if (resourceId == null || resourceId.getUuid() == null)
+                    throw new AttributeException("UUID of Resource Object is missing.", attribute.getUuid(), attribute.getName(), AttributeType.DATA, "");
```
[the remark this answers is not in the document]
klaraf755: Added

src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
```diff
@@ -260,11 +260,18 @@ public void loadResourceObjectContentData(List<DataAttribute> attributes) throws
             if (!AttributeContentType.RESOURCE.equals(attribute.getContentType())) {
                 continue;
             }
-            NameAndUuidDto resourceId = AttributeDefinitionUtils.getNameAndUuidData(attribute.getName(), AttributeDefinitionUtils.getClientAttributes(attributes));
-            if (resourceId == null || resourceId.getUuid() == null)
-                throw new AttributeException("UUID of Resource Object is missing.", attribute.getUuid(), attribute.getName(), AttributeType.DATA, "");
-            ResourceObjectContentData data = getResourceObjectContentData(attribute.getProperties().getResource(), UUID.fromString(resourceId.getUuid()), resourceId.getName());
-            attribute.setContent(List.of(new ResourceObjectContent(resourceId.getName(), data)));
+            List<NameAndUuidDto> resourceIds = AttributeDefinitionUtils.getNameAndUuidDataList(attribute.getName(), AttributeDefinitionUtils.getClientAttributes(attributes));
+            if (resourceIds == null || resourceIds.isEmpty()) {
```
ivosh: resourceIds cannot be null, right? getNameAndUuidDataList returns an empty list at worst. JavaDoc on getNameAndUuidDataList would clarify it.

src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
```diff
@@ -260,11 +260,18 @@ public void loadResourceObjectContentData(List<DataAttribute> attributes) throws
...
+                throw new AttributeException("No Resource Object UUIDs found for attribute: " + attribute.getName(), attribute.getUuid(), attribute.getName(), AttributeType.DATA, "");
+            }
+            List<ResourceObjectContent> contents = new ArrayList<>();
```
ivosh, line 267: Similar fix (using getNameAndUuidDataList) should be also implemented in CredentialServiceImpl. Think how to process the attributes more uniformly.

src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
```diff
@@ -260,11 +260,18 @@ public void loadResourceObjectContentData(List<DataAttribute> attributes) throws
...
+            for (NameAndUuidDto resourceId : resourceIds) {
+                if (resourceId == null || resourceId.getUuid() == null)
+                    throw new AttributeException("UUID of Resource Object is missing.", attribute.getUuid(), attribute.getName(), AttributeType.DATA, "");
```
ivosh, line 270: Should this check rather belong to AttributeDefinitionUtils?

src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
```diff
@@ -260,11 +260,18 @@ public void loadResourceObjectContentData(List<DataAttribute> attributes) throws
...
                 continue;
             }
-            NameAndUuidDto resourceId = AttributeDefinitionUtils.getNameAndUuidData(attribute.getName(), AttributeDefinitionUtils.getClientAttributes(attributes));
-            if (resourceId == null || resourceId.getUuid() == null)
-                throw new AttributeException("UUID of Resource Object is missing.", attribute.getUuid(), attribute.getName(), AttributeType.DATA, "");
-            ResourceObjectContentData data = getResourceObjectContentData(attribute.getProperties().getResource(), UUID.fromString(resourceId.getUuid()), resourceId.getName());
-            attribute.setContent(List.of(new ResourceObjectContent(resourceId.getName(), data)));
+            List<NameAndUuidDto> resourceIds = AttributeDefinitionUtils.getNameAndUuidDataList(attribute.getName(), AttributeDefinitionUtils.getClientAttributes(attributes));
+            if (resourceIds == null || resourceIds.isEmpty()) {
+                throw new AttributeException("No Resource Object UUIDs found for attribute: " + attribute.getName(), attribute.getUuid(), attribute.getName(), AttributeType.DATA, "");
```
ivosh, line 265: This now checks that there needs to be at least one resource selected for a multiselect attribute. What is the proper semantics of the resource object multiselect attribute? I am unsure and docs do not describe it here: https://docs.otilm.com/docs/contributors/attributes#attribute-properties-and-types

src/test/java/com/czertainly/core/service/ResourceServiceTest.java
```diff
@@ -314,28 +319,40 @@ void testListAllResourceEvents() {
     }
 
     @Test
+    @Transactional
     void testLoadResourceObjectContentDataFromDataAttributes() throws NotFoundException, AttributeException, ConnectorException {
         DataAttributeV3 nonResourceAttribute = new DataAttributeV3();
         nonResourceAttribute.setName("name");
         nonResourceAttribute.setContentType(AttributeContentType.DATE);
 
+        Certificate certificate1 = createCertificate();
```
ivosh: The new `certificate1` is merged with the existing one `certificate` created in setup. Why? They should be distinct, right, with distinct UUIDs. The certificates should also have different serialNumber and subjectDn. You can use existing utility classes present in the codebase today to create certificates.

ivosh (changes requested):

src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
klaraf755: Multi select attribute in general just means that the attribute content can have more than one value. It can also have zero if the attribute is not required. But this method should not concern itself with properties of the attribute - it should just add object content to the attributes that have been sent. The intent for this check is that even though attributes are provided, their name and UUID was not retrieved. Anyway, I think that there should be check that attribute has no content before all that, and just return that empty content back, I will add that.

src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
klaraf755: Not necessarily, the method in AttributeDefinition should just parse what is provided, in theory it could be used for some other purpose that does not need UUID to be present.

src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
klaraf755: I was considering to add the fix to credentials as well, but I skipped it since credentials are now outdated way of retrieving secrets (and only working with v2 attributes), the same can be achieved using resource attributes. So while the same issue can occur here and there is no check guarding against multi select credential attribute, I believe the original intent was to just have one credential content and I would leave it as it is. (Or additionally forbid multiselect property for this content type)

ivosh: Ok.

src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
klaraf755: Removed the check for null

src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
ivosh: Ok.

src/test/java/com/czertainly/core/service/ResourceServiceTest.java
klaraf755: I did not notice they use the same UUID, they were meant to have different ones. I added distinct properties to the other certificate.

ivosh (approved):

PR #1456, opened 2026-05-12
Revalidate dependent certificates when CA trust mark changes

### LINKED ISSUE #1513: Signing Profile Creation Uses Stale Certificate Validation Status
## Description

The signing profile creation check relies on the cached `validationStatus` field on the `Certificate` entity instead of performing a fresh trust-chain evaluation at signing time. When a CA certificate is marked or unmarked as trusted, no revalidation event is published for dependent certificates, so their cached status can remain stale for an arbitrary amount of time — up to a day, until the next scheduled revalidation run.

The original discovery path was through API scripting: a TSA certificate issued before its root CA was marked trusted received an INVALID cached status, which the signing profile check then rejected even after the root was trusted. There is a simple workaround for that direction (call `GET /v1/certificates/{uuid}/validate` before creating the profile).

**The more serious risk is the reverse direction:** when a CA is *unmarked* as trusted, dependent certificates keep their VALID cached status until the next scheduled revalidation. During that window, a signing profile can be created (or an existing one used) with a certificate whose chain is no longer trusted. This is a trust-enforcement gap, not just a UX inconvenience.

> **Note on scope:** This PR targets `main`, which may not yet contain the signing profile creation or the actual signing workflow. The underlying bug — stale `validationStatus` after a trust-mark change — exists independently of signing profiles. The fix can be verified in the UI on the certificate list page: re-validation should happen near-instantly after toggling the trusted CA flag. This PR does **not** need to wait for the signing profile feature to land.

---

### Location
`SigningProfileServiceImpl.java:587-592` (`applyScheme`) and `CertificateServiceImpl.java:554-563` (`updateTrustedCaMark`)

### The stale-status check
```java
// SigningProfileServiceImpl.java — eligibility check at profile creation
if (CertificateUtil.isCertificateDigitalSigningAcceptable(certificate, p.getWorkflowType(), ...)) {
    version.setCertificate(certificate);
} else {
    throw new ValidationException("Certificate " + certificate.getUuid() + " is not eligible ...");
}

// CertificateUtil.java:933-935 — reads cached field, not live chain
(!certificate.getValidationStatus().equals(CertificateValidationStatus.VALID)
        && !certificate.getValidationStatus().equals(CertificateValidationStatus.EXPIRING))
```

The `validationStatus` field is persisted in the database. If the certificate is not re-validated, the stale status is served straight from the DB on every subsequent check.

### No revalidation on trust-mark change
```java
// CertificateServiceImpl.java
private void updateTrustedCaMark(SecuredUUID uuid, Boolean trustedCa) throws NotFoundException {
    Certificate certificate = getCertificateEntity(uuid);
    certificate.setTrustedCa(trustedCa);  // ← no CertificateValidationEvent published
}
```

Because no `CertificateValidationEvent` is published, dependent certificates are never revalidated when a CA's trust mark changes. The DB status remains out of sync with reality until the next scheduled revalidation run — which can be up to a day later.

---

## Impact

| Scenario | Severity | Workaround |
|---|---|---|
| Mark CA trusted → create signing profile immediately → INVALID cached status rejects it | Medium | Call `GET /v1/certificates/{uuid}/validate` first |
| Unmark CA trusted → signing profile creation (or use) succeeds during revalidation lag → signs with untrusted cert | High | None — the stale status is VALID, so no error surfaces |

---

## Steps to Reproduce (trust-removal scenario)

1. Issue a TSA certificate from a CA whose root is already trusted.
   - Certificate validates as VALID.
2. Unmark the root CA certificate as trusted via `PATCH /v1/certificates/{root-uuid}` with `{"trustedCa": false}`.
   - Root CA is unmarked successfully.
   - No revalidation is triggered for dependent certificates.
3. Immediately create a signing profile using the TSA certificate.
   - **Expected:** HTTP 422 — certificate is no longer eligible (root untrusted).
   - **Actual:** Profile creation succeeds, using a stale VALID status.

If the signing profile workflow is not yet present in `main`, the same stale-status behavior can be verified in the UI on the certificate list page — dependent certificates should show updated validation status near-instantly after the trust mark changes.

---

## Expected Behavior

When a CA's trust mark changes (either direction), dependent certificates must be synchronously or near-synchronously revalidated before any trust-sensitive operation (signing profile creation, signing itself) can proceed with a stale result. Triggering revalidation on the trust-mark change shortens the inconsistency window from up to a day down to near-instant.

---

## Actual Behavior

Trust-mark changes are not propagated to dependent certificate validation status until the next scheduled revalidation run (up to a day later). During that window, the system enforces the wrong trust state because the stale `validationStatus` is read directly from the database.

---

## Workaround (marking-trusted direction only)

Before creating a signing profile, explicitly trigger validation:
```bash
ilm_curl GET "/v1/certificates/${cert_uuid}/validate" >/dev/null
ilm_curl POST "/v1/signingProfiles" -d '{...}'
```
---

## Severity

High (trust enforcement gap — can sign with an untrusted certificate chain)

## Module

Certificates / Signing Profiles

LukasNajman:


src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -548,6 +561,34 @@ private void updateTrustedCaMark(SecuredUUID uuid, Boolean trustedCa) throws Not
             throw new ValidationException("Trying to mark certificate as trusted CA when certificate is not CA.");
         }
```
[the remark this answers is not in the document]
LukasNajman: fixed

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -548,6 +561,34 @@ private void updateTrustedCaMark(SecuredUUID uuid, Boolean trustedCa) throws Not
...
+        certificateRepository.findAllDescendantCertificatesEligibleForValidation(ca.getUuid(), platformEnabled)
+                .stream()
+                .map(UUID::fromString)
+                .forEach(toRevalidate::add);
+        if (!toRevalidate.isEmpty()) {
+            log.debug(applicationEventPublisher.getClass().getName());
+            applicationEventPublisher.publishEvent(new CertificateValidationEvent(toRevalidate));
+        }
+    }
+
```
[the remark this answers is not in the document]
LukasNajman: implemented

src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java
```diff
@@ -266,6 +267,56 @@ AND NOT (c.uuid = ANY(chain.path))
...
+            INNER JOIN {h-schema}certificate c ON c.uuid = s.uuid
+            LEFT JOIN {h-schema}ra_profile rp ON rp.uuid = c.ra_profile_uuid
+            WHERE c.archived = false
+              AND c.certificate_content_id IS NOT NULL
+              AND c.validation_status NOT IN ('REVOKED', 'EXPIRED')
+              AND (
+                  (rp.validation_enabled IS NULL AND :platformEnabled = true)
+                  OR rp.validation_enabled = true
+              )
+            """, nativeQuery = true)
```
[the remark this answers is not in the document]
LukasNajman: This is ok, the case you have described would be in case the data in db are corrupted, not a normal case.

src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java
```diff
@@ -266,6 +267,56 @@ AND NOT (c.uuid = ANY(chain.path))
...
+                SELECT uuid, ARRAY[uuid] AS path, 0 AS depth
+                FROM {h-schema}certificate
+                WHERE issuer_certificate_uuid = :caUuid
+                UNION ALL
+                SELECT c.uuid, subtree.path || c.uuid, subtree.depth + 1
+                FROM {h-schema}certificate c
+                INNER JOIN subtree ON c.issuer_certificate_uuid = subtree.uuid
+                WHERE subtree.depth < 20
+                  AND NOT (c.uuid = ANY(subtree.path))
+            )
```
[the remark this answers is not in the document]
LukasNajman: fixed

src/main/java/com/czertainly/core/service/handler/CertificateHandler.java
```diff
@@ -39,6 +39,7 @@
 public class CertificateHandler {
 
     private static final Logger logger = LoggerFactory.getLogger(CertificateHandler.class);
+    private static final int VALIDATION_BATCH_SIZE = 10;
```
ivosh: This value shall come from the application properties. We may have some pre-existing batch sizes already - check if they can be reused.

Actually the same constant is defined also in CertificateServiceImpl - unify it.

src/test/java/com/czertainly/core/dao/repository/FindEligibleDescendantCertificatesTest.java
```diff
@@ -0,0 +1,350 @@
...
+
+    @Autowired
+    private CertificateRepository certificateRepository;
+    @Autowired
+    private CertificateContentRepository certificateContentRepository;
+    @Autowired
+    private RaProfileRepository raProfileRepository;
+
+    private Certificate buildCert(Certificate issuer, boolean hasContent,
+                                   CertificateValidationStatus status, boolean archived,
```
ivosh: Check formatting. Better - reformat the whole file.

src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java
```diff
@@ -266,6 +267,58 @@ AND NOT (c.uuid = ANY(chain.path))
...
+    /**
+     * Returns the UUIDs of all certificates issued directly or transitively by the given CA that are eligible for
+     * validation, in a single recursive CTE query.
+     *
+     * <p>Eligibility filters applied:</p>
+     * <ul>
+     *   <li>{@code archived = false} — archived certificates are excluded from all automated validation flows.</li>
+     *   <li>{@code certificate_content_id IS NOT NULL} — cannot validate without content.</li>
+     *   <li>{@code validation_status NOT IN ('REVOKED', 'EXPIRED')} — terminal statuses; CA trust cannot change them.</li>
+     *   <li>{@code (rp.validation_enabled IS NULL AND platformEnabled) OR rp.validation_enabled = true} — mirrors
```
ivosh, line 395: If this logic is copied from another place, could you centralize it so it is not repeated?

src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java
```diff
@@ -266,6 +267,58 @@ AND NOT (c.uuid = ANY(chain.path))
...
+     * <p>The CA certificate itself is not included in the result; callers are expected to add it separately.</p>
+     *
+     * <p>Traversal is capped at {@code maxDepth} levels and uses a path array to break cycles in corrupt data
+     * (e.g. a self-signed root whose {@code issuer_certificate_uuid} points back to itself).</p>
+     *
+     * @param caUuid          UUID of the issuing CA certificate
+     * @param platformEnabled value of the platform-level certificate validation {@code enabled} flag;
+     *                        applied to certificates whose RA profile has no explicit {@code validation_enabled} override
+     * @param maxDepth        maximum number of hops to follow (safety cap against circular references)
+     * @return unordered list of UUID strings for all eligible descendants
```
ivosh: The method actually returns an unordered set.
I'd say callers would expect some order, such as by depth. So I'd suggest to return `List<UUID>` and order the result with `ORDER BY s.depth`.

src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java
```diff
@@ -266,6 +267,58 @@ AND NOT (c.uuid = ANY(chain.path))
...
+            LEFT JOIN {h-schema}ra_profile rp ON rp.uuid = c.ra_profile_uuid
+            WHERE c.archived = false
+              AND c.certificate_content_id IS NOT NULL
+              AND c.validation_status NOT IN ('REVOKED', 'EXPIRED')
+              AND (
+                  (rp.validation_enabled IS NULL AND :platformEnabled = true)
+                  OR rp.validation_enabled = true
+              )
+            """, nativeQuery = true)
+    Set<String> findAllDescendantCertificatesEligibleForValidation(@Param("caUuid") UUID caUuid,
```
ivosh: UUIDs are returned, so if possible, use `UUID` type.

src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java
```diff
@@ -266,6 +267,58 @@ AND NOT (c.uuid = ANY(chain.path))
...
+    @Query(value = """
+            WITH RECURSIVE subtree AS (
+                SELECT uuid, ARRAY[uuid] AS path, 0 AS depth
+                FROM {h-schema}certificate
+                WHERE issuer_certificate_uuid = :caUuid
+                UNION ALL
+                SELECT c.uuid, subtree.path || c.uuid, subtree.depth + 1
+                FROM {h-schema}certificate c
+                INNER JOIN subtree ON c.issuer_certificate_uuid = subtree.uuid
+                WHERE subtree.depth < :maxDepth
```
ivosh, line 424: PostgreSQL 14+ supports a cleaner native cycle-detection syntax via the CYCLE clause. We target PostgreSQL 17 these days (refer to pom.xml under `test`).

src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java
```diff
@@ -266,6 +267,58 @@ AND NOT (c.uuid = ANY(chain.path))
...
+                  AND NOT (c.uuid = ANY(subtree.path))
+            )
+            SELECT s.uuid::text FROM subtree s
+            INNER JOIN {h-schema}certificate c ON c.uuid = s.uuid
+            LEFT JOIN {h-schema}ra_profile rp ON rp.uuid = c.ra_profile_uuid
+            WHERE c.archived = false
+              AND c.certificate_content_id IS NOT NULL
+              AND c.validation_status NOT IN ('REVOKED', 'EXPIRED')
```
ivosh: Do not hard-code these enum values - pass them via a parameter, like the other queries in this repository.

src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java
```diff
@@ -266,6 +267,58 @@ AND NOT (c.uuid = ANY(chain.path))
...
 
+    /**
+     * Returns the UUIDs of all certificates issued directly or transitively by the given CA that are eligible for
+     * validation, in a single recursive CTE query.
+     *
+     * <p>Eligibility filters applied:</p>
+     * <ul>
+     *   <li>{@code archived = false} — archived certificates are excluded from all automated validation flows.</li>
+     *   <li>{@code certificate_content_id IS NOT NULL} — cannot validate without content.</li>
+     *   <li>{@code validation_status NOT IN ('REVOKED', 'EXPIRED')} — terminal statuses; CA trust cannot change them.</li>
```
ivosh, line 394: Referring to https://docs.otilm.com/docs/certificate-key/concept-design/core-components/certificate#certificate-validation-status, I can see three terminal statuses, including Failed.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -539,6 +540,18 @@ public void updateCertificateObjects(SecuredUUID uuid, CertificateUpdateObjectsD
         }
     }
 
+    /**
+     * Sets the trusted-CA mark on the given certificate and immediately triggers revalidation for the CA and all
+     * eligible descendants.
+     *
+     * <p>The {@link CertificateValidationEvent} is published via Spring so that
```
ivosh, line 583: Why repeat what the method actually does? Well-written code is self-describing. I'd say this paragraph is redundant.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -547,7 +560,38 @@ private void updateTrustedCaMark(SecuredUUID uuid, Boolean trustedCa) throws Not
...
+                .stream()
+                .map(UUID::fromString)
+                .forEach(toRevalidate::add);
+        if (!toRevalidate.isEmpty()) {
+            log.debug("Publishing certificate validation event for CA subtree revalidation. caUuid={}, certificateCount={}", ca.getUuid(), toRevalidate.size());
+            applicationEventPublisher.publishEvent(new CertificateValidationEvent(toRevalidate));
+        }
+    }
+
+    private boolean isEligibleForRevalidation(Certificate certificate, boolean platformEnabled) {
```
ivosh, line 626: Should this method behave like findAllDescendantCertificatesEligibleForValidation? If yes, then it is missing check for `archived` and `certificate_content`. 

Regardless, I'd like to see a parameterized par-to-par test similar to certificate eligibility tests from CertificateUtil: https://github.com/OmniTrustILM/core/blob/main/src/test/java/com/czertainly/core/util/CertificateUtilTest.java#L97 or https://github.com/OmniTrustILM/core/blob/main/src/test/java/com/czertainly/core/util/CertificateUtilTest.java#L139

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -123,6 +123,7 @@
 
 @Service(Resource.Codes.CERTIFICATE)
 @Transactional
+@Slf4j
```
ivosh, line 129: @Slf4j adds `log` (good!) but there is an existing `logger` - delete it and use solely `log`.

ivosh (changes requested):
The linked issue frames this PR within the context of a signing profile. However there is no signing profile in `main`, yet. How does the changes in this PR affect the existing functionality? Should this PR land only after signing profile materializes?

Also, the linked issue talks about stale cached validationStatus. How does this PR address that issue?

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
LukasNajman: fixed

LukasNajman:
> The linked issue frames this PR within the context of a signing profile. However there is no signing profile in `main`, yet. How does the changes in this PR affect the existing functionality? Should this PR land only after signing profile materializes?
> 
> Also, the linked issue talks about stale cached validationStatus. How does this PR address that issue?

Actually, the issue is aware of the problem that signing may not be in the main branch before merging this. The behavior can be verified also in UI, on the certificate list page. There is a note about in in the Steps to reproduce section.

```Note: The PR is done to `main` branch that may not yet contain the signing profile creation or the actual signing workflow. In that case, we can verify in UI - in certificate list, that re-validation happens near instantly.```

About the cached validation status - the status is persisted in the database and if the certificate is not re validated, the stale status is served from database. That is why we trigger re-validation on change of trusted certificate flag. This shortens the time the status in db is not accurate. Without it, it could take up to 1 day before the status is aligned with reality.

**Comment by LukasNajman** [comment]
I have updated the bug description once again, hopefully it is more clear now.

File: src/main/java/com/czertainly/core/service/handler/CertificateHandler.java (hunk 5 above)
Comment (LukasNajman) [reply src/main/java/com/czertainly/core/service/handler/CertificateHandler.java]: I have unified it and added it to application properties. Should we make it configurable from helm chart? Seems like overkill to me. I doubt slightly whether making it configurable is really a necessity.

File: src/test/java/com/czertainly/core/dao/repository/FindEligibleDescendantCertificatesTest.java (hunk 6 above)
Comment (LukasNajman) [reply src/test/java/com/czertainly/core/dao/repository/FindEligibleDescendantCertificatesTest.java]: Reformatted

File: src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java (hunk 12 above)
Comment (LukasNajman) [reply src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java:394]: How I see it, Failed is not a terminal status, it just indicated that validation has failed. It may or may not succeed later. It the same logic as in `findCertificatesToCheckStatus`.

File: src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java (hunk 11 above)
Comment (LukasNajman) [reply src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java]: What would be a benefit in parametrizing this? This should one of method for a concrete usecase. We don't have any callers what would benefit from parametrizing it.

File: src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java (hunk 8 above)
Comment (LukasNajman) [reply src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java]: I changed the comment to explicitly mention the set. I prefer the Set semantic here, as it clearly implies that each certificate will be present exactly once. This is important as later we trigger a rather expensive operation (revalidation) for each of them.

File: src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java (hunk 9 above)
Comment (LukasNajman) [reply src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java]: Changed to UUID

File: src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java (hunk 13 above)
Comment (LukasNajman) [reply src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java:583]: I gives the reader a quick context for the next paragraph that describes the intricacies of the publish mechanism in respect to DB transaction.

File: src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java (hunk 15 above)
Comment (LukasNajman) [reply src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java:129]: Removed the explicit logger.

File: src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java (hunk 14 above)
Comment (LukasNajman) [reply src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java:626]: The check is only for the rootCA that us being modified and the archived status was carried about by calling method `updateTrustedCaMark`. The content check was missing.

I have added both so it is consistent with the repository method, though it may be duplicit and can not be tested.

File: src/main/java/com/czertainly/core/service/handler/CertificateHandler.java (hunk 5 above)
Comment (ivosh) [reply src/main/java/com/czertainly/core/service/handler/CertificateHandler.java]: application properties is fine

File: src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java (hunk 8 above)
Comment (ivosh) [reply src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java]: perfect

File: src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java (hunk 11 above)
Comment (ivosh) [reply src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java]: For two reasons:
* Because you do not hard-code codes from the enumeration values.
* Because there is already a precedens how to do it in the codebase. Check CertificateServiceImpl.updateCertificatesStatusScheduled and CertificateRepository.countCertificatesToCheckStatus (skipStatuses).

File: src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java (hunk 7 above)
Comment (LukasNajman) [reply src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java:395]: The logic is similar to the one in `findCertificatesToCheckStatus` but not the same. It is not easy to unify that. I would leave it separate.

File: src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java (hunk 12 above)
Comment (ivosh) [reply src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java:394]: Looking at https://docs.otilm.com/docs/certificate-key/concept-design/core-components/certificate#certificate-validation-status, I can see there is a transition from the `Failed` state directly to the terminal state. Is the documentation wrong, then?

File: src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java (hunk 7 above)
Comment (ivosh) [reply src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java:395]: ok, fine, you've checked and there are tests which cover the logic (I hope).

File: src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java (hunk 13 above)
Comment (ivosh) [reply src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java:583]: ok, fine

File: src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java (hunk 14 above)
Comment (ivosh) [reply src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java:626]: Are you going to provide par-to-par parameterized test, similar to certificate eligibility tests from CertificateUtil?

Comment (LukasNajman) [reply src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java:626]: Yes, the tests are there. A have added one for missing content. The one for archived flag can't reach the actual check - the method `updateTrustedCaMark` throws an exception in this case. The test for the exception throw is present.

File: src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java (hunk 12 above)
Comment (LukasNajman) [reply src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java:394]: We agreed on Failed not being a terminal status

File: src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java (hunk 11 above)
Comment (LukasNajman) [reply src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java]: Replaced the static strings with SpEL expression.

File: src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java (hunk 10 above)
Comment (LukasNajman) [reply src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java:424]: Updated to CYCLE clause

**Review by lubomirw — COMMENTED** [review]
Code review: 2 issue(s) found. Both advisory (non-blocking). Title also tidied to drop the `fix:` prefix per repo convention.

File: src/main/java/com/czertainly/core/service/handler/CertificateHandler.java [hunk 16]
```diff
@@ -40,6 +41,9 @@ public class CertificateHandler {
 
     private static final Logger logger = LoggerFactory.getLogger(CertificateHandler.class);
 
+    @Value("${certificate.validation.batch-size:10}")
+    private int validationBatchSize;
```
lubomirw, line 43: **⚠️ Major** | Correctness

The `@Value` here is on the field, so Spring sets `validationBatchSize` directly and `setValidationBatchSize(...)` never runs at startup — its `<= 0` guard only fires from tests. So if someone sets `certificate.validation.batch-size: 0` in config, nothing stops it, and we end up in `for (int i = 0; i < size; i += validationBatchSize)` with `i` stuck at 0 — an infinite loop pushing empty messages onto the queue. A negative value blows up differently, with an `IllegalArgumentException` from `subList`. Either way it's silent until the first validation event fires, not at boot. Same thing in `CertificateServiceImpl.sendValidationBatches`.

Simplest fix: move the `@Value` annotation onto `setValidationBatchSize` so the guard you already wrote actually runs during injection (the tests call the setter already, so they're unchanged). A bad config then fails fast at startup instead of hanging later.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -547,15 +559,46 @@ private void updateTrustedCaMark(SecuredUUID uuid, Boolean trustedCa) throws Not
...
+            toRevalidate.add(ca.getUuid());
+        }
+        toRevalidate.addAll(certificateRepository.findAllDescendantCertificatesEligibleForValidation(ca.getUuid(), platformEnabled, certificateChainMaxDepth));
+        if (!toRevalidate.isEmpty()) {
+            log.debug("Publishing certificate validation event for CA subtree revalidation. caUuid={}, certificateCount={}", ca.getUuid(), toRevalidate.size());
+            applicationEventPublisher.publishEvent(new CertificateValidationEvent(toRevalidate));
+        }
+    }
+
+    private boolean isEligibleForRevalidation(Certificate certificate, boolean platformEnabled) {
```
lubomirw, line 626: **✳️ Minor** | Maintainability

Eligibility is defined twice now: this Java method handles the CA itself, while the same four rules (archived, content present, not REVOKED/EXPIRED, RA-profile-vs-platform flag) live in the SQL `WHERE` of `findAllDescendantCertificatesEligibleForValidation` for the descendants. They agree today, but a future change to one rule has to remember both, and a mismatch would silently judge the CA differently from its subtree. Cheapest guard without reworking anything: add a cross-reference comment on each pointing at the other so they read as a matched pair. (If you'd rather kill the duplication outright, the CA could be folded into the same query by anchoring the CTE at the CA node — but that's a larger change and a follow-up is fine.)

src/main/java/com/czertainly/core/service/handler/CertificateHandler.java
LukasNajman: Moved to setter.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
LukasNajman: Added comment on both sides to emphasize, the two methods must stay in sync.

lubomirw (approved):

ivosh (approved):

PR #1461, opened 2026-05-14
Use factory methods from BulkActionMessageDto

### LINKED ISSUE #222: Don't leak runtime details to the wire via exception messages
### Description

# Introduction
`CLAUDE.md` now states:

Never forward raw `Exception.getMessage()` to a client / protocol response / external API. Runtime exceptions can carry SQL fragments (`DataIntegrityViolationException`), stack-frame class names,
internal table or column identifiers, or upstream error detail that should not be exposed.

Gate exposure on a controlled domain-exception type:

```java
String safeText = e instanceof OurDomainException && e.getMessage() != null
        ? e.getMessage()           // shaped by us, safe to expose
        : "generic safe placeholder";
```

This applies to PKI/CMP/SCEP `PKIFreeText`, ACME `error.detail`, REST API error responses, and any other surface where the message reaches an external party.

--------

# Problem statement

For checking `e instanceof OurDomainException`, we currently do not have any vehicle to do that. All czertainly-specific exceptions inherit directly from `Exception` or `RuntimeException`.

# Plan

1. Add a marker interface `PlatformException` so we can do `e instanceof PlatformException`.
2. Add ArchUnit rules to check that `Exception.getMessage()` does not leak runtime details to the wire.
3. Fix problematic places.

### Definition of Done

- [x] `PlatformException` available
- [x] `ArchUnit` rules in place
- [x] Problematic places fixed.

ivosh:


src/main/java/com/czertainly/core/service/impl/CbomServiceImpl.java
```diff
@@ -286,13 +286,13 @@ public List<BulkActionMessageDto> bulkDeleteCbom(List<UUID> uuids) {
...
             if (!existingUuids.contains(uuid)) {
-                messages.add(new BulkActionMessageDto(uuid.toString(), "", "CBOM entry not found"));
+                messages.add(BulkActionMessageDto.failureWithMessage(uuid.toString(), "", "CBOM entry not found"));
                 continue;
             }
             try {
                 transactionHandler.runInNewTransaction(() -> cbomRepository.deleteById(uuid));
             } catch (Exception ex) {
-                messages.add(new BulkActionMessageDto(uuid.toString(), "", "Error deleting CBOM entry: %s".formatted(ex.getMessage())));
+                messages.add(BulkActionMessageDto.failure(uuid.toString(), "", ex, "Delete failed"));
```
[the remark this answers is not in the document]
ivosh: Fixed.

src/test/java/com/czertainly/core/architecture/PlatformExceptionTest.java
```diff
@@ -0,0 +1,48 @@
...
+                            "so safeMessage() can gate message exposure at wire boundaries");
+
+    @ArchTest
+    static final ArchRule bulkActionMessageDtoMustUseFactory =
+            noClasses()
+                    .that().resideInAPackage("com.czertainly.core..")
+                    .should().callConstructor(BulkActionMessageDto.class, String.class, String.class, String.class)
+                    .because("BulkActionMessageDto must be created via BulkActionMessageDto.failure() " +
+                            "so message content is controlled at the factory level; " +
+                            "direct constructor calls bypass that gate");
```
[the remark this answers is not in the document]
ivosh: Fixed.

src/test/java/com/czertainly/core/architecture/PlatformExceptionTest.java
```diff
@@ -0,0 +1,55 @@
...
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.noClasses;
+
+/**
+ * Structural safety rules for exception message handling.
+ * <p>
+ * Rule 1: All concrete Throwable subclasses in com.czertainly.core must implement PlatformException so safeMessage()
+ * can gate message exposure at wire boundaries. A similar rule is enforced for exceptions declared in interfaces.
+ * com.czertainly.core.intune is excluded: those exceptions wrap third-party Microsoft API responses whose messages
+ * are not safe to expose via safeMessage().
```
LukasNajman: So, why is it excluded when it is not safe to expose them to the user?

src/test/java/com/czertainly/core/architecture/PlatformExceptionTest.java
```diff
@@ -0,0 +1,55 @@
...
+            noClasses()
+                    .that().resideInAPackage("com.czertainly.core..")
+                    .should(callCodeUnitWhere(DescribedPredicate.describe(
+                            "call a constructor of BulkActionMessageDto",
+                            (JavaCall<?> call) -> BulkActionMessageDto.class.getName().equals(
+                                    call.getTarget().getOwner().getName())
+                                    && "<init>".equals(call.getTarget().getName()))))
+                    .because("BulkActionMessageDto must be created via BulkActionMessageDto.failure() " +
+                            "so message content is controlled at the factory level; " +
+                            "direct constructor calls bypass that gate");
```
LukasNajman, line 63: This also could have been enforced by making the constructor of the BulkActionMessageDto private. I would prefer that as it would guard against using it in the first place, instead of waiting for the test to fail.

LukasNajman (dismissed):

src/test/java/com/czertainly/core/architecture/PlatformExceptionTest.java
ivosh: Nice idea. Let's do it in steps - I'll prepare a follow-up PR once this one is merged and there are no consumers of BulkMessageDto public constructors.

src/test/java/com/czertainly/core/architecture/PlatformExceptionTest.java
ivosh: That's a very good question. I expanded the explanation and added a new rule.

LukasNajman (approved):

PR #1462, opened 2026-05-14
Disable session lookup and creation for TSP protocol requests

### LINKED ISSUE #1463: Disable session lookup and creation for Timestamping Protocol
### Description

Disable sessions for Timestamping Protocol (TSP). It adds an additional database overhead.
TSP is a stateless binary protocol that does not use HTTP sessions.

### User Story

_No response_

### Use Case

_No response_

### Acceptance Criteria

- [] Using TSP endpoints for signing does not establish an HTTP session.
- [] No session cookie is sent for TSP endpoints.
- [] Existing non-TSP endpoints are not affected - they still establish a session.

ivosh:


src/main/java/com/czertainly/core/config/SessionConfig.java
```diff
@@ -92,4 +98,40 @@ SessionRepositoryCustomizer<JdbcIndexedSessionRepository> customizer() {
         };
     }
 
+    /**
+     * Skips session lookup and creation for TSP protocol requests.
+     * TSP is a stateless binary protocol that does not use HTTP sessions.
+     */
+    @Bean
+    public HttpSessionIdResolver httpSessionIdResolver() {
+        CookieHttpSessionIdResolver delegate = new CookieHttpSessionIdResolver();
```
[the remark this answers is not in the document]
ivosh: Fixed: injected the configured `CookieSerializer` bean as a parameter and called `delegate.setCookieSerializer(cookieSerializer)`; also switched `isTspRequest` from `getRequestURI()` to `getServletPath()` so the check is context-path-agnostic and works correctly in production under the `/api` context path.

src/test/java/com/czertainly/core/config/SessionConfigTest.java
```diff
@@ -0,0 +1,153 @@
...
+    @Autowired
+    JdbcTemplate jdbcTemplate;
+
+    @MockitoBean
+    CzertainlyAuthenticationClient authenticationClient;
+
+    @DynamicPropertySource
+    static void overrideProperties(DynamicPropertyRegistry registry) {
+        // Use empty context path so the URL constants above are unambiguous
+        registry.add("server.servlet.context-path", () -> "");
```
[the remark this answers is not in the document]
ivosh: Fixed: switched production code to `getServletPath()` (context-path-agnostic); updated test cookie name from `"SESSION"` to `CookieConfig.COOKIE_NAME` so the resolver can actually parse it; updated context-path-override comment to document why the empty override is still valid; added `tspRequest_expireSession_doesNotClearCookie` and `nonTspRequest_expireSession_clearsCookie` tests to cover the previously missing `expireSession` branches and close the SonarQube coverage gap.

src/main/java/com/czertainly/core/config/SessionConfig.java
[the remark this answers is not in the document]
ivosh: Fixed.

src/test/java/com/czertainly/core/config/SessionConfigTest.java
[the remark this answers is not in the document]
ivosh: Overriding the context path is necessary to correctly simulate production environment which uses `request.getServletPath()` in `SessionConfig#isTspRequest`.

src/test/java/com/czertainly/core/config/SessionConfigTest.java
```diff
@@ -0,0 +1,194 @@
...
+        String sessionId = UUID.randomUUID().toString();
+        var session = sessionRepository.createSession();
+        sessionRepository.save(session);
+
+        var result = mvc.perform(
+                post(NON_TSP_URL)
+                        .contentType(MediaType.APPLICATION_JSON)
+                        .content("{}")
+                        .cookie(sessionCookie(sessionId)))
+                .andReturn();
```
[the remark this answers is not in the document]
ivosh: Fixed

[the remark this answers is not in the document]
ivosh: Fixed.

src/main/java/com/czertainly/core/config/SessionConfig.java
```diff
@@ -92,4 +99,43 @@ SessionRepositoryCustomizer<JdbcIndexedSessionRepository> customizer() {
...
+    /**
+     * Skips session lookup and creation for TSP protocol requests.
+     * TSP is a stateless binary protocol that does not use HTTP sessions.
+     */
+    @Bean
+    public HttpSessionIdResolver httpSessionIdResolver(CookieSerializer cookieSerializer) {
+        CookieHttpSessionIdResolver delegate = new CookieHttpSessionIdResolver();
+        delegate.setCookieSerializer(cookieSerializer);
+
+        return new HttpSessionIdResolver() {
```
lubomirw, line 111: **! Major** | Correctness

Is the prefix match intentional for all paths beginning with `tsp`?

`getServletPath().startsWith("/v1/protocols/tsp")` matches not only `/v1/protocols/tsp/{profile}` but also any sibling path that happens to start with the literal `tsp` — e.g. a future `/v1/protocols/tspAdmin` or `/v1/protocols/tsp-management`. Note that `tspProfiles` is already used as a cache name in `CacheConfig`, so a `/v1/protocols/tspProfiles` admin endpoint is plausible.

If the intent is "any route under the TSP protocol family", the current check is correct and worth a comment to that effect. If the intent is strictly the protocol endpoint, a stricter check would be safer: `equals("/v1/protocols/tsp") || startsWith("/v1/protocols/tsp/")`.

lubomirw, line 111: **. Nitpick** | Maintainability

Forward-looking note (not a request for this PR): the other protocols expose their base path as a constant next to the protocol code — `AcmeConstants.ACME_URI_HEADER`, `CmpConstants.CMP_BASE_CONTEXT`, `ScepServiceImpl.SCEP_URL_PREFIX`. When the TSP protocol endpoint actually lands and there's a second reference point for `/v1/protocols/tsp`, extracting it into a `TspConstants` class and referencing it from here would keep the session-suppression rule in lockstep with the endpoint mapping. Today there's only one user of the literal, so inlining is fine.

src/test/java/com/czertainly/core/config/SessionConfigTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+        assertNull(result.getResponse().getHeader("Set-Cookie"),
+                "TSP expireSession must not produce a Set-Cookie header");
+    }
+
+    /**
+     * expireSession for a non-TSP request must clear the cookie (delegate must be called).
+     */
+    @Test
+    void nonTspRequest_expireSession_clearsCookie() throws Exception {
```
lubomirw: **~ Minor** | Testing

The test name claims `expireSession ... clearsCookie`, but the assertion only verifies that `sessionRepository.findById` was called at least once — which is already covered by `nonTspRequest_resolvesIncomingSessionCookie` above. The created `session` (and its persisted state) is never associated with the request, and the inline comment acknowledges the assertion is really "the resolver delegated", not "cookie was cleared".

Two follow-ups worth considering:

1. If the intent is to verify the delegate's `expireSession` actually runs, capture the resolver bean and spy on it (or assert a `Max-Age=0`/expired Set-Cookie header on the response).
2. If the intent is just "non-TSP path delegates", this test is a duplicate of `nonTspRequest_resolvesIncomingSessionCookie` and the unit test (`SessionConfigUnitTest`) already covers the no-op branches directly. Consider removing the redundancy.

src/test/java/com/czertainly/core/config/SessionConfigTest.java
```diff
@@ -0,0 +1,194 @@
...
+import java.security.Security;
+import java.util.Base64;
+import java.util.List;
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+
+/**
+ * Verifies that {@link SessionConfig#httpSessionIdResolver()} correctly suppresses session handling for TSP protocol
```
lubomirw: **. Nitpick** | Documentation

The `{@link SessionConfig#httpSessionIdResolver()}` reference will not resolve — the actual method signature is `httpSessionIdResolver(CookieSerializer)`. Javadoc requires the parameter types in the link target: `{@link SessionConfig#httpSessionIdResolver(org.springframework.session.web.http.CookieSerializer)}`. Same typo also appears at the top of `SessionConfigUnitTest.java`.

lubomirw (commented):

src/main/java/com/czertainly/core/config/SessionConfig.java
ivosh: Fixed.

src/test/java/com/czertainly/core/config/SessionConfigTest.java
ivosh: Fixed.

src/main/java/com/czertainly/core/config/SessionConfig.java
ivosh: Noted.

src/test/java/com/czertainly/core/config/SessionConfigTest.java
ivosh: Fixed. Test removed.

lubomirw (approved):

PR #1464, opened 2026-05-14
Implement handling of Certificate Uploaded Event

### LINKED ISSUE #1441: Implement handling of Certificate Uploaded Event
### Description

When user manually uploads a certificate, `Certificate Uploaded` event will be fired, which will decide based on triggers set in platform settings, whether the certificate can be uploaded and if so, perform actions defined in following triggers.

## Setting Service
When updating event settings, validate that conditions on triggers assigned to `Certificate Uploaded` only reference fields that are available at upload time — i.e. direct fields on the `Certificate` entity that are derivable from the X.509 content without any DB join.

The rule: a condition item is valid for this event only if its `FilterField` has `fieldResource == null && joinAttributes.isEmpty()`. Fields with a non-null `fieldResource` (RA profile, group, owner, key, location, protocol profiles) or non-empty `joinAttributes` (protocol association, successor/predecessor relations) require DB relations that do not exist at upload time and must be rejected at configuration time.

## Certificate Service
The method for uploading certificate should be split to accommodate both synchronous and asynchronous certificate upload. In both cases, service should check for existence of certificate by fingerprint, and create certificate model without any DB persistence (do not save certificate, content or key). If method is called synchronously, service should then send message to event producer containing the model, custom attributes from the request and certificate content. In the other case, the service should call the certificate uploaded event handler directly.

Uploading the certificate is also done when adding user — upload there should call the synchronous certificate upload.

## Certificate Uploaded Event
Handle event method steps:
1. Retrieve platform settings for the event
2. Evaluate ignore triggers against the transient Certificate model → if there is a match, do not upload certificate and return
3. Save certificate in DB and upload keys
4. Evaluate rest of the triggers
5. Set custom attributes from the upload request (if there is a conflict with platform settings, request should take precedence over settings)
7. Add certificate event history for Upload and publish event for validation
Note: Also save event history here, so in `Certificate Uploaded` event history user can see ignored certificates — in that case, "ignore" trigger history message should contain information about certificate (information in `CertificateEventData`) that was not uploaded. So a message for ignore trigger should be sent to trigger evaluator in case of ignore triggers.

### User Story

_No response_

### Use Case

_No response_

### Acceptance Criteria

- `Certificate Uploaded` event can be configured in settings
- only conditions on direct certificate content fields (`FilterField` with `fieldResource == null && joinAttributes.isEmpty()`) can be assigned to this event; other fields are rejected at configuration time
- certificate is uploaded based on event settings and further actions in settings are performed
- request custom attributes take precedence over attributes set by action triggers
- uploading a certificate when adding a user also goes through the event
- rejected certificates (ignore trigger match) are visible in event history with certificate info; user is created without certificate association

klaraf755:
Closes #1441


src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1202,9 +1205,9 @@ public CertificateContent checkAddCertificateContent(String fingerprint, String
...
     }
 
-    @Override
     @ExternalAuthorization(resource = Resource.CERTIFICATE, action = ResourceAction.CREATE)
-    public CertificateDetailDto upload(UploadCertificateRequestDto request, boolean ignoreCustomAttributes) throws CertificateException, NoSuchAlgorithmException, AlreadyExistException, NotFoundException, AttributeException {
+    @Override
+    public void upload(UploadCertificateRequestDto request, boolean ignoreCustomAttributes, UUID userUuid) throws CertificateException, NoSuchAlgorithmException, AlreadyExistException, NotFoundException, AttributeException {
         X509Certificate certificate = CertificateUtil.parseUploadedCertificateContent(request.getCertificate());
         String fingerprint = CertificateUtil.getThumbprint(certificate);
         if (certificateRepository.findByFingerprint(fingerprint).isPresent()) {
```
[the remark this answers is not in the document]
klaraf755: Added

src/main/java/com/czertainly/core/events/handlers/CertificateUploadedEventHandler.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    public static EventMessage constructEventMessage(CertificateUploadEventMessageData data) {
+        return new EventMessage(ResourceEvent.CERTIFICATE_UPLOADED, Resource.CERTIFICATE, null, data);
+    }
+
+    @Override
+    protected Object getEventData(Certificate object, Object eventMessageData) {
+        CertificateUploadEventMessageData uploadEventMessageData = objectMapper.convertValue(eventMessageData, CertificateUploadEventMessageData.class);
+        return EventDataBuilder.getCertificateUploadedEventData(object, uploadEventMessageData.userUuid());
```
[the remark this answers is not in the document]
klaraf755: I do not think it is important to include also custom attributes in notifications

src/main/java/com/czertainly/core/service/impl/TriggerServiceImpl.java
```diff
@@ -230,6 +231,12 @@ public void createTriggerAssociations(ResourceEvent event, Resource resource, UU
         List<TriggerAssociation> ignoreTriggers = new ArrayList<>();
         for (UUID triggerUuid : triggerUuids) {
             Trigger trigger = getTriggerEntity(triggerUuid.toString());
+            if (event == ResourceEvent.CERTIFICATE_UPLOADED && trigger.getRules().stream()
+                    .flatMap(rule -> rule.getConditions().stream())
+                    .flatMap(condition -> condition.getItems().stream())
+                    .anyMatch(conditionItem -> conditionItem.getFieldSource() != FilterFieldSource.PROPERTY)) {
+                throw new ValidationException("Trigger '%s' cannot be associated with event '%s' as it contains rule with invalid field source, which is not allowed for this event.".formatted(trigger.getName(), event.getLabel()));
```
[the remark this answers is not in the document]
klaraf755: Added the check

src/test/java/com/czertainly/core/service/compliance/ComplianceServiceTest.java
```diff
@@ -148,25 +155,34 @@ void testCheckCompliance() throws Exception {
...
         var certificateChainInfo = CertificateGeneratorHelper.generateCertificateWithIssuer(KeyAlgorithm.RSA, "CN=Test-Ca", "CN=Test-EndEntity", null);
-        UploadCertificateRequestDto uploadRequestDto = new UploadCertificateRequestDto();
-        uploadRequestDto.setCertificate(certificateChainInfo.getCaCertificateBase64Encoded());
-        var certificateDto = certificateService.upload(uploadRequestDto, true);
-        uploadRequestDto.setCertificate(certificateChainInfo.getEndEntityCertificateBase64Encoded());
-        var certificate2Dto = certificateService.upload(uploadRequestDto, true);
+        X509Certificate x509Certificate = CertificateUtil.parseCertificate(certificateChainInfo.getCaCertificateBase64Encoded());
+        String fingerprint = CertificateUtil.getThumbprint(x509Certificate);
+        CertificateUploadEventMessageData eventData = CertificateUploadEventMessageData.builder()
+                .certificateContent(certificateChainInfo.getEndEntityCertificateBase64Encoded())
```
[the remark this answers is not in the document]
klaraf755: Fixed

src/main/java/com/czertainly/core/mapper/workflows/EventHistoryMapper.java
```diff
@@ -45,6 +45,9 @@ public static TriggerHistoryObjectTriggerSummaryDto toTriggerHistoryObjectTrigge
     public static TriggerHistoryObjectSummaryDto toTriggerHistoryObjectSummaryDto(List<TriggerHistory> triggerHistories, UUID objectUuid) {
         TriggerHistoryObjectSummaryDto triggerHistoryObjectSummaryDto = new TriggerHistoryObjectSummaryDto();
         triggerHistoryObjectSummaryDto.setObjectUuid(objectUuid);
+        boolean ignored = triggerHistories.stream().filter(th -> th.getTrigger().isIgnoreTrigger()).anyMatch(TriggerHistory::isActionsPerformed);
+        triggerHistoryObjectSummaryDto.setIgnored(ignored);
+        triggerHistoryObjectSummaryDto.setMatched(ignored || triggerHistories.stream().filter(th -> !th.getTrigger().isIgnoreTrigger()).allMatch(TriggerHistory::isConditionsMatched));
```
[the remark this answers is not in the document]
klaraf755: Fixed

src/main/java/com/czertainly/core/messaging/jms/listeners/NotificationListener.java
```diff
@@ -445,6 +445,17 @@ yield new InternalNotificationEventData("Certificate identified as '%s' with ser
...
                         .formatted(data.getSubjectDn(), data.getSerialNumber(), data.getIssuerDn()), null);
             }
+            case CERTIFICATE_UPLOADED -> {
+                CertificateUploadedEventData data = (CertificateUploadedEventData) eventData;
+                String internalMessage = "Certificate identified as '%s' with serial number '%s' issued by '%s' has been uploaded".formatted(data.getSubjectDn(), data.getSerialNumber(), data.getIssuerDn());
+                if (data.getUserUuid() != null) {
+                    internalMessage += "User identified by %s has been associated with the certificate".formatted(data.getUserUuid());
+                }
+                if (!data.getCustomAttributes().isEmpty()) {
+                    internalMessage += "Custom attributes %s from the upload request have been assigned".formatted(data.getCustomAttributes().toString());
```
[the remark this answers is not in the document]
klaraf755: Fixed

src/main/java/com/czertainly/core/service/impl/EventServiceImpl.java
```diff
@@ -128,14 +128,15 @@ public PaginationResponseDto<EventHistoryDto> getEventHistory(ResourceEvent even
...
-                : triggerHistoryRepository.findByEventHistoryUuidInAndObjectUuidInOrderByEventHistoryUuidAscObjectUuidAscTriggeredAtDesc(eventHistoryUuids, allPaginatedObjectUuids)
-                        .stream().collect(Collectors.groupingBy(
-                                TriggerHistory::getEventHistoryUuid,
-                                LinkedHashMap::new,
-                                Collectors.groupingBy(TriggerHistory::getObjectUuid, LinkedHashMap::new, Collectors.toList())
-                        ));
+        Map<UUID, Map<UUID, List<TriggerHistory>>> triggerHistoriesByEventAndObject = new LinkedHashMap<>();
+        if (!allPaginatedObjectUuids.isEmpty()) {
+            for (TriggerHistory th : triggerHistoryRepository.findByEventHistoryUuidsAndObjectUuids(eventHistoryUuids, allPaginatedObjectUuids)) {
+                triggerHistoriesByEventAndObject
```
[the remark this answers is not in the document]
klaraf755: Fixed

src/main/java/com/czertainly/core/dao/repository/workflows/TriggerHistoryRepository.java
```diff
@@ -75,9 +75,17 @@ List<Object[]> findPaginatedObjectUuidsByEventHistoryUuids(
     /**
      * Fetches all trigger histories for a set of event histories filtered to specific object UUIDs,
      * replacing one query per event history row. Results are grouped in Java by eventHistoryUuid → objectUuid.
+     * Includes rows where objectUuid IS NULL (e.g. ignored certificates that were never persisted).
      */
+    @Query("""
+            SELECT t FROM TriggerHistory t
+            WHERE t.eventHistoryUuid IN :eventHistoryUuids
+              AND (t.objectUuid IN :objectUuids OR t.objectUuid IS NULL)
```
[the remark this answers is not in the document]
klaraf755: Null is supposed to show up

src/main/java/com/czertainly/core/service/impl/TriggerServiceImpl.java
```diff
@@ -230,6 +232,12 @@ public void createTriggerAssociations(ResourceEvent event, Resource resource, UU
         List<TriggerAssociation> ignoreTriggers = new ArrayList<>();
         for (UUID triggerUuid : triggerUuids) {
             Trigger trigger = getTriggerEntity(triggerUuid.toString());
+            if (event == ResourceEvent.CERTIFICATE_UPLOADED && trigger.getRules().stream()
```
ivosh: Could you extract the trigger validation (lines 235-246) into a dedicated method.

src/main/java/com/czertainly/core/events/handlers/CertificateUploadedEventHandler.java
```diff
@@ -0,0 +1,192 @@
...
+
+    @Override
+    protected EventContext<Certificate> prepareContext(EventMessage eventMessage) throws EventException {
+        EventContext<Certificate> context = new EventContext<>(eventMessage, triggerEvaluator, new Certificate(), null);
+        fetchEventTriggers(context, null, null); // triggers without resource and its UUID are platform ones
+        return context;
+    }
+
+    @Override
+    @Transactional
```
ivosh, line 98: The transaction this methods open is way too broad and spans an external HTTP call (actually two). There are several ways how to tackle this. You can either split this method into several ones and annotate only some of them. Or annotate updateUserCertificate with `TransactionalEventListener(phase = AFTER_COMMIT)`. Or decide yourself.

```
CertificateUploadedEventHandler.handleEvent()   ← @Transactional (DB tx opens here)
  │  ... saves certificate, triggers, eventHistory (all inside tx) ...
  │
  └─ userManagementService.updateUserCertificate(userUuid, certUuid, null)   [line 454]
       │  (UserManagementServiceImpl has class-level @Transactional, propagation=REQUIRED
       │   → it JOINS the outer transaction; no @Transactional override on this method)
       │
       ├─ getUser(userUuid)
       │    └─ userManagementApiClient.getUserDetail(userUuid)
       │         └─ WebClient...toEntity(...).block()   ← blocking HTTP call #1 (tx held)
       │
       ├─ builds UpdateUserRequestDto from dto fields
       │
       └─ getUserUpdateRequestPayload(userUuid, request, certUuid, null)   [private]
            ├─ getExistingCertificate(certUuid, null) → DB lookup
            └─ userManagementApiClient.updateUser(userUuid, requestDto)
                 └─ WebClient...toEntity(...).block()   ← blocking HTTP call #2 (tx held)
```

src/test/java/com/czertainly/core/service/CertificateServiceTest.java
```diff
@@ -519,21 +519,11 @@ void findCertificateEntityByUserUuid_returnsEmptyWhenNoCertificateForUser() {
...
-        Assertions.assertEquals("CLIENT1", dto.getCommonName());
-        Assertions.assertEquals("177e75f42e95ecb98f831eb57de27b0bc8c47643", dto.getSerialNumber());
-
-        // test for presence of created public key
-        var newCertificate = certificateRepository.findWithAssociationsByUuid(UUID.fromString(dto.getUuid()));
-        Assertions.assertTrue(newCertificate.isPresent());
-        Assertions.assertEquals("certKey_%s".formatted(dto.getCommonName()), newCertificate.get().getKey().getName());
-        Assertions.assertEquals(1, newCertificate.get().getKey().getItems().size());
-        Assertions.assertEquals(KeyType.PUBLIC_KEY, newCertificate.get().getKey().getItems().stream().findFirst().get().getType());
+        Assertions.assertDoesNotThrow(() -> certificateService.upload(request, null));
```
ivosh: This test has become way too simple to actually test anything useful. I suggest to remove it completely and add new end-to-end test instead: CertificateService.upload() → EventProducer → queue → EventListener.processMessage() → CertificateUploadedEventHandler, using the real RabbitMQ.

The infrastructure is already there: BaseMessagingIntTest + JmsListenerIntegrationTest are already there. The gap is simply that no one extended that base class for CERTIFICATE_UPLOADED.
What a RabbitMQ integration test would catch that the current tests don't:
* Wrong routing key / exchange binding (cert upload message never reaches the handler)
* Serialization/deserialization bugs in CertificateUploadEventMessageData
* EventListener.processMessage() dispatching to the wrong handler
* The real async delivery gap — currently upload() returns void and the test has no way to know if the cert ever lands in the DB

src/test/java/com/czertainly/core/events/EventHandlersTest.java
```diff
@@ -517,6 +535,82 @@ void testScheduledJobFinishedEvent() {
...
+        uploadedCertificate = certificateRepository.findByFingerprint(fingerprint).orElseThrow();
+        certificateDetailDto = certificateService.getCertificate(uploadedCertificate.getSecuredUuid());
+        Assertions.assertFalse(certificateDetailDto.getCustomAttributes().isEmpty());
+        Optional<ResponseAttribute> customAttributeDtoOptional = certificateDetailDto.getCustomAttributes().stream().filter(attr -> CERTIFICATE_CUSTOM_ATTRIBUTE_NAME.equals(attr.getName())).findFirst();
+        Assertions.assertTrue(customAttributeDtoOptional.isPresent());
+        Assertions.assertEquals("fromRequest", ((List<StringAttributeContentV3>) customAttributeDtoOptional.get().getContent()).getFirst().getData());
+        Assertions.assertEquals(userUuid, uploadedCertificate.getUserUuid());
+    }
+
     @Test
```
ivosh, line 640: I am missing the following negative/error tests:
* Duplicate fingerprint (what does the handler return? what status is in EventHistory?)
* Malformed/unparseable certificate content
* updateUserCertificate failing after certificate is already saved - what happens if the auth service fails/rejects the user update? This is exactly the scenario described in CLAUDE.md - state divergence (external system rejects the work) - Core should either roll back or recover.

src/main/java/com/czertainly/core/events/handlers/CertificateUploadedEventHandler.java
```diff
@@ -0,0 +1,192 @@
...
+        applicationEventPublisher.publishEvent(new CertificateValidationEvent(certificate.getUuid()));
+        sendFollowUpEventsNotifications(context);
+    }
+
+    private void saveCertificate(CertificateUploadEventMessageData data, Certificate certificate) {
+        X509Certificate x509Certificate = CertificateUtil.parseUploadedCertificateContent(data.certificateContent());
+        CertificateContent certificateContent = certificateService.checkAddCertificateContent(data.fingerprint(), X509ObjectToString.toPem(x509Certificate));
+        certificate.setCertificateContent(certificateContent);
+        certificate.setCertificateContentId(certificateContent.getId());
+        certificate.setFingerprint(data.fingerprint());
```
ivosh: Certificate content is re-parsed but certificate fingerprint is blindly trusted without recalculation. The method should verify they match.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1202,33 +1204,32 @@ public CertificateContent checkAddCertificateContent(String fingerprint, String
...
+    public void upload(UploadCertificateRequestDto request, UUID userUuid) throws CertificateException, AlreadyExistException {
         X509Certificate certificate = CertificateUtil.parseUploadedCertificateContent(request.getCertificate());
-        String fingerprint = CertificateUtil.getThumbprint(certificate);
+        String fingerprint;
+        try {
+            fingerprint = CertificateUtil.getThumbprint(certificate);
+        } catch (NoSuchAlgorithmException | CertificateEncodingException e) {
+            throw new CertificateException("Failed to calculate certificate fingerprint: " + e.getMessage());
+        }
         if (certificateRepository.findByFingerprint(fingerprint).isPresent()) {
```
ivosh, line 1253: This is a subtle time-of-check-time-of-use (TOCTOU) bug which results from the new asynchronous certificate upload processing. The fingerprint uniqueness is checked in this method but the actual use is in CertificateUploadedEventHandler.handleEvent (via saveCertificate - fingerprint needs to be unique in the database). Recall that CertificateUploadedEventHandler.handleEvent runs in another thread, with a dedicated transaction.

Now immagine two concurrent certificates uploaded with the same fingeprint but for two different users. Both of them get queued (response 204) but one of them fails to be saved. In the event history UI, user  sees a FAILED upload event with no linked certificate (objectUuid = null) and no indication of why — the only hint is a DataIntegrityViolationException found in the logs.

I suggest to think how to give the user a more detailed feedback about potential failures in the asynchronous processing path.

ivosh (changes requested):

src/main/java/com/czertainly/core/api/web/CertificateControllerImpl.java
```diff
@@ -100,17 +96,9 @@ public void bulkUpdateCertificateObjects(MultipleCertificateObjectUpdateDto requ
...
-                .fromCurrentRequest()
-                .path("/{uuid}")
-                .buildAndExpand(dto.getUuid())
-                .toUri();
-        UuidDto responseDto = new UuidDto();
-        responseDto.setUuid(dto.getUuid());
-        return ResponseEntity.created(location).body(responseDto);
+    public void upload(@RequestBody UploadCertificateRequestDto request)
+            throws CertificateException, AlreadyExistException {
+        certificateService.upload(request, null);
```
ivosh: Passing `null` here is a code smell which has two dimensions:
1. Null-as-mode-flag — null is not a meaningful domain value, it's a sentinel that changes the internal behaviour of the method (skip user association). The method signature doesn't communicate this. Callers have to know that null means "anonymous upload" rather than "unset" or "error".
2. Wrong abstraction level — CertificateService.upload() shouldn't need to know who initiated the upload. The user association is an event handler concern, not an upload concern. The userUuid is only needed by the async handler, yet it's threaded through the service method purely to be packed into the message payload.

ivosh (changes requested):

src/main/java/com/czertainly/core/api/web/CertificateControllerImpl.java
klaraf755: I probably did not make clear what the user UUID represents here - it is not the UUID of user that is uploading the certificate, but the UUID of user the will be able to log in to the platform using this certificate. This UUID is here when the service method is being called from the UserManagementService while creating/updating a user. In this case, is the user UUID okay here or should I still reformat the method somehow?

src/main/java/com/czertainly/core/service/impl/TriggerServiceImpl.java
klaraf755: Done

src/main/java/com/czertainly/core/api/web/CertificateControllerImpl.java
ivosh: The intent would be clear if:
* UploadCertificateRequestDto stays used purely for the contract between external API client and ILM, and is not misused also for internal service-to-service communication.

In practice:
* CertificateServiceImpl.upload(UploadCertificateRequestDto request)` would accept just request parameter and translate UploadCertificateRequestDto into an internal request structure.
* UserManagementServiceImpl.uploadNewCertificate prepares the new internal request structure (containing userUuid), *not* UploadCertificateRequestDto
* both CertificateService.upload and uploadNewCertificate call a new method CertificateService.upload(internal request structure)

src/main/java/com/czertainly/core/events/handlers/CertificateUploadedEventHandler.java
klaraf755: Added published event for updateUserCertificate

src/main/java/com/czertainly/core/api/web/CertificateControllerImpl.java
klaraf755: I see now, I added the new internal method

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
klaraf755: I think EventHistory should also have a message column, for this case when there is a failure. But that would be added in separate PR.

src/test/java/com/czertainly/core/events/EventHandlersTest.java
klaraf755: I added tests for the two first cases, but since upload of certificate in user management operations is done synchronously, the last test is not relevant now.

src/test/java/com/czertainly/core/service/CertificateServiceTest.java
klaraf755: Added test for the async path.

src/main/java/com/czertainly/core/events/handlers/CertificateUploadedEventHandler.java
klaraf755: I actually passed the fingerprint along with content in order to not recalculate it again. But since the content is reparsed anyway, I will just remove the fingerprint from the message and recalculate it as well.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1204,33 +1216,56 @@ public CertificateContent checkAddCertificateContent(String fingerprint, String
...
             throw new AlreadyExistException("Certificate already exists with fingerprint " + fingerprint);
         }
 
-        if (!ignoreCustomAttributes) {
-            attributeEngine.validateCustomAttributesContent(Resource.CERTIFICATE, request.getCustomAttributes());
+        if (customAttributes != null && customAttributes.isEmpty()) {
```
ivosh: Double-check that `customAttributes.isEmpty()` is really intended here. It would effectively run the validation check only for empty attributes?

src/main/java/com/czertainly/core/service/impl/TriggerServiceImpl.java
```diff
@@ -263,6 +268,19 @@ public void createTriggerAssociations(ResourceEvent event, Resource resource, UU
...
 
+    private static boolean isCertificateUploadedEventCompatible(ResourceEvent event, Trigger trigger) {
+        return event != ResourceEvent.CERTIFICATE_UPLOADED || trigger.getRules().stream()
+                .flatMap(rule -> rule.getConditions().stream())
+                .flatMap(condition -> condition.getItems().stream())
+                .allMatch(TriggerServiceImpl::isCertificateUploadedEventCompatible);
+    }
+
+    private static boolean isCertificateUploadedEventCompatible(ConditionItem conditionItem) {
+        FilterField filterField = FilterField.valueOf(conditionItem.getFieldIdentifier());
```
ivosh: Double-check what happens if the trigger has condition with fieldSource=CUSTOM (or META/DATA). The guard below on line 281 is not validated yet (fieldSource == PROPERTY), but FilterField.valueOf could be called with an an attribute UUID or name (reflecting CUSTOM/META/DATA attribute).

I'd say the guard from line 281 related to fieldSource == PROPERTY shall be lifted at the beginning of this method.

src/main/java/com/czertainly/core/events/handlers/CertificateUploadedEventHandler.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    @Override
+    @Transactional
+    public void handleEvent(EventMessage eventMessage) throws EventException {
+        EventContext<Certificate> context = prepareContext(eventMessage);
+        EventHistory eventHistory = createEventHistory(ResourceEvent.CERTIFICATE_UPLOADED, null, null);
+        CertificateUploadEventMessageData eventMessageData = objectMapper.convertValue(eventMessage.getData(), CertificateUploadEventMessageData.class);
+
+        X509Certificate x509Certificate;
```
ivosh, line 104: Certificate is parsed two times in CertificateUploadedEventHandler (lines 107 and 173) and also in CertificateServiceImpl:1234.  I understand you maybe do not want to carry the parsed cert in the event from CertificateServiceImpl to CertificateUploadedEventHandler. But this class should definitely parse the cert just once.

src/main/java/com/czertainly/core/events/handlers/CertificateUploadedEventHandler.java
```diff
@@ -0,0 +1,194 @@
...
+import com.czertainly.core.events.transaction.CertificateValidationEvent;
+import com.czertainly.core.messaging.model.CertificateUploadEventMessageData;
+import com.czertainly.core.messaging.model.EventMessage;
+import com.czertainly.core.messaging.model.NotificationMessage;
+import com.czertainly.core.messaging.model.NotificationRecipient;
+import com.czertainly.core.service.CertificateEventHistoryService;
+import com.czertainly.core.service.CertificateService;
+import com.czertainly.core.util.CertificateUtil;
+import com.czertainly.core.util.X509ObjectToString;
+import jakarta.transaction.Transactional;
```
ivosh: Pay attention here because it imports `jakarta.transaction.Transactional` while the majority of the codebase imports `org.springframework.transaction.annotation.Transactional`.

src/main/java/com/czertainly/core/events/handlers/CertificateUploadedEventHandler.java
```diff
@@ -0,0 +1,194 @@
...
+        EventContext<Certificate> context = prepareContext(eventMessage);
+        EventHistory eventHistory = createEventHistory(ResourceEvent.CERTIFICATE_UPLOADED, null, null);
+        CertificateUploadEventMessageData eventMessageData = objectMapper.convertValue(eventMessage.getData(), CertificateUploadEventMessageData.class);
+
+        X509Certificate x509Certificate;
+        try {
+            x509Certificate = CertificateUtil.parseCertificate(eventMessageData.certificateContent());
+        } catch (CertificateException e) {
+            logger.error("Unable to parse certificate {} from uploaded certificate: {}", eventMessageData.certificateContent(), e.getMessage());
+            eventHistory.setStatus(EventStatus.FAILED);
```
ivosh: Could you create a helper for this block of code? It is repeated five times in the handler.

src/main/java/com/czertainly/core/events/handlers/CertificateUploadedEventHandler.java
```diff
@@ -0,0 +1,194 @@
...
+        if (certificateRepository.findByFingerprint(fingerprint).isPresent()) {
+            logger.warn("Certificate with fingerprint {} already exists, skipping upload event", fingerprint);
+            eventHistory.setStatus(EventStatus.FAILED);
+            eventHistory.setFinishedAt(OffsetDateTime.now());
+            eventHistoryRepository.save(eventHistory);
+            return;
+        }
+        Certificate certificate = context.getResourceObjects().getFirst();
+        CertificateUtil.prepareIssuedCertificate(certificate, x509Certificate);
+        CertificateEventData eventData = (CertificateEventData) getEventData(certificate, eventMessageData);
```
ivosh, line 128: I think `getEventData` does not populate correctly the event because it does not have all the information available in `certificate`.
The uuid is known only later, when the certificate is actually persisted (line 142). Availability of fingerprint is also questionable.


I suggest you add a new test which basically exercises the ignore trigger and verifies the produced trigger history. Adapt as needed:
```
@Autowired
private TriggerHistoryRepository triggerHistoryRepository;

@Test
void testCertificateUploadedEventIgnored_triggerHistoryCarriesFingerprint() throws Exception {
    X509Certificate certificate = CertificateGeneratorHelper.generateCACertificate(null, "CN=test");
    String fingerprint = CertificateUtil.getThumbprint(certificate);
    var eventMessageData = CertificateUploadEventMessageData.builder()
            .certificateContent(Base64.getEncoder().encodeToString(certificate.getEncoded()))
            .build();

    createCertificateTriggerAssociation(ResourceEvent.CERTIFICATE_UPLOADED, null, null, true);

    certificateUploadedEventHandler.handleEvent(
            CertificateUploadedEventHandler.constructEventMessage(eventMessageData));

    Assertions.assertFalse(certificateRepository.findByFingerprint(fingerprint).isPresent(),
            "ignored cert must not be persisted");

    List<TriggerHistory> histories = triggerHistoryRepository.findAll();
    Assertions.assertEquals(1, histories.size());
    TriggerHistory th = histories.getFirst();
    Assertions.assertTrue(th.isConditionsMatched());
    Assertions.assertTrue(th.isActionsPerformed());

    Assertions.assertNotNull(th.getMessage());
    Assertions.assertTrue(th.getMessage().contains(fingerprint), "ignore TriggerHistory.message includes the fingerprint");
}
```

ivosh (changes requested):

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
klaraf755: Should definitely be `!customAttributes.isEmpty()`.

src/main/java/com/czertainly/core/service/impl/TriggerServiceImpl.java
klaraf755: Yes, I separated that and also added a test for that in the service.

src/main/java/com/czertainly/core/events/handlers/CertificateUploadedEventHandler.java
klaraf755: Yes, it should parse it just once. I did not carry the parsed X509 certificate because it was not able to be serialised for the event message.

src/main/java/com/czertainly/core/events/handlers/CertificateUploadedEventHandler.java
klaraf755: Changed it

src/main/java/com/czertainly/core/events/handlers/CertificateUploadedEventHandler.java
klaraf755: I extracted it `EventHandler` as protected method.

src/main/java/com/czertainly/core/events/handlers/CertificateUploadedEventHandler.java
klaraf755: I added UUID of certificate to event data after it is persisted, and fingerprint to the certificate before creating event data. I already had test for the ignore trigger, so I just added the check for trigger history there.

ivosh (dismissed):

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1218,18 +1228,20 @@ public CertificateContent checkAddCertificateContent(String fingerprint, String
     @ExternalAuthorization(resource = Resource.CERTIFICATE, action = ResourceAction.CREATE)
     @Transactional(propagation = Propagation.NOT_SUPPORTED)
     public FingerprintDto uploadAsync(UploadCertificateRequestDto request) throws CertificateException, AlreadyExistException {
-        String fingerprint = upload(request.getCertificate(), request.getCustomAttributes(), false);
+        String fingerprint = ((CertificateService) AopContext.currentProxy()).upload(request.getCertificate(), request.getCustomAttributes(), false);
```
ivosh, line 1231: Uh oh. This is smelly. Better is to inject `self` and invoke `self.upload()` to involve the method in a dedicated transaction. Fix it in a follow-up PR.

ivosh (approved):

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
klaraf755: I had it originally like that, but then an issue with circular dependency has occurred, so I will think about how to resolve it differently.

PR #1471, opened 2026-05-16
Split services into *ServiceInternal and *ServiceExternal

ivosh:


ivosh:
I will split this PR into smaller ones.

PR #1473, opened 2026-05-17
Cache Cryptographic Key Item lookups for signing operations

### LINKED ISSUE #1475: Cache cryptographic key items to avoid database lookups
### Description

Caffeine cache for CryptographicKeyItem lookups used by cryptographic operations (sign/verify/encrypt/decrypt). A new CryptographicKeyItemModel record captures only the data needed by these hot paths, the service layer populates/invalidates the cache around every key-item mutation.

ivosh:


src/main/java/com/czertainly/core/service/CryptographicOperationService.java
```diff
@@ -104,6 +116,25 @@ SignDataResponseDto signData(
...
+     * @param request           DTO containing the data to sign a request {@Link SignDataRequestDto}
+     * @return Signed Data {@Link SignDataResponseDto}
+     * @throws NotFoundException when the token instance with the specified UUID is not found
+     */
+    SignDataResponseDto signDataWithoutEventHistory(
+            SecuredParentUUID tokenInstanceUuid,
+            SecuredUUID tokenProfileUUID,
+            UUID uuid, UUID keyItemUuid,
+            SignDataRequestDto request
+    ) throws ConnectorException, NotFoundException;
```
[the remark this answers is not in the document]
ivosh: That's right. The new TSA engine caller will arrive soon.

src/main/java/com/czertainly/core/service/CryptographicOperationService.java
```diff
@@ -104,6 +116,25 @@ SignDataResponseDto signData(
...
+     * @param tokenInstanceUuid UUID of the token instance
+     * @param tokenProfileUUID  UUID of the token profile
+     * @param uuid              UUID of the cryptographic key
+     * @param keyItemUuid       UUID of the Item inside the key Object
+     * @param request           DTO containing the data to sign a request {@Link SignDataRequestDto}
+     * @return Signed Data {@Link SignDataResponseDto}
+     * @throws NotFoundException when the token instance with the specified UUID is not found
+     */
+    SignDataResponseDto signDataWithoutEventHistory(
+            SecuredParentUUID tokenInstanceUuid,
```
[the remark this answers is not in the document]
ivosh: Parameter is used for authorization, via SecuredParentUUID. Cannot remove it.

src/main/java/com/czertainly/core/service/CryptographicOperationService.java
```diff
@@ -104,6 +116,25 @@ SignDataResponseDto signData(
...
+            SecuredUUID tokenProfileUUID,
+            UUID uuid, UUID keyItemUuid,
```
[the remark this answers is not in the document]
ivosh: Cannot remove it. It is a part of the contract defined in interfaces.

src/main/resources/application.yml
```diff
@@ -7,6 +7,10 @@ caching:
   authentication:
     ttl-minutes: ${AUTH_CACHE_TTL_MINUTES:5}   # How long a cached identity is considered valid
     max-size: ${AUTH_CACHE_MAX_SIZE:500}       # Maximum number of cached identities
+  # Cryptographic Key Item cache caches keys and their associations avoid repeated database lookups
```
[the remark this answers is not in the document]
ivosh: Fixed.

src/main/java/com/czertainly/core/config/cache/CryptographicKeyItemCacheProperties.java
```diff
@@ -0,0 +1,15 @@
...
+
+@ConfigurationProperties(prefix = "caching.crypto-keys")
+public record CryptographicKeyItemCacheProperties(
+        int ttlMinutes,
+        int maxSize
+) {
+
+    public CryptographicKeyItemCacheProperties {
+        if (ttlMinutes <= 0) ttlMinutes = 5;
+        if (maxSize <= 0) maxSize = 500;
```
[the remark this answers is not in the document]
ivosh: Fixed.

src/main/java/com/czertainly/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -903,8 +919,65 @@ public CryptographicKeyItem getKeyItemFromKey(CryptographicKey key, KeyType keyT
...
+                keyItem.getState(),
+                keyItem.isEnabled(),
+                keyItem.getUsage(),
+                keyItem.getKeyAlgorithm(),
+                keyItem.getKeyReferenceUuid(),
+                tokenInstanceReference.getConnector().mapToApiClientDtoV1(),
+                tokenInstanceUuid
+        );
+        if (cache != null)
+            cache.put(keyItemUuid, model);
```
lubomirw, line 963: **~ Minor** | Correctness

Worth weighing whether the 5-minute staleness window from the classic cache-aside race is acceptable here.

The race: Thread A reads the cache (miss), queries the DB at T1 and gets V1. Thread B mutates the same key item, commits, and `afterCommit` fires `cache.evict(uuid)` at T2 — a no-op because the entry isn't there yet. Thread A then `cache.put(uuid, V1)` at T3, and the stale value lingers for the full TTL.

Considerations:

- **Window is narrow** — only when a concurrent write commits between A's DB load and A's `cache.put`. Sequential and write-then-read flows are fine.
- **Cached data is security-relevant** — `state`, `enabled`, and `usage` feed `verifyKeyActive` and the per-operation usage check (`!key.usage().contains(KeyUsage.SIGN)` etc.). A compromised or disabled key remaining usable for up to 5 min under the race is the worst case.
- **Fix is cheap** — `cache.get(keyItemUuid, () -> /* the existing DB-and-model-build */)` is atomic on Caffeine: concurrent loaders for the same key are coordinated, and an eviction during the load is visible to the loader. Same hot-path performance, no thundering herd.
- **Equivalent annotation route** — converting `getKeyItemModel` to `@Cacheable` and the manual evicts to `@CacheEvict` (or keeping the existing `afterCommit` evict alongside) achieves the same property with less boilerplate.

If a 5-min worst-case staleness on `state`/`enabled`/`usage` is acceptable operationally (e.g. compromise/disable flows already have other guardrails), the current code is fine. If not, the atomic-loader form is a small change worth making now while the cache is fresh.

src/main/java/com/czertainly/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -903,8 +919,65 @@ public CryptographicKeyItem getKeyItemFromKey(CryptographicKey key, KeyType keyT
...
+        UUID tokenInstanceUuid = UUID.fromString(tokenInstanceReference.getTokenInstanceUuid());
+
+        CryptographicKeyItemModel model = new CryptographicKeyItemModel(
+                keyItem.getUuid(),
+                keyItem.getState(),
+                keyItem.isEnabled(),
+                keyItem.getUsage(),
+                keyItem.getKeyAlgorithm(),
+                keyItem.getKeyReferenceUuid(),
+                tokenInstanceReference.getConnector().mapToApiClientDtoV1(),
```
lubomirw: **~ Minor** | Correctness

Worth considering whether the 5-minute connector-staleness window is worth closing, given that Connector edits (URL changes, credential rotation) are rare.

The cached `CryptographicKeyItemModel` snapshots `Connector` data via `mapToApiClientDtoV1()` — URL, auth credentials, status — but eviction is wired only to **key item** mutations. If a `Connector` is edited via Connector management, the cached `ApiClientConnectorInfo` keeps the old values for up to the TTL. During that window, the signing/encryption operations call the stale URL or use stale credentials and surface as connection failures, auth errors, or — worst case — calls to a host that's been repurposed.

Considerations:

- **Frequency** — if Connector URL/credential edits are genuinely rare, the worst case is encountered seldom and recoverable (operation fails, retry after TTL clears it).
- **Cheap mitigation if desired** — a single `cache.clear()` on any Connector edit closes the gap without touching the hot path. Cold-start cycles are paid only at the rare Connector edit. A more targeted variant (evict only key items linked to the affected connector) needs a join query but bounds the eviction.
- **Don't go the other way** — pulling `Connector` out of the cached model and looking it up live on every signing op would defeat much of the cache's value (extra DB hit per operation on the hot path) and is not worth it for a rare event.

If a 5-min connector-staleness on edits is acceptable operationally — e.g. credential rotations already have a grace overlap, URL changes happen during planned maintenance — the current code is fine and the eviction-on-connector-edit can wait for a follow-up. If not, the `cache.clear()`-on-connector-edit option is small and worth doing here.

src/main/java/com/czertainly/core/service/impl/CryptographicOperationServiceImpl.java
```diff
@@ -241,23 +252,41 @@ public List<BaseAttribute> listSignatureAttributes(SecuredParentUUID tokenInstan
 
     @Override
     @ExternalAuthorization(resource = Resource.CRYPTOGRAPHIC_KEY, action = ResourceAction.SIGN, parentResource = Resource.TOKEN, parentAction = ResourceAction.DETAIL)
+    @Transactional
```
lubomirw, line 263: **~ Minor** | Correctness

Worth filing as a follow-up: the per-method `@Transactional` annotations preserve a pre-existing pattern where `signData`, `encryptData`, `decryptData`, `verifyData`, and `randomData` hold a database transaction across the HTTP call to the connector (HSM / signer). The repo's `CLAUDE.md` (Transactions and external calls) calls this out — a slow or unresponsive connector holds a JDBC connection and any row locks for its entire response time.

Not in scope for this PR — class-level `@Transactional` already did this before, and rewiring all five methods (drop `@Transactional`, push local writes through `addEventHistory`'s own tx, or use the `Propagation.NOT_SUPPORTED` + explicit-tx pattern from `CLAUDE.md`) is its own refactor. But the move from class-level to per-method annotation does make each call site visible, so this is a good moment to capture the follow-up.

Two interesting data points already in this PR:

- `signDataWithoutEventHistory` deliberately omits `@Transactional` — that's the right shape, and it's a useful template for what the other operations could look like.
- The actual local writes in each `@Transactional` method are inside `eventHistoryService.addEventHistory(...)`, which presumably has its own tx boundary. If so, the outer `@Transactional` is contributing nothing except wrapping the HTTP call.

lubomirw (commented):

src/main/java/com/czertainly/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -903,8 +919,65 @@ public CryptographicKeyItem getKeyItemFromKey(CryptographicKey key, KeyType keyT
...
+                    cache.evict(keyItemUuid);
+                }
+            });
+        } else {
+            cache.evict(keyItemUuid);
+        }
+    }
 
     @Override
+    @Transactional(noRollbackFor = ValidationException.class)
```
lubomirw, line 983: **~ Minor** | Maintainability

The newly added method-level `@Transactional(noRollbackFor = ValidationException.class)` on `uploadCertificatePublicKey` is identical to the class-level annotation declared on `CryptographicKeyServiceImpl`. Spring's transaction attribute resolution picks the most specific annotation, so the method-level one wins — but since the attributes are the same, it has no effect.

If there was an intent to override something specific for this method (e.g. different propagation, isolation, readOnly, or removal of `noRollbackFor`), that intent isn't visible in the annotation as written. If there wasn't, the annotation can be removed and the class-level one continues to apply.

Worth either adding a one-line comment explaining the override or removing the redundant annotation.

src/main/java/com/czertainly/core/config/cache/CryptographicKeyItemCacheProperties.java
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "caching.crypto-keys")
+public record CryptographicKeyItemCacheProperties(
+        int ttlMinutes,
+        int maxSize
+) {
+
+    public CryptographicKeyItemCacheProperties {
+        if (ttlMinutes <= 0) ttlMinutes = 5;
```
lubomirw: **~ Minor** | Correctness

What's the intent of the canonical-constructor fallback here?

```java
public CryptographicKeyItemCacheProperties {
    if (ttlMinutes <= 0) ttlMinutes = 5;
    if (maxSize <= 0) maxSize = 10000;
}
```

A few framings, depending on the intent:

- **"Defend against missing config"** — `application.yml` already supplies defaults via `${CRYPTO_KEYS_CACHE_TTL_MINUTES:5}` / `${CRYPTO_KEYS_CACHE_MAX_SIZE:10000}`, so this branch can only fire when an operator explicitly sets the env var to `0` or a negative number. If that's the intent, worth a one-line comment naming it so future readers don't read this as redundant.

- **"Forgive operator typos"** — the fallback silently rewrites `ttlMinutes=0` (which an operator might intend as "disable caching") to `5`, with no log line. If lenient clamping is intentional, a `warn`-level log when the override fires would let the operator notice their value was rejected.

- **"Reject bad input"** — if the intent is just "don't let bad values through", `@Validated` on the bean + `@Min(1)` on each parameter rejects the context at startup with a clear error, which is usually preferable to silent clamping.

If it's just defensive code without a specific scenario in mind, the simpler option is to drop the constructor body and let the `application.yml` defaults stand on their own.

src/main/java/com/czertainly/core/service/impl/CryptographicKeyServiceImpl.java
ivosh: This will be tackled in the connector cache which needs to land first.

src/main/java/com/czertainly/core/service/impl/CryptographicOperationServiceImpl.java
ivosh: Acknowledged.

src/main/java/com/czertainly/core/service/impl/CryptographicKeyServiceImpl.java
ivosh: Fixed.

src/main/java/com/czertainly/core/service/impl/CryptographicKeyServiceImpl.java
ivosh: Weighted. I will not fix it.

The race is real in the abstract. The race is essentially never observable in practice for this workload, because:
* Lifecycle mutations are rare events on any given key.
* They have to collide with the millisecond-scale load window on the same key, for the operations to collide.
* The consequence is bounded by TTL and softened by upstream defenses.

The atomic-loader fix is small but not free — it introduces an exception-unwrapping path (a real source of subtle bugs) and changes hot-path semantics. This isn't obviously a win.

lubomirw (dismissed):

src/main/java/com/czertainly/core/config/cache/CryptographicKeyItemCacheProperties.java
ivosh: I replaced the compact constructor with validation annotations.

src/main/java/com/czertainly/core/service/impl/CryptographicKeyServiceImpl.java
ivosh: Refactored, using connector cache.

lubomirw (dismissed):

lubomirw (approved):

PR #1474, opened 2026-05-17
Split InternalService and ExternalService - part 2

ivosh:


src/test/java/com/czertainly/core/service/ConnectorServiceMockTest.java
```diff
@@ -62,7 +62,7 @@ class ConnectorServiceMockTest {
     private ComplianceProfileRuleRepository complianceProfileRuleRepository;
 
     @Mock
-    private ConnectorAuthService connectorAuthService;
+    private ConnectorAuthExternalService connectorAuthService;
```
lubomirw: **! Major** | Correctness / Testing

Mock type does not match the field/setter in `ConnectorServiceImpl` — `connectorAuthService` is declared as `ConnectorAuthInternalService` there, but this test mocks `ConnectorAuthExternalService`.

`@InjectMocks` matches by type, and the two interfaces are unrelated (neither extends the other; only the impl class implements both). Mockito will silently leave the field `null` rather than inject the mock. The current `testConnect_*` cases happen not to exercise `mergeAndValidateAuthAttributes`, so the test still passes — but the mock is effectively dead and any future test in this class that touches the auth path will NPE.

Change to `@Mock private ConnectorAuthInternalService connectorAuthService;` to match the production wiring.

lubomirw (changes requested):
Code review: 1 issue found.

src/test/java/com/czertainly/core/service/ConnectorServiceMockTest.java
ivosh: Fixed.

lubomirw (approved):

PR #1476, opened 2026-05-18
Add Signing Profile Management

### LINKED ISSUE #197: Merge Signing Profile management from feat/signing into main
### Description

Merge Signing Profile management from feat/signing into main:

* interfaces:
* core
* frontend

LukasNajman:


src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,826 @@
...
+            case ContentSigningWorkflowRequestDto w -> {
+                if (w.getSignatureFormatterConnectorUuid() == null) {
+                    throw new ValidationException("Signature formatter connector is required for content signing workflow");
+                }
+                Connector contentConnector = connectorRepository.findByUuid(w.getSignatureFormatterConnectorUuid())
+                        .orElseThrow(() -> new NotFoundException(Connector.class, w.getSignatureFormatterConnectorUuid()));
+                validateFormatterConnectorFeature(contentConnector, FeatureFlag.CONTENT_SIGNING, SigningWorkflowType.CONTENT_SIGNING);
+                version.setSignatureFormatterConnector(contentConnector);
+            }
+            case RawSigningWorkflowRequestDto w -> {
```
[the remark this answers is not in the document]
LukasNajman: We are targeting Java 21, where unnamed variables are not supported

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,826 @@
...
+            case ContentSigningWorkflowRequestDto w -> {
+                fetchAndUpdateFormatterAttributeDefinitions(w.getSignatureFormatterConnectorUuid());
+                yield attributeEngine.replaceObjectDataAttributesContent(
+                        ObjectAttributeContentInfo.builder(Resource.SIGNING_PROFILE, p.getUuid())
+                                .connector(w.getSignatureFormatterConnectorUuid())
+                                .operation(AttributeOperation.WORKFLOW_FORMATTER)
+                                .version(version.getVersion()).build(),
+                        w.getSignatureFormatterConnectorAttributes());
+            }
+            case RawSigningWorkflowRequestDto w -> {
```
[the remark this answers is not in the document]
LukasNajman: We are targeting Java 21, where unnamed variables are not supported

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,826 @@
...
+        }
+
+        String oldName = profile.getName();
+        profile.setName(request.getName());
+        profile.setDescription(request.getDescription());
+
+        SigningProfileVersion version;
+        version = new SigningProfileVersion();
+        version.setSigningProfile(profile);
+        version.setVersion(profile.getLatestVersion());
```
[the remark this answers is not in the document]
LukasNajman: Fixed

src/main/java/com/czertainly/core/dao/entity/signing/SigningProfile.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.UUID;
+
+@Getter
+@Setter
+@Entity
+@Table(name = "signing_profile")
+public class SigningProfile extends UniquelyIdentifiedAndAudited implements Securable {
+
+    @Column(name = "name", nullable = false)
+    private String name;
```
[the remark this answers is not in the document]
LukasNajman: fixed

src/main/java/com/czertainly/core/api/web/SigningProfileControllerImpl.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    @Autowired
+    public SigningProfileControllerImpl(SigningProfileService signingProfileService) {
+        this.signingProfileService = signingProfileService;
+    }
+
+    @Override
+    @AuthEndpoint(resourceName = Resource.SIGNING_PROFILE)
+    @AuditLogged(module = Module.SIGNING, resource = Resource.SIGNING_PROFILE, operation = Operation.LIST)
```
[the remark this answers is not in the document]
LukasNajman: Fixed

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,826 @@
...
+    private ConnectorRepository connectorRepository;
+
+    // ──────────────────────────────────────────────────────────────────────────
+    // List / search
+    // ──────────────────────────────────────────────────────────────────────────
+
+    @Override
+    @Transactional(readOnly = true)
+    public List<SearchFieldDataByGroupDto> getSearchableFieldInformation() {
+        List<SearchFieldDataByGroupDto> searchFieldDataByGroupDtos = attributeEngine.getResourceSearchableFields(Resource.SIGNING_PROFILE, false);
```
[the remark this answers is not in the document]
LukasNajman: Added mapping into `ResourceToClass`

src/main/java/com/czertainly/core/mapper/signing/SigningProfileMapper.java
```diff
@@ -0,0 +1,198 @@
...
+            );
+        }
+        wf.setValidateTokenSignature(version.getValidateTokenSignature() != null ? version.getValidateTokenSignature() : false);
+        if (header.getTimeQualityConfiguration() != null) {
+            wf.setTimeQualityConfiguration(TimeQualityConfigurationMapper.toDto(header.getTimeQualityConfiguration(), List.of()));
+        }
+        return wf;
+    }
+
+    private static void setFormatterRef(SigningProfileVersion profileVersion, Consumer<NameAndUuidDto> setter) {
```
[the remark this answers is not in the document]
LukasNajman: Null handled gracefully

src/main/java/com/czertainly/core/service/impl/TimeQualityConfigurationServiceImpl.java
```diff
@@ -230,12 +234,15 @@ private void fillTimeQualityConfigurationEntity(TimeQualityConfiguration entity,
 
     private void deleteTimeQualityConfiguration(TimeQualityConfiguration configuration) {
         attributeEngine.deleteObjectAttributeContent(Resource.TIME_QUALITY_CONFIGURATION, configuration.getUuid());
+        signingProfileService.notifyTimeQualityConfigurationChange(configuration.getUuid());
         timeQualityConfigurationRepository.delete(configuration);
```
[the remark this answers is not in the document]
LukasNajman: Added check that will throw ValidationException when trying to delete time profile associated with SigningProfile

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,826 @@
...
+    @Override
+    public List<CertificateDto> listSigningCertificates(SigningWorkflowType signingWorkflowType, boolean qualifiedTimestamp) {
+        return certificateService.listDigitalSigningCertificates(SecurityFilter.create(), signingWorkflowType, qualifiedTimestamp);
+    }
+
+    @Override
+    @Transactional(readOnly = true)
+    public List<BaseAttribute> listSignatureAttributesForCertificate(UUID certificateUuid) throws NotFoundException {
+        Certificate certificate = certificateRepository.findByUuid(certificateUuid)
+                .orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
```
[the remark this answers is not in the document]
LukasNajman: Implemented - the access is now through service layer throughout the SignigProfile service

src/main/java/com/czertainly/core/service/impl/TimeQualityConfigurationServiceImpl.java
```diff
@@ -256,4 +263,14 @@ public void setTimeQualityConfigurationRepository(TimeQualityConfigurationReposi
...
+
+    @Autowired
+    public void setSigningProfileService(SigningProfileService signingProfileService) {
+        this.signingProfileService = signingProfileService;
+    }
+
+    @Autowired
+    public void setApplicationEventPublisher(ApplicationEventPublisher applicationEventPublisher) {
+        this.applicationEventPublisher = applicationEventPublisher;
+    }
```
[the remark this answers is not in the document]
LukasNajman: Removed

src/main/java/com/czertainly/core/service/impl/TspProfileServiceImpl.java
```diff
@@ -269,13 +282,29 @@ public void evaluatePermissionChain(SecuredUUID uuid) throws NotFoundException {
     // Private helpers
     // ──────────────────────────────────────────────────────────────────────────
 
-    private void validateCreateUpdateRequest(TspProfileRequestDto request) throws ValidationException {
+    private static String safeBulkMessage(Exception e) {
+        return e.getClass().getPackageName().startsWith("com.czertainly.api.exception") && e.getMessage() != null
+                ? e.getMessage()
+                : "Operation failed";
+    }
+
```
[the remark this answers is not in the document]
LukasNajman: removed

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,826 @@
...
+        return null;
+    }
+
+    private List<ResponseAttribute> persistSignatureFormatterConnectorAttributes(SigningProfile p, SigningProfileVersion version, WorkflowRequestDto workflow)
+            throws AttributeException, ConnectorException, NotFoundException {
+        return switch (workflow) {
+            case ContentSigningWorkflowRequestDto w -> {
+                fetchAndUpdateFormatterAttributeDefinitions(w.getSignatureFormatterConnectorUuid());
+                yield attributeEngine.replaceObjectDataAttributesContent(
+                        ObjectAttributeContentInfo.builder(Resource.SIGNING_PROFILE, p.getUuid())
```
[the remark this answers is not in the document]
LukasNajman: added attributes validation

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,826 @@
...
+                }
+                if (!certificateService.getCertificateChain(SecuredUUID.fromUUID(certificate.getUuid()), false).isCompleteChain()) {
+                    throw new ValidationException("Certificate " + certificate.getUuid() + " does not represent a complete certificate chain");
+                }
+            }
+            case OneTimeKeyManagedSigningRequestDto s -> {
+                version.setManagedSigningType(ManagedSigningType.ONE_TIME_KEY);
+                version.setTokenProfileUuid(s.getTokenProfileUuid());
+                version.setRaProfileUuid(s.getRaProfileUuid());
+                version.setCsrTemplateUuid(s.getCsrTemplateUuid());
```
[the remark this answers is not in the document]
LukasNajman: Existence of RaProfile and TokenProfile is now validated before the SingingProfile is saved.

LukasNajman:
In this PR, new method `getConnectorEntity` was added to `ConnectorService`. No `@ExternalAuthorization` was used for this method as `getConnector` also does not have it. Is this expected? @ivosh or @lubomirw

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,826 @@
...
+            }
+            case DelegatedSigningRequestDto s -> {
+                version.setDelegatedSignerConnectorUuid(s.getConnectorUuid());
```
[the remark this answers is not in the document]
LukasNajman: The whole service now uses Connector service and the service handles this properly.

src/main/java/com/czertainly/core/service/impl/TspProfileServiceImpl.java
```diff
@@ -269,13 +282,29 @@ public void evaluatePermissionChain(SecuredUUID uuid) throws NotFoundException {
...
+        SigningProfile defaultSigningProfile = null;
+        if (request.getDefaultSigningProfileUuid() != null) {
+            UUID defaultSigningProfileUuid = request.getDefaultSigningProfileUuid();
+            defaultSigningProfile = signingProfileRepository.findByUuid(SecuredUUID.fromUUID(defaultSigningProfileUuid))
+                    .orElseThrow(() -> new NotFoundException("Signing Profile not found: " + defaultSigningProfileUuid));
+        }
+
+        return defaultSigningProfile;
     }
```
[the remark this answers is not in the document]
LukasNajman: Added validation

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,826 @@
...
+        SigningProfile signingProfile = findByUuid(uuid);
+        return SigningProfileMapper.toTspActivationDto(signingProfile);
+    }
+
+    @Override
+    @ExternalAuthorization(resource = Resource.SIGNING_PROFILE, action = ResourceAction.UPDATE)
+    @Transactional
+    public TspActivationDetailDto activateTsp(SecuredUUID signingProfileUuid, SecuredUUID tspProfileUuid) throws NotFoundException {
+        SigningProfile signingProfile = findByUuid(signingProfileUuid);
+        validateSupportedProtocol(signingProfile.getWorkflowType(), SigningProtocol.TSP);
```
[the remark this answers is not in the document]
LukasNajman: Updated, all calls not go through TspService

src/main/java/com/czertainly/core/service/impl/TspProfileServiceImpl.java
```diff
@@ -269,13 +282,29 @@ public void evaluatePermissionChain(SecuredUUID uuid) throws NotFoundException {
...
+                : "Operation failed";
+    }
+
+    private SigningProfile validateCreateUpdateRequest(TspProfileRequestDto request) throws NotFoundException, ValidationException {
         attributeEngine.validateCustomAttributesContent(Resource.TSP_PROFILE, request.getCustomAttributes());
+
+        SigningProfile defaultSigningProfile = null;
+        if (request.getDefaultSigningProfileUuid() != null) {
+            UUID defaultSigningProfileUuid = request.getDefaultSigningProfileUuid();
+            defaultSigningProfile = signingProfileRepository.findByUuid(SecuredUUID.fromUUID(defaultSigningProfileUuid))
```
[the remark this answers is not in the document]
LukasNajman: Access is through service now. The service checks the permissions.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -421,8 +422,8 @@ public CertificateDetailDto getCertificate(SecuredUUID uuid) throws NotFoundExce
         }
         if (dto.getCertificateRequest() != null) {
             dto.getCertificateRequest().setAttributes(attributeEngine.getObjectDataAttributesContent(ObjectAttributeContentInfo.builder(Resource.CERTIFICATE_REQUEST, certificate.getCertificateRequest().getUuid()).build()));
-            dto.getCertificateRequest().setSignatureAttributes(attributeEngine.getObjectDataAttributesContent(ObjectAttributeContentInfo.builder(Resource.CERTIFICATE_REQUEST, certificate.getCertificateRequest().getUuid()).operation(AttributeOperation.SIGN).build()));
-            dto.getCertificateRequest().setAltSignatureAttributes(attributeEngine.getObjectDataAttributesContent(ObjectAttributeContentInfo.builder(Resource.CERTIFICATE_REQUEST, certificate.getCertificateRequest().getUuid()).operation(AttributeOperation.SIGN).purpose(AttributeContentPurpose.CERTIFICATE_REQUEST_ALT_KEY).build()));
+            dto.getCertificateRequest().setSignatureAttributes(attributeEngine.getObjectDataAttributesContent(ObjectAttributeContentInfo.builder(Resource.CERTIFICATE_REQUEST, certificate.getCertificateRequest().getUuid()).operation(AttributeOperation.CERTIFICATE_REQUEST_SIGN).build()));
```
ivosh: Use AttributeOperation.SIGN. CERTIFICATE_REQUEST_SIGN is legacy, for database migrations only.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -421,8 +422,8 @@ public CertificateDetailDto getCertificate(SecuredUUID uuid) throws NotFoundExce
...
+            dto.getCertificateRequest().setAltSignatureAttributes(attributeEngine.getObjectDataAttributesContent(ObjectAttributeContentInfo.builder(Resource.CERTIFICATE_REQUEST, certificate.getCertificateRequest().getUuid()).operation(AttributeOperation.CERTIFICATE_REQUEST_SIGN).purpose(AttributeContentPurpose.CERTIFICATE_REQUEST_ALT_KEY).build()));
```
ivosh: Use AttributeOperation.SIGN. CERTIFICATE_REQUEST_SIGN is legacy, for database migrations only.

ivosh:
> In this PR, new method `getConnectorEntity` was added to `ConnectorService`. No `@ExternalAuthorization` was used for this method as `getConnector` also does not have it. Is this expected? @ivosh or @lubomirw

@LukasNajman I do not see any consumers for `getConnectorEntity`. Please double-check. If confirmed, I recommend to not include `getConnectorEntity` in this PR, to keep its scope manageable. Anyway, this method won't be probably needed after introduction of connector cache.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
LukasNajman: Fixed

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
LukasNajman: Fixed

LukasNajman:
> > In this PR, new method `getConnectorEntity` was added to `ConnectorService`. No `@ExternalAuthorization` was used for this method as `getConnector` also does not have it. Is this expected? @ivosh or @lubomirw
> 
> @LukasNajman I do not see any consumers for `getConnectorEntity`. Please double-check. If confirmed, I recommend to not include `getConnectorEntity` in this PR, to keep its scope manageable. Anyway, this method won't be probably needed after introduction of connector cache.

The method is used in
`SigningProfileServiceImpl`
  - Line 532
  - Line 562
  - Line 574
  - Line 698

I was addressing issues copilot has raised. He has highlighted the fact we are accessing the entities directly through `Repository` without any permissions checks. I replaced the use of repositories with `Service`s.

ivosh:
> > > In this PR, new method `getConnectorEntity` was added to `ConnectorService`. No `@ExternalAuthorization` was used for this method as `getConnector` also does not have it. Is this expected? @ivosh or @lubomirw
> > 
> > 
> > @LukasNajman I do not see any consumers for `getConnectorEntity`. Please double-check. If confirmed, I recommend to not include `getConnectorEntity` in this PR, to keep its scope manageable. Anyway, this method won't be probably needed after introduction of connector cache.
> 
> The method is used in `SigningProfileServiceImpl`
> 
> * Line 532
> * Line 562
> * Line 574
> * Line 698
> 
> I was addressing issues copilot has raised. He has highlighted the fact we are accessing the entities directly through `Repository` without any permissions checks. I replaced the use of repositories with `Service`s.

Cool, thanks.
I tend to agree with Copilot that at leat the authorization to get the connector should be checked. Signing profile creation itself does not check any such permissions, so placing @ExternalAuthorization on `getConnectorEntity` seems  the right choice.

src/main/resources/db/migration/V202605051246__signing.sql
```diff
@@ -0,0 +1,83 @@
...
+-- ── 2. signing_profile_version
+CREATE TABLE "signing_profile_version"
+(
+    "uuid"                               UUID    NOT NULL,
+    "signing_profile_uuid"               UUID    NOT NULL,
+    "version"                            INTEGER NOT NULL,
+    -- scheme (authoritative)
+    "signing_scheme"                     VARCHAR NOT NULL,
+    "managed_signing_type"               VARCHAR,
+    "token_profile_uuid"                 UUID    REFERENCES "token_profile" ("uuid") ON DELETE SET NULL,
```
ivosh: Check all the subsequent FK `DELETE SET NULL`  on this table - if they are set to NULL we can no longer validate and answer auditing questions.

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,834 @@
...
+        return signingProfileRepository.findAllNames();
+    }
+
+    // ──────────────────────────────────────────────────────────────────────────
+    // Get (with optional version)
+    // ──────────────────────────────────────────────────────────────────────────
+
+    @Override
+    @ExternalAuthorization(resource = Resource.SIGNING_PROFILE, action = ResourceAction.DETAIL)
+    public SigningProfile getSigningProfileEntity(SecuredUUID uuid) throws NotFoundException {
```
ivosh, line 234: @Transactional is missing here?

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,834 @@
...
+    @Override
+    @ExternalAuthorization(resource = Resource.SIGNING_PROFILE, action = ResourceAction.ANY)
+    @Transactional(readOnly = true)
+    public List<BaseAttribute> listSignatureFormatterConnectorAttributes(UUID connectorUuid, SecuredUUID signingProfileUuid) throws NotFoundException, ConnectorException, AttributeException {
+        return fetchAndUpdateFormatterAttributeDefinitions(connectorUuid);
+    }
+
+    @Override
+    @ExternalAuthorization(resource = Resource.SIGNING_PROFILE, action = ResourceAction.LIST)
+    public List<String> findAllNames() {
```
ivosh, line 223: @Transactional is missing here?

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,834 @@
...
+                        .operation(AttributeOperation.SIGN)
+                        .version(version.getVersion()).build());
+        return List.of();
+    }
+
+    private List<ResponseAttribute> persistSignatureFormatterConnectorAttributes(SigningProfile p, SigningProfileVersion version, WorkflowRequestDto workflow)
+            throws AttributeException, ConnectorException, NotFoundException {
+        return switch (workflow) {
+            case ContentSigningWorkflowRequestDto w -> {
+                List<BaseAttribute> formatterDefinitions = fetchAndUpdateFormatterAttributeDefinitions(w.getSignatureFormatterConnectorUuid());
```
ivosh: An HTTP call to an external system held inside a database transaction. Refer to CLAUDE.md for details and why it's an anti-pattern.

Fix:
1. read/validate state + prefetch external data - no transaction
2. open transaction, write data, commit

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,834 @@
...
+            case RawSigningWorkflowRequestDto w -> {
+                // Raw signing has no formatter; clean up any formatter attributes that may remain for this version.
+                attributeEngine.deleteOperationObjectAttributesContent(AttributeType.DATA,
+                        ObjectAttributeContentInfo.builder(Resource.SIGNING_PROFILE, p.getUuid())
+                                .operation(AttributeOperation.WORKFLOW_FORMATTER)
+                                .version(version.getVersion()).build());
+                yield null;
+            }
+            case TimestampingWorkflowRequestDto w -> {
+                List<BaseAttribute> formatterDefinitions = fetchAndUpdateFormatterAttributeDefinitions(w.getSignatureFormatterConnectorUuid());
```
ivosh: An HTTP call to an external system held inside a database transaction. Refer to CLAUDE.md for details and why it's an anti-pattern.

Fix:
1. read/validate state + prefetch external data - no transaction
2. open transaction, write data, commit

ivosh (changes requested):

src/main/java/com/czertainly/core/service/SigningProfileService.java
```diff
@@ -0,0 +1,78 @@
...
+
+    PaginationResponseDto<SigningProfileListDto> listSigningProfiles(SearchRequestDto request, SecurityFilter filter);
+
+    List<SimplifiedSigningProfileDto> listSigningProfilesAssociatedTimeQualityConfiguration(SecuredUUID timeQualityConfigurationUuid, SecurityFilter filter);
+
+    SecuredList<SigningProfile> listSigningProfileEntitiesAssociatedTimeQualityConfiguration(SecuredUUID timeQualityConfigurationUuid, SecurityFilter filter);
+
+    SecuredList<SigningProfile> listSigningProfilesAssociatedWithTsp(SecuredUUID tspProfileUuid, SecurityFilter filter);
+
+    void notifyTimeQualityConfigurationChange(UUID timeQualityConfigurationUuid);
```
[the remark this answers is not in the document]
LukasNajman: This is intentional. The implementation that will use it will come in next PR.

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,834 @@
...
+            case ContentSigningWorkflowRequestDto w -> {
+                if (w.getSignatureFormatterConnectorUuid() == null) {
+                    throw new ValidationException("Signature formatter connector is required for content signing workflow");
+                }
+                Connector contentConnector =
+                        connectorService.getConnectorEntity(SecuredUUID.fromUUID(w.getSignatureFormatterConnectorUuid()));
+                validateFormatterConnectorFeature(contentConnector, FeatureFlag.CONTENT_SIGNING, SigningWorkflowType.CONTENT_SIGNING);
+                version.setSignatureFormatterConnector(contentConnector);
+            }
+            case RawSigningWorkflowRequestDto w -> {
```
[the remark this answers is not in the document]
LukasNajman: Unnamed patters are not available in Java21. Will name the variable `ignored`.

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,834 @@
...
+                List<BaseAttribute> formatterDefinitions = fetchAndUpdateFormatterAttributeDefinitions(w.getSignatureFormatterConnectorUuid());
+                attributeEngine.validateUpdateDataAttributes(w.getSignatureFormatterConnectorUuid(), AttributeOperation.WORKFLOW_FORMATTER, formatterDefinitions, w.getSignatureFormatterConnectorAttributes());
+                yield attributeEngine.replaceObjectDataAttributesContent(
+                        ObjectAttributeContentInfo.builder(Resource.SIGNING_PROFILE, p.getUuid())
+                                .connector(w.getSignatureFormatterConnectorUuid())
+                                .operation(AttributeOperation.WORKFLOW_FORMATTER)
+                                .version(version.getVersion()).build(),
+                        w.getSignatureFormatterConnectorAttributes());
+            }
+            case RawSigningWorkflowRequestDto w -> {
```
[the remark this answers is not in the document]
LukasNajman: Unnamed patters are not available in Java21. Will name the variable `ignored`.

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
LukasNajman: Fixed

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
LukasNajman: Fixed

src/main/resources/db/migration/V202605051246__signing.sql
LukasNajman: After discussion with @ivosh, we have decided to leave it as is. If this will be a problem in the future, we can find a way to address the concern.

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,853 @@
...
+            case ContentSigningWorkflowRequestDto w -> {
+                if (w.getSignatureFormatterConnectorUuid() == null) {
+                    throw new ValidationException("Signature formatter connector is required for content signing workflow");
+                }
+                Connector contentConnector =
+                        connectorService.getConnectorEntity(SecuredUUID.fromUUID(w.getSignatureFormatterConnectorUuid()));
+                validateFormatterConnectorFeature(contentConnector, FeatureFlag.CONTENT_SIGNING, SigningWorkflowType.CONTENT_SIGNING);
+                version.setSignatureFormatterConnector(contentConnector);
+            }
+            case RawSigningWorkflowRequestDto ignored -> {
```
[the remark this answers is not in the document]
LukasNajman: This is not supported in java 21!

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
LukasNajman: HTTP call refactored out of the DB transaction.

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
LukasNajman: HTTP call refactored out of the DB transaction.

ivosh (dismissed):

src/main/java/com/czertainly/core/dao/entity/signing/SigningProfileVersion.java
```diff
@@ -0,0 +1,169 @@
...
+        this.delegatedSignerConnectorUuid = delegatedSignerConnector != null ? delegatedSignerConnector.getUuid() : null;
+    }
+
+    public void setSignatureFormatterConnector(Connector signatureFormatterConnector) {
+        this.signatureFormatterConnector = signatureFormatterConnector;
+        this.signatureFormatterConnectorUuid = signatureFormatterConnector != null ? signatureFormatterConnector.getUuid() : null;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
lubomirw, line 144: **! Major** | Maintainability

`equals` and `hashCode` include lazy `@ManyToOne` references (`signingProfile`, `tokenProfile`, `certificate`, `raProfile`, `delegatedSignerConnector`, `signatureFormatterConnector`) alongside their `*_uuid` columns.

This has three concrete problems: (1) computing `hashCode` forces every lazy proxy to initialize, defeating `FetchType.LAZY`; (2) outside an open Hibernate session (e.g. an entity placed in a detached collection, or compared after `tx.commit()`), the proxy access throws `LazyInitializationException`; (3) the same logical row taken from two different sessions will not compare equal because the proxies are different instances. Compare only the `*Uuid` scalar columns — they uniquely identify the same target and are already stored on this entity, so the FK navigation field is pure noise in equality.

src/main/resources/db/migration/V202605051246__signing.sql
lubomirw: **! Major** | Correctness

Question on intent: every FK on `signing_profile_version` (`token_profile_uuid`, `certificate_uuid`, `ra_profile_uuid`, `delegated_signer_connector_uuid`, `signature_formatter_connector_uuid`) is declared `ON DELETE SET NULL`. Each of these fields is **operationally required** for its scheme/workflow — a STATIC_KEY scheme with no certificate, an ONE_TIME_KEY with no token profile, a CONTENT_SIGNING workflow with no formatter connector etc. can't sign. So `SET NULL` doesn't just hurt history on older versions; if it nulls a column on the row pointed to by `signing_profile.latest_version`, the currently-active signing profile silently breaks — the admin only discovers it on the next signing attempt, with a confusing downstream error.

Two precedents in this codebase point in different directions:

- `notification_profile_version.notification_instance_ref_uuid` → `ON DELETE RESTRICT` (block delete while referenced; force the admin to detach or delete the profile first).
- `approval_step.user_uuid` / `role_uuid` / `group_uuid` → no FK at all (deliberately accept silent history drift).

What's the intent here? If the goal is "operator must remediate before the source entity disappears" (the notification-profile model), `ON DELETE RESTRICT` matches and gives a clear error at the offending delete. If the goal is "deletions of source entities should always succeed and we'll track breakage elsewhere", dropping the FK (the approval-step model) is more honest than `SET NULL` and avoids the operational footgun of a half-broken latest version. `SET NULL` falls between the two: it leaves a passing-looking row that doesn't actually work.

Could you note in the migration which intent applies, and align the constraint accordingly?

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -0,0 +1,853 @@
...
+    SigningProfileDto persistCreate(SigningProfileRequestDto request, List<BaseAttribute> formatterDefinitions)
+            throws AttributeException, NotFoundException {
+        SigningProfile profile = new SigningProfile();
+        profile.setName(request.getName());
+        profile.setDescription(request.getDescription());
+        profile.setLatestVersion(1);
+        profile.setSigningScheme(request.getSigningScheme().getSigningScheme());
+        profile.setWorkflowType(request.getWorkflow().getType());
+        profile = signingProfileRepository.save(profile);
+
```
lubomirw, line 278: **. Nitpick** | Performance

`signingProfileRepository.save(profile)` is called twice in `persistCreate` — once before `applyWorkflow` / `applyScheme`, once after. The first save persists the cache columns (`signingScheme`, `workflowType`, `timeQualityConfiguration`) that then get immediately overwritten in-memory by the apply methods, so it's just churn. A single save after the apply calls should be enough — Hibernate assigns the UUID on `persist` before flush, so `v1.setSigningProfile(profile)` still gets a non-null UUID for the FK.

lubomirw (changes requested):

src/main/java/com/czertainly/core/dao/entity/signing/SigningProfileVersion.java
lubomirw, line 144: **! Major** | Maintainability

`equals` and `hashCode` include lazy `@ManyToOne` references (`signingProfile`, `tokenProfile`, `certificate`, `raProfile`, `delegatedSignerConnector`, `signatureFormatterConnector`) alongside their `*_uuid` columns.

This has three concrete problems: (1) computing `hashCode` forces every lazy proxy to initialize, defeating `FetchType.LAZY`; (2) outside an open Hibernate session (e.g. an entity placed in a detached collection, or compared after `tx.commit()`), the proxy access throws `LazyInitializationException`; (3) the same logical row taken from two different sessions will not compare equal because the proxies are different instances. Compare only the `*Uuid` scalar columns — they uniquely identify the same target and are already stored on this entity, so the FK navigation field is pure noise in equality.

src/main/resources/db/migration/V202605051246__signing.sql
lubomirw: **! Major** | Correctness

Question on intent: every FK on `signing_profile_version` (`token_profile_uuid`, `certificate_uuid`, `ra_profile_uuid`, `delegated_signer_connector_uuid`, `signature_formatter_connector_uuid`) is declared `ON DELETE SET NULL`. Each of these fields is **operationally required** for its scheme/workflow — a STATIC_KEY scheme with no certificate, an ONE_TIME_KEY with no token profile, a CONTENT_SIGNING workflow with no formatter connector etc. can't sign. So `SET NULL` doesn't just hurt history on older versions; if it nulls a column on the row pointed to by `signing_profile.latest_version`, the currently-active signing profile silently breaks — the admin only discovers it on the next signing attempt, with a confusing downstream error.

Two precedents in this codebase point in different directions:

- `notification_profile_version.notification_instance_ref_uuid` → `ON DELETE RESTRICT` (block delete while referenced; force the admin to detach or delete the profile first).
- `approval_step.user_uuid` / `role_uuid` / `group_uuid` → no FK at all (deliberately accept silent history drift).

What's the intent here? If the goal is "operator must remediate before the source entity disappears" (the notification-profile model), `ON DELETE RESTRICT` matches and gives a clear error at the offending delete. If the goal is "deletions of source entities should always succeed and we'll track breakage elsewhere", dropping the FK (the approval-step model) is more honest than `SET NULL` and avoids the operational footgun of a half-broken latest version. `SET NULL` falls between the two: it leaves a passing-looking row that doesn't actually work.

Could you note in the migration which intent applies, and align the constraint accordingly?

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
lubomirw, line 278: **. Nitpick** | Performance

`signingProfileRepository.save(profile)` is called twice in `persistCreate` — once before `applyWorkflow` / `applyScheme`, once after. The first save persists the cache columns (`signingScheme`, `workflowType`, `timeQualityConfiguration`) that then get immediately overwritten in-memory by the apply methods, so it's just churn. A single save after the apply calls should be enough — Hibernate assigns the UUID on `persist` before flush, so `v1.setSigningProfile(profile)` still gets a non-null UUID for the FK.

lubomirw (changes requested):

lubomirw (commented):
Follow-up: one more observation.

src/main/java/com/czertainly/core/dao/entity/signing/SigningProfile.java
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Denormalized cache column — mirrors the latest {@link SigningProfileVersion#workflowType}.
+     * Updated on every create/update alongside the version row.
+     */
+    @Column(name = "workflow_type", nullable = false)
+    @Enumerated(EnumType.STRING)
+    private SigningWorkflowType workflowType;
+
+    @Column(name = "latest_version", nullable = false)
+    private Integer latestVersion = 1;
```
lubomirw: **~ Minor** | Maintainability

This `Integer latestVersion` (and `Integer version` at `SigningProfileVersion.java:46`) would be better as primitive `int`. Both columns are declared `NOT NULL` with a default in the migration, and the field has `= 1` initialisation — null is never the intended state. The established codebase convention matches: `ApprovalProfileVersion.version` uses `private int version = 1;` and `NotificationProfile.versionLock` uses `private int versionLock;` for the same NOT NULL counter pattern. `Integer` is reserved for genuinely nullable columns (e.g. `ApprovalProfileVersion.expiry`).

Two small concrete payoffs from switching to `int`:

1. `profile.setLatestVersion(null)` becomes a compile error instead of a DB constraint failure at flush time.
2. The increment in `persistUpdate` — `int nextVersion = profile.getLatestVersion() + 1;` — is currently a hidden unboxing site. If `latestVersion` is ever null (custom JPQL projection, partially-hydrated entity from a future query), this NPEs at autounbox. Primitive `int` removes the risk by construction.

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
LukasNajman: Actually, the claim `v1.setSigningProfile(profile) still gets a non-null UUID for the FK` is not true. When the signing profile is not saved, the `setSigningProfile` will read UUID of `NULL` and save it to local variable. It does not get replaced with real UUID on save.

However, I have restructured the code so a single save if necessary.

src/main/java/com/czertainly/core/dao/entity/signing/SigningProfile.java
LukasNajman: Changed to int

src/main/resources/db/migration/V202605051246__signing.sql
LukasNajman: To be on the safe side, I agree, we should change this to `ON DELETE RESTRICT`. However, I see another issue here. The concept of versions kind of implies the version is immutable. But it points to mutable references (RA Profile, Token Profile) which may "silently" mutate the `SigningProfileVersion` when changed.

I propose to restrict delete for now and create new task that will address the mutability of the version.

@lubomirw and @ivosh do you agree?

ivosh: I certainly agree to do it in a follow-up PR, consistently with the rest of the codebase.

src/main/java/com/czertainly/core/dao/entity/signing/SigningProfileVersion.java
LukasNajman: Fixed, the `equals` and `hashCode`  now delegate to `equals` and `hashCode` of `UniquelyIdentifiedAndAudited`

src/main/resources/db/migration/V202605051246__signing.sql
lubomirw: I agree.

lubomirw (approved):

PR #1481, opened 2026-05-19
Split InternalService and ExternalService - part 3

ivosh:


src/main/java/com/czertainly/core/service/impl/NotificationInstanceServiceImpl.java
```diff
@@ -211,6 +212,7 @@ public void deleteNotificationInstance(UUID uuid) throws NotFoundException {
     }
 
     @Override
+    @ExternalAuthorizationMissing
     public List<DataAttribute> listMappingAttributes(String connectorUuid, String kind) throws ConnectorException, NotFoundException {
         ConnectorDto connector = connectorService.getConnector(SecuredUUID.fromString(connectorUuid));
         return connectorApiFactory.getNotificationInstanceApiClient(connector).listMappingAttributes(connector, kind);
```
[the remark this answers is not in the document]
ivosh: Exactly. This will be done in a follow-up PR.

src/main/java/com/czertainly/core/service/impl/CallbackServiceImpl.java
```diff
@@ -201,6 +202,7 @@ private AttributeResource getAttributeResource(BaseAttribute attribute) {
     }
 
     @Override
+    @ExternalAuthorizationMissing
     public Object resourceCallback(Resource resource, String resourceUuid, RequestAttributeCallback callback) throws ConnectorException, ValidationException, NotFoundException, AttributeException {
```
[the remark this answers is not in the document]
ivosh: Exactly. This will be done in a follow-up PR.

src/main/java/com/czertainly/core/service/CallbackExternalService.java
```diff
@@ -0,0 +1,29 @@
...
+import com.czertainly.api.exception.ConnectorException;
+import com.czertainly.api.exception.NotFoundException;
+import com.czertainly.api.exception.ValidationException;
+import com.czertainly.api.model.common.attribute.common.callback.RequestAttributeCallback;
+import com.czertainly.api.model.core.auth.Resource;
+import com.czertainly.api.model.core.connector.FunctionGroupCode;
+
+import java.util.UUID;
+
+public interface CallbackExternalService {
```
lubomirw, line 13: **~ Minor** | Documentation

Method-level Javadoc was dropped when `CallbackService` was re-created as `CallbackExternalService` (and same for `TrustedCertificateService` → `TrustedCertificateExternalService`).

The other four interfaces in this PR were properly renamed and kept their content. For these two, the previous Javadoc carried non-obvious semantic distinctions that the method signatures alone don't convey — e.g., `callback(String, FunctionGroupCode, String, ...)` was documented as "executes the callback only for the attributes derived from the primary objects of the connector", while `callback(UUID, ...)` was "for the attributes defined by the connector". With three overloads named `callback` / `resourceCallback`, a caller can't tell them apart without reading the impl.

Per this repo's CLAUDE.md ("Method-level Javadoc on non-obvious behaviour over inline blocks"), this is the kind of doc worth keeping. Suggest carrying the prior Javadoc forward — or at least the one-liner that differentiates each overload.

lubomirw, line 13: **~ Minor** | Maintainability

These two interfaces (`CallbackExternalService` and `TrustedCertificateExternalService`) appear in git as `delete` + `add` rather than a rename, while the other four interfaces in this PR are properly tracked as renames (similarity 87–96%). The cause is that these two lost both their Javadoc and their `interface` declaration line, dropping below git's default 50% similarity threshold.

The practical impact: `git log` / `git blame` on `CallbackExternalService.java` and `TrustedCertificateExternalService.java` will not see history before this PR without `--follow`, and even `--follow` is unreliable across delete+add. Restoring the Javadoc (see the other comment) would likely also restore rename detection here, killing two birds with one stone.

lubomirw (commented):

src/main/java/com/czertainly/core/service/impl/AuthServiceImpl.java
```diff
@@ -61,11 +63,13 @@ public UserProfileDetailDto getAuthProfile() {
     }
 
     @Override
+    @SelfPrincipalEndpoint
```
lubomirw: **~ Minor** | Correctness

`@SelfPrincipalEndpoint` may be semantically incorrect here — the method returns a system-wide catalog, not data about the principal.

The annotation's Javadoc states: *"the subject of the operation is the principal itself"* and gives `getAuthProfile()` as the example. `getAuthResources()` returns the full list of authorization resources known to the system — the same response for every caller — so the subject is the resource catalog, not the principal. The other two methods you annotated (`getAuthProfile`, `updateUserProfile`) are genuine self-principal operations; this one isn't.

Suggestion: introduce a new permanent annotation — for example `@AnyPrincipalEndpoint` (alternatives: `@AnyAuthenticatedEndpoint`, `@AuthenticatedOnlyEndpoint`) — to mark endpoints that require an authenticated caller but do not need any resource-level authorization and are not personalized to the principal. Adding it to the set of four permanent annotations referenced by `@ExternalAuthorizationMissing`'s Javadoc would let callers like `getAuthResources()` declare intent correctly, instead of overloading `@SelfPrincipalEndpoint` with a meaning it does not carry.

lubomirw (commented):
Code review: 1 issue found.

src/main/java/com/czertainly/core/service/CallbackExternalService.java
ivosh: Fixed.

ivosh: Fixed.

src/main/java/com/czertainly/core/service/impl/AuthServiceImpl.java
ivosh: Fixed.

src/test/resources/archunit_store/dc319b8f-4aa8-4995-960d-775b2144f2a2
```diff
@@ -1,7 +1,9 @@
...
 Method <com.czertainly.core.service.impl.ConnectorAuthServiceImpl.validateApiKeyAuthAttributes(java.util.List)> is annotated with @ExternalAuthorizationMissing in (ConnectorAuthServiceImpl.java:306)
 Method <com.czertainly.core.service.impl.ConnectorAuthServiceImpl.validateBasicAuthAttributes(java.util.List)> is annotated with @ExternalAuthorizationMissing in (ConnectorAuthServiceImpl.java:117)
 Method <com.czertainly.core.service.impl.ConnectorAuthServiceImpl.validateCertificateAttributes(java.util.List)> is annotated with @ExternalAuthorizationMissing in (ConnectorAuthServiceImpl.java:223)
 Method <com.czertainly.core.service.impl.ConnectorAuthServiceImpl.validateJWTAuthAttributes(java.util.List)> is annotated with @ExternalAuthorizationMissing in (ConnectorAuthServiceImpl.java:319)
+Method <com.czertainly.core.service.impl.NotificationInstanceServiceImpl.listMappingAttributes(java.lang.String, java.lang.String)> is annotated with @ExternalAuthorizationMissing in (NotificationInstanceServiceImpl.java:217)
```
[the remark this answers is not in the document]
ivosh: The proper categorization will be done in a follow-up PR.

lubomirw (approved):

PR #1485, opened 2026-05-20
Implement ResourceExtensionService for Approval

### LINKED ISSUE #1486: Events - Event History: HTTP 501 returned when retrieving event history for Approval resource objects
## Description
The Event History API endpoint (`GET /v1/resource-events/{event}/history`) returns **HTTP 501 Not Supported** when called for an `APPROVAL` resource object. The root cause is that `ApprovalServiceImpl` does not implement `ResourceExtensionService`, so `ResourceServiceImpl.getResourceObject(Resource.APPROVAL, uuid)` finds no registered extension service for the `approval` resource code and throws `NotSupportedException`.

The `EventServiceImpl.getEventHistory` method calls `resourceService.getResourceObject(resource, uuid)` as an existence and permission check before querying the event history. This call unconditionally fails for any approval UUID, making the entire Event History feature non-functional for the Approval resource.

## Steps to Reproduce
1. Ensure at least one Approval exists in the system.
2. Call `GET /v1/resource-events/APPROVAL_CREATED/history` (or any event scoped to `APPROVAL`) with a valid approval UUID as the resource filter.
3. Observe the response.

## Expected Result
- API returns HTTP 200 with paginated event history records for the given Approval object.

## Actual Result
- API returns HTTP 501 Not Supported: `"Cannot retrieve object for requested resource: Approval"`

## Environment

ILM Core:
Commit:
Branch:
Build time:

Client:
- OS:
- Browser:

klaraf755:
Fixes #1486 
Depends on https://github.com/OmniTrustILM/interfaces/pull/680


src/main/java/com/czertainly/core/dao/repository/SecurityFilterRepositoryImpl.java
```diff
@@ -170,14 +170,20 @@ private CriteriaQuery<NameAndUuidDto> createResourceObjectsQuery(SecurityFilter
...
                 cb.construct(
                         NameAndUuidDto.class,
                         root.get("uuid"),
-                        root.get(nameAttribute)
+                        nameExpression
                 )
         );
-        query.orderBy(cb.asc(root.get(nameAttribute)));
+        if (nameAttribute != null) {
+            query.orderBy(cb.asc(root.get(nameAttribute)));
```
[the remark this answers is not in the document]
klaraf755: Added

src/main/java/com/czertainly/core/service/impl/ApprovalServiceImpl.java
```diff
@@ -371,4 +374,26 @@ public void setEventProducer(EventProducer eventProducer) {
...
     }
+
+    @Override
+    public NameAndUuidDto getResourceObjectInternal(UUID objectUuid) throws NotFoundException {
+        return approvalRepository.findResourceObject(objectUuid, null);
+    }
+
+    @Override
+    public NameAndUuidDto getResourceObjectExternal(SecuredUUID objectUuid) throws NotFoundException {
+        return approvalRepository.findResourceObject(objectUuid.getValue(), null);
```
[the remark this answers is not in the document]
klaraf755: Since Approval is also identified in listing on FE only by UUID, I think the null name is OK here.

src/main/java/com/czertainly/core/service/impl/ApprovalServiceImpl.java
```diff
@@ -371,4 +374,26 @@ public void setEventProducer(EventProducer eventProducer) {
...
         this.approvalRecipientHelper = approvalRecipientHelper;
     }
+
+    @Override
+    public NameAndUuidDto getResourceObjectInternal(UUID objectUuid) throws NotFoundException {
+        return approvalRepository.findResourceObject(objectUuid, null);
+    }
+
+    @Override
+    public NameAndUuidDto getResourceObjectExternal(SecuredUUID objectUuid) throws NotFoundException {
```
lubomirw, line 399: **! Major** | Security

`getResourceObjectExternal` missing `@ExternalAuthorization` annotation — bypasses ext-auth check on the external code path.

Every sibling impl annotates this method: `CertificateServiceImpl`, `AcmeProfileServiceImpl`, `AcmeAccountServiceImpl` all use `@ExternalAuthorization(resource = Resource.X, action = ResourceAction.DETAIL)`. Without it here, `ResourceServiceImpl.getResourceObject(Resource.APPROVAL, uuid)` returns approval data to any authenticated caller, regardless of their permission on `APPROVAL`. The bean is the only one Spring resolves for code `approval`, so this path is now reachable platform-wide (event history, custom-attribute APIs, etc). Add `@ExternalAuthorization(resource = Resource.APPROVAL, action = ResourceAction.DETAIL)` to match the pattern, or justify in a comment why approvals deliberately skip the gate.

src/main/java/com/czertainly/core/service/impl/ApprovalServiceImpl.java
```diff
@@ -371,4 +374,26 @@ public void setEventProducer(EventProducer eventProducer) {
     public void setApprovalRecipientHelper(ApprovalRecipientHelper approvalRecipientHelper) {
         this.approvalRecipientHelper = approvalRecipientHelper;
     }
+
+    @Override
+    public NameAndUuidDto getResourceObjectInternal(UUID objectUuid) throws NotFoundException {
```
lubomirw, line 393: **~ Minor** | Maintainability

All three lookup methods pass `nameAttribute = null`, so every `NameAndUuidDto` returned for approvals has `name = null`. Downstream consumers — Event History UI, resource picker dropdowns, audit-log enrichment — display this name field to operators; rows will render blank or "null".

Approvals have no human-friendly name column, but a synthetic descriptor would help. Options: format as `"Approval <short-uuid>"`, or derive from `resource + " / " + objectUuid` (the action target), or use `createdAt` timestamp. Cleanest is probably a dedicated DB-projected expression in a custom repository method rather than null-padding in the generic base. Worth deciding before this ships to the FE, since the FE will likely need a fallback otherwise.

lubomirw (changes requested):

src/main/java/com/czertainly/core/service/impl/ApprovalServiceImpl.java
klaraf755: Added, originally it was missing on getApprovalDetail method as well, but I see it has been added

src/main/java/com/czertainly/core/service/impl/ApprovalServiceImpl.java
klaraf755: Added projected expression

lubomirw (approved):

PR #1488, opened 2026-05-20
Keep event history when a follow-up notification fails

klaraf755:
Follow up PR to https://github.com/OmniTrustILM/core/pull/1437, addressing comment https://github.com/OmniTrustILM/core/pull/1437#pullrequestreview-4235468993


src/test/java/com/czertainly/core/events/EventHistoryTransactionTest.java
```diff
@@ -0,0 +1,95 @@
...
+        association.setEvent(ResourceEvent.DISCOVERY_FINISHED);
+        // resource and objectUuid left null → platform trigger, matched by fetchEventTriggers(ctx, null, null)
+        triggerAssociationRepository.save(association);
+
+        final UUID discoveryUuid = discovery.getUuid();
+        Assertions.assertThrows(RuntimeException.class, () ->
+                discoveryFinishedEventHandler.handleEvent(
+                        DiscoveryFinishedEventHandler.constructEventMessage(
+                                discoveryUuid, null, null,
+                                new DiscoveryResult(DiscoveryStatus.COMPLETED, "Test"))));
```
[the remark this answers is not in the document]
klaraf755: Added the assertion

src/main/java/com/czertainly/core/events/handlers/CertificateDiscoveredEventHandler.java
```diff
@@ -84,11 +82,6 @@ public void setDiscoveryCertificateHandler(CertificateHandler certificateHandler
         this.certificateHandler = certificateHandler;
     }
 
-    @Autowired
-    public void setTransactionHandler(TransactionHandler transactionHandler) {
```
ivosh, line 88: Check if import `com.czertainly.core.events.transaction.TransactionHandler` is now unused.

src/main/java/com/czertainly/core/events/handlers/ApprovalRequestedEventHandler.java
```diff
@@ -43,6 +43,12 @@ protected Object getEventData(Approval approval, Object eventMessageData) {
         return EventDataBuilder.getApprovalRequestedEventData(approval, approvalProfile, approvalStepDto, authHelper.getUserUsername(approval.getCreatorUuid().toString()));
     }
 
+    @Override
+    protected void prefetchForFollowUp(EventContext<Approval> context) {
```
ivosh: ## Problem
The scope of this PR is a follow-up, cleaning up and refactoring the existing technical debt.
We should not introduce more technical debt by keeping the original design where live JPA entities (Approval, DiscoveryHistory, Certificate) are carried out of the transaction inside
  EventContext, and sendFollowUpEventsNotifications then reads their associations after the transaction has closed. The consequence is that every subclass must override prefetchForFollowUp and manually
  re-list, by hand, exactly which lazy associations the sender will later touch.

Nothing enforces that the prefetch list stays in sync with what is later accessed — change one part, forget the prefetch, and you get a production-only lazy-load crash that no compiler catches. That's fragile by construction.

In summary: the next step consumes live entities instead of a self-contained value, so the transaction boundary has to be managed by hand. We will never get this right with the current design.


## Suggestion
  Core idea: stop carrying live entities past the transaction boundary. Build the self-contained notification payload inside the transaction (where lazy loading is free), publish it as an application event, and let
  @TransactionalEventListener(AFTER_COMMIT) do the JMS send. The transaction boundary becomes Spring's job, not hand-written orchestration.

  The flow becomes:
```
  handleEvent  (@Transactional REQUIRED)         ← back to a plain tx method
     ├─ prepareContext
     ├─ processAllTriggers          ← writes EventHistory, commits with the tx
     └─ publishFollowUpEvents(ctx)  ← runs IN-tx: builds NotificationMessage
                                       value objects + publishes them as events
          │
          └─ (tx commits)
               └─ @TransactionalEventListener(AFTER_COMMIT) → notificationProducer.produceMessage
```

ivosh (changes requested):

src/main/java/com/czertainly/core/events/handlers/CertificateDiscoveredEventHandler.java
klaraf755: Since I reverted the changes, TransactionHandler is needed in the class.

src/main/java/com/czertainly/core/events/handlers/ApprovalRequestedEventHandler.java
klaraf755: Changed the fix according to the suggestion.

src/test/java/com/czertainly/core/events/EventHistoryTransactionTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.mockito.Mockito;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.test.context.bean.override.mockito.MockitoBean;
+
+import java.util.List;
+import java.util.UUID;
+
+class EventHistoryTransactionTest extends BaseSpringBootTest {
+
+    @MockitoBean
```
ivosh, line 36: **! Major** | Testing

EventHistoryTransactionTest.java:34 mocks the whole `NotificationProducer` with `@MockitoBean` and stubs `produceMessage(...)` to throw (line 57-58). But the production fix routes notifications through `applicationEventPublisher.publishEvent(notificationMessage)` → the new `@TransactionalEventListener NotificationProducer.onNotificationMessage` (NotificationProducer.java:43-46). Because the bean is replaced by a Mockito mock, Spring's `EventListenerMethodProcessor` does not register `@TransactionalEventListener` on the mock subclass, so `publishEvent(...)` has no subscriber. The stubbed throw is never reached and the EventHistory assertions (lines 91-93) pass independently of the notification path.

This was confirmed empirically by running the test. So the exact regression this PR exists to prevent (notification failure silently dropping/rolling back EventHistory) would still leave this test green. This is the CLAUDE.md "green test that covers nothing" anti-pattern, and it is significant because validating this behaviour is the entire purpose of the PR.

Suggestion: do not mock the whole `NotificationProducer`. Either (a) mock the lower-level collaborator (`JmsTemplate` / the `sendMessage` path) so the real `@TransactionalEventListener` runs and actually throws, or (b) keep the real producer and assert via `Mockito.verify(...)` / an awaitility check that `produceMessage` was invoked AND that the EventHistory still committed despite the throw. Add an explicit `Mockito.verify(notificationProducer).produceMessage(any())` (or equivalent) so a silent-drop regression fails the test.

src/main/java/com/czertainly/core/messaging/jms/producers/NotificationProducer.java
```diff
@@ -39,6 +40,11 @@ private void sendMessage(final NotificationMessage notificationMessage) {
         });
     }
 
+    @TransactionalEventListener
```
ivosh: **~ Minor** | Error handling

NotificationProducer.java:43 uses `@TransactionalEventListener` with no attributes → default `phase = AFTER_COMMIT`, `fallbackExecution = false`. If `publishEvent(NotificationMessage)` is ever called with no active/committing transaction, the listener is silently skipped and the notification is lost with no log or error.

This is NOT a realized bug today: all 7 converted publish sites run inside a committing transaction — each handler is class-level `@Transactional` (REQUIRED), and the two `handleEvent` invokers (EventListener.java:35 and CertificateUploadServiceImpl.java:79) are `NOT_SUPPORTED` but call the handler on a separate proxied bean, so REQUIRED opens and commits a fresh transaction around the publish. So this finding is latent fragility, not data loss now.

It is worth hardening because the codebase's own established pattern is more defensive: `TriggerEvaluator.performSendNotificationAction` (TriggerEvaluator.java:441-450) explicitly checks `TransactionSynchronizationManager.isSynchronizationActive()` and falls back to a synchronous `produceMessage` when no transaction is active. The new listener has neither that fallback nor `fallbackExecution=true`, so any future publish site (or a refactor making an invoker non-transactional) would silently lose notifications. 

Consider `fallbackExecution = true` or a one-line comment documenting the invariant that every `publishEvent(NotificationMessage)` caller must hold a committing transaction.

src/main/java/com/czertainly/core/messaging/jms/producers/NotificationProducer.java
```diff
@@ -39,6 +40,11 @@ private void sendMessage(final NotificationMessage notificationMessage) {
...
+    public void onNotificationMessage(NotificationMessage notificationMessage) {
+        produceMessage(notificationMessage);
```
ivosh, line 46: **~ Minor** | Maintainability

After this PR the 7 event handlers dispatch notifications via `applicationEventPublisher.publishEvent(...)` → the new AFTER_COMMIT listener (NotificationProducer.java:43) → `produceMessage`.

The existing `TriggerEvaluator.performSendNotificationAction` (TriggerEvaluator.java:441-450) still calls `notificationProducer.produceMessage(...)` directly, hand-rolling the same afterCommit logic plus a no-transaction synchronous fallback. Two parallel paths for the same `NotificationMessage` type now exist. 

Could you check (taking into account comments at TriggerEvaluator.java:439-440) if `TriggerEvaluator.performSendNotificationAction` could use the same route through `publishEvent`?

src/test/java/com/czertainly/core/events/EventHistoryTransactionTest.java
```diff
@@ -0,0 +1,95 @@
...
+    @Autowired
+    private TriggerRepository triggerRepository;
+    @Autowired
+    private TriggerAssociationRepository triggerAssociationRepository;
+
+    /**
+     * Verifies the end-to-end fix: when sendFollowUpEventsNotifications throws after successful
+     * trigger processing, the EventHistory row must still be visible and carry FINISHED status —
+     * not rolled back with the enclosing handleEvent invocation, because handleEvent runs with
+     * Propagation.NOT_SUPPORTED, and each repository save commits its own small transaction.
```
ivosh: **~ Minor** | Documentation

The Javadoc states that `handleEvent` runs with `Propagation.NOT_SUPPORTED` and that each repository save commits its own small transaction.

But the test drives `DiscoveryFinishedEventHandler`, which does not override `handleEvent` and so inherits `EventHandler.handleEvent` (EventHandler.java:109) governed by the class-level `@Transactional` default REQUIRED (DiscoveryFinishedEventHandler.java:29, EventHandler.java:37). 

`NOT_SUPPORTED` actually lives only on `CertificateDiscoveredEventHandler.handleEvent` (CertificateDiscoveredEventHandler.java:134) — a different handler. The stated mechanism is therefore factually wrong for the handler under test. 

Either retarget the test at a NOT_SUPPORTED handler if that is the boundary you mean to cover, or correct the comment to describe the REQUIRED transaction that actually wraps this handler.

src/main/java/com/czertainly/core/messaging/jms/producers/NotificationProducer.java
ivosh: **. Nitpick** | Maintainability

NotificationProducer.java:43 writes `@TransactionalEventListener` with no attributes. Other transactional listeners in the codebase (e.g. CertificateEventHistoryServiceImpl.java:81, CertificateHandler/TimeQualityRegisterImpl) spell out `phase = TransactionPhase.AFTER_COMMIT`. The default is already AFTER_COMMIT so behaviour is identical, but stating it explicitly matches the house style and makes the post-commit semantics obvious at the call site. For your consideration.

ivosh (changes requested):
Thank you for the new design! I hope it was worth spending a little more time to get the transaction boundary right :-) I like the fix much more now.

There are only few loose ends to tie - nothing critical.

src/test/java/com/czertainly/core/events/EventHistoryTransactionTest.java
klaraf755: Changed the test to mock the JmsTemplate.

src/main/java/com/czertainly/core/messaging/jms/producers/NotificationProducer.java
klaraf755: I added `fallbackExecution = true`.

src/test/java/com/czertainly/core/events/EventHistoryTransactionTest.java
klaraf755: I changed the comment,previous comment was referencing the original fix.

src/main/java/com/czertainly/core/messaging/jms/producers/NotificationProducer.java
klaraf755: Updated it with the phase.

src/main/java/com/czertainly/core/messaging/jms/producers/NotificationProducer.java
klaraf755: I tested the trigger evaluator with `publishEvent` and the trigger history record got created, so the published event can also be used here.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
```diff
@@ -436,18 +434,9 @@ protected void performSendNotificationAction(Resource resource, ResourceEvent ev
...
 
         NotificationMessage message = new NotificationMessage(event, resource, object.getUuid(), notificationProfileUuids, null, data, triggerHistory.getUuid(), execution.getUuid());
-        // Delay publish until after the current transaction commits so TriggerHistory is
+        // Delay publication until after the current transaction commits, so TriggerHistory is
         // visible to the NotificationListener when it creates TriggerHistoryRecords.
-        if (TransactionSynchronizationManager.isSynchronizationActive()) {
-            TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
-                @Override
-                public void afterCommit() {
-                    notificationProducer.produceMessage(message);
```
ivosh, line 445: **~ Minor** | Maintainability

This PR switches all 7 follow-up handlers from notificationProducer.produceMessage(...) to applicationEventPublisher.publishEvent(...). After the change, no subclass of EventHandler references notificationProducer anymore. That leaves the injected field `protected NotificationProducer notificationProducer;` and the @Autowired setNotificationProducer(...) setter at EventHandler.java:45 and :75-77 as dead injection. Harmless (still compiles, bean resolves) but misleading leftover state.

Please remove the unused field+setter from base EventHandler; NotificationProducer bean stays because produceInternalNotificationMessage is still used elsewhere.

ivosh (changes requested):

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
klaraf755: Done

ivosh (approved):

PR #1491, opened 2026-05-20
Parallelize service and non-service test jobs in build_pr

### LINKED ISSUE #1490: Build performance improvements
### Description

Improvements for the build infrastructure

ivosh:
Tests take ~14 minutes in the CI pipeline and 7:30 minutes on my local PC (MacBook with M5 silicon).
The split tests take roughly 3:50 each on my local PC - there is a small startup overhead paid two times. But still, the overall time is halved. I hope the same performance boost can be observed also for the CI run.

No worries about billing - CI minutes are free for public repos.


src/test/java/com/czertainly/core/architecture/CiTestSplitTest.java
```diff
@@ -0,0 +1,114 @@
...
+                .isEmpty();
+    }
+
+    private static boolean containsTestAnnotation(Path file) {
+        try {
+            String source = Files.readString(file);
+            return source.contains("@Test") || source.contains("@ParameterizedTest");
+        } catch (IOException e) {
+            return false;
+        }
```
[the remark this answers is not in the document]
ivosh: Fixed in 67dd6236.

.github/workflows/build_pr.yml
```diff
@@ -8,9 +8,71 @@ concurrency:
...
+          name: compiled
+
+      - name: Extract compiled output
+        run: tar -xzf compiled.tar.gz
+
+      - name: Run service tests
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
+        run: mvn -B -ntp test -Ptest-services -Dmaven.compiler.skip=true
+
```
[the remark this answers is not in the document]
ivosh: Fixed in 67dd6236.

src/test/java/com/czertainly/core/architecture/CiTestSplitTest.java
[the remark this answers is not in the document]
ivosh: Fixed in a19f7c39.

.github/workflows/build_pr.yml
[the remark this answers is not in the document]
ivosh: Fixed in a19f7c39.

src/test/java/com/czertainly/core/architecture/CiTestSplitTest.java
```diff
@@ -0,0 +1,113 @@
...
+                        They are not picked up by the test-services Maven profile and run in test-rest instead.
+                        Either rename them to match the pattern, or update both <includes> in test-services
+                        and <excludes> in test-rest in pom.xml to cover the new suffix.""")
+                .isEmpty();
+    }
+
+    private static boolean containsTestAnnotation(Path file) {
+        try {
+            String source = Files.readString(file);
+            return source.contains("@Test") || source.contains("@ParameterizedTest");
```
lubomirw: **! Major** | Correctness

`containsTestAnnotation` only detects `@Test` and `@ParameterizedTest`, missing `@RepeatedTest`, `@TestFactory`, and `@TestTemplate`.

A class in `service/` named e.g. `FooBehavior.java` that uses only `@RepeatedTest` would be filtered out of the violation list and silently run in `test-rest` instead of `test-services`, defeating the guard this test is meant to provide. Consider broadening the check to all JUnit 5 test method annotations, or adding `@TestFactory|@RepeatedTest|@TestTemplate` to the matched set so the guard catches every runnable test method type.

.github/workflows/build_pr.yml
```diff
@@ -24,32 +51,134 @@ jobs:
...
+          path: target/jacoco.exec
+
+  test-rest:
+    name: Test (rest)
+    needs: compile
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
```
lubomirw: **~ Minor** | Performance

The "Shallow clones should be disabled for a better relevancy of analysis" comment is a SonarQube convention carried over from the pre-split job. Since SonarQube runs in a separate workflow (triggered after `build_pr` completes) and does its own checkout there, none of the jobs in `build_pr.yml` actually need full history.

Dropping `fetch-depth: 0` from `test-services`, `test-rest`, and `package` (it appears in all three) and removing the now-misleading comment would give each runner a faster, lighter checkout. The `compile` job already does the right thing (no `fetch-depth` override).

pom.xml
```diff
@@ -468,4 +468,82 @@
...
+                            </includes>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
+        <!-- Run all tests except service/** (used by CI test-rest job) -->
+        <profile>
+            <id>test-rest</id>
```
lubomirw: **~ Minor** | Maintainability

Quick question, no strong opinion — would `test-non-services` (or `test-other` / `test-remainder`) be more readable than `test-rest`?

`test-rest` reads naturally as "REST endpoint tests", which is a plausible meaning in a codebase that has REST controllers and REST clients. The intent here is "the rest of the tests" (everything not in `service/`), and the inline comment makes that clear, but the name itself is ambiguous on first read. Happy to leave it as-is if you prefer — just flagging in case it's worth tweaking before the name hardens through CI references and the `CiTestSplitTest` assertions.

.github/workflows/build_pr.yml
```diff
@@ -8,9 +8,36 @@ concurrency:
...
+        run: mvn -B -U -ntp test-compile -Dmaven.compiler.proc=full
+
+      - name: Package compiled output
+        run: tar -czf compiled.tar.gz target/classes target/test-classes
+
+      - name: Upload compiled output
+        uses: actions/upload-artifact@v7
+        with:
+          name: compiled
+          path: compiled.tar.gz
```
lubomirw, line 36: **~ Minor** | Performance

The new intermediate artifacts — `compiled`, `jacoco-exec-services`, `jacoco-exec-rest` — are only consumed within a single workflow run, but they inherit the org default retention (90 days). They'll accumulate across every PR build and add nontrivial storage with no real downstream use.

Adding `retention-days: 7` to those three uploads would still cover the realistic debugging window (Friday failure → Monday investigation, plus a holiday buffer) while letting them age out far sooner than the default. The final artifacts kept for outside consumption (`pr-number`, `compiled-classes`, `coverage-reports`) are fine to leave at the default.

lubomirw (commented):

src/test/java/com/czertainly/core/architecture/CiTestSplitTest.java
ivosh: Fixed in dd6d72dc.

.github/workflows/build_pr.yml
ivosh: Fixed in dd6d72dc.

.github/workflows/build_pr.yml
ivosh: Fixed in dd6d72dc.

pom.xml
ivosh: Fixed in dd6d72dc — went with `test-non-services`.

lubomirw (approved):

PR #1492, opened 2026-05-21
Add certificate-chain cache

### LINKED ISSUE #1499: Certificate chain in-process cache
## Summary

`getCertificateChainForSigning` is called on every signing operation to reconstruct the issuer chain from the database. Under concurrent signing load this becomes a hot read path — the same chains are fetched repeatedly with no reuse.

An in-process Caffeine cache for the certificate chain lookup, with a transaction-safe eviction strategy.

ivosh:


src/main/java/com/czertainly/core/dao/repository/CertificateRepository.java
```diff
@@ -249,25 +268,55 @@ void updateCertificateSubjectDN(@Param("oid") String oid,
...
+                SELECT c.uuid, c.issuer_certificate_uuid, chain.depth + 1
                 FROM {h-schema}certificate c
                 INNER JOIN chain ON chain.issuer_certificate_uuid = c.uuid
                 WHERE chain.depth < :maxDepth
-                  AND NOT (c.uuid = ANY(chain.path))
             )
-            SELECT uuid::text FROM chain ORDER BY depth ASC
+            CYCLE uuid SET is_cycle USING cycle_path
+            SELECT uuid::text FROM chain WHERE NOT is_cycle ORDER BY depth ASC
             """, nativeQuery = true)
```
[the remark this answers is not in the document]
ivosh: We target PostgreSQL 17+ in testcontainers and helm charts.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -865,6 +866,18 @@ private List<Certificate> getCertificateChainInternal(Certificate certificate, b
...
+    @Override
+    @Cacheable(value = CacheConfig.CERTIFICATE_CHAIN_CACHE, key = "#certificateUuid + '_' + #withEndCertificate", sync = true)
+    public List<X509Certificate> getCertificateChainForSigning(UUID certificateUuid, boolean withEndCertificate) throws CertificateException {
+        List<String> contents = certificateRepository.findCertificateChainContents(certificateUuid, certificateChainMaxDepth);
+        int startIdx = withEndCertificate ? 0 : 1;
+        List<X509Certificate> chain = new ArrayList<>(Math.max(0, contents.size() - startIdx));
+        for (int i = startIdx; i < contents.size(); i++) {
+            chain.add(CertificateUtil.parseCertificate(contents.get(i)));
+        }
+        return chain;
```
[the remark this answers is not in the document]
ivosh: Fixed in 4028a142.

src/main/java/com/czertainly/core/aop/CertificateChainCacheEvictor.java
```diff
@@ -0,0 +1,48 @@
...
+    public void evict() {
+        Cache cache = cacheManager.getCache(CacheConfig.CERTIFICATE_CHAIN_CACHE);
+        if (cache == null) return;
+        if (TransactionSynchronizationManager.isActualTransactionActive()) {
+            TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
+                @Override
+                public void afterCommit() {
+                    cache.clear();
+                }
+            });
```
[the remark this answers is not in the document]
ivosh: Fixed in 4028a142.

src/test/java/com/czertainly/core/service/CertificateChainCacheTest.java
```diff
@@ -0,0 +1,255 @@
...
+        // two getCertificateChainForSigning calls under test.
+        Mockito.clearInvocations(certificateRepository);
+
+        List<X509Certificate> first = certificateService.getCertificateChainForSigning(uuid, true);
+        List<X509Certificate> second = certificateService.getCertificateChainForSigning(uuid, true);
+
+        Assertions.assertSame(first, second, "second call must return the exact same cached object reference");
+        Mockito.verify(certificateRepository, Mockito.times(1))
+                .findCertificateChainContents(ArgumentMatchers.eq(uuid), ArgumentMatchers.anyInt());
+    }
```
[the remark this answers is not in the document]
ivosh: Fixed in 4028a142.

ivosh:
Fixed in c3292113.

lubomirw (commented):

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -865,6 +866,18 @@ private List<Certificate> getCertificateChainInternal(Certificate certificate, b
         return certificateChain;
     }
 
+    @Override
+    @Cacheable(value = CacheConfig.CERTIFICATE_CHAIN_CACHE, key = "#certificateUuid + '_' + #withEndCertificate", sync = true)
+    public List<X509Certificate> getCertificateChainForSigning(UUID certificateUuid, boolean withEndCertificate) throws CertificateException {
```
lubomirw, line 870: **! Major** | Maintainability

What is the intended consumer of `getCertificateChainForSigning`?

I searched three places and found no production caller: `src/main` in this branch, this branch's tests (8 hits — all under `src/test`), and `gh search code --owner OmniTrustILM` org-wide (empty result). The existing chain consumers (CMP, SCEP, ACME, REST controllers) all still route through `getCertificateChain(SecuredUUID, boolean)`.

If the consumer lands in a follow-up PR, could you link it from the description so reviewers can confirm the wiring? And if it is preliminary infrastructure, would it be worth noting that in the PR body so the cache/aspect behaviour is reviewed with that context in mind? Right now the only thing exercising the cache contract is the test suite, so a future aspect regression would not be caught by production traffic.

src/main/java/com/czertainly/core/aop/CertificateRepositoryCacheEvictionAspect.java
```diff
@@ -0,0 +1,36 @@
...
+public class CertificateRepositoryCacheEvictionAspect {
+
+    private CertificateChainCacheEvictor certChainCacheEvictor;
+
+    @Autowired
+    public void setCertChainCacheEvictor(CertificateChainCacheEvictor certChainCacheEvictor) {
+        this.certChainCacheEvictor = certChainCacheEvictor;
+    }
+
+    @Pointcut("bean(certificateRepository) && (execution(* save*(..)) || execution(* delete*(..)) || execution(* insert*(..)))")
```
lubomirw: **! Major** | Maintainability

Pointcut binds to the bean *name* `certificateRepository` rather than the repository *type*. Spring Data derives the bean name from the interface name, so today this works — but if someone ever adds `@Repository("certRepo")` to `CertificateRepository`, renames the interface, or wraps it behind a `@Primary` decorator, the pointcut silently stops matching and the cache returns stale chains. Stale chains during signing means the wrong issuer chain is attached to a signature — a correctness bug, not a perf bug, and one with no test to catch it (the cache tests would still pass because they use the same bean name).

Would you consider switching to a type-based pointcut, e.g. `target(com.czertainly.core.dao.repository.CertificateRepository+)` combined with the same `save*/delete*/insert*` execution match? That stays tied to the contract instead of a name.

src/main/java/com/czertainly/core/aop/CertificateRepositoryCacheEvictionAspect.java
```diff
@@ -0,0 +1,36 @@
...
+    private void certificateMutation() {}
+
+    @AfterReturning("certificateMutation()")
+    public void onMutation() {
+        certChainCacheEvictor.evict();
```
lubomirw, line 44: **! Major** | Performance

Is finer-grained eviction feasible here, or did you consider it and reject it?

Every `save*/delete*/insert*` on `CertificateRepository` currently triggers a full `cache.clear()` (deferred to `afterCommit`, deduped per transaction — the dedup is a nice touch). The docstring justifies the blast radius with "cheap to repopulate", but on a write-heavy deployment (ACME or SCEP issuing/uploading thousands of certs) every commit wipes the cache while signing traffic keeps refilling it. The cache becomes effectively a no-op under sustained writes.

A few concrete examples of what tighter eviction could look like:

- **Single `save(Certificate cert)`** — evict only `cert.uuid + "_true"` and `cert.uuid + "_false"`. Covers content changes and the leaf-as-end-cert case. Misses ancestor rewiring (see below).
- **Issuer rewiring** — when `cert.issuerCertificateUuid` changes, every cached chain whose path includes `cert.uuid` is stale. Without a reverse index (ancestor → set of cached leaf UUIDs maintained on cache put), the only safe move is `cache.clear()`. Worth it only if profiling shows ancestor rewiring is rare relative to leaf inserts.
- **Bulk deletes / `deleteAllInBatch`** — `cache.clear()` is genuinely the right call; intercept these separately and keep the wipe.

I don't think this needs to be in this PR, but a sentence in the docstring explaining "we deliberately picked the conservative strategy and here is the breakpoint at which finer-grained eviction becomes worth it" would help future maintainers know when to revisit.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
```diff
@@ -823,6 +823,7 @@ private void updateCertificateChain(Certificate certificate, X509Certificate sub
                     assert nextInChain != null;
                     previousCertificate.setIssuerCertificateUuid(nextInChain.getUuid());
                     previousCertificate.setIssuerSerialNumber(nextInChain.getSerialNumber());
+                    certificateRepository.save(previousCertificate);
```
lubomirw: **~ Minor** | Correctness

What is the reason for the explicit `certificateRepository.save(previousCertificate)` here? It's not mentioned in the commit message or PR body, and it's the only behaviour change in `updateCertificateChain` outside the cache infra.

Before this PR, `previousCertificate` was a managed JPA entity inside an active transaction, so dirty-checking would flush the issuer-link updates at commit. After this PR, every iteration triggers an explicit `save()`, which (a) flushes earlier and (b) — because of the new aspect — registers a `cache.clear()` synchronization. The cache eviction is dedup'd per transaction so the second/third/Nth save in the loop is essentially free for cache purposes; the real change is the flush timing.

If the intent is "make the issuer link visible before the next loop iteration queries it", that's legitimate but worth stating. If the intent is "trigger the cache eviction aspect", dirty-checking would already cover it (the aspect fires on the commit-time flush of the same row). Could you either add a brief comment explaining the intent or drop the line if it slipped in?

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
ivosh: This is preliminary infrastructure — the production consumer (signing service) lands in a follow-up PR; only the cache contract is exercised by the test suite in this branch. I'll add a note to the PR body to flag that context for the cache/aspect review.

src/main/java/com/czertainly/core/aop/CertificateRepositoryCacheEvictionAspect.java
ivosh: Fixed in 43a0991f — pointcut switched to `target(CertificateRepository+)` so it binds to the repository type rather than the bean name.

src/main/java/com/czertainly/core/aop/CertificateRepositoryCacheEvictionAspect.java
ivosh: Fixed in 43a0991f — extended the aspect Javadoc with the conservative-strategy rationale and the breakpoint at which finer-grained eviction (per-UUID + reverse index for issuer rewiring, full wipe retained for bulk paths) becomes worth the bookkeeping.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
ivosh: Fixed in 43a0991f — dropped the explicit `save(previousCertificate)`. It turned out to be redundant on both fronts: (a) the issuer-link columns are flushed at commit by JPA dirty-checking (no intra-loop read depends on the early flush — `previousCertificate` is reassigned to `nextInChain` immediately after), and (b) the cache eviction is already registered by the prior `insertWithFingerprintConflictResolve` inside `createCertificateAtomic` in the same iteration. The `cacheIsEvictedAfterAiaIssuerDownloaded` test still passes.

src/main/java/com/czertainly/core/service/impl/CertificateServiceImpl.java
ivosh: For now, you can refer to https://github.com/OmniTrustILM/core/blob/feat/signing/src/main/java/com/czertainly/core/service/tsa/certificateprovider/StaticManagedKeyCertificateProvider.java#L64

lubomirw (approved):

PR #1498, opened 2026-05-22
Add TSP profile cache

### LINKED ISSUE #1494: Cache TSP Profile to avoid database lookups
### Description

Caffeine cache for TspProfileModel lookups used by timestamping. TspProfileModel record captures only the data needed by the hot path.

ivosh:


src/main/java/com/czertainly/core/service/impl/TspProfileServiceImpl.java
```diff
@@ -104,14 +111,17 @@ public TspProfileDto getTspProfile(SecuredUUID uuid) throws NotFoundException {
...
-        TspProfile tspConfiguration = tspProfileRepository.findWithAssociationsByName(name)
-                .orElseThrow(() -> new NotFoundException("TSP Profile not found: " + name));
+        return self.loadTspProfileModel(name);
+    }
 
-        List<ResponseAttribute> customAttributes = attributeEngine.getObjectCustomAttributesContent(Resource.TSP_PROFILE, tspConfiguration.getUuid());
-        return TspProfileMapper.toModel(tspConfiguration, customAttributes);
+    @Cacheable(value = CacheConfig.TSP_PROFILE_CACHE, key = "#name", sync = true)
+    @Transactional(readOnly = true)
+    TspProfileModel loadTspProfileModel(String name) throws NotFoundException {
```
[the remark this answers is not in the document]
ivosh: No fix needed for this Spring Boot version. Spring Framework 6.0 changed `ProxyCachingConfiguration` to pass `publicMethodsOnly=false` to `AnnotationCacheOperationSource` (commit spring-projects/spring-framework@37bebea, closing gh-25582). The visibility gate in `AbstractFallbackCacheOperationSource` is now skipped entirely for CGLIB proxies, so package-private methods are eligible for `@Cacheable`. This project runs Spring Boot 3.5.13 (Spring 6.x) with CGLIB proxies by default, so the method caches correctly as-is.

src/main/java/com/czertainly/core/service/impl/TspProfileServiceImpl.java
```diff
@@ -104,14 +111,17 @@ public TspProfileDto getTspProfile(SecuredUUID uuid) throws NotFoundException {
     }
 
     @Override
-    @Transactional(readOnly = true)
     @ExternalAuthorization(resource = Resource.TSP_PROFILE, action = ResourceAction.DETAIL)
     public TspProfileModel getTspProfile(String name) throws NotFoundException {
```
klaraf755, line 115: I have just noticed that this method is only used in tests. Does it have some other purpose?

ivosh: The method is declared in the public TspProfileService interface and is the intended hot-path entry point for the timestamping service — the cache (@Cacheable by name) exists  specifically for this lookup. Its only current callers are in TspProfileCacheTest because the consuming component will land in a separate PR. We still have 10k+ lines on feat/signing branch to land :-)

klaraf755 (approved):

PR #1500, opened 2026-05-22
Split InternalService and ExternalService - part 4

ivosh:


src/main/java/com/czertainly/core/service/impl/TriggerServiceImpl.java
```diff
@@ -198,6 +198,7 @@ private void setTriggerRulesAndActions(Trigger trigger, List<String> rulesUuids,
     //region Trigger Associations
 
     @Override
+    @ExternalAuthorizationMissing
     public Map<ResourceEvent, List<UUID>> getTriggersAssociations(Resource resource, UUID associationObjectUuid) {
         List<TriggerAssociation> triggerAssociations = triggerAssociationRepository.findAllByResourceAndObjectUuidOrderByTriggerOrderAsc(resource, associationObjectUuid);
```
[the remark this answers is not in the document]
ivosh: The proper annotation categorization will be done in a follow-up PR.

src/main/java/com/czertainly/core/service/impl/SettingServiceImpl.java
```diff
@@ -190,8 +198,11 @@ public void updatePlatformSettings(PlatformSettingsUpdateDto platformSettings) {
     @Override
     @ExternalAuthorization(resource = Resource.SETTINGS, action = ResourceAction.LIST)
     public EventsSettingsDto getEventsSettings() {
-        Map<ResourceEvent, List<UUID>> eventsTriggerMapping = triggerService.getTriggersAssociations(null, null);
-        return new EventsSettingsDto(eventsTriggerMapping);
+        return loadEventsSettings();
```
lubomirw, line 201: **~ Minor** | Maintainability

`loadEventsSettings()` is a one-line trampoline. Both `getEventsSettings()` and `refreshCache()` call it; the apparent intent is to give internal callers (`refreshCache`) a path that skips the `@ExternalAuthorization` proxy. But Spring self-invocation already bypasses the proxy, so `refreshCache()` calling `getEventsSettings()` directly would behave identically — the split adds no runtime difference, only intent signaling.

Without a comment, a future maintainer will likely inline `loadEventsSettings()` back. Either add a one-liner explaining the intent ("internal callers must not go through the auth-checked entry point") or inline it.

lubomirw (commented):
Code review: 1 issue noted.

lubomirw (dismissed):

src/main/java/com/czertainly/core/service/impl/SettingServiceImpl.java
ivosh: Fixed in 9655d8d6.

lubomirw (approved):

PR #1519, opened 2026-05-27
Add TimeQualityConfiguration cache

ivosh:


src/main/java/com/czertainly/core/service/impl/TimeQualityConfigurationServiceImpl.java
```diff
@@ -112,6 +120,22 @@ public TimeQualityConfigurationDto getTimeQualityConfiguration(SecuredUUID uuid)
...
+    @Override
+    public TimeQualityConfigurationModel getTimeQualityConfigurationModel(UUID uuid)
+            throws NotFoundException {
+        return self.loadTimeQualityConfigurationModel(uuid);
+    }
+
+    @Cacheable(value = CacheConfig.TIME_QUALITY_CONFIGURATION_CACHE, key = "#uuid", sync = true)
+    @Transactional(readOnly = true)
+    TimeQualityConfigurationModel loadTimeQualityConfigurationModel(UUID uuid)
+            throws NotFoundException {
```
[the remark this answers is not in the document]
ivosh: The consumer will be added in a follow-up PR.

src/main/java/com/czertainly/core/service/impl/TimeQualityConfigurationServiceImpl.java
```diff
@@ -145,6 +169,7 @@ public TimeQualityConfigurationDto updateTimeQualityConfiguration(SecuredUUID uu
 
         fillTimeQualityConfigurationEntity(configuration, request);
         TimeQualityConfiguration saved = saveOrTranslateUniqueViolation(configuration, request.getName());
+        evictTimeQualityConfigurationCache(saved.getUuid()); // deferred to afterCommit() by the eviction helper
```
lubomirw, line 179: **~ Minor** | Correctness

Event published before cache eviction is registered — confirm no listener reads the model through the cache.

`saveOrTranslateUniqueViolation()` publishes `TimeQualityConfigChangedEvent` *before* `evictTimeQualityConfigurationCache()` registers its `afterCommit` synchronization. A synchronous `@EventListener` (runs in-transaction, pre-commit) or an `@TransactionalEventListener(AFTER_COMMIT)` registered first would still observe the **stale** cached model if it re-reads via `getTimeQualityConfigurationModel()`, since eviction hasn't run yet. This mirrors `TspProfileServiceImpl`, so it's likely fine if these events only trigger async messaging — could you confirm no listener does a cache read-through here?

lubomirw (commented):
Code review: 1 issue found (advisory). Overall a clean PR that faithfully mirrors the existing TspProfile cache pattern.

src/main/java/com/czertainly/core/service/impl/TimeQualityConfigurationServiceImpl.java
ivosh: Confirmed — the only `@TransactionalEventListener(AFTER_COMMIT)` listener for `TimeQualityConfigChangedEvent` is `TimeQualityMonitorInitializer.onConfigChanged()`, which calls `timeQualityConfigurationRepository.findAll()` directly rather than going through `getTimeQualityConfigurationModel()`. No listener does a cache read-through, so the ordering between the event synchronization and the eviction synchronization has no observable effect.

ivosh: Confirmed — the only @TransactionalEventListener(AFTER_COMMIT) listener for TimeQualityConfigChangedEvent is TimeQualityMonitorInitializer.onConfigChanged(), which calls
  timeQualityConfigurationRepository.findAll() directly rather than going through getTimeQualityConfigurationModel(). No listener does a cache read-through, so the ordering between the event synchronization  and the eviction synchronization has no observable effect.

lubomirw (approved):

PR #1520, opened 2026-05-27
Add source attribute mapping (META/DATA/CUSTOM → CUSTOM) to execution items

### LINKED ISSUE #1509: Implement Data/Custom/Metadata-to-Custom-Attribute Mapping in Execution Items
## Overview

Allow `SET_FIELD` execution items to copy the live value of a source attribute (META, DATA, or CUSTOM) into a target CUSTOM attribute, instead of writing a static value. This enables read-only metadata to be promoted into user-modifiable custom attributes for downstream use.

## Scope

- **Source**: any attribute-based field source — `META`, `DATA`, or `CUSTOM`
- **Target**: always `CUSTOM` (unchanged from existing SET_FIELD behavior)
- Both repos affected: `CZERTAINLY-Core` (entity, evaluator, service) and `CZERTAINLY-Interfaces` (DTOs)

### ExecutionItem entity

Two nullable columns added to the `execution_item` table via Flyway migration:

```java
@Column(name = "source_field_source")
@Enumerated(EnumType.STRING)
private FilterFieldSource sourceFieldSource;   // null = static value (existing behavior)

@Column(name = "source_field_identifier")
private String sourceFieldIdentifier;          // format: attributeName|ContentType
```

When both are `null`, behavior is unchanged (existing static `data` value is used).
When both are non-null, the source attribute's live value is read at execution time.

### sourceFieldIdentifier format

All three source types use the same `name|ContentType` format (e.g., `ejbcaUsername|STRING`), consistent with how META/DATA/CUSTOM identifiers already work in condition items.

## Creation-Time Validation

Performed in `ActionServiceImpl` when creating or updating an execution item with source fields populated:

| # | Check | Error |
|---|-------|-------|
| 1 | `source.contentType ≠ target.contentType` | Content types must match |

Validation requires fetching both attribute definitions (source and target) to read their properties.

**Not checked at creation time:**
- Whether the source attribute actually exists on any specific object (per-object, not schema-level)
- `extensibleList` / predefined list value compatibility — rejected by the attribute engine at write time if invalid
- `protectionLevel` — the `AttributeEngine` decrypts transparently on read; target's storage policy is respected

## Execution Logic

In `TriggerEvaluator.performSetFieldExecution()`, when `executionItem.sourceFieldSource` is non-null:

1. **Resolve source value** by `sourceFieldSource`:
   - `META`: `attributeEngine.getMappedMetadataContent()` — attribute name from `sourceFieldIdentifier`
   - `DATA`: `attributeEngine.getObjectDataAttributesContent()` — attribute name from `sourceFieldIdentifier`
   - `CUSTOM`: `attributeEngine.getObjectCustomAttributesContent()` — find by name from `sourceFieldIdentifier`

2. **Source not found**: mark execution item as failed in trigger history with a descriptive message (e.g., `"Source metadata attribute 'ejbcaUsername' not found on object"`); continue processing remaining items.

3. **Write to target**: pass resolved content as `actionData` into the existing `performSetFieldAttributeExecution()` path, which calls `attributeEngine.updateObjectCustomAttributeContent()`. Any `AttributeException` thrown by the engine (content type mismatch, extensible list violation, etc.) is caught and recorded as a failed execution item — no explicit type check needed in the evaluator.

## Error Handling

All failures are per-item, not per-execution: one failed item does not abort the remaining execution items. Failures are recorded in trigger history with specific messages to aid debugging.

### TriggerEvaluatorTest (runtime behavior)
- META → CUSTOM: source metadata exists, value is copied correctly
- DATA → CUSTOM: source data attribute exists, value is copied correctly
- CUSTOM → CUSTOM: source custom attribute exists, value is copied correctly
- Source attribute not found → item marked failed, other items continue
- AttributeEngine rejects content (e.g., extensible list violation) → item marked failed, other items continue

### ActionServiceImplTest (creation-time validation)
- Valid source fields accepted
- `contentType` mismatch rejected
- `source.multiSelect=true` + `target.multiSelect=false` rejected
- `target.readOnly=true` rejected
- `data` populated alongside source fields rejected
- Only one of `sourceFieldSource`/`sourceFieldIdentifier` provided → rejected

### User Story

_No response_

### Use Case

Use Case 2 - Mapping of metadata to custom attributes in execution

### Acceptance Criteria

- A SET_FIELD execution item targeting a CUSTOM attribute can optionally reference a source attribute (META, DATA, or CUSTOM) via new sourceFieldSource + sourceFieldIdentifier fields instead of a static value; when source fields are set, data must be null and vice versa
- At creation time, the API validates that source and target attribute definitions are compatible: content types must match, a multi-value source cannot map to a single-value target, and a read-only target is rejected
- At execution time, the evaluator reads the live value from the source attribute and writes it to the target custom attribute; if the source attribute is not found or the write is rejected by the attribute engine, the item is marked as failed in trigger history and remaining items continue

klaraf755:
Closes #1509


src/main/java/com/czertainly/core/service/impl/ActionServiceImpl.java
```diff
@@ -178,6 +180,57 @@ private ExecutionItem createSetFieldExecutionItem(Execution execution, Execution
...
+            throw new ValidationException("Source field reference is only supported when target fieldSource is CUSTOM.");
+        }
+        if (dto.getSourceFieldSource() != FilterFieldSource.META
+                && dto.getSourceFieldSource() != FilterFieldSource.DATA
+                && dto.getSourceFieldSource() != FilterFieldSource.CUSTOM) {
+            throw new ValidationException("sourceFieldSource must be META, DATA, or CUSTOM.");
+        }
+        if (dto.getData() != null) {
+            throw new ValidationException("data must be null when sourceFieldSource is set — use source reference or static data, not both.");
+        }
```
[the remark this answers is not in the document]
klaraf755: Originally, I was going to resolve definitions here. But definitions are not static and can change. So event though it would be evaluated as valid here, there can be issues at runtime anyway. Only content types are immutable.

src/main/java/com/czertainly/core/service/impl/ActionServiceImpl.java
```diff
@@ -178,6 +180,57 @@ private ExecutionItem createSetFieldExecutionItem(Execution execution, Execution
...
+        String targetId = dto.getFieldIdentifier();
+        int targetPipe = targetId.indexOf("|");
+        if (targetPipe < 0) {
+            throw new ValidationException("fieldIdentifier must be in format 'name|ContentType', got: " + targetId);
+        }
+        AttributeContentType targetContentType;
+        try {
+            targetContentType = AttributeContentType.valueOf(targetId.substring(targetPipe + 1));
+        } catch (IllegalArgumentException e) {
+            throw new ValidationException("Invalid content type in fieldIdentifier: " + targetId.substring(targetPipe + 1));
```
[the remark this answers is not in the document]
klaraf755: Changed

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
```diff
@@ -328,17 +329,74 @@ public void performActions(Trigger trigger, TriggerHistory triggerHistory, T obj
...
+            case DATA -> {
+                List<ResponseAttribute> dataAttrs = attributeEngine.getObjectDataAttributesContentUnversioned(resource, objectUuid);
+                yield dataAttrs.stream()
+                        .filter(ra -> Objects.equals(ra.getName(), sourceName))
+                        .findFirst()
+                        .map(ra -> ra.<List<AttributeContent>>getContent().stream()
+                                .<BaseAttributeContentV3<?>>map(c -> AttributeVersionHelper.convertAttributeContentToV3(c, ra.getContentType()))
+                                .toList())
+                        .orElse(null);
+            }
```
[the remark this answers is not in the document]
klaraf755: Added try block and check for same content type

src/main/java/com/czertainly/core/events/handlers/CertificateDiscoveredEventHandler.java
```diff
@@ -286,14 +286,15 @@ private void processDiscoveredCertificate(EventContext<Certificate> eventContext
                 eventData.setDiscoveryConnectorUuid(discovery.getConnectorUuid());
                 eventData.setDiscoveryConnectorName(discovery.getConnectorName());
 
+                certificateHandler.updateDiscoveredCertificate(discovery, certificate, discoveryCertificate.getMeta());
```
lubomirw, line 289: **! Major** | Correctness

Moving `updateDiscoveredCertificate` before the trigger loop is required so META-source resolution can read discovery metadata — but it also persists metadata before *every* discovery trigger's condition evaluation, not just the new SET_FIELD-from-source path.

Triggers with metadata-based conditions will now see discovered metadata where they previously did not, which can change existing trigger outcomes during discovery. Also adds the DISCOVERY event-history entry before trigger-history entries. Confirm this is intended and check for condition-evaluation regressions on existing discovery triggers.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
```diff
@@ -328,17 +329,80 @@ public void performActions(Trigger trigger, TriggerHistory triggerHistory, T obj
...
+            objectUuid = (UUID) PropertyUtils.getProperty(object, "uuid");
+        } catch (IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
+            throw new RuleException("Cannot get uuid from resource " + resource + ".");
+        }
+
+        final String sourceName;
+        final AttributeContentType sourceContentType;
+        try {
+            String sourceId = executionItem.getSourceFieldIdentifier();
+            sourceName = sourceId.substring(0, sourceId.indexOf("|"));
```
lubomirw: **~ Minor** | Correctness

If `sourceFieldIdentifier` has no `|`, `indexOf` returns -1 and `substring(0, -1)` throws `StringIndexOutOfBoundsException` — which is not an `IllegalArgumentException`, so this catch never fires for the malformed-separator case. Only the `AttributeContentType.valueOf` failure is actually caught.

Low impact (creation-time validation enforces format, and `performActions` catches generic `Exception`), but the catch is misleading. Reuse the same `split("\\|", -1)` + length check from `ActionServiceImpl.getAttributeContentType` so validation and execution parse identically.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
```diff
@@ -328,17 +329,80 @@ public void performActions(Trigger trigger, TriggerHistory triggerHistory, T obj
...
             String fieldIdentifier = executionItem.getFieldIdentifier();
-            Object actionData = executionItem.getData();
             FilterFieldSource fieldSource = executionItem.getFieldSource();
 
-            // Set a property of the object using setter, the property must be set as settable
-            if (fieldSource == FilterFieldSource.PROPERTY) {
-                performSetFieldPropertyExecution(fieldIdentifier, actionData, object);
-            } else if (fieldSource == FilterFieldSource.CUSTOM) { // Set a custom attribute for the object
-                performSetFieldAttributeExecution(resource, fieldIdentifier, actionData, object);
+            if (executionItem.getSourceFieldSource() != null) {
```
lubomirw, line 400: **~ Minor** | Correctness

`resolveSourceAttributeContent` throws `RuleException` when the source attribute is missing. `performActions` catches that per-*execution*, so any remaining `ExecutionItem`s in the same execution are skipped — a missing source on an early item silently drops later items in that execution.

The `...MissingSource...` test covers the single-item case only. Confirm executions are effectively single-item, or isolate per-item failures so one missing source doesn't abort sibling items.

src/main/java/com/czertainly/core/service/impl/ActionServiceImpl.java
```diff
@@ -201,9 +248,9 @@ private ExecutionItem createSendNotificationExecutionItem(Execution execution, E
         return executionItem;
     }
 
-    //endregion
+//endregion
```
lubomirw: **. Nitpick** | Maintainability

Region comments dedented to column 0 (`//endregion`, `//region Actions`) — unrelated whitespace churn from this feature. Restore the original indentation to keep the diff focused.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
```diff
@@ -292,7 +293,7 @@ private boolean evaluateCustomAttributeConditionItem(Resource resource, UUID obj
 
     @Override
     public void performActions(Trigger trigger, TriggerHistory triggerHistory, T object, Object data) throws RuleException {
-        Class resourceClass = ResourceToClass.getClassByResource(trigger.getResource());
+       Class resourceClass = ResourceToClass.getClassByResource(trigger.getResource());
```
lubomirw: **. Nitpick** | Maintainability

This line's indentation changed from 8 to 7 spaces (now misaligned), plus an unrelated `return true;` reindent in `evaluateMetaAttributeConditionItem`. Neither relates to the feature — revert to keep the diff clean.

src/test/java/com/czertainly/core/service/ActionServiceTest.java
```diff
@@ -0,0 +1,95 @@
...
+        item.setSourceFieldSource(FilterFieldSource.META);
+        item.setSourceFieldIdentifier("someAttr|STRING");
+
+        ExecutionRequestDto request = new ExecutionRequestDto();
+        request.setName("testExecution_" + System.nanoTime());
+        request.setResource(Resource.CERTIFICATE);
+        request.setType(ExecutionType.SET_FIELD);
+        request.setItems(List.of(item));
+        return request;
+    }
```
lubomirw, line 94: **. Nitpick** | Testing

Missing trailing newline at end of file. Add one.

lubomirw (commented):
Code review: 6 finding(s). 2 Major (both advisory questions on correctness/behavior), 2 Minor, 2 Nitpick. No blocking request-changes.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
```diff
@@ -328,17 +329,74 @@ public void performActions(Trigger trigger, TriggerHistory triggerHistory, T obj
...
+            if (executionItem.getSourceFieldSource() != null) {
+                List<BaseAttributeContentV3<?>> resolvedContent = resolveSourceAttributeContent(resource, executionItem, object);
+                performSetFieldAttributeExecution(resource, fieldIdentifier, resolvedContent, object);
+            } else {
+                Object actionData = executionItem.getData();
+                if (fieldSource == FilterFieldSource.PROPERTY) {
+                    performSetFieldPropertyExecution(fieldIdentifier, actionData, object);
+                } else if (fieldSource == FilterFieldSource.CUSTOM) {
+                    performSetFieldAttributeExecution(resource, fieldIdentifier, actionData, object);
+                }
```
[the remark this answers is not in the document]
klaraf755: It is intended to fail on execution as whole.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
```diff
@@ -328,17 +329,74 @@ public void performActions(Trigger trigger, TriggerHistory triggerHistory, T obj
...
+                        .map(ra -> ((ResponseAttributeV3) ra).getContent())
+                        .orElse(null);
+            }
+            default -> throw new RuleException("Unsupported sourceFieldSource: " + sourceFieldSource);
+        };
+
+        if (content == null) {
+            throw new RuleException("Source attribute '" + sourceName + "' of type " + sourceFieldSource + " not found on object.");
+        }
+        return content;
```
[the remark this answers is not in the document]
klaraf755: It is intended to fail on execution as whole.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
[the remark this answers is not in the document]
klaraf755: It is intended to fail on execution as whole.

src/test/java/com/czertainly/core/service/ActionServiceTest.java
klaraf755: Done

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
klaraf755: Done

src/main/java/com/czertainly/core/service/impl/ActionServiceImpl.java
klaraf755: Done

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
klaraf755: Reused the same check

src/main/java/com/czertainly/core/events/handlers/CertificateDiscoveredEventHandler.java
klaraf755: Adding the DISCOVERY event-history entry before trigger-history entries does not change anything - it is created from the discovery info. 
Existing discovery triggers are evaluated only once per discovery on newly created certificates, so there is no change for them.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
klaraf755: This is the current design for trigger records - they are tied to the execution as whole rather to execution item.

lubomirw (approved):

PR #1522, opened 2026-05-28
Add Signing Profile model layer for managed timestamping

### LINKED ISSUE #1523: Signing Profile — Model layer: cache-safe sealed-type model for Signing Profiles
## Description

Introduce an immutable, cache-safe **Signing Profile model** layer that represents a Signing Profile independently of its JPA entity. The model is shaped to be cached safely and consumed by signing/timestamping workflows without dragging entity graphs or peer-cache objects into request execution.

ivosh:


src/main/java/com/czertainly/core/mapper/signing/SigningProfileMapper.java
```diff
@@ -173,6 +212,58 @@ private static TimestampingWorkflowDto buildTimestampingWorkflowDto(
...
+
+    private static ManagedTimestampingWorkflow buildManagedTimestampingWorkflowModel(
+            SigningProfile header, SigningProfileVersion version, List<RequestAttribute> signatureFormatterConnectorAttributes) {
+        return new ManagedTimestampingWorkflow(
+                version.getSignatureFormatterConnectorUuid(),
+                safeList(signatureFormatterConnectorAttributes),
+                version.getQualifiedTimestamp(),
+                header.getTimeQualityConfigurationUuid(),
+                version.getDefaultPolicyId(),
+                safeList(version.getAllowedPolicyIds()),
```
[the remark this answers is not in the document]
ivosh: Fixed in 79612fbe.

src/main/java/com/czertainly/core/mapper/signing/SigningProfileMapper.java
```diff
@@ -173,6 +212,58 @@ private static TimestampingWorkflowDto buildTimestampingWorkflowDto(
...
+                    version.getDelegatedSignerConnectorUuid(),
+                    List.of());
+            case MANAGED -> {
+                if (version.getManagedSigningType() == null) {
+                    throw new IllegalStateException("MANAGED signing profile version has no managedSigningType");
+                }
+                yield switch (version.getManagedSigningType()) {
+                    case STATIC_KEY -> new StaticKeyManagedSigning(
+                            version.getCertificateUuid(),
+                            safeList(signingOperationAttributes));
```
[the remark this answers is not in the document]
ivosh: Fixed in 79612fbe.

src/main/java/com/czertainly/core/mapper/signing/SigningProfileMapper.java
```diff
@@ -173,6 +212,58 @@ private static TimestampingWorkflowDto buildTimestampingWorkflowDto(
...
+                    case ONE_TIME_KEY -> new OneTimeKeyManagedSigning(
+                            version.getRaProfileUuid(),
+                            version.getTokenProfileUuid(),
+                            version.getCsrTemplateUuid(),
+                            safeList(signingOperationAttributes));
```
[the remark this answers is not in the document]
ivosh: Fixed in 79612fbe.

src/main/java/com/czertainly/core/mapper/signing/SigningProfileMapper.java
```diff
@@ -103,6 +110,38 @@ public static SigningProfileDto toDto(SigningProfile header, SigningProfileVersi
...
+     * and never dereferences the lazy JPA associations, so it is safe to invoke on detached entities and outside an open Session.</p>
+     *
+     * @throws IllegalArgumentException if the profile's workflow type is not {@code TIMESTAMPING}
+     *                                  or its signing scheme is not {@code MANAGED}
+     */
+    public static SigningProfileModel<ManagedTimestampingWorkflow, SigningSchemeModel> toManagedTimestampingModel(
+            SigningProfile header,
+            SigningProfileVersion version,
+            List<RequestAttribute> signingOperationAttributes,
+            List<RequestAttribute> signatureFormatterConnectorAttributes) {
```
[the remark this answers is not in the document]
ivosh: Fixed.

lubomirw (commented):

src/main/java/com/czertainly/core/model/signing/scheme/ManagedSigning.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Sealed interface for the managed-signing branch of the scheme model hierarchy.
+ *
+ * <p>Use pattern matching to access type-specific fields:</p>
+ * <pre>{@code
+ * switch (model.signingScheme()) {
+ *     case StaticKeyManagedSigning s -> s.certificateUuid();
+ *     case OneTimeKeyManagedSigning o -> o.raProfileUuid();
+ *     case DelegatedSigning d -> d.connectorUuid();
```
lubomirw: **~ Minor** | Documentation

Pattern-match example includes `DelegatedSigning`, which isn't a permit of `ManagedSigning`

The Javadoc on `ManagedSigning` shows a `switch (model.signingScheme())` example with three cases, the third being `case DelegatedSigning d -> …`. But `ManagedSigning permits StaticKeyManagedSigning, OneTimeKeyManagedSigning` only — `DelegatedSigning` is a sibling under `SigningSchemeModel`, not a permit of `ManagedSigning`. As written, this example doesn't compile when the static type of `signingScheme()` is `ManagedSigning` (and is exhaustive without the `Delegated` case). Either move the example to `SigningSchemeModel` (where all three branches are valid permits and exhaustiveness holds), or drop the `DelegatedSigning` case here and leave just the two managed variants.

src/main/java/com/czertainly/core/mapper/signing/SigningProfileMapper.java
```diff
@@ -103,6 +109,38 @@ public static SigningProfileDto toDto(SigningProfile header, SigningProfileVersi
...
     // ──────────────────────────────────────────────────────────────────────────
 
+    /**
+     * Transforms a {@link SigningProfile} and {@link SigningProfileVersion} pair to a {@link SigningProfileModel} typed
+     * with {@link ManagedTimestampingWorkflow}. The caller must ensure the profile uses a managed timestamping workflow.
+     *
+     * <p>This assembler reads UUID columns only (e.g. {@code version.getCertificateUuid()}, {@code header.getTimeQualityConfigurationUuid()})
+     * and never dereferences the lazy JPA associations, so it is safe to invoke on detached entities and outside an open Session.</p>
+     *
+     * @throws IllegalArgumentException if the profile's workflow type is not {@code TIMESTAMPING}
```
lubomirw, line 119: **~ Minor** | Documentation

`@throws` list omits `IllegalStateException`

`toManagedTimestampingModel` can propagate an `IllegalStateException` from `buildManagedSchemeModel` when a version has `signingScheme == MANAGED` but `managedSigningType == null` — a DB-integrity inconsistency the assembler can't recover from. The choice to use ISE here (vs. IAE on the workflow/scheme dispatch checks) is the right one: IAE signals "wrong assembler for this entity, route differently", while ISE signals "this entity is internally malformed". But the public `@throws` list only documents `IllegalArgumentException`. Please add an `@throws IllegalStateException` line so callers can reason about both failure modes.

src/main/java/com/czertainly/core/mapper/signing/SigningProfileMapper.java
```diff
@@ -103,6 +109,38 @@ public static SigningProfileDto toDto(SigningProfile header, SigningProfileVersi
...
+            List<RequestAttribute> signatureFormatterConnectorAttributes) {
+        if (version.getWorkflowType() != SigningWorkflowType.TIMESTAMPING) {
+            throw new IllegalArgumentException("Signing Profile '%s' does not use a timestamping workflow".formatted(header.getName()));
+        }
+        if (version.getSigningScheme() != SigningScheme.MANAGED) {
+            throw new IllegalArgumentException("Signing Profile '%s' does not use a managed signing scheme".formatted(header.getName()));
+        }
+
+        List<SigningProtocol> protocols = header.getTspProfileUuid() != null ? List.of(SigningProtocol.TSP) : List.of();
+        int ver = version.getVersion() != null ? version.getVersion() : 1;
```
lubomirw: **~ Minor** | Maintainability

The `getVersion() != null ? … : 1` fallback only exists because `SigningProfileVersion.version` is declared `Integer` despite `@Column(nullable = false)`

The new model layer correctly uses `int version` (primitive). The entity, however, declares `private Integer version` with `nullable = false` — which is the only column in `SigningProfileVersion` or `SigningProfile` that boxes a non-nullable primitive. Compare:

- `SigningProfile.enabled` → `boolean` (nullable=false ✓ primitive)
- `SigningProfile.latestVersion` → `int` (nullable=false ✓ primitive — the analogous "version number on a profile" field)
- `SigningProfileVersion.validateTokenSignature` → `Boolean` (nullable ✓ boxed)

Changing `SigningProfileVersion.version` to `int` aligns the entity with its sibling's convention, matches the model's choice, and makes the awkward null-check-or-default question in this assembler simply go away (the ternary no longer compiles). Scope-wise this is an entity-field edit outside this PR's strict scope, but it's a one-character fix that's worth doing here since this PR is the first consumer that has to defend against the inconsistency.

src/main/java/com/czertainly/core/model/signing/scheme/ManagedSigning.java
ivosh: Fixed in e37a5253.

src/main/java/com/czertainly/core/mapper/signing/SigningProfileMapper.java
ivosh: Fixed in e37a5253.

src/main/java/com/czertainly/core/mapper/signing/SigningProfileMapper.java
ivosh: Fixed in e37a5253. The Integer → int change also eliminated the same null-guard in the DTO mapper (line 51), which had the same pattern but hadn't been flagged separately.

lubomirw (approved):

PR #1526, opened 2026-05-28
Add dedicated writer for CRL generation

### LINKED ISSUE #1468: Certificates - Revoke: first Revoke attempt is silently dropped when the RA Profile has a Compliance Profile attached
### Description

When a certificate is issued via an RA Profile that has a **Compliance Profile** attached, the **first** click on **Revoke** is silently dropped: the backend returns `204 No Content`, the success toast is shown in the UI, but the certificate state does **not** transition to `Pending Revoke`, no entry is written to **Event History**, and the certificate remains in `Issued` state.

Only the **second** click on **Revoke** actually initiates the revocation flow — even though the API returns `204 No Content` in both cases. From the user's point of view, the first Revoke appears to have succeeded but in fact nothing happened, which is especially misleading for a critical PKI operation.

This was originally observed alongside #1455 (compliance check failing with `canceling statement due to user request`). Per dev comment the original validation-status / duplicate-history symptom is fixed, but this Revoke first-attempt issue is a related but **separate** bug that needs its own ticket.

Core log evidence on the same flow showed the first `revoke` message arriving on `actionsListener`, the API responding `success`, but the actual state transition to `PENDING_REVOKE` only happening ~2 minutes later — at the exact moment the compliance check finally failed with `canceling statement due to user request`. The compliance check appears to block the first revoke message from being processed.

### Steps to Reproduce

1. Make sure there is an RA Profile that has a **Compliance Profile** attached (External Authority RA Profile, e.g. `external-3keynote`).
2. Issue a certificate via that RA Profile and finalize it so it reaches state `Issued`, validation status `Valid`.
3. Open the certificate detail page and click **Revoke**, fill in the dialog, submit.
4. Observe the success toast and the certificate detail page.
5. Without refreshing, click **Revoke** again.

### Expected Behavior

- The first **Revoke** request actually initiates revocation
- Certificate transitions to `Pending Revoke` immediately
- Event History contains a `Revocation accepted` entry
- The success toast accurately reflects what happened in the backend (no false-positive confirmation)

### Actual Behavior

- First **Revoke**: API returns `204 No Content`, success toast shown, but certificate state stays `Issued`, Event History does not get a new entry — the operation is silently dropped
- Second **Revoke** (without page refresh): API returns `204 No Content` again, this time the certificate transitions to `Pending Revoke` and `Revocation accepted` appears in Event History
- Reproducible multiple times in a row in one session; correlates in time with `canceling statement due to user request` from the compliance check on the same certificate

### Severity

Major — Revoke is a critical security operation. The UI shows a success toast for an operation that did not actually happen, which can mislead the operator into believing a compromised certificate has been revoked when it has not.

### Module

Certificates

### Screenshot

_No response_

### Logs

_No response_

### Environment
ILM Core: 2.17.1-SNAPSHOT (15 May 2026, 13:36)
ILM Frontend: 2.17.0 (15 May 2026, 12:42)
Client: macOS Tahoe 26.3.1 / Chrome 148.0.7778.168 (arm64)

ivosh:
## What the old code got wrong

Before this PR, `CrlServiceImpl.getCurrentCrl` held an ambient JPA transaction across the HTTP download of the CRL. Any slow or unresponsive CRL distribution point would pin a database connection and any row locks acquired earlier in the call chain for the full duration of the HTTP fetch — minutes, in pathological cases — blocking every other operation that needed the same resources.

Inside that ambient transaction, the persistence path was equally fragile. `createCrlAndCrlEntries` rebuilt the `Crl` entity in memory and called `crlRepository.save(crl)` to persist it. `save` delegates to `em.merge`, which writes **every column** back to the database from the in-memory snapshot. Across the slow HTTP window, any concurrent writer (a parallel `getCurrentCrl` racing on the same issuer, an admin-driven CRL refresh) could commit changes that the merge then silently overwrote with a stale view.

The same shape repeated in `addCrlEntries`: a refetch, a mutation of `lastRevocationDate`, and another `save(crl)` — another wholesale merge of every column on a now-doubly-stale snapshot.

```
Thread A: getCurrentCrl(cert)                  Thread B: getCurrentCrl(cert)
─────────────────────────────────              ─────────────────────────────────
BEGIN tx
read crl (crlNumber = 42)
  ... HTTP fetch CDP (slow) ...
                                               BEGIN tx
                                               read crl (crlNumber = 42)
                                                 ... HTTP fetch CDP ...
                                               em.merge(crl with crlNumber = 43)
                                               COMMIT                ← committed
em.merge(crl with crlNumber = 43, stale)  ───► crlNumber = 43 (lost-update window)
COMMIT
```

The delta path had a third variant of the same problem. `updateDeltaCrl` walked `crl.getCrlEntries()` (lazy-loaded under the ambient tx) and called `crlEntryRepository.save` / `delete` once per delta entry, then `crlRepository.save(crl)` on the parent — a per-entry round-trip pattern with a full-row merge of the parent at the end.

## What this PR does, and where it sits in the path to the fix

This PR is **PR 3 of the 5-PR sequence** that converges on the race-free revoke fix. It introduces `CrlWriter` — the writer bean for the CRL ingestion path — and the targeted `@Modifying @Query` methods it needs in `CrlRepository` and `CrlEntryRepository`. The orchestrator (`CrlServiceImpl`) is refactored to do all HTTP I/O and in-memory shaping **outside** any transaction; the writer opens a short transaction only for the local writes.

```
┌────────────────────────────────────────┐    ┌───────────────────────────────────────────┐
│ CrlServiceImpl                         │    │ CrlWriter                                 │
│ • getCurrentCrl                        │    │ (no class annotation)                     │
│   @Transactional(NOT_SUPPORTED)        │    │ • @Transactional                          │
│   — runs WITHOUT an ambient tx so the  │───▶│   persistCrlAndEntries(...)               │
│     HTTP CRL download never holds a    │    │     — for new CRL:                        │
│     connection or locks                │    │       insertWithIssuerConflictResolve     │
│                                        │    │       + targeted entry inserts            │
│ • createCrlAndCrlEntries               │    │       + updateBaseMetadata (3 cols)       │
│   — HTTP fetch, parse, build a List<   │    │     — for refreshed CRL:                  │
│     CrlEntryData> entirely in memory   │    │       deleteAllByCrlUuid                  │
│   → writer.persistCrlAndEntries(...)   │    │       + targeted entry inserts            │
│                                        │    │       + updateBaseMetadata (3 cols)       │
│ • updateDeltaCrl                       │    │ • @Transactional                          │
│   — HTTP fetch, classify entries into  │───▶│   applyDeltaCrl(...)                      │
│     upserts vs removals in memory      │    │     — bulk delete removals                │
│   → writer.applyDeltaCrl(...)          │    │     + bulk upserts (ON CONFLICT DO UPDATE)│
│                                        │    │     + updateDeltaMetadata (3 cols)        │
└────────────────────────────────────────┘    └───────────────────────────────────────────┘
```

`updateBaseMetadata` and `updateDeltaMetadata` are targeted UPDATEs that touch **only** the metadata columns the path actually owns (`crl_number`, `next_update`, `last_revocation_date` for the base; `crl_number_delta`, `next_update_delta`, `last_revocation_date` for the delta). They never read or rewrite columns owned by other paths, so concurrent writers on disjoint columns no longer clobber each other.

The new `CrlEntryRepository` methods follow the same principle: `upsertEntry` does `INSERT … ON CONFLICT DO UPDATE` (so concurrent delta applies converge instead of racing on row existence), `deleteByCrlUuidAndSerialNumber` is a single targeted delete, and `findAllByCrlUuid` replaces the lazy-collection traversal with an explicit fetch the caller controls.

PR sequence to the revoke fix:

```
PR #1512        ──►  PR #1517         ──►  PR #1526 (this) ──►  PR 4         ──►  PR 5
Chain service        Validation            CrlService           Drop redundant     Revoke fix
+ writer             writer                refactor + writer    repo-level         lands cleanly
                                                                @Transactional
```

## Tests

- **`CrlInsertVsValidateTest`** — end-to-end race simulation built on `BaseSpringBootTest` + WireMock. A WireMock CDP serves a real CRL with a 500ms fixed delay; two parallel `getCurrentCrl` calls race on the same issuer. Asserts that exactly one row lands in `crl` and that neither caller observes a torn state — exercises both the `NOT_SUPPORTED` propagation on the orchestrator and the `ON CONFLICT DO NOTHING` resolution inside `persistCrlAndEntries`.
- **`CertificateValidationTest`** — extended to cover the refactored CRL path: base CRL creation, base CRL refresh (new `crlNumber`), and delta CRL application (upserts, `REMOVE_FROM_CRL` removals, `last_revocation_date` advancement). Verifies that `applyDeltaCrl` writes exactly the delta-owned columns and leaves the base-CRL columns intact.


src/main/java/com/czertainly/core/service/impl/CrlServiceImpl.java
```diff
@@ -197,57 +197,61 @@ private void updateCrlAndCrlEntriesFromDeltaCrl(X509Certificate certificate, Crl
...
-        if (crl.getCrlNumberDelta() == null || Integer.parseInt(encodedCrlNumber.toString()) > Integer.parseInt(crl.getCrlNumberDelta())) {
-            List<CrlEntry> crlEntries = crl.getCrlEntries();
-            Date lastRevocationDateNew = crl.getLastRevocationDate();
-            lastRevocationDateNew = addDeltaCrlEntries(crl, deltaCrl, lastRevocationDateNew, crlEntries);
-            // Update last revocation date from new/updated entries
-            crl.setLastRevocationDate(lastRevocationDateNew);
-            crl.setCrlNumberDelta(encodedCrlNumber.toString());
-            crl.setNextUpdateDelta(deltaCrl.getNextUpdate());
-            crlRepository.save(crl);
+        if (crl.getCrlNumberDelta() != null && Integer.parseInt(encodedCrlNumber.toString()) <= Integer.parseInt(crl.getCrlNumberDelta())) {
```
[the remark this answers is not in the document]
ivosh: Fixed in e8bfac73.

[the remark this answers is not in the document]
ivosh: Fixed in e8bfac73.

src/test/java/com/czertainly/core/service/CertificateValidationTest.java
```diff
@@ -387,123 +387,248 @@ void testOcspValidationCheck() throws Exception {
...
+        return keyPairGen.generateKeyPair();
+    }
+
+    private void replaceStub(String urlPath, byte[] body) {
+        mockServer.getStubMappings().stream()
+                .filter(s -> s.getRequest().getUrl() != null && s.getRequest().getUrl().equals(urlPath))
+                .findFirst()
+                .ifPresent(s -> mockServer.removeStubMapping(s));
+        stubCrlPoint(urlPath, body);
+    }
```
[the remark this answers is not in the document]
ivosh: Fixed in 29f24eea.

src/main/java/com/czertainly/core/service/impl/CrlServiceImpl.java
```diff
@@ -87,79 +100,87 @@ public void clearCrlsForCaCertificate(List<UUID> caCertificateUuids) {
...
-                crlRepository.save(crl);
-            }
-
-            crl = addCrlEntries(caCertificateUuid, x509CRL, crl);
+            List<CrlEntryData> entries = new ArrayList<>();
+            Date lastRevocationDate = collectEntries(x509CRL, entries);
 
-            // Managed to process a CRL url and do not need to try other URLs
+            UUID persistedUuid = crlWriter.persistCrlAndEntries(crl, isNewCrl, entries, lastRevocationDate);
+            crl.setUuid(persistedUuid);
```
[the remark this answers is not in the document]
ivosh: Fixed in 29f24eea.

lubomirw (commented):
Code review: 4 comments — 3 inline, 1 general observation below. Mostly nitpicks; no blockers. The race-safety story is sound for a PR that's explicitly foundation for the later revoke fix.

**General observation (not inline because the file isn't in the diff): `src/main/java/com/czertainly/core/dao/entity/Crl.java`**

**. Nitpick** | Maintainability

After this PR, no caller in `src/main` reads `Crl.crlEntries` or `getCrlEntriesMap()`:

- The old delta path consumed `crl.getCrlEntriesMap()` — now replaced by `crlEntryRepository.findAllByCrlUuid(crl.getUuid())`.
- The old `addCrlEntries` populated `crl.setCrlEntries(new ArrayList<>())` — gone.

The mapping still exists with `cascade = CascadeType.ALL`, so it would activate if anyone calls `crlRepository.delete(crl)` or `save(crl)` on a managed instance, but neither happens in the new write path (all writes go through `@Modifying` queries on `CrlWriter`). That makes `getCrlEntriesMap()` dead code and the cascade misleading.

If this is intentional (kept for future use or read-side callers I missed), a one-line comment on the field would help. Otherwise consider removing `getCrlEntriesMap()` and reducing the cascade to what's actually needed, in a follow-up.

src/main/java/com/czertainly/core/service/impl/CrlServiceImpl.java
```diff
@@ -197,57 +216,71 @@ private void updateCrlAndCrlEntriesFromDeltaCrl(X509Certificate certificate, Crl
...
-            crl.setNextUpdateDelta(deltaCrl.getNextUpdate());
-            crlRepository.save(crl);
+        if (crl.getCrlNumberDelta() != null) {
+            int deltaCrlNumber;
+            int storedDeltaCrlNumber;
+            try {
+                deltaCrlNumber = Integer.parseInt(encodedCrlNumber.toString());
+                storedDeltaCrlNumber = Integer.parseInt(crl.getCrlNumberDelta());
+            } catch (NumberFormatException e) {
+                throw new ValidationException("Invalid CRL number format in delta CRL: " + e.getMessage());
```
lubomirw: **~ Minor** | Error handling

Per `core/CLAUDE.md` ("Don't leak runtime details to the wire"), raw `Exception.getMessage()` from a JDK exception shouldn't be forwarded onto a domain exception that may reach a client. `NumberFormatException.getMessage()` is bounded (`"For input string: \"…\""`) but it does echo the offending input verbatim — fine when the input is from a CA-signed CRL extension, but the pattern is the one the guideline calls out.

Two cheap alternatives:

```java
throw new ValidationException("Invalid CRL number format in delta CRL");
// or, if you want to keep diagnostic detail in logs but not on the wire:
logger.warn("Invalid CRL number format in delta CRL: {}", e.getMessage());
throw new ValidationException("Invalid CRL number format in delta CRL");
```

Same applies to `storedDeltaCrlNumber` — if that ever fails to parse, the offending value originated from our own DB, so leaking it is doubly avoidable.

src/main/java/com/czertainly/core/service/writer/CrlWriter.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    @Transactional
+    public UUID persistCrlAndEntries(Crl crl, boolean isNewCrl, List<CrlEntryData> entries, Date lastRevocationDate) {
+        if (isNewCrl) {
+            crlRepository.insertWithIssuerConflictResolve(crl);
+        } else {
+            crlEntryRepository.deleteAllByCrlUuid(crl.getUuid());
+        }
+        UUID persistedUuid = crlRepository.findByIssuerDnAndSerialNumber(crl.getIssuerDn(), crl.getSerialNumber())
+                .map(Crl::getUuid)
```
lubomirw: **. Nitpick** | Maintainability

The `findByIssuerDnAndSerialNumber` lookup is essential on the new-CRL branch (it resolves the winner's UUID after `ON CONFLICT DO NOTHING`). On the refresh branch (`isNewCrl == false`) it is redundant — `crl.getUuid()` was loaded by the caller from the same `(issuer_dn, serial_number)` row a few statements earlier, and `deleteAllByCrlUuid(crl.getUuid())` has already used it.

Splitting the two paths would make the new-CRL branch's intent (recover the winner's UUID) explicit and let the refresh branch skip the extra `SELECT`:

```java
UUID persistedUuid;
if (isNewCrl) {
    crlRepository.insertWithIssuerConflictResolve(crl);
    persistedUuid = crlRepository.findByIssuerDnAndSerialNumber(crl.getIssuerDn(), crl.getSerialNumber())
            .map(Crl::getUuid)
            .orElseThrow(() -> new IllegalStateException("..."));
} else {
    persistedUuid = crl.getUuid();
    crlEntryRepository.deleteAllByCrlUuid(persistedUuid);
}
```

Not a correctness issue — and the unified form may have been deliberate as a defensive re-fetch — but worth a Javadoc note if so.

src/test/java/com/czertainly/core/service/writer/CrlInsertVsValidateTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+        assertNull(err1.get(), "race worker 1 must not throw: " + err1.get());
+        assertNull(err2.get(), "race worker 2 must not throw: " + err2.get());
+
+        List<com.czertainly.core.dao.entity.Crl> rows = crlRepository.findAll();
+        long matchingRows = rows.stream()
+                .filter(c -> issuerDn.equals(c.getIssuerDn()) && issuerSerial.equals(c.getSerialNumber()))
+                .count();
+        assertEquals(1L, matchingRows,
+                "exactly one CRL row must exist for the issuer — ON CONFLICT (issuer_dn, serial_number) DO NOTHING "
```
lubomirw, line 129: **. Nitpick** | Testing

The race assertion pins down the row-count invariant, but the stub serves an *empty* CRL (no revoked entries), so the test would still pass even if `persistCrlAndEntries` silently dropped or duplicated entry inserts during the race. Two complementary assertions would close that gap:

```java
// 1. The winning row should be fully populated, not a half-written stub:
Crl row = crlRepository.findByIssuerDnAndSerialNumber(issuerDn, issuerSerial).orElseThrow();
assertNotNull(row.getCrlNumber(),  "crl_number must be set on the winning row");
assertNotNull(row.getNextUpdate(), "next_update must be set on the winning row");

// 2. If the stub were extended to a CRL with N entries, exactly N rows in crl_entry:
//    (would require switching the @BeforeEach stub from emptyCrl to a CRL with at least one revoked serial)
long entries = crlEntryRepository.findAllByCrlUuid(row.getUuid()).size();
assertEquals(expectedEntries, entries, "entry set must match the served CRL");
```

The first assertion alone is a one-line addition and catches the "metadata UPDATE didn't run on the winner" failure mode. The second is what would make the test guard against ghost / missing entries under the wider race scenarios (different CRL versions between threads); worth as a follow-up if not in scope here.

src/main/java/com/czertainly/core/service/impl/CrlServiceImpl.java
ivosh: Fixed in aa3ca903.

src/main/java/com/czertainly/core/service/writer/CrlWriter.java
ivosh: Fixed in aa3ca903.

src/test/java/com/czertainly/core/service/writer/CrlInsertVsValidateTest.java
ivosh: Fixed in aa3ca903.

lubomirw (approved):

PR #1527, opened 2026-05-28
Replace @ExternalAuthorizationMissing with permanent annotations - part 6

ivosh:
## Summary

Converts 21 of the 25 `@ExternalAuthorizationMissing` markers recorded in the ArchUnit freeze store to the appropriate permanent annotation from the five-annotation set (`@ExternalAuthorization`, `@ProtocolEndpoint`, `@SelfPrincipalEndpoint`, `@AnyPrincipalEndpoint`, `@UnauthenticatedEndpoint`).

The remaining four markers all have the same shape — the protected `Resource` is a runtime parameter, not a compile-time constant, so `@ExternalAuthorization(resource = ...)` cannot bind it. They are deferred to a follow-up that decides whether to add a dynamic-resource annotation, refactor those endpoints to per-resource overloads, or accept `@AnyPrincipalEndpoint` with downstream enforcement. The freeze store now contains exactly those four.

### ResourceServiceImpl → `@AnyPrincipalEndpoint` (4)

`listResources`, `listResourceRuleFilterFields`, `listResourceEvents`, `listAllResourceEvents` all return enum/metadata-derived content that is identical for every authenticated caller. No object-level data leaks, no per-resource authorization decision to make.

### ApprovalServiceImpl → `@SelfPrincipalEndpoint` (2)

`approveApprovalRecipient` / `rejectApprovalRecipient` are the canonical self-referential case: the caller is voting on their own recipient slot. Recipient-membership is enforced inside `validateAndSetPendingApprovalRecipient` (the user must appear in the approval's recipient list and must not be the approval's creator). Deliberately distinct from the admin-side `approveApproval` / `rejectApproval`, which already carry `@ExternalAuthorization(APPROVAL, APPROVE/REJECT)` — a recipient may not hold that global permission.

### SchedulerServiceImpl → `@ExternalAuthorization` (2)

- `updateScheduledJob` → `SCHEDULED_JOB.UPDATE`, symmetric with the adjacent `enableScheduledJob` / `disableScheduledJob`.
- `registerScheduledJob` (5-arg) → `SCHEDULED_JOB.CREATE`. Called from `DiscoveryControllerImpl.scheduleDiscovery`, so it must stay on the external surface. (Initial proposal to refactor back to internal-only was based on missing this caller.)

### ComplianceProfileServiceImpl v1 façade → `@ExternalAuthorization` (13)

All thirteen mirror the annotations already carried by the v2 service, which this façade delegates to. Most are mechanically obvious (DETAIL / CREATE / UPDATE / DELETE on `COMPLIANCE_PROFILE`). Two points worth flagging:

- **`getComplianceRules` and `getComplianceGroups` use `COMPLIANCE_PROFILE.LIST`, not `COMPLIANCE_RULE.LIST` / `COMPLIANCE_GROUP.LIST`.** The `Resource.COMPLIANCE_RULE` and `Resource.COMPLIANCE_GROUP` API codes have no OPA-side counterpart in the internal `core.model.auth.Resource` enum: at context refresh, `ContextRefreshListener.handleContextRefresh` scans every `@ExternalAuthorization` and calls `core.model.auth.Resource.findByCode(...)` on the resource code; the lookup throws "Unknown Resource Name complianceRules" and the application context fails to load. The v2 service annotates the analogous listings with `COMPLIANCE_PROFILE.LIST` for the same reason.
- **`associateProfile` / `disassociateProfile` use a single-resource annotation, not parent/child.** Initial proposal added `parentResource = RA_PROFILE`; dropped to mirror v2's `associateComplianceProfile`, which uses just `COMPLIANCE_PROFILE.UPDATE`.

### Deferred (4) — dynamic resource parameter

- `ResourceServiceImpl.getResourceObjects(Resource, …)`
- `ResourceServiceImpl.updateAttributeContentForObject(Resource, …)`
- `ComplianceServiceImpl.getComplianceCheckResult(Resource, UUID)`
- `ComplianceServiceImpl.checkResourceObjectsComplianceValidation(Resource, List<UUID>)`

All four take a `Resource` parameter that determines the protected resource at runtime. Options for the follow-up: (a) add `@ExternalAuthorizationDynamic(resourceParam = "...", action = ...)` so the annotation processor can resolve the resource from a parameter; (b) split each endpoint into per-resource overloads in the controller layer; (c) accept `@AnyPrincipalEndpoint` and document that authorization is enforced downstream (e.g., `resourceExtensionService.evaluatePermissionChain` for `updateAttributeContentForObject`). No code action in this PR.

### Note on `getResourceObjects` (deferred)

While reviewing, noticed that `getResourceObjects` calls `resourceExtensionService.listResourceObjects(SecurityFilter.create(), …)` — passing a fresh empty `SecurityFilter` instead of one derived from the caller's authorization. This means the downstream listing currently returns every object regardless of caller scope. The same unscoped enumeration also reaches `listResourceRuleFilterFields` (an `@AnyPrincipalEndpoint`) for `LIST` filter fields backed by a `fieldResource`. Tracked as a separate follow-up bug — OmniTrustILM/core#1532 — to be addressed alongside the four deferred markers; it is not changed by this PR.

## ⚠️ Breaking change (release notes)

Scheduling a discovery now requires the `SCHEDULED_JOB.CREATE` permission in addition to discovery permissions. Previously `DiscoveryControllerImpl.scheduleDiscovery` derived its authorization solely from `discoveryService.createDiscovery(...)`; the cross-bean call into the now-externally-authorized `registerScheduledJob` adds a `SCHEDULED_JOB.CREATE` OPA check. No built-in role is affected, but **admins should grant `SCHEDULED_JOB.CREATE` to any custom role that schedules discoveries before upgrading**, otherwise those callers will be denied.

## Test plan

- [x] `mvn -Dmaven.compiler.proc=full clean compile` — passes.
- [x] `mvn -Dmaven.compiler.proc=full test` — 2416 tests, 0 failures, 0 errors, 2 skipped.
- [x] `ExternalServiceAuthorizationArchTest` (the freeze-rule guardian) passes with the freeze store reduced to the four deferred entries.


lubomirw:
Good point regarding dynamic resource parameter for `ExternalAuthorization`. I had it in mind to do something similar by introducing `SecuredResource` class or similar analogy to `SecuredUuid`. But resourceParam could also work since parameters of method are anyway iterated.

👍

src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
```diff
@@ -148,7 +149,7 @@ public List<ResponseAttribute> updateAttributeContentForObject(Resource resource
     }
 
     @Override
-    @ExternalAuthorizationMissing
+    @AnyPrincipalEndpoint
```
lubomirw, line 152: **~ Minor** | Security

`listResourceRuleFilterFields` is being marked permanently `@AnyPrincipalEndpoint`, but for any `LIST` filter field backed by a `fieldResource` it calls `getResourceObjects(filterField.getFieldResource(), null, null)` (line 173) to populate the available filter values — and that method (deferred in this PR) passes a fresh empty `SecurityFilter`, so it returns every object regardless of caller scope. Concretely, `RA_PROFILE_NAME` is a `LIST` field on `CERTIFICATE` backed by `Resource.RA_PROFILE`, so `listResourceRuleFilterFields(CERTIFICATE, …)` hands the full set of RA-profile names + UUIDs to any authenticated caller, with no OPA check.

This doesn't change runtime behavior (it matches the old `@ExternalAuthorizationMissing`), so it needn't block this PR. But please open a follow-up issue to scope the per-filter object listing to the caller's permissions — i.e. derive a real `SecurityFilter` for the `getResourceObjects` call backing each `fieldResource` filter, so the filter-value enumeration only returns objects the caller is allowed to see. That follow-up should be linked to the same work that addresses the deferred `getResourceObjects` marker.

src/main/java/com/czertainly/core/service/impl/SchedulerServiceImpl.java
```diff
@@ -229,7 +228,7 @@ public ScheduledJobDetailDto registerScheduledJob(final Class<? extends Schedule
     }
 
     @Override
-    @ExternalAuthorizationMissing
+    @ExternalAuthorization(resource = Resource.SCHEDULED_JOB, action = ResourceAction.CREATE)
```
lubomirw, line 231: **! Major** | Security

Keeping `registerScheduledJob` on the external surface is the right call, but note it changes the authorization contract of its only non-system caller. `DiscoveryControllerImpl.scheduleDiscovery` (line 112) carries no `@ExternalAuthorization` of its own — its authorization today comes solely from `discoveryService.createDiscovery(...)` (discovery permissions). After this change, the cross-bean call into `SchedulerExternalService` triggers an OPA check, so scheduling a discovery additionally requires `SCHEDULED_JOB.CREATE`.

No built-in role schedules anything, so nothing ships broken — but any customer-defined custom role that can schedule discoveries today without holding `SCHEDULED_JOB.CREATE` will start getting denied after upgrade. The annotation itself is correct (registering a job is genuinely a `SCHEDULED_JOB` create).

Could you add a note about this to the PR body so it gets picked up for the release notes? Something like: "Scheduling a discovery now requires the `SCHEDULED_JOB.CREATE` permission in addition to discovery permissions. Admins should grant `SCHEDULED_JOB.CREATE` to any custom role that schedules discoveries before upgrading." That way the breaking change is documented for operators updating their custom roles.

lubomirw (commented):
Code review: 2 issue(s) found. Both are non-blocking comments — the diff itself is sound and the annotation choices are well-reasoned. These concern permanent classifications/migration notes, not defects introduced here.

lubomirw (approved):

src/main/java/com/czertainly/core/service/impl/ResourceServiceImpl.java
ivosh: Good catch — agreed this needn't block the PR (no runtime change vs. the old `@ExternalAuthorizationMissing`). Opened a follow-up to scope the per-filter object listing to the caller's permissions (derive a real `SecurityFilter` for the `getResourceObjects` call backing each `fieldResource` filter): OmniTrustILM/core#1532. Linked it to the deferred `getResourceObjects` marker work, and referenced it from the PR body.

src/main/java/com/czertainly/core/service/impl/SchedulerServiceImpl.java
ivosh: Agreed — the annotation is correct and the contract change is real. Added a **Breaking change (release notes)** section to the PR body noting that scheduling a discovery now additionally requires `SCHEDULED_JOB.CREATE`, and advising admins to grant it to any custom role that schedules discoveries before upgrading.

PR #1528, opened 2026-05-28
Resolve object recipients when sending notifications

### LINKED ISSUE #1511: Implement Mapped recipient resolution in NotificationListener
### Summary
Wire up the `MAPPED` recipient type introduced in interfaces. When a notification profile uses `MAPPED`, the triggering resource object itself becomes the recipient: its custom attributes are fetched and passed through the existing mapping attributes mechanism to produce the contact details the connector needs. No new DB column or service-level validation is required — the profile stores only `recipientType=MAPPED` with null `recipientUuids`.

### Changes
- **`NotificationListener.getRecipients()`** — add `MAPPED` case: return `new NotificationRecipient(MAPPED, objectUuid)`
- **`NotificationListener.sendExternalNotifications()`** — for `MAPPED` recipients, use the triggering object's `resource` and `recipientUuid` (= `objectUuid`) when calling `getObjectCustomAttributesContent()`, instead of `recipientType.getRecipientResource()`; the rest of the `getMappedAttributes()` path is unchanged
- **`NotificationListener.constructNotificationRecipientDto()`** — add `MAPPED` case returning an empty `NotificationRecipientDto` (contact details come from mapped attributes, not from a platform entity)
- **`NotificationMappedContactIntegrationTest`** — integration test covering: certificate with custom attribute mapped to connector attribute (value appears in notify request), and certificate without the attribute set (connector still called, no mapped value)

### User Story

_No response_

### Use Case

_No response_

### Acceptance Criteria

- Profile with `MAPPED` type sends to all valid email addresses in the named attribute

klaraf755:
Closes #1511 
Also fixes mapped-attribute resolution for USER/ROLE/GROUP recipients, which were previously filtered to empty.


src/main/java/com/czertainly/core/messaging/jms/listeners/NotificationListener.java
```diff
@@ -200,6 +200,10 @@ private boolean proceedWithNotifying(NotificationProfileVersion notificationProf
     }
 
     private List<NotificationRecipient> getRecipients(RecipientType recipientType, List<UUID> recipientUuids, ResourceEvent event, Object data, Resource resource, UUID objectUuid) {
+        if (recipientType == RecipientType.MAPPED) {
+            return List.of(new NotificationRecipient(RecipientType.MAPPED, objectUuid));
```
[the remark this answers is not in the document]
klaraf755: I would keep it as it is to be consistent with other unsupported recipient types.

src/main/java/com/czertainly/core/messaging/jms/listeners/NotificationListener.java
```diff
@@ -285,8 +289,10 @@ private void sendExternalNotifications(UUID notificationInstanceUUID, List<Notif
                     continue;
                 }
 
-                List<ResponseAttribute> recipientCustomAttributes = attributeEngine.getObjectCustomAttributesContent(recipient.getRecipientType().getRecipientResource(), recipient.getRecipientUuid());
-
+                Resource customAttributeResource = recipient.getRecipientType() == RecipientType.MAPPED
+                        ? resource
+                        : recipient.getRecipientType().getRecipientResource();
+                List<ResponseAttribute> recipientCustomAttributes = attributeEngine.getObjectCustomAttributesContentNoAuth(customAttributeResource, recipient.getRecipientUuid());
```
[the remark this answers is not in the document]
klaraf755: Unauthenticated lookup is correct here, since the access to the custom attributes has been resolved when defining mapping attributes.

src/main/java/com/czertainly/core/attribute/engine/AttributeEngine.java
```diff
@@ -721,6 +721,11 @@ public List<ResponseAttribute> getObjectCustomAttributesContent(Resource objectT
         return getObjectCustomAttributesContent(objectType, objectUuid, securityResourceFilter);
     }
 
+    public List<ResponseAttribute> getObjectCustomAttributesContentNoAuth(Resource objectType, UUID objectUuid) {
```
lubomirw: **! Major** | Maintainability

I'd push back a bit on adding `getObjectCustomAttributesContentNoAuth` as a new public method on `AttributeEngine`. The reason it exists here is specific: the notification listener runs from a JMS message handler, so there's no `SecurityContext` on the thread, and the auth-filtered variant ends up applying the "anonymous user" branch in `loadCustomAttributesSecurityResourceFilter()` — which produces a restrictive filter and effectively returns nothing. So the bypass is justified for this caller.

What concerns me is that, once it lives on `AttributeEngine` as a `public` method, anyone wiring up a new service will see it in autocomplete next to the auth-checked one and pick whichever has the friendlier name. There's nothing in the signature that tells them this is meant for system-context paths only.

Two things would help future readers:

1. Rename it to something that names the intent rather than the absence of a check — e.g. `getObjectCustomAttributesContentForSystemContext` or `...Unfiltered`. The `NoAuth` suffix reads as a smell but doesn't actually stop misuse.
2. Add a Javadoc that explicitly says: this is for code paths where there is no authenticated user (JMS listeners, scheduled jobs, etc.), and that any code reached from a controller must use the auth-checked variant.

Not blocking — the method does the right thing for this caller — but worth tightening before it grows new callers.

src/main/java/com/czertainly/core/messaging/jms/listeners/NotificationListener.java
```diff
@@ -279,17 +283,26 @@ private void sendExternalNotifications(UUID notificationInstanceUUID, List<Notif
             logger.debug("Processing recipient {} of type {}.", recipient.getRecipientUuid(), recipient.getRecipientType());
             try {
                 // construct recipient DTO
-                NotificationRecipientDto recipientDto = constructNotificationRecipientDto(recipient, notificationInstanceReference.getKind());
+                NotificationRecipientDto recipientDto = constructNotificationRecipientDto(recipient, notificationInstanceReference.getKind(), resource);
                 if (recipientDto == null) {
                     // this should happen only in case of recipient type NONE
                     continue;
                 }
```
lubomirw, line 290: **! Major** | Correctness

Switch to `NoAuth` changes behavior for ALL recipient types, not just MAPPED.

The previous call routed through `loadCustomAttributesSecurityResourceFilter()`. JMS listener has no `SecurityContext`, so that filter hits the anonymous-user branch and returns no attributes. The old code was almost certainly returning an empty list for USER/ROLE/GROUP recipients too — mapped attributes never actually resolved in this path.

This PR fixes that latent bug while adding MAPPED. Right outcome, but worth surfacing:

1. **PR description** — add a note that the switch also fixes mapped-attribute resolution for USER/ROLE/GROUP recipients, which were previously filtered to empty.

2. **Inline comment** above the call:

   ```java
   // No SecurityContext in JMS listener — auth-filtered variant returns nothing.
   // Use system-context method so mapped attributes resolve for all recipient types.
   ```

src/main/java/com/czertainly/core/messaging/jms/listeners/NotificationListener.java
```diff
@@ -279,17 +283,26 @@ private void sendExternalNotifications(UUID notificationInstanceUUID, List<Notif
             logger.debug("Processing recipient {} of type {}.", recipient.getRecipientUuid(), recipient.getRecipientType());
             try {
                 // construct recipient DTO
-                NotificationRecipientDto recipientDto = constructNotificationRecipientDto(recipient, notificationInstanceReference.getKind());
+                NotificationRecipientDto recipientDto = constructNotificationRecipientDto(recipient, notificationInstanceReference.getKind(), resource);
                 if (recipientDto == null) {
```
lubomirw, line 287: **~ Minor** | Maintainability

MAPPED recipient silently dropped when mapped attributes missing.

For MAPPED, `customAttributeResource` is the triggering event's `resource`, and the lookup uses `(triggerResource, objectUuid)` to fetch custom attributes. If no custom attribute content is found — either because nothing is set on the object, or because the configured `attributeDefinitionUuid` points to an attribute defined on a different resource than the trigger — `getMappedAttributes()` either `continue`s (if `required=false`) or throws `ValidationException` (if `required=true`), and one level up the `catch (Exception e)` skips the recipient entirely. The operator sees no notification and no signal in logs that anything went wrong.

Suggest adding a `warn`-level log when a MAPPED recipient ends up with no mapped attributes resolved (or when it gets dropped because the required attribute was missing), so misconfiguration surfaces in logs instead of as a silent missing notification. The existing `logger.warn(...)` in the `catch` already covers the exception path; the gap is the `required=false` + missing-content case where nothing is logged at all.

src/main/java/com/czertainly/core/messaging/jms/listeners/NotificationListener.java
```diff
@@ -279,17 +283,26 @@ private void sendExternalNotifications(UUID notificationInstanceUUID, List<Notif
...
                     continue;
                 }
 
-                List<ResponseAttribute> recipientCustomAttributes = attributeEngine.getObjectCustomAttributesContent(recipient.getRecipientType().getRecipientResource(), recipient.getRecipientUuid());
-
+                Resource customAttributeResource = recipient.getRecipientType() == RecipientType.MAPPED
+                        ? resource
+                        : recipient.getRecipientType().getRecipientResource();
+                // Unauthenticated lookup is correct here, since the access to the custom attributes has been resolved when defining mapping attributes.
+                List<ResponseAttribute> recipientCustomAttributes = attributeEngine.getObjectCustomAttributesContentNoAuth(customAttributeResource, recipient.getRecipientUuid());
```
lubomirw: **~ Minor** | Maintainability

Comment placement after `case MAPPED ->` reads awkwardly.

Putting the `//` comment on the same line as the arrow, with the assignment dangling on the next line, makes the case harder to parse at a glance — every other arm uses either a block or a single expression on the same line as `->`. The comment itself is valuable (explains *why* the DTO has no name/email), only placement is the concern.

Prefer either:

```java
case MAPPED -> {
    // Connector resolves contact details via mapped attributes — no name/email to set here.
    recipientDto = new NotificationRecipientDto();
}
```

or move the comment above the case label:

```java
// MAPPED: connector resolves contact details via mapped attributes — no name/email to set here.
case MAPPED -> recipientDto = new NotificationRecipientDto();
```

src/test/java/com/czertainly/core/events/NotificationMappedIntegrationTest.java
```diff
@@ -0,0 +1,162 @@
...
+        mockServer = new WireMockServer(0);
+        mockServer.start();
+        WireMock.configureFor("localhost", mockServer.port());
+
+        // Connector declares one string mapping attribute
+        mockServer.stubFor(WireMock.get(WireMock.urlPathMatching("/v1/notificationProvider/[^/]+/attributes/mapping"))
+                .willReturn(WireMock.okJson("""
+                        [{"uuid": "%s", "name": "recipientContact", "type": "data", "version": 3,
+                          "contentType": "string", "properties": {"required": false}}]
+                        """.formatted(MAPPING_ATTRIBUTE_UUID))));
```
lubomirw, line 75: **~ Minor** | Testing

No test coverage for `required=true` mapping attribute path.

Both existing tests stub the connector's mapping attribute as `"required": false`. That means the second test (`...connectorCalledWithoutMappedAttribute`) only exercises the `continue` branch inside `getMappedAttributes()`. The other branch — where `mappingAttribute.getProperties().isRequired()` is true and the recipient's custom attribute is missing — throws `ValidationException`, which is caught in `sendExternalNotifications` and the recipient is silently skipped.

That branch is meaningful for the feature: it's the difference between "operator misconfigured the profile and the platform fails loudly" and "operator misconfigured the profile and the connector is called with no mapped value." Worth adding a third test with `"required": true` + no attribute content on the certificate, asserting either the warn log is produced or the connector is called with no recipients (whichever you decide is correct behavior per the related comment above).

lubomirw (commented):

src/main/java/com/czertainly/core/messaging/jms/listeners/NotificationListener.java
klaraf755: Added both

src/main/java/com/czertainly/core/attribute/engine/AttributeEngine.java
klaraf755: Renamed and added javadoc

src/main/java/com/czertainly/core/messaging/jms/listeners/NotificationListener.java
klaraf755: Added the log for empty `mappedAttributes`

src/main/java/com/czertainly/core/messaging/jms/listeners/NotificationListener.java
klaraf755: It has been moved.

src/test/java/com/czertainly/core/events/NotificationMappedIntegrationTest.java
klaraf755: Added a test

lubomirw (approved):

PR #1531, opened 2026-05-29
Add Signing Profile cache

### LINKED ISSUE #1529: Cache Signing Profile to avoid database lookups
### Description

Caffeine cache for SigningProfileModel lookups used by timestamping. SigningProfileModel record captures only the data needed by the hot path.

ivosh:


src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -302,7 +366,11 @@ public SigningProfileDto updateSigningProfile(SecuredUUID uuid, SigningProfileRe
         validateSigningSchemeCoherence(request.getSigningScheme());
         attributeEngine.validateCustomAttributesContent(Resource.SIGNING_PROFILE, request.getCustomAttributes());
         List<BaseAttribute> formatterDefinitions = fetchFormatterAttributeDefinitions(request.getWorkflow());
-        return self.persistUpdate(uuid, request, formatterDefinitions);
+        String oldName = findByUuid(uuid).getName();
+        SigningProfileDto updated = self.persistUpdate(uuid, request, formatterDefinitions);
+        evictSigningProfileCache(oldName);
+        evictSigningProfileCache(updated.getName());
```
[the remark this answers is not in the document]
ivosh: Fixed in 87345043. The previous name is now captured inside `persistUpdate` after the advisory lock (`oldName = profile.getName()`), and both names are evicted there — so concurrent renames evict the committed source name rather than a stale pre-lock value. Eviction still happens post-commit via the transaction's afterCommit synchronization.

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -235,6 +241,48 @@ public SigningProfile getSigningProfileEntity(SecuredUUID uuid) throws NotFoundE
...
 
+    @Override
+    @ExternalAuthorization(resource = Resource.SIGNING_PROFILE, action = ResourceAction.DETAIL)
+    public SigningProfileModel<?, ?> getSigningProfileModel(String name) throws NotFoundException {
+        return self.loadSigningProfileModel(name);
+    }
+
+    @Cacheable(value = CacheConfig.SIGNING_PROFILE_CACHE, key = "#name", sync = true)
+    @Transactional(readOnly = true)
+    public SigningProfileModel<?, ?> loadSigningProfileModel(String name) throws NotFoundException {
```
lubomirw: **. Nitpick** | Maintainability

Question for consideration, not necessarily an issue: `loadSigningProfileModel` is `public` while being the cache-backed internal of `getSigningProfileModel` (which holds the `@ExternalAuthorization`), and it isn't on the `SigningProfileService` interface. Is `public` intentional here, or would package-private be preferable? Spring's self-proxy invocation works with package-private too, as `TimeQualityConfigurationServiceImpl.loadTimeQualityConfigurationModel` does. That said, `TspProfileServiceImpl.loadTspProfileModel` is `public`, so keeping it `public` for symmetry with TSP is also a defensible choice — flagging only so the visibility is a deliberate decision rather than incidental.

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -235,6 +241,48 @@ public SigningProfile getSigningProfileEntity(SecuredUUID uuid) throws NotFoundE
...
+                ObjectAttributeContentInfo.builder(Resource.SIGNING_PROFILE, profile.getUuid())
+                        .operation(AttributeOperation.SIGN)
+                        .version(currentVersion.getVersion()).build());
+        List<RequestAttribute> signatureFormatterConnectorAttributes = attributeEngine.getRequestObjectDataAttributesContent(
+                ObjectAttributeContentInfo.builder(Resource.SIGNING_PROFILE, profile.getUuid())
+                        .connector(currentVersion.getSignatureFormatterConnectorUuid())
+                        .operation(AttributeOperation.WORKFLOW_FORMATTER)
+                        .version(currentVersion.getVersion()).build());
+
+        // Narrow scope: only managed-timestamping profiles are cacheable for now.
```
lubomirw, line 272: **. Nitpick** | Performance

Forward-looking note (no production caller exists yet, so no impact today): `getSigningProfileModel` returns no cache entry for non-managed-timestamping profiles. `toManagedTimestampingModel` throws for any non-`TIMESTAMPING`/non-`MANAGED` profile, and `@Cacheable(sync = true)` does not store exceptions, so each such call re-runs `findByName` + both `getRequestObjectDataAttributesContent` fetches before throwing. Such profiles are configurable (cf. `buildDelegatedTimestampingRequest`). When the consumer of this method is wired up, consider pre-filtering by profile type before the attribute fetches so non-managed profiles don't repeatedly miss the cache and hit the DB.

lubomirw (commented):
Code review: 2 issue(s) found — both advisory nitpicks; the change is clean and well-tested.

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
ivosh: Fixed in 0f6ff3da — made package-private. It does not need to stay public: it is a self-invoked internal cache loader (reached only via `self.loadSigningProfileModel`, where `self` is typed as the concrete impl, so package-private compiles and CGLIB still proxies it — same as `TimeQualityConfigurationServiceImpl.loadTimeQualityConfigurationModel`).

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
ivosh: Agreed — leaving this as-is for now. There is no production caller yet, and the cache key/loader shape is likely to change when the consumer is wired up (cf. `feat/signing`, where the loader narrows to managed-timestamping profiles and pre-resolves the type before the attribute fetches). Pre-filtering by profile type to avoid the repeated non-managed cache miss is the right move, but it belongs with that consumer wiring rather than speculatively here. Keeping your note for that follow-up.

lubomirw (approved):

PR #1541, opened 2026-06-02
Fix orphaned notification instances breaking notification profiles

### LINKED ISSUE #1505: Settings - Notifications: Orphaned Notification Instance cascades into broken Notification Profile — both records unopenable, instance undeletable
### Description

A Notification Instance can end up in a "zombie" state where Core still lists it on the **Notification Instances** page but the underlying connector (in this case `Webhook-Notification-Provider`) no longer has the record. As a result:

- The instance is rendered in the list as if it existed
- Bulk delete from the list returns HTTP **422** because the instance is still referenced by a Notification Profile
- Opening the **Details** page returns HTTP **404** with `NotFoundException` from the connector
- Opening the **Notification Profile** that references this instance also returns HTTP **404** — the failure cascades upward
- The user has no path to resolve this through the UI: cannot view, cannot delete, cannot re-sync, and the referencing profile is also broken

There is no UI indication that the instance is orphaned — it looks like any other instance until the user interacts with it.

### Steps to Reproduce

1. Navigate to **Settings → Notifications → Notification Instances** (Notification Store page)
2. Select the affected instance via its checkbox and click **Delete** at the top of the list — **422** error returned
3. From the same list, click the instance row to open its **Details** page — **404** error returned
4. Navigate to **Settings → Notifications → Notification Profiles** and open the profile that references this instance — **404** error returned

### Expected Behavior

- Core and connector data stay consistent — an instance shown in the list always has retrievable details
- Profiles that reference the instance should still open and surface a clear warning (e.g. "linked instance is unavailable") rather than failing outright
- If the connector loses the record (manual cleanup, connector reset, etc.), the platform should:
  - Detect the inconsistency and clearly mark the instance as orphaned in the list, AND/OR
  - Allow force-deletion of orphans (with explicit confirmation), AND/OR
  - Offer a re-sync / re-create action
- The user always has a recovery path; no record gets stuck permanently; failures should not cascade to referencing profiles

### Actual Behavior

- **Notification Instances list:** orphaned instance is rendered like any other — no visual indication of inconsistency
- **Bulk Delete from list (checkbox + Delete button):** returns HTTP **422**:
  > *"Failed to delete notification instance (422): Cannot delete notification instance. 1 notification profile version(s) are referencing this notification instance"*

- **Instance Details page:** returns HTTP **404** with response payload:

```json
{
  "message": "...Object of type 'NotificationInstance' identified by 54021af7-84e1-4114-93af-b06e4a287768 is not found...
   Error is related to connector name=Webhook-Notification-Provider, uuid=4a3d3ec6-83bf-4075-97f2-8077ed96bffb."
}
```

Stack trace points to `com.czertainly.np.webhook.service.impl.NotificationInstanceServiceImpl.getNotificationInstance` (line 98) — the webhook provider does not have this UUID.

- **Referencing Notification Profile details:** also returns HTTP **404** — the orphan propagates the failure upward, so the profile that links to this instance also becomes unopenable
- The instance is permanently stuck:
  - Cannot delete (422 — referenced by a Notification Profile)
  - Cannot open details (404 — connector lost the record)
  - Cannot use (the referencing profile is also broken)
- The blast radius extends beyond the orphan itself to any profile that depends on it

### Severity

Critical

### Module

Core

### Screenshot

Image

https://github.com/user-attachments/assets/692c4d05-c67b-47c6-8faa-f34f8c46d552

### Logs

_No response_

### Environment

ILM Core: 2.17.1-SNAPSHOT (25 May 2026, 09:45)
ILM Frontend: 2.17.0 (25 May 2026, 08:49)
Client: macOS Tahoe 26.3.1 / Chrome 148.0.7778.179 (arm64)

klaraf755:
Closes #1505


src/main/java/com/czertainly/core/service/impl/NotificationInstanceServiceImpl.java
```diff
@@ -260,24 +275,41 @@ private void updateMappedAttributes(NotificationInstanceReference savedInstance,
...
         }
 
-        // check notification profiles referencing notification instance
-        Long referencesCount = notificationProfileVersionRepository.countByNotificationInstanceRefUuid(notificationInstanceRef.getUuid());
-        if (referencesCount > 0) {
-            throw new ValidationException("Cannot delete notification instance. %d notification profile version(s) are referencing this notification instance".formatted(referencesCount));
+        // Read before writing: check current-version references that block deletion.
+        // Historical versions are auto-unlinked, but current versions require the user to update the profile first.
+        // Orphaned instances skip this check — all references are unlinked unconditionally.
+        if (!isOrphaned) {
```
lubomirw: **! Major** | Correctness

Blocking-reference check runs *after* the connector-side delete, which can create the very orphan this PR fixes.

In `removeNotificationInstance`, the connector instance is removed (line ~282) before the current-version reference check at line 298. When a current-version profile blocks deletion, the `ValidationException` rolls back the core DB transaction — but the connector-side HTTP delete is an external side effect that is **not** rolled back. Result: the instance is gone in the connector while core keeps the `NotificationInstanceReference` → a fresh orphan. Consider moving the blocking check (and ideally the whole reference resolution) *before* the connector delete so a blocked deletion leaves both sides untouched. Note this ordering predates the PR, but the PR rewrites this exact block.

src/test/java/com/czertainly/core/service/writer/NotificationProfileWriter.java
```diff
@@ -0,0 +1,4 @@
+package com.czertainly.core.service.writer;
+
+public class NotificationProfileWriter {
```
lubomirw: **~ Minor** | Maintainability

Stray empty placeholder class — looks like an accidental commit.

`NotificationProfileWriter` has an empty body, no annotations, no usages. It appears unrelated to the fix (note also the name differs from the real `NotificationProfileVersionWriter`). Please delete this file before merge.

src/main/java/com/czertainly/core/dao/repository/notifications/NotificationProfileVersionRepository.java
```diff
@@ -3,20 +3,51 @@
...
+    void detachHistoricalInstanceReferences(@Param("uuid") UUID uuid);
+
+    @Query("""
+            UPDATE NotificationProfileVersion npv SET npv.notificationInstanceRefUuid = NULL
+            WHERE npv.notificationInstanceRefUuid = :uuid
+            """)
+    @Modifying
+    int detachNotificationInstanceRefUuid(@Param("uuid") UUID uuid);
 }
+
```
lubomirw: **. Nitpick** | Maintainability

Extra blank line after the closing brace at end of file. Drop it.

src/main/java/com/czertainly/core/service/writer/NotificationProfileVersionWriter.java
```diff
@@ -0,0 +1,29 @@
...
+
+    private final NotificationProfileVersionRepository notificationProfileVersionRepository;
+
+    @Autowired
+    public NotificationProfileVersionWriter(NotificationProfileVersionRepository notificationProfileVersionRepository) {
+        this.notificationProfileVersionRepository = notificationProfileVersionRepository;
+    }
+
+    @Transactional
+    public void detachHistoricalInstanceReferencesByNotificationInstanceRefUuid(UUID notificationInstanceRefUuid) {
```
lubomirw, line 21: **. Nitpick** | Maintainability

Wrapper/repo method names are inconsistent. The writer method `detachHistoricalInstanceReferencesByNotificationInstanceRefUuid` wraps repo `detachHistoricalInstanceReferences`, while the other pair (`detachNotificationInstanceRefUuid`) shares the same name on both. Align the two for consistency — either keep the `...ByNotificationInstanceRefUuid` suffix on both wrappers or drop it from both.

lubomirw (changes requested):
Code review: 4 issue(s) found (1 major, 1 minor blocking, 2 nitpicks).

src/main/java/com/czertainly/core/service/impl/NotificationInstanceServiceImpl.java
klaraf755: Moved the validation check at the start of method.

src/main/java/com/czertainly/core/dao/repository/notifications/NotificationProfileVersionRepository.java
klaraf755: Done

src/main/java/com/czertainly/core/service/writer/NotificationProfileVersionWriter.java
klaraf755: Done

src/test/java/com/czertainly/core/service/writer/NotificationProfileWriter.java
klaraf755: Removed

lubomirw (approved):

PR #1544, opened 2026-06-03
Fix custom attribute condition evaluation and negated operators in TriggerEvaluator

### LINKED ISSUE #1515: Workflows - Conditions: Property "Common Name" with CONTAINS operator yields false-positive — matches certs whose CN does not contain the value
### Description

Discovered while testing Epic [#115 — Categorization of uploaded certificates](https://github.com/OmniTrustILM/ilm/issues/115). An ignore-trigger configured with `Property → Common Name → CONTAINS → test` rejects certificates whose Common Name does not contain the substring `test`. The substring check appears to match against more than just the CN field (possibly the entire encoded Subject DN, the full cert blob, or something else), which breaks the entire selective-reject feature: any cert with the wrong substring elsewhere gets blocked regardless of CN.

### Steps to Reproduce

1. Generate a self-signed cert with a CN that **does not contain** the literal substring `test`. Example (HARICA root, used in our repro):
   ```
   Subject: CN=HARICA Client ECC Root CA 2021, O=Hellenic Academic and Research Institutions CA, C=GR
   ```
2. Create a Workflow **Condition**: Resource=Certificate, Filter Field Source=**Property**, Filter Field=**Common Name**, Filter Condition=**CONTAINS**, Filter Value=`test`
3. Wrap in a **Rule**, attach to an **ignore-trigger** on event `Certificate Uploaded`
4. Upload the cert from step 1 via Certificate Inventory → Upload Certificate

### Expected Behavior

- Cert is added to inventory (CN does not contain `test` — condition should not match)
- Event Firing Details shows the condition evaluating to `false`

### Actual Behavior

- Cert is **rejected** — Event History shows `Obj Ignored = 1` for the trigger named "qa trigger-ignore certificate with common names contains test"
- Cert never reaches inventory
- Same trigger correctly rejects cert with CN containing `test` — so the trigger is wired up, the matching logic itself is broken

### Severity

Major

### Module

Core

### Screenshot

_No response_

### Logs

_No response_

### Environment

ILM Core: 2.17.1-SNAPSHOT (25 May 2026, 16:33)
ILM Frontend: 2.17.0 (25 May 2026, 17:59)
Client: macOS Tahoe 26.3.1 / Chrome 148.0.7778.179 (arm64)

klaraf755:
Closes #1515

### DIFF
--- a/src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
+++ b/src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
@@ -265,7 +265,7 @@ public class TriggerEvaluator<T extends UniquelyIdentifiedObject> implements ITr
     private boolean evaluateMetaAttributeConditionItem(Resource resource, String fieldIdentifier, UUID objectUuid, Object conditionValue, FilterConditionOperator operator) throws RuleException {
         // If the Field Source is Meta Attribute, we expect Field Identifier to be formatted as follows 'name|contentType', since there can be multiple Meta Attributes with the same name, the Content Type must be specified
         String[] parts = parseNameAndContentType(fieldIdentifier);
-        AttributeContentType fieldAttributeContentType = AttributeContentType.valueOf(parts[1]);
+        AttributeContentType fieldAttributeContentType = parseAttributeContentType(fieldIdentifier);
         String fieldIdentifierName = parts[0];
         // From all Metadata of the object, find those with matching Name and Content Type and evaluate condition on these, return true for the first satisfying attribute, otherwise continue wit next
         List<MetadataResponseDto> metadata = attributeEngine.getMappedMetadataContent(ObjectAttributeContentInfo.builder(resource, objectUuid).build());
@@ -339,14 +339,9 @@ public class TriggerEvaluator<T extends UniquelyIdentifiedObject> implements ITr
         }
 
         final String sourceName;
-        final AttributeContentType sourceContentType;
         String[] sourceParts = parseNameAndContentType(executionItem.getSourceFieldIdentifier());
         sourceName = sourceParts[0];
-        try {
-            sourceContentType = AttributeContentType.valueOf(sourceParts[1]);
-        } catch (IllegalArgumentException e) {
-            throw new RuleException("Cannot parse source field identifier %s from execution item: %s".formatted(executionItem.getSourceFieldIdentifier(), e.getMessage()));
-        }
+        final AttributeContentType sourceContentType = parseAttributeContentType(sourceParts[1]);
         FilterFieldSource sourceFieldSource = executionItem.getSourceFieldSource();
 
         List<BaseAttributeContentV3<?>> content = switch (sourceFieldSource) {
@@ -486,6 +481,15 @@ public class TriggerEvaluator<T extends UniquelyIdentifiedObject> implements ITr
             throw new RuleException("Field identifier is not in correct format 'name|contentType', got: " + fieldIdentifier);
         }
         return parts;
+    }
+
+    private AttributeContentType parseAttributeContentType(String contentType) throws RuleException {
+        AttributeContentType sourceContentType;
+        try {
+            return sourceContentType = AttributeContentType.valueOf(contentType);
+        } catch (IllegalArgumentException e) {
+            throw new RuleException("Cannot parse content type %s from execution item: %s".formatted(contentType, e.getMessage()));
+        }
     }
 
     private static final Map<FilterConditionOperator, BiFunction<Object, Object, Boolean>> commonOperatorFunctionMap;
@@ -637,20 +641,21 @@ public class TriggerEvaluator<T extends UniquelyIdentifiedObject> implements ITr
         };
         boolean isNegated = effectiveOperator != operator;
 
-        // If the attribute is a list, iterate through each item and check if any match the condition.
+        // If the attribute is a list, iterate through each item and short-circuit on the first definitive result.
         // If the attribute is not a list, there is only one item in the content list, so only one check will be done.
-        boolean contentMatched = false;
         for (AttributeContent attributeContent : content) {
             Object attributeValue = contentType.isFilterByData() ? attributeContent.getData() : attributeContent.getReference();
             try {
                 if (Boolean.TRUE.equals(fieldTypeToOperatorActionMap.get(contentTypeToFieldType(contentType)).get(effectiveOperator).apply(attributeValue, conditionValue)))
-                    contentMatched = true;
+                    // Positive match found: for non-negated ops return true, for negated ops return false (a match disqualifies NOT_EQUALS/NOT_CONTAINS/NOT_MATCHES)
+                    return !isNegated;
             } catch (Exception e) {
                 throw new RuleException("Cannot evaluate operator %s on attribute value '%s' with condition value '%s' (contentType: %s): %s"
                         .formatted(operator, attributeValue, conditionValue, contentType, e.getMessage()));
             }
         }
-        return isNegated != contentMatched;
+        // No positive match found: for non-negated ops return false, for negated ops return true
+        return isNegated;
     }
 
 
--- a/src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
+++ b/src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
@@ -265,7 +265,7 @@ public class TriggerEvaluator<T extends UniquelyIdentifiedObject> implements ITr
     private boolean evaluateMetaAttributeConditionItem(Resource resource, String fieldIdentifier, UUID objectUuid, Object conditionValue, FilterConditionOperator operator) throws RuleException {
         // If the Field Source is Meta Attribute, we expect Field Identifier to be formatted as follows 'name|contentType', since there can be multiple Meta Attributes with the same name, the Content Type must be specified
         String[] parts = parseNameAndContentType(fieldIdentifier);
-        AttributeContentType fieldAttributeContentType = parseAttributeContentType(fieldIdentifier);
+        AttributeContentType fieldAttributeContentType = parseAttributeContentType(parts[1]);
         String fieldIdentifierName = parts[0];
         // From all Metadata of the object, find those with matching Name and Content Type and evaluate condition on these, return true for the first satisfying attribute, otherwise continue wit next
         List<MetadataResponseDto> metadata = attributeEngine.getMappedMetadataContent(ObjectAttributeContentInfo.builder(resource, objectUuid).build());
@@ -484,9 +484,8 @@ public class TriggerEvaluator<T extends UniquelyIdentifiedObject> implements ITr
     }
 
     private AttributeContentType parseAttributeContentType(String contentType) throws RuleException {
-        AttributeContentType sourceContentType;
         try {
-            return sourceContentType = AttributeContentType.valueOf(contentType);
+            return AttributeContentType.valueOf(contentType);
         } catch (IllegalArgumentException e) {
             throw new RuleException("Cannot parse content type %s from execution item: %s".formatted(contentType, e.getMessage()));
         }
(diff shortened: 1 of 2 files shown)


src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
```diff
@@ -283,9 +282,11 @@ private boolean evaluateMetaAttributeConditionItem(Resource resource, String fie
     }
 
     private boolean evaluateCustomAttributeConditionItem(Resource resource, UUID objectUuid, String fieldIdentifier, Object conditionValue, FilterConditionOperator operator) throws RuleException {
-        // If source is Custom Attribute, retrieve custom attributes of this object and find the attribute which has Name equal to Field Identifier
+        // If source is Custom Attribute, Field Identifier is formatted as 'name|contentType'; extract name to look up the attribute
+        String[] parts = parseNameAndContentType(fieldIdentifier);
+        String attributeName = parts[0];
         List<ResponseAttribute> responseAttributes = attributeEngine.getObjectCustomAttributesContent(resource, objectUuid);
-        ResponseAttributeV3 attributeToCompare = (ResponseAttributeV3) responseAttributes.stream().filter(rad -> Objects.equals(rad.getName(), fieldIdentifier)).findFirst().orElse(null);
+        ResponseAttributeV3 attributeToCompare = (ResponseAttributeV3) responseAttributes.stream().filter(rad -> Objects.equals(rad.getName(), attributeName)).findFirst().orElse(null);
```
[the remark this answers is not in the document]
klaraf755: Since custom attributes have unique name, content type does not need to be taken into consideration.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
```diff
@@ -615,16 +620,37 @@ private static int compareNumbers(Number objectNumber, Object conditionNumber) {
...
+            case NOT_MATCHES -> FilterConditionOperator.MATCHES;
+            default -> operator;
+        };
+        boolean isNegated = effectiveOperator != operator;
+
+        // If the attribute is a list, iterate through each item and check if any match the condition.
+        // If the attribute is not a list, there is only one item in the content list, so only one check will be done.
+        boolean contentMatched = false;
         for (AttributeContent attributeContent : content) {
             Object attributeValue = contentType.isFilterByData() ? attributeContent.getData() : attributeContent.getReference();
```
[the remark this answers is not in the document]
klaraf755: Fixed

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
```diff
@@ -264,10 +264,9 @@ private boolean evaluateItemsInCollection(FilterConditionOperator operator, Obje
 
     private boolean evaluateMetaAttributeConditionItem(Resource resource, String fieldIdentifier, UUID objectUuid, Object conditionValue, FilterConditionOperator operator) throws RuleException {
         // If the Field Source is Meta Attribute, we expect Field Identifier to be formatted as follows 'name|contentType', since there can be multiple Meta Attributes with the same name, the Content Type must be specified
-        String[] split = fieldIdentifier.split("\\|");
-        if (split.length < 2) throw new RuleException("Field identifier is not in correct format.");
-        AttributeContentType fieldAttributeContentType = AttributeContentType.valueOf(split[1]);
-        String fieldIdentifierName = split[0];
+        String[] parts = parseNameAndContentType(fieldIdentifier);
+        AttributeContentType fieldAttributeContentType = AttributeContentType.valueOf(parts[1]);
+        String fieldIdentifierName = parts[0];
```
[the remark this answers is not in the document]
klaraf755: Added

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
```diff
@@ -615,22 +623,44 @@ private static int compareNumbers(Number objectNumber, Object conditionNumber) {
...
     private boolean evaluateConditionOnAttribute(List<? extends AttributeContent> content, AttributeContentType contentType, Object conditionValue, FilterConditionOperator operator) throws RuleException {
+        if (operator == FilterConditionOperator.EMPTY) {
+            return content.isEmpty();
+        }
+        if (operator == FilterConditionOperator.NOT_EMPTY) {
+            return !content.isEmpty();
+        }
+
+        // For negated operators, evaluate the positive counterpart across all items and negate the result:
+        // "none of the items satisfies EQUALS/CONTAINS/MATCHES" — mirrors FilterPredicatesBuilder's NOT EXISTS semantics.
```
lubomirw, line 632: **! Major** | Correctness

Negated and EMPTY operators evaluate to `false` for objects that don't have the attribute at all, contradicting the NOT EXISTS semantics this comment claims to mirror.

In `FilterPredicatesBuilder` (line 77/107), `NOT_EQUALS`, `NOT_CONTAINS`, `NOT_MATCHES`, and `EMPTY` are NOT-EXISTS conditions — an object **without** the attribute satisfies them. Here, `evaluateCustomAttributeConditionItem` returns `false` when `attributeToCompare == null` (and `evaluateConditionItem` returns `false` for objects without a UUID), so the negation logic is never reached for missing attributes. Result: a list filter for `customAttr NOT_EQUALS x` includes objects lacking the attribute, but the same condition in a trigger rule excludes them. Consider handling the missing-attribute case before the null-check, e.g. returning `true` for `EMPTY`/`NOT_*` operators when the attribute is absent, so trigger semantics match the DB filter semantics.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
```diff
@@ -283,9 +282,11 @@ private boolean evaluateMetaAttributeConditionItem(Resource resource, String fie
     }
 
     private boolean evaluateCustomAttributeConditionItem(Resource resource, UUID objectUuid, String fieldIdentifier, Object conditionValue, FilterConditionOperator operator) throws RuleException {
-        // If source is Custom Attribute, retrieve custom attributes of this object and find the attribute which has Name equal to Field Identifier
+        // If source is Custom Attribute, Field Identifier is formatted as 'name|contentType'; extract name to look up the attribute
+        String[] parts = parseNameAndContentType(fieldIdentifier);
```
lubomirw: **~ Minor** | Correctness

Custom-attribute condition items with a bare attribute name (no `|contentType`) now throw `RuleException` instead of evaluating.

Previously this method matched `fieldIdentifier` directly against the attribute name, so condition items stored with a bare name (e.g. created via the API rather than the UI — the old test itself used `"custom"`) evaluated successfully. After this change they throw, the exception message is recorded in trigger history, and the condition fails. UI-created identifiers use `name|contentType` (`SearchHelper`), so those are fine — but if bare-name condition items exist in any deployment's DB, this is a silent behavior break with no data migration. Consider a fallback (treat a pipe-less identifier as the name) or a migration normalizing stored identifiers.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
lubomirw, line 632: **~ Minor** | Correctness

For meta attributes, the negated-operator semantics introduced here interact with the per-entry OR loop in `evaluateMetaAttributeConditionItem` — worth documenting the intended behavior.

When multiple metadata entries share the same name + contentType (e.g. the same meta attribute from different connectors), `evaluateMetaAttributeConditionItem` returns `true` for the first entry that satisfies the condition. With `NOT_EQUALS`/`NOT_CONTAINS`/`NOT_MATCHES`, an entry that merely *lacks* the value satisfies the condition even when another entry *contains* it, whereas `FilterPredicatesBuilder` evaluates a single NOT EXISTS across all matching entries and would return false. This is an edge case and the per-entry semantics may well be the intended one for triggers — but since the new comment explicitly references `FilterPredicatesBuilder`'s NOT EXISTS semantics, please add a short code comment in `evaluateMetaAttributeConditionItem` stating that negated operators are evaluated per metadata entry, so the divergence is documented rather than discovered. It is also worth opening a follow-up issue to align this behavior with the DB-side filter semantics properly, so the gap is tracked rather than left implicit.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
```diff
@@ -615,22 +623,44 @@ private static int compareNumbers(Number objectNumber, Object conditionNumber) {
...
             Object attributeValue = contentType.isFilterByData() ? attributeContent.getData() : attributeContent.getReference();
             try {
-                if (Boolean.TRUE.equals(fieldTypeToOperatorActionMap.get(contentTypeToFieldType(contentType)).get(operator).apply(attributeValue, conditionValue)))
-                    return true;
+                if (Boolean.TRUE.equals(fieldTypeToOperatorActionMap.get(contentTypeToFieldType(contentType)).get(effectiveOperator).apply(attributeValue, conditionValue)))
+                    // Positive match found: for non-negated ops return true, for negated ops return false (a match disqualifies NOT_EQUALS/NOT_CONTAINS/NOT_MATCHES)
+                    return !isNegated;
             } catch (Exception e) {
-                throw new RuleException("Invalid condition.");
+                throw new RuleException("Cannot evaluate operator %s on attribute value '%s' with condition value '%s' (contentType: %s): %s"
```
lubomirw: **~ Minor** | Security

The improved error message embeds the raw attribute value, which can leak sensitive content into trigger history.

`RuleException` messages from condition evaluation are recorded as trigger history records (`getConditionEvaluationResult`), which are user-visible. Since `contentTypeToFieldType` maps `SECRET` and `CREDENTIAL` to `STRING`, a failed evaluation on such an attribute would write its value into the history record. Please mask `attributeValue` in the message when `contentType` is `SECRET` or `CREDENTIAL` (e.g. replace it with `*****`) — operator, contentType, and the cause message are enough to diagnose the failure for those types.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
```diff
@@ -483,6 +475,22 @@ private boolean getConditionEvaluationResult(ConditionItem conditionItem, T obje
...
+            throw new RuleException("Field identifier is not in correct format 'name|contentType', got: " + fieldIdentifier);
+        }
+        return parts;
+    }
+
+    private AttributeContentType parseAttributeContentType(String contentType) throws RuleException {
+        try {
+            return AttributeContentType.valueOf(contentType);
+        } catch (IllegalArgumentException e) {
+            throw new RuleException("Cannot parse content type %s from execution item: %s".formatted(contentType, e.getMessage()));
```
lubomirw: **. Nitpick** | Maintainability

The error message says "from execution item", but this helper is now also called from condition-evaluation paths.

`parseAttributeContentType` is invoked from `evaluateMetaAttributeConditionItem` and `resolveSourceAttributeContent`, so a malformed content type in a *condition item* would produce a message blaming an execution item — misleading when it lands in trigger history. Suggest a neutral wording, e.g. `"Cannot parse content type %s from field identifier: %s"`. (The method could also be `static`, like its sibling `parseNameAndContentType`.)

lubomirw (changes requested):

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
klaraf755: For secret/credential, attribute value is reference instead of data, and reference does not need to be masked. But actually this problem would occur with encrypted attribute content instead. So I will fix it for encrypted attribute.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
klaraf755: Modified it so it matches `FilterPredicateBuilder` logic.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
klaraf755: Fixed

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
klaraf755: Added check if fieldIdentifier contains `|`.

src/main/java/com/czertainly/core/evaluator/TriggerEvaluator.java
klaraf755: The correct approach in both cases is to also add connector to field identifier, which would guarantee the uniqueness of metadata attribute. I will add the comment for now.

lubomirw (approved):

PR #1547, opened 2026-06-03
Record every signing operation as a signing record

### LINKED ISSUE #143: Signing Record
- create or not create
- what to persist (signature, document...)
- retention policy
- (?) delete after retrieval connected to asynchronous CSC API call

LukasNajman:


src/main/java/com/czertainly/core/dao/repository/signing/SigningRecordRepository.java
```diff
@@ -0,0 +1,44 @@
...
+    int deleteExpiredByRetention(@Param("limit") int limit);
+
+    @Modifying
+    @Query(value = """
+            DELETE FROM {h-schema}signing_record sr
+            USING {h-schema}signing_profile sp
+            WHERE sr.signing_profile_uuid = sp.uuid
+              AND sp.delete_after_retrieval = true
+              AND sr.signed_document_retrieved_at IS NOT NULL
+            """, nativeQuery = true)
```
ivosh, line 51: To discuss: do we want to introduce LIMIT here, like deleteExpiredByRetention does?

src/main/java/com/czertainly/core/dao/repository/signing/SigningRecordRepository.java
```diff
@@ -0,0 +1,44 @@
...
+    int deleteExpiredByRetention(@Param("limit") int limit);
+
+    @Modifying
+    @Query(value = """
+            DELETE FROM {h-schema}signing_record sr
+            USING {h-schema}signing_profile sp
+            WHERE sr.signing_profile_uuid = sp.uuid
+              AND sp.delete_after_retrieval = true
+              AND sr.signed_document_retrieved_at IS NOT NULL
```
ivosh: To discuss: there is no database index on signed_document_retrieved_at.
Introduce a partial index?

```
CREATE INDEX idx_sr_retrieved_at ON signing_record (signed_document_retrieved_at)
      WHERE signed_document_retrieved_at IS NOT NULL;
```

src/main/java/com/czertainly/core/mapper/signing/SigningRecordMapper.java
```diff
@@ -0,0 +1,43 @@
...
+public class SigningRecordMapper {
+
+    private SigningRecordMapper() {
+    }
+
+    public static SigningRecordDto toDto(SigningRecord record) {
+        SigningRecordDto dto = new SigningRecordDto();
+        dto.setUuid(record.getUuid().toString());
+        dto.setName(record.getName());
+        Instant signingTimeZoned = record.getSigningTime() != null ? record.getSigningTime() : null;
```
ivosh: To discuss: What is the intent here? The same values are returned?

src/main/java/com/czertainly/core/mapper/signing/SigningRecordMapper.java
```diff
@@ -0,0 +1,43 @@
...
+            dto.setSignedDocumentRetrievedAt(record.getSignedDocumentRetrievedAt());
+        }
+        return dto;
+    }
+
+    public static SigningRecordListDto toListDto(SigningRecord record) {
+        SigningRecordListDto dto = new SigningRecordListDto();
+        dto.setUuid(record.getUuid().toString());
+        dto.setName(record.getName());
+        Instant signingTime = record.getSigningTime() != null ? record.getSigningTime() : null;
```
ivosh: To discuss: the same as for line 19.

src/main/java/com/czertainly/core/service/impl/SigningRecordServiceImpl.java
```diff
@@ -0,0 +1,151 @@
...
+        SigningRecord record = getSigningRecordEntity(uuid);
+        return SigningRecordMapper.toDto(record);
+    }
+
+    @Override
+    @ExternalAuthorization(resource = Resource.SIGNING_RECORD, action = ResourceAction.DETAIL)
+    @Transactional(readOnly = true)
+    public SigningRecordValidationResultDto validateSigningRecord(SecuredUUID uuid) throws NotFoundException {
+        getSigningRecord(uuid);
+        throw new UnsupportedOperationException("Signing record validation not yet implemented");
```
ivosh: To discuss: should not be a part of this PR (likewise controller and interfaces).

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -686,6 +713,42 @@ private void applyWorkflow(SigningProfile p, SigningProfileVersion version, Work
...
 
+    private boolean contentPolicydDifferFromVersion(SigningProfileVersion v, SigningRecordPolicyRequestDto p) {
+        if (p == null)
+            return false;
+        return v.isRecordMetadata() != p.isRecordMetadata()
+                || v.isRecordRequestMetadata() != p.isRecordRequestMetadata()
+                || v.isRecordSignature() != p.isRecordSignature()
+                || v.isRecordSignedDocument() != p.isRecordSignedDocument()
+                || v.isRecordDtbs() != p.isRecordDtbs();
+    }
```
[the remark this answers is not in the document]
ivosh: Fixed.

src/main/java/com/czertainly/core/service/writer/signingrecord/ImmediateSigningRecordWriter.java
```diff
@@ -0,0 +1,41 @@
...
+    public void record(SigningRecordInput input) {
+        if (!SigningRecordPolicy.hasAnyRecordableContent(input.getSigningProfile().recordPolicy())) {
+            metrics.skippedNoContentPolicy().increment();
+            return;
+        }
+        metrics.timed("IMMEDIATE", () -> {
+            try {
+                repository.save(mapper.toRecord(input));
+                metrics.created("IMMEDIATE").increment();
+            } catch (RuntimeException e) {
```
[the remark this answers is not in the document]
ivosh: Fixed in 85a8cdfb.

src/main/java/com/czertainly/core/service/writer/signingrecord/OutboxSigningRecordWriter.java
```diff
@@ -0,0 +1,119 @@
...
+        if (!SigningRecordPolicy.hasAnyRecordableContent(input.getSigningProfile().recordPolicy())) {
+            metrics.skippedNoContentPolicy().increment();
+            return;
+        }
+        metrics.timed("DEFERRED_DURABLE", () -> {
+            try {
+                outboxRepository.save(mapper.toOutbox(input));
+                metrics.outboxEnqueued().increment();
+                metrics.created("DEFERRED_DURABLE").increment();
+            } catch (RuntimeException e) {
```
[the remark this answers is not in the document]
ivosh: Fixed in 85a8cdfb.

src/main/java/com/czertainly/core/dao/entity/signing/SigningRecord.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.UUID;
+
+@Getter
+@Setter
+@Entity
+@Table(name = "signing_record")
+public class SigningRecord extends UniquelyIdentifiedAndAudited implements Securable {
+
+    @Column(name = "name")
+    private String name;
```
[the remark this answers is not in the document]
ivosh: Fixed in 85a8cdfb.

src/main/java/com/czertainly/core/service/impl/SigningRecordServiceImpl.java
LukasNajman: Removed

src/test/java/com/czertainly/core/signing/record/SigningRecordPolicyTest.java
```diff
@@ -0,0 +1,61 @@
...
+                Arguments.of("recordRequestMetadata", aSigningRecordPolicy().recordRequestMetadata(true).build()),
+                Arguments.of("recordSignature", aSigningRecordPolicy().recordSignature(true).build()),
+                Arguments.of("recordSignedDocument", aSigningRecordPolicy().recordSignedDocument(true).build()),
+                Arguments.of("recordDtbs", aSigningRecordPolicy().recordDtbs(true).build()));
+    }
+
+    @ParameterizedTest(name = "{0}")
+    @MethodSource("singleRecordFlagPolicies")
+    void hasAnyRecordableContent_returnsTrue_whenAnySingleRecordFlagIsSet(
+            String flagName, SigningRecordPolicyModel policyWithSingleFlag) {
```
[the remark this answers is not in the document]
LukasNajman: Named used

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -373,14 +378,36 @@ SigningProfileDto persistUpdate(SecuredUUID uuid, SigningProfileRequestDto reque
...
+                        v.setVersion(profileRef.getLatestVersion());
+                        return v;
+                    });
+        }
 
         applyWorkflow(profile, version, request.getWorkflow());
         applyScheme(profile, version, request.getSigningScheme());
+        applyContentPolicyRecordToVersion(version, request.getRecordPolicy());
+        applyOperationalPolicyToProfile(profile, request.getRecordPolicy());
+        mirrorContentPolicyRecordToProfileCache(profile, version);
```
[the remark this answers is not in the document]
LukasNajman: The DTO is always fully populated

src/main/java/com/czertainly/core/dao/repository/signing/SigningRecordRepository.java
LukasNajman: Limited

src/main/java/com/czertainly/core/mapper/signing/SigningRecordMapper.java
LukasNajman: Ternary removed

src/main/java/com/czertainly/core/mapper/signing/SigningRecordMapper.java
LukasNajman: Ternary removed

src/main/java/com/czertainly/core/dao/repository/signing/SigningRecordRepository.java
LukasNajman: Index added

CLAUDE.md
```diff
@@ -100,6 +100,22 @@ Use pessimistic locking on the read-and-update path: a custom repository finder
 
 Don't span the lock across slow external calls (see "Transactions and external calls" above) — split the transaction so the lock covers only the local writes that need atomicity.
 
+## New `@Value` parameters belong in application.yml
```
ivosh: I suggest to remove this section because it is not aligned with the direction from the CEO - settings should be configurable via Settings UI, not through env. variables, unless there is a specific use case fit.

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -440,6 +458,15 @@ private void deleteSigningProfile(SigningProfile signingProfile) throws Validati
             );
         }
 
+        if (signingRecordService.doesSigningRecordExistForProfileInternal(signingProfile.getUuid())) {
```
ivosh, line 461: **MINOR** · _Correctness_

**Profile-delete guard ignores pending DEFERRED_DURABLE outbox rows, orphaning them into a poison loop**

SigningProfileServiceImpl.java:461 blocks deletion only when signing_record rows exist (signingRecordService.doesSigningRecordExistForProfileInternal -> SigningRecordRepository.existsBySigningProfileUuid, which queries only the signing_record table). It does NOT check signing_record_outbox. For DEFERRED_DURABLE mode, a freshly-signed record is durably staged in signing_record_outbox but copied into signing_record only later by SigningRecordOutboxDrainer. If a profile (and its versions) is deleted while outbox rows for it are still undrained, those rows survive (no FK from signing_record_outbox to signing_profile_version).

On the next drain, SigningRecordWriter.saveRecordAndDeleteOutbox inserts into signing_record, which violates the (signing_profile_uuid, signing_profile_version) FK declared ON DELETE RESTRICT (migration lines 36-37). Every drain attempt then throws, attemptDrain (SigningRecordOutboxDrainer.java:127) records the failure and increments attempts until the row crosses poison-threshold and is abandoned forever — the staged signing record is silently lost, and the operator who set DEFERRED_DURABLE for durability gets neither the record nor an error.

There is also in-memory queue (BEST_EFFORT) to consider - there is no good way how to check entries there.

This is not a blocking finding: we can discuss what should happen in this edge case -  If this is intended (best-effort once the profile is gone), then document it; otherwise the delete guard should also reject when undrained outbox rows exist for the profile (an existsBySigningProfileUuid on SigningRecordOutboxRepository), mirroring the signing_record check.

src/main/java/com/czertainly/core/service/impl/SigningRecordServiceImpl.java
```diff
@@ -0,0 +1,185 @@
...
+                        (root, cb) -> cb.desc(root.get(Audited_.CREATED)))
+                .stream().map(SigningRecordMapper::toListDto).toList();
+        long totalItems = signingRecordRepository.countUsingSecurityFilter(filter, predicate);
+        return PaginationResponseMapper.toDto(dtos, request.getPageNumber(), request.getItemsPerPage(), totalItems);
+    }
+
+    @Override
+    @ExternalAuthorization(resource = Resource.SIGNING_RECORD, action = ResourceAction.DETAIL)
+    @Transactional(readOnly = true)
+    public SigningRecordDto getSigningRecord(SecuredUUID uuid) throws NotFoundException {
```
ivosh: 🔴 **MAJOR** · _Correctness_

**Serving the signed document never marks it retrieved, so delete-after-retrieval never fires**

SigningRecordServiceImpl.getSigningRecord (src/main/java/com/czertainly/core/service/impl/SigningRecordServiceImpl.java:104-107) is the only production path that returns the signed document to a client — SigningRecordMapper.toDto sets dto.signedDocument (SigningRecordMapper.java:54). But it never calls SigningRecordRetrievalHook.onSignedDocumentServed, which is the only code that sets signedDocumentRetrievedAt (SigningRecordRetrievalHook.java:50-52). I grep-confirmed onSignedDocumentServed has no production caller (only the unit test references it). Consequence: when an operator enables the opt-in delete_after_retrieval flag on a SigningProfileVersion (migration V202606021000__signing_record.sql:13, default FALSE), the document IS served but signedDocumentRetrievedAt stays NULL forever. Both deletion paths gate on `signed_document_retrieved_at IS NOT NULL` — the per-retrieval afterCommit delete in planSigningRecordDeletion and the fallback sweep query SigningRecordRepository.java:48 (`AND sr.signed_document_retrieved_at IS NOT NULL`). So a record's signed document is disclosed via the API yet the record is never deleted: the delete-after-retrieval feature is effectively inert, and the sensitive blob lives on indefinitely against operator intent. The scheduler/Javadoc explicitly calls runFallbackSweep the 'fallback' to a 'per-retrieval' delete (SigningRecordRetrievalDeletionScheduler.java:9), confirming the per-retrieval marking was meant to happen on serve. If a separate not-yet-merged document-download endpoint is intended to call onSignedDocumentServed, say so in a comment / plan note and keep this PR's getSigningRecord from exposing signedDocument; otherwise wire getSigningRecord (or the dedicated retrieval endpoint) to call onSignedDocumentServed so retrieval is recorded and deletion can fire.

src/main/java/com/czertainly/core/signing/record/SigningRecordOutboxDrainer.java
```diff
@@ -0,0 +1,149 @@
...
+    private boolean attemptDrain(UUID uuid) {
+        try {
+            SigningRecordOutbox row = outboxRepo.findById(uuid).orElse(null);
+            if (row == null) {
+                return false;
+            }
+            metrics.persist(SigningRecordPersistenceMode.DEFERRED_DURABLE.name()).increment();
+            writer.saveRecordAndDeleteOutbox(SigningRecordMapper.toRecord(row));
+            return true;
+        } catch (RuntimeException drainError) {
```
ivosh, line 136: 🟡 **MINOR** · _Performance_

**Outbox blob entities accumulate in the long-lived outer drain transaction's persistence context**

`drainOnce()` is `@Transactional(REQUIRES_NEW)` (line 71), so one persistence context (PC) stays open for the whole `drainBatches()` loop. `attemptDrain()` (line 127) is a **private, self-invoked** method, so `outboxRepo.findById(uuid)` (line 129) runs in that **outer** PC. `SigningRecordOutbox` carries three `byte[]` blobs (`signatureValue`, `signedDocument`, `dtbs`) that are plain `@Basic` (no `@Lob`, no `fetch = LAZY`), so they load **eagerly** and the blob-bearing entity becomes managed in the outer PC. The per-row copy `saveRecordAndDeleteOutbox` runs in its own `REQUIRES_NEW` tx (separate PC), so committing/deleting a row does **not** evict the outbox entity from the outer PC, and `SigningRecordMapper.toRecord` only copies the array *references*. Nothing calls `entityManager.clear()`/`detach()` in the loop.

**Working-set bound:** `maxBatchesPerRun × maxBatchSize` = 10 × 200 = **2000** blob entities pinned in heap until `drainOnce()` commits (`application.yml`: `max-batch-size: 200`, `max-batches-per-run: 10`).

**Likelihood — a conditional scaling cliff, not a steady-state problem:**
- **Steady state: negligible.** The scheduler fires `fixedDelay = 500ms`; each run drains only what accrued since the last and the PC is discarded at commit (per-run, not cumulative). Hitting the cap needs ~4000 DEFERRED_DURABLE signings/second sustained; normal runs hold a handful of entities.
- **Backlog recovery: moderate and potentially serious.** The cap is reached only when a backlog built up (drain disabled / node down / cluster-lock contention / DB blip) and then catches up at 2000 rows/run. At ~50 KB avg payload that is ~300 MB transient heap/run; with multi-MB signed documents it is multiple GB → real OOM risk, precisely when the system is already stressed recovering.

Note `maxBatchesPerRun` already bounds the lock/transaction *hold duration* (Javadoc lines 84–92), but it bounds time, not the per-run memory working set — that is the gap here.

**Fix (cheapest first):**
1. Call `entityManager.clear()` after each batch in `drainBatches()` (or `detach(row)` in `attemptDrain`'s `finally`) — caps the outer PC to one batch.
2. Move `findById` + `toRecord` into the writer's `REQUIRES_NEW` tx (`saveRecordAndDeleteOutbox(uuid)`) so the outbox entity lives only in the short-lived per-row PC — outer PC growth drops to zero. Retry/poison semantics are unchanged (a read failure still surfaces to `attemptDrain` and is recorded).

Option 1 is sufficient and low-risk. Not blocking.

src/main/java/com/czertainly/core/signing/record/AbstractSigningRecordStrategy.java
```diff
@@ -0,0 +1,42 @@
...
+ * implement {@link #doRecord(SigningRecordInput)} (the mode-specific persistence dispatch) and {@link #mode()}
+ * (the metric tag / persistence-mode identity).
+ *
+ * <p>{@code recordingEnabled} is the sole gate on whether a record is written: when it is on, every signing
+ * operation produces a record carrying at least its intrinsic metadata (who, when, which profile/version),
+ * with the per-field {@code record*} content toggles deciding only which optional payload columns are filled.</p>
+ */
+public abstract class AbstractSigningRecordStrategy implements SigningRecordStrategy {
+
+    private static final Logger logger = LoggerFactory.getLogger(AbstractSigningRecordStrategy.class);
```
ivosh: ⚪ **NITPICK** · _Maintainability_

**Logger declared via LoggerFactory while every sibling strategy uses @Slf4j**

AbstractSigningRecordStrategy.java:19 declares `private static final Logger logger = LoggerFactory.getLogger(...)` by hand, whereas every concrete neighbour in the same package obtains its logger from Lombok's `@Slf4j` (BestEffortSigningRecordStrategy.java:21, SigningRecordOutboxDrainer.java:38, SigningRecordRetentionSweeper.java:11, SigningRecordRetrievalHook.java:21, SigningRecordBestEffortFlusher.java:14). Both work; this is purely a convention drift (also note the field is named `logger` here vs Lombok's `log`). For consistency with the established neighbour pattern, annotate the class `@Slf4j` and drop the manual field, or leave as-is if there is a reason. Effective Java Item 67 is unrelated; this is a Google Java Style consistency point (uniform logging idiom across a package). Non-blocking.

ivosh (approved):
The PR is very well organized.
I posted few minor findings but they can be deferred to a follow-up PR.
Couple of tests in SigningProfileServiceImplTest were disabled - they will be fixed in the follow-up PR introducing refactored tests.

ivosh (changes requested):
The PR is very well organized.
I posted few minor findings but they can be deferred to a follow-up PR.
Couple of tests in SigningProfileServiceImplTest were disabled - they will be fixed in the follow-up PR introducing refactored tests.

Address the the one finding related to CLAUDE.md. Everything else could be deferred to a follow-up.

src/main/java/com/czertainly/core/signing/record/AbstractSigningRecordStrategy.java
LukasNajman: fixed

CLAUDE.md
LukasNajman: Edited so that the description includes the direction proposed by the CEO

src/main/java/com/czertainly/core/signing/record/SigningRecordOutboxDrainer.java
LukasNajman: `entityManager.clear()` is now called after each `drainBatches()`

src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java
LukasNajman: agree, lets discuss this and decide later.

LukasNajman: Documented in https://github.com/OmniTrustILM/ilm/issues/243

ivosh (dismissed):

ivosh (approved):

PR #1549, opened 2026-06-03
Fix missing actor in audit logs on auth cache hits

### LINKED ISSUE #1548: ACME: All public ACME endpoints fail with NullPointerException in audit logging (LoggingHelper expects non-null AuthMethod for unauthenticated requests)
### Description

The ACME protocol is currently unusable on this environment. Any request to a public ACME endpoint (`/directory`, `/newNonce`, `/newAccount`) fails with HTTP 400 because the request crashes inside the audit logging filter **before it ever reaches the ACME controller**.

Root cause: `LoggingHelper.getActorInfo` calls `AuthMethod.valueOf(null)` on requests that have no authenticated actor. ACME endpoints are public per RFC 8555 (the protocol uses JWS-signed payloads, not server-side authentication), so the actor's AuthMethod is null by design. The audit code does not handle this case, throws `NullPointerException`, and the AOP wrapper turns the failure into a generic 400 response.

This blocks all ACME usage — no client can fetch the directory, request a nonce, or register an account. As a result, the retest of #1351 (Default RA Profile change propagation) is blocked, and any other ACME flow on this build is broken.

### Steps to Reproduce

1. Configure any ACME Profile in ILM with a Default RA Profile
2. Send a request to the ACME directory URL via any ACME client (e.g. `acme.sh --register-account --server <ACME-directory-URL>`)
3. Or hit directly with curl: `curl -v <ACME-directory-URL>`

### Expected Behavior

- Directory request returns 200 with the JSON list of ACME endpoints
- Subsequent registration succeeds and returns the account location URL
- Audit log records the request as an anonymous/public action without crashing

### Actual Behavior

- All ACME endpoints return HTTP 400
- Core log shows:

```
ERROR [tomcat-handler-XXXX] [com.czertainly.core.api.AcmeExceptionHandlingAdvice]: Name is null
java.lang.NullPointerException: Name is null
    at java.base/java.lang.Enum.valueOf(Unknown Source)
    at com.czertainly.api.model.core.logging.enums.AuthMethod.valueOf(AuthMethod.java:12)
    at com.czertainly.core.logging.LoggingHelper.getActorInfo(LoggingHelper.java:60)
    at com.czertainly.core.aop.AuditLogAspect.log(AuditLogAspect.java:79)
    ...
    at com.czertainly.core.api.acme.AcmeControllerImpl$$SpringCGLIB$$0.getDirectory(<generated>)
```

- ACME flow cannot start

### Severity

Critical

### Module

Core

### Screenshot

_No response_

### Logs

[core_3.6.26_logs.txt](https://github.com/user-attachments/files/28550212/core_3.6.26_logs.txt)

### Environment

ILM Core: 2.17.1-SNAPSHOT (2 Jun 2026, 16:19)
Commit: 29ba651
Branch: main
Build time: 2026-06-02T14:19:37Z
ILM Frontend: 2.17.0 (3 Jun 2026, 09:51)
Client:
- macOS Tahoe 26.3.1
- Chrome 148.0.7778.179 (arm64)

ivosh:


src/main/java/com/czertainly/core/security/authn/client/CzertainlyAuthenticationClient.java
```diff
@@ -60,24 +60,39 @@ public CzertainlyAuthenticationClient(
...
+     * {@code AuditLogAspect} reads actor identity exclusively from the MDC.
+     * Restores the actor type, auth method, user UUID and username; on a cache miss the loader has already set
+     * identical values, so reapplying them is a no-op.
+     * Anonymous results are never cached, so the loader's anonymous MDC path is unaffected.
+     */
+    private static AuthenticationInfo restoreActorMdc(AuthenticationInfo authInfo) {
+        if (!authInfo.isAnonymous()) {
+            LoggingHelper.putActorInfoWhenNull(ActorType.USER, authInfo.getAuthMethod());
+            LoggingHelper.putActorInfoWhenNull(ActorType.USER, authInfo.getUserUuid(), authInfo.getUsername());
+        }
```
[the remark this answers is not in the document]
ivosh: Fixed in 85857f97.

lubomirw (commented):

src/main/java/com/czertainly/core/security/authn/client/CzertainlyAuthenticationClient.java
```diff
@@ -60,24 +60,40 @@ public CzertainlyAuthenticationClient(
...
+                jti, () -> authenticate(AuthMethod.TOKEN, claims, false)));
+    }
+
+    /**
+     * Replays the actor MDC side effects of {@link #authenticate} so that cache hits leave the MDC in the same state as cache misses.
+     * {@code AuditLogAspect} reads actor identity exclusively from the MDC.
+     * Restores the actor type, user UUID and username only when the MDC carries no actor yet.
+     * Anonymous results are never cached, so the loader's anonymous MDC path is unaffected.
+     */
+    private static AuthenticationInfo restoreActorMdc(AuthenticationInfo authInfo) {
```
lubomirw, line 88: **❗ Major** | Correctness

`restoreActorMdc` does not reach parity with the cache-miss path for the methods' real callers.

This fixes the cache-hit case where MDC had *no* actor, but the two real callers of these methods — `AuthHelper.authenticateAsSystemUser` and `authenticateAsUser` — always set the actor in MDC just before calling. So at the actual call sites `hasActorInfo()` is already true, and the hit and miss paths still disagree:

- **System / protocol user:** on a cache miss the loader's `createAuthenticationInfo` calls the 3-arg `putActorInfoWhenNull(USER, uuid, name)`, which writes unconditionally and overwrites the actor to `USER` + the real uuid/username. On a cache hit `restoreActorMdc` keeps the `CORE`/`PROTOCOL` actor that AuthHelper set, with the system username and no uuid. Same user, different audit actor depending on whether the entry was cached.
- **authenticateAsUser:** miss logs the real username, hit logs `null` (AuthHelper passes a uuid but no name).

So the audit actor flips the first time an entry gets cached — which is the inconsistency this PR set out to remove. Worth deciding which actor is canonical and making both paths land there. If the preserved (`CORE`/`PROTOCOL`) actor is the intended one, then the loader's unconditional overwrite on the miss path is the real bug to fix; if `USER` + real identity is intended, the hit path should overwrite too. Either way the Javadoc's "same state as cache misses" is not true for these callers today.

src/test/java/com/czertainly/core/security/authn/client/CzertainlyAuthenticationClientTest.java
```diff
@@ -1,6 +1,8 @@
 package com.czertainly.core.security.authn.client;
 
 import com.czertainly.api.model.core.logging.enums.AuthMethod;
+import com.czertainly.api.model.core.logging.records.ActorRecord;
+import com.czertainly.core.logging.LoggingHelper;
```
lubomirw, line 6: **✴ Nitpick** | Maintainability

`ActorType` is referenced by its fully-qualified name inline three times in the two new tests (`...doesNotClobberExistingActorMdc` and `...anonymousResult...`). `AuthMethod` and `ActorRecord` are already imported at the top — add `import com.czertainly.api.model.core.logging.enums.ActorType;` and use the short name for consistency and readability.

src/main/java/com/czertainly/core/security/authn/client/CzertainlyAuthenticationClient.java
ivosh: Investigation: before introducing the authentication cache, every successful proxy auth ran the loader and always ended at USER + real identity; CORE/PROTOCOL never survived a successful auth. So the canonical
  actor is USER + identity. I strayed from the right path by Copilot's "don't clobber" concern which is moot (clobbering is historical behaviour).

src/test/java/com/czertainly/core/security/authn/client/CzertainlyAuthenticationClientTest.java
ivosh: Fixed.

lubomirw (approved):

PR #1550, opened 2026-06-03
Persist event history for failed certificate operations that roll back

### LINKED ISSUE #1545: Failed bulk "Override RA Profile" on certificates writes no History log record
### Description

In the certificate listing, selecting one or more certificates and applying the bulk "Override RA Profile" action with an RA Profile that cannot identify the selected certificates causes the operation to fail. The failed bulk operation is not recorded in the History log.

### Steps to Reproduce

1. Open the certificate listing
2. Select one or more certificates
3. Choose the bulk "Override RA Profile" action
4. Select an RA Profile that will not identify the certificates
5. Run the action (it fails)
6. Open the History log

### Expected Behavior

The failed bulk "Override RA Profile" attempt is recorded in the History log (operation, affected certificates, outcome).

### Actual Behavior

No record is created in the History log when the bulk operation fails.

### Severity

Major

### Module

Certificates

ivosh:


src/main/java/com/czertainly/core/events/transaction/UpdateCertificateHistoryEvent.java
```diff
@@ -3,9 +3,18 @@
...
 import java.util.UUID;
 
+/**
+ * Records a certificate history entry. Publish this only for records describing committed work — a rollback of the publishing
+ * transaction discards the event, so the history record dies together with the work it describes.
+ *
+ * <p>Do NOT publish this for FAILED records on a code path whose transaction is about to roll back: the event would be
+ * silently dropped and the failure would leave no trace. Use {@link CertificateEventHistoryInternalService#addEventHistorySurvivingRollback}
+ * instead, which commits independently of the caller's transaction.</p>
+ */
```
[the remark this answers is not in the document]
ivosh: JavaDoc with full import is not readable by humans.

src/main/java/com/czertainly/core/service/impl/CertificateEventHistoryServiceImpl.java
```diff
@@ -64,6 +64,12 @@ public void addEventHistory(UUID certificateUuid, CertificateEvent event, Certif
         certificateEventHistoryRepository.save(history);
     }
 
+    @Override
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public void addEventHistorySurvivingRollback(UUID certificateUuid, CertificateEvent event, CertificateEventStatus status, String message, String additionalInformation) {
```
lubomirw, line 69: **~ Minor** | Correctness

Question on the `REQUIRES_NEW` + FK interaction: this method suspends the caller's transaction and inserts a `certificate_event_history` row whose FK references the certificate. If a caller ever invokes it *after* the outer transaction has already UPDATEd/locked that certificate row, the new transaction's FK check would wait on a lock the suspended outer transaction holds — a self-deadlock on a single thread.

Both current call sites look safe (delete is refused before any mutation; the RA-profile failure is caught before the row is updated), so this likely isn't a bug today. Did you confirm that no current caller holds a write lock on the certificate row before calling? If so, worth a one-line Javadoc note — e.g. "call before the outer transaction mutates/locks the certificate row" — so a future post-mutation caller doesn't reintroduce the hazard.

lubomirw (commented):
Code review: 1 issue found (advisory).

src/main/java/com/czertainly/core/service/impl/CertificateEventHistoryServiceImpl.java
ivosh: Fixed in 80d006e2. Confirmed both current callers invoke this before any mutation: the refused-delete path calls it immediately (before any certificate row write), and the failed RA-profile switch calls it inside the catch blocks, before `certificate.setRaProfile(...)` / `certificateRepository.save(...)`. So no self-deadlock today. Added the one-line Javadoc note to guard against a future post-mutation caller reintroducing the hazard.

lubomirw (approved):

PR #1554, opened 2026-06-04
Restore certificate state to Issued when revocation approval rejected

### LINKED ISSUE #1553: Rejected revocation approval leaves certificate stuck in Pending approval state
### Description

Certificate belongs to an RA profile that has an approval profile assigned. Revoking such a certificate opens an approval request. When the approver rejects the approval, the certificate remains in "Pending approval" state even though the certificate history records the approval as rejected. Certificate state and history are inconsistent.

### Steps to Reproduce

1. Configure an RA profile with an approval profile assigned.
2. Revoke a certificate issued under that RA profile.
3. Revocation enters the approval workflow (certificate goes to "Pending approval").
4. Reject the approval.
5. Inspect the certificate state and its history.

### Expected Behavior

After rejection, revocation is not performed: the certificate returns to its prior state and is no longer in "Pending approval". State matches history.

### Actual Behavior

Certificate stays in "Pending approval" state while history says approval was rejected. State and history disagree.

### Severity

Major

### Module

Workflows

ivosh:


src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1010,6 +1010,18 @@ public void revokeCertificateAction(final UUID certificateUuid, ClientCertificat
...
+    @Override
+    public void revokeCertificateRejectedAction(final UUID certificateUuid) throws NotFoundException {
+        final Certificate certificate = certificateRepository.findByUuid(certificateUuid).orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
+        if (certificate.getState() != CertificateState.PENDING_APPROVAL) {
+            logger.debug("Certificate {} is in state {}, not PENDING_APPROVAL; skipping revoke-rejection state restore", certificateUuid, certificate.getState().getLabel());
+            return;
+        }
+        certificate.setState(CertificateState.ISSUED);
+        certificateRepository.save(certificate);
+        certificateEventHistoryService.addEventHistory(certificate.getUuid(), CertificateEvent.REVOKE, CertificateEventStatus.FAILED, "Revocation approval was rejected; certificate restored to Issued.", "");
```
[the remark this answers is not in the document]
ivosh: Fixed in c4aad2ad.

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1010,6 +1010,18 @@ public void revokeCertificateAction(final UUID certificateUuid, ClientCertificat
...
     }
 
+    @Override
+    public void revokeCertificateRejectedAction(final UUID certificateUuid) throws NotFoundException {
+        final Certificate certificate = certificateRepository.findByUuid(certificateUuid).orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
+        if (certificate.getState() != CertificateState.PENDING_APPROVAL) {
+            logger.debug("Certificate {} is in state {}, not PENDING_APPROVAL; skipping revoke-rejection state restore", certificateUuid, certificate.getState().getLabel());
+            return;
+        }
+        certificate.setState(CertificateState.ISSUED);
```
3keyroman, line 1020: What if it was in other state than `ISSUED` before that? Is this valid assumption?

lubomirw (approved):

src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: That should not be case. Revoke should start in `Issued` state.

 Evidence in ClientOperationServiceImpl.java:
  - Line 1054 — [`validateOldCertificateForOperation`](https://github.com/OmniTrustILM/core/blob/c66ae20ae573ea8d187ea71606176ccf3663cd20/src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java#L1054) (gate for REVOKE) throws unless state == ISSUED. PENDING/archived/any-other rejected. So revoke can only start from ISSUED.
  - Line 1651 — existing [cancel-revoke](https://github.com/OmniTrustILM/core/blob/c66ae20ae573ea8d187ea71606176ccf3663cd20/src/main/java/com/czertainly/core/service/v2/impl/ClientOperationServiceImpl.java#L1651) path also restores to ISSUED. Same precedent.

lubomirw: When we will introduce state machine, this questions will be much easier to answer 😃

PR #1575, opened 2026-06-10
Rename packages to com.otilm and Czertainly classes to Platform prefix

### LINKED ISSUE #1566: Rebrand CZERTAINLY to ILM
### Description

We need to drop the CZERTAINLY name from Core. Three things to do, plus some infra strings that need care.

**Move the packages and groupId.**
Everything under `com.czertainly.core.*` moves to `com.otilm.core.*`, and we bump Core's Maven `groupId` to `com.otilm` (`pom.xml`). That's ~1200 files across 233 package dirs. Note we still depend on `com.czertainly.api` and the `com.czertainly:dependencies` parent — those live in the `interfaces`/`dependencies` repos and stay as-is until they're rebranded, so don't touch those imports.

**Rename the `Czertainly*` classes.**
33 of them (plus tests) — the JCA provider (`CzertainlyProvider`, cipher/signature services + SPIs, `CzertainlyPrivateKey`/`PublicKey`), the auth layer (`CzertainlyAuthenticationFilter`, `...Token`, `...Client`, `CzertainlyUserDetails`, the OAuth2 handlers), `CzertainlyJmsPoolConnectionFactory`, and `CzertainlyX500NameStyle`. Use a generic prefix like `Platform`, not `ILM`. The new name should reflect the role rather than the brand and survives any future rename.

**Migrate the scheduled-job class names.**
This is the one that bites. `scheduled_job.job_class_name` stores the full class name and we load it with `Class.forName(...)` (`SchedulerServiceImpl.java:243`). The task classes (`UpdateCertificateStatusTask`, `UpdateIntuneRevocationRequestsTask`, `CbomSyncTask`, `DiscoveryCertificateTask`) live in `com.czertainly.core.tasks`, so moving the package breaks every existing job row unless we ship a Flyway migration that rewrites the stored names (`REPLACE(job_class_name, 'com.czertainly.core.tasks', 'com.otilm.core.tasks')`). Check the `scheduler` service too, and grep for any other class names persisted in the DB.

**Infra/wire strings.**
> [!IMPORTANT]
> Things here needs to be analyzed and considered if they will not break some functionality and if they do not need to go hand > in hand with changes in othere repos like `scheduler`, `helm-charts`, `development-environment`, and FE.

We're rebranding these too, but several are shared contracts — rename them together with their consumers, don't do it piecemeal: the RabbitMQ exchange/vhost (`czertainly`, `czertainly-proxy`), the `czertainly-session` cookie (changing it logs everyone out), the JCA provider name string and the SCEP/Intune `"CZERTAINLY-V…"` string (sent to external clients — confirm interop first), the `/etc/czertainly-backend` config path, the `CZERTAINLY_LOG_LEVEL` env var, logback's `com.czertainly` logger, the Hibernate `FunctionContributor` SPI file, and the Dockerfile user/paths/labels. This part touches `scheduler`, `helm-charts`, `development-environment`, and FE.

Best done as a few stacked PRs rather than one giant one.
- Update interfaces to 2.18.1-SNAPSHOT and do the updates
- move classes to `com.otilm.*` package and rename classes
- DB migrations

### Definition of Done

- Packages moved to `com.otilm.core.*`, groupId bumped; builds and all tests pass.
- All `Czertainly*` classes renamed to a generic prefix; no behavior change.
- Flyway migration rewrites `scheduled_job.job_class_name`; existing jobs still resolve and run after upgrade (tested against pre-rename rows). System jobs re-register cleanly.
- Any other persisted class names found and migrated.
- Logback config, the Hibernate SPI file, and all package-path references updated.
- Infra/wire strings rebranded with their cross-repo consumers cut over together; anything deliberately left as `com.czertainly` (external artifacts) noted with a reason.
- No `com.czertainly.core` left in the repo except the documented external dependencies.

klaraf755:


src/test/resources/archunit_store/73ede010-447b-4223-95de-15077e7d1715
```diff
@@ -1,33 +1,33 @@
-com.czertainly.core.attribute.engine.AttributeEngine calls @Modifying AttributeContent2ObjectRepository.deleteAllOperationAttributesByVersion() from outside ..service.writer..
```
3keyroman, line 1: Why is this file in the repo?

klaraf755: This was here also before. Javadoc for Rule C in `TransactionalBoundaryArchTest` mentions that pre-existing violations are frozen so they can be resolved later. This is that list of frozen violations.

pom.xml
```diff
@@ -26,16 +26,16 @@
...
         <tomcat.version>10.1.55</tomcat.version>
 
         <sonar.host.url>https://sonarcloud.io</sonar.host.url>
         <sonar.organization>ilm</sonar.organization>
         <sonar.projectKey>ilm_core</sonar.projectKey>
         <sonar.sources>src/main/java</sonar.sources>
-        <sonar.cpd.exclusions>src/main/java/com/czertainly/core/dao/entity/**</sonar.cpd.exclusions>
-        <sonar.coverage.exclusions>src/main/java/com/czertainly/core/dao/entity/**,src/main/java/com/czertainly/core/api/**,src/main/java/com/czertainly/core/util/converter/**</sonar.coverage.exclusions>
+        <sonar.cpd.exclusions>src/main/java/com/otilm/core/dao/entity/**</sonar.cpd.exclusions>
+        <sonar.coverage.exclusions>src/main/java/com/otilm/core/dao/entity/**,src/main/java/com/ilm/core/api/**,src/main/java/com/ilm/core/util/converter/**</sonar.coverage.exclusions>
```
lubomirw: **! Major** | Correctness

`sonar.coverage.exclusions` uses the wrong package path: `com/ilm/core/...` instead of `com/otilm/core/...`.

The second and third patterns (`src/main/java/com/ilm/core/api/**` and `src/main/java/com/ilm/core/util/converter/**`) won't match any files after the rename, so the `api` and `util/converter` packages silently start counting against the coverage metric and may fail the Sonar quality gate. Also worth noting: `ilm` is exactly the brand-derived prefix the originating issue recommended avoiding. Please change both to `com/otilm/core/...`.

src/main/java/com/otilm/core/provider/PlatformProvider.java
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * JCA provider for cryptographic operations using Cryptographic Provider interface.
+ */
+public class PlatformProvider extends Provider {
+
+    public static final String PROVIDER_NAME = "PlatformProvider";
+    private static final Logger logger = LoggerFactory.getLogger(PlatformProvider.class);
+
+    private PlatformProvider(String name, CryptographicOperationsSyncApiClient apiClient) {
+        super(name, "1.0", "Platform Provider");
```
lubomirw, line 19: **~ Minor** | Maintainability

A few brand strings survived inside the renamed classes — worth sweeping while the rename is fresh.

Specifically: the JCA provider info string here is still `"CZERTAINLY Provider"` (the class itself became `PlatformProvider`); `PlatformPrivateKey.getFormat()` still returns `"CZERTAINLY"`; and `AuthHelper` still uses the variable names `czertainlyAuthenticationClient` (line 83) and `czertainlyAuthenticationToken` (line 162) for what are now `Platform*` types. None of these are persisted or contractual, so renaming them is safe — e.g. `"Platform Provider"` / `"PLATFORM"` and matching variable names.

lubomirw (changes requested):
Code review: 5 issue(s) found. Three are posted as inline/file comments; two file-level questions follow as separate comments (their lines are unchanged context). One more note that cannot be attached inline because the Dockerfile is not part of this diff:

**~ Minor** | Maintainability

Question: is the deployment surface intentionally left czertainly-branded for a follow-up PR?

The container user/paths (`/opt/czertainly`), env var names (`CZERTAINLY_LOG_LEVEL`), broker exchange/vhost defaults (`czertainly`), and the `org.opencontainers.image.authors` label (`support@czertainly.com`) are all still on the old brand. Keeping the runtime contracts is the safe choice since they're shared with helm charts and other components — if a follow-up PR is planned to cover these, could you link it here and mention the scope boundary in the PR description?

src/main/java/com/otilm/core/service/impl/SchedulerServiceImpl.java
lubomirw, line 1: **!! Critical** | Correctness

Question: will the `scheduled_job.job_class_name` data migration be handled in a follow-up PR?

`job_class_name` is persisted via `ScheduledJobTask.getJobClassName()` which returns `this.getClass().getName()`, so existing databases hold values like `com.czertainly.core.tasks.UpdateCertificateStatusTask`. After this PR, `runScheduledJob` hits `Class.forName(...)` → `ClassNotFoundException`, and every pre-existing job (certificate status update, discovery, Intune revocation, CBOM sync) would be recorded as FAILED "Unknown scheduled task" on every trigger. `registerScheduler` returns early when the job name already exists, so startup re-registration does not heal the rows. If a follow-up PR is planned for the upgrade path (e.g. `UPDATE scheduled_job SET job_class_name = REPLACE(job_class_name, 'com.czertainly.', 'com.otilm.');`), could you link it here?

src/main/java/com/otilm/core/config/SessionConfig.java
lubomirw, line 1: **! Major** | Correctness

Question: will migration of persisted HTTP sessions be handled in a follow-up PR?

Spring Session JDBC (`spring_session` table) serializes session attributes as JSON with `@class` type metadata via `SecurityJackson2Modules`, so rows written before this release can reference `com.czertainly.core.security.authn.*` FQCNs (e.g. the authentication token / user details stored in the security context). After the package move, loading such a session would throw on deserialization instead of resuming it. If a follow-up PR is planned to handle the upgrade path (e.g. clearing `spring_session` so users get a clean re-login), could you link it here? Otherwise this PR may need that migration itself.

pom.xml
klaraf755: Renamed

src/main/java/com/otilm/core/provider/PlatformProvider.java
klaraf755: This has already been updated

src/main/java/com/otilm/core/service/impl/SchedulerServiceImpl.java
klaraf755: This will be addressed in follow-up PR https://github.com/OmniTrustILM/core/pull/1584.

src/main/java/com/otilm/core/config/SessionConfig.java
klaraf755: This will be addressed in follow-up PR https://github.com/OmniTrustILM/core/pull/1584.

klaraf755:
> Code review: 5 issue(s) found. Three are posted as inline/file comments; two file-level questions follow as separate comments (their lines are unchanged context). One more note that cannot be attached inline because the Dockerfile is not part of this diff:
> 
> **~ Minor** | Maintainability
> 
> Question: is the deployment surface intentionally left czertainly-branded for a follow-up PR?
> 
> The container user/paths (`/opt/czertainly`), env var names (`CZERTAINLY_LOG_LEVEL`), broker exchange/vhost defaults (`czertainly`), and the `org.opencontainers.image.authors` label (`support@czertainly.com`) are all still on the old brand. Keeping the runtime contracts is the safe choice since they're shared with helm charts and other components — if a follow-up PR is planned to cover these, could you link it here and mention the scope boundary in the PR description?

This will be addressed in follow-up PR https://github.com/OmniTrustILM/core/pull/1584.

lubomirw (approved):

PR #1583, opened 2026-06-10
Expose enabled protocols in Signing Profile list DTO

### LINKED ISSUE #1581: Include enabled protocols in Signing Profile list endpoint response
## Description
The Signing Profile list endpoint (`listSigningProfiles`, returning `PaginationResponseDto<SigningProfileListDto>`) does not include the list of enabled protocols for each profile. The detail endpoint (`getSigningProfile`, returning `SigningProfileDto`) already exposes this via the `enabledProtocols` field (`List<SigningProtocol>`).

Add `enabledProtocols` to `SigningProfileListDto` so the list response carries the same protocol information as the detail response. This avoids having to call the detail endpoint per profile just to learn which protocols (e.g. `TSP`) each profile is enabled for.

Relevant code:
- List mapper: `SigningProfileMapper.toListDto(...)` (`src/main/java/com/czertainly/core/mapper/signing/SigningProfileMapper.java:163`) — does **not** populate protocols.
- Detail mapper: `SigningProfileMapper.toDto(...)` (`src/main/java/com/czertainly/core/mapper/signing/SigningProfileMapper.java:105`) — populates `enabledProtocols` from the unversioned header (e.g. `header.getTspProfileUuid() != null` → adds `SigningProtocol.TSP`).
- Service: `SigningProfileServiceImpl.listSigningProfiles` (`src/main/java/com/czertainly/core/service/impl/SigningProfileServiceImpl.java:162`).

The protocol data is derived from unversioned header fields, so it is available on the `SigningProfile` entity used by `toListDto` — the same detection logic from `toDto` can be reused.

## User Story
As an API consumer listing signing profiles, I want each profile in the list response to include its enabled protocols so that I can display or filter by protocol without issuing a separate detail request per profile.

## Use Case
A UI or integration renders a table of signing profiles and needs to show, per row, which protocols (e.g. `TSP`) the profile is enabled for. Today this requires N additional detail calls; with `enabledProtocols` on the list DTO it is a single list call.

## Acceptance Criteria
- [ ] `SigningProfileListDto` includes an `enabledProtocols` field of type `List<SigningProtocol>`.
- [ ] `SigningProfileMapper.toListDto(...)` populates `enabledProtocols` using the same header-derived detection logic as `toDto(...)` (e.g. TSP when `getTspProfileUuid() != null`, plus any other protocols already detected in the detail mapper).
- [ ] The list endpoint response returns protocols consistently with the detail endpoint for the same profile.
- [ ] Existing tests pass and coverage is added for the populated field in the list mapper.

ivosh:


lubomirw (commented):

src/main/java/com/otilm/core/mapper/signing/SigningProfileMapper.java
```diff
@@ -168,6 +171,7 @@ public static SigningProfileListDto toListDto(SigningProfile profile) {
         dto.setVersion(profile.getLatestVersion());
         dto.setSigningWorkflowType(profile.getWorkflowType());
         dto.setEnabled(profile.isEnabled());
+        dto.setEnabledProtocols(detectEnabledProtocols(profile));
```
lubomirw, line 175: **~ Minor** | Correctness

Question: is `signingScheme` on the list DTO intended to come in a later PR, or why is it not set here?

`SigningProfileListDto` declares `signingScheme` as `RequiredMode.REQUIRED`, but `toListDto` never populates it, so list responses return `null` for a required field. The `SigningProfile` entity already carries the denormalized `signingScheme` cache column, so `dto.setSigningScheme(profile.getSigningScheme())` would close the gap in this same pass — unless there is a reason it is deliberately deferred.

src/main/java/com/otilm/core/mapper/signing/SigningProfileMapper.java
```diff
@@ -123,6 +120,13 @@ public static SigningProfileDto toDto(SigningProfile header, SigningProfileVersi
         return policy;
     }
 
+    /**
+     * Detects the protocols a Signing Profile is enabled for.
```
lubomirw: **. Nitpick** | Documentation

The deleted inline comment carried an invariant the new javadoc lost: protocol activation is **header-level (unversioned)** state.

The old code said `// Enabled protocols from header (unversioned)`. That distinction matters in this mapper, where most fields come from `SigningProfileVersion`. Suggest:

```java
/**
 * Detects the protocols a Signing Profile is enabled for.
 * Protocol activation is header-level (unversioned) state — it does not vary across profile versions.
 */
```

src/main/java/com/otilm/core/mapper/signing/SigningProfileMapper.java
```diff
@@ -123,6 +120,13 @@ public static SigningProfileDto toDto(SigningProfile header, SigningProfileVersi
...
+     */
+    private static List<SigningProtocol> detectEnabledProtocols(SigningProfile header) {
+        return header.getTspProfileUuid() != null ? List.of(SigningProtocol.TSP) : List.of();
```
lubomirw, line 127: **. Nitpick** | Maintainability

`detectEnabledProtocols` returns an immutable `List.of(...)`, which `setEnabledProtocols` swaps in over the DTO's mutable `new ArrayList<>()` default.

Any future caller following the pattern this PR just removed — `dto.getEnabledProtocols().add(...)` — will hit `UnsupportedOperationException` at runtime. All current usages are read-only, so nothing breaks today; this is just a footgun to be aware of. If a second protocol gets added by accumulating into the list, the helper will need to build a mutable list or callers must always go through it.

src/main/java/com/otilm/core/mapper/signing/SigningProfileMapper.java
ivosh: Signing scheme alone is not worth displaying because it would just distinguish delegated/managed. What would be worth exposing is the flattened combination: delegate/managed-static/managed-one-time/managed-session-time.

I'd defer it until we support more signing schemes in practice and we have a good understanding of all the possibilities.

lubomirw: Shouldn't it be then marked as NOT_REQUIRED in Schema for OpenAPI?

src/main/java/com/otilm/core/mapper/signing/SigningProfileMapper.java
ivosh: Fixed in 1ff56a76.

src/main/java/com/otilm/core/mapper/signing/SigningProfileMapper.java
ivosh: Acknowledged — the immutable `List.of(...)` return is intentional. All current callers are read-only; if a future change accumulates protocols, the helper will build a mutable list at that point.

lubomirw (dismissed):

src/main/java/com/otilm/core/mapper/signing/SigningProfileMapper.java
ivosh: Fixed in fd6421a5 — `toListDto` now populates the required `signingScheme` from the profile's denormalized cache column.

lubomirw (approved):

PR #1584, opened 2026-06-11
Finish ILM rebrand: Docker paths, broker names, persisted class-name migration

### LINKED ISSUE #1566: Rebrand CZERTAINLY to ILM

klaraf755:


src/main/resources/application.yml
```diff
@@ -264,9 +264,9 @@ proxy:
   # Feature flag to enable/disable proxy messaging support
   enabled: ${PROXY_ENABLED:false}
   # Azure Service Bus topic / RabbitMQ exchange name for proxy messages
-  exchange: ${BROKER_PROXY_EXCHANGE:czertainly-proxy}
+  exchange: ${BROKER_PROXY_EXCHANGE:ilm-proxy}
   # Azure Service Bus subscription / RabbitMQ queue name for receiving responses
-  response-queue: ${BROKER_PROXY_RESPONSE_QUEUE:core}
+  response-queue: ${BROKER_PROXY_RESPONSE_QUEUE:ilm}
   # Unique identifier for this Core instance (defaults to hostname/pod name)
```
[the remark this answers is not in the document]
klaraf755: It should have stayed as core there.

hooks/build_win.bat
```diff
@@ -1,4 +1,4 @@
-echo Build of CZERTAINLY CORE
+echo Build of ILM CORE
 
 echo Cleaning containers
 docker rm czertainlycont
```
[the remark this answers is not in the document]
klaraf755: Done

src/test/java/com/otilm/core/messaging/jms/producers/TimeQualityConfigurationProducerTest.java
```diff
@@ -43,7 +43,7 @@ void publishSnapshot_sendsMessageWithAllConfigurations() {
 
         ArgumentCaptor<Object> messageCaptor = ArgumentCaptor.forClass(Object.class);
         verify(jmsTemplate).convertAndSend(
-                eq("/exchanges/czertainly/time-quality.config"),
+                eq("/exchanges/core/time-quality.config"),
                 messageCaptor.capture(),
                 any(MessagePostProcessor.class));
```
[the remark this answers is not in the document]
klaraf755: Done

src/main/resources/db/migration/V202606101414__class_name_update.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE scheduled_job SET job_class_name = REPLACE(job_class_name, 'com.czertainly.core.tasks', 'com.otilm.core.tasks');
+
+UPDATE spring_session_attributes SET attribute_bytes = REPLACE(attribute_bytes, 'com.czertainly.core', 'com.otilm.core');
```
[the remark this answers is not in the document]
klaraf755: Done

src/main/java/com/otilm/core/messaging/proxy/ProxyProperties.java
```diff
@@ -17,14 +17,14 @@
...
          */
         String exchange,
 
         /**
          * Azure Service Bus subscription / RabbitMQ queue name for receiving
          * fire-and-forget messages (health checks, connector registration).
-         * Default: core
+         * Default: ilm
          */
         String responseQueue,
```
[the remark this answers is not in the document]
klaraf755: Fixed so it remains `core`.

src/main/java/com/otilm/core/messaging/proxy/ProxyProperties.java
```diff
@@ -57,10 +57,10 @@ public record ProxyProperties(
 ) {
     public ProxyProperties {
         if (exchange == null) {
-            exchange = "czertainly-proxy";
+            exchange = "ilm-proxy";
         }
         if (responseQueue == null) {
-            responseQueue = "core";
+            responseQueue = "ilm";
```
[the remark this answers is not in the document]
klaraf755: Fixed so it remains core.

src/test/java/com/otilm/core/messaging/proxy/ProxyPropertiesTest.java
```diff
@@ -76,8 +76,8 @@ void constructor_defaults_areApplied() {
                 null, null, null
         );
         assertThat(props.instanceId()).isEqualTo("test-instance");
-        assertThat(props.exchange()).isEqualTo("czertainly-proxy");
-        assertThat(props.responseQueue()).isEqualTo("core");
+        assertThat(props.exchange()).isEqualTo("ilm-proxy");
+        assertThat(props.responseQueue()).isEqualTo("ilm");
         assertThat(props.requestTimeout()).isEqualTo(Duration.ofSeconds(30));
```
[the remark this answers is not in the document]
klaraf755: Fixed so it remains core.

src/main/resources/db/migration/V202606101414__class_name_update.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE scheduled_job SET job_class_name = REPLACE(job_class_name, 'com.czertainly.core.tasks', 'com.otilm.core.tasks');
+
+UPDATE spring_session_attributes SET attribute_bytes = REPLACE(attribute_bytes, 'com.czertainly.core', 'com.otilm.core')::jsonb;
```
lubomirw: **!! Critical** | Correctness

`REPLACE()` called directly on a JSONB column — this migration fails and blocks application startup.

`attribute_bytes` is `JSONB` (defined in `V202410231035__create_session_tables.sql`), and PostgreSQL has no `replace(jsonb, text, text)` overload in any version. Verified against PostgreSQL 17 (the version used by the test suite):

```
ERROR:  function replace(jsonb, unknown, unknown) does not exist
```

This fails the Flyway migration and prevents the application from starting. Explicit casts fix it (verified working):

```sql
UPDATE spring_session_attributes SET attribute_bytes = REPLACE(attribute_bytes::text, 'com.czertainly.core', 'com.otilm.core')::jsonb;
```

Alternatively, since sessions are ephemeral, consider simply deleting rows from `spring_session` / `spring_session_attributes` instead — users re-authenticate and you avoid string surgery on serialized session state entirely.

Dockerfile
```diff
@@ -59,10 +60,10 @@ ENV HEADER_NAME=ssl-client-cert
 ENV SCHEDULED_TASKS_ENABLED=true
 ENV BROKER_HOST=
 ENV BROKER_PORT=5672
-ENV BROKER_VIRTUAL_HOST=czertainly
+ENV BROKER_VIRTUAL_HOST=/
```
lubomirw, line 63: **~ Minor** | Correctness

Image default vhost changed from `czertainly` to `/` — a semantic change beyond renaming, and it now diverges from the rest of the platform.

The Helm charts currently provision and inject vhost `czertainly` (`messaging-rabbitmq` definitions and `core-deployment.yaml` both default to `"czertainly"`), so under Helm this default is never used — but anyone running the image directly now silently lands on the RabbitMQ root vhost `/`. If the target end-state is vhost `/` (matching `src/test/resources/rabbitMQ-definitions.json` in this PR), please make sure the helm-charts rebrand updates `virtualHost` defaults in lockstep and that the change is called out in upgrade notes; otherwise keep the image default aligned with the charts until they move together.

docker/opt/otilm/entry.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+otilmHome="/opt/otilm"
+source ${otilmHome}/static-functions
+
+log "INFO" "Launching the Core"
+java $JAVA_OPTS -jar ./app.jar
```
lubomirw: **~ Minor** | Correctness

`java` is launched without `exec`, so the shell stays PID 1 and the JVM never receives SIGTERM — on `docker stop` / pod termination, the container hangs until the grace period expires and gets SIGKILLed, skipping graceful shutdown (in-flight requests, scheduled-job state, broker connection cleanup).

The old `entry.sh` had the same flaw, but since this file is being rewritten anyway, this is the moment to fix it:

```sh
exec java $JAVA_OPTS -jar ./app.jar
```

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -96,8 +97,7 @@ public class ScepServiceImpl implements ScepService {
             "SCEPStandard"
     );
 
-    @Value("${app.version}")
-    private String appVersion;
+    private IntuneConfigProperties intuneConfigProperties;
```
lubomirw, line 99: **. Nitpick** | Maintainability

The removed `@Value("${app.version}")` fields were the only `@Value` usages in both `ScepServiceImpl` and `UpdateIntuneRevocationRequestsTask`, but the `org.springframework.beans.factory.annotation.Value` imports were left behind in both files. Remove the now-unused imports.

Dockerfile
```diff
@@ -37,20 +37,21 @@ ENV PATH="${JAVA_HOME}/bin:${PATH}"
 # copy optimized JRE
 COPY --from=optimize /javaruntime $JAVA_HOME
 
-LABEL org.opencontainers.image.authors="CZERTAINLY <support@czertainly.com>"
+# support mail is not updated yet
+LABEL org.opencontainers.image.authors="ILM <support@czertainly.com>"
```
lubomirw: **. Nitpick** | Documentation

Old-brand support email survives the rebrand with only a loose comment. "Not updated yet" comments rot — nothing links this to follow-up work. Either reference a tracking issue in the comment, or use a neutral placeholder that won't need a second pass (e.g. the omnitrust.com domain once the mailbox exists).

lubomirw (changes requested):

src/main/resources/db/migration/V202606101414__class_name_update.sql
klaraf755: Fixed it

Dockerfile
klaraf755: `/` is the target end-state, in `develop-environment`, the vhost name is also `/`, so it is already non-label specific name being used.

docker/opt/otilm/entry.sh
klaraf755: Fixed

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
klaraf755: Removed imports.

Dockerfile
klaraf755: Changed to otilm placeholder.

lubomirw (approved):

PR #1586, opened 2026-06-11
Add signing URL to TSP Profile list DTO

### LINKED ISSUE #1579: Add signingUrl to TspProfileListDto for the TSP Profile listing
## Description
Expose the **Signing URL** of a TSP (Timestamping) profile on the **list** DTO so the frontend can render a `Signing URL` column in the TSP Profile listing, matching how the ACME/SCEP/CMP profile listings expose their protocol URLs.

Today the value is only available on the **detail** DTO, not the list DTO:

- `TspProfileDto` (detail) already carries `signingUrl` — set by `TspProfileMapper.toDto(...)` to `<contextPath>/v1/protocols/tsp/<name>/sign` (only when a default signing profile is present).
- `TspProfileListDto` (list) has **no** URL field — only `description`, `defaultSigningProfile`, `enabled` (plus `name`/`uuid` from `NameAndUuidDto`). `TspProfileMapper.toListDto(...)` therefore never sets it.

Compare with `AcmeProfileListDto`, which exposes `directoryUrl` on its list DTO — the TSP listing should follow the same pattern.

Without this, the frontend (`OmniTrustILM/fe-administrator#1730`) cannot render the column from the list endpoint and would otherwise have to fetch each profile's detail per row, which is inconsistent with how ACME/SCEP/CMP listings work.

## Definition of Done
- [ ] Add a `signingUrl` (String, NOT_REQUIRED) field to `TspProfileListDto` in `OmniTrustILM/interfaces` (`com.czertainly.api.model.client.signing.protocols.tsp.TspProfileListDto`), with a `@Schema` description mirroring `AcmeProfileListDto.directoryUrl`.
- [ ] Populate it in `TspProfileMapper.toListDto(...)` (`OmniTrustILM/core`) using the same construction as `toDto(...)` — `<contextPath>/v1/protocols/tsp/<name>/sign`, guarded on `defaultSigningProfile != null` so list and detail behave identically.
- [ ] Regenerate the OpenAPI spec / generated clients so `TspProfileListDto.signingUrl` is available to consumers (incl. `fe-administrator` TS model).
- [ ] Update/extend `TspProfileServiceImplTest` (and search/list tests as needed) to assert the list DTO returns the expected `signingUrl`.
- [ ] Coordinate with `OmniTrustILM/fe-administrator#1730` which consumes this field.

ivosh:


lubomirw (commented):
Code review: 1 issue(s) found.

src/main/java/com/otilm/core/mapper/signing/TspProfileMapper.java
```diff
@@ -16,6 +16,11 @@ public class TspProfileMapper {
     private TspProfileMapper() {
     }
 
+    private static String buildSigningUrl(TspProfile profile) {
+        return ServletUriComponentsBuilder.fromCurrentContextPath().build().toUriString()
```
lubomirw, line 20: **~ Minor** | Maintainability

Profile name is concatenated into the URL path without encoding.

If a TSP profile name can contain characters that are reserved in a URL path segment (spaces, `/`, `?`, `#`), the resulting `signingUrl` will be malformed or point to a different path. Since this PR centralizes the URL construction into `buildSigningUrl`, it is a good opportunity to build the URL with `UriComponentsBuilder.pathSegment("v1", "protocols", "tsp", profile.getName(), "sign")`, which percent-encodes each segment. If profile names are already validated to a URL-safe character set, feel free to resolve this as-is.

lubomirw (approved):

src/main/java/com/otilm/core/mapper/signing/TspProfileMapper.java
ivosh: Names are validated at the API boundary: `TspProfileRequestDto.name` carries `@NotBlank @ValidName`, and `ValidNameValidator` accepts only RFC 3986 unreserved characters (`[A-Za-z0-9._~-]`). Reserved path-segment characters (space, `/`, `?`, `#`) are rejected before persistence, so the concatenated `signingUrl` can't be malformed. Resolving as-is per your note.

PR #1592, opened 2026-06-12
Add authentication methods and Basic credentials to TSP Profiles

### LINKED ISSUE #1593: TSP Profile authentication management (core implementation)
## Description
Backend (core) implementation of authentication management for TSP (Timestamping Protocol) Profiles, consuming the interface contracts from `OmniTrustILM/interfaces`. A TSP profile can now declare its allowed authentication methods and own a set of vault-backed **Basic** credentials (username/password) used to authenticate timestamping clients.

Scope:
- New entity `TspProfileBasicCredential` with a `(tsp_profile_uuid, username)` unique constraint; the password is stored as a vault secret.
- New CRUD service + controller family under `/v1/tspProfiles/{tspProfileUuid}/basicCredentials`, gated by a new authorization resource `TSP_PROFILE_BASIC_CREDENTIAL` (parent `TSP_PROFILE`).
- `TspProfile` gains `vaultProfileUuid` and `allowedAuthenticationMethods`; create/update now validate and dereference these.
- Flyway migration `V202606020900__tsp_profile_authentication.sql` adds the columns/constraint and backfills existing profiles with `CLIENT_CERTIFICATE`.
- Secret lifecycle wired through `SecretService` / `VaultProfileService` (create/rotate/delete); profile teardown deletes associated credential secrets, tolerating per-credential vault failures.

Parent: OmniTrustILM/ilm#145 — "Dedicated authentication stack for TSP protocol".

## Definition of Done
- [ ] `TspProfile` exposes `vaultProfileUuid` and `allowedAuthenticationMethods`; create/update validate them and existing profiles are backfilled with `CLIENT_CERTIFICATE`.
- [ ] Basic credentials can be listed, retrieved, created, updated, and deleted via `/v1/tspProfiles/{tspProfileUuid}/basicCredentials`, all gated by `TSP_PROFILE_BASIC_CREDENTIAL`.
- [ ] Credential passwords are persisted as vault secrets and rotated/deleted in sync with the credential row.
- [ ] Deleting a TSP profile removes its credential rows and associated vault secrets.
- [ ] Unit tests for service, mapper, and model pass.
- [ ] Release notes / role setup document the new `TSP_PROFILE_BASIC_CREDENTIAL` resource, the required `SECRET`/`VAULT` grants, the new request/response fields, and the `CLIENT_CERTIFICATE` backfill default.

ivosh:
PR for interfaces: https://github.com/OmniTrustILM/interfaces/pull/708

### DIFF
--- a/src/main/java/com/otilm/core/dao/entity/signing/TspProfile.java
+++ b/src/main/java/com/otilm/core/dao/entity/signing/TspProfile.java
@@ -1,17 +1,27 @@
 package com.otilm.core.dao.entity.signing;
 
+import com.otilm.api.model.core.signing.TspAuthenticationMethod;
 import com.otilm.core.dao.entity.UniquelyIdentifiedAndAudited;
+import com.otilm.core.dao.entity.VaultProfile;
 import com.otilm.core.service.model.Securable;
+import jakarta.persistence.CascadeType;
 import jakarta.persistence.Column;
 import jakarta.persistence.Entity;
+import jakarta.persistence.EnumType;
+import jakarta.persistence.Enumerated;
 import jakarta.persistence.FetchType;
 import jakarta.persistence.JoinColumn;
 import jakarta.persistence.ManyToOne;
+import jakarta.persistence.OneToMany;
 import jakarta.persistence.Table;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.ToString;
+import org.hibernate.annotations.JdbcTypeCode;
+import org.hibernate.type.SqlTypes;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.UUID;
 
 @Getter
@@ -29,6 +39,28 @@ public class TspProfile extends UniquelyIdentifiedAndAudited implements Securabl
 
     @Column(name = "enabled", nullable = false)
     private boolean enabled = false;
+
+    @Column(name = "allowed_authentication_methods", nullable = false)
+    @Enumerated(EnumType.STRING)
+    @JdbcTypeCode(SqlTypes.ARRAY)
+    private List<TspAuthenticationMethod> allowedAuthenticationMethods = new ArrayList<>();
+
+    @OneToMany(mappedBy = "tspProfile", cascade = CascadeType.ALL, orphanRemoval = true, fetch = FetchType.LAZY)
+    @ToString.Exclude
+    private List<TspProfileBasicCredential> basicCredentials = new ArrayList<>();
+
+    @Column(name = "vault_profile_uuid")
+    private UUID vaultProfileUuid;
+
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "vault_profile_uuid", insertable = false, updatable = false)
+    @ToString.Exclude
+    private VaultProfile vaultProfile;
+
+    public void setVaultProfile(VaultProfile vaultProfile) {
+        this.vaultProfile = vaultProfile;
+        this.vaultProfileUuid = vaultProfile != null ? vaultProfile.getUuid() : null;
+    }
 
     @Column(name = "default_signing_profile_uuid")
     private UUID defaultSigningProfileUuid;
--- a/src/main/java/com/otilm/core/mapper/signing/TspProfileBasicCredentialMapper.java
+++ b/src/main/java/com/otilm/core/mapper/signing/TspProfileBasicCredentialMapper.java
@@ -1,0 +1,23 @@
+package com.otilm.core.mapper.signing;
+
+import com.otilm.api.model.client.signing.protocols.tsp.TspBasicCredentialDto;
+import com.otilm.api.model.common.NameAndUuidDto;
+import com.otilm.core.dao.entity.signing.TspProfileBasicCredential;
+
+import java.util.List;
+
+public class TspProfileBasicCredentialMapper {
+    private TspProfileBasicCredentialMapper() {
+    }
+
+    public static TspBasicCredentialDto mapToDto(TspProfileBasicCredential credential, String mappedUserName) {
+        TspBasicCredentialDto dto = new TspBasicCredentialDto();
+        dto.setUuid(credential.getUuid());
+        dto.setUsername(credential.getUsername());
+        NameAndUuidDto mapped = new NameAndUuidDto();
+        mapped.setUuid(credential.getMappedUserUuid().toString());
+        mapped.setName(mappedUserName);
+        dto.setMappedUser(mapped);
+        return dto;
+    }
+}
--- a/src/main/java/com/otilm/core/mapper/signing/TspProfileBasicCredentialMapper.java
+++ b/src/main/java/com/otilm/core/mapper/signing/TspProfileBasicCredentialMapper.java
@@ -4,8 +4,6 @@ import com.otilm.api.model.client.signing.protocols.tsp.TspBasicCredentialDto;
 import com.otilm.api.model.common.NameAndUuidDto;
 import com.otilm.core.dao.entity.signing.TspProfileBasicCredential;
 
-import java.util.List;
-
 public class TspProfileBasicCredentialMapper {
     private TspProfileBasicCredentialMapper() {
     }
--- a/src/main/java/com/otilm/core/service/TspProfileBasicCredentialService.java
+++ b/src/main/java/com/otilm/core/service/TspProfileBasicCredentialService.java
@@ -25,8 +25,7 @@ public interface TspProfileBasicCredentialService {
     void delete(SecuredParentUUID tspProfileUuid, SecuredUUID uuid) throws AttributeException, ConnectorCommunicationException, NotFoundException;
 
     /**
-     * Deletes the vault secret backing every Basic credential of the given TSP profile and evicts the
-     * corresponding credential-verification cache entries.
+     * Deletes the vault secret backing every Basic credential of the given TSP profile.
      * Runs without an ambient transaction so the vault HTTP calls never hold a database transaction open.
      */
     void deleteSecretsForProfile(UUID tspProfileUuid);
--- a/src/main/java/com/otilm/core/service/TspProfileBasicCredentialService.java
+++ b/src/main/java/com/otilm/core/service/TspProfileBasicCredentialService.java
@@ -1,0 +1,33 @@
+package com.otilm.core.service;
+
+import com.otilm.api.exception.AlreadyExistException;
+import com.otilm.api.exception.AttributeException;
+import com.otilm.api.exception.ConnectorCommunicationException;
+import com.otilm.api.exception.NotFoundException;
+import com.otilm.api.model.client.signing.protocols.tsp.TspBasicCredentialDto;
+import com.otilm.api.model.client.signing.protocols.tsp.TspBasicCredentialRequestDto;
+import com.otilm.core.security.authz.SecuredParentUUID;
+import com.otilm.core.security.authz.SecuredUUID;
+
+import java.util.List;
+import java.util.UUID;
+
+public interface TspProfileBasicCredentialService {
+
+    List<TspBasicCredentialDto> list(SecuredParentUUID tspProfileUuid) throws NotFoundException;
+
+    TspBasicCredentialDto get(SecuredParentUUID tspProfileUuid, SecuredUUID uuid) throws NotFoundException;
+
+    TspBasicCredentialDto create(SecuredParentUUID tspProfileUuid, TspBasicCredentialRequestDto request) throws AlreadyExistException, AttributeException, ConnectorCommunicationException, NotFoundException;
+
+    TspBasicCredentialDto update(SecuredParentUUID tspProfileUuid, SecuredUUID uuid, TspBasicCredentialRequestDto request) throws AlreadyExistException, AttributeException, ConnectorCommunicationException, NotFoundException;
+
+    void delete(SecuredParentUUID tspProfileUuid, SecuredUUID uuid) throws AttributeException, ConnectorCommunicationException, NotFoundException;
+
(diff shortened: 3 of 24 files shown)


src/main/java/com/otilm/core/mapper/signing/TspProfileBasicCredentialMapper.java
```diff
@@ -0,0 +1,23 @@
+package com.otilm.core.mapper.signing;
+
+import com.otilm.api.model.client.signing.protocols.tsp.TspBasicCredentialDto;
+import com.otilm.api.model.common.NameAndUuidDto;
+import com.otilm.core.dao.entity.signing.TspProfileBasicCredential;
+
+import java.util.List;
+
```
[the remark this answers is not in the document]
ivosh: Fixed in 8e92ed68.

src/main/java/com/otilm/core/service/impl/SecretServiceImpl.java
```diff
@@ -586,6 +586,21 @@ public List<SecretVersionDto> getSecretVersions(UUID uuid) throws NotFoundExcept
...
 
+    @Override
+    public Map<UUID, String> getLatestFingerprintsByUuid(List<UUID> secretUuids) {
+        if (secretUuids == null || secretUuids.isEmpty()) {
+            return Map.of();
+        }
+        Map<UUID, String> result = new HashMap<>();
+        for (Secret secret : secretRepository.findByUuidIn(secretUuids)) {
+            SecretVersion latest = secret.getLatestVersion();
+            if (latest != null && latest.getFingerprint() != null) {
```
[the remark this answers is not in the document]
ivosh: Fixed in 8e92ed68 — added a dedicated `findWithLatestVersionByUuidIn` finder with `@EntityGraph(attributePaths = {"latestVersion"})` and switched `getLatestFingerprintsByUuid` to it, so the latest version is fetched in one query. Left the shared `findByUuidIn` untouched to avoid affecting its other callers.

src/main/java/com/otilm/core/service/TspProfileBasicCredentialService.java
```diff
@@ -0,0 +1,33 @@
...
+
+    TspBasicCredentialDto update(SecuredParentUUID tspProfileUuid, SecuredUUID uuid, TspBasicCredentialRequestDto request) throws AlreadyExistException, AttributeException, ConnectorCommunicationException, NotFoundException;
+
+    void delete(SecuredParentUUID tspProfileUuid, SecuredUUID uuid) throws AttributeException, ConnectorCommunicationException, NotFoundException;
+
+    /**
+     * Deletes the vault secret backing every Basic credential of the given TSP profile and evicts the
+     * corresponding credential-verification cache entries.
+     * Runs without an ambient transaction so the vault HTTP calls never hold a database transaction open.
+     */
```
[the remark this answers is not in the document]
ivosh: Fixed in 8e92ed68 — corrected the Javadoc. The method only deletes the vault secrets; profile-level cache eviction is handled by the caller during profile teardown (`TspProfileServiceImpl#deleteTspProfile` -> `evictTspProfileCache`).

src/main/java/com/otilm/core/dao/entity/signing/TspProfile.java
```diff
@@ -30,6 +40,28 @@ public class TspProfile extends UniquelyIdentifiedAndAudited implements Securabl
...
     private boolean enabled = false;
 
+    @Column(name = "allowed_authentication_methods", nullable = false)
+    @Enumerated(EnumType.STRING)
+    @JdbcTypeCode(SqlTypes.ARRAY)
+    private List<TspAuthenticationMethod> allowedAuthenticationMethods = new ArrayList<>();
+
+    @OneToMany(mappedBy = "tspProfile", cascade = CascadeType.ALL, orphanRemoval = true, fetch = FetchType.LAZY)
+    @ToString.Exclude
+    private List<TspProfileBasicCredential> basicCredentials = new ArrayList<>();
```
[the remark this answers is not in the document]
ivosh: Not changing this. `basicCredentials` is a JPA-managed `@OneToMany(cascade = ALL, orphanRemoval = true)` collection — Hibernate replaces it with a `PersistentBag` and relies on the getter returning that same instance for dirty-tracking and orphan removal. Returning `List.copyOf(...)` would break lazy loading and cascade/orphan-removal semantics. Defensive copies on managed entity collections are an anti-pattern; this rule doesn't account for JPA.

src/main/java/com/otilm/core/service/impl/TspProfileBasicCredentialServiceImpl.java
```diff
@@ -0,0 +1,300 @@
...
+        VaultProfile vaultProfile = resolveVaultProfile(profile.getVaultProfileUuid());
+        UUID vaultInstanceUuid = vaultProfile.getVaultInstanceUuid();
+        SecretRequestDto secretRequest = new SecretRequestDto();
+        secretRequest.setName("tsp-basic-" + profile.getUuid() + "-" + username + "-" + UUID.randomUUID().toString().substring(0, 8));
+        secretRequest.setSecret(new BasicAuthSecretContent(username, password));
+        try {
+            SecretDetailDto created = secretService.createSecret(secretRequest,
+                    SecuredParentUUID.fromUUID(vaultProfile.getUuid()), SecuredUUID.fromUUID(vaultInstanceUuid));
+            return UUID.fromString(created.getUuid());
+        } catch (ConnectorException e) {
```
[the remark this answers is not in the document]
ivosh: Intentional. Reusing the authorized `SecretService` is a deliberate defense-in-depth choice — these are real vault secrets, so requiring SECRET + vault-profile membership is appropriate. Introducing an internal authorization-bypassing secret CRUD path is a larger architectural decision out of scope for this PR.

src/main/java/com/otilm/core/service/TspProfileBasicCredentialService.java
```diff
@@ -0,0 +1,32 @@
...
+    TspBasicCredentialDto update(SecuredParentUUID tspProfileUuid, SecuredUUID uuid, TspBasicCredentialRequestDto request) throws AlreadyExistException, AttributeException, ConnectorCommunicationException, NotFoundException;
+
+    void delete(SecuredParentUUID tspProfileUuid, SecuredUUID uuid) throws AttributeException, ConnectorCommunicationException, NotFoundException;
+
+    /**
+     * Deletes the vault secret backing every Basic credential of the given TSP profile.
+     * Runs without an ambient transaction so the vault HTTP calls never hold a database transaction open.
+     */
+    void deleteSecretsForProfile(UUID tspProfileUuid);
+}
```
lubomirw, line 32: **~ Minor** | Documentation

Question: the javadoc says this method "evicts the corresponding credential-verification cache entries", but the implementation only deletes vault secrets, and no credential-verification cache is registered in `CacheConfig`. Is this doc incorrect, or is the cache planned to arrive with the TSP Authentication Filter in a follow-up PR?

If it's coming later, consider trimming the sentence (or marking it as forward-looking) until the cache and its eviction actually exist — otherwise the contract promises security-relevant behavior the method doesn't deliver, and it's easy to forget to wire the eviction in when the cache lands.

src/main/java/com/otilm/core/service/impl/TspProfileBasicCredentialServiceImpl.java
```diff
@@ -0,0 +1,300 @@
...
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    public TspBasicCredentialDto update(SecuredParentUUID tspProfileUuid, SecuredUUID uuid, TspBasicCredentialRequestDto request) throws AlreadyExistException, AttributeException, ConnectorCommunicationException, NotFoundException {
+        TspProfile profile = getTspProfile(tspProfileUuid);
+        TspProfileBasicCredential credential = getCredentialScoped(tspProfileUuid, uuid);
+        ensureUsernameAvailable(profile.getUuid(), request.getUsername(), credential.getUuid());
+        validateMappedUserExists(request.getMappedUserUuid());
+
+        boolean rotate = request.getPassword() != null && !request.getPassword().isBlank();
+        if (rotate) {
+            rotateVaultSecret(credential.getSecretUuid(), request.getUsername(), request.getPassword());
```
lubomirw, line 123: **~ Minor** | Correctness

Renaming a credential without supplying a password leaves the old username inside the vault secret.

The vault secret stores a `BasicAuthSecretContent(username, password)` pair, but rotation only happens when a password is provided. An update that changes only the username updates the DB row while the secret content keeps the previous username — the two stores diverge permanently (the next password rotation happens to heal it, since rotation writes the new username too).

If verification only ever reads the password/fingerprint from the secret and takes the username from the credential row, this is harmless but confusing for anyone inspecting the vault. If anything ever consumes the username from the secret content, it's a real bug. Worth either rewriting the secret content on rename, or documenting that the username inside the secret is informational-only and may be stale.

src/main/java/com/otilm/core/service/impl/TspProfileBasicCredentialServiceImpl.java
```diff
@@ -0,0 +1,300 @@
...
+        row.setTspProfile(profile);
+        row.setUsername(request.getUsername());
+        row.setSecretUuid(secretUuid);
+        row.setMappedUserUuid(request.getMappedUserUuid());
+        try {
+            row = credentialWriter.insert(row);
+        } catch (DataIntegrityViolationException e) {
+            deleteVaultSecretQuietly(secretUuid);
+            throw new AlreadyExistException("A Basic credential with username '" + request.getUsername() + "' already exists on this profile.");
+        } catch (RuntimeException e) {
```
lubomirw, line 99: **~ Minor** | Error handling

Question: is remapping `RuntimeException` to `ValidationException` here intentional?

The catch block itself makes sense — it guarantees the orphaned vault secret gets cleaned up on any persistence failure. But the rethrow converts infrastructure failures (DB outage, pool exhaustion, anything unexpected) into a 422 validation error, telling the client their request was invalid when it wasn't. Elsewhere in core, `catch (RuntimeException)` blocks rethrow the original exception after their side effects (e.g. the signing-record strategies), letting the generic handler map it to 5xx.

If the remap is deliberate (e.g. to avoid leaking internals), a short comment would help; otherwise consider `throw e` after the cleanup so monitoring and clients get an accurate signal.

src/main/java/com/otilm/core/service/impl/TspProfileServiceImpl.java
```diff
@@ -323,13 +346,20 @@ private SigningProfile validateCreateUpdateRequest(TspProfileRequestDto request)
             }
         }
 
-        return defaultSigningProfile;
+        VaultProfile vaultProfile = null;
+        if (request.getVaultProfileUuid() != null) {
+            vaultProfile = vaultProfileService.getVaultProfileEntity(SecuredUUID.fromUUID(request.getVaultProfileUuid()));
```
lubomirw, line 352: **~ Minor** | Correctness

A profile update can remove or change `vaultProfileUuid` while Basic credentials still exist.

The bean-level `@VaultProfileRequiredForBasicPassword` validator only enforces the vault profile when `BASIC_PASSWORD` is in the allowed methods — and it cannot see the database. So an update that drops `BASIC_PASSWORD` (or swaps the vault profile UUID) passes validation while existing credential rows and their vault secrets remain, now tied to a vault profile the TSP profile no longer references. Those credentials become unreachable zombies: they can't authenticate (method disallowed) but still hold live secrets in the old vault.

Consider a server-side guard in `validateCreateUpdateRequest`: reject removing/changing the vault profile (or removing `BASIC_PASSWORD`) while credentials exist, or cascade-clean the credentials as part of the update.

lubomirw (commented):

src/main/java/com/otilm/core/service/impl/TspProfileBasicCredentialServiceImpl.java
ivosh: Fixed in 85295515 — the catch now rethrows the original exception after the vault cleanup + log, so a DB/pool failure surfaces as 5xx instead of a misleading 422.

src/main/java/com/otilm/core/service/impl/TspProfileBasicCredentialServiceImpl.java
ivosh: Fixed in 85295515 — documented the invariant. The secret content can't be rewritten on a username-only change since we don't hold the stored password, so the username inside the secret is informational-only; verification reads the username from the credential row, and a later password rotation heals it.

src/main/java/com/otilm/core/service/impl/TspProfileServiceImpl.java
ivosh: Fixed in 85295515 — added a guard in updateTspProfile that rejects changing or removing the vault profile while Basic credentials exist (those secrets live in the profile's vault and would be orphaned). Removing BASIC_PASSWORD is intentionally left allowed: the credentials are retained (hidden) and become usable again if the method is re-added — they only become unreachable if the vault profile itself changes, which is what the guard now blocks.

src/main/java/com/otilm/core/service/TspProfileBasicCredentialService.java
ivosh: The javadoc was already corrected in an earlier push — it no longer claims any credential-verification cache eviction. It now documents only that the method deletes the backing vault secret for each Basic credential and runs without an ambient transaction so the vault HTTP calls don't hold a DB transaction open. The verification cache (and its eviction) will land with the TSP Authentication Filter in a follow-up PR.

lubomirw (approved):

PR #1594, opened 2026-06-12
Expose callable RFC 3161 timestamping endpoints

### LINKED ISSUE #175: Check enable/disable flag for Signing Profile and TSP Profile
### Description

The signing request must check if the corresponding TSP Profile and Signing Profile are enabled before progressing further. If either of them is disabled, the signing request must fail. This variant holds regardless how the request arrives (via TSP Profile or via Signing Profile).
Naturally, methods for enable/disable TSP Profile and Signing Profile must evict the database cache.

This is consitent with how RA Profile and ACME/SCEP/CMP profiles are gated today. Both the RA Profile and ACME/SCEP/CMP profile must be enabled.

## Core Rule: Dual-Gate
All three protocols (ACME, SCEP, CMP) require BOTH the protocol profile AND the RA Profile to be enabled. Whether the request arrives via the RA Profile URL or via the protocol profile URL, both are checked.
### ACME
• RA Profile-based request → checks: RA Profile enabled first, then ACME Profile enabled
• Direct ACME Profile request → checks: ACME Profile enabled first, then RA Profile enabled
• Failure: AcmeProblemDocumentException / HTTP 400, problem type MALFORMED
• Null = disabled
### SCEP
• Same validation path for both entry points (validateScepProfile() then validateRaProfile())
• Failure: ScepException with FailInfo.BAD_REQUEST
• Null = disabled
### CMP
• Same validation path for both entry points (validateCmpProfile() then validateRaProfile())
• Failure: CmpConfigurationException → wrapped as CmpCrmfValidationException or CmpProcessingException with PKIFailureInfo.systemFailure
• Null = enabled ← differs from ACME/SCEP

### User Story

_No response_

### Use Case

_No response_

### Acceptance Criteria

- [] Signing request arrives on the TSP Profile. Both the TSP Profile and the linked default Signing Profile must be enabled for the request to proceed. If either of them is disabled, the request must fail.
- [] Signing request arrives on the Signing Profile. Both the TSP Profile and the linked default Signing Profile must be enabled for the request to proceed. If either of them is disabled, the request must fail.

### LINKED ISSUE #214: Remove `/sign` component from the TSP Signing URL
### Description

Timestamping Protocol (TSP) defines only `sign` operation, therefore `.../sign` component in the Timestamping signing URL is redundant.

This affects both `interfaces` (example values) and `core` (where the property is populated).

### Severity

Minor

### LINKED ISSUE #1564: Introduce TSP Controller sign operation
## Description
Introduce the **TSP Controller** sign operation from the `feat/signing` branch into `main`. The scope is limited to the timestamping **sign** entry point (`TspController.timestamp`) and the controller-layer call chain it invokes — directly or indirectly — to produce a TSP response.

This is the protocol-facing sign path only. TSP profile management endpoints (`TspProfileControllerImpl`, `TspSigningProfileControllerImpl`) are out of scope and tracked separately. The managed TSP timestamping engine (`TsaService` / `TsaServiceImpl` / `ManagedTimestampEngine`) is already merged into `main` and is **consumed, not introduced** by this task.

In scope (sign-operation call chain):
- `TspController` interface and `TspControllerImpl` — the `timestamp(tspProfileName, request)` endpoint, `@AuditLogged` with `Module.PROTOCOLS` / `Operation.SIGN`.
- `TspRequestParser` and `TspRequestParsingException` — RFC 3161 request parsing.
- `TSPResponseBuilder` — success and rejection response serialization.
- Wiring of the sign path into the already-merged `TsaService.processTspRequestForTspProfile(...)`.
- Associated integration coverage (`TspProtocolFlowITest`) for the sign flow.

Out of scope:
- Managed TSP timestamping engine (`TsaService`, `TsaServiceImpl`, `ManagedTimestampEngine`) — already in `main`.
- TSP profile management controllers (`TspProfileControllerImpl`, `TspSigningProfileControllerImpl`).

## Definition of Done
- [ ] TSP Controller sign operation and its controller-layer call chain merged into `main` from `feat/signing`.
- [ ] `TspController.timestamp` endpoint exposed and wired to the existing `TsaService`.
- [ ] Request parsing and response building paths integrated and building with `-Dmaven.compiler.proc=full`.
- [ ] Error paths return correctly built TSP rejections (`TspException` / `NotFoundException` / unexpected) without leaking raw runtime detail to the wire.
- [ ] Audit logging applied to the sign operation (`Module.PROTOCOLS`, `Resource.SIGNING_RECORD`, `Operation.SIGN`).
- [ ] Integration test for the sign flow passes against `main`.
- [ ] Out-of-scope items (TSA engine, profile management controllers) explicitly excluded from this merge.

LukasNajman:


src/main/java/com/otilm/core/util/TspProtocolUrlFactory.java
```diff
@@ -0,0 +1,30 @@
...
+    public static String forTspProfile(String baseUrl, String tspProfileName) {
+        return baseUrl + TSP_PROTOCOL_BASE_PATH + "/" + tspProfileName;
+    }
+
+    /**
+     * Endpoint a client posts timestamp requests to for a TSP-activated signing profile.
+     */
+    public static String forSigningProfile(String baseUrl, String signingProfileName) {
+        return baseUrl + TSP_PROTOCOL_BASE_PATH + "/signingProfiles/" + signingProfileName;
+    }
```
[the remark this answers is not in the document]
LukasNajman: This is false positive, the name is @ValidName, where it's javadoc says

```
 Validates that a name contains only unreserved URI characters as defined in RFC 3986:
 alphanumeric characters (a-z, A-Z, 0-9), hyphen (-), period (.), underscore (_), and tilde (~).
 ```

File: src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java [hunk 2]
```diff
@@ -40,18 +41,40 @@ public TspResponse processTspRequestForTspProfile(String tspProfileName, TspRequ
...
     @Override
     public TspResponse processTspRequestForSigningProfile(String signingProfileName, TspRequest request) throws NotFoundException, TspException {
         SigningProfileModel<?, ?> signingProfile = signingProfileService.getSigningProfileModel(signingProfileName);
 
+        TspProfileModel tspProfile = tspProfileService.getTspProfile(signingProfile.tspProfileName());
+        if (!tspProfile.enabled()) {
+            throw new TspException(TspFailureInfo.BAD_REQUEST, "TSP profile '%s' is disabled".formatted(tspProfile.name()), "The TSP profile is disabled and cannot be used for timestamping.");
+        }
+
+        return processTspRequest(signingProfile, request);
```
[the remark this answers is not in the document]
LukasNajman: Added a check for this

src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java
```diff
@@ -40,18 +41,40 @@ public TspResponse processTspRequestForTspProfile(String tspProfileName, TspRequ
 
         TspProfileModel tspProfile = tspProfileService.getTspProfile(tspProfileName);
 
-        return processTspRequestForSigningProfile(tspProfile.defaultSigningProfileName(), request);
+        if (!tspProfile.enabled()) {
+            throw new TspException(TspFailureInfo.BAD_REQUEST, "TSP profile '%s' is disabled".formatted(tspProfileName), "The TSP profile is disabled and cannot be used for timestamping.");
+        }
+
+        SigningProfileModel<?, ?> signingProfile = signingProfileService.getSigningProfileModel(tspProfile.defaultSigningProfileName());
+        return processTspRequest(signingProfile, request);
```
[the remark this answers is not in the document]
LukasNajman: Added a check for this

src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java
```diff
@@ -37,21 +40,49 @@ public TsaServiceImpl(TspRequestValidator tspRequestValidator, SigningProfileSer
 
     @Override
     public TspResponse processTspRequestForTspProfile(String tspProfileName, TspRequest request) throws NotFoundException, TspException {
-
         TspProfileModel tspProfile = tspProfileService.getTspProfile(tspProfileName);
+        LoggingHelper.putLogResourceInfo(Resource.TSP_PROFILE, true, tspProfile.uuid().toString(), tspProfile.name());
+
+        if (tspProfile.defaultSigningProfileName() == null) {
+            var message = "TSP profile '%s' does not have a default signing profile".formatted(tspProfile.name());
+            throw new TspException(TspFailureInfo.BAD_REQUEST, message, message);
```
ivosh, line 48: **! Major** | Security

TsaServiceImpl.java:48, :62, :73, :78 each build a `message` string interpolating a runtime value (`tspProfile.name()` / `signingProfile.name()`, plus config state) and pass that same `message` as BOTH the log message AND the `clientMessage` (3rd arg) of `new TspException(...)`. The clientMessage is forwarded into TspResponseBuilder.buildRejection -> PKIFreeText in the DER TimeStampResp. Both TspController and TspSigningProfileController extend the new InbandProtocolController, so this text reaches an anonymous wire surface. 

Violates the TspException Javadoc (clientMessage must be a static hardcoded string) and CLAUDE.md 'Don't leak runtime details to the wire'. Sibling case at TsaServiceImpl.java:83-86 uses a static clientMessage.

What is a genuine database-sourced disclosure is the configuration state — "this profile exists but has no default signing profile configured." An unauthenticated caller learns server-side config details (which named profiles exist and how they're wired) by probing. That's the leak this finding is about.

 Fix: keep interpolated message as log arg only, pass static clientMessage.

src/main/java/com/otilm/core/api/tsp/parser/TspRequestParser.java
```diff
@@ -0,0 +1,111 @@
...
+            return null;
+        }
+    }
+
+    /**
+     * BouncyCastle leaves an extension's {@code extnValue} as an opaque OCTET STRING — it never decodes the
+     * bytes inside. Confirm that inner content is at least well-formed DER; per-OID type checking is out of
+     * scope because for now we accept all extensions.
+     */
+    private static void ensureValidExtension(Extension extension) throws TspRequestParsingException {
```
ivosh, line 82: **~ Minor** | Documentation

TspRequestParser.java:82-90 (ensureValidExtension) only verifies extnValue is well-formed DER; criticality and per-OID support are not checked, and getExtensions() carries every extension through to the connector formatter (isCritical forwarded verbatim). RFC 3161 2.4.2 requires the TSA to reject a request with unacceptedExtension when it gets an unsupported critical extension.

Add a section to the documentation describing this as intentional.

src/main/java/com/otilm/core/mapper/signing/TspProfileMapper.java
```diff
@@ -6,10 +6,9 @@
 import com.otilm.api.model.client.signing.protocols.tsp.TspProfileListDto;
 import com.otilm.core.dao.entity.signing.SigningProfile;
 import com.otilm.core.dao.entity.signing.TspProfile;
-import com.otilm.core.dao.entity.signing.TspProfileBasicCredential;
 import com.otilm.core.model.signing.TspProfileModel;
+import com.otilm.core.util.TspProtocolUrlFactory;
 import com.otilm.core.model.signing.TspProfileModel.BasicCredentialRef;
```
ivosh, line 10: **. Nitpick** | Maintainability

TspProfileMapper.java:11 inserts import com.otilm.core.util.TspProtocolUrlFactory; before import com.otilm.core.model.signing.TspProfileModel.BasicCredentialRef; (line 12). com.otilm.core.util sorts after com.otilm.core.model, so the block is no longer ASCII-alphabetical (Google Java Style 3.3.3). Rest of block is sorted, so this is a regression. Move the TspProtocolUrlFactory import below the TspProfileModel imports. Non-blocking.

src/main/java/com/otilm/core/api/tsp/TspControllerImpl.java
```diff
@@ -0,0 +1,68 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@Slf4j
+public class TspControllerImpl implements TspController {
+
+    private TsaService tsaService;
```
ivosh: **. Nitpick** | Maintainability

TspControllerImpl.java:26-28 (and the identical TspSigningProfileControllerImpl.java:26-28) declare `private TsaService tsaService;` / `private AuditResultOverride auditResultOverride;` as non-final fields wired via `@Autowired` setters. The established neighbour pattern for controllers in this codebase is constructor injection with `final` fields — e.g. `SigningProfileControllerImpl.java:52` and `TspProfileControllerImpl.java:37` both use an `@Autowired` constructor. Constructor injection keeps the dependencies final/immutable (Effective Java Item 17) and is the prevailing controller idiom here. 

Non-blocking style note.

src/main/java/com/otilm/core/model/signing/SigningProfileModel.java
```diff
@@ -25,6 +25,11 @@
  * @param version           Current version number.
  * @param enabled           Whether the profile is currently enabled.
  * @param enabledProtocols  Protocols enabled on this profile (e.g. TSP).
+ * @param tspProfileUuid    UUID of the TSP Profile this signing profile is linked to (TSP activation),
```
ivosh, line 28: .Minor | Documentation

I'd suggest to trim down the parameter description to just one line. There is no need to re-iterate again the rationale stated at line 18.

ivosh (changes requested):

src/main/java/com/otilm/core/api/tsp/parser/TspRequestParser.java
LukasNajman: We have a follow-up task for this, that will address the issue that all well-formed exceptions are now accepted. https://github.com/OmniTrustILM/ilm/issues/249

src/main/java/com/otilm/core/mapper/signing/TspProfileMapper.java
LukasNajman: reordered

src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java
LukasNajman: **1. The caller is authenticated, so there is no anonymous wire surface.** TSP endpoints fall under
  `.anyRequest().authenticated()` (`SecurityConfig.java:49-50`) — they are not in the permit-all list. And unlike
  ACME/SCEP/CMP, which authenticate as a single shared protocol pseudo-user, TSP will authenticate each caller as a named user with their own password. The "unauthenticated caller probes the server for config" scenario the finding is premised on does not exist.

  **2. The message content is fully known and contains no implementation details.** These strings are not derived from runtime internals — no exception text, SQL, stack frames, or table/column names. They contain only the profile name (which the authenticated caller is operating against) and its high-level configuration state (enabled/disabled, whether a TSP profile is associated). That is intentional, actionable feedback for the authenticated user (e.g. during the initial setup and testing when the user will probably have enough rights to do something about it).
 
The `CLAUDE.md` "don't leak runtime details to the wire" rule targets *uncontrolled* `Exception.getMessage()` forwarding; these messages are controlled, static-shaped strings we author, so the rule isn't violated in spirit.

src/main/java/com/otilm/core/model/signing/SigningProfileModel.java
LukasNajman: trimmed down

src/main/java/com/otilm/core/api/tsp/TspControllerImpl.java
LukasNajman: Used constructor injection

ivosh (approved):

Issue #1603, opened 2026-06-14
Implement v3 certificate register + availableOperations operator flow


lubomirw:
### Description

Implement the v3 operator-facing **register** flow + **capability discovery** on top of S3's endpoints/adapter and S4's polling. The `register` and `listAvailableOperations` REST endpoints exist as `NotSupportedException` stubs (S3 #1614); this issue makes them real. `cancelPendingCertificateOperation` (S3) and the insert-on-202 poll scheduling (S4 #1627) are already done.

Part of the authority-provider-v3 epic OmniTrustILM/ilm#110.

### Scope

- **`registerCertificate`** (`ClientOperationServiceImpl` + service interface; wire the controller stub): `instanceof RegisterCapability` gate → **no-CSR placeholder** (identity from `ClientCertificateRegistrationDto` DN/SAN/extensions → `X500Name` → `setSubjectDNParams`, state `REQUESTED`) → `@Transactional(NOT_SUPPORTED)` + `adapter.register(...)` → branch `ASYNC_ACCEPTED` (SM `REQUESTED→PENDING_REGISTRATION` + `scheduleStatusPoll(cert, REGISTER)`) vs `SYNC_OK` (SM `→PENDING_REGISTRATION→REGISTERED`) → meta persist → `ClientCertificateDataResponseDto`.
- **`listAvailableOperations`** (+ wire the controller stub): resolve the adapter, build one `OperationSupport(kind, supported, asyncSupported, cancelSupported)` per `CertificateOperationKind` via the capability `instanceof` checks → `AvailableOperationsDto`.
- **REGISTERED-state issue** — issue a CSR against a `REGISTERED` placeholder (`addCertificateRequestToExisting` / `issueExistingCertificate` with CSR attach; `REGISTERED→PENDING_ISSUE`). **Verify** the released-interfaces `issueExistingCertificate` `consumes=application/json` vs bodiless-REQUESTED 415 risk (or file a follow-up interfaces PR relaxing `consumes`).
- **`manuallyIssueCertificate` v3 hook** — post-commit best-effort `adapter.cancel(...)` for an in-flight async tracking.
- Inject `AuthorityProviderAdapterFactory` + `CertificateStateMachine` into `ClientOperationServiceImpl` (mirror how S3's cancel reaches them).

### Already shipped (not in scope)

- `cancelPendingCertificateOperation` + the register/availableOperations endpoint stubs — S3 (#1614).
[...]

PR #1605, opened 2026-06-14
Add certificate state machine for lifecycle transitions

### LINKED ISSUE #1599: Introduce certificate state machine for lifecycle transitions
### Description

Replace the scattered `cert.setState(...)` calls in `service/v2/impl/ClientOperationServiceImpl` with an explicit, validated certificate state machine:

- `CertificateStateTransition` — the (from → to) transition table with the default audit event per row.
- `CertificateStateMachine` — validates the requested transition, mutates state, and writes the audit / event-history entry.
- `InvalidTransitionException` — thrown on an undeclared transition.

Behavior-preserving: every existing transition keeps the same outcome and audit semantics. Packages are `com.otilm.*`.

**Review focus:** transition-table completeness; audit-event parity; no live transition newly rejected.

### Definition of Done

- [ ] `CertificateStateTransition`, `CertificateStateMachine`, `InvalidTransitionException` added with unit tests
- [ ] All `cert.setState(...)` sites in `service/v2/impl/ClientOperationServiceImpl` routed through the state machine
- [ ] Audit-event parity verified — no existing transition changes behavior
- [ ] PR merged to `main`

lubomirw:
First slice of landing `feat/authority-v3` on `main` as a reviewable stacked-PR chain. Part of #1599 (epic #110).

## What

Adds the certificate lifecycle state machine — three new classes plus unit tests:

- `CertificateStateTransition` — the `(from → to)` transition table with a default audit event per row.
- `CertificateStateMachine` — validates a requested transition, mutates state, and writes the audit / event-history entry (runs in the caller's transaction).
- `InvalidTransitionException` — thrown on an undeclared transition, carrying the attempted `(from, to)` pair.

## Notes for review

- **Purely additive.** Routing `ClientOperationServiceImpl`'s `setState(...)` calls through the state machine lands in the **activation slice** (#1603): the hub file is a single file and is carried whole there to keep each PR's diff to one slice. Nothing calls the state machine yet in this PR — it compiles and is tested in isolation.
- Code originates from `feat/authority-v3` (already reviewed there); this re-slicing is for incremental review, not new development.
- Behavior-preserving for the platform as a whole; no wire or schema change.

## Review addressed

Copilot's auto-review findings folded in:
- `transition(...)` is now `@Transactional` so the state write and audit write commit atomically.
- Null-guards on `cert` / `toState`.
- Removed a brittle line-number reference from the transition-table Javadoc.

## Verification

`mvn test` for the three test classes: **8 tests, 0 failures**.


src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachine.java
```diff
@@ -0,0 +1,72 @@
...
+     * @param auditEventOverride if non-null, overrides the row's defaultAuditEvent
+     * @param reasonMessage if non-null, used as the audit-history message; otherwise auto-generated
+     */
+    public void transition(Certificate cert, CertificateState toState,
+                           @Nullable CertificateEvent auditEventOverride,
+                           @Nullable String reasonMessage) {
+        CertificateState fromState = cert.getState();
+        CertificateStateTransition row = CertificateStateTransition.lookup(fromState, toState)
+            .orElseThrow(() -> new InvalidTransitionException(
+                Resource.CERTIFICATE, cert.getUuid(), fromState, toState));
```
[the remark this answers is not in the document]
lubomirw: Addressed in 061de2a — added `Objects.requireNonNull(cert, "cert")` and `Objects.requireNonNull(toState, "toState")` at the top of `transition(...)`, so a null input fails fast with a clear message instead of surfacing as an NPE while building the exception.

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachine.java
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * <p>SM does NOT drive side effects (location cleanup, predecessor relations, dual-cert event
+ * history) that v2's handleFailedOrRejectedEvent bundles. Caller remains responsible for those.</p>
+ *
+ * <p>SM governs state CHANGES, not initial assignment at row creation. Paths that set state
+ * directly on a new Certificate (CertificateServiceImpl.createPlaceholder,
+ * CertificateUtil.prepareIssuedCertificate) bypass the SM by design.</p>
+ */
+@Service
+public class CertificateStateMachine {
```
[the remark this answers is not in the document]
lubomirw: Addressed in 061de2a — `transition(...)` is now `@Transactional` (REQUIRED), so the state write and the audit-history insert commit atomically. The state machine performs only local writes and no external calls, so joining the caller's ambient transaction is safe and matches the codebase's 'transactional boundaries live on services' guidance.

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateTransition.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Optional;
+
+import static com.otilm.api.model.core.certificate.CertificateState.*;
+
+/**
+ * Allowed certificate state transitions, keyed on (from, to) pairs.
+ * Every distinct lifecycle transition has a unique (from, to) pair.
+ * The reason for a transition (e.g., issue-failed vs cancelled vs timed out) is captured
+ * via the reasonMessage parameter on {@link CertificateStateMachine#transition} — not via
+ * separate rows. This matches v2's line 386 pattern of free-text audit-history reasons.
```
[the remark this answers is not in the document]
lubomirw: Addressed in 061de2a — removed the brittle `line 386` reference; the Javadoc now describes the free-text audit-history reason pattern without pointing at a specific line.

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachine.java
```diff
@@ -0,0 +1,82 @@
...
+     * @param auditEventOverride if non-null, overrides the row's defaultAuditEvent
+     * @param reasonMessage if non-null, used as the audit-history message; otherwise auto-generated
+     */
+    @Transactional
+    public void transition(Certificate cert, CertificateState toState,
+                           @Nullable CertificateEvent auditEventOverride,
+                           @Nullable String reasonMessage) {
+        Objects.requireNonNull(cert, "cert");
+        Objects.requireNonNull(toState, "toState");
+        CertificateState fromState = cert.getState();
```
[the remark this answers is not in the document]
lubomirw: Addressed in 76398caba — made `InvalidTransitionException` null-safe: it now renders a missing state as "none" via a private `label(...)` helper instead of calling `getLabel()` on a null `fromState`. The exception can no longer NPE while formatting its own message, regardless of caller.

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateTransition.java
```diff
@@ -0,0 +1,74 @@
...
+    PENDING_APPROVAL_TO_PENDING_REVOKE(PENDING_APPROVAL,    PENDING_REVOKE,       CertificateEvent.APPROVAL_CLOSE),
+    PENDING_APPROVAL_TO_ISSUED       (PENDING_APPROVAL,     ISSUED,               CertificateEvent.APPROVAL_CLOSE),
+
+    // v3 register lifecycle
+    REQUESTED_TO_PENDING_REGISTRATION(REQUESTED,            PENDING_REGISTRATION, CertificateEvent.UPDATE_STATE),
+    PENDING_REGISTRATION_TO_REGISTERED(PENDING_REGISTRATION,REGISTERED,           CertificateEvent.UPDATE_STATE),
+    PENDING_REGISTRATION_TO_FAILED   (PENDING_REGISTRATION, FAILED,               CertificateEvent.UPDATE_STATE),
+    REGISTERED_TO_PENDING_ISSUE      (REGISTERED,           PENDING_ISSUE,        CertificateEvent.ISSUE),
+
+    // Compliance rejection (v2 line 481)
```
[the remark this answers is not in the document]
lubomirw: Addressed in 76398caba — removed the second `v2 line 481` reference (the comment now reads just "Compliance rejection"). Swept the rest of the file; no other line-number references remain.

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachine.java
```diff
@@ -0,0 +1,91 @@
...
+ *   <li>Validate (from, to) pair against {@link CertificateStateTransition} table — throw on miss</li>
+ *   <li>Mutate cert.state + persist via repository</li>
+ *   <li>Write audit history entry via {@link CertificateEventHistoryInternalService}</li>
+ * </ol>
+ *
+ * <p>SM does NOT call entityManager.flush() — callers can mutate the same entity post-transition
+ * within the same transaction; mutations flush at commit (Hibernate dirty-tracking).</p>
+ *
+ * <p>SM does NOT drive side effects (location cleanup, predecessor relations, dual-cert event
+ * history) that v2's handleFailedOrRejectedEvent bundles. Caller remains responsible for those.</p>
```
ivosh: What is v2 here?

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachine.java
```diff
@@ -0,0 +1,91 @@
...
+    private final CertificateRepository certificateRepository;
+    private final CertificateEventHistoryInternalService eventHistoryService;
+
+    public CertificateStateMachine(CertificateRepository certificateRepository,
+                                   CertificateEventHistoryInternalService eventHistoryService) {
+        this.certificateRepository = certificateRepository;
+        this.eventHistoryService = eventHistoryService;
+    }
+
+    /** Default: SM resolves the audit event from the transition row's defaultAuditEvent. */
```
ivosh: You could mention that also reason message is auto-generated here.

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachine.java
```diff
@@ -0,0 +1,91 @@
...
+
+        cert.setState(toState);
+        certificateRepository.save(cert);
+
+        CertificateEvent auditEvent = auditEventOverride != null ? auditEventOverride : row.defaultAuditEvent;
+        String message = reasonMessage != null
+            ? reasonMessage
+            : "State changed from %s to %s".formatted(fromState.getLabel(), toState.getLabel());
+        eventHistoryService.addEventHistory(cert.getUuid(), auditEvent,
+            CertificateEventStatus.SUCCESS, message, "");
```
ivosh: applyTransition (CertificateStateMachine.java:88-89) unconditionally writes CertificateEventStatus.SUCCESS for every transition, including failure transitions the table explicitly models: PENDING_ISSUE_TO_FAILED, REQUESTED_TO_FAILED, PENDING_APPROVAL_TO_FAILED, PENDING_REGISTRATION_TO_FAILED and the rejection rows PENDING_APPROVAL_TO_REJECTED / ISSUED_TO_REJECTED / REQUESTED_TO_REJECTED. The established neighbour pattern this SM is meant to absorb does the opposite: ClientOperationServiceImpl.handleFailedOrRejectedEvent records CertificateEventStatus.FAILED whenever state becomes FAILED (ClientOperationServiceImpl.java:503-507) and for REJECTED-with-message (line 514), reserving SUCCESS only for the message-less REJECTED case (line 512). issueCertificateRejectedAction writes status FAILED for the ISSUED->REJECTED audit (line 1016). When activation slice #1603 routes those calls through transition(), failure transitions will be mis-audited as SUCCESS. Suggest deriving status from toState (FAILED/REJECTED -> FAILED) or threading an explicit status.

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachine.java
```diff
@@ -0,0 +1,91 @@
...
+                                 @Nullable String reasonMessage) {
+        Objects.requireNonNull(cert, "cert");
+        Objects.requireNonNull(toState, "toState");
+        CertificateState fromState = cert.getState();
+        CertificateStateTransition row = CertificateStateTransition.lookup(fromState, toState)
+            .orElseThrow(() -> new InvalidTransitionException(
+                Resource.CERTIFICATE, cert.getUuid(), fromState, toState));
+
+        cert.setState(toState);
+        certificateRepository.save(cert);
```
ivosh, line 84: applyTransition reads fromState=cert.getState() so it is susceptible for TOCTOU race. Have you considered a scenario with parallel certificate transitions? The last-to-write wins?

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateTransition.java
```diff
@@ -0,0 +1,74 @@
...
+    // Compliance rejection
+    REQUESTED_TO_REJECTED            (REQUESTED,            REJECTED,             CertificateEvent.UPDATE_STATE),
+
+    // Issue / renew / rekey failure (v2 handleFailedOrRejectedEvent)
+    REQUESTED_TO_FAILED              (REQUESTED,            FAILED,               CertificateEvent.ISSUE),
+    PENDING_APPROVAL_TO_FAILED       (PENDING_APPROVAL,     FAILED,               CertificateEvent.ISSUE),
+
+    // Approval-rejected transition for certs that were ISSUED before the approval was raised
+    // (v2 issueCertificateRejectedAction; covers the ISSUED→REJECTED path in tests)
+    ISSUED_TO_REJECTED               (ISSUED,               REJECTED,             CertificateEvent.UPDATE_STATE),
```
ivosh: CertificateStateTransition.java REQUESTED->FAILED, PENDING_APPROVAL->FAILED, and ISSUED->REJECTED are present in the enum but no test references them. allCriticalTransitionsArePresent (CertificateStateTransitionTest.java:25-54) asserts 18 rows; these three are omitted. The comment claims ISSUED->REJECTED 'covers the ISSUED->REJECTED path in tests' but it does not.

src/test/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachineTest.java
```diff
@@ -0,0 +1,82 @@
...
+        sm.transition(cert, CertificateState.PENDING_ISSUE);
+
+        assertEquals(CertificateState.PENDING_ISSUE, cert.getState());
+        verify(certificateRepository).save(cert);
+        verify(eventHistoryService).addEventHistory(eq(cert.getUuid()), eq(CertificateEvent.ISSUE),
+            eq(CertificateEventStatus.SUCCESS), anyString(), eq(""));
+    }
+
+    @Test
+    void invalidTransitionThrowsInvalidTransitionException() {
```
ivosh, line 46: invalidTransitionThrowsInvalidTransitionException verifies save and addEventHistory are never called, but never asserts cert.getState() is still REVOKED after the throw.

src/test/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateTransitionTest.java
```diff
@@ -0,0 +1,56 @@
...
+            CertificateStateTransition.lookup(CertificateState.REQUESTED, CertificateState.PENDING_ISSUE);
+        assertTrue(row.isPresent());
+        assertEquals(CertificateEvent.ISSUE, row.get().defaultAuditEvent);
+    }
+
+    @Test
+    void lookupReturnsEmptyForUnknownTransition() {
+        Optional<CertificateStateTransition> row =
+            CertificateStateTransition.lookup(CertificateState.REVOKED, CertificateState.ISSUED);
+        assertFalse(row.isPresent());
```
ivosh, line 27: allCriticalTransitionsArePresent (CertificateStateTransitionTest.java:25-54) asserts only .isPresent() for 18 rows; only lookupReturnsRowForKnownTransition (line 12-17) checks a single defaultAuditEvent (ISSUE for REQUESTED→PENDING_ISSUE). The audit-event column drives which CertificateEvent gets written to history (CertificateStateMachine.java:84), so a wrong mapping — e.g. an APPROVAL_CLOSE row mislabeled REVOKE — would pass this suite. Consider parameterizing the presence assertions to also check the expected defaultAuditEvent, turning 18 weak presence checks into 18 meaningful ones.

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateTransition.java
```diff
@@ -0,0 +1,74 @@
...
+    // Issue approval flow
+    REQUESTED_TO_PENDING_APPROVAL    (REQUESTED,            PENDING_APPROVAL,     CertificateEvent.APPROVAL_REQUEST),
+    PENDING_APPROVAL_TO_PENDING_ISSUE(PENDING_APPROVAL,     PENDING_ISSUE,        CertificateEvent.APPROVAL_CLOSE),
+    PENDING_APPROVAL_TO_REJECTED     (PENDING_APPROVAL,     REJECTED,             CertificateEvent.APPROVAL_CLOSE),
+
+    // Revoke (sync + async)
+    ISSUED_TO_PENDING_REVOKE         (ISSUED,               PENDING_REVOKE,       CertificateEvent.REVOKE),
+    PENDING_REVOKE_TO_REVOKED        (PENDING_REVOKE,       REVOKED,              CertificateEvent.REVOKE),
+    PENDING_REVOKE_TO_ISSUED         (PENDING_REVOKE,       ISSUED,               CertificateEvent.REVOKE),
+    ISSUED_TO_REVOKED                (ISSUED,               REVOKED,              CertificateEvent.REVOKE),
```
ivosh: The public doc's Pending Issue description reads 'when the certification authority accepts an issue OR RENEW request but cannot complete it synchronously'. A renew acts on an already-Issued certificate, which would imply an ISSUED -> PENDING_ISSUE transition — but the table has no such row. If async renewal transitions the existing certificate in place, transition(issuedCert, PENDING_ISSUE) would throw InvalidTransitionException. Question: is renewal modelled as minting a NEW certificate (no edge needed, and this is correct as-is) or as an in-place transition of the existing one (edge missing)? Worth confirming before the activation slice (#1603) routes renew through the SM. Non-blocking.

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachine.java
```diff
@@ -0,0 +1,91 @@
...
+    @Transactional
+    public void transition(Certificate cert, CertificateState toState,
+                           @Nullable CertificateEvent auditEventOverride,
+                           @Nullable String reasonMessage) {
+        applyTransition(cert, toState, auditEventOverride, reasonMessage);
+    }
+
+    /**
+     * Shared logic for both public overloads. Private so the {@code @Transactional} entry points
+     * are reached through the Spring proxy rather than by self-invocation; it runs within whichever
```
ivosh: This is a normal code flow, not something exotic. I recommend trimming this down or the reader remains perplexed why documeting a normal behaviour?

ivosh (changes requested):
Thanks for the state machine!
Do not forget to update the documentation with the new transitions: 
https://docs.otilm.com/docs/certificate-key/concept-design/core-components/certificate#certificate-state

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachine.java
lubomirw: Clarified in 642ca0b4a — reworded to "the existing v2 client-operations service's failure handling" (the SM is the new path that absorbs it).

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachine.java
lubomirw: Done in 642ca0b4a — the 2-arg overload's javadoc now states both the audit event and the reason message are auto-derived.

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachine.java
lubomirw: Good catch — fixed in 642ca0b4a, and it went deeper than a toState rule: `PENDING_REVOKE → ISSUED` (revoke failed/cancelled restore) lands on ISSUED but must audit FAILED, so the destination state alone is insufficient. Each transition row now carries a `defaultStatus` (parallel to `defaultAuditEvent`); the SM writes `row.defaultStatus`. FAILED for the failure/rejection/failed-restore rows (incl. PENDING_REVOKE→ISSUED and PENDING_APPROVAL→ISSUED), SUCCESS otherwise. Note: REJECTED rows are uniformly FAILED (controlled failure of the cert operation); the reason message carries the wording. Covered by new tests.

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachine.java
lubomirw: By design — the SM is intentionally lock-free; single-cert concurrency is the caller's responsibility. Operator paths and the async poll listener acquire a pessimistic lock (findAndLockWithAssociationsByUuid) and re-assert state before calling transition(), so this read-modify-write runs under that lock. Added a class-javadoc note in 642ca0b4a; the locked call sites land in the activation slice (#1603).

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateTransition.java
lubomirw: Fixed in 642ca0b4a — REQUESTED→FAILED, PENDING_APPROVAL→FAILED and ISSUED→REJECTED are now asserted (all 21 rows are), and the misleading "covers the ISSUED→REJECTED path in tests" clause was removed from the row comment.

src/test/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachineTest.java
lubomirw: Done in 642ca0b4a — the test now asserts cert.getState() is still REVOKED after the throw.

src/test/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateTransitionTest.java
lubomirw: Done in 642ca0b4a — replaced the presence-only checks with `everyTransitionHasExpectedAuditEventAndStatus`, which asserts the expected defaultAuditEvent AND defaultStatus for all 21 rows via an assertRow(...) helper. A mislabeled event or status now fails the suite.

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateTransition.java
lubomirw: Confirmed: renewal mints a NEW successor certificate (REQUESTED→PENDING_ISSUE→ISSUED); there is no in-place ISSUED→PENDING_ISSUE transition. The adapter signature renew(oldCert, newCert) reflects this — newCert is the successor. So correct as-is, no edge needed. (The public doc wording will be reconciled per your review-body note.)

src/main/java/com/otilm/core/service/handler/authority/lifecycle/CertificateStateMachine.java
lubomirw: Trimmed in 642ca0b4a — the applyTransition javadoc is now a single line about proxy-reachability.

lubomirw:
Thanks for the review! On the docs: agreed — the new `PENDING_REGISTRATION`/`REGISTERED` states and the register/async transitions will be added to the certificate state-design page. Since the new states only become operator-visible when the **activation slice (#1603)** lands, I'll raise the docs update on the `documentation` repo to merge alongside it (tracked so it isn't lost). All inline comments addressed in 642ca0b4a; replied per-thread.

lubomirw:
Docs follow-up is now tracked as documentation#307 (sub-issue of #110), to merge alongside the activation slice (#1603).

ivosh (dismissed):

ivosh (approved):

PR #1607, opened 2026-06-15
Add authority-provider adapter layer and version-dispatch factory

### LINKED ISSUE #1600: Add authority-provider adapter layer and version-dispatch factory
### Description

Introduce a version-dispatch adapter layer so the v2 client-operation service no longer calls the connector API client directly:

- `AuthorityProviderAdapter` interface + `AbstractAuthorityProviderAdapter` base
- `AuthorityProviderV2Adapter` (wraps the existing v2 path) and `AuthorityProviderV3Adapter`
- `AuthorityProviderAdapterFactory` — dispatch by connector interface version
- capability interfaces `RegisterCapability`, `AsyncOperationCapability`
- result types: `AdapterOperationResult` / `AdapterOperationOutcome`, `CancelResult` / `CancelOutcome`, `StatusPollResult`, `CertificateOperation`; plus `ConnectorOperationErrorCodes`, `UnsupportedAuthorityVersionException`, `ConnectorAcceptedButLocalFailureException`

Route the v2 `ClientOperationServiceImpl` call sites through the factory. The V2 path is behavior-preserving; the V3 adapter is present but inert (no v3 authority can exist until the activation work), so it is safe to merge ahead of activation. Largest slice and the fan-out bottleneck for the rest of the work.

**Depends on:** #1599
**Review focus:** `V2Adapter` equals the legacy call path bit-for-bit; factory falls back to v2 on a null connector interface; `AbstractAuthorityProviderAdapter` extraction.

### Definition of Done

- [ ] Adapter interface + abstract base + V2/V3 adapters + factory + capability interfaces + result records added with tests
- [ ] v2 `ClientOperationServiceImpl` call sites routed through the factory
- [ ] `V2Adapter` verified equivalent to the legacy connector call path (behavior-preserving)
- [ ] Factory routes a null connector interface to the v2 adapter
- [ ] PR merged to `main`

lubomirw:
Second slice of landing `feat/authority-v3` on `main` as a reviewable stacked-PR chain. Part of #1600 (epic #110). **Stacked on #1605 (S1)** — review that first; this PR's base retargets to `main` automatically once S1 merges.

## What

Introduces the version-dispatch adapter layer so the v2 client-operation service no longer calls the connector API client directly:

- `AuthorityProviderAdapter` + `AbstractAuthorityProviderAdapter` + `AuthorityProviderV2Adapter` / `AuthorityProviderV3Adapter` + `AuthorityProviderAdapterFactory`
- capability interfaces `RegisterCapability`, `AsyncOperationCapability`
- result types `AdapterOperationResult`/`AdapterOperationOutcome`, `CancelResult`/`CancelOutcome`, `StatusPollResult`, `CertificateOperation`, `ConnectorOperationErrorCodes`
- exceptions `UnsupportedAuthorityVersionException`, `ConnectorAcceptedButLocalFailureException`
- v3 connector client beans (`ApplicationConfig`, `ProxyClientConfig`, `ConnectorApiFactory`) — REST + MQ
- `AuthorityInstanceReference.connector_interface_uuid` FK + Flyway migration (deterministic, idempotent backfill)
- adapter unit tests

## Notes for review

- **Purely additive / dark merge.** Nothing calls the adapters yet — routing the v2 service through the factory lands in the activation slice (#1603). Compiles against `main` + the state-machine slice; the v3 client beans reference only the released interfaces lib. 39 adapter tests pass.
- Code originates from `feat/authority-v3` (already reviewed there); re-sliced for incremental review.

## Pre-review addressed (guided-pr-review + Copilot — both SHIP after fixes)

- **[MEDIUM] V2 `renew()` NPE on reuse-key (no-CSR) renewal** — `AuthorityProviderV2Adapter.renew` accessed `newCert.getCertificateRequest()` unconditionally where V3 guards it. Fixed (omit request/format when absent) + regression test `renew_withoutCsr_omitsRequestInsteadOfNpe`.
- **Factory null-version** — `forAuthority` now throws `UnsupportedAuthorityVersionException` for a null version instead of NPE, consistent with its Javadoc.
- **[LOW] "register() doesn't replay meta" — dismissed (false positive).** The v3 register DTO (`CertificateRegistrationRequestDtoV3`) has no `meta` field: register is the originating op that *produces* the tracking meta (returned in its 202), so there is nothing to replay on the request.

## Known items (intentional, not changed here)

- The migration FK uses `ON DELETE RESTRICT`; once backfilled it gates deletion of a referenced `connector_interface` row. Confirm connector-interface teardown paths before later slices rely on the FK.
- The migration version (`V202605251200…`) is back-dated relative to June migrations on the branch; safe because `spring.flyway.out-of-order` is enabled.

## Verification

`mvn test` across the adapter suite: **39 tests, 0 failures**.

### DIFF
--- a/src/main/java/com/otilm/core/service/handler/authority/AsyncOperationCapability.java
+++ b/src/main/java/com/otilm/core/service/handler/authority/AsyncOperationCapability.java
@@ -1,0 +1,14 @@
+package com.otilm.core.service.handler.authority;
+
+import com.otilm.api.exception.ConnectorException;
+import com.otilm.core.dao.entity.Certificate;
+
+/**
+ * v3-only capability: poll status and cancel in-flight async operations.
+ */
+public interface AsyncOperationCapability {
+
+    StatusPollResult pollStatus(Certificate cert, CertificateOperation op) throws ConnectorException;
+
+    CancelResult cancel(Certificate cert, CertificateOperation op) throws ConnectorException;
+}
--- a/src/main/java/com/otilm/core/service/handler/authority/AsyncOperationCapability.java
+++ b/src/main/java/com/otilm/core/service/handler/authority/AsyncOperationCapability.java
@@ -8,7 +8,19 @@ import com.otilm.core.dao.entity.Certificate;
  */
 public interface AsyncOperationCapability {
 
+    /**
+     * Polls the connector for the status of an async operation. The {@link StatusPollResult} carries
+     * the operation status (in-progress / completed / failed); the caller maps it onto the certificate
+     * state per the operation's transition matrix.
+     */
     StatusPollResult pollStatus(Certificate cert, CertificateOperation op) throws ConnectorException;
 
+    /**
+     * Requests cancellation of an in-flight async operation. The {@link CancelResult}'s
+     * {@link CancelOutcome} drives the caller's rollback decision: {@code CANCELLED} and
+     * {@code NOT_TRACKED} are both soft successes — the operation is no longer running, so the caller
+     * rolls local state back — while {@code REFUSED_PAST_POINT_OF_NO_RETURN} means the connector could
+     * not cancel and local state must be left unchanged.
+     */
     CancelResult cancel(Certificate cert, CertificateOperation op) throws ConnectorException;
 }
--- a/src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapter.java
+++ b/src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapter.java
@@ -19,6 +19,17 @@ import java.util.List;
  */
 public interface AuthorityProviderAdapter {
 
+    /**
+     * Issues a certificate through the authority provider.
+     *
+     * @param cert the certificate to issue — provides the CSR ({@code getCertificateRequest()}) and
+     *             the RA-profile association used to resolve connector attributes.
+     * @param req  the operator-level sign request DTO. Currently informational at the adapter layer:
+     *             the wire body is reconstructed from the persisted certificate plus the attribute
+     *             engine (Core persists the operator's CSR + request attributes before the adapter
+     *             runs, then the adapter re-reads them). Kept for contract symmetry with
+     *             {@code renew}/{@code revoke}/{@code register}, which do consume their DTOs.
+     */
     AdapterOperationResult issue(Certificate cert, ClientCertificateSignRequestDto req) throws ConnectorException;
 
     /**
@@ -29,8 +40,10 @@ public interface AuthorityProviderAdapter {
      * @param newCert the successor certificate — provides the new CSR
      *                ({@code getCertificateRequest().getContent()} / {@code .getCertificateRequestFormat()})
      *                and the RA profile association used to look up connector attributes.
-     * @param req     the operator-level renew request DTO (carries client-visible fields such as
-     *                {@code replaceInLocations}; wire-level fields are derived from oldCert/newCert).
+     * @param req     the operator-level renew request DTO. Not consumed by the adapter: client-visible
+     *                fields (e.g. {@code replaceInLocations}) are handled by the service layer and the
+     *                wire body is derived from oldCert/newCert. Kept for contract symmetry with the
+     *                other operations.
      */
     AdapterOperationResult renew(Certificate oldCert, Certificate newCert, ClientCertificateRenewRequestDto req) throws ConnectorException;
 
--- a/src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapter.java
+++ b/src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapter.java
@@ -1,0 +1,70 @@
+package com.otilm.core.service.handler.authority;
+
+import com.otilm.api.exception.ConnectorException;
+import com.otilm.api.exception.ValidationException;
+import com.otilm.api.model.client.attribute.RequestAttribute;
+import com.otilm.api.model.common.attribute.common.BaseAttribute;
+import com.otilm.api.model.core.v2.ClientCertificateRenewRequestDto;
+import com.otilm.api.model.core.v2.ClientCertificateRevocationDto;
+import com.otilm.api.model.core.v2.ClientCertificateSignRequestDto;
+import com.otilm.core.dao.entity.AuthorityInstanceReference;
+import com.otilm.core.dao.entity.Certificate;
+import com.otilm.core.dao.entity.RaProfile;
+
+import java.util.List;
+
+/**
+ * Common authority-provider operations supported by every interface version (v1 not implemented).
+ * Methods take Core's operator DTOs; adapter translates to version-specific wire DTOs internally.
+ */
+public interface AuthorityProviderAdapter {
+
+    AdapterOperationResult issue(Certificate cert, ClientCertificateSignRequestDto req) throws ConnectorException;
+
+    /**
+     * Renews (or rekeys) a certificate through the authority provider.
+     *
+     * @param oldCert the predecessor certificate — provides the old cert content
+     *                ({@code getCertificateContent().getContent()}) and its connector-issued metadata.
+     * @param newCert the successor certificate — provides the new CSR
+     *                ({@code getCertificateRequest().getContent()} / {@code .getCertificateRequestFormat()})
+     *                and the RA profile association used to look up connector attributes.
+     * @param req     the operator-level renew request DTO (carries client-visible fields such as
+     *                {@code replaceInLocations}; wire-level fields are derived from oldCert/newCert).
+     */
+    AdapterOperationResult renew(Certificate oldCert, Certificate newCert, ClientCertificateRenewRequestDto req) throws ConnectorException;
+
+    AdapterOperationResult revoke(Certificate cert, ClientCertificateRevocationDto req) throws ConnectorException;
+
+    /**
+     * Authority-instance attribute definitions. v2 lists via the function-group attribute endpoint
+     * (`/v1/authorityProvider/{kind}/attributes`); v3 via the stateless
+     * `/v3/authorityProvider/authorities/attributes`. Used by authority create/edit so the operator
+     * input is validated against the version-correct schema (legacy connectors do NOT use this —
+     * they stay on the function-group path in ConnectorService).
+     */
+    List<BaseAttribute> listAuthorityInstanceAttributes(AuthorityInstanceReference authority) throws ConnectorException;
+
+    /**
+     * Dynamic issue-attribute schema scoped to a specific RA profile. v3 carries both
+     * {@code authorityAttributes} (auth/identity to the upstream CA) and {@code raProfileAttributes}
+     * (which profile/template determines the schema). v2 ignores {@code raProfile} — its endpoint
+     * is keyed by {@code authorityInstanceUuid} alone and returns a single per-authority schema.
+     */
+    List<BaseAttribute> listIssueAttributes(AuthorityInstanceReference authority, RaProfile raProfile) throws ConnectorException;
+
+    /** See {@link #listIssueAttributes} — identical semantics for revoke. */
+    List<BaseAttribute> listRevokeAttributes(AuthorityInstanceReference authority, RaProfile raProfile) throws ConnectorException;
+
+    /**
+     * Connector-side validation of operator-supplied issue attributes. v2 calls the connector's
+     * {@code /validate} endpoint; v3 is a no-op (the v3 contract dropped {@code /validate} — Core
+     * validates structurally against the listed definitions instead).
+     */
+    void validateIssueAttributes(AuthorityInstanceReference authority, List<RequestAttribute> attributes) throws ValidationException, ConnectorException;
+
+    /** Connector-side validation of operator-supplied revoke attributes. See {@link #validateIssueAttributes}. */
+    void validateRevokeAttributes(AuthorityInstanceReference authority, List<RequestAttribute> attributes) throws ValidationException, ConnectorException;
+
+    void checkAuthorityConnection(AuthorityInstanceReference authority, List<RequestAttribute> attributes) throws ValidationException, ConnectorException;
+}
--- a/src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapterFactory.java
+++ b/src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapterFactory.java
@@ -1,0 +1,61 @@
+package com.otilm.core.service.handler.authority;
+
+import com.otilm.core.dao.entity.AuthorityInstanceReference;
+import com.otilm.core.dao.entity.ConnectorInterfaceEntity;
+import com.otilm.core.exception.UnsupportedAuthorityVersionException;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * Dispatches authority operations to the adapter matching the authority's connector interface
+ * version. v2 → {@link AuthorityProviderV2Adapter}, v3 → {@link AuthorityProviderV3Adapter}.
+ *
+ * <p>Defensive — legacy v1-authority connectors ({@code FunctionGroupCode.LEGACY_AUTHORITY_PROVIDER})
+ * flow through the separate legacy service path and never reach this factory.</p>
+ *
+ * <p>{@link ConnectorInterfaceEntity#getVersion()} returns the version string as reported by the
+ * connector's info endpoint (e.g. {@code "v2"}, {@code "v3"}). A {@code null} interface is treated
+ * as v2: framework-v1 connectors that speak the v2 authority wire protocol (e.g. ejbca-ng) declare
+ * no interface row, so they route to {@link AuthorityProviderV2Adapter} (see {@link #forAuthority}).
+ * Any other non-null value (an unrecognized or bare-decimal version such as {@code "2"}) results in
+ * an {@link UnsupportedAuthorityVersionException}.</p>
+ */
+@Component
+public class AuthorityProviderAdapterFactory {
+
+    private final AuthorityProviderV2Adapter v2Adapter;
+    private final AuthorityProviderV3Adapter v3Adapter;
+
+    @Autowired
+    public AuthorityProviderAdapterFactory(AuthorityProviderV2Adapter v2Adapter,
+                                           AuthorityProviderV3Adapter v3Adapter) {
+        this.v2Adapter = v2Adapter;
+        this.v3Adapter = v3Adapter;
+    }
+
+    public AuthorityProviderAdapter forAuthority(AuthorityInstanceReference authority) {
+        ConnectorInterfaceEntity iface = authority.getConnectorInterface();
+        if (iface == null) {
+            // Connector declared no AUTHORITY interface row in connector_interface. This is
+            // expected for framework-v1 connectors that implement the v2 authority wire
+            // protocol (e.g. ejbca-ng) — their /v1/info endpoint predates the v2 interface
+            // descriptor and the M0 backfill cannot create a row for them. Route to v2
+            // since that is the wire protocol they speak. Legacy v1-authority connectors
+            // (FunctionGroupCode.LEGACY_AUTHORITY_PROVIDER) never reach this factory; they
+            // route through the separate legacy service path by design.
+            return v2Adapter;
+        }
+        String version = iface.getVersion();
+        if (version == null) {
+            throw new UnsupportedAuthorityVersionException(
+                    "Authority connector interface has no version (authority " + authority.getUuid() + ")");
+        }
+        return switch (version) {
+            case "v2" -> v2Adapter;
+            case "v3" -> v3Adapter;
+            default -> throw new UnsupportedAuthorityVersionException(
+                    "Unsupported authority connector interface version: " + version
+                            + " (authority " + authority.getUuid() + ")");
+        };
+    }
+}
--- a/src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapterFactory.java
+++ b/src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapterFactory.java
@@ -36,13 +36,7 @@ public class AuthorityProviderAdapterFactory {
     public AuthorityProviderAdapter forAuthority(AuthorityInstanceReference authority) {
         ConnectorInterfaceEntity iface = authority.getConnectorInterface();
         if (iface == null) {
-            // Connector declared no AUTHORITY interface row in connector_interface. This is
-            // expected for framework-v1 connectors that implement the v2 authority wire
-            // protocol (e.g. ejbca-ng) — their /v1/info endpoint predates the v2 interface
-            // descriptor and the M0 backfill cannot create a row for them. Route to v2
-            // since that is the wire protocol they speak. Legacy v1-authority connectors
-            // (FunctionGroupCode.LEGACY_AUTHORITY_PROVIDER) never reach this factory; they
-            // route through the separate legacy service path by design.
+            // No interface row → a framework-v1 connector speaking the v2 authority protocol (see class Javadoc).
             return v2Adapter;
         }
         String version = iface.getVersion();
--- a/src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
+++ b/src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
@@ -24,6 +24,7 @@ import com.otilm.api.model.core.v2.ClientCertificateRenewRequestDto;
 import com.otilm.api.model.core.v2.ClientCertificateRevocationDto;
 import com.otilm.api.model.core.v2.ClientCertificateSignRequestDto;
 import com.otilm.core.attribute.engine.AttributeEngine;
+import com.otilm.core.attribute.engine.AttributeOperation;
 import com.otilm.core.attribute.engine.records.ObjectAttributeContentInfo;
 import com.otilm.core.client.ConnectorApiFactory;
 import com.otilm.core.dao.entity.AuthorityInstanceReference;
@@ -294,6 +295,7 @@ public class AuthorityProviderV3Adapter
         return attributeEngine.getRequestObjectDataAttributesContent(
                 ObjectAttributeContentInfo.builder(Resource.CERTIFICATE, cert.getUuid())
                         .connector(authority.getConnectorUuid())
+                        .operation(AttributeOperation.CERTIFICATE_ISSUE)
                         .build());
     }
 
--- a/src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
+++ b/src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
@@ -1,0 +1,353 @@
+package com.otilm.core.service.handler.authority;
+
+import com.otilm.api.clients.ApiClientConnectorInfo;
+import com.otilm.api.exception.ConnectorException;
+import com.otilm.api.exception.ConnectorProblemException;
+import com.otilm.api.exception.ValidationException;
+import com.otilm.api.interfaces.client.v3.AuthoritySyncApiClient;
+import com.otilm.api.interfaces.client.v3.CertificateSyncApiClient;
+import com.otilm.api.model.client.attribute.RequestAttribute;
+import com.otilm.api.model.common.attribute.common.BaseAttribute;
+import com.otilm.api.model.common.error.ErrorCode;
+import com.otilm.api.model.connector.v3.certificate.CertificateAttributeListRequestDtoV3;
+import com.otilm.api.model.connector.v3.certificate.CertificateDataResponseDto;
+import com.otilm.api.model.connector.v3.certificate.CertificateOperationCancelRequestDtoV3;
+import com.otilm.api.model.connector.v3.certificate.CertificateOperationStatusRequestDtoV3;
+import com.otilm.api.model.connector.v3.certificate.CertificateOperationStatusResponseDto;
+import com.otilm.api.model.connector.v3.certificate.CertificateRegistrationRequestDtoV3;
+import com.otilm.api.model.connector.v3.certificate.CertificateRenewRequestDtoV3;
+import com.otilm.api.model.connector.v3.certificate.CertificateRevocationRequestDtoV3;
+import com.otilm.api.model.connector.v3.certificate.CertificateSignRequestDtoV3;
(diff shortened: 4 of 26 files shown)


src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -0,0 +1,353 @@
...
+            throw e;
+        }
+    }
+
+    private List<RequestAttribute> issueAttributesFor(Certificate cert, AuthorityInstanceReference authority) {
+        return attributeEngine.getRequestObjectDataAttributesContent(
+                ObjectAttributeContentInfo.builder(Resource.CERTIFICATE, cert.getUuid())
+                        .connector(authority.getConnectorUuid())
+                        .build());
+    }
```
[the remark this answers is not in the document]
lubomirw: Fixed in 53bbc871d — `issueAttributesFor` now passes `.operation(AttributeOperation.CERTIFICATE_ISSUE)`, matching V2Adapter so the engine returns issue-scoped attributes. Verified V3 renew/revoke take caller-supplied `req.getAttributes()` (no engine load), so this was the only gap. Added regression test `issueLoadsCertificateRequestAttributesScopedToIssueOperation`.

src/main/resources/db/migration/V202605251200__add_connector_interface_uuid_to_authority_instance_reference.sql
```diff
@@ -0,0 +1,30 @@
...
+-- correct for the framework-v1 + authority-v2 case. Legacy v1-authority connectors flow
+-- through a separate service path that never consults this column.
+UPDATE authority_instance_reference air
+SET connector_interface_uuid = (
+    SELECT ci.uuid
+    FROM connector_interface ci
+    WHERE ci.connector_uuid = air.connector_uuid
+      AND ci.interface = 'AUTHORITY'
+    ORDER BY ci.version DESC
+    LIMIT 1
```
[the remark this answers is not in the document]
lubomirw: Fixed in 53bbc871d — the backfill now prefers v2 explicitly: `ORDER BY CASE WHEN ci.version = 'v2' THEN 0 ELSE 1 END, ci.version`. Every backfilled authority predates v3 (v3 authorities set the FK explicitly and are excluded by the WHERE), so resolving them to v2 is correct regardless of which connector versions happen to be registered at migration time — no longer relying on deployment timing.

src/main/resources/db/migration/V202605251200__add_connector_interface_uuid_to_authority_instance_reference.sql
```diff
@@ -0,0 +1,30 @@
...
+    ADD COLUMN connector_interface_uuid UUID
+    REFERENCES connector_interface(uuid) ON UPDATE CASCADE ON DELETE RESTRICT;
+
+-- Backfill: link each authority instance to its connector's AUTHORITY interface row.
+--
+-- A connector may expose multiple AUTHORITY interface rows when it advertises both v2
+-- and v3 side-by-side during migration. Pick the highest version deterministically so
+-- the UPDATE never aborts with "more than one row returned by a subquery". Existing
+-- authorities are pre-v3 so this resolves to v2 on coexisting connectors. v3 authorities
+-- created via the service from M2 onward set the FK explicitly and bypass this backfill.
```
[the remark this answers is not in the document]
lubomirw: Fixed in 53bbc871d — reworded the comment to state the actual rule: backfilled (pre-v3) authorities resolve to the v2 AUTHORITY interface; v3 authorities set the FK explicitly via the service and are excluded from the backfill.

src/test/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapterFactoryTest.java
```diff
@@ -0,0 +1,64 @@
...
+    void missingInterfaceFallsBackToV2() {
+        // A null connector interface is expected for framework-v1 connectors that implement the
+        // v2 authority wire protocol (e.g. ejbca-ng) — they declare no connector_interface row,
+        // so the M0 backfill leaves the FK null. The factory routes these to the v2 adapter
+        // (the protocol they speak). Legacy v1-authority connectors never reach this factory.
+        AuthorityInstanceReference auth = new AuthorityInstanceReference();
+        auth.setUuid(UUID.randomUUID());
+        // connector interface left null
+        assertSame(v2, factory.forAuthority(auth));
+    }
```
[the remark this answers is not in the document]
lubomirw: Added in 16e04e4a7 — `rejectsNullVersion` asserts `forAuthority` throws `UnsupportedAuthorityVersionException` for a non-null interface with a null version (distinct from the null-interface→v2 fallback). Closes the coverage gap from the null-version guard.

src/test/java/com/otilm/core/service/handler/authority/AuthorityProviderV3AdapterTest.java
```diff
@@ -0,0 +1,333 @@
...
+    void issueWrapsConnectorAcceptedLocalFailure() throws ConnectorException {
+        // Simulate connector returns 200 but body is null -> mapIssueRenewRegisterResponse throws NPE
+        // We can't make mapIssueRenewRegisterResponse throw internally easily without reflections,
+        // but we can verify the guard: if the client throws RuntimeException after 200 is observed,
+        // the adapter wraps it. Use a spy to inject the failure after acceptance.
+
+        // Direct test: issue throws RuntimeException from within try after connectorAccepted = true
+        // We simulate this by making the certClient return a response that causes a NPE in the map step.
+        // The simplest way: return a ResponseEntity whose getStatusCode() returns 200 but body.getMeta()
+        // throws — that happens in mapIssueRenewRegisterResponse after connectorAccepted = true.
```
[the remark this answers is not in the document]
lubomirw: Fixed in 16e04e4a7 — replaced the misleading/verbose comment. The test makes the 200 response body throw on `getCertificateData()` during the mapping step (after connectorAccepted=true), so the adapter wraps it in `ConnectorAcceptedButLocalFailureException`. Comment now describes exactly that.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapter.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.List;
+
+/**
+ * Common authority-provider operations supported by every interface version (v1 not implemented).
+ * Methods take Core's operator DTOs; adapter translates to version-specific wire DTOs internally.
+ */
+public interface AuthorityProviderAdapter {
+
+    AdapterOperationResult issue(Certificate cert, ClientCertificateSignRequestDto req) throws ConnectorException;
```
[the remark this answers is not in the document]
lubomirw: Keeping this parameter (won't-fix). The finding is accurate — `req` is unused by `issue`/`renew` today — but removal isn't the right call: (1) the adapter derives the wire body from the persisted certificate + attribute engine (Core persists the operator's CSR + request attributes and commits *before* the adapter runs, then re-reads them), so `req` is informational here by architecture, not oversight; (2) it keeps the interface symmetric with `revoke`/`register`, which do consume their DTOs; (3) removing it is a contract change rippling to the not-yet-merged activation slice's callers. Documented the WHY in a `@param` Javadoc on both methods in 462bbbb25 so the intent is explicit. Note: SonarCloud's java:S1172 exempts interface declarations and interface-implementation methods, so this is advisory-only and does not affect the quality gate.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapter.java
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @param oldCert the predecessor certificate — provides the old cert content
+     *                ({@code getCertificateContent().getContent()}) and its connector-issued metadata.
+     * @param newCert the successor certificate — provides the new CSR
+     *                ({@code getCertificateRequest().getContent()} / {@code .getCertificateRequestFormat()})
+     *                and the RA profile association used to look up connector attributes.
+     * @param req     the operator-level renew request DTO (carries client-visible fields such as
+     *                {@code replaceInLocations}; wire-level fields are derived from oldCert/newCert).
+     */
+    AdapterOperationResult renew(Certificate oldCert, Certificate newCert, ClientCertificateRenewRequestDto req) throws ConnectorException;
```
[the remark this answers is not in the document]
lubomirw: Keeping this parameter (won't-fix). The finding is accurate — `req` is unused by `issue`/`renew` today — but removal isn't the right call: (1) the adapter derives the wire body from the persisted certificate + attribute engine (Core persists the operator's CSR + request attributes and commits *before* the adapter runs, then re-reads them), so `req` is informational here by architecture, not oversight; (2) it keeps the interface symmetric with `revoke`/`register`, which do consume their DTOs; (3) removing it is a contract change rippling to the not-yet-merged activation slice's callers. Documented the WHY in a `@param` Javadoc on both methods in 462bbbb25 so the intent is explicit. Note: SonarCloud's java:S1172 exempts interface declarations and interface-implementation methods, so this is advisory-only and does not affect the quality gate.

lubomirw:
A fresh guided-review pass on the rebased head (after the earlier Copilot rounds) caught a **revoke-reason parity bug** the mock-based tests had hidden: `ClientCertificateRevocationDto.reason` is optional (no `@NotNull`), but the v3 wire DTO `CertificateRevocationRequestDtoV3.reason` is `@NotNull`. V2 defaults a null reason to `UNSPECIFIED` (matching the legacy service); V3 passed `null` through → validation failure on identical input. Fixed in 660856e69 to mirror V2, with a regression test (`revokeDefaultsNullReasonToUnspecified`).

One MINOR point reviewed and intentionally **kept**: in `executeWithAcceptanceGuard`, an unexpected-but-2xx status (e.g. 201/206 — a connector contract violation) sets `connectorAccepted=true`, so the mapper's `IllegalStateException` is wrapped as `ConnectorAcceptedButLocalFailureException` (i.e. don't roll back). This is deliberate: on an ambiguous success we must not risk rolling back work the connector may have performed (the worse divergence). Flagging here rather than commenting the code.

src/main/java/com/otilm/core/service/handler/authority/CertificateOperation.java
```diff
@@ -0,0 +1,39 @@
...
+import com.otilm.api.model.core.certificate.CertificateState;
+
+public enum CertificateOperation {
+    ISSUE,
+    RENEW,
+    REVOKE,
+    REGISTER;
+
+    /** The PENDING state a certificate sits in while this operation is async-in-flight. */
+    public CertificateState pendingState() {
```
ivosh, line 12: CertificateOperation.pendingState() (line 12), terminalSuccessState() (line 21), and terminalFailureState() (line 33) encode the per-operation certificate state-transition matrix. None of the four new test files exercise them: AdapterOperationResultTest.certificateOperationAllValues only asserts the enum constants are non-null. terminalFailureState() deliberately maps REVOKE->ISSUED (a failed/cancelled revoke leaves the cert valid, NOT FAILED). A regression that flipped REVOKE failure to FAILED would mark valid certificates as failed with no test to catch it. Add a parameterized test asserting the full matrix.

src/test/java/com/otilm/core/service/handler/authority/AdapterOperationResultTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Test
+    void syncNoContentFactory() {
+        AdapterOperationResult r = AdapterOperationResult.syncNoContent();
+        assertEquals(AdapterOperationOutcome.SYNC_NO_CONTENT, r.outcome());
+        assertFalse(r.isAsync());
+    }
+
+    @Test
+    void cancelOutcomeAllValues() {
```
ivosh: cancelOutcomeAllValues (line 36-40) and certificateOperationAllValues (line 42-48) only assert that enum constants are non-null. Enum constants are non-null by language guarantee, so these tests are green-by-construction and cover no behavior. Either delete them or replace them with tests that exercise the logic keyed off these enums. Verify the test bodies actually only assert non-null.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -0,0 +1,356 @@
...
+        AuthorityInstanceReference authority = raProfile.getAuthorityInstanceReference();
+        ApiClientConnectorInfo connectorDto = connectorForApiClient(authority);
+
+        CertificateOperationStatusRequestDtoV3 wire = new CertificateOperationStatusRequestDtoV3();
+        wire.setMeta(loadMeta(cert, authority));
+        wire.setAuthorityAttributes(authorityAttributesFor(authority));
+        wire.setRaProfileAttributes(raProfileAttributesFor(raProfile, authority));
+
+        CertificateSyncApiClient v3Client = connectorApiFactory.getCertificateApiClientV3(connectorDto);
+        CertificateOperationStatusResponseDto resp = switch (op) {
```
ivosh, line 215: AuthorityProviderV3Adapter.pollStatus (switch at line 223-227) and cancel (switch at line 244-248) route to different connector endpoints per CertificateOperation. The tests only cover pollStatus for ISSUE and cancel for ISSUE and REGISTER. The REVOKE/RENEW branches are never asserted. Verify the actual test coverage and switch branches in the worktree.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -0,0 +1,356 @@
...
+    /**
+     * Issues 3 separate {@code attribute_content_2_object} reads (cert meta + authority
+     * attrs + RA profile attrs) plus the connector HTTP call per poll message. Bounded
+     * constant per message (not per cert), so the absolute round-trip cost is small —
+     * but in high-throughput async pipelines (thousands of in-flight cert ops) a single
+     * batched lookup would halve the DB load from this listener. Tracked as M4-M9
+     * follow-up; the API surface change in AttributeEngine is out of scope for this batch.
+     */
+    @Override
+    public StatusPollResult pollStatus(Certificate cert, CertificateOperation op) throws ConnectorException {
```
ivosh: AuthorityProviderV3Adapter.pollStatus(...) (line 212) and cancel(...) (line 232), declared via the new AsyncOperationCapability interface, are reached only from AuthorityProviderV3AdapterTest. Since the factory itself is uncalled, there is no production path that could produce an ASYNC_ACCEPTED outcome and poll it. Confirm via grep/gh search there is no production caller.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -0,0 +1,356 @@
...
+
+    @Override
+    public List<BaseAttribute> listRegisterAttributes(AuthorityInstanceReference authority, RaProfile raProfile) throws ConnectorException {
+        ApiClientConnectorInfo connectorDto = connectorForApiClient(authority);
+        return connectorApiFactory.getCertificateApiClientV3(connectorDto)
+                .listRegisterAttributes(connectorDto, attributeListRequest(authority, raProfile));
+    }
+
+    @Override
+    public AdapterOperationResult register(Certificate cert, ClientCertificateRegistrationDto req) throws ConnectorException {
```
ivosh: AuthorityProviderV3Adapter.register(...) (line 181), backing the new RegisterCapability interface, is invoked only from AuthorityProviderV3AdapterTest. Same root cause as the factory finding — no production dispatch reaches it. Confirm via grep/gh search there is no production caller.

src/main/resources/db/migration/V202605251200__add_connector_interface_uuid_to_authority_instance_reference.sql
```diff
@@ -0,0 +1,31 @@
+ALTER TABLE authority_instance_reference
+    ADD COLUMN connector_interface_uuid UUID
+    REFERENCES connector_interface(uuid) ON UPDATE CASCADE ON DELETE RESTRICT;
```
ivosh, line 3: The new connector_interface_uuid column is a foreign key joined via @ManyToOne. No index is created on authority_instance_reference(connector_interface_uuid). ON DELETE RESTRICT FK checks and entity joins will do sequential scans. The peer column secrets.connector_interface_uuid (V202602200935) follows the same un-indexed pattern. Verify the migration creates no index and the peer pattern.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -0,0 +1,356 @@
...
+     * Connector-side failures (the call itself throwing) propagate unchanged.
+     */
+    private AdapterOperationResult executeWithAcceptanceGuard(
+            CertificateOperation op, ConnectorCall call,
+            java.util.function.Function<ResponseEntity<CertificateDataResponseDto>, AdapterOperationResult> mapper)
+            throws ConnectorException {
+        boolean connectorAccepted = false;
+        try {
+            ResponseEntity<CertificateDataResponseDto> response = call.call();
+            connectorAccepted = response.getStatusCode().is2xxSuccessful();
```
ivosh, line 274: executeWithAcceptanceGuard (AuthorityProviderV3Adapter.java:279-293) sets connectorAccepted = response.getStatusCode().is2xxSuccessful() and only the mapper rejects unexpected codes. If a connector client ever returns a non-2xx ResponseEntity instead of throwing (e.g. a 4xx surfaced as a value), connectorAccepted stays false and the mapper's IllegalStateException("Unexpected v3 status") propagates raw as a RuntimeException rather than a ConnectorException — callers expecting only ConnectorException on the failure path could see an unchecked exception escape. This is the correct rollback semantic (connector did not accept), so it is not a state-divergence bug; flagging only that the unexpected-status path produces an unchecked exception type. Confirm the v3 clients always throw on non-2xx so this branch is unreachable.

src/main/java/com/otilm/core/service/handler/authority/AsyncOperationCapability.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.otilm.api.exception.ConnectorException;
+import com.otilm.core.dao.entity.Certificate;
+
+/**
+ * v3-only capability: poll status and cancel in-flight async operations.
+ */
+public interface AsyncOperationCapability {
+
+    StatusPollResult pollStatus(Certificate cert, CertificateOperation op) throws ConnectorException;
```
ivosh, line 16: AsyncOperationCapability declares two non-obvious public methods, pollStatus (line 11) and cancel (line 13), with only an interface-level comment. Their contract is non-trivial — pollStatus dispatches by CertificateOperation and returns a StatusPollResult that callers interpret against the transition matrix; cancel returns a CancelResult whose CancelOutcome (CANCELLED vs NOT_TRACKED vs REFUSED_PAST_POINT_OF_NO_RETURN) drives caller rollback decisions.

The sibling RegisterCapability documents both of its methods, and AuthorityProviderAdapter documents every method, so this is an inconsistency with the established pattern in the same package. Add method-level Javadoc describing the outcome semantics, particularly that NOT_TRACKED is a soft-success for cancel.

Although there is a JavaDoc on the method implementation, it does not document the contract.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -0,0 +1,356 @@
...
+
+    /**
+     * Runs a v3 connector call and maps the response, applying the post-acceptance failure rule:
+     * once the connector returns 2xx, a failure in the local mapping is surfaced as
+     * {@link ConnectorAcceptedButLocalFailureException} (callers must NOT roll back state).
+     * Connector-side failures (the call itself throwing) propagate unchanged.
+     */
+    private AdapterOperationResult executeWithAcceptanceGuard(
+            CertificateOperation op, ConnectorCall call,
+            java.util.function.Function<ResponseEntity<CertificateDataResponseDto>, AdapterOperationResult> mapper)
```
ivosh: executeWithAcceptanceGuard at line 277 spells the mapper parameter as java.util.function.Function<...> fully qualified inline rather than via an import. Every other type in this file is imported; the inline FQN here makes the already-dense generic signature harder to read. Add an import java.util.function.Function and use the short name. Trivial, non-blocking.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapterFactory.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public AuthorityProviderAdapter forAuthority(AuthorityInstanceReference authority) {
+        ConnectorInterfaceEntity iface = authority.getConnectorInterface();
+        if (iface == null) {
+            // Connector declared no AUTHORITY interface row in connector_interface. This is
+            // expected for framework-v1 connectors that implement the v2 authority wire
+            // protocol (e.g. ejbca-ng) — their /v1/info endpoint predates the v2 interface
+            // descriptor and the M0 backfill cannot create a row for them. Route to v2
+            // since that is the wire protocol they speak. Legacy v1-authority connectors
```
ivosh: Nitpick: 
This repeats the same information from class-level JavaDoc but is not relevant here.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapterFactory.java
```diff
@@ -0,0 +1,61 @@
...
+        this.v3Adapter = v3Adapter;
+    }
+
+    public AuthorityProviderAdapter forAuthority(AuthorityInstanceReference authority) {
+        ConnectorInterfaceEntity iface = authority.getConnectorInterface();
+        if (iface == null) {
+            // Connector declared no AUTHORITY interface row in connector_interface. This is
+            // expected for framework-v1 connectors that implement the v2 authority wire
+            // protocol (e.g. ejbca-ng) — their /v1/info endpoint predates the v2 interface
+            // descriptor and the M0 backfill cannot create a row for them. Route to v2
```
ivosh: **~ Minor** | Maintainability

`AuthorityProviderAdapterFactory.java:42` — the comment references "the M0 backfill", an internal epic-milestone label. To a future reader (or anyone outside epic #110) "M0" is opaque and will rot once the milestone naming is forgotten. Prefer describing what the backfill does (or link the migration/issue that performs it) rather than the milestone tag. The same applies to the sibling comment in `AuthorityProviderAdapterFactoryTest.java:50`.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    // ---- AsyncOperationCapability ----
+
+    /**
+     * Issues 3 separate {@code attribute_content_2_object} reads (cert meta + authority
+     * attrs + RA profile attrs) plus the connector HTTP call per poll message. Bounded
+     * constant per message (not per cert), so the absolute round-trip cost is small —
+     * but in high-throughput async pipelines (thousands of in-flight cert ops) a single
+     * batched lookup would halve the DB load from this listener. Tracked as M4-M9
```
ivosh: **~ Minor** | Maintainability

`AuthorityProviderV3Adapter.java:208` — "Tracked as M4-M9 follow-up" points at an internal milestone range rather than a linkable tracker. A "tracked as" note is only actionable if it links the actual issue; "M4-M9" will be opaque to a future maintainer. Replace with a GitHub issue link (or drop the milestone tag and keep just the rationale).

src/test/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapterFactoryTest.java
```diff
@@ -0,0 +1,72 @@
...
+        // (rather than NPE on the switch). Distinct from the null-interface fallback below.
+        AuthorityInstanceReference auth = authorityWithVersion(null);
+        assertThrows(UnsupportedAuthorityVersionException.class, () -> factory.forAuthority(auth));
+    }
+
+    @Test
+    void missingInterfaceFallsBackToV2() {
+        // A null connector interface is expected for framework-v1 connectors that implement the
+        // v2 authority wire protocol (e.g. ejbca-ng) — they declare no connector_interface row,
+        // so the M0 backfill leaves the FK null. The factory routes these to the v2 adapter
```
ivosh: **~ Minor** | Maintainability

`AuthorityProviderAdapterFactoryTest.java:50` — same internal-milestone reference as the production comment: "so the M0 backfill leaves the FK null". "M0" is an opaque epic-milestone label that will rot; describe the backfill, link it or simple delete it instead. Pairs with the flag on `AuthorityProviderAdapterFactory.java:42`.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -0,0 +1,356 @@
...
+    @Override
+    public List<BaseAttribute> listRevokeAttributes(AuthorityInstanceReference authority, RaProfile raProfile) throws ConnectorException {
+        ApiClientConnectorInfo connectorDto = connectorForApiClient(authority);
+        return connectorApiFactory.getCertificateApiClientV3(connectorDto)
+                .listRevokeAttributes(connectorDto, attributeListRequest(authority, raProfile));
+    }
+
+    @Override
+    public void validateIssueAttributes(AuthorityInstanceReference authority, List<RequestAttribute> attributes) {
+        // v3 dropped the connector-side /validate endpoint (interfaces #560). Structural
```
ivosh: Doc: this information is retained in git history and not relevant here. I'd suggest to drop the interface PR reference.

ivosh (changes requested):
Good work with the adapter!
Some minor issues found.

I was not able to review how the new adaptation layer compares to the existing code flow - this will come in future PRs. I'd be only speculating now.

src/main/java/com/otilm/core/service/handler/authority/CertificateOperation.java
lubomirw: Added `CertificateOperationTest` in 2d5014dc — asserts all four operations × pendingState/terminalSuccessState/terminalFailureState, including the REVOKE→ISSUED failure mapping you called out.

src/test/java/com/otilm/core/service/handler/authority/AdapterOperationResultTest.java
lubomirw: Deleted both in 06eca36c2 — the enum logic is now exercised by `CertificateOperationTest` and the cancel/outcome tests, so the non-null assertions were redundant green-by-construction noise.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
lubomirw: Covered in 2d5014dc: `pollStatusUsesRevokeEndpointForRevoke`, `pollStatusUsesRegisterEndpointForRegister`, and `cancelUsesCancelRevokeForRevoke`. RENEW shares the ISSUE switch arm, so it's exercised by the ISSUE tests.

src/main/resources/db/migration/V202605251200__add_connector_interface_uuid_to_authority_instance_reference.sql
lubomirw: Keeping it un-indexed. `authority_instance_reference` holds ~one row per authority, so an index's selectivity benefit is negligible while it still adds write/maintenance cost — it lowers usability for what it buys. Matches the un-indexed peer `secrets.connector_interface_uuid`. Worth adding only if the table ever grows materially.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
lubomirw: Confirmed: the v3 BaseApiClient pipeline throws `ConnectorException` (RFC 9457 problem+json → `ConnectorProblemException`) on every non-2xx, so a non-2xx never surfaces as a returned value — this branch is only reachable for an *unexpected 2xx*, which is the intentional accepted-but-local-failure (don't roll back) path. No change.

src/main/java/com/otilm/core/service/handler/authority/AsyncOperationCapability.java
lubomirw: Added method-level Javadoc in 06eca36c2 for both pollStatus and cancel, documenting the StatusPollResult/CancelOutcome semantics — including that NOT_TRACKED is a soft success for cancel.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
lubomirw: Fixed in 06eca36c2 — added `import java.util.function.Function` and use the short name.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapterFactory.java
lubomirw: Trimmed in 06eca36c2 — the inline block duplicated the class Javadoc; reduced to a one-line pointer.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapterFactory.java
lubomirw: Fixed in 06eca36c2 — dropped the 'M0' milestone tag; the comment now describes the framework-v1 / no-interface-row case without the internal label.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
lubomirw: Removed in 06eca36c2 — dropped the speculative perf note and the 'M4-M9' tag. A future optimization like this belongs in a tracked issue, not a code comment.

src/test/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapterFactoryTest.java
lubomirw: Fixed in 06eca36c2 — same de-milestoned wording in the test comment.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
lubomirw: Fixed in 06eca36c2 — dropped the interfaces PR reference; kept the 'v3 has no connector-side /validate' rationale.

ivosh (approved):

Issue #719, opened 2024-05-15
Add trigger history to store details and results of trigger evaluations


lubomirw:
### Interfaces

Add endpoint to retrieve trigger history: `GET /v1/rules/triggers/{triggerUuid}/history/{trigger_object_uuid}` with 2 optional query.
It allows to filter trigger history based on `RuleTrigger2Object` association.

### Implement data model regarding trigger evaluation history:
- implement endpoint to retrieve history
- implement data model entities and repositories
- add logic to `RuleEvaluator` to create history while evaluating triggers
  - create `RuleTriggerHistory` item for each resource object evaluated
  - create `RuleTriggerHistoryRecord` for each condition or action failed for particular object
  - `conditions_matched` and `actions_performed` are true when all conditions or actions are successfully performed

```mermaid
---
title: Discovery Rules (basis for rules engine)
---
erDiagram
    RuleTrigger ||--o{ RuleTriggerHistory: has
    RuleTrigger2Object |o--o{ RuleTriggerHistory: has
    RuleTriggerHistory {
        UUID uuid PK
        UUID trigger_uuid
        UUID trigger_association_uuid "NULL"
        bool conditions_matched
        bool actions_performed
        UUID object_uuid
        timestamp triggered_at
        text message "NULL"
    }
    RuleTriggerHistory ||--o{ RuleTriggerHistoryRecord: has
    RuleTriggerHistoryRecord {
        UUID uuid PK
        UUID trigger_history_uuid
        UUID rule_condition_uuid "NULL"
        UUID rule_action_uuid "NULL"
        text message "NULL"
    }
    RuleTrigger {
    }
   RuleTrigger2Object {
    }
```

PR #1614, opened 2026-06-16
Add v3 authority-instance support (lifecycle and provider operations)

### LINKED ISSUE #1601: Add v3 authority-instance support (lifecycle and provider operations)
### Description

Joins the v3 client-bean wiring with the schema migration and v3 authority-creation path (both small, same v3-enablement theme). All of it is inert until the activation work, so it is safe to merge ahead of it.

- Wire the v3 REST + MQ `CertificateApiClient` / `AuthorityApiClient` beans in `ApplicationConfig` / `ProxyClientConfig`, exposed via `ConnectorApiFactory`.
- Flyway migration adding `authority_instance_reference.connector_interface_uuid` (FK) with a **deterministic** backfill, plus the entity mapping; the FK is set on v2 authority create.
- `AuthorityInstanceServiceImpl` v3 creation path: skips the connector `createAuthorityInstance` call and resolves the interface by explicit `interfaceUuid`.

Disjoint from `service/v2/impl/ClientOperationServiceImpl`, so it can merge independently.

**Depends on:** #1600
**Review focus:** backfill determinism / idempotency (the only stateful slice); FK set on v2 create; REST/MQ bean symmetry.

### Definition of Done

- [ ] v3 REST + MQ client beans wired and exposed via `ConnectorApiFactory` (symmetric)
- [ ] Flyway migration + entity FK with a deterministic, idempotent backfill; FK set on v2 create
- [ ] `AuthorityInstanceServiceImpl` v3 creation path (skips `createAuthorityInstance`; resolves by `interfaceUuid`) with test
- [ ] PR merged to `main`

lubomirw:
Third slice of landing `feat/authority-v3` on `main` as a reviewable stacked-PR chain. Closes #1601. Part of epic #110. Builds on the merged adapter layer (#1607) and state machine (#1605) — both already on `main`, so this PR targets `main` directly.

## What

Adds the v3 authority-instance create/edit path to `AuthorityInstanceServiceImpl`, routing through the version-dispatch adapter from #1607:

- v3 authorities are **stateless** — no connector `createAuthorityInstance` / `updateAuthorityInstance` call. Attributes are validated locally against the definitions listed from the v3 attributes endpoint (`adapter.listAuthorityInstanceAttributes`), and connectivity is probed via `adapter.checkAuthorityConnection`.
- v1/v2 keep the legacy function-group attribute path unchanged.
- `resolveAuthorityInterface` binds the authority to a connector interface: an explicit `interfaceUuid` on the request selects v3 on a connector exposing both v2 and v3; absent it, creation falls back to the sole AUTHORITY interface and rejects ambiguity (more than one) with a clear error.
- editing a v3 authority re-probes connectivity (there is no update endpoint) and persists the attribute content as the effective "update".

The v3 connector client beans and the `connector_interface_uuid` FK migration named in #1601 already landed with the adapter slice (#1607); this PR completes #1601 with the authority-creation path.

## Notes for review

- Compiles against `main`; the v3 branch is reached only for connectors carrying a v3 AUTHORITY interface (none can exist until the activation slice #1603), so this is effectively additive. 9 unit tests pass.
- The credential field intentionally stays `CredentialInternalService` — the `feat/authority-v3` rename to a `CredentialService` interface is an orthogonal refactor, deliberately excluded from this slice.
- Code originates from `feat/authority-v3` (already reviewed there); re-sliced for incremental review.

## Pre-review addressed (guided-pr-review + Copilot — both SHIP after fixes)

- **[MEDIUM] Non-deterministic interface pick** — `resolveAuthorityInterface` returned `findFirst()` from a Hibernate `Set`, so a connector exposing more than one AUTHORITY interface with no `interfaceUuid` got an arbitrary pick. Now throws a clear `ValidationException` demanding an explicit `interfaceUuid`; regression test added.
- **Missing v3 edit-path / local-validation coverage** — added tests asserting the edit path re-probes (and never calls the connector update client), and that v3 create validates against the exact definitions the adapter lists. Plus the three `resolveAuthorityInterface` branches (explicit / unknown / ambiguous interface).
- Comment hygiene: dropped inline comments that merely restated the method Javadoc.

## Known items (intentional, not changed here)

- **Transaction boundary — deferred.** `createAuthorityInstance` / `editAuthorityInstance` carry the class-level `@Transactional` and run the connector probe (`checkAuthorityConnection` / `listAuthorityInstanceAttributes`) inside it, holding a DB connection across an external HTTP call (against the repo's tx/external-call convention). This is **pre-existing**: the legacy v1/v2 path already calls the connector inside the same transaction. A correct fix is whole-method (`NOT_SUPPORTED` + an explicit short transaction around the local writes) and changes the legacy path's semantics too, so it is out of scope for this focused slice. Tracked as a follow-up to refactor the boundary holistically across both paths, with legacy coverage.

## Verification

`mvn test` on `AuthorityInstanceServiceImplV3Test`: **9 tests, 0 failures**.


src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -175,6 +199,10 @@ public AuthorityInstanceDto createAuthorityInstance(com.otilm.api.model.client.a
         authorityInstanceRef.setConnectorUuid(connectorUuid.getValue());
         authorityInstanceRef.setKind(request.getKind());
         authorityInstanceRef.setConnectorName(connector.getName());
+        if (iface != null) {
+            authorityInstanceRef.setConnectorInterface(iface);
+            authorityInstanceRef.setConnectorInterfaceUuid(iface.getUuid());
+        }
```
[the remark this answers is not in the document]
lubomirw: Done — removed; `setConnectorInterface(iface)` already sets `connectorInterfaceUuid`.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -347,6 +383,98 @@ public void evaluatePermissionChain(SecuredUUID uuid) throws NotFoundException {
...
+    private AuthorityInstanceReference transientAuthorityRef(SecuredUUID connectorUuid, ConnectorDto connector,
+                                                             ConnectorInterfaceEntity iface, String kind) {
+        AuthorityInstanceReference ref = new AuthorityInstanceReference();
+        ref.setConnectorUuid(connectorUuid.getValue());
+        ref.setConnectorName(connector.getName());
+        ref.setKind(kind);
+        if (iface != null) {
+            ref.setConnectorInterface(iface);
+            ref.setConnectorInterfaceUuid(iface.getUuid());
+        }
```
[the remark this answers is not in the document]
lubomirw: Done — removed; the setter already populates `connectorInterfaceUuid`.

src/test/java/com/otilm/core/service/impl/AuthorityInstanceServiceImplV3Test.java
```diff
@@ -0,0 +1,275 @@
...
+        Set<ConnectorInterfaceEntity> interfaces = new HashSet<>();
+        interfaces.add(v3Iface);
+
+        connectorEntity = new Connector();
+        connectorEntity.setName("v3-connector");
+
+        // Use reflection to set interfaces (field is populated via Hibernate in prod, not a setter)
+        java.lang.reflect.Field ifacesField = Connector.class.getDeclaredField("interfaces");
+        ifacesField.setAccessible(true);
+        ifacesField.set(connectorEntity, interfaces);
```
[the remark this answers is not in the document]
lubomirw: Done — using `connectorEntity.getInterfaces().add(...)` (the collection is initialized); reflection dropped.

src/test/java/com/otilm/core/service/impl/AuthorityInstanceServiceImplV3Test.java
```diff
@@ -0,0 +1,275 @@
...
+    void createRejectsMultipleAuthorityInterfacesWithoutInterfaceUuid() throws Exception {
+        ConnectorInterfaceEntity secondIface = new ConnectorInterfaceEntity();
+        secondIface.setInterfaceCode(ConnectorInterface.AUTHORITY);
+        secondIface.setVersion("v2");
+        secondIface.setConnectorUuid(connectorUuid);
+        java.lang.reflect.Field ifacesField = Connector.class.getDeclaredField("interfaces");
+        ifacesField.setAccessible(true);
+        @SuppressWarnings("unchecked")
+        Set<ConnectorInterfaceEntity> ifaces = (Set<ConnectorInterfaceEntity>) ifacesField.get(connectorEntity);
+        ifaces.add(secondIface);
```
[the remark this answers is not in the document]
lubomirw: Done — switched to the `getInterfaces()` accessor instead of reflecting into the private field.

lubomirw:
Latest push (`0c466b4f`): addressed the Copilot review (dropped the redundant `setConnectorInterfaceUuid`; tests use `Connector.getInterfaces()` instead of field reflection), and `AuthorityInstanceReference.mapToDto()` now populates `connector` + `connectorInterface` — covering list/detail/edit/create via the shared mapper, with the deprecated `connectorUuid`/`connectorName` still populated.

Note for CI: this consumes the `connector`/`connectorInterface` fields added by interfaces #719/#720 (both merged), so it needs the `com.otilm:interfaces:2.18.1-SNAPSHOT` that includes them. If the build races a stale snapshot, a re-run once the interfaces snapshot publishes will resolve it.

src/main/java/com/otilm/core/dao/entity/AuthorityInstanceReference.java
```diff
@@ -83,8 +83,12 @@ public AuthorityInstanceDto mapToDto() {
         dto.setKind(kind);
         dto.setConnectorName(this.connectorName);
         if (this.connectorUuid != null) {
+            // connector supersedes the deprecated connectorUuid/connectorName pair, which stays
+            // populated until consumers migrate off it.
             dto.setConnectorUuid(this.connectorUuid.toString());
+            dto.setConnector(new NameAndUuidDto(this.connectorUuid, this.connectorName));
         }
+        dto.setConnectorInterface(this.connectorInterface == null ? null : this.connectorInterface.mapToDto());
         return dto;
```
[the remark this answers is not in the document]
lubomirw: Done — `listAuthorityInstances` now fetch-joins `connectorInterface` via `findUsingSecurityFilter(filter, List.of("connectorInterface"), null)` (LEFT join, so legacy v1 authorities with no interface are still returned), eliminating the per-row N+1. The association stays LAZY everywhere else; single-entity paths (detail/edit/create) map within the `@Transactional` method, so no LazyInitializationException.

src/test/java/com/otilm/core/service/impl/AuthorityInstanceServiceImplV3Test.java
```diff
@@ -0,0 +1,272 @@
...
+        verify(v3Adapter).checkAuthorityConnection(any(), any());
+    }
+
+    @Test
+    void createV3AuthoritySavesWithConnectorInterfaceUuid() throws Exception {
+        UUID ifaceUuid = UUID.randomUUID();
+        // Set UUID on the interface entity via reflection (UniquelyIdentified has a protected uuid field)
+        java.lang.reflect.Field uuidField = findUuidField(ConnectorInterfaceEntity.class);
+        uuidField.setAccessible(true);
+        uuidField.set(v3Iface, ifaceUuid);
```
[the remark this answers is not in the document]
lubomirw: Done — set directly via `v3Iface.setUuid(ifaceUuid)`.

src/test/java/com/otilm/core/service/impl/AuthorityInstanceServiceImplV3Test.java
```diff
@@ -0,0 +1,272 @@
...
+        verify(v3Adapter).checkAuthorityConnection(any(), any());                  // re-probe instead
+        verify(authorityInstanceReferenceRepository).save(existing);
+    }
+
+    @Test
+    void createV3AuthorityWithExplicitInterfaceUuidSelectsThatInterface() throws Exception {
+        UUID ifaceUuid = UUID.randomUUID();
+        java.lang.reflect.Field uuidField = findUuidField(ConnectorInterfaceEntity.class);
+        uuidField.setAccessible(true);
+        uuidField.set(v3Iface, ifaceUuid);
```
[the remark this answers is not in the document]
lubomirw: Done — set directly via `v3Iface.setUuid(ifaceUuid)`.

src/test/java/com/otilm/core/service/impl/AuthorityInstanceServiceImplV3Test.java
```diff
@@ -0,0 +1,272 @@
...
+        Class<?> current = clazz;
+        while (current != null) {
+            try {
+                return current.getDeclaredField("uuid");
+            } catch (NoSuchFieldException e) {
+                current = current.getSuperclass();
+            }
+        }
+        throw new IllegalStateException("Could not find 'uuid' field in hierarchy of " + clazz.getName());
+    }
```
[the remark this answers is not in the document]
lubomirw: Done — removed the `findUuidField` helper.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -347,6 +383,97 @@ public void evaluatePermissionChain(SecuredUUID uuid) throws NotFoundException {
...
+            AuthorityInstanceReference authorityInstanceRef,
+            List<?> dataAttributes)
+            throws ConnectorException, AttributeException, ValidationException, NotFoundException {
+        authorityInstanceRef.setName(request.getName());
+        authorityInstanceRef.setStatus("connected");
+
+        AuthorityProviderAdapter adapter = adapterFactory.forAuthority(authorityInstanceRef);
+        adapter.checkAuthorityConnection(authorityInstanceRef, AttributeDefinitionUtils.getClientAttributes(dataAttributes));
+
+        authorityInstanceReferenceRepository.save(authorityInstanceRef);
```
ivosh, line 492: **~ Minor** | Correctness — v3 create is reachable, but read/delete/list still hit the v2 client

**Known** (verified against the PR head + merged #1607):
- `createV3Authority` (~458-475) and `transientAuthorityRef` (~396-406) never set `authorityInstanceUuid` — correct, v3 is stateless.
- `getAuthorityInstance` (~123), `removeAuthorityInstance` (~491) and `listEndEntityProfiles`/`listCertificateProfiles`/`listCAsInProfile`/`listRAProfileAttributes` call the v2 client `connectorApiFactory.getAuthorityInstanceApiClient(...)` with `getAuthorityInstanceUuid()` and have **no `isV3`/version branch** — they guard only on `getConnector()`.
- Core does **not** gate interface versions at registration: `ConnectorV2Adapter.updateConnectorFunctions` (lines 111-131) persists whatever the connector advertises, `setVersion("v3")` included; `validateConnection` checks only mandatory INFO/HEALTH/METRICS.
- The adapter factory (merged in #1607) already accepts `"v3"`.

**Unknown:**
- Whether a later, currently-unfiled slice wires read/delete/list through `adapterFactory.forAuthority(ref)` for v3. #1601 is creation-only; #1603 is endpoints + capability gating — neither names this routing.
- Whether any deployed connector can advertise a v3 AUTHORITY interface at its V2 info endpoint today.

**Assumption (from the PR body):**
- *"the v3 branch is reached only for connectors carrying a v3 AUTHORITY interface (none can exist until the activation slice #1603), so this is effectively additive."* This safety net rests on #1603's **capability gating**, which is still **OPEN/unmerged**. With only #1607 + this PR merged, there is no capability gate in core.

**Consequences if the assumption doesn't hold:**
- In the window between this PR merging and #1603's gating landing, a connector advertising a v3 AUTHORITY interface lets an operator create a v3 authority (null `authorityInstanceUuid` persisted). A later GET-detail / DELETE / profile-listing then calls the v2 client with a null UUID — wrong client, null arg — surfacing as a failed operation; DELETE in particular can leave an undeletable row.
- Blast radius is bounded to v3 authorities; v1/v2 are unaffected.

**Action** (any one):
- Land #1603's capability gating before this reaches an environment where a connector can advertise v3; or
- Gate `createV3Authority` behind the same capability check so create and read/delete/list activate atomically; or
- Add the v3 read/delete/list adapter routing to this series — and file it if not already tracked.

Non-blocking for the create-path slice, but the "additive/unreachable" reassurance is **not** guaranteed by this PR's merged state — worth an explicit decision rather than an implicit dependency on an unmerged PR.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -153,14 +170,22 @@ public AuthorityInstanceDto createAuthorityInstance(com.otilm.api.model.client.a
...
+        ConnectorInterfaceEntity iface = resolveAuthorityInterface(connectorUuid.getValue(), request.getInterfaceUuid());
+        AuthorityInstanceReference probeRef = transientAuthorityRef(connectorUuid, connector, iface, request.getKind());
         attributeEngine.validateCustomAttributesContent(Resource.AUTHORITY, request.getCustomAttributes());
-        connectorService.mergeAndValidateAttributes(connectorUuid, codeToSearch, request.getAttributes(), request.getKind());
+        mergeAndValidateAuthorityAttributes(probeRef, iface, codeToSearch, request.getAttributes(), request.getKind());
 
         // Load complete credential data and resource data
         var dataAttributes = attributeEngine.getDataAttributesByContent(connectorUuid.getValue(), request.getAttributes());
         credentialService.loadFullCredentialData(dataAttributes);
         resourceService.loadResourceObjectContentData(dataAttributes);
```
ivosh, line 191: **. Nitpick** | Testing

Could you add a test for 'exactly one AUTHORITY interface, no interfaceUuid' fallback (the size!=>1, non-null return) — the explicit/unknown/ambiguous branches are tested but the happy single-interface fallback is not. Minor/optional: add one test for the single-interface auto-select. Non-blocking nitpick.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -347,6 +383,97 @@ public void evaluatePermissionChain(SecuredUUID uuid) throws NotFoundException {
...
+        }
+        if (interfaces.size() > 1) {
+            throw new ValidationException(
+                    "Connector " + connectorUuid + " exposes multiple AUTHORITY interfaces; supply interfaceUuid to select one.");
+        }
+        return interfaces.stream().findFirst().orElse(null);
+    }
+
+    private AuthorityInstanceDto createV3Authority(
+            com.otilm.api.model.client.authority.AuthorityInstanceRequestDto request,
```
ivosh: **. Nitpick** | Maintainability

createV3Authority at src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java:459 declares its parameter as the inline FQN `com.otilm.api.model.client.authority.AuthorityInstanceRequestDto request`. 

Add `import com.otilm.api.model.client.authority.AuthorityInstanceRequestDto;` and using the simple name in the new method. Note the preexisting createAuthorityInstance signature (line 142) already uses the same inline FQN and is outside this PR's diff; folding both onto the import in this PR would remove the inconsistency the new method otherwise propagates.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -347,6 +383,97 @@ public void evaluatePermissionChain(SecuredUUID uuid) throws NotFoundException {
         // Since there are is no parent to the Authority, exclusive parent permission evaluation need not be done
     }
 
+    private static boolean isV3(ConnectorInterfaceEntity iface) {
```
ivosh, line 409: **. Nitpick** | Maintainability

isV3 (src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java:386-388) hardcodes "v3".equals(iface.getVersion()), and AuthorityProviderAdapterFactory.forAuthority independently switches on "v2"/"v3". Two places now encode the version-string vocabulary. If a connector reports a non-canonical version (e.g. a bare "3"), isV3 returns false and the create/edit flow silently takes the legacy v2/connector path here, while the factory would throw UnsupportedAuthorityVersionException elsewhere — divergent handling of the same malformed input. Low risk given the factory's class Javadoc says versions are reported as "v2"/"v3". Action: consider centralizing the version vocabulary (enum or shared constant) so isV3 and the factory cannot drift. Non-blocking nit.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -347,6 +383,97 @@ public void evaluatePermissionChain(SecuredUUID uuid) throws NotFoundException {
...
+                    .filter(i -> interfaceUuid.equals(i.getUuid()))
+                    .findFirst()
+                    .orElseThrow(() -> new ValidationException(
+                            "Connector " + connectorUuid + " has no AUTHORITY interface with UUID " + interfaceUuid));
+        }
+        if (interfaces.size() > 1) {
+            throw new ValidationException(
+                    "Connector " + connectorUuid + " exposes multiple AUTHORITY interfaces; supply interfaceUuid to select one.");
+        }
+        return interfaces.stream().findFirst().orElse(null);
```
ivosh, line 478: **. Nitpick** | Maintainability

You could consider returning `Optional<ConnectorInterfaceEntity` instead of null/non-null. Optional would be marginally clearer for future AI readers (per CLAUDE.md "Writing for future AI sessions"), but given the tight, fully-guarded usage this is optional polish, not a correctness issue.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -92,10 +98,21 @@ public void setAttributeEngine(AttributeEngine attributeEngine) {
...
+    public void setAdapterFactory(AuthorityProviderAdapterFactory adapterFactory) {
+        this.adapterFactory = adapterFactory;
+    }
+
     @Override
     @ExternalAuthorization(resource = Resource.AUTHORITY, action = ResourceAction.LIST)
     public List<AuthorityInstanceDto> listAuthorityInstances(SecurityFilter filter) {
-        return authorityInstanceReferenceRepository.findUsingSecurityFilter(filter)
+        // fetch-join connectorInterface so mapToDto does not lazy-load it per row (avoids N+1)
+        return authorityInstanceReferenceRepository.findUsingSecurityFilter(filter, List.of("connectorInterface"), null)
```
ivosh: .Nitpick | Maintainability

Could you use a generated JPA name, such as `ConnectorInterface_` here, instead of a plain string?

ivosh (dismissed):

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
lubomirw: Done — added `createV3AuthorityFallsBackToSoleAuthorityInterfaceWhenNoInterfaceUuid`, asserting the sole AUTHORITY interface is selected when no interfaceUuid is supplied.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
lubomirw: Done — imported `AuthorityInstanceRequestDto`; both `createAuthorityInstance` and `createV3Authority` now use the simple name.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
lubomirw: Agreed in principle, but deferring: centralizing the v2/v3 vocabulary touches the already-merged factory (#1600), and the drift risk is low — the factory's own Javadoc pins versions to "v2"/"v3". Better as a small standalone cleanup than widening this slice; I'll track it.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
lubomirw: Keeping null here — `resolveAuthorityInterface` is private with fully-guarded callers, and `Optional` would ripple into `isV3`. Noted as optional polish per your comment.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
lubomirw: Done — now `List.of(AuthorityInstanceReference_.CONNECTOR_INTERFACE)`.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
lubomirw: Delivered in full in this PR rather than deferred (per discussion). Every v3 authority-instance operation is now v3-aware, so none mis-calls the v2 client:
- get/delete: v3 is stateless → skip the connector call, operate locally (no undeletable-row risk).
- listEndEntityProfiles/listCertificateProfiles/listCAsInProfile: v3 has no EJBCA end-entity-profile model → return empty.
- listRAProfileAttributes: routed through a new `AuthorityProviderAdapter.listRaProfileAttributes` (v3 → stateless `/raProfile/attributes`); validateRAProfileAttributes validates locally for v3.

So the "additive/unreachable" reassurance is now moot — v3 authorities are fully manageable end-to-end. (`validateRAProfileAttributes` declaring `AttributeException` needs interfaces OmniTrustILM/interfaces#724.)

lubomirw:
Latest push (`dc21f5960`) expands this PR to the **full v3 authority-instance lifecycle + provider operations** (per review discussion): get/delete v3-local, EEP-listings empty for v3, and listRAProfileAttributes/validate routed to v3 via a new adapter method. Reviewed by Copilot CLI + a guided pass; the one finding (validateRAProfileAttributes swallowing the validation error) is fixed — it now declares/propagates `AttributeException`.

**CI dependency:** that declaration requires interfaces **#724** (open). This PR needs the interfaces `2.18.1-SNAPSHOT` that includes #724 — a CI re-run after #724 merges + publishes will go green.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -262,15 +314,21 @@ public List<NameAndIdDto> listCAsInProfile(SecuredUUID uuid, Integer endEntityPr
...
     @ExternalAuthorization(resource = Resource.AUTHORITY, action = ResourceAction.ANY)
-    public Boolean validateRAProfileAttributes(SecuredUUID uuid, List<RequestAttribute> attributes) throws ConnectorException, NotFoundException {
+    public Boolean validateRAProfileAttributes(SecuredUUID uuid, List<RequestAttribute> attributes) throws ConnectorException, AttributeException, NotFoundException {
         AuthorityInstanceReference authorityInstance = getAuthorityInstanceReferenceEntity(uuid);
+        if (isV3(authorityInstance.getConnectorInterface())) {
+            // v3 is stateless: validate locally against the listed RA-profile definitions. Invalid
+            // content throws ValidationException and a definition/schema failure throws
+            // AttributeException — both surface to the caller, matching the v2 connector-validate path.
+            List<BaseAttribute> definitions = adapterFactory.forAuthority(authorityInstance).listRaProfileAttributes(authorityInstance);
+            attributeEngine.validateUpdateDataAttributes(authorityInstance.getConnectorUuid(), null, definitions, attributes);
```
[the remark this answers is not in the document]
lubomirw: Done — null-coalesce `attributes` to an empty list before `validateUpdateDataAttributes`, matching the guard in `mergeAndValidateAuthorityAttributes`. Added a test passing null.

klaraf755 (approved):

PR #1616, opened 2026-06-16
Persist signing records for granted TSP timestamps

### LINKED ISSUE #1567: Integrate Signing Record into the TSA engine (producer wiring + end-to-end integration test)
## Description
The signing-record subsystem — entity, repositories, persistence strategies (`ImmediateSigningRecordStrategy`, `DeferredDurableSigningRecordStrategy`, `BestEffortSigningRecordStrategy`), the outbox drainer, retention/retrieval sweepers, the `SigningRecordController`, and metrics — is fully built and unit/integration-tested in isolation on branch `feat/signing-record` (PR #1547). However, it currently has **no production caller**:

- The TSA signing flow never invokes `SigningRecordStrategyFactory.strategyFor(version).recordSigning(...)`, so no signing operation produces a signing record in production.
- `SigningRecordRetrievalHook.onSignedDocumentServed(...)` is never called when a signed document is served, so `signed_document_retrieved_at` is never stamped and delete-after-retrieval never fires.

The existing tests (including `SigningRecordEndToEndTest`) drive the subsystem by calling the strategy factory directly, so they verify the record machinery but **not** its integration with the production signing path.

This task wires the subsystem into the TSA engine and adds a true end-to-end integration test that exercises it through the production signing entry point.

## Definition of Done
- [ ] The TSA signing path invokes `SigningRecordStrategyFactory.strategyFor(version).recordSigning(input)` so that a signing operation produces a signing record according to the profile's configured persistence mode (immediate / deferred-durable / best-effort).
- [ ] The `recordSigning` call is placed so its persistence semantics respect the documented transaction boundaries (no DB transaction or row lock held across the timestamping/external call).
- [ ] The signed-document retrieval path invokes `SigningRecordRetrievalHook.onSignedDocumentServed(uuid)` within an ambient transaction (the hook is `@Transactional(MANDATORY)`), so `signed_document_retrieved_at` is stamped and the delete-after-retrieval flow can fire.
- [ ] A new end-to-end integration test drives a signing operation through the **production TSA entry point** (not `SigningRecordStrategyFactory` directly) and asserts a record is created; and drives a retrieval through the production endpoint and asserts the document is stamped retrieved and, for a delete-after-retrieval profile, eventually deleted.
- [ ] `SigningRecordEndToEndTest` is updated (or superseded) so its name reflects what it actually covers — no green test that bypasses the production wiring it claims to verify.
- [ ] The DEFERRED_DURABLE producer path is reconciled with profile deletion: deleting a profile with undrained `signing_record_outbox` rows does not orphan them into a poison loop (see PR #1547 review finding).

LukasNajman:


src/main/java/com/otilm/core/signing/tsa/TspSigningRecordFactory.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public SigningRecordInput build(SigningProfileModel<?, ?> signingProfile, TspRequest request,
+                                    BigInteger serialNumber, Instant genTime, byte[] encodedToken) {
+        String serialHex = serialNumber.toString(16);
+
+        // The timestamp token is the self-contained signed artifact: it already embeds the signature value and the
+        // signed attributes (DTBS), so both are recoverable from it. Storing them again under signature/dtbs would
+        // duplicate substrings of the token, so only signedDocument is populated for the TSP path.
+        return SigningRecordInput.builder()
+                .signingProfile(signingProfile)
```
lubomirw, line 58: **‼️ Critical** | Correctness

`build()` never sets `protocol`, so the assembled `SigningRecordInput` carries `protocol == null`.

`protocol` is a `NOT NULL` column on both `signing_record` and `signing_record_outbox` (migration `V202606161000__signing_record_protocol.sql`, and `@Column(nullable = false)` on the entities). `SigningRecordInputMapper.toRecord/toOutbox` maps `setProtocol(input.getProtocol())`, so every insert from the TSP path violates the constraint. Because `ManagedTimestampEngine.recordSigning(...)` wraps the call in `catch (Exception)` and only logs, the timestamp is still granted but **no signing record is ever persisted** — silent audit-data loss in production, and the cause of the red `Test (non-services)` check (`TsaServiceImplTest.persistsSigningRecord_whenTimestampGranted` → `Expected size: 1 but was: 0`).

This is the first production path to build a `SigningRecordInput` since the protocol column was introduced, so nothing else caught it. Fix: add `.protocol(SigningProtocol.TSP)` to the builder. Affects all three persistence modes (the outbox column is NOT NULL too), so the single factory fix covers them all.

src/main/java/com/otilm/core/signing/tsa/ManagedTimestampEngine.java
```diff
@@ -101,4 +114,18 @@ public TspResponse process(TspRequest request, ResolvedManagedTimestampingProfil
         }
 
     }
+
+    /**
+     * Records the granted timestamp. The signature has already been produced by the managed key, so a recording
+     * failure must never downgrade the response: it is logged and swallowed, leaving the granted token intact.
+     * The {@code recordingEnabled} policy gate lives inside the strategy, so disabled profiles are a no-op.
+     */
+    private void recordSigning(SigningProfileModel<?, ?> signingProfile, TspRequest request, BigInteger serialNumber, Instant genTime, byte[] encodedToken) {
```
lubomirw, line 125: **~ Minor** | Performance

`tspSigningRecordFactory.build(...)` is invoked on every granted timestamp before the `recordingEnabled` gate is reached. That gate lives inside the strategy (per this method's own javadoc), so for profiles with recording disabled the full input — including the `ObjectMapper` serialization of `requestMetadataJson` — is assembled and then discarded. On the TSA hot path this is avoidable work.

Consider short-circuiting when `!signingProfile.recordPolicy().recordingEnabled()` before building the input, so disabled profiles skip both the factory call and the strategy lookup entirely.

src/test/java/com/otilm/core/signing/tsa/TsaServiceImplTest.java
```diff
@@ -0,0 +1,341 @@
...
+import java.util.UUID;
+
+import static com.otilm.core.signing.tsa.messages.TspRequestBuilder.aTspRequest;
+import static com.otilm.core.util.builders.ConnectorRequestDtoBuilder.aV1ConnectorRequest;
+import static com.otilm.core.util.builders.ConnectorRequestDtoBuilder.aV2ConnectorRequest;
+import static com.otilm.core.util.builders.KeyPairRequestDtoBuilder.aKeyPairRequest;
+import static com.otilm.core.util.builders.SearchRequestDtoBuilder.aSearchRequest;
+import static com.otilm.core.util.builders.SigningProfileRequestDtoBuilder.aSigningProfileRequest;
+import static com.otilm.core.util.builders.SigningRecordPolicyRequestDtoBuilder.aSigningRecordPolicyRequest;
+import static com.otilm.core.util.builders.SigningRecordPolicyRequestDtoBuilder.recordingEverything;
```
lubomirw: **. Nitpick** | Maintainability

`recordingEverything` is statically imported but never used in this test — the profile setup uses `aSigningRecordPolicyRequest()` with explicit `with*` calls. Remove the unused import.

lubomirw (changes requested):
Code review: 3 issue(s) found. The Critical one is the cause of the failing `Test (non-services)` check.

src/main/java/com/otilm/core/signing/tsa/TspSigningRecordFactory.java
LukasNajman: Protocol is now added into the RecordInput. The gap was introduced by merging other PRs to  main.

src/test/java/com/otilm/core/signing/tsa/TsaServiceImplTest.java
LukasNajman: Removed

src/main/java/com/otilm/core/signing/tsa/ManagedTimestampEngine.java
LukasNajman: A new callback interface is not used to build the record only when needed and to not skew any record related metrics

lubomirw (approved):

PR #1625, opened 2026-06-18
Add authorization to the timestamping protocol endpoints

### LINKED ISSUE #1617: Enforce authorization on the TSA timestamping signing path (TSP_SIGN)
## Description
The TSA timestamping signing path in `TsaServiceImpl` should enforce authorization before processing a timestamp request. Add `@ExternalAuthorization(resource = TSP_PROFILE, action = TSP_SIGN)` to the signing entry points (`authorizeAndProcessForTspProfile`, `authorizeAndProcessForSigningProfile`).

Because these methods are invoked from within the same bean, they must be routed through the Spring AOP proxy via the `@Lazy self`-injection pattern — direct self-invocation would bypass the authorization advice and silently skip the access check.

Note: the feature is present on `feat/signing` branch - it needs to be extracted and polished into its own PR.

## Definition of Done
- [ ] `@ExternalAuthorization(resource = TSP_PROFILE, action = TSP_SIGN)` applied to the TSA signing entry points
- [ ] `@Lazy self`-injection in place so authorization advice is not bypassed by self-invocation
- [ ] Test coverage for both authorized and unauthorized timestamping requests (`TsaServiceAuthzTest`, plus supporting `TsaServiceImplTest`, `TsaServiceImplUnitTest`, `TimestampTokenTestUtil`)
- [ ] Independent review pass run on the PR diff (per `CLAUDE.md`)
- [ ] PR merged to `main`

LukasNajman:


src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java
```diff
@@ -38,9 +48,25 @@ public TsaServiceImpl(TspRequestValidator tspRequestValidator, SigningProfileSer
...
         TspProfileModel tspProfile = tspProfileService.getTspProfile(tspProfileName);
+        try {
+            return self.authorizeAndProcessForTspProfile(SecuredUUID.fromUUID(tspProfile.uuid()), tspProfile, request);
+        } catch (AccessDeniedException e) {
+            throw notFoundForDeniedTimestamping(tspProfile.name(), e);
+        }
+    }
+
+    @ExternalAuthorization(resource = Resource.TSP_PROFILE, action = ResourceAction.TIMESTAMP)
+    public TspResponse authorizeAndProcessForTspProfile(SecuredUUID tspProfileUuid, TspProfileModel tspProfile, TspRequest request)
```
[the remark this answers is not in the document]
LukasNajman: Used by @ExternalAuthorizatio

src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java
```diff
@@ -57,8 +83,26 @@ public TspResponse processTspRequestForSigningProfile(String signingProfileName,
...
+            // distinguish it from a non-existent profile (enumeration defense). See notFoundForDeniedTimestamping.
+            log.warn("Signing profile '{}' has no associated TSP profile; cannot authorize timestamping", signingProfile.name());
+            throw new NotFoundException(SigningProfileService.class, signingProfileName);
+        }
+
+        try {
+            return self.authorizeAndProcessForSigningProfile(SecuredUUID.fromUUID(signingProfile.tspProfileUuid()), signingProfile, request);
+        } catch (AccessDeniedException e) {
+            throw notFoundForDeniedTimestamping(signingProfile.name(), e);
+        }
```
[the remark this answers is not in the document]
LukasNajman: Fixed - now the class is passed as argument to the method

lubomirw (commented):
Code review — 2 advisory comments, no blockers. The main one is about timestamping ending up behind two authz checks (DETAIL + TIMESTAMP); mostly I want to understand the intended authorization model before suggesting a change. Details inline.

src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java
```diff
@@ -38,9 +48,25 @@ public TsaServiceImpl(TspRequestValidator tspRequestValidator, SigningProfileSer
...
+    @Lazy
+    @Autowired
+    public void setSelf(TsaServiceImpl self) {
+        this.self = self;
+    }
+
     @Override
     public TspResponse processTspRequestForTspProfile(String tspProfileName, TspRequest request) throws NotFoundException, TspException {
         TspProfileModel tspProfile = tspProfileService.getTspProfile(tspProfileName);
+        try {
```
lubomirw: **! Major** | Security / Maintainability

Heads up — this ends up gating timestamping behind *two* authz checks, which I don't think is what we want. To get the profile we call `getTspProfile(name)` / `getSigningProfileModel(name)`, and those already carry `@ExternalAuthorization(… DETAIL)`. That runs before (and outside) the try block, so a caller now needs both `DETAIL` *and* the new `TIMESTAMP` action to timestamp anything.

Everywhere else we do the opposite: SIGN/ISSUE/REVOKE (e.g. `ClientOperationServiceImpl`, `CryptographicOperationServiceImpl`) authorize only the parent resource and then load the target with a plain non-authorizing loader, specifically so we don't stack a second check on top. This path breaks from that.

Couple of things I'd want to nail down:

- The `tsp` protocol principal will have to be granted `DETAIL` as well as `TIMESTAMP`, otherwise timestamping just fails. And right now `ProtocolValidationFilter` doesn't even have a `tsp` branch (only scep/acme/cmp), nor is there a `TSP_USERNAME` — so whatever principal ends up serving `/v1/protocols/tsp/**` needs to be wired up with both. Is that coming in a separate change?
- The `DETAIL` denial on the lookup never hits `notFoundForDeniedTimestamping` — it bubbles up to the controller's `catch (Exception)` and comes back as `SYSTEM_FAILURE` ("unexpected error"), while a `TIMESTAMP` denial comes back as the not-found rejection we intended. So the "looks like it doesn't exist" masking only half works.

So mainly I want to understand the intent here: was the `DETAIL` check on the lookup meant to be part of the authorization story, or is `TIMESTAMP` supposed to be the single gate and the `DETAIL` is just incidental from reusing the public getter? If it's the latter, resolving the profile through a non-authorizing loader would line this up with the rest of the codebase and make the not-found masking uniform — but I'd rather hear your thinking before assuming.

src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java
```diff
@@ -57,8 +83,26 @@ public TspResponse processTspRequestForSigningProfile(String signingProfileName,
...
-            var message = "Signing profile '%s' does not have a TSP profile associated.".formatted(signingProfile.name());
+        if (signingProfile.tspProfileUuid() == null) {
+            // No linked TSP profile means there is no TSP_PROFILE resource to run authorization against, so we
+            // cannot tell whether the caller is permitted. Surface it as not-found so an unauthorized caller cannot
+            // distinguish it from a non-existent profile (enumeration defense). See notFoundForDeniedTimestamping.
+            log.warn("Signing profile '{}' has no associated TSP profile; cannot authorize timestamping", signingProfile.name());
+            throw new NotFoundException(SigningProfileService.class, signingProfileName);
+        }
+
+        try {
```
lubomirw: **~ Minor** | Correctness

The `try` wraps the whole `authorizeAndProcess*` call, not just the authorization step — so *any* `AccessDeniedException` raised anywhere inside gets turned into a not-found. Today that includes the inner `getTspProfile(uuid)` lookup (also `DETAIL`), and in principle anything the resolver/engine might throw later. It happens to do the right thing for the lookup, but it's broader than "deny the timestamping action → hide it": a future authz check deep in the processing path would be silently masked as not-found too.

Not urgent, and it partly resolves itself if we go the non-authorizing-loader route from the other comment. Just flagging that the catch is wider than the intent it documents.

src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java
LukasNajman: As discussed with @lubomirw, the two authz checks are expected and we already have precedence for this in other operations e.g. `com.otilm.core.service.impl.ClientOperationServiceImpl.issueCertificate()`.

src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java
LukasNajman: This is expected, I have also moved the try-catch to the controller so all access denied exceptions are handled equally as Not Found (to prevent enumeration attacks).

src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java
```diff
@@ -57,8 +68,18 @@ public TspResponse processTspRequestForSigningProfile(String signingProfileName,
         SigningProfileModel<?, ?> signingProfile = signingProfileService.getSigningProfileModel(signingProfileName);
         LoggingHelper.putLogResourceInfo(Resource.SIGNING_PROFILE, true, signingProfile.uuid().toString(), signingProfile.name());
 
-        if (!signingProfile.enabledProtocols().contains(SigningProtocol.TSP) || signingProfile.tspProfileUuid() == null) {
-            var message = "Signing profile '%s' does not have a TSP profile associated.".formatted(signingProfile.name());
+        if (signingProfile.tspProfileUuid() == null) {
```
lubomirw, line 71: **⚠ Minor** | Security

This `tspProfileUuid() == null` check runs before the authorization call, so it leaks profile existence to unauthorized callers.

A non-existent signing profile and an existing-but-unauthorized one both return the generic not-found rejection, but an existing profile with no linked TSP profile returns a distinct `BAD_REQUEST` ("does not have the TSP protocol enabled") with no permission required. An unauthorized caller can thus distinguish "exists without TSP" from "absent / no access," partially undermining the enumeration-defense goal stated in this PR. The TSP-profile endpoint does not have this asymmetry (only existence is pre-auth, and it collapses to not-found). This is inherent — a null UUID cannot be authorized against — so either collapse this case into the same not-found rejection, or document the accepted residual leak.

src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java
```diff
@@ -57,8 +68,18 @@ public TspResponse processTspRequestForSigningProfile(String signingProfileName,
...
+            throw new TspException(TspFailureInfo.BAD_REQUEST, message, message);
+        }
+
+        permissionEvaluator.tspProfileTimestamping(SecuredUUID.fromUUID(signingProfile.tspProfileUuid()));
+        return processForSigningProfile(signingProfile, request);
+    }
+
+    private TspResponse processForSigningProfile(SigningProfileModel<?, ?> signingProfile, TspRequest request) throws NotFoundException, TspException {
+        if (!signingProfile.enabledProtocols().contains(SigningProtocol.TSP)) {
+            var message = "Signing profile '%s' does not have the TSP protocol enabled.".formatted(signingProfile.name());
```
lubomirw, line 82: **⚠ Minor** | Maintainability

Two distinct conditions now emit the identical message "does not have the TSP protocol enabled.": (1) `tspProfileUuid() == null` (no TSP profile linked) and (2) `!enabledProtocols().contains(TSP)` (protocol not in the enabled set).

These are different misconfigurations, so an identical message makes operator diagnosis harder. The message is also inaccurate for case (1) when the protocol is enabled but no profile is linked. The prior message distinguished them ("does not have a TSP profile associated"). Suggest keeping distinct, accurate messages for the two branches.

lubomirw (commented):

src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java
LukasNajman: This is actually not true. Before the null check, calling `signingProfileService.getSigningProfileModel()` verifies callers right to access the Signing Profile detail. So when the null check is reached, we have verified the user is authorized to see the details of the signing profile, so reporting a not-linked TSP profile (TSP protocol enabled) is not exposing any information the caller would otherwise not be authorized to see.

The authorization for TIMESTAMP is run later only because we need the TSP Profile id to do it. Otherwise it would be ideal to run it as first check.

src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java
LukasNajman: In normal circumstances, the two information (1) TSP protocol present in protocols and (2) TSP Profile UUID set go hand in hand, so either both are set or none. For one check to trigger and the second not, the data would have to be corrupted. In normal situation, when not set, the both mean the protocol was not enabled on the Signing Profile.

lubomirw (approved):

PR #1627, opened 2026-06-19
Add async certificate-status polling (due-time table + sweep)

### LINKED ISSUE #1602: Add async certificate-status polling over JMS
### Description

Add the `provider-status-poll` JMS queue and the machinery to poll a connector for async (HTTP 202) certificate operations and apply terminal transitions through the state machine. Nothing enqueues poll messages until the activation work, so it is safe to merge ahead of it.

- `CertificateStatusPollMessage`, `CertificateStatusPollProducer`, `CertificateStatusPollListener`
- `StatusPollProperties`, `PollJmsEndpointConfig`, and additions to `MessagingProperties` / `JmsConfig` / `JmsListenersConfigurerImpl` / `MessagingConcurrencyProperties`
- `CertificateRevocationFinalizer`, `events/transaction/TransactionHandler`, a single-query poll finder on `CertificateRepository`
- `application.yml` defaults + per-operation (by-kind) overrides

Can be authored concurrently with #1601.

**Depends on:** #1599, #1600
**Review focus:** re-enqueue backoff / max-attempts; terminal transition via the state machine; meta-write split from the state transition; transaction rollback via `TransactionHandler` (no DB transaction held across the connector call).

### Definition of Done

- [ ] Poll message / producer / listener, `StatusPollProperties`, JMS endpoint config, finalizer, `TransactionHandler`, poll finder, and yml defaults added with tests
- [ ] Backoff / max-attempts re-enqueue behavior covered
- [ ] Terminal transition applied via the state machine; meta-write split from the transition; rollback via `TransactionHandler`
- [ ] PR merged to `main`

lubomirw:
Splits the async-status-polling slice of OmniTrustILM/ilm#110 out of `feat/authority-v3`.

Adds the machinery to poll a connector for async (HTTP 202) certificate operations (issue/renew/revoke/register), applies terminal transitions through the certificate state machine, **and activates it**: the 202 paths (`transitionToPendingIssue` / `transitionToPendingRevoke`) now schedule a due-now poll row via `CertificateStatusPollWriter.schedule`, so the sweep has rows to act on. Register's own 202 path lands with the register PR #1603.

## Design change from the issue

#1602 originally specced broker-driven backoff (RabbitMQ `x-delay` on a `rabbitmq_delayed_message_exchange`). That plugin is not provisioned in any environment, needs a custom broker image, and is RabbitMQ-only (no Azure Service Bus path). After analysis we switched to a **DB due-time table + scheduled sweep**: broker-portable, no plugin/custom image, the DB row is the source of truth, and it reuses the existing scheduled-sweep + advisory-lock pattern (cf. `SigningRecordOutboxDrainer`). The spec (`docs/superpowers/.../core/design.md` §4.5) is reconciled to match.

## What's added

- `certificate_status_poll` due-time table (Flyway `V202606181000`) — one in-flight poll per certificate (`UNIQUE(certificate_uuid)`, FK `ON DELETE CASCADE`); entity + repository + `CertificateStatusPollWriter`.
- `CertificateStatusPollClaimer` (`@Transactional`): acquires a cluster-wide advisory lock (`pg_try_advisory_xact_lock`), reads the due batch, and advances `attempt`/`next_poll_at` by the backoff curve — database only.
- `CertificateStatusPollSweeper` (`@Scheduled`, 10s default): sends the claimed poll messages **outside** the lock/transaction, so no broker round-trip pins the lock or a DB connection.
- `CertificateStatusPollListener`: polls the connector outside any transaction; on IN_PROGRESS leaves the row for the next sweep; on the last attempt times out; on COMPLETED/FAILED applies the terminal transition under a pessimistic lock and deletes the poll row.
- Producer / message / endpoint config / `StatusPollProperties` (the backoff curve now feeds `next_poll_at`; the producer no longer sets `x-delay`).

## Review focus (from #1602), as built

- **Backoff / max-attempts:** the curve drives `next_poll_at`; the listener times out at `attempt + 1 >= maxAttempts`.
- **Terminal transition via the state machine**, with meta written post-commit so a meta failure does not roll back the transition.
- **No DB transaction held across the connector call:** claim (lock + DB writes) is separated from the connector poll and the JMS sends.
- **State divergence:** COMPLETED issue/renew with no certificate content → FAILED (not ISSUED-with-empty-cert); a failed/timed-out revoke clears the pending-revoke fields.

## HA

`@Scheduled` fires on every replica; a transaction-scoped advisory lock keeps exactly one sweeper (auto-released on commit/rollback/crash). The idempotent listener (pessimistic re-read + re-assert pending) and `scheduleIfAbsent` (`INSERT … ON CONFLICT DO NOTHING`) are correctness backstops.

## ⚠️ Deploy prerequisites (separate repos)

These are needed before polling is **activated** (#1603); they do not block this additive slice:

- Broker topology needs a **plain durable `provider.status-poll` queue + binding** (no plugin, no delayed exchange): OmniTrustILM/helm-charts#305, OmniTrustILM/development-environment#21.
- New env overrides to wire if non-default: `PROVIDER_STATUS_POLL_SWEEP_INTERVAL_MS` / `_BATCH_SIZE` / `_MAX_BATCHES_PER_RUN`, `BROKER_QUEUE_PROVIDER_STATUS_POLL`, `BROKER_ROUTINGKEY_PROVIDER_STATUS_POLL`.

Reviewed locally (Copilot CLI + a guided pass, two cycles converged). 20 poll-related unit tests green.

Closes #1602.


src/main/java/com/otilm/core/dao/repository/CertificateStatusPollRepository.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.UUID;
+
+@Repository
+public interface CertificateStatusPollRepository extends JpaRepository<CertificateStatusPoll, UUID> {
+
+    /**
+     * Returns rows due for polling ({@code next_poll_at} reached), soonest-due first. No row locking:
+     * cross-node mutual exclusion is provided by the cluster-wide advisory lock held by
+     * {@link CertificateStatusPollSweeper}, so a plain ordered read is enough.
+     */
```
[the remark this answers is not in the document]
lubomirw: Fixed in f052b7b — dropped the messaging-layer import from the DAO and describe the sweep in prose, so the repository no longer references the JMS layer.

src/main/java/com/otilm/core/messaging/jms/listeners/poll/CertificateStatusPollSweeper.java
```diff
@@ -0,0 +1,71 @@
...
+        int enqueued = 0;
+        int batchesRun = 0;
+        int batchCount;
+        do {
+            List<CertificateStatusPollMessage> due = pollClaimer.claimDueBatch(batchSize);
+            batchCount = due.size();
+            for (CertificateStatusPollMessage message : due) {
+                pollProducer.produceMessage(message);
+                enqueued++;
+            }
```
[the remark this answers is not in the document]
lubomirw: Fixed in f052b7b — each send is now wrapped in a per-message try/catch, so one broker hiccup no longer aborts the rest of the batch. The row is already rescheduled, so it is re-enqueued when next due.

src/main/java/com/otilm/core/messaging/jms/configuration/StatusPollProperties.java
```diff
@@ -0,0 +1,29 @@
...
+        PollSchedule defaults,
+        Map<CertificateOperation, PollSchedule> byKind
+) {
+    public record PollSchedule(List<Duration> delays, int maxAttempts) {
+        public Duration delayFor(int attempt) {
+            if (attempt <= 0) return delays.get(0);
+            int idx = Math.min(attempt - 1, delays.size() - 1);
+            return delays.get(idx);
+        }
+    }
```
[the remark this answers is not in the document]
lubomirw: Deferred. Adding @Validated/@NotEmpty here would require seeding a provider.status-poll curve into every context-loading test profile (the defaults ship only in the main application.yml, and the context tests don't exercise the curve), which risks re-breaking unrelated tests as just happened with the spring.messaging binding. The shipped defaults are well-formed; tracking startup-time validation as a follow-up hardening rather than expanding this slice.

src/main/java/com/otilm/core/messaging/jms/listeners/poll/PollJmsEndpointConfig.java
```diff
@@ -0,0 +1,42 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.context.annotation.Profile;
+import org.springframework.jms.config.SimpleJmsListenerEndpoint;
+import org.springframework.retry.support.RetryTemplate;
+import org.springframework.stereotype.Component;
+
+
+@Component
+@Profile("!test")
+public class PollJmsEndpointConfig extends AbstractJmsEndpointConfig<CertificateStatusPollMessage> {
```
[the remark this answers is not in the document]
lubomirw: Not gating, by decision. The broker-provisioning PRs (helm-charts#305, development-environment#21) will be merged and rolled out before this slice is deployed, so the provider.status-poll queue exists when the listener starts. The sweep is one indexed SELECT on an empty table every 10s behind a single-node advisory lock (negligible), and nothing inserts poll rows until the activation slice — so there is no work to do regardless. A default-off flag would only add coupling (activation must remember to flip it) without buying anything once broker-first ordering is guaranteed.

src/main/java/com/otilm/core/messaging/jms/listeners/poll/CertificateStatusPollSweepScheduler.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Owns the periodic trigger that drives {@link CertificateStatusPollSweeper}. The sweeper holds all
+ * advisory-lock/batch orchestration; this class holds only the schedule and the {@code scheduled-tasks.enabled}
+ * gate. Keeping them apart mirrors the signing-record sweepers and lets the sweeper be unit-tested without
+ * standing up the scheduling machinery.
+ */
+@Component
+@ConditionalOnProperty(value = "scheduled-tasks.enabled", matchIfMissing = true, havingValue = "true")
+public class CertificateStatusPollSweepScheduler {
```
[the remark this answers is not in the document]
lubomirw: Not gating, by decision. The broker-provisioning PRs (helm-charts#305, development-environment#21) will be merged and rolled out before this slice is deployed, so the provider.status-poll queue exists when the listener starts. The sweep is one indexed SELECT on an empty table every 10s behind a single-node advisory lock (negligible), and nothing inserts poll rows until the activation slice — so there is no work to do regardless. A default-off flag would only add coupling (activation must remember to flip it) without buying anything once broker-first ordering is guaranteed.

src/main/resources/application.yml
```diff
@@ -64,6 +64,31 @@ discovery:
     sleep-time-ms: ${DISCOVERY_PROVIDER_SLEEP_TIME_MS:5000}
     max-wait-time-seconds: ${DISCOVERY_PROVIDER_MAX_WAIT_TIME_SECONDS:21600}
 
+# Authority-provider async (HTTP 202) operation status polling
+provider:
+  status-poll:
+    # How often each node scans the certificate_status_poll due-time table (one node sweeps at a time).
+    sweep-interval-ms: ${PROVIDER_STATUS_POLL_SWEEP_INTERVAL_MS:10000}
+    sweep-batch-size: ${PROVIDER_STATUS_POLL_SWEEP_BATCH_SIZE:200}
+    sweep-max-batches-per-run: ${PROVIDER_STATUS_POLL_SWEEP_MAX_BATCHES_PER_RUN:10}
```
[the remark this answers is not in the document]
lubomirw: Not gating, by decision. The broker-provisioning PRs (helm-charts#305, development-environment#21) will be merged and rolled out before this slice is deployed, so the provider.status-poll queue exists when the listener starts. The sweep is one indexed SELECT on an empty table every 10s behind a single-node advisory lock (negligible), and nothing inserts poll rows until the activation slice — so there is no work to do regardless. A default-off flag would only add coupling (activation must remember to flip it) without buying anything once broker-first ordering is guaranteed.

src/test/java/com/otilm/core/messaging/jms/listeners/JmsEndpointConfigTest.java
```diff
@@ -37,7 +37,7 @@
 @MockitoSettings(strictness = Strictness.LENIENT)
 class JmsEndpointConfigTest {
 
-    private static final String EXCHANGE = "ilm";
+    private static final String EXCHANGE = "czertainly";
```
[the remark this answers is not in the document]
lubomirw: Added in c008986 — JmsEndpointConfigTest now has a PollEndpointConfigTests nested class asserting the RabbitMQ (/queues/provider.status-poll, no subscription/selector) and Service Bus (topic destination + subscription + routingKey selector) wiring for the poll endpoint.

src/main/java/com/otilm/core/messaging/jms/listeners/CertificateStatusPollListener.java
```diff
@@ -0,0 +1,344 @@
...
+
+    /**
+     * Sync-path equivalence for a completed issue/renew: parse + persist cert content + ISSUE event.
+     * {@code issueRequestedCertificate} sets state=ISSUED via prepareIssuedCertificate (matching the sync v2
+     * path). Meta is passed empty here and written after commit so a meta-persistence failure does not roll
+     * back the transition — the connector accepted COMPLETED, so state must reflect that.
+     */
+    private void persistIssuedCertificate(Certificate locked, CertificateOperation op, StatusPollResult status) {
+        try {
+            certificateService.issueRequestedCertificate(locked.getUuid(), status.certificateData(), List.of());
```
lubomirw, line 226: **~ Minor** | Correctness

Persist failure on a COMPLETED issue/renew rethrows and never deletes the poll row.

The connector returned COMPLETED, but if `issueRequestedCertificate` throws (e.g. parse error, `AlreadyExistException` on redelivery), this rethrows -> tx rolls back -> poll row survives -> next sweep re-polls -> connector returns COMPLETED again -> same throw. The cert burns every remaining attempt stuck in `PENDING_ISSUE`, then times out to `FAILED` even though upstream succeeded. A deterministic persist failure cannot be recovered by retry. Consider: on a non-transient persist error, transition straight to `terminalFailureState()` with a distinct reason (like the `isTerminalPollError` path) rather than looping to timeout.

src/main/java/com/otilm/core/messaging/jms/listeners/CertificateStatusPollListener.java
```diff
@@ -0,0 +1,344 @@
...
+    private void applyTerminalTransition(Certificate cert, CertificateOperation op, StatusPollResult status) {
+        // Holders for post-commit side effects run OUTSIDE the lock so a slow connector doesn't extend the
+        // row lock duration. applied distinguishes the no-op drop (state no longer pending) from a real
+        // transition so the post-commit blocks are skipped on a drop.
+        CertificateRevocationFinalizer.KeyCleanup[] revokeCleanupHolder =
+                { CertificateRevocationFinalizer.KeyCleanup.NONE };
+        boolean[] applied = {false};
+
+        transactionHandler.runInNewTransaction(() -> {
+            Certificate locked = certificateRepository.findAndLockWithAssociationsByUuid(cert.getUuid())
```
lubomirw: **~ Minor** | Performance

`findAndLockWithAssociationsByUuid`'s entity graph loads `raProfile` but not `raProfile.authorityInstanceReference`. The REVOKE finalization path (`prepareRevokeFinalization` -> `getAuthorityInstanceReference().getConnectorUuid()`) therefore triggers an extra lazy-load query while the pessimistic row lock is held. Correct (session is open), but it widens the lock window. Consider adding `raProfile.authorityInstanceReference` to the graph, or reusing the eagerly-loaded `cert` for the connector UUID where possible.

src/main/java/com/otilm/core/messaging/model/CertificateStatusPollMessage.java
```diff
@@ -0,0 +1,13 @@
+package com.otilm.core.messaging.model;
+
+import com.otilm.api.model.core.auth.Resource;
+import com.otilm.core.service.handler.authority.CertificateOperation;
+
+import java.util.UUID;
+
+public record CertificateStatusPollMessage(
+        Resource resourceType,
```
lubomirw, line 16: **. Nitpick** | Maintainability

`resourceType` is always `Resource.CERTIFICATE` and is never read by `CertificateStatusPollListener` (it uses `resourceUuid`/`op`/`attempt`). If it is not needed for routing/future polymorphism, dropping it simplifies the message contract; if it is intentional, a one-line comment would clarify.

src/main/java/com/otilm/core/dao/entity/CertificateStatusPoll.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Column(name = "attempt", nullable = false)
+    private int attempt;
+
+    @Column(name = "next_poll_at", nullable = false)
+    private OffsetDateTime nextPollAt;
+
+    @CreationTimestamp
+    @Column(name = "i_cre", nullable = false, updatable = false)
+    private OffsetDateTime created;
```
lubomirw, line 47: **. Nitpick** | Maintainability

`@CreationTimestamp created` is never exercised: rows are inserted via the native `scheduleIfAbsent` query (which relies on the `i_cre DEFAULT now()`), and there is no JPA `persist` path for this entity. The annotation is effectively dead. Harmless, but worth a note or removal to avoid implying a JPA write path that does not exist.

src/test/java/com/otilm/core/messaging/jms/producers/CertificateStatusPollProducerTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private CertificateStatusPollProducer producer;
+
+    @BeforeEach
+    void setUp() {
+        when(messagingProperties.produceDestinationProviderStatusPoll()).thenReturn("/exchanges/ilm/provider.status-poll");
+
+        MessagingProperties.RoutingKey routingKey = new MessagingProperties.RoutingKey(
+                "actions", "audit-logs", "event", "notification", "scheduler",
+                "validation", "time-quality-config-request", "time-quality-config",
```
lubomirw: **. Nitpick** | Testing

The `RoutingKey` fixture uses hyphenated values (`time-quality-config-request`, `time-quality-config`) while the real config and the other test fixtures use dotted keys (`time-quality.config-request`). Test-only, but aligning with the dotted convention avoids confusion for the next reader.

lubomirw (commented):
Code review: 5 issue(s) found (all advisory). Strong slice — REQUIRES_NEW locked re-reads, advisory-lock claim separated from connector call, idempotent listener, broad test coverage. Notes below are minor/nitpick polish.

src/main/java/com/otilm/core/messaging/jms/listeners/CertificateStatusPollListener.java
lubomirw: Fixed in 65017ffa3. The persist now classifies the failure: a transient `TransientDataAccessException` still propagates (so it retries on the next poll), but a deterministic failure (parse error, already-exists on redelivery, …) throws an internal `DeterministicPersistException` that the caller catches and resolves straight to `terminalFailureState()` in a fresh transaction (the locked tx rolled back) with reason "completed at the authority but the certificate could not be persisted; reconcile manually" — no more looping to timeout. Tests: completedIssueDeterministicPersistFailureFailsFast / completedIssueTransientPersistFailurePropagatesForRetry.

src/main/java/com/otilm/core/messaging/jms/listeners/CertificateStatusPollListener.java
lubomirw: Fixed in 65017ffa3. Added `raProfile.authorityInstanceReference` to `findAndLockWithAssociationsByUuid`'s @EntityGraph, so the revoke finalization's connector-UUID access is eagerly loaded in the lock query instead of lazy-loading under the pessimistic lock. Additive for the operator revoke/cancel callers too.

src/main/java/com/otilm/core/messaging/model/CertificateStatusPollMessage.java
lubomirw: Kept + documented in 65017ffa3 — it's a deliberate discriminator so other provider interfaces can share the provider.status-poll queue for their own resources later without changing the message contract. Added a Javadoc saying so.

src/main/java/com/otilm/core/dao/entity/CertificateStatusPoll.java
lubomirw: Fixed in 65017ffa3. Dropped the misleading @CreationTimestamp and made `created` a read-only mapping (`insertable=false, updatable=false`) — the row inserts via the native scheduleIfAbsent upsert and the i_cre column has a DB `DEFAULT now()`, so JPA only reads it now.

src/test/java/com/otilm/core/messaging/jms/producers/CertificateStatusPollProducerTest.java
lubomirw: Fixed in 65017ffa3 — the routing-key fixture now uses the dotted `time-quality.config-request` / `time-quality.config` / `time-quality.results` to match the real config and the other fixtures.

lubomirw:
Folded in the insert-on-202 activation wiring (commit c98ecc4c4): `transitionToPendingIssue` / `transitionToPendingRevoke` now enqueue a due-now poll row via `CertificateStatusPollWriter.schedule` (best-effort, idempotent). S4 now ships **activated end-to-end** — the scheduled sweep gets rows for the existing async issue/renew/rekey/revoke paths, instead of running against an empty table. Re-reviewing the combined diff.

lubomirw:
Both review passes — Copilot and an independent reviewer — returned **SHIP** on the activation wiring (best-effort 202 scheduling; `REQUIRES_NEW` runs after the already-committed `PENDING_*` transition under the `NOT_SUPPORTED` callers; `REKEY→RENEW` correct; all four async paths covered). Applied the one consistency nit: the poll timestamp now uses `OffsetDateTime.now(ZoneOffset.UTC)`, matching the sweeper/claimer.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1050,6 +1064,25 @@ private void transitionToPendingRevoke(Certificate certificate, ClientCertificat
...
 
+    /**
+     * Best-effort enqueue of an async status-poll row (due now) after a connector accepts an operation
+     * with HTTP 202. The certificate is already committed to its {@code PENDING_*} state — per the
+     * state-divergence rule a scheduling failure must NOT roll that back, so we log and continue. The
+     * write is idempotent on the certificate UUID, so a duplicate (e.g. a retried request) is a no-op.
+     */
+    private void scheduleStatusPoll(Certificate certificate, CertificateOperation operation) {
+        try {
+            pollWriter.schedule(certificate.getUuid(), operation, OffsetDateTime.now(ZoneOffset.UTC));
```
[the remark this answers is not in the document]
lubomirw: Intentional — the activation was deliberately folded into this PR (commit e75d913) so S4 ships activated end-to-end rather than as a separate micro-PR (the sweep machinery alone would otherwise be dead code). PR description updated to match. The scheduling is best-effort and idempotent, and runs in a REQUIRES_NEW transaction after the already-committed PENDING_* transition (callers are NOT_SUPPORTED), so a scheduling failure cannot roll back the accepted transition. An independent review pass and Copilot's latest review both returned SHIP. Register's own 202 path lands with #1603.

src/main/java/com/otilm/core/messaging/jms/listeners/CertificateStatusPollListener.java
```diff
@@ -0,0 +1,397 @@
...
+        // Resolved — by us or by a racing actor — so stop polling it either way.
+        pollWriter.delete(cert.getUuid());
+        if (!resolution.applied()) {
+            return;
+        }
+        // Post-commit, outside the lock/tx: best-effort key destruction, then the meta write (a meta failure
+        // must not roll back the committed transition — state-divergence rule).
+        revocationFinalizer.destroyKeyIfRequested(resolution.keyCleanup(), cert.getUuid());
+        updateMetaAfterCommit(cert, op, status);
+    }
```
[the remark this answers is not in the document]
lubomirw: Fixed in ef072d803 — the post-commit side effects (key destruction + meta) now run **before** the poll-row delete, and the delete is guarded (log + continue). A transient delete failure can no longer skip key destruction for a completed revoke: the row survives and the next poll (cert no longer pending) drops it.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1050,6 +1064,25 @@ private void transitionToPendingRevoke(Certificate certificate, ClientCertificat
         logger.info("Certificate {} transitioned to PENDING_REVOKE", certificate.getUuid());
     }
 
+    /**
+     * Best-effort enqueue of an async status-poll row (due now) after a connector accepts an operation
+     * with HTTP 202. The certificate is already committed to its {@code PENDING_*} state — per the
+     * state-divergence rule a scheduling failure must NOT roll that back, so we log and continue. The
+     * write is idempotent on the certificate UUID, so a duplicate (e.g. a retried request) is a no-op.
+     */
```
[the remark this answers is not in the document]
lubomirw: The Javadoc is accurate as written. The four callers — `issueCertificateAction`, `renewCertificateAction`, `rekeyCertificateAction`, `revokeCertificateAction` — are annotated `@Transactional(propagation = NOT_SUPPORTED)`, which overrides the class-level `@Transactional`. So `certificateRepository.save()` commits in its own transaction **before** `scheduleStatusPoll` runs; the PENDING_* state is already committed. Clarified the Javadoc in ef072d803 to name NOT_SUPPORTED explicitly so it isn't re-flagged.

src/main/java/com/otilm/core/messaging/jms/listeners/CertificateStatusPollListener.java
```diff
@@ -0,0 +1,387 @@
...
+                return;
+            }
+            // A failed revoke returns to ISSUED — clear pending-revoke params (see applyResolvedState).
+            if (op == CertificateOperation.REVOKE) {
+                revocationFinalizer.clearPendingRevokeFields(locked);
+            }
+            stateMachine.transition(locked, op.terminalFailureState(), null, reason);
+        });
+        // Resolved (failed, or already resolved by a racing actor) — stop polling it.
+        pollWriter.delete(cert.getUuid());
```
klaraf755: `applyTerminalTransition` wraps its delete in a try/catch with a warn log, but applyFailure calls pollWriter.delete bare. Wrap the delete the same way as in applyTerminalTransition.

src/main/java/com/otilm/core/messaging/jms/listeners/CertificateStatusPollListener.java
```diff
@@ -0,0 +1,387 @@
...
+        }
+
+        if (completed && op == CertificateOperation.REVOKE) {
+            // destroyKey is captured here and run post-commit by the finalizer (outside the lock).
+            CertificateRevocationFinalizer.KeyCleanup cleanup = revocationFinalizer.prepareRevokeFinalization(locked);
+            stateMachine.transition(locked, op.terminalSuccessState(), null, reasonFor(op, status));
+            return cleanup;
+        }
+
+        // Failure paths (and the COMPLETED-REGISTER success). A failed/cancelled REVOKE returns the cert to
```
klaraf755: The last block handles both failure cases and REGISTER COMPLETED, with a single comment trying to explain both. At minimum the comment should be split.

src/main/java/com/otilm/core/messaging/jms/listeners/CertificateStatusPollListener.java
```diff
@@ -0,0 +1,387 @@
...
+            return;
+        } catch (RuntimeException e) {
+            // Add cert/op context — the JMS adapter logs only the bare exception. The rollback left the poll
+            // row, so the sweep retries (a transient blip recovers; a should-not-happen error retries to maxAttempts).
+            throw new IllegalStateException(
+                    "Async " + op + " terminal transition for cert " + cert.getUuid()
+                            + " did not apply; poll row left for the sweep to retry", e);
+        }
+
+        if (resolution.applied()) {
```
klaraf755, line 162: When `resolution.applied()` is false, the code still deletes the poll row. It's correct because the unique constraint means at most one row per cert UUID exists regardless of which operation put it there — but a future reader will find it non-obvious. Worth a comment.

src/main/java/com/otilm/core/messaging/jms/listeners/CertificateStatusPollListener.java
lubomirw: Fixed in e21951e98 — extracted a shared `stopPolling(cert, op)` helper, so `applyFailure` now guards its poll-row delete the same way `applyTerminalTransition` does.

src/main/java/com/otilm/core/messaging/jms/listeners/CertificateStatusPollListener.java
lubomirw: Fixed in e21951e98 — split into a block-scope comment (failures → failure state; a COMPLETED REGISTER → REGISTERED) and a separate comment on the failed-revoke field clearing.

src/main/java/com/otilm/core/messaging/jms/listeners/CertificateStatusPollListener.java
lubomirw: Fixed in e21951e98 — added a one-line note that deleting on the race-lost (not-applied) branch is safe because the unique constraint means one poll row per certificate.

klaraf755 (approved):

PR #1647, opened 2026-06-23
Add integration test for TSP signing profile timestamp flow

### LINKED ISSUE #1620: Introduce TspProtocolFlowITest end-to-end integration test for the RFC 3161 TSP flow
## Description
Add an end-to-end integration test (`TspProtocolFlowITest`, under `src/test/java/com/otilm/core/api/tsp/integration/`) covering the full RFC 3161 Timestamp Protocol production path through the service/controller layer.

The test extends `BaseSpringBootTest` and exercises the complete flow:
1. Builds shared infrastructure — connector, token instance, token profile, cryptographic key, TSA certificate, signing profile, TSP profile.
2. Constructs a TSP request with BouncyCastle `TimeStampRequestGenerator`.
3. Invokes `TspControllerImpl#timestamp`.
4. Parses the RFC 3161 `TimeStampResponse` and asserts PKI status `GRANTED` with a SHA-256 imprint.

External connectors (signer, signature-formatter) are stubbed with **WireMock**, including a `ResponseDefinitionTransformerV2` that signs on the fly with the registered private key so the produced token verifies cryptographically.

Coverage is parameterized over all supported signing algorithms — **RSA, ECDSA, FALCON-1024, ML-DSA-65, SLH-DSA-SHA2-128F** — in two variants:
- `withoutSignatureValidation` — asserts GRANTED + SHA-256 imprint.
- `withSignatureValidation` — same, plus verifies the cryptographic signature on the timestamp token (token-signature validation enabled).

Note: the test is present on feat/signing branch - it needs to be extracted and polished into its own PR.

## Definition of Done
- [ ] `TspProtocolFlowITest` added, extending `BaseSpringBootTest`, with WireMock-stubbed signer/formatter connectors.
- [ ] Parameterized (`@MethodSource`) over RSA, ECDSA, FALCON-1024, ML-DSA-65, SLH-DSA-SHA2-128F.
- [ ] Both `withoutSignatureValidation` and `withSignatureValidation` paths covered, asserting PKI status GRANTED + SHA-256 imprint.
- [ ] Key material pre-generated once in `setUp()` (no per-invocation key generation).
- [ ] Test is green in CI and self-contained (no external network dependency).

LukasNajman:


src/test/java/com/otilm/core/api/tsp/integration/TspSigningProfileControllerIntegrationTest.java
```diff
@@ -0,0 +1,444 @@
...
+    /**
+     * Branch unique to the signing-profile path: a signing profile that exists and is enabled but was never
+     * activated for a TSP profile ({@code tspProfileUuid == null}) must be rejected as BAD_REQUEST, not granted.
+     * The TSP-profile entry point cannot reach this state.
+     */
+    @Test
+    public void rejectsBadRequest_whenSigningProfileHasTspProtocolDisabled() throws Exception {
+        // given: an enabled signing profile that was NOT activated for any TSP profile
+        boolean validateTokenSignature = false;
+        String signingProfileName = createEnabledSigningProfileWithoutTspActivation("RSA", KeyAlgorithm.RSA, validateTokenSignature);
```
ivosh, line 292: **. Nitpick** | Documentation

The Javadoc for `rejectsBadRequest_whenSigningProfileHasTspProtocolDisabled` (src/.../TspSigningProfileControllerIntegrationTest.java:287-292) states the profile 'must be rejected as BAD_REQUEST'. The test asserts `PKIStatus.REJECTION` (line ~301) plus the exact statusString. In RFC 3161, BAD_REQUEST is a PKIFailureInfo bit, not a PKIStatus value (confirmed in TspResponseBuilder.buildRejection which always sets PKIStatus.rejection and encodes the failureInfo separately). The test never asserts the BAD_REQUEST failure-info bit, so it does not actually verify the BAD_REQUEST classification the Javadoc and method name claim — it only verifies a rejection with the expected message. The behaviour is distinguished adequately by the statusString, so this is documentation/intent drift rather than a wrong assertion; either soften the wording or add an assertion on the decoded failure-info bit. The same applies to the unknown-profile test's reliance on statusString alone.

src/test/java/com/otilm/core/api/tsp/integration/TspSigningProfileControllerIntegrationTest.java
```diff
@@ -0,0 +1,444 @@
...
+                            .withName("tsp-sp-key-" + spec.label().toLowerCase())
+                            .build());
+
+            // Uploading the leaf associates it (by public-key fingerprint) with the token-backed key
+            tsaCertificates.put(spec.keyAlgorithm(),
+                    trustedCa.issueTimestampingCertificate(keyPair, "CN=Test SP TSA " + spec.label()));
+        }
+    }
+
+    @AfterEach
```
ivosh, line 218: **. Nitpick** | Maintainability

This file's tearDown (line 216-224) null-guards both mock fields, and setUp carries a comment (line 151-153) explaining the assign-then-stub ordering exists so a throwing stub step still leaves a started server reachable for tearDown to stop. The sibling TspControllerIntegrationTest tearDown (line 206-210) calls `cryptographyProviderMock.stop()` directly with no guard and chains start+stub in one expression (line 146-149), so a stub failure there leaks the port and masks the real error with an NPE — exactly what this file's comment describes avoiding. The hardening here is genuinely better; the smell is that two sibling tests that should mirror each other now diverge silently. Either backport the guard to the sibling or this defensive ordering will read as unexplained inconsistency to the next reader.

ivosh (dismissed):
Nice PR. Two nit-picks, non-blocking.

src/test/java/com/otilm/core/api/tsp/integration/TspSigningProfileControllerIntegrationTest.java
LukasNajman: Aligned between tests

src/test/java/com/otilm/core/api/tsp/integration/TspSigningProfileControllerIntegrationTest.java
LukasNajman: BAD_REQUEST asserrted

ivosh (approved):

PR #1655, opened 2026-06-23
Validate request fieldMapping and exempt value-source attrs from callback

### LINKED ISSUE #1629: Validate fieldMapping at registration and gate by context
### Description
Validate a request attribute's `fieldMapping` when its definition is created/updated, and exempt value-source attributes from the connector-callback requirement. Part of #1572 (epic OmniTrustILM/ilm#131).

### Scope — what ships
- `AttributeEngine.validateFieldMapping` / `validateMappedField` / `isRequestOperation`, called from `updateDataAttributeDefinition` — objectType present, fields coherent, OIDs resolve; ignore-and-warn outside request scopes (backward compatible).
- Relax `validateResourceAttributeProperties` so a `valueSource` (COLLECTION / STATIC_LIST) attribute is exempt from the connector-callback requirement (prerequisite for the COLLECTION provider, tracked under ilm#132).

### Dependencies
- interfaces #730 only — SNAPSHOT-first. Independent of CO-0b.

### Acceptance Criteria
- `@SpringBootTest` tests; new-code coverage ≥80%; SonarCloud gate clean.

### Review focus
Backward-compat (no v2 / abstract-attribute change); correct request-context gating.

klaraf755:
Closes #1629


src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -1075,11 +1093,76 @@ private static void validateResourceAttributeProperties(BaseAttribute attribute,
...
+            throw new AttributeException("fieldMapping.fields must not be empty",
+                    attribute.getUuid(), attribute.getName(), attribute.getType(), connectorUuidStr);
+        for (MappedField field : fieldMapping.getFields())
+            validateMappedField(attribute, field, connectorUuidStr);
+    }
+
+    private static void validateMappedField(DataAttributeV3 attribute, MappedField field, String connectorUuidStr) throws AttributeException {
+        if (field.getFieldType() == null)
+            throw new AttributeException("fieldMapping field is missing fieldType",
+                    attribute.getUuid(), attribute.getName(), attribute.getType(), connectorUuidStr);
```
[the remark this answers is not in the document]
klaraf755: It gets called only from non null context, otherwise we do not need to validate, so mapped field does not have to be checked.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -1075,11 +1093,76 @@ private static void validateResourceAttributeProperties(BaseAttribute attribute,
...
+                if (ext.getExtensionOid() == null || ext.getExtensionOid().isBlank())
+                    throw new AttributeException("fieldMapping EXTENSION field is missing extensionOid",
+                            attribute.getUuid(), attribute.getName(), attribute.getType(), connectorUuidStr);
+                // Extension OID must be registered so the platform knows defaultCritical and valueEncoding
+                String extOid = ext.getExtensionOid();
+                SystemOid systemOid = SystemOid.fromOID(extOid);
+                if ((systemOid == null || systemOid.getCategory() != OidCategory.CERTIFICATE_EXTENSION)
+                        && OidHandler.getOidCache(OidCategory.CERTIFICATE_EXTENSION).get(extOid) == null)
+                    throw new AttributeException("fieldMapping EXTENSION OID '%s' is not registered in the OID registry".formatted(extOid),
+                            attribute.getUuid(), attribute.getName(), attribute.getType(), connectorUuidStr);
```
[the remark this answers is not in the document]
klaraf755: Fixed

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -1075,11 +1093,76 @@ private static void validateResourceAttributeProperties(BaseAttribute attribute,
...
+            }
+            default ->
+                    throw new IllegalStateException("Unexpected MappedField subtype: " + field.getClass().getSimpleName());
```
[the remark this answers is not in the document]
klaraf755: Fixed.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -1075,11 +1093,76 @@ private static void validateResourceAttributeProperties(BaseAttribute attribute,
         if (attribute instanceof DataAttribute dataAttribute && dataAttribute.getContentType() == AttributeContentType.RESOURCE) {
             if (attributeResource == null)
                 throw new AttributeException("Attribute with Resource Content Type is missing resource type in properties", attribute.getUuid(), attribute.getName(), attribute.getType(), connectorUuidStr);
-            if (!hasCallback)
+            // A Core-side valueSource (STATIC_LIST) resolves content without a connector callback
+            boolean hasCoreValueSource = attribute instanceof DataAttributeV3 v3
```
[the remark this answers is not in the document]
klaraf755: Done.

lubomirw (commented):
Code review: 3 issue(s) found. All non-blocking comments — code is correct, gating is sound, v2 backward-compat preserved.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -1075,11 +1092,80 @@ private static void validateResourceAttributeProperties(BaseAttribute attribute,
...
+                    attribute.getUuid(), attribute.getName(), attribute.getType(), connectorUuidStr);
+        switch (field) {
+            case RdnMappedField rdn -> {
+                if (rdn.getRdn() == null || rdn.getRdn().isBlank())
+                    throw new AttributeException("fieldMapping RDN field is missing rdn",
+                            attribute.getUuid(), attribute.getName(), attribute.getType(), connectorUuidStr);
+                // Dotted-decimal OIDs are always valid; short codes must resolve via OidHandler at build time
+                String rdnValue = rdn.getRdn();
+                boolean isOid = rdnValue.matches("\\d+(\\.\\d+)+");
+                if (!isOid && !OidHandler.getCodeToOidMap().containsKey(rdnValue))
```
lubomirw: **Minor** | Performance

`getCodeToOidMap()` rebuilds the entire code→OID reverse map on every RDN field.

`OidHandler.getCodeToOidMap()` iterates the full `RDN_ATTRIBUTE_TYPE` cache and allocates a fresh `HashMap` on each call. `validateMappedField` calls it once per RDN field, and `validateFieldMapping` runs on the issue/sign request-validation path (`validateUpdateDataAttributes`), not only at registration — so this repeats per request per field. Consider building the reverse map once per `validateFieldMapping` call (or per request) and passing it down, rather than recomputing inside the per-field loop.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -1075,11 +1092,80 @@ private static void validateResourceAttributeProperties(BaseAttribute attribute,
...
+                    throw new AttributeException("fieldMapping RDN code '%s' is not a known RDN code".formatted(rdnValue),
+                            attribute.getUuid(), attribute.getName(), attribute.getType(), connectorUuidStr);
+            }
+            case SanMappedField san -> {
+                if (san.getGeneralNameType() == null)
+                    throw new AttributeException("fieldMapping SAN field is missing generalNameType",
+                            attribute.getUuid(), attribute.getName(), attribute.getType(), connectorUuidStr);
+                if (san.getGeneralNameType() == GeneralNameType.OTHER_NAME
+                        && (san.getOtherNameOid() == null || san.getOtherNameOid().isBlank()))
```
lubomirw: **Minor** | Maintainability

OTHER_NAME `otherNameOid` is only checked for blank, not for OID format.

The EXTENSION branch validates its OID against the registry, and RDN accepts dotted-decimal via regex — but `otherNameOid` accepts any non-blank string. A typo'd OID (`1.2.x`) passes registration and fails later at cert build. Registry lookup isn't appropriate here (OtherName type-ids are arbitrary, not platform-registered extensions), but a dotted-decimal format check (same `\d+(\.\d+)+` pattern used for RDN) would catch typos at registration. Optional.

src/test/java/com/otilm/core/attribute/AttributeEngineTest.java
```diff
@@ -1125,4 +1142,303 @@ void testGetResourceSettableFields() throws AttributeException {
...
+            }
+        }
+
+        @BeforeEach
+        void setUp() {
+            // Re-seed OID entries that field-mapping tests depend on. The OID cache is loaded from DB
+            // on Spring context startup; entries added here are not in the DB, so they must be
+            // explicitly placed in the cache each time setUp() runs (before each test).
+            ensureOidCached(OidCategory.RDN_ATTRIBUTE_TYPE, "2.5.4.3", new OidRecord("Common Name", "CN", List.of()));
+            ensureOidCached(OidCategory.CERTIFICATE_EXTENSION, REGISTERED_EXTENSION_OID, new OidRecord("Test Extension", null, null));
```
lubomirw: **Nitpick** | Testing

Tests mutate the shared static `OidHandler` cache without cleanup.

`ensureOidCached` writes into the process-wide `OidHandler` cache. The fake extension `9.9.9.1` is never removed, so it leaks into `CERTIFICATE_EXTENSION` for any test class running later in the same JVM. Mitigated within this class by `@BeforeEach` re-seeding, and the scheduled `refreshCache` can't wipe mid-run (test `refresh-interval` is 1 year), so it's not flaky today — but an `@AfterEach`/`@AfterAll` `removeCachedOid` cleanup would keep the global cache clean for sibling tests.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
klaraf755: Changed to build the map before looping through attributes.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
klaraf755: Added check for OID to also other name.

src/test/java/com/otilm/core/attribute/AttributeEngineTest.java
klaraf755: Added the clean-up.

lubomirw (dismissed):

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -482,12 +493,13 @@ public void updateDataAttributeDefinitions(UUID connectorUuid, String operation,
     }
 
     public void updateAttributeDefinitionsWithCallback(UUID connectorUuid, List<? extends BaseAttribute> attributes) throws AttributeException {
+        Map<String, String> codeToOidMap = OidHandler.getCodeToOidMap();
```
lubomirw: **~ Minor** | Performance

`getCodeToOidMap()` is built eagerly here, but this path always passes `operation = null`. `validateFieldMapping` only runs when `isRequestOperation(operation)` is true, so the reverse map is never consumed on the callback path — it's rebuilt (full scan of the RDN OID cache) on every connector callback for nothing.

Consider computing the map lazily, only when a request-scoped `DataAttributeV3` with a non-null `fieldMapping` is actually present. Same applies to the eager build in `updateDataAttributeDefinitions` — most attributes won't have a `fieldMapping`.

lubomirw (commented):
Code review: 3 issue(s) found (all non-blocking).

src/test/java/com/otilm/core/attribute/AttributeEngineTest.java
```diff
@@ -1125,4 +1142,324 @@ void testGetResourceSettableFields() throws AttributeException {
...
+        static void initOidCache() {
+            for (OidCategory category : OidCategory.values()) {
+                if (OidHandler.getOidCache(category) == null) {
+                    OidHandler.cacheOidCategory(category, new HashMap<>());
+                }
+            }
+        }
+
+        @AfterAll
+        static void clearOidCache() {
```
lubomirw: **~ Minor** | Testing

`OidHandler`'s cache is a global static loaded once at context startup (`CustomOidEntryServiceImpl` constructor -> `refreshCache`). This `@AfterAll` overwrites *every* category with an empty map, discarding the DB-loaded system OIDs for the rest of the JVM run. Any test class that executes later in the same Surefire JVM and depends on the OID cache (e.g. `PlatformX500NameStyle`, `CustomOidEntry*` tests) can then see empty caches until the `@Scheduled` refresh happens to fire — an order-dependent flakiness risk.

Snapshot the prior cache contents in `@BeforeAll` and restore them in `@AfterAll`, or call `refreshCache()` to repopulate from DB, rather than clearing to empty maps.

src/test/java/com/otilm/core/attribute/AttributeEngineTest.java
```diff
@@ -1125,4 +1142,324 @@ void testGetResourceSettableFields() throws AttributeException {
...
+        @Test
+        void testFieldMapping_missingObjectType_throws() {
+            DataAttributeV3 attr = fieldMappingAttribute("fm_no_obj_type");
+            FieldMapping fm = new FieldMapping();
+            fm.setObjectType(null);
+            fm.setFields(List.of(rdnField("CN")));
+            attr.setFieldMapping(fm);
+
+            UUID connectorUuid = connectorAuthority.getUuid();
+            Assertions.assertThrows(AttributeException.class,
```
lubomirw: **~ Minor** | Testing

The negative tests assert only `AttributeException.class`, not the message. Per this repo's own guidance (CLAUDE.md, "Tests that look green but cover nothing"), a type-only assertion can pass for the wrong reason — e.g. an `AttributeException` raised by `validateAttributeDefinition` upstream rather than the `fieldMapping` rule under test. The green then proves nothing about the rule.

Use `assertThrows(...)` + assert a substring of the returned message (e.g. `"objectType is required"`, `"is not a known RDN code"`) so each test pins the failure to its specific rule.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
ivosh: Fixed in 9fd26bf5.

src/test/java/com/otilm/core/attribute/AttributeEngineTest.java
ivosh: Fixed in 9fd26bf5.

src/test/java/com/otilm/core/attribute/AttributeEngineTest.java
ivosh: Fixed in 9fd26bf5.

lubomirw (dismissed):

lubomirw (approved):

PR #1662, opened 2026-06-24
Fix SCEP profile edit: keep secrets, toggle challenge password, clear stale Intune config

### LINKED ISSUE #1663: SCEP profile edit requires re-entering challenge password and Intune application key
### Description

Editing a SCEP profile in Core (`ScepProfileServiceImpl.editScepProfile`) forces re-entry of write-only secrets that the Edit form does not prefill, so a normal edit either fails or silently destroys data:

- **Challenge password** — the guard `if (scepProfile.getChallengePassword() != null) scepProfile.setChallengePassword(request.getChallengePassword())` overwrites the stored password with blank/null when the request omits it, silently wiping the existing challenge password. It also can never set a password on a profile that previously had none.
- **Intune application key** — validation requires `intuneApplicationKey` whenever Intune is enabled and the setter assigns it unconditionally. Editing an Intune-enabled profile without retyping the key returns 422, and a blank value wipes the stored key.

Expected behaviour is write-only secret semantics (mirroring OAuth2 `clientSecret`): a blank or omitted value keeps the stored secret; a non-empty value updates it.

Part of OmniTrustILM/ilm#238. Fix in progress: OmniTrustILM/core#1662. This is the Core half of the frontend fix in OmniTrustILM/fe-administrator#1253 — the FE had to revert making the Intune key optional on edit until Core preserves omitted secrets (review thread: https://github.com/OmniTrustILM/fe-administrator/pull/1808#discussion_r3467089035).

### Steps to Reproduce

1. Create a SCEP profile with a challenge password (and optionally Intune enabled with an application key).
2. Open the profile and click Edit.
3. Leave the challenge password / Intune application key fields blank (the form does not prefill them).
4. Click Update.

### Expected Behavior

The profile updates successfully; the stored challenge password and Intune application key are preserved when their fields are left blank, and updated only when a new value is supplied.

### Actual Behavior

Editing an Intune-enabled profile without retyping the key fails with HTTP 422. Editing a profile with a challenge password while leaving the field blank silently wipes the stored challenge password.

lubomirw:
## Problem

Editing a SCEP profile mishandled its write-only secrets, which the Edit form does not prefill:

- **Challenge password** — a blank request silently wiped the stored password, and a profile that had none could never gain one on edit.
- **Intune application key** — required on every edit while Intune was enabled (→ 422), and a blank value wiped the stored key.
- There was **no way to remove** an optional challenge password, and **no signal** telling the client whether one is set.

Core half of the FE fix in OmniTrustILM/fe-administrator#1253 ([review thread](https://github.com/OmniTrustILM/fe-administrator/pull/1808#discussion_r3467089035)). Tracked by #1663, part of OmniTrustILM/ilm#238.

## Fix

**Write-only secrets, keep-on-omit.** Omitting the Intune key while Intune stays enabled keeps the stored key; a non-empty value updates it.

**Challenge password — tri-state `enableChallengePassword` toggle** (shared `applyChallengePassword` helper for create + edit):

| toggle | behavior |
|---|---|
| `null` (absent) | keep stored — legacy/opt-out, never wipes |
| `true` + value | set |
| `true` + blank | keep stored, or 400 if none stored |
| `false` | clear |

Nullable-by-design so existing clients that don't send the field never wipe a password. The toggle is **not** normalized like `enableIntune` (`Boolean.TRUE.equals`) — doing so would turn an absent toggle into a wipe; an isolated helper + Javadoc guard against that regression.

**Intune disabled → clear sub-config.** When `enableIntune=false`, tenant + appId + key are nulled so no stale secret lingers or gets silently reused on re-enable. (Matches prior behavior — old code already cleared Intune config when the flag was absent.)

**Response presence flag.** `ScepProfileDto.enableChallengePassword` = `challengePassword != null`, so the FE can initialize the switch. The password value itself remains write-only and is never returned.

## Backward compatibility

The toggle is purely opt-in. Clients that omit it get exactly the keep-on-omit behavior — no wipe, no 400 on passwordless profiles. The SCEP profile admin API has no generated SDK consumer (go-sdk is connector-only), and the field is `NOT_REQUIRED`, so codegen/TS clients omit it when unset rather than sending `false`.

## Dependency

Requires **OmniTrustILM/interfaces#743** (new DTO fields, `2.18.1-SNAPSHOT`) — must merge/publish before this PR's CI gate passes.

## Tests

`ScepProfileServiceTest` — keep-on-omit, set/keep/clear via toggle, reject-when-enabled-and-nothing-stored, null-toggle on passwordless profile (no wipe, no 400), Intune-disable clears config, response flag, and create-path rejection. **33/33 pass.**

Closes #1663
Part of OmniTrustILM/ilm#238

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
@@ -153,9 +153,12 @@ public class ScepProfileServiceImpl implements ScepProfileExternalService, ScepP
         scepProfile.setRequireManualApproval(false);
         scepProfile.setCaCertificateUuid(UUID.fromString(request.getCaCertificateUuid()));
         scepProfile.setIntuneEnabled(intuneEnabled);
-        scepProfile.setIntuneTenant(request.getIntuneTenant());
-        scepProfile.setIntuneApplicationId(request.getIntuneApplicationId());
-        scepProfile.setIntuneApplicationKey(request.getIntuneApplicationKey());
+        // Unset fields stay null on a fresh profile, so a disabled profile stores no stale Intune secret.
+        if (intuneEnabled) {
+            scepProfile.setIntuneTenant(request.getIntuneTenant());
+            scepProfile.setIntuneApplicationId(request.getIntuneApplicationId());
+            scepProfile.setIntuneApplicationKey(request.getIntuneApplicationKey());
+        }
         scepProfile.setRaProfile(raProfile);
 
         if (request.getCertificateAssociations() != null && !request.getCertificateAssociations().isEmpty()) {
--- a/src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
@@ -273,21 +273,6 @@ public class ScepProfileServiceImpl implements ScepProfileExternalService, ScepP
     }
 
     /**
-     * Applies the write-only challenge password according to the tri-state {@code enableChallengePassword} toggle.
-     * The toggle MUST be treated as keep-when-null: an absent toggle (legacy clients, or clients that do not send
-     * the field) must never wipe a stored password. Do NOT collapse it with {@code Boolean.TRUE.equals(...)} the way
-     * {@code enableIntune} is handled — that would turn a missing toggle into {@code false} and silently clear the
-     * stored secret.
-     *
-     * <ul>
-     *   <li>toggle {@code null} — set when a value is supplied, otherwise leave the entity untouched
-     *       (keep on edit, no password on create).</li>
-     *   <li>toggle {@code false} — clear the challenge password.</li>
-     *   <li>toggle {@code true} + non-blank value — set the new password.</li>
-     *   <li>toggle {@code true} + blank value — keep the stored password, or reject when none is stored.</li>
-     * </ul>
-     */
-    /**
      * Persists the Intune sub-config, shared by create and edit. When Intune is enabled the tenant and
      * application id are set from the request and the application key is treated as a write-only secret —
      * kept as stored when the request omits it (blank on a fresh entity means no key). When Intune is
@@ -312,6 +297,21 @@ public class ScepProfileServiceImpl implements ScepProfileExternalService, ScepP
         return request.getIntuneApplicationKey() != null && !request.getIntuneApplicationKey().isBlank();
     }
 
+    /**
+     * Applies the write-only challenge password according to the tri-state {@code enableChallengePassword} toggle.
+     * The toggle MUST be treated as keep-when-null: an absent toggle (legacy clients, or clients that do not send
+     * the field) must never wipe a stored password. Do NOT collapse it with {@code Boolean.TRUE.equals(...)} the way
+     * {@code enableIntune} is handled — that would turn a missing toggle into {@code false} and silently clear the
+     * stored secret.
+     *
+     * <ul>
+     *   <li>toggle {@code null} — set when a value is supplied, otherwise leave the entity untouched
+     *       (keep on edit, no password on create).</li>
+     *   <li>toggle {@code false} — clear the challenge password.</li>
+     *   <li>toggle {@code true} + non-blank value — set the new password.</li>
+     *   <li>toggle {@code true} + blank value — keep the stored password, or reject when none is stored.</li>
+     * </ul>
+     */
     private void applyChallengePassword(ScepProfile scepProfile, BaseScepProfileRequestDto request) {
         Boolean enable = request.getEnableChallengePassword();
         boolean valueProvided = request.getChallengePassword() != null && !request.getChallengePassword().isBlank();
--- a/src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
@@ -127,7 +127,9 @@ public class ScepProfileServiceImpl implements ScepProfileExternalService, ScepP
         }
 
         boolean intuneKeyProvided = request.getIntuneApplicationKey() != null && !request.getIntuneApplicationKey().isBlank();
-        if (intuneEnabled && (request.getIntuneTenant() == null || request.getIntuneApplicationId() == null || !intuneKeyProvided)) {
+        if (intuneEnabled && (request.getIntuneTenant() == null || request.getIntuneTenant().isBlank()
+                || request.getIntuneApplicationId() == null || request.getIntuneApplicationId().isBlank()
+                || !intuneKeyProvided)) {
             throw new ValidationException(ValidationError.create("Invalid Intune configuration. Intune tenant, application ID and application key are required when Intune is enabled"));
         }
 
@@ -199,7 +201,8 @@ public class ScepProfileServiceImpl implements ScepProfileExternalService, ScepP
         // The Intune application key is write-only: when Intune stays enabled and the request omits the key,
         // keep the stored one (the form does not prefill it). Requiring re-entry otherwise would 422 or wipe it.
         boolean intuneKeyProvided = request.getIntuneApplicationKey() != null && !request.getIntuneApplicationKey().isBlank();
-        if (intuneEnabled && (request.getIntuneTenant() == null || request.getIntuneApplicationId() == null
+        if (intuneEnabled && (request.getIntuneTenant() == null || request.getIntuneTenant().isBlank()
+                || request.getIntuneApplicationId() == null || request.getIntuneApplicationId().isBlank()
                 || (!intuneKeyProvided && scepProfile.getIntuneApplicationKey() == null))) {
             throw new ValidationException(ValidationError.create("Invalid Intune configuration. Intune tenant, application ID and application key are required when Intune is enabled (the application key may be omitted only if one is already stored)"));
         }
--- a/src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
@@ -126,8 +126,9 @@ public class ScepProfileServiceImpl implements ScepProfileExternalService, ScepP
             throw new ValidationException(ValidationError.create("CA Certificate is not acceptable as SCEP CA certificate for this profile"));
         }
 
-        if (intuneEnabled && (request.getIntuneTenant() == null || request.getIntuneApplicationId() == null || request.getIntuneApplicationKey() == null)) {
-            throw new ValidationException(ValidationError.create("Invalid Intune configuration. Missing Intune tenant and/or intune app identification"));
+        boolean intuneKeyProvided = request.getIntuneApplicationKey() != null && !request.getIntuneApplicationKey().isEmpty();
+        if (intuneEnabled && (request.getIntuneTenant() == null || request.getIntuneApplicationId() == null || !intuneKeyProvided)) {
+            throw new ValidationException(ValidationError.create("Invalid Intune configuration. Intune tenant, application ID and application key are required when Intune is enabled"));
         }
 
         RaProfile raProfile = null;
@@ -197,7 +198,7 @@ public class ScepProfileServiceImpl implements ScepProfileExternalService, ScepP
         boolean intuneKeyProvided = request.getIntuneApplicationKey() != null && !request.getIntuneApplicationKey().isEmpty();
         if (intuneEnabled && (request.getIntuneTenant() == null || request.getIntuneApplicationId() == null
                 || (!intuneKeyProvided && scepProfile.getIntuneApplicationKey() == null))) {
-            throw new ValidationException(ValidationError.create("Invalid Intune configuration. Missing Intune tenant and/or intune app identification"));
+            throw new ValidationException(ValidationError.create("Invalid Intune configuration. Intune tenant, application ID and application key are required when Intune is enabled (the application key may be omitted only if one is already stored)"));
         }
 
         attributeEngine.validateCustomAttributesContent(Resource.SCEP_PROFILE, request.getCustomAttributes());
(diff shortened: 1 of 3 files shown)


src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
```diff
@@ -189,7 +189,14 @@ public ScepProfileDetailDto editScepProfile(SecuredUUID uuid, ScepProfileEditReq
...
+        ScepProfile scepProfile = getScepProfileEntity(uuid);
+
+        // Secrets (Intune application key, challenge password) are write-only on edit: a blank or omitted
+        // value keeps the stored secret, mirroring OAuth2 clientSecret handling. The form does not prefill
+        // them, so requiring re-entry on every edit would 422 or silently wipe the existing value.
+        boolean intuneKeyProvided = request.getIntuneApplicationKey() != null && !request.getIntuneApplicationKey().isEmpty();
+        if (intuneEnabled && (request.getIntuneTenant() == null || request.getIntuneApplicationId() == null
+                || (!intuneKeyProvided && scepProfile.getIntuneApplicationKey() == null))) {
             throw new ValidationException(ValidationError.create("Invalid Intune configuration. Missing Intune tenant and/or intune app identification"));
         }
```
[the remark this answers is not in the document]
lubomirw: Fixed in 23ec2b5. The Intune validation message now names the required fields explicitly: "Intune tenant, application ID and application key are required when Intune is enabled (the application key may be omitted only if one is already stored)". The create path uses the same clarified message.

src/test/java/com/otilm/core/service/ScepProfileServiceTest.java
```diff
@@ -318,6 +318,44 @@ void testEditScepProfile() throws ConnectorException, AttributeException, NotFou
...
+
+        ScepProfileEditRequestDto request = new ScepProfileEditRequestDto();
+        request.setCaCertificateUuid(certificate.getUuid().toString());
+        request.setChallengePassword("newChallenge");
+
+        scepProfileService.editScepProfile(scepProfile.getSecuredUuid(), request);
+
+        ScepProfile updated = scepProfileRepository.findByUuid(scepProfile.getUuid()).orElseThrow();
+        Assertions.assertEquals("newChallenge", updated.getChallengePassword());
+    }
```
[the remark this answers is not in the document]
lubomirw: Added in 23ec2b5: `testEditScepProfile_updatesIntuneKeyWhenProvided` sets an Intune-enabled profile with a stored key, edits with a new key, and asserts the key changed. The keep-when-omitted side is covered by `testEditScepProfile_keepsSecretsWhenOmitted`.

src/test/java/com/otilm/core/service/ScepProfileServiceTest.java
```diff
@@ -318,6 +318,44 @@ void testEditScepProfile() throws ConnectorException, AttributeException, NotFou
...
+        request.setIntuneTenant("tenant");
+        request.setIntuneApplicationId("appId");
+        // challengePassword and intuneApplicationKey omitted (form does not prefill secrets)
+
+        scepProfileService.editScepProfile(scepProfile.getSecuredUuid(), request);
+
+        ScepProfile updated = scepProfileRepository.findByUuid(scepProfile.getUuid()).orElseThrow();
+        Assertions.assertEquals("originalChallenge", updated.getChallengePassword());
+        Assertions.assertEquals("originalKey", updated.getIntuneApplicationKey());
+    }
```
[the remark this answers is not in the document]
lubomirw: Added in 23ec2b5: `testEditScepProfile_keepsChallengePasswordWhenBlankValueAndNullToggle` sends an empty-string challengePassword (not just omitted) and asserts the stored password is kept, matching the documented blank=keep behavior.

src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
```diff
@@ -125,8 +126,9 @@ public ScepProfileDetailDto createScepProfile(ScepProfileRequestDto request) thr
             throw new ValidationException(ValidationError.create("CA Certificate is not acceptable as SCEP CA certificate for this profile"));
         }
 
-        if (intuneEnabled && (request.getIntuneTenant() == null || request.getIntuneApplicationId() == null || request.getIntuneApplicationKey() == null)) {
-            throw new ValidationException(ValidationError.create("Invalid Intune configuration. Missing Intune tenant and/or intune app identification"));
+        boolean intuneKeyProvided = request.getIntuneApplicationKey() != null && !request.getIntuneApplicationKey().isEmpty();
```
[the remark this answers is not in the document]
lubomirw: Fixed in 3f62f31 — switched to `isBlank()` so whitespace-only values are treated as blank (not persisted), matching the codebase credential-validation convention. Applied to all three spots (create Intune key, edit Intune key, challenge password). Added `testEditScepProfile_treatsWhitespaceChallengeAsBlankAndKeeps` to lock it.

src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
```diff
@@ -189,8 +194,14 @@ public ScepProfileDetailDto editScepProfile(SecuredUUID uuid, ScepProfileEditReq
             throw new ValidationException(ValidationError.create("CA Certificate is not acceptable as SCEP CA certificate for this profile"));
         }
 
-        if (intuneEnabled && (request.getIntuneTenant() == null || request.getIntuneApplicationId() == null || request.getIntuneApplicationKey() == null)) {
-            throw new ValidationException(ValidationError.create("Invalid Intune configuration. Missing Intune tenant and/or intune app identification"));
+        ScepProfile scepProfile = getScepProfileEntity(uuid);
+
+        // The Intune application key is write-only: when Intune stays enabled and the request omits the key,
+        // keep the stored one (the form does not prefill it). Requiring re-entry otherwise would 422 or wipe it.
+        boolean intuneKeyProvided = request.getIntuneApplicationKey() != null && !request.getIntuneApplicationKey().isEmpty();
```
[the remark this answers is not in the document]
lubomirw: Fixed in 3f62f31 — switched to `isBlank()` so whitespace-only values are treated as blank (not persisted), matching the codebase credential-validation convention. Applied to all three spots (create Intune key, edit Intune key, challenge password). Added `testEditScepProfile_treatsWhitespaceChallengeAsBlankAndKeeps` to lock it.

src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
```diff
@@ -271,6 +290,40 @@ private ScepProfileDetailDto updateAndMapDtoAttributes(ScepProfile scepProfile,
...
+     *       (keep on edit, no password on create).</li>
+     *   <li>toggle {@code false} — clear the challenge password.</li>
+     *   <li>toggle {@code true} + non-blank value — set the new password.</li>
+     *   <li>toggle {@code true} + blank value — keep the stored password, or reject when none is stored.</li>
+     * </ul>
+     */
+    private void applyChallengePassword(ScepProfile scepProfile, BaseScepProfileRequestDto request) {
+        Boolean enable = request.getEnableChallengePassword();
+        boolean valueProvided = request.getChallengePassword() != null && !request.getChallengePassword().isEmpty();
+
```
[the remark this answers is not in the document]
lubomirw: Fixed in 3f62f31 — switched to `isBlank()` so whitespace-only values are treated as blank (not persisted), matching the codebase credential-validation convention. Applied to all three spots (create Intune key, edit Intune key, challenge password). Added `testEditScepProfile_treatsWhitespaceChallengeAsBlankAndKeeps` to lock it.

src/test/java/com/otilm/core/service/ScepProfileServiceTest.java
```diff
@@ -318,6 +318,286 @@ void testEditScepProfile() throws ConnectorException, AttributeException, NotFou
         Assertions.assertNotNull(dto.getCertificateAssociations());
     }
 
+    @Test
```
ivosh, line 326: **~ Minor** | Testing

Every Intune edit test sets `enableIntune` explicitly to true or false; none omits it on a profile that currently has Intune enabled. Because the edit endpoint is `@PutMapping` (full-replace), an omitted `enableIntune` collapses to `false` and clears the stored tenant/appId/key — that's intended PUT semantics, not a defect, but it is currently unlocked by any test.

Add an edit test that omits `enableIntune` on an Intune-enabled stored profile and asserts the sub-config is cleared, to document and pin the PUT contract.

src/test/java/com/otilm/core/service/ScepProfileServiceTest.java
```diff
@@ -318,6 +318,286 @@ void testEditScepProfile() throws ConnectorException, AttributeException, NotFou
...
+        scepProfile.setChallengePassword(null);
+        scepProfileRepository.save(scepProfile);
+
+        ScepProfileEditRequestDto request = new ScepProfileEditRequestDto();
+        request.setCaCertificateUuid(certificate.getUuid().toString());
+        request.setEnableChallengePassword(true);
+        // no value and nothing stored -> reject
+
+        SecuredUUID uuid = scepProfile.getSecuredUuid();
+        Assertions.assertThrows(ValidationException.class,
```
ivosh: **~ Minor** | Testing

The new negative tests assert only `ValidationException.class`. `ValidationException` is thrown by several earlier branches (the CA-cert checks, the Intune-config check) before the challenge-password check, so a regression that made an earlier check fire would leave the test green while the intended branch is never reached — JaCoCo would then show those target lines uncovered.

Pin the assertion to the message, e.g. `assertThatThrownBy(...).isInstanceOf(ValidationException.class).hasMessageContaining("Challenge password is required")`, so the test proves which validation actually triggered.

src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
```diff
@@ -271,6 +290,40 @@ private ScepProfileDetailDto updateAndMapDtoAttributes(ScepProfile scepProfile,
...
+        if (enable == null) {
+            if (valueProvided) scepProfile.setChallengePassword(request.getChallengePassword());
+            return;
+        }
+        if (!enable) {
+            scepProfile.setChallengePassword(null);
+            return;
+        }
+        if (valueProvided) {
```
ivosh, line 327: **~ Minor** | Testing

`applyChallengePassword` is exercised from both create and edit, but the new tests only cover the create path's *reject* branch. No create test asserts that `enableChallengePassword=true` + a non-blank value actually persists the password on a fresh entity (`testAddScepProfile` sets a password but never asserts it was stored).

Add a create success test for the toggle-true-with-value path.

src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
```diff
@@ -147,13 +149,16 @@ public ScepProfileDetailDto createScepProfile(ScepProfileRequestDto request) thr
...
         scepProfile.setCaCertificateUuid(UUID.fromString(request.getCaCertificateUuid()));
         scepProfile.setIntuneEnabled(intuneEnabled);
-        scepProfile.setIntuneTenant(request.getIntuneTenant());
-        scepProfile.setIntuneApplicationId(request.getIntuneApplicationId());
-        scepProfile.setIntuneApplicationKey(request.getIntuneApplicationKey());
+        // Unset fields stay null on a fresh profile, so a disabled profile stores no stale Intune secret.
+        if (intuneEnabled) {
+            scepProfile.setIntuneTenant(request.getIntuneTenant());
+            scepProfile.setIntuneApplicationId(request.getIntuneApplicationId());
+            scepProfile.setIntuneApplicationKey(request.getIntuneApplicationKey());
```
ivosh: **. Nitpick** | Maintainability

ScepProfileServiceImpl.java:155-161 (create) and :230-243 (edit) duplicate the Intune sub-config persistence logic inline, while the parallel challenge-password logic was cleanly extracted into the well-documented applyChallengePassword(...) helper (:308). The two flows share the same tri-state intent (enabled vs disabled vs write-only key), and the blank-check expression `request.getIntuneApplicationKey() != null && !request.getIntuneApplicationKey().isBlank()` is copy-pasted at :129 and :201. Per Clean Code (small focused methods, DRY) and to match the helper-method design already adopted for the challenge password, consider an applyIntuneConfig(scepProfile, request, intuneEnabled) helper. Purely a consistency/readability nitpick — the inline form is correct and the existing pattern in this file mixes both styles, so this need not block the PR.

ivosh (commented):
Nice and clean PR. Just few minor comments to address.

src/test/java/com/otilm/core/service/ScepProfileServiceTest.java
lubomirw: Done in f6e83cb — added `testEditScepProfile_omittedEnableIntuneClearsStoredConfig`: stores an Intune-enabled profile, edits with `enableIntune` omitted entirely, and asserts tenant/appId/key are cleared, pinning the PUT full-replace contract.

src/test/java/com/otilm/core/service/ScepProfileServiceTest.java
lubomirw: Done in f6e83cb — the negative tests now capture the exception and assert the message: challenge-password rejects check `contains("Challenge password is required")`, Intune rejects check `contains("Invalid Intune configuration")`. A regression in an earlier CA/Intune branch can no longer leave them green while the intended branch is skipped.

src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
lubomirw: Done in f6e83cb — added `testCreateScepProfile_setsChallengePasswordWhenToggleTrue`: create with `enableChallengePassword=true` + a value, asserting the password is persisted on the fresh entity and the response presence flag is true.

src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
lubomirw: Done in f6e83cb — extracted `applyIntuneConfig(scepProfile, request, intuneEnabled)` (shared by create and edit) plus `isIntuneKeyProvided(request)` for the blank-check, mirroring the `applyChallengePassword` helper. The duplicated inline sub-config logic and the copy-pasted blank-check at the two validation sites are gone.

ivosh (approved):

PR #1666, opened 2026-06-25
TsaServiceAuthzTest + system-user guard tests

### LINKED ISSUE #1665: Backport signing-branch test coverage to main: TsaServiceAuthzTest + system-user guard tests
## Description
Backport test coverage that exists on the `feat/signing` branch but is missing on `main`, even though the production code it exercises already lives on `main`. Two areas are affected:

1. **`TsaServiceAuthzTest`** (new file under `src/test/java/com/otilm/core/signing/tsa/`) — an integration test that asserts **object-level OPA authorization scoping** for TSP timestamping. It verifies that `opaClient.checkResourceAccess` is queried with the correct TSP-profile `UUID` and `TIMESTAMP` action for both routes:
   - direct: `processTspRequestForTspProfile`
   - indirect: `processTspRequestForSigningProfile` (authorizes against the linked TSP profile)

   It also covers denial (`AccessDeniedException`) and disabled-profile rejection (`TspException` / `BAD_REQUEST`). `main`'s existing `TsaServiceImplUnitTest` only stubs a denial and never verifies *which* object/action OPA was asked about, and `TsaServiceImplTest` does not touch OPA at all — so this fills a real gap. The production `TsaServiceImpl` is identical on both branches.

2. **`TspProfileBasicCredentialServiceImplTest`** — two new tests for the system-user guard (which exists on `main` but is currently untested there):
   - `throwsValidationAndCreatesNoSecret_whenMappedToSystemUser` — rejects credential **create** when the mapped user resolves to a system user, before any vault secret is provisioned.
   - `throwsValidation_whenRemappedToSystemUser` — rejects credential **update** that remaps to a system user, with no secret rotation attempted.

Excluded from the backport (already covered on `main`): the `evictBySecretUuid` delete-path assertions (covered by `TspProfileBasicCredentialServiceImplEvictionTest`) and the `resolveForAuthentication` direct test (covered indirectly by `TspRouteResolverTest` / `TspAuthenticationFilterTest`).

Verified locally: `TsaServiceAuthzTest` 9/9 pass; `TspProfileBasicCredentialServiceImplTest` 16/16 pass (including the 2 new tests), 0 failures, 0 errors.

## Definition of Done
- [ ] `TsaServiceAuthzTest` ported to `main` and passing
- [ ] Two system-user guard tests added to `TspProfileBasicCredentialServiceImplTest` on `main` and passing
- [ ] No production-code changes required (test-only PR)
- [ ] PR merged to `main`

ivosh:
Test-only backport of coverage that exists on `feat/signing` but is missing on `main`, even though the production code it exercises already lives on `main`.

## What's ported
- **`TsaServiceAuthzTest`** (new) — integration test asserting **object-level OPA authorization scoping** for TSP timestamping: verifies `opaClient.checkResourceAccess` is queried with the correct TSP-profile `UUID` and `TIMESTAMP` action for both the direct (`processTspRequestForTspProfile`) and indirect (`processTspRequestForSigningProfile`, via the linked TSP profile) routes, plus denial (`AccessDeniedException`) and disabled-profile rejection (`TspException`/`BAD_REQUEST`). `main`'s existing `TsaServiceImplUnitTest` only stubs a denial and never verifies *which* object/action OPA was asked about; `TsaServiceImplTest` doesn't touch OPA. Production `TsaServiceImpl` is identical on both branches.
- **`TspProfileBasicCredentialServiceImplTest`** — two tests for the system-user guard (present but untested on `main`): reject credential **create** and **update** mapped to a system user, with no secret provisioned/rotated.

## Excluded (already covered on `main`)
- `evictBySecretUuid` delete-path assertions → `TspProfileBasicCredentialServiceImplEvictionTest`
- `resolveForAuthentication` direct test → covered by `TspRouteResolverTest` / `TspAuthenticationFilterTest`

## Verification
- `TsaServiceAuthzTest`: 9/9 pass
- `TspProfileBasicCredentialServiceImplTest`: 16/16 pass (incl. the 2 new), 0 failures / 0 errors
- No production-code changes.

Refs #1665

### DIFF
--- a/src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
+++ b/src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
@@ -166,7 +166,8 @@ class TsaServiceAuthzTest extends BaseSpringBootTest {
             denyTimestampForObject(tspProfileA.getUuid());
 
             // when / then
-            assertThatThrownBy(() -> tsaService.processTspRequestForTspProfile("tsp-a", aTspRequest().build()))
+            var request = aTspRequest().build();
+            assertThatThrownBy(() -> tsaService.processTspRequestForTspProfile("tsp-a", request))
                     .isInstanceOf(AccessDeniedException.class);
         }
 
@@ -248,7 +249,8 @@ class TsaServiceAuthzTest extends BaseSpringBootTest {
             denyTimestampForObject(linkedTspProfile.getUuid());
 
             // when / then
-            assertThatThrownBy(() -> tsaService.processTspRequestForSigningProfile("sp-indirect-denied", aTspRequest().build()))
+            var request = aTspRequest().build();
+            assertThatThrownBy(() -> tsaService.processTspRequestForSigningProfile("sp-indirect-denied", request))
                     .isInstanceOf(AccessDeniedException.class);
         }
 
--- a/src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
+++ b/src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
@@ -1,0 +1,282 @@
+package com.otilm.core.signing.tsa;
+
+import com.otilm.api.interfaces.core.tsp.error.TspException;
+import com.otilm.api.interfaces.core.tsp.error.TspFailureInfo;
+import com.otilm.api.model.client.signing.profile.scheme.ManagedSigningType;
+import com.otilm.api.model.client.signing.profile.scheme.SigningScheme;
+import com.otilm.api.model.client.signing.profile.workflow.SigningWorkflowType;
+import com.otilm.api.model.core.auth.Resource;
+import com.otilm.api.model.core.signing.SigningProtocol;
+import com.otilm.core.model.signing.SigningCertificateBuilder;
+import com.otilm.core.dao.entity.signing.SigningProfile;
+import com.otilm.core.dao.entity.signing.SigningProfileVersion;
+import com.otilm.core.dao.entity.signing.TspProfile;
+import com.otilm.core.dao.repository.signing.SigningProfileRepository;
+import com.otilm.core.dao.repository.signing.SigningProfileVersionRepository;
+import com.otilm.core.dao.repository.signing.TspProfileRepository;
+import com.otilm.core.model.auth.ResourceAction;
+import com.otilm.core.model.signing.SigningProfileModel;
+import com.otilm.core.model.signing.resolved.ResolvedManagedTimestampingProfile;
+import com.otilm.core.model.signing.resolved.ResolvedStaticKeyManagedSigning;
+import com.otilm.core.model.signing.timequality.LocalClockTimeQualityConfiguration;
+import com.otilm.core.security.authz.opa.dto.OpaRequestedResource;
+import com.otilm.core.security.authz.opa.dto.OpaResourceAccessResult;
+import com.otilm.core.signing.tsa.messages.TspResponse;
+import com.otilm.core.signing.tsa.resolver.SigningProfileResolverFactory;
+import com.otilm.core.util.BaseSpringBootTest;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.access.AccessDeniedException;
+import org.springframework.test.context.bean.override.mockito.MockitoBean;
+
+import java.util.List;
+
+import static com.otilm.core.signing.tsa.messages.TspRequestBuilder.aTspRequest;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.atLeastOnce;
+import static org.mockito.Mockito.lenient;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+class TsaServiceAuthzTest extends BaseSpringBootTest {
+
+    @Autowired
+    private TsaService tsaService;
+
+    @MockitoBean
+    private ManagedTimestampEngine managedTimestampEngine;
+
+    @MockitoBean
+    private SigningProfileResolverFactory signingProfileResolverFactory;
+
+    @Autowired
+    private SigningProfileRepository signingProfileRepository;
+
+    @Autowired
+    private SigningProfileVersionRepository signingProfileVersionRepository;
+
+    @Autowired
+    private TspProfileRepository tspProfileRepository;
+
+    @BeforeEach
+    void stubEngineAndResolver() throws TspException {
+        lenient().when(managedTimestampEngine.process(any(), any(), any()))
+                .thenReturn(TspResponse.granted(new byte[]{1, 2, 3}));
+
+        lenient().when(signingProfileResolverFactory.resolve(any())).thenAnswer(invocation -> {
+            SigningProfileModel<?, ?> model = invocation.getArgument(0);
+            return new ResolvedManagedTimestampingProfile(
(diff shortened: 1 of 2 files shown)


src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
```diff
@@ -0,0 +1,284 @@
...
+                    LocalClockTimeQualityConfiguration.INSTANCE, null,
+                    new ResolvedStaticKeyManagedSigning(SigningCertificateBuilder.valid(), List.of(), null, List.of()));
+        });
+    }
+
+    // ── Helpers ───────────────────────────────────────────────────────────────
+
+    private SigningProfile createTimestampingSigningProfile(String name, boolean enabled) {
+        SigningProfile profile = new SigningProfile();
+        profile.setName(name);
```
LukasNajman, line 87: Personal preference: I do prefer using builders as i find the fluent API easier to read.

src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
```diff
@@ -0,0 +1,284 @@
...
+            SigningProfile signingProfile = createTimestampingSigningProfile("sp-authz", true);
+            TspProfile tspProfile = createTspProfileFor("tsp-authz", true, signingProfile);
+
+            // when
+            tsaService.processTspRequestForTspProfile("tsp-authz", aTspRequest().build());
+
+            // then
+            verify(opaClient, atLeastOnce()).checkResourceAccess(
+                    any(),
+                    org.mockito.ArgumentMatchers.argThat(req -> isTimestampFor(req, tspProfile.getUuid())),
```
LukasNajman: Is the FQN neccessary?

src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
```diff
@@ -0,0 +1,284 @@
...
+        return tspProfileRepository.saveAndFlush(profile);
+    }
+
+    /** Establish the signing-profile → TSP-profile back-link the indirect route authorizes against. */
+    private void linkTspProfile(SigningProfile signingProfile, TspProfile tspProfile) {
+        signingProfile.setTspProfile(tspProfile);
+        signingProfileRepository.saveAndFlush(signingProfile);
+    }
+
+    /** Deny the object-level OPA check only for the given TSP profile UUID + TIMESTAMP action. */
```
LukasNajman, line 121: Consider extracting this to general helper so it can be used in other tests in future. `BaseSpringBootTest` already has some helper methods regarding the OPA, so those could be places to the helper too.

src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
```diff
@@ -0,0 +1,284 @@
...
+        void authorizesObjectLevel_withTspProfileUuid_andTimestampAction() throws Exception {
+            // given
+            SigningProfile signingProfile = createTimestampingSigningProfile("sp-authz", true);
+            TspProfile tspProfile = createTspProfileFor("tsp-authz", true, signingProfile);
+
+            // when
+            tsaService.processTspRequestForTspProfile("tsp-authz", aTspRequest().build());
+
+            // then
+            verify(opaClient, atLeastOnce()).checkResourceAccess(
```
LukasNajman, line 154: This could go to the OPA helper too.

src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
```diff
@@ -0,0 +1,284 @@
...
+                    any(),
+                    org.mockito.ArgumentMatchers.argThat(req -> isTimestampFor(req, tspProfile.getUuid())),
+                    any(), any());
+            verify(managedTimestampEngine).process(any(), any(), any());
+        }
+
+        @Test
+        void deniesObjectLevel_whenNotAuthorizedForThatTspProfile() {
+            // given
+            SigningProfile signingProfileA = createTimestampingSigningProfile("sp-a", true);
```
LukasNajman: I prefer builders over the factory methods as from this, it is not clear what the parameters mean. Alternatively, you could use variables to named the parameters.

src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
```diff
@@ -0,0 +1,284 @@
...
+            SigningProfile signingProfileB = createTimestampingSigningProfile("sp-b", true);
+            TspProfile tspProfileA = createTspProfileFor("tsp-a", true, signingProfileA);
+            createTspProfileFor("tsp-b", true, signingProfileB);
+
+            denyTimestampForObject(tspProfileA.getUuid());
+
+            // when / then
```
LukasNajman, line 172: Consider splitting the /when /then by wrapping the function call by lambda as saving it to `Executable` variable. This applies to all occurrences of the same pattern in this PR.

LukasNajman (dismissed):

src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
ivosh: Good catch — no, the FQNs weren't necessary. Replaced with imports (`java.util.UUID`, static `org.mockito.ArgumentMatchers.argThat`) and unqualified all usages. Fixed in 6c6429b8.

src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
ivosh: I drafted an entity builder to compare. Because these are persisted JPA entities (the signing-profile helper writes a two-entity graph with an FK ordering), the builder has to carry the repositories into a `persistWith(repo…)` terminal, which made it net more verbose at the call site than the factory helper — so it didn't clearly win here. To address the underlying readability concern (the opaque positional `true`/`false`), I named the args via `enabled`/`disabled` locals at the call sites instead. 6c6429b8

src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
ivosh: Went with your named-variables alternative: `createTspProfileFor("tsp-a", enabled, signingProfile)` rather than a bare `true`. (Builder comparison written up on the helpers thread above — it came out more verbose for these persisted entities.) 6c6429b8

src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
ivosh: Left as-is for now. `denyTimestampForObject`/`isTimestampFor` are specialised to TSP_PROFILE + TIMESTAMP and used only in this class, whereas the existing `BaseSpringBootTest` OPA helpers are resource-level. I'd rather hoist it to the base class when a second caller actually appears, to avoid speculative generalisation — happy to do that follow-up if/when another test needs it.

src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
ivosh: Same reasoning as the `denyTimestampForObject` thread — `isTimestampFor` is a single-use, TSP-specific predicate, so I'd prefer to move it to the OPA helper once there's a second consumer rather than now.

src/test/java/com/otilm/core/signing/tsa/TsaServiceAuthzTest.java
ivosh: Kept the current form here. With AssertJ, `assertThatThrownBy(() -> tsaService.xxx(...))` already expresses when (the lambda) and then (the fluent assertion chain) together; extracting it into a separate `Executable` variable leans toward the older JUnit `assertThrows` style and reads less cleanly in this file. Where pulling the call out aids clarity I did factor out a local (`var request = …`).

LukasNajman (approved):

PR #1673, opened 2026-06-25
Project v3 attribute field mappings into generated PKCS#10 requests

### LINKED ISSUE #1630: Build PKCS#10 from structured content (Mode A)
### Description
Build a PKCS#10 CSR from structured certificate-request content (full RDN/SAN/extension support), replacing the subject-only limitation of the current flat CSR attributes. Part of #1572 (epic OmniTrustILM/ilm#131).

### Scope — what ships
- Pure kernels: `RequestAttributeProjector` (attribute values → `CertificateRequestContent`) and `X509RequestContentRenderer` (content → BouncyCastle `X500Name` / `Extensions`).
- `CsrRequestAttributes` — the v3 default request-attribute set (re-expresses the legacy `CsrAttributes` six subject fields as `DataAttributeV3` with explicit `fieldMapping`; seed for the platform default set in CO-3).
- `CryptographicOperationServiceImpl.generateCsr` overload taking `Extensions`; Mode-A wiring in `ClientOperationServiceImpl.generateBase64EncodedCsr`.
- Also ships `getCertificateApiClientV3` + v3 routing (consumed by CO-4).

### Dependencies
- interfaces #730 (SNAPSHOT-first), CO-0b, CO-0c. Produces the default set CO-2 and CO-3 consume.

### Acceptance Criteria
- Pure-kernel unit tests (no Docker) + integration; new-code coverage ≥80%; SonarCloud gate clean.

### Review focus
No regression on the existing subject-only path; `criticalOverridable` must-stay-critical blocklist.

klaraf755:
Closes #1630


src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
```diff
@@ -0,0 +1,133 @@
...
+            gen.addExtension(Extension.subjectAlternativeName, false, new GeneralNames(names));
+        }
+
+        List<RequestedExtension> extensions = x509.getExtensions() == null ? List.of() : x509.getExtensions();
+        for (RequestedExtension ext : extensions) {
+            ASN1ObjectIdentifier oid = new ASN1ObjectIdentifier(ext.getOid());
+            boolean critical = effectiveCritical(ext.getOid(), ext.getCritical());
+            byte[] derValue = Base64.decode(ext.getValue());
+            gen.addExtension(oid, critical, derValue);
+        }
```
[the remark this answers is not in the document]
ivosh: Fixed in 94420806.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
```diff
@@ -0,0 +1,133 @@
...
+        return switch (encoding) {
+            case IA5_STRING -> new DERIA5String(value);
+            case OCTET_STRING -> new DEROctetString(value.getBytes(java.nio.charset.StandardCharsets.UTF_8));
+            // DER: caller supplies a base64-encoded DER blob
+            case DER -> org.bouncycastle.util.encoders.Base64.decode(value).length > 0
+                    ? new org.bouncycastle.asn1.DEROctetString(org.bouncycastle.util.encoders.Base64.decode(value))
+                    : new DERUTF8String(value);
+            default -> new DERUTF8String(value);
+        };
```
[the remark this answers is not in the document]
ivosh: Fixed in 94420806.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1367,8 +1385,54 @@ private void validateSubjectDnForCertificate(String certificateContent, Certific
...
+            return defaults;
+        }
+
+        Map<String, String> codeToOid = OidHandler.getCodeToOidMap();
+
+        Set<String> claimedRdns = connectorDefs.stream()
+                .flatMap(d -> d.getFieldMapping().getFields().stream())
+                .filter(f -> f.getFieldType() == FieldType.RDN)
+                .map(f -> normalizeRdn(((RdnMappedField) f).getRdn(), codeToOid))
+                .collect(java.util.stream.Collectors.toSet());
```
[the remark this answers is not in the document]
ivosh: Fixed in 94420806.

src/test/java/com/otilm/core/service/CertificateServiceTest.java
```diff
@@ -1,10 +1,12 @@
 package com.otilm.core.service;
 
+import com.otilm.api.model.common.attribute.v3.content.StringAttributeContentV3;
 import com.otilm.core.attribute.CsrAttributes;
 import com.otilm.api.exception.*;
 import com.otilm.api.model.client.attribute.RequestAttributeV2;
 import com.otilm.api.model.client.attribute.RequestAttributeV3;
 import com.otilm.api.model.core.workflows.*;
+import com.otilm.core.attribute.CsrAttributes;
```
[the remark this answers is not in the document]
ivosh: Fixed in 94420806.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
```diff
@@ -0,0 +1,170 @@
...
+        }
+        return switch (encoding) {
+            case DER -> Base64.decode(value);
+            case UTF8_STRING -> new DERUTF8String(value).getEncoded(ASN1Encoding.DER);
+            case IA5_STRING -> new DERIA5String(value).getEncoded(ASN1Encoding.DER);
+            case PRINTABLE_STRING -> new DERPrintableString(value).getEncoded(ASN1Encoding.DER);
+            case OCTET_STRING -> new DEROctetString(value.getBytes(StandardCharsets.UTF_8)).getEncoded(ASN1Encoding.DER);
+            case BIT_STRING -> new DERBitString(value.getBytes(StandardCharsets.UTF_8)).getEncoded(ASN1Encoding.DER);
+        };
+    }
```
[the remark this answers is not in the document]
ivosh: Fixed in aedb13bf.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1408,8 +1426,65 @@ private void validateSubjectDnForCertificate(String certificateContent, Certific
...
+                    .map(DataAttributeV3.class::cast)
+                    .toList();
+        } catch (Exception e) {
+            logger.debug("Could not fetch connector issue attributes for RA profile {}; using default CSR attribute set: {}",
+                    raProfile.getName(), e.getMessage());
+            return defaults;
+        }
+
+        return mergeIssuanceDefinitions(defaults, connectorDefs, OidHandler.getCodeToOidMap());
+    }
```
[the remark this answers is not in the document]
ivosh: Fixed in aedb13bf.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1408,8 +1426,65 @@ private void validateSubjectDnForCertificate(String certificateContent, Certific
...
+                .toList();
+
+        List<DataAttributeV3> merged = new ArrayList<>(connectorDefs);
+        merged.addAll(filteredDefaults);
+        return merged;
+    }
+
+    private static String normalizeRdn(String rdn, Map<String, String> codeToOid) {
+        return codeToOid.getOrDefault(rdn, rdn);
+    }
```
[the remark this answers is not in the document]
ivosh: Fixed in aedb13bf.

src/main/java/com/otilm/core/attribute/CertificateRequestAttributeProjector.java
```diff
@@ -0,0 +1,100 @@
...
+                    .toList();
+
+            for (MappedField field : sortedFields) {
+                switch (field) {
+                    case RdnMappedField rdn ->
+                            subject.add(new RdnEntry(rdn.getRdn(), value));
+                    case SanMappedField san ->
+                            subjectAltNames.add(new GeneralNameEntry(san.getGeneralNameType(), value, san.getGeneralNameType() == GeneralNameType.OTHER_NAME ? san.getOtherNameOid() : null, san.getOtherNameValueEncoding()));
+                    case ExtensionMappedField ext ->
+                            extensions.add(new RequestedExtension(ext.getExtensionOid(), false, null, value));
```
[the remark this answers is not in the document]
ivosh: Fixed in aedb13bf.

src/test/java/com/otilm/core/util/X509RequestContentRendererTest.java
```diff
@@ -0,0 +1,493 @@
...
+            Map<String, OidRecord> existing = OidHandler.getOidCache(OidCategory.RDN_ATTRIBUTE_TYPE);
+            savedRdnCache = existing == null ? null : new HashMap<>(existing);
+        }
+
+        @AfterAll
+        static void restoreRdnCache() {
+            if (savedRdnCache != null) {
+                OidHandler.cacheOidCategory(OidCategory.RDN_ATTRIBUTE_TYPE, savedRdnCache);
+            }
+        }
```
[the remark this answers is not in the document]
ivosh: Fixed in aedb13bf.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -245,7 +254,14 @@ public CertificateDetailDto submitCertificateRequest(ClientCertificateRequestDto
             throw new ValidationException("Cannot submit certificate request without specifying key or uploaded request content");
         }
 
-        String certificateRequest = generateBase64EncodedCsr(request.getRequest(), request.getFormat(), request.getCsrAttributes(), request.getKeyUuid(), request.getTokenProfileUuid(), request.getSignatureAttributes(), request.getAltKeyUuid(), request.getAltTokenProfileUuid(), request.getAltSignatureAttributes());
+        RaProfile raProfile = request.getRaProfileUuid() != null
+                ? raProfileRepository.findByUuid(request.getRaProfileUuid()).orElse(null)
+                : null;
+        if (raProfile == null) {
```
lubomirw: **[High]** | Correctness

The `raProfile == null` guard fires *before* branching on `uploadedRequest`, so a request that uploads a pre-built PKCS#10 without an RA profile — previously valid — now fails with a `ValidationException`. The guided reviewer also traced this to **SCEP enrollment**: a SCEP profile with no RA profile bound (`ScepServiceImpl` only sets one when it exists) would break.

_Suggested:_ scope the RA-profile requirement to the key-generation path only — move the guard into the `uploadedRequest == null` branch (inside `generateBase64EncodedCsr` / `resolveIssuanceDefinitions`) so uploaded-CSR and RA-profile-less SCEP flows keep working.

<sub>flagged by guided-pr-review + Copilot CLI + superpowers (3 independent reviewers)</sub>

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -245,7 +254,14 @@ public CertificateDetailDto submitCertificateRequest(ClientCertificateRequestDto
             throw new ValidationException("Cannot submit certificate request without specifying key or uploaded request content");
         }
 
-        String certificateRequest = generateBase64EncodedCsr(request.getRequest(), request.getFormat(), request.getCsrAttributes(), request.getKeyUuid(), request.getTokenProfileUuid(), request.getSignatureAttributes(), request.getAltKeyUuid(), request.getAltTokenProfileUuid(), request.getAltSignatureAttributes());
+        RaProfile raProfile = request.getRaProfileUuid() != null
+                ? raProfileRepository.findByUuid(request.getRaProfileUuid()).orElse(null)
```
lubomirw, line 287: **[Low]** | Correctness

This uses `raProfileRepository.findByUuid(...)` rather than `findByUuidAndEnabledIsTrue(...)`, so a **disabled** RA profile passes `submitCertificateRequest`, whereas `issueCertificate` explicitly rejects disabled profiles — an inconsistency that lets a disabled profile drive CSR / issue-attribute resolution.

_Suggested:_ use `findByUuidAndEnabledIsTrue` (or add the same disabled-profile guard used in `issueCertificate`).

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -814,6 +830,7 @@ private void createRequestFromKeys(ClientCertificateRekeyRequestDto request, Cer
                 keyUuid,
                 getTokenProfileUuid(request.getTokenProfileUuid(), oldCertificate),
                 principal,
+                null,
```
lubomirw, line 1058: **[Follow-up]** | Correctness — out of scope for this PR, not a blocker

The re-key-from-key path (`createRequestFromKeys`) passes `null` for the projected `Extensions`, so a re-keyed cert whose CSR is generated server-side from a key carries no SAN/extension-mapped attributes. On deeper review this is **pre-existing** behavior — the base commit's low-level overload had no extensions parameter at all; this PR only wired projection into the issue/submit path. Renew is **unaffected** (it reuses the prior CSR verbatim via `getExistingCsr`).

_Suggested (follow-up):_ run the same `resolveIssuanceDefinitions → CertificateRequestAttributeProjector.project → X509RequestContentRenderer.toExtensions` pipeline here (or carry forward the old certificate's extensions) instead of `null`, and add a re-key-from-key test asserting SAN/extension presence. Worth confirming product intent first: should server-generated re-key re-project CSR attributes, or only rotate the key while the authority re-issues with its own policy-driven extensions?

<sub>flagged by Copilot CLI; scope/regression corrected on deeper review (renew is fine; re-key gap predates this PR)</sub>

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1408,8 +1427,76 @@ private void validateSubjectDnForCertificate(String certificateContent, Certific
...
+        if (raProfile == null) {
+            return defaults;
+        }
+        List<DataAttributeV3> connectorDefs;
+        try {
+            connectorDefs = extendedAttributeService.listIssueCertificateAttributes(raProfile).stream()
+                    .filter(DataAttributeV3.class::isInstance)
+                    .map(DataAttributeV3.class::cast)
+                    .toList();
+        } catch (Exception e) {
```
lubomirw: **[Medium]** | Error handling

`resolveIssuanceDefinitions` wraps the connector issue-attribute fetch in `catch (Exception)` and logs only at `debug`, then falls back to the static default CSR set. A genuine connector failure (timeout, 5xx, parse error) therefore produces a CSR missing connector-required SANs/extensions with no error surfaced at request time — only later rejected by the CA.

_Suggested:_ distinguish "connector has no v3 attributes" (empty → fall back) from "connector unavailable" (exception → propagate / warn), failing closed on genuine errors rather than silently substituting defaults.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1408,8 +1427,76 @@ private void validateSubjectDnForCertificate(String certificateContent, Certific
...
+     */
+    private List<DataAttributeV3> resolveIssuanceDefinitions(RaProfile raProfile) {
+        List<DataAttributeV3> defaults = CsrAttributes.csrAttributesAsDataAttributesV3();
+        if (raProfile == null) {
+            return defaults;
+        }
+        List<DataAttributeV3> connectorDefs;
+        try {
+            connectorDefs = extendedAttributeService.listIssueCertificateAttributes(raProfile).stream()
+                    .filter(DataAttributeV3.class::isInstance)
```
lubomirw: **[Low]** | Correctness — works as designed; only the silence is the issue

`.filter(DataAttributeV3.class::isInstance)` silently discards connector attributes that aren't v3. A **v2** authority adapter returns `DataAttributeV2`, so for a v2 connector all connector-supplied subject/SAN/extension fields are dropped and the CSR is built from the 6 default RDNs only.

On deeper review this is **by design** — Mode A (#1630) is v3-only — and it is **not a regression** (the pre-PR code never consulted connector attributes at all). The only gap is that the drop is silent.

_Suggested:_ add a `debug` / one-time log counting the dropped non-v3 attributes (e.g. "N connector issue attributes ignored for structured CSR building; RA profile {} uses a v2 authority connector — structured CSR enrichment requires v3"). A hard error would be a worse regression than the v2 status quo.

<sub>flagged by guided-pr-review; confirmed v3-by-design on deeper review</sub>

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1419,16 +1506,32 @@ private String generateBase64EncodedCsr(String uploadedRequest, CertificateReque
...
+            // validate and update definitions of certificate request attributes with the attribute engine
             attributeEngine.validateUpdateDataAttributes(null, null, definitions, csrAttributes);
-            // TODO: return CertificateRequest object instead of Base64 encoded CSR
+
+            X509RequestContent requestContent = CertificateRequestAttributeProjector.project(definitions, csrAttributes);
+            Extensions extensions;
+            try {
+                extensions = X509RequestContentRenderer.toExtensions(requestContent);
+            } catch (IOException e) {
+                throw new CertificateException("Failed to build CSR extensions: " + e.getMessage());
```
lubomirw: **[Medium]** | Security

This — and the sibling `throw new CertificateException("Failed to build CSR subject: " + e.getMessage())` at line 1527 — forwards raw exception text (which can include connector-supplied OIDs and internal parse detail) into an externally visible API response (info disclosure). It also **drops the original cause** `e`, losing the stack trace for diagnosis.

_Suggested:_ log the full exception server-side and throw a fixed external message (e.g. `"Failed to build CSR extensions"`) that **wraps `e` as the cause**, without appending `e.getMessage()`.

<sub>flagged by Copilot CLI (info leak) + superpowers (dropped cause)</sub>

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
```diff
@@ -0,0 +1,184 @@
...
+        if (!sans.isEmpty()) {
+            GeneralName[] names = sans.stream()
+                    .map(X509RequestContentRenderer::toGeneralName)
+                    .toArray(GeneralName[]::new);
+            gen.addExtension(Extension.subjectAlternativeName, false, new GeneralNames(names));
+        }
+
+        List<RequestedExtension> extensions = x509.getExtensions() == null ? List.of() : x509.getExtensions();
+        for (RequestedExtension ext : extensions) {
+            ASN1ObjectIdentifier oid = new ASN1ObjectIdentifier(ext.getOid());
```
lubomirw: **[Medium]** | Error handling

`new ASN1ObjectIdentifier(ext.getOid())` here — and `new ASN1ObjectIdentifier(e.getOtherNameOid())` at line 140 — runs on connector-supplied data with no null/format check, unlike `resolveOid` and `decodeBase64Der`, which validate/translate. A null OID throws `NullPointerException` and a malformed OID throws `IllegalArgumentException`, both of which bypass the declared `throws IOException` contract and surface as an unhandled 500.

_Suggested:_ validate against `OID_PATTERN` (and non-null) before constructing the `ASN1ObjectIdentifier`, throwing a controlled `IOException` for invalid/missing OIDs, mirroring `resolveOid()`.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
```diff
@@ -0,0 +1,184 @@
...
+     */
+    public static Extensions toExtensions(X509RequestContent x509) throws IOException {
+        ExtensionsGenerator gen = new ExtensionsGenerator();
+
+        List<GeneralNameEntry> sans = x509.getSubjectAltNames() == null ? List.of() : x509.getSubjectAltNames();
+        if (!sans.isEmpty()) {
+            GeneralName[] names = sans.stream()
+                    .map(X509RequestContentRenderer::toGeneralName)
+                    .toArray(GeneralName[]::new);
+            gen.addExtension(Extension.subjectAlternativeName, false, new GeneralNames(names));
```
lubomirw: **[Low]** | Correctness

The SAN extension is added with `critical = false` unconditionally. RFC 5280 §4.2.1.6 requires the SAN to be marked **critical** when the subject DN is empty — which this renderer can now produce (subject may be empty while SANs are present).

_Suggested:_ when the rendered subject is empty and SANs exist, add the SAN extension as critical.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
```diff
@@ -0,0 +1,184 @@
...
+            throw new IOException("Extension value is required");
+        }
+        ExtensionValueEncoding effective = encoding == null ? ExtensionValueEncoding.DER : encoding;
+        return switch (effective) {
+            case DER -> decodeBase64Der(value);
+            case UTF8_STRING -> new DERUTF8String(value).getEncoded(ASN1Encoding.DER);
+            case IA5_STRING -> new DERIA5String(value).getEncoded(ASN1Encoding.DER);
+            case PRINTABLE_STRING -> new DERPrintableString(value).getEncoded(ASN1Encoding.DER);
+            case OCTET_STRING -> new DEROctetString(value.getBytes(StandardCharsets.UTF_8)).getEncoded(ASN1Encoding.DER);
+            case BIT_STRING -> new DERBitString(value.getBytes(StandardCharsets.UTF_8)).getEncoded(ASN1Encoding.DER);
```
lubomirw: **[Low]** | Maintainability

`new DERBitString(value.getBytes(UTF_8))` embeds the value's literal UTF-8 text bytes as bit-string content (0 unused bits) — unlike the `DER` branch, which base64-decodes a pre-encoded blob. This **is reachable**: an operator can register a custom extension OID with `valueEncoding = BIT_STRING` via the OID registry (`CustomOidEntryServiceImpl` stores it with no validation), and the result is a cert that issues successfully but carries a silently malformed extension. Untested on the extension path (BIT_STRING / UTF8_STRING / PRINTABLE_STRING all lack coverage).

_Suggested:_ either reject `BIT_STRING` explicitly (`IOException`) or treat it like `DER` (base64-decode via `decodeBase64Der`); add an `ExtensionValueEncodingTest` case to pin the contract.

<sub>flagged by superpowers; confirmed reachable (not theoretical) on deeper review</sub>

lubomirw (commented):
## Multi-reviewer triage

Reviewed against `main` (merge-base `708d402`) by four independent reviewers — **Copilot CLI**, **guided-pr-review**, **superpowers**, and a **hygiene** scan (hygiene came back clean). The inline comments below are deduped and categorized; cross-source agreement is noted per comment.

Five "needs-judgment" findings were then re-investigated against the shipped code, and the refined verdicts are baked into the relevant inline comments:

- **#2** corrected — *renew* is fine (it reuses the prior CSR), and the re-key-from-key gap is **pre-existing**, so it's flagged as a follow-up, not a blocker.
- **#6** confirmed a real, likely-accidental inconsistency (the legacy guard is still on its three sibling methods).
- **#7** confirmed **v3-by-design** and not a regression — softened to a logging suggestion.
- **#12** confirmed **reachable** (custom OID registry), not theoretical.

Two items were dropped as not actionable: a theoretical null-input path in the projector (guarded upstream by `validateUpdateDataAttributes`) and three cosmetic nits (a migration UUID comment, dead params, and inline fully-qualified `java.util.stream` references). A documentation gap is tracked in a separate issue.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    private static boolean effectiveCritical(String oid, boolean requestedCritical) {
+        if (oid != null && CRITICALITY_FORCED_OIDS.contains(oid)) {
+            return true;
+        }
+        return requestedCritical;
+    }
+
+    private static GeneralName toGeneralName(GeneralNameEntry e) {
```
lubomirw: **[Low]** | Testing

`toGeneralName` renders non-OTHER_NAME SANs via BouncyCastle's `GeneralName(int, String)` string constructor, but the renderer tests only cover DNS / EMAIL / URI / OTHER_NAME. The **IP** (OCTET STRING), **DIRECTORY_NAME** (X.500 name), and **REGISTERED_ID** (OID) tags supported by `GeneralNameType` go through that same string path untested — a malformed string would corrupt or throw at CSR build time.

_Suggested:_ add renderer tests for `GeneralNameType.IP`, `DIRECTORY_NAME`, and `REGISTERED_ID` asserting the correct ASN.1 tag/value, or explicitly reject the unsupported subset with a clear error.

src/main/java/com/otilm/core/attribute/CertificateRequestAttributeProjector.java
```diff
@@ -0,0 +1,122 @@
...
+        if (value == null) return null;
+        return value;
+    }
+
+    private static Map<UUID, String> extractStringValues(List<? extends RequestAttribute> attributes) {
+        Map<UUID, String> map = new HashMap<>();
+        for (RequestAttribute attr : attributes) {
+            List<?> content = attr.getContent();
+            if (content == null || content.isEmpty()) continue;
+            Object first = content.getFirst();
```
lubomirw, line 123: **[Low]** | Correctness

`extractStringValues` reads only `content.getFirst()` and only when its data is a `String`, so a multi-value attribute (e.g. multiple SANs) keeps only the first entry, and non-string content types are dropped without any signal.

_Suggested:_ iterate all string content items and emit one mapped field per value; or, if multi-value mapping is out of scope, explicitly reject list-valued mapped attributes with a validation error (or document the single-value limitation) so the truncation is intentional and discoverable.

<sub>flagged by guided-pr-review + Copilot CLI</sub>

src/main/java/com/otilm/core/service/v2/impl/ExtendedAttributeServiceImpl.java
```diff
@@ -53,16 +61,10 @@ public void setAttributeEngine(AttributeEngine attributeEngine) {
...
+        if (authorityRef.getConnector() == null) {
             throw new NotFoundException("Connector of the Authority is not available / deleted");
         }
-        validateLegacyConnector(connector);
-
-        ApiClientConnectorInfo connectorDto = connectorService.getConnectorForApiClient(connector.getUuid());
-        return connectorApiFactory.getCertificateApiClientV2(connectorDto).listIssueCertificateAttributes(
-                connectorDto,
-                authorityRef.getAuthorityInstanceUuid());
+        return authorityProviderAdapterFactory.forAuthority(authorityRef).listIssueAttributes(authorityRef, raProfile);
```
lubomirw, line 69: **[Medium]** | Correctness

`listIssueCertificateAttributes` no longer calls `validateLegacyConnector(connector)` before dispatch — it now delegates straight to `authorityProviderAdapterFactory.forAuthority(...)`. The factory routes a legacy authority (null `connectorInterface`) to the **v2 adapter**, which performs no legacy detection (the "legacy never reaches here" guarantee is only a javadoc comment, not code). So a legacy `LegacyEjbca` authority on this path now fails with a confusing connector-side error instead of the explicit `NotFoundException("Legacy Authority. V2 Implementation not found...")`.

This looks like an **oversight**: the three sibling methods (`validateIssueCertificateAttributes`, `listRevokeCertificateAttributes`, `validateRevokeCertificateAttributes`) and the caller at `ClientOperationServiceImpl:1127` all still call `validateLegacyConnector`.

_Suggested:_ restore `validateLegacyConnector(authorityRef.getConnector())` before the `forAuthority(...)` dispatch (matches the siblings, low-risk) — or, if the adapter framework is meant to be the single enforcement point, move the legacy check into the factory and remove it from all four sites consistently.

<sub>flagged by Copilot CLI; confirmed a real, likely-accidental inconsistency on deeper review</sub>

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
ivosh: Fixed in e5b50c33 — moved the RA-profile requirement into the key-generation branch of `generateBase64EncodedCsr`, so uploaded-CSR and RA-profile-less SCEP flows pass again. Confirmed against `ScepServiceImpl` (only sets the RA-profile UUID when one is bound).

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
ivosh: Fixed in e5b50c33 — added an explicit disabled-RA-profile rejection in `submitCertificateRequest` (mirrors `issueCertificate`). Used an explicit `getEnabled()` check rather than `findByUuidAndEnabledIsTrue` so a supplied-but-disabled profile is rejected outright instead of being silently treated as absent on the now-optional uploaded path.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
ivosh: Fixed in e5b50c33 — `resolveIssuanceDefinitions` no longer swallows connector failures: a genuine error now propagates (fail closed) instead of silently substituting the default CSR set. An empty connector result still falls back naturally via `mergeIssuanceDefinitions`.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
ivosh: Fixed in e5b50c33 — added a debug log counting dropped non-v3 connector attributes (kept the v3-only-by-design filter).

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
ivosh: Fixed in e5b50c33 — both sites now log the full exception server-side and throw a fixed message wrapping `e` as the cause, with no `e.getMessage()` on the wire.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
ivosh: Fixed in e5b50c33 — added `parseOid()` validating non-null + `OID_PATTERN` for both the extension OID and the otherName type-id, throwing a controlled `IOException` instead of letting an NPE/IAE escape as a 500.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
ivosh: Fixed in e5b50c33 — the SAN is now marked critical when the subject DN is empty (RFC 5280 §4.2.1.6); added tests for both the empty- and present-subject cases.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
ivosh: Fixed in e5b50c33 — `BIT_STRING` extension value encoding is now rejected with a controlled `IOException` rather than embedding literal UTF-8 bytes; added an `ExtensionValueEncodingTest` case. (The only legitimate `DERBitString` use in the codebase wraps real signature bytes, not text.)

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
ivosh: Fixed in e5b50c33 — added renderer tests for `GeneralNameType.IP`, `DIRECTORY_NAME`, and `REGISTERED_ID` asserting the correct ASN.1 tag.

src/main/java/com/otilm/core/service/v2/impl/ExtendedAttributeServiceImpl.java
ivosh: Fixed in e5b50c33 — restored `validateLegacyConnector(connector)` before the `forAuthority(...)` dispatch in `listIssueCertificateAttributes`, matching the three sibling methods.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
ivosh: Agreed — leaving as a follow-up, not changing it in this PR. Confirmed against the design plan: renew/re-key content seeding is Phase 4 scope ("renew-seeding"), and renew is unaffected here (it reuses the prior CSR). Tracking the re-key extension projection separately.

src/main/java/com/otilm/core/attribute/CertificateRequestAttributeProjector.java
ivosh: Documented in e5b50c33 — added a Javadoc note on `extractStringValues` making the single-value limitation explicit. This matches the legacy `buildSubject` baseline (one value per RDN via `getSingleItemAttributeContentValue`), so it is not a regression. Implementing multi-valued RDNs/SANs (spec §8 / §13 edge-case 4) is deferred to a follow-up rather than expanding this PR's scope.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
ivosh: https://github.com/OmniTrustILM/core/issues/1688

src/main/java/com/otilm/core/attribute/CertificateRequestAttributeProjector.java
ivosh: https://github.com/OmniTrustILM/core/issues/1688

lubomirw (approved):

PR #1676, opened 2026-06-26
Batch the signing-record outbox drain and bypass merge SELECTs

### LINKED ISSUE #1675: SigningRecord batch insert: `saveAll()` issues SELECT per entity for application-assigned UUIDs, preventing JDBC batching
### Description

`SigningRecordWriter.insertBatch` uses Spring Data's `saveAll()`, which internally calls `merge()` on each entity. Because signing record UUIDs are assigned in application code (not by the database), Hibernate cannot determine whether each entity is new — so it issues a `SELECT … FOR UPDATE` per record to check existence before inserting. This serializes what should be a single batched INSERT and locks rows under load, degrading throughput on the signing record write path.

### Steps to Reproduce

1. Generate a burst of signing operations that fills the best-effort queue.
2. Let the background flush thread call `SigningRecordWriter.insertBatch` with a batch of `SigningRecord` entities (UUIDs assigned before the call).
3. Enable Hibernate SQL logging (`logging.level.org.hibernate.SQL=DEBUG`) or a DB slow-query log.
4. Observe one `SELECT … FOR UPDATE` per entity instead of a single batched `INSERT`.

### Expected Behavior

Hibernate issues a single batched `INSERT` for the entire batch (respecting `hibernate.jdbc.batch_size`), with no preceding SELECT, because the records are always new at the point `insertBatch` is called.

### Actual Behavior

Spring Data's `save()`/`saveAll()` calls `EntityManager.merge()` for each entity. Because the UUID is application-assigned (not `@GeneratedValue`), Hibernate treats the entity as potentially detached and emits a `SELECT … FOR UPDATE` per record before inserting. Under load this serializes inserts, holds row locks, and prevents JDBC batch insert optimisation — measurably degrading signing record write throughput.

### Suggested Fix
Inject `EntityManager` directly into `com.otilm.core.service.writer.signingrecord.SigningRecordWriter` and replace `saveAll()` with `records.forEach(em::persist)`. `persist()` tells Hibernate the entity is definitely new — no SELECT is issued and Hibernate can group the INSERTs into a single JDBC batch.

### Severity

Major

### Module

Signing

### Screenshot

_No response_

### Logs

_No response_

### Environment

_No response_

LukasNajman:
### Problem                                                                                                                                                  
                                                                                                                                                               
  Two related bottlenecks on the signing record write path degraded throughput under load:                                                                     
                                                                                                                                                               
  **1. `SigningRecordWriter` — SELECT before every INSERT**                                                                                                    
                                                                                                                                                               
  `saveAll()` / `saveAndFlush()` routes through Spring Data's `merge()` path. Because signing record UUIDs are assigned in application code (not by the        
  database), Hibernate cannot determine whether each entity is new — so it issues a `SELECT … FOR UPDATE` per record before inserting. This serializes what    
  should be a batched INSERT, holds row locks, and prevents JDBC batch optimisation entirely.                                                                  
                                                                                                                                                               
  **2. `SigningRecordOutboxDrainer` — four round-trips and one transaction per outbox row**                                                                    
                                                                                                                                                               
  The drainer called `SigningRecordWriter.saveRecordAndDeleteOutbox()` for each row individually. That method runs:                                            
  1. SELECT the outbox row                                                                                                                                     
  2. SELECT … FOR UPDATE (again, merge-path for application-assigned UUID)                                                                                     
  3. INSERT into `signing_record`                                                                                                                              
  4. DELETE from `signing_record_outbox`                                                                                                                       
                                                                                                                                                               
  With a batch of N rows: 4×N round-trips and N transactions. Under sustained load the drainer could not keep up with the inflow, the outbox grew unboundedly, 
  and the connection pool was saturated with tiny single-row transactions.                                                                                     
                                                                                                                                                               
  ---                                                                                                                                                          
                                                                                                                                                               
 ### Fix                                                                                                                                                      
                                                                                                                                                               
  **`SigningRecordWriter`** — replace all `save()`/`saveAndFlush()` calls with `EntityManager.persist()` + explicit `flush()`. `persist()` tells Hibernate the 
  entity is definitely new — no SELECT is issued and the JDBC batch optimisation kicks in.                                                                     
                                                                                                                                                               
  **New `SigningRecordWriter.persistBatchAndDeleteOutbox()`** — bulk drain method that persists an entire batch via Hibernate JDBC batching and removes all    
  originating outbox rows with a single `DELETE … WHERE uuid IN (…)`. Cost: **two round-trips per batch** regardless of batch size, in one transaction.        
                                                                                                                                                               
  **`SigningRecordOutboxDrainer`** restructured to use the fast path by default:                                                                               
  - Load all batch rows in one query, map them all, call `persistBatchAndDeleteOutbox()`.                                                                      
  - If that throws (crash-recovery case: a node died after inserting into `signing_record` but before deleting the outbox row, leaving a duplicate), fall back 
  to the existing per-row `saveRecordAndDeleteOutbox()` path — which uses an idempotent merge copy and preserves individual retry/poison accounting per row.   
  - Fallback fires are counted via a new `signing_record.outbox.batch_drain.fallback` metric.                                                                  
                                                                                                                                                               
  **`BestEffortSigningRecordQueue.enqueueDropping()`** — two related fixes spotted during the same investigation:                                              
  - Previously logged one WARN per evicted record inside the loop — under sustained full-queue conditions this produced thousands of log lines per second.     
  Changed to collect all evicted UUIDs and emit a single summary line per enqueue call.                                                                        
  - Added a fast-path early return when `offer()` succeeds on the first try, avoiding an `ArrayList` allocation on every non-full enqueue.


src/main/java/com/otilm/core/signing/record/SigningRecordOutboxDrainer.java
```diff
@@ -105,38 +103,70 @@ private void drainBatches() {
...
-            metrics.persist(SigningRecordPersistenceMode.DEFERRED_DURABLE.name()).increment();
-            writer.saveRecordAndDeleteOutbox(SigningRecordMapper.toRecord(row));
+            writer.saveRecordAndDeleteOutbox(signingRecord);
             return true;
         } catch (RuntimeException drainError) {
             metrics.persistFailed(SigningRecordPersistenceMode.DEFERRED_DURABLE.name()).increment();
-            log.warn("Failed to drain outbox row {}", uuid, drainError);
-            recordFailure(uuid, drainError.getMessage());
+            log.warn("Failed to drain outbox row {}", signingRecord.uuid, drainError);
+            recordFailure(signingRecord.uuid, drainError.toString());
```
ivosh: **~ Minor** | Maintainability

SigningRecordOutboxDrainer.java:168-169 read `signingRecord.uuid` via the public field inherited from UniquelyIdentifiedAndAudited, whereas the rest of this PR uses the accessor `SigningRecord::getUuid` (e.g. SigningRecordWriter.java:103, the recordFor matcher at the test's :283). Both compile (the superclass field is public + @Getter), but mixing field access and getter on the same property reads inconsistently. Prefer signingRecord.getUuid() to match the neighbouring code.

src/main/java/com/otilm/core/service/writer/signingrecord/SigningRecordWriter.java
```diff
@@ -30,6 +32,9 @@ public class SigningRecordWriter {
      */
     private static final int MAX_ERROR_LENGTH = 1000;
 
+    @PersistenceContext
```
ivosh: **. Nitpick** | Maintainability

SigningRecordWriter.java:35-36 introduces `@PersistenceContext private EntityManager entityManager` (field injection) alongside the existing constructor-injected `recordRepository`/`outboxRepository` (SigningRecordWriter.java:41-45). The sibling consumer of an EntityManager in this subsystem, SigningRecordOutboxDrainer, constructor-injects it instead (SigningRecordOutboxDrainer.java:50,56). Mixing field and constructor injection in one bean is the inconsistency worth a note; field injection also makes the dependency non-final and harder to set in a plain unit test. Note that `@PersistenceContext` on a field is itself a legitimate, common JPA idiom (it carries proxy/extended-context semantics that plain `@Autowired` does not), so this is a consistency nit rather than a defect — if field injection is the deliberate choice for the proxy semantics, leave it; otherwise move it to the constructor to match the drainer. Take-it-or-leave-it.

src/main/java/com/otilm/core/signing/record/SigningRecordOutboxDrainer.java
```diff
@@ -105,38 +103,70 @@ private void drainBatches() {
         }
     }
 
+    /**
+     * Drains a batch using the fast bulk path: loads all rows in one query, maps them, and hands them to
+     * {@link SigningRecordWriter#persistBatchAndDeleteOutbox} which persists all records via Hibernate JDBC batching
+     * and removes the outbox rows with a single bulk DELETE — two round-trips per batch instead of four per row.
+     *
+     * <p>Falls back to per-row draining when the batch path throws. The only expected failure is a
+     * {@link PersistenceException} on the flush in crash-recovery: a node crashed after writing to
```
ivosh: **~ Minor** | Documentation

SigningRecordOutboxDrainer.java:112 — the new drainRows Javadoc says "The only expected failure is a {@link PersistenceException} on the flush in crash-recovery". PersistenceException (jakarta.persistence.PersistenceException) is not imported in this file (imports verified: only EntityManager from jakarta.persistence is imported, lines 3-17), and there is no other PersistenceException on the classpath of this package, so the {@link} cannot resolve. javadoc will emit a 'reference not found' warning, and the link renders as plain text. Either add `import jakarta.persistence.PersistenceException;` and keep the {@link}, or downgrade to {@code PersistenceException} if a live link is not wanted. Non-blocking.

src/test/java/com/otilm/core/signing/record/SigningRecordOutboxDrainerUnitTest.java
```diff
@@ -242,6 +254,27 @@ void drainOnce_queriesUsingTheConfiguredBatchSizeAndPoisonThreshold() {
...
+                .persistBatchAndDeleteOutbox(anyList());
+
+        // when
+        createDrainer().drainOnce();
+
+        // then each row is drained individually via the per-row fallback
+        verify(writer).saveRecordAndDeleteOutbox(recordFor(uuid1));
+        verify(writer).saveRecordAndDeleteOutbox(recordFor(uuid2));
+        verify(writer, never()).recordFailure(any(), any());
+        assertEquals(2, atemptedDrainedCount());
```
ivosh: **~ Minor** | Testing

SigningRecordOutboxDrainerUnitTest.java:275 (and :133, :172) assert `atemptedDrainedCount()` == the batch size. That helper reads the `signing_record.persist` counter, but the drainer now increments that counter up-front for the whole batch — SigningRecordOutboxDrainer.java:125 `metrics.persist(...).increment(rows.size())` runs *before* persistBatchAndDeleteOutbox is even attempted. So `atemptedDrainedCount()` equals rows.size() whether the batch succeeds, falls back, or every row fails; it no longer reflects rows actually drained, only rows attempted. The real draining is verified by the `verify(writer).saveRecordAndDeleteOutbox(...)` calls in these tests, so coverage is not lost, but the count assertion is now tautological and its name ("...DrainedCount") is misleading post-change. Consider renaming the helper to attemptedCount and dropping it where it adds no signal beyond the verify(...) calls. Also fix the typo `atemptedDrainedCount` ->  `attemptedDrainedCount`.  Non-blocking.

src/main/java/com/otilm/core/signing/record/SigningRecordOutboxDrainer.java
```diff
@@ -105,38 +103,70 @@ private void drainBatches() {
...
+            return 0;
+
+        metrics.persist(SigningRecordPersistenceMode.DEFERRED_DURABLE.name()).increment(rows.size());
+
+        List<SigningRecord> records = rows.stream().map(this::mapAndEvict).toList();
+
+        try {
+            writer.persistBatchAndDeleteOutbox(records);
+            return records.size();
+        } catch (RuntimeException e) {
```
ivosh, line 137: **~ Minor** | Error handling

SigningRecordOutboxDrainer.java:132 — drainRows() catches `RuntimeException` from `persistBatchAndDeleteOutbox` and treats it as the crash-recovery-duplicate case, falling back to per-row draining. But the catch also fires on transient/infrastructure failures unrelated to a duplicate: a DB connection drop, a statement/lock timeout, or a serialization error on the bulk INSERT/flush. When that happens the fallback runs N more REQUIRES_NEW transactions (saveRecordAndDeleteOutbox) which will each hit the same transient failure, and attemptDrain() records a failed attempt via recordFailure() for every one — so a single transient outage increments `attempts` on the entire (otherwise healthy) batch, pushing rows toward the poison threshold. The transaction handling itself is sound (the inner REQUIRES_NEW rolls back cleanly and the merge fallback is idempotent), and per-row attempts also advanced on transient errors before this PR, so this is not a regression introduced by the diff — but the new bulk path amplifies it from one row to the whole batch per failed run. If the duplicate case is truly the only expected failure, consider narrowing the catch (e.g. distinguish a constraint/duplicate exception from a transient one) or not counting a whole-batch transient failure toward the per-row poison budget. Non-blocking.

ivosh (changes requested):
Thank you for the PR. There is also Copilot's review finding to address.

src/main/java/com/otilm/core/service/writer/signingrecord/SigningRecordWriter.java
```diff
@@ -30,6 +32,9 @@ public class SigningRecordWriter {
...
+    private EntityManager em;
+
```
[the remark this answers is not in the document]
LukasNajman: Fixed in 2ce3af55 — moved to constructor injection and named the field `entityManager` (matching `AttributeEngine`/`AuditLogServiceImpl`).

src/main/java/com/otilm/core/signing/record/SigningRecordOutboxDrainer.java
LukasNajman: Fixed in f7f553bd — switched to `signingRecord.getUuid()` to match the accessor used elsewhere in the drainer/writer.

src/main/java/com/otilm/core/service/writer/signingrecord/SigningRecordWriter.java
LukasNajman: Fixed in 2ce3af55 — moved `EntityManager` to constructor injection to match `SigningRecordOutboxDrainer`; the dependency is now `final`.

src/main/java/com/otilm/core/signing/record/SigningRecordOutboxDrainer.java
LukasNajman: Fixed in 93595dee — added the `jakarta.persistence.PersistenceException` import so the `{@link}` resolves.

src/test/java/com/otilm/core/signing/record/SigningRecordOutboxDrainerUnitTest.java
LukasNajman: Fixed in bb8b8852 — renamed the helper (and fixed the typo) to `attemptedCount`, since it only reflects rows attempted, not rows actually drained.

src/main/java/com/otilm/core/signing/record/SigningRecordOutboxDrainer.java
LukasNajman: Fixed in 67e600af — the fallback now only fires when the cause chain contains a Hibernate `ConstraintViolationException` (the actual duplicate-key signal). Any other failure (dropped connection, lock timeout, serialization error) is rethrown so the whole run aborts and the batch is retried whole on the next scheduled run, instead of driving every row toward the poison threshold via per-row `recordFailure()` calls.

ivosh (approved):

PR #1678, opened 2026-06-26
Add reference expander with per-object authorization

### LINKED ISSUE #1624: Reference expander with per-object authorization
### Description

Net-new `AttributeReferenceExpander`: expands RESOURCE/CREDENTIAL references in scope blobs and current attribute values with per-object calling-user authorization (a new guarded blob-loader per resource kind), secret-flow rules, recursion/cycle/depth caps, and fail-closed multi-select. Wired into callback dispatch and the v3 operation path (the operation path does not expand references today), coordinated with the in-flight authority-v3 work.

This is the larger, partly-open piece: the exact guarded entrypoint and required action per resource kind are settled in the core implementation plan, after which this issue is broken into task sub-issues.

### User Story

As a platform security owner, references to stored objects (credentials, authorities) are expanded to full material only for callers authorized for each object, and only server-side.

### Use Case

A stateless connector callback or operation that needs an authority's credential to authenticate to its upstream system.

### Acceptance Criteria

- [ ] **Two authorization modes, selected by request kind.** Callback mode (`expandForCaller`) authorizes the **calling user** per referenced object via a **two-step gate-then-load**: credentials gate on `getResourceObjectExternal(SecuredUUID)` (`CREDENTIAL:DETAIL`) — which returns only a name/uuid, **not** the blob — then the credential content loads only behind a passing gate for that UUID; AUTHORITY/ENTITY/LOCATION use a new `getAuthorizedObjectAttributes(SecuredUUID)` (`<KIND>:DETAIL`) that returns the content directly. The expander enforces the gate-precedes-load ordering (tested). Operation mode (`expandAsSystem`) resolves via a **fenced internal loader with no `@ExternalAuthorization`** — no per-object check, no elevated principal (the operation is authorized at the operation level; the value is the authority's own infrastructure credential). A system principal holding blanket `CREDENTIAL:DETAIL` is rejected.
- [ ] Fail closed: 403 if the caller lacks rights to any selected object (including multi-select).
- [ ] Recursion with cycle-safety and a depth cap; user-typed secrets pass through, stored secrets enter only via expansion.
- [ ] **Outbound containment check (v1).** The expander records the secret-typed values it expands per call (`expandedSecrets`); callback dispatch **rejects** any response whose `content`/`attributes` leaf values echo one of them (no expanded secret reaches the FE) and logs the connector contract violation. Bounded backstop (catches echo of this-call secrets, not general DLP); connector binding rule + startup self-validation remain primary.
- [ ] Wired into callback dispatch (caller mode) and the v3 operation path (system mode); the internal loader is package-private, operation-adapters-only, with a **reachability test** asserting it is unreachable from callback/controller paths; add a protocol/system-principal issuance regression test; coordinated with the authority-v3 work.
- [ ] Open mechanism decisions (guarded entrypoint + action per kind) recorded in the core plan; task sub-issues created afterwards.
- [ ] Depends on the NG dispatch rebuild (core #1621) and the interfaces contract (interfaces #725).

vyskocilm:
Fixes #1624

Add reference expander. Callback need resolve RESOURCE attribute references to real object attributes for connector. Must not leak - each reference load behind per-object KIND:DETAIL guard, fail-closed.

What:
- `expandForCaller` walk RESOURCE attribute references. Each load via SecuredUUID + `@ExternalAuthorization(KIND, DETAIL)` loader (new `getAuthorizedObjectAttributes` on Credential/Authority/Entity/Location). One unauthorized element abort whole call, no partial blob, no principal swap.
- Single-level expansion: only references directly in passed attributes resolved. Nested-in-blob references not wired (`attributesOf` return empty). Depth-cap + cycle-guard kernel = unit-tested scaffolding, become live with op-path follow-up.
- Typed loader registry + load-bearing ArchUnit fence keep callback path away from unguarded `findByUuid` / public list-loaders / `AuthHelper` (visibility fence impossible, those public).
- D19 outbound containment: record secret leaves expansion made (incl credential blob nested `SecretAttributeContentV2`, not only inline `ResourceSecretContentData`), reject response that echo one or carry populated secret shape.

Deferred to follow-up PR: operation-path / system-identity wiring. Known limit: LOCATION loader parent ENTITY:DETAIL is resource-level (loader carry only object uuid); LOCATION:DETAIL object gate is per-object.

Review: deep multi-agent + Copilot. Fixed before PR - honest single-level docs (recursion was inert), credential-blob secret recording, `SecretAttributeContentV2` structural reject, truthful LOCATION comment. 23/23 tests green.


vyskocilm:
## Independent review — Copilot CLI (done)

An independent Copilot CLI review of this PR's delta (`origin/main..feature/1624-reference-expander-authz`) was performed on 2026-06-26.

**Verdict: FIX-FIRST.** The primary control is confirmed solid (per-object `KIND:DETAIL` fires through the Spring proxy, fail-closed, no system-bypass; single-level expansion is documented-inactive scaffolding, not a bug). One new finding to resolve before merge:

- **`OutboundSecretContainment.rejectStructuralSecret()` doesn't descend POJO/DTO wrappers.** It recurses only `Iterable`/array/`Map`, so a `SecretAttributeContentV2`/`CredentialAttributeContentV2` nested inside the response DTO wrappers (`AttributeCallbackResponseDto` → `content`/`attributes` → `ResponseAttribute`) is not reached by the structural guard — asymmetric with `recordSecretLeaves()`, which does descend those wrappers. The value-echo arm (serializes the full response tree) is unaffected, so an echoed secret *value* is still caught; this gap is the secret-*shape* check.

Verified against the bytecode/source. Tracked separately: LOCATION parent-scope, ArchUnit direct-only fence.

vyskocilm:
## Update — review findings addressed; independent Copilot re-review: SHIP

Deep agentic review + independent Copilot CLI re-review of the fixes (round 2):

- Unified the structural-rejection and value-recording checks onto ONE wrapper-descending traversal (`walkSecretGraph`), closing the earlier gap where `rejectStructuralSecret` descended only collections and was a no-op on the real `AttributeCallbackResponseDto`.
- A Claude review agent then caught two issues in that unification, both fixed:
  - **Descend the universal `BaseAttribute.getContent()`** (was `DataAttribute`-only) so a secret nested in a `Metadata`/`Custom` attribute of the response `attributes` arm is reached.
  - **Strip the `@JsonTypeInfo "type"` discriminator** when recording a `SecretContent`, so a benign token like `"apiKey"` can't false-positive the value-echo check and reject a legitimate response.

Independent **Copilot CLI re-review: SHIP** (shared traversal closes the DTO-wrapper gap, recording limited to secret leaves, fail-closed, bounded/symmetric). Containment suite is 10 tests; 23/23 #1624 tests green. Branch force-pushed.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
```diff
@@ -0,0 +1,188 @@
+package com.otilm.core.attribute.engine;
+
+import com.otilm.api.exception.AttributeException;
+import com.otilm.api.exception.ConnectorException;
+import com.otilm.api.exception.NotFoundException;
+import com.otilm.api.model.client.attribute.RequestAttribute;
+import com.otilm.api.model.common.attribute.common.content.AttributeContentType;
+import com.otilm.api.model.common.attribute.v3.content.BaseAttributeContentV3;
```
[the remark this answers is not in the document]
vyskocilm: Fixed in 2d8fdc78 — removed the unused `AttributeContentType` import. (Minor: an unused import is a javac *warning*, not a compilation error; the module compiles and the 26 #1624 tests pass. Removed for cleanliness.)

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.UUID;
+
+/**
+ * Net-new shared facility (#1624): expands RESOURCE references inside a {@code List<RequestAttribute>} into the
+ * referenced object's full connector-consumable attribute blob, so a stateless NG connector can authenticate
+ * upstream without holding the credential material itself.
+ *
+ * <h2>Callback mode — the only mode implemented here</h2>
+ * {@link #expandForCaller(List)} authorizes the <em>ambient calling user</em> per referenced object through the
+ * resource's object-scoped {@code @ExternalAuthorization(<KIND>, DETAIL)} guarded entrypoint (a {@link SecuredUUID}
```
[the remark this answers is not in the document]
vyskocilm: Fixed in 2d8fdc78 — link corrected to `#expandForCaller(List, Set)`.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
```diff
@@ -0,0 +1,188 @@
...
+ * argument the auth aspect resolves the concrete object from). It <strong>fails closed</strong>: any
+ * {@code AccessDeniedException} from a DETAIL gate propagates and aborts the whole expansion — no partial blob is
+ * emitted, and multi-select selecting one unauthorized element among authorized ones fails the entire call.
+ * <p>
+ * The aspect reads the ambient Spring {@code SecurityContext}; there is no API to pass a context in, so there is
+ * deliberately no {@code SecurityContext} parameter, and this class performs <strong>no</strong>
+ * {@code setAuthentication} / principal swap of any kind (plan N3 invariant — kept closed; asserted by
+ * {@code expanderInvokesNoSetAuthentication}).
```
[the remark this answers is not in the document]
vyskocilm: Fixed in 2d8fdc78 — corrected to the actual test `AttributeReferenceExpanderIntegrationTest.expanderDoesNotMutateAmbientPrincipal`.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
```diff
@@ -0,0 +1,188 @@
...
+     * expanded into {@code expandedSecrets} for the outbound containment check on the connector's response.
+     *
+     * @param attributes      the request attributes to expand in place
+     * @param expandedSecrets out-param collecting secret values this call materialized (for outbound containment)
+     */
+    public List<RequestAttribute> expandForCaller(List<RequestAttribute> attributes, Set<String> expandedSecrets)
+            throws NotFoundException, AttributeException, ConnectorException {
+        expand(attributes, new java.util.HashSet<>(), 0, expandedSecrets);
+        return attributes;
+    }
```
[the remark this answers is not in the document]
vyskocilm: By design: #1624 is the facility slice. The production call sites are in the stacked PR #1679 (`CallbackServiceImpl.dispatchNg` and `AttributeCallbackScopeResolver.resolveScopeChain`), which threads the `expandedSecrets` set into `assertNoExpandedSecretOutbound`. This PR ships the facility; #1679 wires it into callback dispatch. Merge order: #1678 first, then #1679.

src/main/java/com/otilm/core/attribute/engine/OutboundSecretLeakException.java
```diff
@@ -0,0 +1,14 @@
+package com.otilm.core.attribute.engine;
+
+/**
+ * Raised by {@link OutboundSecretContainment} when an FE-bound callback response would carry secret material the
+ * expander materialized server-side — either by echoing an expanded secret value or by structurally containing a
+ * secret-bearing content shape. The dispatch arm (#1621) maps this to a connector contract violation
+ * (502 {@code CONNECTOR_PROTOCOL_VIOLATION}) and does not forward the response.
```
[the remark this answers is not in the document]
vyskocilm: Fixed in 2d8fdc78 — reworded to describe actual behaviour (thrown before forwarding, fail-closed; message is a fixed, secret-free string). The 502 `CONNECTOR_PROTOCOL_VIOLATION` mapping is not implemented in this PR; a dedicated connector-contract status is now noted as a follow-up rather than claimed.

src/main/java/com/otilm/core/service/impl/LocationServiceImpl.java
```diff
@@ -747,6 +750,22 @@ public NameAndUuidDto getResourceObjectExternal(SecuredUUID objectUuid) throws N
         return new NameAndUuidDto(String.valueOf(objectUuid), location.getName());
     }
 
+    @Override
+    @ExternalAuthorization(resource = Resource.LOCATION, action = ResourceAction.DETAIL, parentResource = Resource.ENTITY, parentAction = ResourceAction.DETAIL)
```
lubomirw: **[Medium]** Parent `ENTITY:DETAIL` gate here is resource-level, not object-scoped.

This loader declares `parentResource=ENTITY` but takes only a plain `SecuredUUID` (no `SecuredParentUUID`), so the auth aspect evaluates `ENTITY:DETAIL` with no object UUID — unlike the sibling guards (`getLocation`/`editLocation`/…) which scope the parent to the location's owning entity. A caller with broad `ENTITY:DETAIL` but no access to *this* location's owning entity passes the parent gate. The object-scoped `LOCATION:DETAIL` gate still holds, so this is a parent-dimension weakening, not a full bypass — and the inline comment already flags it as deferred.

_Suggested:_ thread the owning-entity UUID as a `SecuredParentUUID` so the parent check is object-scoped, or drop the parent annotation rather than leave a check that reads as enforcement it doesn't provide. If the resource-level interim is intended, worth explicit sign-off.

<sub>flagged by guided-pr-review + superpowers code review</sub>

src/main/java/com/otilm/core/attribute/engine/OutboundSecretContainment.java
```diff
@@ -0,0 +1,212 @@
...
+     * blob with the same traversal the structural check uses records those too (O1).
+     */
+    void recordExpandedSecrets(ResourceObjectContentData blob, Set<String> expandedSecrets) {
+        if (expandedSecrets == null) {
+            return;
+        }
+        walkSecretGraph(blob, 0, node -> recordIfSecretLeaf(node, expandedSecrets));
+    }
+
+    private void recordIfSecretLeaf(Object node, Set<String> sink) {
```
lubomirw, line 79: **[Low]** `recordIfSecretLeaf` records every scalar of a `SecretContent`, including non-secret identifiers.

For a `ResourceSecretContentData` it serializes the whole `SecretContent` and `collectScalarStrings` adds all scalars (only the `type` discriminator is stripped), so a non-secret field like a basic-auth **username** becomes an echo needle. A connector that legitimately echoes that username then trips the value-echo check and has its whole response refused (false positive). Fails safe (refuse, not leak), so non-blocking.

_Suggested:_ record only genuinely secret-bearing fields per `SecretContent` subtype (password/content/key/token), skip low-entropy identifiers — or document the over-strict behavior and pin it with a test.

src/main/java/com/otilm/core/attribute/engine/OutboundSecretContainment.java
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * Walk the secret-relevant object graph of a content payload, invoking {@code visitor} on every node, and
+     * descend the wrappers that can nest secret content: collections AND the typed DTO wrappers
+     * (AttributeCallbackResponseDto / ResponseAttribute / DataAttribute / ResourceObjectContent /
+     * ResourceSimpleContentData / Credential + Secret content) that a Jackson value-tree walk cannot type-detect.
+     * Both the value-recording and structural-rejection checks share this traversal so they cannot diverge.
+     */
+    private void walkSecretGraph(Object node, int depth, java.util.function.Consumer<Object> visitor) {
+        if (node == null || depth > MAX_WALK_DEPTH) {
```
lubomirw: **[Low]** Structural secret-shape walk fails *open* past `MAX_WALK_DEPTH` (12).

`walkSecretGraph` returns silently when `depth > 12`, so a secret-bearing shape (e.g. populated `ResourceSecretContentData` or `CredentialAttributeContentV2`) nested deeper than 12 wrappers escapes structural rejection — the opposite fail-direction from the expander's depth cap, which throws. Value-echo is unbounded so it still catches verbatim recorded values, but the structural check (the one that fires with no recorded values) has the silent gap. Payloads >12 deep are unlikely; the fail-open direction in a security control is the concern.

_Suggested:_ on exceeding `MAX_WALK_DEPTH`, fail closed (throw `OutboundSecretLeakException` — "response too deep to verify"), consistent with `ReferenceExpansionException.depthExceeded`.

<sub>flagged by guided-pr-review + superpowers code review</sub>

src/main/java/com/otilm/core/attribute/engine/OutboundSecretContainment.java
```diff
@@ -0,0 +1,212 @@
...
+ * </ol>
+ * Refuse (do not strip) is the fail-closed default: a connector that returns expanded secret content is violating
+ * the contract, and silently stripping would mask the violation.
+ */
+@Component
+public class OutboundSecretContainment {
+
+    private static final Logger logger = LoggerFactory.getLogger(OutboundSecretContainment.class);
+
+    private final ObjectMapper objectMapper = new ObjectMapper();
```
lubomirw: **[Low]** Bespoke `new ObjectMapper()` instead of the Spring-configured bean.

`valueToTree`/serialization here uses a private mapper whose config (naming, `@JsonInclude`, custom serializers, polymorphic handling) can differ from the application mapper that produces the actual wire bytes — so the value-echo comparison may serialize a secret differently than it appears on the wire and miss an echo, or throw on a type the default mapper can't serialize (a 500 rather than a containment decision).

_Suggested:_ constructor-inject the managed `ObjectMapper`; guard `valueToTree` against serialization exceptions.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
```diff
@@ -0,0 +1,187 @@
...
+    // DEFERRED to the operation-path PR (Task 3) — do NOT wire the 8 AuthorityProviderV3Adapter callsites.
+    //
+    // Operation mode must resolve through a fenced loader carrying NO @ExternalAuthorization and performing NO
+    // setAuthentication (it runs under the operation path's existing principal). The live operation primitive is
+    // the public ConnectorRequestAttributesBuilder (plan-review P4); the safety rests on extending the ArchUnit
+    // fence to cover the operation entrypoint, NOT on a new unguarded twin. Until that PR:
+    //   - no expandViaOperationLoader is exposed,
+    //   - no adapter callsite is changed,
+    //   - the v2-adapter-does-not-expand / noDoubleExpansion / protocol-principal regression tests live there.
+    // See ingest/plans-v2/1624-...-plan-v2.md §"Task 3" and plan-review.md P2/P4.
```
lubomirw: **[Low]** Internal planning-artifact path leaks into a shipped comment.

`// See ingest/plans-v2/1624-...-plan-v2.md §"Task 3" and plan-review.md P2/P4.` points at the planning workspace, which won't exist for readers of the shipped repo.

_Suggested:_ drop the path reference; keep the surrounding behavioral note about deferred operation mode.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
```diff
@@ -0,0 +1,187 @@
...
+ * <h2>Callback mode — the only mode implemented here</h2>
+ * {@link #expandForCaller(List, Set)} authorizes the <em>ambient calling user</em> per referenced object through the
+ * resource's object-scoped {@code @ExternalAuthorization(<KIND>, DETAIL)} guarded entrypoint (a {@link SecuredUUID}
+ * argument the auth aspect resolves the concrete object from). It <strong>fails closed</strong>: any
+ * {@code AccessDeniedException} from a DETAIL gate propagates and aborts the whole expansion — no partial blob is
+ * emitted, and multi-select selecting one unauthorized element among authorized ones fails the entire call.
+ * <p>
+ * The aspect reads the ambient Spring {@code SecurityContext}; there is no API to pass a context in, so there is
+ * deliberately no {@code SecurityContext} parameter, and this class performs <strong>no</strong>
+ * {@code setAuthentication} / principal swap of any kind (plan N3 invariant — kept closed; asserted by
```
lubomirw: **[Nit]** Generalize the internal plan/review tags to describe the invariant in code terms (keep the prose).

Several Javadoc tags anchor rationale to internal planning artifacts rather than the code: `plan N3 invariant` (this line), `plan-review P1` (L50), `DEFERRED — operation mode (Task 3)` (L53), plus `plan Q3-device-1` in `CallerAuthorizedReferenceLoaderRegistry` (L14) and `#1624 AC4 / D19` in `OutboundSecretContainment` (L26).

_Suggested:_ reword each tag to state the invariant/decision in code terms (e.g. "invariant: the expander never swaps the principal" instead of "plan N3 invariant") and drop the artifact identifiers, keeping the explanatory prose.

src/main/java/com/otilm/core/service/impl/CredentialServiceImpl.java
```diff
@@ -290,6 +293,23 @@ public NameAndUuidDto getResourceObjectExternal(SecuredUUID objectUuid) throws N
...
 
+    @Override
+    @ExternalAuthorization(resource = Resource.CREDENTIAL, action = ResourceAction.DETAIL)
+    public ResourceObjectContentData getAuthorizedObjectAttributes(SecuredUUID objectUuid) throws NotFoundException {
+        // Single guarded gate-then-load: CREDENTIAL:DETAIL on this exact object (the SecuredUUID arg) is
+        // enforced by the aspect before the body runs, so the credential material below is only loaded behind
+        // a passing per-object check. The reference expander resolves credentials through THIS method only —
+        // never the unguarded private getCredentialEntity/findByUuid nor the resource-level
+        // loadFullCredentialData(List) (enforced by the ArchUnit fence test).
+        Credential credential = credentialRepository.findByUuid(objectUuid)
```
lubomirw: **[Nit]** Inconsistent `findByUuid` argument vs the sibling loaders.

This passes the `SecuredUUID` directly to `findByUuid(objectUuid)` (the `SecurityFilterRepository.findByUuid(SecuredUUID)` overload), whereas the AUTHORITY/ENTITY/LOCATION loaders pass `objectUuid.getValue()`. Both compile and behave the same, but the divergence invites confusion about whether an extra security filter applies.

_Suggested:_ use `objectUuid.getValue()` consistently across all four loaders.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /**
+     * The attributes nested inside an already-loaded blob, as the {@code List<RequestAttribute>} the recursion
+     * would consume. <strong>Returns empty unconditionally today</strong>: nested-blob resolution is not yet
+     * wired, so expansion is single-level. A loaded blob's nested attributes are {@code ResponseAttribute}, not
+     * {@code RequestAttribute}, and are not re-resolvable in place; translating them into recursion input is the
+     * operation-path follow-up's job. Until then a reference nested inside a loaded blob stays a bare reference.
+     */
+    private static List<RequestAttribute> attributesOf(ResourceObjectContentData blob) {
```
lubomirw, line 174: **[Nit]** Dead branch in `attributesOf`.

It checks `instanceof ResourceSimpleContentData simple && simple.getAttributes() != null` but returns `List.of()` in both the matched and fall-through paths, leaving the `simple` binding unused — Sonar will flag the unused pattern binding and constant return, and it obscures that nested resolution is a deliberate no-op today.

_Suggested:_ collapse to a single `return List.of();` (keep the Javadoc explaining nested resolution is deferred); restore real extraction in the op-path follow-up.

<sub>flagged by guided-pr-review (+ orchestrator)</sub>

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    private void expandElement(ResourceObjectContent element, Set<String> visited, int depth, Set<String> expandedSecrets)
+            throws NotFoundException, AttributeException, ConnectorException {
+        ResourceObjectContentData ref = element.getData();
+        if (ref == null || ref.getResource() == null || ref.getUuid() == null) {
+            return;
+        }
+        AttributeResource kind = ref.getResource();
+        UUID uuid = UUID.fromString(ref.getUuid());
```
lubomirw: **[Nit]** `UUID.fromString` on a malformed reference UUID throws raw `IllegalArgumentException`.

A connector/client-supplied bad UUID surfaces as an unchecked 500 rather than a clean validation error.

_Suggested:_ validate or catch and raise `AttributeException`/`NotFoundException` so a malformed reference fails as 4xx, not 500.

<sub>flagged by orchestrator (diff read)</sub>

src/main/java/com/otilm/core/service/impl/LocationServiceImpl.java
vyskocilm: Took option (b) in a005f6d3 — dropped the `parentResource=ENTITY` annotation. Object-scoped `LOCATION:DETAIL` on objectUuid stays, now consistent with the Credential/Authority/Entity loaders (which carry no parent). The resource-level parent check added no object-scoped protection, so removing the annotation that reads as enforcement it can't provide is the honest fix. Per-entity scoping would need the owning-entity uuid threaded through the `loadAuthorized(SecuredUUID)` contract — deferred with the op-path.

src/main/java/com/otilm/core/attribute/engine/OutboundSecretContainment.java
vyskocilm: Fixed in a005f6d3 — `recordIfSecretLeaf` strips `username` + `keyStoreType` (plus the `type` discriminator), so low-entropy identifiers aren't recorded as echo needles; the secret fields (password/content) still record. Test `recordsSecretFieldButNotLowEntropyIdentifier`. Kept strip-by-name over per-subtype allowlisting to avoid a fail-open regression when a new secret type/field is added.

src/main/java/com/otilm/core/attribute/engine/OutboundSecretContainment.java
vyskocilm: Fixed in a005f6d3 — `walkSecretGraph` takes `failClosedOnDepth`: the structural reject path throws `OutboundSecretLeakException` past `MAX_WALK_DEPTH` (fail-closed, matching `ReferenceExpansionException.depthExceeded`). The record path stays best-effort (silent stop) so a deep legitimate expansion isn't spuriously aborted — the unbounded value-echo scan is its net. Test `structuralCheckFailsClosedWhenResponseTooDeeplyNested`.

src/main/java/com/otilm/core/attribute/engine/OutboundSecretContainment.java
vyskocilm: Fixed in a005f6d3 — constructor-injects `@Qualifier("jacksonObjectMapper")` (the wire bean) instead of `new ObjectMapper()`; both `valueToTree` calls guarded — reject path throws (fail-closed), record path skips (best-effort).

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
vyskocilm: Fixed in a005f6d3 — dropped the `ingest/plans-v2` path; kept the behavioral note about deferred operation mode.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
vyskocilm: Fixed in a005f6d3 — reworded the plan/Q/P/Task/AC/D19 tags to code-term invariants across the engine package (AttributeReferenceExpander, CallerAuthorizedReferenceLoader[Registry], OutboundSecretContainment, ReferenceExpansionException).

src/main/java/com/otilm/core/service/impl/CredentialServiceImpl.java
vyskocilm: Fixed in a005f6d3 — `findByUuid(objectUuid.getValue())`, consistent with the other loaders. Confirmed the `SecuredUUID` overload applied no extra filter (`SecurityFilterRepositoryImpl` delegates with a null where-clause), so behavior is unchanged.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
vyskocilm: Fixed in a005f6d3 — collapsed `attributesOf` to a single `return List.of();` and removed the now-unused `ResourceSimpleContentData` import. The Javadoc (nested resolution deferred) is kept.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
vyskocilm: Fixed in a005f6d3 — `UUID.fromString` is wrapped; a malformed reference uuid now raises `AttributeException` (400), not a raw `IllegalArgumentException` (500). Test `malformedReferenceUuidFailsAsValidationNotServerError`.

src/test/java/com/otilm/core/architecture/AttributeReferenceExpanderFenceArchTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /** The classes that make up callback mode's resolution path. */
+    private static final List<Class<?>> CALLBACK_PATH = List.of(
+            AttributeReferenceExpander.class,
+            CallerAuthorizedReferenceLoaderRegistry.class);
+
+    @ArchTest
+    static final ArchRule callback_path_must_not_call_unguarded_credential_list_loader =
+            noClasses()
+                    .that().belongToAnyOf(AttributeReferenceExpander.class, CallerAuthorizedReferenceLoaderRegistry.class)
```
lubomirw: **[Low]** The ArchUnit fence only constrains two *named* classes, not the whole engine package.

`@AnalyzeClasses` (L35) imports `com.otilm.core.attribute.engine`, but all four rules use `belongToAnyOf(AttributeReferenceExpander.class, CallerAuthorizedReferenceLoaderRegistry.class)`. A **future** helper/adapter added to this package that calls `loadFullCredentialData` / `loadResourceObjectContentData` / `ConnectorRequestAttributesBuilder` would satisfy none of the rule predicates and silently bypass the fence. Since this test is documented as "THE load-bearing security boundary … carries the ENTIRE fence weight," pinning only today's two classes undercuts that guarantee.

_Suggested:_ change the predicate to `that().resideInAPackage("com.otilm.core.attribute.engine..")` so the fence covers the whole package, not just the classes named at authoring time.

<sub>flagged by Copilot CLI review</sub>

src/main/java/com/otilm/core/service/impl/LocationServiceImpl.java
lubomirw: Additional fix option (from a second-pass Copilot review): instead of threading a `SecuredParentUUID`, an explicit in-body ownership check before returning the blob also works — resolve `location.getEntityInstanceReference().getUuid()` and verify the caller's `ENTITY:DETAIL` against *that* entity. Either way the goal is to bind the parent gate to this location's owning entity rather than leaving it resource-scoped. This finding is now corroborated by 3 independent reviewers.

src/main/java/com/otilm/core/attribute/engine/OutboundSecretContainment.java
lubomirw: A second-pass Copilot review rated this **Medium** rather than Low, emphasizing the leak-direction risk: if the bare `ObjectMapper` serializes a `SecretContent` subtype differently than the app-configured mapper (missing modules/mixins/custom serializers), the recorded scalar may not match the wire form — so the value-echo check **silently misses** a secret the connector echoes (fail-open, not merely a 500). Two reviewers independently flagged this, so worth weighting toward Medium.

src/main/java/com/otilm/core/service/impl/LocationServiceImpl.java
vyskocilm: Implemented the stronger option in 8fd28ad5 — kept `@ExternalAuthorization(LOCATION, DETAIL)` and added an in-body owning-entity gate: `if (location.getEntityInstanceReference() != null) permissionEvaluator.entityInstance(location.getEntityInstanceReference().getSecuredUuid())` before building the blob, failing closed (AccessDeniedException) when the caller lacks ENTITY:DETAIL on the owning entity. Added a new `PermissionEvaluator.entityInstance` = `@ExternalAuthorization(ENTITY, DETAIL)` (the existing `authorityInstance` is AUTHORITY:DETAIL — wrong resource). Null owner skips (mirrors evaluatePermissionChain). New test `locationExpansionFailsClosedWhenDeniedOwningEntityDetail` pins it.

src/test/java/com/otilm/core/architecture/AttributeReferenceExpanderFenceArchTest.java
vyskocilm: Fixed in 8fd28ad5 — broadened all four rules to `resideInAPackage("com.otilm.core.attribute.engine..")`. Note the literal whole-package form would have broken the build: `AttributeEngine` (engine pkg) legitimately uses `AuthHelper` (rule 4) and `ConnectorRequestAttributesBuilder` legitimately calls the list-loaders / is the fenced op-mode primitive (rule 3). So the rules exclude exactly those two by class (`.areNotAssignableTo(...)`) — any NEW engine helper is now fenced by default and must be explicitly, reviewably excluded. Dropped the dead CALLBACK_PATH field. Fence 4/4 green.

src/main/java/com/otilm/core/attribute/engine/OutboundSecretContainment.java
vyskocilm: Already resolved in a005f6d3 (the previous round): `OutboundSecretContainment` now constructor-injects `@Qualifier("jacksonObjectMapper")` — the same managed mapper that produces the wire bytes — so the recorded scalar matches the wire form and the value-echo can't silently miss (no fail-open), and `valueToTree` is guarded fail-closed. No bespoke `new ObjectMapper()` remains.

lubomirw (approved):

lubomirw:
## ✅ Verification — reference-expander happy path (PASS)

Ran the expander integration suite against this PR's head (`8fd28ad5`) to confirm the happy path executes through the **real** per-object authorization aspect, not a mock of it.

**Method:** built the branch in an isolated worktree and ran `AttributeReferenceExpanderIntegrationTest` via Maven — full ILM Core Spring context (Spring Boot 3.5.14, Java 21) against a **testcontainers Postgres** (`ddl-auto: create-drop`), with `OpaClient` mocked (default-allow; flipped to deny per case). So the `@ExternalAuthorization(…, DETAIL)` aspect and a real DB are exercised end-to-end.

**Result:**
```
[INFO] Running com.otilm.core.attribute.engine.AttributeReferenceExpanderIntegrationTest
[INFO] Tests run: 5, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 114.3 s
[INFO] BUILD SUCCESS
```

Methods exercised (all green):
- `authorizedCallerGetsCredentialBlobExpanded` — the happy path
- `authorizedCallerExpandsAuthorityEntityAndLocationBlobs` — AUTHORITY/ENTITY/LOCATION happy paths
- `deniedCallerNeverReceivesBlob_failsClosed`
- `locationExpansionFailsClosedWhenDeniedOwningEntityDetail` — covers the in-body owning-entity `ENTITY:DETAIL` fix
- `expanderDoesNotMutateAmbientPrincipal`

**Scope / caveat:** this is **not** an HTTP callback-API call. In this PR the expander has no production caller (the `CallbackServiceImpl → expander → outbound containment` dispatch wiring lands in the follow-up #1679), so the callback HTTP seam can't be exercised here — only the expander's behavior through Spring + DB + the authz aspect. A true end-to-end callback-API run should be done on #1679.

<sub>Verification run locally against `8fd28ad5`.</sub>

PR #1679, opened 2026-06-26
Add Attributes v2 NG callback dispatch and registry resolution

### LINKED ISSUE #1621: NG callback dispatch and scope resolver
### Description

Rewrite Core's callback dispatch (`CallbackServiceImpl.getCallbackObject` / `resourceCallback`) into the per-callback dispatch ladder (core-handled -> legacy `callbackContext` -> NG `dependsOn`) plus a table-driven scope-chain resolver. Rebuilds the resource-scoped callback route, which is currently broken for v3 authorities (it calls the v1 client with a null `authorityInstanceUuid`) and throws "not supported" for `TOKEN_PROFILE`. The NG branch routes to the envelope builder/dispatch delivered in #1622.

### User Story

As an operator, dynamic attribute callbacks work on stateless NG-connector forms (RA profile, issuance, token profile, location) with the correct scope context injected.

### Use Case

RA-profile form under a v3 authority; certificate issuance form (raProfile -> authority scope chain).

### Acceptance Criteria

- [ ] Ladder order matches the current core-handled paths (`core/getCredentials` -> RESOURCE -> legacy `callbackContext` -> NG `dependsOn`); the v1/v2 path is unchanged (one new branch in the default arm, no new rung structure).
- [ ] Per-callback gating by declaration shape (`callbackContext` -> legacy, `dependsOn` -> NG).
- [ ] Table-driven scope resolver: RA_PROFILE -> [authority]; CERTIFICATE -> [authority, raProfile]; TOKEN_PROFILE -> [tokenInstance]; CRYPTOGRAPHIC_KEY -> [tokenInstance, tokenProfile]; LOCATION -> [entity]; connector-scoped -> [].
- [ ] v3 authority scoped callback succeeds (regression for the null-uuid breakage); `TOKEN_PROFILE` no longer throws; issuance scope walks raProfile -> authority.
- [ ] NG branch routes to the #1622 envelope dispatch; legacy `attributeCallback` default unchanged.
- [ ] Depends on the interfaces contract and clients (interfaces #725, #726).

### Scope boundary (vs #1622)

This issue and #1622 share the `CallbackServiceImpl` rewrite and are reviewed and merged together. **This issue (#1621)** owns the dispatch-ladder branch, the table-driven scope resolver, and the `resourceCallback` rebuild. **#1622** owns `ConnectorApiFactory.getAttributesApiClientV2`, the definition-resolution ladder, and the NG envelope builder/dispatch incl. the `X-ILM-Attribute` header.

### LINKED ISSUE #1622: Attribute definition registry resolution
### Description

Name-keyed definition-resolution ladder with a `GET /v2/attributes` fallback on cache miss; add `ConnectorApiFactory.getAttributesApiClientV2`; build and dispatch the NG callback envelope; ingest GROUP-callback child definitions; derive `connectorInterface`/`interfaceVersion` from the form context (no new DB columns). When a connector returns `ATTRIBUTE_DEFINITION_NOT_FOUND`, refresh that connector's registry once and retry.

### User Story

As Core, I can resolve attribute definitions that were never listed (API-only clients, never-materialised GROUP children, metadata) without re-listing whole schemas.

### Use Case

Validating an operation request from a stored API/cURL client that references a definition Core has not previously listed.

### Acceptance Criteria

- [ ] Resolution ladder: stored (name-keyed) -> miss -> `GET /v2/attributes?uuids=` fetch + ingest -> still unresolved -> 422 naming the UUID(s).
- [ ] **Definition keying.** Registry-origin definitions ingest as `operation = null` (global) and are read back with an explicit `operation = null` filter; the finder is operation-scoped (or prefers the global row) so multiple `(connectorUuid, name)` rows differing by operation can **never** throw `IncorrectResultSizeDataAccessException` (500). Core neither relies on nor enforces connector UUID-uniqueness in its own table.
- [ ] `ATTRIBUTE_DEFINITION_NOT_FOUND` from a connector -> refresh registry once + retry.
- [ ] NG callback envelope built and dispatched via `getAttributesApiClientV2`. (The non-binding `X-ILM-Attribute` observability header is split to a follow-up: core #1696.)
- [ ] GROUP callback arm ingests child definitions.
- [ ] `connectorInterface`/`interfaceVersion` derived from form context; no `attribute_definition` schema columns added.
- [ ] `getAttributesApiClientV2` added to `ConnectorApiFactory`.
- [ ] METADATA-definition miss-path resolution handled (separate ingest path).
- [ ] Depends on the interfaces clients (interfaces #726).

### Scope boundary (vs #1621)

This issue and #1621 share the `CallbackServiceImpl` rewrite and are reviewed and merged together. **This issue (#1622)** owns `ConnectorApiFactory.getAttributesApiClientV2`, the definition-resolution ladder, and the NG envelope builder/dispatch. **#1621** owns the dispatch-ladder branch, the scope resolver, and the `resourceCallback` rebuild.

vyskocilm:
Fixes #1621
Fixes #1622

Add Attributes v2 NG callback dispatch + registry resolution. Builds on the reference expander from #1624 (now merged to main), so this PR stands alone — its diff is just the NG-dispatch change.

#1621 — NG callback dispatch + scope resolver:
- Route `dependsOn` callbacks (no legacy callbackContext) to new `NgCallbackDispatcher`; legacy callbacks unchanged.
- `AttributeCallbackScopeResolver` builds the parent-first scope chain per resource, expands each scope object's attributes for the caller (via the merged #1624 expander), threading one `expandedSecrets` set into the outbound containment gate.
- `resourceCallback` handles v3 authorities (skip the null-uuid v1 call), adds TOKEN_PROFILE/CERTIFICATE arms, stamps `connectorInterface` per resource.

#1622 — registry resolution + dispatcher:
- POST the v2 callback envelope outside any transaction (collaborator bean, `NOT_SUPPORTED`); ingest response child defs in a short tx — containment gate runs BEFORE ingest.
- `AttributeDefinitionResolver`: resolve referenced def, fetch from the connector registry once on a miss, 422 (naming the uuid) when unresolvable.
- UUID-first strict accessor closes the non-unique `(type,connector,name)` -> 500 class (initial dispatch + retry); name-only fallback preserves legacy.
- `getAttributesApiClientV2` factory + REST/MQ bean registration.

Reviewed: 4 deep agents + Copilot CLI across rounds; all findings fixed. Affected suite green on current main.


vyskocilm:
## Independent review — Copilot CLI (done)

An independent Copilot CLI review of this PR's own delta (the NG-dispatch commit, `feature/1624-reference-expander-authz..feature/1621-1622-ng-callback-dispatch-registry`) was performed on 2026-06-26.

**Verdict: FIX-FIRST.** Confirmed good: connector POST outside any tx (NOT_SUPPORTED collaborator bean), no raw-message leak, legacy callbacks pay no scope-chain auth. Three new findings to resolve before merge (all verified against source/bytecode):

- **Initial NG resolution is name-only.** `getBaseAttributeFromExistingDefinition` resolves the definition by `callback.getName()` and ignores `RequestAttributeCallback.uuid` (which exists). UUID-first strict resolution only runs on the `ATTRIBUTE_DEFINITION_NOT_FOUND` retry — so a duplicate `(type, connector, name)` can dispatch the **wrong `attributeUuid`** on the first attempt.
- **Containment runs after ingest.** `NgCallbackDispatcher.dispatchNgCallback` calls `handleResponseArms` (child-definition ingest + commit) **before** `assertNoExpandedSecretOutbound`. A response that violates containment is partly persisted before the fail-closed gate fires.
- **`connectorInterface` is sent null but is `@NotNull`/REQUIRED** on `AttributeCallbackRequestDto`. `interfaceVersion` is threaded; `connectorInterface` is hard-null in `NgDispatchContext`.

Depends on #1678 (stacked).

vyskocilm:
## Update — review findings addressed (deep agents + independent Copilot CLI)

Round-2 fixes applied and re-reviewed:

- **Initial NG resolution prefers the referenced attributeUuid** (strict accessor) with name-only **fallback** when the uuid is absent or doesn't match a stored `(attributeUuid,name)` row — closes the duplicate-name wrong-row risk on the first dispatch while preserving legacy callers (`RequestAttributeCallback.uuid` is overloaded; some legacy paths put the connector uuid there).
- **Containment runs BEFORE child-definition ingest**, so a response that violates containment is never partly persisted.
- **`connectorInterface` is stamped from form context** for every `resourceCallback` arm — `interfaceCodeOf(authority)` for RA_PROFILE/CERTIFICATE, `CRYPTOGRAPHY` for TOKEN_PROFILE/CRYPTOGRAPHIC_KEY, `ENTITY` for LOCATION (the independent Copilot re-review flagged the non-authority arms still sent `null` into the `@NotNull` envelope; now fixed). The bare connector-level entrypoint has no form context, so `connectorInterface` stays null there — not the FE NG path.
- Null-name guard on the strict accessor (no `List.of(null)` NPE).

A Claude review agent returned SHIP on the resolution + connectorInterface dimension. 39/39 affected tests green. Depends on #1678 (stacked). Branch force-pushed.

src/main/java/com/otilm/core/service/callback/AttributeDefinitionResolver.java
```diff
@@ -0,0 +1,142 @@
...
+            return;
+        }
+
+        TransactionStatus tx = transactionManager.getTransaction(new DefaultTransactionDefinition());
+        try {
+            ingest(connectorUuid, type, definitions);
+            transactionManager.commit(tx);
+        } catch (RuntimeException e) {
+            transactionManager.rollback(tx);
+            throw e;
```
lubomirw: **[High]** | Error handling

`fetchAndIngest` catches `RuntimeException`, rolls back, and rethrows `e` raw. A `DataIntegrityViolationException` from the commit carries SQL/constraint text that then propagates to the API response — unlike `ingest` and `NgCallbackDispatcher.handleResponseArms`, which wrap into a sanitised `ValidationException`.

_Suggested:_ after rollback, log `e.getMessage()` at DEBUG and throw a sanitised `ValidationException` (e.g. "Fetched attribute definitions could not be ingested for connector `<uuid>`"), mirroring `handleResponseArms`.

<sub>flagged by Copilot CLI reviewer</sub>

src/main/java/com/otilm/core/service/callback/AttributeDefinitionResolver.java
```diff
@@ -0,0 +1,142 @@
...
+     * @param type          DATA, GROUP or META
+     * @return the resolved definition
+     * @throws ValidationException (422) if unresolvable after the single registry refresh
+     */
+    public BaseAttribute resolve(ApiClientConnectorInfo connector, UUID attributeUuid, String name, AttributeType type)
+            throws ConnectorException, ValidationException {
+        UUID connectorUuid = UUID.fromString(connector.getUuid());
+
+        BaseAttribute stored = readStored(connectorUuid, attributeUuid, name, type);
+        if (stored != null) {
```
lubomirw, line 78: **[Medium]** | Correctness

On `ATTRIBUTE_DEFINITION_NOT_FOUND`, `NgCallbackDispatcher` calls `resolve()` to "refresh" before its single retry — but `resolve()` reads stored first and returns the still-present local definition here, without ever calling `fetchAndIngest` (the registry refresh only runs when `stored == null`). So the retry rebuilds an identical envelope and re-POSTs the payload the connector just rejected, defeating the single-shot refresh.

_Suggested:_ give the not-found retry a forced-refresh path (e.g. `resolveForced`/`forceRefresh` that runs `fetchAndIngest` before re-reading). Note `NgCallbackDispatcherTest` masks this by mocking `resolve()` to return a *different* object — add a test where the stored row is unchanged and assert `fetchAndIngest` runs on retry.

src/main/java/com/otilm/core/service/impl/CallbackServiceImpl.java
```diff
@@ -123,12 +157,31 @@ public Object callback(UUID connectorUuid, RequestAttributeCallback callback) th
...
         AttributeCallback attributeCallback = getAttributeCallback(attribute);
 
         AttributeResource attributeResource = getAttributeResource(attribute);
+
+        // NG (dependsOn) dispatch: routed only when the definition declares a dependsOn callback and carries no
+        // legacy callbackContext. It is checked BEFORE the legacy callback validation because the NG shape has no
+        // callbackContext/callbackMethod (which AttributeDefinitionUtils.validateCallback requires); the NG
+        // declaration is validated at ingest instead (#1622 Task 4b). Both-set is rejected at ingest, so the
+        // callbackContext == null conjunct here is defensive. Legacy callbacks fall through unchanged.
+        if (isNgCallback(attributeCallback)) {
```
lubomirw, line 188: **[Medium]** | Correctness

`isNgCallback(attributeCallback)` is called before `validateCallback`, but `getAttributeCallback` can return `null` for a DATA attribute stored without a callback, and `isNgCallback` immediately dereferences `attributeCallback.getDependsOn()` — turning a controlled 400 into an unhandled 500 NPE.

_Suggested:_ guard with `attributeCallback != null && isNgCallback(attributeCallback)`, or make `if (attributeCallback == null) return false;` the first line of `isNgCallback`.

<sub>flagged by Copilot CLI reviewer</sub>

src/main/java/com/otilm/core/service/callback/NgCallbackDispatcher.java
```diff
@@ -0,0 +1,193 @@
...
+        if (response == null) {
+            throw new ValidationException(ValidationError.create(
+                    "Connector returned an empty attribute-callback response"));
+        }
+
+        // Fail-closed leak gate (#1624) runs BEFORE any ingest/commit: a connector must not echo back a secret
+        // value Core expanded server-side, nor carry a secret-bearing shape. Rejecting first means a violating
+        // response never has its child definitions persisted (handleResponseArms commits to the registry).
+        outboundContainment.assertNoExpandedSecretOutbound(response, expandedSecrets);
+        handleResponseArms(connectorUuid, response);
```
lubomirw, line 146: **[Low]** | Correctness

`AttributeCallbackResponseDto.isExactlyOneArmSet` documents "Core validates the response on receipt," but no check is enforced here. Both arms set → `handleResponseArms` persists the attributes arm *and* `dispatchNg` returns the content arm (a contract-violating response both mutates the registry and returns data). Both arms null → `dispatchNg` (CallbackServiceImpl:310) returns `null` to the FE, where the legacy path returned the raw response.

_Suggested:_ after the containment gate and before `handleResponseArms`, enforce the XOR — reject with `ValidationException` when `(content == null) == (attributes == null)` (this single check covers both the both-set and both-null edges).

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -215,21 +215,91 @@ private static CustomAttribute getCustomAttributeWithDecryptedContentFromRelatio
...
+    }
+
+    /**
+     * Name-only resolution for legacy callers that have no referenced UUID in scope. Uses a List finder plus a
+     * deterministic tiebreak so it can never throw {@code IncorrectResultSizeDataAccessException}: prefer the row
+     * whose {@code operation} matches the active operation context, else the {@code operation == null}
+     * (connector/registry) row, else the lexicographically smallest {@code attributeUuid} (a stable tiebreak,
+     * not {@code updatedAt} which same-batch ingests share).
+     */
+    private AttributeDefinition selectByNameDeterministic(AttributeType type, UUID connectorUuid, String name, String operation) {
```
lubomirw: **[Low]** | Maintainability

All three callers pass `operation = null` (lines 218 / 226 / 267), so the comparator's "matches active operation context" tier and the `operation == null ? 1` branch can never fire — dead code implying selection behavior no caller exercises, and the Javadoc overstates it.

_Suggested:_ either thread a real operation context through the callers, or drop the `operation` parameter and simplify the comparator to (null-operation first, then lexicographic `attributeUuid`); trim the Javadoc to match.

src/main/java/com/otilm/core/service/callback/AttributeDefinitionResolver.java
vyskocilm: Fixed in 11190c54. The `fetchAndIngest` catch now rethrows an already-shaped `ValidationException` unchanged and wraps any other `RuntimeException` — the commit-time `DataIntegrityViolationException` — into a generic `ValidationException` after rollback, logging `e.getMessage()` at DEBUG, mirroring `ingest`/`handleResponseArms`. Test `commitFailureDoesNotLeakRawSqlToTheWire` asserts the SQL fragment never reaches the message.

src/main/java/com/otilm/core/service/callback/AttributeDefinitionResolver.java
vyskocilm: Fixed in 11190c54. Added `resolveForced`, which skips the stored-first short-circuit and always runs `fetchAndIngest` before re-reading; the not-found retry now calls it instead of `resolve`. Added `resolveForcedFetchesAndIngestsEvenWhenAStoredRowIsPresent` — stored row left in place, asserts `listDefinitions` + `updateDataAttributeDefinitions` both run on the forced path (the masking you flagged).

src/main/java/com/otilm/core/service/impl/CallbackServiceImpl.java
vyskocilm: Fixed in 11190c54 — and stronger than the `isNgCallback` null-check alone: `validateCallback` -> `validateContextAndMethod` also dereferences the callback (`callback.getCallbackContext()`), so a null falling through there would still 500. Added an explicit `attributeCallback == null` guard in `getCallbackObject` that throws a controlled `ValidationException` before either path, plus made `isNgCallback` null-safe as defense. Test `dataAttributeWithoutCallbackIsRejectedAsValidationNotNpe`.

src/main/java/com/otilm/core/service/callback/NgCallbackDispatcher.java
vyskocilm: Fixed in 11190c54. Enforced the XOR after the containment gate and before `handleResponseArms`: reject when `(content == null) == (attributes == null)`, covering both both-set and both-null. Tests `bothArmsNullIsRejectedAsValidation` and `bothArmsSetIsRejectedBeforeAnyRegistryWrite` (the latter asserts no registry write happens).

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
vyskocilm: Fixed in 11190c54. Dropped the always-null `operation` parameter and the dead comparator tier it guarded; the comparator is now (null-operation first, then lexicographic `attributeUuid`). Trimmed the Javadoc to match and updated all three callers.

vyskocilm:
## Regression testing report

Scope: verify the review-finding fixes (`11190c54`) and the operation-param removal introduce no regression in the affected callback codebase. Tested at `1b0c11ca`.

### Automated suites (all green)

| Suite | Tests | Result |
|-------|-------|--------|
| `AttributesV2CallbackDispatchTest` | 10 | ✅ |
| `NgCallbackDispatcherTest` | 10 | ✅ |
| `AttributeDefinitionResolverTest` | 4 | ✅ |
| `AttributeCallbackScopeResolverTest` (new) | 4 | ✅ |
| Architecture fences (`*ArchTest`) | 16 | ✅ |

Plus the pre-existing `CallbackServiceTest` (legacy `callbackContext` dispatch for GROUP/DATA, RESOURCE-to-body mapping) and `CoreCallbackServiceTest` continue to pass — these exercise the legacy branch of `getCallbackObject`, so they confirm the new `isNgCallback` check and the null-callback guard do not divert or break legacy callbacks (a legacy callback has `callbackContext != null`, so `isNgCallback` returns false and it falls through unchanged).

### Added regression coverage

- `nameOnlyResolutionPicksDeterministicallyAcrossDuplicateRows` — pins `selectByNameDeterministic` after the `operation` parameter was dropped: two DATA rows sharing `(type, connector, name)` must resolve name-only without `IncorrectResultSizeDataAccessException` and must pick the lexicographically smallest `attributeUuid` deterministically.
- New scope-resolver unit tests cover the per-kind walkers and the fail-closed `<KIND>:DETAIL` authorization arms (certificate-issuance, cryptographic-key, location, unmapped-resource).

A "both `dependsOn` and `callbackContext` routes legacy" test was attempted and then dropped: ingest (`AttributeEngine.validateCallbackDeclaration`) already rejects a both-set definition (`Attribute callback cannot declare both dependsOn and callbackContext`), so that state is unreachable through normal ingest — confirming the `callbackContext == null` conjunct in `isNgCallback` is purely defensive.

### Smoke test against a real Core

Ran this branch as a live Core (dependencies in Docker, Core via `spring-boot:run`) and exercised it over HTTP with client-certificate auth:

- **Boots clean.** Flyway applied all 26 migrations to schema `core`; the full Spring context started.
- **Bean wiring verified.** The context includes the changed-constructor `AttributeCallbackScopeResolver` (now also injecting `PermissionEvaluator` and the authority/raProfile/tokenProfile/entity repositories). Unit tests construct this bean manually, so a real context start is what proves the new constructor dependencies are satisfiable.
- Authenticated `GET /v1/auth/profile`, `GET /v1/connectors`, `GET /v1/settings/platform` all return 200.
- `POST /v2/connectors/{unknown-uuid}/callback` returns a controlled **404** (`Object of type 'Connector' ... is not found.`), not a 500 — the NG-capable entrypoint maps errors cleanly.

Full end-to-end NG dispatch (`POST /v2/attributes/callback` on the connector) was not exercised live because no shipped connector yet declares a `dependsOn` attribute (the FE form lands in fe-administrator#1764); that path is covered by the integration tests against a WireMock connector stub (`AttributesV2CallbackDispatchTest`, `NgCallbackDispatcherTest`).

### Conclusion

No regression observed in build, Flyway migration, bean wiring, the legacy connector-callback entrypoints, or error mapping. All affected suites and architecture fences are green.

vyskocilm:
### Live HTTP verification of the changed callback API

Followed up the in-process suite by driving the **changed endpoints over real HTTP** against this branch running as a live Core (deps in Docker, Core via `spring-boot:run`). A registered connector was repointed at a minimal mock that serves attribute definitions, the NG `POST /v2/attributes/callback`, a legacy callback, and health — so `getCallbackObject` and `NgCallbackDispatcher` execute end-to-end through the controller.

Endpoint: `POST /api/v1/connectors/{uuid}/authorityProvider/EJBCA/callback`

| Scenario (request) | Result | Covers |
|---|---|---|
| NG `dependsOn` attribute (`{"name":"ngAttr"}`) | **200**, content arm returned; mock received `POST /v2/attributes/callback` with envelope `{"attributeUuid":"…","attributeName":"ngAttr","contextAttributes":[]}` | NG routing + dispatch |
| attribute with no callback (`{"name":"noCbAttr"}`) | **422** `["Attribute 'noCbAttr' has no callback definition to invoke"]` | fix #3 — controlled error, not a 500 NPE |
| legacy `callbackContext` attribute (`{"name":"legacyAttr"}`) | **200** `{"property":"legacy-value"}` | regression — legacy path unchanged |
| NG, connector returns BOTH arms | **422** `["Connector callback response must set exactly one of content or attributes"]` | fix #4 — response-arm XOR |

The NG envelope captured at the connector confirms the dispatch path actually fired (not short-circuited), and the legacy attribute on the same connector still routes to the legacy callback — so the `isNgCallback` insertion discriminates correctly at runtime. No regression observed; the changed behaviours match the integration-suite assertions over real HTTP.

vyskocilm:
### Mock connector used for the live HTTP checks (reproducible)

The live verification above used a ~50-line stdlib-only Python HTTP server standing in for a connector, plus three JSON bodies in the **library's exact wire format** (emitted via `AttributeDefinitionUtils.serialize` / the DTO `ObjectMapper`, so Core deserializes them as-is).

**Setup**

```sh
# 1. deps up, Core running on :8280 (this branch), an admin cert enrolled
# 2. write the bodies below to /tmp/mock-bodies/
# 3. start the mock and repoint a registered connector at it:
python3 mockconn.py 9099                            # valid single-arm NG response
#   psql: UPDATE core.connector SET url='http://localhost:9099' WHERE name='<connector>';
# 4. drive the changed endpoint:
CERT="ssl-client-cert: $(openssl x509 -in admin.cert.pem -outform PEM | base64 -w0)"
U=<connector-uuid>
curl -s -X POST "http://localhost:8280/api/v1/connectors/$U/authorityProvider/EJBCA/callback" \
     -H "$CERT" -H 'Content-Type: application/json' -d '{"name":"ngAttr"}'      # -> 200 NG dispatch
# name=noCbAttr -> 422 (fix #3); name=legacyAttr -> 200 legacy; restart with the
# both-arms body to get the XOR 422 (fix #4):
python3 mockconn.py 9099 ng-response-bothargs.json
```

**`mockconn.py`**

```python
#!/usr/bin/env python3
"""Minimal CZERTAINLY connector mock for exercising the changed callback API."""
import json, sys
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer

BODIES = "/tmp/mock-bodies"
STATE = {"ng_body": "ng-response.json"}

def load(name):
    with open(f"{BODIES}/{name}") as f:
        return f.read()

class H(BaseHTTPRequestHandler):
    def _send(self, code, body, ctype="application/json"):
        b = body.encode() if isinstance(body, str) else body
        self.send_response(code)
        self.send_header("Content-Type", ctype)
        self.send_header("Content-Length", str(len(b)))
        self.end_headers()
        self.wfile.write(b)

    def log_message(self, fmt, *args):
        sys.stderr.write("MOCK %s - %s\n" % (self.command, self.path))

    def do_GET(self):
        p = self.path
        if p.endswith("/attributes"):
            self._send(200, load("attributes.json"))
        elif "health" in p or p in ("/v1", "/v2/info", "/v1/info"):
            self._send(200, '{"status":"ok"}')
        elif p.startswith("/legacy-callback"):
            self._send(200, '{"property":"legacy-value"}')
        else:
            self._send(200, '{"status":"ok"}')

    def do_POST(self):
        n = int(self.headers.get("Content-Length", 0))
        raw = self.rfile.read(n) if n else b""
        if self.path == "/v2/attributes/callback":
            sys.stderr.write("MOCK   /v2/attributes/callback body: %s\n" % raw.decode("utf-8", "replace"))
            self._send(200, load(STATE["ng_body"]))
        else:
            self._send(200, "{}")

if __name__ == "__main__":
    port = int(sys.argv[1]) if len(sys.argv) > 1 else 9099
    if len(sys.argv) > 2:
        STATE["ng_body"] = sys.argv[2]
    print(f"mock connector on :{port} ng_body={STATE['ng_body']}", flush=True)
    ThreadingHTTPServer(("0.0.0.0", port), H).serve_forever()
```

**`/tmp/mock-bodies/attributes.json`** (served on `GET /v1/{fg}/{kind}/attributes`)

```json
[{"uuid":"09d47d69-27b5-39b7-ab7e-d50ee6a461ca","name":"ngAttr","version":2,"type":"data","contentType":"string","properties":{"label":"ngAttr","visible":true,"required":false,"readOnly":false,"list":false,"multiSelect":false,"protectionLevel":"none","extensibleList":false},"attributeCallback":{"dependsOn":["dep"]}},
 {"uuid":"304520b1-7c96-3fef-8dc6-9bb71afcc3a4","name":"legacyAttr","version":2,"type":"data","contentType":"string","properties":{"label":"legacyAttr","visible":true,"required":false,"readOnly":false,"list":false,"multiSelect":false,"protectionLevel":"none","extensibleList":false},"attributeCallback":{"callbackContext":"/legacy-callback","callbackMethod":"GET"}},
 {"uuid":"d52f802a-de30-3362-be58-bdf309acbbb0","name":"noCbAttr","version":2,"type":"data","contentType":"string","properties":{"label":"noCbAttr","visible":true,"required":false,"readOnly":false,"list":false,"multiSelect":false,"protectionLevel":"none","extensibleList":false}}]
```

**`/tmp/mock-bodies/ng-response.json`** (valid single-arm — `POST /v2/attributes/callback`)

```json
{"content":[{"reference":"ng-resolved-value","data":"ng-resolved-value","contentType":"string"}]}
```

**`/tmp/mock-bodies/ng-response-bothargs.json`** (XOR violation — both arms set)

```json
{"content":[{"reference":"x","data":"x","contentType":"string"}],"attributes":[{"uuid":"1b7d5726-533a-3525-a876-0351e9b5e415","name":"child","version":2,"type":"data","contentType":"string","properties":{"label":"child","visible":true,"protectionLevel":"none"}}]}
```

> Bodies were generated from the interfaces DTOs themselves (not hand-written), so the polymorphic `type`/`version` discriminators match what Core expects. The connector is registered with `authType: none`; the mock answers health so the status poll keeps it `connected`.

lubomirw (approved):

src/main/java/com/otilm/core/service/impl/LocationServiceImpl.java
```diff
@@ -747,6 +750,22 @@ public NameAndUuidDto getResourceObjectExternal(SecuredUUID objectUuid) throws N
         return new NameAndUuidDto(String.valueOf(objectUuid), location.getName());
     }
 
+    @Override
+    @ExternalAuthorization(resource = Resource.LOCATION, action = ResourceAction.DETAIL, parentResource = Resource.ENTITY, parentAction = ResourceAction.DETAIL)
+    public ResourceObjectContentData getAuthorizedObjectAttributes(SecuredUUID objectUuid) throws NotFoundException {
```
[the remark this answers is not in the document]
lubomirw: Addressed on the current head: `getAuthorizedObjectAttributes(SecuredUUID)` is now `@ExternalAuthorization(resource = LOCATION, action = DETAIL)` — object-scoped only, no `parentResource`/`parentAction`, matching `getResourceObjectExternal` and the AUTHORITY/ENTITY/CREDENTIAL loaders. The empty-parent-UUID deny-all path no longer applies. Good catch when posted — resolving.

src/main/java/com/otilm/core/service/callback/AttributeCallbackScopeResolver.java
```diff
@@ -0,0 +1,188 @@
...
+        return scoped;
+    }
+
+    // --- walkers ------------------------------------------------------------------------------------------
+
+    private List<ScopeStep> walkRaProfile(UUID raProfileUuid) throws NotFoundException {
+        AuthorityInstanceReference authority = authorityInstanceReferenceRepository.findByUuid(raProfileUuid)
+                .orElseThrow(() -> notFound(AuthorityInstanceReference.class, raProfileUuid));
+        return List.of(authorityStep(authority));
+    }
```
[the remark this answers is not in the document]
lubomirw: Behaviour is correct: for the `RA_PROFILE` route the scope UUID is the authority (`resourceCallback` resolves `AuthorityInstanceReference.findByUuid(resourceUuid)`), so the walker intentionally loads the authority. Agreed the parameter name reads misleadingly — noting as an optional rename (e.g. `scopeObjectUuid`). Non-blocking; resolving.

src/main/java/com/otilm/core/service/impl/CallbackServiceImpl.java
```diff
@@ -200,11 +249,44 @@ private AttributeResource getAttributeResource(BaseAttribute attribute) {
...
+        List<ScopedAttributes> contextAttributes = scopeResource == null
+                ? List.of()
+                : scopeResolver.resolveScopeChain(scopeResource, scopeResourceUuid, expandedSecrets);
+        List<RequestAttribute> currentAttributes = callback.getAttributes();
+        if (currentAttributes != null && !currentAttributes.isEmpty()) {
+            attributeReferenceExpander.expandForCaller(currentAttributes, expandedSecrets);
+        }
+        NgCallbackDispatcher.NgDispatchContext context = new NgCallbackDispatcher.NgDispatchContext(
+                attribute, (ConnectorInterface) null, interfaceVersion, contextAttributes, currentAttributes);
+        AttributeCallbackResponseDto response = ngCallbackDispatcher.dispatchNgCallback(connector, context, callback, expandedSecrets);
```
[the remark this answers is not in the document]
lubomirw: `dispatchNg` does not hard-code null — `connectorInterface` is a parameter the resource-scoped routes stamp: `RA_PROFILE` → `interfaceCodeOf(authorityInstance)`, `CERTIFICATE`/issuance → `interfaceCodeOf(issuanceAuthority)`. It is null only by design where there is no interface context (the connector-only callback route and non-authority kinds such as `TOKEN_PROFILE`), covered by the `bothNullInterfaceContext…` test. Authority callback envelopes are stamped, so this does not apply — resolving.

Issue #1604, opened 2026-06-14
Add v3 integration test suite and update v2/ACME fixtures


lubomirw:
### Description

Add the cross-stack v3 integration tests under `service/v3/*`:

- sync issue (connector 200); async issue (202 → poll → COMPLETED → ISSUED)
- register-then-issue; cancel variants (204 / 404 / 422)
- race conditions (concurrent completion vs. poll); meta-persistence failure
- capability gating; v3 authority creation

Update existing v2 / ACME test fixtures for the connector-interface changes introduced earlier in the work.

**Depends on:** #1599, #1600, #1601, #1602, #1603
**Review focus:** coverage breadth of the async, race, and register-then-issue paths.

### Definition of Done

- [ ] `service/v3/*` integration tests covering sync issue, async poll→completed, register-then-issue, cancel 204/404/422, race, meta-failure, capability gating, authority creation
- [ ] Existing v2 / ACME fixtures updated for the connector-interface changes; full suite green
- [ ] New-code coverage ≥85% line+branch; SonarCloud gate clean
- [ ] PR merged to `main`

PR #1681, opened 2026-06-26
Add v3 certificate registration and route state changes through the state machine

### LINKED ISSUE #1603: Implement v3 certificate register + availableOperations operator flow

### LINKED ISSUE #1638: Add ConnectorCapabilityService + FeatureFlag gating for v3 operations
### Description

Add the **service-layer capability-gating tier** for v3 authority operations — defense-in-depth layer 2 (design §6): a `ConnectorCapabilityService` that gates **ENFORCED** `FeatureFlag`s with opt-in semantics and passes **INFORMATIONAL** flags through, resolved per the authority's bound connector interface.

The `FeatureFlag` infrastructure is merged in interfaces — `FeatureFlag` (`com.otilm.api.model.client.connector.v2`) carries `FeatureFlagBehavior` (`ENFORCED`/`INFORMATIONAL`), `getBehavior()`, per-flag `applicableInterfaces`, and the `CERTIFICATE_REGISTRATION` flag (`ENFORCED`, `AUTHORITY`). This follow-up builds the Core gating tier that consumes that metadata, and adds **one new flag** (`CERTIFICATE_STATUS_POLLING`) to interfaces for the async-polling case below.

Today only layer 1 (protocol-level `instanceof <Capability>`) and layer 3 (connector-side `OPERATION_NOT_SUPPORTED`) exist; this adds the missing middle tier. **Carved out of core#1603** — the register pass ships with protocol-level gating in the interim, so there is no correctness hole; this is the policy/opt-in layer.

Part of the authority-provider-v3 epic OmniTrustILM/ilm#110.

### Scope

**ConnectorCapabilityService (core):**
- `ConnectorCapabilityService.supports(AuthorityInstanceReference authority, FeatureFlag flag)` (+ connector / interface / version overloads) — opt-in for `ENFORCED` (supported only if the connector advertises the flag on its applicable interface), pass-through for `INFORMATIONAL`; resolved against the authority's connector-interface advertised `features`.
- Gate the v3 operations (register, etc.) on the relevant flag (e.g. `CERTIFICATE_REGISTRATION`) at the service layer, alongside the existing protocol-level `instanceof` checks.

**Async status-polling gating (folded in from the S4 review):**
- Add a `CERTIFICATE_STATUS_POLLING` flag to the interfaces `FeatureFlag` enum (`ENFORCED`, `AUTHORITY`) — a v3 connector advertises it when it actually supports being polled for async completion. *(Small interfaces change — the only non-core part of this issue.)*
- Gate the activation in `ClientOperationServiceImpl` so `scheduleStatusPoll(...)` only schedules a poll when the authority is async-pollable, composing the interface-version check with the flag:
  ```
  adapter instanceof AsyncOperationCapability                          // v3 — the status endpoint exists
    && capabilityService.supports(authority, CERTIFICATE_STATUS_POLLING)   // connector says "poll me"
  ```
- Why both conjuncts: `instanceof` only says the connector is v3 (the poll endpoint exists), not whether it wants to be polled. A connector can return 202 for **out-of-band / manual completion** (an external/offline authority) and 404 its status endpoint. Without the flag the platform would poll it and — because the listener treats a 404 (`RESOURCE_NOT_FOUND`) on `pollStatus` as a terminal poll error — would wrongly move the certificate to `FAILED`.
- Companion refinement: when polling isn't advertised, don't fail-on-404 — the certificate stays `PENDING_*` for manual / out-of-band completion.

**Motivating case:** `external-authority-provider` returns 202 on every operation and 404s its status endpoint (manual-completion model). It's **v2** today, so S4's protocol-level `instanceof` check already skips polling it — but S4 ships with **unconditional** scheduling (per the S4 review), so until this lands every v2-202 operation incurs a harmless-but-noisy cycle: a poll row → one RabbitMQ message → listener WARN ("adapter does not implement AsyncOperationCapability") → delete. This issue removes that churn and makes the gating correct for future **v3** manual-completion connectors (where `instanceof` alone is insufficient).

### Dependencies

- The interfaces `FeatureFlag` / `FeatureFlagBehavior` API is merged; this adds one enum value (`CERTIFICATE_STATUS_POLLING`). Relates to **core#1603** (register pass — protocol-level gating until this lands) and **core#1602 / PR #1627** (S4 async polling — ships scheduling unconditional; this gates it).

### Acceptance Criteria

- `ConnectorCapabilityService` with opt-in `ENFORCED` / pass-through `INFORMATIONAL` semantics, resolved per connector interface.
- v3 operations gated at the service layer (in addition to protocol-level `instanceof`).
- `CERTIFICATE_STATUS_POLLING` added to the interfaces `FeatureFlag` enum.
- `scheduleStatusPoll` schedules a poll only when the authority is v3 **and** advertises `CERTIFICATE_STATUS_POLLING`; a v2 / non-advertising authority schedules no poll (no row, no message, no WARN).
- A v3 connector that returns 202 without advertising polling is **not** polled and **not** failed-on-404 — it stays `PENDING_*` for manual completion.
- New-code coverage ≥85% line+branch; SonarCloud gate clean; PR merged.

### LINKED ISSUE #1677: Cancelling a pending certificate operation logs a SUCCESS history entry that contradicts the resulting state
### Summary

Cancelling a **pending certificate operation** (the operator "cancel pending operation" action) moves the certificate to a terminal state but records a **`SUCCESS`** entry in the certificate history under the **operation's own event kind**, so the history contradicts the certificate's final state.

Observed: an external (offline) authority issue → certificate `PENDING_ISSUE` → manual **cancel** → certificate state **`FAILED`**, but the certificate **history shows an `ISSUE` event with status `SUCCESS`** ("Pending issue cancelled"). A `FAILED` certificate should never show a successful `ISSUE` event.

The `PENDING_REVOKE` cancel has the same shape: the certificate returns to `ISSUED` and the history records a `REVOKE` event with `SUCCESS` — less alarming, but equally inconsistent (revocation did not actually complete).

### Steps to reproduce

1. Connect an authority whose operations are accepted asynchronously (e.g. the external-authority connector — returns `202`, finalized manually).
2. Issue a certificate → it lands in `PENDING_ISSUE` (polling, if any, is abandoned for a non-pollable connector).
3. Operator → **cancel pending operation**.
4. Certificate state = **`FAILED`**; certificate history's latest entry = **`ISSUE` / `SUCCESS`** ("Pending issue cancelled").

### Root cause

`ClientOperationServiceImpl.cancelPendingCertificateOperation` (~L1655):
- `determineCancelTarget` (~L1691) maps `PENDING_ISSUE` → target state `FAILED`, **event kind `CertificateEvent.ISSUE`** (and `PENDING_REVOKE` → `ISSUED`, event kind `REVOKE`).
- After `commitLocalCancel` applies the terminal state, the history is written (~L1670) as `target.eventKind()` + a hardcoded **`CertificateEventStatus.SUCCESS`** — the status reflects "the cancel action succeeded", but it is recorded against the *operation's* event kind, so it reads as "issuance succeeded" on a `FAILED` cert.

Underlying cause: **`CertificateEvent` has no `CANCEL` value** (only `ISSUE`/`RENEW`/`REKEY`/`REVOKE`), so cancels reuse the operation's event kind and the status then contradicts the outcome.

**Introduced in PR #1439 ("Asynchronous authority provider operations"), merged to `main` (2026-05-06) — independent of the in-flight async-polling PR #1627.**

### Expected

The certificate history must not contradict the certificate's final state. A cancelled issuance ending `FAILED` must not surface a `SUCCESS` `ISSUE` event; a cancelled revocation leaving the cert `ISSUED` must not surface a `SUCCESS` `REVOKE` event implying revocation happened.

### Candidate fixes (decide during analysis)

**(a) Add a dedicated `CANCEL` event — cleanest, cross-repo.** Add `CANCEL` to the `CertificateEvent` enum (`interfaces`) and record cancel history as `CANCEL / SUCCESS`. The cancel action's success then coexists cleanly with the resulting state (`FAILED` / `ISSUED`) — no contradiction. Requires an `interfaces` change plus the FE tolerating the new enum value (the codebase already tolerates unknown enum values).

**(b) Status reflects the operation outcome — core-only.** Record the cert history with the *operation's* outcome: `FAILED` for cancel-issue (cert → `FAILED`), and consistently for cancel-revoke (cert → `ISSUED`). Keep the cancel **action's** success in the audit **event log** (`Operation.CANCEL` / `SUCCESS`). Note: the failure paths (`recordCancelFailure` / `recordCancelSoftFailure`) already emit an `Operation.CANCEL` event-log entry, but the **success path currently omits it** — add it so the cancel action stays auditable either way.

### Acceptance criteria

- Cancelling a `PENDING_ISSUE` op: cert `FAILED`, and its history contains no entry implying issuance succeeded.
- Cancelling a `PENDING_REVOKE` op: cert `ISSUED`, and its history contains no entry implying revocation succeeded.
- The cancel action is auditable via the event log on both the success and failure paths.
- Tests cover the resulting state + history for both cancel paths.

### References

`core` · `service/v2/impl/ClientOperationServiceImpl.java` — `cancelPendingCertificateOperation` (~L1655), `determineCancelTarget` (~L1691), history write (~L1670). Part of the authority-provider-v3 epic OmniTrustILM/ilm#110.

### LINKED ISSUE #1682: Route ClientOperationServiceImpl state transitions through CertificateStateMachine
### Description

`CertificateStateMachine` (+ `CertificateStateTransition`) was introduced as new files (S1, epic OmniTrustILM/ilm#110), but the planned routing of the existing state-write call sites in `service/v2/impl/ClientOperationServiceImpl.java` through it was deferred. That file still uses direct `cert.setState(...)` + hand-written `addEventHistory(...)` at ~9 sites across issue / approval / revoke / cancel / restore / manual-confirm / commit-cancel. Only the new v3 register-lifecycle transitions (PR #1681) go through the SM today.

Route the remaining `setState` sites through `stateMachine.transition(...)` so the audit-history status comes from the transition table instead of being hand-written. No behaviour change is intended beyond correct/consistent audit status; the surrounding side-effects (location cleanup, predecessor deletion, dual-cert history) must be preserved.

This also fixes #1677 — "cancel logs a SUCCESS history entry on a FAILED cert" is a direct symptom of the un-migrated `cancelPendingCertificateOperation` calling `addEventHistory(..., SUCCESS, ...)` regardless of the resulting state.

### Acceptance criteria

- All `cert.setState(...)` call sites in `ClientOperationServiceImpl` (and the helpers it delegates to) go through `CertificateStateMachine.transition(...)`.
- Audit-history entries derive their status from the transition row (no hand-written SUCCESS on a failed/restore transition).
- #1677 is resolved (cancelling a pending issue records a FAILED issue event, not SUCCESS).
- Existing v2 issue / renew / rekey / revoke / cancel / manual flows are behaviour-preserving (full `mvn verify` green).

### Notes

- Do **after** PR #1681 (register slice) merges, to avoid conflicts in `ClientOperationServiceImpl`.
- Part of epic OmniTrustILM/ilm#110. This is the original "M1" milestone from the planning `plan.md` (see its "Reconciliation — as shipped" section); the state machine itself already exists, only the call-site routing remains.

lubomirw:
## Summary

Implements the operator-facing v3 authority **certificate registration** lifecycle, part of the authority-provider-v3 epic (#110). It also folds in the **M1 state-machine migration** (#1682) — originally a stacked follow-up, delivered together here — routing the remaining `ClientOperationServiceImpl` state changes through `CertificateStateMachine`, and the **capability + FeatureFlag gating slice** (#1638), which edits the very methods this PR introduces.

## What's included

- **`registerCertificate`** — capability-gated (rejects non-v3 authorities, leaving no placeholder behind), creates a no-CSR identity placeholder, transitions `REQUESTED → PENDING_REGISTRATION` via `CertificateStateMachine`, calls the connector with no transaction held, then `→ REGISTERED` (sync) or schedules a REGISTER status poll (async 202). Connector metadata is persisted; a persistence failure is tolerated per the state-divergence rule, and a `ConnectorAcceptedButLocalFailureException` leaves the cert `PENDING_REGISTRATION` rather than rolling it back to FAILED.
- **`listAvailableOperations`** — advertises per-operation support flags (issue/renew/revoke always; register + async + cancel for v3) from adapter capabilities **and** the advertised FeatureFlags (see capability gating below).
- **Capability + FeatureFlag gating (#1638, folded in)** — new `ConnectorCapabilityService`, a service-layer (defense-in-depth layer 2) gate between the protocol-level `instanceof <Capability>` check and the connector-side `OPERATION_NOT_SUPPORTED`: ENFORCED flags are opt-in (supported only when the authority's connector interface advertises them), INFORMATIONAL flags pass through. `registerCertificate` and `listAvailableOperations` gate on `CERTIFICATE_REGISTRATION`; `scheduleStatusPoll` gates on `CERTIFICATE_STATUS_POLLING`, so a v3 connector that 202s without advertising polling is left `PENDING` for manual/out-of-band completion rather than polled (and logs accordingly). The gate is evaluated best-effort, so a connector-accepted operation is never rolled back. Depends on the `CERTIFICATE_STATUS_POLLING` flag added in interfaces#750.
- **REGISTERED → issue completion** — `issueRequestedCertificate` renamed to `issueExistingCertificate`, now state-aware / body-optional: `REQUESTED` uses its existing (protocol-attached) CSR with no body; `REGISTERED` requires the operator's CSR, attached via the new `CertificateService.addCertificateRequestToExisting` (the registration identity on the row is preserved). The `issueCertificateAction` state guard accepts `REGISTERED` so the existing sync/async issue tail completes it.
- **Manual-issue hook** — `manuallyIssueCertificate` makes a post-issue best-effort `adapter.cancel(ISSUE)` for v3 async certs so an out-of-band manual completion does not leave an orphaned upstream operation.

New helpers: `CertificateUtil.applyRegistrationSubject`, `CertificateService.createRegistrationPlaceholder`, `RaProfileRepository.findWithAuthorityByUuid`.

## State transitions

All of `ClientOperationServiceImpl`'s operator state changes now go through `CertificateStateMachine`, so audit status comes from the transition table. The register-lifecycle transitions plus — via the folded-in M1 migration — the issue / revoke / approval / cancel transitions all route through the SM (the one no-op revoke connector-failure restore stays a direct `setState` by design). Two table rows the original set omitted were added: `PENDING_APPROVAL → REVOKED` and `REGISTERED → FAILED`. The cancel audit now records `FAILED` from the row (closes #1677).

## Tests

`ClientOperationServiceRegisterTest` (26 tests): register sync / async / unsupported-v2 / connector-failure / accepted-but-local-failure, SAN-only registration, `listAvailableOperations` v2/v3, REGISTERED→issue (validation + CSR attach), plus the #1638 gating cases (registration rejected when the capability is not advertised, async registration skips the poll when status polling is not advertised, `listAvailableOperations` gated on advertised flags). New `ConnectorCapabilityServiceTest` (3 tests) covers the ENFORCED-opt-in / INFORMATIONAL-pass-through resolution. The folded-in M1 migration adds `CertificateStateMachine` unit tests (the new rows, `transitionAuditedExternally`, `canTransition`, `auditDetail`) and `ClientOperationServiceV2Test` assertions (cancel → FAILED, redelivery idempotency). Affected suites green locally (ConnectorCapabilityServiceTest 3, ClientOperationServiceRegisterTest 26, ClientOperationServiceV2Test 70); CI re-runs the full suite + SonarCloud on the combined PR.

## Follow-ups (not in this PR)

- **#1683**: complete `CertificateStateMachine` coverage for the remaining cert-state writers (`prepareIssuedCertificate` → ISSUED incl. async-issue completion; `revokeCertificate(serialNumber)` → REVOKED).
- **#1684**: unify approval-driven event-history ownership (the `transitionAuditedExternally` seam + the handlers' cross-resource TODO).
- **#130**: registration-completion authorization (challenge / verification / events) layers on `issueExistingCertificate` + `addCertificateRequestToExisting`; flat SAN/DN identity persistence on the placeholder.
- **Docs**: `interface-documentation` v2 OpenAPI is stale (still `issueRequestedCertificate`, no `register`/`availableOperations`); docs.otilm.com operator guide.
- **S6 (#1604)**: v3 integration tests — full register→issue round-trip and the manual-issue cancel hook.

Closes #1603.
Closes #1677.
Closes #1682.
Closes #1638.


src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -315,6 +334,107 @@ public ClientCertificateDataResponseDto issueCertificate(final SecuredParentUUID
...
+        }
+
+        persistRegistrationMeta(certificate, result.meta());
+        if (result.isAsync()) {
+            scheduleStatusPoll(certificate, CertificateOperation.REGISTER);
+            logger.info("Certificate {} registration accepted by authority; awaiting asynchronous completion", certificate.getUuid());
+        } else {
+            stateMachine.transition(certificate, CertificateState.REGISTERED);
+            logger.info("Certificate {} registered by authority", certificate.getUuid());
+        }
```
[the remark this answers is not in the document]
lubomirw: Fixed in 8d6e4ddc8 — the REGISTER status poll is now scheduled only when the adapter implements AsyncOperationCapability; an async-accepted result from a non-polling adapter leaves the certificate in PENDING_REGISTRATION with no abandoned poll. Added regression test asyncRegistrationWithoutAsyncCapabilityDoesNotSchedulePoll.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -536,11 +656,37 @@ private void handleFailedOrRejectedEvent(Certificate certificate, UUID oldCertif
...
+        // enqueued; otherwise the async consumer could read the placeholder before the CSR is visible and fail it.
+        RaProfile raProfile = raProfileRepository.findByUuid(raProfileUuid.getValue())
+                .orElseThrow(() -> new NotFoundException(RaProfile.class, raProfileUuid));
+        if (Boolean.FALSE.equals(raProfile.getEnabled())) {
+            throw new ValidationException("Cannot issue certificate with disabled RA profile. Ra Profile: %s".formatted(raProfile.getName()));
+        }
         Certificate certificate = certificateService.getCertificateEntity(SecuredUUID.fromString(certificateUuid));
-        if (certificate.getState() != CertificateState.REQUESTED) {
-            throw new ValidationException(ValidationError.create(String.format("Cannot issue requested certificate with status %s. Certificate: %s", certificate.getState().getLabel(), certificate)));
+        CertificateState state = certificate.getState();
```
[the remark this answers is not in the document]
lubomirw: Fixed in 8d6e4ddc8 — issueExistingCertificate now rejects a certificate that belongs to a different RA profile than the request (eager getRaProfileUuid check). Added regression test issueExistingCertificateRejectsCertFromDifferentRaProfile.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1094,6 +1096,51 @@ public Certificate createCertificateEntity(X509Certificate certificate) {
...
+    public void addCertificateRequestToExisting(UUID certificateUuid, ClientCertificateSignRequestDto signRequest)
+            throws CertificateRequestException, NoSuchAlgorithmException, NotFoundException {
+        Certificate certificate = getCertificateEntity(SecuredUUID.fromUUID(certificateUuid));
+
+        byte[] decodedCsr;
+        try {
+            decodedCsr = Base64.getDecoder().decode(signRequest.getRequest());
+        } catch (IllegalArgumentException e) {
+            throw new CertificateRequestException("Certificate signing request is not valid Base64");
+        }
```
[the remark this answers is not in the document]
lubomirw: Fixed in 8d6e4ddc8 — addCertificateRequestToExisting now rejects a null or blank CSR with a controlled CertificateRequestException instead of throwing an NPE.

src/main/java/com/otilm/core/service/v2/ClientOperationService.java
```diff
@@ -127,4 +128,15 @@ CertificateDetailDto cancelPendingCertificateOperation(
...
     ) throws NotFoundException;
+
+    ClientCertificateDataResponseDto registerCertificate(
+            SecuredParentUUID authorityUuid,
+            SecuredUUID raProfileUuid,
+            ClientCertificateRegistrationDto request
+    ) throws NotFoundException, ConnectorException;
+
+    AvailableOperationsDto listAvailableOperations(
+            SecuredParentUUID authorityUuid,
```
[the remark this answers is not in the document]
lubomirw: Not changed — false positive. The SecuredParentUUID authorityUuid is consumed reflectively by the @ExternalAuthorization aspect (parentResource = AUTHORITY) for parent-resource authorization, even though the method body does not reference it directly. This matches the other operator methods such as listIssueCertificateAttributes and cancelPendingCertificateOperation; removing it would drop the parent-resource authorization check.

src/main/java/com/otilm/core/service/v2/ClientOperationService.java
```diff
@@ -127,4 +128,15 @@ CertificateDetailDto cancelPendingCertificateOperation(
             String certificateUuid,
             CancelPendingCertificateRequestDto request
     ) throws NotFoundException;
+
+    ClientCertificateDataResponseDto registerCertificate(
+            SecuredParentUUID authorityUuid,
```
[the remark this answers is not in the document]
lubomirw: Not changed — false positive. The SecuredParentUUID authorityUuid is consumed reflectively by the @ExternalAuthorization aspect (parentResource = AUTHORITY) for parent-resource authorization, even though the method body does not reference it directly. This matches the other operator methods such as listIssueCertificateAttributes and cancelPendingCertificateOperation; removing it would drop the parent-resource authorization check.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -315,6 +334,113 @@ public ClientCertificateDataResponseDto issueCertificate(final SecuredParentUUID
...
+    public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID authorityUuid, SecuredUUID raProfileUuid,
+                                                                ClientCertificateRegistrationDto request) throws NotFoundException, ConnectorException {
+        // Connector call below holds no transaction (NOT_SUPPORTED), so load the authority graph eagerly —
+        // every association the adapter dereferences must be initialized before the session closes.
+        RaProfile raProfile = raProfileRepository.findWithAuthorityByUuid(raProfileUuid.getValue())
+                .orElseThrow(() -> new NotFoundException(RaProfile.class, raProfileUuid));
+        if (Boolean.FALSE.equals(raProfile.getEnabled())) {
+            throw new ValidationException("Cannot register certificate with disabled RA profile. Ra Profile: %s".formatted(raProfile.getName()));
+        }
+
```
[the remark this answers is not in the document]
lubomirw: Fixed in e05bca6b1 — registerCertificate now calls assertRaProfileUnderAuthority, which rejects a request whose RA profile does not belong to the authority named in the path (eager authorityInstanceReferenceUuid comparison). Added regression test registerRejectsRaProfileNotUnderRequestedAuthority.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -315,6 +334,113 @@ public ClientCertificateDataResponseDto issueCertificate(final SecuredParentUUID
...
+    }
+
+    @Override
+    @Transactional(readOnly = true)
+    @ExternalAuthorization(resource = Resource.RA_PROFILE, action = ResourceAction.DETAIL, parentResource = Resource.AUTHORITY, parentAction = ResourceAction.DETAIL)
+    public AvailableOperationsDto listAvailableOperations(SecuredParentUUID authorityUuid, SecuredUUID raProfileUuid) throws NotFoundException {
+        RaProfile raProfile = raProfileRepository.findWithAuthorityByUuid(raProfileUuid.getValue())
+                .orElseThrow(() -> new NotFoundException(RaProfile.class, raProfileUuid));
+        AuthorityProviderAdapter adapter = adapterFactory.forAuthority(raProfile.getAuthorityInstanceReference());
+
```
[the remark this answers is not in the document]
lubomirw: Fixed in e05bca6b1 — listAvailableOperations now applies the same assertRaProfileUnderAuthority check against the path authority.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -536,11 +662,40 @@ private void handleFailedOrRejectedEvent(Certificate certificate, UUID oldCertif
...
-    public ClientCertificateDataResponseDto issueRequestedCertificate(final SecuredParentUUID authorityUuid, final SecuredUUID raProfileUuid, final String certificateUuid) throws ConnectorException, NotFoundException {
+    public ClientCertificateDataResponseDto issueExistingCertificate(final SecuredParentUUID authorityUuid, final SecuredUUID raProfileUuid, final String certificateUuid, final ClientCertificateSignRequestDto request) throws NotFoundException {
+        // NOT_SUPPORTED so the CSR attach below commits in its own transaction before the ISSUE action is
+        // enqueued; otherwise the async consumer could read the placeholder before the CSR is visible and fail it.
+        RaProfile raProfile = raProfileRepository.findByUuid(raProfileUuid.getValue())
+                .orElseThrow(() -> new NotFoundException(RaProfile.class, raProfileUuid));
+        if (Boolean.FALSE.equals(raProfile.getEnabled())) {
+            throw new ValidationException("Cannot issue certificate with disabled RA profile. Ra Profile: %s".formatted(raProfile.getName()));
+        }
```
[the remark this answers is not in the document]
lubomirw: Fixed in e05bca6b1 — issueExistingCertificate now applies assertRaProfileUnderAuthority in addition to the existing RA-profile-ownership check on the certificate.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -332,7 +458,7 @@ public void issueCertificateAction(final UUID certificateUuid, boolean isApprove
         final Certificate certificate = certificateRepository.findWithAssociationsByUuid(certificateUuid).orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
         if (certificate.isArchived())
             throw new ValidationException(ValidationError.create(String.format("Cannot issue requested certificate that has been archived. Certificate: %s", certificate.toStringShort())));
-        if (certificate.getState() != CertificateState.REQUESTED && certificate.getState() != CertificateState.PENDING_APPROVAL) {
+        if (certificate.getState() != CertificateState.REQUESTED && certificate.getState() != CertificateState.PENDING_APPROVAL && certificate.getState() != CertificateState.REGISTERED) {
             throw new ValidationException(ValidationError.create(String.format("Cannot issue requested certificate with state %s. Certificate: %s", certificate.getState().getLabel(), certificate)));
         }
```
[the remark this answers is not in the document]
lubomirw: Fixed in e05bca6b1 — the issueCertificateAction state-guard message no longer says requested; it now reads "Cannot issue certificate with state X".

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -315,6 +334,107 @@ public ClientCertificateDataResponseDto issueCertificate(final SecuredParentUUID
...
+        // v3-only, and asynchronous completion + cancellation come together with AsyncOperationCapability.
+        // Finer per-connector advertisement (FeatureFlag-driven) is layered on in the capability-service slice.
+        boolean register = adapter instanceof RegisterCapability;
+        boolean async = adapter instanceof AsyncOperationCapability;
+        List<OperationSupport> operations = List.of(
+                new OperationSupport(CertificateOperationKind.ISSUE, true, async, async),
+                new OperationSupport(CertificateOperationKind.RENEW, true, async, async),
+                new OperationSupport(CertificateOperationKind.REVOKE, true, async, async),
+                new OperationSupport(CertificateOperationKind.REGISTER, register, async, async)
+        );
```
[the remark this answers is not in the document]
lubomirw: Fixed in 8d6e4ddc8 — for REGISTER, asyncSupported and cancelSupported are now gated on register support as well (register AND async). An async-capable authority that does not implement RegisterCapability therefore no longer advertises REGISTER as async or cancellable. Covered by the listAvailableOperations v2 and v3 tests.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1494,11 +1649,34 @@ public CertificateDetailDto manuallyIssueCertificate(
...
+            if (certificate == null) {
+                return;
+            }
+            AuthorityProviderAdapter adapter = adapterFactory.forAuthority(certificate.getRaProfile().getAuthorityInstanceReference());
+            if (adapter instanceof AsyncOperationCapability async) {
+                async.cancel(certificate, CertificateOperation.ISSUE);
+            }
+        } catch (Exception e) {
+            logger.warn("Best-effort cancel of in-flight async issue after manual issue failed (cert {}): {}", certificateUuid, e.getMessage());
+        }
```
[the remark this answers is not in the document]
lubomirw: Fixed in e05bca6b1 — the WARN now passes the exception as the final logger argument, so the stack trace is logged. The failure is still swallowed (best-effort, the certificate is already issued locally), but it is now diagnosable.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -513,34 +662,75 @@ private void handleFailedOrRejectedEvent(Certificate certificate, UUID oldCertif
             }
```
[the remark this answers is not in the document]
lubomirw: Fixed in 5a948c5b4 — the location-removal catch now logs the actual caught exception (`ex`) instead of the unrelated outer `message` variable.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1094,6 +1096,59 @@ public Certificate createCertificateEntity(X509Certificate certificate) {
...
+        CertificateRequest request = CertificateRequestUtils.createCertificateRequest(decodedCsr, signRequest.getFormat());
+
+        // Attach the operator-supplied CSR to the placeholder; the registration identity already on the row
+        // (subject DN / SAN) is intentionally left untouched here — the issued certificate's identity is
+        // written from the CA response at issuance.
+        CertificateRequestEntity certificateRequestEntity = certificate.prepareCertificateRequest(signRequest.getFormat());
```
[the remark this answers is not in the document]
lubomirw: Fixed in 5a948c5b4 — `addCertificateRequestToExisting` now rejects a null CSR format with a clear `CertificateRequestException` before parsing (rather than NPE-ing in the `createCertificateRequest` `switch(format)`); added a regression test.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -315,11 +334,130 @@ public ClientCertificateDataResponseDto issueCertificate(final SecuredParentUUID
...
 
+    @Override
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    @ExternalAuthorization(resource = Resource.RA_PROFILE, action = ResourceAction.DETAIL, parentResource = Resource.AUTHORITY, parentAction = ResourceAction.DETAIL)
+    public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID authorityUuid, SecuredUUID raProfileUuid,
+                                                                ClientCertificateRegistrationDto request) throws NotFoundException, ConnectorException {
+        // Connector call below holds no transaction (NOT_SUPPORTED), so load the authority graph eagerly —
+        // every association the adapter dereferences must be initialized before the session closes.
+        RaProfile raProfile = raProfileRepository.findWithAuthorityByUuid(raProfileUuid.getValue())
+                .orElseThrow(() -> new NotFoundException(RaProfile.class, raProfileUuid));
```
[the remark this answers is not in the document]
lubomirw: Fixed in 1288b1fc8 — `registerCertificate` now rejects a null registration request with a `ValidationException`, so a missing body returns a controlled 4xx error instead of an NPE.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1094,6 +1096,62 @@ public Certificate createCertificateEntity(X509Certificate certificate) {
...
 
+    @Override
+    @Transactional
+    public Certificate createRegistrationPlaceholder(RaProfile raProfile, ClientCertificateRegistrationDto request) {
+        // Identity-only placeholder: no key/CSR/content yet. The authoritative subject, SAN and key
+        // material are recorded when the follow-up CSR issuance completes against this record, so only
+        // the subject DN identity from the registration request is captured here.
+        Certificate certificate = new Certificate();
+        CertificateUtil.applyRegistrationSubject(certificate, request.getSubjectDn());
+        certificate.setState(CertificateState.REQUESTED);
```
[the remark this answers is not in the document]
lubomirw: Addressed in 1288b1fc8 — the null-request guard is enforced at `registerCertificate`, the only caller, so `createRegistrationPlaceholder` is not reachable with a null request through the operator flow.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -332,8 +470,8 @@ public void issueCertificateAction(final UUID certificateUuid, boolean isApprove
         final Certificate certificate = certificateRepository.findWithAssociationsByUuid(certificateUuid).orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
         if (certificate.isArchived())
             throw new ValidationException(ValidationError.create(String.format("Cannot issue requested certificate that has been archived. Certificate: %s", certificate.toStringShort())));
-        if (certificate.getState() != CertificateState.REQUESTED && certificate.getState() != CertificateState.PENDING_APPROVAL) {
-            throw new ValidationException(ValidationError.create(String.format("Cannot issue requested certificate with state %s. Certificate: %s", certificate.getState().getLabel(), certificate)));
+        if (certificate.getState() != CertificateState.REQUESTED && certificate.getState() != CertificateState.PENDING_APPROVAL && certificate.getState() != CertificateState.REGISTERED) {
+            throw new ValidationException(ValidationError.create(String.format("Cannot issue certificate with state %s. Certificate: %s", certificate.getState().getLabel(), certificate)));
```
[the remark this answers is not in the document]
lubomirw: Fixed in 1288b1fc8 — the issueCertificateAction validation messages (archived / no RA profile / no CSR) no longer say "requested certificate", since the method now also handles REGISTERED and PENDING_APPROVAL.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1078,15 +1290,66 @@ private void transitionToPendingRevoke(Certificate certificate, ClientCertificat
...
+     * <p>Best-effort: the connector has already accepted the operation, so neither the gate evaluation nor the
+     * scheduling may roll it back (state-divergence rule). Any failure degrades to "no poll".
+     *
+     * @return {@code true} iff a poll was scheduled.
+     */
+    private boolean scheduleStatusPoll(Certificate certificate, CertificateOperation operation, BooleanSupplier pollable) {
         try {
+            if (!pollable.getAsBoolean()) {
+                logger.debug("Authority for certificate {} does not advertise CERTIFICATE_STATUS_POLLING; "
+                        + "no poll scheduled (left PENDING for manual/out-of-band completion)", certificate.getUuid());
```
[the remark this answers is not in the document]
lubomirw: Fixed in d0e004578: the `scheduleStatusPoll` debug log now reads "is not pollable (not v3, or does not advertise CERTIFICATE_STATUS_POLLING)", covering both reasons `pollable` is false rather than implying only the unadvertised-flag case.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -505,42 +677,92 @@ private boolean isRequestNotCompliant(UUID certificateUuid, UUID certificateRequ
...
+            throw new ValidationException("Cannot issue certificate with disabled RA profile. Ra Profile: %s".formatted(raProfile.getName()));
+        }
+        assertRaProfileUnderAuthority(raProfile, authorityUuid);
         Certificate certificate = certificateService.getCertificateEntity(SecuredUUID.fromString(certificateUuid));
-        if (certificate.getState() != CertificateState.REQUESTED) {
-            throw new ValidationException(ValidationError.create(String.format("Cannot issue requested certificate with status %s. Certificate: %s", certificate.getState().getLabel(), certificate)));
+        if (!raProfileUuid.getValue().equals(certificate.getRaProfileUuid())) {
+            throw new ValidationException("Cannot issue a certificate that belongs to a different RA profile. Certificate: %s".formatted(certificate.toStringShort()));
+        }
+        CertificateState state = certificate.getState();
```
ivosh, line 748: **~ Minor** | Correctness

ClientOperationServiceImpl.java:732-775 (issueExistingCertificate) is a new operator-driven read-and-update path: it reads the certificate via the unlocked certificateService.getCertificateEntity (line 741), asserts state==REGISTERED, then attaches an operator CSR (addCertificateRequestToExisting, line 762) and enqueues an ISSUE action (line 775). The method runs @Transactional(NOT_SUPPORTED), so each read is a plain, lock-free read. Two concurrent issueExistingCertificate calls on the same REGISTERED placeholder can both read REGISTERED, both pass the guard, both attach a CSR (the second overwriting the first's certificateRequest in CertificateServiceImpl.addCertificateRequestToExisting, which re-checks state==REGISTERED but the row is still REGISTERED — the state is not advanced to PENDING_ISSUE until the async consumer runs), and both enqueue an ISSUE ActionMessage — yielding a double issuance attempt for one certificate. This is exactly the CLAUDE.md "Race conditions on multi-actor endpoints" hazard: "concurrent transactions can both pass a state assertion ... and then both write." The sibling operator endpoints handle this correctly — manuallyConfirmRevoke (line 1875) and cancelPendingCertificateOperation/commitLocalCancel re-read under certificateRepository.findAndLockWithAssociationsByUuid (CertificateRepository.java:76-79, @Lock(PESSIMISTIC_WRITE), whose Javadoc explicitly names these operator pending-state endpoints), then assert state inside the locked tx. The register path even calls out its own no-race reasoning (line 376-378), but issueExistingCertificate is silent on why it is exempt. If concurrent issue-existing on the same placeholder is considered impossible/acceptable, add a one-line comment stating why; otherwise route the REGISTERED read-attach-enqueue through a short explicit transaction using findAndLockWithAssociationsByUuid with the state assertion held under the lock, matching the sibling pattern. Note the connector HTTP call is in the async consumer, not here, so the lock would not span an external call.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -315,11 +344,152 @@ public ClientCertificateDataResponseDto issueCertificate(final SecuredParentUUID
...
+        // Re-load with the full adapter graph for the transaction-less connector call, as the poll listener does.
+        // No pessimistic lock here (unlike the cancel/poll paths): the placeholder was just created and its UUID
+        // is not yet known to any concurrent actor, so the read-modify-write below cannot race.
+        Certificate certificate = certificateRepository.findForPollingByUuid(placeholder.getUuid())
+                .orElseThrow(() -> new NotFoundException(Certificate.class, placeholder.getUuid()));
+        stateMachine.transition(certificate, CertificateState.PENDING_REGISTRATION);
+
+        AdapterOperationResult result;
+        try {
+            result = registerCapability.register(certificate, request);
```
ivosh, line 385: **~ Minor** | Correctness

ClientOperationServiceImpl.java:383-399 — registerCertificate() runs @Transactional(NOT_SUPPORTED) and commits PENDING_REGISTRATION in its own REQUIRED tx (stateMachine.transition, line 381) before calling registerCapability.register(...). The try at line 384 catches only ConnectorAcceptedButLocalFailureException (leave PENDING, intended) and ConnectorException (transition to FAILED). An unchecked RuntimeException from register() is NOT caught, so the placeholder is orphaned in PENDING_REGISTRATION with no FAILED transition and no poll row (scheduleStatusPoll at line 405 is past the try).

This is reachable, not hypothetical. In the only implementation, AuthorityProviderV3Adapter.register (lines 188-206), three paths throw unchecked and non-ConnectorException: (1) the attribute-engine lookups authorityAttributesFor / raProfileAttributesFor (lines 200-201, bodies at 302-307) run BEFORE the acceptance guard and propagate any RuntimeException (e.g. a JPA PersistenceException or NPE) raw; (2) executeWithAcceptanceGuard's catch (RuntimeException e) rethrows the exception raw when connectorAccepted==false (line 290) — e.g. an HttpStatusCodeException/RestClientException from the HTTP client; (3) mapIssueRenewRegisterResponse throws IllegalStateException for a non-200/202 status (line 335), also pre-acceptance.

Remediation (refined after tracing the adapter): the guard converts every POST-acceptance runtime failure into ConnectorAcceptedButLocalFailureException (lines 286-288), so any RAW RuntimeException that reaches registerCertificate from this implementation is provably PRE-acceptance — no upstream work is in flight, so transitioning to FAILED is safe (a blanket catch is NOT a divergence risk here, contrary to a naive reading). Add a catch (RuntimeException e) mirroring the ConnectorException branch (transition to FAILED, then rethrow). Because registerCertificate is coded against the RegisterCapability interface rather than this concrete adapter, either note the pre-acceptance assumption in a one-line comment, or — better — hoist the 'raw RuntimeException => pre-acceptance (post-acceptance failures must be ConnectorAcceptedButLocalFailureException)' guarantee into RegisterCapability.register's Javadoc so every future adapter must uphold it. Non-blocking but worth resolving before merge.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1094,6 +1096,73 @@ public Certificate createCertificateEntity(X509Certificate certificate) {
...
+        // cannot overwrite the request of an ISSUED / REQUESTED / pending certificate.
+        if (certificate.getState() != CertificateState.REGISTERED) {
+            throw new ValidationException("A certificate signing request can only be attached to a REGISTERED certificate. Certificate: %s".formatted(certificate.toStringShort()));
+        }
+
+        byte[] decodedCsr;
+        try {
+            decodedCsr = Base64.getDecoder().decode(signRequest.getRequest());
+        } catch (IllegalArgumentException e) {
+            throw new CertificateRequestException("Certificate signing request is not valid Base64");
```
ivosh: **~ Minor** | Error handling

CertificateServiceImpl.java:1136-1137 catches `IllegalArgumentException e` from `Base64.getDecoder().decode(...)` and rethrows `new CertificateRequestException("Certificate signing request is not valid Base64")` without passing `e` as the cause. CertificateRequestException supports a `(message, cause)` constructor and the established neighbour pattern always chains it: `Pkcs10CertificateRequest.java:33` (`throw new CertificateRequestException("Cannot process PKCS#10 request", e)`) and `CrmfCertificateRequest.java:33`. Dropping the cause discards the decoder's diagnostic detail and breaks the exception chain (Effective Java Item 73 — throw exceptions appropriate to the abstraction, but preserve the lower-level cause). Fix: `throw new CertificateRequestException("Certificate signing request is not valid Base64", e);` to mirror the sibling request classes. Non-blocking idiom correction.

src/test/java/com/otilm/core/service/ClientOperationServiceRegisterTest.java
```diff
@@ -0,0 +1,561 @@
...
+        String certUuid = registerSyncRegistered();
+        ClientCertificateSignRequestDto signRequest = new ClientCertificateSignRequestDto();
+        signRequest.setRequest("!!!not-valid-base64!!!");
+        Assertions.assertThrows(ValidationException.class, () -> clientOperationService.issueExistingCertificate(
+                authorityParent,
+                securedRaProfile, certUuid, signRequest));
+    }
+
+    @Test
+    void registerPersistsConnectorMetadata() throws Exception {
```
ivosh, line 521: **~ Minor** | Testing

ClientOperationServiceRegisterTest.java:500 — the test is named registerPersistsConnectorMetadata and stubs register() to return one Mockito.mock(MetadataAttribute.class), but its only assertion is cert.getState() == REGISTERED. attributeEngine is NOT a @MockitoBean here, so persistRegistrationMeta (ClientOperationServiceImpl.java:~430) calls the real engine with a bare mock MetadataAttribute; per the method's own contract that failure is caught, logged, and the flow proceeds. The test therefore passes whether or not metadata is persisted correctly — JaCoCo would show the successful updateMetadataAttributes branch as uncovered. This is the CLAUDE.md 'looks green but covers nothing' pattern: a type/state-only assertion that does not check the origin of the behavior. If verifying persistence is intended, mock attributeEngine and verify(attributeEngine).updateMetadataAttributes(...) (or assert the persisted metadata via a query); otherwise rename the test to reflect that it only asserts registration completes despite a metadata payload. Not a merge blocker.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -981,8 +1203,13 @@ public void revokeCertificateAction(final UUID certificateUuid, ClientCertificat
...
 
-            certificate.setState(CertificateState.REVOKED);
-            certificateRepository.save(certificate);
+            // Connector revoked synchronously (200). Drive state to REVOKED before the fallible
+            // attribute write so a later local failure cannot leave the cert out of sync with an
+            // authority that has already revoked it (state-divergence rule). The SM gates the
+            // transition but writes no audit here; the REVOKE/SUCCESS history is recorded only after
+            // the attribute write succeeds, so an attribute failure surfaces as the single
+            // connector-accepted-but-local-failure entry below, not a misleading SUCCESS+FAILED pair.
+            stateMachine.transitionAuditedExternally(certificate, CertificateState.REVOKED);
```
ivosh, line 1215: **~ Minor** | Testing

ClientOperationServiceImpl.java:1212 — the synchronous-200 revoke path was changed to stateMachine.transitionAuditedExternally(...) precisely so the REVOKE/SUCCESS history row is written only after the attribute write succeeds, avoiding (per the in-diff comment) 'a misleading SUCCESS+FAILED pair'. The pre-existing sync-revoke tests (ClientOperationServiceV2Test.java:1178 onward, revokeCertificateAction_completesSynchronously_*) assert only the resulting state (REVOKED); none assert the cert-event history, so the very behavior this change exists to guarantee — exactly one REVOKE/SUCCESS entry, recorded after the attribute write, no duplicate/ordering regression — is unverified. The PR added analogous history assertions for the cancel paths (cancelPendingCertificateOperation_* now check ISSUE/FAILED and REVOKE/FAILED) but not for this revoke change. Suggest adding a history assertion to one sync-revoke test. Not a merge blocker.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -331,15 +501,15 @@ public void issueCertificateAction(final UUID certificateUuid, boolean isApprove
 
         final Certificate certificate = certificateRepository.findWithAssociationsByUuid(certificateUuid).orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
         if (certificate.isArchived())
-            throw new ValidationException(ValidationError.create(String.format("Cannot issue requested certificate that has been archived. Certificate: %s", certificate.toStringShort())));
-        if (certificate.getState() != CertificateState.REQUESTED && certificate.getState() != CertificateState.PENDING_APPROVAL) {
-            throw new ValidationException(ValidationError.create(String.format("Cannot issue requested certificate with state %s. Certificate: %s", certificate.getState().getLabel(), certificate)));
+            throw new ValidationException(ValidationError.create(String.format("Cannot issue certificate that has been archived. Certificate: %s", certificate.toStringShort())));
+        if (certificate.getState() != CertificateState.REQUESTED && certificate.getState() != CertificateState.PENDING_APPROVAL && certificate.getState() != CertificateState.REGISTERED) {
```
ivosh, line 508: **. Nitpick** | Correctness

ClientOperationServiceImpl.java:505 now lets issueCertificateAction accept REGISTERED, and the synchronous (HTTP 200) issuance path eventually calls certificateService.issueRequestedCertificate(...), which sets state directly via CertificateUtil.prepareIssuedCertificate -> modal.setState(ISSUED) (CertificateUtil.java:379) — bypassing CertificateStateMachine entirely. The transition table (CertificateStateTransition.java) has no REGISTERED->ISSUED row, only REGISTERED->PENDING_ISSUE and REGISTERED->PENDING_APPROVAL. This is pre-existing behaviour for the sync issue path (REQUESTED->ISSUED is also not in the table and goes through the same direct setState), so it is not a defect introduced here, and async issuance correctly routes REGISTERED->PENDING_ISSUE->ISSUED through the SM.

Confirm this is intentional.

src/main/java/com/otilm/core/util/CertificateUtil.java
```diff
@@ -578,6 +578,23 @@ private static void setIssuerDNParams(Certificate modal, X500Name issuerDN) {
...
+     * permitted by RFC 5280 §4.1.2.6).
+     */
+    public static void applyRegistrationSubject(Certificate modal, String subjectDn) {
+        if (subjectDn == null || subjectDn.isBlank()) {
+            return;
+        }
+        try {
+            setSubjectDNParams(modal, new X500Name(new PlatformX500NameStyle(false), subjectDn));
+        } catch (IllegalArgumentException e) {
+            throw new ValidationException(ValidationError.create("Invalid subject DN '%s': %s".formatted(subjectDn, e.getMessage())));
```
ivosh, line 594: **. Nitpick** | Security

CertificateUtil.java:594 builds `Invalid subject DN '%s': %s` from the operator-supplied `subjectDn` and the raw `IllegalArgumentException.getMessage()` of BouncyCastle's X500Name parser, then throws ValidationException. ValidationException is a controlled domain type and is rendered to the REST client as HTTP 422 with its message (ExceptionHandlingAdvice.java:209), so the parser message reaches the wire. This is low risk: the cause is a library-level DN-syntax parse error describing the caller's own input (not SQL/stack/internal identifiers), and echoing back the user's own malformed input is acceptable. If you want to be conservative, drop the `: %s` and keep just `Invalid subject DN '%s'`. Take-it-or-leave-it; need not block the PR.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1094,6 +1096,73 @@ public Certificate createCertificateEntity(X509Certificate certificate) {
...
+    @Transactional
+    public void addCertificateRequestToExisting(UUID certificateUuid, ClientCertificateSignRequestDto signRequest)
+            throws CertificateRequestException, NoSuchAlgorithmException, NotFoundException {
+        if (signRequest == null || signRequest.getRequest() == null || signRequest.getRequest().isBlank()) {
+            throw new CertificateRequestException("A certificate signing request is required to complete a registered certificate");
+        }
+        if (signRequest.getFormat() == null) {
+            throw new CertificateRequestException("A certificate signing request format (PKCS10 or CRMF) is required");
+        }
+        Certificate certificate = getCertificateEntity(SecuredUUID.fromUUID(certificateUuid));
```
ivosh, line 1125: **. Nitpick** | Maintainability

CertificateServiceImpl.java:1125-1131: addCertificateRequestToExisting (@Transactional REQUIRED) loads the certificate via the unlocked getCertificateEntity and then asserts state==REGISTERED before overwriting the certificateRequest. The Javadoc calls this a "defense-in-depth" guard, but because the read takes no PESSIMISTIC_WRITE lock and the state remains REGISTERED throughout this method, the check does not actually serialize two concurrent attaches against the same placeholder — the second still sees REGISTERED and overwrites the first's certificateRequest/keyUuid. This is the same underlying race surfaced at the orchestrator in the major finding on issueExistingCertificate; flagged here as the deeper write site. If the orchestrator-level lock is added, this guard is fine as-is as a cheap secondary check; if not, consider loading via a locking finder here so the guard is genuinely atomic. Need not block on its own.

src/main/java/com/otilm/core/dao/repository/RaProfileRepository.java
```diff
@@ -12,6 +13,18 @@ public interface RaProfileRepository extends SecurityFilterRepository<RaProfile,
 
     Optional<RaProfile> findByUuid(UUID uuid);
 
+    /**
+     * Loads an RA profile with its authority, connector and connector interface eagerly. The v3 register
```
ivosh: Nitpick

I'd suggest to drop the second sentence because it restates the purpose of @EntityGraph. The first sentence is usefull, though.

ivosh (dismissed):
A couple of non-blocking findings, worth checking if they are intentional or need to be addressed.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Confirmed — and we'd independently flagged this same race in our own multi-reviewer pass before your review: two concurrent `issueExistingCertificate` calls both pass the unlocked `REGISTERED` guard and both enqueue ISSUE → double issuance. You're right that the `state == REGISTERED` guard added in this PR is only a cheap secondary check, not serialization.

We deferred the full fix — routing the REGISTERED read-attach-enqueue through `findAndLockWithAssociationsByUuid` with the assertion held under the lock, mirroring `manuallyConfirmRevoke` / `commitLocalCancel` — rather than grow this already-large PR, reasoning that: (a) it's a pre-existing shape (the operator issue path has never locked, and the connector call is in the async consumer, outside any lock, so the lock here is a purely local read-attach-enqueue guard); (b) the window needs two near-simultaneous operator issues on the *same* placeholder; (c) it's a contained change better reviewed on its own. Tracked as a dedicated follow-up: #1687 (linked under the authority-provider-v3 epic).

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
lubomirw: Agreed — this is the deeper write-site of the same race flagged on `issueExistingCertificate`. The `state == REGISTERED` guard here is a cheap secondary check, not a serialization point, as you note. When the orchestrator-level lock lands in `issueExistingCertificate` (follow-up #1687), this guard stays as the secondary check rather than adding a second lock. (Separately, the Base64 cause-chaining you flagged just below is fixed in 090ae25a4.)

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Confirmed intentional. The sync-200 path setting ISSUED directly via `prepareIssuedCertificate` is pre-existing (REQUESTED→ISSUED does the same), and async correctly routes REGISTERED→PENDING_ISSUE→ISSUED through the SM. Completing SM coverage for the remaining direct-`setState` writers (`prepareIssuedCertificate`→ISSUED and `revokeCertificate(serialNumber)`→REVOKED) is tracked as #1683.

src/main/java/com/otilm/core/util/CertificateUtil.java
lubomirw: Leaving as-is, per your read: the echoed text is BouncyCastle's DN-syntax error describing the caller's own malformed `subjectDn` (not internal identifiers or stack), surfaced as 422 — acceptable. Noted for a broader 422-body-hygiene pass if we tighten connector error messages later.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Fixed in 090ae25a4: added a `catch (RuntimeException)` that fails the placeholder (the rejection-path treatment) instead of orphaning it in PENDING_REGISTRATION, and hoisted the post-acceptance contract into `RegisterCapability.register`'s Javadoc so a raw runtime is provably pre-acceptance for every adapter. Thanks for tracing the three pre-acceptance throw sites.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
lubomirw: Fixed in 090ae25a4 — `new CertificateRequestException("Certificate signing request is not valid Base64", e)`, chaining the cause like `Pkcs10CertificateRequest` / `CrmfCertificateRequest`.

src/test/java/com/otilm/core/service/ClientOperationServiceRegisterTest.java
lubomirw: Fixed in 090ae25a4 — the test now mocks `AttributeEngine` and verifies `updateMetadataAttributes(...)`, so it exercises the persistence branch instead of passing on a swallowed real-engine failure.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Fixed in 090ae25a4 — added a history assertion to a sync-200 revoke test: exactly one REVOKE entry, status SUCCESS (no SUCCESS+FAILED pair).

src/main/java/com/otilm/core/dao/repository/RaProfileRepository.java
lubomirw: Done in 090ae25a4 — dropped the second sentence, kept the first.

ivosh (approved):

Issue #748, opened 2024-05-29
Audit log purge timeout


3keyroman:
**Describe the bug**
When there are many audit logs and I want to purge the audit log, the operation will timeout. The same behaviour is on the UI; when I click on Purge button, the wheel is spinning until I get the `Failed to purge audit logs (504): ajax error 504`.

**To Reproduce**
1. Go to Audit logs or use the API
2. Click on Purge
3. Wait until timeout

**Expected behavior**
Purging of audit logs should not produce timeout also in case there are many logs, I believe it should be handled asynchronously.

**Screenshots**
N/A

**Additional context**
Version 2.11.0

PR #1689, opened 2026-06-30
Add RA-profile request-attribute configuration with value-source bindings

### LINKED ISSUE #1632: Add RA-profile static request-attribute set, value-source binding, resolver, and value-source SPI
### Description
Persist and resolve the RA-Profile static certificate-request-attribute set, the value-source binding, and the editable platform default set; expose the value-source provider SPI seam. Part of #1572 (epic OmniTrustILM/ilm#131).

### Scope — what ships
- `RaProfileCertificateRequestAttribute` + `RaProfileValueSourceBinding` serialized-attribute entities (mirror `RaProfileProtocolAttribute`); editable **platform default set** as a `Setting` singleton, seeded from CO-1's `CsrRequestAttributes` (so admins can add/remove/reorder the defaults that are hardcoded today in `CsrAttributes`) — surfaced through the platform `CertificateSettings` (`/platform`) per interfaces #731's fold, not a bespoke endpoint.
- Pure `RequestAttributeSetResolver` — merge modes (`AttributeSetMergeMode`: STATIC_ONLY / CONNECTOR_ONLY / MERGE), key = UUID with name fallback, value-source-binding application.
- Orchestrator (`NOT_SUPPORTED`, no class-level `@Transactional`) + `…Writer` bean; Flyway.
- **Collections SPI seam:** define a `CollectionValueSourceProvider` interface; the resolver calls a registered provider for a `valueSource.type=COLLECTION`, and with no provider registered (no collections subsystem yet) it is a validated passthrough/no-op. The actual provider impl is tracked under ilm#132 (`blocked-by` this issue).

### Dependencies
- interfaces #730 + #731 (set DTOs / merge-mode enum) — SNAPSHOT-first; CO-1 (default-set seed, soft).

### Acceptance Criteria
- Exhaustive resolver kernel unit tests + Testcontainers entity/orchestrator tests; new-code coverage ≥80%; SonarCloud gate clean.

### Review focus
Two-bean transactional split; merge-key (UUID/name) semantics; the SPI contract is stable for the ilm#132 provider.

> Flyway: re-stamp the migration id to sort after the current latest at execution.

ivosh:


src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
```diff
@@ -0,0 +1,244 @@
...
+        this.requestAttributeRepository = requestAttributeRepository;
+        this.valueSourceBindingRepository = valueSourceBindingRepository;
+        this.writer = writer;
+        this.extendedAttributeService = extendedAttributeService;
+        this.settingRepository = settingRepository;
+        this.objectMapper = objectMapper;
+    }
+
+    @Override
+    public List<BaseAttribute> resolveIssueAttributeSet(RaProfile raProfile, AttributeSetMergeMode mode)
```
lubomirw, line 68: **[Medium]** | Performance

`resolveIssueAttributeSet` makes an outbound connector HTTP call (`listConnectorIssueAttributes` → `extendedAttributeService.listIssueCertificateAttributes`) with no `@Transactional(propagation = NOT_SUPPORTED)`. Invoked under the class-level-`@Transactional` `RaProfileServiceImpl`, the JDBC connection is held open across the round-trip → connection-pool exhaustion risk with a slow/hanging connector. Latent — no caller in this PR yet (new SPI).

_Suggested:_ annotate both overloads `@Transactional(propagation = Propagation.NOT_SUPPORTED)` and pre-fetch the static set + bindings before the HTTP call.

src/main/java/com/otilm/core/certificate/request/RequestAttributeSetResolver.java
```diff
@@ -0,0 +1,132 @@
...
+                                                               List<ValueSourceBindingSpec> bindings) {
+        if (definitions == null) {
+            return new ArrayList<>();
+        }
+        List<ValueSourceBindingSpec> bindingList = bindings == null ? List.of() : bindings;
+        for (BaseAttribute def : definitions) {
+            if (!(def instanceof DataAttributeV3 v3)) {
+                continue;
+            }
+            findValueSourceBinding(v3, bindingList).ifPresent(binding -> v3.setValueSource(toValueSource(binding)));
```
lubomirw, line 109: **[Low]** | Correctness

`applyValueSourceBindings` unconditionally calls `v3.setValueSource(...)` on any matched `DataAttributeV3`, overwriting a `valueSource` a connector may already supply. Low impact today (connector defs don't carry one), latent otherwise.

_Suggested:_ only set when `v3.getValueSource() == null`, or document the override as intentional; add a test for a connector def that already carries a `valueSource`.

<sub>flagged by Superpowers review</sub>

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
```diff
@@ -0,0 +1,244 @@
...
+            throws ConnectorException, NotFoundException {
+        List<BaseAttribute> staticSet = getStaticSet(raProfile);
+        List<BaseAttribute> connectorSet =
+                RequestAttributeSetResolver.effectiveMode(mode) == AttributeSetMergeMode.STATIC_ONLY
+                        ? List.of()
+                        : listConnectorIssueAttributes(raProfile);
+
+        List<BaseAttribute> merged = RequestAttributeSetResolver.merge(staticSet, connectorSet, mode);
+        if (merged.isEmpty()) {
```
lubomirw, line 89: **[Low]** | Correctness

The empty-merge fallback to `getDefaultSet()` fires regardless of merge mode. `CONNECTOR_ONLY` with an empty connector result (or `STATIC_ONLY` with an empty static set) returns the platform default CSR attributes — neither connector- nor profile-sourced. A test documents this as intended, but it may surprise an operator who deliberately chose a single-source mode.

_Suggested:_ gate the fallback on the effective mode, or document it explicitly on the interface method + add a `CONNECTOR_ONLY`-with-empty-connector test.

<sub>flagged by Superpowers + Guided</sub>

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    @Override
+    public List<BaseAttribute> resolveIssueAttributeSet(RaProfile raProfile) throws ConnectorException, NotFoundException {
+        return resolveIssueAttributeSet(raProfile, getStoredMergeMode(raProfile));
+    }
+
+    private List<BaseAttribute> listConnectorIssueAttributes(RaProfile raProfile) throws ConnectorException, NotFoundException {
+        if (raProfile.getAuthorityInstanceReference() == null
+                || raProfile.getAuthorityInstanceReference().getConnectorUuid() == null) {
```
lubomirw: **[Low]** | Error handling

This guard checks `authorityInstanceReference == null || getConnectorUuid() == null`, but `ExtendedAttributeServiceImpl.listIssueCertificateAttributes` guards on `getConnector() == null` and throws `NotFoundException`. A profile with a non-null `connectorUuid` but a deleted/unresolved Connector passes this guard, reaches the call, and surfaces `NotFoundException` instead of the graceful empty-set this resolver intends for offline/external authorities.

_Suggested:_ align to `getConnector() == null`, or catch `NotFoundException` and treat it as no-dynamic-set; add a test for the connectorUuid-set-but-Connector-missing case.

lubomirw (commented):
Multi-reviewer pass (Copilot CLI · Guided · Superpowers · Hygiene), deduped and verified against interfaces #748. 9 findings — 8 inline below, plus one cross-repo item here.

_Excluded as likely-noise after verification:_ NPE on null `valueSourceType` in `toBindingEntities` (unreachable — the endpoint is `@Valid` and `valueSourceType` is `@NotNull`, so a null is a 400 before it reaches the mapper); `updateConfiguration` non-atomic across two writers (atomic today — the sole caller `RaProfileServiceImpl` is class-level `@Transactional`, so both writers join one ambient tx; adding `@Transactional` to `updateConfiguration` is cheap self-enforcing hardening); value-source binding repo not SecurityFilter-scoped (consistent with the protocol-attribute pattern, gated by `@ExternalAuthorization` at the service); in-place `DataAttributeV3` mutation in `applyValueSourceBindings` (documented, and all current inputs are freshly built).

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
```diff
@@ -0,0 +1,244 @@
...
+            // fall back to the editable platform default set when nothing resolved.
+            merged = new ArrayList<>(getDefaultSet());
+        }
+
+        return RequestAttributeSetResolver.applyValueSourceBindings(merged, loadValueSourceBindings(raProfile));
+    }
+
+    @Override
+    public List<BaseAttribute> resolveIssueAttributeSet(RaProfile raProfile) throws ConnectorException, NotFoundException {
+        return resolveIssueAttributeSet(raProfile, getStoredMergeMode(raProfile));
```
lubomirw: **[Low]** | Performance

The no-arg `resolveIssueAttributeSet(raProfile)` calls `getStoredMergeMode` (one `findByRaProfileUuid`) then `resolveIssueAttributeSet(profile, mode)` → `getStaticSet` (a second `findByRaProfileUuid` for the same row).

_Suggested:_ fetch the `RaProfileCertificateRequestAttribute` row once and pass the merge mode + parsed static set down.

<sub>flagged by Superpowers review</sub>

src/main/java/com/otilm/core/dao/entity/RaProfile.java
```diff
@@ -271,6 +275,13 @@ public RaProfileProtocolAttribute getProtocolAttribute() {
         return protocolAttribute;
     }
 
+    public RaProfileCertificateRequestAttribute getRequestAttribute() {
```
lubomirw, line 278: **[Low]** | Maintainability

`getRequestAttribute()` returns a freshly-constructed transient `RaProfileCertificateRequestAttribute` when the association is null, and has no main-code callers (the `RaProfileDto` field is populated via `requestAttributeService.getConfiguration`, not this getter). Footgun: a future `getRequestAttribute().setX()` would mutate a throwaway object expecting persistence. Diverges from the sibling `getProtocolAttribute` pattern.

_Suggested:_ remove the unused getter, or return the raw field / `Optional` and route persistence through the writer.

<sub>flagged by Superpowers review</sub>

src/main/resources/db/migration/V202606271400__ra_profile_request_attribute.sql
```diff
@@ -0,0 +1,35 @@
...
+    CONSTRAINT uq_ra_profile_request_attribute_ra_profile UNIQUE (ra_profile_uuid)
+);
+
+ALTER TABLE ra_profile_request_attribute
+    ADD CONSTRAINT ra_profile_request_attribute_to_ra_profile
+    FOREIGN KEY (ra_profile_uuid)
+    REFERENCES ra_profile
+    ON UPDATE NO ACTION ON DELETE CASCADE;
+
+CREATE TABLE ra_profile_value_source_binding (
```
lubomirw, line 17: **[Low]** | Maintainability

`ra_profile_value_source_binding` has only PK(uuid) + index(ra_profile_uuid) — no uniqueness on `(ra_profile_uuid, attribute_uuid)` / `(ra_profile_uuid, attribute_name)`. `replaceValueSourceBindings` is delete-then-insert so it can't self-duplicate, but two bindings targeting the same attribute in one payload are accepted and `findValueSourceBinding` applies the first, silently ignoring the rest. (The both-null-target case is already blocked by `ValueSourceBindingDto.@AssertTrue isValid()`.)

_Suggested:_ validate duplicate attribute targets in `updateConfiguration`, or add a partial unique constraint.

src/main/java/com/otilm/core/api/web/RAProfileManagementControllerImpl.java
```diff
@@ -83,6 +84,12 @@ public RaProfileDto updateRaProfileValidationConfiguration(String authorityUuid,
         return raProfileService.updateRaProfileValidationConfiguration(SecuredParentUUID.fromString(authorityUuid), SecuredUUID.fromString(raProfileUuid), request);
     }
 
+    @Override
+    @AuditLogged(module = Module.CERTIFICATES, resource = Resource.RA_PROFILE, affiliatedResource = Resource.AUTHORITY, operation = Operation.UPDATE)
+    public RaProfileDto updateRaProfileRequestAttributesConfiguration(String authorityUuid, String raProfileUuid, RaProfileCertificateRequestAttributesUpdateDto request) throws NotFoundException {
```
lubomirw, line 89: **[Low]** | Documentation

This new operator-facing endpoint, the `RaProfileDto.certificateRequestAttributes` field, two new `PLATFORM_CERTIFICATES` settings (`certificatesRequestAttributes`, `certificatesRequestAttributesStrict`), and the `CollectionValueSourceProvider` SPI are user-facing. docs.otilm.com (RA Profile management, Platform Settings › Certificates) and interface-documentation will be stale until updated.

_Suggested:_ open or link a companion docs task covering the new RA-profile request-attribute configuration, the merge-mode/strictness semantics, the two platform settings, and the value-source SPI.

<sub>flagged by Guided / Docs check</sub>

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
ivosh: Fixed in cf94c30e — both `resolveIssueAttributeSet` overloads now run with `@Transactional(propagation = NOT_SUPPORTED)`, so the connector HTTP call is not made while holding a DB connection.

src/main/java/com/otilm/core/certificate/request/RequestAttributeSetResolver.java
ivosh: Documented as intentional in cf94c30e — a matched binding deliberately overrides any connector-supplied `valueSource` (the operator's explicit Core-side binding is authoritative).

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
ivosh: Fixed in cf94c30e — the guard now checks `getConnector() == null` (matching `ExtendedAttributeServiceImpl`), so a profile whose Connector is deleted resolves to the graceful empty set instead of surfacing `NotFoundException`. Added a regression test (`connectorUuidSetButConnectorMissingSkipsConnectorSetGracefully`).

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
ivosh: Fixed in b98210e4 — both overloads now fetch the `ra_profile_request_attribute` row once via `loadStoredSet` and pass the merge mode + static set into a shared `resolve()` path, so the no-arg overload no longer reads the row twice.

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
```diff
@@ -0,0 +1,244 @@
...
+
+    @Override
+    public RaProfileCertificateRequestAttributesDto getConfiguration(RaProfile raProfile) {
+        RaProfileCertificateRequestAttributesDto dto = new RaProfileCertificateRequestAttributesDto();
+        RaProfileCertificateRequestAttribute set = requestAttributeRepository.findByRaProfileUuid(raProfile.getUuid())
+                .orElse(null);
+        if (set != null) {
+            dto.setRequestAttributes(deserializeOrEmpty(set.getRequestAttributes()));
+            dto.setExternalCsrValidationStrict(set.getExternalCsrValidationStrict());
+        }
```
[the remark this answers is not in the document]
ivosh: Fixed in cf94c30e — `getConfiguration()` now always sets `requestAttributes` to a list (empty when no row), so the DTO shape is consistent.

src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
```diff
@@ -137,11 +141,37 @@ public PlatformSettingsDto getPlatformSettings() {
...
+    private CertificateRequestAttributesSettingsDto readRequestAttributesSettings(Map<String, Setting> certificateSettings) {
+        CertificateRequestAttributesSettingsDto dto = new CertificateRequestAttributesSettingsDto();
+        Setting definitions = certificateSettings == null ? null : certificateSettings.get(DefaultRequestAttributeSet.SETTING_NAME);
+        // resolve() seeds the built-in default set (CsrAttributes) when nothing has been stored yet.
+        dto.setRequestAttributes(DefaultRequestAttributeSet.resolve(definitions == null ? null : definitions.getValue()));
+
+        Setting strict = certificateSettings == null ? null : certificateSettings.get(DefaultRequestAttributeSet.STRICT_SETTING_NAME);
+        if (strict != null && strict.getValue() != null) {
+            dto.setExternalCsrValidationStrict(Boolean.valueOf(strict.getValue()));
+        }
```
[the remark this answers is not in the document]
ivosh: Fixed in cf94c30e — a blank setting value is now treated as unset (trimmed + `isBlank()` check) rather than parsed as `false`.

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
ivosh: Leaving as-is, intentionally. The empty-merge fallback to the platform default set is a deliberate safety net so an issue/sign operation never proceeds with no resolvable attributes — including `CONNECTOR_ONLY` with an empty connector result or `STATIC_ONLY` with an empty static set. The existing test documents this. 

Design spec §4.3 defines the resolution order as static set → connector dynamic set → platform default set.

src/main/java/com/otilm/core/dao/entity/RaProfile.java
ivosh: Leaving as-is. It intentionally mirrors the sibling `getProtocolAttribute()` (identical null→transient pattern), so it's consistent rather than divergent. The `RaProfileDto` field is populated via `requestAttributeService.getConfiguration`, not this getter. If you'd like both getters changed to return `Optional`/raw field, that's a separate cleanup spanning the protocol-attribute path too.

Also refer to Phase 3 plan line 452: "the null-safe getter must return a fresh empty entity (mirrors getProtocolAttribute)."

src/main/resources/db/migration/V202606271400__ra_profile_request_attribute.sql
ivosh: Handled at the contract layer in interfaces #748 (commit 7634d649): `RaProfileCertificateRequestAttributesUpdateDto` now carries an `@AssertTrue` (`isValueSourceBindingsUnique`) rejecting two bindings that target the same attribute by uuid or by name, with tests. So duplicate targets are a 400 before they reach persistence; no core-side change needed here.

src/main/java/com/otilm/core/api/web/RAProfileManagementControllerImpl.java
ivosh: Acknowledged — not a code change. Tracking as a separate docs follow-up (docs.otilm.com RA Profile management + Platform Settings › Certificates, and interface-documentation) covering the new endpoint, the `certificateRequestAttributes` field, the two PLATFORM_CERTIFICATES settings, and the value-source SPI. Not addressed in this PR.

src/main/java/com/otilm/core/service/RaProfileService.java
```diff
@@ -207,4 +208,14 @@ List<CertificateDetailDto> getAuthorityCertificateChain(SecuredParentUUID author
...
+
+    /**
+     * Update the static request-attribute configuration (definitions, merge mode, value-source bindings, strictness) of the RA Profile.
+     *
+     * @param authorityUuid UUID of the authority associated with the RA profile
+     * @param raProfileUuid UUID of the RA Profile
+     * @param request Request-attribute configuration request
+     * @return Edited RA Profile
+     */
+    RaProfileDto updateRaProfileRequestAttributesConfiguration(SecuredParentUUID authorityUuid, SecuredUUID raProfileUuid, RaProfileCertificateRequestAttributesUpdateDto request) throws NotFoundException;
```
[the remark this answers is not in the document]
ivosh: Not a defect. `authorityUuid` is a `SecuredParentUUID` consumed by the `@ExternalAuthorization(parentResource = AUTHORITY, parentAction = DETAIL)` aspect for parent-resource authorization — not dead. It matches the sibling `updateRaProfileValidationConfiguration` signature; removing it would drop the parent authz check.

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
```diff
@@ -0,0 +1,244 @@
...
+        for (ValueSourceBindingDto binding : bindings) {
+            RaProfileValueSourceBinding entity = new RaProfileValueSourceBinding();
+            entity.setRaProfileUuid(raProfile.getUuid());
+            entity.setAttributeUuid(binding.getAttributeUuid());
+            entity.setAttributeName(binding.getAttributeName());
+            entity.setValueSourceType(binding.getValueSourceType().name());
+            entity.setCollectionRef(binding.getCollectionRef());
+            entity.setParams(serializeParams(binding.getParams()));
+            entities.add(entity);
+        }
```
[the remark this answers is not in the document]
ivosh: Not reachable. The endpoint is `@Valid`, `ValueSourceBindingDto.valueSourceType` is `@NotNull`, and `@AssertTrue isValid()` already requires uuid-or-name — a malformed binding is a 400 before it reaches `toBindingEntities`.

src/main/java/com/otilm/core/certificate/request/RequestAttributeSetResolver.java
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * A Core-side value-source binding, decoupled from the {@code RaProfileValueSourceBinding} entity so the
+     * kernel stays Spring/JPA-free. {@code attributeUuid} is the primary key; {@code attributeName} is the fallback.
+     */
+    public record ValueSourceBindingSpec(String attributeUuid,
+                                         String attributeName,
+                                         ValueSourceType valueSourceType,
+                                         String collectionRef,
+                                         List<SourceParam> params) {
+    }
```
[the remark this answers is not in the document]
ivosh: Not lost. `ValueSource` carries only `kind` + `params`, so `collectionRef` cannot be projected onto it; it is persisted and round-tripped through the configuration DTO and consumed server-side by the `CollectionValueSourceProvider` SPI (resolution deferred). The merge kernel intentionally does not apply it.

src/main/java/com/otilm/core/dao/repository/RaProfileCertificateRequestAttributeRepository.java
```diff
@@ -0,0 +1,13 @@
+package com.otilm.core.dao.repository;
+
+import com.otilm.core.dao.entity.RaProfileCertificateRequestAttribute;
+import org.springframework.stereotype.Repository;
+
+import java.util.Optional;
+import java.util.UUID;
+
+@Repository
+public interface RaProfileCertificateRequestAttributeRepository extends SecurityFilterRepository<RaProfileCertificateRequestAttribute, Long> {
```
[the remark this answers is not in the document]
ivosh: Matches the established repository convention here — `RaProfileRepository`, `RaProfileProtocolAttributeRepository`, and `CertificateProtocolAssociationRepository` all declare `Long` despite a `UUID` `@Id`. The typed `findById`/`deleteById` are never used (custom `findByRaProfileUuid`). Changing only this repo would make it inconsistent with its siblings.

src/main/java/com/otilm/core/dao/repository/RaProfileValueSourceBindingRepository.java
```diff
@@ -0,0 +1,15 @@
+package com.otilm.core.dao.repository;
+
+import com.otilm.core.dao.entity.RaProfileValueSourceBinding;
+import org.springframework.stereotype.Repository;
+
+import java.util.List;
+import java.util.UUID;
+
+@Repository
+public interface RaProfileValueSourceBindingRepository extends SecurityFilterRepository<RaProfileValueSourceBinding, Long> {
```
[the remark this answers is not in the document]
ivosh: Same as the sibling repo: `Long` follows the established convention across the RA-profile repositories, and the typed `findById`/`deleteById` are never used (custom `findByRaProfileUuid`/`deleteByRaProfileUuid`).

src/main/java/com/otilm/core/api/web/RAProfileManagementControllerImpl.java
ivosh: Correction to my reply above: this is already tracked, not a new follow-up. The operator-facing docs (RA-profile authoring — static set, default set, value-source bindings, merge mode — plus the external-CSR strict/lenient subsection) are covered by OmniTrustILM/documentation#310, under the umbrella OmniTrustILM/documentation#309, blocked by core#1632. The CollectionValueSourceProvider SPI's concept doc is deferred to the Collections epic (ilm#132) per #309's deferral note. No code change in this PR.

lubomirw (approved):

PR #1694, opened 2026-07-01
Validate uploaded (Mode B) CSRs against RA-profile request-attribute policy

### LINKED ISSUE #1631: Validate external CSRs against the resolved request-attribute set (Mode B)
### Description
Validate a client-supplied (external) CSR against the resolved request-attribute set. Part of #1572 (epic OmniTrustILM/ilm#131).

### Scope — what ships
- Pure `X509RequestContentParser` (CSR/CRMF → `X509RequestContent`, the inverse of CO-1).
- `CertificateRequestContentValidator` (static kernel + thin instance method; reuses `AttributeDefinitionUtils.validateConstraints`) — the canonical validator CO-5 reuses.
- Parse+validate wired into the upload-CSR issue branch. Strict rejects via a shaped domain exception (no raw leak); lenient warns.
- Strict/lenient policy is **read from** the Phase 3 request-attribute config (`RaProfileCertificateRequestAttribute.externalCsrValidationStrict`, accessed via `raProfile.getRequestAttribute()`); `null` inherits the platform default setting `certificatesRequestAttributesStrict`. **No new column/entity/migration on `RaProfile`** — that persistence shipped in #1632 (PR #1689).

### Dependencies
- interfaces #730 (SNAPSHOT-first); CO-1 (soft, documented fallback). Its validator is a hard prerequisite for CO-5.
- **#1632 (PR #1689)** — provides the persisted strict flag and the resolved set. The validation wiring consumes `RaProfileCertificateRequestAttributeService.resolveIssueAttributeSet(raProfile)` (static → connector → platform default, per merge mode, with value-source bindings applied) rather than a hardcoded default set.

### Acceptance Criteria
- Kernel unit tests + integration; new-code coverage ≥80%; SonarCloud gate clean.

### Review focus
The canonical validator contract; no runtime detail leaked to the wire on rejection.

### Engineering notes
- `resolveIssueAttributeSet(...)` performs a **connector call** to fetch the dynamic set, so the resolve+validate step must run **outside any DB transaction** (no tx across an external call). The Phase 2 plan's "local CPU only / no external HTTP" assumption no longer holds once consuming the real resolver.
- No Flyway migration is added by this issue — the `external_csr_validation_strict` column already ships in `V202606271400__ra_profile_request_attribute.sql` (#1689).

> **Plan reconciliation (post #1689):** the original scope persisted `externalCsrValidationStrict` directly on `RaProfile` (+ its own Flyway migration). #1632/PR #1689 instead placed the flag on the dedicated `RaProfileCertificateRequestAttribute` entity (spec §4.3 / sibling-entity decision) and supplied `RaProfileCertificateRequestAttributeService` as the resolution surface. Phase 2 now reads the flag and consumes the resolver service; the standalone `RaProfile` column + migration (plan Task 5) is dropped.

ivosh:


src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1764,12 +1783,77 @@ private static String normalizeRdn(String rdn, Map<String, String> codeToOid) {
...
+    private void validateUploadedRequestAttributes(String csr, CertificateRequestFormat requestFormat, RaProfile raProfile)
+            throws CertificateException {
+        if (raProfile == null) {
+            return;
+        }
+        List<BaseAttribute> definitions;
+        try {
+            definitions = requestAttributeService.resolveIssueAttributeSet(raProfile);
+        } catch (ConnectorException | NotFoundException e) {
+            logger.debug("Could not resolve request-attribute set for Mode B validation (RA profile {})", raProfile.getName(), e);
```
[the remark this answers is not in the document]
ivosh: Fixed in 78c28f3c. Went beyond the log level: since the design intent is strict = reject, a strict RA profile now **fails closed** on a resolution failure (WARN + reject issuance) rather than silently skipping. Lenient profiles proceed unvalidated but now log at WARN.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1764,12 +1783,77 @@ private static String normalizeRdn(String rdn, Map<String, String> codeToOid) {
...
+                logger.warn("Request-attribute validation failed for uploaded CSR (RA profile {}): {}",
+                        raProfile.getName(), result.getErrors());
+                throw new CertificateRequestValidationException(
+                        "Uploaded certificate request does not satisfy the request-attribute policy of RA profile '%s'"
+                                .formatted(raProfile.getName()),
+                        result.getErrors());
+            }
+        } catch (CertificateRequestException e) {
+            logger.debug("Failed to parse uploaded CSR for request-attribute validation", e);
+            throw new CertificateException("Uploaded certificate request could not be parsed for validation");
```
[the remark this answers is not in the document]
ivosh: Fixed in 78c28f3c.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1764,12 +1783,77 @@ private static String normalizeRdn(String rdn, Map<String, String> codeToOid) {
...
+        } catch (CertificateRequestValidationException e) {
+            throw new CertificateException(e.getMessage());
+        }
```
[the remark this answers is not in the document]
ivosh: Fixed in 78c28f3c.

src/main/java/com/otilm/core/certificate/request/X509RequestContentParser.java
```diff
@@ -0,0 +1,209 @@
...
+        }
+        // Reverse because PlatformX500NameStyle.DEFAULT reverses RDN order for display
+        Collections.reverse(result);
+        return result;
+    }
+
+    private static List<GeneralNameEntry> parseSans(CertificateRequest request) {
+        List<GeneralNameEntry> result = new ArrayList<>();
+        Map<String, List<String>> sans = CertificateUtil.getSAN(request);
+        for (Map.Entry<String, List<String>> e : sans.entrySet()) {
```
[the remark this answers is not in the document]
ivosh: Fixed in 78c28f3c. parseSans() now guards a CRMF request whose CertTemplate has no extensions, returning no SANs instead of hitting GeneralNames.fromExtensions(null, ...).

src/test/java/com/otilm/core/certificate/request/X509RequestContentParserTest.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    static {
+        // Seed the OidHandler with standard RDN attribute types for PlatformX500NameStyle
+        OidHandler.cacheOidCategory(OidCategory.RDN_ATTRIBUTE_TYPE, new HashMap<>());
+        OidHandler.cacheOid(OidCategory.RDN_ATTRIBUTE_TYPE, "2.5.4.3",
+                OidRecord.builder().displayName("Common Name").code("CN").build());
+        OidHandler.cacheOid(OidCategory.RDN_ATTRIBUTE_TYPE, "2.5.4.10",
+                OidRecord.builder().displayName("Organization").code("O").build());
+    }
```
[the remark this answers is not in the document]
ivosh: Fixed in 78c28f3c. The seeding moved into @BeforeAll after the snapshot, so the original global cache is captured and restored.

src/test/java/com/otilm/core/service/v2/CertificateRequestModeBValidationTest.java
```diff
@@ -0,0 +1,202 @@
...
+    @Autowired
+    private CertificateRepository certificateRepository;
+
+    private WireMockServer mockServer;
+    private RaProfile raProfile;
+    private String csrMissingCommonName;
+
+    @BeforeEach
+    void wireStrictRaProfileRequiringCommonName() throws Exception {
+        // given — an RA profile with no authority connector attached (offline authority)
```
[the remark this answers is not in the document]
ivosh: Fixed in 78c28f3c.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1764,12 +1783,84 @@ private static String normalizeRdn(String rdn, Map<String, String> codeToOid) {
...
+        List<BaseAttribute> definitions;
+        try {
+            definitions = requestAttributeService.resolveIssueAttributeSet(raProfile);
+        } catch (ConnectorException | NotFoundException e) {
+            if (resolveExternalCsrValidationStrict(raProfile)) {
+                logger.warn("Could not resolve request-attribute set for Mode B validation; strict RA profile {} rejects issuance during the connector outage",
+                        raProfile.getName(), e);
+                throw new CertificateException(
+                        "Request-attribute set is unavailable; strict RA profile '%s' cannot validate the uploaded certificate request"
+                                .formatted(raProfile.getName()));
```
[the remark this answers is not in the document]
ivosh: Fixed in c9dab32d.

src/test/java/com/otilm/core/service/v2/UploadedCsrValidationTest.java
```diff
@@ -0,0 +1,227 @@
...
+        config.setExternalCsrValidationStrict(Boolean.TRUE);
+        requestAttributeService.updateConfiguration(raProfile, config);
+
+        csrMissingCommonName = pemEncodedCsrWithSubject("O=Acme,C=US");
+    }
+
+    @AfterEach
+    void stopMockServer() {
+        mockServer.stop();
+    }
```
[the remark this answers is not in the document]
ivosh: Fixed in c9dab32d.

src/main/java/com/otilm/core/certificate/request/X509RequestContentParser.java
```diff
@@ -0,0 +1,212 @@
...
+        }
+        // Reverse because PlatformX500NameStyle.DEFAULT reverses RDN order for display
+        Collections.reverse(result);
+        return result;
+    }
+
+    private static List<GeneralNameEntry> parseSans(CertificateRequest request) {
+        List<GeneralNameEntry> result = new ArrayList<>();
+        if (request instanceof CrmfCertificateRequest && extractExtensions(request) == null) {
+            return result; // CMRF request with no extensions -> no SANs
```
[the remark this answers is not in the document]
ivosh: Fixed in c9dab32d.

lubomirw (commented):
## Multi-angle review — 12 findings (each verified against the code)

Four independent reviewers (Copilot CLI, guided-pr-review, superpowers, hygiene) ran over the diff, then every finding was adversarially re-verified against the actual code. The 12 confirmed items are inline, sorted by severity: **2 High, 3 Medium, 4 Low, 3 Nit**.

**Two confirmed exploitable issues stand out** — both let the policy engine and the CA see *different* content, because the uploaded CSR is forwarded verbatim while the parser feeds a lossy view to validation:
- **#1** subject-DN parsing splits an unescaped display string, so a value with a literal `,`/`+` mis-splits → attribute-policy bypass.
- **#2** iPAddress SANs are compared in hex-DER form (`#0a000001`), so dotted-IP rules never match.

**Documentation (Low, not inline — different repo):** `documentation` → `docs/certificate-key/concept-design/core-components/ra-profile.md` documents lifecycle validation only. It should also cover the new `externalCsrValidationStrict` (Mode-B) semantics this PR activates: strict rejects non-conforming uploads and fails closed on connector outage; lenient logs warnings and proceeds.

**Checked and cleared (raised but refuted on verification):** an otherName `=`-split concern (value extraction is actually correct — the delimiter is the first `=`, and the OID prefix is `=`-free) and a "strictness resolved twice" concern (the two calls sit on mutually-exclusive branches). One minor *genuine* redundancy remains, for awareness: on the success path the RA-profile request-attribute row is read twice (once in `resolveIssueAttributeSet`, once in `resolveExternalCsrValidationStrict`).

**PR title:** accurate — no change suggested.

src/main/java/com/otilm/core/certificate/request/X509RequestContentParser.java
```diff
@@ -0,0 +1,267 @@
...
+        return x509;
+    }
+
+    private static List<RdnEntry> parseSubject(CertificateRequest request) {
+        List<RdnEntry> result = new ArrayList<>();
+        X500Name subject = request.getSubject();
+        if (subject == null) {
+            return result;
+        }
+        String dn = PlatformX500NameStyle.DEFAULT.toString(subject);
```
lubomirw: **[High]** | Security

`parseSubject` builds the DN with `PlatformX500NameStyle.DEFAULT.toString(subject)` then splits it on `,`/`+` via `splitOnUnescaped`. That style appends `value.toString()` with **no RFC 4514 escaping**, and `splitOnUnescaped` only treats backslash as an escape — so an RDN value containing a literal `,`/`+` (e.g. `CN=Acme, Inc.`) is split into fabricated/truncated RDNs. The policy engine then validates the corrupted values while the original CSR is re-encoded verbatim and forwarded to the CA with the true value → whitelist/REGEXP/required checks can be bypassed (and benign commas in O/OU cause false rejects).

_Suggested:_ Don't round-trip through a display string for a security decision — walk `subject.getRDNs()`/`AttributeTypeAndValue` objects directly, or render with an RFC 4514-escaping style and split with an escape-aware parser.

<sub>flagged by Copilot CLI + superpowers; verified against BouncyCastle behavior</sub>

src/main/java/com/otilm/core/certificate/request/X509RequestContentParser.java
```diff
@@ -0,0 +1,267 @@
...
+        entry.setValue(value);
+        return entry;
+    }
+
+    private static List<GeneralNameEntry> parseSans(CertificateRequest request) {
+        List<GeneralNameEntry> result = new ArrayList<>();
+        if (request instanceof CrmfCertificateRequest && extractExtensions(request) == null) {
+            return result; // CRMF request with no extensions -> no SANs
+        }
+        Map<String, List<String>> sans = CertificateUtil.getSAN(request);
```
lubomirw: **[High]** | Correctness

iPAddress SAN values arrive hex-DER-encoded, not dotted: `CertificateUtil.getSAN` returns `value.toString()`, and BC `ASN1OctetString.toString()` renders `10.0.0.1` as `#0a000001`. Any dotted-IP REGEXP/whitelist/required rule silently never matches — strict mode falsely rejects legitimate IP-SAN CSRs, and IP allow-lists are unenforceable.

_Suggested:_ Decode the iPAddress octets to canonical dotted/colon form (e.g. `InetAddress.getByAddress(...)`) before building the SAN entry; add a test asserting the parsed value is human-readable. (`getSAN` is unchanged by this PR, but the new Mode-B validation now depends on its output form.)

<sub>flagged by superpowers; verified against BouncyCastle behavior</sub>

src/main/java/com/otilm/core/certificate/request/CertificateRequestContentValidator.java
```diff
@@ -0,0 +1,270 @@
...
+                                                     Set<String> mappedRdnKeys,
+                                                     Set<GeneralNameType> mappedSanTypes,
+                                                     Set<String> mappedExtensionOids,
+                                                     Map<String, String> codeToOid) {
+        return switch (field) {
+            case RdnMappedField rdn -> collectMatched(
+                    canonicalRdnKey(rdn.getRdn(), codeToOid), mappedRdnKeys, subject,
+                    (key, e) -> key.equalsIgnoreCase(canonicalRdnKey(e.getType(), codeToOid)), RdnEntry::getValue);
+            case SanMappedField san -> collectMatched(
+                    san.getGeneralNameType(), mappedSanTypes, sans,
```
lubomirw, line 198: **[Medium]** | Security

otherName SANs are matched (here) and whitelisted (`checkWhitelist`, ~L224) by `GeneralNameType` alone — `SanMappedField.getOtherNameOid()` is never read. So a required otherName mapped to OID A is satisfied by a CSR otherName of any OID, and once any otherName mapping exists the whitelist accepts every otherName regardless of OID → policy bypass (e.g. a UPN-only policy accepts an arbitrary otherName type).

_Suggested:_ For OTHER_NAME `SanMappedField`, match and whitelist on the (`GeneralNameType`, `otherNameOid`) pair; track mapped otherName OIDs separately. Add a test where the policy maps OID A, the CSR carries OID B, and must be rejected.

<sub>flagged by superpowers; verified against code</sub>

src/main/java/com/otilm/core/certificate/request/CertificateRequestContentValidator.java
```diff
@@ -0,0 +1,270 @@
...
+                continue;
+            }
+            boolean required = v3.getProperties() != null && v3.getProperties().isRequired();
+
+            for (MappedField field : v3.getFieldMapping().getFields()) {
+                List<String> matchedValues = collectMatchedValues(field, subject, sans, extensions,
+                        mappedRdnKeys, mappedSanTypes, mappedExtensionOids, codeToOid);
+
+                if (required && matchedValues.isEmpty()) {
+                    recordViolation(result, policy, "Missing required mapped field for attribute '%s' (%s)"
```
lubomirw, line 126: **[Medium]** | Correctness

The required-attribute check records a "Missing required mapped field" violation **per field** inside the loop over `getFieldMapping().getFields()`. `FieldMapping.fields` is documented 1-to-many ("CN + dNSName from a single attribute"), so a required attribute whose value is present in one target (e.g. CN) but absent in a sibling target (e.g. dNSName) is wrongly rejected in strict mode.

_Suggested:_ Evaluate "required" at the attribute level — record a violation only when no mapped field for the attribute has a non-empty match (at-least-one), rather than per field.

<sub>flagged by Copilot CLI; verified — 1-to-many mapping confirmed in the interface schema</sub>

src/main/java/com/otilm/core/certificate/request/X509RequestContentParser.java
```diff
@@ -0,0 +1,267 @@
...
+    private static List<GeneralNameEntry> toGeneralNameEntries(String sanTypeKey, List<String> values) {
+        if (values == null || values.isEmpty()) {
+            return List.of();
+        }
+        if ("otherName".equals(sanTypeKey)) {
+            return toOtherNameEntries(values);
+        }
+        GeneralNameType type;
+        try {
+            type = GeneralNameType.fromCode(mapSanKeyToGeneralNameTypeCode(sanTypeKey));
```
lubomirw: **[Medium]** | Security

SAN kinds the parser can't model are silently dropped and thus invisible to the whitelist (fail-open): `toGeneralNameEntries` catches `IllegalArgumentException` from `GeneralNameType.fromCode(...)` and returns `List.of()` for kinds with no enum counterpart (x400Address, ediPartyName), and `CertificateUtil.getSAN` swallows per-SAN parse errors with only a warning. Since the CSR is forwarded verbatim, a dropped SAN reaches the CA unchecked against a closed-set whitelist.

_Suggested:_ Under strict/whitelist policy, treat any SAN the parser can't represent (or that fails to parse) as a violation instead of dropping it — surface a count of unrepresented SANs so the validator can fail closed. (Lower urgency: those kinds are rarely honored by CAs.)

<sub>flagged by superpowers; verified against code</sub>

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1764,12 +1783,89 @@ private static String normalizeRdn(String rdn, Map<String, String> codeToOid) {
...
+                    raProfile.getName(), e);
+            return;
+        }
+        if (definitions == null || definitions.isEmpty()) {
+            return;
+        }
+        boolean strict = resolveExternalCsrValidationStrict(raProfile);
+        try {
+            CertificateRequest parsed = CertificateRequestUtils.createCertificateRequest(csr, requestFormat);
+            X509RequestContent content = X509RequestContentParser.parse(parsed);
```
lubomirw: **[Low]** | Error handling

The `try` wrapping the parse + `X509RequestContentParser.parse` catches only `CertificateRequestException`/`CertificateRequestValidationException`. Most malformed-input errors are already shaped to 4xx by the global `@RestControllerAdvice` (`IllegalArgumentException`→400), but a few unchecked paths — `ArrayIndexOutOfBoundsException` from empty `extensionRequest` attrValues, or a null CRMF `CertTemplate` NPE — fall through to HTTP 500 for what is really bad client input.

_Suggested:_ Catch `RuntimeException` around parse+validate and rethrow as the client-facing 400/422, so malformed uploads never surface as 500.

<sub>flagged by guided-pr-review; verified — residual is narrow (global advice handles the common cases)</sub>

src/main/java/com/otilm/core/certificate/request/CertificateRequestContentValidator.java
```diff
@@ -0,0 +1,270 @@
...
+ * Validates a parsed {@link X509RequestContent} against the resolved request-attribute definitions.
+ */
+@Component
+public class CertificateRequestContentValidator {
+
+    /**
+     * Parse the supplied CSR, validate against the resolved definitions, and throw on a strict-policy
+     * failure. Whitelist enforcement is driven through {@link #validate(List, X509RequestContent, RequestAttributePolicy)}.
+     */
+    public void validate(CertificateRequest request, List<? extends BaseAttribute> definitions, boolean lenient)
```
lubomirw, line 47: **[Low]** | Maintainability

This `@Component` instance `validate(CertificateRequest, definitions, boolean)` is called only by tests; production (`ClientOperationServiceImpl.validateUploadedRequestAttributes`) calls the **static** overload and re-implements parse + strict/lenient result shaping inline. So the unit-tested convenience path isn't the one that ships (the two already differ in policy construction and logging), and this bean is never injected.

_Suggested:_ Either route the service through the instance method (dropping the inline duplication) or delete the instance method + `@Component` and keep only the static kernel; align the tests with whichever path ships.

<sub>flagged by guided-pr-review + superpowers; verified against code</sub>

src/test/java/com/otilm/core/certificate/request/X509RequestContentParserTest.java
```diff
@@ -0,0 +1,343 @@
...
+
+            // when
+            X509RequestContent content = X509RequestContentParser.parse(request);
+
+            // then
+            assertThat(content.getSubject()).isEmpty();
+        }
+
+        @Test
+        void splitsMultiValuedRdn_intoSeparateEntries() throws Exception {
```
lubomirw, line 104: **[Low]** | Testing

Two real coverage gaps on the highest-risk paths: (1) no test exercises RDN values containing `,` `+` `=` `\` — the `splitOnUnescaped`/`unescape` logic behind finding #1 (this `splitsMultiValuedRdn` test uses plain values only); (2) the only CRMF parser test carries no SANs and no extensions, so CRMF SAN/extension extraction is unverified.

_Suggested:_ Add parser tests for special-character RDN values and for a CRMF request carrying SANs + extensions. (The originally-suggested iPAddress-decode and otherName-OID-discrimination tests are moot — that behavior isn't implemented.)

<sub>flagged by superpowers; verified — 2 of 4 claimed gaps are real</sub>

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1764,12 +1783,89 @@ private static String normalizeRdn(String rdn, Map<String, String> codeToOid) {
...
+     */
+    private void validateUploadedRequestAttributes(String csr, CertificateRequestFormat requestFormat, RaProfile raProfile)
+            throws CertificateException {
+        if (raProfile == null) {
+            return;
+        }
+        List<BaseAttribute> definitions;
+        try {
+            definitions = requestAttributeService.resolveIssueAttributeSet(raProfile);
+        } catch (ConnectorException | NotFoundException e) {
```
lubomirw, line 1799: **[Low]** | Error handling

`resolveIssueAttributeSet` is wrapped in `catch (ConnectorException | NotFoundException)`, which correctly covers connector availability. Other unchecked failures from the resolve path escape the strict/lenient gate; a `ValidationException` (corrupt stored binding) is still shaped to 422, but a bare `RuntimeException` surfaces as an unshaped 500.

_Suggested:_ **Keep the narrow catch** — it's intentional fail-closed design; widening it to catch-all would mask genuine data/logic bugs as "connector unavailable." Just ensure an unexpected `RuntimeException` is mapped to a shaped error rather than a raw 500.

<sub>flagged by Copilot CLI; verified — narrow catch is deliberate, fix reframed</sub>

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1764,12 +1783,89 @@ private static String normalizeRdn(String rdn, Map<String, String> codeToOid) {
...
+            logger.warn("Could not resolve request-attribute set for Mode B validation (RA profile {}); lenient validation skipped",
+                    raProfile.getName(), e);
+            return;
+        }
+        if (definitions == null || definitions.isEmpty()) {
+            return;
+        }
+        boolean strict = resolveExternalCsrValidationStrict(raProfile);
+        try {
+            CertificateRequest parsed = CertificateRequestUtils.createCertificateRequest(csr, requestFormat);
```
lubomirw: **[Nit]** | Performance

On the validated uploaded-CSR path the CSR is ASN.1-parsed twice — here in `validateUploadedRequestAttributes`, and again in the base64 re-encode block further down. Negligible (issuance isn't a high-QPS path; a single parse is sub-millisecond), and there's already an in-code `TODO` to thread a `CertificateRequest` object through.

_Suggested:_ When you act on that TODO, parse once and reuse the `CertificateRequest` for both validation and re-encoding.

<sub>flagged by guided-pr-review + Copilot CLI + superpowers; verified — real but negligible</sub>

src/main/java/com/otilm/core/certificate/request/CertificateRequestContentValidator.java
```diff
@@ -0,0 +1,270 @@
...
+                && mapping.getObjectType() == ObjectType.X509_CERTIFICATE
+                && mapping.getFields() != null;
+    }
+
+    /**
+     * Canonicalizes an RDN identifier to its OID form for matching: a dotted OID is returned as-is,
+     * a known short code is resolved via {@code codeToOid}, and an unknown code falls back to itself
+     * (so code-vs-code comparison still works when the OID registry is unseeded).
+     */
+    private static String canonicalRdnKey(String rdn, Map<String, String> codeToOid) {
```
lubomirw, line 155: **[Nit]** | Performance

`canonicalRdnKey` linearly scans `codeToOid` with `equalsIgnoreCase` per RDN, and `checkWhitelist` streams `mappedRdnKeys` similarly. `codeToOid` is a small fixed RDN-type registry and RDN counts are small, so this is not a realistic performance/DoS concern.

_Suggested:_ Optional cleanup — build a case-insensitive code→OID lookup once and canonicalize whitelist keys into a `Set` for O(1) membership.

<sub>flagged by superpowers; verified — negligible</sub>

src/main/java/com/otilm/core/oid/OidHandler.java
```diff
@@ -5,12 +5,21 @@
...
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Pattern;
 
 public class OidHandler {
 
     private OidHandler() {
     }
 
+    /** Dotted-decimal OID form (e.g. {@code 2.5.4.3}); anything else is treated as a short RDN code. */
+    public static final Pattern OID_PATTERN = Pattern.compile("^[0-2](\\.(0|[1-9]\\d{0,38})){1,127}$");
```
lubomirw: **[Nit]** | Maintainability

`OID_PATTERN` is declared `public static final`, but every consumer uses `isOid(...)` — no external caller references the raw pattern.

_Suggested:_ Make `OID_PATTERN` `private static final`; expose only `isOid(String)`.

<sub>flagged by superpowers; verified — no external caller</sub>

src/main/java/com/otilm/core/certificate/request/X509RequestContentParser.java
ivosh: Fixed in e80e1cdf. `parseSubject` now walks `subject.getRDNs()`/`AttributeTypeAndValue` directly — no rendered-DN round-trip; the split/unescape helpers are gone. RFC 4514 special-character values are pinned by a new parser test.

src/main/java/com/otilm/core/certificate/request/X509RequestContentParser.java
ivosh: Fixed in e80e1cdf. The parser no longer routes SANs through `CertificateUtil.getSAN`; it extracts `GeneralNames` per tag and decodes iPAddress octets via `InetAddress` to dotted/colon form. IPv4 + IPv6 decode tests added; a non-4/16-octet address now fails closed instead of passing through in hex.

src/main/java/com/otilm/core/certificate/request/CertificateRequestContentValidator.java
ivosh: Fixed in e80e1cdf. otherName mappings now match and whitelist on the (`GeneralNameType`, `otherNameOid`) pair via a separate mapped-OID set, so a UPN-only policy no longer accepts arbitrary otherName types. Added the policy-maps-OID-A / CSR-carries-OID-B rejection test.

src/main/java/com/otilm/core/certificate/request/CertificateRequestContentValidator.java
ivosh: Keeping the per-field semantics — it's the pinned design: Mode B is the reverse of the Mode A projection, and a required attribute mapped 1-to-many (e.g. FQDN → CN + dNSName) projects into *all* of its targets on the build path, so a conforming CSR must carry all of them. A CSR with the CN but missing the sibling dNSName deviates from the mapping and strict mode should say so. Added a test in e80e1cdf pinning this as intentional.

src/main/java/com/otilm/core/certificate/request/X509RequestContentParser.java
ivosh: Fixed in e80e1cdf. The parser now reports every SAN it cannot represent or decode (x400Address, ediPartyName, undecodable entries) via `ParsedRequestContent.unsupportedSans`, and the whitelist pass records a violation for each — strict mode fails closed while lenient behavior is unchanged.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
ivosh: Fixed in e80e1cdf. Parse+validate is now guarded by `catch (RuntimeException)` rethrowing a shaped 422 `ValidationException` with a platform-authored message (original exception logged server-side, nothing leaks to the wire). The validator's instance entry point got the same guard so protocol adapters inherit it. Integration test with a hand-assembled empty-extensionRequest PKCS#10 covers the AIOOBE path end-to-end.

src/main/java/com/otilm/core/certificate/request/CertificateRequestContentValidator.java
ivosh: Keeping the instance overload — it isn't dead code: it's the designated entry point the Phase-5 protocol adapters (CMP/SCEP/ACME/EST) consume per the implementation program, and the unit tests hold that contract in place until Phase 5 lands. The REST issue flow stays on the static kernel because it additionally logs lenient warnings and shapes the REST error. The overload now also gains the malformed-CSR guard in e80e1cdf, so future protocol callers inherit the no-leak behavior.

src/test/java/com/otilm/core/certificate/request/X509RequestContentParserTest.java
ivosh: Fixed in e80e1cdf. Added: RDN values with unescaped `,` `+` `=` (pinning the new direct ASN.1 walk), a CRMF request carrying SANs + extensions, plus IPv4/IPv6 iPAddress decode and otherName OID-discrimination tests — the latter two are no longer moot since that behavior is now implemented.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
ivosh: No change, as suggested — the narrow catch stays. The residual concern is already covered: an unexpected RuntimeException from the resolve path reaches the global `@ExceptionHandler(Exception.class)`, which returns a shaped `ErrorMessageDto("Internal server error.")` — nothing unshaped or leaking, and 500 is the honest status for a genuine server-side bug (vs. masking it as connector unavailability).

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
ivosh: Agreed, deferred to the existing TODO — when the `CertificateRequest` object is threaded through, it will be parsed once and reused for both validation and re-encoding.

src/main/java/com/otilm/core/certificate/request/CertificateRequestContentValidator.java
ivosh: Fixed in e80e1cdf. `codeToOid` is wrapped once per `validate()` in a case-insensitive `TreeMap` and mapped RDN keys live in a case-insensitive `TreeSet`, so `canonicalRdnKey` and the whitelist check are O(1) lookups instead of linear scans.

src/main/java/com/otilm/core/oid/OidHandler.java
ivosh: Fixed in e80e1cdf. `OID_PATTERN` is now private; `isOid(String)` remains the only public surface.

lubomirw (approved):

PR #1699, opened 2026-07-01
Implement v3 authority instance attribute listing endpoint

lubomirw:
## What

Implements the operator endpoint `GET /v1/authorities/{connectorUuid}/attributes` (contract in OmniTrustILM/interfaces#753) that lists a stateless **v3 authority connector**'s attribute schema for the create-authority flow, mirroring `VaultInstanceServiceImpl.listVaultInstanceAttributes`.

- Routes through `AuthorityProviderV3Adapter.listAuthorityInstanceAttributes` → the connector's kind-less `GET /v3/authorityProvider/authorities/attributes`, then persists the definitions via `AttributeEngine.updateDataAttributeDefinitions` (same as the Vault flow).
- **v3-only:** a non-v3 (or ambiguous-interface) connector resolves to a `ValidationException` (422); v1/v2 connectors continue to use the function-group connector-attributes endpoint.
- Service takes a plain `UUID connectorUuid` with type-level `@ExternalAuthorization(AUTHORITY, ANY)` — **no per-connector object-access evaluation**, matching the Vault service. Both path/query UUIDs parse via `UUID.fromString` (consistent 400 on malformed input).

## Why

Exposes the already-present v3 adapter method as an operator endpoint so the FE (and API clients) can render the create-authority form for stateless v3 connectors. Part of the Authority Provider v3 epic (OmniTrustILM/ilm#110). FE wiring is tracked in OmniTrustILM/fe-administrator#1796.

## Coordination

⚠️ **Paired with OmniTrustILM/interfaces#753 — merge / version-bump together**, or this fails to compile against a new `interfaces` release.

## Testing

`mvn compile` — BUILD SUCCESS. (Full `verify` not run here.)

### DIFF
--- a/src/main/java/com/otilm/core/api/web/AuthorityInstanceControllerImpl.java
+++ b/src/main/java/com/otilm/core/api/web/AuthorityInstanceControllerImpl.java
@@ -23,11 +23,13 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.servlet.support.ServletUriComponentsBuilder;
 
 import java.net.URI;
 import java.util.List;
+import java.util.UUID;
 
 @RestController
 public class AuthorityInstanceControllerImpl implements AuthorityInstanceController {
@@ -90,6 +92,13 @@ public class AuthorityInstanceControllerImpl implements AuthorityInstanceControl
     @AuditLogged(module = Module.CERTIFICATES, resource = Resource.END_ENTITY_PROFILE, affiliatedResource = Resource.AUTHORITY, operation = Operation.LIST_CAS)
     public List<NameAndIdDto> listCAsInProfile(@LogResource(uuid = true, affiliated = true) @PathVariable String uuid, @PathVariable Integer endEntityProfileId) throws ConnectorException, NotFoundException {
         return authorityInstanceService.listCAsInProfile(SecuredUUID.fromString(uuid), endEntityProfileId);
+    }
+
+    @Override
+    @AuditLogged(module = Module.CERTIFICATES, resource = Resource.ATTRIBUTE, name = "authority", affiliatedResource = Resource.CONNECTOR, operation = Operation.LIST_ATTRIBUTES)
+    public List<BaseAttribute> listAuthorityInstanceAttributes(@LogResource(uuid = true, affiliated = true) @PathVariable String connectorUuid, @RequestParam(required = false) String interfaceUuid) throws ConnectorException, NotFoundException, AttributeException {
+        return authorityInstanceService.listAuthorityInstanceAttributes(SecuredUUID.fromString(connectorUuid),
+                interfaceUuid != null ? UUID.fromString(interfaceUuid) : null);
     }
 
     @Override
--- a/src/main/java/com/otilm/core/api/web/AuthorityInstanceControllerImpl.java
+++ b/src/main/java/com/otilm/core/api/web/AuthorityInstanceControllerImpl.java
@@ -95,7 +95,7 @@ public class AuthorityInstanceControllerImpl implements AuthorityInstanceControl
 
     @Override
     @AuditLogged(module = Module.CERTIFICATES, resource = Resource.ATTRIBUTE, name = "authority", affiliatedResource = Resource.CONNECTOR, operation = Operation.LIST_ATTRIBUTES)
-    public List<BaseAttribute> listAuthorityInstanceAttributes(@LogResource(uuid = true, affiliated = true) @PathVariable String connectorUuid, @RequestParam(required = false) String interfaceUuid) throws ConnectorException, NotFoundException, AttributeException {
+    public List<BaseAttribute> listAuthorityInstanceAttributes(@LogResource(uuid = true, affiliated = true) @PathVariable("connectorUuid") String connectorUuid, @RequestParam(name = "interfaceUuid", required = false) String interfaceUuid) throws ConnectorException, NotFoundException, AttributeException {
         return authorityInstanceService.listAuthorityInstanceAttributes(SecuredUUID.fromString(connectorUuid),
                 interfaceUuid != null ? SecuredUUID.fromString(interfaceUuid).getValue() : null);
     }
--- a/src/main/java/com/otilm/core/api/web/AuthorityInstanceControllerImpl.java
+++ b/src/main/java/com/otilm/core/api/web/AuthorityInstanceControllerImpl.java
@@ -94,8 +94,8 @@ public class AuthorityInstanceControllerImpl implements AuthorityInstanceControl
     }
 
     @Override
-    @AuditLogged(module = Module.CERTIFICATES, resource = Resource.ATTRIBUTE, name = "authority", affiliatedResource = Resource.CONNECTOR, operation = Operation.LIST_ATTRIBUTES)
-    public List<BaseAttribute> listAuthorityInstanceAttributes(@LogResource(uuid = true, affiliated = true) @PathVariable("connectorUuid") String connectorUuid, @RequestParam(name = "interfaceUuid", required = false) String interfaceUuid) throws ConnectorException, NotFoundException, AttributeException {
+    @AuditLogged(module = Module.CERTIFICATES, resource = Resource.ATTRIBUTE, name = "authority", affiliatedResource = Resource.AUTHORITY, operation = Operation.LIST_ATTRIBUTES)
+    public List<BaseAttribute> listAuthorityInstanceAttributes(@PathVariable("connectorUuid") String connectorUuid, @RequestParam(name = "interfaceUuid", required = false) String interfaceUuid) throws ConnectorException, NotFoundException, AttributeException {
         return authorityInstanceService.listAuthorityInstanceAttributes(SecuredUUID.fromString(connectorUuid),
                 interfaceUuid != null ? SecuredUUID.fromString(interfaceUuid).getValue() : null);
     }
--- a/src/main/java/com/otilm/core/api/web/AuthorityInstanceControllerImpl.java
+++ b/src/main/java/com/otilm/core/api/web/AuthorityInstanceControllerImpl.java
@@ -97,7 +97,7 @@ public class AuthorityInstanceControllerImpl implements AuthorityInstanceControl
     @Override
     @AuditLogged(module = Module.CERTIFICATES, resource = Resource.ATTRIBUTE, name = "authority", affiliatedResource = Resource.CONNECTOR, operation = Operation.LIST_ATTRIBUTES)
     public List<BaseAttribute> listAuthorityInstanceAttributes(@LogResource(uuid = true, affiliated = true) @PathVariable String connectorUuid, @RequestParam(required = false) String interfaceUuid) throws ConnectorException, NotFoundException, AttributeException {
-        return authorityInstanceService.listAuthorityInstanceAttributes(SecuredUUID.fromString(connectorUuid),
+        return authorityInstanceService.listAuthorityInstanceAttributes(UUID.fromString(connectorUuid),
                 interfaceUuid != null ? UUID.fromString(interfaceUuid) : null);
     }
 
--- a/src/main/java/com/otilm/core/api/web/AuthorityInstanceControllerImpl.java
+++ b/src/main/java/com/otilm/core/api/web/AuthorityInstanceControllerImpl.java
@@ -29,7 +29,6 @@ import org.springframework.web.servlet.support.ServletUriComponentsBuilder;
 
 import java.net.URI;
 import java.util.List;
-import java.util.UUID;
 
 @RestController
 public class AuthorityInstanceControllerImpl implements AuthorityInstanceController {
@@ -97,8 +96,8 @@ public class AuthorityInstanceControllerImpl implements AuthorityInstanceControl
     @Override
     @AuditLogged(module = Module.CERTIFICATES, resource = Resource.ATTRIBUTE, name = "authority", affiliatedResource = Resource.CONNECTOR, operation = Operation.LIST_ATTRIBUTES)
     public List<BaseAttribute> listAuthorityInstanceAttributes(@LogResource(uuid = true, affiliated = true) @PathVariable String connectorUuid, @RequestParam(required = false) String interfaceUuid) throws ConnectorException, NotFoundException, AttributeException {
-        return authorityInstanceService.listAuthorityInstanceAttributes(UUID.fromString(connectorUuid),
-                interfaceUuid != null ? UUID.fromString(interfaceUuid) : null);
+        return authorityInstanceService.listAuthorityInstanceAttributes(SecuredUUID.fromString(connectorUuid),
+                interfaceUuid != null ? SecuredUUID.fromString(interfaceUuid).getValue() : null);
     }
 
     @Override
--- a/src/main/java/com/otilm/core/service/AuthorityInstanceExternalService.java
+++ b/src/main/java/com/otilm/core/service/AuthorityInstanceExternalService.java
@@ -31,7 +31,14 @@ public interface AuthorityInstanceExternalService {
 
     List<NameAndIdDto> listCAsInProfile(SecuredUUID uuid, Integer endEntityProfileId) throws ConnectorException, NotFoundException;
 
-    List<BaseAttribute> listAuthorityInstanceAttributes(UUID connectorUuid, UUID interfaceUuid) throws ConnectorException, AttributeException, NotFoundException;
+    /**
+     * Lists the authority attribute schema for a stateless v3 authority connector (used by the
+     * create-authority form), keyed by connector UUID. v3-only: throws ValidationException when the
+     * connector has no AUTHORITY interface or a non-v3 one. {@code interfaceUuid} is optional and only
+     * disambiguates a connector that exposes more than one AUTHORITY interface. Side effect: persists
+     * the returned definitions via the attribute engine for later validation and content preparation.
+     */
+    List<BaseAttribute> listAuthorityInstanceAttributes(SecuredUUID connectorUuid, UUID interfaceUuid) throws ConnectorException, AttributeException, NotFoundException;
 
     List<BaseAttribute> listRAProfileAttributes(SecuredUUID uuid) throws ConnectorException, NotFoundException;
 
--- a/src/main/java/com/otilm/core/service/AuthorityInstanceExternalService.java
+++ b/src/main/java/com/otilm/core/service/AuthorityInstanceExternalService.java
@@ -12,6 +12,7 @@ import com.otilm.core.security.authz.SecuredUUID;
 import com.otilm.core.security.authz.SecurityFilter;
 
 import java.util.List;
+import java.util.UUID;
 
 public interface AuthorityInstanceExternalService {
     List<AuthorityInstanceDto> listAuthorityInstances(SecurityFilter filter);
@@ -29,6 +30,8 @@ public interface AuthorityInstanceExternalService {
     List<NameAndIdDto> listCertificateProfiles(SecuredUUID uuid, Integer endEntityProfileId) throws ConnectorException, NotFoundException;
 
     List<NameAndIdDto> listCAsInProfile(SecuredUUID uuid, Integer endEntityProfileId) throws ConnectorException, NotFoundException;
+
+    List<BaseAttribute> listAuthorityInstanceAttributes(SecuredUUID connectorUuid, UUID interfaceUuid) throws ConnectorException, AttributeException, NotFoundException;
 
     List<BaseAttribute> listRAProfileAttributes(SecuredUUID uuid) throws ConnectorException, NotFoundException;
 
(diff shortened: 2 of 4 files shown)


src/main/java/com/otilm/core/api/web/AuthorityInstanceControllerImpl.java
```diff
@@ -92,6 +94,13 @@ public List<NameAndIdDto> listCAsInProfile(@LogResource(uuid = true, affiliated
         return authorityInstanceService.listCAsInProfile(SecuredUUID.fromString(uuid), endEntityProfileId);
     }
 
+    @Override
+    @AuditLogged(module = Module.CERTIFICATES, resource = Resource.ATTRIBUTE, name = "authority", affiliatedResource = Resource.CONNECTOR, operation = Operation.LIST_ATTRIBUTES)
+    public List<BaseAttribute> listAuthorityInstanceAttributes(@LogResource(uuid = true, affiliated = true) @PathVariable String connectorUuid, @RequestParam(required = false) String interfaceUuid) throws ConnectorException, NotFoundException, AttributeException {
+        return authorityInstanceService.listAuthorityInstanceAttributes(UUID.fromString(connectorUuid),
+                interfaceUuid != null ? UUID.fromString(interfaceUuid) : null);
+    }
```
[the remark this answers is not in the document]
lubomirw: By design. This endpoint mirrors the stateless Vault flow (`VaultInstanceController`), whose typed `UUID` path var also yields 400, and the service takes a plain `UUID` (no connector object-access evaluation). Keeping 400 here rather than the controller-family 422 to preserve the Vault-mirror semantics.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -312,6 +312,29 @@ public List<NameAndIdDto> listCAsInProfile(SecuredUUID uuid, Integer endEntityPr
...
+            // v1/v2 connectors resolve their attributes by function group + kind through the connector
+            // attributes endpoint; direct callers there rather than guess a kind here.
+            throw new ValidationException(
+                    "Connector " + connectorUuid + " has no stateless (v3) AUTHORITY interface; "
+                            + "list attributes for v1/v2 authority connectors via the connector attributes endpoint (function group and kind).");
+        }
+        // v3 is stateless and kind-less: definitions come from GET /v3/authorityProvider/authorities/attributes.
+        AuthorityInstanceReference probeRef = transientAuthorityRef(securedConnectorUuid, connector, iface, null);
+        List<BaseAttribute> attributes = adapterFactory.forAuthority(probeRef).listAuthorityInstanceAttributes(probeRef);
+        // Persist the definitions so later validation and content preparation can resolve them, mirroring
```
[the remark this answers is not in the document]
lubomirw: Leaving as-is. `getConnector(...)` is kept for its 404-on-missing-connector semantics ahead of the v3 gate (without it a missing connector would fall through to the 422 non-v3 path). The adapter's `getConnectorForApiClient(...)` is the shared resolution path every v3 op uses (`listRAProfileAttributes`, cert ops); bypassing the adapter to save one lookup on a create-form load path would break that abstraction.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -312,6 +312,29 @@ public List<NameAndIdDto> listCAsInProfile(SecuredUUID uuid, Integer endEntityPr
...
+        // VaultInstanceServiceImpl.listVaultInstanceAttributes for the stateless secret-provider flow.
+        attributeEngine.updateDataAttributeDefinitions(connectorUuid, null, attributes);
+        return attributes;
+    }
```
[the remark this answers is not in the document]
lubomirw: Added in 5a786aa82 — `AuthorityInstanceServiceImplV3Test` now covers `listAuthorityInstanceAttributes`: v3 lists + persists definitions, non-v3 → `ValidationException`, multiple AUTHORITY interfaces require `interfaceUuid`, and explicit `interfaceUuid` selection.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -312,6 +312,29 @@ public List<NameAndIdDto> listCAsInProfile(SecuredUUID uuid, Integer endEntityPr
...
+            // Only stateless v3 authority connectors list attributes without a bound instance. Legacy
+            // v1/v2 connectors resolve their attributes by function group + kind through the connector
+            // attributes endpoint; direct callers there rather than guess a kind here.
+            throw new ValidationException(
+                    "Connector " + connectorUuid + " has no stateless (v3) AUTHORITY interface; "
+                            + "list attributes for v1/v2 authority connectors via the connector attributes endpoint (function group and kind).");
+        }
+        // v3 is stateless and kind-less: definitions come from GET /v3/authorityProvider/authorities/attributes.
+        AuthorityInstanceReference probeRef = transientAuthorityRef(securedConnectorUuid, connector, iface, null);
+        List<BaseAttribute> attributes = adapterFactory.forAuthority(probeRef).listAuthorityInstanceAttributes(probeRef);
```
ivosh, line 343: **~ Minor** | Performance

AuthorityInstanceServiceImpl.java:331 makes the external connector HTTP call `adapterFactory.forAuthority(probeRef).listAuthorityInstanceAttributes(probeRef)` while the ambient transaction from the class-level `@Transactional` (AuthorityInstanceServiceImpl.java:55) is open, then writes with `attributeEngine.updateDataAttributeDefinitions(...)` at line 334 still inside that same tx. This is the anti-pattern called out in CLAUDE.md 'Transactions and external calls': a slow/unresponsive connector holds a DB connection (and any locks) for its full response time. 

Note that a neighbour pattern behaves similarly: the Vault sibling `VaultInstanceServiceImpl.listVaultInstanceAttributes` (VaultInstanceServiceImpl.java:214, also class-level `@Transactional`) and the peer `listRAProfileAttributes` right below (line ~356) do exactly the same. Split so the HTTP call runs outside the tx and only `updateDataAttributeDefinitions` is wrapped (and do the same for the sibling in a follow-up).

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -312,6 +312,29 @@ public List<NameAndIdDto> listCAsInProfile(SecuredUUID uuid, Integer endEntityPr
...
+    @ExternalAuthorization(resource = Resource.AUTHORITY, action = ResourceAction.ANY)
+    public List<BaseAttribute> listAuthorityInstanceAttributes(UUID connectorUuid, UUID interfaceUuid) throws ConnectorException, AttributeException, NotFoundException {
+        SecuredUUID securedConnectorUuid = SecuredUUID.fromUUID(connectorUuid);
+        ConnectorDto connector = connectorService.getConnector(securedConnectorUuid);
+        ConnectorInterfaceEntity iface = resolveAuthorityInterface(connectorUuid, interfaceUuid);
+        if (!isV3(iface)) {
+            // Only stateless v3 authority connectors list attributes without a bound instance. Legacy
+            // v1/v2 connectors resolve their attributes by function group + kind through the connector
+            // attributes endpoint; direct callers there rather than guess a kind here.
+            throw new ValidationException(
```
ivosh, line 337: **. Nitpick** | Error handling

AuthorityInstanceServiceImpl.java:320-327: `resolveAuthorityInterface` returns null when the connector exposes zero AUTHORITY interfaces (findFirst().orElse(null) at line 515), and `isV3(null)` is false, so any connector that is not an authority connector at all falls into the ValidationException at line 325 whose text tells the caller to 'list attributes for v1/v2 authority connectors via the connector attributes endpoint (function group and kind)'. For a connector with no AUTHORITY function group that advice is misleading. Behaviourally it still returns a 422 (via ExceptionHandlingAdvice), so this is cosmetic, not a defect. Consider distinguishing 'connector has no AUTHORITY interface' from 'AUTHORITY interface is not v3' so the message is actionable. Non-blocking.

src/test/java/com/otilm/core/service/impl/AuthorityInstanceServiceImplV3Test.java
```diff
@@ -333,6 +333,66 @@ void validateRAProfileAttributesToleratesNullAttributesForV3() throws Exception
         assertThat(service.validateRAProfileAttributes(SecuredUUID.fromUUID(existing.uuid), null)).isTrue();
     }
 
+    @Test
+    void listAuthorityInstanceAttributesListsAndPersistsForV3() throws Exception {
```
ivosh, line 342: **~ Minor** | Documentation

AuthorityInstanceServiceImplV3Test.java:49-56 the class Javadoc states "Unit tests for the v3-specific branch in AuthorityInstanceServiceImpl#createAuthorityInstance" and elaborates only on the create/checkAuthorityConnection path. 

This PR adds four tests for a different method, listAuthorityInstanceAttributes (lines 337-394), and the class already also tests validateRAProfileAttributes — so the Javadoc's scope claim is doc-vs-code drift: false for the class as it now stands. Non-blocking. Broaden the class comment to say it covers the v3-specific branches across the service (create, validateRAProfile, listAuthorityInstanceAttributes), or drop the method-specific sentence so a future reader isn't misled about what is verified here.

src/test/java/com/otilm/core/service/impl/AuthorityInstanceServiceImplV3Test.java
```diff
@@ -333,6 +333,66 @@ void validateRAProfileAttributesToleratesNullAttributesForV3() throws Exception
...
+        secondIface.setVersion("v2");
+        secondIface.setUuid(UUID.randomUUID());
+        secondIface.setConnectorUuid(connectorUuid);
+        connectorEntity.getInterfaces().add(secondIface);
+        List<BaseAttribute> definitions = List.of(mock(BaseAttribute.class));
+        when(v3Adapter.listAuthorityInstanceAttributes(any())).thenReturn(definitions);
+
+        List<BaseAttribute> result = service.listAuthorityInstanceAttributes(connectorUuid, ifaceUuid);
+
+        assertThat(result).isSameAs(definitions);
```
ivosh, line 412: **. Nitpick** | Testing

AuthorityInstanceServiceImplV3Test.java:378-394 listAuthorityInstanceAttributesWithExplicitInterfaceUuidSelectsThatInterface only asserts result.isSameAs(definitions). The name promises that the *explicit* interfaceUuid was the one bound, but that is only proven indirectly (selecting the v2 interface would make isV3 false and throw). The sibling test createV3AuthorityWithExplicitInterfaceUuidSelectsThatInterface at line 205-227 in this same class does it properly: it captures the AuthorityInstanceReference via ArgumentCaptor and asserts getConnectorInterface().isSameAs(v3Iface).

Mirror that captor pattern here (capture the ref passed to adapterFactory.forAuthority / v3Adapter.listAuthorityInstanceAttributes and assert its interface is the v3 one) and also verify(attributeEngine).updateDataAttributeDefinitions(...), so the test verifies what its name claims rather than a coincidence of ordering.

src/main/java/com/otilm/core/service/AuthorityInstanceExternalService.java
```diff
@@ -30,6 +31,8 @@ public interface AuthorityInstanceExternalService {
 
     List<NameAndIdDto> listCAsInProfile(SecuredUUID uuid, Integer endEntityProfileId) throws ConnectorException, NotFoundException;
 
+    List<BaseAttribute> listAuthorityInstanceAttributes(UUID connectorUuid, UUID interfaceUuid) throws ConnectorException, AttributeException, NotFoundException;
```
ivosh: **. Nitpick** | Maintainability

AuthorityInstanceExternalService.java:34 listAuthorityInstanceAttributes(UUID, UUID) has no Javadoc, yet its behaviour is not obvious from the name: it is v3-only (throws ValidationException for v1/v2 or for an ambiguous connector), the interfaceUuid arg is nullable and only needed to disambiguate multiple AUTHORITY interfaces, and — despite being a "list" method — it writes, calling attributeEngine.updateDataAttributeDefinitions as a side effect (impl AuthorityInstanceServiceImpl.java:334). The sibling VaultInstanceServiceImpl.listVaultInstanceAttributes has the same list-that-persists surprise and is likewise undocumented (only a TODO), so this matches the neighbour pattern rather than regressing it.

 Per CLAUDE.md "Writing for future AI sessions" (method-level Javadoc on non-obvious behaviour), a one-line Javadoc noting the v3-only contract and the definition-persistence side effect would help. Non-blocking; the interfaces-repo @Operation/@Parameter docs already cover the HTTP contract.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -312,6 +312,29 @@ public List<NameAndIdDto> listCAsInProfile(SecuredUUID uuid, Integer endEntityPr
                 authorityInstanceRef.getAuthorityInstanceUuid(), endEntityProfileId);
     }
 
+    @Override
+    @ExternalAuthorization(resource = Resource.AUTHORITY, action = ResourceAction.ANY)
```
ivosh: **~ Minor** | Security

AuthorityInstanceServiceImpl.java:316 — listAuthorityInstanceAttributes is annotated `@ExternalAuthorization(resource = AUTHORITY, action = ANY)` with a plain UUID connectorUuid. ExternalMethodAuthorizationManager collects the objects to authorize only from `SecuredUUID` arguments, so a plain `UUID` sends OPA no object and the declared check degrades to a resource-type presence check — "caller holds some AUTHORITY permission", unscoped. It performs no per-object authorization, and it gates on the wrong resource: this endpoint is connector-scoped (keyed by connector UUID; no authority instance exists yet).

The only thing that actually scopes access to this connector is the incidental nested call `connectorService.getConnector(securedConnectorUuid)` at line 319 → `@ExternalAuthorization(CONNECTOR, DETAIL), SecuredUUID)`. So the effective gate is (any AUTHORITY permission) AND (CONNECTOR/DETAIL on this connector), with the object-level half being a side effect of getConnector rather than of this method's own annotation — fragile: a future refactor that fetches the connector differently silently drops the object-level check.

The true analog is the connector's own attribute-schema methods, which do it correctly: `ConnectorServiceImpl.getAttributes(SecuredUUID uuid, ...)` (line 207) and `getAllAttributesOfConnector(SecuredUUID uuid)` (line 251) use `@ExternalAuthorization(resource = CONNECTOR, action = ANY)` with a SecuredUUID, so OPA receives the connector object and the object-level check happens at the boundary.

Option A (matches the connector-attributes convention): type the parameter as SecuredUUID and gate on CONNECTOR:
`@ExternalAuthorization(resource = Resource.CONNECTOR, action = ResourceAction.ANY)`
`public List<BaseAttribute> listAuthorityInstanceAttributes(SecuredUUID connectorUuid, UUID interfaceUuid)`
getConnector then becomes redundant belt-and-braces rather than the sole gate.

 Option B (if the "authority operator" intent must be kept): use the documented nested form — `resource = AUTHORITY` with `parentResource = CONNECTOR`, `parentAction = DETAIL` and a `SecuredParentUUID connectorUuid` — so the connector object-level check is declared, not incidental.

Note VaultInstanceServiceImpl.listVaultInstanceAttributes (line 213) shares the identical shape (VAULT/ANY + plain UUID + incidental getConnector) — this PR copied that suboptimal pattern, so the fix should cover both (same PR, or a paired follow-up for Vault) rather than treating Vault as the template.

ivosh (changes requested):

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
lubomirw: Fixed in 725ca5210 (Option A) — now takes `SecuredUUID connectorUuid` + `@ExternalAuthorization(CONNECTOR, ANY)`, so the object-level connector check is declared (matching `ConnectorServiceImpl.getAttributes`). Malformed UUIDs are 422 again. `VaultInstanceServiceImpl` shares the shape — paired follow-up to come.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
lubomirw: Fixed in 725ca5210 — `@Transactional(NOT_SUPPORTED)`; the connector HTTP call runs outside any tx, the lazy interface read is wrapped in `TransactionHandler.runInTransaction`, and the definition write opens its own tx via `AttributeEngine`. Vault / `listRAProfileAttributes` share the pattern — paired follow-up to come.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
lubomirw: Fixed in 725ca5210 — "no AUTHORITY interface" (not an authority connector) and "v1/v2 AUTHORITY interface" (use the connector attributes endpoint) are now distinct messages; added a negative test for the no-interface case.

src/main/java/com/otilm/core/service/AuthorityInstanceExternalService.java
lubomirw: Fixed in 725ca5210 — added a Javadoc noting the v3-only contract, the nullable `interfaceUuid`, and the definition-persistence side effect.

src/test/java/com/otilm/core/service/impl/AuthorityInstanceServiceImplV3Test.java
lubomirw: Fixed in 725ca5210 — broadened the class Javadoc to the v3 branches it now covers (create/edit, validateRAProfile, listAuthorityInstanceAttributes).

src/test/java/com/otilm/core/service/impl/AuthorityInstanceServiceImplV3Test.java
lubomirw: Fixed in 725ca5210 — the test now captures the probe ref, asserts the bound interface is the v3 one, and verifies the definition persistence (mirrors the createV3 captor test).

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -313,25 +321,34 @@ public List<NameAndIdDto> listCAsInProfile(SecuredUUID uuid, Integer endEntityPr
     }
 
     @Override
-    @ExternalAuthorization(resource = Resource.AUTHORITY, action = ResourceAction.ANY)
-    public List<BaseAttribute> listAuthorityInstanceAttributes(UUID connectorUuid, UUID interfaceUuid) throws ConnectorException, AttributeException, NotFoundException {
-        SecuredUUID securedConnectorUuid = SecuredUUID.fromUUID(connectorUuid);
-        ConnectorDto connector = connectorService.getConnector(securedConnectorUuid);
-        ConnectorInterfaceEntity iface = resolveAuthorityInterface(connectorUuid, interfaceUuid);
+    // Connector-scoped: a SecuredUUID connectorUuid lets ExternalMethodAuthorizationManager perform the
```
ivosh: **~ Minor** | Documentation

Do not restate the obvious.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -313,25 +321,34 @@ public List<NameAndIdDto> listCAsInProfile(SecuredUUID uuid, Integer endEntityPr
...
+    // object-level CONNECTOR check (a plain UUID would degrade this to an unscoped resource-type check).
+    @ExternalAuthorization(resource = Resource.CONNECTOR, action = ResourceAction.ANY)
+    // NOT_SUPPORTED so the connector HTTP call below runs outside any DB transaction (core CLAUDE.md,
```
ivosh: **~ Minor** | Documentation

Do not restate the obvious.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -313,25 +321,34 @@ public List<NameAndIdDto> listCAsInProfile(SecuredUUID uuid, Integer endEntityPr
...
+                            + "list its attributes via the connector attributes endpoint (function group and kind).");
         }
         // v3 is stateless and kind-less: definitions come from GET /v3/authorityProvider/authorities/attributes.
-        AuthorityInstanceReference probeRef = transientAuthorityRef(securedConnectorUuid, connector, iface, null);
+        AuthorityInstanceReference probeRef = transientAuthorityRef(connectorUuid, connector, iface, null);
         List<BaseAttribute> attributes = adapterFactory.forAuthority(probeRef).listAuthorityInstanceAttributes(probeRef);
-        // Persist the definitions so later validation and content preparation can resolve them, mirroring
-        // VaultInstanceServiceImpl.listVaultInstanceAttributes for the stateless secret-provider flow.
-        attributeEngine.updateDataAttributeDefinitions(connectorUuid, null, attributes);
+        // Persist the definitions (AttributeEngine opens its own tx) so later validation and content
```
ivosh: **~ Minor** | Documentation

Do not restate the obvious. Referencing a sibling does not add any additional trust.

ivosh (changes requested):

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
lubomirw: Removed in 9834b1cd2.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
lubomirw: Removed in 9834b1cd2.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
lubomirw: Removed in 9834b1cd2.

ivosh (approved):

PR #1707, opened 2026-07-02
Complete CertificateStateMachine coverage for the remaining cert-state writers

### LINKED ISSUE #1683: Complete CertificateStateMachine coverage for the remaining cert-state writers
### Description

#1682 (M1) routed every certificate `setState` in `ClientOperationServiceImpl` through `CertificateStateMachine`, so operator-driven transitions are now gated, validated against the transition table, and audited consistently. Two certificate-state writes remain **outside** the SM, in other classes. Both are pre-existing v2 paths (not v3 functionality), so this is state-machine **coverage hardening** that can land independently of the authority-provider-v3 epic (ilm#110).

**Route these through the state machine:**

1. **`CertificateUtil.prepareIssuedCertificate(...)` → `ISSUED`** (issue completion). Today a static, repository-less stamp that also sets serial/validity. Routing it needs the certificate + repository and from-state handling (`REQUESTED`/`PENDING_ISSUE` → `ISSUED`). Bonus: both synchronous and async-poll issue completion funnel through this method, so routing it also brings async-issue completion under the SM.
2. **`CertificateServiceImpl.revokeCertificate(String serialNumber)` → `REVOKED`** (internal "mark revoked by serial"; protocol/internal use, no connector round-trip). It currently revokes whatever it finds; routing it needs a from-state guard, like the operator revoke paths got in #1682.

**Intentionally out of scope (by-design SM exemptions):** initial-state assignment on newly-created entities (`createRegistrationPlaceholder`, CSR submission); the no-op self-restore in the revoke connector-failure path; `X509CertificateValidator`'s out-of-band revoke (atomic conditional UPDATE — a deliberately different concurrency model than the SM's lock-free read-modify-write).

Follow-up to #1682.

### Definition of Done

- [ ] `CertificateUtil.prepareIssuedCertificate` issue-completion transition routed through `CertificateStateMachine` with from-state validation (covers synchronous and async-poll completion)
- [ ] `CertificateServiceImpl.revokeCertificate(String serialNumber)` routed through the SM with a from-state guard
- [ ] Existing state outcomes and audit history preserved, verified by existing and new tests
- [ ] No remaining direct certificate `setState` outside the SM except the documented by-design exemptions

lubomirw:
## Summary

Completes #1682 by bringing the two remaining certificate-state writers under `CertificateStateMachine`, so every ISSUED / REVOKED transition is validated against the transition table and audited consistently.

## Changes

- **`CertificateUtil`** — extracted a field-only `stampIssuedFields` out of `prepareIssuedCertificate`; `prepareIssuedCertificate = stampIssuedFields(...) + setState(ISSUED)`, so its creation/import callers and all test fixtures are unchanged.
- **`ClientOperationServiceImpl`** — the sync issue/renew/rekey handlers transition the certificate to `PENDING_ISSUE` via the state machine **before** the connector call, so both the sync (200) and async (202) outcomes leave it in `PENDING_ISSUE`; the shared `transitionToPendingIssue` helper is stripped of its (now-redundant) transition and renamed `onAsyncAccepted`.
- **`CertificateServiceImpl.issueRequestedCertificate`** — routes `PENDING_ISSUE → ISSUED` through the state machine (which owns the save + `ISSUE/SUCCESS` audit); the manual save + audit are removed. Covers both the sync issue/renew tails and the async-poll completion path.
- **`CertificateServiceImpl.revokeCertificate(String)`** (legacy v1 revoke) — routed through the state machine **idempotently**: a `canTransition` guard logs and skips a no-op/illegal-from-state revoke rather than throwing. The legacy path calls the connector first, so a hard local throw would diverge local state from an already-revoked authority; the status-changed event + auth-cache eviction now fire only on a real transition.

No new `CertificateStateTransition` rows — the required `→ PENDING_ISSUE`, `PENDING_ISSUE → ISSUED`, and `ISSUED → REVOKED` arcs already exist.

## Out of scope

The third REVOKED writer — `X509CertificateValidator` (OCSP/CRL-driven revoke via its own `CertificateValidationWriter`) — is deferred to a follow-up.

## Testing

Per-method unit/integration tests (valid transition + illegal-from-state behavior), a sync `PENDING_ISSUE → FAILED` failure test, and the existing v2 / upload / import / ACME / async-poll suites as the regression net. Branch synced with `main`.

Closes #1683
Part of epic ilm#110.


src/test/java/com/otilm/core/service/CertificateServiceTest.java
```diff
@@ -60,6 +60,7 @@
 import org.bouncycastle.operator.OperatorCreationException;
 import org.jetbrains.annotations.NotNull;
 import org.jspecify.annotations.NonNull;
+import com.otilm.core.service.handler.authority.lifecycle.InvalidTransitionException;
 import org.junit.jupiter.api.AfterEach;
```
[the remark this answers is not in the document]
lubomirw: Moved the `InvalidTransitionException` import into the `com.otilm.core` group. Fixed in e34576741.

src/test/java/com/otilm/core/service/ClientOperationServiceV2Test.java
```diff
@@ -1153,6 +1153,67 @@ void issueCertificateAction_recordsEventHistoryEntry_on202() throws Exception {
...
+                "expected the state-machine's PENDING_ISSUE audit row (ISSUE/SUCCESS, "
+                        + "\"Issuance in progress\") to precede ISSUED on the sync path");
+    }
+
+    /**
+     * When the connector returns an error (no certificate data) on a synchronous issue call,
+     * the certificate must end in FAILED and the audit history must contain an ISSUE/FAILED row.
+     * The failure-from-state is PENDING_ISSUE (the pre-connector transition already ran), so this
+     * exercises the new PENDING_ISSUE -> FAILED arc introduced by this branch.
+     */
```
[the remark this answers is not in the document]
lubomirw: Reworded — the `PENDING_ISSUE -> FAILED` arc is pre-existing; this branch only reroutes the sync failure through it (previously from REQUESTED). Fixed in e34576741.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -601,13 +607,16 @@ public void issueCertificateAction(final UUID certificateUuid, boolean isApprove
...
+     * it via the state machine before calling the connector — so this method no longer transitions.
      * The {@code originatingAction} (ISSUE / RENEW / REKEY) drives the event-log
      * {@code CERTIFICATE_ACTION_PERFORMED} message so subscribers see the operation that produced
      * this state — not always {@code ISSUE}.
      */
-    private void transitionToPendingIssue(Certificate certificate, CertificateDataResponseDto acceptedBody, ResourceAction originatingAction) {
+    private void onAsyncAccepted(Certificate certificate, CertificateDataResponseDto acceptedBody, ResourceAction originatingAction) {
         if (acceptedBody != null && acceptedBody.getMeta() != null && !acceptedBody.getMeta().isEmpty()) {
             try {
                 attributeEngine.updateMetadataAttributes(acceptedBody.getMeta(),
```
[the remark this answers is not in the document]
lubomirw: Reworded — the helper no longer transitions; the certificate is already in `PENDING_ISSUE` before the connector call, so the comment now says we keep it there. Fixed in e34576741.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -559,6 +559,12 @@ public void issueCertificateAction(final UUID certificateUuid, boolean isApprove
         caRequest.setRaProfileAttributes(attributeEngine.getRequestObjectDataAttributesContent(ObjectAttributeContentInfo.builder(Resource.RA_PROFILE, certificate.getRaProfile().getUuid()).connector(certificate.getRaProfile().getAuthorityInstanceReference().getConnectorUuid()).build()));
 
         try {
+            // Move to PENDING_ISSUE before calling the connector so every path (sync 200 or async
+            // 202) reaches issueRequestedCertificate / the poll cycle from a uniform PENDING_ISSUE
+            // state via the state machine.
+            stateMachine.transition(certificate, CertificateState.PENDING_ISSUE, CertificateEvent.ISSUE,
```
ivosh, line 566: **~ Minor** | Correctness

- File: src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java:565
   - What: The pre-connector transition commits PENDING_ISSUE in its own transaction before the connector call. On the sync (200) path no status poll is scheduled (scheduleStatusPoll runs only in onAsyncAccepted, i.e. only on 202). If the process dies between the committed PENDING_ISSUE and the terminal ISSUED/FAILED transition, the cert is stranded in PENDING_ISSUE with nothing scheduled to reconcile it.
   - Why: Before this PR the cert stayed in REQUESTED, a retriable state. Now PENDING_ISSUE semantically implies an async op is in flight, but on the sync path there is none.
   - Fix: Confirm a sweep/reaper re-drives orphaned PENDING_ISSUE certs regardless of a poll row, or document the trade-off in the handler comment.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1296,8 +1303,12 @@ public void revokeCertificate(String serialNumber) {
         try {
             certificate = certificateRepository.findBySerialNumberIgnoreCase(serialNumber).orElseThrow(() -> new NotFoundException(Certificate.class, serialNumber));
             oldStatus = certificate.getValidationStatus();
-            certificate.setState(CertificateState.REVOKED);
-            certificateRepository.save(certificate);
+            if (!stateMachine.canTransition(certificate.getState(), CertificateState.REVOKED)) {
```
ivosh: **~ Minor** | Correctness

- File: src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java:1306
   - What: In the v1 revoke, the connector revoke call succeeds first, then revokeCertificate(serial) runs. The new canTransition guard silently skips (log.warn + return) when the from-state has no ->REVOKED arc. For a non-revocable-but-not-already-REVOKED state, the external CA has already revoked while local state is left unchanged.
   - Why: The previous unconditional setState(REVOKED) always kept local state aligned with the just-completed connector revocation; the lenient guard trades that for idempotency and masks surprising states with only a warn.
   - Fix: Distinguish "already REVOKED (benign skip)" from "unexpected non-revocable state after connector already revoked" — force-reconcile or raise a visible alert for the latter.

src/test/java/com/otilm/core/service/ClientOperationServiceV2Test.java
```diff
@@ -1153,6 +1153,68 @@ void issueCertificateAction_recordsEventHistoryEntry_on202() throws Exception {
...
+    @Test
+    void issueCertificateAction_transitionsToFailed_onSyncConnectorError() throws Exception {
+        UUID certUuid = prepareCertificateForIssuance();
+        // Connector returns 200 but with an empty body — no certificate data
+        mockServer.stubFor(WireMock
+                .post(WireMock.urlPathMatching("/v2/authorityProvider/authorities/[^/]+/certificates/issue"))
+                .willReturn(WireMock.okJson("{}")));
+
+        Assertions.assertThrows(
```
ivosh, line 1201: **~ Minor** | Testing

- File: src/test/java/com/otilm/core/service/ClientOperationServiceV2Test.java:1201
   - What: ClientOperationServiceV2Test adds sync-200 and sync-error coverage for issueCertificateAction only; the identical pre-connector transitions added to renewCertificateAction and rekeyCertificateAction have only 202-async tests.
   - Why: renew/rekey start from a different from-state (REQUESTED/PENDING_APPROVAL) than issue; only issue's path is tested end-to-end.
   - Fix: Add sync-path renew (and rekey) tests mirroring the issue tests.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1296,8 +1303,12 @@ public void revokeCertificate(String serialNumber) {
...
+                log.warn("Certificate {} is in state {}; skipping revoke transition (already revoked or not revocable)",
+                        certificate.getUuid(), certificate.getState());
+                return;
```
ivosh: **~ Minor** | Security

CertificateServiceImpl.java:1306-1319 — the new `if (!stateMachine.canTransition(getState(), REVOKED)) { log.warn(...); return; }` guard returns at line 1309 *before* the auth-cache eviction block (line 1317 `authenticationCache.evictByCertificateFingerprint(...)`) and the status-changed event (line 1319). Previously the method unconditionally did `setState(REVOKED)`+`save`, then always evicted. So eviction of the positive-authentication cache (AuthenticationCache, keyed by SHA-256 cert fingerprint — the entry that lets a client-cert principal skip re-authentication) now fires ONLY on a real PENDING/ISSUED->REVOKED transition.

Security-relevant window: for a user-mapped cert (`getUserUuid() != null`) that is in a non-revocable state, `revokeCertificate` now silently no-ops (warn + return) AND leaves the positive-auth entry live until its TTL expires, whereas before it was evicted immediately. It also removes the prior idempotent safety-net: re-invoking revoke on an already-REVOKED cert can no longer force-flush a lingering cached auth entry (e.g. one repopulated after a race or a prior partial eviction). The AuthenticationCache Javadoc at AuthenticationCache.java:33 still asserts the cert entry is 'evicted immediately on revocation via CertificateServiceImpl' — the guard can now falsify that.

Suggest to move the eviction/status-event out of the early-return path so a revoke request always evicts a user-mapped cert's cached authentication regardless of the SM outcome.

src/test/java/com/otilm/core/service/CertificateServiceTest.java
```diff
@@ -1908,6 +1909,120 @@ private Certificate setUpRequestedIssueChain() throws Exception {
...
+
+            assertThat(certificateEventHistoryRepository.findByCertificateOrderByCreatedDesc(reloaded))
+                    .as("no-op revoke must not write a REVOKE audit row")
+                    .noneMatch(h -> h.getEvent() == CertificateEvent.REVOKE);
+        }
+
+        @Test
+        void revokeCertificateBySerial_notFound_warnsAndReturns() {
+            // when / then — unknown serial triggers the NotFoundException path (warn + return)
+            assertThatCode(() -> certificateService.revokeCertificate("deadbeef-not-a-real-serial"))
```
ivosh, line 2047: **. Nitpick** | Testing

CertificateServiceTest.java:2019-2024 (`revokeCertificateBySerial_notFound_warnsAndReturns`) asserts only `assertThatCode(...).doesNotThrowAnyException()`. This is a green-but-covers-nothing risk (CLAUDE.md "Tests that look green but cover nothing"): the test would pass even if `revokeCertificate` silently returned before reaching the intended NotFoundException catch branch, or if a future refactor stopped calling the SM entirely. It never confirms the NotFound path was actually taken. Strengthen it by asserting the observable consequences of that branch: no cert-event-history REVOKE row was written and no `CertificateStatusChangedEvent` was produced (the `certificate != null` block is skipped when nothing is found).

src/test/java/com/otilm/core/service/ClientOperationServiceV2Test.java
ivosh, line 1201: **. Nitpick** | Testing

ClientOperationServiceV2Test.java:1201-1204 (`issueCertificateAction_transitionsToFailed_onSyncConnectorError`) uses `Assertions.assertThrows(Exception.class, ...)`, while every sibling test in this file asserts the specific type (`CertificateOperationException`, `ValidationException`, `NotFoundException`, e.g. lines 290, 305, 975). The broad type would let the test pass even if the throw originated somewhere other than the intended sync-connector-error path. The follow-on assertions on FAILED state and the ISSUE/FAILED row do ground it, so this is only a nitpick — narrow to `CertificateOperationException.class` to match the neighbour pattern. Non-blocking.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -909,6 +915,12 @@ public void renewCertificateAction(final UUID certificateUuid, ClientCertificate
         HashMap<String, Object> additionalInformation = new HashMap<>();
         additionalInformation.put("New Certificate UUID", certificate.getUuid());
         try {
+            // Move the new certificate to PENDING_ISSUE before calling the connector so every path
```
ivosh, line 916: **. Nitpick** | Testing

This PR moves the PENDING_ISSUE transition to *before* the connector call in all three handlers, and the new sync-path tests (`issueCertificateAction_transitionsThroughPendingIssue_on200Sync`, `..._onSyncConnectorError`) cover only the issue path. The identical new ordering added to `renewCertificateAction` (ClientOperationServiceImpl.java:918-922) and `rekeyCertificateAction` (1133-1137) — where a sync 200 or a connector error now also routes through PENDING_ISSUE first — is exercised only indirectly. Existing renew/rekey tests assert PENDING_ISSUE only on the 202 branch (ClientOperationServiceV2Test.java:1382, 1458). Since the issue path is tested and renew/rekey mirror it exactly, this is a nitpick; add a renew (or rekey) sync-200 test asserting the "Issuance in progress" PENDING_ISSUE audit row precedes ISSUED to lock in the new ordering.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -601,13 +607,16 @@ public void issueCertificateAction(final UUID certificateUuid, boolean isApprove
     }
 
     /**
-     * Transitions the certificate to {@code PENDING_ISSUE}, persists any metadata the connector
-     * returned in its {@code 202 Accepted} response body, and records an event-history entry.
+     * Records the connector's asynchronous acceptance ({@code 202 Accepted}) of an issue / renew /
+     * rekey: persists any metadata the connector returned in its response body, schedules the status
+     * poll, raises the action-performed event, and writes the async-acceptance event-log entry. The
+     * certificate is already in {@code PENDING_ISSUE} by the time this runs — the handlers transition
+     * it via the state machine before calling the connector — so this method no longer transitions.
```
ivosh: **~Minor | Documentation**

Do not narrate the story which led to the current shape. That story is available from the history.

ivosh (changes requested):
The state-machine design itself is sound: all three sync handlers are `@Transactional(NOT_SUPPORTED)`, so each stateMachine.transition commits in its own short tx before the connector HTTP call — the feared lock-across-HTTP does not occur, and the "no new transition rows" claim checks out. The findings cluster around one genuinely new regression (the pre-connector PENDING_ISSUE commit) plus revoke-guard side-effects and test gaps.

One general remark: some of the comments are too wordy and they actually put a strain on the human brain to distil useful information from them. Consider trimming them down.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Documented the trade-off at the issue handler in 12214d34c — no poll row is created on the sync path, so a crash before the terminal ISSUED/FAILED transition strands the cert in PENDING_ISSUE (the poll sweep is poll-row-driven; `UpdateCertificateStatusTask` does not re-drive lifecycle state). A reaper for poll-less PENDING_ISSUE certs would be a separate follow-up.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
lubomirw: Fixed in 12214d34c — the guard now logs benign "already REVOKED" at debug and a surprising non-revocable state at warn (the authority may already have revoked upstream). Kept lenient rather than force-reconciling, to respect the SM model.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
lubomirw: Fixed in 12214d34c — moved the auth-cache eviction and the status-changed event out of the early-return path, so a revoke always flushes a user-mapped cert's positive-auth entry regardless of the SM outcome (restores the AuthenticationCache Javadoc contract and the idempotent safety-net). Added `evictsCertCache_whenRevokeSkipsTransition_forNonRevocableUserCert` to lock it in.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Fixed in 12214d34c — reworded the `onAsyncAccepted` Javadoc and the metadata-failure comment to state the current invariant (the cert is already PENDING_ISSUE; this method leaves the state unchanged) without narrating the change.

src/test/java/com/otilm/core/service/ClientOperationServiceV2Test.java
lubomirw: Added renew + rekey sync-200 tests in 12214d34c (`renewCertificateAction_transitionsThroughPendingIssue_on200Sync` / `rekeyCertificateAction_transitionsThroughPendingIssue_on200Sync`), each asserting the "Issuance in progress" PENDING_ISSUE audit row precedes ISSUED.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Covered in 12214d34c — added renew + rekey sync-200 tests asserting the "Issuance in progress" PENDING_ISSUE audit precedes ISSUED (same fix as the sibling testing comment).

src/test/java/com/otilm/core/service/CertificateServiceTest.java
lubomirw: Strengthened in 12214d34c — now asserts the cert-event-history count is unchanged, proving the not-found branch writes no REVOKE row. (`eventProducer` is a real @Autowired bean here, not spyable, so the count delta stands in for "no status event produced" — both live in the same skipped `certificate != null` block.)

src/test/java/com/otilm/core/service/ClientOperationServiceV2Test.java
lubomirw: Narrowed to `CertificateOperationException.class` in 12214d34c to match the sibling tests.

ivosh (approved):

PR #1708, opened 2026-07-02
Dereference SECRET references in NG reference-expansion (callback + v3 operation path)

### LINKED ISSUE #1697: Dereference SECRET references in NG reference-expansion (callback loader + operation-path expandAsSystem)
### Description

The NG Attributes v2 reference-expansion facility (`AttributeReferenceExpander`, added in #1624 / PR #1678) resolves reference-typed attribute values into inline content before Core sends them to a stateless connector. **Two gaps remain**, both surfaced by the first real adopter (`otpki-connector`), whose authority authenticates upstream via an `oauthClient` attribute — a `RESOURCE`-content value referencing a **SECRETS** object (a stored Basic-Auth secret). The connector expects Core to deliver it **dereferenced** (inline `{type:basicAuth, username, password}`); a bare reference is unusable to a stateless connector (it models both a dereferenced view and a reference-only `SecretRef`).

**Gap 1 — SECRET references are not dereferenced on the callback path.** `CallerAuthorizedReferenceLoaderRegistry` registers loaders only for `CREDENTIAL`, `AUTHORITY`, `ENTITY`, `LOCATION`. For `AttributeResource.SECRET`, `AttributeReferenceExpander.expandElement` finds no loader and passes the reference through unchanged. The NG callback scope-blob assembly (`AttributeCallbackScopeResolver.buildScopedAttributes`) and `CallbackServiceImpl.dispatchNg` rely solely on `expandForCaller`, so an authority scope blob's `oauthClient` reaches the connector as a bare reference. **This shipped in #1679** (merged 2026-07-01; now live on `main`); otpki callbacks that must reach upstream (e.g. dropdowns querying the CA) cannot authenticate until this lands. **This now gates the otpki adoption (#25):** otpki authenticates via a SECRET (`oauthClient`), so it cannot complete end-to-end (callbacks or operations) until this is done.

**Gap 2 — the v3 operation path expands nothing.** `AuthorityProviderV3Adapter` builds `authorityAttributes` / `raProfileAttributes` via `getRequestObjectDataAttributesContent(...)` with no expansion; the expander's operation entry point (`expandAsSystem`) is an unimplemented placeholder. So issue/renew/revoke through a stateless v3 authority delivers unresolved references (SECRET and otherwise) → the connector cannot authenticate upstream → operations fail. Authority-v3 is now **finalized**, so the contract is stable and this is ready to implement.

The fix is **path-dependent** — callback mode authorizes the calling user per object; operation mode resolves under a system identity — and reuses the platform's existing secret-content read so the established authorization model is preserved.

### Fix A — Callback path (unblocks otpki callbacks)

Keep the dereference **inside** `AttributeReferenceExpander` so outbound secret containment and fail-closed authorization compose automatically:

- Register a SECRET loader in `CallerAuthorizedReferenceLoaderRegistry`: `AttributeResource.SECRET` → a loader that takes the `SecuredUUID` the expander already builds and delegates to the caller-authorized secret read `SecretService.getResourceObjectContent(uuid)`. That read enforces per-object authorization internally via `permissionEvaluator.vaultProfileMembers(SecuredUUID)` (`VAULT_PROFILE:MEMBERS`) on the ambient caller, fail-closed.
- **No change** to `AttributeCallbackScopeResolver.buildScopedAttributes` or `CallbackServiceImpl.dispatchNg` — they already route through `expandForCaller`, so the secret now dereferences and `expandElement`'s `recordExpandedSecrets(...)` records it, **arming the value-echo arm of the outbound containment check** (which is blind to any secret dereferenced off-path).
- Wrap the **single-object** `getResourceObjectContent`, **NOT** the `loadResourceObjectContentData(List)` overload (the latter is forbidden by `AttributeReferenceExpanderFenceArchTest`).
- Fix the `CallerAuthorizedReferenceLoaderRegistry` Javadoc, which wrongly claims SECRET has "no connector-consumable blob."

**Authorization-shape caveat:** secrets are authorized by `VAULT_PROFILE:MEMBERS`, not `<KIND>:DETAIL` like the other four kinds — this is the correct model for secret content. Do **not** rely on the `@ExternalAuthorization(SECRET, GET_SECRET_CONTENT)` annotation on `getResourceObjectContent` as the per-object gate: with a plain `UUID` argument its object-scoping degrades to a resource-level check; the genuine per-object gate is the internal `vaultProfileMembers` call, which must be preserved.

### Fix B — Operation path (v3 authority)

Implement operation-mode expansion and wire it into `AuthorityProviderV3Adapter`:

- Resolve references (including SECRET) under a **system identity** — no per-caller authorization; the operation is already authorized at the operation level and runs as protocol/scheduler principals — via the existing core resource-content resolver `ResourceService.loadResourceObjectContentData(List)` / `getResourceObjectContentData` (which already handles SECRET and FILTER mappings), matching the existing reuse in `ConnectorRequestAttributesBuilder`.
- Override the attribute builders in `AuthorityProviderV3Adapter` (`authorityAttributesFor` / `raProfileAttributesFor`) to run through this resolution; leave the shared base adapter and the v2 adapter **untouched** (v2 connectors are stateful — byte-identical).
- Extend `AttributeReferenceExpanderFenceArchTest` to permit this specific operation entrypoint rather than adding an unguarded twin.

### Definition of Done

- **Fix A:** SECRET loader registered; an authority scope blob's `oauthClient` is delivered dereferenced (inline Basic-Auth) in a callback's `contextAttributes`; a caller lacking vault-profile membership fails closed (no connector call); a connector echoing the dereferenced secret back is rejected by outbound containment (value-echo arm armed); fence test green; multi-select denial fails closed (propagates, no partial expansion).
- **Fix B:** issue/renew/revoke through a stateless v3 authority (otpki) whose `oauthClient` is a stored secret authenticates upstream successfully; a protocol/scheduler principal succeeds (not fail-closed); v2 authority path byte-identical (boundary test); fence extended, no unguarded twin.
- `CallerAuthorizedReferenceLoaderRegistry` Javadoc corrected (SECRET is connector-consumable).

### Context / links

- **Ships-with-gap:** #1679 — Gap 1 is a known limitation there; PR stays mergeable, this is the tracked follow-up.
- **Builds on:** #1624 / PR #1678 (the reference expander) and the merged secure-listing work (#1623).
- Coordinates with the finalized authority-v3 work (ilm#110); part of the Attributes v2 feature (ilm#251).

lubomirw:
Closes #1697.

Stateless NG connectors receive stored resource references they cannot resolve themselves. The first adopter (otpki, otpki#25) authenticates upstream via `oauthClient` — a stored Basic-Auth SECRET referenced from its authority attributes — and expects Core to deliver it dereferenced. This closes both gaps left open after #1679.

## Fix A — callback path
Registers a caller-authorized SECRET loader in `CallerAuthorizedReferenceLoaderRegistry`, so a stored SECRET reference carried in a scope blob resolves to inline content on the callback path instead of reaching the connector as a bare reference. The loader reads through `SecretServiceImpl.getAuthorizedObjectAttributes`, guarded by `SECRET:GET_SECRET_CONTENT` plus the vault-profile-membership gate — so expansion fails closed when the caller lacks membership, and the outbound value-echo containment arms automatically as the secret flows through the expander.

**@vyskocilm — this reverses a deliberate stance from #1679.** The registry's Javadoc previously excluded SECRET ("SECRET's inline content is the connector's responsibility … never re-fetch a user-typed SECRET value"). That reasoning holds for *user-typed inline secrets* (first-delivery values in `currentAttributes` — not `ResourceObjectContent` references, so the expander never walks them), but not for a *stored* SECRET reference that a stateless connector cannot resolve. Only the latter is dereferenced here; user-typed inline secrets stay untouched, and CERTIFICATE stays pass-through. Flagging for your review.

## Fix B — v3 operation path
The v3 authority operation path (issue/renew/revoke/register, status poll/cancel, and the attribute-list requests) sent the stored authority + ra-profile attributes to the connector with their references intact. Adds a dereference-only primitive to the fence-exempt `ConnectorRequestAttributesBuilder` (resolves CREDENTIAL + RESOURCE incl. SECRET, without re-validating — the attributes were validated when persisted) and routes both v3 attribute blobs through it under the operation principal (no per-object authorization; the operation is authorized at the operation level). v2 authority connectors are stateful and resolve their own references, so the base adapter's non-dereferencing helpers still serve v2 unchanged.

## Tests
- **Registry** — SECRET loader registered; CERTIFICATE stays a pass-through null loader.
- **Callback expander (unit)** — a stored SECRET reference resolves to inline content and arms containment.
- **Callback expander (integration)** — SECRET expansion fails closed when the caller lacks vault-profile membership (real `@ExternalAuthorization` aspect via OPA).
- **Builder (unit)** — dereferences credential + resource content without re-running `validateUpdateDataAttributes`.
- **v3 adapter** — issue and attribute-list requests carry dereferenced authority + ra-profile attributes.
- **v2 adapter** — byte-identical guard: stored ra-profile attributes pass through unchanged (no dereference).
- **Reference-expander fence ArchTest** stays green (the adapter → builder dependency is outside the engine package).


lubomirw:
## Follow-up: operation-path deref now runs under a scoped system identity

The multi-review + a deeper investigation showed the original operation-path change (commit `6ab6c8424`) resolved an authority's own infrastructure references (credential/secret) in the **wrong authorization context**:

- the **async status-poll** thread runs with **no principal** → guarded reads threw `AuthenticationCredentialsNotFoundException`;
- **operator / protocol robots** (ACME/SCEP/CMP) don't hold per-object rights on the authority's *infrastructure* secret → denied.

The secret-content fetch is guarded recursively (`getConnector` `CONNECTOR:DETAIL`, `vaultProfileMembers`, a nested guarded builder call), so "un-gating the reads" is a dead end. Commit `4c2322cfa` instead resolves these references as the **platform**, authorized at the operation level:

- **`AuthHelper.runAsSystem`** — scoped elevation (snapshot Authentication → fresh context → restore/clear in `finally`, with MDC-actor discipline) so no system principal/actor leaks onto a pooled thread.
- **`OperationAttributeResolver`** — the single home for system-mode operation-path resolution; elevates to a dedicated least-privilege **`attribute-resolver`** identity and dereferences via the existing guarded `ConnectorRequestAttributesBuilder`. The v3 adapter delegates to it; **v2 stays byte-identical**.
- **`attribute-resolver`** is seeded by a Flyway migration with only `CONNECTOR:DETAIL`, `CREDENTIAL:DETAIL`, `SECRET:GET_SECRET_CONTENT`, `VAULT_PROFILE:MEMBERS`, `CERTIFICATE:DETAIL` (the object-config kinds resolve via an unguarded engine read and need no grant).

**@vyskocilm — two things that need your eyes**, since this extends the SECRET authz model you set:
1. **Privilege-elevation path.** `runAsSystem` is a new scoped elevation; the callback expander is fenced from it (ArchUnit) and elevation is confined to `OperationAttributeResolver`. Please sanity-check the trust boundary.
2. **E2E grant verification (top risk).** Integration tests mock OPA allow-all and Flyway is disabled in tests, so they prove the *mechanism*, not that the `attribute-resolver` role's grants are complete against the real OPA policy. This must be verified in a real environment before relying on the identity; if the migration can't land this release, the interim is `superadmin` behind the same constant.

Audit note: authorization runs as `attribute-resolver` while the audit actor stays the caller (operator/protocol) — deliberate, so it reads as "the platform resolved the authority's own infra secret on the caller's behalf."

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -299,11 +307,32 @@ private List<RequestAttribute> issueAttributesFor(Certificate cert, AuthorityIns
...
+     * level). v2 stays stateful and is untouched — the base adapter's non-dereferencing helpers still serve it.
+     */
+    private List<RequestAttribute> dereferenceForConnector(UUID connectorUuid, List<RequestAttribute> stored)
+            throws ConnectorException {
+        try {
+            return connectorRequestAttributesBuilder.dereferenceForConnectorRequest(connectorUuid, stored);
+        } catch (AttributeException | NotFoundException e) {
+            throw new ConnectorException("Unable to resolve stored attribute references for connector request", e);
+        }
     }
```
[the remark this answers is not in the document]
lubomirw: Fixed in d762e36ef — the resolution-failure message now includes the connector UUID.

src/test/java/com/otilm/core/service/handler/authority/AuthorityProviderV3AdapterTest.java
```diff
@@ -167,6 +173,62 @@ void issueLoadsCertificateRequestAttributesScopedToIssueOperation() throws Conne
...
+
+        ArgumentCaptor<CertificateAttributeListRequestDtoV3> dtoCaptor =
+                ArgumentCaptor.forClass(CertificateAttributeListRequestDtoV3.class);
+        verify(certClientV3).listIssueAttributes(eq(connectorInfo), dtoCaptor.capture());
+        CertificateAttributeListRequestDtoV3 dto = dtoCaptor.getValue();
+        assertSame(resolved, dto.getAuthorityAttributes(),
+                "the attribute-list request must carry resolved authority attributes");
+        assertSame(resolved, dto.getRaProfileAttributes(),
+                "the attribute-list request must carry resolved ra-profile attributes");
+    }
```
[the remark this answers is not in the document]
lubomirw: Fixed in 363e64a32 — the test now stubs distinct authority and RA-profile resolved lists and asserts each request slot carries its own, so an authority↔RA-profile cross is caught.

src/main/java/com/otilm/core/attribute/engine/ConnectorRequestAttributesBuilder.java
```diff
@@ -40,6 +40,32 @@ public void setCredentialService(CredentialInternalService credentialService) {
...
+     * definitions are supplied here; per-attribute content validation still runs inside
+     * {@code getDataAttributesByContent}. Callers resolving an authority's own infrastructure references go through
+     * {@code OperationAttributeResolver}, which elevates to the platform's attribute-content-resolver system identity
+     * for the duration of this call (authorized at the operation level, not per acting caller).
+     * <p>
+     * This does not arm the callback path's outbound-secret value-echo containment. For issue/renew/revoke/register the
+     * connector response is certificate data, not a caller-reflected surface. The attribute-list endpoints
+     * (list{Issue,Revoke,Register,RaProfile}Attributes) do return connector-supplied content to the operator, so a
+     * malicious connector could echo a resolved secret there — but it already holds that secret to authenticate to the
+     * CA, so the residual exposure is limited; hardening those responses with containment is tracked as a follow-up.
```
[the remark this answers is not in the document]
lubomirw: Fixed in af1e41b6e — the javadoc now describes the shipped behavior: AuthorityProviderV3Adapter routes the four list*Attributes responses through OutboundSecretContainment (record-then-assert, fail-closed), so it's no longer a follow-up.

src/test/java/com/otilm/core/integration/attribute/engine/AttributeReferenceExpanderITest.java
```diff
@@ -181,4 +208,74 @@ void expanderDoesNotMutateAmbientPrincipal() throws Exception {
...
+        content.setPassword("oauth-client-secret");
+        secretVersion.setFingerprint(SecretsUtil.calculateSecretContentFingerprint(content));
+        secretVersionRepository.save(secretVersion);
+
+        secret.setLatestVersion(secretVersion);
+        secretRepository.save(secret);
+
+        secretVersion.setSecretUuid(secret.getUuid());
+        secretVersionRepository.save(secretVersion);
+        return secret;
```
ivosh, line 258: **. Nitpick** | Testing

The integration test (secretExpansionFailsClosedWhenDeniedVaultProfileMembership) only covers the
   DENY path for SECRET; there is no positive end-to-end test that a stored SECRET actually resolves
   to inline content (the fixture deliberately stubs no vault content). The happy path is only
   unit-tested with mocked loaders — understandable given the vault-connector dependency, but the
   successful dereference is the feature's core and is currently unproven end-to-end. Separately,
   OperationAttributeResolverTest does not assert that a `ConnectorException` thrown by
   dereferenceForConnectorRequest propagates unwrapped (not double-wrapped) — a cheap test given
   ConnectorException is the one checked type deliberately left uncaught in the lambda.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
```diff
@@ -49,10 +49,14 @@
...
- * NOT implemented here. See {@code #expandViaOperationLoader} placeholder below and the follow-up PR. This
- * facility slice ships callback mode only.
+ * <h2>Operation mode lives outside this class — by design</h2>
+ * The v3 operation path (issue/renew/revoke/register, status poll/cancel) does <strong>not</strong> run through
+ * this fenced callback expander. {@code OperationAttributeResolver} (from {@code AuthorityProviderV3Adapter})
+ * resolves an authority's own infrastructure references there by elevating to the platform's
+ * attribute-content-resolver system identity for the dereference only — authorized at the operation level, not per
+ * acting caller. The ArchUnit
+ * fence forbids this package from depending on {@code AuthHelper}, so the caller-authorized (callback) mode here can
+ * never elevate; the two modes stay separated.
```
ivosh, line 59: **~ Minor** | Documentation

AttributeReferenceExpander.java:70 still reads that the depth cap/cross-frame cycle guard "becomes live when the operation-path follow-up implements nested resolution." This PR deliberately rewrote that narrative everywhere else: the operation path shipped in this PR (OperationAttributeResolver, called from AuthorityProviderV3Adapter), it does not run through this class, and it does not make attributesOf live. The changed lines confirm it — line 41 now says "Nested resolution is future work ... operation mode does not run through this expander," line 59 states the two modes stay separated, and the attributesOf comment (line 179) was scrubbed to "a future nested-resolution change." Line 70 is the one spot that missed the scrub, so it now asserts a false coupling: the operation-path work has landed without touching nested resolution. 

Recommendation: drop the references to future work altogether or retarget this occurrence to "a future nested-resolution change" to match lines 41/59/179.

Anchored on the nearest changed line (59) because line 70 itself was not touched by this PR.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
```diff
@@ -38,7 +38,7 @@
  * <strong>Single-level only today.</strong> Only the references directly present in the passed attribute list
  * are resolved (depth 0). Resolving references nested <em>inside</em> an already-loaded blob (depth 1+) is not
  * wired — {@link #attributesOf} returns empty, so the recursion below is a no-op and the depth/cycle machinery
- * is inactive at runtime. Nested resolution is deferred to the operation-path follow-up (see DEFERRED below).
+ * is inactive at runtime. Nested resolution is future work (see below); operation mode does not run through this expander.
```
ivosh, line 41: **~ Minor** | Documentation

AttributeReferenceExpander.java:28 still hardcodes the callback gate as "the resource's object-scoped @ExternalAuthorization(<KIND>, DETAIL) guarded entrypoint" — a single uniform DETAIL gate across kinds. This PR breaks that uniformity: it registers a SECRET loader (CallerAuthorizedReferenceLoaderRegistry.java:46 → SecretServiceImpl.getAuthorizedObjectAttributes) gated on @ExternalAuthorization(SECRET, GET_SECRET_CONTENT) PLUS a vault-profile-membership check, not DETAIL. 

The registry's OWN Javadoc was correctly updated to describe this ("SECRET resolves via SECRET:GET_SECRET_CONTENT plus vault-profile membership"), but the expander's class Javadoc at line 28 was left describing the old DETAIL-only model, so the two docs now disagree.

Recommendation: do not restate the obvious. These low-level details are available easily from the source codebase and do not need to be repeated by inline comments.

Anchored on the nearest changed line (41) because line 28 itself was not touched by this diff.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -190,8 +206,25 @@ public void checkAuthorityConnection(AuthorityInstanceReference authority, List<
...
-                .listRegisterAttributes(connectorDto, attributeListRequest(authority, raProfile));
+        CertificateAttributeListRequestDtoV3 request = attributeListRequest(authority, raProfile);
+        List<BaseAttribute> response = connectorApiFactory.getCertificateApiClientV3(connectorDto)
+                .listRegisterAttributes(connectorDto, request);
+        return contained(response, request.getAuthorityAttributes(), request.getRaProfileAttributes());
+    }
+
+    /**
+     * Fail closed if the connector echoed back a secret Core expanded into the request. The v3 attribute-list
+     * responses flow to the operator's attribute-selection UI (unlike issue/revoke/register, whose response is
```
ivosh: **~ Minor** | Security

AuthorityProviderV3Adapter.java:217 asserts issue/revoke/register responses are 'certificate data' and so are exempt from OutboundSecretContainment, while the four list* endpoints (lines 164/173/182/212) run their responses through contained(). But issue/renew/register/issueRegistered do return a connector-controlled surface that is NOT certificate data: the meta bag. mapIssueRenewRegisterResponse (line 406) forwards body.getMeta() verbatim into AdapterOperationResult, where it is persisted on the certificate and replayed/shown to operators. Before this PR the v3 wire carried BARE references, so the connector never received the resolved secret and could not echo it; now authorityAttributes/raProfileAttributes carry inline SECRET/credential content (lines 100-101,124-125,143-144,245-246,265-266), so a buggy connector that reflects a received secret into meta would land that plaintext in Core's DB and operator UI — the exact accidental-echo the list* containment is built to catch. Question: is the meta surface deliberately excluded because the connector is the legitimate recipient of the secret (so echo is not a NEW disclosure to an untrusted party), or should meta on the issue/renew/register/issueRegistered path also pass through OutboundSecretContainment (or be recorded and value-echo-checked) for symmetry with the list* guard? If the former, tighten the Javadoc at line 217 to say 'certificate data plus a connector-owned meta bag we deliberately do not scan because the connector already holds the secret'; if the latter, run the persisted meta through containment. Low marginal risk (connector already possesses the plaintext), hence minor, not blocking.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -251,7 +284,7 @@ public StatusPollResult pollStatus(Certificate cert, CertificateOperation op) th
         CertificateOperationStatusRequestDtoV3 wire = new CertificateOperationStatusRequestDtoV3();
         wire.setMeta(loadMeta(cert, authority));
         wire.setAuthorityAttributes(authorityAttributesFor(authority));
-        wire.setRaProfileAttributes(raProfileAttributesFor(raProfile, authority));
+        wire.setRaProfileAttributes(resolvedRaProfileAttributes(raProfile, authority));
```
ivosh, line 294: **~ Minor** | Security

AuthorityProviderV3Adapter.java:287 (pollStatus) now calls resolvedRaProfileAttributes(raProfile, authority) — the same OperationAttributeResolver path the attribute-list endpoints use, which dereferences a stored authority/ra-profile SECRET reference (e.g. the authority's OAuth-client secret) into inline content on the connector request. Unlike listRaProfile/Issue/Revoke/RegisterAttributes (:164/:173/:182/:212), pollStatus does NOT wrap its connector response in the contained(...) guard (:221): it returns resp.getMeta()/getCertificateData()/getReason() straight into StatusPollResult (:295), and meta is a connector-controlled surface that reaches the operator-visible status record. The javadoc carve-out justifying the exemption (ConnectorRequestAttributesBuilder.java:151-167 and the contained() javadoc at :215-218) names only issue/renew/revoke/register as 'response is certificate data, not a caller-reflected surface' — it says nothing about pollStatus, so a connector echoing a resolved secret into the poll meta would not be caught, and the asymmetry is undocumented. (Verification note: cancel was also flagged originally but its response is only a CancelResult enum — no connector meta/data is returned or persisted — so cancel is NOT a leak surface and is excluded here.) Two readings, please confirm which: (a) intentional carve-out extension — poll responses are non-reflective status/certificate data; if so, extend the contained()/dereferenceForConnectorRequest javadoc to name pollStatus explicitly so the omission is a documented decision. (b) genuine gap — the connector-echoed meta is operator-visible and should get the same containment; if so, route resp.getMeta() through outboundContainment.assertNoExpandedSecretOutbound as the list endpoints do. Non-blocking either way.

pom.xml
```diff
@@ -34,7 +34,7 @@
         <sonar.projectKey>ilm_core</sonar.projectKey>
         <sonar.sources>src/main/java</sonar.sources>
         <sonar.cpd.exclusions>src/main/java/com/otilm/core/dao/entity/**</sonar.cpd.exclusions>
-        <sonar.coverage.exclusions>src/main/java/com/otilm/core/dao/entity/**,src/main/java/com/otilm/core/api/**,src/main/java/com/otilm/core/util/converter/**</sonar.coverage.exclusions>
+        <sonar.coverage.exclusions>src/main/java/com/otilm/core/dao/entity/**,src/main/java/com/otilm/core/api/**,src/main/java/com/otilm/core/util/converter/**,src/main/java/db/migration/**</sonar.coverage.exclusions>
```
ivosh: **~ Minor** | Testing

pom.xml:37 adds `src/main/java/db/migration/**` to `sonar.coverage.exclusions`. The pre-existing exclusion list is entity/api/converter boilerplate only (`dao/entity/**`, `api/**`, `util/converter/**`), and no sibling OmniTrustILM/3KeyCompany Java repo excludes a migration tree (checked via `gh search code sonar.coverage.exclusions`: interfaces excludes generated stubs, common-credential-provider its provider pkg, scheduler its controllers — none exclude migrations). This is a new, broad convention: the same migration this PR adds contains real logic (role/user seeding, resource-action grant map, `DatabaseAuthMigration.seedResources/createRole/createUser`) that no test exercises (no test references `V202607031200__...`). Excluding the whole tree hides that logic — and every future logic-bearing migration — from the new-code coverage gate rather than the boilerplate-only intent of the existing entries. 

Question for the author: is a tree-wide migration exclusion the intended standing convention, or should logic-bearing migrations like this one keep coverage (e.g. via the static-kernel extraction pattern the repo CLAUDE.md recommends)?

src/main/java/com/otilm/core/logging/LoggingHelper.java
```diff
@@ -148,6 +150,33 @@ public static void clearActorInfo() {
         MDC.remove(LOG_ACTOR_NAME);
     }
 
+    private static final String[] ACTOR_KEYS = {LOG_ACTOR_TYPE, LOG_ACTOR_AUTH_METHOD, LOG_ACTOR_UUID, LOG_ACTOR_NAME};
```
ivosh: **. Nitpick** | Maintainability

LoggingHelper.java:153 introduces `private static final String[] ACTOR_KEYS` interleaved between clearActorInfo() (line 146) and snapshotActorInfo() (line 160). Every other private static final constant in this class is grouped at the top (the LOG_* block, lines 27-44), so this placement breaks the class's own established field-ordering convention (Google Java Style §3.4.2 — members grouped logically, and the neighbour pattern here is a single top constant block). Not a correctness issue — the field is private and read-only. Suggest moving it up beside the LOG_ACTOR_* constants it derives from. The utility-class shape is otherwise intact: the private constructor at line 46 is preserved and the new methods are static, matching the established shape.

src/main/java/com/otilm/core/service/handler/OperationAttributeResolver.java
```diff
@@ -0,0 +1,70 @@
...
+        this.connectorRequestAttributesBuilder = connectorRequestAttributesBuilder;
+    }
+
+    public List<RequestAttribute> resolveForConnectorRequestAsSystem(UUID connectorUuid, List<RequestAttribute> stored)
+            throws ConnectorException {
+        if (stored == null || stored.stream().noneMatch(OperationAttributeResolver::isInfrastructureReference)) {
+            // No infrastructure reference to resolve: skip the system-identity elevation (and its auth-service
+            // round-trip) and hand the connector the stored attributes unchanged, as it receives them for a
+            // reference-free authority.
+            return stored;
```
ivosh, line 55: **. Nitpick** | Maintainability

OperationAttributeResolver.java:48-52 short-circuits `if (stored == null || ...) return stored;`, so a null argument is returned as-is rather than normalized to an empty list (Effective Java Item 54: return an empty collection, not null). This is a pass-through/identity-on-null rather than manufacturing a null, and the sole real caller path (AuthorityProviderV3Adapter -> authorityAttributesFor/raProfileAttributesFor -> AttributeEngine.getRequestObjectDataAttributesContent at AttributeEngine.java:854, which collects a stream and never returns null) never actually passes null — so a resolved null cannot reach the wire DTO setters today.

Recommendation: do not confuse the readers - normalize to `List.of()`.

src/main/java/com/otilm/core/util/AuthHelper.java
```diff
@@ -96,6 +103,47 @@ public void authenticateAsSystemUser(String username) {
         logger.debug("User with username '{}' has been successfully authenticated as system user proxy.", authUserInfo.getUsername());
     }
 
+    /**
```
ivosh, line 106: **~ Minor** | Documentation

The second paragraph of the runAsSystem Javadoc (AuthHelper.java:112-118) is a single ~60-word sentence chaining a colon, a semicolon, an em-dash and a nested parenthetical: "Afterwards both are put back exactly as the caller left them: the caller's Authentication is restored, or the holder is cleared when the caller had none (so no system principal lingers on a principal-less pooled thread - the async status-poll listener runs with no SecurityContext); and the actor MDC snapshot is restored, so an operator/protocol caller keeps its own audit attribution and an actor-less caller is left actor-less." It fuses two independent restoration concerns (the SecurityContext lifecycle and the actor-MDC lifecycle) with the subject lost mid-clause, and there are no bold lead-ins. Restructure to mirror the dereferenceForConnectorRequest exemplar: keep the one-line summary sentence, then split the restoration guarantees into two short bolded paragraphs, e.g. `<p><b>SecurityContext:</b> a fresh empty context is installed for the elevation window; on exit the caller's Authentication is restored, or the holder is cleared if the caller had none (so no system principal lingers on a pooled thread).` and `<p><b>Actor MDC:</b> the caller's actor snapshot is restored on exit, so an operator/protocol caller keeps its own audit attribution and an actor-less caller stays actor-less.` Cap each sentence at ~20 words.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
```diff
@@ -168,24 +172,22 @@ private static List<BaseAttributeContentV3<?>> resourceElements(RequestAttribute
...
-    // DEFERRED to the operation-path follow-up — do NOT wire the AuthorityProviderV3Adapter callsites here.
-    //
-    // Operation mode must resolve through a fenced loader carrying NO @ExternalAuthorization and performing NO
-    // setAuthentication (it runs under the operation path's existing principal). The live operation primitive is
-    // the public ConnectorRequestAttributesBuilder; safety rests on extending the ArchUnit fence to cover the
-    // operation entrypoint, NOT on a new unguarded twin. Until that follow-up:
-    //   - no expandViaOperationLoader is exposed,
-    //   - no adapter callsite is changed,
-    //   - the v2-adapter-does-not-expand / noDoubleExpansion / protocol-principal regression tests live there.
+    // Operation mode is NOT implemented in this fenced expander — by design. The v3 operation path resolves an
```
ivosh: **. Nitpick** | Documentation

The trailing comment block (AttributeReferenceExpander.java:185-192) is a 7-line inline wall that largely duplicates the class Javadoc's own "Operation mode lives outside this class" section, and it over-documents neighbours: most of it describes what OperationAttributeResolver and AuthorityProviderV3Adapter do (system-identity elevation, operation-level authorization) rather than what this class does. The core sentence "...resolves an authority's own infrastructure references through OperationAttributeResolver (called from AuthorityProviderV3Adapter), which elevates to the platform's attribute-content-resolver system identity for the dereference only - authorized at the operation level, not per acting caller" is ~35 words chaining an em-dash and nested parenthetical, and the block has no structural lead-ins. Since the class Javadoc already covers this in a well-structured `<h2>`, trim this to the one fact that belongs at the code site: e.g. "// Operation mode is deliberately not implemented here (see class Javadoc). The only piece still deferred is nested-blob resolution via attributesOf(), above." Drop the re-description of OperationAttributeResolver/AuthHelper.

src/main/java/com/otilm/core/service/handler/OperationAttributeResolver.java
```diff
@@ -0,0 +1,70 @@
...
+import com.otilm.api.model.common.attribute.common.content.AttributeContentType;
+import com.otilm.core.attribute.engine.ConnectorRequestAttributesBuilder;
+import com.otilm.core.util.AuthHelper;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.List;
+import java.util.UUID;
+
+/**
```
ivosh, line 17: **. Nitpick** | Documentation

The middle paragraph of the class Javadoc (OperationAttributeResolver.java:22-27) packs four distinct concerns into one `<p>` with no bold lead-ins: which system identity is used, who the acting callers are, the trust-model justification, and the scope/restoration of the elevation. Its first sentence is ~35 words with a nested parenthetical: "...so an authority's own infrastructure secret resolves without gating on the acting caller - an operator, a low-privilege protocol robot (ACME/SCEP/CMP), or the principal-less async status-poll thread." The `<p>` structure is a good start; finish it to the dereferenceForConnectorRequest bar by giving each concern its own bolded lead-in and ~20-word sentences, e.g. `<p><b>Authorization:</b> the dereference runs under the attribute-content-resolver system identity (via AuthHelper.runAsSystem), authorized at the operation level, not per acting caller - so it works for operators, protocol robots, and the principal-less status-poll thread alike.` and `<p><b>Scope:</b> the elevation covers the dereference only; runAsSystem restores the caller's context so the rest of the operation runs as the caller.`

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -334,11 +367,26 @@ private List<RequestAttribute> issueAttributesFor(Certificate cert, AuthorityIns
                         .build());
     }
 
-    private List<RequestAttribute> authorityAttributesFor(AuthorityInstanceReference authority) {
-        return attributeEngine.getRequestObjectDataAttributesContent(
+    /**
```
ivosh, line 377: **. Nitpick** | Documentation

The authorityAttributesFor Javadoc (AuthorityProviderV3Adapter.java:370-377) is one unbroken paragraph with no `<p>` grouping or lead-ins, fusing three concerns: what/why (v3 stateless, references can't self-resolve), the auth model (system identity, operation-level authorization, list of caller types), and the v2 contrast. Its middle sentence is ~35 words with an em-dash and a nested parenthetical, and the auth-model half restates what OperationAttributeResolver's own Javadoc already says (neighbour over-documentation). Restructure to a one-line summary plus per-concern paragraphs, e.g. `<p>` opening "v3 connectors are stateless, so stored authority/ra-profile attributes carry CREDENTIAL/RESOURCE (incl. SECRET) references the connector cannot resolve itself." then `<p><b>Resolution:</b> Core inlines them via OperationAttributeResolver (see its Javadoc for the system-identity trust model).` then `<p><b>v2:</b> unchanged - the base adapter's non-dereferencing helpers still serve the stateful v2 path.` Defer the caller-type enumeration to the resolver's Javadoc rather than repeating it here.

src/main/java/db/migration/V202607031200__CreateAttributeContentResolverUserAndPermissions.java
```diff
@@ -0,0 +1,73 @@
...
+import com.otilm.core.util.DatabaseMigration;
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+
+import java.util.ArrayList;
+import java.util.EnumMap;
+import java.util.List;
+import java.util.Map;
+
+/**
```
ivosh, line 16: **. Nitpick** | Documentation

The second paragraph of the migration Javadoc (V202607031200...java:24-31) is a wall with no lead-ins that fuses three distinct rationales: (1) why the role covers every referenceable kind (fail-safe for currently-unguarded AUTHORITY/ENTITY/LOCATION reads), (2) why CONNECTOR:DETAIL and VAULT_PROFILE:MEMBERS are transitively needed by a SECRET dereference, and (3) why the granted pairs must be seeded before the role is created (catalog sync runs only after Flyway). Each is a separate ~35-40-word sentence chaining em-dashes and parentheticals, so the safety-critical seed-ordering fact is buried last. 

The first paragraph's summary line is fine; break the second into bolded per-rationale paragraphs mirroring the exemplar, e.g. `<p><b>Broad coverage (fail-safe):</b> ...`, `<p><b>Transitive grants:</b> CONNECTOR:DETAIL and VAULT_PROFILE:MEMBERS are required by a SECRET dereference...`, `<p><b>Seed order:</b> pairs are seeded before the role because the auth service rejects unknown resource/actions and catalog sync runs only after Flyway.`

ivosh (changes requested):

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
lubomirw: Went with option (a): documented the carve-out in contained()/dereferenceForConnectorRequest rather than scanning meta. The connector already holds the resolved secret (echo is no new disclosure to an untrusted party), and unlike the read-only list fetch, failing closed after the connector has committed the operation would diverge Core's state from the connector's. Done in 693c8c51b.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
lubomirw: Same carve-out, now documented to name pollStatus explicitly (its connector-owned meta; cancel returns no connector content). Option (a), not containment — failing closed on a poll/post-commit response isn't safe. Done in 693c8c51b.

pom.xml
lubomirw: Narrowed the exclusion from db/migration/** to this one integration-only migration, so it is not a standing convention — future logic-bearing migrations should follow CLAUDE.md's static-kernel extraction rather than auto-escape coverage. Done in 693c8c51b.

src/test/java/com/otilm/core/integration/attribute/engine/AttributeReferenceExpanderITest.java
lubomirw: Added the ConnectorException-unwrapped propagation test (OperationAttributeResolverTest). The positive end-to-end SECRET-resolve needs a live vault connector, so it stays a unit-level proof for now — noted as a limitation, not faked. 693c8c51b.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
lubomirw: Fixed in 693c8c51b — retargeted to "a future nested-resolution change" to match lines 41/59/179.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
lubomirw: Fixed in 693c8c51b — dropped the hardcoded @ExternalAuthorization(<KIND>, DETAIL) detail; the per-kind gate now points to CallerAuthorizedReferenceLoaderRegistry.

src/main/java/com/otilm/core/logging/LoggingHelper.java
lubomirw: Fixed in 693c8c51b — moved ACTOR_KEYS up beside the LOG_ACTOR_* constants.

src/main/java/com/otilm/core/service/handler/OperationAttributeResolver.java
lubomirw: Fixed in 693c8c51b — a null argument is now normalized to List.of().

src/main/java/com/otilm/core/util/AuthHelper.java
lubomirw: Fixed in 693c8c51b — split into bolded "SecurityContext:" and "Actor MDC:" paragraphs.

src/main/java/com/otilm/core/attribute/engine/AttributeReferenceExpander.java
lubomirw: Fixed in 693c8c51b — trimmed to a one-line pointer to the class Javadoc.

src/main/java/com/otilm/core/service/handler/OperationAttributeResolver.java
lubomirw: Fixed in 693c8c51b — added bolded "Authorization:" / "Scope:" / "Placement:" lead-ins.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
lubomirw: Fixed in 693c8c51b — restructured with "Resolution:" / "v2:" lead-ins; caller-type enumeration deferred to the resolver's Javadoc.

src/main/java/db/migration/V202607031200__CreateAttributeContentResolverUserAndPermissions.java
lubomirw: Fixed in 693c8c51b — split into bolded "Broad coverage:" / "Transitive grants:" / "Seed order:" paragraphs.

src/main/java/com/otilm/core/attribute/engine/CallerAuthorizedReferenceLoaderRegistry.java
```diff
@@ -13,12 +14,16 @@
...
- * Kinds absent from the map (CERTIFICATE, SECRET) have no caller-side connector-consumable blob to expand and are
- * passed through as plain references by {@link AttributeReferenceExpander}. SECRET's inline content is the
- * connector's responsibility, not Core's to expand; a credential's stored material reaches the wire only through
- * the CREDENTIAL loader below, never by re-fetching a user-typed SECRET value.
+ * CREDENTIAL/AUTHORITY/ENTITY/LOCATION resolve via {@code <KIND>:DETAIL}; SECRET resolves via
+ * {@code SECRET:GET_SECRET_CONTENT} plus vault-profile membership (the established secret-content gate). SECRET is
+ * dereferenced because a <strong>stored</strong> SECRET reference carried in a scope blob (for example an
+ * authority's OAuth-client secret) is unusable to a stateless connector as a bare reference — Core must resolve it.
+ * This is distinct from a <strong>user-typed</strong> secret in {@code currentAttributes}, which is an inline
+ * first-delivery value (not a {@code ResourceObjectContent} reference) and so is never walked by the expander.
```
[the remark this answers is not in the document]
lubomirw: Fixed in 84b400089 — the getAuthorizedObjectAttributes contract javadoc no longer hardcodes DETAIL; it now describes the per-kind object-read gate (DETAIL for AUTHORITY/ENTITY/LOCATION; GET_SECRET_CONTENT + vault-profile membership for SECRET), matching the registry doc and the actual SECRET loader.

ivosh (approved):

PR #1711, opened 2026-07-02
Fix information leak in CMP error messages

### LINKED ISSUE #1635: Fix CMP PKIFreeText information leak
### Description
`CmpServiceImpl.validateProfile` forwards raw exception text into the CMP `PKIFreeText` response, which can leak internal runtime detail (SQL fragments, internal identifiers, upstream error text). Gate the forwarded text on the controlled domain-exception type. Part of #1572 (epic OmniTrustILM/ilm#131).

### Scope — what ships
- Gate the forwarded text: `e instanceof CmpProcessingException ? e.getMessage() : "<safe placeholder>"`, mirroring `RevocationMessageHandler`. No behavior change beyond the message shaping.

### Dependencies
- None — independent, mergeable immediately (no interfaces / phase dependency).

### Acceptance Criteria
- Leak-path test proving non-domain exception text is not forwarded; new-code coverage ≥80%; SonarCloud gate clean.

### Review focus
No SQL / stack-frame / internal identifier reaches CMP `PKIFreeText`.

ivosh:


src/main/java/com/otilm/core/service/cmp/impl/CmpServiceImpl.java
```diff
@@ -1,482 +1,503 @@
...
+            case V2 -> new CmpConfigurationContext(cmpProfile, raProfile, pkiRequest,
+                    certificateKeyServiceImpl, issueAttributes, revokeAttributes);
+            /*rfc9483*/
+            case V3 -> throw new UnsupportedOperationException("not implemented");
+        };
+
+        try {
+            PKIMessage pkiResponse;
+            int bodyType = pkiRequest.getBody().getType();
+            validateProfile(tid, bodyType, profileName);
```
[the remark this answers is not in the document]
ivosh: Fixed in b8b709fa. Moved profile validation ahead of the `cmpProfile.getVariant()` switch; on failure it now returns an unprotected shaped error via `e.toPKIBody()`.

src/main/java/com/otilm/core/service/cmp/impl/CmpServiceImpl.java
```diff
@@ -208,6 +209,16 @@ public ResponseEntity<byte[]> handlePost(String profileName, byte[] request) thr
 
         // -- (processing) part
         init(profileName);
+        int bodyType = pkiRequest.getBody().getType();
+
+        // Profile validation must run before cmpProfile is dereferenced below.
+        try {
+            validateProfile(tid, bodyType, profileName);
```
lubomirw, line 216: **[Medium]** | Error handling

Moving `validateProfile` into `try { … } catch (CmpBaseException e)` narrows the net vs. the old flow, where it ran inside the main `try { … } catch (Exception e)`. `validateCmpProfile` throws a plain `IllegalArgumentException` on a cert-content parse failure (SIGNATURE profiles — the *"This should not happen"* branch), and `certificateService.getCertificateChain(...)` can raise a runtime/DB exception — neither is a `CmpBaseException`, so both now **escape `handlePost`** to Spring's default handler (a non-CMP 500) instead of a well-formed CMP error. Reachability is low (SIGNATURE profiles with a cert-chain runtime error; `SHARED_SECRET` profiles only throw `CmpBaseException`), but it's the exact exception class this PR aims to shape.

_Suggested:_ add a catch-all safety net at the call site so these route through the same shaping:

```java
} catch (CmpBaseException e) {
    return errorResponse(tid, logPrefix, requestAsString, "profile validation", e,
            PkiMessageError.unprotectedMessage(pkiRequest.getHeader(), e.toPKIBody()));
} catch (Exception e) {
    return errorResponse(tid, logPrefix, requestAsString, "profile validation", e,
            safeUnprotectedError(pkiRequest.getHeader(), e));
}
```

<sub>flagged by Copilot CLI · verified in source + against pre-PR <code>main</code></sub>

src/main/java/com/otilm/core/service/cmp/impl/CmpServiceImpl.java
```diff
@@ -208,6 +209,16 @@ public ResponseEntity<byte[]> handlePost(String profileName, byte[] request) thr
...
         // -- (processing) part
         init(profileName);
+        int bodyType = pkiRequest.getBody().getType();
+
+        // Profile validation must run before cmpProfile is dereferenced below.
+        try {
+            validateProfile(tid, bodyType, profileName);
+        } catch (CmpBaseException e) {
+            return errorResponse(tid, logPrefix, requestAsString, "profile validation", e,
+                    PkiMessageError.unprotectedMessage(pkiRequest.getHeader(), e.toPKIBody()));
```
lubomirw, line 219: **[Low]** | Correctness

Because `validateProfile` now runs before the `ConfigurationContext` switch, a profile that *resolves* but is disabled/misconfigured returns an **unprotected** error here (`PkiMessageError.unprotectedMessage(...)`), whereas the old flow reached the downstream `catch (CmpBaseException)` and produced a **protected** `PkiMessageBuilder` response. RFC 4210 permits unprotected error messages, so this is likely intentional — flagging to confirm.

_Suggested:_ confirm the change is intended and note it in the PR description; consider a test that locks in protected-vs-unprotected for a resolved-but-disabled profile.

src/test/java/com/otilm/core/service/cmp/impl/CmpServiceImplHandlePostTest.java
```diff
@@ -0,0 +1,152 @@
...
+        // when
+        ResponseEntity<byte[]> response = service.handlePost(PROFILE_NAME, revocationRequest());
+
+        // then: the shaped domain message reaches the wire; the response is a well-formed CMP error
+        assertThat(response.getBody()).isNotNull();
+        assertThat(wireStatusText(response.getBody())).contains("Requested CMP Profile not found");
+    }
+
+    @Test
+    void handlePost_replacesRawMessageWithGenericDetail_whenProcessingThrowsNonDomainException() throws Exception {
```
lubomirw, line 87: **[Low]** | Testing

`cmpTransactionService.findByTransactionId(...)` is mocked to return an empty list, so `handleTrxError`'s `if (!trx.isEmpty())` branch never runs. The refactor funnels the transaction-fail logic (`state=FAILED` + `save(...)`) through `errorResponse` for all three catch blocks, but that path isn't exercised by these tests.

_Suggested:_ add a case where `findByTransactionId` returns a `CmpTransaction` and `verify(cmpTransactionService).save(...)` with `CmpTransactionState.FAILED`, to lock in that the refactor preserved transaction-fail semantics.

src/test/java/com/otilm/core/service/cmp/impl/CmpServiceImplHandlePostTest.java
```diff
@@ -0,0 +1,152 @@
...
+                PKIHeader.CMP_2000,
+                new GeneralName(new X500Name("CN=user")),
+                new GeneralName(new X500Name("CN=ManagementCA")));
+        headerBuilder.setTransactionID(new byte[]{1, 2, 3, 4});
+        headerBuilder.setSenderNonce("12345".getBytes(StandardCharsets.UTF_8));
+        PKIBody body = CmpTestUtil.createRevocationBody(BigInteger.ONE);
+        return new PKIMessage(headerBuilder.build(), body).getEncoded();
+    }
+
+    private static String wireStatusText(byte[] response) {
```
lubomirw, line 258: **[Nit]** | Maintainability

The `wireStatusText` PKIFreeText-extraction helper (`instanceof ErrorMsgContent ? … : ((CertRepMessage)…).getResponse()[0].getStatus()`) is duplicated in both new test classes.

_Suggested:_ extract a shared `wireStatusText(PKIBody)` into `CmpTestUtil` (both tests already import it) to avoid drift.

lubomirw (commented):
## Multi-reviewer pass — 1 item worth a look, rest minor

Reviewed via four parallel reviewers (Copilot CLI, guided-pr-review, superpowers, hygiene). This is a solid, well-tested leak fix: the old catch-all `PkiMessageError.unprotectedMessage(header, e)` put raw `e.getMessage()` straight into the wire `PKIFreeText`, and `safeUnprotectedError` / `safeCmpDetail` close that — faithfully implementing the documented org idiom (*"Don't leak runtime details to the wire"*). Moving `validateProfile` ahead of the `cmpProfile.getVariant()` switch also fixes a latent NPE when the profile is unresolved. 👍

**One item worth addressing** (inline, `[Medium]`): unchecked exceptions from `validateProfile` can now escape `handlePost`. Two minor test/style notes are inline as well (`[Low]`/`[Nit]`).

Considered but **not** raised as blocking:
- **`safeCmpDetail` `instanceof` gate** — a `CmpBaseException` built via the 4-arg constructor carries its cause's message in `getMessage()`, so the gate isn't full sanitization in general. Not reachable at either current call site (the `validateProfile` catch only sees a cause-less `CmpConfigurationException`; the catch-all never receives a `CmpBaseException`), and it matches the sanctioned pattern — informational only.
- **Processing-path `catch (CmpBaseException)`** still forwards `e.toPKIBody()` unshaped, and **`handleTrxError`** can NPE on a null exception message / null transactionID — both are **pre-existing on `main`** and unchanged by this PR (the refactor only relocated the identical `handleTrxError` calls). Possible follow-ups, not part of this diff.

_Nothing here blocks; the `[Medium]` item is the only one I'd suggest resolving (or consciously accepting) before merge._

src/main/java/com/otilm/core/service/cmp/impl/CmpServiceImpl.java
ivosh: Fixed in 792cb78c. Added a `catch (Exception e)` catch-all after the `CmpBaseException` catch at the `validateProfile` call site, routing unchecked exceptions through `safeUnprotectedError(...)`. This covers both the SIGNATURE "This should not happen" `IllegalArgumentException` and any `getCertificateChain(...)` runtime/DB error — they now return a shaped, well-formed CMP error instead of escaping to Spring's default 500. As a bonus it closes a leak: that `IllegalArgumentException` embeds raw certificate content in its message, which `safeUnprotectedError` now sanitizes.

src/main/java/com/otilm/core/service/cmp/impl/CmpServiceImpl.java
ivosh: Confirmed intentional, and now locked in by a test in 792cb78c (`handlePost_returnsUnprotectedShapedError_whenResolvedProfileDisabled`).

Trace of the old vs. new flow for a resolved-but-disabled profile:
- **Old (main):** `configuration` was built from the switch *before* validation, so when `validateProfile` threw a `CmpBaseException` it was caught with a live `configuration` in scope and produced a **protected** `PkiMessageBuilder` response.
- **New:** `validateProfile` had to move ahead of `cmpProfile.getVariant()` to fix the NPE when the profile is *unresolved* (`cmpProfile == null`). At that point no `ConfigurationContext` exists yet, so the error is **unprotected**.

This is fine per RFC 4210 (§5.1.3 / §5.3.21 permit unprotected error messages), and arguably more correct — protecting an error with the credentials of a profile we just declared disabled/misconfigured is questionable, and a disabled SIGNATURE profile lacking a valid signing cert couldn't have protected the response anyway. Noted in the PR description.

src/test/java/com/otilm/core/service/cmp/impl/CmpServiceImplHandlePostTest.java
ivosh: Fixed in 792cb78c. Added `handlePost_failsTransaction_whenProfileValidationFails`: `findByTransactionId` returns a `CmpTransaction`, and the test verifies `cmpTransactionService.save(...)` is called with state `FAILED`, exercising the `handleTrxError` `if (!trx.isEmpty())` branch through `errorResponse`.

src/test/java/com/otilm/core/service/cmp/impl/CmpServiceImplHandlePostTest.java
ivosh: Fixed in 792cb78c. Extracted `wireStatusText(PKIBody)` as a `public static` helper in `CmpTestUtil`; both test classes now delegate to it (their `byte[]`/`PKIMessage`/`CmpBaseException` overloads are thin wrappers).

lubomirw (approved):

PR #1718, opened 2026-07-03
Reap orphaned PENDING_ISSUE certificates that have no status-poll row

### LINKED ISSUE #1716: Reap orphaned PENDING_ISSUE certificates that have no status-poll row
### Description

The synchronous (HTTP 200) issue/renew/rekey path transitions a certificate to `PENDING_ISSUE` before the blocking connector call. Because the action methods run without an ambient transaction (`@Transactional(NOT_SUPPORTED)`), the state-machine save auto-commits, so `PENDING_ISSUE` is durably persisted before the connector responds. If the process dies after that commit but before the terminal `ISSUED`/`FAILED` transition, the certificate is stranded in `PENDING_ISSUE`.

Unlike the asynchronous (202) path, the synchronous path creates no `certificate_status_poll` row, so `CertificateStatusPollSweeper` never re-drives it, and nothing else reconciles lifecycle state (`UpdateCertificateStatusTask` only updates validation status). A stranded certificate cannot be re-issued (`PENDING_ISSUE` is not an allowed issue from-state), so it requires manual intervention (operator cancel or DB fix).

Impact is low: a narrow process-crash window, single-certificate blast radius, synchronous-completing connectors only, and a manual recovery path exists. It is a crash-recovery ergonomics regression versus prior behaviour, where such a certificate stayed in the retriable `REQUESTED` state.

Surfaced during review of the CertificateStateMachine coverage slice (#1683). Part of the authority-provider-v3 epic (ilm#110).

### Definition of Done

- A scheduled sweep transitions `PENDING_ISSUE` certificates older than a configurable threshold **that have no poll row** to `FAILED` (re-driving is impossible on the sync path — the connector response is already lost), making them actionable/retriable.
- `PENDING_ISSUE` certificates that have a live poll row (the 202 async path) are left untouched.
- The sweep is multi-instance safe, reusing the advisory-lock/batch orchestration of the existing `CertificateStatusPollSweeper`.
- Covered by a test.

lubomirw:
## What

Adds a reaper that fails certificates left orphaned in `PENDING_ISSUE` by a crash on the **synchronous** (HTTP 200) issue/renew/rekey path. That path commits `PENDING_ISSUE` before the blocking connector call but, unlike the async (202) path, creates no `certificate_status_poll` row — so the poll sweep (driven by those rows) never re-drives it, and nothing else reconciles the lifecycle state. Such a certificate can't be re-issued (`PENDING_ISSUE` isn't an allowed issue from-state), so today it requires manual intervention.

## How

`PendingIssueReaper` runs as a **final phase of the existing `CertificateStatusPollSweeper.sweep()`** — no new scheduled task, config namespace, or advisory-lock key:

- Selects stale, poll-less `PENDING_ISSUE` certificates (`state = PENDING_ISSUE AND updated < now - stale-after AND NOT EXISTS(poll row)`, oldest-first) under the existing `PROVIDER_STATUS_POLL_SWEEP` advisory lock.
- Per certificate: pessimistic-lock (`findAndLockWithAssociationsByUuid`) + re-assert `PENDING_ISSUE` + `stateMachine.transition(FAILED, ISSUE, …)` in its own transaction — re-drive is impossible (the connector response is lost), so `FAILED` makes it actionable/retriable.
- Cross-node correctness rests on the per-certificate lock + re-assert; the advisory lock only avoids redundant selection.
- A **partial index** (`idx_certificate_pending_issue_reaper ON certificate (i_upd) WHERE state = 'PENDING_ISSUE'`) backs the recurring selection so the sweep does an index range scan, not a full table scan.

New config: `provider.status-poll.reap-stale-after` (default `PT30M`).

### Notes / deliberate choices

- **Staleness uses `updated` (`i_upd`)** — there is no dedicated "entered PENDING_ISSUE at" column; the state-machine's write stamps `i_upd` and a stuck certificate isn't mutated afterwards, so it approximates entry time. A later write only *delays* reaping, never triggers it early.
- **A reaped renew/rekey successor leaves its `PENDING` predecessor relation in place** (the full failure funnel that also clears relations is out of scope for this safety-net).

## Testing

- `PendingIssueReaperTest` (integration, real Postgres): stale + no poll row → `FAILED` + `ISSUE`/`FAILED` audit row; fresh, has-poll-row (202 async), and non-`PENDING_ISSUE` certificates left untouched.
- `PendingIssueReaperUnitTest`: under-lock re-assert skip; one certificate's failure doesn't abort the batch.
- `CertificateStatusPollSweeperTest`: every sweep invokes the reap phase.

Closes #1716


src/main/java/com/otilm/core/dao/repository/CertificateRepository.java
```diff
@@ -163,6 +163,22 @@ List<UUID> findCertificatesToCheckStatus(@Param("statusValidityEndTimestamp") Of
...
+    @Query("""
+            SELECT c.uuid FROM Certificate c
+                WHERE c.state = :state AND c.updated < :threshold
+                    AND NOT EXISTS (SELECT 1 FROM CertificateStatusPoll p WHERE p.certificateUuid = c.uuid)
+                ORDER BY c.updated ASC
+            """
+    )
+    List<UUID> findStalePendingIssueWithoutPollRow(@Param("state") CertificateState state,
+                                                   @Param("threshold") OffsetDateTime threshold,
+                                                   Pageable pageable);
```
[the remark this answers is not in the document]
lubomirw: Fixed in 92d16a140 — the JPQL now hard-codes `PENDING_ISSUE` via the SpEL enum literal (matching the sibling queries in this repo) and the `state` parameter is removed, so the method can only ever hit the partial index.

src/main/java/com/otilm/core/messaging/jms/listeners/poll/PendingIssueReaper.java
```diff
@@ -0,0 +1,106 @@
...
+    private boolean reapOne(UUID uuid) {
+        try {
+            return Boolean.TRUE.equals(transactionHandler.runInNewTransaction(() -> {
+                Certificate locked = certificateRepository.findAndLockWithAssociationsByUuid(uuid).orElse(null);
+                // Re-assert under the row lock: the certificate may have completed or been reaped by
+                // another node between selection and locking.
+                if (locked == null || locked.getState() != CertificateState.PENDING_ISSUE) {
+                    return false;
+                }
+                stateMachine.transition(locked, CertificateState.FAILED, CertificateEvent.ISSUE,
```
ivosh, line 102: **! Major** | Correctness

reapOne() (PendingIssueReaper.java:96) fails a stale PENDING_ISSUE certificate with only `stateMachine.transition(locked, FAILED, ISSUE, ...)`. But the sync issue/renew/rekey paths that put a certificate into PENDING_ISSUE (ClientOperationServiceImpl.issueCertificateAction:591, renewCertificateAction:950, rekeyCertificateAction:1165) do NOT fail through the SM alone — they call handleFailedOrRejectedEvent(), which additionally (ClientOperationServiceImpl.java:733-770): removes the cert from its locations, `certificateRelationRepository.deleteAll(certificate.getPredecessorRelations())`, and for RENEW/REKEY records a FAILED history entry against the predecessor certificate. CertificateStateMachine's own Javadoc (lines 28-30) states the SM 'does NOT drive the side effects (location cleanup, predecessor-relation deletion, dual-cert event history)... The caller remains responsible for those.' The reaper is such a caller and performs none of them.

  The setup (what's on disk before the crash):

  A renew (or rekey) is a two-step operation. The PENDING predecessor relation is created in an earlier, already-committed step — renewCertificateAction doesn't create it, it finds it and throws if it's missing:

  // ClientOperationServiceImpl.java:898
  CertificateRelation certificateRelation = certificateRelationRepository
      .findFirstBy...RelationTypeOrderByCreatedAtAsc(certificateUuid, CertificateRelationType.PENDING)
      .orElseThrow(...);

  So the relation predecessor(ISSUED) → successor(new cert) is already durable when the action runs.

  The crash window

  // renewCertificateAction, line 919 — runs in its OWN transaction and commits
  stateMachine.transition(certificate, PENDING_ISSUE, ISSUE, "Issuance in progress");
  // line 921 — the blocking connector HTTP call
  connectorApiFactory...renewCertificate(...);   // ← node crashes here (sync path)

  The enclosing method is @Transactional(NOT_SUPPORTED) and stateMachine.transition is @Transactional, so PENDING_ISSUE commits before the connector call. On the sync (200) path no certificate_status_poll row is ever
  written. A crash here therefore leaves, persisted: successor = PENDING_ISSUE + live PENDING predecessor relation + no poll row — precisely what the reaper's selection query targets, and it cannot tell a renew/rekey orphan
  apart from a pure-issue one.

  What the reaper does vs. what it should do:

  reapOne does only the bare state transition:

  // PendingIssueReaper.java:96
  stateMachine.transition(locked, FAILED, ISSUE, "Operation did not complete; reaped...");

  But the state machine's own Javadoc is explicit that this is not enough:

  ▎ CertificateStateMachine.java:28 — "SM does NOT drive the side effects (location cleanup, predecessor-relation deletion, dual-cert event history)… The caller remains responsible for those."

  Every other path that moves PENDING_ISSUE → FAILED discharges that responsibility:

  - handleFailedOrRejectedEvent:761 — certificateRelationRepository.deleteAll(certificate.getPredecessorRelations()) + predecessor RENEW/REKEY FAILED audit.
  - commitLocalCancel:2370 — same deletion, with a Javadoc that names this exact bug (line 2333): "it deletes predecessor relations to prevent dangling PENDING relations on the predecessor — a renew/rekey predecessor would
  otherwise still link to a now-FAILED successor."

  The reaper is the third such path and does none of it.


Action: route the reap through the same failure-handling side effects (delete predecessorRelations, remove from locations, and for renew/rekey audit the predecessor) rather than a bare SM transition — e.g. delegate to a shared failure handler, or at minimum delete predecessor relations under the lock. If the intent is that the reaper only ever handles pure-ISSUE orphans, that is not enforced: renew/rekey reach PENDING_ISSUE identically and are indistinguishable to the selection query.

src/test/java/com/otilm/core/messaging/jms/listeners/poll/PendingIssueReaperUnitTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+        assertThatCode(() -> reaper.reapStaleOrphans()).doesNotThrowAnyException();
+
+        // The healthy certificate is still reaped despite the earlier failure.
+        verify(stateMachine).transition(eq(healthyCert), eq(CertificateState.FAILED),
+                eq(CertificateEvent.ISSUE), anyString());
+    }
+
+    private void doThrowOnTransition(Certificate cert) {
+        org.mockito.Mockito.doThrow(new RuntimeException("lock timeout"))
```
ivosh: **. Nitpick** | Maintainability

src/test/java/com/otilm/core/messaging/jms/listeners/poll/PendingIssueReaperUnitTest.java:94 calls `org.mockito.Mockito.doThrow(...)` with the fully-qualified class name inline, while every other Mockito API in this file is statically imported (lines 26-29: lenient/never/verify/when, plus `doThrow`'s siblings). Google Java Style §3.3.1 says to import rather than fully-qualify. Add `import static org.mockito.Mockito.doThrow;` and drop the qualifier for consistency with the rest of the file. Nitpick only — test-scoped and behaviourally irrelevant.

src/main/java/com/otilm/core/messaging/jms/listeners/poll/PendingIssueReaper.java
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * Selects one bounded batch of stale, poll-less PENDING_ISSUE certificates under the shared advisory
+     * lock, then fails each in its own transaction outside the lock. One batch per sweep is sufficient:
+     * the sweep cadence drains any backlog, and a certificate is not a candidate until it has been stuck
+     * for {@code stale-after}.
+     */
+    public void reapStaleOrphans() {
+        List<UUID> candidates = transactionHandler.runInNewTransaction(() -> {
+            if (!clusterSynchronizer.tryLock(ClusterOperationSynchronizer.Operation.PROVIDER_STATUS_POLL_SWEEP)) {
```
ivosh, line 74: **. Nitpick** | Testing

reapStaleOrphans() returns an empty candidate list (and does no work) when clusterSynchronizer.tryLock(...) is false — PendingIssueReaper.java:68-70. The class Javadoc (PendingIssueReaper.java:36-40) makes this a documented contract ("the advisory lock only avoids redundant selection"), yet neither PendingIssueReaperUnitTest nor PendingIssueReaperTest exercises the false branch: the unit test stubs tryLock to true (PendingIssueReaperUnitTest.java:59-60) and never overrides it. A regression that ran the selection/reap regardless of the lock (e.g. moving the query out of the guard) would leave every existing test green while breaking the cross-node exclusion the Javadoc promises. Add a unit test asserting that when tryLock returns false, findStalePendingIssueWithoutPollRow is never called and no transition occurs.

ivosh (changes requested):

src/main/java/com/otilm/core/messaging/jms/listeners/poll/PendingIssueReaper.java
lubomirw: Fixed in 12bf0180d — `reapOne` now deletes predecessor relations under the lock (`certificateRelationRepository.deleteAll(locked.getPredecessorRelations())`), mirroring `handleFailedOrRejectedEvent`/`commitLocalCancel`: a no-op for a pure-issue orphan (no relation), and it drops the dangling PENDING relation for a renew/rekey orphan. Added an integration test asserting the relation is gone and the predecessor stays ISSUED.

Two side effects I deliberately did **not** mirror:
- **Location removal** — the reaper's targets crashed *before* `issueRequestedCertificate` ran, so the successor was never pushed to a location (push happens after issue); it would be a no-op, and doing it would add connector HTTP calls inside the reaper's locked transaction, against the sweep's design of never pinning the lock across connector round-trips.
- **Predecessor RENEW/REKEY audit** — the reaper cannot tell RENEW from REKEY (the relation is still PENDING; it is only flipped to RENEWAL/REKEY on successful issuance, which an orphan never reaches), so a guessed event would be misleading. The successor itself still gets its ISSUE/FAILED audit with the reap reason from the SM transition. Happy to add a generic predecessor entry if you want that observability.

src/test/java/com/otilm/core/messaging/jms/listeners/poll/PendingIssueReaperUnitTest.java
lubomirw: Fixed in 12bf0180d — added `import static org.mockito.Mockito.doThrow;` and dropped the fully-qualified name.

src/main/java/com/otilm/core/messaging/jms/listeners/poll/PendingIssueReaper.java
lubomirw: Added in 12bf0180d — `skipsAllWorkWhenAdvisoryLockNotAcquired` stubs `tryLock` to false and asserts `findStalePendingIssueWithoutPollRow` is never called and no transition occurs.

ivosh (approved):

PR #1721, opened 2026-07-03
Split services into internal and external interfaces - part 18

ivosh:
Split CryptographicKeyService (Mixed) into a flat CryptographicKeyExternalService (29 controller-callable methods) and CryptographicKeyInternalService (4 service-only methods, extends ResourceExtensionService). Six controller endpoints that were callable but unannotated (getSearchableFieldInformation and the *KeyItems bulk operations) move onto the external interface with @ExternalAuthorizationMissing, documenting a pre-existing authorization gap for follow-up triage.

Rename the pure-external TsaService to TsaExternalService and mark its two TSP protocol methods @ExternalAuthorizationMissing. Widen ExternalServiceAuthorizationArchTest @AnalyzeClasses to com.otilm.core.signing.tsa so the contract enforces on it.

Freeze store grows 7 -> 15 (8 new @ExternalAuthorizationMissing entries).


src/main/java/com/otilm/core/service/CryptographicKeyInternalService.java
```diff
@@ -0,0 +1,47 @@
...
+import java.security.PublicKey;
+import java.util.UUID;
+
+public interface CryptographicKeyInternalService extends ResourceExtensionService {
+
+    /**
+     * Function to get the key based on the sha 256 key fingerprint
+     *
+     * @param fingerprint SHA 256 fingerprint of the key
+     * @return Cryptographic Key UUID
```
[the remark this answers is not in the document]
ivosh: Fixed in 5eabd722.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -737,6 +741,7 @@ public void destroyKey(List<String> uuids) throws ConnectorException, NotFoundEx
     }
 
     @Override
+    @ExternalAuthorizationMissing
```
lubomirw, line 744: **[Low]** | Security

The bulk key-item operations now surfaced on `CryptographicKeyExternalService` — `destroyKeyItems`, `compromiseKeyItems`, `enableKeyItems`, `disableKeyItems`, `updateKeyItemUsages` (and read-only `getSearchableFieldInformation`), plus the two `TsaServiceImpl` TSP methods — are frozen as `@ExternalAuthorizationMissing`, i.e. reachable via the controller with no service-layer object-level `@ExternalAuthorization`. This is pre-existing (the base commit had no auth annotation on them either), so this mechanical split introduces no regression — it only documents/freezes the gap. **Not blocking this PR.**

_Suggested:_ Track a follow-up to add real `@ExternalAuthorization` (resource=CRYPTOGRAPHIC_KEY) to the destructive bulk key-item ops — `destroyKeyItems`/`compromiseKeyItems` are irreversible — or route them through an authorized parent-key path, then shrink the `archunit_store` baseline.

<sub>flagged by guided-pr-review, superpowers, Copilot CLI (3/4 reviewers)</sub>

src/test/java/com/otilm/core/architecture/ExternalServiceAuthorizationArchTest.java
```diff
@@ -31,7 +31,7 @@
 import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
 import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.noMethods;
 
-@AnalyzeClasses(packages = "com.otilm.core.service", importOptions = ImportOption.DoNotIncludeTests.class)
+@AnalyzeClasses(packages = {"com.otilm.core.service", "com.otilm.core.signing.tsa"}, importOptions = ImportOption.DoNotIncludeTests.class)
```
lubomirw: **[Nit]** | Maintainability

`@AnalyzeClasses` now uses a hand-maintained package set `{"com.otilm.core.service", "com.otilm.core.signing.tsa"}` — this PR had to add `com.otilm.core.signing.tsa` because `TsaServiceImpl` became an external service in a new package. Any future `*ExternalService` in yet another package will silently escape the authorization-annotation contract until someone remembers to extend this list.

_Suggested:_ Scope the scan to the common root `com.otilm.core` (or otherwise derive it) so the contract auto-covers new external services without a manual edit each split.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
ivosh: Agreed. They will be tackled in a follow-up PR.

src/test/java/com/otilm/core/architecture/ExternalServiceAuthorizationArchTest.java
ivosh: Fixed in f73f72e5. Scoped @AnalyzeClasses to the common root `com.otilm.core` (matching PlatformExceptionTest / ConnectorApiClientArchTest / TransactionalBoundaryArchTest / NativeQuerySchemaArchTest), so new *ExternalService classes in any package are auto-covered without a manual package-list edit.

lubomirw (approved):

PR #1723, opened 2026-07-03
Lock the operator issue path against concurrent double-issue

### LINKED ISSUE #1687: Lock the operator issue path against concurrent double-issue
### Description

`issueExistingCertificate` (and the operator issue path generally) reads a certificate via the unlocked `getCertificateEntity`, asserts `state == REGISTERED`, attaches the operator CSR, and enqueues an ISSUE action — all under `@Transactional(NOT_SUPPORTED)`, so each read is lock-free. Two concurrent calls on the same REGISTERED placeholder can both pass the state guard, both attach a CSR (the second overwriting the first in `addCertificateRequestToExisting`, whose own `state == REGISTERED` guard is only a cheap secondary check that does not serialize), and both enqueue an ISSUE action — yielding a double issuance for one certificate.

Independently flagged in the PR #1681 multi-reviewer pass and confirmed by @ivosh's review. Deferred from #1681 deliberately: the operator issue path has never locked, the connector call lives in the async consumer (outside any lock), and this is a contained change better reviewed on its own.

### Definition of Done

- The REGISTERED read-attach-enqueue in `issueExistingCertificate` runs through a short explicit transaction using `findAndLockWithAssociationsByUuid` (PESSIMISTIC_WRITE) with the `state == REGISTERED` assertion held under the lock, mirroring `manuallyConfirmRevoke` / `commitLocalCancel`.
- A second concurrent issue on the same placeholder is rejected (or no-ops) rather than producing a second ISSUE action or overwriting the attached CSR.
- The `addCertificateRequestToExisting` `state == REGISTERED` guard stays as the cheap secondary check (no double lock).
- A test exercises the concurrent path (or asserts the lock + under-lock assertion) per the codebase's race-test conventions.
- The connector HTTP call remains outside the lock (it runs in the async consumer).

lubomirw:
## What

Under concurrency (operator double-click / retry / two admins), one certificate could reach an **ISSUED outcome more than once**. `Certificate` has no `@Version` and no `@DynamicUpdate`, and the issue-completion paths read unlocked and `save` full-column, so two racing paths could both reach the connector, or a stale write could clobber a committed claim.

Closes #1687 (sub-issue of the authority-provider-v3 epic ilm#110).

## How

Serialize the three issue-completion paths with the shipped **pessimistic-lock + re-assert in a short transaction, connector call outside the lock** pattern (`manuallyConfirmRevoke`, `PendingIssueReaper`). The lock (`findAndLockWithAssociationsByUuid`, `SELECT … FOR UPDATE`) is always released before any connector HTTP call.

1. **`issueCertificateAction`** (async ISSUE consumer, queue concurrency 10) — read + guards + `REGISTERED→PENDING_ISSUE` claim now run under the row lock, committed before the connector call. A second concurrent action blocks, re-reads `PENDING_ISSUE`, and **benign-skips**. The skip is deliberately **outside** the connector-failure catch: `PENDING_ISSUE→FAILED` is a *legal* transition, so failing there would mark the concurrent winner's in-flight issuance FAILED. Data needed after the claim (connector ids, sign request, location ids) is captured inside the lock since the entity detaches on commit; the failure path re-reads a managed entity for its collection cleanup.
2. **`addCertificateRequestToExisting`** — its read is now the locking finder (authorization preserved). The existing `state == REGISTERED` re-assert becomes authoritative: a racing CSR attach blocks, re-reads a changed state, and throws — closing the full-column write-back that would otherwise reset `state=REGISTERED` over a committed claim.
3. **`manuallyIssueCertificate`** (operator upload-finalize) — the `PENDING_ISSUE→ISSUED` finalize runs under the lock with a re-assert; the connector identify runs before, the location push/cancel after.

### Scope / determinations
- **renew/rekey are out of scope** — each call creates a *new successor* (a two-certs shape), not one cert issued twice.
- **No `@Version` needed** — with these locks every issue-completion state write is lock-guarded and re-asserts; `@Version` would only guard unrelated incidental edits (a separate, repo-wide concern).

## Testing

227 tests green (`ClientOperationServiceV2Test` + `CertificateServiceTest`), including new deterministic guards for each path:
- `issueCertificateAction` benign-skip on an already-`PENDING_ISSUE` cert — asserts the connector is **not** called (`verify(0)`) and the cert is not driven to FAILED.
- `addCertificateRequestToExisting` rejects a non-`REGISTERED` cert (the authoritative re-assert).
- `manuallyIssueCertificate` locked re-assert — stubs only the locked read to return `ISSUED` (simulating a concurrent winner the outer guard can't observe) and asserts rejection + the real row stays `PENDING_ISSUE`.

The existing happy-path issue/renew/upload tests are the regression net for the transaction/lazy-loading restructure.

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
@@ -1140,6 +1140,13 @@ public class CertificateServiceImpl implements CertificateService, AttributeReso
                 SecuredParentUUID.fromUUID(raProfileUuid));
         Certificate certificate = certificateRepository.findAndLockWithAssociationsByUuid(certificateUuid)
                 .orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
+        // The RA profile can change concurrently (switchRaProfile is allowed on REGISTERED and takes no row
+        // lock) between the authorization above and this locked read. Authorization was evaluated against
+        // raProfileUuid, so reject if the locked row now belongs to a different profile — otherwise a caller
+        // authorized on the old profile could attach a CSR to a certificate under one they do not control.
+        if (!raProfileUuid.equals(certificate.getRaProfileUuid())) {
+            throw new ValidationException("Certificate's RA profile changed during authorization; retry the operation. Certificate: %s".formatted(certificate.toStringShort()));
+        }
         // Defense-in-depth: a CSR is attached only while completing a registered placeholder. The sole
         // caller (issueExistingCertificate) already gates on this, but guard here too so this public method
         // cannot overwrite the request of an ISSUED / REQUESTED / pending certificate.
--- a/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
@@ -1129,10 +1129,15 @@ public class CertificateServiceImpl implements CertificateService, AttributeReso
         if (signRequest.getFormat() == null) {
             throw new CertificateRequestException("A certificate signing request format (PKCS10 or CRMF) is required");
         }
-        // Authorize on an unlocked read first — evaluateCertificateRaProfilePermissions makes an external
-        // OPA call and must not run while holding the row lock — then take the lock (SELECT ... FOR UPDATE)
-        // so a concurrent attach blocks here and the state re-assert below is authoritative.
-        getCertificateEntity(SecuredUUID.fromUUID(certificateUuid));
+        // Authorize before locking WITHOUT managing the entity: the RA-profile check makes an external OPA
+        // call (must not run while holding the row lock), and loading the full entity first would make the
+        // locking query below return the already-managed, stale-state instance — defeating the under-lock
+        // re-assert. Fetch only the RA-profile UUID (a scalar projection), authorize, then take the lock so
+        // findAndLockWithAssociationsByUuid loads fresh state under SELECT ... FOR UPDATE.
+        UUID raProfileUuid = certificateRepository.findRaProfileUuidByUuid(certificateUuid)
+                .orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
+        raProfileService.evaluateCertificateRaProfilePermissions(SecuredUUID.fromUUID(certificateUuid),
+                SecuredParentUUID.fromUUID(raProfileUuid));
         Certificate certificate = certificateRepository.findAndLockWithAssociationsByUuid(certificateUuid)
                 .orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
         // Defense-in-depth: a CSR is attached only while completing a registered placeholder. The sole
--- a/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
@@ -1129,7 +1129,13 @@ public class CertificateServiceImpl implements CertificateService, AttributeReso
         if (signRequest.getFormat() == null) {
             throw new CertificateRequestException("A certificate signing request format (PKCS10 or CRMF) is required");
         }
-        Certificate certificate = getCertificateEntity(SecuredUUID.fromUUID(certificateUuid));
+        // Locked read (SELECT ... FOR UPDATE): a concurrent attach on the same certificate blocks here
+        // instead of racing an unlocked read, so the state re-assert below is authoritative rather than
+        // a check the other transaction can still slip past.
+        SecuredUUID securedCertificateUuid = SecuredUUID.fromUUID(certificateUuid);
+        Certificate certificate = certificateRepository.findAndLockWithAssociationsByUuid(certificateUuid)
+                .orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
+        raProfileService.evaluateCertificateRaProfilePermissions(securedCertificateUuid, SecuredParentUUID.fromUUID(certificate.getRaProfileUuid()));
         // Defense-in-depth: a CSR is attached only while completing a registered placeholder. The sole
         // caller (issueExistingCertificate) already gates on this, but guard here too so this public method
         // cannot overwrite the request of an ISSUED / REQUESTED / pending certificate.
--- a/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
@@ -1129,13 +1129,12 @@ public class CertificateServiceImpl implements CertificateService, AttributeReso
         if (signRequest.getFormat() == null) {
             throw new CertificateRequestException("A certificate signing request format (PKCS10 or CRMF) is required");
         }
-        // Locked read (SELECT ... FOR UPDATE): a concurrent attach on the same certificate blocks here
-        // instead of racing an unlocked read, so the state re-assert below is authoritative rather than
-        // a check the other transaction can still slip past.
-        SecuredUUID securedCertificateUuid = SecuredUUID.fromUUID(certificateUuid);
+        // Authorize on an unlocked read first — evaluateCertificateRaProfilePermissions makes an external
+        // OPA call and must not run while holding the row lock — then take the lock (SELECT ... FOR UPDATE)
+        // so a concurrent attach blocks here and the state re-assert below is authoritative.
+        getCertificateEntity(SecuredUUID.fromUUID(certificateUuid));
         Certificate certificate = certificateRepository.findAndLockWithAssociationsByUuid(certificateUuid)
                 .orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
-        raProfileService.evaluateCertificateRaProfilePermissions(securedCertificateUuid, SecuredParentUUID.fromUUID(certificate.getRaProfileUuid()));
         // Defense-in-depth: a CSR is attached only while completing a registered placeholder. The sole
         // caller (issueExistingCertificate) already gates on this, but guard here too so this public method
         // cannot overwrite the request of an ISSUED / REQUESTED / pending certificate.
--- a/src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
@@ -539,37 +539,72 @@ public class ClientOperationServiceImpl implements ClientOperationExternalServic
             certificateService.checkIssuePermissions();
         }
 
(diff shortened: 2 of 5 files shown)


src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1129,7 +1129,12 @@ public void addCertificateRequestToExisting(UUID certificateUuid, ClientCertific
         if (signRequest.getFormat() == null) {
             throw new CertificateRequestException("A certificate signing request format (PKCS10 or CRMF) is required");
         }
-        Certificate certificate = getCertificateEntity(SecuredUUID.fromUUID(certificateUuid));
+        // Authorize on an unlocked read first — evaluateCertificateRaProfilePermissions makes an external
+        // OPA call and must not run while holding the row lock — then take the lock (SELECT ... FOR UPDATE)
+        // so a concurrent attach blocks here and the state re-assert below is authoritative.
+        getCertificateEntity(SecuredUUID.fromUUID(certificateUuid));
+        Certificate certificate = certificateRepository.findAndLockWithAssociationsByUuid(certificateUuid)
+                .orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
```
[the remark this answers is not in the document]
lubomirw: Fixed in 28364057b — good catch. Authorization now uses a scalar `raProfileUuid` projection (`findRaProfileUuidByUuid`) instead of loading the full entity, so nothing is managed before the locking read; `findAndLockWithAssociationsByUuid` then loads fresh state under the lock and the state re-assert is authoritative. This also keeps the OPA call off the lock.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -588,21 +623,34 @@ public void issueCertificateAction(final UUID certificateUuid, boolean isApprove
...
-            handleFailedOrRejectedEvent(certificate, null, CertificateState.FAILED, CertificateEvent.ISSUE, new HashMap<>(), e.getMessage());
+            // The certificate is detached here — the claim transaction committed (to release the row
+            // lock) before the connector call, so handleFailedOrRejectedEvent cannot lazy-load the
+            // collections it needs (locations, predecessor relations). Run the failure handling against
+            // a managed entity re-read inside a short transaction.
+            TransactionStatus failTx = transactionManager.getTransaction(new DefaultTransactionDefinition());
+            try {
+                Certificate managed = certificateRepository.findWithAssociationsByUuid(certificateUuid)
+                        .orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
+                handleFailedOrRejectedEvent(managed, null, CertificateState.FAILED, CertificateEvent.ISSUE, new HashMap<>(), e.getMessage());
```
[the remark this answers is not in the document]
lubomirw: Fixed in 28364057b — the failure path now re-reads with `findAndLockWithAssociationsByUuid` under the row lock, so handleFailedOrRejectedEvent observes the latest state; its `canTransition` guard then skips when a concurrent actor has already finalized the certificate (ISSUED -> FAILED is not a legal transition), so a concurrent finalize is no longer clobbered.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -588,21 +623,34 @@ public void issueCertificateAction(final UUID certificateUuid, boolean isApprove
...
+                transactionManager.commit(failTx);
+            } catch (Exception failEx) {
+                transactionManager.rollback(failTx);
+                logger.error("Failed to record issuance failure for certificate {}: {}", certificateUuid, failEx.getMessage());
+            }
```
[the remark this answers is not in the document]
lubomirw: Fixed in 809889596 — the failure log now passes the exception as the final argument, so the full stack trace is emitted (matching the onAsyncAccepted / handleFailedOrRejectedEvent convention).

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1129,7 +1129,17 @@ public void addCertificateRequestToExisting(UUID certificateUuid, ClientCertific
...
+        // Authorize before locking WITHOUT managing the entity: the RA-profile check makes an external OPA
+        // call (must not run while holding the row lock), and loading the full entity first would make the
+        // locking query below return the already-managed, stale-state instance — defeating the under-lock
+        // re-assert. Fetch only the RA-profile UUID (a scalar projection), authorize, then take the lock so
+        // findAndLockWithAssociationsByUuid loads fresh state under SELECT ... FOR UPDATE.
+        UUID raProfileUuid = certificateRepository.findRaProfileUuidByUuid(certificateUuid)
+                .orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
+        raProfileService.evaluateCertificateRaProfilePermissions(SecuredUUID.fromUUID(certificateUuid),
+                SecuredParentUUID.fromUUID(raProfileUuid));
+        Certificate certificate = certificateRepository.findAndLockWithAssociationsByUuid(certificateUuid)
```
ivosh, line 1141: **! Major** | Security

`addCertificateRequestToExisting()` now authorizes against the scalar `raProfileUuid` read at line 1137 and then mutates the certificate returned by the later locking read at line 1141 without verifying the RA profile is still the same. Because `switchRaProfile()` is allowed on `REGISTERED` certificates, a concurrent RA-profile switch between those two reads lets a caller who was authorized on the old profile attach a CSR after the certificate has moved to a different profile they may not control. After taking the lock, compare `certificate.getRaProfileUuid()` to the pre-authorized UUID and reject if it changed, or re-run the permission check against the locked row before applying the CSR.

**How it materializes** — certificate `C` is `REGISTERED` under RA profile `P1`. Actor `A` has CSR-attach rights on `P1` but **not** on `P2`; an admin/automation concurrently reassigns `C` to `P2` (allowed on `REGISTERED`, and `switchRaProfile` takes **no** row lock, so it does not serialize against the attach path):

```
Thread A  (addCertificateRequestToExisting)          Thread B  (switchRaProfile → P2)
────────────────────────────────────────────────    ──────────────────────────────────
1137  raProfileUuid = findRaProfileUuidByUuid(C)
        → returns P1
1139  evaluateCertificateRaProfilePermissions(
        cert=C, parent=P1)  → PASS
        (OPA HTTP round-trip — widens the window)
                                                      switchRaProfile(C, P2): C.raProfile = P2
                                                      commit   ← no lock, not serialized
1141  certificate = findAndLockWithAssociations(C)
        → C, now under P2   (SELECT … FOR UPDATE)
1146  re-assert  state == REGISTERED   → still true
        (the switch left state unchanged)
1148+ attach A's CSR to C, persist
        → C now carries A's CSR while under P2,
          which A was never authorized for
```

The under-lock re-assert only guards `state`, so it cannot catch a profile change: A was authorized against `P1` (step 1139) but the write lands on `C` under `P2`. The pre-PR code did a single `getCertificateEntity(...)` read — it authorized against the *same* entity instance it then mutated, so authz-target and write-target could not diverge. Splitting the read into scalar-authz-then-locked-load is what opened the window, and the OPA HTTP call between the two reads makes it comfortably wide.

**Fix:** after acquiring the lock at 1141, re-derive the RA profile from the locked row and either (a) reject if `certificate.getRaProfileUuid()` != the pre-authorized `raProfileUuid`, or (b) re-run `evaluateCertificateRaProfilePermissions(...)` against the locked row before attaching the CSR.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1129,7 +1129,17 @@ public void addCertificateRequestToExisting(UUID certificateUuid, ClientCertific
         if (signRequest.getFormat() == null) {
             throw new CertificateRequestException("A certificate signing request format (PKCS10 or CRMF) is required");
         }
-        Certificate certificate = getCertificateEntity(SecuredUUID.fromUUID(certificateUuid));
+        // Authorize before locking WITHOUT managing the entity: the RA-profile check makes an external OPA
+        // call (must not run while holding the row lock), and loading the full entity first would make the
+        // locking query below return the already-managed, stale-state instance — defeating the under-lock
+        // re-assert. Fetch only the RA-profile UUID (a scalar projection), authorize, then take the lock so
+        // findAndLockWithAssociationsByUuid loads fresh state under SELECT ... FOR UPDATE.
+        UUID raProfileUuid = certificateRepository.findRaProfileUuidByUuid(certificateUuid)
```
ivosh, line 1137: **~ Nitpick** | Performance

CertificateServiceImpl is class-level @Transactional (line 131) and addCertificateRequestToExisting is @Transactional (line 1123), i.e. REQUIRED. Inside the method the projection query findRaProfileUuidByUuid (line 1137) opens the transaction/acquires the pooled DB connection, and the very next call evaluateCertificateRaProfilePermissions (line 1139) is an external OPA authorization call (executed by the @ExternalAuthorization AOP advice on RaProfileServiceImpl.evaluateCertificateRaProfilePermissions, whose body is empty). So a pooled DB connection is held across that HTTP call — the exact anti-pattern CLAUDE.md 'Transactions and external calls' warns against. The sibling operator paths in ClientOperationServiceImpl (issueCertificateAction, manuallyIssueCertificate, manuallyConfirmRevoke) all use @Transactional(propagation=NOT_SUPPORTED) + an explicit short tx precisely to avoid this. Two mitigating facts keep this non-blocking: (1) the PESSIMISTIC_WRITE row lock is acquired only AFTER the OPA call (line 1141), so the row lock never spans HTTP — the primary hazard is avoided and the PR's reordering is correct; (2) this connection-across-OPA is pre-existing (the prior getCertificateEntity(...) call did findByUuid then OPA within the same @Transactional). 

Consider switching this method to the NOT_SUPPORTED + explicit-tx shape to match the established pattern and keep the connection off the OPA round-trip.

ivosh (changes requested):

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
lubomirw: Fixed in 33d13412d — after acquiring the lock, `addCertificateRequestToExisting` now rejects (`ValidationException`) if the locked row's RA profile differs from the one authorization was evaluated against, closing the `switchRaProfile` TOCTOU. Added a deterministic regression test (`addCertificateRequestToExisting_reassertUnderLock_rejectsWhenRaProfileChangedDuringAuthorization`).

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
lubomirw: Agreed — valid, but pre-existing (the prior `getCertificateEntity` read held the connection across the same OPA call) and non-blocking as you note, since the row lock no longer spans OPA. Leaving the `NOT_SUPPORTED` + explicit-tx refactor of this method as a separate cleanup rather than folding a transaction-semantics change onto the concurrency fix.

ivosh (dismissed):

ivosh (approved):

PR #1729, opened 2026-07-07
Require a functional interface by category in connector validation

lubomirw:
## What

`ConnectorV2Adapter.validateConnection` gated the "connector must implement at least one functional interface" rule with a **size comparison** — `implementedInterfaces.size() == mandatoryInterfaces.size()` (3). It inferred "no functional interface" from "exactly 3 interfaces", which only held while the sole common interfaces were `{INFO, HEALTH, METRICS}`.

`interfaces#755` added `ATTRIBUTES` as a **common** interface (plus an `InterfaceCategory` axis on every `ConnectorInterface`). A connector advertising `{INFO, HEALTH, METRICS, ATTRIBUTES}` now has size 4 ≠ 3, so it **wrongly passed** the check with zero functional interfaces.

## Change

Gate on category instead of count:

```java
boolean hasFunctionalInterface = implementedInterfaces.stream()
        .anyMatch(i -> i.getCategory() == ConnectorInterface.InterfaceCategory.FUNCTIONAL);
```

The mandatory-interface presence check is unchanged. Depends on the `getCategory()` API from `interfaces:2.18.1-SNAPSHOT` (already pinned).

## Tests

New `ConnectorV2AdapterTest` — pure unit test (no Spring/Mockito; `validateConnection(ConnectInfo)` touches no collaborators), 5 cases:

- missing mandatory → throws (message asserts "mandatory")
- `{INFO, HEALTH, METRICS}` → throws (message asserts "functional")
- `{INFO, HEALTH, METRICS, ATTRIBUTES}` → throws — **regression guard** for the exact case the old size-check let through
- `{…, AUTHORITY}` and `{…, ATTRIBUTES, AUTHORITY}` → pass

`mvn test -Dtest=ConnectorV2AdapterTest` → 5/5 green.

Refs OmniTrustILM/interfaces#755


src/test/java/com/otilm/core/service/handler/ConnectorV2AdapterTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+class ConnectorV2AdapterTest {
+
+    private final ConnectorV2Adapter adapter = new ConnectorV2Adapter();
+
+    @Test
+    void missingMandatoryInterface_throws() {
+        ValidationException ex = assertThrows(ValidationException.class,
+                () -> adapter.validateConnection(connectInfo(ConnectorInterface.INFO, ConnectorInterface.HEALTH)));
+        assertTrue(ex.getMessage().toLowerCase().contains("mandatory"), ex.getMessage());
```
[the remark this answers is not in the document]
lubomirw: Fixed in c2c9a787d — dropped `toLowerCase()` and assert on a branch-unique message substring instead.

src/test/java/com/otilm/core/service/handler/ConnectorV2AdapterTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Test
+    void onlyMandatoryCommons_throwsMissingFunctional() {
+        ValidationException ex = assertThrows(ValidationException.class,
+                () -> adapter.validateConnection(connectInfo(
+                        ConnectorInterface.INFO, ConnectorInterface.HEALTH, ConnectorInterface.METRICS)));
+        assertTrue(ex.getMessage().toLowerCase().contains("functional"), ex.getMessage());
```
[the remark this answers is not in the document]
lubomirw: Fixed in c2c9a787d — dropped `toLowerCase()` and assert on a branch-unique message substring instead.

src/test/java/com/otilm/core/service/handler/ConnectorV2AdapterTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /** ATTRIBUTES is a common interface, so it must not count toward the functional-interface requirement. */
+    @Test
+    void commonsWithAttributes_throwsMissingFunctional() {
+        ValidationException ex = assertThrows(ValidationException.class,
+                () -> adapter.validateConnection(connectInfo(
+                        ConnectorInterface.INFO, ConnectorInterface.HEALTH,
+                        ConnectorInterface.METRICS, ConnectorInterface.ATTRIBUTES)));
+        assertTrue(ex.getMessage().toLowerCase().contains("functional"), ex.getMessage());
```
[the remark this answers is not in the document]
lubomirw: Fixed in c2c9a787d — dropped `toLowerCase()` and assert on a branch-unique message substring instead.

src/main/java/com/otilm/core/service/handler/ConnectorV2Adapter.java
```diff
@@ -83,7 +83,9 @@ public ConnectInfo validateConnection(ConnectInfo connectInfoV2) throws Connecto
             throw new ValidationException("Connector is missing mandatory interfaces: " + String.join(", ", mandatoryInterfaces.stream().map(ConnectorInterface::getLabel).toList()));
         }
 
-        if (implementedInterfaces.size() == mandatoryInterfaces.size()) {
+        boolean hasFunctionalInterface = implementedInterfaces.stream()
+                .anyMatch(i -> i.getCategory() == ConnectorInterface.InterfaceCategory.FUNCTIONAL);
```
ivosh, line 92: **~ Minor** | Error handling

The new functional-interface check dereferences every `ConnectorInterface` via `i.getCategory()`, but `ConnectorInterfaceInfo.code` comes from connector-supplied `/v2/info` JSON and can be `null` when a connector returns malformed metadata. With the mandatory interfaces present plus one bad entry, this path now throws an uncaught `NullPointerException` instead of the expected `ValidationException`, so connector discovery/create flows surface a 500 rather than a user-facing validation error. Reject `null` interface codes explicitly before building `implementedInterfaces` (or at minimum filter them out before `getCategory()` and throw a `ValidationException` describing the malformed interface list).

src/test/java/com/otilm/core/service/handler/ConnectorV2AdapterTest.java
```diff
@@ -0,0 +1,74 @@
...
+                ConnectorInterface.INFO, ConnectorInterface.HEALTH,
+                ConnectorInterface.METRICS, ConnectorInterface.ATTRIBUTES);
+        ValidationException ex = assertThrows(ValidationException.class,
+                () -> adapter.validateConnection(connectInfo));
+        assertTrue(ex.getMessage().contains("functional"), ex.getMessage());
+    }
+
+    @Test
+    void mandatoryPlusFunctional_passes() throws ConnectorException {
+        assertNotNull(adapter.validateConnection(connectInfo(
```
ivosh: **~ Minor** | Testing

`mandatoryPlusFunctional_passes` (ConnectorV2AdapterTest.java:52-56) and `attributesPlusFunctional_passes` (line 58-63) only do `assertNotNull(adapter.validateConnection(...))`. But the production method returns the exact `connectInfo` object it was handed (ConnectorV2Adapter.java:75 casts the input, line 92 `return connectInfo;` returns that same non-null instance). So the return value is *structurally* incapable of being null when validation passes — the `assertNotNull` verifies nothing about the code under test. The real behaviour these tests intend to pin is "a connector with mandatory + one functional interface validates *without throwing* ValidationException", and the only thing actually giving that coverage is the implicit fact that a thrown exception would fail the test. Make the contract explicit and meaningful: e.g. `assertThatCode(() -> adapter.validateConnection(ci)).doesNotThrowAnyException()`, or assert the returned object preserves the input interfaces (`assertThat(result).isSameAs(ci)` / interface codes equal). As written this is a type/null-only assertion — the must-flag "green test asserting the wrong thing" pattern. (Contrast the throwing-case tests at lines 26-28/35-37/46-48, which correctly pin origin via `ex.getMessage().contains(...)`.)

src/test/java/com/otilm/core/service/handler/ConnectorV2AdapterTest.java
```diff
@@ -0,0 +1,74 @@
...
+import com.otilm.api.exception.ValidationException;
+import com.otilm.api.model.client.connector.v2.ConnectorInterface;
+import com.otilm.api.model.client.connector.v2.ConnectorInterfaceInfo;
+import com.otilm.api.model.core.connector.v2.ConnectInfoV2;
+import org.junit.jupiter.api.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
```
ivosh: **. Nitpick** | Maintainability

ConnectorV2AdapterTest.java:13-15 imports JUnit `Assertions.assertNotNull/assertThrows/assertTrue`. The documented OmniTrust Core standard is AssertJ (`assertThat(...)`), and the newest sibling in this very package, CertificateHandlerValidationBatchingTest.java:18-19, uses `assertThatThrownBy(...).hasMessageContaining(...)` — which is exactly the shape of the three exception assertions here (`assertThrows(...)` + `assertTrue(ex.getMessage().contains("functional"))` at lines 26-28/35-37/46-48 would collapse to `assertThatThrownBy(...).isInstanceOf(ValidationException.class).hasMessageContaining("functional")`). HONESTY / calibration: the codebase is genuinely split (local grep: ~224 test files use JUnit `Assertions.*` vs ~98 using AssertJ), and the closest sibling `ConnectorCapabilityServiceTest.java:11-12` in the same package uses JUnit `assertTrue/assertFalse`. So this is not a clear-cut violation of an enforced norm — hence nitpick, not a blocker. Both readings: (a) match the AssertJ direction the align-test-style/refactor-java-tests skills prescribe and the modern sibling follows; (b) leave as-is, consistent with the older sibling. Action: prefer AssertJ `assertThatThrownBy(...).hasMessageContaining(...)` for the throwing cases since it reads better and matches the package's newest test; if left as JUnit, that is a defensible choice given the split.

src/main/java/com/otilm/core/service/handler/ConnectorV2Adapter.java
```diff
@@ -83,7 +83,9 @@ public ConnectInfo validateConnection(ConnectInfo connectInfoV2) throws Connecto
             throw new ValidationException("Connector is missing mandatory interfaces: " + String.join(", ", mandatoryInterfaces.stream().map(ConnectorInterface::getLabel).toList()));
         }
 
-        if (implementedInterfaces.size() == mandatoryInterfaces.size()) {
+        boolean hasFunctionalInterface = implementedInterfaces.stream()
```
ivosh, line 91: **. Nitpick** | Documentation

The overload the new test drives, `validateConnection(ConnectInfo connectInfoV2)` (ConnectorV2Adapter.java:74-93), opens with an unchecked downcast `(ConnectInfoV2) connectInfoV2` (line 75) and then applies two independent gates (mandatory-interface presence, then the functional-category requirement changed by this PR). Neither the impl nor the interface declaration `ConnectorAdapter.java:24` carries Javadoc; the contract that callers must pass a V2-typed ConnectInfo (else ClassCastException) and that the method returns the same instance on success is not stated anywhere. Is this method obvious enough to need no doc? Two readings: (a) it's a straightforward internal adapter override and the sibling overrides (getInfo/checkHealth, lines 47/52) are also undocumented, so no doc is consistent; (b) the silent cast + this-PR's category gate + the this-returns-input semantic are exactly the non-obvious behaviours the repo's own 'Writing for future AI sessions' guidance says to capture in method-level Javadoc. 

Action: consider a one-line Javadoc on ConnectorAdapter.java:24 documenting the expected concrete type and the return-the-validated-input contract.

ivosh (changes requested):

src/main/java/com/otilm/core/service/handler/ConnectorV2Adapter.java
lubomirw: Fixed in 601ed89fe — a null interface code is now rejected with a `ValidationException` ("...no code.") before the category check, so malformed `/v2/info` yields a validation error rather than an NPE/500. Added a `malformedInterfaceWithNullCode_throws` regression test.

src/test/java/com/otilm/core/service/handler/ConnectorV2AdapterTest.java
lubomirw: Fixed in 601ed89fe — the passing tests now assert the return contract (`assertThat(...).isSameAs(connectInfo)`), which also fails if it throws, instead of the vacuous `assertNotNull`.

src/test/java/com/otilm/core/service/handler/ConnectorV2AdapterTest.java
lubomirw: Done in 601ed89fe — converted to AssertJ (`assertThatThrownBy(...).isInstanceOf(...).hasMessageContaining(...)` and `assertThat(...).isSameAs(...)`), matching the package’s newer sibling `CertificateHandlerValidationBatchingTest`.

src/main/java/com/otilm/core/service/handler/ConnectorV2Adapter.java
lubomirw: Added in 601ed89fe — Javadoc on `ConnectorAdapter.validateConnection(ConnectInfo)` documenting the expected version-appropriate subtype and the validated-return / `ValidationException`-on-invalid contract.

ivosh (approved):

PR #1733, opened 2026-07-07
Allow Custom Attribute conditions on Certificate uploaded event triggers

### LINKED ISSUE #1705: Core - Event Triggers: Cannot associate trigger with Custom Attribute condition to "Certificate uploaded" event — backend rejects with 422 "invalid field source"
### Description

In **Settings → Events → Configure Event Triggers**, when the operator tries to associate a trigger with the **Certificate uploaded** event, the backend rejects the association with HTTP **422** if the trigger's rule contains a condition on a **Custom Attribute** field source:

```
[
  "Trigger <trigger_name> cannot be associated with event 'Certificate uploaded' as it contains rule with invalid field source, which is not allowed for this event."
]
```

This restriction is inconsistent with how Certificate uploads actually work. **Custom attributes are part of the certificate upload payload** — the operator can set them at the moment of upload. Filtering by these values is a valid and expected use case (e.g. auto-reject uploads whose custom attribute `Criticality = Low`, or apply categorization based on custom attribute values — which is directly in line with the intent of Epic #115 "Categorization of uploaded certificates").

Blocking Custom Attribute conditions for this event undermines the entire "categorize / reject based on the uploaded cert's properties" scenario that the platform advertises.

### Steps to Reproduce

1. Create a **Condition** with `Resource = Certificate`, `Filter Field Source = Custom Attribute`, any field / operator / value
2. Wrap it in a **Rule** and attach the rule to a **Trigger** (e.g. an ignore trigger or a regular trigger with an action)
3. Navigate to **Settings → Events → Certificate uploaded → Configure Event Triggers**
4. In the trigger dropdown select the trigger from step 2
5. Click **Add** / **Save**

### Expected Behavior

- Backend accepts the association
- The trigger fires on `Certificate uploaded` events and evaluates the Custom Attribute condition against the uploaded certificate's custom attributes (which were sent as part of the upload payload)
- Categorization / rejection based on Custom Attribute values works end-to-end

### Actual Behavior

- Backend returns HTTP **422**:
  ```
  ["Trigger <trigger_name> cannot be associated with event 'Certificate uploaded' as it contains rule with invalid field source, which is not allowed for this event."]
  ```
- Association is refused
- Operator cannot express a valid business rule based on custom attributes provided at upload time

### Severity

Major

### Module

Core

### Attachments

Image

Image

Image

### Logs

_No response_

### Environment

ILM Core: 2.18.1-SNAPSHOT (2 Jul 2026, 09:00)
Commit: 633d1f0
Branch: main
Build time: 2026-07-02T07:00:20Z
ILM Frontend: 2.18.0 (1 Jul 2026, 18:28)
Client:
- macOS Tahoe 26.5.1

klaraf755:
Fixes #1705


src/test/java/com/otilm/core/integration/events/EventHandlersITest.java
```diff
@@ -318,6 +318,111 @@ private void createCertificateTriggerAssociation(ResourceEvent event, Resource e
...
+
+        // The ignore-trigger matched the custom attribute supplied in the request — the attribute was never persisted anywhere,
+        // and the certificate itself was never saved either, so this proves the CUSTOM condition evaluated against request content.
+        Assertions.assertFalse(certificateRepository.findByFingerprint(fingerprint).isPresent());
+
+        List<TriggerHistory> histories = triggerHistoryRepository.findAll();
+        Assertions.assertEquals(1, histories.size());
+        Assertions.assertTrue(histories.getFirst().isConditionsMatched());
+        Assertions.assertTrue(histories.getFirst().isActionsPerformed());
+        mockServer.stop();
```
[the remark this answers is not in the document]
klaraf755: Fixed.

src/test/java/com/otilm/core/integration/events/EventHandlersITest.java
```diff
@@ -318,6 +318,111 @@ private void createCertificateTriggerAssociation(ResourceEvent event, Resource e
...
+
+        // The EMPTY condition matched because no attribute was supplied at all — proves CertificateUploadedEventHandler
+        // normalizes a null customAttributes() to an empty list rather than passing null through to the evaluator.
+        Assertions.assertFalse(certificateRepository.findByFingerprint(fingerprint).isPresent());
+
+        List<TriggerHistory> histories = triggerHistoryRepository.findAll();
+        Assertions.assertEquals(1, histories.size());
+        Assertions.assertTrue(histories.getFirst().isConditionsMatched());
+        Assertions.assertTrue(histories.getFirst().isActionsPerformed());
+        mockServer.stop();
```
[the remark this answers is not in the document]
klaraf755: Fixed.

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
```diff
@@ -278,13 +292,29 @@ private boolean evaluateMetaAttributeConditionItem(Resource resource, String fie
...
-        List<ResponseAttribute> responseAttributes = attributeEngine.getObjectCustomAttributesContent(resource, objectUuid);
-        ResponseAttributeV3 attributeToCompare = (ResponseAttributeV3) responseAttributes.stream().filter(rad -> Objects.equals(rad.getName(), attributeName)).findFirst().orElse(null);
+        List<? extends AttributeContent> attributeContent = null;
+        AttributeContentType attributeContentType = null;
+        // Pending (request-supplied) content takes precedence for this specific attribute name — it may not exist in
+        // the DB yet (e.g. a certificate upload request, evaluated before its attributes are persisted). If this
+        // attribute isn't in the pending list, fall through to the DB, which may hold a value written by an earlier
+        // trigger's action in this same evaluation pass (e.g. a SET_FIELD execution).
+        RequestAttribute requestAttribute = pendingCustomAttributes == null ? null : pendingCustomAttributes.stream().filter(ra -> Objects.equals(ra.getName(), attributeName)).findFirst().orElse(null);
```
lubomirw, line 297: Persisted custom-attribute evaluation goes through `attributeEngine.getObjectCustomAttributesContent(...)`, which applies the current trigger user's attribute ACL via `loadCustomAttributesSecurityResourceFilter()`. This new pending path bypasses that and evaluates `requestAttribute.getContent()` directly after `EventHandler` has switched to `triggerAssociation.getTriggeredBy()`. As a result, a `CERTIFICATE_UPLOADED` trigger can match/ignore/send notifications based on a custom attribute the trigger user would not be allowed to read once persisted. Please apply the same custom-attribute visibility filtering to pending attributes before evaluating them, or evaluate them through `AttributeEngine` under the trigger user.

lubomirw (commented):
Multi-pass review found one issue in the new pending custom-attribute evaluation path.

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
lubomirw, line 297: **[High]** | Security

The pending path reads `requestAttribute.getContent()` directly, bypassing the custom-attribute ACL the DB path applies: `attributeEngine.getObjectCustomAttributesContent(resource, objectUuid)` runs `loadCustomAttributesSecurityResourceFilter()` (allowed/forbidden filtering, `AttributeEngine.java:811-838`). Trigger evaluation runs as the trigger's `triggeredBy` user, so a `CERTIFICATE_UPLOADED` trigger can match / ignore / notify on a custom attribute that user is not authorized to read.

_Suggested:_ resolve pending attributes through the same permission logic (filter against the attribute definition's ACL) and treat unauthorized attributes as absent, rather than trusting the raw request payload.

<sub>flagged by Copilot CLI + guided review; matches the existing open thread on this line</sub>

lubomirw (commented):
Batched multi-review of #1733 (Copilot CLI, guided-pr-review, superpowers code review, pr-hygiene) — 5 inline comments below.

**Docs follow-up (Low, Documentation):** the `documentation` repo's `event.md` (Certificate Uploaded section) doesn't yet mention that Custom Attribute field-source conditions are now accepted on Certificate Uploaded triggers (previously rejected 422, per #1705). Worth a sentence noting they're evaluated against the upload request payload; META remains unsupported.

src/main/java/com/otilm/core/events/handlers/CertificateUploadedEventHandler.java
```diff
@@ -136,7 +143,7 @@ public void handleEvent(EventMessage eventMessage) throws EventException {
             // Retroactively link trigger histories of the ignore triggers to the certificate
             triggerHistoryRepository.updateObjectUuidAndObjectResource(certificate.getUuid(), Resource.CERTIFICATE, eventHistory.getUuid());
 
-            evaluateTriggers(context, context.getPlatformTriggers(), certificate, eventData, eventHistory);
+            evaluateTriggers(context, context.getPlatformTriggers(), certificate, eventData, eventHistory, pendingCustomAttributes);
```
lubomirw, line 146: **[Medium]** | Correctness

Now that normal triggers run with pending custom attributes, note the persist ordering: request custom attributes are written via full-replace `updateObjectCustomAttributesContent(...)` at line 157 — *after* this `evaluateTriggers(...)` call — so a trigger's `SET_FIELD` custom-attribute write here can be silently overwritten by the request payload. Correspondingly, the evaluator gives the pending value precedence over a same-name DB value written by an earlier `SET_FIELD` in the same pass. Self-consistent (request payload wins at persist too) but untested for the same-name case.

_Suggested:_ confirm the precedence is intended and add a test pinning the same-name `SET_FIELD`-then-condition behaviour; or persist validated request attributes before running normal trigger actions.

<sub>flagged by Copilot CLI + superpowers review</sub>

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
```diff
@@ -278,13 +292,29 @@ private boolean evaluateMetaAttributeConditionItem(Resource resource, String fie
...
+        if (requestAttribute != null) {
+            attributeContentType = requestAttribute.getContentType();
```
lubomirw, line 299: **[Low]** | Correctness

The pending lookup matches on attribute name only, then evaluates using `requestAttribute.getContentType()` as-is, discarding the content type in the field identifier (e.g. the `STRING` in `criticality|STRING`). A request attribute with a null/mismatched content type evaluates against the wrong type. Matches the pre-existing DB path (also name-only), so not a regression — unlike META, which matches name + content type.

_Suggested:_ optionally match on name + parsed field-identifier content type (as `evaluateMetaAttributeConditionItem` does), or use the attribute definition's content type.

<sub>flagged by Copilot CLI + guided review</sub>

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
```diff
@@ -77,8 +78,12 @@ public void setAttributeEngine(AttributeEngine attributeEngine) {
 
     @Override
     public TriggerHistory evaluateTrigger(Trigger trigger, TriggerAssociation triggerAssociation, T object, UUID referenceObjectUuid, Object data, EventHistory eventHistory) throws RuleException {
+        return evaluateTrigger(trigger, triggerAssociation, object, referenceObjectUuid, data, eventHistory, null);
+    }
+
+    public TriggerHistory evaluateTrigger(Trigger trigger, TriggerAssociation triggerAssociation, T object, UUID referenceObjectUuid, Object data, EventHistory eventHistory, List<RequestAttribute> pendingCustomAttributes) throws RuleException {
```
lubomirw, line 80: **[Low]** | Maintainability

The new `pendingCustomAttributes` overloads (`evaluateTrigger`, `evaluateRules`, `evaluateConditionItem`) are public on the concrete `TriggerEvaluator` but not on `ITriggerEvaluator`, which still declares only the base signatures. Callers must depend on the concrete class to reach the capability, so the interface is an incomplete contract.

_Suggested:_ add the overloads to `ITriggerEvaluator` with a `default` method delegating to the existing signature (passing `null`).

<sub>flagged by guided review</sub>

src/test/java/com/otilm/core/integration/evaluator/TriggerEvaluatorITest.java
```diff
@@ -768,6 +765,113 @@ void testCertificateRuleEvaluatorCustomAttributeAbsent() throws AlreadyExistExce
...
+        // objectUuid is null AND the pending list doesn't contain this attribute name — no DB to fall back to, so it's absent
+        ConditionItem emptyCondition = new ConditionItem();
+        emptyCondition.setFieldSource(FilterFieldSource.CUSTOM);
+        emptyCondition.setFieldIdentifier("otherAttribute|STRING");
+        emptyCondition.setOperator(FilterConditionOperator.EMPTY);
+        Assertions.assertTrue(certificateTriggerEvaluator.evaluateConditionItem(emptyCondition, unpersistedCertificate, Resource.CERTIFICATE, pendingCustomAttributes));
+
+        // A non-null but EMPTY pending list (the request explicitly supplied zero custom attributes) means the same thing:
+        // no content is possible, so EMPTY is satisfied. This is the case CertificateUploadedEventHandler must produce
+        // when eventMessageData.customAttributes() is null, by normalizing to List.of() rather than passing null through (see Task 4).
```
lubomirw: **[Low]** | Maintainability

Test comment leaks an internal planning reference — `(see Task 4)`.

_Suggested:_ drop the parenthetical so the sentence ends `...rather than passing null through.`

<sub>flagged by pr-hygiene</sub>

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
klaraf755: Name for custom attribute is unique, so no need to check also via content type.

src/test/java/com/otilm/core/integration/evaluator/TriggerEvaluatorITest.java
klaraf755: Dropped it.

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
klaraf755: Added the overloads.

src/main/java/com/otilm/core/events/handlers/CertificateUploadedEventHandler.java
klaraf755: This is intended, and the test already exists - `testCertificateUploadedEvent` in `EventHandlersITest` covers this path, added comments to the test to clarify that.

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
klaraf755: Added security filter.

klaraf755: Added security filter.

lubomirw (approved):

PR #1734, opened 2026-07-07
Stop hourly error log when CBOM repository is unconfigured

### LINKED ISSUE #1724: CbomSyncTask logs UnexpectedRollbackException every hour when CBOM repository client is not configured
### Description

The hourly `CbomSyncTask` scheduled job logs an ERROR with a full `UnexpectedRollbackException` stack trace on **every run** whenever the CBOM repository client is not configured — the common case for deployments that don't use the CBOM repository integration.

Root cause (traced in `SchedulerListener`, `SchedulerServiceImpl`, and `CbomSyncTask`):

- `SchedulerListener.processMessage` is `@Transactional`, so the whole scheduler message is handled in a single transaction.
- When the CBOM repository client is not configured, `CbomSyncTask.performJob` throws `ScheduledJobSkippedException` (a `RuntimeException`). Because `CbomSyncTask` is `@Transactional`, the escaping runtime exception marks the shared transaction **rollback-only** (Spring's default rollback behavior).
- `SchedulerServiceImpl.runScheduledJob` catches `ScheduledJobSkippedException`, treats it as a normal skip, deletes the STARTED history entry, and returns successfully.
- The listener transaction then attempts to commit, but it is marked rollback-only, so Spring throws `UnexpectedRollbackException`. This surfaces at `AbstractJmsEndpointConfig` as ERROR "Unexpected error in endpoint 'schedulerListener'".

Net effect: the job functionally skips as intended, but each hourly run (cron `0 0 * ? * *`) emits a misleading ERROR and stack trace. This pollutes logs and can trigger false monitoring alerts or mask real scheduler failures.

Likely fix direction: don't mark the transaction rollback-only for the skip case — e.g. `@Transactional(noRollbackFor = ScheduledJobSkippedException.class)`, signal "skipped" via a return status instead of throwing across the transactional boundary, or check `cbomService.isCbomRepositoryClientConfigured()` before the transactional task runs.

### Steps to Reproduce

1. Deploy Core with the CBOM repository client not configured (the default when the CBOM repository integration is not set up).
2. Wait for the `CbomSyncTask` scheduled job to fire — cron `0 0 * ? * *`, i.e. every hour on the hour.
3. Inspect the scheduler logs for that run.

### Expected Behavior

When CBOM is not configured, `CbomSyncTask` skips cleanly: the job is marked finished/skipped, the history entry is handled, and no ERROR or stack trace is logged.

### Actual Behavior

The job reports as finished, but the scheduler-listener transaction fails to commit and logs an ERROR on every run:

```
ERROR [schedulerListener-1] [...AbstractJmsEndpointConfig - 105]: Unexpected error in endpoint 'schedulerListener'
org.springframework.transaction.UnexpectedRollbackException: Transaction silently rolled back because it has been marked as rollback-only
    at org.springframework.transaction.support.AbstractPlatformTransactionManager.processCommit(AbstractPlatformTransactionManager.java:804)
    at org.springframework.transaction.support.AbstractPlatformTransactionManager.commit(AbstractPlatformTransactionManager.java:758)
    at org.springframework.transaction.interceptor.TransactionAspectSupport.commitTransactionAfterReturning(TransactionAspectSupport.java:698)
    ...
    at com.czertainly.core.messaging.jms.listeners.SchedulerListener$$SpringCGLIB$$0.processMessage(<generated>)
    at com.czertainly.core.messaging.jms.listeners.AbstractJmsEndpointConfig.lambda$listenerEndpointInternal$0(AbstractJmsEndpointConfig.java:96)
    ...
```

The preceding DEBUG/INFO lines confirm the job was actually skipped before the error:

```
DEBUG [...SchedulerServiceImpl - 302]: Skipping scheduled job 'CbomSyncTask', removing history entry
INFO  [...SchedulerServiceImpl - 321]: Scheduled job 'CbomSyncTask' has finished
```

So the skip itself succeeds; the ERROR is raised afterwards, on transaction commit.

klaraf755:
Fixes #1724

### DIFF
--- a/src/main/java/com/otilm/core/api/ScheduledJobSkippedException.java
+++ b/src/main/java/com/otilm/core/api/ScheduledJobSkippedException.java
@@ -2,5 +2,15 @@ package com.otilm.core.api;
 
 import com.otilm.api.exception.PlatformException;
 
+/**
+ * Signals that a {@link com.otilm.core.tasks.ScheduledJobTask#performJob} run was skipped
+ * (not an error) and should be treated as such by the caller.
+ * <p>
+ * If the throwing task class is {@code @Transactional}, it must declare
+ * {@code @Transactional(noRollbackFor = ScheduledJobSkippedException.class)} — otherwise the
+ * exception marks the ambient transaction (shared with {@code SchedulerListener}) rollback-only
+ * before it is caught, causing {@code UnexpectedRollbackException} on commit even though the
+ * skip itself is handled correctly.
+ */
 public class ScheduledJobSkippedException extends RuntimeException implements PlatformException {
 }
--- a/src/main/java/com/otilm/core/api/ScheduledJobSkippedException.java
+++ b/src/main/java/com/otilm/core/api/ScheduledJobSkippedException.java
@@ -2,15 +2,5 @@ package com.otilm.core.api;
 
 import com.otilm.api.exception.PlatformException;
 
-/**
- * Signals that a {@link com.otilm.core.tasks.ScheduledJobTask#performJob} run was skipped
- * (not an error) and should be treated as such by the caller.
- * <p>
- * If the throwing task class is {@code @Transactional}, it must declare
- * {@code @Transactional(noRollbackFor = ScheduledJobSkippedException.class)} — otherwise the
- * exception marks the ambient transaction (shared with {@code SchedulerListener}) rollback-only
- * before it is caught, causing {@code UnexpectedRollbackException} on commit even though the
- * skip itself is handled correctly.
- */
 public class ScheduledJobSkippedException extends RuntimeException implements PlatformException {
 }
--- a/src/main/java/com/otilm/core/tasks/CbomSyncTask.java
+++ b/src/main/java/com/otilm/core/tasks/CbomSyncTask.java
@@ -16,7 +16,6 @@ import org.springframework.transaction.annotation.Transactional;
 
 @Component
 @NoArgsConstructor
-@Transactional
 public class CbomSyncTask implements ScheduledJobTask {
 
     public static final String NAME = "CbomSyncTask";
@@ -51,6 +50,7 @@ public class CbomSyncTask implements ScheduledJobTask {
     }
 
     @Override
+    @Transactional(noRollbackFor = ScheduledJobSkippedException.class)
     public ScheduledTaskResult performJob(final ScheduledJobInfo scheduledJobInfo, final Object taskData) {
         if (!cbomService.isCbomRepositoryClientConfigured()) {
             throw new ScheduledJobSkippedException();
--- a/src/main/java/com/otilm/core/tasks/CbomSyncTask.java
+++ b/src/main/java/com/otilm/core/tasks/CbomSyncTask.java
@@ -12,6 +12,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 
 @Component
@@ -50,7 +51,7 @@ public class CbomSyncTask implements ScheduledJobTask {
     }
 
     @Override
-    @Transactional(noRollbackFor = ScheduledJobSkippedException.class)
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
     public ScheduledTaskResult performJob(final ScheduledJobInfo scheduledJobInfo, final Object taskData) {
         if (!cbomService.isCbomRepositoryClientConfigured()) {
             throw new ScheduledJobSkippedException();
--- a/src/test/java/com/otilm/core/integration/messaging/jms/listeners/SchedulerListenerITest.java
+++ b/src/test/java/com/otilm/core/integration/messaging/jms/listeners/SchedulerListenerITest.java
@@ -1,0 +1,46 @@
+package com.otilm.core.integration.messaging.jms.listeners;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.test.context.bean.override.mockito.MockitoBean;
+
+import com.otilm.api.model.scheduler.SchedulerJobExecutionMessage;
+import com.otilm.core.dao.entity.ScheduledJob;
+import com.otilm.core.dao.repository.ScheduledJobsRepository;
+import com.otilm.core.messaging.jms.listeners.SchedulerListener;
+import com.otilm.core.service.impl.CbomServiceImpl;
+import com.otilm.core.tasks.CbomSyncTask;
+import com.otilm.core.util.BaseSpringBootTest;
+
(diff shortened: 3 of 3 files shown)


src/main/java/com/otilm/core/api/ScheduledJobSkippedException.java
```diff
@@ -2,5 +2,15 @@
...
 
+/**
+ * Signals that a {@link com.otilm.core.tasks.ScheduledJobTask#performJob} run was skipped
+ * (not an error) and should be treated as such by the caller.
+ * <p>
+ * If the throwing task class is {@code @Transactional}, it must declare
+ * {@code @Transactional(noRollbackFor = ScheduledJobSkippedException.class)} — otherwise the
+ * exception marks the ambient transaction (shared with {@code SchedulerListener}) rollback-only
+ * before it is caught, causing {@code UnexpectedRollbackException} on commit even though the
+ * skip itself is handled correctly.
```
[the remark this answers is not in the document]
klaraf755: Spring transactions are preferred, so javadoc referencing only those is enough.

lubomirw (dismissed):

src/test/java/com/otilm/core/integration/messaging/jms/listeners/SchedulerListenerITest.java
```diff
@@ -0,0 +1,46 @@
...
+        scheduledJob.setEnabled(true);
+        scheduledJob.setOneTime(false);
+        scheduledJob.setSystem(true);
+        scheduledJobsRepository.save(scheduledJob);
+
+        Mockito.when(cbomService.isCbomRepositoryClientConfigured()).thenReturn(false);
+
+        SchedulerJobExecutionMessage message = new SchedulerJobExecutionMessage(CbomSyncTask.NAME, CbomSyncTask.class.getName());
+
+        assertDoesNotThrow(() -> schedulerListener.processMessage(message));
```
lubomirw, line 58: **[Medium]** | Testing

This test asserts only that no exception is thrown. `SchedulerListener.processMessage` swallows `NotFoundException`/`SchedulerException` in its own catch, so if job lookup or bean resolution ever broke, the skip branch would never run and `assertDoesNotThrow` would still pass — a false green. Nothing here proves the skip path executed or that the ambient transaction actually committed.

_Suggested:_ add `Mockito.verify(cbomService).isCbomRepositoryClientConfigured();` to prove the path ran, and/or assert the `STARTED` `ScheduledJobHistory` entry was removed (`finalizeFinishedScheduledJob` deletes it on skip) to confirm the commit.

src/test/java/com/otilm/core/integration/messaging/jms/listeners/SchedulerListenerITest.java
```diff
@@ -0,0 +1,46 @@
...
+        ScheduledJob scheduledJob = new ScheduledJob();
+        scheduledJob.setJobName(CbomSyncTask.NAME);
+        scheduledJob.setJobClassName(CbomSyncTask.class.getName());
+        scheduledJob.setCronExpression("0 0 * ? * *");
+        scheduledJob.setEnabled(true);
+        scheduledJob.setOneTime(false);
+        scheduledJob.setSystem(true);
+        scheduledJobsRepository.save(scheduledJob);
+
+        Mockito.when(cbomService.isCbomRepositoryClientConfigured()).thenReturn(false);
```
lubomirw: **[Low]** | Testing

`CbomSyncTask` throws `ScheduledJobSkippedException` from two sites — unconfigured client (line 56) and a 503 `CbomRepositoryException` (line 64). This test exercises only the first through the listener; the 503 site's reliance on `noRollbackFor` is covered only by a direct `performJob` call in `CbomSyncTaskITest`, not through the shared-transaction listener path.

_Suggested (optional):_ add a second case — stub `isCbomRepositoryClientConfigured()==true` and `cbomService.sync()` to throw a 503 `CbomRepositoryException`, then assert `processMessage` does not throw.

src/main/java/com/otilm/core/tasks/CbomSyncTask.java
```diff
@@ -51,6 +50,7 @@ public boolean isSystemJob() {
     }
 
     @Override
+    @Transactional(noRollbackFor = ScheduledJobSkippedException.class)
```
lubomirw: **[Medium]** | Maintainability

The root cause is that `SchedulerListener` runs the whole dispatch in one shared `@Transactional`, which `performJob` (and any nested `@Transactional` service method) participates in. This fix correctly annotates `CbomSyncTask.performJob` and is complete today — `CbomSyncTask` is the only task that throws the skip exception. But any future `@Transactional` task that throws `ScheduledJobSkippedException` would silently reintroduce the `UnexpectedRollbackException`, relying on each author remembering `noRollbackFor`.

_Suggested (design):_ consider centralizing so it can't be forgotten — run `performJob` via `Propagation.REQUIRES_NEW`, or catch/translate the skip in `SchedulerServiceImpl.runScheduledJob` before it crosses a rollback-marking boundary. Otherwise, keeping the new Javadoc as the documented contract is a reasonable minimum.

src/main/java/com/otilm/core/api/ScheduledJobSkippedException.java
```diff
@@ -2,5 +2,15 @@
 
 import com.otilm.api.exception.PlatformException;
 
+/**
+ * Signals that a {@link com.otilm.core.tasks.ScheduledJobTask#performJob} run was skipped
+ * (not an error) and should be treated as such by the caller.
+ * <p>
+ * If the throwing task class is {@code @Transactional}, it must declare
```
lubomirw: **[Nit]** | Documentation

This says "If the throwing task **class** is `@Transactional`…", but the fix applies `noRollbackFor` at the `performJob` **method** level. The wording could mislead a future author who annotates at a different level.

_Suggested:_ reword to reference the task's transactional boundary generally (class- or method-level `@Transactional` on `performJob`).

lubomirw (commented):
Automated multi-reviewer pass (Copilot CLI, guided-pr-review, superpowers code-review, hygiene). The fix itself is correct and complete for the current codebase — `CbomSyncTask` is the only task that throws `ScheduledJobSkippedException`, both throw-sites are covered by `noRollbackFor`, and the new integration test reproduces the regression. Copilot and hygiene found nothing. The comments below are test-hardening and forward-looking maintainability notes.

src/test/java/com/otilm/core/integration/messaging/jms/listeners/SchedulerListenerITest.java
klaraf755: Added assertions that it followed the correct path.

src/test/java/com/otilm/core/integration/messaging/jms/listeners/SchedulerListenerITest.java
klaraf755: Added the second case.

src/main/java/com/otilm/core/tasks/CbomSyncTask.java
klaraf755: Used `Propagation.REQUIRES_NEW` in this case instead.

src/main/java/com/otilm/core/api/ScheduledJobSkippedException.java
klaraf755: Since I removed `noRollbackFor`, this javadoc is no longer present.

lubomirw (approved):

PR #1720, opened 2026-07-03
Issue certificates from a pre-registered request

ivosh:

Issue #1570, opened 2026-06-09
Pre-register certificate with challenge verified on issue/rekey


lubomirw:
### Description

Core support for pre-registering a certificate in a `REGISTERED` placeholder state that carries its
intended identity (RA Profile, DN, SAN, extensions), an operator-supplied challenge secret, and an
issuance window — then verifying the challenge and the window when the certificate is later issued
**or rekeyed**. Part of the ILM-owned self-service issuance flow (`ilm#130`); builds on the shipped
register→issue binding and consumes the interface contract in `interfaces#765`.

The challenge is stored **encrypted at rest** (via `SecretsUtil`, the pattern already used for SCEP/CMP
profile secrets) behind a dedicated store service — never as plaintext and never on the certificate row;
the certificate UUID identifies the authorization (username = certificate UUID). **The operator always
supplies the challenge — the platform never generates one.**

### User Story

As a certificate manager, I want to pre-register a certificate in ILM with a challenge and an issuance
window, so a user with no platform account can complete issuance (or rekey) themselves later — with the
credential lifecycle owned and verified by the platform rather than the upstream CA.

### Use Case

Certificates are frequently issued to parties with no platform account — external users, service or VPN
endpoints — via a public self-service portal. Moving the challenge and issuance-window lifecycle into ILM
makes the flow CA-agnostic and platform-governed, replacing the arrangement where the upstream CA owned
and verified the credential.

### Acceptance Criteria

- An operator can pre-register a certificate (`REGISTERED` placeholder) with its identity, an optional
  challenge, and an issuance window; it is visible in inventory with its properties.
- The challenge is stored encrypted at rest (never plaintext, never on the certificate row) and compared
  in constant time; the certificate UUID is the credential's username.
- Issuing a pre-registered certificate requires a matching challenge and a non-expired issuance window; an
  optional CSR may be supplied (or the key generated as in normal issue). Repeated failed attempts lock the
  authorization.
- Rekey of a pre-registered (now issued) certificate reuses the same challenge verification.
- Renewing a certificate that still has an active registration is refused (fail-closed) until challenge-gated
  renew is added separately.
- The challenge never appears in API responses, `toString`, logs, audit records, tracing, or approval records.
- Existing certificate issuance (with a CSR, no pre-registration) is unaffected.
- Persisting the pre-registration's SAN on the placeholder is out of scope here (display-only until protocol
  matching needs it) and is tracked with the later SCEP work.

PR #1735, opened 2026-07-07
Add registration-authorization entity and challenge store

lubomirw:
## What

Adds the durable **registration-authorization** foundation for the external self-service issuance flow (ilm#130): a per-certificate record holding the encrypted challenge, issuance window, failed-attempt lockout, and lifecycle state, plus the service that manages the challenge secret.

New:
- `CertificateRegistrationAuthorization` entity + `RegistrationState` enum — table `certificate_registration_authorization`, 1:1 with the certificate (FK `ON DELETE CASCADE`).
- `CertificateRegistrationAuthorizationRepository` — lookup + a `PESSIMISTIC_WRITE` finder (for the future verify / failed-attempt transaction) + delete.
- `RegistrationChallengeStore` — encrypts (via `SecretsUtil`), verifies **constant-time** (`MessageDigest.isEqual`, not `String.equals`), and resolves the plaintext (internal-only). The plaintext never touches the entity.
- Flyway migration `V202607071200__certificate_registration_authorization.sql`.

## Why a separate record

It coexists with the existing transient `CertificateRegistration` register→issue binding (#1720), which is cleared at issuance. This authorization record **survives** issuance, so a later renew/rekey can be authorized against the same challenge.

## Scope

**Dark / inert** — nothing wires this into the register/issue flow yet (verification is R1b, gated on interfaces#699). No existing files changed.

## Tests

- `RegistrationChallengeStoreTest` — encrypt→verify/resolve round-trip; wrong / null / absent-challenge rejection.
- `CertificateRegistrationAuthorizationRepositoryTest` — persistence, `findByCertificateUuid`, pessimistic finder, delete, and cascade on certificate delete.

8 tests green locally.

Part of #1570.

### DIFF
--- a/src/main/java/com/otilm/core/dao/entity/CertificateRegistrationAuthorization.java
+++ b/src/main/java/com/otilm/core/dao/entity/CertificateRegistrationAuthorization.java
@@ -15,10 +15,10 @@ import java.util.UUID;
 
 /**
  * Durable authorization for a pre-registered certificate: the challenge, issuance window, failed-attempt lockout,
- * and lifecycle state that let a later renew or rekey prove it is the registered subject. It coexists with the
- * transient {@link CertificateRegistration} register->issue binding but, unlike that binding, survives issuance —
- * the binding is cleared when the certificate is first issued, while this record remains to authorize follow-up
- * operations.
+ * and lifecycle state used to authorize a later renew or rekey.
+ *
+ * <p>Unlike the transient {@link CertificateRegistration} register-to-issue binding, which is cleared once the
+ * certificate is first issued, this record survives issuance so it can authorize follow-up operations.
  */
 @Getter
 @Setter
--- a/src/main/java/com/otilm/core/dao/entity/CertificateRegistrationAuthorization.java
+++ b/src/main/java/com/otilm/core/dao/entity/CertificateRegistrationAuthorization.java
@@ -1,0 +1,62 @@
+package com.otilm.core.dao.entity;
+
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.EnumType;
+import jakarta.persistence.Enumerated;
+import jakarta.persistence.Table;
+import jakarta.persistence.UniqueConstraint;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.ToString;
+
+import java.time.OffsetDateTime;
+import java.util.UUID;
+
+/**
+ * Durable authorization for a pre-registered certificate: the challenge, issuance window, failed-attempt lockout,
+ * and lifecycle state that let a later renew or rekey prove it is the registered subject. It coexists with the
+ * transient {@link CertificateRegistration} register->issue binding but, unlike that binding, survives issuance —
+ * the binding is cleared when the certificate is first issued, while this record remains to authorize follow-up
+ * operations.
+ */
+@Getter
+@Setter
+@ToString
+@Entity
+// A certificate has at most one durable authorization record.
+@Table(name = "certificate_registration_authorization",
+        uniqueConstraints = @UniqueConstraint(name = "uq_certificate_registration_authorization_certificate",
+                columnNames = "certificate_uuid"))
+public class CertificateRegistrationAuthorization extends UniquelyIdentifiedAndAudited {
+
+    @Column(name = "certificate_uuid", nullable = false)
+    private UUID certificateUuid;
+
+    // Ciphertext only; the plaintext challenge never touches the entity (see RegistrationChallengeStore).
+    // Excluded from toString because a tracing aspect records entity toString into spans.
+    @ToString.Exclude
+    @Column(name = "challenge", nullable = false, length = Integer.MAX_VALUE)
+    private String challenge;
+
+    // Null means no issuance deadline.
+    @Column(name = "expires_at")
+    private OffsetDateTime expiresAt;
+
+    @Column(name = "failed_attempts", nullable = false)
+    private int failedAttempts;
+
+    @Enumerated(EnumType.STRING)
+    @Column(name = "state", nullable = false)
+    private RegistrationState state;
+
+    @Override
+    public boolean equals(Object o) {
+        return super.equals(o);
+    }
+
+    @Override
+    public int hashCode() {
+        return super.hashCode();
+    }
+}
--- a/src/main/java/com/otilm/core/service/registration/RegistrationChallengeStore.java
+++ b/src/main/java/com/otilm/core/service/registration/RegistrationChallengeStore.java
@@ -21,6 +21,11 @@ public class RegistrationChallengeStore {
      * the row within its own transaction.
      */
     public void store(CertificateRegistrationAuthorization row, String plaintext) {
+        // Fail fast: a null/blank challenge would encode to null and only surface as an opaque NOT NULL
+        // violation when the row is saved.
+        if (plaintext == null || plaintext.isBlank()) {
+            throw new IllegalArgumentException("A registration challenge is required");
+        }
         row.setChallenge(SecretsUtil.encryptAndEncodeSecretString(plaintext, SecretEncodingVersion.V1));
     }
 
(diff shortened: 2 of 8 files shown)


src/main/java/com/otilm/core/service/registration/RegistrationChallengeStore.java
```diff
@@ -0,0 +1,48 @@
...
+@Service
+public class RegistrationChallengeStore {
+
+    /**
+     * Encrypts {@code plaintext} and writes the ciphertext onto {@code row}. Does not persist — the caller saves
+     * the row within its own transaction.
+     */
+    public void store(CertificateRegistrationAuthorization row, String plaintext) {
+        row.setChallenge(SecretsUtil.encryptAndEncodeSecretString(plaintext, SecretEncodingVersion.V1));
+    }
```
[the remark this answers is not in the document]
lubomirw: Fixed in c4c09865b — `store` now rejects a null/blank challenge up front (IllegalArgumentException) instead of encoding to null.

src/main/java/com/otilm/core/service/registration/RegistrationChallengeStore.java
```diff
@@ -0,0 +1,48 @@
...
+                stored.getBytes(StandardCharsets.UTF_8));
+    }
+
+    /**
+     * Decrypts and returns the plaintext challenge. Internal-only — for building the future registration event
+     * payload (SCEP / CMP); never exposed by any API.
+     */
+    public String resolvePlaintext(CertificateRegistrationAuthorization row) {
+        return SecretsUtil.decodeAndDecryptSecretString(row.getChallenge(), SecretEncodingVersion.V1);
+    }
```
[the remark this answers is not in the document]
lubomirw: Won't fix — `resolvePlaintext` is internal-only and only ever called on a persisted row, whose `challenge` column is NOT NULL, so the null case can't arise.

src/test/java/com/otilm/core/dao/repository/CertificateRegistrationAuthorizationRepositoryTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Test
+    void deletingParentCertificateCascadesTheAuthorization() {
+        // The certificate_uuid association is intentionally unmapped, so Hibernate's create-drop test schema omits
+        // the ON DELETE CASCADE foreign key that the migration ships. Recreate it here to exercise the cascade.
+        jdbcTemplate.execute("ALTER TABLE core.certificate_registration_authorization "
+                + "DROP CONSTRAINT IF EXISTS fk_certificate_registration_authorization_certificate");
+        jdbcTemplate.execute("ALTER TABLE core.certificate_registration_authorization "
+                + "ADD CONSTRAINT fk_certificate_registration_authorization_certificate "
+                + "FOREIGN KEY (certificate_uuid) REFERENCES core.certificate (uuid) ON DELETE CASCADE");
```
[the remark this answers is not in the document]
lubomirw: Fixed in c4c09865b — dropped that cascade test (it hand-recreated the FK, which ddl-auto omits, rather than exercising the migration) and moved the file into the integration root as an *ITest.

src/main/java/com/otilm/core/service/registration/RegistrationChallengeStore.java
```diff
@@ -0,0 +1,53 @@
...
+                presented.getBytes(StandardCharsets.UTF_8),
+                stored.getBytes(StandardCharsets.UTF_8));
+    }
+
+    /**
+     * Decrypts and returns the plaintext challenge. Internal-only — for building the future registration event
+     * payload (SCEP / CMP); never exposed by any API.
+     */
+    public String resolvePlaintext(CertificateRegistrationAuthorization row) {
+        return SecretsUtil.decodeAndDecryptSecretString(row.getChallenge(), SecretEncodingVersion.V1);
```
ivosh, line 56: **. Nitpick** | Error handling

RegistrationChallengeStore.java:51 calls SecretsUtil.decodeAndDecryptSecretString(row.getChallenge(), ...) with no null check on row.getChallenge(). Its sibling verify() (line 165) explicitly guards `row.getChallenge() == null` and returns false. If resolvePlaintext is ever called on a transient/unsaved authorization (no challenge set), SecretsUtil.isSecretStringV1 will NPE on null.split rather than surfacing a clear domain error. Because the challenge column is NOT NULL, any persisted row always has a value, so this cannot trigger on a loaded row today — hence nitpick. If the two accessors are meant to behave consistently on unpopulated rows, add the same null guard (or throw IllegalStateException with a clear message); otherwise leave as-is since the column invariant prevents it in practice.

src/main/java/com/otilm/core/dao/entity/CertificateRegistrationAuthorization.java
```diff
@@ -0,0 +1,62 @@
...
+import jakarta.persistence.UniqueConstraint;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.ToString;
+
+import java.time.OffsetDateTime;
+import java.util.UUID;
+
+/**
+ * Durable authorization for a pre-registered certificate: the challenge, issuance window, failed-attempt lockout,
```
ivosh, line 17: **. Nitpick** | Documentation

The class Javadoc (CertificateRegistrationAuthorization.java:16-22) hits multiple density markers flagged for OmniTrust Core: it fuses 3+ distinct concerns (what the record is, its field set, and its full relationship-and-lifecycle contrast with CertificateRegistration) into one wall-of-text with no structural lead-ins, and the second sentence is a ~40-word chain joined by an em-dash. The sibling `CertificateRegistration` Javadoc covers the same kind of ground far more readably. Suggest a one-line summary plus one concern per paragraph with bold lead-ins (e.g. purpose / relationship-to-CertificateRegistration). Non-blocking.

src/main/java/com/otilm/core/service/registration/RegistrationChallengeStore.java
```diff
@@ -0,0 +1,53 @@
+package com.otilm.core.service.registration;
+
+import com.otilm.core.dao.entity.CertificateRegistrationAuthorization;
+import com.otilm.core.util.SecretEncodingVersion;
+import com.otilm.core.util.SecretsUtil;
+import org.springframework.stereotype.Service;
+
+import java.security.MessageDigest;
+import java.nio.charset.StandardCharsets;
```
ivosh, line 8: **. Nitpick** | Maintainability

RegistrationChallengeStore.java:8-9 places `import java.security.MessageDigest` before `import java.nio.charset.StandardCharsets`, so `nio` sorts after `security` — out of ASCII/alphabetical order. Google Java Style 3.3.3 requires imports within a group to be ordered ASCII-sort. The codebase already follows this (e.g. CertificateHandler.java lists java.nio.charset.StandardCharsets before java.security.*). Swap the two lines. Trivial; an IDE optimize-imports pass fixes it.

ivosh (dismissed):

src/main/java/com/otilm/core/service/registration/RegistrationChallengeStore.java
lubomirw: Added the guard in 736cadbdb — `resolvePlaintext` now throws `IllegalStateException` with a clear message on an unpopulated row, consistent with `verify()`.

src/main/java/com/otilm/core/dao/entity/CertificateRegistrationAuthorization.java
lubomirw: Rewritten in 736cadbdb — a one-line summary plus a separate paragraph for the `CertificateRegistration` relationship.

src/main/java/com/otilm/core/service/registration/RegistrationChallengeStore.java
lubomirw: Fixed in 736cadbdb — swapped so `java.nio` sorts before `java.security`.

ivosh (approved):

PR #1741, opened 2026-07-08
Forward certificate custom attributes to upload on user create/update

### LINKED ISSUE #1725: Certificate custom attributes ignored when creating or updating a user with an uploaded certificate
### Description

When a user is created or updated with a newly uploaded certificate, Core does not forward the certificate's custom attributes to the certificate upload method — they are silently dropped and ignored.

Root cause (verified in Core `main` @ `a3ed76103`):
- `AddUserRequestDto` / `UpdateUserRequestDto` (`interfaces/src/main/java/com/otilm/api/model/client/auth/`) have no field for *certificate* custom attributes; their only `customAttributes` field is validated/stored against `Resource.USER`, not `Resource.CERTIFICATE`.
- The shared helper `UserManagementServiceImpl.addUserCertificate(userUuid, certificateUuid, certificateData)` has no parameter to carry certificate custom attributes.
- At `UserManagementServiceImpl.java:352` the new-upload branch calls `certificateUploadService.upload(certificateData, null, true)` — the custom-attributes argument is hard-coded to `null`.

Because `null` is passed, `CertificateUploadServiceImpl.upload(...)` (`CertificateUploadServiceImpl.java:68-70`) skips certificate custom-attribute validation (guarded on a non-empty list) and `CertificateUploadedEventHandler` (`CertificateUploadedEventHandler.java:148-154`) never persists any. The attributes are therefore neither validated nor stored on this path.

Consequence for required attributes: there is no way to supply a required certificate custom attribute when uploading a certificate through user create/update — the flow cannot satisfy required certificate custom attributes even when a value is entered in the UI or API. The fix must both carry the certificate custom attributes through the DTO + helper and forward them to `upload(...)` so they are validated (`AttributeEngine.validateCustomAttributesContent(Resource.CERTIFICATE, ...)`) and persisted. Compare the working direct cert-upload API at `CertificateServiceImpl.java:1229` / `:1237`, which passes the list through.

Reported by @soloviovmax on fe-administrator#1836 — https://github.com/OmniTrustILM/fe-administrator/issues/1836#issuecomment-4881117999. As a workaround, the certificate upload dialog's custom-attribute fields were disabled in fe-administrator; a Core backend change is required.

Note: this also touches `fe-administrator` (#1836), where the certificate upload dialog's custom-attribute fields were disabled pending the backend fix.

### Steps to Reproduce

1. Configure one or more custom attributes for the Certificate resource; mark at least one as required.
2. In the administrator UI, create or edit a user and upload a new certificate for that user, filling in the certificate custom attribute value(s).
3. Submit (equivalently via API: create/update user with `certificateData` set).
4. Inspect the uploaded certificate's custom attributes.

### Expected Behavior

Certificate custom attributes provided during user create/update are forwarded to the certificate upload method, validated against the Certificate resource definitions, and persisted on the uploaded certificate. Required certificate custom attributes can be supplied through this flow.

### Actual Behavior

The certificate custom attributes are not forwarded — `upload(...)` is called with `null`. They are silently dropped: never validated, never stored on the certificate. Required certificate custom attributes cannot be provided through the user create/update flow at all, even when entered in the UI or API.

### Severity

Major

### Module

Certificates

### Environment

Verified in Core `main` @ commit `a3ed76103`.

klaraf755:
Closes #1725

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/UserManagementServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/UserManagementServiceImpl.java
@@ -1,6 +1,7 @@
 package com.otilm.core.service.impl;
 
 import com.otilm.api.exception.*;
+import com.otilm.api.model.client.attribute.RequestAttribute;
 import com.otilm.api.model.client.auth.AddUserRequestDto;
 import com.otilm.api.model.client.auth.UpdateUserRequestDto;
 import com.otilm.api.model.client.auth.UserIdentificationRequestDto;
@@ -143,7 +144,7 @@ public class UserManagementServiceImpl implements UserManagementExternalService,
         UserRequestDto requestDto = new UserRequestDto();
         Certificate certificate = null;
         if ((request.getCertificateUuid() != null && !request.getCertificateUuid().isEmpty()) || (request.getCertificateData() != null && !request.getCertificateData().isEmpty())) {
-            certificate = addUserCertificate(null, request.getCertificateUuid(), request.getCertificateData());
+            certificate = addUserCertificate(null, request.getCertificateUuid(), request.getCertificateData(), request.getCertificateCustomAttributes());
             requestDto.setCertificateUuid(certificate.getUuid().toString());
             requestDto.setCertificateFingerprint(certificate.getFingerprint());
         }
@@ -155,9 +156,11 @@ public class UserManagementServiceImpl implements UserManagementExternalService,
         requestDto.setDescription(request.getDescription());
 
         List<NameAndUuidDto> groups = new ArrayList<>();
-        for (String groupUuid : request.getGroupUuids()) {
-            GroupDto groupDto = groupService.getGroup(SecuredUUID.fromString(groupUuid));
-            groups.add(new NameAndUuidDto(groupDto.getUuid(), groupDto.getName()));
+        if (request.getGroupUuids() != null) {
+            for (String groupUuid : request.getGroupUuids()) {
+                GroupDto groupDto = groupService.getGroup(SecuredUUID.fromString(groupUuid));
+                groups.add(new NameAndUuidDto(groupDto.getUuid(), groupDto.getName()));
+            }
         }
         requestDto.setGroups(groups);
 
@@ -326,7 +329,7 @@ public class UserManagementServiceImpl implements UserManagementExternalService,
         getUser(uuid.toString());
     }
 
-    private Certificate addUserCertificate(String userUuid, String certificateUuid, String certificateData) throws CertificateException, NotFoundException {
+    private Certificate addUserCertificate(String userUuid, String certificateUuid, String certificateData, List<RequestAttribute> certificateCustomAttributes) throws CertificateException, NotFoundException {
         Certificate certificate = null;
         boolean uploadCertificate = false;
         if (StringUtils.isNotBlank(certificateUuid)) {
@@ -349,7 +352,7 @@ public class UserManagementServiceImpl implements UserManagementExternalService,
 
         if (uploadCertificate) {
             try {
-                String fingerprint = certificateUploadService.upload(certificateData, null, true);
+                String fingerprint = certificateUploadService.upload(certificateData, certificateCustomAttributes, true);
                 certificate = certificateService.getCertificateEntityByFingerprint(fingerprint);
                 logger.getLogger().debug("New Certificate uploaded for the user");
             } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/otilm/core/service/impl/UserManagementServiceImpl.java
```diff
@@ -143,7 +144,7 @@ public UserDetailDto createUser(AddUserRequestDto request) throws CertificateExc
         UserRequestDto requestDto = new UserRequestDto();
         Certificate certificate = null;
         if ((request.getCertificateUuid() != null && !request.getCertificateUuid().isEmpty()) || (request.getCertificateData() != null && !request.getCertificateData().isEmpty())) {
```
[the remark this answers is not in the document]
klaraf755: Changed to use StringUtils.

src/main/java/com/otilm/core/service/impl/UserManagementServiceImpl.java
```diff
@@ -373,7 +384,7 @@ private UserDetailDto getUserUpdateRequestPayload(String userUuid, UpdateUserReq
         UserUpdateRequestDto requestDto = new UserUpdateRequestDto();
 
         if ((request.getCertificateUuid() != null && !request.getCertificateUuid().isEmpty()) || (request.getCertificateData() != null && !request.getCertificateData().isEmpty())) {
```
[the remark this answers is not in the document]
klaraf755: Changed to use StringUtils.

src/main/java/com/otilm/core/service/impl/UserManagementServiceImpl.java
```diff
@@ -349,7 +357,7 @@ private Certificate addUserCertificate(String userUuid, String certificateUuid,
 
         if (uploadCertificate) {
             try {
-                String fingerprint = certificateUploadService.upload(certificateData, null, true);
+                String fingerprint = certificateUploadService.upload(certificateData, certificateCustomAttributes, true);
```
lubomirw: **[Medium]** | Error handling

`certificateUploadService.upload(...)` now validates the certificate custom attributes and can throw `ValidationException` (mapped to HTTP 422). But `addUserCertificate`'s `catch (Exception e)` (L356) re-wraps it as `CertificateException` (mapped to HTTP 500), so invalid client-supplied attributes surface as an internal server error instead of a client validation error. Newly reachable — previously `null` was passed, so this validation never ran.

_Suggested:_ catch `ValidationException` (and other client-input exceptions) before the broad catch and rethrow unchanged, or validate the attributes before entering the upload `try` block so the original 422 is preserved.

<sub>flagged by Copilot CLI review; verified against the catch body</sub>

src/main/java/com/otilm/core/service/impl/UserManagementServiceImpl.java
```diff
@@ -326,7 +322,19 @@ public void evaluatePermissionChain(SecuredUUID uuid) throws NotFoundException {
...
+            return groups;
+        }
+        for (String groupUuid : groupUuids) {
+            GroupDto groupDto = groupService.getGroup(SecuredUUID.fromString(groupUuid));
+            groups.add(new NameAndUuidDto(groupDto.getUuid(), groupDto.getName()));
+        }
+        return groups;
+    }
+
+    private Certificate addUserCertificate(String userUuid, String certificateUuid, String certificateData, List<RequestAttribute> certificateCustomAttributes) throws CertificateException, NotFoundException {
```
lubomirw, line 337: **[Medium]** | Testing

This bug fix (the #1725 regression — custom attributes hard-coded to `null`) ships with no regression test. Nothing asserts that `upload(...)` now receives the forwarded `RequestAttribute` list, nor covers the new "attributes ignored, certificate already exists" WARN branch or the `resolveGroups` null-guard.

_Suggested:_ extend `UserManagementServiceTest`: (a) create/update with `certificateData` + custom attributes → assert `upload(...)` is invoked with the forwarded list; (b) existing cert (by UUID and by fingerprint) + attributes → assert they are not applied and the WARN branch is hit; (c) `groupUuids=null` on create → user created with empty groups, no NPE.

lubomirw (commented):
Automated multi-reviewer pass (Copilot CLI · guided-pr-review · superpowers · hygiene). Two actionable findings inline below.

src/main/java/com/otilm/core/service/impl/UserManagementServiceImpl.java
klaraf755: Added rethrow.

src/main/java/com/otilm/core/service/impl/UserManagementServiceImpl.java
klaraf755: Added tests.

lubomirw (dismissed):

lubomirw (approved):

PR #1747, opened 2026-07-08
Fix 500 errors when deleting connectors with dependent records

### LINKED ISSUE #1730: Connectors - Delete: 500 Internal Server Error when connector still referenced by Notification Instances (FK constraint instead of pre-check)
### Description

`DELETE /api/v2/connectors/{uuid}` returns HTTP **500 Internal Server Error** when the connector is still referenced by one or more Notification Instances. The backend does not check for dependent Notification Instances before attempting the deletion — instead, the SQL `DELETE` propagates all the way to PostgreSQL, which rejects it with a foreign key constraint violation. The exception surfaces to the user as a generic 500 with no actionable message.

### Steps to Reproduce

1. Configure any connector that provides notifications (e.g. **Email-Notification-Provider**)
2. Create at least one **Notification Instance** using that connector
3. Navigate to **Inventory → Connectors** and attempt to delete the connector from step 1

### Expected Behavior

- Backend detects the dependent Notification Instance(s) **before** executing the DB delete
- Returns a proper 4xx response (e.g. `HTTP 409 Conflict` or `422 Unprocessable Entity`) with a clear message listing which Notification Instances still reference the connector
- Consistent with existing pre-delete validation applied elsewhere in the platform (e.g. Notification Instance itself refuses delete when referenced by a Notification Profile)

### Actual Behavior

- Backend returns HTTP **500 Internal Server Error** — no user-friendly message
- Core log shows PostgreSQL error propagated all the way up:

```
ERROR [tomcat-handler-928] [SqlExceptionHelper]: ERROR: update or delete on table "connector"
violates foreign key constraint "notification_instance_reference_connector_uuid_fkey"
on table "notification_instance_reference"
Detail: Key (uuid)=(<connector-uuid>) is still referenced from table "notification_instance_reference".
```

Audit log records the operation as `operationResult: failure` with a generic message; the user has no way to know *why* the delete failed without pulling core logs.

### Severity

Major

### Module

Core

### Attachments

Image

[7.7.26_core_logs.txt](https://github.com/user-attachments/files/29744136/7.7.26_core_logs.txt)

### Logs

_No response_

### Environment

ILM Core: 2.18.1-SNAPSHOT (7 Jul 2026, 10:59)
Commit: 95dc6a7
Branch: main
Build time: 2026-07-07T08:59:10Z
ILM Frontend: 2.18.0 (7 Jul 2026, 08:18)
Client:
- macOS Tahoe 26.5.1

### LINKED ISSUE #1732: Connectors - Force Delete: 500 Internal Server Error - dependent attribute_definition rows not cleared, "force" fails to force
### Description

`DELETE /api/v1/connectors/force` is intended as the escape hatch when regular connector deletion is blocked by dependencies — it should remove the connector regardless of what still references it. In practice, the endpoint fails with **HTTP 500 Internal Server Error** whenever the target connector is still referenced by rows in `attribute_definition`: the delete statement propagates all the way to PostgreSQL, which rejects it with a foreign key constraint violation on `fk_attribute_definition_connector`. No dependent rows are cleared before the delete, and no user-facing error is returned — the response is a generic 500.

This defeats the purpose of the "force" variant of the endpoint. It also mirrors the class of bug already seen on regular delete (see #1730), but on a different endpoint (`v1/force`) and different FK (`attribute_definition` vs `notification_instance_reference`).

### Steps to Reproduce

1. Register any connector that produces attribute definitions (e.g. **EJBCA-NG-Connector**)
2. Configure any entity that persists attributes originating from this connector (Authority Instance / RA Profile / Notification Instance / etc.), which populates `core.attribute_definition` rows pointing at the connector UUID
3. Trigger force delete of that connector via:
   - **UI:** Inventory → Connectors → select → Force Delete
   - **API:** `DELETE /api/v1/connectors/force` with body `{"uuids":["<connector-uuid>"]}`

### Expected Behavior

The force-delete endpoint should honor its "force" semantics — one of:

- **Cascade approach:** delete dependent `attribute_definition` (and any other referencing) rows first, then delete the connector, returning `2xx`
- **Validation approach:** if force cannot cascade, return a proper `4xx` response (e.g. `409 Conflict` / `422 Unprocessable Entity`) with a clear message listing which tables still reference the connector, so the user knows what to clean up manually

Either way, the user must not see a generic `500`.

### Actual Behavior

- Response: HTTP **500 Internal Server Error**, no user-facing message
- Core log shows the constraint violation propagated from the DB:

```
ERROR [tomcat-handler-995] [SqlExceptionHelper]: ERROR: update or delete on table "connector"
violates foreign key constraint "fk_attribute_definition_connector" on table "attribute_definition"
Detail: Key (uuid)=(<connector-uuid>) is still referenced from table "attribute_definition".
```

- Audit log records `operation: forceDelete`, `operationResult: failure`, generic message body
- Connector remains in the inventory; user has no indication of what to remove to make deletion succeed

### Severity

Major

### Module

Core

### Attachments

[7.7.26_core2_logs.txt](https://github.com/user-attachments/files/29744476/7.7.26_core2_logs.txt)

### Logs

_No response_

### Environment

ILM Core: 2.18.1-SNAPSHOT (7 Jul 2026, 10:59)
Commit: 95dc6a7
Branch: main
Build time: 2026-07-07T08:59:10Z
ILM Frontend: 2.18.0 (7 Jul 2026, 08:18)
Client:
- macOS Tahoe 26.5.1

klaraf755:
Fixes #1730 
Fixes #1732


src/test/java/com/otilm/core/service/ConnectorServiceTest.java
```diff
@@ -346,6 +350,17 @@ void testRemoveConnector_notFound() {
         Assertions.assertThrows(NotFoundException.class, () -> connectorService.deleteConnector(SecuredUUID.fromString("abfbc322-29e1-11ed-a261-0242ac120002")));
     }
 
+    @Test
+    void testRemoveConnector_withNotificationInstanceReferences_throws() {
+        NotificationInstanceReference notificationInstance = new NotificationInstanceReference();
+        notificationInstance.setName("testNotificationInstance");
+        notificationInstance.setConnectorUuid(connector.getUuid());
+        notificationInstanceReferenceRepository.save(notificationInstance);
+        SecuredUUID connectorSecuredUuid = connector.getSecuredUuid();
```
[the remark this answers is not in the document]
klaraf755: It will not fail, the test suite does not use migration schemas, but creates the DB with Hibernate. The `NotificationInstanceReference` entity does not have the field as nullable.

src/test/java/com/otilm/core/service/ConnectorServiceTest.java
```diff
@@ -346,6 +356,34 @@ void testRemoveConnector_notFound() {
         Assertions.assertThrows(NotFoundException.class, () -> connectorService.deleteConnector(SecuredUUID.fromString("abfbc322-29e1-11ed-a261-0242ac120002")));
     }
 
+    @Test
+    void testRemoveConnector_withNotificationInstanceReferences_throws() {
+        NotificationInstanceReference notificationInstance = new NotificationInstanceReference();
+        notificationInstance.setName("testNotificationInstance");
+        notificationInstance.setConnectorUuid(connector.getUuid());
+        notificationInstanceReferenceRepository.save(notificationInstance);
```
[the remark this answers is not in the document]
klaraf755: Same as https://github.com/OmniTrustILM/core/pull/1747#discussion_r3544487230.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -1413,6 +1413,12 @@ public void deleteConnectorAttributeDefinitionsContent(UUID connectorUuid) {
         Long deletedDefinitions = attributeDefinitionRepository.deleteByTypeAndConnectorUuid(AttributeType.DATA, connectorUuid);
         logger.debug("Deleted {} data attribute definitions for connector with UUID {}", deletedDefinitions, connectorUuid);
 
+        // delete group attributes, since the attribute content items are not never created for GROUP attributes,
+        // it is enough to delete the attribute definitions
```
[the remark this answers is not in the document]
klaraf755: Fixed.

src/main/java/com/otilm/core/service/v2/impl/ConnectorServiceImpl.java
```diff
@@ -616,6 +617,10 @@ private void deleteConnector(Connector connector) {
             errors.add("Dependent Vault Instances: " + String.join(", ", connector.getVaultInstances().stream().map(VaultInstance::getName).collect(Collectors.toSet())));
         }
 
+        if (!connector.getNotificationInstanceReferences().isEmpty()) {
```
lubomirw, line 636: **[Medium]** | Correctness

`forceDeleteConnector` calls `removeConnectorAssociations()` and then this private `deleteConnector()`. `removeConnectorAssociations` detaches credentials, authority/entity/token instance refs, vault instances, and compliance rules — but not notification instance references. So this new validation branch now also blocks a *force* delete of a connector referenced by a notification instance. This is not a regression (that path previously failed on an FK-constraint 500), but force delete is meant to override such dependencies. Pre-existing / arguably out of scope for this fix — flagging for a decision.

_Suggested:_ Either add a detach loop for notification references in `removeConnectorAssociations` (mirroring the vault/token handling) so force delete can remove them, or document that notification instances are a deliberate hard blocker even under force delete.

src/test/java/com/otilm/core/service/ConnectorServiceTest.java
```diff
@@ -346,6 +350,17 @@ void testRemoveConnector_notFound() {
         Assertions.assertThrows(NotFoundException.class, () -> connectorService.deleteConnector(SecuredUUID.fromString("abfbc322-29e1-11ed-a261-0242ac120002")));
     }
 
+    @Test
+    void testRemoveConnector_withNotificationInstanceReferences_throws() {
```
lubomirw, line 360: **[Low]** | Testing

The added test covers only single `deleteConnector`. No test exercises `forceDeleteConnector` / `bulkDeleteConnector` against a connector that has notification instance references, so the behavior tied to the point above is unverified.

_Suggested:_ Add a test asserting the intended force/bulk delete behavior for a notification-referenced connector (either it succeeds and detaches the reference, or it reports the expected failure message).

lubomirw (commented):
Automated multi-reviewer triage — 2 findings (COMMENT). The code change is correct, consistent with the sibling dependency checks, and tested; these are non-blocking design/coverage notes about force-delete for the newly-checked notification-instance dependency.

src/main/java/com/otilm/core/service/v2/impl/ConnectorServiceImpl.java
klaraf755: Added the branch for notification instances in force delete.

src/test/java/com/otilm/core/service/ConnectorServiceTest.java
klaraf755: Added tests.

lubomirw (approved):

PR #1756, opened 2026-07-09
Speed up CodeQL and Sonar CI jobs

lubomirw:
## What

Reduce CI time-to-green by trimming two workflows that sit on (or near) the PR critical path.

### CodeQL (`codeql.yml`)
Switch to `build-mode: none` — CodeQL analyzes source directly instead of running a full `mvn package`. Removes the JDK setup, Maven cache, GH Packages credential steps, and the build step. The analyze job no longer depends on dependency resolution and finishes in roughly half the time.

**Trade-off (accepted):** under `build-mode: none`, annotation-processor output (Lombok accessors, QueryDSL Q-classes) and dependency bytecode are not analyzed, so dataflow through generated code is weaker than under the previous compiled analysis. Hand-written Java — the real attack surface — is still fully scanned, and `none` is GitHub's recommended default. Note: `build-mode: none` scans Java only; the repo has no Kotlin today, but if Kotlin is added it would need `autobuild`/`manual` to be analyzed.

### Sonar (`sonar.yaml`)
- Add a Maven dependency cache so the job stops re-downloading the scanner plugin and its dependencies on every `workflow_run` invocation.
- Add an `actions/cache` for the SonarCloud scanner engine (`~/.sonar/cache`).
- Pin the scan JDK to Temurin 21, aligning it with the JDK used to compile in `build_pr.yml` (previously the job relied on the runner default).

## Not touched
`build_pr.yml` is unchanged — `-U` (snapshot update) is deliberately kept on all Maven calls so PRs always build against the freshest `interfaces` snapshot.

## Verification
CodeQL, Sonar, and the existing Build PR checks run on this PR against the real diff.


.github/workflows/sonar.yaml
```diff
@@ -63,6 +63,17 @@ jobs:
...
+        with:
+          java-version: 21
+          distribution: 'temurin'
+          cache: maven
+      - name: Cache SonarCloud packages
+        uses: actions/cache@v5
+        with:
+          path: ~/.sonar/cache
+          key: ${{ runner.os }}-sonar
       # SonarCloud Scan analysis and push
```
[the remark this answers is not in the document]
lubomirw: Fixed in b6175fe85 — keyed the cache on `${{ github.run_id }}` with a `restore-keys: ${{ runner.os }}-sonar` prefix, so each run restores the latest cache and saves a fresh one as the scanner engine changes.

.github/workflows/codeql.yml
```diff
@@ -32,66 +21,17 @@ jobs:
...
-        language: [ 'java' ]
-        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
-        # Use only 'java' to analyze code written in Java, Kotlin or both
-        # Use only 'javascript' to analyze code written in JavaScript, TypeScript or both
-        # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
+        language: [ 'java' ]  # build-mode: none scans Java source only; Kotlin would need autobuild/manual
 
     steps:
     - name: Checkout repository
       uses: actions/checkout@v6
```
ivosh: Nitpick: action is not pinned to a digest but rather to a version. Confirm this is deliberate.

.github/workflows/codeql.yml
```diff
@@ -32,66 +21,17 @@ jobs:
...
-
-    #   If the Autobuild fails above, remove it and uncomment the following three lines.
-    #   modify them (or add more) to build your code if your project, please refer to the EXAMPLE below for guidance.
-
-    # - run: |
-    #     echo "Run, Build Application using script"
-    #     ./location_of_script_within_repo/buildscript.sh
-    
-    - name: Set up JDK 21
-      uses: actions/setup-java@v5
```
ivosh, line 76: Nitpick: action is not pinned to a digest but rather to a version. Confirm this is deliberate.

.github/workflows/codeql.yml
```diff
@@ -32,66 +21,17 @@ jobs:
...
-      with:
-        java-version: 21
-        distribution: 'temurin'
-        server-id: github # value of repository/id field of the pom.xml
-        server-username: GITHUB_USER_REF
-        server-password: GITHUB_TOKEN_REF
-    - name: Cache Maven packages
-      uses: actions/cache@v5
```
ivosh, line 84: Nitpick: action is not pinned to a digest but rather to a version. Confirm this is deliberate.

.github/workflows/sonar.yaml
```diff
@@ -63,6 +63,18 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           run-id: ${{ github.event.workflow_run.id }}
           path: target/site/jacoco
+      - name: Set up JDK 21
+        uses: actions/setup-java@v5
```
ivosh: Nitpick: action is not pinned to a digest but rather to a version. Confirm this is deliberate.

.github/workflows/sonar.yaml
```diff
@@ -63,6 +63,18 @@ jobs:
...
+        with:
+          java-version: 21
+          distribution: 'temurin'
+          cache: maven
+      - name: Cache SonarCloud packages
+        uses: actions/cache@v5
```
ivosh: Nitpick: action is not pinned to a digest but rather to a version. Confirm this is deliberate.

.github/workflows/sonar.yaml
```diff
@@ -63,6 +63,18 @@ jobs:
...
+        with:
+          path: ~/.sonar/cache
+          key: ${{ runner.os }}-sonar-${{ github.run_id }}
```
ivosh: Minor: Sonar cache key includes `github.run_id`, so the primary key never hits.
Restore falls through to restore-keys (still works) but every run writes a new cache entry.

Fix: drop `run_id`, use a stable key: `${{ runner.os }}-sonar`.
SonarSource's own documented example uses the stable key: `${{ runner.os }}-sonar` precisely to avoid this.

ivosh (changes requested):

.github/workflows/sonar.yaml
ivosh: Upon cross-checking with https://github.com/OmniTrustILM/core/pull/1756#discussion_r3551329017, I withdraw this finding. Copilot's concern is more relevant for our use case.

ivosh (dismissed):

.github/workflows/codeql.yml
lubomirw: Deliberate now: pinned to a full commit SHA with a version comment in 26eafd00f, matching the org's shared-workflow convention (`.github/.github/workflows/containers-test.yml`). Major versions unchanged.

.github/workflows/codeql.yml
lubomirw: Deliberate now: pinned to a full commit SHA with a version comment in 26eafd00f, matching the org's shared-workflow convention (`.github/.github/workflows/containers-test.yml`). Major versions unchanged.

.github/workflows/codeql.yml
lubomirw: Deliberate now: pinned to a full commit SHA with a version comment in 26eafd00f, matching the org's shared-workflow convention (`.github/.github/workflows/containers-test.yml`). Major versions unchanged.

.github/workflows/sonar.yaml
lubomirw: Deliberate now: pinned to a full commit SHA with a version comment in 26eafd00f, matching the org's shared-workflow convention (`.github/.github/workflows/containers-test.yml`). Major versions unchanged.

.github/workflows/sonar.yaml
lubomirw: Deliberate now: pinned to a full commit SHA with a version comment in 26eafd00f, matching the org's shared-workflow convention (`.github/.github/workflows/containers-test.yml`). Major versions unchanged.

.github/workflows/sonar.yaml
lubomirw: Agreed — reverted to the stable `${{ runner.os }}-sonar` key in 26eafd00f. You're right that the scanner cache is content-addressed and low-churn, so keying on run_id only caused a fresh cache write every run with no primary-key hit. Matches SonarSource's documented example.

.github/workflows/sonar.yaml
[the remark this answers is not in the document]
lubomirw: Update: reverted the run_id key back to the stable `${{ runner.os }}-sonar` in 26eafd00f. On reflection (and per reviewer feedback citing SonarSource's documented example), the scanner cache is content-addressed and low-churn, so a stable key is the right call here — the run_id variant wrote a new cache entry every run without ever hitting the primary key.

.github/workflows/sonar.yaml
lubomirw: Settled both comments with a monthly rolling key in 7824879cb:
```yaml
key: ${{ runner.os }}-sonar-${{ steps.cache-month.outputs.yearmonth }}   # yearmonth = date +%Y-%m
restore-keys: |
  ${{ runner.os }}-sonar-
```
Verified mechanics of `actions/cache`: an exact primary-key hit saves nothing at job end; a restore-keys-only hit *does* save a new entry. So both observations were mechanically correct — a constant key never refreshes (Copilot), and a `run_id` key writes a new entry every run (ivosh).

Why this key: `~/.sonar/cache` is content-addressed, so a stale entry never yields wrong analysis — only an incremental re-download. This repo scans **SonarCloud**, whose engine/analyzer versions roll server-side ~monthly, so a permanently constant key silently decays and no repo-derived version handle exists. The month-stamped key gives exact hits (zero upload) for all runs within a month, and one restore+save on the first run of each month — keeping the near-zero churn a stable key gives (this job runs in default-branch context, so `run_id` would upload ~100-200MB per PR update and LRU-evict the larger Maven caches) while still following SonarCloud's rollouts. Thanks both.

.github/workflows/sonar.yaml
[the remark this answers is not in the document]
lubomirw: Settled both comments with a monthly rolling key in 7824879cb:
```yaml
key: ${{ runner.os }}-sonar-${{ steps.cache-month.outputs.yearmonth }}   # yearmonth = date +%Y-%m
restore-keys: |
  ${{ runner.os }}-sonar-
```
Verified mechanics of `actions/cache`: an exact primary-key hit saves nothing at job end; a restore-keys-only hit *does* save a new entry. So both observations were mechanically correct — a constant key never refreshes (Copilot), and a `run_id` key writes a new entry every run (ivosh).

Why this key: `~/.sonar/cache` is content-addressed, so a stale entry never yields wrong analysis — only an incremental re-download. This repo scans **SonarCloud**, whose engine/analyzer versions roll server-side ~monthly, so a permanently constant key silently decays and no repo-derived version handle exists. The month-stamped key gives exact hits (zero upload) for all runs within a month, and one restore+save on the first run of each month — keeping the near-zero churn a stable key gives (this job runs in default-branch context, so `run_id` would upload ~100-200MB per PR update and LRU-evict the larger Maven caches) while still following SonarCloud's rollouts. Thanks both.

ivosh (approved):

PR #1760, opened 2026-07-09
Fix evaluation of conditions for attributes with list content

### LINKED ISSUE #1757: Workflows - Rules - Conditions: Condition on attribute with list content is evaluated as false when it should be true
### Description

When a workflow rule contains a condition that compares a custom attribute whose content is a list of values (multi-value attribute content), the condition is evaluated incorrectly. The condition resolves to false even when the attribute content contains a value that satisfies the condition, so the rule is not applied and the associated workflow actions are never executed.

Conditions on single-value attributes evaluate as expected; the problem is specific to attributes with list content. This silently breaks workflow automation — rules appear correctly configured in the UI, but never trigger.

### Steps to Reproduce

1. Define a custom attribute whose content is a list of values (multi-value content).
2. Assign the attribute to a resource object (e.g. a certificate) with content that includes a matching value.
3. In Workflows, create a Rule with a Condition on that custom attribute (e.g. equals one of the values present in the list).
4. Create a workflow action and associate it with the rule.
5. Trigger the corresponding event so the rule is evaluated against the object.

### Expected Behavior

- The condition evaluates to `true` when the attribute's list content contains a value satisfying the condition.

- The rule is applied and the associated actions are executed.

### Actual Behavior

- The condition evaluates to `false` even though the attribute's list content contains a satisfying value.
- The rule is not applied and the associated actions are not executed.

### Severity

Major

### Module

None

### Screenshot

_No response_

### Logs

_No response_

### Environment

_No response_

klaraf755:
Fixes #1757


src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
```diff
@@ -657,14 +657,21 @@ private boolean evaluateConditionOnAttribute(List<? extends AttributeContent> co
...
             Object attributeValue = contentType.isFilterByData() ? attributeContent.getData() : attributeContent.getReference();
             try {
-                if (Boolean.TRUE.equals(fieldTypeToOperatorActionMap.get(contentTypeToFieldType(contentType)).get(effectiveOperator).apply(attributeValue, conditionValue)))
-                    // Positive match found: for non-negated ops return true, for negated ops return false (a match disqualifies NOT_EQUALS/NOT_CONTAINS/NOT_MATCHES)
-                    return !isNegated;
+                for (Object conditionValueItem : conditionValueList) {
+                    if (Boolean.TRUE.equals(fieldTypeToOperatorActionMap.get(contentTypeToFieldType(contentType)).get(effectiveOperator).apply(attributeValue, conditionValueItem)))
+                        // Positive match found: for non-negated ops return true, for negated ops return false (a match disqualifies NOT_EQUALS/NOT_CONTAINS/NOT_MATCHES)
+                        return !isNegated;
+                }
```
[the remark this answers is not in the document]
klaraf755: Fixed. Also, since extracting the BiFunction has surfaced a SonarQube issue, as a result changed BiFunction to BiPredicate.

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
```diff
@@ -657,14 +655,22 @@ private boolean evaluateConditionOnAttribute(List<? extends AttributeContent> co
         // For non-negated operators, attributes without content do not match the operator.
         if (missingContent) return false;
 
+        // A multi-select list attribute condition arrives as a JSON array of selected values; normalize a scalar
+        // value to a one-element list so EQUALS means "equals any of" (IN semantics), mirroring
+        // FilterPredicatesBuilder.prepareAttributeFilterValues. Collections.singletonList permits a null value.
+        List<?> conditionValueList = conditionValue instanceof List ? (List<?>) conditionValue : Collections.singletonList(conditionValue);
```
lubomirw, line 668: **[Medium]** | Correctness

List-valued `conditionValue` is now expanded for every operator. That gives `CONTAINS`, `STARTS_WITH`, `MATCHES`, numeric/date comparisons, and similar operators any-of semantics in trigger evaluation, while `FilterPredicatesBuilder` uses only the first value for custom-attribute non-equality operators.

_Suggested:_ Restrict list expansion to `EQUALS`/`NOT_EQUALS`, or intentionally align query-side semantics and add tests for non-equality list values.

<sub>flagged by Copilot CLI; verdict: Needs judgment - query parity risk</sub>

lubomirw (commented):
Automated multi-reviewer triage - 1 finding (COMMENT).

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
klaraf755: Restricted list `conditionValue` only `EQUALS/NOT_EQUALS`.

lubomirw (approved):

PR #1765, opened 2026-07-10
Introduce AuthorizationEnforcer, the imperative authorization front door

### LINKED ISSUE #1750: Replace PermissionEvaluator AOP trampoline with imperative AuthorizationEnforcer
## Description

`com.otilm.core.service.PermissionEvaluator` is an imperative authorization API implemented as an AOP trampoline: all 14 methods have empty bodies, and the actual OPA check runs only because `AuthPermissionEvaluationServiceImpl` methods carry `@ExternalAuthorization` and are invoked through the Spring proxy. This design has structural problems:

- **Allow-by-default failure modes.** The security effect depends entirely on proxy interception. Self-invocation, plain instantiation in tests, or a forgotten `@ExternalAuthorization` on a newly added evaluator method silently degrades to an empty method that grants everything.
- **Method-per-(resource, action) explosion.** Each new permission pair grows the interface with another empty copy-paste method whose name encodes data (`tokenInstanceMembers` = TOKEN/MEMBERS). Call sites do not reveal what is checked.
- **Lying contracts.** Javadoc declares `@throws NotFoundException when ... not found`, which an empty body can never do; the real effect (`AccessDeniedException` from the proxy) is invisible in the signature. The name collides with Spring Security's `PermissionEvaluator` and the interface sits in `com.otilm.core.service` although it is authz infrastructure.

The root cause: the check logic (OPA query + group/owner-association fallback) lives only inside `ExternalMethodAuthorizationManager` and is reachable exclusively via the method-security interceptor. The evaluator exists because the check has no direct entry point.

**Target design** — give the check logic a front door:

1. Introduce `com.otilm.core.security.authz.AuthorizationEnforcer` with `void enforce(Resource resource, ResourceAction action, SecuredUUID... objectUuids)` (plus a `List` overload), throwing `AccessDeniedException` on denial. Extract the core check (OPA request construction, group/owner-association fallback, deny-on-error semantics) from `ExternalMethodAuthorizationManager` into its implementation.
2. Keep `ExternalMethodAuthorizationManager` as a thin adapter: extract (resource, action, object UUIDs, `SecurityFilter` presence, parent attributes) from the annotation and `MethodInvocation`, delegate to the same core. The declarative `@ExternalAuthorization` / `@ExternalAuthorizationDynamic` path is unchanged in behavior.
3. Migrate all `PermissionEvaluator` call sites (~35 across ~12 services plus `AttributeCallbackScopeResolver`) to explicit `enforcer.enforce(...)` calls, using the mapping frozen in `AuthPermissionEvaluationServiceImpl`: `tokenProfile`→TOKEN_PROFILE/DETAIL, `tokenProfiles`→TOKEN_PROFILE/LIST, `tokenInstance`→TOKEN/DETAIL, `tokenInstanceMembers`→TOKEN/MEMBERS, `certificate`→CERTIFICATE/DETAIL, `authorityInstance`→AUTHORITY/DETAIL, `raProfile`→RA_PROFILE/DETAIL, `entityInstance`→ENTITY/DETAIL, `vaultProfileMembers`→VAULT_PROFILE/MEMBERS, `acmeProfile`→ACME_PROFILE/DETAIL, `signingProfile(s)`→SIGNING_PROFILE/DETAIL, `vaultInstance`→VAULT/DETAIL, `tspProfileTimestamping`→TSP_PROFILE/TIMESTAMP.
4. Delete `PermissionEvaluator` and `AuthPermissionEvaluationServiceImpl`.

This is a zero-behavior-change refactor: same OPA policy (`method`), same principal handling (`PlatformAuthenticationToken` and `AnonymousAuthenticationToken`), same group/owner fallback, same exception type reaching callers.

## Definition of Done

- [ ] `AuthorizationEnforcer` interface + implementation in `com.otilm.core.security.authz`; core check logic shared with `ExternalMethodAuthorizationManager` (single implementation of the semantics)
- [ ] Declarative annotation path behavior unchanged (existing authorization tests green)
- [ ] All `PermissionEvaluator` call sites migrated to `enforcer.enforce(...)` with the resource/action mapping above
- [ ] `PermissionEvaluator` and `AuthPermissionEvaluationServiceImpl` deleted
- [ ] Unit tests for `AuthorizationEnforcer` (grant, deny, group/owner fallback, anonymous principal, deny-on-OPA-error)
- [ ] Affected service tests updated to mock/verify `AuthorizationEnforcer` explicitly
- [ ] Full build green; new-lines coverage meets the SonarCloud gate

ivosh:
Extract ExternalAuthorizationCore from ExternalMethodAuthorizationManager so the OPA request construction, parent-resource-first check, owner/group-association fallback, and deny-on-error semantics operate on a resolved AuthorizationRequest and are shared by both the declarative @ExternalAuthorization AOP path and a new imperative AuthorizationEnforcer.enforce(...) front door (throws AccessDeniedException on denial, fails closed on unknown/absent authentication).

Delete the PermissionEvaluator interface and AuthPermissionEvaluationServiceImpl (empty @ExternalAuthorization-annotated methods enforced via AOP) and migrate every call site to authorizationEnforcer.enforce(Resource, ResourceAction, uuid), preserving each site's exact resource/action.

Add @ExternalAuthorizationProgrammatic to keep programmatically-enforced resource/ action pairs in the startup auth-service sync payload; apply it to the two TSP timestamping methods (TSP_PROFILE:TIMESTAMP) and scan it in ContextRefreshListener.

## Deliberate behavior note

`ExternalAuthorizationCore.decide()` wraps the owner/group-association lookup in `try/catch(Exception)` and votes to deny on failure. This is intentional and fail-closed: an association-lookup error (e.g. a DB failure) is masked as access-denied rather than surfacing as a 500, matching the class's documented deny-on-error contract. This changes error visibility for the pre-existing declarative path, which previously let such errors propagate.


src/main/java/com/otilm/core/security/authz/AuthorizationRequest.java
```diff
@@ -0,0 +1,44 @@
...
+        List<SecuredUUID> parentUuids,
+        boolean hasSecurityFilter,
+        Optional<Class<ParentUUIDGetter>> parentUuidGetterClass,
+        String contextLabel) {
+
+    public AuthorizationRequest {
+        properties = properties == null ? Map.of() : Map.copyOf(properties);
+        objectUuids = objectUuids == null ? List.of() : List.copyOf(objectUuids);
+        parentUuids = parentUuids == null ? List.of() : List.copyOf(parentUuids);
+    }
```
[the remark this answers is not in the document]
ivosh: Fixed in edfed203.

src/test/java/com/otilm/core/signing/tsa/impl/TsaServiceImplUnitTest.java
```diff
@@ -227,10 +231,10 @@ void propagatesValidationException_fromValidator() throws Exception {
 
         @Test
         void propagatesAuthorizationDenial() throws Exception {
-            // given — the @ExternalAuthorization aspect denies access; the service propagates the denial unchanged so
+            // given — the AuthorizationEnforcer denies access; the service propagates the denial unchanged so
             // the controller can collapse it into the same generic not-found rejection (enumeration defense)
             doThrow(new AccessDeniedException("Access is denied"))
-                    .when(permissionEvaluator).tspProfileTimestamping(any());
+                    .when(authorizationEnforcer).enforce(eq(Resource.TSP_PROFILE), eq(ResourceAction.TIMESTAMP), any(SecuredUUID.class));
```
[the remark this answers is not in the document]
ivosh: Not fixing — false positive. Every production `enforce(...)` call passes exactly one `SecuredUUID` (see `AttributeCallbackScopeResolver`, `TsaServiceImpl`, `SigningRecordServiceImpl`), and `any(SecuredUUID.class)` matches a single vararg element. These tests pass in CI: a non-matching `doThrow(...).when(...)` stub would make the denial test fail (it asserts `AccessDeniedException`), and a non-matching `verify(...)` would fail outright — a silently-passing test isn't possible here.

src/test/java/com/otilm/core/signing/tsa/impl/TsaServiceImplUnitTest.java
```diff
@@ -430,10 +434,10 @@ void throwsSystemFailure_whenWorkflowIsNotManagedTimestamping() throws NotFoundE
...
```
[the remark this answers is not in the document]
ivosh: Not fixing — false positive. Every production `enforce(...)` call passes exactly one `SecuredUUID` (see `AttributeCallbackScopeResolver`, `TsaServiceImpl`, `SigningRecordServiceImpl`), and `any(SecuredUUID.class)` matches a single vararg element. These tests pass in CI: a non-matching `doThrow(...).when(...)` stub would make the denial test fail (it asserts `AccessDeniedException`), and a non-matching `verify(...)` would fail outright — a silently-passing test isn't possible here.

src/test/java/com/otilm/core/service/callback/AttributeCallbackScopeResolverTest.java
```diff
@@ -84,8 +86,8 @@ void certificateIssuanceAuthorizesAuthorityThenRaProfile() throws Exception {
 
         resolver.resolveScopeChain(Resource.CERTIFICATE, raProfileUuid, new HashSet<>());
 
-        Mockito.verify(permissionEvaluator).authorityInstance(any(SecuredUUID.class));
-        Mockito.verify(permissionEvaluator).raProfile(any(SecuredUUID.class));
+        Mockito.verify(authorizationEnforcer).enforce(eq(Resource.AUTHORITY), eq(ResourceAction.DETAIL), any(SecuredUUID.class));
+        Mockito.verify(authorizationEnforcer).enforce(eq(Resource.RA_PROFILE), eq(ResourceAction.DETAIL), any(SecuredUUID.class));
```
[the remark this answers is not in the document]
ivosh: Not fixing — false positive. Every production `enforce(...)` call passes exactly one `SecuredUUID` (see `AttributeCallbackScopeResolver`, `TsaServiceImpl`, `SigningRecordServiceImpl`), and `any(SecuredUUID.class)` matches a single vararg element. These tests pass in CI: a non-matching `doThrow(...).when(...)` stub would make the denial test fail (it asserts `AccessDeniedException`), and a non-matching `verify(...)` would fail outright — a silently-passing test isn't possible here.

src/test/java/com/otilm/core/service/callback/AttributeCallbackScopeResolverTest.java
```diff
@@ -102,8 +104,8 @@ void cryptographicKeyAuthorizesTokenInstanceThenTokenProfile() throws Exception
 
         resolver.resolveScopeChain(Resource.CRYPTOGRAPHIC_KEY, tokenProfileUuid, new HashSet<>());
 
-        Mockito.verify(permissionEvaluator).tokenInstance(any(SecuredUUID.class));
-        Mockito.verify(permissionEvaluator).tokenProfile(any(SecuredUUID.class));
+        Mockito.verify(authorizationEnforcer).enforce(eq(Resource.TOKEN), eq(ResourceAction.DETAIL), any(SecuredUUID.class));
+        Mockito.verify(authorizationEnforcer).enforce(eq(Resource.TOKEN_PROFILE), eq(ResourceAction.DETAIL), any(SecuredUUID.class));
```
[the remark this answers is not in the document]
ivosh: Not fixing — false positive. Every production `enforce(...)` call passes exactly one `SecuredUUID` (see `AttributeCallbackScopeResolver`, `TsaServiceImpl`, `SigningRecordServiceImpl`), and `any(SecuredUUID.class)` matches a single vararg element. These tests pass in CI: a non-matching `doThrow(...).when(...)` stub would make the denial test fail (it asserts `AccessDeniedException`), and a non-matching `verify(...)` would fail outright — a silently-passing test isn't possible here.

src/test/java/com/otilm/core/service/callback/AttributeCallbackScopeResolverTest.java
```diff
@@ -117,6 +119,6 @@ void locationAuthorizesEntity() throws Exception {
 
         resolver.resolveScopeChain(Resource.LOCATION, entityUuid, new HashSet<>());
 
-        Mockito.verify(permissionEvaluator).entityInstance(any(SecuredUUID.class));
+        Mockito.verify(authorizationEnforcer).enforce(eq(Resource.ENTITY), eq(ResourceAction.DETAIL), any(SecuredUUID.class));
     }
```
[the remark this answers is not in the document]
ivosh: Not fixing — false positive. Every production `enforce(...)` call passes exactly one `SecuredUUID` (see `AttributeCallbackScopeResolver`, `TsaServiceImpl`, `SigningRecordServiceImpl`), and `any(SecuredUUID.class)` matches a single vararg element. These tests pass in CI: a non-matching `doThrow(...).when(...)` stub would make the denial test fail (it asserts `AccessDeniedException`), and a non-matching `verify(...)` would fail outright — a silently-passing test isn't possible here.

src/main/java/com/otilm/core/security/authz/AuthorizationRequest.java
```diff
@@ -0,0 +1,47 @@
...
+        parentUuids = parentUuids == null ? List.of() : List.copyOf(parentUuids);
+        Objects.requireNonNull(parentUuidGetterClass,
+                "parentUuidGetterClass must not be null; pass Optional.empty()");
+    }
+
+    /**
+     * Builds the request for a direct {@code enforce(resource, action, uuids)} call.
+     */
+    public static AuthorizationRequest forDirectCheck(Resource resource, ResourceAction action, List<SecuredUUID> objectUuids) {
+        Map<String, String> properties = Map.of(
```
lubomirw, line 34: **[Low]** | Maintainability

`forDirectCheck` hardcodes the OPA property keys (`"action"/"name"/"parentAction"/"parentName"`) that `ExternalAuthorizationCore` reads through its own *private* constants (`NAME_PROP_NAME` etc.). The two string sets must agree, but nothing links them and the tests assert the literals rather than the constants — a rename in the core would silently desync the imperative path (malformed OPA request → fail-closed deny of legitimate calls) with no test catching it.

_Suggested:_ Extract the four keys to shared package-private constants and reference them from both `forDirectCheck` and `ExternalAuthorizationCore`.

src/main/java/com/otilm/core/security/authz/ExternalAuthorizationProgrammatic.java
```diff
@@ -0,0 +1,58 @@
...
+ *     ...
+ * }
+ * }</pre>
+ */
+@Documented
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ExternalAuthorizationProgrammatic {
+
+    /** The resource type the method's programmatic check targets. */
```
lubomirw, line 44: **[Low]** | Maintainability

The annotation drives the auth-service sync registration while enforcement is a separate hand-written `enforce(...)` call in the method body. Nothing verifies the two agree, so if the call is removed or its resource/action drifts from the declared pair, the action stays registered but is no longer enforced as declared — a silent authorization gap. The Javadoc relies on developer discipline, and the ITest only checks the registration side.

_Suggested:_ Add an ArchUnit/reflection test asserting every `@ExternalAuthorizationProgrammatic` method body references `AuthorizationEnforcer.enforce`, or a per-method test verifying `enforce` is invoked with the declared Resource/ResourceAction.

src/main/java/com/otilm/core/security/authz/ExternalAuthorizationCore.java
```diff
@@ -0,0 +1,245 @@
...
+        if (!(authentication instanceof PlatformAuthenticationToken || authentication instanceof AnonymousAuthenticationToken)) {
+            log.trace("Authentication is not of type 'PlatformAuthenticationToken' or 'AnonymousAuthenticationToken'. Cannot authorize.");
+            return new AuthorizationDecision(false);
+        }
+        if (authentication instanceof PlatformAuthenticationToken token) {
+            AuthorizationDecision result = check(token.getPrincipal().getRawData(), request);
+            if (!result.isGranted()) {
+                try {
+                    return checkGroupOwnerAssociations(token.getPrincipal(), request);
+                } catch (Exception e) {
```
lubomirw, line 72: **[Low]** | Error handling

`decide()` wraps `checkGroupOwnerAssociations` in `try/catch(Exception)` → logged `AuthorizationDecision(false)`. The old `ExternalMethodAuthorizationManager.checkInternal` had no try/catch here, so a DB failure in owner/group lookups previously propagated (≈500); now it becomes a 403-style deny. Fail-closed is the secure choice and matches the deny-on-error Javadoc — note it also changes error visibility for the pre-existing declarative path.

_Suggested:_ No code change required if intentional — confirm in the PR description that masking internal association-lookup errors as access-denied (rather than a 500) is deliberate.

src/main/java/com/otilm/core/security/authz/AuthorizationEnforcer.java
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Imperative front door to the platform authorization check.
+ *
+ * <p>Verifies that the currently authenticated principal may perform {@code action} on {@code resource} for the given
+ * object UUIDs, using the same OPA {@code method} policy and group/owner-association fallback as the declarative
+ * {@code @ExternalAuthorization} path. Throws {@link AccessDeniedException} on denial; returns normally on grant.
+ *
+ * <p><b>Runtime cost:</b> every call performs a blocking outbound HTTP request to OPA, and a denied direct check falls
+ * back to owner/group repository reads plus one additional OPA request per object UUID. Do not hold a database row
+ * lock or long transaction across a call (see CLAUDE.md, "Transactions and external calls").
```
lubomirw: **[Low]** | Maintainability

Shipped interface Javadoc points to an agent-guidance file: `(see CLAUDE.md, "Transactions and external calls")`. The section does exist in `core/CLAUDE.md`, so it resolves for repo developers, but a downstream consumer reading the compiled interface cannot follow it.

_Suggested:_ Drop the parenthetical, keep the runtime-cost guidance: "…Do not hold a database row lock or long transaction across a call."

lubomirw (commented):
Automated multi-reviewer triage — 4 findings (COMMENT). All Low severity; each is inline below. Reviewers: Copilot CLI, guided-pr-review, superpowers, hygiene.

src/main/java/com/otilm/core/security/authz/AuthorizationRequest.java
ivosh: Fixed in bcac3cb5. Extracted a package-private `AuthorizationProperties` holder for the four keys; `forDirectCheck`, `ExternalAuthorizationCore`, and `ObjectFilterAspect` (which had a third copy) now all source the literals from it.

src/main/java/com/otilm/core/security/authz/ExternalAuthorizationProgrammatic.java
ivosh: Fixed in bcac3cb5. Added an ArchUnit rule `programmatic_auth_methods_must_call_enforce` to `ExternalServiceAuthorizationArchTest`: every `@ExternalAuthorizationProgrammatic` method must call `AuthorizationEnforcer.enforce`, so dropping the call fails the build. It doesn't verify the call arguments match the declared Resource/ResourceAction pair — static-analysing arg values is brittle — so that half stays a review concern, noted in the rule Javadoc.

src/main/java/com/otilm/core/security/authz/AuthorizationEnforcer.java
ivosh: Fixed in bcac3cb5. Dropped the `(see CLAUDE.md, …)` parenthetical, kept the runtime-cost guidance.

src/main/java/com/otilm/core/security/authz/ExternalAuthorizationCore.java
ivosh: Intentional — no code change. Fail-closed on an association-lookup error is deliberate and matches the class's documented deny-on-error contract; masking the internal error as access-denied (rather than surfacing a 500) is the secure choice here. Adding a note to the PR description as requested.

lubomirw (approved):

PR #1776, opened 2026-07-10
Add regression test for TSP profile rename signing outage

### LINKED ISSUE #1654: TSP profile rename causes up-to-5-minute signing outage via stale signing-profile cache
### Description

  Renaming a TSP profile causes all signing profiles linked to it to reject TSP signing requests
  for up to 5 minutes. `SigningProfileModel` (cached in `SIGNING_PROFILE_CACHE`) bakes in the TSP
  profile name by value at load time (`SigningProfileMapper.toManagedTimestampingModel(SigningProfileMapper.java:161)`).
  When the TSP profile is renamed, `updateTspProfile` evicts only `TSP_PROFILE_CACHE` — it never
  evicts `SIGNING_PROFILE_CACHE`. The reverse eviction hook exists (signing-profile mutations call
  `tspProfileService.evictAllCachedModels()`) but there is no forward hook from TSP profile
  mutations back to the signing-profile cache.

### Steps to Reproduce

 1. Create a TSP profile named `tsa-a` and link it to one or more signing profiles.
  2. Issue several successful TSP signing requests via those signing profiles to prime `SIGNING_PROFILE_CACHE`.
  3. Rename `tsa-a` → `tsa-b` via `PUT /tsp/tspProfiles/{uuid}`.
  4. Within the next `SIGNING_PROFILES_CACHE_TTL_MINUTES` (default: 5 minutes), send a TSP signing
     request via any of the linked signing profiles.

### Expected Behavior

TSP signing requests succeed immediately after the rename.

### Actual Behavior

All TSP signing requests through affected signing profiles fail with `NotFoundException` — the
  cached `SigningProfileModel` still holds `tspProfileName = "tsa-a"`, and the downstream lookup
  `getTspProfile("tsa-a")` finds no profile. Requests fail for the remainder of the 5-minute cache
  TTL, despite the UUID linkage being fully intact. The system self-heals after TTL expiry with no
  operator action.

### Severity

Major

### Module

Signing

### Screenshot

_No response_

### Logs

_No response_

### Environment

2.18.1-SNAPSHOT

ivosh:
**This PR adds an end-to-end regression test for issue #1654; the bug itself was already fixed by #1594, so there is no production code change here.**

### What #1654 reported

Renaming a TSP profile made every linked signing profile reject TSP signing requests for up to 5 minutes.

The cached `SigningProfileModel` baked in the TSP profile **name** by value, so after a rename the downstream `getTspProfile("old-name")` lookup failed with `NotFoundException` until the 5-minute cache TTL expired.

### Why it no longer happens

`SigningProfileModel` now holds the TSP profile's **UUID**, not its name (`SigningProfileModel.tspProfileUuid`).

The signing path resolves the TSP profile by that UUID (`TsaServiceImpl` → `getTspProfile(signingProfile.tspProfileUuid())`).

A UUID never changes on rename, so the cached signing profile stays valid and the request succeeds immediately.

This was introduced by #1594 ("Introduce callable TSP endpoints"), which landed on 2026-06-16 — four days after #1654 was filed.

### What this PR does

It adds `TsaServiceImplITest.grantsTimestamp_afterLinkedTspProfileRenamed`, which reproduces the issue's exact steps end-to-end:

1. Create a signing profile linked to an enabled TSP profile.
2. Issue a granted timestamp to prime the caches.
3. Rename the linked TSP profile.
4. Issue another timestamp through the same signing profile and assert it still succeeds.

On the pre-#1594 code this test would fail with `NotFoundException`; on current code it is Granted.

A cache-level guard already existed (`TspProfileCacheTest.uuidLookupSurvivesRename`); this adds the missing end-to-end guard through the full signing flow.

### Verification

`mvn -Dmaven.compiler.proc=full -Dtest=TsaServiceImplITest test` → **Tests run: 9, Failures: 0, Errors: 0**.

Test-only change. No production code touched.

Closes #1654


src/test/java/com/otilm/core/integration/signing/tsa/TsaServiceImplITest.java
```diff
@@ -258,6 +262,25 @@ void grantsTimestamp_whenRequestValid() throws Exception {
...
+            SigningProfileDto profile = createTimestampingSigningProfile("rename-sp");
+            assertThat(tsaService.processTspRequestForSigningProfile(profile.getName(), aTspRequest().build()))
+                    .isInstanceOf(TspResponse.Granted.class);
+
+            // when — the linked TSP profile is renamed
+            var linkedTspProfile = tspProfileRepository.findByName("rename-sp-tsp").orElseThrow();
+            tspProfileService.updateTspProfile(
+                    SecuredUUID.fromUUID(linkedTspProfile.getUuid()),
+                    aTspProfileRequest().withName("rename-sp-tsp-renamed").build(),
+                    "http://localhost");
```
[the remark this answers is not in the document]
ivosh: Fixed in eece576c.

lubomirw (commented):
Automated multi-reviewer triage - 1 finding (COMMENT).

src/test/java/com/otilm/core/integration/signing/tsa/TsaServiceImplITest.java
```diff
@@ -258,6 +262,26 @@ void grantsTimestamp_whenRequestValid() throws Exception {
...
+            // when — the linked TSP profile is renamed
+            String linkedTspProfileName = profile.getName() + "-tsp";
+            var linkedTspProfile = tspProfileRepository.findByName(linkedTspProfileName).orElseThrow();
+            tspProfileService.updateTspProfile(
+                    SecuredUUID.fromUUID(linkedTspProfile.getUuid()),
+                    aTspProfileRequest().withName(linkedTspProfileName + "-renamed").build(),
+                    "http://localhost");
+
+            // then — a request through the same signing profile with renamed TSP profile still succeeds
+            TspResponse response = tsaService.processTspRequestForSigningProfile(profile.getName(), aTspRequest().build());
```
lubomirw: **[Low]** | Testing

The regression only asserts that signing still grants after the linked TSP profile is renamed. On the signing-profile path, `TsaServiceImpl` reloads the TSP profile by UUID and only checks `enabled()` before signing, so a stale UUID-cache entry with the old name but still enabled could still pass this assertion.

_Suggested:_ If this test is meant to prove cache coherence, mutate an observable TSP profile property after the rename (for example disable the profile) and assert the second request observes the new state. If the goal is only end-to-end "rename does not break signing," consider noting that lower-level `TspProfileCacheTest.uuidLookupSurvivesRename()` covers the cache-refresh guarantee.

<sub>flagged by Superpowers review; triaged as needs judgment</sub>

ivosh: Fixed in 6e49ff2f — but by removing this test, not strengthening it.

**The signing-profile path cannot observe a rename at all**, so this test could never prove cache coherence. It resolves the TSP profile by UUID and reads only `enabled()`; a rename changes neither the UUID nor `enabled()`, so a stale cache entry and a fresh one both grant. I confirmed it by mutation: with UUID cache eviction disabled in `TspProfileServiceImpl`, `grantsTimestamp_afterLinkedTspProfileRenamed` stayed green.

Your "disable and assert the new state is observed" suggestion is the right instinct, so I took it as a **new** test rather than an edit: `stopsGranting_afterLinkedTspProfileDisabled` warms the UUID-keyed cache with a real grant, disables the linked TSP profile, then asserts the next request is rejected (`BAD_REQUEST`, "TSP profile … is disabled"). Under the same mutation this one **fails** — so it genuinely guards hot-path cache coherence, which no existing test covered end-to-end through `TsaService` (prior coverage was either mocked gate logic or the name-key entry via a direct service call).

The rename-refresh guarantee itself is covered directly by `TspProfileCacheTest.uuidLookupSurvivesRename` (asserts the stale UUID entry is evicted and repopulated with the new name).

lubomirw (approved):

PR #1777, opened 2026-07-10
Fix RA profile create and edit for v3 authority connectors

### LINKED ISSUE #1779: Core - RA Profiles - Create/Edit: attribute operations bypass version-aware authority adapter, breaking v3 authority connectors
### Description

RA profile attribute operations in `RaProfileServiceImpl` bypass the version-aware authority adapter layer (`AuthorityProviderAdapterFactory` / `AuthorityProviderAdapter`) and call the v2 connector API client directly.

The private method `mergeAndValidateAttributes`, invoked from both **RA Profile create** and **RA Profile edit**, unconditionally uses `connectorApiFactory.getAuthorityInstanceApiClient(...)` to:

1. validate RA profile attributes via the v2 endpoint `POST /v1/authorityProvider/authorities/{uuid}/raProfile/attributes/validate`, and
2. list RA profile attribute definitions via the v2 endpoint `GET /v1/authorityProvider/authorities/{uuid}/raProfile/attributes`.

Authorities backed by a **v3 authority connector** do not serve these v2 endpoints. The v3 wire contract is stateless — attribute definitions come from `POST /v3/authorityProvider/authorities/raProfile/attributes` and there is no connector-side `/validate` endpoint at all (validation is performed by Core against the listed definitions).

As a result, creating or editing an RA profile under a v3 authority instance fails with a connector error, even though the same operations work correctly for v2 (and legacy framework-v1) authority connectors. This is a leftover from the v2/v3 unification behind capability-driven adapter dispatch: sibling flows (`AuthorityInstanceServiceImpl`, `ClientOperationServiceImpl`, `ExtendedAttributeServiceImpl`) already dispatch through `AuthorityProviderAdapterFactory.forAuthority(...)`, while `RaProfileServiceImpl` was not migrated.

### Steps to Reproduce

1. Register an authority connector implementing the **v3** authority provider interface.
2. Create an authority instance for this connector.
3. Navigate to **RA Profiles** and create a new RA profile under the v3 authority instance (equivalently, call `POST /v1/authorities/{authorityUuid}/raProfiles`).
4. Observe the request fail; the same failure occurs when editing an existing RA profile under a v3 authority.

### Expected Behavior

- RA profile create and edit resolve the correct adapter for the authority's connector interface version via `AuthorityProviderAdapterFactory.forAuthority(...)`.
- For v3 authorities, attribute definitions are fetched from the stateless v3 endpoint and no connector-side validate call is made (Core validates structurally against the listed definitions).
- For v2 authorities, behaviour is unchanged: connector-side validate followed by definition listing.

### Actual Behavior

- `RaProfileServiceImpl.mergeAndValidateAttributes` always calls the v2 API client endpoints, regardless of the connector interface version.
- For v3 authority connectors the calls hit non-existent v2 endpoints, so RA profile create and edit fail with a connector error.

### Severity

Major

### Module

None

### Screenshot

_No response_

### Logs

_No response_

### Environment

_No response_

klaraf755:
Fixes #1779


src/main/java/com/otilm/core/service/impl/RaProfileServiceImpl.java
```diff
@@ -681,15 +684,15 @@ private void mergeAndValidateAttributes(AuthorityInstanceReference authorityInst
...
         }
 
-        ApiClientConnectorInfo connectorDto = connectorService.getConnectorForApiClient(authorityInstanceRef.getConnectorUuid());
+        AuthorityProviderAdapter adapter = authorityProviderAdapterFactory.forAuthority(authorityInstanceRef);
 
         // validate first by connector
-        if (Boolean.FALSE.equals(connectorApiFactory.getAuthorityInstanceApiClient(connectorDto).validateRAProfileAttributes(connectorDto, authorityInstanceRef.getAuthorityInstanceUuid(), attributes))) {
+        if (Boolean.FALSE.equals(adapter.validateRAProfileAttributes(authorityInstanceRef, attributes))) {
             throw new ValidationException(ValidationError.create("RA profile attributes validation failed."));
         }
```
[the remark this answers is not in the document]
klaraf755: Change to check if it returns `Boolean.TRUE`.

src/main/java/com/otilm/core/service/impl/RaProfileServiceImpl.java
```diff
@@ -681,15 +684,15 @@ private void mergeAndValidateAttributes(AuthorityInstanceReference authorityInst
             throw new ValidationException(ValidationError.create("Connector of the Authority is not available / deleted"));
         }
 
-        ApiClientConnectorInfo connectorDto = connectorService.getConnectorForApiClient(authorityInstanceRef.getConnectorUuid());
+        AuthorityProviderAdapter adapter = authorityProviderAdapterFactory.forAuthority(authorityInstanceRef);
 
         // validate first by connector
-        if (Boolean.FALSE.equals(connectorApiFactory.getAuthorityInstanceApiClient(connectorDto).validateRAProfileAttributes(connectorDto, authorityInstanceRef.getAuthorityInstanceUuid(), attributes))) {
+        if (!Boolean.TRUE.equals(adapter.validateRAProfileAttributes(authorityInstanceRef, attributes))) {
```
lubomirw: **[Critical]** | Correctness

The connector-side `validateRAProfileAttributes` is a **`void`** endpoint (`interfaces` `connector/AuthorityInstanceController.java`; confirmed in `ejbca-ng-connector` and `ejbca-legacy-ca-connector`): success = HTTP 200 with an empty body, so the v2 client's `.toEntity(Boolean.class).getBody()` returns **`null`**. The old guard `Boolean.FALSE.equals(null)` passed; the new `!Boolean.TRUE.equals(null)` **throws** `"RA profile attributes validation failed."` — rejecting RA-profile create/edit on the happy path for every real v2 authority connector. (v3 returns `Boolean.TRUE`, which the old guard also accepted, so the rewrite was not needed for the v3 fix.)

_Suggested:_ preserve null-tolerant semantics — keep `Boolean.FALSE.equals(...)` (reject only an explicit `false`), or coalesce null→success in `AuthorityProviderV2Adapter` (`return result == null ? Boolean.TRUE : result;`) mirroring v3. Cleanest is the `void`+throw redesign in #3.

<sub>flagged by Copilot CLI + guided-pr-review + superpowers — verified against connector impls</sub>

src/test/java/com/otilm/core/integration/service/RaProfileServiceITest.java
```diff
@@ -196,6 +199,47 @@ void testAddRaProfile() throws ConnectorException, AlreadyExistException, Attrib
         Assertions.assertEquals(request.getName(), dto.getName());
     }
 
+    @Test
+    void testAddRaProfile_connectorRejectsAttributes() {
+        // v2 adapter path: the connector's RA-profile /validate returning false must surface as
+        // ValidationException and abort the create before anything is persisted.
+        mockServer.stubFor(WireMock
+                .post(WireMock.urlPathMatching("/v1/authorityProvider/authorities/[^/]+/raProfile/attributes/validate"))
+                .willReturn(WireMock.okJson("false")));
```
lubomirw: **[High]** | Testing

These stubs return `okJson("true")` / `okJson("false")` — a JSON boolean **no real connector emits**. The `/validate` wire contract is `void`: success is an empty 200 (client returns `null`), failure is an error status. Because the happy-path stubs in `testAddRaProfile` / `testEditRaProfile` return `"true"`, the null-on-success path is never exercised and the guard regression (#1) passes CI green.

_Suggested:_ stub `/validate` with an empty 200 (`WireMock.aResponse().withStatus(200)`) so the `null` return is reproduced, and assert the create succeeds; model rejection as an error status, not `okJson("false")`. Also add a `null`-return case to `AuthorityProviderV2AdapterTest` (it currently covers only `true`/`false`).

<sub>flagged by Copilot CLI + superpowers + guided-pr-review</sub>

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapter.java
```diff
@@ -65,6 +65,8 @@ public interface AuthorityProviderAdapter {
      */
     List<BaseAttribute> listRaProfileAttributes(AuthorityInstanceReference authority) throws ConnectorException;
 
+    Boolean validateRAProfileAttributes(AuthorityInstanceReference authority, List<RequestAttribute> attributes) throws ConnectorException;
```
lubomirw: **[Medium]** | Maintainability

`validateRAProfileAttributes` returns a tri-state `Boolean` while its siblings `validateIssueAttributes` / `validateRevokeAttributes` return `void` and throw `ValidationException`. The connector contract is `void`+throw (a literal `FALSE` never occurs — success is `null`, failure is an exception), so the `Boolean` (TRUE/FALSE/null) is the root cause of the null footgun in #1.

_Suggested:_ make this return `void` and throw `ValidationException` on failure (v2 throws on connector reject/null; v3 no-op), matching the sibling validate methods and the wire contract; drop the Boolean guard at the call site. This dissolves #1 and #4.

lubomirw: **[Low]** | Documentation

This method has no Javadoc; every other method in `AuthorityProviderAdapter` (including the two sibling validate methods) is documented. The return/null contract and the v2-vs-v3 divergence are left undocumented — the exact ambiguity behind #1.

_Suggested:_ add Javadoc stating the return semantics, the null/empty-body handling, and that v3 does no connector round-trip (returns `TRUE`).

lubomirw: **[Nit]** | Maintainability

Casing inconsistency within one interface: `listRaProfileAttributes` (`Ra`) vs the new `validateRAProfileAttributes` (`RA`). It does match the `interfaces` client method name, so it's defensible — your call.

_Suggested:_ pick one casing for the `RaProfile` token across the adapter interface, or keep it to mirror the client contract.

lubomirw (changes requested):
Automated multi-reviewer triage — 5 findings (REQUEST_CHANGES).

**Blocker (#1):** the guard rewrite `Boolean.FALSE.equals(...)` → `!Boolean.TRUE.equals(...)` rejects the empty-body `null` that v2 authority connectors return on *successful* validation, breaking RA-profile create/edit on the happy path for every real v2 connector. Verified against the `void` connector contract (`interfaces` + ejbca-ng/ejbca-legacy impls). Fixing #3 (return `void`+throw) dissolves both #1 and its documentation gap. See inline comments.

Sources: Copilot CLI, guided-pr-review, superpowers, hygiene (clean).

src/main/java/com/otilm/core/service/impl/RaProfileServiceImpl.java
klaraf755: Change to redesign mentioned in https://github.com/OmniTrustILM/core/pull/1777#discussion_r3571384492.

src/test/java/com/otilm/core/integration/service/RaProfileServiceITest.java
klaraf755: Change to stubs with response as in real connectors.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapter.java
klaraf755: Changed to return `void`.

klaraf755: Added javadoc similar to other validation methods.

klaraf755: Used `Ra`.

lubomirw (approved):

PR #1780, opened 2026-07-10
Project all values of multi-valued request attributes into repeated RDNs and SANs

### LINKED ISSUE #1771: Project all values of multi-valued request attributes into repeated RDNs and SANs
## Description

Dedicated issue for item 2 of #1688 (deferred from PR #1673, Mode A — platform-built PKCS#10).

`CertificateRequestAttributeProjector.extractStringValues` reads only `content.getFirst()`, and only when its data is a `String`. A multi-valued attribute (e.g. several SANs or repeated OUs under one mapping) keeps only its first entry, and an attribute whose first content item is non-string is dropped entirely. This matches the legacy `buildSubject` baseline (single value per RDN via `getSingleItemAttributeContentValue`), so it is **not a regression** — the limitation is documented in the projector's Javadoc as of PR #1673. But the Request Attributes design spec explicitly requires multi-value projection: §8 "ordered / multi-valued RDNs, e.g. two OUs"; §13 edge-case 4 "1→many (FQDN → CN + dNSName) and many→1 (repeated OU / multiple SANs), ordered via `MappedField.order`".

The fix surface is narrow. `project(...)` already sorts a definition's `MappedField`s by `order`, and `X509RequestContentRenderer.toX500Name` already iterates the full `RdnEntry` list — only the value extraction collapses to a single string. The extraction must emit **every** content value per attribute, and the projection loop must add one entry per (mapped field × content value): fields ordered by `MappedField.order`, values in content-list order within a field.

One behavior needs an explicit decision as part of this work: multi-valued content mapped to an **`EXTENSION`** field. RFC 5280 requires extension OIDs to be unique within `Extensions`, so repeated values for one extension OID cannot render to a valid structure (`ExtensionsGenerator.addExtension` throws on a duplicate OID). Rejecting this combination with a clear validation error at projection time is preferred over failing deep inside the renderer.

## User Story

As a certificate administrator, I want every value of a multi-valued request attribute (several SANs, repeated OUs) to be projected into the platform-built certificate request, so that the issued certificate carries the complete identity I configured rather than only the first value.

## Use Case

An RA profile maps a multi-value `dns_names` attribute to `dNSName` SAN entries and a two-value `organizational_unit` attribute to `OU` RDNs. An operator requests a certificate on the Mode A path (platform-generated CSR from a key pair). The generated PKCS#10 contains all configured DNS SANs and both `OU` RDNs, ordered by `MappedField.order` and then by content order.

## Acceptance Criteria

- [ ] `extractStringValues` (or its replacement) emits every content value of a multi-valued attribute; `project(...)` adds one RDN/SAN entry per (mapped field × content value), ordered by `MappedField.order` then content-list order (spec §13 edge-case 4, both 1→many and many→1).
- [ ] Non-string content items are skipped individually — a mixed-content list no longer drops the whole attribute; the behavior is documented on the method.
- [ ] Multi-valued content mapped to an `EXTENSION` field has an explicit, tested outcome (recommended: projection-time validation error naming the attribute and OID, since duplicate extension OIDs cannot render to valid RFC 5280 `Extensions`).
- [ ] Projector unit tests cover: repeated OUs (many→1), multiple DNS SANs, the existing 1→many case (single FQDN → CN + `dNSName`) staying green, mixed multi-value × multi-field ordering, and the non-string skip.
- [ ] The single-value limitation Javadoc in `CertificateRequestAttributeProjector` is removed.
- [ ] New-code coverage ≥ 80 % (line + branch, SonarCloud parity); quality gates per the Request Attributes design §17 Definition of Done.

ivosh:


src/main/java/com/otilm/core/attribute/CertificateRequestAttributeProjector.java
```diff
@@ -39,34 +39,53 @@ private CertificateRequestAttributeProjector() {}
...
+                    }
+                    case ExtensionMappedField ext -> {
+                        // An extension may appear only once in a request (RFC 5280).
+                        if (attributeValues.size() != 1) {
+                            throw new ValidationException(
+                                    "Attribute '%s' supplies %d values for certificate extension OID %s; an extension must map to exactly one value (RFC 5280)"
+                                            .formatted(def.getName(), attributeValues.size(), ext.getExtensionOid()));
+                        }
+                        extensions.add(toRequestedExtension(ext.getExtensionOid(), attributeValues.getFirst()));
+                    }
```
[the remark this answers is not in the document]
ivosh: Fixed in 0de57d3f. Duplicate extension OIDs (same OID mapped by more than one field/definition) are now rejected with a `ValidationException` before being added, so an invalid RFC 5280 Extensions sequence can no longer be produced.

On the message: the count is only reachable with a value ≥ 2 — `attributeValues.isEmpty()` is filtered out earlier (`continue`), so the branch that throws only fires when `size >= 2`. It can never render "1 values", so the plural wording is always correct.

src/main/java/com/otilm/core/attribute/CertificateRequestAttributeProjector.java
```diff
@@ -39,48 +39,92 @@ private CertificateRequestAttributeProjector() {}
...
+    }
+
+    private static void projectExtension(ExtensionMappedField ext, List<String> attributeValues, DataAttributeV3 def, ProjectionSink sink) {
+        // An extension may appear only once in a request (RFC 5280).
+        if (attributeValues.size() != 1) {
+            throw new ValidationException(
+                    "Attribute '%s' supplies %d values for certificate extension OID %s; an extension must map to exactly one value (RFC 5280)"
+                            .formatted(def.getName(), attributeValues.size(), ext.getExtensionOid()));
+        }
+        if (!sink.seenExtensionOids.add(ext.getExtensionOid())) {
```
lubomirw, line 146: **[Medium]** | Correctness

SAN mappings render as extension OID 2.5.29.17, but duplicate detection only tracks explicit EXTENSION mappings; a SAN mapping plus an explicit extension mapping to 2.5.29.17 can produce duplicate subjectAltName extensions.

_Suggested:_ Reserve/reject Extension.subjectAlternativeName.getId() when SAN entries are emitted, or reject explicit extension mappings to the SAN OID when SAN mappings are present; add a test for SAN plus extension 2.5.29.17.

src/main/java/com/otilm/core/attribute/CertificateRequestAttributeProjector.java
```diff
@@ -39,48 +39,92 @@ private CertificateRequestAttributeProjector() {}
...
+    private static List<MappedField> sortFields(FieldMapping fm) {
+        return fm.getFields().stream()
+                .sorted(Comparator.comparingInt(f -> f.getOrder() != null ? f.getOrder() : 0))
+                .toList();
+    }
+
+    private static void applyMappedField(MappedField field, List<String> attributeValues, DataAttributeV3 def, ProjectionSink sink) {
+        switch (field) {
+            case RdnMappedField rdn -> {
+                for (String value : attributeValues) {
```
lubomirw, line 119: **[Medium]** | Security

RDN/SAN projection now emits every string content item, while the pre-projector validation path does not appear to enforce properties.list or multiSelect cardinality for request content; a nominally single-value attribute can produce duplicate CN/RDN/SAN entries.

_Suggested:_ Decide the intended contract and either enforce one content item for non-list definitions before projection, or document that request attributes may be multi-content regardless of list; if enforcing, add coverage for non-list CN with two content items.

lubomirw (commented):
Automated multi-reviewer triage - 2 selected findings (COMMENT).

src/main/java/com/otilm/core/attribute/CertificateRequestAttributeProjector.java
ivosh: Fixed in 6f6fce01. Because SAN entries render into the subjectAltName extension (2.5.29.17), the projector now rejects mapping SAN entries together with an explicit extension mapping to that OID — both would produce a duplicate subjectAltName extension (RFC 5280). The check is a post-projection cross-check, so it catches the collision regardless of field order. Previously this surfaced late as an opaque BouncyCastle `ExtensionsGenerator.addExtension` `IllegalArgumentException` that the caller did not catch. Added a test (SAN + explicit extension to 2.5.29.17).

src/main/java/com/otilm/core/attribute/CertificateRequestAttributeProjector.java
ivosh: Fixed in 6f6fce01. The projector now enforces the contract: only definitions declared as a `list` may supply more than one value; a non-list attribute carrying multiple values is rejected with a `ValidationException`. This matches the platform's own invariant (`AttributeEngine.validateAttributeProperties` already requires `list` for `multiSelect`), and closes the gap where the pre-projector content validation did not check content cardinality against `list`. Definitions without properties are left unchecked (the flag is unknown). Added a test for a non-list CN with two content items.

lubomirw (approved):

PR #1789, opened 2026-07-11
Accept structured csrAttributes on the register path

### LINKED ISSUE #1788: Accept structured request attributes on the register path
### Description

Today the register operator DTO carries flat `subjectDn`/`subjectAltName`/`extensions`, and core projects that flat identity into the typed registration content on the connector wire. This adds an **optional structured request-attribute input (`csrAttributes`) to the register request**, mirroring the issue path.

When the operator supplies structured request attributes, core projects them into the typed registration identity using the existing request-attribute projector (reused from epic `ilm#131`) and validates the result, then carries it onto the register connector wire under the structured-request capability — exactly as the current register dual-wire does for flat input. The flat fields remain as the simple/compatibility shape.

No CSR or key is built at register time (registration is pre-CSR) — only the attribute-to-content projection is used. The shipped register dual-wire, projector, and validators are reused unchanged; the change is an additive request field plus an additive branch in the register resolution.

This makes register consistent with issue-new (both accept structured request attributes) and lets the admin UI use one shared identity editor for both forms (`fe-administrator#1693` consumes it).

Note: this also touches `interfaces` for the additive `csrAttributes` field on the register request DTO.

### User Story

As a certificate manager, I want to describe a pre-registration's identity with the same friendly, structured request attributes I use when issuing a certificate, so I do not have to hand-craft a flat distinguished name and the admin UI can offer one consistent identity editor for both issuing and pre-registration.

### Use Case

When pre-registering a certificate, an operator enters subject, SAN, and extension content through the structured request-attribute inputs (as on the issue form). The platform projects those into the typed registration identity, persists it on the placeholder, and sends it to structured-capable connectors — falling back to the flat shape for connectors and callers that use the simple form.

### Acceptance Criteria

- The register request accepts an optional structured request-attribute input (`csrAttributes`) alongside the existing flat `subjectDn`/`subjectAltName`/`extensions`; additive and backward-compatible.
- When structured request attributes are supplied, core projects them into the registration identity via the existing request-attribute projector and validates the result at the orchestrator boundary; when absent, the current flat projection is used unchanged.
- The projected identity is persisted on the placeholder (subject now; SAN with the SCEP/SAN work) and carried on the register connector wire under the structured-request capability, identical to the existing register dual-wire.
- Supplying both flat and structured identity is rejected or follows a defined precedence; challenge/authorization behavior is unchanged.
- The shipped register dual-wire, projector, and content validators are reused, not rewritten; existing flat-only register requests behave exactly as before.

lubomirw:
Closes #1788.

Accept structured `csrAttributes` on the certificate register path, mirroring the issue path; the flat `subjectDn`/`subjectAltName`/`extensions` stay as the simple shape.

- Orchestrator (`registerCertificate`) projects `csrAttributes` **once** (resolve definitions → `validateUpdateDataAttributes` → `CertificateRequestAttributeProjector.project`), derives the placeholder DN from that same content, rejects a request supplying both flat and structured identity, and threads the content to `register()`.
- `AuthorityProviderV3Adapter.register` uses the pre-built content when present, else builds flat identity itself. Signature gains a `@Nullable X509RequestContent` (mirrors `issueRegistered`).
- Shared `IssuanceDefinitionResolver` extracted from the issue path (single projection, one definition set).
- Fails closed when structured content cannot ride a non-structured connector's flat wire (non-DER extension / non-UTF8 `otherName`), instead of silently dropping.

Depends on interfaces#769 (`csrAttributes` on `ClientCertificateRegistrationDto`) — merged.

Reviewed branch-only via multi-reviewer pass before push; four findings applied. Targeted register/issue tests green (95/95); full verify runs in CI.

### DIFF
--- a/src/main/java/com/otilm/core/certificate/request/RegisterWireBuilder.java
+++ b/src/main/java/com/otilm/core/certificate/request/RegisterWireBuilder.java
@@ -239,13 +239,18 @@ public final class RegisterWireBuilder {
     }
 
     /**
-     * Fails closed when the request content carries an entry the flat register wire cannot represent — a non-DER
-     * extension value, or an otherName SAN with a non-UTF8 value encoding. Reached only for connectors without
-     * {@code CERTIFICATE_REQUEST_STRUCTURED}, where the flat renderers would otherwise drop such entries with only
-     * a warn log, reducing the registered identity below what the operator requested. Flat operator input never
-     * trips this — {@link #buildContent} forces DER extensions and UTF8 otherNames.
+     * Fails closed when projected content cannot be represented on the flat register wire.
+     *
+     * <p>Rejects a non-DER extension value, or an otherName SAN with a non-UTF8 encoding — throwing
+     * rather than dropping it.
+     *
+     * <p><b>Only reached</b> for connectors without {@code CERTIFICATE_REQUEST_STRUCTURED}, where the
+     * flat renderers would otherwise silently drop such entries (warn-log only) and register an identity
+     * weaker than the operator asked for.
+     *
+     * <p>Flat operator input never trips this — {@link #buildContent} forces DER extensions and UTF8 otherNames.
      */
-    private static void assertFlatRepresentable(X509RequestContent content) {
+    public static void assertFlatRepresentable(X509RequestContent content) {
         if (content.getExtensions() != null) {
             for (RequestedExtension ext : content.getExtensions()) {
                 if (ext.getEncoding() != null && ext.getEncoding() != ExtensionValueEncoding.DER) {
--- a/src/main/java/com/otilm/core/certificate/request/RegisterWireBuilder.java
+++ b/src/main/java/com/otilm/core/certificate/request/RegisterWireBuilder.java
@@ -87,6 +87,9 @@ public final class RegisterWireBuilder {
         if (connectorSupportsStructured) {
             dto.setRequestContent(content);
         } else {
+            // No structured wire carries the content here, so anything the flat wire cannot represent would be
+            // silently dropped — fail closed rather than register an identity reduced below the operator's request.
+            assertFlatRepresentable(content);
             dto.setExtensions(renderFlatExtensions(content.getExtensions()));
         }
         return dto;
@@ -233,6 +236,35 @@ public final class RegisterWireBuilder {
             return null;
         }
         return "otherName:" + entry.getOtherNameOid() + ";UTF8:" + entry.getValue();
+    }
+
+    /**
+     * Fails closed when the request content carries an entry the flat register wire cannot represent — a non-DER
+     * extension value, or an otherName SAN with a non-UTF8 value encoding. Reached only for connectors without
+     * {@code CERTIFICATE_REQUEST_STRUCTURED}, where the flat renderers would otherwise drop such entries with only
+     * a warn log, reducing the registered identity below what the operator requested. Flat operator input never
+     * trips this — {@link #buildContent} forces DER extensions and UTF8 otherNames.
+     */
+    private static void assertFlatRepresentable(X509RequestContent content) {
+        if (content.getExtensions() != null) {
+            for (RequestedExtension ext : content.getExtensions()) {
+                if (ext.getEncoding() != null && ext.getEncoding() != ExtensionValueEncoding.DER) {
+                    throw new ValidationException(("Extension %s has %s value encoding, which the flat register wire "
+                            + "cannot represent; the authority must advertise CERTIFICATE_REQUEST_STRUCTURED")
+                            .formatted(ext.getOid(), ext.getEncoding()));
+                }
+            }
+        }
+        if (content.getSubjectAltNames() != null) {
+            for (GeneralNameEntry san : content.getSubjectAltNames()) {
+                if (san.getType() == GeneralNameType.OTHER_NAME
+                        && san.getValueEncoding() != null && san.getValueEncoding() != ExtensionValueEncoding.UTF8_STRING) {
+                    throw new ValidationException(("otherName SAN %s has %s value encoding, which the flat register "
+                            + "wire cannot represent; the authority must advertise CERTIFICATE_REQUEST_STRUCTURED")
+                            .formatted(san.getOtherNameOid(), san.getValueEncoding()));
+                }
+            }
+        }
     }
 
     private static List<CertificateExtension> renderFlatExtensions(List<RequestedExtension> extensions) {
--- a/src/main/java/com/otilm/core/certificate/request/RegisterWireBuilder.java
+++ b/src/main/java/com/otilm/core/certificate/request/RegisterWireBuilder.java
@@ -182,7 +182,11 @@ public final class RegisterWireBuilder {
 
     // ── flat-wire rendering ─────────────────────────────────────────────────
 
-    private static String renderSubjectDn(X509RequestContent content) {
+    /**
+     * Renders the RFC 4514 subject DN string for the given content, or null when it carries no subject.
+     * Shared with the register orchestrator so the persisted placeholder DN matches the wire anchor exactly.
+     */
+    public static String renderSubjectDn(X509RequestContent content) {
         if (content.getSubject() == null || content.getSubject().isEmpty()) {
             return null;
         }
--- a/src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
@@ -515,7 +515,16 @@ public class ClientOperationServiceImpl implements ClientOperationExternalServic
         } catch (AttributeException e) {
             throw new ValidationException("Invalid csrAttributes for certificate registration: " + e.getMessage());
         }
-        return CertificateRequestAttributeProjector.project(definitions, csrAttributes);
+        X509RequestContent content = CertificateRequestAttributeProjector.project(definitions, csrAttributes);
+        boolean hasSubject = content.getSubject() != null && !content.getSubject().isEmpty();
+        boolean hasSan = content.getSubjectAltNames() != null && !content.getSubjectAltNames().isEmpty();
+        if (!hasSubject && !hasSan) {
+            // Optional/unmapped/extension-only csrAttributes can project to no subject and no SAN; reject
+            // locally rather than create a placeholder and register an identity-less certificate.
+            throw new ValidationException(
+                    "csrAttributes did not yield a subject or subjectAltName for the pre-registration identity");
+        }
+        return content;
     }
 
     private static boolean hasStructuredIdentity(List<RequestAttribute> csrAttributes) {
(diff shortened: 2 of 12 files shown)


src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
```diff
@@ -556,6 +572,65 @@ void registerPersistsConnectorMetadata() throws Exception {
...
+
+    @Test
+    void registerRejectsStructuredCsrAttributesThatProjectNoIdentity() throws Exception {
+        // Unmapped/extension-only csrAttributes project to no subject and no SAN — reject before placeholder creation.
+        when(issuanceDefinitionResolver.resolve(Mockito.any())).thenReturn(List.of());
+        ClientCertificateRegistrationDto request = new ClientCertificateRegistrationDto();
+        request.setCsrAttributes(List.of(new RequestAttributeV3(UUID.randomUUID(), "unmapped",
+                AttributeContentType.STRING, List.<BaseAttributeContentV3<?>>of(new StringAttributeContentV3("x")))));
+
+        Assertions.assertThrows(ValidationException.class, () -> clientOperationService.registerCertificate(
```
ivosh: **! Major** | Testing

The test at `ClientOperationServiceRegisterITest.java:620-632` intends to cover `buildStructuredRegistrationContent`'s empty-projection rejection (`ClientOperationServiceImpl.java:519-527`), but the `ValidationException` it asserts is actually thrown much earlier by the capability gate.

This test never stubs `registeringAdapter()`/`adapterFactory` (unlike the sibling happy-path `registerProjectsStructuredCsrAttributesAndForwardsContent`, which stubs it at line 604), and `setUpRegistrationFixtures()` only stubs `capabilityService`. So `adapterFactory.forAuthority(authority)` returns null, and the gate at `ClientOperationServiceImpl.java:424-427` (`adapter instanceof RegisterCapability` is false) throws `ValidationException("Certificate registration is not supported...")` — which runs BEFORE `buildStructuredRegistrationContent` is called at line 431.

Consequences: (a) the empty-projection branch (`ClientOperationServiceImpl.java:519-527`) is never executed — JaCoCo shows those lines uncovered despite a green test; (b) the `when(issuanceDefinitionResolver.resolve(...)).thenReturn(List.of())` stub is dead — it is never invoked; (c) `assertEquals(0, certificateRepository.count())` passes trivially because the gate short-circuits before placeholder creation, not because the empty-projection path declines to persist. This is exactly the CLAUDE.md 'tests that look green but cover nothing' / 'assertion checks the type of an exception without checking the origin' failure.

Fix: stub `registeringAdapter()` (as the happy-path test does) so the flow passes the gate and reaches the empty-projection check, and/or assert on the specific message so a gate-origin exception no longer satisfies the test.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -423,7 +426,13 @@ public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID au
...
         }
 
-        Certificate placeholder = certificateService.createRegistrationPlaceholder(raProfile, request);
+        // Project structured csrAttributes once (validated) so the placeholder DN and the connector wire derive
+        // from the same content; a flat request yields null here and the adapter builds its identity itself.
+        X509RequestContent registrationContent = buildStructuredRegistrationContent(raProfile, request);
+        String effectiveSubjectDn = registrationContent != null
+                ? RegisterWireBuilder.renderSubjectDn(registrationContent)
+                : request.getSubjectDn();
+        Certificate placeholder = certificateService.createRegistrationPlaceholder(raProfile, effectiveSubjectDn);
```
ivosh, line 435: **~ Minor** | Error handling

**The placeholder state is safe (FAILED, not orphaned) — but this local, deterministic validation fires one step too late, leaving a committed FAILED certificate where every other pre-registration validation leaves no row.**

Trace: for a structured `csrAttributes` request, `buildStructuredRegistrationContent` (ClientOperationServiceImpl.java:431) projects the content, then `createRegistrationPlaceholder` (line 435) commits a REQUESTED cert, then `register()` (line 445) calls `RegisterWireBuilder.buildRegistration(content, structured=false)`, which invokes `assertFlatRepresentable` (RegisterWireBuilder.java:92/248) and throws `ValidationException` when the projected content carries a non-DER extension or a non-UTF8 otherName SAN and the connector does not advertise `CERTIFICATE_REQUEST_STRUCTURED`.

I confirmed the divergence-safety part is correct: `ValidationException extends RuntimeException`, so it is caught by the pre-acceptance `catch (ConnectorException | RuntimeException e)` at line 455 and the placeholder is transitioned to FAILED (not orphaned in PENDING_REGISTRATION). No external commitment happened, so there is no state divergence.

The observation is a consistency asymmetry: the two sibling local validations on this path — `rejectAmbiguousRegistrationIdentity` (line 408) and the empty-projection check in `buildStructuredRegistrationContent` (line 521) — both run BEFORE `createRegistrationPlaceholder`, and their tests assert `certificateRepository.count() == 0`. The representability check depends only on the already-projected content plus the connector's `CERTIFICATE_REQUEST_STRUCTURED` capability (both knowable before line 435), yet it rejects only after a placeholder is committed, persisting a FAILED cert for a request that never reached the connector.

Two readings: (a) intended — a FAILED audit row for any post-placeholder rejection is acceptable and uniform with connector-rejection failures; leave as-is. (b) worth aligning — hoist the representability check (e.g. resolve the structured capability and run `assertFlatRepresentable` inside `buildStructuredRegistrationContent`, before the placeholder) so a purely-local rejection leaves no row, matching the other two validations.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -475,6 +484,63 @@ public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID au
...
+        if (!hasStructuredIdentity(request.getCsrAttributes())) {
+            return null;
+        }
+        // The gate tolerates null elements (anyMatch); drop them before validation/projection so a
+        // partially-null list projects the real attributes instead of NPE-ing in the projector.
+        List<RequestAttribute> csrAttributes = request.getCsrAttributes().stream().filter(Objects::nonNull).toList();
+        List<DataAttributeV3> definitions = issuanceDefinitionResolver.resolve(raProfile);
+        try {
+            attributeEngine.validateUpdateDataAttributes(null, null, definitions, csrAttributes);
+        } catch (AttributeException e) {
```
ivosh, line 515: **~ Minor** | Testing

The catch that maps `AttributeException` from `attributeEngine.validateUpdateDataAttributes(...)` to a client-facing `ValidationException` (`ClientOperationServiceImpl.java:513-517`) is untested. The happy-path structured test (`registerProjectsStructuredCsrAttributesAndForwardsContent`) mocks `attributeEngine` as a no-op so `validateUpdateDataAttributes` never throws, and `registerRejectsStructuredCsrAttributesThatProjectNoIdentity` never reaches this method at all (see the major finding). So neither the validation-failure branch nor the empty-projection branch of the new structured path has real coverage. Suggest a test that stubs `attributeEngine.validateUpdateDataAttributes(...)` to throw `AttributeException`, drives `registerCertificate` with a structured csrAttributes request through a register-capable adapter, and asserts the wrapped `ValidationException` ("Invalid csrAttributes for certificate registration: ...").

src/main/java/com/otilm/core/certificate/request/RegisterWireBuilder.java
```diff
@@ -231,6 +238,35 @@ private static String renderOtherName(GeneralNameEntry entry) {
         return "otherName:" + entry.getOtherNameOid() + ";UTF8:" + entry.getValue();
     }
 
+    /**
+     * Fails closed when the request content carries an entry the flat register wire cannot represent — a non-DER
```
ivosh: **. Nitpick** | Documentation

`RegisterWireBuilder.java:241-247` — the `assertFlatRepresentable` Javadoc is the one comment in this PR that packs too much into a single unbroken paragraph. It fuses four concerns — what it does, the two shapes it rejects, when it's reached plus the silent-drop behaviour it replaces, and the boundary that flat input never trips it — with no `<p>` breaks or bold lead-ins.

It stays followable (the load-bearing "fails closed" fact is up front and no sentence exceeds ~30 words), so this is take-it-or-leave-it, not a defect. Every other comment the PR adds is a clear one- or two-sentence WHY comment.

Optional restructure — nothing dropped, just unpacked into one concern per `<p>` with a bold lead-in on the load-bearing condition:

```java
/**
 * Fails closed when projected content cannot be represented on the flat register wire.
 *
 * <p>Rejects a non-DER extension value, or an otherName SAN with a non-UTF8 encoding — throwing
 * rather than dropping it.
 *
 * <p><b>Only reached</b> for connectors without {@code CERTIFICATE_REQUEST_STRUCTURED}, where the
 * flat renderers would otherwise silently drop such entries (warn-log only) and register an identity
 * weaker than the operator asked for.
 *
 * <p>Flat operator input never trips this — {@link #buildContent} forces DER extensions and UTF8 otherNames.
 */
```

ivosh (dismissed):

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
lubomirw: Fixed in 4f70de970 — the test now calls `registeringAdapter()` so the flow passes the capability gate and actually reaches the empty-projection check, and asserts the specific message (`did not yield a subject or subjectAltName`) so a gate-origin exception no longer satisfies it. The dead `resolve(...)` stub is now live.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Adopted reading (b) in 4f70de970 — hoisted the representability check into `buildStructuredRegistrationContent` (before the placeholder), so a purely-local rejection leaves no row, uniform with the other two pre-placeholder validations. `assertFlatRepresentable` is now public; the adapter still re-checks when it builds the flat wire, so this is a consistency guard, not the sole enforcement.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Added `registerWrapsCsrAttributeValidationFailure` in 4f70de970 — stubs `attributeEngine.validateUpdateDataAttributes(...)` to throw `AttributeException` through a register-capable adapter and asserts the wrapped `ValidationException` (`Invalid csrAttributes for certificate registration...`) plus zero placeholders.

src/main/java/com/otilm/core/certificate/request/RegisterWireBuilder.java
lubomirw: Applied your restructure verbatim in 4f70de970 — one concern per `<p>` with the load-bearing "fails closed" fact up front.

3keyroman (approved):

Issue #1571, opened 2026-06-09
Add Certificate Registered event


lubomirw:
### Description

Add a **Certificate Registered** event to the platform's event/notification framework, fired when a
certificate is successfully pre-registered (part of `ilm#130`). It fires **exactly once per registration**,
after the transaction commits, when the certificate reaches `REGISTERED` — on both the synchronous
registration path and the asynchronous poll-completion path — and again when a renew/rekey copies a
registration to its successor. There is no resend or reset in v1.

By default it notifies the certificate **owner** (recipient configurable via the notification framework).
The credential the user presents on the public portal — the challenge stored by the pre-registration work
(`core#1570`) — is delivered **only** to the external notification-provider dispatch; it is **excluded**
from internal notifications (which persist to the database), from event-history, and from logs.

### User Story

As a certificate manager, I want a notification sent automatically when a certificate is pre-registered —
carrying the credentials, instructions, and the issuance deadline — so the user knows how and by when to
complete issuance, without me sending anything manually.

### Use Case

When a certificate is pre-registered for a user who has no platform account, the platform itself notifies
that user with the credentials and instructions to complete issuance on the public portal — replacing the
arrangement where the upstream CA generated and emailed the password.

### Acceptance Criteria

- A "Certificate Registered" event type exists and fires on successful pre-registration, exactly once per
  registration, after commit.
- It fires on both the synchronous and asynchronous registration-completion paths, and on renew/rekey copy
  creation; nothing re-fires it (no resend/reset in v1).
- Default recipient is the certificate owner; the recipient is configurable via the notification framework.
- The external-provider notification payload includes the credential to complete issuance on the portal,
  the instructions, and the issuance-window deadline.
- The credential is excluded from internal notifications, event-history, and logs.
- No event fires for ordinary (non-pre-registered) issuance.
- Credential delivery requires an external notification provider configured on the event; otherwise the
  operator who set the challenge relays it manually (documented for operators).

PR #1791, opened 2026-07-13
Verify pre-registration challenge on issue; fail-closed renew/rekey

### LINKED ISSUE #1570: Pre-register certificate with challenge verified on issue/rekey

lubomirw:
Part of #1570. Adds the platform-owned challenge verification layer for pre-registered certificates.

## What it does

- **Register opt-in** — when the request carries an `authorizationSecret`, a durable ACTIVE `CertificateRegistrationAuthorization` (encrypted challenge, issuance window) is created **before** the connector call. Fail-closed: a connector-accepted-but-local failure leaves the cert reconcilable to `REGISTERED`, so a row written only on the success path would let the issue gate treat a protected registration as unprotected. A past `expiresAt` is rejected at the top of `registerCertificate`, before any placeholder exists.
- **Issue gate** — verifies the issuance window then the operator challenge (constant-time) under a per-row pessimistic lock. The failed-attempt increment + lockout are **committed before** the 422 is thrown, so the counter survives (otherwise a same-transaction rollback would erase it → unlimited retries). Window expiry flips the row to `EXPIRED`; `LOCKED`/`EXPIRED` deny; `CLOSED`/absent pass through unchanged.
- **Renew + rekey** of a certificate with an ACTIVE authorization are **fail-closed denied** (one shared guard). Also covers the CMP-`kur` and location-renewal funnels.
- **Secret hygiene** — `TracingAspect` now masks `@Sensitive` parameters and `RegistrationChallengeStore.store`/`verify` carry it (closes a latent span-attribute leak; tracing is off by default). The rekey `authorizationSecret` field was removed (interfaces#771) since rekey verification is deferred, so there is no rekey secret to strip.

## Scope

Does close #1570. Remaining acceptance criteria land separately: the `CERTIFICATE_REGISTERED` event (#1571, stacked next), the inventory detail block, no-CSR completion, and rekey challenge verification (deferred — verifying rekey without copying the authorization to the successor would leave the rekeyed successor unprotected, so the two land together; rekey is fail-closed-denied here in the meantime). The issuance-window default (7d) and max failed attempts (5) are constants and become operator-configurable settings in a follow-up.

## Tests

`ClientOperationServiceRegisterITest` 49/49 (+10): register-with-secret creates an encrypted ACTIVE authorization; explicit + default window; past window → 422 with no cert; correct challenge issues; wrong challenge → 422 with the increment surviving a fresh re-read; repeated wrong → LOCKED; expired window → 422 + EXPIRED; renew/rekey of an ACTIVE-authorization cert → 422. New `TracingAspectTest` proves a `@Sensitive` parameter records as `***`. The rekey secret-strip line is not ITest-covered — it needs a full non-registered rekey round-trip this suite defers to the v3 integration tests.


src/main/java/com/otilm/core/aop/TracingAspect.java
```diff
@@ -54,10 +56,16 @@ public Object traceMethod(ProceedingJoinPoint joinPoint) throws Throwable {
...
             for (int i = 0; i < parameterNames.length; i++) {
                 String paramName = parameterNames[i];
+                // A @Sensitive parameter (e.g. a raw secret) must never be recorded verbatim into a span.
+                if (i < parameters.length && parameters[i].isAnnotationPresent(Sensitive.class)) {
+                    span.setAttribute("function.param." + paramName, "***");
+                    continue;
+                }
                 Object paramValue = parameterValues[i];
                 span.setAttribute("function.param."+paramName, paramValue != null ? paramValue.toString() : "null");
             }
```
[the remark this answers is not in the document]
lubomirw: Fixed in 10fe134f7 — the loop now skips when `getParameterNames()` is null and iterates only `min(names, values, parameters)`, so it can't NPE or index out of bounds.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -535,6 +559,131 @@ private X509RequestContent buildStructuredRegistrationContent(RaProfile raProfil
         return content;
     }
 
+    /**
+     * Creates the durable challenge authorization for a self-service pre-registration when the operator supplied an
+     * {@code authorizationSecret}. Created before the connector call, not after: a connector-accepted-but-local
+     * failure leaves the certificate reconcilable to REGISTERED, and a row written only on the success path would be
+     * missing there — the later issue gate would then treat a secret-protected registration as unprotected and issue
+     * with no challenge. Absent secret means no row, so the operator register→issue flow is unchanged.
+     */
```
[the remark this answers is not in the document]
lubomirw: Already resolved on the current head — the past-window validation was hoisted into its own `rejectPastRegistrationWindow` method (4620f25d8), so there is no longer a dangling block: the Javadoc at this location is attached to `maybeCreateRegistrationAuthorization`.

src/main/java/com/otilm/core/aop/TracingAspect.java
```diff
@@ -54,10 +56,16 @@ public Object traceMethod(ProceedingJoinPoint joinPoint) throws Throwable {
 
             String[] parameterNames = signature.getParameterNames();
             Object[] parameterValues = joinPoint.getArgs();
+            Parameter[] parameters = method.getParameters();
 
             // Add method parameters as span attributes
```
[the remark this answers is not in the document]
lubomirw: The null/length guard was added in 10fe134f7 — the loop is wrapped in `if (parameterNames != null)` and bounded by `Math.min(names, values, parameters)`, so it cannot NPE or index out of bounds. When names are unavailable it simply records no parameter attributes (tracing still runs); recovering names from arg indices isn't worth the added complexity for that edge.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1345,6 +1504,10 @@ public void renewCertificateAction(final UUID certificateUuid, ClientCertificate
     public ClientCertificateDataResponseDto rekeyCertificate(SecuredParentUUID authorityUuid, SecuredUUID raProfileUuid, String certificateUuid, ClientCertificateRekeyRequestDto request) throws NotFoundException, CertificateException, CertificateOperationException, CertificateRequestException {
         Certificate oldCertificate = validateOldCertificateForOperation(certificateUuid, raProfileUuid.toString(), ResourceAction.REKEY);
 
+        // Fail-closed: rekey is not challenge-gated yet (its verification and successor copy come later), so
+        // refuse to rekey a certificate whose registration is still ACTIVE.
+        rejectRenewOrRekeyOfActiveRegistration(oldCertificate.getUuid());
```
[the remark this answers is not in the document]
lubomirw: Correct — the strip was removed because the rekey `authorizationSecret` field itself was removed (interfaces#771; rekey verification is deferred). Updated the PR description to match — there is no rekey secret to strip.

src/test/java/com/otilm/core/aop/TracingAspectTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * The aspect records every traced method parameter's value into a span attribute, so a parameter carrying a
+ * secret must be masked. This locks in that a {@link Sensitive} parameter is recorded as {@code ***} while an
+ * ordinary parameter keeps its value.
+ */
+class TracingAspectTest {
+
+    @SuppressWarnings("unused") // reflected as the traced-method fixture
+    private void sampleMethod(@Sensitive String secret, String visible) {
```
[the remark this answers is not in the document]
lubomirw: Restructured in ab35d4d81 — the fixture method now returns `visible + secret`, so both parameters are used. It is a reflection-only fixture (the aspect reads its parameter names/annotations; the method is never invoked), so the body is inconsequential to the test.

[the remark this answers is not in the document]
lubomirw: Restructured in ab35d4d81 — the fixture method now returns `visible + secret`, so both parameters are used. It is a reflection-only fixture (the aspect reads its parameter names/annotations; the method is never invoked), so the body is inconsequential to the test.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -442,6 +461,10 @@ public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID au
 
         AdapterOperationResult result;
         try {
+            // Persist the challenge authorization before the connector call — on every arc that can leave the
+            // certificate reconcilable to REGISTERED (see the method) — but inside the try, so a store/save
+            // failure fails the placeholder via the catch below rather than orphaning it in PENDING_REGISTRATION.
+            maybeCreateRegistrationAuthorization(certificate, request);
```
klaraf755, line 474: This catch fails the placeholder but leaves the authorization row behind: `maybeCreateRegistrationAuthorization` commits in its own tx (method is `NOT_SUPPORTED`), so a connector rejection leaves an ACTIVE authorization pointing at a FAILED cert. It's inert but retains an encrypted secret for a dead registration. `deleteByCertificateUuid` already exists on the repository — worth calling it here. This arc is also untested: `registrationStoreFailureFailsPlaceholderInsteadOfOrphaning` covers failure before the save, not rejection after it.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1144,6 +1292,13 @@ public ClientCertificateDataResponseDto issueExistingCertificate(final SecuredPa
         if (!registered && hasCsr) {
             throw new ValidationException(ValidationError.create("This certificate already has a signing request and cannot accept another. Certificate: %s".formatted(certificate.toStringShort())));
         }
+
+        // Self-service gate: verify the operator challenge before the CSR attach and the async enqueue, so a bad
+        // challenge rejects the caller synchronously and the secret never rides the ActionMessage. No-op when the
+        // certificate carries no registration authorization.
+        verifyRegistrationChallenge(certificate.getUuid(),
```
klaraf755, line 1325: The window is enforced only at this synchronous gate. If the RA profile requires approval, `issueCertificateAction` can complete issuance after `expiresAt` with no re-check — but the interfaces-repo DTO documents `expiresAt` as the deadline "by which the pre-registered certificate must be issued." Which semantics are intended? If the deadline binds issuance, the re-check belongs in the action; if it only gates verification, the DTO description needs softening (interfaces PR).

klaraf755 (commented):
Reviewed with focus on correctness, transaction boundaries, and secret handling. The core mechanics check out: the failed-attempt increment genuinely commits before the 422 (all entry points are `NOT_SUPPORTED`, so the explicit tx is independent), the pessimistic lock covers only local writes, all ISSUE/RENEW/REKEY enqueue points sit behind the gate/guard (including CMP `kur` and location renewal), and the secret-hygiene layering (constant-time compare, `@ToString.Exclude`, write-only DTO field, span masking) is real.

One change requested (inline: orphaned authorization row on connector rejection), plus three discussion points below.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -535,6 +558,131 @@ private X509RequestContent buildStructuredRegistrationContent(RaProfile raProfil
...
+    }
+
+    /**
+     * Fail-closed guard: a certificate with an ACTIVE registration authorization must not be renewed or rekeyed
+     * until those paths are challenge-gated (together with copying the authorization to the successor), so neither
+     * can silently exit the challenge regime. This also denies platform-automation renewals (locations, workflows,
+     * scheduled renew) of such certificates. Issue completion remains the challenge-verified path
+     * (verifyRegistrationChallenge).
+     */
+    private void rejectRenewOrRekeyOfActiveRegistration(UUID certificateUuid) {
```
klaraf755, line 703: Nothing sets `CLOSED` yet, and once a cert is ISSUED the only code that transitions the authorization (the issue gate) is unreachable for it — so every challenge-issued cert stays un-renewable/un-rekeyable, including via scheduled and location renewals, until the follow-up lands. Matches the PR description, just confirming this interim operational surface is accepted.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -535,6 +558,131 @@ private X509RequestContent buildStructuredRegistrationContent(RaProfile raProfil
...
+        }
+    }
+
+    private String evaluateLockedAuthorization(CertificateRegistrationAuthorization authorization, String presentedSecret) {
+        UUID certificateUuid = authorization.getCertificateUuid();
+        RegistrationState state = authorization.getState();
+        if (state == RegistrationState.CLOSED) {
+            return null;
+        }
+        if (state == RegistrationState.LOCKED) {
```
klaraf755, line 660: Wrong-challenge and expiry denials write event history; attempts against an already-LOCKED authorization don't. That's the scenario (persistent hammering) where the audit trail matters most — consider recording it.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Good catch — fixed in c981dc3a7. The pre-acceptance failure arc now removes the authorization (`deleteByCertificateUuid`, best-effort in its own transaction since the method is `NOT_SUPPORTED`), so a connector rejection no longer leaves an encrypted secret on a dead certificate. Added `connectorRejectionAfterAuthorizationSavedRemovesTheAuthorization` covering the reject-after-save arc (the store-failure test only covered failure before the save).

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Decision: the window gates the **completion request**, not final issuance — a valid challenge must be presented before `expiresAt`; approval/async processing may finalize issuance shortly after, and the operator's approval latency shouldn't retroactively void a timely submission. Rather than add an issuance-time re-check, I've softened the interfaces `expiresAt` description to state this explicitly (ships with the certificate-registered event branch). So no action-path change here.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Confirmed — accepted interim. Until the follow-up lands (operator close-registration action + challenge-gated renew/rekey with copy-to-successor), an issued cert that carries an active registration is intentionally not renewable/rekeyable (fail-closed), including scheduled and location renewals. Called out in the PR description.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Fixed in c981dc3a7 — an attempt against an already-LOCKED authorization now writes event history too, so persistent hammering is audited (previously only wrong-challenge and expiry denials were recorded).

klaraf755 (approved):

PR #1806, opened 2026-07-14
Fix discovery stuck in Processing status

### LINKED ISSUE #1800: Discovery - Network Discovery Provider: discovery stuck in "Processing" status, unable to finish, duration keeps increasing indefinitely
### Description

A discovery finishes on the connector side — **Discovery Provider Status** shows `Completed` and the discovered certificates appear on the discovery's *Certificates* tab — but the **top-level Discovery Status** displayed in the UI remains `Processing`. The *Duration* counter keeps incrementing indefinitely, as if the run never ended. There is no way for the user to see that the discovery actually finished, and no way to close it out.

Effectively the two status fields are out of sync: the provider (connector) sub-status transitions to `Completed`, but Core never rolls that completion up into the discovery's own overall status.

### Steps to Reproduce

1. Configure a Discovery against the **Network Discovery Provider** connector
2. Trigger the run (manual or scheduled)
3. Wait for the connector to finish — the certificates appear on the discovery detail page
4. Observe the discovery detail page:
   - **Discovery Provider Status** = `Completed`
   - Certificates tab is populated with the discovered items
   - **Status** (top-level, header of the page) = still `Processing`
   - **Duration** continues to grow

### Expected Behavior

- When the Discovery Provider Status becomes `Completed` (and certificates have been ingested), the top-level Discovery Status must transition to `Completed` as well
- **Duration** stops incrementing at the moment the run finishes on the provider side
- The two status fields on the same page never disagree — a single source of truth is displayed consistently

### Actual Behavior

- **Discovery Provider Status** = `Completed`, certificates ingested successfully
- **Discovery Status** (top-level) remains `Processing`
- **Duration** counter keeps growing indefinitely
- No user-facing indication that the discovery has, in fact, finished — appears to still be running

### Severity

Major

### Module

Discovery

### Attachments

Image

current version

Image

2.18.0
Image

### Logs

_No response_

### Environment

ILM Core: 2.18.1-SNAPSHOT (14 Jul 2026, 07:45)
Commit: 2b4a8e4
Branch: main
Build time: 2026-07-14T05:45:42Z
ILM Frontend: 2.18.0 (9 Jul 2026, 22:58)
Client:
- macOS Tahoe 26.5.1

lubomirw:
## Summary
- roll persisted discovery status from `PROCESSING` to `COMPLETED` when provider and finish event both completed
- set discovery end time so duration stops increasing
- preserve failed, running-provider, and legacy processing-event behavior

## Test plan
- `mvn test -q`

Closes #1800

### DIFF
--- a/src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
+++ b/src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
@@ -151,9 +151,28 @@ public class CertificateDiscoveredEventHandler extends EventHandler<Certificate>
         List<DiscoveryCertificate> discoveredCertificates = discoveryCertificateRepository.findByDiscoveryUuidAndNewlyDiscovered(eventMessage.getOverrideObjectUuid(), true, Pageable.unpaged());
         logger.debug("Going to process {} triggers on {} discovered certificates", mergedIgnoreTriggers.size() + mergedTriggers.size(), context.getResourceObjects().size());
 
+        // The DISCOVERY_FINISHED event is the only signal that rolls the top-level status out of PROCESSING, so it
+        // must be emitted even when post-processing fails partway; otherwise the discovery is stranded in PROCESSING
+        // with a growing duration and can never be finalized.
+        boolean discoveryFinishEmitted = false;
+        try {
+            discoveryFinishEmitted = handleDiscoveredCertificates(context, discovery, originalMessage, discoveredCertificates, mergedIgnoreTriggers, mergedTriggers);
+        } catch (Exception e) {
+            logger.error("Post-processing of discovered certificates for discovery {} did not complete: {}", discovery.getName(), e.getMessage(), e);
+        } finally {
+            if (!discoveryFinishEmitted) {
+                emitDiscoveryFinished(discovery, context, DiscoveryStatus.WARNING,
+                        "Discovery post-processing did not complete; some certificates may not have been processed.");
+            }
+        }
+    }
+
+    private boolean handleDiscoveredCertificates(EventContext<Certificate> context, DiscoveryHistory discovery, String originalMessage,
+                                                 List<DiscoveryCertificate> discoveredCertificates, List<TriggerAssociation> mergedIgnoreTriggers,
+                                                 List<TriggerAssociation> mergedTriggers) {
         if (discoveredCertificates.isEmpty()) {
-            eventProducer.produceMessage(DiscoveryFinishedEventHandler.constructEventMessage(discovery.getUuid(), context.getUserUuid(), context.getScheduledJobInfo(), new DiscoveryResult(DiscoveryStatus.PROCESSING, originalMessage)));
-            return;
+            emitDiscoveryFinished(discovery, context, DiscoveryStatus.PROCESSING, originalMessage);
+            return true;
         }
 
         EventHistory eventHistoryDiscovery = createEventHistory(ResourceEvent.CERTIFICATE_DISCOVERED, Resource.DISCOVERY, discovery.getUuid());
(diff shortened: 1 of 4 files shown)


src/main/java/com/otilm/core/events/handlers/DiscoveryFinishedEventHandler.java
```diff
@@ -40,12 +40,17 @@ protected EventContext<DiscoveryHistory> prepareContext(EventMessage eventMessag
...
-        // set discovery status to completed when discovery is in preprocessing state coming from certificate discovered event
-        if (discoveryResult.getDiscoveryStatus() == DiscoveryStatus.PROCESSING) {
-            String message = discoveryResult.getMessage() == null ? "Discovery completed successfully." : "Discovery completed successfully. " + discoveryResult.getMessage();
-            discovery.setMessage(message);
-            discovery.setStatus(DiscoveryStatus.COMPLETED);
+        // Certificate post-processing reports back once the discovered certificates have been handled, signalling
+        // PROCESSING on a clean run or WARNING when some certificates failed; only then is the top-level status
+        // finalized. COMPLETED/FAILED payloads originate from the discovery service, which has already persisted
+        // that terminal state, so they are ignored here. Acting only on a not-yet-terminal discovery keeps this
+        // idempotent against event redelivery and leaves a service-finalized discovery untouched.
```
klaraf755: The comment says this guard "keeps this idempotent against event redelivery," but that only protects the `discovery.status`/`endTime`/`message` write. `EventHandler.handleEvent` (base class) unconditionally calls `sendFollowUpEventsNotifications` after `prepareContext` regardless of whether this guard skipped the write — so a genuinely redelivered `DISCOVERY_FINISHED` message (the JMS retry template does retry on `JMSException`) would still re-publish the user notification and re-fire `ScheduledJobFinishedEvent`, even though the DB write is correctly skipped. Suggest either scoping the comment to "idempotent for the persisted status only" or short-circuiting notification dispatch too when the discovery was already terminal on entry.

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
```diff
@@ -151,8 +151,28 @@ public void handleEvent(EventMessage eventMessage) throws EventException {
...
+        // The DISCOVERY_FINISHED event is the only signal that rolls the top-level status out of PROCESSING, so it
+        // must be emitted even when post-processing fails partway; otherwise the discovery is stranded in PROCESSING
+        // with a growing duration and can never be finalized.
+        boolean discoveryFinishEmitted = false;
+        try {
+            handleDiscoveredCertificates(context, discovery, originalMessage, discoveredCertificates, mergedIgnoreTriggers, mergedTriggers);
+            discoveryFinishEmitted = true;
+        } catch (Exception e) {
+            logger.error("Post-processing of discovered certificates for discovery {} did not complete: {}", discovery.getName(), e.getMessage(), e);
+        } finally {
```
klaraf755, line 166: Catching bare `Exception` here swallows everything, including bugs, not just certificate-processing failures — worth a one-line justification in a comment since it's non-obvious why this is safe. I traced the JMS dispatch path: exceptions escaping `handleEvent` never triggered redelivery or a DLQ before this change either (the listener just logs and drops them), so this catch doesn't sacrifice any retry safety net — it replaces "silently strand the discovery in `PROCESSING` forever" with "finalize as `WARNING`," which is strictly better. Might be worth noting that reasoning inline so a future reader doesn't have to re-derive it.

src/main/java/com/otilm/core/events/handlers/DiscoveryFinishedEventHandler.java
lubomirw: Correct — reworded in da23bdac2 to scope the claim to the persisted write and note the base handler still dispatches notifications either way.

Didn't take the short-circuit option, though: the service-finalized path (`finalizeDiscoveryInTx` emits COMPLETED/FAILED with the row *already* terminal) relies on that same dispatch for its legitimate completion notification, and there's no message-id dedup to distinguish that first delivery from a redelivery — so gating notification dispatch on terminal-on-entry would drop real completions. The duplicate scheduled-job SUCCESS on genuine redelivery is tracked separately in #1817.

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
lubomirw: Added that rationale inline in da23bdac2 (broad catch is intentional; escaping exceptions are only logged+dropped by the JMS listener, so finalizing as WARNING beats stranding in PROCESSING).

klaraf755 (approved):

PR #1808, opened 2026-07-14
Dispatch empty dependsOn as NG fire-on-mount callback

### LINKED ISSUE #1807: Empty dependsOn must dispatch as NG fire-on-mount, not fall to legacy 400
### Description

Delivered Core treats an empty (non-null) `dependsOn` list as **not** an NG (Attributes v2) callback, contradicting the feature's own design.

`CallbackServiceImpl.isNgCallback` (`:295-301`) gates on `dependsOn != null && !dependsOn.isEmpty() && callbackContext == null`. But the core design dispatch ladder specifies rung 3 as just `callback.dependsOn != null`, and both the interfaces design (§2) and the shipped `AttributeCallback` Javadoc state: *"empty (non-null) list = fire once when the form opens."* `AttributeEngine`'s own Javadoc (`:1126-1127`) likewise says "the NG branch can safely gate on `dependsOn != null`."

**Consequence:** an NG attribute whose callback depends only on the parent scope — e.g. an RA-profile certificate-template dropdown that reads the authority connection blob from `contextAttributes`, never from `currentAttributes` — must declare `dependsOn: []`. But an empty list fails `isNgCallback`, falls through to the legacy rung, and `AttributeDefinitionUtils.validateCallback` throws a **400** ("Callback context not set" / "Callback method not set"). This permanently blocks scope-only dropdowns for **every stateless (authority provider v3 / NG) connector**. First affected: the ms-adcs-ng certificate-template dropdown.

### Fix scope (Definition of Done)

- [ ] `isNgCallback` gates on `dependsOn != null && callbackContext == null` — drop the `!isEmpty()` conjunct — matching the design ladder. Optional hardening: also exclude `contentType == RESOURCE`, since the connector-level live-listed callback route bypasses ingest validation.
- [ ] Reorder `AttributeEngine.validateCallbackDeclaration` (`:1130-1140`) so the "both `dependsOn` and `callbackContext` set = invalid" and "RESOURCE content type cannot declare `dependsOn`" checks run for an **empty** list too. The current early-return on null-or-empty skips them, so once an empty list routes NG a RESOURCE attribute could evade the RESOURCE guard.
- [ ] Coalesce `currentAttributes` null → empty list in `dispatchNg` / `NgCallbackDispatcher.buildEnvelope`. The envelope field is `@NotNull` and the class is `@JsonInclude(NON_NULL)`, so a null `currentAttributes` is dropped from the JSON and conformant connectors (Java `@Valid`, C# `required`) reject the body. A fire-on-mount callback with no in-form dependency must still send `currentAttributes: []`.
- [ ] Fix first-open definition resolution. Initial NG resolution `getBaseAttributeFromExistingDefinition` (`:254-278`) throws `NotFound` on a stored-definition miss with no registry fallback; `AttributeDefinitionResolver.resolve` (`:73-82`) implements the stored-miss → registry-fetch ladder but is currently unwired (only `resolveForced` is used, for the connector-404 retry). RA-profile definitions are not ingested at `listRAProfileAttributes` (contrast `listAuthorityInstanceAttributes` `:348-351`, which does ingest), so the very first RA-profile form open fires the callback before any definition row exists → 404. Either ingest at listing (parity with `listAuthorityInstanceAttributes`) or wire `AttributeDefinitionResolver.resolve` into initial NG resolution, so fire-on-mount does not deterministically 404 on a fresh system.

### Regression bar

- Legacy callbacks (`callbackContext` set, `dependsOn` null): unchanged.
- Non-empty `dependsOn` NG callbacks: unchanged.
- RESOURCE / GROUP / no-callback attributes: unchanged.
- Only an empty list flips from 400 to NG-fire-on-mount. Nothing ships an empty list successfully today, so there is no regression given the ingest reorder.

### Tests

- Extend `AttributeDeclarationValidityITest` with empty-list variants: empty + `callbackContext` rejected; empty + RESOURCE rejected; empty-only ingests and stores.
- Extend `AttributesV2CallbackDispatchITest` so an empty-`dependsOn` definition dispatches to `POST /v2/attributes/callback`, and assert the envelope carries `currentAttributes: []` (no delivered test touches `currentAttributes` today).

### Epic / siblings

Part of the Connector Attributes v2 API epic **ilm#251**. Sibling fixes tracked separately:
- interfaces — `AttributeCallback` `@Schema` wording (add the empty = fire-on-mount semantics).
- fe-administrator#1764 — fire-on-mount for empty `dependsOn` + emit `attributes: []`.
- ms-adcs-ng connector — emit an empty `attributeCallback`, validate dispatchability for an empty list.

Design references: Connector Attributes v2 API overview §5.1 and the core / interfaces / fe component designs.

### Steps to Reproduce

1. Define an NG attribute (attribute schema v3, DATA) whose `AttributeCallback` sets `dependsOn: []` — a scope-only dropdown, e.g. RA-profile `data_certificateTemplate` on a stateless authority provider v3 connector.
2. Open the host form so the FE fires the mount callback → `POST /v1/{resource}/{parentObjectUuid}/callback`.
3. Observe Core routing: `isNgCallback` returns false for the empty list; the request falls to the legacy rung.

### Expected Behavior

Empty (non-null) `dependsOn` dispatches on the NG path (fire-on-mount): Core builds the `AttributeCallbackRequestDto` envelope with the resolved `contextAttributes` scope chain and `currentAttributes: []`, and POSTs it to the connector's `/v2/attributes/callback`. Matches the design ladder (`dependsOn != null`) and the "empty = fire on form open" contract.

### Actual Behavior

The empty list fails `isNgCallback` (the `!isEmpty()` conjunct), falls to the legacy rung, and `AttributeDefinitionUtils.validateCallback` throws 400 "Callback context not set / method not set". The dropdown never loads. (Separately, on a fresh system the first form-open would 404 at definition resolution even if routing were fixed — see Fix scope item 4.)

### Severity

Major

### Module

Connectors

### Environment

ILM platform, Connector Attributes v2 API (core#1621 / core#1622 shipped). Affects all stateless authority provider v3 / NG connectors that expose scope-only dynamic dropdowns.

lubomirw:
Closes #1807.

**Depends on** OmniTrustILM/interfaces#775 — that PR widens the
`AuthorityInstanceController.listRAProfileAttributes` contract with
`throws AttributeException`; this branch will not compile against `main`
interfaces until #775 is merged and the snapshot is published.

## Problem

An empty (non-null) `dependsOn` declares an Attributes v2 callback that
fires once when the form opens — a dropdown reading only the parent
scope via `contextAttributes`, never a sibling field. Delivered Core
rejected it: `isNgCallback` required a *non-empty* list, so an empty one
fell through to the legacy rung and 400'd (`Callback context not set`).
This permanently blocked scope-only dropdowns for every stateless
(authority provider v3 / NG) connector; first hit is the ms-adcs-ng
certificate-template dropdown.

## Fix

- **`isNgCallback`** — gate on `dependsOn != null` (drop the `!isEmpty()`
  conjunct), matching the design dispatch ladder. Also exclude RESOURCE
  content so a connector-level live-listed definition (which bypasses the
  ingest-time declaration guard) cannot route a RESOURCE attribute
  through NG — RESOURCE is answered by the core resource path (rung 1).
- **`validateCallbackDeclaration`** — run the both-set and RESOURCE
  guards for an empty list too; the early return no longer short-circuits
  on empty.
- **`buildEnvelope`** — coalesce a null `currentAttributes` to `[]`. The
  field is `@NotNull` and the DTO is `@JsonInclude(NON_NULL)`, so a
  fire-on-mount callback with no in-form dependency would otherwise drop
  the field and a conformant connector would reject the body.
- **`listRAProfileAttributes`** — v3 authorities ingest their RA-profile
  definitions at list time (mirroring `listAuthorityInstanceAttributes`),
  so the name-only fire-on-mount callback resolves the definition from
  Core storage instead of 404'ing on a fresh system. (The FE fire-on-mount
  payload is name-only, so a UUID-keyed registry resolve would not fire.)

## Regression bar

Legacy callbacks (`callbackContext` set, `dependsOn` null) unchanged;
non-empty `dependsOn` NG unchanged; RESOURCE / GROUP / no-callback
unchanged. Only empty flips from 400 to NG fire-on-mount — nothing ships
an empty list successfully today.

## Tests

- `AttributeDeclarationValidityITest` — empty-list variants (empty +
  callbackContext rejected; empty + RESOURCE rejected; empty-only
  ingests).
- `AttributesV2CallbackDispatchITest` — an empty-`dependsOn` definition
  dispatches to `POST /v2/attributes/callback` and the envelope carries
  `currentAttributes: []`.

Part of the Connector Attributes v2 API epic (ilm#251). Sibling fixes:
interfaces `@Schema` wording (#775), fe-administrator#1764, ms-adcs-ng#26.

### DIFF
--- a/src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
+++ b/src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
@@ -1122,13 +1122,15 @@ public class AttributeEngine {
 
     /**
      * Declaration validity for the NG ({@code dependsOn}) callback shape at the ingest
-     * choke point. {@code dependsOn} (NG, scope-resolved) and {@code callbackContext} (legacy, body-mapped)
-     * are mutually exclusive: a definition declaring both is ambiguous to dispatch, so the NG branch can
-     * safely gate on {@code dependsOn != null}. {@code dependsOn} on a RESOURCE-content attribute is also
-     * rejected — RESOURCE content is resolved through the core resource path, not an NG callback.
+     * choke point. A non-null {@code dependsOn} — including an empty list ("fire once on form open") — is an NG
+     * declaration, so the guards below must run for the empty case too. {@code dependsOn} (NG, scope-resolved) and
+     * {@code callbackContext} (legacy, body-mapped) are mutually exclusive: a definition declaring both is ambiguous
+     * to dispatch, so the NG branch can safely gate on {@code dependsOn != null}. {@code dependsOn} on a
+     * RESOURCE-content attribute is also rejected — RESOURCE content is resolved through the core resource path,
+     * not an NG callback.
      */
     private static void validateCallbackDeclaration(BaseAttribute attribute, AttributeCallback callback, String connectorUuidStr) throws AttributeException {
-        if (callback == null || callback.getDependsOn() == null || callback.getDependsOn().isEmpty()) {
+        if (callback == null || callback.getDependsOn() == null) {
             return;
         }
         if (callback.getCallbackContext() != null) {
--- a/src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
+++ b/src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
@@ -1121,13 +1121,16 @@ public class AttributeEngine {
     }
 
     /**
-     * Declaration validity for the NG ({@code dependsOn}) callback shape at the ingest
-     * choke point. A non-null {@code dependsOn} — including an empty list ("fire once on form open") — is an NG
-     * declaration, so the guards below must run for the empty case too. {@code dependsOn} (NG, scope-resolved) and
-     * {@code callbackContext} (legacy, body-mapped) are mutually exclusive: a definition declaring both is ambiguous
-     * to dispatch, so the NG branch can safely gate on {@code dependsOn != null}. {@code dependsOn} on a
-     * RESOURCE-content attribute is also rejected — RESOURCE content is resolved through the core resource path,
-     * not an NG callback.
+     * Declaration validity for the NG ({@code dependsOn}) callback shape at the ingest choke point.
+     *
+     * <p><b>Empty list is still NG.</b> A non-null {@code dependsOn} — including an empty list ("fire once on form
+     * open") — is an NG declaration, so the guards below must run for the empty case too.
+     *
+     * <p><b>Mutual exclusion.</b> {@code dependsOn} (NG, scope-resolved) and {@code callbackContext} (legacy,
+     * body-mapped) cannot both be set; that lets the dispatch path gate NG on {@code dependsOn != null}.
+     *
+     * <p><b>RESOURCE is rejected.</b> {@code dependsOn} on a RESOURCE-content attribute is refused — RESOURCE content
+     * resolves through the core resource path, not an NG callback.
      */
     private static void validateCallbackDeclaration(BaseAttribute attribute, AttributeCallback callback, String connectorUuidStr) throws AttributeException {
         if (callback == null || callback.getDependsOn() == null) {
--- a/src/main/java/com/otilm/core/service/callback/NgCallbackDispatcher.java
+++ b/src/main/java/com/otilm/core/service/callback/NgCallbackDispatcher.java
@@ -167,7 +167,9 @@ public class NgCallbackDispatcher {
         }
         envelope.setAttributeName(definition.getName());
         envelope.setContextAttributes(context.contextAttributes());
-        envelope.setCurrentAttributes(context.currentAttributes());
+        // currentAttributes is @NotNull and the DTO is @JsonInclude(NON_NULL): a null (fire-on-mount callback with
+        // no in-form dependency) would be dropped from the JSON, so a conformant connector rejects the body. Send [].
+        envelope.setCurrentAttributes(context.currentAttributes() == null ? List.of() : context.currentAttributes());
         envelope.setPagination(callback.getPagination());
         return envelope;
     }
--- a/src/main/java/com/otilm/core/service/callback/NgCallbackDispatcher.java
+++ b/src/main/java/com/otilm/core/service/callback/NgCallbackDispatcher.java
@@ -27,6 +27,7 @@ import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.transaction.support.DefaultTransactionDefinition;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 import java.util.UUID;
@@ -168,8 +169,9 @@ public class NgCallbackDispatcher {
         envelope.setAttributeName(definition.getName());
         envelope.setContextAttributes(context.contextAttributes());
         // currentAttributes is @NotNull and the DTO is @JsonInclude(NON_NULL): a null (fire-on-mount callback with
-        // no in-form dependency) would be dropped from the JSON, so a conformant connector rejects the body. Send [].
-        envelope.setCurrentAttributes(context.currentAttributes() == null ? List.of() : context.currentAttributes());
+        // no in-form dependency) would be dropped from the JSON, so a conformant connector rejects the body. Send an
+        // empty list — mutable, to match the pass-through branch's Jackson-deserialized ArrayList.
+        envelope.setCurrentAttributes(context.currentAttributes() == null ? new ArrayList<>() : context.currentAttributes());
         envelope.setPagination(callback.getPagination());
         return envelope;
     }
--- a/src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
@@ -353,9 +353,17 @@ public class AuthorityInstanceServiceImpl implements AuthorityInstanceExternalSe
 
     @Override
     @ExternalAuthorization(resource = Resource.AUTHORITY, action = ResourceAction.ANY)
-    public List<BaseAttribute> listRAProfileAttributes(SecuredUUID uuid) throws ConnectorException, NotFoundException {
+    public List<BaseAttribute> listRAProfileAttributes(SecuredUUID uuid) throws ConnectorException, NotFoundException, AttributeException {
         AuthorityInstanceReference authorityInstance = getAuthorityInstanceReferenceEntity(uuid);
-        return adapterFactory.forAuthority(authorityInstance).listRaProfileAttributes(authorityInstance);
+        List<BaseAttribute> attributes = adapterFactory.forAuthority(authorityInstance).listRaProfileAttributes(authorityInstance);
+        if (isV3(authorityInstance.getConnectorInterface())) {
+            // v3 is stateless: the resourceCallback RA-profile arm has no connector-side authorityInstanceUuid to
+            // re-list with, so a fire-on-mount callback resolves its definition from Core storage by name. Ingest at
+            // list time — mirroring listAuthorityInstanceAttributes — so resolution succeeds on a fresh system;
+            // otherwise the first callback 404s before any definition row exists.
+            attributeEngine.updateDataAttributeDefinitions(authorityInstance.getConnectorUuid(), null, attributes);
+        }
+        return attributes;
     }
 
     @Override
--- a/src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
@@ -353,11 +353,10 @@ public class AuthorityInstanceServiceImpl implements AuthorityInstanceExternalSe
 
     @Override
     @ExternalAuthorization(resource = Resource.AUTHORITY, action = ResourceAction.ANY)
-    // NOT_SUPPORTED (as in listAuthorityInstanceAttributes): the connector list call is HTTP and must not hold a DB
-    // connection across the round-trip; the v3 ingest below opens its own short transaction via
-    // updateDataAttributeDefinitions. Lazy entity reads ride the open-session-in-view session.
-    @Transactional(propagation = Propagation.NOT_SUPPORTED)
     public List<BaseAttribute> listRAProfileAttributes(SecuredUUID uuid) throws ConnectorException, NotFoundException, AttributeException {
+        // Runs under the class-level (REQUIRED) transaction: adapter selection and the v3 check below dereference the
+        // LAZY connectorInterface, so a persistence context must be open — do not switch this to NOT_SUPPORTED without
+        // eagerly loading that association first.
         AuthorityInstanceReference authorityInstance = getAuthorityInstanceReferenceEntity(uuid);
         List<BaseAttribute> attributes = adapterFactory.forAuthority(authorityInstance).listRaProfileAttributes(authorityInstance);
         if (isV3(authorityInstance.getConnectorInterface())) {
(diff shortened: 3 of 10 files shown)


src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
```diff
@@ -353,9 +353,21 @@ public List<BaseAttribute> listAuthorityInstanceAttributes(SecuredUUID connector
...
-    public List<BaseAttribute> listRAProfileAttributes(SecuredUUID uuid) throws ConnectorException, NotFoundException {
+    // NOT_SUPPORTED (as in listAuthorityInstanceAttributes): the connector list call is HTTP and must not hold a DB
+    // connection across the round-trip; the v3 ingest below opens its own short transaction via
+    // updateDataAttributeDefinitions. Lazy entity reads ride the open-session-in-view session.
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    public List<BaseAttribute> listRAProfileAttributes(SecuredUUID uuid) throws ConnectorException, NotFoundException, AttributeException {
         AuthorityInstanceReference authorityInstance = getAuthorityInstanceReferenceEntity(uuid);
-        return adapterFactory.forAuthority(authorityInstance).listRaProfileAttributes(authorityInstance);
+        List<BaseAttribute> attributes = adapterFactory.forAuthority(authorityInstance).listRaProfileAttributes(authorityInstance);
+        if (isV3(authorityInstance.getConnectorInterface())) {
```
ivosh, line 362: **~ Minor** | Correctness

AuthorityInstanceServiceImpl.java:363 dereferences the LAZY association `authorityInstance.getConnectorInterface()` while the method runs under `@Transactional(propagation = NOT_SUPPORTED)` (line 359) on a class that is otherwise `@Transactional` (line 57). `connectorInterface` is `@ManyToOne(fetch = FetchType.LAZY)` (AuthorityInstanceReference.java:56-59) and `getAuthorityInstanceReferenceEntity`->`findByUuid` (line 566) carries no `@EntityGraph`, so the entity comes back with the association unloaded and no ambient transaction/persistence-context is open under NOT_SUPPORTED. The added comment (line 358) bets on this: 'Lazy entity reads ride the open-session-in-view session.'

Two problems ground this as a likely defect:
1. DIVERGENCE FROM THE ADJACENT SIBLING. `listAuthorityInstanceAttributes` (line 331), also NOT_SUPPORTED, does NOT rely on OSIV: it wraps its LAZY read in `transactionHandler.runInTransaction(...)` (line 336) with the comment 'reads the connector's LAZY interfaces collection, so it must run inside a transaction.' That explicit wrap is direct in-file evidence the team does not trust OSIV for lazy reads under NOT_SUPPORTED. The new method reads a LAZY association the sibling would have wrapped.
2. OSIV IS NOT PINNED ON. `spring.jpa.open-in-view` is not set in src/main/resources/application.yml (or anywhere). Spring Boot defaults it to enabled, so the HTTP-only entry path (AuthorityInstanceControllerImpl.listRAProfileAttributes is the only caller found) likely works today — but the moment OSIV is disabled (Spring Boot's own recommended setting, logged as a startup warning) this throws `LazyInitializationException` -> 500. The unit test `AuthorityInstanceServiceImplV3Test` mocks the repository and hands back a fully-populated entity, so it exercises none of the real lazy-load and would stay green through the failure.

Recommendation: match the sibling: read/`isV3` the interface inside a short `transactionHandler.runInTransaction(...)` (as line 336 does) before the connector HTTP call, so the read is correct regardless of OSIV.

src/test/java/com/otilm/core/integration/attribute/AttributeDeclarationValidityITest.java
```diff
@@ -92,6 +92,43 @@ void dependsOnOnly_ingestsCleanly() {
...
+
+    @Test
+    void emptyDependsOnOnResourceContent_rejected() {
+        // The RESOURCE guard must fire for an empty dependsOn too — RESOURCE content is never resolved via NG.
+        DataAttributeV2 a = base(AttributeContentType.RESOURCE);
+        AttributeCallback callback = new AttributeCallback();
+        callback.setDependsOn(List.of());
+        a.setAttributeCallback(callback);
+
+        Assertions.assertThrows(AttributeException.class,
```
ivosh: **~ Minor** | Testing

AttributeDeclarationValidityITest.java:117 — emptyDependsOnOnResourceContent_rejected asserts only `assertThrows(AttributeException.class, ...)`, with no check on the message. The intended target is the RESOURCE guard in AttributeEngine.validateCallbackDeclaration (AttributeEngine.java:1139-1141, message "Attribute with Resource Content Type cannot declare a dependsOn callback"). But updateDataAttributeDefinitions runs validateAttributeProperties BEFORE validateCallbackDeclaration (AttributeEngine.java:1116-1118), and that step also throws AttributeException — so a future change that makes the fixture's properties invalid would let this test stay green while never exercising the RESOURCE guard (the CLAUDE.md "tests that look green but cover nothing" hazard: the assertion checks the type, not the origin). The sibling both-set test one method up (line 106) does the right thing: `assertTrue(ex.getMessage().contains("dependsOn"))`. Recommend capturing the exception and asserting `getMessage().contains("Resource")` here too. Non-blocking. Note the pre-existing sibling dependsOnOnResourceContent_rejected (line 81) has the same gap, so this only continues an existing local pattern rather than introducing a new smell.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -1122,13 +1122,15 @@ private void validateAttributeDefinition(BaseAttribute attribute, UUID connector
 
     /**
      * Declaration validity for the NG ({@code dependsOn}) callback shape at the ingest
-     * choke point. {@code dependsOn} (NG, scope-resolved) and {@code callbackContext} (legacy, body-mapped)
-     * are mutually exclusive: a definition declaring both is ambiguous to dispatch, so the NG branch can
-     * safely gate on {@code dependsOn != null}. {@code dependsOn} on a RESOURCE-content attribute is also
-     * rejected — RESOURCE content is resolved through the core resource path, not an NG callback.
+     * choke point. A non-null {@code dependsOn} — including an empty list ("fire once on form open") — is an NG
```
ivosh: **. Nitpick** | Documentation

AttributeEngine.java:1123-1131 — the updated Javadoc is accurate but reads as a single wall-of-text paragraph that fuses four distinct concerns (what the choke point validates; that an empty list is still an NG declaration so the guards run; the dependsOn/callbackContext mutual-exclusion and the resulting `!= null` gate; the RESOURCE rejection) with 30+ word em-dash/colon/parenthetical-chained sentences and no per-concern lead-ins. Density markers (a), (b) and (d) from the lens all hit. Content is good WHY material — restructure, don't trim: one summary sentence, then a `<p>` per concern with a bold lead-in (e.g. `<b>Empty list is still NG.</b>`, `<b>Mutual exclusion:</b>`, `<b>RESOURCE is rejected:</b>`), sentences ~20 words. Nitpick, non-blocking (private method, not a public API).

src/main/java/com/otilm/core/service/callback/NgCallbackDispatcher.java
```diff
@@ -167,7 +167,9 @@ private AttributeCallbackRequestDto buildEnvelope(NgDispatchContext context, Req
         }
         envelope.setAttributeName(definition.getName());
         envelope.setContextAttributes(context.contextAttributes());
-        envelope.setCurrentAttributes(context.currentAttributes());
+        // currentAttributes is @NotNull and the DTO is @JsonInclude(NON_NULL): a null (fire-on-mount callback with
+        // no in-form dependency) would be dropped from the JSON, so a conformant connector rejects the body. Send [].
+        envelope.setCurrentAttributes(context.currentAttributes() == null ? List.of() : context.currentAttributes());
```
ivosh: **. Nitpick** | Maintainability

NgCallbackDispatcher.java:172 coalesces a null `currentAttributes` to the immutable `List.of()`, while the same ternary's else-branch passes through `context.currentAttributes()` — which originates as `callback.getAttributes()` (CallbackServiceImpl.java:326), a mutable Jackson-deserialized `ArrayList`. So the field is assigned an immutable list on the fire-on-mount path and a mutable one otherwise. This is the immutable-collection-into-a-mutable-slot smell (inverse of the defensive-copy item; cf. Effective Java Item 50 context): a future consumer that follows a mutate-in-place pattern (`envelope.getCurrentAttributes().add(...)`) would work on the non-null path and throw `UnsupportedOperationException` on the empty path. 

**Safe today**; there is no `getCurrentAttributes()` caller — `AttributeCallbackRequestDto.currentAttributes` is a write-only outbound DTO field with no in-field initializer, populated then serialized to JSON, never read back or mutated in Core. The PR body discloses the `[]` intent, so the behaviour is mot likely correct. If you want branch-consistency (both branches hand the DTO a mutable list), use `new ArrayList<>()` instead of `List.of()`; otherwise leave it — it is a genuine, deliberate empty list on a serialize-only field.

ivosh (changes requested):

src/test/java/com/otilm/core/service/impl/AuthorityInstanceServiceImplV3Test.java
```diff
@@ -304,6 +304,26 @@ void listRAProfileAttributesRoutesThroughAdapterForV3() throws Exception {
...
+        v2Iface.setVersion("v2");
+        v2Iface.setConnectorUuid(connectorUuid);
+        existing.setConnectorInterface(v2Iface);
+        when(authorityInstanceReferenceRepository.findByUuid(any(SecuredUUID.class))).thenReturn(Optional.of(existing));
+        when(v3Adapter.listRaProfileAttributes(existing)).thenReturn(List.of(mock(BaseAttribute.class)));
+
+        service.listRAProfileAttributes(SecuredUUID.fromUUID(existing.uuid));
+
+        // Legacy (v1/v2) authorities re-list on the callback path, so list-time ingest must stay v3-only.
+        verify(attributeEngine, never()).updateDataAttributeDefinitions(any(), any(), any());
```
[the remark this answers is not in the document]
lubomirw: Fixed in 5d25e33 — the test now asserts the returned list is the adapter's definitions, not just that ingest is skipped.

src/main/java/com/otilm/core/service/impl/AuthorityInstanceServiceImpl.java
lubomirw: Addressed in 5d25e33. Reverted to the class-level (REQUIRED) transaction rather than wrapping the read in runInTransaction: adapter selection (forAuthority) dereferences the LAZY connectorInterface too, not just the isV3 check, so the whole method needs an open persistence context. The connector-call-in-transaction is pre-existing for every adapter-backed list method and out of scope here, so I kept the safe ambient-tx behaviour instead of a NOT_SUPPORTED that leans on OSIV.

src/test/java/com/otilm/core/integration/attribute/AttributeDeclarationValidityITest.java
lubomirw: Fixed in 5d25e33 — captures the exception and asserts the message contains "dependsOn", pinning the RESOURCE guard rather than the earlier validateAttributeProperties step.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
lubomirw: Fixed in 5d25e33 — restructured into one <p> per concern (empty-is-NG / mutual-exclusion / RESOURCE) with bold lead-ins.

src/main/java/com/otilm/core/service/callback/NgCallbackDispatcher.java
lubomirw: Fixed in 5d25e33 — now new ArrayList<>() so both branches hand the DTO a mutable list.

ivosh (approved):

PR #1810, opened 2026-07-15
Align certificate and certificate request DB schema with entities

### LINKED ISSUE #1805: Database - Entities: Certificate and CertificateRequest schema drift from migrations
### Description

The JPA entities `Certificate` and `CertificateRequestEntity` have drifted from the effective database schema produced by the Flyway migrations (SQL migrations in `src/main/resources/db/migration` plus Java migrations in `src/main/java/db/migration`). The drift falls into three categories:

**1. Dead columns — present in the database, absent from the entity**

`certificate` table:
- `discovery_uuid BIGINT` — created in `V202201041211__init.sql`, never dropped. All other init-era legacy columns (`id`, `status`, `ra_profile_id`, `group_id`, `owner`, `meta`, `entity_id`, `basic_constraints`) were removed by later migrations; this one was missed. It is still `BIGINT` from the pre-UUID-PK era, so it cannot reference anything.

`certificate_request` table:
- `subject_dn_normalized VARCHAR` — added by `V202311071500__IssuerAndSubjectDnMigration`, but `CertificateRequestEntity` never got the corresponding field (the `Certificate` entity has one). The column is populated once by the migration and then never written again, so it silently goes stale for new rows — it should either be mapped and maintained, or dropped.

**2. Nullability understated in entity annotations (DB is stricter than JPA)**

These columns are `NOT NULL` in the database but the entity `@Column` lacks `nullable = false`, so nothing at the JPA level prevents persisting a null — it fails only at insert time with a database constraint violation:

- `Certificate`: `public_key_algorithm`, `subject_dn` (both `NOT NULL DEFAULT ''` since init; note `issuer_dn` had its NOT NULL dropped in `V202303200000` but `subject_dn` did not), `state`, `validation_status` (`SET NOT NULL` in `V202310131425__split_certificate_status.sql`), `certificate_type` (`SET NOT NULL` in `V202306141313__certificate_request.sql`)
- `CertificateRequestEntity`: `subject_dn`, `public_key_algorithm`, `fingerprint`, `certificate_type`, `content`, `certificate_request_format` (all `NOT NULL` since `V202306141313`)

For contrast, `subject_type` and `compliance_status` are annotated correctly (`nullable = false` plus a Java-side default) — the columns above should get the same treatment.

**3. Type-level mismatches**

- `certificate.status_validation_timestamp` is `timestamp` (without time zone, added by `V202301311500__PublicKeyMigration`) but is mapped as `OffsetDateTime`, which Hibernate 6 expects as `timestamptz`. Offsets are silently normalized through the JVM zone. Newer tables (e.g. `certificate_registration`) already use `TIMESTAMPTZ`.
- `key_usage` on both tables was rebuilt as a plain nullable `INTEGER` (drop + re-add, no `NOT NULL`, no default) by `V202508261555__EnumCollectionsColumnsBitmask` / `V202509041555__CertificateRequestEntityBitmask`, while both entities map it as primitive `int`. The migrations backfilled existing rows, but any future NULL in that column (raw SQL, later migration) makes entity hydration throw.
- `certificate.hybrid_certificate` and `certificate.archived` are nullable `BOOLEAN` columns (with `DEFAULT false`) mapped to primitive `boolean` — same NULL-read hazard, mitigated by the column default.

### Steps to Reproduce

1. Build the effective schema of the `certificate` and `certificate_request` tables by replaying all Flyway migrations (SQL in `src/main/resources/db/migration` and Java in `src/main/java/db/migration`), or inspect a migrated database with `\d certificate` / `\d certificate_request`.
2. Compare against the column mappings in `com.otilm.core.dao.entity.Certificate` and `com.otilm.core.dao.entity.CertificateRequestEntity`.

### Expected Behavior

- Every database column is mapped by the corresponding entity, or has been dropped by a migration.
- Entity `@Column(nullable = ...)` annotations match the database NOT NULL constraints.
- Column types match what Hibernate 6 expects for the mapped Java types.

### Actual Behavior

- 2 dead columns across the two tables (`discovery_uuid`, and effectively-stale `subject_dn_normalized` on `certificate_request`)
- 11 columns whose NOT NULL constraint is not reflected in the entity annotations.
- `timestamp` vs `OffsetDateTime` mismatch on `status_validation_timestamp`; nullable `INTEGER`/`BOOLEAN` columns mapped to Java primitives.

### Severity

Minor

### Module

None

### Screenshot

_No response_

### Logs

_No response_

### Environment

_No response_

klaraf755:
Closes #1805

### DIFF
--- a/src/main/resources/db/migration/V202607141200__align_certificate_schema_with_entities.sql
+++ b/src/main/resources/db/migration/V202607141200__align_certificate_schema_with_entities.sql
@@ -24,16 +24,12 @@ UPDATE certificate SET archived = false WHERE archived IS NULL;
 ALTER TABLE certificate
     ALTER COLUMN key_usage SET DEFAULT 0,
     ALTER COLUMN key_usage SET NOT NULL,
+    ALTER COLUMN hybrid_certificate SET DEFAULT false,
     ALTER COLUMN hybrid_certificate SET NOT NULL,
+    ALTER COLUMN archived SET DEFAULT false,
     ALTER COLUMN archived SET NOT NULL;
 
 UPDATE certificate_request SET key_usage = 0 WHERE key_usage IS NULL;
 ALTER TABLE certificate_request
     ALTER COLUMN key_usage SET DEFAULT 0,
     ALTER COLUMN key_usage SET NOT NULL;
-
--- The entity maps status_validation_timestamp as OffsetDateTime, which pairs with TIMESTAMPTZ.
--- Existing values are interpreted in the session time zone, which the JDBC driver sets to the
--- JVM default — the same zone Hibernate normalized the wall-clock values to when writing them.
-ALTER TABLE certificate
-    ALTER COLUMN status_validation_timestamp TYPE TIMESTAMPTZ;
--- a/src/main/resources/db/migration/V202607141200__align_certificate_schema_with_entities.sql
+++ b/src/main/resources/db/migration/V202607141200__align_certificate_schema_with_entities.sql
@@ -1,0 +1,33 @@
+-- Align certificate and certificate_request schema with the JPA entities.
+-- Relaxed columns: registration creates identity-only placeholders without key material
+-- (public_key_algorithm) and, for SAN-only identities, without a subject DN; the CSR-attach
+-- path copies both into certificate_request. Dropped columns have no readers in code.
+ALTER TABLE certificate
+    ALTER COLUMN public_key_algorithm DROP NOT NULL,
+    ALTER COLUMN public_key_algorithm DROP DEFAULT,
+    ALTER COLUMN subject_dn DROP NOT NULL,
+    ALTER COLUMN subject_dn DROP DEFAULT,
+    ALTER COLUMN state DROP NOT NULL,
+    ALTER COLUMN validation_status DROP NOT NULL,
+    ALTER COLUMN certificate_type DROP NOT NULL,
+    DROP COLUMN discovery_uuid;
+
+ALTER TABLE certificate_request
+    ALTER COLUMN subject_dn DROP NOT NULL,
+    ALTER COLUMN public_key_algorithm DROP NOT NULL,
+    DROP COLUMN subject_dn_normalized;
+
+-- Columns mapped to Java primitives can never hold null; backfill defensively, then constrain.
+UPDATE certificate SET key_usage = 0 WHERE key_usage IS NULL;
+UPDATE certificate SET hybrid_certificate = false WHERE hybrid_certificate IS NULL;
+UPDATE certificate SET archived = false WHERE archived IS NULL;
+ALTER TABLE certificate
+    ALTER COLUMN key_usage SET DEFAULT 0,
+    ALTER COLUMN key_usage SET NOT NULL,
+    ALTER COLUMN hybrid_certificate SET NOT NULL,
+    ALTER COLUMN archived SET NOT NULL;
+
+UPDATE certificate_request SET key_usage = 0 WHERE key_usage IS NULL;
+ALTER TABLE certificate_request
+    ALTER COLUMN key_usage SET DEFAULT 0,
+    ALTER COLUMN key_usage SET NOT NULL;
--- a/src/main/resources/db/migration/V202607141200__align_certificate_schema_with_entities.sql
+++ b/src/main/resources/db/migration/V202607141200__align_certificate_schema_with_entities.sql
@@ -31,3 +31,9 @@ UPDATE certificate_request SET key_usage = 0 WHERE key_usage IS NULL;
 ALTER TABLE certificate_request
     ALTER COLUMN key_usage SET DEFAULT 0,
     ALTER COLUMN key_usage SET NOT NULL;
+
+-- The entity maps status_validation_timestamp as OffsetDateTime, which pairs with TIMESTAMPTZ.
+-- Existing values are interpreted in the session time zone, which the JDBC driver sets to the
+-- JVM default — the same zone Hibernate normalized the wall-clock values to when writing them.
+ALTER TABLE certificate
+    ALTER COLUMN status_validation_timestamp TYPE TIMESTAMPTZ;
--- a/src/test/java/com/otilm/core/integration/service/ClientOperationServiceV2ITest.java
+++ b/src/test/java/com/otilm/core/integration/service/ClientOperationServiceV2ITest.java
@@ -58,7 +58,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.test.context.bean.override.mockito.MockitoBean;
 import org.springframework.test.context.bean.override.mockito.MockitoSpyBean;
-import org.testcontainers.shaded.org.checkerframework.checker.units.qual.C;
 
 import java.io.IOException;
 import java.io.InputStream;
--- a/src/test/java/com/otilm/core/integration/service/ClientOperationServiceV2ITest.java
+++ b/src/test/java/com/otilm/core/integration/service/ClientOperationServiceV2ITest.java
@@ -40,6 +40,7 @@ import com.otilm.core.util.CertificateTestUtil;
 import com.otilm.core.util.CertificateUtil;
 import com.github.tomakehurst.wiremock.WireMockServer;
 import com.github.tomakehurst.wiremock.client.WireMock;
+import com.otilm.core.util.builders.CertificateRequestEntityBuilder;
 import org.bouncycastle.asn1.x500.X500Name;
 import org.bouncycastle.asn1.x509.Extension;
 import org.bouncycastle.cert.X509v3CertificateBuilder;
@@ -79,8 +80,6 @@ import static org.mockito.Mockito.when;
 
 @SpringBootTest
 class ClientOperationServiceV2ITest extends BaseSpringBootTest {
-
-    public static final String RA_PROFILE_NAME = "testRaProfile1";
 
     private static final String SAMPLE_PKCS10 = """
             -----BEGIN CERTIFICATE REQUEST-----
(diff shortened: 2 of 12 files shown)


src/test/java/com/otilm/core/util/builders/CertificateRequestEntityBuilder.java
```diff
@@ -1,6 +1,8 @@
 package com.otilm.core.util.builders;
 
+import com.otilm.api.model.core.certificate.CertificateType;
 import com.otilm.api.model.core.enums.CertificateRequestFormat;
+import com.otilm.core.dao.entity.Certificate;
 import com.otilm.core.dao.entity.CertificateRequestEntity;
```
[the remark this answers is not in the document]
klaraf755: Removed.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceV2ITest.java
```diff
@@ -62,6 +58,7 @@
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.test.context.bean.override.mockito.MockitoBean;
 import org.springframework.test.context.bean.override.mockito.MockitoSpyBean;
+import org.testcontainers.shaded.org.checkerframework.checker.units.qual.C;
```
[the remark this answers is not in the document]
klaraf755: Removed.

src/main/resources/db/migration/V202607141200__align_certificate_schema_with_entities.sql
```diff
@@ -0,0 +1,39 @@
+-- Align certificate and certificate_request schema with the JPA entities.
```
ivosh, line 1: **~ Minor** | Correctness

— `V202607141200__…sql:38-39`. `ALTER COLUMN status_validation_timestamp TYPE TIMESTAMPTZ` is not binary-coercible in PostgreSQL, so it rewrites the entire `certificate` table under an `ACCESS EXCLUSIVE` lock; the three `SET NOT NULL` statements additionally each require a full validating scan. `certificate` is typically one of the largest tables in a deployment, and there is no `CONCURRENTLY` path for `ALTER … TYPE`, so this is effectively blocking downtime for the duration. Why it matters: on a large production DB this can lock out all certificate operations for a long time, and Flyway wraps it in one transaction holding the lock throughout. 


Fix: nothing to change in the SQL itself, but flag this in the PR/release notes as a maintenance-window migration, and confirm with ops on the largest deployment's row count what the expected rewrite time is. If it's prohibitive, consider splitting the TYPE change out and staging it.

src/main/resources/db/migration/V202607141200__align_certificate_schema_with_entities.sql
```diff
@@ -0,0 +1,39 @@
...
+    ALTER COLUMN certificate_type DROP NOT NULL,
+    DROP COLUMN discovery_uuid;
+
+ALTER TABLE certificate_request
+    ALTER COLUMN subject_dn DROP NOT NULL,
+    ALTER COLUMN public_key_algorithm DROP NOT NULL,
+    DROP COLUMN subject_dn_normalized;
+
+-- Columns mapped to Java primitives can never hold null; backfill defensively, then constrain.
+UPDATE certificate SET key_usage = 0 WHERE key_usage IS NULL;
```
ivosh, line 21: **~ Minor** | Testing

V202607141200__align_certificate_schema_with_entities.sql:21-38 performs three data transforms on EXISTING rows: the defensive backfill UPDATEs (key_usage/hybrid_certificate/archived -> 0/false, certificate_request.key_usage -> 0) and the `status_validation_timestamp TYPE TIMESTAMPTZ` conversion whose own comment (lines 35-37) admits a non-trivial session-time-zone assumption. 

In the test suite Flyway applies this migration against an EMPTY schema (BaseSpringBootTest bootstrap), so none of these transforms ever runs against a row -- they pass green while covering nothing (see CLAUDE.md 'Tests that look green but cover nothing'). 

The repo already has the right pattern: EnumCollectionsColumnsBitmaskITest / MigrateToComplianceProfilesV2ITest seed pre-migration rows (simulateOldEnvironment / raw INSERT) then assert post-migration values. 

Recommend a migration ITest that seeds a certificate row with NULL key_usage and a naive-timestamp status_validation_timestamp, runs V202607141200, and asserts key_usage=0 and the timestamp's instant is preserved. Non-blocking, but the TIMESTAMPTZ instant-preservation is exactly the kind of subtle transform that warrants a data-bearing test.

src/main/resources/db/migration/V202607141200__align_certificate_schema_with_entities.sql
```diff
@@ -0,0 +1,39 @@
...
+UPDATE certificate SET hybrid_certificate = false WHERE hybrid_certificate IS NULL;
+UPDATE certificate SET archived = false WHERE archived IS NULL;
+ALTER TABLE certificate
+    ALTER COLUMN key_usage SET DEFAULT 0,
+    ALTER COLUMN key_usage SET NOT NULL,
+    ALTER COLUMN hybrid_certificate SET NOT NULL,
```
ivosh, line 28: **. Nitpick** | Correctness

V202607141200__align_certificate_schema_with_entities.sql:24-28. The block backfills and then constrains three primitive-mapped columns, but treats them inconsistently: `key_usage` gets `ALTER COLUMN key_usage SET DEFAULT 0` (line 25) before `SET NOT NULL`, while `hybrid_certificate` (line 27) and `archived` (line 28) get `SET NOT NULL` with no `SET DEFAULT false`.

The header comment on line 20 ("Columns mapped to Java primitives can never hold null; backfill defensively, then constrain") states the same rationale for all three, so the missing defaults contradict the stated intent. Consequence: any INSERT that omits these columns fails the NOT NULL check for hybrid_certificate/archived but silently defaults key_usage to 0. 

This is not a live bug — the JPA entity always supplies them (primitive fields initialised `= false`) and the native insert `CertificateRepository.insertWithFingerprintConflictResolve` binds all three — but the raw-SQL certificate INSERT in MigrateToComplianceProfilesV2ITest had to be amended in this same PR to add all three columns, showing raw inserts exist. If hybrid_certificate/archived are intended to have DB defaults like key_usage, add `SET DEFAULT false` for both; otherwise consider dropping the `SET DEFAULT 0` on key_usage so the three are uniform.

src/test/java/com/otilm/core/util/builders/CertificateRequestEntityBuilder.java
```diff
@@ -56,6 +57,7 @@ public CertificateRequestEntity build() {
                 throw new IllegalStateException("Failed to set certificate request content", e);
             }
         }
+        entity.setCertificateType(CertificateType.X509);
```
ivosh: **. Nitpick** | Maintainability

CertificateRequestEntityBuilder.java:60 adds an unconditional `entity.setCertificateType(CertificateType.X509)` in build() (to satisfy the new NOT NULL). But the class Javadoc (lines 10-12) still says only 'The request format defaults to PKCS10' and 'tests override only the fields whose values drive the assertion under test' -- it now under-documents a second silent default, and there is no `withCertificateType(...)` override, so a test needing a non-X509 request type cannot use this builder. 

Either add a `withCertificateType` setter defaulting to X509 and mention it in the Javadoc, or trim the implementation details from JavaDoc.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceV2ITest.java
```diff
@@ -973,6 +969,8 @@ void testHandleFailedOrRejectedEvent() throws NotFoundException, NoSuchAlgorithm
         certificate.setState(CertificateState.REQUESTED);
         CertificateRequestEntity certificateRequest = new CertificateRequestEntity();
         certificateRequest.setContent("content");
+        certificateRequest.setCertificateType(CertificateType.X509);
```
ivosh: **. Nitpick** | Maintainability

ClientOperationServiceV2ITest.java:972 (and lines 1089/1441/1727/2221, plus V3CancelITest, V3RenewRevokeITest, EnumCollectionsColumnsBitmaskITest) each repeat the same `new CertificateRequestEntity()` + `setCertificateType(X509)` (+ often `setCertificateRequestFormat(PKCS10)` / `setContent(...)`) boilerplate now forced by the new NOT NULL columns. 

A CertificateRequestEntityBuilder already exists to centralize exactly these required-field defaults, but only 2 test classes use `aCertificateRequest()`; 10 test files construct the entity directly. This PR patches each raw call site individually, so the next NOT NULL column added will again require touching every one of them.

Recommendation: route raw `new CertificateRequestEntity()` fixtures through the builder so required-field defaults live in one place.

ivosh (changes requested):

src/test/java/com/otilm/core/integration/service/ClientOperationServiceV2ITest.java
klaraf755: Used the builder for the mentioned constructors of `CertificateRequestEntity`.

src/test/java/com/otilm/core/util/builders/CertificateRequestEntityBuilder.java
klaraf755: Added the default in the same way as for the format (updated description and added it as property of the builder).

src/main/resources/db/migration/V202607141200__align_certificate_schema_with_entities.sql
klaraf755: Used default values also for `archived` and `hybrid_certificate`.

src/main/resources/db/migration/V202607141200__align_certificate_schema_with_entities.sql
klaraf755: Added a test.

src/main/resources/db/migration/V202607141200__align_certificate_schema_with_entities.sql
klaraf755: Since this migration is needed for certificate registration to work, it is not just maintenance migration. I will split it for now, so it is not part of this PR.

ivosh (approved):

PR #1811, opened 2026-07-15
Reject duplicate certificate extension OIDs (RFC 5280)

### LINKED ISSUE #1778: Reject duplicate certificate extension OIDs across attribute mappings at projection time
## Description

**Follow-up to #1771: the projector rejects one multi-valued attribute mapped to an extension, but not the two other ways the same duplicate-OID situation arises.**

#1771 added a guard. A single multi-valued attribute mapped to one `EXTENSION` field is now rejected, because RFC 5280 requires each extension OID to be unique within `Extensions`.

Two paths are still unguarded:

- **(a)** two distinct attribute definitions whose `FieldMapping`s each declare an `ExtensionMappedField` with the same `extensionOid`.
- **(b)** a single `FieldMapping` that declares two `ExtensionMappedField`s with the same `extensionOid`.

Both append two `RequestedExtension` entries carrying the same OID to the request.

Nothing upstream stops them. `FieldMapping.fields` is a plain list with no uniqueness constraint. `mergeIssuanceDefinitions` dedups only by RDN claim (`ClientOperationServiceImpl.java:2044`). `attributeEngine.validateUpdateDataAttributes` validates attribute values, not mapping topology.

Downstream it fails badly. `X509RequestContentRenderer.toExtensions` calls BouncyCastle `ExtensionsGenerator.addExtension`, which throws an **unchecked** `IllegalArgumentException("extension <oid> already added")` (`X509RequestContentRenderer.java:84`). The caller catches only `IOException` (`ClientOperationServiceImpl.java:2122`), so the request dies with an uncontrolled runtime exception instead of a clean `ValidationException` — and the BouncyCastle message can reach the caller, brushing the "don't leak runtime details to the wire" rule.

This is a misconfiguration path, not a request-data surface: the extension OIDs are operator/connector-configured, never requester-supplied, and the gap predates #1771. But the error message #1771 added now advertises "an extension may appear only once (RFC 5280)", so the remaining paths should be closed — or explicitly tracked — rather than left implied-closed by that message.

## Definition of Done

- [ ] `project(...)` (or `X509RequestContentRenderer.toExtensions`) detects a duplicate extension OID across the assembled extension set — covering both (a) two definitions with the same OID and (b) one mapping with two same-OID extension fields — and throws a `ValidationException` naming the duplicated OID, before the BouncyCastle renderer is reached.
- [ ] No raw `IllegalArgumentException` (or other uncontrolled runtime exception) can escape the projection/render path for a duplicate-OID request.
- [ ] Projector unit tests cover both path (a) and path (b), asserting a `ValidationException` whose message contains the duplicated OID.
- [ ] Decide and document whether OID uniqueness should also be enforced earlier (attribute-definition save / RA-profile validation), so a misconfiguration is rejected before request time rather than at issuance; implement or leave a tracked note.
- [ ] New-code coverage ≥ 80 % (line + branch, SonarCloud parity).

ivosh:


lubomirw (commented):
Automated multi-reviewer triage — 4 findings (COMMENT). Copilot CLI: SHIP. Hygiene: clean. Context: duplicate-extension-OID rejection now exists at three layers — AttributeEngine (definition-time), CertificateRequestAttributeProjector (project-time, pre-existing), X509RequestContentRenderer (render-time). The renderer guard is a backstop; the projector dedups before it on the sole production caller.

src/test/java/com/otilm/core/util/X509RequestContentRendererTest.java
```diff
@@ -286,6 +286,41 @@ void throwsIoException_whenExtensionOidIsMalformed() {
                     .isInstanceOf(IOException.class)
                     .hasMessageContaining("Invalid or missing extension OID");
         }
+
+        @Test
+        void throwsIoException_whenTwoExtensionsShareTheSameOid() {
```
lubomirw, line 291: **[Medium]** | Testing

Both duplicate-rejection paths are tested, but the *allowed* case is not: an empty/null SAN list plus an explicit extension carrying `2.5.29.17` is legal and must render. The guard hinges on `seenOids.add(subjectAlternativeName)` sitting inside `if (!sans.isEmpty())` — a future refactor hoisting that `add` out of the conditional would silently reject valid requests with no failing test.

_Suggested:_ add a test — `X509RequestContent` with null/empty `subjectAltNames` + a `RequestedExtension` for `2.5.29.17` renders without `IOException` and yields the SAN extension.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -1251,6 +1246,20 @@ private static void validateFieldMapping(DataAttributeV3 attribute, String conne
...
+        rejectDuplicateExtensionOids(attribute, fieldMapping, connectorUuidStr);
+    }
+
+    /**
+     * Rejects a single mapping that declares the same extension OID on more than one EXTENSION field.
+     */
+    private static void rejectDuplicateExtensionOids(DataAttributeV3 attribute, FieldMapping fieldMapping, String connectorUuidStr) throws AttributeException {
+        Set<String> seenExtensionOids = new HashSet<>();
+        for (MappedField field : fieldMapping.getFields()) {
+            if (field instanceof ExtensionMappedField ext && !seenExtensionOids.add(ext.getExtensionOid()))
```
lubomirw, line 1269: **[Low]** | Maintainability

`rejectDuplicateExtensionOids` dedups only EXTENSION-vs-EXTENSION within a single mapping. It does not catch a SAN field colliding with an explicit `subjectAltName`-OID field, nor the same OID across two definitions — those are deferred to project-time (`CertificateRequestAttributeProjector`). Defensible (at definition time you cannot know both fields will be populated), but the asymmetry with the projector is implicit.

_Suggested:_ tighten the Javadoc to state the scope ("single mapping, extension-vs-extension only"); note SAN/cross-definition collisions are intentionally deferred to request time.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
```diff
@@ -66,19 +67,22 @@ public static X500Principal toX500Principal(X509RequestContent x509) throws IOEx
...
                 names[i] = toGeneralName(sans.get(i));
             }
+            seenOids.add(Extension.subjectAlternativeName.getId());
             gen.addExtension(Extension.subjectAlternativeName, isSubjectEmpty(x509), new GeneralNames(names));
         }
 
         List<RequestedExtension> extensions = x509.getExtensions() == null ? List.of() : x509.getExtensions();
         for (RequestedExtension ext : extensions) {
             ASN1ObjectIdentifier oid = parseOid(ext.getOid());
+            rejectDuplicateOid(seenOids, oid);
```
lubomirw, line 90: **[Low]** | Maintainability

The sole production caller (`ClientOperationServiceImpl.toExtensions`) runs `CertificateRequestAttributeProjector.project()` first, which already rejects duplicate OIDs and the SAN/explicit collision as `ValidationException`. So this render-time guard is a backstop, unexercised on the current production path — reasonable defense-in-depth for a public kernel method, but the three-layer redundancy should be explicit.

_Suggested:_ add a one-line comment on `toExtensions` noting the projector normally dedups upstream, so readers do not treat this as the primary enforcement point.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
```diff
@@ -89,6 +93,16 @@ public static Extensions toExtensions(X509RequestContent x509) throws IOExceptio
         return gen.generate();
     }
 
+    /**
+     * Rejects a second occurrence of an extension OID before it reaches {@code ExtensionsGenerator.addExtension}.
+     */
+    private static void rejectDuplicateOid(Set<String> seenOids, ASN1ObjectIdentifier oid) throws IOException {
+        if (!seenOids.add(oid.getId())) {
+            throw new IOException("Duplicate certificate extension OID " + oid.getId()
```
lubomirw, line 106: **[Nit]** | Maintainability

The three layers phrase the same RFC 5280 violation differently ("declares … more than once" / "is mapped by more than one attribute field" / "Duplicate certificate extension OID"). Tests assert on substrings, so wording drift could break assertions and hurts log/UX correlation.

_Suggested:_ optional — extract a shared constant for the "(RFC 5280) an extension may appear only once" suffix.

src/test/java/com/otilm/core/util/X509RequestContentRendererTest.java
ivosh: Fixed in e7bf4419. Added `rendersSanExtension_whenExplicitSubjectAltNameOidHasNoCompetingSanList` to the ToExtensions nested class: null SAN list + an explicit extension mapped to 2.5.29.17 renders without IOException and yields the subjectAltName extension. This pins the invariant so hoisting the `seenOids.add` out of the `!sans.isEmpty()` guard would now fail a test.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
ivosh: Fixed in e7bf4419. Expanded the Javadoc to state the scope explicitly — extension-vs-extension within a single FieldMapping — and to note that SAN/explicit and cross-definition collisions are intentionally deferred to request time in CertificateRequestAttributeProjector, because at definition time we cannot know whether both fields will actually be populated.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
ivosh: Fixed in e7bf4419. Added a Javadoc paragraph on toExtensions noting the sole production caller runs the projector first (which already dedups), so this render-time guard is defense-in-depth for the public kernel method rather than the primary enforcement point.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
ivosh: Left as-is (you flagged it optional). The tests assert on the OID substring (`hasMessageContaining(oid)`), not on the RFC-5280 suffix, so wording drift won't break assertions. A shared constant would have to live somewhere both AttributeEngine (attribute engine) and X509RequestContentRenderer (util) can reach, coupling two otherwise-unrelated classes; the three messages are also deliberately phrased for their distinct layers (definition-time / project-time / render-time). Happy to extract one if you'd prefer the consistency.

lubomirw (approved):

PR #1814, opened 2026-07-15
Fix issuing a preregistered certificate using existing CSR

### LINKED ISSUE #1815: Certificates - Registration - Issue Preregistered: External CSR attach fails
### Description

Issuing a preregistered certificate with an externally supplied CSR does not work. Completing a pre-registered (`REGISTERED`) placeholder certificate by attaching an operator-supplied CSR and then issuing it against the RA profile/connector fails.

### Steps to Reproduce

1. Pre-register a certificate (create a `REGISTERED` placeholder without an initial CSR/key material).
2. Attach an externally supplied CSR to complete the registration.
3. Attempt to issue the certificate against the RA profile/connector.

### Expected Behavior

- The certificate should be issued successfully using the externally supplied CSR.

### Actual Behavior

- Issuance of the preregistered certificate fails.

### Severity

Major

### Module

None

### Screenshot

_No response_

### Logs

_No response_

### Environment

_No response_

klaraf755:
Closes #1815

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
@@ -1180,7 +1180,7 @@ public class CertificateServiceImpl implements CertificateExternalService, Certi
         try {
             request = CertificateRequestUtils.createCertificateRequest(issueRequest.getRequest(), issueRequest.getFormat());
         } catch (IllegalArgumentException e) {
-            throw new CertificateRequestException("Certificate signing request is not valid Base64", e);
+            throw new CertificateRequestException("Error when creating a Certificate Request from provided content " + issueRequest.getRequest(), e);
         }
 
         // Attach the operator-supplied CSR to the placeholder; the registration identity already on the row
--- a/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
@@ -1203,6 +1203,7 @@ public class CertificateServiceImpl implements CertificateExternalService, Certi
         CertificateRequestEntity certificateRequestEntity = certificateRequestRepository.findByFingerprint(fingerprint).orElse(null);
         if (certificateRequestEntity == null) {
             certificateRequestEntity = certificate.prepareCertificateRequest(issueRequest.getFormat());
+            certificateRequestEntity.setCertificateType(CertificateType.X509);
             certificateRequestEntity.setFingerprint(fingerprint);
             certificateRequestEntity.setContent(Base64.getEncoder().encodeToString(normalizedDer));
             setCertificateRequestEntitySignatureAlgorithms(request, certificateRequestEntity);
--- a/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
@@ -1176,24 +1176,33 @@ public class CertificateServiceImpl implements CertificateExternalService, Certi
             throw new ValidationException("A certificate signing request can only be attached to a REGISTERED certificate. Certificate: %s".formatted(certificate.toStringShort()));
         }
 
-        byte[] decodedCsr;
+        CertificateRequest request;
         try {
-            decodedCsr = Base64.getDecoder().decode(issueRequest.getRequest());
+            request = CertificateRequestUtils.createCertificateRequest(issueRequest.getRequest(), issueRequest.getFormat());
         } catch (IllegalArgumentException e) {
             throw new CertificateRequestException("Certificate signing request is not valid Base64", e);
         }
-        CertificateRequest request = CertificateRequestUtils.createCertificateRequest(decodedCsr, issueRequest.getFormat());
 
         // Attach the operator-supplied CSR to the placeholder; the registration identity already on the row
         // (subject DN / SAN) is intentionally left untouched here — the issued certificate's identity is
         // written from the CA response at issuance. Get-or-create by fingerprint, mirroring the canonical
         // CSR-attach path: an identical CSR already stored is shared, not duplicated.
-        final String fingerprint = CertificateUtil.getThumbprint(decodedCsr);
+        //
+        // Fingerprint and stored content are derived from request.getEncoded() — the normalized DER that
+        // createCertificateRequest above actually parsed — not the raw operator-supplied string. The
+        // operator-supplied CSR is not guaranteed to be plain base64(DER): PEM armor and double base64
+        // encoding are common client mistakes, and createCertificateRequest already normalizes those.
+        // Storing the raw string here would silently persist the un-normalized (and possibly
+        // PEM-wrapped) encoding, which is later replayed verbatim to the connector on register-bound
+        // issue (AuthorityProviderV3Adapter.issueRegistered), producing a request the connector cannot
+        // parse as DER.
+        byte[] normalizedDer = request.getEncoded();
+        final String fingerprint = CertificateUtil.getThumbprint(normalizedDer);
         CertificateRequestEntity certificateRequestEntity = certificateRequestRepository.findByFingerprint(fingerprint).orElse(null);
         if (certificateRequestEntity == null) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1176,24 +1176,33 @@ public void addCertificateRequestToExisting(UUID certificateUuid, ClientCertific
             throw new ValidationException("A certificate signing request can only be attached to a REGISTERED certificate. Certificate: %s".formatted(certificate.toStringShort()));
         }
 
-        byte[] decodedCsr;
+        CertificateRequest request;
         try {
-            decodedCsr = Base64.getDecoder().decode(issueRequest.getRequest());
+            request = CertificateRequestUtils.createCertificateRequest(issueRequest.getRequest(), issueRequest.getFormat());
         } catch (IllegalArgumentException e) {
             throw new CertificateRequestException("Certificate signing request is not valid Base64", e);
```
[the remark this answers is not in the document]
klaraf755: Updated the error message.

src/test/java/com/otilm/core/integration/service/CertificateServiceITest.java
```diff
@@ -2026,6 +2026,68 @@ void rejectsNonRegisteredCertificate() {
...
+            assertThat(stored.getContent()).isEqualTo(Base64.getEncoder().encodeToString(expectedDer));
+        }
+
+        @Test
+        void attachesDoubleBase64EncodedCsr_normalizesToPlainDerBeforeStorage() throws Exception {
+            // given — a REGISTERED placeholder and the same CSR, but base64-encoded twice (another
+            // common client mistake)
+            Certificate registered = certificateRepository.save(
+                    aCertificate().withRaProfile(raProfile).withState(CertificateState.REGISTERED).build());
+            String doubleEncodedCsr = Base64.getEncoder().encodeToString(SAMPLE_PKCS10.getBytes());
```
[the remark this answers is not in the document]
klaraf755: Used encoding.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1176,24 +1176,33 @@ public void addCertificateRequestToExisting(UUID certificateUuid, ClientCertific
...
         }
 
-        byte[] decodedCsr;
+        CertificateRequest request;
         try {
-            decodedCsr = Base64.getDecoder().decode(issueRequest.getRequest());
+            request = CertificateRequestUtils.createCertificateRequest(issueRequest.getRequest(), issueRequest.getFormat());
         } catch (IllegalArgumentException e) {
-            throw new CertificateRequestException("Certificate signing request is not valid Base64", e);
+            throw new CertificateRequestException("Error when creating a Certificate Request from provided content " + issueRequest.getRequest(), e);
```
lubomirw: **[Medium]** | Security

The wrapped exception concatenates the full operator-supplied CSR (`+ issueRequest.getRequest()`). `ExceptionHandlingAdvice.handleCertificateRequestException` logs `getMessage()` at ERROR and returns it verbatim in the 400 body — so the entire unbounded, attacker-controlled request is echoed to the client and the log (multiline log-injection, oversized response). Violates the repo `Dont leak runtime details to the wire` rule. Also: the catch only handles `IllegalArgumentException`; a valid-base64/invalid-DER input throws `CertificateRequestException` from the ctor and bypasses this wrap, giving an inconsistent terse message.

_Suggested:_ Use a constant, content-free message identifying the cert by UUID (keep the original as cause); handle both exception types consistently.

<sub>flagged by Copilot CLI, guided-pr-review, superpowers</sub>

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1176,24 +1176,33 @@ public void addCertificateRequestToExisting(UUID certificateUuid, ClientCertific
...
+        //
+        // Fingerprint and stored content are derived from request.getEncoded() — the normalized DER that
+        // createCertificateRequest above actually parsed — not the raw operator-supplied string. The
+        // operator-supplied CSR is not guaranteed to be plain base64(DER): PEM armor and double base64
+        // encoding are common client mistakes, and createCertificateRequest already normalizes those.
+        // Storing the raw string here would silently persist the un-normalized (and possibly
+        // PEM-wrapped) encoding, which is later replayed verbatim to the connector on register-bound
+        // issue (AuthorityProviderV3Adapter.issueRegistered), producing a request the connector cannot
+        // parse as DER.
+        byte[] normalizedDer = request.getEncoded();
```
lubomirw, line 1201: **[Low]** | Maintainability

The adjacent comment says this get-or-create is `mirroring the canonical CSR-attach path`, but the canonical path (`submitCertificateRequest`) fingerprints single-base64-decoded bytes and stores the **raw** operator string — the opposite of this new normalized-DER derivation. So this path now diverges rather than mirrors, and normalization is applied to only one of the two CSR-storage paths.

_Suggested:_ Correct the comment; consider a shared helper so both paths derive content/fingerprint identically.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1176,24 +1176,33 @@ public void addCertificateRequestToExisting(UUID certificateUuid, ClientCertific
...
+        final String fingerprint = CertificateUtil.getThumbprint(normalizedDer);
         CertificateRequestEntity certificateRequestEntity = certificateRequestRepository.findByFingerprint(fingerprint).orElse(null);
         if (certificateRequestEntity == null) {
             certificateRequestEntity = certificate.prepareCertificateRequest(issueRequest.getFormat());
             certificateRequestEntity.setFingerprint(fingerprint);
-            certificateRequestEntity.setContent(issueRequest.getRequest());
+            certificateRequestEntity.setContent(Base64.getEncoder().encodeToString(normalizedDer));
```
lubomirw, line 1208: **[Low]** | Correctness

Normalization happens at attach time only. Certs that already had a CSR attached before this fix keep un-normalized (PEM / double-base64) content and are still replayed verbatim to the connector via `AuthorityProviderV3Adapter.issueRegistered`. No backfill/migration — already-affected placeholders stay broken until re-attached.

_Suggested:_ No code change strictly required — call out in release notes; consider a one-off migration or re-normalize-on-read if existing rows are in scope for #1815.

lubomirw (changes requested):
Automated multi-reviewer triage — 3 findings (REQUEST_CHANGES). **Merge blocker:** this PR does not compile until interfaces #776 is merged/released and the core `interfaces` dependency is bumped (CI Compile fails: `cannot find symbol: method copy()`). Reviewers: Copilot CLI, guided-pr-review, superpowers, hygiene (clean), + CI.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
klaraf755: The get-or-create-by-fingerprint pattern is the same as `submitCertificateRequest`, but you're right that the input normalization differs — and my comment didn't make that clear. `submitCertificateRequest'`s callers (`generateBase64EncodedCsr`) already normalize the CSR before calling it, so a plain `Base64.decode()` there is sufficient. This endpoint receives the operator's CSR directly with no such upstream normalization, so it has to call `createCertificateRequest` and derive from `getEncoded()` here. I've updated the comment to spell that out instead of just saying "mirroring."

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
klaraf755: Since this is a new feature and the behaviour was broken until this fix, there is no existing data to be migrated.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
klaraf755: Added consistent exception with reference to the certificate UUID.

lubomirw (approved):

PR #1820, opened 2026-07-15
Allow platform-level pre-registration without connector support

### LINKED ISSUE #1816: Support platform-level pre-registration on connectors without registration capability
### Description

Pre-registration currently requires the RA Profile's connector to advertise the `CERTIFICATE_REGISTRATION` capability: `registerCertificate` calls the connector `/register` operation and otherwise rejects with *"Certificate registration is not supported by the authority…"*. This gates pre-registration on the connector so CA-side validation can happen while the operator is present.

This enhancement lets an operator pre-register and control a certificate purely at the platform level — skipping the connector `/register` call — when the connector does not support registration. The certificate reaches `REGISTERED`, holds the platform-owned challenge and issuance window, and completes at issue time through whichever connector the RA Profile uses.

Requirement to preserve: nothing may assume `REGISTERED` implies a CA-side end-entity exists.

### Use Case

An operator pre-registers a certificate (challenge + issuance window, controlled by the platform) on an RA Profile whose connector has no registration support. Today this is rejected; it should succeed as a platform-level pre-registration and later complete through the normal issue path.

### Definition of Done

- `registerCertificate` no longer rejects when the connector lacks `RegisterCapability` / the `CERTIFICATE_REGISTRATION` flag; instead it creates the `REGISTERED` placeholder plus its authorization at the platform level and skips the connector `/register` call.
- Flat identity (DN) is persisted on the placeholder and the Certificate Registered event fires, as on the connector-backed path.
- Issue-time completion works through the plain issue path (already connector-agnostic: `issueCertificateAction` only routes to the register-bound path for `RegisterCapability` adapters).
- No code assumes `REGISTERED` implies a CA-side end-entity exists.
- Integration test: pre-register then issue on an RA Profile whose connector does not advertise `CERTIFICATE_REGISTRATION`.

lubomirw:
Lets an operator pre-register and control a certificate purely at the platform level when the RA Profile's connector does not support registration. The placeholder reaches `REGISTERED` with no connector `/register` call and is completed later through the normal issue path.

- `registerCertificate` takes a platform-level branch — create the placeholder + challenge authorization, persist the subject DN, fire the Certificate Registered event — when the adapter is not a `RegisterCapability` or the authority does not advertise `CERTIFICATE_REGISTRATION`. Connector-backed registration is unchanged.
- A challenge store/save failure fails the placeholder (→ FAILED + best-effort authorization cleanup), mirroring the connector path.
- Nothing assumes `REGISTERED` implies a CA-side end-entity exists: issue-time routing already branches on `RegisterCapability`, so a platform-level placeholder completes through the plain issue path.

Tests: platform-level registration on a v2 authority and on a `RegisterCapability` authority without the flag (asserts no connector `register()` call), plus a platform-level register → issue-through-the-plain-path completion. Register ITest 60 green.

Closes #1816


src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -518,6 +522,38 @@ public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID au
         return response;
     }
 
+    /**
+     * Platform-level pre-registration for an authority whose connector does not support registration: the
+     * placeholder and its challenge authorization are created and owned entirely by the platform, with no connector
+     * {@code /register} call. It reaches {@code REGISTERED} directly and is completed later through the normal issue
+     * path (which routes to the register-bound path only for {@code RegisterCapability} adapters). Nothing here
+     * implies a CA-side end-entity exists.
```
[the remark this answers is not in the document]
lubomirw: Fixed in 15411f4e3.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -518,6 +522,38 @@ public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID au
...
+        try {
+            maybeCreateRegistrationAuthorization(certificate, request);
+        } catch (RuntimeException e) {
+            // A challenge store/save failure must fail the placeholder rather than orphan it in PENDING_REGISTRATION.
+            stateMachine.transition(certificate, CertificateState.FAILED, null, "Registration failed: " + e.getMessage());
+            deleteRegistrationAuthorizationBestEffort(certificate.getUuid());
+            throw e;
+        }
+        stateMachine.transition(certificate, CertificateState.REGISTERED);
+        logger.info("Certificate {} pre-registered at the platform level (authority connector does not support registration)", certificate.getUuid());
```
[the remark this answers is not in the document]
lubomirw: Fixed in 15411f4e3.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -518,6 +522,38 @@ public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID au
...
+                                                                   String effectiveSubjectDn) throws NotFoundException {
+        Certificate placeholder = certificateService.createRegistrationPlaceholder(raProfile, effectiveSubjectDn);
+        Certificate certificate = certificateRepository.findForPollingByUuid(placeholder.getUuid())
+                .orElseThrow(() -> new NotFoundException(Certificate.class, placeholder.getUuid()));
+        stateMachine.transition(certificate, CertificateState.PENDING_REGISTRATION);
+        try {
+            maybeCreateRegistrationAuthorization(certificate, request);
+        } catch (RuntimeException e) {
+            // A challenge store/save failure must fail the placeholder rather than orphan it in PENDING_REGISTRATION.
+            stateMachine.transition(certificate, CertificateState.FAILED, null, "Registration failed: " + e.getMessage());
```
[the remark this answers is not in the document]
lubomirw: Fixed in 15411f4e3.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -440,23 +440,27 @@ public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID au
...
+        // A connector that does not advertise registration (not a RegisterCapability, or without the opt-in
+        // CERTIFICATE_REGISTRATION flag) still supports platform-level pre-registration: the placeholder and its
+        // challenge authorization are created and controlled entirely by the platform, with no connector /register
+        // call, and are completed later through the normal issue path. Nothing here implies a CA-side end-entity
+        // exists. Otherwise the registration is connector-backed below (early CA-side validation while the operator
+        // is present).
+        if (!(adapter instanceof RegisterCapability registerCapability)
+                || !capabilityService.supports(authority, FeatureFlag.CERTIFICATE_REGISTRATION)) {
+            return registerPlatformLevel(raProfile, request, effectiveSubjectDn);
+        }
```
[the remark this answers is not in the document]
lubomirw: Deferred by decision, not fixed here. `listAvailableOperations` REGISTER support intentionally reflects *connector-backed* register (RegisterCapability + CERTIFICATE_REGISTRATION); platform-level pre-registration is always available and does not flow through that capability flag. Reconciling the two (whether the FE should surface a platform-level register action) is an FE-contract decision tracked as a follow-up, out of scope for this PR.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1401,11 +1438,13 @@ public void issueCertificateRejectedAction(final UUID certificateUuid) throws No
...
                 .orElseThrow(() -> new NotFoundException(Certificate.class, certificateUuid));
         // A pre-registered placeholder whose issuance approval was rejected is restored to REGISTERED so the holder
-        // can retry — keyed on the register->issue binding, the same discriminator issueCertificateAction uses to
-        // route the placeholder (a no-secret pre-registration has a binding but no challenge authorization). Any
-        // challenge authorization stays ACTIVE. Non-registered certs keep the terminal REJECTED behaviour.
+        // can retry, rather than terminally rejected. A placeholder is identified by its register->issue binding
+        // (connector-backed pre-registration) OR its challenge authorization — a platform-level pre-registration has
+        // no binding, so the authorization is its marker. Any challenge authorization stays ACTIVE. Non-registered
+        // certs keep the terminal REJECTED behaviour.
         if (certificate.getState() == CertificateState.PENDING_APPROVAL
```
ivosh, line 1458: **! Major** | Correctness

The restore-to-`REGISTERED` guard identifies a placeholder by *either* the register->issue binding row *or* the challenge authorization row. A connector-backed pre-registration always has a binding (written by `persistRegistrationBinding`, `:502`, regardless of challenge). But `registerPlatformLevel` never writes a binding (no connector call) and only writes an authorization when a secret was supplied. So a **platform-level, no-secret** placeholder has *neither* marker: on approval rejection the condition is false and it falls through to `handleFailedOrRejectedEvent(... REJECTED ...)` (`:1452`). `REJECTED` is not a claimable state for `issuePlainCertificateAction` (`:926` allows only `REQUESTED`/`PENDING_APPROVAL`/`REGISTERED`), so the placeholder is stuck terminally and cannot be retried in place — directly contradicting this method's own comment and the connector-backed sibling's behavior. This is realistic: any RA profile with an ISSUE approval profile hits it.

Fix: give the platform path a placeholder marker independent of the challenge. Either write a binding-equivalent row for platform-level pre-registrations, or add an explicit "platform-level registration" flag/row that `issueCertificateRejectedAction` (and the routing in `issueCertificateAction:897`) can key on. Then the no-secret case restores symmetrically.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
```diff
@@ -250,13 +250,100 @@ void asyncRegistrationStaysPendingAndSchedulesRegisterPoll() throws Exception {
...
+    void platformLevelRegistrationStoreFailureFailsPlaceholder() {
+        // Platform-level path (v2 authority): a challenge-store failure must fail the placeholder and leave no
+        // authorization, mirroring the connector-backed path's pre-acceptance cleanup.
+        when(adapterFactory.forAuthority(Mockito.any())).thenReturn(mock(AuthorityProviderAdapter.class));
+        doThrow(new IllegalStateException("challenge encryption failed"))
+                .when(registrationChallengeStore).store(Mockito.any(), Mockito.any());
+        ClientCertificateRegistrationDto request = registrationRequest();
+        request.setAuthorizationSecret(CHALLENGE);
+
+        Assertions.assertThrows(RuntimeException.class, () -> clientOperationService.registerCertificate(
```
ivosh: **. Nitpick** | Testing

ClientOperationServiceRegisterITest.java:340 platformLevelRegistrationStoreFailureFailsPlaceholder asserts assertThrows(RuntimeException.class, ...) then only two negative facts: authorizationRepository.count()==0 (343) and no cert in PENDING_REGISTRATION (344-346). The stated intent — 'fail the placeholder, not orphan it in PENDING_REGISTRATION' — is never positively verified: both assertions pass vacuously if registerPlatformLevel threw BEFORE creating the placeholder (e.g. from createRegistrationPlaceholder), so the catch block in registerPlatformLevel (ClientOperationServiceImpl.java:542-546) could be uncovered while the test stays green (cf. CLAUDE.md 'Tests that look green but cover nothing'). The sibling connectorFailureLeavesPlaceholderFailed (356-358) does this correctly: it asserts certs.size()==1 and state==FAILED. Fix: mirror the sibling — assert exactly one cert exists and its state is FAILED, and tighten assertThrows to IllegalStateException.class (the mock throws IllegalStateException at 335, and siblings at 354/368 assert the concrete type).

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -518,6 +522,39 @@ public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID au
         return response;
     }
 
+    /**
```
ivosh, line 521: **. Nitpick** | Documentation

ClientOperationServiceImpl.java:525-531 is a single 7-line paragraph with no lead-ins that fuses five concerns (when it applies + the two alternative conditions in one ~40-word em-dash sentence, then platform ownership, no /register call, reaches REGISTERED, completed via the issue path with a parenthetical routing caveat, and 'nothing implies a CA-side end-entity exists').

It also near-duplicates the inline guard comment at 453-458, which states the same conditions and the same 'nothing implies a CA-side end-entity' point right where the branch is taken.

Unfortunately the class's existing dense-comment norm is a bad precendent here. 
Per CLAUDE.md 'AI-written code' guidance: restructure the Javadoc into a one-line summary plus one concern per `<p>` with bold lead-ins (~20-word sentences), and thin the 453-458 guard comment to the one fact the call site needs (why this branch is taken) since the method Javadoc already carries the rationale.

Suggested: perform the same surgery on the whole file.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -440,23 +440,27 @@ public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID au
...
+
+        // A connector that does not advertise registration (not a RegisterCapability, or without the opt-in
+        // CERTIFICATE_REGISTRATION flag) still supports platform-level pre-registration: the placeholder and its
+        // challenge authorization are created and controlled entirely by the platform, with no connector /register
+        // call, and are completed later through the normal issue path. Nothing here implies a CA-side end-entity
+        // exists. Otherwise the registration is connector-backed below (early CA-side validation while the operator
+        // is present).
+        if (!(adapter instanceof RegisterCapability registerCapability)
+                || !capabilityService.supports(authority, FeatureFlag.CERTIFICATE_REGISTRATION)) {
+            return registerPlatformLevel(raProfile, request, effectiveSubjectDn);
```
ivosh, line 457: **~ Minor** | Documentation

ClientOperationServiceImpl.java:461 changes registerCertificate so a v2 authority, or a v3 authority without CERTIFICATE_REGISTRATION, now creates a platform-level REGISTERED placeholder instead of throwing ValidationException. 

The companion contract surface in the interfaces repo still documents the removed behavior and was not updated: OmniTrustILM/interfaces src/main/java/com/otilm/api/interfaces/core/client/v2/ClientOperationController.java:354 says registration is 'Only supported on v3 authorities advertising the CERTIFICATE_REGISTRATION feature flag', and line 361 declares @ApiResponse(responseCode = "422", description = "Authority does not support registration") — a response this branch no longer emits for those authorities.

The summary/description (lines 348, 350: 'Pre-register a certificate with the upstream CA' / 'Reserves a slot at the CA') also no longer hold for the platform-level path, which the PR explicitly states does NOT imply a CA-side end-entity. This is an interfaces->core lockstep the PR body does not mention. 

Fix: update the ClientOperationController.java @Operation description (drop the v3+flag-only restriction) and the 422 ApiResponse for the register endpoint in the interfaces repo to match the new platform-level behavior.

ivosh (changes requested):

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Fixed in 034cefb6d. registerPlatformLevel now writes a register->issue binding (empty meta), so a no-secret platform-level placeholder carries the same marker as a connector-backed one and restores to REGISTERED on approval rejection. To keep the binding from mis-routing issuance, issueCertificateAction now gates the register-bound path on the CERTIFICATE_REGISTRATION flag too, so a platform-level placeholder issues plainly. The restore keys on the binding again (single marker). Added approvalRejectedRestoresPlatformLevelNoSecretPlaceholder.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
lubomirw: Fixed in 034cefb6d. Mirrors connectorFailureLeavesPlaceholderFailed: asserts exactly one cert in FAILED and tightens to IllegalStateException, so the registerPlatformLevel catch is positively covered.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Fixed in 034cefb6d. Javadoc restructured into a one-line summary plus one <p> per concern with bold lead-ins; the gate comment thinned to the one fact the call site needs. Kept the whole-file surgery out of this PR's scope.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Agreed — the interfaces ClientOperationController register @Operation description and the 422 'Authority does not support registration' ApiResponse are now stale for the platform-level path. That's an interfaces-repo change; opening a companion interfaces PR to reword the @Operation (drop the v3+flag-only restriction) and the 422, since core consumes it in lockstep.

lubomirw: Companion interfaces PR: https://github.com/OmniTrustILM/interfaces/pull/777

ivosh (dismissed):

ivosh (dismissed):

ivosh (approved):

PR #1821, opened 2026-07-15
Project the resolved request-attribute set when issuing and registering certificates

### LINKED ISSUE #1812: Issue/pre-register projection ignores the configured request-attribute set
## Description
The certificate **issuance** and **pre-registration** build path projects certificate request content from a hardcoded seed plus connector attributes only. It ignores both the **RA-Profile static request-attribute set** and the **editable platform default request-attribute set** that operators configure. As a result, request attributes configured under **Settings → Request Attributes** or on an RA Profile's **Request Attributes** tab never shape an issued or pre-registered certificate.

**Evidence (branch `main`, repo `core`):**
- `IssuanceDefinitionResolver.resolve(...)` merges `CsrAttributes.csrAttributesAsDataAttributesV3()` (the fixed `CN/OU/O/L/ST/C` seed) with v3 connector definitions only — no RA-Profile static set, no editable platform default: `src/main/java/com/otilm/core/certificate/request/IssuanceDefinitionResolver.java:48-64`.
- This is the resolver the issue and register paths use: `ClientOperationServiceImpl.java:2364` (issue) and `:546` (register / pre-register `buildStructuredRegistrationContent`).
- The resolver that DOES honour the configured sets (static ⊕ connector ⊕ editable-default fallback), `RaProfileCertificateRequestAttributeServiceImpl.resolveIssueAttributeSet(...)`, is wired ONLY to external-CSR validation (`ProtocolRequestAttributeValidator.java:54`) and EST (`EstCsrAttrsServiceImpl.java:26`) — never to the issue/register projection.

**Design vs. implementation: implementation divergence, not a design flaw.** The design specifies a single resolution order — RA-Profile static set → connector set → platform default — feeding the build path (`request-attributes-design-and-plans/core-design/2026-06-01-request-attributes-design.md`, §4.3; Mode A "project values → content → build", §7; operations flow, §9). Phase 3 was to make the default set admin-editable and fold the static set into that resolution (design §14). The storage and a correct resolver were implemented, but the issue/register projection calls the wrong, seed-only resolver (`IssuanceDefinitionResolver`), so the specified resolution order is never applied on the build path.

**Proposed fix:** point the issue/register projection at a single resolver that applies the §4.3 order (RA-Profile static ⊕ connector ⊕ editable-default fallback, value-source bindings applied). This resolver is shared with the resolved-set endpoint tracked in the companion Core issue (no-endpoint gap).

_Companion issues: the missing resolved-set endpoint (Core) and the frontend rendering gap (fe-administrator) are filed separately and cross-linked in a comment._

## Steps to Reproduce
1. In **Settings → Request Attributes**, add a request attribute (e.g. `TEST Server FQDN` mapped to RDN `CN`) and save.
2. On an RA Profile, open the **Request Attributes** tab, add e.g. `TEST Department` mapped to RDN `OU`, and save.
3. Issue (or pre-register) a certificate through that RA Profile via the platform-key path, then inspect the projected request content / resulting certificate.

## Expected Behavior
The projected request content reflects the configured request-attribute set — RA-Profile static set ⊕ connector set ⊕ editable platform default (fallback), per design §4.3.

## Actual Behavior
Only the hardcoded `CN/OU/O/L/ST/C` seed plus connector-supplied attributes are projected. The configured RA-Profile static set and the editable platform default set are ignored on the issue and register paths.

## Severity
Major

## Module
Core

## Environment
ILM version: 2.19.0 (develop-02), Browser: n/a, OS: n/a, Deployment: Kubernetes (develop-02)

ivosh:
Resolves #1812.

Certificate issuance and registration now project the resolved request-attribute set (static + connector, per merge mode, narrowed to v3) rather than a fixed CN/OU/O/L/ST/C seed.

📋 **Release note / operator-migration details:** https://github.com/OmniTrustILM/core/pull/1821#issuecomment-4985730839


src/main/java/com/otilm/core/certificate/request/IssuanceDefinitionResolver.java
```diff
@@ -2,106 +2,71 @@
...
-            return defaults;
-        }
-        List<BaseAttribute> connectorAttrs = extendedAttributeService.listIssueCertificateAttributes(raProfile);
-        List<DataAttributeV3> connectorDefs = connectorAttrs.stream()
-                .filter(DataAttributeV3.class::isInstance)
-                .map(DataAttributeV3.class::cast)
-                .toList();
-        int droppedNonV3 = connectorAttrs.size() - connectorDefs.size();
+        Objects.requireNonNull(raProfile, "raProfile is required to resolve issuance definitions");
+        List<BaseAttribute> resolved = requestAttributeService.resolveIssueAttributeSet(raProfile);
```
lubomirw, line 45: **[Medium]** | Correctness

Re-pointing the resolver to `resolveIssueAttributeSet` drops the old always-on merge of the hardcoded CN/OU/O/L/ST/C seed over connector v3 attrs. Profiles with **no stored static set** no longer get those RDNs projected. The changed integration test (`projectsConnectorSanAndExtensionIntoStoredCsr_...` now has to add `saveStaticSet(..., MERGE, null)` to keep CN) confirms the implicit seed is gone.

_Suggested:_ Intended per #1812 — confirm and document operator migration for existing profiles, or gate the legacy auto-merge behind a compatibility flag.

<sub>flagged by Copilot CLI, corroborated by the test diff</sub>

src/main/java/com/otilm/core/certificate/request/IssuanceDefinitionResolver.java
```diff
@@ -2,106 +2,71 @@
...
+        List<DataAttributeV3> definitions = onlyV3(resolved);
```
lubomirw, line 46: **[Low]** | Correctness

The old resolver dropped any default whose RDN was already claimed by a connector def (with OID normalization). `RequestAttributeSetResolver.mergeUnion` dedups **only by uuid/name** (`mergeKey`), never by mapped RDN — so a connector def and a static/default def mapping the same RDN (both → CN) both survive into the projection.

_Suggested:_ Confirm `CertificateRequestContentValidator`/the projector tolerates two definitions on one RDN without emitting a conflicting subject RDN; if not, restore RDN-level collision handling in the shared resolver.

<sub>flagged by guided-pr-review (verified against RequestAttributeSetResolver.java:71)</sub>

src/main/java/com/otilm/core/certificate/request/IssuanceDefinitionResolver.java
```diff
@@ -2,106 +2,71 @@
...
-     * null {@code fieldMapping} or a mapping with null {@code fields}.
-     */
-    private static Stream<RdnMappedField> rdnFields(DataAttributeV3 def) {
-        FieldMapping fm = def.getFieldMapping();
-        if (fm == null || fm.getFields() == null) {
-            return Stream.empty();
+        if (definitions.isEmpty() && !resolved.isEmpty()) {
+            logger.warn("Resolved request-attribute set for RA profile {} carries no v3 definitions; falling back to the platform default set",
+                    raProfile.getName());
+            definitions = onlyV3(requestAttributeService.getDefaultSet(raProfile));
```
lubomirw, line 55: **[Low]** | Maintainability

"Fall back to the editable default set" now lives in two places with different triggers: the service falls back when the merged set is empty; the resolver falls back separately (here) when the merged set is non-empty but carries no v3 defs.

_Suggested:_ Consider consolidating into a single v3-projectable resolve variant on the service so the resolver has one fallback contract to reason about.

src/test/java/com/otilm/core/certificate/request/IssuanceDefinitionResolverTest.java
```diff
@@ -0,0 +1,86 @@
...
+        doReturn(List.of(legacy)).when(requestAttributeService).resolveIssueAttributeSet(raProfile);
+        DataAttributeV3 fqdn = aMappedDataAttribute().withName("fqdn").mappingRdn("CN").build();
+        InfoAttributeV2 legacyDefault = new InfoAttributeV2();
+        legacyDefault.setName("legacy-default");
+        doReturn(List.of(fqdn, legacyDefault)).when(requestAttributeService).getDefaultSet(raProfile);
+
+        // when
+        List<DataAttributeV3> resolved = resolver.resolve(raProfile);
+
+        // then — the platform default set shapes the projection instead of resolving empty (design §13 edge-case 1)
```
lubomirw: **[Low]** | Maintainability

The `(design §13 edge-case 1)` parenthetical is a planning-workspace artifact with no meaning in shipped code (per CLAUDE.md: no comments referring to plan/PR artifacts).

_Suggested:_ Drop the parenthetical, keep the rationale: `// then — the platform default set shapes the projection instead of resolving empty`.

<sub>flagged by pr-hygiene (high confidence)</sub>

lubomirw (commented):
Automated multi-reviewer triage — 5 findings (COMMENT). 4 inline, 1 below (off-diff).

**Off-diff finding:**
- `IssuanceDefinitionResolverTest.java` — **[Nit] Testing.** The empty-resolved-set branch (`resolveIssueAttributeSet` returns empty → the `!resolved.isEmpty()` guard skips the fallback → straight to ValidationException) is untested. _Suggested:_ add a case returning `List.of()` and assert ValidationException while `getDefaultSet(raProfile)` is never consulted (verifyNoInteractions).

src/test/java/com/otilm/core/certificate/request/IssuanceDefinitionResolverTest.java
ivosh: Fixed in 153e2d2e. Dropped the parenthetical, kept the rationale: `// then — the platform default set shapes the projection instead of resolving empty`.

src/main/java/com/otilm/core/certificate/request/IssuanceDefinitionResolver.java
ivosh: **Intended per #1812, and no migration is needed.**

The projection now follows the *resolved* request-attribute set instead of a hardcoded seed — that is the point of #1812.

Profiles keep the RDNs in every case but one. `DefaultRequestAttributeSet.resolve(null)` returns the CN/OU/O/L/ST/C seed, so the code default already *is* the seed. When a profile has no static set and the connector supplies no v3 defs, the merged set is empty and the service falls back to that seed — RDNs preserved.

The only changed case: the connector supplies some v3 defs but not CN, there is no static set, and mode is MERGE. There the implicit CN seed is gone, by design.

Migration story for an affected operator: add CN to that profile's static set with MERGE, or edit the platform default set. I'll capture this in the PR/release notes.

src/main/java/com/otilm/core/certificate/request/IssuanceDefinitionResolver.java
ivosh: **The projector tolerates two definitions mapping the same RDN, so no RDN-level collision handling is needed.**

`CertificateRequestAttributeProjector` appends one `RdnEntry` per value with no dedup — that is deliberate multi-valued-RDN support. There is no "conflicting subject RDN" error path, and multiple RDNs of the same type are valid.

The only once-only guards are for extension OIDs and the SAN/explicit-extension collision — neither is an RDN.

`mergeKey` dedups on definition identity (uuid/name), not on the mapped RDN slot, on purpose: two distinct definitions are two distinct attributes even when they map to the same RDN. So both surviving into the projection is fine.

src/main/java/com/otilm/core/certificate/request/IssuanceDefinitionResolver.java
ivosh: **The two fallbacks serve different consumers, so keeping them split is deliberate.**

The service's empty-fallback is shared across every caller of `resolveIssueAttributeSet` — including EST `/csrattrs` and external-CSR validation, which need the non-v3 definitions too.

The v3-narrowing plus no-v3 fallback is issuance-projection-specific and lives in the resolver.

Consolidating into a single v3-projectable service method would leak a projection concern into the shared service API and require a second service method returning a narrowed type. So the split is intentional rather than accidental duplication.

ivosh:
@lubomirw Off-diff testing nit addressed in 153e2d2e: added `rejectsResolution_whenResolvedSetIsEmpty` to `IssuanceDefinitionResolverTest` — resolves to `List.of()`, asserts `ValidationException`, and `verify(..., never()).getDefaultSet(raProfile)` to prove the empty resolved set skips the default-set fallback.

Replies to the four inline threads are posted; the three [Correctness]/[Maintainability] ones are no-fix with rationale (intended per #1812 + no migration needed; projector tolerates multiple defs on one RDN; the two fallbacks serve different consumers). Left the threads unresolved for you.

ivosh:
## Release note — Certificate issuance/registration now projects the *resolved* request-attribute set (#1812)

**Structured CSR projection during issue and register now follows each RA profile's resolved request-attribute set, instead of always merging a fixed CN/OU/O/L/ST/C seed over the connector's attributes.**

The resolved set is the same one used by CSR validation: the profile's stored static set combined with the connector's dynamic set per the profile's merge mode, narrowed to the v3 definitions the projection consumes.

The merge mode defaults to **MERGE** — a profile with no explicitly stored mode is treated as MERGE. So the changed case below also covers profiles that never set a mode at all.

**This is a behavior change for a narrow set of profiles.** In almost all cases the projected subject is unchanged.

### When nothing changes
- A profile with a stored static set — the projection follows that set as configured.
- A profile whose connector supplies the CN definition — CN is projected from the connector.
- A profile with no static set whose connector supplies no v3 definitions — the resolved set is empty, so the platform default set applies, and that default set still contains CN/OU/O/L/ST/C.

### When the projected subject changes
- A profile with **no stored static set**, whose **connector supplies some v3 definitions but not CN**, running in **MERGE** mode (the default, including profiles with no mode stored). Previously the fixed CN seed was merged on top and CN was always projected. Now the projection reflects only the connector's definitions, so CN is no longer added implicitly.

### What operators do if affected
- Add CN (and any other needed RDNs) to that profile's static request-attribute set with merge mode MERGE, **or**
- Edit the platform default request-attribute set in Settings.

### No data migration
No data migration ships with this change. The platform default set already resolves to the CN/OU/O/L/ST/C seed when unset, so there is nothing to populate. A per-profile migration would re-inject the implicit seed this change deliberately removes, and there is no reliable way to detect which profiles depended on it.

lubomirw (approved):

PR #1824, opened 2026-07-15
Skip redundant metadata definition rewrites (discovery import performance bottleneck)

### LINKED ISSUE #1819: Discovery import is slow (because concurent locks on DB)
### Description

Discovery of ~12,6k certs from EJBCA takes about 10minutes.

But importing them from connector to core takes about 40min - it seems to be very slow. According slow queries log on DB server it is caused by concurrent locks. Some of them have to wait **~62s**.

```
2026-07-15 12:55:03.942 UTC [2131128] cmluser@cmldb LOG:  process 2131128 acquired ShareLock on transaction 46749058 after 38272.564 ms
...
2026-07-15 13:13:24.108 UTC [2206158] cmluser@cmldb LOG:  process 2206158 acquired ShareLock on transaction 46756541 after 62766.098 ms
```

### Steps to Reproduce

1. Create single shot discovery on EJBCA, select only ACTIVE certiticates (if you keep all preselect states you will find nothing, see bug https://github.com/OmniTrustILM/ejbca-ng-connector/issues/120)
2. Wait, wait, ... and wait for discovery to complete import. ;) 

### Expected Behavior

I would expect much faster import time. If EJBCA is capable exporting witin 10min we should be able to reach at least same time. I think core should serialize requests to not make such long lock waits.

### Actual Behavior

Slooow.

### Severity

Minor

### Module

Discovery

### Screenshot

[Screenshot at 2026-07-15 15-15-20.png](https://github.com/user-attachments/assets/440bfdf0-fc43-48d9-886e-245ae661c00d)
[Screenshot at 2026-07-15 15-16-23.png](https://github.com/user-attachments/assets/ad83629e-4daf-467e-840a-09a95542710a)

### Logs

[ejbca-ng-connector-deployment-2026-07-15-1434.log](https://github.com/user-attachments/files/30049373/ejbca-ng-connector-deployment-2026-07-15-1434.log)
[postgresql-ejbca-discovery-2026-07-15-1434.log](https://github.com/user-attachments/files/30049422/postgresql-ejbca-discovery-2026-07-15-1434.log)
[core-ejbca-discovery-2026-07-15-1434.log.gz](https://github.com/user-attachments/files/30049428/core-ejbca-discovery-2026-07-15-1434.log.gz)

### Environment

ILM version: 2.18.0
EJBCA server: https://ejbca.3key.company/
It is running on internal lab10.

ivosh:


src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -669,11 +670,28 @@ public AttributeDefinition updateMetadataAttributeDefinition(MetadataAttribute m
...
                 throw new AttributeException(String.format("Metadata attribute content type changed to %s while stored attribute definition have content type %s", metadataAttribute.getContentType().getLabel(), attributeDefinition.getContentType().getLabel()), metadataAttribute.getUuid(), metadataAttribute.getName(), metadataAttribute.getType(), connectorUuid == null ? null : connectorUuid.toString());
             }
+            // The same metadata definition gets (re-)sent for every repeated operation. The definition rarely changes,
+            // so re-saving it unchanged only issues a redundant UPDATE on this row and contends the transactions.
+            // When nothing changed, skip the write and mark the entity read-only: Hibernate cannot dirty-check the
+            // jsonb `definition`, so a merely-loaded definition would otherwise be re-UPDATEd on flush.
+            if (Objects.equals(attributeDefinition.getLabel(), label) && sameSerializedDefinition(attributeDefinition.getDefinition(), definitionCopy)) {
+                entityManager.unwrap(Session.class).setReadOnly(attributeDefinition, true);
+                return attributeDefinition;
```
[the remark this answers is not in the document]
ivosh: Fixed in 9c39f0e9. Added `updateMetadataAttributeDefinition_skipsUpdateWhenUnchangedButPersistsOnChange` to `AttributeEngineITest`: it registers a definition in its own committed transaction, then resubmits the identical definition in a fresh transaction with an explicit flush and asserts (via Hibernate entity-update statistics) that no UPDATE is issued for the read-only, unchanged row — and asserts a changed label still persists an UPDATE, so the skip can't silently swallow genuine changes.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -685,20 +703,26 @@ public AttributeDefinition updateMetadataAttributeDefinition(MetadataAttribute m
...
+     */
+    private static boolean sameSerializedDefinition(Object stored, Object candidate) {
+        try {
+            return ATTRIBUTES_OBJECT_MAPPER.writeValueAsString(stored)
+                    .equals(ATTRIBUTES_OBJECT_MAPPER.writeValueAsString(candidate));
+        } catch (JsonProcessingException e) {
+            // If either side cannot be rendered, fall back to writing the definition — correctness over the optimization.
+            return false;
+        }
+    }
```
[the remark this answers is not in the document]
ivosh: Declining this one. Both sides are serialized by the same `ATTRIBUTES_OBJECT_MAPPER`, so field-ordering insensitivity doesn't apply here. And `valueToTree` isn't cheaper — it serializes into a `TokenBuffer` and builds a full `JsonNode` tree, comparable-or-greater allocation than two Strings. This comparison runs once per call and is negligible against the DB round-trip (redundant UPDATE + row lock) it exists to avoid, which was the actual #1819 bottleneck. Keeping the current form as it's correct and simpler.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -669,11 +670,28 @@ public AttributeDefinition updateMetadataAttributeDefinition(MetadataAttribute m
...
                 throw new AttributeException(String.format("Metadata attribute content type changed to %s while stored attribute definition have content type %s", metadataAttribute.getContentType().getLabel(), attributeDefinition.getContentType().getLabel()), metadataAttribute.getUuid(), metadataAttribute.getName(), metadataAttribute.getType(), connectorUuid == null ? null : connectorUuid.toString());
             }
+            // The same metadata definition gets (re-)sent for every repeated operation. The definition rarely changes,
+            // so re-saving it unchanged only issues a redundant UPDATE on this row and contends the transactions.
+            // When nothing changed, skip the write and mark the entity read-only: Hibernate cannot dirty-check the
+            // jsonb `definition`, so a merely-loaded definition would otherwise be re-UPDATEd on flush.
+            if (Objects.equals(attributeDefinition.getLabel(), label) && sameSerializedDefinition(attributeDefinition.getDefinition(), definitionCopy)) {
+                entityManager.unwrap(Session.class).setReadOnly(attributeDefinition, true);
```
lubomirw: **[High]** | Correctness

`setReadOnly(entity, true)` is persistence-context-wide and sticky, and is never reset. The same managed `AttributeDefinition` is returned by identity on a repeat `findBy...` in one context; if the same definition is re-sent unchanged-then-changed within a single transaction (amplified by class `@Transactional` + `open-in-view=true`), the later `setLabel`/`save` runs on a read-only entity and the UPDATE is silently dropped at flush.

_Suggested:_ call `setReadOnly(attributeDefinition, false)` on the update path before `save()`, or `detach`/evict the unchanged entity instead of marking it read-only, or fix jsonb dirty-checking so an unchanged definition is never dirty and the skip can just return.

<sub>flagged by Copilot, superpowers, guided-pr-review</sub>

src/test/java/com/otilm/core/integration/attribute/AttributeEngineITest.java
```diff
@@ -978,6 +994,64 @@ void updateMetadataAttributeDefinition_doesNotMutateCallerContent() throws Attri
         Assertions.assertTrue(reloadedDefinition.getContent().isEmpty());
     }
 
+    @Test
+    void updateMetadataAttributeDefinition_skipsUpdateWhenUnchangedButPersistsOnChange() {
```
lubomirw, line 999: **[Medium]** | Testing

Every call runs in `REQUIRES_NEW`, so each `updateMetadata` gets a fresh persistence context. The test therefore cannot reach the same-context silent-drop path from the read-only finding, giving false confidence that genuine updates are not swallowed.

_Suggested:_ in one transaction, call `updateMetadata` unchanged then with a changed label, flush, reload in a separate transaction, and assert the change persisted (fails today, passes after the read-only fix).

<sub>flagged by superpowers, Copilot</sub>

src/test/java/com/otilm/core/integration/attribute/AttributeEngineITest.java
```diff
@@ -978,6 +994,64 @@ void updateMetadataAttributeDefinition_doesNotMutateCallerContent() throws Attri
...
+
+        // when: the label changes and the definition is resubmitted
+        statistics.clear();
+        inNewTransaction(() -> {
+            updateMetadata(attributeUuid, "Skip Fix Meta (changed)");
+            entityManager.flush();
+        });
+
+        // then: the change is persisted, proving the skip is not swallowing genuine updates
+        assertThat(statistics.getEntityUpdateCount())
```
lubomirw: **[Medium]** | Testing

`statistics.getEntityUpdateCount()` is global across all entities, so `isPositive()` can be satisfied by an unrelated update, and the change case never reads back the label to prove the definition itself was persisted.

_Suggested:_ scope to the entity via `getEntityStatistics(AttributeDefinition.class.getName())`, or reload the row and assert `label` equals the new value; assert an exact delta rather than `isZero`/`isPositive`.

<sub>flagged by superpowers, guided-pr-review</sub>

src/test/java/com/otilm/core/integration/attribute/AttributeEngineITest.java
```diff
@@ -978,6 +994,64 @@ void updateMetadataAttributeDefinition_doesNotMutateCallerContent() throws Attri
...
     }
 
+    @Test
+    void updateMetadataAttributeDefinition_skipsUpdateWhenUnchangedButPersistsOnChange() {
+        // given: a metadata definition already registered in its own committed transaction
+        UUID attributeUuid = UUID.randomUUID();
+        inNewTransaction(() -> updateMetadata(attributeUuid, "Skip Fix Meta"));
+
+        Statistics statistics = entityManagerFactory.unwrap(SessionFactory.class).getStatistics();
+        statistics.setStatisticsEnabled(true);
```
lubomirw, line 1006: **[Low]** | Testing

`setStatisticsEnabled(true)` mutates the shared `SessionFactory` statistics singleton and is never restored, leaking enabled state and overhead into later tests in the same JVM (flakiness risk).

_Suggested:_ capture the prior enabled state and restore it in an `@AfterEach`/finally.

<sub>flagged by superpowers, guided-pr-review</sub>

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -685,20 +703,26 @@ public AttributeDefinition updateMetadataAttributeDefinition(MetadataAttribute m
...
 
         return attributeDefinition;
     }
 
+    /**
+     * Compares definitions by their serialized form. The attribute model has no value-based equals on its nested types.
+     */
+    private static boolean sameSerializedDefinition(Object stored, Object candidate) {
+        try {
+            return ATTRIBUTES_OBJECT_MAPPER.writeValueAsString(stored)
```
lubomirw, line 714: **[Low]** | Performance

Serialize-to-compare is map-order sensitive (`ORDER_MAP_ENTRIES_BY_KEYS` is not set); map-bearing definitions can serialize to different strings and defeat the skip, forcing a redundant UPDATE. Safe (never a false skip) but weakens the intended savings.

_Suggested:_ compare canonical trees (`mapper.valueToTree(stored).equals(mapper.valueToTree(candidate))`) or use a dedicated mapper with ordered map entries.

<sub>flagged by superpowers, guided-pr-review</sub>

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -685,20 +703,26 @@ public AttributeDefinition updateMetadataAttributeDefinition(MetadataAttribute m
...
+                    .equals(ATTRIBUTES_OBJECT_MAPPER.writeValueAsString(candidate));
+        } catch (JsonProcessingException e) {
+            // If either side cannot be rendered, fall back to writing the definition — correctness over the optimization.
+            return false;
```
lubomirw, line 719: **[Nit]** | Error handling

`JsonProcessingException` is swallowed and returns `false` with no log; if serialization consistently fails, the optimization silently no-ops with no diagnostic signal.

_Suggested:_ log the exception at debug/warn before returning false.

lubomirw (changes requested):
Automated multi-reviewer triage — 6 findings (REQUEST_CHANGES). Copilot CLI, guided-pr-review, superpowers, and hygiene, deduped. Headline: sticky read-only can silently drop a later legitimate UPDATE to the same definition in one transaction (#1), and the new test cannot catch that path (#2).

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
ivosh: **Fixed in f6dda36c — at the root, via your third suggested option.** The sticky `setReadOnly` is gone entirely.

The real cause: the attribute model has no value-based `equals()`. Hibernate's snapshot dirty-check therefore saw every *loaded* definition as dirty (the snapshot deep copy never `equals()` the live instance) and re-UPDATEd the jsonb column on every flush.

The fix maps `AttributeDefinition.definition` with `@Mutability(Immutability.class)`. A merely-loaded definition is never dirty, so the unchanged path simply returns — no session-wide state, nothing to reset. A genuine change still persists, because the update path replaces the value (`setDefinition(newInstance)`), which the immutable mapping detects.

The contract — **mutate by replacement only, never in place** — is documented on the field. All five existing `setDefinition` call sites already comply; no in-place mutation exists in main or tests.

Side benefits: every other load path of `AttributeDefinition` stops emitting redundant UPDATEs too, and `updated_at` now only moves on real changes.

Your unchanged-then-changed-in-one-transaction scenario is covered by a new regression test (see the test thread).

src/test/java/com/otilm/core/integration/attribute/AttributeEngineITest.java
ivosh: **Fixed in f6dda36c.** Added `updateMetadataAttributeDefinition_persistsChangeAfterUnchangedResubmitInSameTransaction`: one transaction resubmits the definition unchanged, then changes the label on the same managed entity; a separate transaction reloads and asserts the change persisted. This is exactly the silent-drop scenario — it would fail against the sticky read-only approach and passes with the mapping-level fix.

src/test/java/com/otilm/core/integration/attribute/AttributeEngineITest.java
ivosh: **Fixed in f6dda36c.** The assertions are now scoped via `getEntityStatistics(AttributeDefinition.class.getName())`, use exact deltas (0 for unchanged, 1 for changed), and the change case reloads the row and asserts the new label value.

src/test/java/com/otilm/core/integration/attribute/AttributeEngineITest.java
ivosh: **Fixed in f6dda36c.** The prior enabled state is captured before the test and restored in a `finally` block.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
ivosh: **Declining — a metadata definition cannot contain a `Map`, so map-order sensitivity cannot materialize on this path.** I checked the model: `MetadataAttributeV3` and its `MetadataAttributeProperties` hold scalars and lists throughout, and `content` is stripped to an empty list on both compared sides. The only `Map<String, Serializable>` fields in the attribute model live on `RequestAttributeCallback`, which is never part of a metadata definition. Both sides also serialize through the same mapper instance, so the serialized forms are deterministic here and the skip cannot be defeated.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
ivosh: **Fixed in f6dda36c.** The fallback now logs the `JsonProcessingException` at debug before returning false.

lubomirw (approved):

PR #1828, opened 2026-07-16
Validate and persist certificate custom attributes when registering certificate

### LINKED ISSUE #1829: Client Operations - Certificate Registration: Custom attributes are not persisted for pre-registered certificates
### Description

When a certificate is **pre-registered** via `POST /v2/{authorityUuid}/raProfile/{raProfileUuid}/certificates/register`, any **Custom Attributes** supplied in the request (`customAttributes` field of the registration payload) are silently dropped: they are neither validated nor persisted against the resulting placeholder certificate.

This is inconsistent with the regular issuance/submit flow (`submitCertificateRequest` / `issueCertificate`), where custom attributes are explicitly validated via `AttributeEngine.validateCustomAttributesContent` and persisted via `AttributeEngine.updateObjectCustomAttributesContent` right after the certificate record is created. The registration flow (`ClientOperationServiceImpl.registerCertificate`) never calls either method, so a caller supplying `customAttributes` on the register request gets no error, but the values never show up on the certificate.

### Steps to Reproduce

1. Define a **Custom Attribute** associated with the **Certificate** resource.
2. Call the certificate pre-registration endpoint (`registerCertificate`) for an RA profile, supplying a value for that custom attribute in the `customAttributes` field of the request body.
3. Inspect the resulting placeholder certificate's custom attributes (e.g. via the certificate detail API/UI).

### Expected Behavior

- The submitted custom attribute value is validated against its definition.
- The submitted custom attribute value is persisted and visible on the pre-registered placeholder certificate, consistent with the regular certificate issuance flow.

### Actual Behavior

- The request succeeds and the certificate is pre-registered.
- The supplied `customAttributes` are silently ignored — no validation error, and no custom attribute is stored on the certificate.

### Severity

Minor

### Module

None

### Screenshot

_No response_

### Logs

_No response_

### Environment

_No response_

klaraf755:
Closes #1829


src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -465,6 +467,8 @@ public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID au
                 .orElseThrow(() -> new NotFoundException(Certificate.class, placeholder.getUuid()));
         stateMachine.transition(certificate, CertificateState.PENDING_REGISTRATION);
 
+        attributeEngine.updateObjectCustomAttributesContent(Resource.CERTIFICATE, certificate.getUuid(), request.getCustomAttributes());
+
```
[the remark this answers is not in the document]
klaraf755: Moved to the try/catch block.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -425,10 +425,12 @@ public ClientCertificateDataResponseDto issueCertificate(final SecuredParentUUID
     @Transactional(propagation = Propagation.NOT_SUPPORTED)
     @ExternalAuthorization(resource = Resource.RA_PROFILE, action = ResourceAction.DETAIL, parentResource = Resource.AUTHORITY, parentAction = ResourceAction.DETAIL)
     public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID authorityUuid, SecuredUUID raProfileUuid,
-                                                                ClientCertificateRegistrationDto request) throws NotFoundException, ConnectorException {
+                                                                ClientCertificateRegistrationDto request) throws NotFoundException, ConnectorException, AttributeException {
         if (request == null) {
             throw new ValidationException("A certificate registration request is required.");
         }
+
+        attributeEngine.validateCustomAttributesContent(Resource.CERTIFICATE, request.getCustomAttributes());
```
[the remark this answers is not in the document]
klaraf755: Added persistence of custom attributes there.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
```diff
@@ -710,6 +711,62 @@ void registerPersistsConnectorMetadata() throws Exception {
...
+        ClientCertificateRegistrationDto request = registrationRequest();
+        List<RequestAttribute> customAttributes = List.of(mock(RequestAttribute.class));
+        request.setCustomAttributes(customAttributes);
+
+        ClientCertificateDataResponseDto response = clientOperationService.registerCertificate(
+                authorityParent, securedRaProfile, request);
+
+        UUID certUuid = UUID.fromString(response.getUuid());
+        verify(attributeEngine).validateCustomAttributesContent(Resource.CERTIFICATE, customAttributes);
+        verify(attributeEngine).updateObjectCustomAttributesContent(Resource.CERTIFICATE, certUuid, customAttributes);
```
lubomirw, line 735: **[Low]** | Testing

This test asserts custom attributes are *actually persisted against the placeholder*, but `attributeEngine` is a full `@MockitoBean` — `updateObjectCustomAttributesContent` is a no-op, so the test only verifies the call was made with the right args, not that any content row was written.

_Suggested:_ Soften the comment to say the service delegates validate/persist to the engine (interaction test), or add a separate ITest exercising the real `AttributeEngine` (definition + relation fixture) and asserting `getObjectCustomAttributesContent` returns the persisted content.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
```diff
@@ -710,6 +711,62 @@ void registerPersistsConnectorMetadata() throws Exception {
         verify(attributeEngine).updateMetadataAttributes(Mockito.anyList(), Mockito.any());
     }
 
+    @Test
+    void registerValidatesAndPersistsCustomAttributes() throws Exception {
```
lubomirw, line 721: **[Low]** | Testing

The three new tests cover happy-path validate+persist and post-placeholder persistence failure on both paths, but none exercises the up-front `validateCustomAttributesContent` rejection (invalid content thrown *before* the placeholder exists).

_Suggested:_ Add a test that stubs `validateCustomAttributesContent` to throw and asserts the request fails with `AttributeException` and no `Certificate` persisted (`certificateRepository.findAll()` empty).

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: **[Low]** | Correctness

`submitCertificateRequest` gates custom-attribute validate/persist behind `!AuthHelper.isLoggedProtocolUser()` (ACME/SCEP/CMP protocol users deliberately skip custom attributes); `registerCertificate` validates and persists unconditionally. No protocol-context caller of the register endpoint was found, so this is likely safe — but it is an intentional inconsistency with the flow this PR mirrors.

_Suggested:_ Confirm the register endpoint is never reachable in a logged-protocol-user context; if it can be, mirror the `!isLoggedProtocolUser()` gate. Otherwise add a one-line comment explaining why register does not need the guard.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -467,6 +469,11 @@ public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID au
 
         AdapterOperationResult result;
         try {
+            // Custom attributes were already validated up front (before the placeholder existed); persisting them
+            // here can still fail (e.g. a definition removed concurrently), and no upstream work is in flight yet,
+            // so a failure here must fail the placeholder via the catch below rather than orphan it silently
+            // attribute-less in PENDING_REGISTRATION.
+            attributeEngine.updateObjectCustomAttributesContent(Resource.CERTIFICATE, certificate.getUuid(), request.getCustomAttributes());
```
lubomirw: **[Nit]** | Performance

`updateObjectCustomAttributesContent` is invoked on every registration (here and again at line 556, platform-level path) regardless of whether `request.getCustomAttributes()` is null/empty; with an empty list it still opens a transaction and runs a delete query, adding overhead to every register carrying no custom attributes.

_Suggested:_ Guard the persistence call on a non-empty custom-attribute list (the up-front validate can stay unconditional). Confirm this matches the submit/issue flow convention first, to keep paths consistent.

lubomirw (commented):
Automated multi-reviewer triage — 5 findings (COMMENT). Selected subset: test coverage/accuracy, protocol-user gating, empty-list persistence overhead, and import order. Higher-severity items (interfaces-snapshot build gate, OPA-in-transaction) were reviewed but not included in this review at the author's request.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
```diff
@@ -8,6 +8,7 @@
 import com.otilm.api.model.client.connector.v2.ConnectorVersion;
 import com.otilm.api.model.client.connector.v2.FeatureFlag;
 import com.otilm.api.model.core.certificate.CertificateState;
+import com.otilm.api.model.core.auth.Resource;
```
lubomirw, line 10: **[Nit]** | Maintainability

The added import `com.otilm.api.model.core.auth.Resource` is inserted between certificate-model imports rather than in sorted order, which some checkstyle/import-order configs flag.

_Suggested:_ Reorder the import to match the project's import-ordering convention.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Added guards for non-empty custom attributes list.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Current protocols do not use the register operation, but added the check anyway, in case a protocol will need it in the future.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
klaraf755: Added a test (in a new class, since this test needs not mocked `AttributeEngine`).

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
klaraf755: Added a test (in a new class, since this test needs not mocked `AttributeEngine`).

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
klaraf755: Fixed order.

lubomirw (approved):

PR #1831, opened 2026-07-16
Resolve the request-attribute set over the CSR-attributes endpoint

### LINKED ISSUE #1813: No API returns the resolved set for an RA Profile
## Description
No REST endpoint returns the **resolved request-attribute set** (RA-Profile static set ⊕ connector set ⊕ editable platform default fallback, with value-source bindings applied) for a given RA Profile. A client that wants to render the fields the platform will actually project/validate has nothing to call. Consequently the issue/pre-register form cannot display configured request attributes (see the companion frontend issue), and operators cannot preview the effective set anywhere in the UI.

**Evidence (branch `main`, repo `core`):**
- `GET /v1/certificates/csr/attributes` returns the fixed seed only and takes no RA-Profile/authority parameter: `CertificateController.java:169-170` → `CertificateServiceImpl.getCsrGenerationAttributes():1577-1581` → `CsrAttributes.csrAttributes()`.
- `GET /v2/operations/authorities/{authorityUuid}/raProfiles/{raProfileUuid}/attributes/issue` returns connector-supplied issue attributes only: `ClientOperationControllerImpl.java:91-94` → `ExtendedAttributeServiceImpl.listIssueCertificateAttributes:46-54` (`adapter.listIssueAttributes(...)`; throws if no connector). No merge with the static/default set.
- `GET /v1/authorities/.../raProfiles/{uuid}` exposes the RA-Profile **static** set only (`RaProfileDto.certificateRequestAttributes`); `GET /v1/settings/platform` exposes the **editable default** in isolation. Neither is the resolved set.
- The resolvers that produce the resolved set (`RaProfileCertificateRequestAttributeServiceImpl.resolveIssueAttributeSet`, `IssuanceDefinitionResolver.resolve`) are consumed internally only; no controller returns their output.

**Design vs. implementation: implementation divergence.** The frontend implementation plan (`request-attributes-design-and-plans/fe-administrator/2026-06-01-request-attributes-fe.md`, Group 1) explicitly depended on Core Phase 1 surfacing "the resolved request-attribute set on the existing CSR-attributes endpoint (`getCsrAttributes` → descriptors now carry `fieldMapping`)" (line 21) and flagged confirming "whether the endpoint becomes RA-Profile-scoped" (lines 313-315). That backend contract change was never implemented — `getCsrGenerationAttributes()` still returns the fixed seed and takes no arguments.

**Proposed fix:** expose the resolved set for an RA Profile — either by making the issue attribute-listing (`.../attributes/issue`) return the resolved definitions carrying `fieldMapping`, or by adding a dedicated resolved-set endpoint keyed by `(authorityUuid, raProfileUuid, operation)`. Reuse the same resolver that the projection path should adopt (companion Core projection issue).

_Companion issues: the projection-side gap (Core) and the frontend rendering gap (fe-administrator) are cross-linked in a comment._

## Steps to Reproduce
1. Configure a request-attribute set on an RA Profile (and/or under **Settings → Request Attributes**).
2. Call `GET /v1/certificates/csr/attributes`.
3. Call `GET /v2/operations/authorities/{authorityUuid}/raProfiles/{raProfileUuid}/attributes/issue`.

## Expected Behavior
An endpoint returns the resolved request-attribute set for the RA Profile, with definitions carrying `fieldMapping`, so a client can render exactly the fields the platform will project/validate.

## Actual Behavior
`csr/attributes` returns the fixed 6-attribute seed (no RA-Profile context); `attributes/issue` returns connector attributes only; the RA-Profile detail and platform settings expose their sets in isolation. No endpoint returns the resolved set.

## Severity
Major

## Module
Core

## Environment
ILM version: 2.19.0 (develop-02), Browser: n/a, OS: n/a, Deployment: Kubernetes (develop-02)

ivosh:
Extend GET /v1/certificates/csr/attributes with an optional raProfileUuid query parameter so one endpoint serves all three fe-administrator call sites:

- With raProfileUuid: return the resolved request-attribute set for the RA profile (RA-Profile static set merged with the connector set per the profile's merge mode, falling back to the editable platform default set), narrowed to the projectable v3 definitions via IssuanceDefinitionResolver. Object-level RA_PROFILE authorization; runs NOT_SUPPORTED and loads the authority->connector graph eagerly (findWithAuthorityByUuid) so the connector call does not hit a LazyInitializationException; disabled and unknown profiles surface NotFoundException, an empty projection surfaces ValidationException (HTTP 422).

- Without raProfileUuid: return the editable platform default set (RaProfileCertificateRequestAttributeService.getDefaultSet) instead of the hardcoded CsrAttributes.csrAttributes() seed. Behaviour is identical to the old seed until an operator edits the default set. Drop the now-dead csrAttributes() definition.

### DIFF
--- a/src/main/java/com/otilm/core/api/web/CertificateControllerImpl.java
+++ b/src/main/java/com/otilm/core/api/web/CertificateControllerImpl.java
@@ -165,11 +165,11 @@ public class CertificateControllerImpl implements CertificateController {
 
     @Override
     @AuditLogged(module = Module.CERTIFICATES, resource = Resource.ATTRIBUTE, name = "csr", affiliatedResource = Resource.CERTIFICATE, operation = Operation.LIST_ATTRIBUTES)
-    public List<BaseAttribute> getCsrGenerationAttributes(String raProfileUuid) throws NotFoundException, ConnectorException {
-        if (raProfileUuid == null || raProfileUuid.isBlank()) {
+    public List<BaseAttribute> getCsrGenerationAttributes(UUID raProfileUuid) throws NotFoundException, ConnectorException {
+        if (raProfileUuid == null) {
             return certificateService.getCsrGenerationAttributes();
         }
-        return certificateService.getCsrGenerationAttributes(SecuredUUID.fromString(raProfileUuid));
+        return certificateService.getCsrGenerationAttributes(SecuredUUID.fromUUID(raProfileUuid));
     }
 
     @Override
--- a/src/main/java/com/otilm/core/api/web/CertificateControllerImpl.java
+++ b/src/main/java/com/otilm/core/api/web/CertificateControllerImpl.java
@@ -164,8 +164,8 @@ public class CertificateControllerImpl implements CertificateController {
     }
 
     @Override
-    @AuditLogged(module = Module.CERTIFICATES, resource = Resource.ATTRIBUTE, name = "csr", affiliatedResource = Resource.CERTIFICATE, operation = Operation.LIST_ATTRIBUTES)
-    public List<BaseAttribute> getCsrGenerationAttributes(UUID raProfileUuid) throws NotFoundException, ConnectorException {
+    @AuditLogged(module = Module.CERTIFICATES, resource = Resource.ATTRIBUTE, name = "csr", affiliatedResource = Resource.RA_PROFILE, operation = Operation.LIST_ATTRIBUTES)
+    public List<BaseAttribute> getCsrGenerationAttributes(@LogResource(uuid = true, affiliated = true) UUID raProfileUuid) throws NotFoundException, ConnectorException {
         if (raProfileUuid == null) {
             return certificateService.getCsrGenerationAttributes();
         }
--- a/src/main/java/com/otilm/core/api/web/CertificateControllerImpl.java
+++ b/src/main/java/com/otilm/core/api/web/CertificateControllerImpl.java
@@ -165,8 +165,11 @@ public class CertificateControllerImpl implements CertificateController {
 
     @Override
     @AuditLogged(module = Module.CERTIFICATES, resource = Resource.ATTRIBUTE, name = "csr", affiliatedResource = Resource.CERTIFICATE, operation = Operation.LIST_ATTRIBUTES)
-    public List<BaseAttribute> getCsrGenerationAttributes() {
-        return certificateService.getCsrGenerationAttributes();
+    public List<BaseAttribute> getCsrGenerationAttributes(String raProfileUuid) throws NotFoundException, ConnectorException {
+        if (raProfileUuid == null || raProfileUuid.isBlank()) {
+            return certificateService.getCsrGenerationAttributes();
+        }
+        return certificateService.getCsrGenerationAttributes(SecuredUUID.fromString(raProfileUuid));
     }
 
     @Override
--- a/src/main/java/com/otilm/core/dao/repository/RaProfileRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/RaProfileRepository.java
@@ -14,12 +14,16 @@ public interface RaProfileRepository extends SecurityFilterRepository<RaProfile,
     Optional<RaProfile> findByUuid(UUID uuid);
 
     /**
-     * Loads an RA profile with its authority, connector and connector interface eagerly.
+     * Loads an RA profile with its authority, connector, connector interface and the connector's
+     * function groups eagerly. The function groups are fetched so callers that traverse the connector
+     * outside an active transaction (e.g. request-attribute resolution under
+     * {@code Propagation.NOT_SUPPORTED}) do not depend on open-session-in-view for the lazy collection.
      */
     @EntityGraph(attributePaths = {
             "authorityInstanceReference",
             "authorityInstanceReference.connectorInterface",
-            "authorityInstanceReference.connector"
+            "authorityInstanceReference.connector",
+            "authorityInstanceReference.connector.functionGroups"
     })
     Optional<RaProfile> findWithAuthorityByUuid(UUID uuid);
 
--- a/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
@@ -1,7 +1,6 @@
 package com.otilm.core.service.impl;
 
 import com.otilm.api.model.common.UuidDto;
-import com.otilm.core.attribute.CsrAttributes;
 import com.otilm.api.exception.*;
 import com.otilm.api.model.client.attribute.RequestAttribute;
 import com.otilm.api.model.client.attribute.ResponseAttribute;
@@ -35,6 +34,7 @@ import com.otilm.core.attribute.engine.AttributeContentPurpose;
 import com.otilm.core.attribute.engine.AttributeEngine;
 import com.otilm.core.attribute.engine.AttributeOperation;
 import com.otilm.core.attribute.engine.records.ObjectAttributeContentInfo;
+import com.otilm.core.certificate.request.IssuanceDefinitionResolver;
 import com.otilm.core.comparator.SearchFieldDataComparator;
 import com.otilm.core.config.cache.CacheConfig;
 import com.otilm.core.dao.entity.*;
@@ -149,6 +149,8 @@ public class CertificateServiceImpl implements CertificateExternalService, Certi
     private CertificateValidationWriter validationWriter;
     private CertificateRequestRepository certificateRequestRepository;
     private RaProfileRepository raProfileRepository;
+    private IssuanceDefinitionResolver issuanceDefinitionResolver;
+    private RaProfileCertificateRequestAttributeService requestAttributeService;
     private CertificateRegistrationAuthorizationRepository registrationAuthorizationRepository;
     private RaProfileInternalService raProfileService;
     private GroupRepository groupRepository;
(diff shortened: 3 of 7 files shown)


src/main/java/com/otilm/core/api/web/CertificateControllerImpl.java
```diff
@@ -165,8 +165,11 @@ public CertificateValidationResultDto getCertificateValidationResult(@LogResourc
...
     @Override
     @AuditLogged(module = Module.CERTIFICATES, resource = Resource.ATTRIBUTE, name = "csr", affiliatedResource = Resource.CERTIFICATE, operation = Operation.LIST_ATTRIBUTES)
-    public List<BaseAttribute> getCsrGenerationAttributes() {
-        return certificateService.getCsrGenerationAttributes();
+    public List<BaseAttribute> getCsrGenerationAttributes(String raProfileUuid) throws NotFoundException, ConnectorException {
+        if (raProfileUuid == null || raProfileUuid.isBlank()) {
+            return certificateService.getCsrGenerationAttributes();
+        }
+        return certificateService.getCsrGenerationAttributes(SecuredUUID.fromString(raProfileUuid));
     }
```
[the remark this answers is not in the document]
ivosh: Fixed by using UUID instead of String.

src/test/java/com/otilm/core/api/web/CertificateControllerImplTest.java
```diff
@@ -0,0 +1,73 @@
...
+        // when
+        List<BaseAttribute> result = controller.getCsrGenerationAttributes(raProfileUuid.toString());
+
+        // then
+        assertThat(result).isSameAs(resolved);
+        ArgumentCaptor<SecuredUUID> captor = ArgumentCaptor.forClass(SecuredUUID.class);
+        verify(certificateService).getCsrGenerationAttributes(captor.capture());
+        assertThat(captor.getValue().getValue()).isEqualTo(raProfileUuid);
+    }
+}
```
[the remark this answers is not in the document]
ivosh: Fixed by using UUID instead of String.

src/main/java/com/otilm/core/dao/repository/RaProfileRepository.java
```diff
@@ -14,12 +14,16 @@ public interface RaProfileRepository extends SecurityFilterRepository<RaProfile,
...
+     * outside an active transaction (e.g. request-attribute resolution under
+     * {@code Propagation.NOT_SUPPORTED}) do not depend on open-session-in-view for the lazy collection.
      */
     @EntityGraph(attributePaths = {
             "authorityInstanceReference",
             "authorityInstanceReference.connectorInterface",
-            "authorityInstanceReference.connector"
+            "authorityInstanceReference.connector",
+            "authorityInstanceReference.connector.functionGroups"
     })
```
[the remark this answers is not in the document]
ivosh: Keeping the single shared finder. Of its 3 callers, 2 run under `Propagation.NOT_SUPPORTED` (`CertificateServiceImpl.getCsrGenerationAttributes` and `ClientOperationServiceImpl.registerCertificate`) and genuinely need `functionGroups` initialized before the session closes — that eager fetch is the method's purpose. The lone read-only caller (`listAvailableOperations`) has an open session and tolerates it harmlessly. The graph fetches only one collection (everything else is to-one), so there is no cartesian product — just one bounded LEFT JOIN over a single profile's function groups, negligible for a by-UUID lookup. A dedicated minimal finder would add surface area and a LazyInitializationException footgun for future NOT_SUPPORTED callers, so the tradeoff isn't worth it here.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1589,7 +1601,19 @@ public void checkRevokePermissions() {
     @Override
     @ExternalAuthorization(resource = Resource.CERTIFICATE, action = ResourceAction.ANY)
     public List<BaseAttribute> getCsrGenerationAttributes() {
-        return CsrAttributes.csrAttributes();
+        return requestAttributeService.getDefaultSet();
```
lubomirw, line 1604: **[High]** | Correctness

Removing `@CoreAttributeDefinitions csrAttributes()` drops the only path that registered the six built-in CSR DN attribute definitions (connectorUuid=null) on context refresh. Migration `V202606251000` only UPDATEs existing rows — no INSERT anywhere — and `AttributeEngine.updateObjectDataAttributesContent` (AttributeEngine.java:972) throws `NotFoundException(AttributeDefinition)` when the definition is absent. On a fresh install, submitting a CSR carrying DN attributes will fail.

_Suggested:_ Keep a registration path (a `@CoreAttributeDefinitions` bean returning `csrAttributesAsDataAttributesV3()`, or a migration inserting the six rows) and add a fresh-DB ITest that submits a CSR with DN attributes.

<sub>flagged by superpowers code review; verified against AttributeEngine.java:972 + migration V202606251000</sub>

src/test/java/com/otilm/core/integration/service/CsrGenerationAttributesITest.java
```diff
@@ -0,0 +1,204 @@
...
+        // when
+        List<BaseAttribute> resolved = certificateService.getCsrGenerationAttributes(SecuredUUID.fromUUID(raProfile.getUuid()));
+
+        // then: the platform default set (built-in seed while unset)
+        assertThat(resolved).extracting(BaseAttribute::getName)
+                .containsExactlyElementsOf(DefaultRequestAttributeSet.seed().stream().map(BaseAttribute::getName).toList());
+    }
+
+    @Test
+    void withProfileMergesPersistedConnectorSet() throws Exception {
```
lubomirw, line 109: **[Low]** | Testing

`withProfileMergesPersistedConnectorSet` stubs `ExtendedAttributeService` via `@MockitoBean`, so `validateLegacyConnector` — which iterates the newly eager-fetched `connector.functionGroups` — never runs against persisted entities. The entity-graph change that motivates this PR is therefore unexercised.

_Suggested:_ Add an ITest driving the real `ExtendedAttributeService` against a persisted connector with ≥1 function group (ideally with `open-in-view=false`) asserting no `LazyInitializationException`.

src/main/java/com/otilm/core/api/web/CertificateControllerImpl.java
```diff
@@ -165,8 +165,11 @@ public CertificateValidationResultDto getCertificateValidationResult(@LogResourc
 
     @Override
     @AuditLogged(module = Module.CERTIFICATES, resource = Resource.ATTRIBUTE, name = "csr", affiliatedResource = Resource.CERTIFICATE, operation = Operation.LIST_ATTRIBUTES)
-    public List<BaseAttribute> getCsrGenerationAttributes() {
-        return certificateService.getCsrGenerationAttributes();
+    public List<BaseAttribute> getCsrGenerationAttributes(UUID raProfileUuid) throws NotFoundException, ConnectorException {
```
lubomirw: **[Nit]** | Maintainability

`@AuditLogged` (resource=ATTRIBUTE, affiliatedResource=CERTIFICATE) is unchanged, so a call carrying `raProfileUuid` does not record which RA profile scoped the resolved set — default-set vs profile-scoped calls are indistinguishable in the audit trail.

_Suggested:_ Log the `raProfileUuid` (e.g. `@LogResource` on the parameter, or `affiliatedResource = RA_PROFILE`) for the resolved branch.

lubomirw (changes requested):
Automated multi-reviewer triage — 3 findings (REQUEST_CHANGES). Two additional findings were assessed and dropped as likely noise (interface-signature compile risk — CI is green; nested `functionGroup` lazy load — query binds by identifier).

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
ivosh: Fixed in df40bf90 by re-adding `@CoreAttributeDefinitions` to `csrAttributesAsDataAttributesV3()`, which restores the context-refresh registration of the six built-in CSR DN attribute definitions (connectorUuid=null) that was lost when `csrAttributes()` was removed. Confirmed empirically: on a fresh DB the six definitions were absent, so a CSR carrying DN attributes hit the `NotFoundException` at AttributeEngine:972. The annotation is behavior-identical to main; the line-972 lookup doesn't filter by operation, so registration operation is irrelevant.

src/main/java/com/otilm/core/api/web/CertificateControllerImpl.java
ivosh: Fixed in df40bf90. Switched `affiliatedResource` to `RA_PROFILE` and added `@LogResource(uuid = true, affiliated = true)` on `raProfileUuid`, matching the TspProfile parameter-affiliation pattern. Profile-scoped calls now record the RA profile UUID in the audit trail; default-set calls pass null and the aspect skips null UUIDs (`getResourceUuidsFromParameter` returns null), so that branch is unaffected.

src/test/java/com/otilm/core/integration/service/CsrGenerationAttributesITest.java
ivosh: Leaving as-is for now. The sibling ITests in this class (`withProfileFallsBackToPlatformDefaultSetWhenNothingConfigured`, the configured-set case) also stub `ExtendedAttributeService` via `@MockitoBean`, so driving the real service against a persisted connector with function groups (open-in-view=false) would be new test infrastructure not established anywhere in this suite. I'd rather not introduce it as a one-off in this PR. Happy to file a follow-up if you want the LazyInitialization guard covered properly. Leaving this thread open for you to resolve.

lubomirw (approved):

PR #1836, opened 2026-07-16
Consolidate Spring test contexts behind shared mock modules

### LINKED ISSUE #1704: Consolidate Spring test-context configurations to reduce context-cache fragmentation
## Description
Test-suite wall-clock time is dominated by how many distinct Spring contexts boot, not how many tests run: every unique combination of `@MockitoBean` fields, active profiles, and property sources is a separate context-cache key that forces a full multi-second context boot.

Current fragmentation signals in `src/test/java`:
- 36 files use `@MockitoBean`, with the same beans mocked repeatedly in different files (`NotificationProducer` in 5 files, `PollFeature` in 3, several others in 2) — each spelling a fresh context configuration.
- The messaging tests spell `@ActiveProfiles` three different ways (`value = {"messaging-int-test"}, inheritProfiles = false` / `{"messaging-int-test"}` / `value = {"messaging-int-test"}`), which may split the cache further.

Consolidate recurring mock sets into shared `@TestConfiguration` on the context base classes (`BaseSpringBootTest`, `BaseMessagingIntTest`), normalize the `@ActiveProfiles` spellings, and finish with a guard test capping the distinct context-configuration count so fragmentation cannot silently regress.

Best executed after the per-package migration under #1693 is largely complete, when context-loading tests live under the single `com.otilm.core.integration` root and the change surface is one package tree. Complements #1693: without it, the dedicated integration CI job just relocates the slowness.

## Definition of Done
- [ ] Census script/report enumerating distinct Spring context configurations (baseline count recorded)
- [ ] Recurring mock sets consolidated into shared `@TestConfiguration` on the context base classes
- [ ] `@ActiveProfiles` spellings normalized to one canonical form
- [ ] Guard test asserts the distinct context-configuration count stays at or below the consolidated baseline
- [ ] Integration-job wall-clock and context-boot count measured before/after consolidation

ivosh:


lubomirw (commented):
Automated multi-reviewer triage — 1 finding (COMMENT). See the inline comment on `ProducerMocks.java` (latent `@Primary`-shadowing risk). Not blocking; posted for the record. cc @ivosh

src/test/java/com/otilm/core/util/mockbeans/ProducerMocks.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Mocks the JMS producers (external-boundary I/O that tests never exercise for real).
+ */
+@TestConfiguration
+public class ProducerMocks {
+
+    @Bean
+    @Primary
+    NotificationProducer mockNotificationProducer() {
```
lubomirw, line 21: **[High]** | Correctness

`@Bean @Primary` **shadows** but does not **replace** the real `NotificationProducer`. It is still `@Component`-scanned and instantiated, so its `@TransactionalEventListener(phase = AFTER_COMMIT, fallbackExecution = true)` stays registered and can fire a real `jmsTemplate.convertAndSend` after the real commits these tests perform — reintroducing the external JMS I/O this mock exists to suppress. `@Primary` only steers `@Autowired` injection; unlike `@MockitoBean` it does not remove the bean definition.

**Not a live break today:** the cache/parity ITests that import `ProducerMocks` don't publish `NotificationMessage` events, so the listener never fires (CI is green). It is **latent**: any future test that both commits and publishes such an event while importing this module would get real sends.

_Suggested:_ for beans carrying framework-registered side effects (`@TransactionalEventListener` / `@Scheduled` / `@JmsListener`), replace them via `@MockitoBean` (definition override) rather than shadow them with `@Bean @Primary` — or exclude the real `@Component` from the scan so no second live listener remains.

cc @ivosh

<sub>flagged by superpowers code review — multi-reviewer triage</sub>

lubomirw (approved):

src/test/java/com/otilm/core/util/mockbeans/ProducerMocks.java
ivosh: Addressed in https://github.com/OmniTrustILM/core/pull/1931.

PR #1839, opened 2026-07-17
Fix CMP/SCEP handling of pending validation and CRMF replies

### LINKED ISSUE #1830: Protocols - CMP: IR/CR request returns "rejection" to client when async validation status is "Not checked" at response time - client never receives issued certificate, cert stays orphaned on platform
### Description

A CMP **Initialization Request** (`ir`) / **Certification Request** (`cr`) can end in a broken state where **the certificate is actually issued and stored on the platform** (later reaches `State: Issued`, `Validation Status: Valid`), but the **CMP response sent back to the client is `rejection`**, so the requesting client never receives its own certificate.

The race is between three concurrent flows: (a) issuance on EJBCA, (b) async certificate validation on the platform, (c) construction of the CMP `IP`/`CP` response. If (c) evaluates the certificate's validation status at a moment when (b) has not yet run, the status is `Not checked` — which the CMP handler interprets as "cert not valid" and returns `PKIStatus: rejection`, `PKIFailureInfo: systemFailure` to the client. A few seconds later (b) catches up and marks the cert `Valid`, but by then the CMP response has already been finalised.

Observed rejection message returned by the platform:

```
CMP error: received "rejection" status rather than cert
CMP error: request rejected by server:
  PKIStatus: rejection
  PKIFailureInfo: systemFailure
  StatusString: "SN=… | Certificate is not valid.
                 UUID: <uuid>, Fingerprint: <fp>, Status: Not checked"
cannot extract certificate from response
```

Meanwhile the same cert (matching UUID / fingerprint) appears in the Certificate Inventory as `State: Issued`, `Validation Status: Valid`.

### Steps to Reproduce

1. Configure a CMP Profile (Shared Secret PBM, both request and response protection = Shared Secret) with Default RA Profile pointing at a working EJBCA RA
2. Use the profile-based URL directly
3. From a client that supports CMP v2 (e.g. OpenSSL 3.x `openssl cmp`), send an `ir` with correct PBM auth and PoP:

   ```
   openssl cmp -cmd ir \
     -server https://<host>/api/v1/protocols/cmp/<cmpProfileName> \
     -recipient "/CN=CMP" \
     -ref "<any>" \
     -secret 'pass:<shared-secret>' \
     -newkey <newkey.pem> \
     -subject "/CN=<cn>.example.com" \
     -sans "<cn>.example.com" \
     -certout out.crt \
     -poll_count 5 -total_timeout 60
   ```
4. Observe on the client: `rejection` with `Certificate is not valid … Status: Not checked`
5. Observe on the platform: the corresponding certificate exists in Certificate Inventory with `State: Issued` and (after a short delay) `Validation Status: Valid`

### Expected Behavior

- The CMP response returned to the client must be based on **issuance success**, not on the transient state of an asynchronous validation
- If the underlying certificate was issued successfully by EJBCA (State becomes `Issued`), the CMP handler returns `PKIStatus: accepted` with the cert in `IP`/`CP` — regardless of whether async validation has completed at that instant
- `Validation Status = Not checked` is a normal transient state and must not be treated as a rejection reason
- Client receives its certificate; the platform record and the client-side certificate match; no orphaned cert is left behind

### Actual Behavior

- Cert is created on the platform and eventually reaches `State: Issued`, `Validation Status: Valid`
- CMP handler constructs the response too early, sees `Validation Status: Not checked`, and returns `PKIStatus: rejection` / `PKIFailureInfo: systemFailure` with StatusString `Certificate is not valid`
- CMP client (`openssl cmp`) cannot extract the certificate from the rejection response
- The freshly issued cert is orphaned — it exists on the platform (and in EJBCA as an End Entity + issued cert), but the requesting client has no copy of it
- Retrying from the client re-triggers a `End Entity … already exists` error in EJBCA, requiring manual EJBCA cleanup

### Impact / Production Consequences

- **CMP end-to-end broken** for its main use case (unattended provisioning of routers / IoT / MDM-managed devices) — the client that asked for a cert never gets it
- **Orphaned certs / End Entities** accumulate on the platform and in EJBCA, requiring manual cleanup after every failed attempt
- **No safe retry** — the client cannot simply repeat the `ir` because the End Entity is already registered
- **Silent divergence** — platform inventory shows the cert as valid and issued; operator sees no visible issue, but the client device sits without a certificate

### Severity

Major

### Module

Protocols

### Attachments

_No response_

### Logs

_No response_

### Environment

ILM Core: 2.18.1-SNAPSHOT (16 Jul 2026, 08:54)
Commit: 5f7bc88
Branch: main
Build time: 2026-07-16T06:54:47Z
ILM Frontend: 2.18.0 (16 Jul 2026, 10:25)
Client:
- macOS Tahoe 26.5.1

### LINKED ISSUE #1834: Protocols - SCEP: enrollment response stuck in PENDING then FAILURE while cert is actually issued on platform - client never receives the certificate; also violates RFC 8894 §3.3.1
### Description

A SCEP enrollment (`PKCSReq`) can end in a broken state where **the certificate is actually issued and stored on the platform** (later verified in Certificate Inventory as `State: Issued`, `Validation Status: Valid`), but the SCEP client polling for the enrollment response never receives the certificate.

The observed sequence:

1. Client sends `PKCSReq` — server responds with `pkiStatus: PENDING`
2. Client polls with `GetCertInitial` (same transaction ID) — server continues to respond `PENDING`
3. After a few polls the server changes the response to `pkiStatus: FAILURE` with `failInfo: badRequest` and reason string **"Transaction not permitted or supported"**
4. The corresponding certificate exists in the Certificate Inventory as `State: Issued`, `Validation Status: Valid` — but the client cannot retrieve it via SCEP anymore

There are two related defects visible here:

- **Primary — PENDING never transitions to SUCCESS.** The SCEP handler's response state is decoupled from the actual certificate state; even after the cert has been issued and validated, the SCEP handler continues to report `PENDING`.
- **Secondary — non-compliance with RFC 8894 §3.3.1.** Once a transaction has been completed successfully on the server side, subsequent polls (or re-issued `PKCSReq` with the same transaction ID) MUST return the previously issued certificate. Instead the server rejects with `FAILURE / Transaction not permitted or supported`, leaving the client with no way to recover the cert.

Observed `sscep` output (fresh key, fresh transaction ID, with polling enabled):

```
sscep: sending certificate request
sscep: valid response from server
sscep: reply transaction id: 6B01E0C282E4C07829320AE5C37D9E9E
sscep: pkistatus: PENDING
sscep: requesting certificate (#1)
sscep: pkistatus: PENDING
sscep: requesting certificate (#2)
sscep: pkistatus: FAILURE
sscep: reason: Transaction not permitted or supported
```

Same TID's corresponding cert on the platform: `State: Issued`, `Validation Status: Valid`.

### Steps to Reproduce

1. Configure a SCEP Profile (Enable Challenge Password, Include CA Certificate Chain, Default RA Profile pointing at a working EJBCA RA)
2. Get the CA chain via `sscep getca` (or the equivalent HTTP GET) — succeeds, chain is returned correctly
3. Generate an ECDSA P-256 key + CSR with:
   - `Subject: CN=<any>.example.com` (no `O` — required by the EJBCA End Entity Profile)
   - `subjectAltName = DNS:<any>.example.com`
   - `challengePassword` attribute = the value configured in the SCEP Profile
4. Run enrollment with polling:

   ```
   sscep enroll \
     -u "http://<host>/api/v1/protocols/scep/<scepProfileName>/pkiclient.exe" \
     -c <ra-cert> \
     -k <key> \
     -r <csr> \
     -l <out.crt> \
     -t 5 -n 12
   ```
5. Observe: first response `PENDING`, subsequent polls remain `PENDING`, then transition to `FAILURE / Transaction not permitted or supported`
6. Observe on the platform: a certificate matching the CSR's subject is in the Certificate Inventory as `State: Issued`, `Validation Status: Valid`

### Expected Behavior

- Once the certificate is successfully issued on the platform (State reaches `Issued`), the SCEP handler must transition the transaction's response state to `SUCCESS` and return the certificate to the client on the next poll (`GetCertInitial`)
- If the response happens to still be `PENDING` when a client polls, subsequent polls must continue returning `PENDING` (or `SUCCESS`) — **never** `FAILURE / Transaction not permitted or supported` while the cert is being / has been issued
- Per RFC 8894 §3.3.1, if the transaction ID identifies a completed transaction, the server must return the previously issued cert
- The client receives its certificate through the SCEP flow; the platform record and the client-side certificate match; no orphan is left behind

### Actual Behavior

- SCEP handler always returns `PENDING` on the initial `PKCSReq`, regardless of whether the underlying issuance completed immediately or asynchronously
- Polling never yields `SUCCESS` — response stays `PENDING` for a few polls, then flips to `FAILURE (Transaction not permitted or supported)`
- Certificate is issued on the platform (`State: Issued`, `Validation Status: Valid`) but the SCEP client has no way to retrieve it
- The transaction becomes unrecoverable — client cannot re-issue `PKCSReq` with same TID (server rejects), and cannot poll further (also rejected)

### Impact / Production Consequences

- **SCEP end-to-end broken** for its main use case — enrolling client devices (routers, printers, IoT, MDM-managed devices) that rely on SCEP for automated provisioning; the device never receives its certificate
- **Orphaned certs and End Entities** accumulate on the platform and in EJBCA on every failed enrollment
- **No retry possible** — same key / same TID gets rejected as "Transaction not permitted"; client is forced to generate a fresh key pair for every attempt, which contradicts the SCEP idempotency contract
- **Silent divergence** — platform inventory reports the cert as valid and issued; operator sees no visible issue, but the client device sits without a certificate
- Same class as #1830 (CMP IR) and #1833 (CMP RR) — protocol handlers' state machines are decoupled from actual certificate state across all three protocols

### Severity

Major

### Module

Protocols

### Attachments

_No response_

### Logs

_No response_

### Environment

ILM Core: 2.18.1-SNAPSHOT (16 Jul 2026, 08:54)
Commit: 5f7bc88
Branch: main
Build time: 2026-07-16T06:54:47Z
ILM Frontend: 2.18.0 (16 Jul 2026, 10:25)
Client:
- macOS Tahoe 26.5.1

klaraf755:
Fixes #1830
Fixes #1834

### DIFF
--- a/src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
+++ b/src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
@@ -45,7 +45,10 @@ import java.util.*;
  * @see CrmfKurMessageHandler
  */
 @Component
-@Transactional//(propagation = Propagation.REQUIRES_NEW)
+// noRollbackFor keeps Spring's default (no rollback on checked exceptions): CMP surfaces
+// CmpBaseException as a protocol error response, and this handler shares the caller's
+// transaction, so rolling back on it would mark that transaction rollback-only.
+@Transactional(noRollbackFor = CmpBaseException.class)
 public class CrmfMessageHandler implements MessageHandler<PKIMessage> {
 
     private static final Logger LOG = LoggerFactory.getLogger(CrmfMessageHandler.class.getName());
--- a/src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
+++ b/src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
@@ -52,9 +52,10 @@ public class CrmfMessageHandler implements MessageHandler<PKIMessage> {
 
     /**
      * How long to wait for the event-driven post-issuance validation to move the
-     * certificate off {@code NOT_CHECKED} before validating inline as a fallback.
-     * The validation listener typically lands within ~100 ms of issuance; 3 s covers
-     * slow OCSP/CRL checks without stretching the request unreasonably.
+     * certificate off {@code NOT_CHECKED}. If it does not resolve within this budget the
+     * transient {@code NOT_CHECKED} is accepted (it is not a rejection reason). The
+     * validation listener typically lands within ~100 ms of issuance; 3 s covers slow
+     * OCSP/CRL checks without stretching the request unreasonably.
      */
     private static final long VALIDATION_STATUS_WAIT_MS = 3_000L;
 
--- a/src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
+++ b/src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
@@ -13,6 +13,7 @@ import com.otilm.core.dao.entity.Certificate;
 import com.otilm.core.dao.entity.CertificateContent;
 import com.otilm.core.dao.entity.cmp.CmpTransaction;
 import com.otilm.core.service.CertificateInternalService;
+import com.otilm.core.service.handler.CertificateValidationStatusPoller;
 import com.otilm.core.service.cmp.configurations.ConfigurationContext;
 import com.otilm.core.service.cmp.message.CmpTransactionService;
 import com.otilm.core.service.cmp.message.PkiMessageDumper;
@@ -76,6 +77,13 @@ public class CrmfMessageHandler implements MessageHandler<PKIMessage> {
     @Autowired
     public void setPollFeature(PollFeature pollFeature) {
         this.pollFeature = pollFeature;
+    }
+
+    private CertificateValidationStatusPoller validationStatusPoller;
+
+    @Autowired
+    public void setValidationStatusPoller(CertificateValidationStatusPoller validationStatusPoller) {
+        this.validationStatusPoller = validationStatusPoller;
     }
 
     private CmpTransactionService cmpTransactionService;
@@ -439,7 +447,7 @@ public class CrmfMessageHandler implements MessageHandler<PKIMessage> {
             return certificate.getValidationStatus();
         }
         try {
-            CertificateValidationStatus resolved = pollFeature.pollValidationStatus(
+            CertificateValidationStatus resolved = validationStatusPoller.pollValidationStatus(
                     certificate.getUuid(), VALIDATION_STATUS_WAIT_MS);
             LOG.debug("TID={}, UUID={} | validation status after wait: {}",
                     tid, certificate.getUuid(), resolved);
--- a/src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
+++ b/src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
@@ -53,15 +53,6 @@ public class CrmfMessageHandler implements MessageHandler<PKIMessage> {
 
     private static final Logger LOG = LoggerFactory.getLogger(CrmfMessageHandler.class.getName());
 
-    /**
-     * How long to wait for the event-driven post-issuance validation to move the
-     * certificate off {@code NOT_CHECKED}. If it does not resolve within this budget the
-     * transient {@code NOT_CHECKED} is accepted (it is not a rejection reason). The
-     * validation listener typically lands within ~100 ms of issuance; 3 s covers slow
-     * OCSP/CRL checks without stretching the request unreasonably.
-     */
-    private static final long VALIDATION_STATUS_WAIT_MS = 3_000L;
-
     private static final List<Integer> ALLOWED_TYPES = List.of(
             PKIBody.TYPE_INIT_REQ,          // ir       [0]  CertReqMessages,       --Initialization Req
             PKIBody.TYPE_CERT_REQ,          // cr       [2]  CertReqMessages,       --Certification Req
@@ -408,16 +399,22 @@ public class CrmfMessageHandler implements MessageHandler<PKIMessage> {
         }
 
         // -- parse found CA certificate(s)
+        String leafCertificateUuid = leafCertificate.getUuid().toString();
         for (CertificateDetailDto certificate : caChain) {
-            // Only a certificate with a definitively bad status is rejected. NOT_CHECKED is
-            // a normal transient state — a certificate freshly issued in this same request
-            // stays NOT_CHECKED until the async validation lands — so it is briefly waited
-            // on and, if still unresolved, accepted: the response must reflect issuance
-            // success, not the progress of an asynchronous validation.
-            CertificateValidationStatus validationStatus = ensureValidationStatus(tid, certificate);
-            if (validationStatus != CertificateValidationStatus.VALID
-                    && validationStatus != CertificateValidationStatus.EXPIRING
-                    && validationStatus != CertificateValidationStatus.NOT_CHECKED) {
+            // Only the freshly-issued leaf may legitimately be NOT_CHECKED: its async
+            // validation may not have landed yet, so wait briefly and tolerate NOT_CHECKED
+            // if it never resolves (the response must reflect issuance success, not the
+            // progress of an asynchronous validation). CA / issuer certs are long-lived and
+            // must already be VALID/EXPIRING — a NOT_CHECKED (or worse) CA cert is rejected
+            // as before, and never waited on.
+            boolean isLeaf = certificate.getUuid().equals(leafCertificateUuid);
+            CertificateValidationStatus validationStatus = isLeaf
+                    ? validationStatusPoller.resolveOrKeep(certificate)
+                    : certificate.getValidationStatus();
+            boolean acceptable = validationStatus == CertificateValidationStatus.VALID
+                    || validationStatus == CertificateValidationStatus.EXPIRING
+                    || (isLeaf && validationStatus == CertificateValidationStatus.NOT_CHECKED);
+            if (!acceptable) {
                 throw new CmpCrmfValidationException(tid, bodyType, PKIFailureInfo.systemFailure,
                         String.format("SN=%s | Certificate is not valid. UUID: %s, Fingerprint: %s, Status: %s",
                                 leafCertificateSerialNumber,
@@ -436,31 +433,6 @@ public class CrmfMessageHandler implements MessageHandler<PKIMessage> {
             }
         }
         return certificateChain;
-    }
-
-    /**
-     * Certificate validation status is advanced asynchronously (event-driven after issuance,
-     * hourly batch as fallback), so a certificate issued moments ago can still be
-     * {@code NOT_CHECKED} when this response is built. Wait briefly for the in-flight
-     * validation to land so a definitively bad status can still be caught; if it does not
-     * land within the budget, the caller accepts {@code NOT_CHECKED} — it is a transient
-     * state, not a verdict, and must not fail an issuance that already succeeded.
-     */
-    private CertificateValidationStatus ensureValidationStatus(ASN1OctetString tid, CertificateDetailDto certificate) {
-        if (certificate.getValidationStatus() != CertificateValidationStatus.NOT_CHECKED) {
-            return certificate.getValidationStatus();
-        }
-        try {
-            CertificateValidationStatus resolved = validationStatusPoller.pollValidationStatus(
-                    certificate.getUuid(), VALIDATION_STATUS_WAIT_MS);
-            LOG.debug("TID={}, UUID={} | validation status after wait: {}",
-                    tid, certificate.getUuid(), resolved);
-            return resolved;
-        } catch (NotFoundException e) {
-            LOG.warn("TID={}, UUID={} | certificate not found while waiting for validation status",
-                    tid, certificate.getUuid());
-            return certificate.getValidationStatus();
-        }
     }
 
 }
--- a/src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
+++ b/src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
@@ -50,12 +50,12 @@ public class CrmfMessageHandler implements MessageHandler<PKIMessage> {
     private static final Logger LOG = LoggerFactory.getLogger(CrmfMessageHandler.class.getName());
 
     /**
-     * RFC 4210 §5.3.22 {@code checkAfter} (seconds). Tells the CMP client how long to wait
-     * before retrying the {@code pollReq} when the authority provider connector accepted the
-     * operation asynchronously. 60 s balances "client doesn't hammer the server" against
-     * "operator doesn't wait too long for state to propagate".
+     * How long to wait for the event-driven post-issuance validation to move the
+     * certificate off {@code NOT_CHECKED} before validating inline as a fallback.
+     * The validation listener typically lands within ~100 ms of issuance; 3 s covers
+     * slow OCSP/CRL checks without stretching the request unreasonably.
      */
-    private static final long POLL_REP_CHECK_AFTER_SECONDS = 60L;
+    private static final long VALIDATION_STATUS_WAIT_MS = 3_000L;
 
     private static final List<Integer> ALLOWED_TYPES = List.of(
             PKIBody.TYPE_INIT_REQ,          // ir       [0]  CertReqMessages,       --Initialization Req
@@ -266,9 +266,12 @@ public class CrmfMessageHandler implements MessageHandler<PKIMessage> {
     }
 
     /**
-     * Connector accepted the operation asynchronously (HTTP 202): persist the transaction
-     * so a subsequent {@code pollReq} can be correlated back to the in-flight cert, then
-     * emit a CMP {@code pollRep} so the client knows to retry later (RFC 4210 §5.2.6).
+     * The certificate is still pending after the poll budget: persist the transaction so a
+     * subsequent {@code pollReq} can be correlated back to the in-flight cert, then answer
+     * the ir/cr/kur with an ip/cp/kup whose {@code PKIStatusInfo} is {@code waiting}. Per
+     * RFC 4210 §5.3.22 the polling exchange is initiated by exactly this response — the
+     * client then sends {@code pollReq} (handled by {@link PollReqMessageHandler}). A bare
+     * {@code pollRep} here would be out-of-state and conformant clients reject it.
      */
     private PKIMessage handleAsynchronousAcceptance(ASN1OctetString tid, PKIMessage request,
                                                     ConfigurationContext configuration, String msgBodyType,
@@ -281,20 +284,23 @@ public class CrmfMessageHandler implements MessageHandler<PKIMessage> {
                 trxState,
                 request.getBody().getType()));
         try {
-            LOG.info("TID={} | CRMF {} accepted asynchronously (cert {}); returning pollRep",
+            LOG.info("TID={} | CRMF {} still pending after poll budget (cert {}); returning response with status 'waiting'",
                     tid, msgBodyType, requestedCert.getUuid());
+            CertResponse waitingResponse = new CertResponse(
+                    crmf.getCertReqId(),
+                    new PKIStatusInfo(PKIStatus.waiting));
             return new PkiMessageBuilder(configuration)
                     .addHeader(PkiMessageBuilder.buildBasicHeaderTemplate(request))
-                    .addBody(PkiMessageBuilder.createPollRepBody(
-                            crmf.getCertReqId(),
-                            POLL_REP_CHECK_AFTER_SECONDS,
-                            "Awaiting asynchronous completion"))
+                    .addBody(PkiMessageBuilder.createIpCpKupBody(
+                            request.getBody(),
+                            new CertResponse[]{waitingResponse},
+                            null))
                     .addExtraCerts(null)
                     .build();
         } catch (Exception e) {
-            LOG.error("TID={} | CRMF pollRep message cannot be built (type={})", tid, msgBodyType, e);
+            LOG.error("TID={} | CRMF waiting response cannot be built (type={})", tid, msgBodyType, e);
             throw new CmpCrmfValidationException(tid, request.getBody().getType(), PKIFailureInfo.systemFailure,
-                    "CRMF pollRep cannot be built, type=" + PkiMessageDumper.msgTypeAsString(request.getBody().getType()));
+                    "CRMF waiting response cannot be built, type=" + PkiMessageDumper.msgTypeAsString(request.getBody().getType()));
         }
     }
 
@@ -391,15 +397,21 @@ public class CrmfMessageHandler implements MessageHandler<PKIMessage> {
 
         // -- parse found CA certificate(s)
         for (CertificateDetailDto certificate : caChain) {
-            // only certificate with valid status should be used
-            if (certificate.getValidationStatus() != CertificateValidationStatus.VALID
-                    && certificate.getValidationStatus() != CertificateValidationStatus.EXPIRING) {
+            // Only a certificate with a definitively bad status is rejected. NOT_CHECKED is
+            // a normal transient state — a certificate freshly issued in this same request
+            // stays NOT_CHECKED until the async validation lands — so it is briefly waited
+            // on and, if still unresolved, accepted: the response must reflect issuance
+            // success, not the progress of an asynchronous validation.
+            CertificateValidationStatus validationStatus = ensureValidationStatus(tid, certificate);
+            if (validationStatus != CertificateValidationStatus.VALID
+                    && validationStatus != CertificateValidationStatus.EXPIRING
+                    && validationStatus != CertificateValidationStatus.NOT_CHECKED) {
                 throw new CmpCrmfValidationException(tid, bodyType, PKIFailureInfo.systemFailure,
                         String.format("SN=%s | Certificate is not valid. UUID: %s, Fingerprint: %s, Status: %s",
                                 leafCertificateSerialNumber,
                                 certificate.getUuid(),
                                 certificate.getFingerprint(),
-                                certificate.getValidationStatus().getLabel()));
+                                validationStatus.getLabel()));
             }
             try {
                 certificateChain.add(CertificateUtil.parseCertificate(certificate.getCertificateContent()));
@@ -412,6 +424,31 @@ public class CrmfMessageHandler implements MessageHandler<PKIMessage> {
             }
         }
         return certificateChain;
+    }
+
+    /**
+     * Certificate validation status is advanced asynchronously (event-driven after issuance,
+     * hourly batch as fallback), so a certificate issued moments ago can still be
+     * {@code NOT_CHECKED} when this response is built. Wait briefly for the in-flight
+     * validation to land so a definitively bad status can still be caught; if it does not
+     * land within the budget, the caller accepts {@code NOT_CHECKED} — it is a transient
+     * state, not a verdict, and must not fail an issuance that already succeeded.
+     */
+    private CertificateValidationStatus ensureValidationStatus(ASN1OctetString tid, CertificateDetailDto certificate) {
+        if (certificate.getValidationStatus() != CertificateValidationStatus.NOT_CHECKED) {
+            return certificate.getValidationStatus();
+        }
+        try {
+            CertificateValidationStatus resolved = pollFeature.pollValidationStatus(
+                    certificate.getUuid(), VALIDATION_STATUS_WAIT_MS);
+            LOG.debug("TID={}, UUID={} | validation status after wait: {}",
+                    tid, certificate.getUuid(), resolved);
+            return resolved;
+        } catch (NotFoundException e) {
+            LOG.warn("TID={}, UUID={} | certificate not found while waiting for validation status",
+                    tid, certificate.getUuid());
+            return certificate.getValidationStatus();
+        }
     }
 
 }
--- a/src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
+++ b/src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
@@ -123,7 +123,11 @@ public class PollFeature {
                             String.format("SN=%s | polling timed out after %d ms — cert is in transitional state %s, expected %s",
                                     serialNumber, timeoutMs, current, expectedState));
                 }
-                TimeUnit.MILLISECONDS.sleep(POLL_INTERVAL_MS);
+                // Clamp to the remaining budget so the loop never overshoots timeoutMs by up to a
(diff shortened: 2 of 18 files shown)


src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
```diff
@@ -50,12 +51,12 @@ public class CrmfMessageHandler implements MessageHandler<PKIMessage> {
...
 
     /**
-     * RFC 4210 §5.3.22 {@code checkAfter} (seconds). Tells the CMP client how long to wait
-     * before retrying the {@code pollReq} when the authority provider connector accepted the
-     * operation asynchronously. 60 s balances "client doesn't hammer the server" against
-     * "operator doesn't wait too long for state to propagate".
+     * How long to wait for the event-driven post-issuance validation to move the
+     * certificate off {@code NOT_CHECKED} before validating inline as a fallback.
+     * The validation listener typically lands within ~100 ms of issuance; 3 s covers
+     * slow OCSP/CRL checks without stretching the request unreasonably.
```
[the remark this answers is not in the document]
klaraf755: Updated comment.

src/main/java/com/otilm/core/service/handler/CertificateValidationStatusPoller.java
```diff
@@ -0,0 +1,81 @@
...
+     */
+    public CertificateValidationStatus pollValidationStatus(String uuid, long budgetMs) throws NotFoundException {
+        Certificate certificate = certificateService.getCertificateEntity(SecuredUUID.fromString(uuid));
+        long start = System.currentTimeMillis();
+        try {
+            while (certificate.getValidationStatus() == CertificateValidationStatus.NOT_CHECKED
+                    && System.currentTimeMillis() - start < budgetMs) {
+                TimeUnit.MILLISECONDS.sleep(POLL_INTERVAL_MS);
+                entityManager.refresh(certificate);
+            }
```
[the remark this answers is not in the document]
klaraf755: Added deadline.

src/test/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandlerTest.java
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    private static CertificateDetailDto chainCertificateDto(CertificateValidationStatus status) {
+        CertificateDetailDto dto = new CertificateDetailDto();
+        dto.setUuid(UUID.randomUUID().toString());
+        dto.setFingerprint("aa:bb:cc");
+        dto.setValidationStatus(status);
+        dto.setCertificateContent(generateSelfSignedCertBase64());
+        return dto;
+    }
```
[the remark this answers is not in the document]
klaraf755: Used static variable.

src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
```diff
@@ -391,15 +409,21 @@ private List<X509Certificate> loadCaCertificateChain(ASN1OctetString tid, int bo
...
-                    && certificate.getValidationStatus() != CertificateValidationStatus.EXPIRING) {
+            // Only a certificate with a definitively bad status is rejected. NOT_CHECKED is
+            // a normal transient state — a certificate freshly issued in this same request
+            // stays NOT_CHECKED until the async validation lands — so it is briefly waited
+            // on and, if still unresolved, accepted: the response must reflect issuance
+            // success, not the progress of an asynchronous validation.
+            CertificateValidationStatus validationStatus = ensureValidationStatus(tid, certificate);
+            if (validationStatus != CertificateValidationStatus.VALID
+                    && validationStatus != CertificateValidationStatus.EXPIRING
+                    && validationStatus != CertificateValidationStatus.NOT_CHECKED) {
```
lubomirw: **[Medium]** | Security

The NOT_CHECKED relaxation is applied to every certificate in `getCertificateChain(leaf, true)` — the leaf **plus** issuer/intermediate/root. Previously CA-chain certs required VALID/EXPIRING; an unvalidated CA cert can now be advertised where it was rejected before. Only the freshly-issued leaf is legitimately transient-NOT_CHECKED.

_Suggested:_ apply the NOT_CHECKED fallback only when the DTO UUID equals the freshly-issued leaf UUID; keep rejecting unresolved NOT_CHECKED for issuer/CA entries.

<sub>flagged by Copilot CLI + guided + superpowers</sub>

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -857,13 +866,54 @@ private void sendIntuneFailureMessage(IntuneScepServiceClient client, ScepReques
...
-                && certificate.getValidationStatus() != CertificateValidationStatus.EXPIRING) {
+        // Only a definitively bad status is rejected. NOT_CHECKED is a normal transient
+        // state — a certificate freshly issued in this same transaction stays NOT_CHECKED
+        // until the async validation lands — so it is briefly waited on and, if still
+        // unresolved, accepted: the response must reflect issuance success, not the
+        // progress of an asynchronous validation.
+        CertificateValidationStatus validationStatus = ensureValidationStatus(certificate);
+        if (validationStatus != CertificateValidationStatus.VALID
+                && validationStatus != CertificateValidationStatus.EXPIRING
+                && validationStatus != CertificateValidationStatus.NOT_CHECKED) {
```
lubomirw: **[Medium]** | Security

Same issue as the CMP path: `loadCertificateChain` loops `checkCertificateValidity` over the whole `getCertificateChain(leaf, true)`, so NOT_CHECKED is now accepted for CA/intermediate/root entries, not just the freshly-issued leaf.

_Suggested:_ restrict the NOT_CHECKED fallback to the newly-issued leaf UUID; keep enforcing VALID/EXPIRING for issuer/CA chain entries.

<sub>flagged by Copilot CLI + guided + superpowers</sub>

src/main/java/com/otilm/core/service/handler/CertificateValidationStatusPoller.java
```diff
@@ -0,0 +1,86 @@
...
+        long deadline = System.currentTimeMillis() + budgetMs;
+        try {
+            while (certificate.getValidationStatus() == CertificateValidationStatus.NOT_CHECKED) {
+                long remaining = deadline - System.currentTimeMillis();
+                if (remaining <= 0) {
+                    break;
+                }
+                // Clamp to the remaining budget so total wait never exceeds budgetMs, even
+                // when budgetMs is smaller than one sampling interval.
+                TimeUnit.MILLISECONDS.sleep(Math.min(POLL_INTERVAL_MS, remaining));
```
lubomirw, line 90: **[Medium]** | Performance

This sleep runs inside the caller's `@Transactional` boundary (`CrmfMessageHandler.ensureValidationStatus`, `ScepServiceImpl.checkCertificateValidity`), so a pooled DB connection + the request thread are pinned for up to `VALIDATION_STATUS_WAIT_MS` (3s) whenever the validation listener lags. core/CLAUDE.md "Transactions and external calls" forbids holding a tx across a slow/blocking operation — under a concurrent enrollment burst with a backlogged listener this can exhaust the connection pool.

_Suggested:_ run the wait outside the ambient tx (`Propagation.NOT_SUPPORTED`, or re-read each sample in a short separate tx).

<sub>flagged by guided + superpowers</sub>

src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
```diff
@@ -100,17 +105,20 @@ public PollResult pollCertificate(ASN1OctetString tid, String serialNumber, Stri
...
                 if (isTerminal(current)) {
                     log.warn("TID={}, SN={} | certificate uuid={} diverted to {} while waiting for {}",
                             tid, serialNumber, certUUID, current, expectedState);
                     return new PollResult.Diverted(current);
                 }
                 if (System.currentTimeMillis() - startRequest >= timeoutMs) {
+                    // PENDING_* is ridden out for the whole budget: even a synchronously-
+                    // completing connector transits PENDING_ISSUE on the actions-listener
+                    // thread, so only budget exhaustion makes "still pending" a verdict.
+                    if (current == CertificateState.PENDING_ISSUE || current == CertificateState.PENDING_REVOKE) {
```
lubomirw, line 117: **[Medium]** | Performance

PENDING_* is now ridden out for the full poll budget (`cmp.protocol.poll.feature.timeout`, default 10s) inside the `@Transactional` request, so a genuinely async (HTTP 202) issuance holds a DB connection for up to 10s — stacking on the 3s validation wait (~13s worst case). The rationale for riding it out (every issuance transits PENDING_ISSUE on the actions-listener thread) is sound; the transaction hold is the fixable part.

_Suggested:_ don't hold the request tx across the poll loop; verify pool/thread sizing tolerates the worst-case hold.

<sub>flagged by guided + superpowers</sub>

lubomirw (commented):
Automated multi-reviewer triage — 8 findings (COMMENT event; no confirmed Critical/High). Sources: Copilot CLI, guided-pr-review, superpowers, hygiene.

The protocol fix itself (answering pending ir/cr with ip(waiting) per RFC 4210 §5.3.22 instead of a bare pollRep, and exempting the waiting response from the 3GPP extraCerts requirement) is sound and well-tested. Findings below are the deduped, cross-verified issues.

_Not posted inline:_ a READ_COMMITTED-isolation dependency note (`entityManager.refresh` only observes the listener's commit under READ_COMMITTED — Postgres default, satisfied today; degrades gracefully) worth a one-line comment; and a Copilot 3GPP-PBM extraCerts concern that targets `PollReqMessageHandler` (the post-poll issued response, outside this diff) — verify that path separately if the end-to-end pending flow is untested.

src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
```diff
@@ -44,18 +45,22 @@
  * @see CrmfKurMessageHandler
  */
 @Component
-@Transactional//(propagation = Propagation.REQUIRES_NEW)
+// noRollbackFor keeps Spring's default (no rollback on checked exceptions): CMP surfaces
+// CmpBaseException as a protocol error response, and this handler shares the caller's
+// transaction, so rolling back on it would mark that transaction rollback-only.
+@Transactional(noRollbackFor = CmpBaseException.class)
```
lubomirw, line 51: **[Low]** | Maintainability

`CmpBaseException` is checked (`extends Exception`), and Spring `@Transactional` never rolls back on checked exceptions by default — so `noRollbackFor` changes nothing at runtime. The comment even says it "keeps Spring's default," yet the annotation implies a rollback hazard that does not exist.

_Suggested:_ drop `noRollbackFor` + the comment; or, if the real risk is a RuntimeException path (JPA / `PkiMessageBuilder.build`) rolling back the just-saved `CmpTransaction`, address that concretely — this annotation does not cover unchecked exceptions.

<sub>flagged by guided + superpowers (exception hierarchy verified)</sub>

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -78,7 +79,9 @@
 import java.util.*;
 
 @Service
-@Transactional
+// noRollbackFor keeps Spring's default (no rollback on checked exceptions): SCEP surfaces
+// these as protocol error responses rather than treating them as transaction failures.
+@Transactional(noRollbackFor = {ScepException.class, NotFoundException.class})
```
lubomirw, line 84: **[Low]** | Maintainability

Same no-op as the CMP handler: `ScepException` and `NotFoundException` both `extends Exception` (checked), so `@Transactional(noRollbackFor = {...})` has no runtime effect — checked exceptions never trigger rollback by default.

_Suggested:_ drop `noRollbackFor` + the comment, or state the RuntimeException-refactor guard explicitly if that is the intent.

<sub>flagged by guided + superpowers (exception hierarchy verified)</sub>

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -857,13 +866,54 @@ private void sendIntuneFailureMessage(IntuneScepServiceClient client, ScepReques
...
     }
+
+    /**
+     * How long to wait for the event-driven post-issuance validation to move the
+     * certificate off {@code NOT_CHECKED}. If it does not resolve within this budget the
+     * transient {@code NOT_CHECKED} is accepted (it is not a rejection reason). The
+     * validation listener typically lands within ~100 ms of issuance; 3 s covers slow
+     * OCSP/CRL checks without stretching the request unreasonably.
+     */
+    private static final long VALIDATION_STATUS_WAIT_MS = 3_000L;
```
lubomirw: **[Low]** | Maintainability

`VALIDATION_STATUS_WAIT_MS` (3s), its Javadoc, and the entire `ensureValidationStatus` helper (resolve / poll / catch-NotFound / fall-back-to-NOT_CHECKED) are copy-pasted verbatim into both this class and `CrmfMessageHandler` (~line 449). Both already inject `CertificateValidationStatusPoller`.

_Suggested:_ move the budget + resolve-with-fallback onto `CertificateValidationStatusPoller` (e.g. `resolveOrKeep(dto, budgetMs)`) so both handlers call one shared method and the budget lives in one place.

src/test/java/com/otilm/core/service/handler/CertificateValidationStatusPollerTest.java
```diff
@@ -0,0 +1,104 @@
...
+                .thenReturn(cert);
+
+        CertificateValidationStatus resolved = poller.pollValidationStatus(UUID.randomUUID().toString(), 3_000L);
+
+        assertThat(resolved).isEqualTo(CertificateValidationStatus.VALID);
+        verify(entityManager, never()).refresh(any());
+    }
+
+    @Test
+    void waitsForAsyncValidation_andReturnsResolvedStatus() throws Exception {
```
lubomirw: **[Nit]** | Testing

This test (and `returnsNotChecked_whenBudgetExhausted`, plus `PollFeatureTest.ridesThroughPendingIssue...`) uses real wall-clock sleeps (250ms–1s), so the suite spends real seconds and couples correctness to timing rather than observable behavior. core `mvn verify` is already ~52min.

_Suggested:_ inject the sampling interval (or a `Clock`/sleeper) so the flip and budget-exhaustion are driven deterministically without real sleeps.

src/test/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandlerTest.java
```diff
@@ -0,0 +1,303 @@
...
+import java.util.Base64;
+import java.util.Date;
+import java.util.List;
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * Unit tests for {@link CrmfMessageHandler#loadCaCertificateChain} (issue #1830).
```
lubomirw, line 66: **[Low]** | Maintainability

Rotting issue reference in the class Javadoc — `(issue #1830)` carries no behavioral information and rots (core/CLAUDE.md: comments referring to PR/issue numbers rot fast).

_Suggested:_ drop the parenthetical: `Unit tests for {@link CrmfMessageHandler#loadCaCertificateChain}.`

<sub>flagged by hygiene</sub>

src/test/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandlerTest.java
```diff
@@ -0,0 +1,303 @@
...
+                .thenReturn(CertificateValidationStatus.VALID);
+
+        List<X509Certificate> chain = invokeLoadCaCertificateChain(leaf);
+
+        assertThat(chain).hasSize(1);
+    }
+
+    @Test
+    void accepts_whenStillNotCheckedAfterWait() throws NotFoundException {
+        // NOT_CHECKED is a transient state, not a verdict (issue #1830 expected behavior):
```
lubomirw: **[Low]** | Maintainability

Inline comment embeds `(issue #1830 expected behavior)`; the rationale stands on its own without the ref.

_Suggested:_ keep the WHY, drop the ref: `// NOT_CHECKED is a transient state, not a verdict:`

<sub>flagged by hygiene</sub>

src/test/java/com/otilm/core/service/scep/impl/ScepServiceImplValidationStatusTest.java
```diff
@@ -0,0 +1,135 @@
...
+import org.springframework.test.util.ReflectionTestUtils;
+
+import java.lang.reflect.UndeclaredThrowableException;
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * Unit tests for {@link ScepServiceImpl#checkCertificateValidity} (issue #1834).
```
lubomirw: **[Low]** | Maintainability

Rotting issue reference `(issue #1834)` in the class Javadoc.

_Suggested:_ drop the parenthetical: `Unit tests for {@link ScepServiceImpl#checkCertificateValidity}.`

<sub>flagged by hygiene</sub>

src/test/java/com/otilm/core/service/scep/impl/ScepServiceImplValidationStatusTest.java
```diff
@@ -0,0 +1,135 @@
...
+        when(validationStatusPoller.pollValidationStatus(eq(dto.getUuid()), anyLong()))
+                .thenReturn(CertificateValidationStatus.VALID);
+
+        assertThatCode(() -> ReflectionTestUtils.invokeMethod(service, "checkCertificateValidity", dto))
+                .doesNotThrowAnyException();
+    }
+
+    @Test
+    void accepts_whenStillNotCheckedAfterWait() throws NotFoundException {
+        // NOT_CHECKED is a transient state, not a verdict (issue #1834 expected behavior):
```
lubomirw: **[Low]** | Maintainability

Inline comment embeds `(issue #1834 expected behavior)`.

_Suggested:_ keep the WHY, drop the ref: `// NOT_CHECKED is a transient state, not a verdict:`

<sub>flagged by hygiene</sub>

src/test/java/com/otilm/core/service/cmp/configurations/variants/Mobile3gppProfileContextTest.java
```diff
@@ -0,0 +1,112 @@
...
+    private Mobile3gppProfileContext context() {
+        // validateOnCrmfResponse reads only the response argument, not constructor state,
+        // so the collaborators can be null for this unit test.
+        return new Mobile3gppProfileContext(null, null, null, null, null, null);
+    }
+
+    @Test
+    void acceptsWaitingResponse_withoutExtraCerts() {
+        PKIMessage waiting = certRepMessage(PKIBody.TYPE_INIT_REP,
+                new CertResponse(new org.bouncycastle.asn1.ASN1Integer(0), new PKIStatusInfo(PKIStatus.waiting)),
```
lubomirw: **[Nit]** | Maintainability

`org.bouncycastle.asn1.ASN1Integer` is used fully-qualified inline here (also lines 63 and 78); the simple name is collision-free in this file.

_Suggested:_ add `import org.bouncycastle.asn1.ASN1Integer;` and use the simple name `ASN1Integer` at all three sites.

<sub>flagged by hygiene</sub>

src/test/java/com/otilm/core/service/cmp/message/handler/PollFeatureTest.java
```diff
@@ -71,10 +81,34 @@ void returnsStillPending_whenCertReachesPendingIssue() throws Exception {
...
 
     @Test
-    void returnsStillPending_whenCertReachesPendingRevoke() throws Exception {
+    void ridesThroughPendingIssue_andReturnsReached_whenStateFlipsWithinBudget() throws Exception {
+        // The actions listener parks the cert in PENDING_ISSUE for the duration of the
+        // connector call — even a synchronously-completing connector transits this state.
+        // The poll must keep sampling instead of giving up on the first PENDING_ISSUE read.
+        UUID certUuid = UUID.randomUUID();
+        Certificate cert = certificateInState(certUuid, CertificateState.PENDING_ISSUE);
+        java.util.concurrent.atomic.AtomicInteger reads = new java.util.concurrent.atomic.AtomicInteger();
```
lubomirw: **[Nit]** | Maintainability

`java.util.concurrent.atomic.AtomicInteger` is used fully-qualified inline for both declaration and instantiation; the simple name is collision-free.

_Suggested:_ add `import java.util.concurrent.atomic.AtomicInteger;` and use `AtomicInteger reads = new AtomicInteger();`

<sub>flagged by hygiene</sub>

klaraf755: Fixed.

src/test/java/com/otilm/core/service/cmp/configurations/variants/Mobile3gppProfileContextTest.java
klaraf755: Fixed.

src/test/java/com/otilm/core/service/scep/impl/ScepServiceImplValidationStatusTest.java
klaraf755: Fixed

src/test/java/com/otilm/core/service/scep/impl/ScepServiceImplValidationStatusTest.java
klaraf755: Fixed.

src/test/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandlerTest.java
klaraf755: Fixed.

src/test/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandlerTest.java
klaraf755: Fixed.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
klaraf755: Moved the method to `CertificateValidationStatusPoller`.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
klaraf755: The annotation is present because of SonarQube flagging it as an issue otherwise.

src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
klaraf755: The annotation is present because of SonarQube flagging it as an issue otherwise.

src/main/java/com/otilm/core/service/cmp/message/handler/CrmfMessageHandler.java
klaraf755: Fixed. The NOT_CHECKED tolerance is now leaf-only.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
klaraf755: Fixed, same shape as the CMP path.

src/main/java/com/otilm/core/service/handler/CertificateValidationStatusPoller.java
klaraf755: Fixed. The wait was extracted into `CertificateValidationStatusPoller.resolveOrKeep`, which is annotated `@Transactional(propagation = Propagation.NOT_SUPPORTED)` — the caller's transaction is suspended for the duration of the wait, so no pooled connection is pinned across the sleeps.

src/test/java/com/otilm/core/service/handler/CertificateValidationStatusPollerTest.java
klaraf755: Kept real sleeps but shrank them so wall-clock cost is negligible.

src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
klaraf755: Valid, and distinct from the validation-status wait (that one is now NOT_SUPPORTED). Investigating this: the transaction that pins a connection across the poll is owned at the protocol entry — `CmpServiceImpl` is class-level `@Transactional` and wraps the whole request (profile load → validation → handler dispatch → response). So the hold isn't in the handler or `PollFeature`; moving those to `NOT_SUPPORTED `only suspends the entry tx, which keeps its connection parked rather than freeing it. Properly fixing it means changing the boundary at `CmpServiceImpl`, which removes whole-request atomicity for all CMP operations — larger than this PR's validation-gate scope.

Tracking as a follow-up.

src/test/java/com/otilm/core/service/cmp/message/handler/PollFeatureTest.java
```diff
@@ -10,7 +10,9 @@
 import org.bouncycastle.asn1.DEROctetString;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.mockito.Mockito;
+import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
```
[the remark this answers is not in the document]
klaraf755: Removed.

src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
```diff
@@ -100,17 +105,20 @@ public PollResult pollCertificate(ASN1OctetString tid, String serialNumber, Stri
...
                 if (isDivergentTerminal(current, expectedState)) {
                     log.warn("TID={}, SN={} | certificate uuid={} diverted to {} while waiting for {}",
                             tid, serialNumber, certUUID, current, expectedState);
                     return new PollResult.Diverted(current);
                 }
                 if (System.currentTimeMillis() - startRequest >= timeoutMs) {
+                    // PENDING_* is ridden out for the whole budget: even a synchronously-
+                    // completing connector transits PENDING_ISSUE on the actions-listener
+                    // thread, so only budget exhaustion makes "still pending" a verdict.
+                    if (current == CertificateState.PENDING_ISSUE || current == CertificateState.PENDING_REVOKE) {
```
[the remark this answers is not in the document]
klaraf755: Fixed — the loop now clamps the sleep to the remaining budget.

lubomirw (approved):

PR #1840, opened 2026-07-17
Fix CMP revocation falsely rejecting as diverted-to-ISSUED

### LINKED ISSUE #1833: Protocols - CMP: RR (Revocation Request) returns "rejection" to client while backend actually completes the revocation successfully - same async-state race as #1830 but on the revocation flow
### Description

A CMP **Revocation Request** (`rr`) can end in the same broken state as observed on issuance (#1830), but on the opposite side of the certificate lifecycle: **the revocation is actually performed on the platform** (certificate later reaches `State: Revoked`), but the CMP `RP` response returned to the client is `rejection`, so the requesting client incorrectly concludes that the revocation failed.

The race is between (a) the backend state transition to `REVOKED`, and (b) the construction of the CMP `RP` response. If (b) reads the certificate state at a moment when (a) has not yet applied the `REVOKED` transition, the CMP handler observes the current state as `ISSUED`, decides that the revocation is "no longer in progress", and returns `PKIStatus: rejection`, `PKIFailureInfo: systemFailure` to the client. A short moment later the state transition completes and the cert is properly `Revoked` — but the CMP response has already been finalised as a rejection.

Observed rejection message returned by the platform:

```
CMP info: sending RR
CMP info: received RP
CMP error: request rejected by server:
  PKIStatus: rejection
  PKIFailureInfo: systemFailure
  StatusString: "(TID=#…) SN=… | certificate diverted to ISSUED
                 while waiting for REVOKED — revocation no longer in progress"
```

Meanwhile the same cert appears in the Certificate Inventory as `State: Revoked`, `Validation Status: Valid`.

### Steps to Reproduce

1. Have a certificate previously issued through the platform (via CMP, ACME, or manual issue against the same RA Profile)
2. From an OpenSSL 3.x client, send `rr` protected by the same CMP Profile PBM shared secret:

   ```
   openssl cmp -cmd rr \
     -server https://<host>/api/v1/protocols/cmp/<cmpProfileName> \
     -recipient "/CN=CMP" \
     -ref "<any>" \
     -secret 'pass:<shared-secret>' \
     -oldcert <cert.pem> \
     -revreason 5 \
     -poll_count 5 -total_timeout 60
   ```
3. Observe on the client: `RP` received with `rejection`, StatusString `certificate diverted to ISSUED while waiting for REVOKED — revocation no longer in progress`
4. Observe on the platform: the same certificate reaches `State: Revoked` (and appears in the CRL in the underlying CA) shortly after — i.e. the revocation succeeded

### Expected Behavior

- The CMP handler waits for the backend state transition to reach `REVOKED` before constructing the `RP` response — using a bounded wait / polling loop consistent with `-poll_count` / `-total_timeout` on the client side
- The `RP` returned to the client accurately reflects the terminal outcome:
  - `accepted` if the revocation completed (cert reached `REVOKED`)
  - `rejection` only if the revocation genuinely failed (e.g. permission denied, cert unknown, invalid revocation reason)
- No "diverted to ISSUED / no longer in progress" pseudo-error while the backend is still transitioning
- Client's view of the revocation outcome matches the platform's view

### Actual Behavior

- Cert becomes `Revoked` on the platform
- CMP handler returns `PKIStatus: rejection` / `PKIFailureInfo: systemFailure` with StatusString `certificate diverted to ISSUED while waiting for REVOKED — revocation no longer in progress`
- CMP client treats the revocation as failed even though the platform revoked the cert
- Silent divergence between client and platform state; risk of unnecessary retries or of a client continuing to use / trust a cert that has actually been revoked

### Impact / Production Consequences

- **CMP revocation flow unusable end-to-end** — clients that automate revocation cannot get a reliable success signal
- **Silent state divergence** — client believes cert is still valid; platform knows it is revoked; downstream trust decisions on the client side become unsafe
- **False retries** — clients that retry on rejection may hit follow-up errors (cert already revoked, etc.), increasing operator load
- Same class of race as #1830 (IR/CR flow returning rejection while cert is issued successfully) — fix likely benefits from a shared solution

### Severity

Major

### Module

Protocols

### Attachments

_No response_

### Logs

_No response_

### Environment

ILM Core: 2.18.1-SNAPSHOT (16 Jul 2026, 08:54)
Commit: 5f7bc88
Branch: main
Build time: 2026-07-16T06:54:47Z
ILM Frontend: 2.18.0 (16 Jul 2026, 10:25)
Client:
- macOS Tahoe 26.5.1

klaraf755:
Closes #1833 

**Operational note:** a genuine cancel-revoke that returns the certificate to `ISSUED` is now rejected via poll timeout (up to `cmp.protocol.poll.feature.timeout`, default 10 s) instead of an immediate `Diverted` rejection. The client-visible failure (`PKIFailureInfo.systemFailure`) is unchanged; only latency in that rare case increases.

### DIFF
--- a/src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
+++ b/src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
@@ -141,11 +141,15 @@ public class PollFeature {
      * (e.g. an operator cancel flipped an in-flight issue to FAILED). The one exception is
      * {@code ISSUED} while waiting for {@code REVOKED}: {@code ISSUED} is the resting state a
      * certificate occupies <em>before</em> a revocation transition lands, so observing it
-     * means "revocation not applied yet", not a divergence. Treating it as divergence made a
-     * revocation poll reject on its first sample before the async {@code ISSUED -> REVOKED}
-     * transition completed (issue #1833). Such a poll instead keeps waiting and, if the
-     * transition never lands within the budget, ends as a timeout — a rejection either way,
-     * but never a false "diverted to ISSUED".
+     * means "revocation not applied yet", not a divergence — treating it as one would reject
+     * a revocation poll on its first sample, before the async {@code ISSUED -> REVOKED}
+     * transition can land. Such a poll instead keeps waiting and, if the transition never
+     * lands within the budget, ends as a timeout — a rejection either way, but never a false
+     * "diverted to ISSUED".
+     *
+     * <p>The {@code current == expectedState} guard is unreachable from the poll loop (which
+     * returns {@code Reached} before calling this) but keeps the predicate correct on its
+     * own terms: the expected state is never divergent.</p>
      */
     private static boolean isDivergentTerminal(CertificateState current, CertificateState expectedState) {
         if (!isTerminal(current) || current == expectedState) {
--- a/src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
+++ b/src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
@@ -105,7 +105,7 @@ public class PollFeature {
                             tid, serialNumber, certUUID, current);
                     return new PollResult.StillPending(current);
                 }
-                if (isTerminal(current)) {
+                if (isDivergentTerminal(current, expectedState)) {
                     log.warn("TID={}, SN={} | certificate uuid={} diverted to {} while waiting for {}",
                             tid, serialNumber, certUUID, current, expectedState);
                     return new PollResult.Diverted(current);
@@ -134,5 +134,23 @@ public class PollFeature {
                 || state == CertificateState.REVOKED
                 || state == CertificateState.FAILED
                 || state == CertificateState.REJECTED;
+    }
+
+    /**
+     * A terminal state that is not the expected one usually means the operation diverged
+     * (e.g. an operator cancel flipped an in-flight issue to FAILED). The one exception is
+     * {@code ISSUED} while waiting for {@code REVOKED}: {@code ISSUED} is the resting state a
+     * certificate occupies <em>before</em> a revocation transition lands, so observing it
+     * means "revocation not applied yet", not a divergence. Treating it as divergence made a
+     * revocation poll reject on its first sample before the async {@code ISSUED -> REVOKED}
+     * transition completed (issue #1833). Such a poll instead keeps waiting and, if the
+     * transition never lands within the budget, ends as a timeout — a rejection either way,
+     * but never a false "diverted to ISSUED".
+     */
+    private static boolean isDivergentTerminal(CertificateState current, CertificateState expectedState) {
+        if (!isTerminal(current) || current == expectedState) {
+            return false;
+        }
+        return !(expectedState == CertificateState.REVOKED && current == CertificateState.ISSUED);
     }
 }
--- a/src/test/java/com/otilm/core/service/cmp/message/handler/PollFeatureTest.java
+++ b/src/test/java/com/otilm/core/service/cmp/message/handler/PollFeatureTest.java
@@ -18,6 +18,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.when;
 
 /**
  * Unit tests for {@link PollFeature}.
@@ -61,7 +62,7 @@ class PollFeatureTest {
     void returnsStillPending_whenCertReachesPendingIssue() throws Exception {
         UUID certUuid = UUID.randomUUID();
         Certificate cert = certificateInState(certUuid, CertificateState.PENDING_ISSUE);
-        Mockito.when(certificateService.getCertificateEntity(Mockito.any(SecuredUUID.class)))
+        when(certificateService.getCertificateEntity(Mockito.any(SecuredUUID.class)))
                 .thenReturn(cert);
 
         PollResult result = pollFeature.pollCertificate(
@@ -75,7 +76,7 @@ class PollFeatureTest {
     void returnsStillPending_whenCertReachesPendingRevoke() throws Exception {
         UUID certUuid = UUID.randomUUID();
         Certificate cert = certificateInState(certUuid, CertificateState.PENDING_REVOKE);
-        Mockito.when(certificateService.getCertificateEntity(Mockito.any(SecuredUUID.class)))
+        when(certificateService.getCertificateEntity(Mockito.any(SecuredUUID.class)))
                 .thenReturn(cert);
 
         PollResult result = pollFeature.pollCertificate(
@@ -89,7 +90,7 @@ class PollFeatureTest {
     void returnsReached_whenCertAlreadyInExpectedState() throws Exception {
         UUID certUuid = UUID.randomUUID();
         Certificate cert = certificateInState(certUuid, CertificateState.ISSUED);
-        Mockito.when(certificateService.getCertificateEntity(Mockito.any(SecuredUUID.class)))
+        when(certificateService.getCertificateEntity(Mockito.any(SecuredUUID.class)))
                 .thenReturn(cert);
 
         PollResult result = pollFeature.pollCertificate(
@@ -106,7 +107,7 @@ class PollFeatureTest {
         // Caller must reject the operation cleanly rather than time out with systemFailure.
         UUID certUuid = UUID.randomUUID();
         Certificate cert = certificateInState(certUuid, CertificateState.FAILED);
-        Mockito.when(certificateService.getCertificateEntity(Mockito.any(SecuredUUID.class)))
+        when(certificateService.getCertificateEntity(Mockito.any(SecuredUUID.class)))
                 .thenReturn(cert);
 
         PollResult result = pollFeature.pollCertificate(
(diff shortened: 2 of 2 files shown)


src/test/java/com/otilm/core/service/cmp/message/handler/PollFeatureTest.java
```diff
@@ -132,19 +133,59 @@ void returnsDiverted_whenCertEndsInRejectedButExpectedIssued() throws Exception
...
         UUID certUuid = UUID.randomUUID();
         Certificate cert = certificateInState(certUuid, CertificateState.ISSUED);
         Mockito.when(certificateService.getCertificateEntity(Mockito.any(SecuredUUID.class)))
                 .thenReturn(cert);
 
+        assertThatThrownBy(() -> pollFeature.pollCertificate(
+                new DEROctetString(new byte[]{1}), "01", certUuid.toString(), CertificateState.REVOKED))
+                .isInstanceOf(CmpProcessingException.class)
+                .hasMessageContaining("polling timed out");
+    }
```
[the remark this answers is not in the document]
klaraf755: Added.

src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
```diff
@@ -105,7 +105,7 @@ public PollResult pollCertificate(ASN1OctetString tid, String serialNumber, Stri
                             tid, serialNumber, certUUID, current);
                     return new PollResult.StillPending(current);
                 }
-                if (isTerminal(current)) {
+                if (isDivergentTerminal(current, expectedState)) {
```
lubomirw, line 108: **[Nit]** | Performance

A genuine cancel-revoke that flips the cert back to `ISSUED` no longer returns `Diverted` here immediately — the poll now rides the full budget (default 10s), holding the CMP request thread, then rejects via timeout. This is deliberate and documented: `rejectIfNotReached` maps both `Diverted` and the timeout to the same `PKIFailureInfo.systemFailure`, so the client outcome is unchanged; only added latency in the rare genuine-divergence case.

_Suggested:_ Not a bug — worth a line in the PR/ops notes so reviewers know revoke rejections can now take up to the full budget before rejecting.

<sub>flagged by guided-pr-review; trade-off independently confirmed by Copilot CLI</sub>

src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
```diff
@@ -135,4 +135,22 @@ private static boolean isTerminal(CertificateState state) {
...
     }
+
+    /**
+     * A terminal state that is not the expected one usually means the operation diverged
+     * (e.g. an operator cancel flipped an in-flight issue to FAILED). The one exception is
+     * {@code ISSUED} while waiting for {@code REVOKED}: {@code ISSUED} is the resting state a
+     * certificate occupies <em>before</em> a revocation transition lands, so observing it
+     * means "revocation not applied yet", not a divergence. Treating it as divergence made a
+     * revocation poll reject on its first sample before the async {@code ISSUED -> REVOKED}
+     * transition completed (issue #1833). Such a poll instead keeps waiting and, if the
```
lubomirw: **[Low]** | Maintainability

This Javadoc mixes sound WHY rationale with a past-tense sentence narrating the old buggy behavior plus an `(issue #1833)` tracker reference. Issue/PR refs in comments rot (repo `CLAUDE.md` discourages them), and the history narration isn't code rationale.

_Suggested:_ Drop the `(issue #1833)` ref and the "Treating it as divergence made … completed" history sentence; keep the invariant explanation (ISSUED is the resting state a cert occupies before a revocation transition lands). Judgment call — issue-as-WHY is borderline defensible for a regression invariant.

<sub>flagged by Hygiene</sub>

src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
```diff
@@ -135,4 +135,22 @@ private static boolean isTerminal(CertificateState state) {
...
+     * transition never lands within the budget, ends as a timeout — a rejection either way,
+     * but never a false "diverted to ISSUED".
+     */
+    private static boolean isDivergentTerminal(CertificateState current, CertificateState expectedState) {
+        if (!isTerminal(current) || current == expectedState) {
```
lubomirw, line 155: **[Nit]** | Maintainability

The `current == expectedState` short-circuit can never be true from the sole caller — the poll loop already returns `Reached` when `expectedState.equals(current)` before this helper is reached. Harmless defensive check.

_Suggested:_ Optionally drop it, or add a brief "defensive" note so future readers don't puzzle over the unreachable branch. Fine as-is.

src/test/java/com/otilm/core/service/cmp/message/handler/PollFeatureTest.java
```diff
@@ -132,25 +134,69 @@ void returnsDiverted_whenCertEndsInRejectedButExpectedIssued() throws Exception
     }
 
     @Test
-    void returnsDiverted_whenCertEndsInIssuedButExpectedRevoked() throws Exception {
-        // Equivalent race in the revoke direction: a concurrent cancelPendingCertificateOperation
-        // (cancel-revoke) sets the cert back to ISSUED while this poll waits for REVOKED.
+    void ridesThroughIssued_andReturnsReached_whenRevocationLandsWithinBudget() throws Exception {
```
lubomirw, line 137: **[Nit]** | Testing

This test needs a 2nd poll iteration to observe the REVOKED flip, so it sleeps one `POLL_INTERVAL_MS` (1000ms, a hardcoded constant) per run — a real ~1s wall-clock cost added to the unit suite.

_Suggested:_ Non-blocking. If suite speed matters, make `POLL_INTERVAL_MS` injectable/overridable so the interval can be shortened in tests, mirroring how `pollFeatureTimeout` is set via reflection.

src/test/java/com/otilm/core/service/cmp/message/handler/PollFeatureTest.java
```diff
@@ -132,25 +134,69 @@ void returnsDiverted_whenCertEndsInRejectedButExpectedIssued() throws Exception
...
+        // Issue #1833: ISSUED is the resting state a certificate occupies before a revocation
```
lubomirw: **[Low]** | Maintainability

Comment opens with the `Issue #1833:` tracker prefix, which rots (repo `CLAUDE.md` discourages issue/PR refs in comments). The remaining rationale is valid.

_Suggested:_ Drop the `Issue #1833: ` prefix; keep the explanation of ISSUED being the resting state before the async `ISSUED -> REVOKED` transition. Same judgment call as the Javadoc ref.

<sub>flagged by Hygiene</sub>

lubomirw (commented):
Automated multi-reviewer triage — 6 findings (COMMENT). Sources: Copilot CLI, guided-pr-review, superpowers code-review, hygiene — deduped. The fix itself reads as correct and well-scoped; these are documentation/wording/nit observations.

Two findings anchor to **unchanged context lines** (not on the diff), so they can't be inline and are listed here:

- **[Medium] | Documentation — `PollFeature.java:55-59`**: The class Javadoc's `Diverted` bullet still says the cert diverts on any terminal state "one of `ISSUED`, `REVOKED`, `FAILED`, `REJECTED` that is not the expected one" — now contradicted by this fix, where `ISSUED` while awaiting `REVOKED` is intentionally exempt. Add a clause noting the ISSUED-while-awaiting-REVOKED exception (cross-ref `isDivergentTerminal`) so the contract doc matches behavior. _(guided-pr-review + superpowers)_
- **[Low] | Error handling — `PollFeature.java:115`**: A revoke poll stuck at `ISSUED` now reaches this timeout branch, whose message hardcodes `"cert is in transitional state %s"` — but `ISSUED` is terminal, so the `StatusString` returned to the CMP client is inaccurate on this new path. Reword to not assert "transitional", e.g. `"polling timed out after %d ms — cert is in state %s, expected %s"`. _(guided-pr-review + superpowers)_

src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
klaraf755: Deliberate, as noted — the mapping to `systemFailure` is identical for `Diverted` and timeout, so only latency changes in the genuine-divergence case. Added an operational note to the PR description covering the up-to-full-budget rejection latency.

src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
klaraf755: Dropped the `(issue #1833)` ref and reworded the history sentence into present-tense rationale ("treating it as one would reject a revocation poll on its first sample…"). The invariant explanation is kept.

src/main/java/com/otilm/core/service/cmp/message/handler/PollFeature.java
klaraf755: Kept the guard: it is unreachable from the poll loop today, but without it `isDivergentTerminal(X, X)` would return true for a terminal state equal to the expected one — wrong for the predicate on its own terms, and a trap if it ever gains a second caller. Added a Javadoc note stating exactly that so the branch does not read as an oversight.

src/test/java/com/otilm/core/service/cmp/message/handler/PollFeatureTest.java
klaraf755: Leaving as-is. Shortening the interval would mean turning the production constant into an instance field solely so tests can override it — reshaping production code for test speed is not worth one second of suite time. If the interval ever becomes genuinely variable, injectability can come with that change.

src/test/java/com/otilm/core/service/cmp/message/handler/PollFeatureTest.java
klaraf755: Dropped the prefix; the resting-state rationale is kept.

lubomirw (approved):

PR #1841, opened 2026-07-17
Support optional owner and groups on certificate registration

### LINKED ISSUE #1835: Support optional owner and groups on certificate registration
### Description

Certificate registration has no option to set the certificate owner and groups. Add owner and groups as optional inputs on registration, backed by optional properties on the corresponding registration DTO. Honor them at issuance so that when a registered certificate is issued, existing owner/groups are not overwritten if no values are provided.

### Steps to Reproduce

1. Register a certificate (via API / registration flow).
2. Note there is no field/property to set owner or groups.
3. Issue the registered certificate.
4. Owner/groups cannot be preset at registration.

### Expected Behavior

Registration exposes optional owner and groups inputs, present as optional properties on the registration DTO. At issuance, owner/groups set during registration are preserved and not overwritten when no values are provided.

### Actual Behavior

No owner/groups option exists at registration; the DTO lacks these properties; issuance overwrites owner/groups regardless of registration values.

### Severity

Minor

### Module

Certificates

vyskocilm:
Closes #1835

## What

Certificate registration can now preset the certificate **owner** and **groups**, and those values are preserved when the pre-registered certificate is later issued.

- New `applyRegistrationAssociations(Certificate, ClientCertificateRegistrationDto)` helper in `ClientOperationServiceImpl`, called in both registration paths (connector-backed `registerCertificate` and `registerPlatformLevel`).
- Explicit `ownerUuid` → set as owner; when absent → `setOwnerFromProfile` (the registering user becomes owner, mirroring the plain `POST /issue` path).
- Non-empty `groupUuids` → replace the certificate's group set.
- Applied through the existing `ResourceObjectAssociationService`, reusing the `setProtocolAssociations` precedent.

## Placement / failure semantics

The helper runs inside the registration `try` block, **after** custom-attribute persistence and **before** the connector `register()` call. An invalid owner/group UUID (`NotFoundException`) or a malformed owner UUID string (`IllegalArgumentException`) therefore fails the placeholder **pre-acceptance** (→ `FAILED`), never diverging from a connector that has already accepted the registration.

## Issuance preservation

No production change needed: the register→issue path (`issueExistingCertificate` → `addCertificateRequestToExisting` → `issueRequestedCertificate`) reuses the same entity and never resets owner/groups. A regression test locks this in.

## Tests

5 new tests in `ClientOperationServiceRegisterITest` (explicit owner+groups on connector and platform-level paths; default-to-registering-user when omitted; unknown group fails the placeholder pre-acceptance; issuance of a registered cert preserves owner+groups). Full class 74/74 green.

## Dependency

Requires interfaces PR OmniTrustILM/interfaces#781 (adds `ownerUuid`/`groupUuids` to the DTO). Merge/publish that SNAPSHOT before this PR's CI.

## Review

Independent review pass (Claude reviewer + Copilot CLI) — both returned SHIP; no correctness, race, state-divergence, transaction-boundary, or information-leak issues.

### DIFF
--- a/src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
@@ -51,6 +51,7 @@ import com.otilm.core.logging.LoggerWrapper;
 import com.otilm.core.messaging.jms.producers.ActionProducer;
 import com.otilm.core.messaging.jms.producers.EventProducer;
 import com.otilm.core.messaging.model.ActionMessage;
+import com.otilm.core.service.ResourceObjectAssociationService;
 import com.otilm.core.service.handler.authority.AdapterOperationOutcome;
 import com.otilm.core.service.handler.authority.AdapterOperationResult;
 import com.otilm.core.service.handler.authority.AsyncOperationCapability;
@@ -164,6 +165,12 @@ public class ClientOperationServiceImpl implements ClientOperationExternalServic
     private RegistrationChallengeStore registrationChallengeStore;
     private CertificateRegistrationAuthorizationRepository registrationAuthorizationRepository;
     private CertificateRegistrationAuthorizationWriter registrationAuthorizationWriter;
+    private ResourceObjectAssociationService objectAssociationService;
+
+    @Autowired
+    public void setObjectAssociationService(ResourceObjectAssociationService objectAssociationService) {
+        this.objectAssociationService = objectAssociationService;
+    }
 
     @Autowired
     public void setProtocolRequestAttributeValidator(ProtocolRequestAttributeValidator protocolRequestAttributeValidator) {
@@ -480,6 +487,9 @@ public class ClientOperationServiceImpl implements ClientOperationExternalServic
             if (createCustomAttributes && request.getCustomAttributes() != null && !request.getCustomAttributes().isEmpty()) {
                 attributeEngine.updateObjectCustomAttributesContent(Resource.CERTIFICATE, certificate.getUuid(), request.getCustomAttributes());
             }
+            // Apply owner/groups before the connector call, alongside custom attributes: an invalid owner/group
+            // UUID fails the placeholder here (pre-acceptance) rather than after a connector has already accepted.
+            applyRegistrationAssociations(certificate, request);
             // Persist the challenge authorization before the connector call — on every arc that can leave the
             // certificate reconcilable to REGISTERED (see the method) — but inside the try, so a store/save
             // failure fails the placeholder via the catch below rather than orphaning it in PENDING_REGISTRATION.
@@ -563,6 +573,7 @@ public class ClientOperationServiceImpl implements ClientOperationExternalServic
             if (createCustomAttributes && request.getCustomAttributes() != null && !request.getCustomAttributes().isEmpty()) {
                 attributeEngine.updateObjectCustomAttributesContent(Resource.CERTIFICATE, certificate.getUuid(), request.getCustomAttributes());
             }
+            applyRegistrationAssociations(certificate, request);
             maybeCreateRegistrationAuthorization(certificate, request);
             // Write the register->issue binding (no connector meta) so a platform-level placeholder carries the same
             // marker as a connector-backed one — the discriminator both the issue routing and the approval-reject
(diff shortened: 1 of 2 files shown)


src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -585,6 +596,32 @@ private ClientCertificateDataResponseDto registerPlatformLevel(RaProfile raProfi
...
+     * <p>Called before the connector call (alongside custom attributes) so an invalid owner/group UUID fails the
+     * placeholder pre-acceptance rather than diverging from a connector that has already accepted the
+     * registration. The values are preserved when the pre-registered certificate is later issued, because the
+     * register-&gt;issue path reuses the same entity and does not reset owner or groups.</p>
+     */
+    private void applyRegistrationAssociations(Certificate certificate, ClientCertificateRegistrationDto request)
+            throws NotFoundException {
+        UUID certificateUuid = certificate.getUuid();
+        if (request.getOwnerUuid() != null && !request.getOwnerUuid().isBlank()) {
+            objectAssociationService.setOwner(Resource.CERTIFICATE, certificateUuid, UUID.fromString(request.getOwnerUuid()));
```
lubomirw, line 623: **[Medium]** | Error handling

`request.getOwnerUuid()` is an unvalidated `String`. A non-blank but malformed value reaches `UUID.fromString(...)` here, throws `IllegalArgumentException`, is caught by the `RuntimeException` clause, and transitions the just-created placeholder to FAILED — surfacing as a generic 500 ("Invalid UUID string") rather than a clean 4xx. Asymmetric with `groupUuids` (`Set<UUID>`, rejected cleanly at deserialization), and it litters a FAILED certificate row.

_Suggested:_ Validate the owner-id format at the boundary (a UUID `@Pattern` or `UUID` type on the DTO field), or parse defensively in `applyRegistrationAssociations` and convert `IllegalArgumentException` into a `ValidationException`, so a malformed `ownerUuid` yields a 4xx without creating a FAILED placeholder.

<sub>flagged by Copilot CLI + guided-pr-review (2 sources, high confidence)</sub>

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
```diff
@@ -934,6 +955,117 @@ void issueExistingRejectsCertInNonIssuableState() {
...
+
+        UUID certUuid = UUID.fromString(response.getUuid());
+        Assertions.assertEquals(ownerUuid.toString(),
+                objectAssociationService.getOwner(Resource.CERTIFICATE, certUuid).getUuid());
+        Assertions.assertEquals(List.of(group.getUuid()),
+                objectAssociationService.getGroupUuids(Resource.CERTIFICATE, certUuid));
+    }
+
+    @Test
+    void registrationWithUnknownGroupFailsPlaceholderPreAcceptance() throws Exception {
```
lubomirw, line 1029: **[Low]** | Testing

The unknown-*group* pre-acceptance failure is covered here, but the symmetric owner cases are not: (a) a well-formed `ownerUuid` that resolves to no user (`setOwner` → `NotFoundException`), and (b) a malformed `ownerUuid` string (see the Error-handling finding on the impl). These are the paths most likely to regress given the manual `String` parsing and cross-service owner resolution. No test asserts association state on a FAILED placeholder either.

_Suggested:_ Add a test for an unresolvable explicit `ownerUuid` (assert FAILED placeholder + connector `register` never called) and one for a malformed `ownerUuid` (assert a validation error).

<sub>flagged by Copilot CLI + guided-pr-review + superpowers (3 sources)</sub>

lubomirw (commented):

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
vyskocilm: Fixed. Added a boundary guard `rejectMalformedOwnerUuid(request)` (alongside `rejectAmbiguousRegistrationIdentity`/`rejectPastRegistrationWindow`), so a malformed `ownerUuid` is rejected with a `ValidationException` (4xx) **before** the placeholder is created — no orphaned FAILED row.

Note: converting the `IllegalArgumentException` to a `ValidationException` inside `applyRegistrationAssociations` alone would not have sufficed — `ValidationException` is a `RuntimeException` and would still be caught by the pre-acceptance catch that transitions the placeholder to FAILED. Validating at the boundary (pre-placeholder) is what yields the clean 4xx with no FAILED row. `UUID.fromString` in `applyRegistrationAssociations` stays, now guaranteed well-formed.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
vyskocilm: Added both. In `ClientOperationServiceRegisterITest`:

- `registrationWithUnresolvableOwnerFailsPlaceholderPreAcceptance` — well-formed `ownerUuid` resolving to no user (`setOwner` → `NotFoundException`); asserts FAILED placeholder and connector `register` never called.
- `registrationWithMalformedOwnerUuidIsRejectedWithoutPlaceholder` — malformed `ownerUuid`; asserts a `ValidationException` and that **no** placeholder is persisted (rejected at the boundary, see the impl thread).

PR #1843, opened 2026-07-18
Dispatch NG callbacks on parent-less connector forms

### LINKED ISSUE #1842: Dispatch NG callbacks on scope-less connector forms (authority instance)
### Description

NG (`dependsOn`) attribute callbacks fail on forms that have no parent object — notably the authority-instance create/edit form, where a GROUP attribute declares a `dependsOn` on a sibling field on the same form. With no `parentObjectUuid`, the frontend falls back to the legacy callback.

The connector-scoped route `POST /v2/connectors/{uuid}/callback` already routes NG by the resolved definition's `dependsOn`, but it stamps a both-null `connectorInterface`/`interfaceVersion` pair — nothing on that route supplies the form's interface. Because the callback envelope marks both fields required and omits nulls from the JSON, a conformant NG connector rejects the body with a 400.

Fix: carry the connector-interface identity on the callback so Core stamps the envelope correctly, and fold in the related hardening below. Depends on the interfaces change that adds an optional `interfaceUuid` to the callback request DTO.

Part of the Connector Attributes v2 API epic ilm#251.

### User Story

As a connector author, I want NG callbacks to work on parent-less forms (the authority-instance create/edit form) so a group or dropdown there resolves against the operator's in-form values, instead of the platform silently falling back to the legacy callback.

### Use Case

The authority-instance form has a GROUP attribute whose callback depends on a sibling selector field. The frontend fires it on the connector route with the interface the form was listed under; Core stamps that interface, sends the in-form values as `currentAttributes` with an empty `contextAttributes`, and the connector returns the selected branch's child attributes.

### Acceptance Criteria

- [ ] `RequestAttributeCallback` carries an optional `interfaceUuid`. On `/v2/connectors/{uuid}/callback`, when the resolved definition is NG: require `interfaceUuid` (422 if absent), load the connector-interface entity, assert it belongs to the route connector, and stamp `connectorInterface`/`interfaceVersion` from its columns; `contextAttributes = []`.
- [ ] Parallel NG interface versions on one connector (e.g. v3 and v4) resolve by the interface-row UUID the form was listed under — no version inference or heuristic.
- [ ] Completeness guard: every `dependsOn` name has a value in `currentAttributes`, else 422 and the connector is not called.
- [ ] Rename the callback service's `resourceUuid` / `scopeResourceUuid` to `parentObjectUuid` / `scopeParentUuid` — the UUID is the parent scope object, not an object of the named resource.
- [ ] The token-profile callback route uses the token-instance UUID as its parent (not the token-profile), fixing the latent 404 where the arm loads a token profile by that UUID.
- [ ] `buildEnvelope` rejects a both-null interface pair on NG dispatch (422); update the pinned both-null test. The token-profile / cryptographic-key / location arms stay non-NG (they have no connector-interface entity today) or adopt `interfaceUuid`.
- [ ] No dedicated authority resource-callback arm and no create/edit split; a self-scope authority walker is noted as the future vehicle for callbacks that need persisted object state.
- [ ] Tests: interface stamped from the row (v3); belongs-to-connector rejection; NG with missing `interfaceUuid` → 422; missing `dependsOn` value → 422; both-null NG → 422; regression on legacy and resource-scoped callbacks.

lubomirw:
Closes OmniTrustILM/core#1842. Part of the Connector Attributes v2 API epic ilm#251.

NG (`dependsOn`) callbacks failed on forms with no parent object — notably the authority-instance create-edit form, where a group/dropdown depends on a sibling field on the same form. With no parent, the frontend dispatches on the connector callback route, which stamped a both-null `connectorInterface`/`interfaceVersion` pair; a conformant NG connector rejects that body.

## Connector-route NG stamping
`POST /v2/connectors/{uuid}/callback`, when the resolved definition is NG:
- requires `interfaceUuid` (422 if absent) — nothing else on this route supplies the form's interface;
- loads the connector-interface row, asserts it belongs to the route connector (rejects a foreign row);
- stamps `connectorInterface`/`interfaceVersion` from its columns; `contextAttributes = []`.

Parallel NG interface versions on one connector (e.g. v3 and v4) resolve by the interface-row UUID — no version inference.

## Folded-in hardening
- **Completeness guard** — every `dependsOn` name must have a value in `currentAttributes`, else 422 before the connector is called.
- **buildEnvelope both-null forbid** — every NG dispatch must stamp a full interface pair; a both-null / half-pair shape fails closed.
- **Token-profile parent fix** — the token-profile callback route (route arm + scope walker) resolves its parent as the token instance, not a token profile, fixing a latent 404.
- **Rename** — `resourceUuid`/`scopeResourceUuid` → `parentObjectUuid`/`scopeParentUuid` (the UUID is the parent scope object, not an object of the named resource).

No dedicated authority resource-callback arm and no create/edit split.

## Tests
`AttributesV2CallbackDispatchITest`: interface stamped from the row (v3), belongs-to-connector rejection, missing `interfaceUuid` → 422, missing `dependsOn` value → 422, both-null NG → 422; legacy + resource-scoped regression unchanged. Unit coverage in `NgCallbackDispatcherTest` (both-null rejected) and `AttributeCallbackScopeResolverTest` (token-profile → token-instance chain).

Depends on OmniTrustILM/interfaces#782 (the `interfaceUuid` field).

### DIFF
--- a/src/main/java/com/otilm/core/service/callback/AttributeCallbackScopeResolver.java
+++ b/src/main/java/com/otilm/core/service/callback/AttributeCallbackScopeResolver.java
@@ -190,8 +190,8 @@ public class AttributeCallbackScopeResolver {
         // TOKEN_PROFILE -> [{tokenInstance}]), so the parent UUID identifies the instance, not a token profile.
         TokenInstanceReference tokenInstance = tokenInstanceReferenceRepository.findByUuid(tokenInstanceUuid)
                 .orElseThrow(() -> notFound(TokenInstanceReference.class, tokenInstanceUuid));
-        return List.of(new ScopeStep(Resource.TOKEN, tokenInstance.getUuid(),
-                tokenInstance.getConnector() == null ? null : tokenInstance.getConnector().getUuid()));
+        // Read the connectorUuid column directly rather than walking the LAZY connector association.
+        return List.of(new ScopeStep(Resource.TOKEN, tokenInstance.getUuid(), tokenInstance.getConnectorUuid()));
     }
 
     private List<ScopeStep> walkCryptographicKey(UUID tokenProfileUuid) throws NotFoundException {
--- a/src/main/java/com/otilm/core/service/callback/AttributeCallbackScopeResolver.java
+++ b/src/main/java/com/otilm/core/service/callback/AttributeCallbackScopeResolver.java
@@ -14,10 +14,12 @@ import com.otilm.core.attribute.engine.records.ObjectAttributeContentInfo;
 import com.otilm.core.dao.entity.AuthorityInstanceReference;
 import com.otilm.core.dao.entity.EntityInstanceReference;
 import com.otilm.core.dao.entity.RaProfile;
+import com.otilm.core.dao.entity.TokenInstanceReference;
 import com.otilm.core.dao.entity.TokenProfile;
 import com.otilm.core.dao.repository.AuthorityInstanceReferenceRepository;
 import com.otilm.core.dao.repository.EntityInstanceReferenceRepository;
 import com.otilm.core.dao.repository.RaProfileRepository;
+import com.otilm.core.dao.repository.TokenInstanceReferenceRepository;
 import com.otilm.core.dao.repository.TokenProfileRepository;
 import com.otilm.core.model.auth.ResourceAction;
 import com.otilm.core.security.authz.AuthorizationEnforcer;
@@ -64,6 +66,7 @@ public class AttributeCallbackScopeResolver {
     private final AuthorityInstanceReferenceRepository authorityInstanceReferenceRepository;
     private final RaProfileRepository raProfileRepository;
     private final TokenProfileRepository tokenProfileRepository;
+    private final TokenInstanceReferenceRepository tokenInstanceReferenceRepository;
     private final EntityInstanceReferenceRepository entityInstanceReferenceRepository;
 
     private final Map<Resource, ScopeWalker> walkers = new EnumMap<>(Resource.class);
@@ -71,7 +74,7 @@ public class AttributeCallbackScopeResolver {
     /** A scope-chain walker that may fail with a 404 when a referenced object is absent. */
     @FunctionalInterface
     private interface ScopeWalker {
-        List<ScopeStep> walk(UUID resourceUuid) throws NotFoundException;
+        List<ScopeStep> walk(UUID parentObjectUuid) throws NotFoundException;
     }
 
     public AttributeCallbackScopeResolver(AttributeEngine attributeEngine,
@@ -80,6 +83,7 @@ public class AttributeCallbackScopeResolver {
                                           AuthorityInstanceReferenceRepository authorityInstanceReferenceRepository,
                                           RaProfileRepository raProfileRepository,
                                           TokenProfileRepository tokenProfileRepository,
+                                          TokenInstanceReferenceRepository tokenInstanceReferenceRepository,
                                           EntityInstanceReferenceRepository entityInstanceReferenceRepository) {
         this.attributeEngine = attributeEngine;
         this.expander = expander;
@@ -87,6 +91,7 @@ public class AttributeCallbackScopeResolver {
         this.authorityInstanceReferenceRepository = authorityInstanceReferenceRepository;
         this.raProfileRepository = raProfileRepository;
         this.tokenProfileRepository = tokenProfileRepository;
+        this.tokenInstanceReferenceRepository = tokenInstanceReferenceRepository;
         this.entityInstanceReferenceRepository = entityInstanceReferenceRepository;
         registerWalkers();
     }
@@ -107,7 +112,7 @@ public class AttributeCallbackScopeResolver {
      * Resolve the ordered (parent-first) scope chain for the given scoped resource into expanded
      * {@link ScopedAttributes} blobs. An unmapped resource fails closed with a {@link ValidationException}.
      */
-    public List<ScopedAttributes> resolveScopeChain(Resource resource, UUID resourceUuid, Set<String> expandedSecrets)
+    public List<ScopedAttributes> resolveScopeChain(Resource resource, UUID parentObjectUuid, Set<String> expandedSecrets)
             throws NotFoundException, AttributeException, ConnectorException {
         ScopeWalker walker = walkers.get(resource);
         if (walker == null) {
@@ -117,7 +122,7 @@ public class AttributeCallbackScopeResolver {
                     "Callback scope chain is not supported for resource " + resource));
         }
         List<ScopedAttributes> chain = new ArrayList<>();
-        for (ScopeStep step : walker.walk(resourceUuid)) {
+        for (ScopeStep step : walker.walk(parentObjectUuid)) {
             chain.add(buildScopedAttributes(step, expandedSecrets));
         }
         return chain;
@@ -180,10 +185,13 @@ public class AttributeCallbackScopeResolver {
         return List.of(authorityStep(authority), raProfileStep(raProfile));
     }
 
-    private List<ScopeStep> walkTokenProfile(UUID tokenProfileUuid) throws NotFoundException {
-        TokenProfile tokenProfile = tokenProfileRepository.findByUuid(tokenProfileUuid)
-                .orElseThrow(() -> notFound(TokenProfile.class, tokenProfileUuid));
-        return List.of(tokenInstanceStep(tokenProfile));
+    private List<ScopeStep> walkTokenProfile(UUID tokenInstanceUuid) throws NotFoundException {
+        // The token-profile create form is scoped by its parent token INSTANCE (per the scope table
+        // TOKEN_PROFILE -> [{tokenInstance}]), so the parent UUID identifies the instance, not a token profile.
+        TokenInstanceReference tokenInstance = tokenInstanceReferenceRepository.findByUuid(tokenInstanceUuid)
+                .orElseThrow(() -> notFound(TokenInstanceReference.class, tokenInstanceUuid));
+        return List.of(new ScopeStep(Resource.TOKEN, tokenInstance.getUuid(),
+                tokenInstance.getConnector() == null ? null : tokenInstance.getConnector().getUuid()));
     }
 
     private List<ScopeStep> walkCryptographicKey(UUID tokenProfileUuid) throws NotFoundException {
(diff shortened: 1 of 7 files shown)


src/main/java/com/otilm/core/service/impl/CallbackServiceImpl.java
```diff
@@ -333,21 +366,62 @@ private Object dispatchNg(ConnectorDetailDto connector, BaseAttribute attribute,
...
+    private static void assertDependsOnNamesPresent(List<String> dependsOn, List<RequestAttribute> currentAttributes) {
+        if (dependsOn == null || dependsOn.isEmpty()) {
+            return;
+        }
+        Set<String> present = new HashSet<>();
+        if (currentAttributes != null) {
+            for (RequestAttribute current : currentAttributes) {
+                present.add(current.getName());
+            }
+        }
```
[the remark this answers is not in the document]
lubomirw: Fixed in 900f1497a — a null element (or null name) is now skipped, so the dependsOn name it was meant to satisfy stays unsatisfied and the guard fails closed (422) instead of NPE-ing into a 500.

src/main/java/com/otilm/core/service/callback/AttributeCallbackScopeResolver.java
```diff
@@ -180,10 +185,13 @@ private List<ScopeStep> walkCertificateIssuance(UUID raProfileUuid) throws NotFo
...
-        TokenProfile tokenProfile = tokenProfileRepository.findByUuid(tokenProfileUuid)
-                .orElseThrow(() -> notFound(TokenProfile.class, tokenProfileUuid));
-        return List.of(tokenInstanceStep(tokenProfile));
+    private List<ScopeStep> walkTokenProfile(UUID tokenInstanceUuid) throws NotFoundException {
+        // The token-profile create form is scoped by its parent token INSTANCE (per the scope table
+        // TOKEN_PROFILE -> [{tokenInstance}]), so the parent UUID identifies the instance, not a token profile.
+        TokenInstanceReference tokenInstance = tokenInstanceReferenceRepository.findByUuid(tokenInstanceUuid)
+                .orElseThrow(() -> notFound(TokenInstanceReference.class, tokenInstanceUuid));
+        return List.of(new ScopeStep(Resource.TOKEN, tokenInstance.getUuid(),
+                tokenInstance.getConnector() == null ? null : tokenInstance.getConnector().getUuid()));
```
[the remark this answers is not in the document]
lubomirw: Fixed in 900f1497a — reads the token instance's connectorUuid column directly instead of walking the lazy connector association.

src/main/java/com/otilm/core/service/impl/CallbackServiceImpl.java
```diff
@@ -333,21 +365,60 @@ private Object dispatchNg(ConnectorDetailDto connector, BaseAttribute attribute,
...
+        List<String> missing = dependsOn.stream().filter(name -> !present.contains(name)).toList();
+        if (!missing.isEmpty()) {
+            throw new ValidationException(ValidationError.create(
+                    "NG attribute callback is missing current values for dependsOn attributes: " + missing));
+        }
+    }
```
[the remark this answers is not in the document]
lubomirw: Deliberate. The guard is defence-in-depth behind the FE trigger rule (a dependsOn callback only fires once its named siblings are populated); presence-by-name is the intended check. Requiring non-empty content here would 422 a legitimately empty/falsy value (empty string, false, cleared optional). Comment was reworded to state presence-by-name explicitly (bf8c3f44d).

src/test/java/com/otilm/core/service/callback/AttributeCallbackScopeResolverTest.java
```diff
@@ -92,6 +95,22 @@ void certificateIssuanceAuthorizesAuthorityThenRaProfile() throws Exception {
...
 
+    @Test
+    void tokenProfileAuthorizesTokenInstance() throws Exception {
+        // The token-profile create form is scoped by its parent token INSTANCE, so the parent UUID identifies the
+        // instance; the chain is [{tokenInstance}] and only TOKEN:DETAIL is enforced.
+        UUID tokenInstanceUuid = UUID.randomUUID();
+        Connector connector = connectorWithUuid();
+        TokenInstanceReference tokenInstance = mock(TokenInstanceReference.class);
+        when(tokenInstance.getUuid()).thenReturn(tokenInstanceUuid);
+        when(tokenInstance.getConnector()).thenReturn(connector);
```
klaraf755: The walker this test covers reads `getConnectorUuid()` (the column, per the LAZY-avoidance comment in `walkTokenProfile`), not `getConnector()` — so this stub is dead and the `ScopeStep` under test carries a `null` connectorUuid, unlike production. Stub `getConnectorUuid()` instead (and drop the now-unused `connectorWithUuid()` line) so the test stays aligned with the code path it documents.

src/main/java/com/otilm/core/service/impl/CallbackServiceImpl.java
```diff
@@ -312,18 +329,34 @@ private static boolean isNgCallback(BaseAttribute attribute, AttributeCallback a
...
+
+        // Defence-in-depth behind the FE trigger rule: a dependsOn callback only fires once its named siblings are
+        // populated, so every dependsOn name must be present in currentAttributes. This checks presence by name (a
+        // value cleared to empty still counts as present); reject before touching the connector.
+        assertDependsOnNamesPresent(attributeCallback.getDependsOn(), currentAttributes);
+
+        // Stamp the envelope interface. Resource-scoped forms pass it pre-resolved from the scoped object; the
+        // connector-scoped route leaves it null and carries the form's connector-interface row UUID on the request
+        // instead. Resolving it here (rather than the entrypoint) keeps the lookup on the NG branch only.
+        if (connectorInterface == null) {
```
klaraf755, line 350: When a pre-resolved route (RA_PROFILE/CERTIFICATE with a stored `ConnectorInterfaceEntity`) also receives a request-supplied `interfaceUuid`, the request field is silently ignored — the stored interface wins. If the FE ends up always sending `interfaceUuid` (likely, since the connector route requires it), a version mismatch between the form's interface row and the authority's stored row would be invisible. Worth either validating consistency when both are present, or at minimum stating the precedence in this comment.

src/test/java/com/otilm/core/integration/service/AttributesV2CallbackDispatchITest.java
```diff
@@ -355,20 +384,16 @@ void ngScopeChainFailsClosedWhenParentDetailDenied() throws Exception {
     }
 
     @Test
-    void ngTokenProfileRouteDispatchesWithBothNullInterfaceContext() throws Exception {
-        // TOKEN_PROFILE/CRYPTOGRAPHIC_KEY/LOCATION have no stored interface version (only authorities carry one), so
-        // the arm emits a both-null interface envelope. A dependsOn callback on this route must still dispatch — the
-        // interface-version fail-fast must not trip on the both-null shape, and the envelope omits connectorInterface.
+    void ngTokenProfileRouteWithoutInterfaceUuidIsRejected() throws Exception {
```
klaraf755, line 388: The token-profile 404 fix only has its *rejection* path covered at integration level — there's no test proving that `TOKEN_PROFILE` + token-instance UUID + valid `interfaceUuid` actually dispatches, with `contextAttributes` carrying the token-instance scope. The route arm and the scope walker resolving the same UUID as an *instance* is exactly the agreement that was broken before; an end-to-end happy path (asserting the instance UUID appears in the first scope step of the dispatched envelope) would pin it.

src/test/java/com/otilm/core/service/callback/NgCallbackDispatcherTest.java
```diff
@@ -70,7 +70,8 @@ private NgCallbackDispatcher.NgDispatchContext context() {
         def.setUuid(UUID.randomUUID().toString());
         def.setName("ngAttr");
         def.setContentType(AttributeContentType.STRING);
-        return new NgCallbackDispatcher.NgDispatchContext(def, null, "v2", List.of(), List.of());
+        return new NgCallbackDispatcher.NgDispatchContext(def,
+                com.otilm.api.model.client.connector.v2.ConnectorInterface.AUTHORITY, "v3", List.of(), List.of());
```
klaraf755: Optional nit: `ConnectorInterface` could be imported instead of fully qualified — there's no name clash in this file. Same applies to the new `com.otilm.api.exception.ValidationException` occurrences in `AttributesV2CallbackDispatchITest` (lines 407/465/486/508). Both follow a pre-existing occurrence in each file, though, so feel free to resolve as style-only — but if the fully-qualified form keeps accumulating, an import cleanup touching all occurrences would be tidier.

klaraf755 (commented):
Reviewed the diff plus surrounding code. The change is correct and well-guarded — the foreign-interface ownership check, the enumeration-safe collapsed error, and the fail-closed both-null envelope guard are all solid, and the rejection-path tests verifying no connector POST fired are appreciated. Four inline notes below; nothing blocking.

src/test/java/com/otilm/core/service/callback/AttributeCallbackScopeResolverTest.java
lubomirw: Fixed in e702cc671 — stubs getConnectorUuid() and drops the now-dead getConnector()/connectorWithUuid(), so the ScopeStep carries the same connectorUuid path production uses.

src/main/java/com/otilm/core/service/impl/CallbackServiceImpl.java
lubomirw: Documented in e702cc671. Precedence stated in the comment: a pre-resolved interface (RA_PROFILE/CERTIFICATE, from the authority row) wins and a request interfaceUuid is ignored there — those forms derive the interface from the scoped object, so the FE does not send one; the request field is consulted only when nothing is pre-resolved. Kept it to a documented precedence rather than a cross-check to avoid an extra query for a case the FE contract shouldn't produce.

src/test/java/com/otilm/core/integration/service/AttributesV2CallbackDispatchITest.java
lubomirw: Added in e702cc671 — ngTokenProfileRouteDispatchesWithTokenInstanceScope: TOKEN_PROFILE + token-instance UUID + valid interfaceUuid dispatches, asserting contextAttributes[0].scope=tokens and objectUuid = the instance UUID (pins the arm/walker agreement).

src/test/java/com/otilm/core/service/callback/NgCallbackDispatcherTest.java
lubomirw: Cleaned up in e702cc671 — imported ConnectorInterface (NgCallbackDispatcherTest) and ValidationException (AttributesV2CallbackDispatchITest) and dropped the fully-qualified forms.

klaraf755 (approved):

PR #1849, opened 2026-07-21
Fix duplicate notification profile version numbers on concurrent edits

3keyroman:
## What changed

Concurrent edits of a notification profile could assign the same version number twice: `editNotificationProfile` computed the next version from the latest existing row without any serialization, and the schema had no uniqueness guarantee on `(notification_profile_uuid, version)`. Duplicate version numbers made version resolution for notifications nondeterministic.

- `editNotificationProfile` now takes a pessimistic lock on the profile row (`SELECT … FOR UPDATE`) while assigning the next version number, so concurrent edits serialize and each produces a distinct version.
- New unique constraint `uq_notification_profile_version` on `notification_profile_version (notification_profile_uuid, version)`, declared on the entity and added by a Flyway migration. The migration renumbers already-existing duplicate version numbers chronologically and remaps `pending_notification` version references accordingly.
- Recipient info is resolved before the row lock is taken, keeping the auth service call outside the lock window.
- A violation of the new unique constraint is reported as a validation error (profile was concurrently modified); other integrity violations surface unchanged.
- Description changes are persisted even when the edit does not result in a new version.

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
@@ -166,10 +166,10 @@ public class NotificationProfileServiceImpl implements NotificationProfileExtern
     @ExternalAuthorization(resource = Resource.NOTIFICATION_PROFILE, action = ResourceAction.UPDATE)
     @Transactional(propagation = Propagation.NOT_SUPPORTED)
     public NotificationProfileDetailDto editNotificationProfile(SecuredUUID uuid, NotificationProfileUpdateRequestDto updateRequestDto) throws NotFoundException {
+        validateNotificationInstanceExists(updateRequestDto.getNotificationInstanceUuid());
         // Resolve recipient info from the request before opening the write transaction: recipient lookup
         // can call the auth service over HTTP and must not hold a DB connection or the profile row lock.
         List<NameAndUuidDto> recipients = resolveRecipients(updateRequestDto.getRecipientType(), updateRequestDto.getRecipientUuids());
-        validateNotificationInstanceExists(updateRequestDto.getNotificationInstanceUuid());
 
         // The transaction boundary comes from the self-proxied call; invoking persistEditedVersion directly
         // on `this` would skip the @Transactional advice and reintroduce the version race.
--- a/src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
@@ -8,9 +8,11 @@ import com.otilm.api.model.common.NameAndUuidDto;
 import com.otilm.api.model.core.auth.Resource;
 import com.otilm.api.model.core.notification.RecipientType;
 import com.otilm.api.model.core.scheduler.PaginationRequestDto;
+import com.otilm.core.dao.entity.notifications.NotificationInstanceReference;
 import com.otilm.core.dao.entity.notifications.NotificationProfile;
 import com.otilm.core.dao.entity.notifications.NotificationProfileVersion;
 import com.otilm.core.dao.entity.workflows.Execution;
+import com.otilm.core.dao.repository.notifications.NotificationInstanceReferenceRepository;
 import com.otilm.core.dao.repository.notifications.NotificationProfileRepository;
 import com.otilm.core.dao.repository.notifications.NotificationProfileVersionRepository;
 import com.otilm.core.dao.repository.workflows.ExecutionRepository;
@@ -48,6 +50,7 @@ public class NotificationProfileServiceImpl implements NotificationProfileExtern
     private NotificationProfileServiceImpl self;
     private NotificationProfileRepository notificationProfileRepository;
     private NotificationProfileVersionRepository notificationProfileVersionRepository;
+    private NotificationInstanceReferenceRepository notificationInstanceReferenceRepository;
 
     private ExecutionRepository executionRepository;
     private ResourceObjectAssociationService resourceObjectAssociationService;
@@ -66,6 +69,11 @@ public class NotificationProfileServiceImpl implements NotificationProfileExtern
     @Autowired
     public void setNotificationProfileVersionRepository(NotificationProfileVersionRepository notificationProfileVersionRepository) {
         this.notificationProfileVersionRepository = notificationProfileVersionRepository;
+    }
+
+    @Autowired
+    public void setNotificationInstanceReferenceRepository(NotificationInstanceReferenceRepository notificationInstanceReferenceRepository) {
+        this.notificationInstanceReferenceRepository = notificationInstanceReferenceRepository;
     }
 
     @Autowired
@@ -131,6 +139,7 @@ public class NotificationProfileServiceImpl implements NotificationProfileExtern
         if (notificationProfileRepository.findByName(requestDto.getName()).isPresent()) {
             throw new AlreadyExistException("Notification profile with name " + requestDto.getName() + " already exists.");
         }
+        validateNotificationInstanceExists(requestDto.getNotificationInstanceUuid());
 
         NotificationProfile notificationProfile = new NotificationProfile();
         notificationProfile.setName(requestDto.getName());
@@ -160,7 +169,10 @@ public class NotificationProfileServiceImpl implements NotificationProfileExtern
         // Resolve recipient info from the request before opening the write transaction: recipient lookup
         // can call the auth service over HTTP and must not hold a DB connection or the profile row lock.
         List<NameAndUuidDto> recipients = resolveRecipients(updateRequestDto.getRecipientType(), updateRequestDto.getRecipientUuids());
+        validateNotificationInstanceExists(updateRequestDto.getNotificationInstanceUuid());
 
+        // The transaction boundary comes from the self-proxied call; invoking persistEditedVersion directly
+        // on `this` would skip the @Transactional advice and reintroduce the version race.
         return self.persistEditedVersion(uuid, updateRequestDto, recipients);
     }
 
@@ -205,6 +217,17 @@ public class NotificationProfileServiceImpl implements NotificationProfileExtern
         }
 
         return notificationProfileVersion.mapToDetailDto(recipients);
+    }
+
+    /**
+     * Fails fast with a not-found error for a client-supplied notification instance reference that does not
+     * exist; without this check the insert would fail the foreign key and surface as a server error. The
+     * foreign key still guards the race with a concurrent instance deletion.
+     */
+    private void validateNotificationInstanceExists(UUID notificationInstanceUuid) throws NotFoundException {
+        if (notificationInstanceUuid != null && notificationInstanceReferenceRepository.findByUuid(notificationInstanceUuid).isEmpty()) {
+            throw new NotFoundException(NotificationInstanceReference.class, notificationInstanceUuid);
+        }
     }
 
     private static boolean isUniqueVersionViolation(DataIntegrityViolationException e) {
--- a/src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
@@ -28,7 +28,9 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Pageable;
+import org.springframework.context.annotation.Lazy;
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 
 import java.util.ArrayList;
@@ -38,16 +40,23 @@ import java.util.UUID;
 import java.util.stream.Collectors;
 
 @Service
-@Transactional
+@Transactional(rollbackFor = Exception.class)
 public class NotificationProfileServiceImpl implements NotificationProfileExternalService {
 
     private static final Logger logger = LoggerFactory.getLogger(NotificationProfileServiceImpl.class);
 
+    private NotificationProfileServiceImpl self;
     private NotificationProfileRepository notificationProfileRepository;
     private NotificationProfileVersionRepository notificationProfileVersionRepository;
 
     private ExecutionRepository executionRepository;
     private ResourceObjectAssociationService resourceObjectAssociationService;
+
+    @Lazy
+    @Autowired
+    public void setSelf(NotificationProfileServiceImpl self) {
+        this.self = self;
+    }
 
     @Autowired
     public void setNotificationProfileRepository(NotificationProfileRepository notificationProfileRepository) {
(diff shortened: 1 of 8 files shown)


src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
```diff
@@ -144,19 +147,26 @@ public NotificationProfileDetailDto createNotificationProfile(NotificationProfil
...
+        // the auth service over HTTP and must not extend the lock window.
+        List<NameAndUuidDto> recipients = resolveRecipients(updateRequestDto.getRecipientType(), updateRequestDto.getRecipientUuids());
 
-        if (areVersionsEqual(currentVersion, updateRequestDto)) {
-            logger.debug("Current version of notification profile {} is same as in request. New version is not created", notificationProfile.getName());
-            return getNotificationProfileDetailDto(currentVersion);
-        }
+        // The row lock serializes concurrent edits; without it, both could read the same latest version and
+        // insert duplicate version numbers.
+        NotificationProfile notificationProfile = notificationProfileRepository.findAndLockByUuid(uuid.getValue()).orElseThrow(() -> new NotFoundException(NotificationProfile.class, uuid));
```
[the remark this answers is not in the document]
3keyroman: Restructured in a83d3dfe6: `editNotificationProfile` now runs with `@Transactional(propagation = NOT_SUPPORTED)`, resolves recipients without a transaction, and delegates the locked version assignment to `persistEditedVersion`, a `@Transactional` method invoked through the lazily self-injected proxy (same pattern as `SigningProfileServiceImpl.persistUpdate`). `getRecipientObjectInfo` additionally suppresses the class-level transaction of its own service, so the auth service HTTP call runs with no DB transaction at all.

src/main/resources/db/migration/V202607211000__notification_profile_version_unique_version.sql
```diff
@@ -0,0 +1,62 @@
+-- Flyway runs this file in a single transaction; the explicit locks block concurrent writers (e.g. the
```
lubomirw, line 1: **[Medium]** | Testing

The Flyway repair migration's data-repair SQL (version renumbering + `pending_notification` remap) never runs in CI: the integration suite uses `spring.flyway.enabled=false` and `hibernate.ddl-auto=create-drop` (`src/test/resources/application.yml`), so the unique constraint the new tests exercise is generated by Hibernate from the `@UniqueConstraint` entity annotation — not by this migration. The riskiest part of the change is entirely unverified.

_Suggested:_ Add a migration integration test (the repo already has `integration/migration/BaseMigrationTest`) that seeds duplicate `(notification_profile_uuid, version)` rows plus referencing `pending_notification` rows, runs this migration with Flyway enabled, and asserts the dense renumbering and the `pending_notification` remap — including the ambiguous duplicate-version (MAX/latest-created) case.

<sub>flagged by superpowers code-review + Copilot CLI</sub>

src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
```diff
@@ -167,22 +192,44 @@ public NotificationProfileDetailDto editNotificationProfile(SecuredUUID uuid, No
...
         notificationProfileVersion.setRepetitions(updateRequestDto.getRepetitions());
-        notificationProfileVersion = notificationProfileVersionRepository.save(notificationProfileVersion);
+        try {
+            notificationProfileVersion = notificationProfileVersionRepository.saveAndFlush(notificationProfileVersion);
+        } catch (DataIntegrityViolationException e) {
+            // Backstop for the unique (notification_profile_uuid, version) constraint, reachable only if a
+            // writer bypasses the row lock above. Other integrity violations (e.g. a foreign key on a
+            // concurrently deleted notification instance) are not concurrent-edit collisions and must surface as-is.
+            if (isUniqueVersionViolation(e)) {
+                throw new ValidationException("Notification profile %s was concurrently modified. Retry the edit.".formatted(notificationProfile.getName()));
```
lubomirw, line 214: **[Low]** | Testing

The branch mapping a duplicate-version `DataIntegrityViolationException` to a `ValidationException` (`isUniqueVersionViolation == true`) is never exercised: the row lock prevents duplicates in the concurrency test, `testEditWithUnknownNotificationInstanceFailsWithIntegrityError` covers only the rethrow branch, and `testDuplicateVersionInsertIsRejectedByUniqueConstraint` only pins the constraint name at the repository level.

_Suggested:_ Unit-test the mapping directly — synthesize a `DataIntegrityViolationException` wrapping a Hibernate `ConstraintViolationException` whose `constraintName` is `NotificationProfileVersion.UNIQUE_VERSION_CONSTRAINT`, assert `persistEditedVersion` surfaces `ValidationException`; add a negative case (different constraint name → rethrow).

<sub>flagged by guided-pr-review + superpowers code-review</sub>

src/test/java/com/otilm/core/integration/service/NotificationProfileServiceITest.java
```diff
@@ -221,6 +236,81 @@ void testUpdateNotificationProfile() throws NotFoundException {
         mockServer.stop();
     }
 
+    @Test
+    void testConcurrentEditsAssignDistinctVersions() throws Exception {
```
lubomirw, line 245: **[Low]** | Testing

`testConcurrentEditsAssignDistinctVersions` synchronizes only the start (single `CyclicBarrier`, 2 threads, 1 iteration); the read-latest-version-then-insert contention window is tiny, so the test passes deterministically because the lock works — but would frequently still pass by chance even if the pessimistic lock were removed, giving weak regression protection for the feature's core mechanism.

_Suggested:_ Loop the scenario over many iterations and/or use more editor threads, or hold a latch inside `persistEditedVersion` during the test, to force the interleaving the lock is meant to serialize.

src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
```diff
@@ -167,22 +192,44 @@ public NotificationProfileDetailDto editNotificationProfile(SecuredUUID uuid, No
...
+            }
+            throw e;
```
lubomirw, line 216: **[Low]** | Error handling

A client-supplied unknown notification-instance reference produces a foreign-key `DataIntegrityViolationException` that is rethrown as-is here and then caught by the generic `@ExceptionHandler(Exception.class)`, yielding HTTP 500 for what is really invalid input (no data leak — the generic handler returns a fixed message). Pre-existing behavior, now locked in by `testEditWithUnknownNotificationInstanceFailsWithIntegrityError`.

_Suggested:_ Validate the notification-instance reference before insert (or map the FK violation to a 4xx domain exception) so bad input returns 404/422 instead of 500 — kept distinct from the concurrent-modification mapping above.

src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
```diff
@@ -143,20 +155,33 @@ public NotificationProfileDetailDto createNotificationProfile(NotificationProfil
...
-        NotificationProfileVersion currentVersion = notificationProfile.getCurrentVersion();
+        // Resolve recipient info from the request before opening the write transaction: recipient lookup
+        // can call the auth service over HTTP and must not hold a DB connection or the profile row lock.
+        List<NameAndUuidDto> recipients = resolveRecipients(updateRequestDto.getRecipientType(), updateRequestDto.getRecipientUuids());
 
-        if (areVersionsEqual(currentVersion, updateRequestDto)) {
-            logger.debug("Current version of notification profile {} is same as in request. New version is not created", notificationProfile.getName());
-            return getNotificationProfileDetailDto(currentVersion);
-        }
+        return self.persistEditedVersion(uuid, updateRequestDto, recipients);
```
lubomirw, line 176: **[Nit]** | Maintainability

`persistEditedVersion`'s `@Transactional` (which scopes the `PESSIMISTIC_WRITE` lock) takes effect because it is invoked through the injected `self` proxy. A later refactor to a direct `this.persistEditedVersion(...)` call would drop the transaction and reintroduce the version race. (Package-private is fine — Spring 6.2 + CGLIB advises it.)

_Suggested:_ Add a one-line comment on this `self.persistEditedVersion(...)` call noting the transaction boundary relies on the self-proxied invocation, so the indirection isn't "simplified away" in a future edit.

lubomirw (commented):
Automated multi-reviewer triage — 5 findings (COMMENT). Author-selected from a Copilot CLI + guided-pr-review + superpowers + hygiene run; deduped and independently verified.

A separately-claimed **High** ("package-private `@Transactional` gets no advice → lock not held") was **refuted**: Spring Framework 6.2 advises package-visible methods on CGLIB class-based proxies by default, this project sets no `publicMethodsOnly=true` override, and the method is invoked via the `self` proxy — so the `PESSIMISTIC_WRITE` lock is held across the read-modify-write as intended.

src/main/resources/db/migration/V202607211000__notification_profile_version_unique_version.sql
3keyroman: Added `NotificationProfileVersionUniqueMigrationITest` in 33d5869f5: it drops the entity-generated constraint, seeds duplicate `(notification_profile_uuid, version)` rows with controlled `created_at` plus referencing `pending_notification` rows, executes the migration SQL file in a single transaction (mirroring Flyway), and asserts the dense renumbering, the remap — including the ambiguous duplicated-version → latest-created case — the untouched clean profile, and that the restored constraint rejects duplicates.

src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
3keyroman: Added `NotificationProfileServiceImplPersistEditTest` in 33d5869f5: it drives `persistEditedVersion` with synthesized `DataIntegrityViolationException`s — matching constraint name maps to `ValidationException` with the retry message; a different constraint name and a null constraint name are rethrown unchanged (asserted same instance).

src/test/java/com/otilm/core/integration/service/NotificationProfileServiceITest.java
3keyroman: Strengthened in 33d5869f5: four editors over three rounds, each round on a fresh profile, asserting the exact dense sequence `[1..5]` per round. I left out the latch-inside-`persistEditedVersion` option — it would add test-only state to production code; the widened schedule fails reliably without the lock (the original two-thread variant is what reproduced the duplicate before the fix).

src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
3keyroman: Fixed in 33d5869f5: `createNotificationProfile` and `editNotificationProfile` validate the notification instance reference up front and throw `NotFoundException` for an unknown reference, so clients get 404 instead of 500. The foreign key stays as the backstop for a concurrent instance deletion (that rethrow branch is covered by the new unit test). The former integrity-error test now asserts not-found on both paths, including that a failed create leaves no partial profile behind.

src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
3keyroman: Added in 33d5869f5 — the call site now notes that the transaction boundary relies on the self-proxied invocation.

src/main/java/com/otilm/core/service/impl/ResourceObjectAssociationServiceImpl.java
```diff
@@ -179,6 +180,9 @@ public void bulkRemoveObjectAssociations(Resource resource, List<UUID> objectUui
     }
 
     @Override
+    // The USER and ROLE branches call the auth service over HTTP; suppressing the class-level
+    // @Transactional keeps transaction-less callers from opening a DB transaction around that call.
```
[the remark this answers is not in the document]
3keyroman: Reworded in fc25a3d94 — the comment now states that a caller's active transaction is suspended for the duration of the call rather than joined.

src/test/java/com/otilm/core/integration/service/NotificationProfileServiceITest.java
```diff
@@ -221,6 +241,102 @@ void testUpdateNotificationProfile() throws NotFoundException {
...
+                        .filter(version -> version.getNotificationProfileUuid().equals(profileUuid.getValue()))
+                        .map(NotificationProfileVersion::getVersion)
+                        .sorted()
+                        .toList();
+                Assertions.assertEquals(expectedVersions, versions,
+                        "Concurrent edits must serialize version assignment and never produce duplicate version numbers (round " + round + ")");
+            }
+        } finally {
+            executor.shutdownNow();
+        }
```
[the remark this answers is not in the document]
3keyroman: Changed in fc25a3d94 to orderly `shutdown()` + `awaitTermination(30s)` with `shutdownNow()` as the timeout fallback.

src/test/java/com/otilm/core/integration/migration/NotificationProfileVersionUniqueMigrationITest.java
```diff
@@ -0,0 +1,140 @@
...
+
+            // The unique constraint is in place again and rejects duplicates.
+            try (Statement statement = connection.createStatement()) {
+                SQLException rejection = Assertions.assertThrows(SQLException.class, () -> statement.execute("""
+                        INSERT INTO notification_profile_version
+                            (uuid, notification_profile_uuid, version, recipient_type, internal_notification, created_at)
+                        VALUES ('c0000000-0000-0000-0000-000000000001', '%s', 1, 'OWNER', true, now())
+                        """.formatted(PROFILE_A)));
+                Assertions.assertTrue(rejection.getMessage().contains("uq_notification_profile_version"),
+                        "Duplicate insert should be rejected by the unique constraint, but was: " + rejection.getMessage());
```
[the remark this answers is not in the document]
3keyroman: Added in fc25a3d94 — the test now asserts SQLSTATE 23505 alongside the constraint name (the name check stays to pin *which* constraint fired).

src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
```diff
@@ -143,20 +164,36 @@ public NotificationProfileDetailDto createNotificationProfile(NotificationProfil
...
     @Override
     @ExternalAuthorization(resource = Resource.NOTIFICATION_PROFILE, action = ResourceAction.UPDATE)
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
     public NotificationProfileDetailDto editNotificationProfile(SecuredUUID uuid, NotificationProfileUpdateRequestDto updateRequestDto) throws NotFoundException {
-        NotificationProfile notificationProfile = notificationProfileRepository.findByUuid(uuid).orElseThrow(() -> new NotFoundException(NotificationProfile.class, uuid));
-        NotificationProfileVersion currentVersion = notificationProfile.getCurrentVersion();
+        // Resolve recipient info from the request before opening the write transaction: recipient lookup
+        // can call the auth service over HTTP and must not hold a DB connection or the profile row lock.
+        List<NameAndUuidDto> recipients = resolveRecipients(updateRequestDto.getRecipientType(), updateRequestDto.getRecipientUuids());
+        validateNotificationInstanceExists(updateRequestDto.getNotificationInstanceUuid());
```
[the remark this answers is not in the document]
3keyroman: Reordered in the follow-up commit — the instance existence check now runs before recipient resolution, so an invalid instance UUID fails fast without any auth service call.

lubomirw (approved):

PR #1854, opened 2026-07-21
Validate fieldMapping regardless of operation

### LINKED ISSUE #1850: fieldMapping validation skipped on the register build (operation=null)
## Symptom

Every structured pre-registration logs:

```
DataAttribute '<name>' has fieldMapping but is registered outside a request operation context (operation=null); fieldMapping validation skipped
```

## Root cause

`buildStructuredRegistrationContent` calls the attribute engine with `operation=null`:

```java
// ClientOperationServiceImpl.java:659
attributeEngine.validateUpdateDataAttributes(null, null, definitions, csrAttributes);
```

The engine gates `fieldMapping` validation on a request operation:

```java
// AttributeEngine.java:595 / isRequestOperation:1247
if (dataAttribute instanceof DataAttributeV3 v3 && v3.getFieldMapping() != null) {
    if (isRequestOperation(operation)) {           // CERTIFICATE_ISSUE / SIGN only
        validateFieldMapping(v3, ...);
    } else {
        logger.warn("... fieldMapping validation skipped");
    }
}
```

With `operation=null`, `isRequestOperation` is false, so `validateFieldMapping` (`AttributeEngine.java:1266`) is skipped on the register path. A malformed `fieldMapping` is therefore not rejected when registering, and mapped content is not engine-validated.

## Impact

- A malformed request-attribute `fieldMapping` passes silently through pre-registration instead of being rejected.
- Low functional impact on the happy path, but the missing validation masks authoring errors (e.g. a SAN field with a missing `generalNameType`).

## Fix

Pass a real request operation (`CERTIFICATE_ISSUE`, or a register-appropriate request operation) to `validateUpdateDataAttributes` in `buildStructuredRegistrationContent` so `validateFieldMapping` runs on the register build.

Note: `validateFieldMapping` only validates (it does not mutate the mapping or the submitted content), so this fix does not by itself change projection behaviour.

klaraf755:
Fixes #1850.

### Problem

`validateFieldMapping` was gated on `isRequestOperation(operation)` — only `certificateIssue` and `sign` triggered it. The register build (and the platform-built CSR path) call `validateUpdateDataAttributes(null, null, definitions, csrAttributes)` with a null operation, so a malformed `fieldMapping` on an issuance definition passed silently through pre-registration, and every structured registration logged a "fieldMapping validation skipped" warning.

### Fix

Validate `fieldMapping` unconditionally whenever a registered definition carries one. The operation gate, the skip-branch warning, and `isRequestOperation` are removed.

The rationale: whether a `fieldMapping` is well-formed is intrinsic to the definition — a malformed mapping is an authoring error regardless of which operation the definition registers under. Gating validation on caller-supplied context is exactly what caused this bug, and any variant of it keeps the failure mode.

### Why not the alternatives

- **Passing a request operation on the register path** (the fix suggested in the issue) is unsafe. The `operation` value is persisted: new definition rows are created with it, and `validateDataAttributesContent` overwrites the stored operation on definitions it loads by request attribute. The stored operation is load-bearing — `AttributeContent2ObjectRepository` joins content reads through `ad.operation`, and csrAttributes content on `CERTIFICATE_REQUEST` is written and read with `operation=null`. Retagging the issuance definitions would make those null-operation reads come back empty.

### Behavior change

Registration of any fieldMapping-bearing definition — including connector-supplied attributes under non-request operations — now rejects a malformed mapping instead of warn-and-tolerate. `fieldMapping` is a recent v3 feature (gate introduced in #1655), so no deployed connectors are expected to be affected; if one is, the failure is loud and names the attribute and defect, and the mapping would have misbehaved at projection time anyway.

### Tests

`testFieldMapping_nonRequestOperation_skipsValidation` is flipped to `testFieldMapping_nonRequestOperation_validatesFieldMapping`: a malformed mapping now throws under `operation=null`, which is the register path's exact shape. The remaining `ValidateMappedField` suite is unchanged and still passes.

### DIFF
--- a/src/test/java/com/otilm/core/integration/attribute/AttributeEngineITest.java
+++ b/src/test/java/com/otilm/core/integration/attribute/AttributeEngineITest.java
@@ -1622,11 +1622,13 @@ class AttributeEngineITest extends BaseSpringBootTest {
             Assertions.assertTrue(ex.getMessage().contains("fieldMapping is only valid for attributes with STRING or TEXT content type"), ex::getMessage);
         }
 
-        // ── isRequestOperation gating ─────────────────────────────────────────────
+        // ── fieldMapping validated regardless of operation ───────────────────────
 
         @Test
-        void testFieldMapping_nonRequestOperation_skipsValidation() {
-            // Invalid fieldMapping (no objectType) should NOT throw when operation is not a request operation
+        void testFieldMapping_nonRequestOperation_validatesFieldMapping() {
+            // A fieldMapping declares projection intent; a malformed one is an authoring error
+            // whatever operation the definition registers under (issuance definitions register
+            // with operation=null), so validation must not be gated on the operation
             DataAttributeV3 attr = fieldMappingAttribute("fm_non_req_op");
             FieldMapping fm = new FieldMapping();
             fm.setObjectType(null);
@@ -1634,9 +1636,9 @@ class AttributeEngineITest extends BaseSpringBootTest {
             attr.setFieldMapping(fm);
 
             UUID connectorUuid = connectorAuthority.getUuid();
-            // null operation → skips validateFieldMapping, so no exception
-            Assertions.assertDoesNotThrow(
+            AttributeException ex = Assertions.assertThrows(AttributeException.class,
                     () -> attributeEngine.updateDataAttributeDefinitions(connectorUuid, null, List.of(attr)));
+            Assertions.assertTrue(ex.getMessage().contains("fieldMapping.objectType is required"), ex::getMessage);
         }
 
         @Test
--- a/src/test/java/com/otilm/core/integration/attribute/AttributeEngineITest.java
+++ b/src/test/java/com/otilm/core/integration/attribute/AttributeEngineITest.java
@@ -1642,6 +1642,18 @@ class AttributeEngineITest extends BaseSpringBootTest {
         }
 
         @Test
+        void testFieldMapping_nonRequestOperation_acceptsValidMapping() {
+            // The register build registers issuance definitions with operation=null; a well-formed
+            // mapping must be accepted there, not merely rejected when malformed
+            DataAttributeV3 attr = fieldMappingAttribute("fm_non_req_op_valid");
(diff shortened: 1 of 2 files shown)


src/test/java/com/otilm/core/integration/attribute/AttributeEngineITest.java
```diff
@@ -1622,21 +1622,23 @@ void testFieldMapping_nonStringContentType_throws() {
             Assertions.assertTrue(ex.getMessage().contains("fieldMapping is only valid for attributes with STRING or TEXT content type"), ex::getMessage);
         }
 
-        // ── isRequestOperation gating ─────────────────────────────────────────────
+        // ── fieldMapping validated regardless of operation ───────────────────────
 
         @Test
-        void testFieldMapping_nonRequestOperation_skipsValidation() {
-            // Invalid fieldMapping (no objectType) should NOT throw when operation is not a request operation
+        void testFieldMapping_nonRequestOperation_validatesFieldMapping() {
```
lubomirw, line 1625: **[Low]** | Testing

This flipped test covers only the negative path (missing `objectType` throws under `operation=null`). Every happy-path fieldMapping test still runs under `CERTIFICATE_ISSUE`. The core point of this change — the register path (`operation=null`) must validate **and still accept** a well-formed mapping — is not locked in by any test, so a future over-tightening on the register path would go uncaught.

_Suggested:_ add an `assertDoesNotThrow` case that registers a well-formed fieldMapping-bearing `DataAttributeV3` via `updateDataAttributeDefinitions(connectorUuid, null, List.of(attr))`, mirroring the register-build shape.

lubomirw (commented):
Automated multi-reviewer triage — 1 finding (COMMENT). Copilot CLI, superpowers, and hygiene raised no blocking issues; title and label handled separately.

src/test/java/com/otilm/core/integration/attribute/AttributeEngineITest.java
klaraf755: Added a test.

lubomirw (approved):

PR #1855, opened 2026-07-21
Persist registered SAN and surface dropped extensions on pre-registration

### LINKED ISSUE #1851: Register drops projected SAN / extensions; only the subject DN is persisted
## Symptom

A request attribute mapped to a SAN (e.g. a DNS name) or a certificate extension is lost during pre-registration: the certificate is registered with the subject DN only, no SAN / no extension.

## Root cause

The register flow projects the full structured identity (subject DN + SAN + extensions) into `registrationContent`, then collapses it to the DN and never carries the rest:

```java
// ClientOperationServiceImpl.java:468-471
X509RequestContent registrationContent = buildStructuredRegistrationContent(raProfile, request); // DN + SAN + extensions
String effectiveSubjectDn = registrationContent != null
        ? RegisterWireBuilder.renderSubjectDn(registrationContent)   // DN only — SAN/extensions discarded
        : request.getSubjectDn();
```

- `createRegistrationPlaceholder(RaProfile, String effectiveSubjectDn)` (`CertificateServiceImpl.java:1142`) takes only a DN string — the placeholder row cannot record a SAN or extensions.
- `registerPlatformLevel` (`ClientOperationServiceImpl.java:568`) receives only `effectiveSubjectDn`; the projected `registrationContent` (carrying the SAN) is never passed in. On the **platform-level** path the SAN and extensions are **silently dropped** — not stored, not sent anywhere.
- The **connector-backed** path does forward `registrationContent` to the adapter (`ClientOperationServiceImpl.java:504` → `RegisterWireBuilder.buildRegistration` → flat `subjectAltName` → the connector's create-end-entity call), so the CA receives the SAN — but the platform's own REGISTERED placeholder still shows the DN only.

## Impact

- Platform-level pre-registration silently drops a mapped SAN / extension; the operator's mapped identity is reduced to the subject DN with no error.
- On both paths the REGISTERED placeholder certificate persists only the subject DN, so the platform view of a pre-registered identity is incomplete.

## Fix

Stop reducing the projected identity to the DN on the register path — carry the whole `X509RequestContent`:

- Extend `createRegistrationPlaceholder` to persist the projected SAN and extensions on the placeholder.
- Pass `registrationContent` into `registerPlatformLevel` so the platform-level path stores/forwards the SAN and extensions instead of dropping them.

klaraf755:
Fixes #1851

 ## Problem
The register flow projected the full structured identity (subject DN + SAN + extensions) into `X509RequestContent`, then collapsed it to the DN. The placeholder row never recorded the SAN, and the platform-level path dropped the SAN and extensions entirely — silently, with no error and no record.
>
 ## Changes
> - **Single identity projection for both request forms.** Structured `csrAttributes` go through the issuance-definition projector as before; flat requests now go through the same `RegisterWireBuilder.buildContent` the adapter uses for the wire, in the orchestrator. The adapter receives the pre-built content on both forms (its flat fallback remains as a defensive contract). The placeholder DN keeps its historical derivation, so no stored DN formats change.
> - **The placeholder persists the SAN.** `createRegistrationPlaceholder` takes the projected content; new `CertificateUtil.applyRegistrationSan` writes `subject_alternative_names` in the same serialized map format used for issued certificates (shared otherName formatter, so the two can't drift). The stored SAN is the platform's record of the pre-registered identity until issuance overwrites it from the CA response.
> - **The certificate detail maps the SAN unconditionally**, not only when `certificateContent` is present, so a content-less REGISTERED placeholder exposes its registered SAN.
> - **Platform-level extensions are recorded, not dropped silently.** On the platform-level path there is no connector wire to forward extensions to and no column to record them, and the authoritative extensions arrive with the CSR at issuance — so the registration succeeds, and the drop is made operator-visible: a certificate event-history entry listing the extension OIDs with guidance to supply them in the CSR, plus a warn log. (Event history rather than logs alone, since SaaS operators can't read server logs.)
>
 ## Behavior changes
> - Malformed flat SAN/DN now rejects **before** the placeholder exists (no FAILED row), uniform with the sibling register validations.
> - Platform-level registrations with an unparsable SAN string now return a validation error instead of silently dropping it.
> - Platform-level registrations with extensions remain accepted, now with an audit record in the certificate event history.

### DIFF
--- a/src/main/java/com/otilm/core/mapper/certificate/CertificateDetailDtoMapper.java
+++ b/src/main/java/com/otilm/core/mapper/certificate/CertificateDetailDtoMapper.java
@@ -30,7 +30,9 @@ public class CertificateDetailDtoMapper {
         final CertificateDetailDto dto = new CertificateDetailDto();
         dto.setCommonName(CertificateUtil.formatCommonName(certificate.getCommonName()));
         dto.setIssuerCommonName(resolveIssuerCommonName(certificate));
-        dto.setSubjectAlternativeNames(CertificateUtil.deserializeSans(certificate.getSubjectAlternativeNames()));
+        if (certificate.getSubjectAlternativeNames() != null) {
+            dto.setSubjectAlternativeNames(CertificateUtil.deserializeSans(certificate.getSubjectAlternativeNames()));
+        }
         if (certificate.getCertificateContent() != null) {
             dto.setCertificateContent(certificate.getCertificateContent().getContent());
             dto.setIssuerDn(certificate.getIssuerDn());
--- a/src/main/java/com/otilm/core/mapper/certificate/CertificateDetailDtoMapper.java
+++ b/src/main/java/com/otilm/core/mapper/certificate/CertificateDetailDtoMapper.java
@@ -30,6 +30,7 @@ public class CertificateDetailDtoMapper {
         final CertificateDetailDto dto = new CertificateDetailDto();
         dto.setCommonName(CertificateUtil.formatCommonName(certificate.getCommonName()));
         dto.setIssuerCommonName(resolveIssuerCommonName(certificate));
+        dto.setSubjectAlternativeNames(CertificateUtil.deserializeSans(certificate.getSubjectAlternativeNames()));
         if (certificate.getCertificateContent() != null) {
             dto.setCertificateContent(certificate.getCertificateContent().getContent());
             dto.setIssuerDn(certificate.getIssuerDn());
@@ -39,7 +40,6 @@ public class CertificateDetailDtoMapper {
             dto.setExtendedKeyUsage(MetaDefinitions.deserializeArrayString(certificate.getExtendedKeyUsage()));
             dto.setKeyUsage(certificate.getKeyUsage().stream().toList());
             dto.setFingerprint(certificate.getFingerprint());
-            dto.setSubjectAlternativeNames(CertificateUtil.deserializeSans(certificate.getSubjectAlternativeNames()));
             dto.setIssuerSerialNumber(certificate.getIssuerSerialNumber());
             dto.setSerialNumber(certificate.getSerialNumber());
 
--- a/src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
@@ -11,6 +11,7 @@ import com.otilm.api.model.client.location.PushToLocationRequestDto;
 import com.otilm.api.model.common.attribute.common.BaseAttribute;
 import com.otilm.api.model.common.attribute.v3.DataAttributeV3;
 import com.otilm.api.model.connector.v3.certificate.CertificateRequestContent;
+import com.otilm.api.model.connector.v3.certificate.RequestedExtension;
 import com.otilm.api.model.connector.v3.certificate.X509RequestContent;
 import com.otilm.api.exception.ConnectorClientException;
 import com.otilm.api.exception.ConnectorCommunicationException;
@@ -114,6 +115,7 @@ import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.util.*;
 import java.util.function.BooleanSupplier;
+import java.util.stream.Collectors;
 
 @Service("clientOperationServiceImplV2")
 @Transactional
@@ -463,21 +465,28 @@ public class ClientOperationServiceImpl implements ClientOperationExternalServic
         AuthorityInstanceReference authority = raProfile.getAuthorityInstanceReference();
         AuthorityProviderAdapter adapter = adapterFactory.forAuthority(authority);
 
-        // Project structured csrAttributes once (validated) so the placeholder DN and the connector wire derive
-        // from the same content; a flat request yields null here and the adapter builds its identity itself.
+        // Project the registration identity once (validated) so the placeholder row and the connector wire
+        // derive from the same content: structured csrAttributes through the issuance-definition projector,
+        // a flat request through the same builder the adapter uses for the wire. The placeholder DN keeps its
+        // historical derivation (rendered for structured, the raw request DN for flat).
         X509RequestContent registrationContent = buildStructuredRegistrationContent(raProfile, request);
-        String effectiveSubjectDn = registrationContent != null
-                ? RegisterWireBuilder.renderSubjectDn(registrationContent)
-                : request.getSubjectDn();
+        String effectiveSubjectDn;
+        if (registrationContent != null) {
+            effectiveSubjectDn = RegisterWireBuilder.renderSubjectDn(registrationContent);
+        } else {
+            effectiveSubjectDn = request.getSubjectDn();
+            registrationContent = RegisterWireBuilder.buildContent(
+                    request.getSubjectDn(), request.getSubjectAltName(), request.getExtensions());
+        }
 
         // No connector-backed registration for this authority (not a RegisterCapability, or the flag is not
         // advertised) -> platform-level pre-registration (see registerPlatformLevel); otherwise connector-backed below.
         if (!(adapter instanceof RegisterCapability registerCapability)
                 || !capabilityService.supports(authority, FeatureFlag.CERTIFICATE_REGISTRATION)) {
-            return registerPlatformLevel(raProfile, request, effectiveSubjectDn, createCustomAttributes);
+            return registerPlatformLevel(raProfile, request, effectiveSubjectDn, registrationContent, createCustomAttributes);
         }
 
-        Certificate placeholder = certificateService.createRegistrationPlaceholder(raProfile, effectiveSubjectDn);
+        Certificate placeholder = certificateService.createRegistrationPlaceholder(raProfile, effectiveSubjectDn, registrationContent);
         // Re-load with the full adapter graph for the transaction-less connector call, as the poll listener does.
         // No pessimistic lock here (unlike the cancel/poll paths): the placeholder was just created and its UUID
         // is not yet known to any concurrent actor, so the read-modify-write below cannot race.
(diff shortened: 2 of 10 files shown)


src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -586,6 +600,7 @@ private ClientCertificateDataResponseDto registerPlatformLevel(RaProfile raProfi
             // marker as a connector-backed one — the discriminator both the issue routing and the approval-reject
             // restore key on, which makes a no-secret placeholder restorable.
             certificateRegistrationWriter.upsert(certificate.getUuid(), List.of());
+            noteUnrecordedPlatformLevelExtensions(certificate.getUuid(), registrationContent);
```
lubomirw: **[Low]** | Error handling

`noteUnrecordedPlatformLevelExtensions` runs inside the placeholder try-block, so a failure writing the purely-advisory event-history entry transitions an otherwise-functional registration to FAILED and deletes its authorization (catch at line 604) — the essential setup already succeeded.

_Suggested:_ make the note best-effort — catch+log inside the method (mirroring `deleteRegistrationAuthorizationBestEffort`) so an advisory-note write failure can't roll back done work.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -463,21 +465,28 @@ public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID au
...
         X509RequestContent registrationContent = buildStructuredRegistrationContent(raProfile, request);
-        String effectiveSubjectDn = registrationContent != null
-                ? RegisterWireBuilder.renderSubjectDn(registrationContent)
-                : request.getSubjectDn();
+        String effectiveSubjectDn;
+        if (registrationContent != null) {
+            effectiveSubjectDn = RegisterWireBuilder.renderSubjectDn(registrationContent);
+        } else {
+            effectiveSubjectDn = request.getSubjectDn();
+            registrationContent = RegisterWireBuilder.buildContent(
```
lubomirw, line 476: **[Low]** | Error handling

The flat-request path calls `RegisterWireBuilder.buildContent(...)`, whose `mapExtensions` guards a null list but not a null *element* — a request with `"extensions":[null]` yields an unhandled NPE/500. Pre-existing in the builder, but newly reachable on the platform-level flat path here.

_Suggested:_ skip/reject null elements in `mapExtensions`, or annotate the DTO list `List<@NotNull CertificateExtension>` so a null element becomes a 400.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -603,6 +618,31 @@ private ClientCertificateDataResponseDto registerPlatformLevel(RaProfile raProfi
...
+    private void noteUnrecordedPlatformLevelExtensions(UUID certificateUuid, X509RequestContent registrationContent) {
+        if (registrationContent == null || registrationContent.getExtensions() == null
+                || registrationContent.getExtensions().isEmpty()) {
+            return;
+        }
+        String oids = registrationContent.getExtensions().stream()
+                .map(RequestedExtension::getOid)
+                .collect(Collectors.joining(", "));
+        certificateEventHistoryService.addEventHistory(certificateUuid, CertificateEvent.REQUEST, CertificateEventStatus.SUCCESS,
+                ("Requested certificate extensions (%s) are not recorded on a platform-level pre-registration "
```
lubomirw: **[Low]** | Security

Operator-supplied extension OIDs are joined straight into a persisted (UI-rendered) certificate-event message and a `logger.warn` without validation — `CertificateExtension.oid` carries no `@Pattern`. Main risk is log injection (CR/LF) into the warn line; impact limited since the operator is authenticated.

_Suggested:_ validate each OID (e.g. `OidHandler.isOid`) before formatting, or add a `@Pattern` on the DTO field.

src/main/java/com/otilm/core/mapper/certificate/CertificateDetailDtoMapper.java
```diff
@@ -30,6 +30,7 @@ private static CertificateDetailDto buildDetailDto(Certificate certificate, bool
         final CertificateDetailDto dto = new CertificateDetailDto();
         dto.setCommonName(CertificateUtil.formatCommonName(certificate.getCommonName()));
         dto.setIssuerCommonName(resolveIssuerCommonName(certificate));
+        dto.setSubjectAlternativeNames(CertificateUtil.deserializeSans(certificate.getSubjectAlternativeNames()));
```
lubomirw: **[Low]** | Maintainability

Moving `setSubjectAlternativeNames` out of the `certificateContent != null` block sets it for every content-less certificate. `deserializeSans(null)` returns an empty map, so the detail DTO field flips from null → `{}` for all content-less states (PENDING_ISSUE, FAILED, …), a wider contract change than the placeholder fix needs.

_Suggested:_ confirm null→`{}` is intended for all states; if placeholders-only was the intent, gate the set — otherwise note the widened contract.

src/main/java/com/otilm/core/util/CertificateUtil.java
```diff
@@ -605,6 +606,37 @@ public static void applyRegistrationSubject(Certificate modal, String subjectDn)
...
+        Map<String, List<String>> sans = buildEmptySans();
+        for (GeneralNameEntry entry : subjectAltNames) {
+            // Projected entries carry a validated type; guard anyway so a malformed persisted mapping
+            // surfaces as a controlled rejection instead of an NPE inside the placeholder transaction.
+            if (entry.getType() == null) {
+                throw new ValidationException(ValidationError.create(
+                        "A projected subject alternative name entry carries no type; check the SAN field mappings of the issuance attributes"));
+            }
+            String value = entry.getType() == GeneralNameType.OTHER_NAME
+                    ? formatOtherNameSan(entry.getOtherNameOid(), entry.getValue())
```
lubomirw, line 633: **[Low]** | Correctness

OTHER_NAME persistence in `applyRegistrationSan`: the `type==null` guard exists but `otherNameOid` is unguarded (a null oid serializes literal `"null=value"`), and the projected textual value ignores `valueEncoding`, so a non-UTF8 otherName could diverge from the issued-cert form. Edge cases only — content is internally built and the column is overwritten at issuance.

_Suggested:_ add an `otherNameOid` null/blank check beside the type guard; optionally canonicalize the otherName through the same ASN.1 path issued certs use.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
```diff
@@ -882,6 +894,121 @@ void registerProjectsStructuredCsrAttributesAndForwardsContent() throws Exceptio
         Assertions.assertEquals(CertificateState.REGISTERED, cert.getState());
     }
 
+    // ── registered identity (SAN / extensions) on the placeholder ────────────
```
lubomirw, line 894: **[Low]** | Testing

Tests cover DNS/IP flat + projected SANs but not the OTHER_NAME persistence branch (the shared `formatOtherNameSan`) nor the null-type `ValidationException` branch in `applyRegistrationSan`.

_Suggested:_ add an OTHER_NAME persistence case (assert the `oid=value` serialized form) and, if practical, the null-type rejection.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
```diff
@@ -882,6 +894,121 @@ void registerProjectsStructuredCsrAttributesAndForwardsContent() throws Exceptio
...
+        Assertions.assertEquals(List.of("device-1.example.com"), sans.get("dNSName"),
+                "the flat DNS SAN must be persisted on the placeholder row");
+        Assertions.assertEquals(List.of("10.0.0.1"), sans.get("iPAddress"),
+                "the flat IP SAN must be persisted on the placeholder row");
+    }
+
+    @Test
+    void platformLevelRegistrationPersistsSanOnPlaceholder() throws Exception {
+        // Platform-level path (no connector-backed registration): the requested SAN must be stored on the
+        // placeholder, not silently dropped — before this fix the identity was reduced to the subject DN.
```
lubomirw: **[Low]** | Maintainability

This comment mixes assertion rationale with change-history narration ('before this fix the identity was reduced to the subject DN') — it records past behavior rather than explaining the code.

_Suggested:_ drop the history clause, keep the rationale.

<sub>flagged by hygiene</sub>

lubomirw (commented):
Automated multi-reviewer triage (Copilot CLI, guided-pr-review, superpowers, hygiene) — 7 findings, all Low/Nit (COMMENT). Deduped and verified against the diff. No blocking issues. Nothing off-diff, so all findings are inline.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
klaraf755: Dropped the history clause.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
klaraf755: Added those two tests.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Removed the note because it will be shown in request attributes tied to registered certificate.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Skips null elements.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Removed the method for creating the note in event history, so no longer relevant.

src/main/java/com/otilm/core/mapper/certificate/CertificateDetailDtoMapper.java
klaraf755: Added check to set it only for not-null SAN.

src/main/java/com/otilm/core/util/CertificateUtil.java
klaraf755: Added validation check for null OID.

lubomirw (approved):

PR #1868, opened 2026-07-23
Accept platform RDN codes such as EMAIL in subject DNs

### LINKED ISSUE #1863: Subject DNs using platform RDN codes (EMAIL) are rejected on re-parse and by strict RFC 4514 connectors
### Description

The platform renders subject DNs with its OID-registry codes (e.g. EMAIL for 1.2.840.113549.1.9.1 per SystemOid), but two surfaces reject exactly that form:

1. **Parse side (core):** `PlatformX500NameStyle` extends `BCStrictStyle` and does not consult the OID registry when parsing, so a registration subject DN like `EMAIL=mail@mail.com, CN=reg` — the same form the platform itself displays — fails with HTTP 422 "Unknown object id - EMAIL - passed to distinguished name" (`CertificateUtil.applyRegistrationSubject`, `X509RequestContentParser.parseSubjectDn`). Runtime-registered custom OID codes are likewise unparseable, and `SN` (rendered for surname 2.5.4.4) re-parses as serialNumber 2.5.4.5, silently corrupting round-trips.

2. **Wire side (v3 register):** `RegisterWireBuilder` renders the flat `subjectDn` for connectors with the platform display style, emitting non-RFC 4514 keywords (`EMAIL=`) with no value escaping. Strict connector parsers reject it — otpki returns HTTP 400 'Invalid RFC 4514 Subject DN: unknown attribute type "EMAIL"'. Comma-containing values (`O=Org, s.r.o.`) also produce an ambiguous unescaped wire string.

### Steps to Reproduce

1. Register a certificate (no-CSR path) with subject DN `EMAIL=mail@mail.com, CN=reg` → HTTP 422
2. With parsing fixed, register via an RA profile backed by a strict RFC 4514 connector (otpki) → HTTP 400 on the register wire

### Expected Behavior

DN strings the platform renders are accepted back by the platform, and the connector wire carries a strict RFC 4514/2253 subject DN (registered keywords or dotted-decimal OIDs with hexstring values, escaped).

### Actual Behavior

HTTP 422 "Invalid subject DN ... Unknown object id - EMAIL" on core parse; HTTP 400 "Invalid RFC 4514 Subject DN: unknown attribute type \"EMAIL\"" from connector otpki.

### Severity

Major

### Module

Certificates

lubomirw:
## What

Follow-up hardening to the strict-wire-DN / registry-first-OID change (merged into `main` via #1839's merge commit). Addresses correctness, concurrency, and test-hygiene items surfaced in review.

### `OidHandler`
- Serialize writers on a private `WRITE_LOCK` monitor instead of the public `OidHandler.class` object, so foreign code cannot contend on the same lock.
- Guard `cacheOid` / `removeCachedOid` with `getOrDefault(category, Map.of())` so the first write to an as-yet-uncached category cannot NPE.
- Expose `getRdnCodeToOidMap()` — the published immutable, case-insensitive snapshot — so callers can capture it once.

### Wire DN rendering
- `X509RequestContentRenderer.toX500Principal` captures one registry snapshot per subject and resolves every RDN against it, and `resolveOid` now matches RDN codes **case-insensitively** — the same rule the parse path uses (`PlatformX500NameStyle.attrNameToOID`). A code that parses in now also renders back out, and all RDNs of one subject resolve against a single, consistent snapshot.

### Tests
- `AttributeEngineITest.ensureOidCached` routes through `cacheOid` instead of mutating the `getOidCache()` map in place, honoring the copy-on-write contract.
- New `OidHandlerTest`: null and case-insensitive code/altCode lookup, code de-registration via `removeCachedOid`, and uncached-category NPE-guard coverage (with the uncached precondition enforced).
- New case-insensitive wire-rendering test.

## Why

The render path resolved RDN codes case-sensitively while the parse path was case-insensitive, so a lower-case structured code could parse but fail to render. The per-RDN live lookup also let one subject resolve against multiple registry snapshots. The lock and NPE-guard changes tighten the copy-on-write registry the platform relies on.

## Testing

`mvn test` for the affected unit tests passes (`OidHandlerTest`, `X509RequestContentRendererTest`, `RegisterWireBuilderTest`, `X509RequestContentParserTest`, `CertificateRequestContentValidatorTest`). Spring integration tests (`PlatformX500NameStyleITest`, `AttributeEngineITest`, `CustomOidEntryServiceITest`) were not run locally — please run the ITest suite in CI before merge.

### DIFF
--- a/src/main/java/com/otilm/core/oid/OidHandler.java
+++ b/src/main/java/com/otilm/core/oid/OidHandler.java
@@ -33,27 +33,48 @@ public class OidHandler {
     private static final AtomicReference<Map<String, String>> rdnCodeToOid =
             new AtomicReference<>(Collections.emptyMap());
 
+    /**
+     * Serializes writers so that the read-copy-publish of a per-category map and the rebuild of the
+     * derived {@link #rdnCodeToOid} index happen as one unit. A private monitor is used rather than
+     * the {@code OidHandler.class} object so foreign code cannot contend on the same lock.
+     */
+    private static final Object WRITE_LOCK = new Object();
+
     public static Map<String, OidRecord> getOidCache(OidCategory oidCategory) {
         return oidCache.get(oidCategory);
     }
 
-    public static synchronized void cacheOidCategory(OidCategory category, Map<String, OidRecord> oidRecordMap) {
-        oidCache.put(category, oidRecordMap);
-        refreshRdnCodeLookup(category);
+    public static void cacheOidCategory(OidCategory category, Map<String, OidRecord> oidRecordMap) {
+        synchronized (WRITE_LOCK) {
+            oidCache.put(category, oidRecordMap);
+            refreshRdnCodeLookup(category);
+        }
     }
 
-    public static synchronized void cacheOid(OidCategory category, String oid, OidRecord oidRecord) {
-        // Copy-on-write: published per-category maps are iterated lock-free by readers
-        // (getCodeToOidMap, style snapshots), so never mutate one in place.
-        Map<String, OidRecord> next = new HashMap<>(oidCache.get(category));
-        next.put(oid, oidRecord);
-        oidCache.put(category, next);
-        refreshRdnCodeLookup(category);
+    public static void cacheOid(OidCategory category, String oid, OidRecord oidRecord) {
+        synchronized (WRITE_LOCK) {
+            // Copy-on-write: published per-category maps are iterated lock-free by readers
+            // (getCodeToOidMap, style snapshots), so never mutate one in place. getOrDefault
+            // keeps the first write to an as-yet-uncached category from throwing.
+            Map<String, OidRecord> next = new HashMap<>(oidCache.getOrDefault(category, Map.of()));
+            next.put(oid, oidRecord);
+            oidCache.put(category, next);
+            refreshRdnCodeLookup(category);
+        }
     }
 
     /** OID for an RDN code or alternative code, matched case-insensitively; {@code null} when unknown. */
     public static String getOidForRdnCode(String code) {
         return code == null ? null : rdnCodeToOid.get().get(code);
+    }
+
+    /**
+     * The published, immutable, case-insensitive RDN code/altCode → OID snapshot. Capture it once
+     * and reuse it for a whole DN so every RDN of one subject resolves against the same registry
+     * state, rather than re-reading the reference per attribute.
+     */
+    public static Map<String, String> getRdnCodeToOidMap() {
+        return rdnCodeToOid.get();
     }
 
     private static void refreshRdnCodeLookup(OidCategory category) {
@@ -93,10 +114,12 @@ public class OidHandler {
     }
 
 
-    public static synchronized void removeCachedOid(OidCategory category, String oid) {
-        Map<String, OidRecord> next = new HashMap<>(oidCache.get(category));
-        next.remove(oid);
-        oidCache.put(category, next);
-        refreshRdnCodeLookup(category);
+    public static void removeCachedOid(OidCategory category, String oid) {
+        synchronized (WRITE_LOCK) {
+            Map<String, OidRecord> next = new HashMap<>(oidCache.getOrDefault(category, Map.of()));
+            next.remove(oid);
+            oidCache.put(category, next);
+            refreshRdnCodeLookup(category);
+        }
     }
 }
--- a/src/main/java/com/otilm/core/oid/OidHandler.java
+++ b/src/main/java/com/otilm/core/oid/OidHandler.java
@@ -40,25 +40,30 @@ public class OidHandler {
      */
     private static final Object WRITE_LOCK = new Object();
 
+    /** The published per-category map, or {@code null} when the category is not loaded. Immutable — see {@link #cacheOidCategory}. */
     public static Map<String, OidRecord> getOidCache(OidCategory oidCategory) {
         return oidCache.get(oidCategory);
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/otilm/core/oid/OidHandler.java
```diff
@@ -93,10 +114,12 @@ public static Map<String, String> getCodeToOidMap() {
...
-        refreshRdnCodeLookup(category);
+    public static void removeCachedOid(OidCategory category, String oid) {
+        synchronized (WRITE_LOCK) {
+            Map<String, OidRecord> next = new HashMap<>(oidCache.getOrDefault(category, Map.of()));
+            next.remove(oid);
+            oidCache.put(category, next);
+            refreshRdnCodeLookup(category);
+        }
     }
 }
```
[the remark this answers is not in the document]
lubomirw: Good catch — fixed in 5ddbdbb75. `removeCachedOid` now returns early when the category map is absent, so removal from a never-loaded category is a true no-op and `getOidCache(category)` stays `null` (the NPE guard is kept for cached categories). Added an assertion that the category remains null after a no-op remove.

src/test/java/com/otilm/core/oid/OidHandlerTest.java
```diff
@@ -0,0 +1,133 @@
...
+            saved.put(category, existing == null ? null : new HashMap<>(existing));
+        }
+    }
+
+    @AfterAll
+    static void restoreTouchedCategories() {
+        for (OidCategory category : TOUCHED) {
+            Map<String, OidRecord> original = saved.get(category);
+            if (original == null) {
+                evictCategory(category);
```
klaraf755, line 51: **Stale `rdnCodeToOid` snapshot leaks out of this class when a touched category was originally unloaded.**

`evictCategory` bypasses `refreshRdnCodeLookup`, so this branch removes the `oidCache` entry but leaves the derived `rdnCodeToOid` index holding whatever the last test registered. Concretely: fresh JVM where this class runs first → `@BeforeAll` saves `null` for `RDN_ATTRIBUTE_TYPE` → after `@AfterAll`, `getOidCache(RDN_ATTRIBUTE_TYPE)` is `null` but `getOidForRdnCode("email")` still returns `1.2.840.113549.1.9.1`. Later test classes going through `PlatformX500NameStyle.attrNameToOID` or `getRdnCodeToOidMap()` see phantom entries from a registry that claims to be empty — the exact contamination this snapshot/restore dance is meant to prevent.

The non-null branch is fine (`cacheOidCategory` refreshes the index); only this path leaks. Suggested fix:

```java
if (original == null) {
    // Refresh the derived rdnCodeToOid index to empty before dropping the entry —
    // evictCategory bypasses refreshRdnCodeLookup.
    OidHandler.cacheOidCategory(category, new HashMap<>());
    evictCategory(category);
} else {
    OidHandler.cacheOidCategory(category, original);
}
```

(Harmless no-op for the two non-RDN categories, since `refreshRdnCodeLookup` ignores them.)

src/main/java/com/otilm/core/oid/OidHandler.java
```diff
@@ -33,29 +33,50 @@ public static boolean isOid(String value) {
...
     public static Map<String, OidRecord> getOidCache(OidCategory oidCategory) {
         return oidCache.get(oidCategory);
     }
 
-    public static synchronized void cacheOidCategory(OidCategory category, Map<String, OidRecord> oidRecordMap) {
-        oidCache.put(category, oidRecordMap);
-        refreshRdnCodeLookup(category);
+    public static void cacheOidCategory(OidCategory category, Map<String, OidRecord> oidRecordMap) {
+        synchronized (WRITE_LOCK) {
+            oidCache.put(category, oidRecordMap);
```
klaraf755: **Optional hardening: the copy-on-write contract is still enforceable only by convention here.**

`cacheOidCategory` publishes the caller's map by reference, so mutate-after-publish still desynchronizes `oidCache` from the derived `rdnCodeToOid` index — exactly the failure mode the `AttributeEngineITest.ensureOidCached` fix in this PR removes. Since this is a hardening pass anyway, consider making the contract structural:

```java
oidCache.put(category, Collections.unmodifiableMap(new HashMap<>(oidRecordMap)));
```

Then the next `ensureOidCached`-style shortcut fails fast instead of silently going stale. Non-blocking.

klaraf755 (commented):
Two findings from review — one test-state leak worth fixing before merge, one optional hardening suggestion. Details inline.

src/test/java/com/otilm/core/oid/OidHandlerTest.java
lubomirw: Fixed in 2ba694ef4. Teardown now refreshes the derived `rdnCodeToOid` index to empty via `cacheOidCategory(category, new HashMap<>())` before `evictCategory`, so the null-original branch no longer leaves phantom RDN codes visible to later test classes. Applied your suggested fix verbatim.

src/main/java/com/otilm/core/oid/OidHandler.java
lubomirw: Applied in 2ba694ef4 — went structural across all three publish points (`cacheOidCategory`, `cacheOid`, `removeCachedOid`) rather than just `cacheOidCategory`, so `getOidCache` uniformly returns an unmodifiable map and any future mutate-after-publish fails fast. Verified all current `getOidCache` callers are read-only; 207 unit tests pass.

klaraf755 (dismissed):

PR #1871, opened 2026-07-23
Validate request-attribute definitions before persisting (platform settings & RA-profile static sets)

### LINKED ISSUE #1870: Request-attribute definitions are persisted without any backend validation (platform settings & RA-profile static sets)
### Description

Operator-authored request-attribute definitions are persisted without validation on both authoring paths. The platform default set (Settings → Certificates → Request Attributes) is serialized straight into the setting in `SettingServiceImpl.updateCertificateSettings` with no checks at all; the RA-profile static set path (`RaProfileCertificateRequestAttributeServiceImpl.updateConfiguration`) only checks that `properties` is non-null. The engine's definition validation (`AttributeEngine.validateAttributeDefinition` / `validateFieldMapping`) runs only on the connector-registration path, and nothing downstream re-validates authored definitions before certificate issue — the phase-3 design plan intended engine validation to be invoked at authoring, but it was never wired in.

This allows persisting definitions that are impossible or broken at request time, e.g.:

- required + read-only with no default value — the certificate request form becomes unsubmittable (FE counterpart: OmniTrustILM/fe-administrator#1913)
- definitions without a `fieldMapping` — dead weight, never projected into request content
- incoherent mappings, e.g. an unknown RDN code — fails only at PKCS#10 build during issuance
- default content not matching the declared content type, or violating the definition's constraints
- definitions with null `properties` — `NullPointerException` instead of a validation error

### Steps to Reproduce

1. In platform Settings → Certificates → Request Attributes (or on an RA profile's request-attribute set), add a definition marked Required and Read Only with no default value — it is accepted and persisted
2. Open a certificate request form for an RA profile that resolves this set
3. The required read-only field can never be filled; the request cannot be submitted. Variants: author a definition without a `fieldMapping` or with an unknown RDN code — accepted at authoring, inert or failing only at issuance

### Expected Behavior

Both authoring paths validate each definition at save time via the attribute engine: engine shape rules (label, valid UUID, read-only requires default content, read-only cannot be a list, multiselect requires list, properties present), a declared `fieldMapping` with at least one coherent field (resolvable RDN code / registered extension OID), and — when default content is provided — well-formed content (data present, conforming to the declared content type, satisfying constraints). Invalid definitions are rejected with an actionable `ValidationException`.

### Actual Behavior

Definitions are persisted unvalidated; the failures surface later as an unsubmittable certificate request form, silently ignored definitions, runtime errors at PKCS#10 build, or a `NullPointerException`.

### Severity

Major

### Module

Certificates

klaraf755:
Fixes #1870

### DIFF
--- a/src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
+++ b/src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
@@ -1280,6 +1280,10 @@ public class AttributeEngine {
         }
         validateResourceAttributeProperties(attribute, connectorUuidStr, attributeResource, hasCallback);
 
+        validateReadOnlyAttributeProperties(attribute, connectorUuidStr, readOnly, hasCallback, hasContent, list);
+    }
+
+    private static void validateReadOnlyAttributeProperties(BaseAttribute attribute, String connectorUuidStr, boolean readOnly, boolean hasCallback, boolean hasContent, boolean list) throws AttributeException {
         if (readOnly) {
             if (hasCallback) {
                 throw new AttributeException("Read only attribute cannot have callback", attribute.getUuid(), attribute.getName(), attribute.getType(), connectorUuidStr);
--- a/src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
+++ b/src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
@@ -402,6 +402,70 @@ public class AttributeEngine {
         }
     }
 
+    /**
+     * Authoring-time validation for platform-owned request-attribute definitions (the platform default set in
+     * Settings and RA-profile static sets). Every definition must be a v3 data attribute declaring a
+     * {@code fieldMapping} with at least one coherent field: an unmapped definition is never projected into
+     * request content (see CertificateRequestAttributeProjector), so in a platform-authored set it can only be
+     * dead weight — and nothing downstream re-validates an authored mapping before certificate issue. Default
+     * content, when provided, must be well-formed: non-null data, conforming to the declared content type, and
+     * satisfying the definition's constraints. Connector-registered attribute sets are validated on the
+     * registration path instead and may stay unmapped.
+     *
+     * @param definitions the authored definitions; {@code null} means "not updating" and is a no-op
+     * @throws ValidationException when any definition is malformed; messages are platform-authored and safe to expose
+     */
+    public static void validateRequestAttributeDefinitions(List<BaseAttribute> definitions) {
+        if (definitions == null) {
+            return;
+        }
+        Supplier<Map<String, String>> codeToOidMap = lazyCodeToOidMap();
+        for (BaseAttribute definition : definitions) {
+            String name = definition == null || definition.getName() == null ? "?" : definition.getName();
+            if (!(definition instanceof DataAttributeV3 v3)) {
+                throw new ValidationException("Request attribute definition '%s' must be a v3 data attribute".formatted(name));
+            }
+            // No explicit type check needed: the DataAttributeV3 constructor pins type = DATA, and the same
+            // field is Jackson's polymorphic discriminator, so class and type can never disagree on the wire.
+            if (v3.getFieldMapping() == null || v3.getFieldMapping().getFields() == null || v3.getFieldMapping().getFields().isEmpty()) {
+                throw new ValidationException("Request attribute definition '%s' must declare a field mapping with at least one mapped field".formatted(name));
+            }
+            try {
+                validateAttributeDefinition(v3, null);
+                validateFieldMapping(v3, null, codeToOidMap);
+            } catch (AttributeException e) {
+                // AttributeException messages are authored inside this class — safe to surface.
+                throw new ValidationException(e.getMessage());
+            }
+            validateDefaultContent(v3);
+        }
+    }
+
+    /**
+     * Default content of an authored definition, when present, must be usable at request time: every item
+     * carries data, conforms to the declared content type, and satisfies the definition's constraints.
+     */
+    private static void validateDefaultContent(DataAttributeV3 definition) {
+        List<? extends AttributeContent> content = definition.getContent();
+        if (content == null || content.isEmpty()) {
+            return;
+        }
+        for (AttributeContent item : content) {
+            if (item == null || item.getData() == null) {
+                throw new ValidationException("Request attribute definition '%s' default content is malformed and does not contain data".formatted(definition.getName()));
+            }
+            if (item.getContentType() != null && item.getContentType() != definition.getContentType()) {
+                throw new ValidationException("Request attribute definition '%s' default content does not match content type %s".formatted(definition.getName(), definition.getContentType().getLabel()));
+            }
+        }
+        List<ValidationError> constraintErrors = AttributeDefinitionUtils.validateConstraints(definition, content);
+        if (!constraintErrors.isEmpty()) {
+            throw new ValidationException("Request attribute definition '%s' default content violates constraints: %s".formatted(
+                    definition.getName(),
+                    constraintErrors.stream().map(ValidationError::getErrorDescription).collect(Collectors.joining("; "))));
+        }
+    }
+
     public AttributeDefinition updateCustomAttributeDefinition(CustomAttributeV3 customAttribute, List<Resource> resources) throws AttributeException {
         validateAttributeDefinition(customAttribute, null);
 
@@ -1120,7 +1184,7 @@ public class AttributeEngine {
         }
     }
 
-    private void validateAttributeDefinition(BaseAttribute attribute, UUID connectorUuid) throws AttributeException {
+    private static void validateAttributeDefinition(BaseAttribute attribute, UUID connectorUuid) throws AttributeException {
         String connectorUuidStr = connectorUuid == null ? null : connectorUuid.toString();
         if (attribute.getUuid() == null || !UUID_REGEX.matcher(attribute.getUuid()).matches()) {
             throw new AttributeException("Attribute does not have valid UUID", attribute.getUuid(), attribute.getName(), attribute.getType(), connectorUuidStr);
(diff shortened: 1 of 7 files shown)


src/test/java/com/otilm/core/attribute/engine/RequestAttributeDefinitionValidationTest.java
```diff
@@ -0,0 +1,184 @@
...
+                        .build());
+            }
+        }
+        OidHandler.cacheOidCategory(OidCategory.RDN_ATTRIBUTE_TYPE, rdn);
+    }
+
+    @AfterAll
+    static void clearRdnRegistry() {
+        OidHandler.cacheOidCategory(OidCategory.RDN_ATTRIBUTE_TYPE, new HashMap<>());
+    }
```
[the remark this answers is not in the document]
klaraf755: Added snapshots.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -402,6 +402,70 @@ public void updateCustomAttributeResources(UUID uuid, List<Resource> resources)
...
+     * request content (see CertificateRequestAttributeProjector), so in a platform-authored set it can only be
+     * dead weight — and nothing downstream re-validates an authored mapping before certificate issue. Default
+     * content, when provided, must be well-formed: non-null data, conforming to the declared content type, and
+     * satisfying the definition's constraints. Connector-registered attribute sets are validated on the
+     * registration path instead and may stay unmapped.
+     *
+     * @param definitions the authored definitions; {@code null} means "not updating" and is a no-op
+     * @throws ValidationException when any definition is malformed; messages are platform-authored and safe to expose
+     */
+    public static void validateRequestAttributeDefinitions(List<BaseAttribute> definitions) {
```
lubomirw, line 418: **[Low]** | Correctness _(revised after checking the frontend)_

This validation is materially stricter than the `validateDefinitionShape` it replaces (old check: `properties != null` only): it now requires a `DataAttributeV3` with STRING/TEXT content type, a non-empty valid `fieldMapping`, read-only⇒content, and known RDN codes. It runs on the **write** path, so a definition persisted before these rules existed will fail re-validation on the next save.

**Correction:** I originally worried that editing an unrelated certificate setting could be blocked by a legacy-invalid definition. Checked fe-administrator — that does **not** happen:
- Validation/registration are saved from a separate form (`CertificateSettingsForm.tsx`) whose payload **omits** `certificates.requestAttributes`, and the backend null-gates on it (`if (requestAttributes != null)`), so it isn't re-validated.
- Request-attribute definitions are edited on a **separate** screen (`RequestAttributesSettings.tsx`) that does a full read-modify-write of the whole definition set on every (auto-saved) change.

**Residual (narrow):** on the request-attributes screen, the full set is re-submitted, so if any definition persisted before these rules is now invalid, the operator can't save *any* change there until the legacy one is fixed. That's the screen that owns those definitions, so re-validating the set is largely intended. The seeded default set passes validation.

_Suggested:_ Confirm no production-stored default/static sets violate the new rules; if any might, add a one-time reconciliation. Otherwise no action needed — the broad regression risk is not present.

<sub>flagged by guided-pr-review + superpowers; revised after FE cross-check</sub>

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -402,6 +402,70 @@ public void updateCustomAttributeResources(UUID uuid, List<Resource> resources)
...
+    private static void validateDefaultContent(DataAttributeV3 definition) {
+        List<? extends AttributeContent> content = definition.getContent();
+        if (content == null || content.isEmpty()) {
+            return;
+        }
+        for (AttributeContent item : content) {
+            if (item == null || item.getData() == null) {
+                throw new ValidationException("Request attribute definition '%s' default content is malformed and does not contain data".formatted(definition.getName()));
+            }
+            if (item.getContentType() != null && item.getContentType() != definition.getContentType()) {
```
lubomirw: **[Low]** | Correctness

`item.getContentType() != null && item.getContentType() != definition.getContentType()` — a default-content item that carries data but reports a **null** contentType bypasses the type-conformance guard entirely. Concrete content classes return a non-null type so the tested paths work, but the null branch is an unchecked hole.

_Suggested:_ Document why a null item contentType is treated as conforming, or drop the null guard / reject null (the declared definition type is always non-null here, since `validateFieldMapping` already enforced STRING/TEXT).

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -402,6 +402,70 @@ public void updateCustomAttributeResources(UUID uuid, List<Resource> resources)
...
+            // field is Jackson's polymorphic discriminator, so class and type can never disagree on the wire.
+            if (v3.getFieldMapping() == null || v3.getFieldMapping().getFields() == null || v3.getFieldMapping().getFields().isEmpty()) {
+                throw new ValidationException("Request attribute definition '%s' must declare a field mapping with at least one mapped field".formatted(name));
+            }
+            try {
+                validateAttributeDefinition(v3, null);
+                validateFieldMapping(v3, null, codeToOidMap);
+            } catch (AttributeException e) {
+                // AttributeException messages are authored inside this class — safe to surface.
+                throw new ValidationException(e.getMessage());
```
lubomirw: **[Low]** | Error handling

`throw new ValidationException(e.getMessage())` discards the original `AttributeException` as cause, dropping its stack trace and structured context (uuid, name, type, connector) that the rest of the codebase surfaces.

_Suggested:_ Chain `e` as the cause (or log it at debug), preserving diagnostics while still surfacing the safe message to the caller.

src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
```diff
@@ -265,6 +266,7 @@ private void updateCertificateSettings(PlatformSettingsUpdateDto platformSetting
         CertificateRequestAttributesSettingsUpdateDto requestAttributes = platformSettings.getCertificates().getRequestAttributes();
         if (requestAttributes != null) {
             Setting definitionsSetting = certificateSetting(certificateSettings, DefaultRequestAttributeSet.SETTING_NAME);
+            AttributeEngine.validateRequestAttributeDefinitions(requestAttributes.getRequestAttributes());
```
lubomirw, line 268: **[Low]** | Error handling

Validation runs here, after earlier platform-settings sections have already called `settingRepository.save(...)`, so it is not fail-fast. No data corruption results (the class is `@Transactional` and `ValidationException` is a `RuntimeException`, so the write rolls back), but validating up front reads more clearly.

_Suggested:_ Optional — validate all request-attribute definitions at the top of the update method before any `save()`.

src/test/java/com/otilm/core/attribute/engine/RequestAttributeDefinitionValidationTest.java
```diff
@@ -0,0 +1,190 @@
...
+/**
+ * Plain unit test (no Spring context) for the authoring-time request-attribute definition validation.
+ * RDN-code mapping validation resolves against the OidHandler registry, seeded from SystemOid here.
+ */
+class RequestAttributeDefinitionValidationTest {
+
+    private static Map<String, OidRecord> savedRdnRegistry;
+
+    @BeforeAll
+    static void seedRdnRegistry() {
```
lubomirw, line 46: **[Low]** | Testing

`seedRdnRegistry` writes into the process-wide static `OidHandler` RDN cache in `@BeforeAll` and restores it in `@AfterAll`. Fragile under JUnit parallel execution, and the restore falls back to an empty map rather than the original `null` state when no cache existed. Fine for serial execution.

_Suggested:_ Document the shared-state assumption or isolate the OID seeding per-test.

<sub>flagged by superpowers + Copilot</sub>

src/test/java/com/otilm/core/attribute/engine/RequestAttributeDefinitionValidationTest.java
```diff
@@ -0,0 +1,190 @@
...
+        assertRejected(definition, "does not match content type");
+    }
+
+    @Test
+    void defaultContentViolatingConstraintsRejected() {
+        // commonNameAttribute carries a regexp constraint: max 64 characters
+        DataAttributeV3 definition = CsrAttributes.commonNameAttribute();
+        definition.setContent(List.of(new StringAttributeContentV3("x".repeat(65))));
+        assertRejected(definition, "violates constraints");
+    }
```
lubomirw, line 196: **[Nit]** | Maintainability

This new test file ends without a trailing newline.

_Suggested:_ Add a trailing newline at end of file.

lubomirw (commented):
Automated multi-reviewer triage — 6 findings (COMMENT). Reviewers: Copilot CLI, guided-pr-review, superpowers, hygiene. No High/Critical; all findings inline.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
klaraf755: Currently, there are no production stored default sets, since this is a new feature. I will check the develop environment after merge to see if anything needs to be adjusted.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
klaraf755: Removed the non-null guard and changed the type to `BaseAttributeContentV3` - v3 content must always declare its content type, otherwise it does get serialised, so the check for non-null is not needed.

src/test/java/com/otilm/core/attribute/engine/RequestAttributeDefinitionValidationTest.java
klaraf755: Added.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
klaraf755: Wrapped the exception with the cause.

src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
klaraf755: Moving it to the top of updatePlatformSettings would duplicate the getCertificates()/getRequestAttributes() null navigation this method already owns, just to fail a few statements earlier inside a single transaction. I've made the validation the first statement of the request-attributes block instead, so nothing in that block runs before it.

src/test/java/com/otilm/core/attribute/engine/RequestAttributeDefinitionValidationTest.java
klaraf755: Documented the shared-state and serial-execution assumption on the class instead. The empty-map fallback on restore is behaviorally identical to the original null state — getCodeToOidMap (and every other RDN-category reader) treats an absent category as empty — so exact-null restoration would need reflective eviction of the private cache, which isn't warranted here.

lubomirw (approved):

PR #1872, opened 2026-07-23
Reject unsupported merge mode and value-source bindings in RA Profile request attributes

### LINKED ISSUE #1862: Reject unsupported merge mode and value-source bindings in RA Profile request attributes
### Description

Connector request-attribute handling needs improvements before the RA Profile request-attributes configuration can fully support merge modes and value-source bindings. Until those improvements land, Core should reject any RA Profile request-attribute configuration that uses a merge mode other than `Static only`, or that includes value-source bindings, by throwing an operation-not-supported error. Keep the supporting code in place (do not delete) so it can be re-enabled when the improvements are delivered.

Note: this is the backend counterpart to the frontend change in fe-administrator#1908 (which hides the Merge mode selector and Value-source bindings section). Coordinate rollout with FE.

### Definition of Done

- [ ] Throw operation-not-supported when merge mode is not `Static only`
- [ ] Throw operation-not-supported when value-source bindings are present
- [ ] Supporting code retained (not removed) for future re-enable
- [ ] Change coordinated with FE (fe-administrator#1908)

klaraf755:
Closes #1862

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
@@ -2,10 +2,7 @@ package com.otilm.core.service.impl;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.otilm.api.exception.ConnectorException;
-import com.otilm.api.exception.NotFoundException;
-import com.otilm.api.exception.ValidationError;
-import com.otilm.api.exception.ValidationException;
+import com.otilm.api.exception.*;
 import com.otilm.api.model.common.attribute.common.BaseAttribute;
 import com.otilm.api.model.common.attribute.v3.DataAttributeV3;
 import com.otilm.api.model.common.attribute.v3.mapping.SourceParam;
@@ -131,6 +128,13 @@ public class RaProfileCertificateRequestAttributeServiceImpl implements RaProfil
     @Override
     public void updateConfiguration(RaProfile raProfile, RaProfileCertificateRequestAttributesUpdateDto request) {
         validateDefinitionShape(request.getRequestAttributes());
+        AttributeSetMergeMode effectiveMode = RequestAttributeSetResolver.effectiveMode(request.getMergeMode());
+        if (effectiveMode != AttributeSetMergeMode.STATIC_ONLY) {
+            throw new NotSupportedException(String.format("Merge mode %s is not supported. Use `Static Only` mode.", effectiveMode));
+        }
+        if (request.getValueSourceBindings() != null && !request.getValueSourceBindings().isEmpty()) {
+            throw new NotSupportedException("Value-source bindings are not supported in this version. Use the `Static Only` mode without bindings.");
+        }
         writer.saveStaticSet(
                 raProfile,
                 AttributeDefinitionUtils.serialize(request.getRequestAttributes()),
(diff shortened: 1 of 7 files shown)


src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
```diff
@@ -131,6 +128,13 @@ public RaProfileCertificateRequestAttributesDto getConfiguration(RaProfile raPro
     @Override
     public void updateConfiguration(RaProfile raProfile, RaProfileCertificateRequestAttributesUpdateDto request) {
         validateDefinitionShape(request.getRequestAttributes());
+        AttributeSetMergeMode effectiveMode = RequestAttributeSetResolver.effectiveMode(request.getMergeMode());
+        if (effectiveMode != AttributeSetMergeMode.STATIC_ONLY) {
```
lubomirw, line 132: **[Medium]** | Correctness

Read view advertises a merge mode the write path now rejects. `getConfiguration` returns `effectiveMode(...)` = `MERGE` for any never-configured profile, but this gate 501s on anything != `STATIC_ONLY` (including an omitted mode). A naive GET->modify->PUT, or PUT-back unchanged, breaks; the FE must always send `STATIC_ONLY`.

_Suggested:_ make read and write consistent while gated — expose `STATIC_ONLY` from the read/default path, or confirm the FE never echoes the read `mergeMode` and always submits `STATIC_ONLY`.

<sub>flagged by Copilot CLI + superpowers code-review</sub>

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
```diff
@@ -131,6 +128,13 @@ public RaProfileCertificateRequestAttributesDto getConfiguration(RaProfile raPro
...
+            throw new NotSupportedException(String.format("Merge mode %s is not supported. Use `Static Only` mode.", effectiveMode));
```
lubomirw: **[Low]** | Error handling

`NotSupportedException` maps to HTTP 501 Not Implemented (`ExceptionHandlingAdvice:549`). A client-supplied but currently-disallowed value is a client error, not a missing server capability — 501 invites 5xx retry logic.

_Suggested:_ map this rejection to HTTP 422 (a validation-style exception) instead of `NotSupportedException`/501.

<sub>flagged by superpowers code-review (+ Copilot CLI)</sub>

lubomirw: **[Low]** | Error handling

When `mergeMode` is omitted, `effectiveMode()` resolves to `MERGE`, so the error reads `Merge mode MERGE is not supported` though the caller never sent MERGE (locked in by `omittedMergeModeIsRejectedAsEffectiveMerge`). It also surfaces the raw enum vs the `Static Only` label used in the same sentence.

_Suggested:_ distinguish the null case — e.g. `Merge mode is required and must be Static Only`.

lubomirw (commented):
Automated multi-reviewer triage — 4 findings (COMMENT). 3 inline, 1 off-diff below.

**Off-diff (docs):**
- `interface-documentation` — `openapi/doc-openapi-core-ra-profile.yaml:4469-4494` — **[Low] | Documentation** — `RaProfileCertificateRequestAttributesUpdateDto` still shows `default: merge`, advertises `connectorOnly`/`merge`, and presents `valueSourceBindings` as freely configurable — all now rejected by this gate. _Suggested:_ update the source annotations in the interfaces DTO to state only `staticOnly` is accepted this version, omitting the mode is rejected, and bindings must be empty. (flagged by guided-pr-review docs check)

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
klaraf755: Merge mode is not required and defaults to Static Only now, so null in request is expected work like `Static Only`.

klaraf755: Changed to `ValidationException`.

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
klaraf755: `getConfiguration` has been changed to return `STATIC_ONLY` as default for not configured profile.

lubomirw (approved):

PR #1881, opened 2026-07-24
Add connector WebClient tuning configuration

### LINKED ISSUE #1878: Tune the connector WebClient: response/connect timeouts and connection pool
## Problem

The shared `WebClient` used for all connector calls is built by `BaseApiClient.prepareWebClient()` (`interfaces` repo, `BaseApiClient.java:192-201`) and wired as a single bean in `core` (`ApplicationConfig.webClient()`, `ApplicationConfig.java:75-78`). It sets only the codec max-in-memory size — **no response/read timeout, no connect timeout, and no explicit `ConnectionProvider`**. Calls block synchronously via `.block()` (`BaseApiClient.java:209-229`).

Consequences:
- A hung or slow connector/CA response blocks the caller indefinitely (and, combined with the transaction-boundary issue, can pin a DB connection) — there is no fail-fast.
- The default Reactor-Netty per-host pool (~16 connections) throttles connector concurrency below the target issuance concurrency.
- `SslContext` is rebuilt (keystore base64-decode + parse) on **every** CERTIFICATE-auth request (`BaseApiClient.java:100-104`), wasting CPU per call.

## Change

Configure the connector WebClient (in `core`'s `ApplicationConfig.webClient()`, or the shared `prepareWebClient()` if kept common):
- `responseTimeout` ~35 s (authority connector `PerCallSeconds` 30 + margin) and a connect timeout (~3 s).
- An explicit `ConnectionProvider` with `maxConnections` >= the actions-listener concurrency, plus a short `pendingAcquireTimeout`.
- Cache the `SslContext` per connector for CERTIFICATE auth instead of rebuilding it per request.

## Acceptance criteria

- A hung connector call fails fast within the configured response timeout rather than blocking indefinitely.
- The connection pool is sized to the configured issuance concurrency (not the default ~16).
- `SslContext` is not rebuilt on every CERTIFICATE-auth request.

lubomirw:
## What

Expose the connector `WebClient` tuning as configuration and wire it into the shared `webClient` bean.

- New `connector.api-client.*` properties (`ConnectorApiClientProperties`): `response-timeout` (35s), `connect-timeout` (3s), `max-connections` (20), `pending-acquire-timeout` (10s) — all env-overridable, following the `provisioning.api` / `validation.crl` precedent. Validated at startup (durations strictly positive, `max-connections` positive) so misconfiguration fails fast.
- `ApplicationConfig.webClient()` maps them to `BaseApiClient.prepareWebClient(ClientTuning)`.
- `ConnectorServiceImpl.evictConnectorCache` also evicts the interfaces-side CERTIFICATE WebClient cache, so a deleted or reconfigured connector does not retain its TLS client.

## Why yml/env, not a Settings-UI tunable

This is boot-time infrastructure sizing by design. The interfaces side applies tuning **write-once per JVM** — the Reactor-Netty `ConnectionProvider` (and the timeouts baked into the shared `HttpClient`) cannot be swapped live once the pool is in use — so these values genuinely cannot be a runtime setting. They belong in yml/env, applied at startup.

## Follow-up

The four `CONNECTOR_API_CLIENT_*` env overrides still need wiring into the Helm charts (values + deployment env, same pattern as `CBOM_CLIENT_MAX_BUFFER_SIZE`) to be reachable in deployments — separate `helm-charts` PR. Defaults apply until then, so deployments remain functional.

## Depends on

Requires the `interfaces` change that adds `ClientTuning`, `prepareWebClient(ClientTuning)`, and `evictCertificateClient`: OmniTrustILM/interfaces#788. Merge and publish that first, or CI here will not resolve them.

Relates to #1878

### DIFF
--- a/src/main/java/com/otilm/core/config/ApplicationConfig.java
+++ b/src/main/java/com/otilm/core/config/ApplicationConfig.java
@@ -4,6 +4,7 @@ import com.otilm.api.clients.AttributeApiClient;
 import com.otilm.api.clients.AuthorityInstanceApiClient;
 import com.otilm.api.clients.BaseApiClient;
 import com.otilm.api.clients.CertificateApiClient;
+import com.otilm.api.clients.ClientTuning;
 import com.otilm.api.clients.ComplianceApiClient;
 import com.otilm.api.clients.ConnectorApiClient;
 import com.otilm.api.clients.DiscoveryApiClient;
@@ -42,7 +43,7 @@ import javax.net.ssl.TrustManager;
 
 @Configuration
 @EnableJpaAuditing(auditorAwareRef = "auditorAware")
-@EnableConfigurationProperties(DiscoveryProperties.class)
+@EnableConfigurationProperties({DiscoveryProperties.class, ConnectorClientProperties.class})
 @ComponentScan(basePackages = "com.otilm.core",
         excludeFilters = @ComponentScan.Filter(type = FilterType.CUSTOM, classes = TypeExcludeFilter.class))
 public class ApplicationConfig {
@@ -73,8 +74,12 @@ public class ApplicationConfig {
     }
 
     @Bean
-    public WebClient webClient() {
-        return BaseApiClient.prepareWebClient();
+    public WebClient webClient(ConnectorClientProperties connectorClientProperties) {
+        return BaseApiClient.prepareWebClient(new ClientTuning(
+                connectorClientProperties.connectTimeout(),
+                connectorClientProperties.responseTimeout(),
+                connectorClientProperties.maxConnections(),
+                connectorClientProperties.pendingAcquireTimeout()));
     }
 
     @Bean
--- a/src/main/java/com/otilm/core/config/ApplicationConfig.java
+++ b/src/main/java/com/otilm/core/config/ApplicationConfig.java
@@ -43,7 +43,7 @@ import javax.net.ssl.TrustManager;
 
 @Configuration
 @EnableJpaAuditing(auditorAwareRef = "auditorAware")
-@EnableConfigurationProperties({DiscoveryProperties.class, ConnectorClientProperties.class})
+@EnableConfigurationProperties({DiscoveryProperties.class, ConnectorApiClientProperties.class})
 @ComponentScan(basePackages = "com.otilm.core",
         excludeFilters = @ComponentScan.Filter(type = FilterType.CUSTOM, classes = TypeExcludeFilter.class))
 public class ApplicationConfig {
@@ -74,12 +74,12 @@ public class ApplicationConfig {
     }
 
     @Bean
-    public WebClient webClient(ConnectorClientProperties connectorClientProperties) {
+    public WebClient webClient(ConnectorApiClientProperties connectorApiClientProperties) {
         return BaseApiClient.prepareWebClient(new ClientTuning(
-                connectorClientProperties.connectTimeout(),
-                connectorClientProperties.responseTimeout(),
-                connectorClientProperties.maxConnections(),
-                connectorClientProperties.pendingAcquireTimeout()));
+                connectorApiClientProperties.connectTimeout(),
+                connectorApiClientProperties.responseTimeout(),
+                connectorApiClientProperties.maxConnections(),
+                connectorApiClientProperties.pendingAcquireTimeout()));
     }
 
     @Bean
--- a/src/main/java/com/otilm/core/config/ConnectorClientProperties.java
+++ b/src/main/java/com/otilm/core/config/ConnectorClientProperties.java
@@ -1,0 +1,17 @@
+package com.otilm.core.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+import java.time.Duration;
+
+/**
+ * Tuning for the shared connector WebClient — response/connect timeouts and the connection pool
+ * used for all HTTP calls to connectors and authorities. Bound from {@code connector.client.*}.
+ */
+@ConfigurationProperties(prefix = "connector.client")
+public record ConnectorClientProperties(
+        Duration connectTimeout,
+        Duration responseTimeout,
+        int maxConnections,
+        Duration pendingAcquireTimeout) {
+}
--- a/src/main/java/com/otilm/core/config/ConnectorClientProperties.java
+++ b/src/main/java/com/otilm/core/config/ConnectorClientProperties.java
@@ -1,17 +1,22 @@
 package com.otilm.core.config;
 
+import jakarta.validation.constraints.NotNull;
+import jakarta.validation.constraints.Positive;
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
 
 import java.time.Duration;
 
 /**
- * Tuning for the shared connector WebClient — response/connect timeouts and the connection pool
- * used for all HTTP calls to connectors and authorities. Bound from {@code connector.client.*}.
+ * Tuning for the shared WebClient — response/connect timeouts and the connection pool — used for all
+ * outbound API-client calls. Bound from {@code connector.client.*}; validated at startup so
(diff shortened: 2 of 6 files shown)


src/main/java/com/otilm/core/config/ConnectorClientProperties.java
```diff
@@ -0,0 +1,17 @@
...
+ * Tuning for the shared connector WebClient — response/connect timeouts and the connection pool
+ * used for all HTTP calls to connectors and authorities. Bound from {@code connector.client.*}.
+ */
+@ConfigurationProperties(prefix = "connector.client")
+public record ConnectorClientProperties(
+        Duration connectTimeout,
+        Duration responseTimeout,
+        int maxConnections,
+        Duration pendingAcquireTimeout) {
+}
```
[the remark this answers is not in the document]
lubomirw: Fixed in b477a24e6 — ConnectorClientProperties is now @Validated with @NotNull on the durations and @Positive on maxConnections, so missing/invalid tuning fails fast at startup (matching ProvisioningApiProperties). The interfaces-side ClientTuning also validates positivity at construction as a second line of defense.

src/main/resources/application.yml
```diff
@@ -331,6 +331,17 @@ validation:
     read-timeout: 1000
     connect-timeout: 1000
 
+# Shared WebClient tuning applied to every outbound API-client call (any provider implementation)
+connector:
```
klaraf755, line 335: These four env vars (`CONNECTOR_CLIENT_RESPONSE_TIMEOUT`, `CONNECTOR_CLIENT_CONNECT_TIMEOUT`, `CONNECTOR_CLIENT_MAX_CONNECTIONS`, `CONNECTOR_CLIENT_PENDING_ACQUIRE_TIMEOUT`) also need to be wired into the Helm charts (values + deployment env, same pattern as `CBOM_CLIENT_MAX_BUFFER_SIZE`) — without that, the override isn't reachable in deployments.

src/main/resources/application.yml
```diff
@@ -331,6 +331,17 @@ validation:
...
+  client:
+    # Fail fast when a remote endpoint stops responding instead of blocking the caller indefinitely
+    response-timeout: ${CONNECTOR_CLIENT_RESPONSE_TIMEOUT:35s}
```
klaraf755: The defaults here duplicate `ClientTuning.defaults()` in interfaces (35s/3s/20/10s): tests and untuned callers use that, deployments use these. If one side changes later, they silently diverge — worth a keep-in-sync cross-reference comment on one side.

src/main/java/com/otilm/core/config/ConnectorClientProperties.java
```diff
@@ -0,0 +1,22 @@
...
+import java.time.Duration;
+
+/**
+ * Tuning for the shared WebClient — response/connect timeouts and the connection pool — used for all
+ * outbound API-client calls. Bound from {@code connector.client.*}; validated at startup so
+ * misconfiguration fails fast with a clear error rather than a runtime NPE or an invalid pool.
+ */
+@ConfigurationProperties(prefix = "connector.client")
+@Validated
+public record ConnectorClientProperties(
```
klaraf755: Worth stating in the PR description why this is yml/env rather than a Settings-UI tunable: the interfaces side made tuning write-once per JVM (the `ConnectionProvider` can't be swapped live), so this is boot-time infrastructure sizing by design and genuinely can't be a runtime setting. Putting that rationale on record pre-empts the question from future readers.

src/main/java/com/otilm/core/config/ApplicationConfig.java
```diff
@@ -73,8 +74,12 @@ public AuditorAware<String> auditorAware() {
     }
```
klaraf755, line 75: Follow-up once interfaces exposes the cert-client cache evict hook (see the review on OmniTrustILM/interfaces#788): call it from `com.otilm.core.service.v2.impl.ConnectorServiceImpl` — `deleteConnector`, `forceDeleteConnector`, and `editConnector` — so a deleted connector doesn't leave its cached TLS client (`SslContext` with parsed key material) in memory until restart. Evicting unconditionally on edit is the simplest rule: credential rotation already self-invalidates via the hash, and the entry rebuilds lazily on the next request.

klaraf755 (commented):
Clean and minimal — validation is two-layered and both layers fire at startup (jakarta at binding, `ClientTuning`'s constructor for zero/negative durations), and this bean is the only `prepareWebClient` call site in core, so the write-once semantics are safe. Four inline notes below.

src/main/java/com/otilm/core/config/ConnectorClientProperties.java
```diff
@@ -0,0 +1,22 @@
...
+        @NotNull Duration connectTimeout,
+        @NotNull Duration responseTimeout,
+        @Positive int maxConnections,
+        @NotNull Duration pendingAcquireTimeout) {
+}
```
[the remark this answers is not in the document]
lubomirw: Fixed in 2987fd1f2 — the record constructor now rejects zero/negative (and null) durations at binding, so 0s/-1s fails fast with a clear error instead of disabling timeouts. maxConnections keeps @Positive.

src/main/resources/application.yml
lubomirw: Addressed in 2987fd1f2 — added a keep-in-sync note on the ConnectorClientProperties Javadoc cross-referencing ClientTuning.defaults().

src/main/java/com/otilm/core/config/ConnectorClientProperties.java
lubomirw: Added to the PR description ("Why yml/env, not a Settings-UI tunable"): tuning is write-once per JVM because the ConnectionProvider can't be swapped live, so it's boot-time infra sizing by design.

src/main/java/com/otilm/core/config/ApplicationConfig.java
lubomirw: Already wired in d425e2db8: ConnectorServiceImpl.evictConnectorCache (hit by editConnector, deleteConnector, forceDeleteConnector, bulk) now calls BaseApiClient.evictCertificateClient(uuid). Evicts unconditionally as you suggested; credential rotation also self-invalidates via the hash.

src/main/resources/application.yml
lubomirw: Valid — the four CONNECTOR_CLIENT_* overrides need Helm wiring (values + deployment env, like CBOM_CLIENT_MAX_BUFFER_SIZE) to be reachable. Tracking as a separate helm-charts PR; noted in the PR description under Follow-up. Defaults apply meanwhile so deployments stay functional. Leaving this thread open until the helm-charts PR lands.

src/main/resources/application.yml
```diff
@@ -331,6 +331,17 @@ validation:
...
+# Shared WebClient tuning applied to every outbound API-client call (any provider implementation)
+connector:
+  api-client:
+    # Fail fast on unresponsive endpoints; keep connect + response timeouts below the 120s transaction timeout.
+    response-timeout: ${CONNECTOR_API_CLIENT_RESPONSE_TIMEOUT:35s}
+    connect-timeout: ${CONNECTOR_API_CLIENT_CONNECT_TIMEOUT:3s}
+    # Connection-pool limit per remote host; size to the expected concurrent load
+    max-connections: ${CONNECTOR_API_CLIENT_MAX_CONNECTIONS:20}
+    # Shed load quickly under pool saturation rather than queueing behind the response timeout
+    pending-acquire-timeout: ${CONNECTOR_API_CLIENT_PENDING_ACQUIRE_TIMEOUT:10s}
```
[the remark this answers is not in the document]
lubomirw: Aligned the PR description with the implemented keys: connector.api-client.* / CONNECTOR_API_CLIENT_* (and the ConnectorApiClientProperties class name). The Helm follow-up note now references the correct env var names too.

3keyroman (approved):

PR #1890, opened 2026-07-26
Keep connector round-trips out of the issuance transaction

### LINKED ISSUE #1877: Keep the issue-attribute connector round-trip out of the request DB transaction
## Problem

On `POST .../certificates` (issue), `ClientOperationServiceImpl.issueCertificate` opens a manual DB transaction (`ClientOperationServiceImpl.java:413`) and calls `submitCertificateRequest` inside it (`:415`, committing at `:416`). That path reaches `CertificateServiceImpl.submitCertificateRequest:1738` → `ExtendedAttributeService.mergeAndValidateIssueAttributes`, whose implementation (`ExtendedAttributeServiceImpl.java:82-98`) makes **two outbound connector HTTP calls** — `adapter.validateIssueAttributes(...)` (`:92`) and `adapter.listIssueAttributes(...)` (`:95`) — **while the DB transaction (and its Hikari connection) is held open**.

Under parallel issuance this holds a pooled DB connection for the duration of each connector round-trip. With virtual threads enabled (unbounded inbound concurrency) and Hikari `maximumPoolSize=40`, the pool exhausts and requests block up to `connectionTimeout` (30 s). Observed in throughput testing: Core became blocked and "ran out of transactions in the pool" under many parallel workers; **increasing the pool size did not recover throughput** — the constraint is connection *hold time* across the HTTP call, not pool size.

Note the actual CA issue submit is already correctly outside any transaction (claim-commit-before-HTTP on the JMS path); this issue is specifically the attribute validate/list round-trip on the synchronous request path.

## Change

Remove the connector HTTP round-trip from the open transaction, either by:
- moving `mergeAndValidateIssueAttributes` (the `validateIssueAttributes` + `listIssueAttributes` calls) out of the manual transaction boundary, or
- caching the issue-attribute definitions per (connector, RA profile) with a short TTL — the authority connector's issue-attribute registry is static, so this removes the round-trip from the hot path entirely.

## Acceptance criteria

- No DB connection is held across a connector HTTP call on the issuance request path.
- Under parallel issuance load, `hikaricp.connections.pending` stays ~0 and the pool no longer exhausts.

lubomirw:
## Problem

On `POST .../certificates` the certificate-request path held a pooled Hikari DB connection across connector HTTP calls — the issue-attribute merge (`validate` + `list`) and the CSR build (uploaded-CSR validation, key-generation signing) all ran inside the request transaction. Under parallel issuance with virtual threads and a 40-connection pool, connection hold-time exhausted the pool; enlarging the pool did not help.

## Change

`ClientOperationServiceImpl.submitCertificateRequest` now builds the CSR and merges/validates issue-attributes **before** opening a `REQUIRED` transaction that wraps only the local persistence, and rethrows raw. `issueCertificate` / `renewCertificate` / `rekeyCertificate` (all `NOT_SUPPORTED`) drop their own manual transaction and self-invoke it inside their existing `CertificateOperationException` / `RequestAttributePolicyViolationException` mapping catch. The merge moves out of `CertificateServiceImpl.submitCertificateRequest`; `CryptographicOperationServiceImpl.generateCsr` becomes `NOT_SUPPORTED` so key-generation signing holds no connection. The original order (uploaded-CSR validation before the issue-attribute merge) is preserved.

Result: on the self-invoked issue/renew/rekey (REST v2) paths no DB connection is held across any connector HTTP call. Proxied callers (REST v1 submit, SCEP manual-approval) run under the class-level transaction as before — unchanged, not a regression.

## Regression safety

Exception contract for every protocol consumer (SCEP auto/manual-approval, CMP ir-cr/kur, ACME finalize, Locations) is preserved: only `CertificateOperationException` / `RequestAttributePolicyViolationException` / declared checked types escape.

## Testing

- New integration guard asserting no transaction is active when the issue-attribute merge runs (verified failing pre-fix, passing after).
- Regression: protocol / crypto / issuance suites green (SCEP, CMP, ACME, `CryptographicOperationServiceITest`, Locations, uploaded-CSR and attribute-propagation).
- Context-signature guard unchanged.

## Deferred (own follow-ups, not this PR)

- Extending the fix to the proxied REST v1 / SCEP manual-approval paths (converting `submitCertificateRequest` to `NOT_SUPPORTED`) — blocked on the SCEP manual-approval atomicity question (the certificate row would commit before the `scep_transaction` row).
- Protocol-service outer-transaction suspension (ACME / SCEP / CMP hold a suspended-transaction connection across `issueCertificate`).
- TTL cache of issue-attribute definitions.

Closes #1877

### Attribute-definition upserts

The issue-attribute merge refreshes the CERTIFICATE_ISSUE attribute definitions. Now that the merge runs before the persistence transaction on the self-invoked paths, those definition writes commit independently: if the issuance transaction later rolls back, the refreshed definitions remain. That is intended and harmless — the definitions mirror the authority connector's schema, the upsert is idempotent, and the same upsert already runs from the RA/ACME/SCEP/CMP profile-configuration paths.


lubomirw:
### Deferred follow-up — SCEP manual-approval atomicity (before extending the fix to the proxied paths)

This PR keeps the connector round-trips out of the transaction on the **self-invoked** REST v2 `issueCertificate`/`renewCertificate`/`rekeyCertificate` paths (the `POST .../certificates` hot path in #1877). The two **proxied** callers — REST v1 submit (`CertificateControllerImpl`) and SCEP manual-approval (`ScepServiceImpl.generateCsr`) — still enter `submitCertificateRequest` under the class-level `@Transactional`, so they hold a connection across the build + merge exactly as they did before this PR (unchanged, not a regression).

Extending the fix to them by making `submitCertificateRequest` itself `@Transactional(NOT_SUPPORTED)` is **deliberately deferred**, because it changes atomicity on the SCEP manual-approval path:

- Today the SCEP manual-approval flow runs the certificate-request persistence and the `scep_transaction` insert (`addTransactionEntity`) inside the one SCEP outer transaction.
- If `submitCertificateRequest` opened its own committed transaction, the certificate row would commit **before** the `scep_transaction` row. A failure between them would leave an orphaned `REQUESTED` certificate with no SCEP transaction record, and a client retry (which keys off `existsByTransactionId`) would create a **duplicate** certificate — certificate rows are not fingerprint-deduped, only `certificate_request` is.

So the proxied-path improvement should be a separate change that first decides the SCEP manual-approval commit ordering (e.g. persist the `scep_transaction` row in the same explicit transaction as the certificate, or make the SCEP submit idempotent on retry). Tracking as a follow-up rather than folding it in here.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -409,17 +427,15 @@ public ClientCertificateDataResponseDto issueCertificate(final SecuredParentUUID
...
             certificate = submitCertificateRequest(certificateRequestDto, protocolInfo);
-            transactionManager.commit(status);
         } catch (RequestAttributePolicyViolationException e) {
-            transactionManager.rollback(status);
             // A client-facing validation failure; must reach the protocol unchanged.
             throw e;
         } catch (Exception e) {
-            transactionManager.rollback(status);
             throw new CertificateOperationException("Failed to submit certificate request: " + e.getMessage());
         }
```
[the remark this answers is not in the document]
lubomirw: Fixed in 551b82c7b — the three catch sites (issue/renew/rekey) now go through a `submitAndShapeFailure` helper that logs the cause (with stack) and surfaces a generic `CertificateOperationException` message instead of appending `e.getMessage()`, so the raw exception message no longer reaches the error response.

3keyroman (dismissed):

klaraf755 (commented):
Reviewed the transaction restructuring — mechanics look correct (commit-time guard, eager RA-profile fetch, exception contract for protocol consumers all check out). A few suggestions inline.

One nit that GitHub won't let me anchor (line outside the diff): the class-level `@Transactional` comment on `ClientOperationServiceImpl` ("the methods governed here are reads") is now slightly misleading — `submitCertificateRequest` is a write that stays under the class-level transaction on the proxied paths (REST v1, SCEP manual approval). Since this PR re-shapes exactly these boundaries, worth a one-line amendment.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -364,25 +363,61 @@ public CertificateDetailDto submitCertificateRequest(ClientCertificateRequestDto
...
+     * {@link RequestAttributePolicyViolationException} passes through unchanged; any other failure is logged with
+     * its cause and surfaced as a {@link CertificateOperationException} carrying the generic {@code failureMessage},
+     * so a raw exception message cannot leak internal detail through the error response.
+     */
+    private CertificateDetailDto submitAndShapeFailure(ClientCertificateRequestDto request, CertificateProtocolInfo protocolInfo, String failureMessage) throws CertificateOperationException {
+        try {
+            return submitCertificateRequest(request, protocolInfo);
+        } catch (RequestAttributePolicyViolationException e) {
+            throw e;
+        } catch (Exception e) {
```
klaraf755, line 422: Dropping the raw message is right, but this also swallows our own domain exceptions: a `ValidationException` from the merge ("connector unavailable", disabled RA profile hit inside submit) or a `NotFoundException` now reaches ACME/CMP/REST clients as an unactionable generic message — previously the `": " + e.getMessage()` suffix carried the reason. Consider the gated-exposure pattern instead of all-or-nothing:

```java
} catch (Exception e) {
    logger.error(failureMessage, e);
    String detail = (e instanceof ValidationException || e instanceof NotFoundException) && e.getMessage() != null
            ? failureMessage + ": " + e.getMessage()
            : failureMessage;
    throw new CertificateOperationException(detail);
}
```

Keeps the type contract for protocol consumers, keeps runtime detail off the wire, restores actionable errors for client mistakes.

src/main/java/com/otilm/core/service/impl/CryptographicOperationServiceImpl.java
```diff
@@ -433,7 +434,9 @@ public RandomDataResponseDto randomData(SecuredUUID tokenInstanceUuid, RandomDat
     }
 
     @Override
-    @Transactional
+    // Read-only (key reads + connector signing over HTTP); NOT_SUPPORTED keeps the DB connection out of the
+    // crypto-connector round-trip so it is not held while signing (certificate key-generation path).
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
```
klaraf755, line 439: Under `NOT_SUPPORTED`, `getPublicAndPrivateKey` iterates `key.getItems()` and the signer traverses `privateKeyItem.getKey().getTokenInstanceReference()` — all LAZY, on an entity whose loading transaction has closed. Today this survives via default OSIV on web threads (and the ITest is green), but `findWithAuthorityByUuid`'s Javadoc in this same PR establishes "don't depend on open-session-in-view" as the convention. Suggest the same treatment here: an `@EntityGraph(attributePaths = {"items", "tokenProfile", "tokenInstanceReference"})` finder on `CryptographicKeyRepository`. Also removes the per-lazy-load connection acquisitions.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -364,25 +363,61 @@ public CertificateDetailDto submitCertificateRequest(ClientCertificateRequestDto
...
+        // must run before the issue-attribute merge.
         String certificateRequest = generateBase64EncodedCsr(request.getRequest(), request.getFormat(), request.getCsrAttributes(), request.getKeyUuid(), request.getTokenProfileUuid(), request.getSignatureAttributes(), request.getAltKeyUuid(), request.getAltTokenProfileUuid(), request.getAltSignatureAttributes(), raProfile);
-        CertificateDetailDto certificate = certificateService.submitCertificateRequest(certificateRequest, request.getFormat(), request.getSignatureAttributes(), request.getAltSignatureAttributes(), request.getCsrAttributes(), request.getIssueAttributes(), request.getKeyUuid(), request.getAltKeyUuid(), request.getRaProfileUuid(), request.getSourceCertificateUuid(),
-                protocolInfo);
-
-        // create custom Attributes
-        if (createCustomAttributes) {
-            certificate.setCustomAttributes(attributeEngine.updateObjectCustomAttributesContent(Resource.CERTIFICATE, UUID.fromString(certificate.getUuid()), request.getCustomAttributes()));
+        if (raProfile != null) {
+            extendedAttributeService.mergeAndValidateIssueAttributes(raProfile, request.getIssueAttributes());
```
klaraf755, line 383: `validateUpdateDataAttributes` inside the merge persists refreshed connector attribute definitions. Now that this runs before the transaction on the self-invoked paths, those writes commit independently — if the persistence transaction later rolls back, the definition updates stay. Almost certainly fine (they mirror the connector and look idempotent), but worth confirming that's intended and maybe a sentence in the PR description.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1734,8 +1727,9 @@ public CertificateDetailDto submitCertificateRequest(
             UUID predecessorCertificateUuid,
             CertificateProtocolInfo protocolInfo
     ) throws NoSuchAlgorithmException, ConnectorException, AttributeException, CertificateRequestException, NotFoundException {
+        // Issue-attribute merge/validation is performed by the caller before the request transaction opens,
```
klaraf755, line 1730: This precondition ("caller merges/validates issue attributes first") lives only in an impl-side comment. A future caller of `CertificateService.submitCertificateRequest` that doesn't know it will silently skip connector-side issue-attribute validation. Suggest stating it in the interface method's Javadoc so it's visible at the call site.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Fixed in 7ef7b89b5 — gated exposure as suggested: `ValidationException` / `NotFoundException` messages are appended, everything else stays generic. Narrowed to those two rather than the `PlatformException` marker on purpose, since `ConnectorException` implements it too and `ConnectorServerException` relays upstream connector text. Covered both directions with tests (`issueCertificate_keepsDomainExceptionDetailInFailureMessage`, `issueCertificate_wrapsPersistenceFailureAsGenericCertificateOperationException`).

src/main/java/com/otilm/core/service/impl/CryptographicOperationServiceImpl.java
lubomirw: Fixed in 7ef7b89b5 — added `findWithKeyItemsAndTokenByUuid` with `@EntityGraph({"tokenProfile", "items", "tokenInstanceReference"})` and switched `getPublicAndPrivateKey` to it. All three paths are dereferenced on the signing path (`tokenProfile` at the profile check, `items` in the keypair loop, `tokenInstanceReference` via `privateKeyItem.getKey()` when resolving the connector), and `generateCsr` is the only caller of that helper, so the scope is contained.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
lubomirw: Intended, yes. The definitions mirror the connector schema and the upsert is idempotent — the same one runs from the profile-CRUD paths — so an issuance rollback leaving refreshed definitions behind is harmless. Called out in the PR description now.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
lubomirw: Fixed in 7ef7b89b5 — the precondition is now in the `CertificateInternalService.submitCertificateRequest` Javadoc, stating that the caller must merge/validate the issue attributes first (both for connector-side validation and to register the definitions the method writes content against) and that callers do it outside their persistence transaction.

lubomirw:
Also addressed the unanchored nit in 7ef7b89b5: the class-level `@Transactional` comment on `ClientOperationServiceImpl` no longer claims the methods it governs are reads. It now says the issue/renew/rekey entry points override it with `NOT_SUPPORTED` and manage the persistence transaction internally, while `submitCertificateRequest` stays under it when reached through the proxy (REST v1 submit, SCEP manual approval) and runs with no ambient transaction when self-invoked.

klaraf755 (approved):

PR #1891, opened 2026-07-27
Allow deleting scheduled jobs with history

### LINKED ISSUE #1801: Scheduled Jobs - Delete: 422 "unable to delete job with existing history" - user cannot remove a scheduled job once it has run at least once
### Description

Deleting a **Scheduled Job** fails with **HTTP 422** and message `unable to delete job with existing history` whenever the job has any execution history — i.e. whenever it has fired at least once. In practice this means: any scheduled job that has *ever* run cannot be removed through the UI or API. The only way to get rid of it is manual DB intervention.

There is no user-facing option to purge the history first, no "force delete" flow, no bulk cleanup — the operation is simply blocked. Over time this causes unusable, obsolete scheduled jobs to accumulate in the list.

### Steps to Reproduce

1. Create a **Scheduled Job** (e.g. a scheduled Discovery) with a schedule that fires soon
2. Wait for the job to fire at least once → it now has execution history
3. Navigate to **Scheduled Jobs** and attempt to delete the job (UI or `DELETE /api/v1/scheduler/jobs/{uuid}`)
4. Observe the response: **HTTP 422 — `unable to delete job with existing history`**
5. Job remains in the list; no way to remove it via the UI

### Expected Behavior

- User can delete a scheduled job regardless of whether it has execution history — one of:
  - **Cascade approach:** deleting the job also deletes (or archives) its history records
  - **Explicit-choice approach:** UI/API offers a "delete history first" step, or a `force=true` parameter that cascades on demand
- If a real reason exists to preserve history after job deletion, the history should live independently and not block cleanup of the job definition itself
- No orphan scheduled jobs are left behind that the user cannot manage

### Actual Behavior

- Response: **HTTP 422 — `unable to delete job with existing history`**
- Job cannot be removed through any user-facing path
- No option to purge history and retry
- Cleanup requires backend/DB intervention

### Severity

Major

### Module

Core

### Attachments

Image

### Logs

_No response_

### Environment

ILM Core: 2.18.1-SNAPSHOT (14 Jul 2026, 07:45)
Commit: 2b4a8e4
Branch: main
Build time: 2026-07-14T05:45:42Z
ILM Frontend: 2.18.0 (9 Jul 2026, 22:58)
Client:
- macOS Tahoe 26.5.1

klaraf755:
Fixes #1801

### DIFF
--- a/src/main/java/com/otilm/core/dao/entity/ScheduledJobHistory.java
+++ b/src/main/java/com/otilm/core/dao/entity/ScheduledJobHistory.java
@@ -32,8 +32,9 @@ public class ScheduledJobHistory extends UniquelyIdentified {
     @Enumerated(EnumType.STRING)
     private SchedulerJobExecutionStatus schedulerExecutionStatus;
 
-    @ManyToOne(fetch = FetchType.LAZY)
+    // Mirrors ON DELETE CASCADE from migration for test environment
     @OnDelete(action = OnDeleteAction.CASCADE)
+    @ManyToOne(fetch = FetchType.LAZY)
     @JoinColumn(name = "scheduled_job_uuid", insertable = false, updatable = false)
     @ToString.Exclude
     private ScheduledJob scheduledJob;
--- a/src/main/java/com/otilm/core/dao/entity/ScheduledJobHistory.java
+++ b/src/main/java/com/otilm/core/dao/entity/ScheduledJobHistory.java
@@ -5,6 +5,8 @@ import com.otilm.api.model.core.scheduler.ScheduledJobHistoryDto;
 import com.otilm.api.model.scheduler.SchedulerJobExecutionStatus;
 import jakarta.persistence.*;
 import lombok.*;
+import org.hibernate.annotations.OnDelete;
+import org.hibernate.annotations.OnDeleteAction;
 import org.hibernate.proxy.HibernateProxy;
 
 import java.util.Date;
@@ -31,6 +33,7 @@ public class ScheduledJobHistory extends UniquelyIdentified {
     private SchedulerJobExecutionStatus schedulerExecutionStatus;
 
     @ManyToOne(fetch = FetchType.LAZY)
+    @OnDelete(action = OnDeleteAction.CASCADE)
     @JoinColumn(name = "scheduled_job_uuid", insertable = false, updatable = false)
     @ToString.Exclude
     private ScheduledJob scheduledJob;
--- a/src/main/java/com/otilm/core/service/impl/SchedulerServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/SchedulerServiceImpl.java
@@ -139,11 +139,6 @@ public class SchedulerServiceImpl implements SchedulerExternalService, Scheduler
                 throw new ValidationException(ValidationError.create("Unable to delete system job."));
             }
 
-            if (scheduledJobHistoryRepository.existsByScheduledJobUuid(UUID.fromString(uuid))) {
-                logger.warn("Unable to delete job with existing history.");
-                throw new ValidationException(ValidationError.create("Unable to delete job with existing history."));
-            }
-
             try {
                 schedulerApiClient.deleteScheduledJob(scheduledJob.getJobName());
                 scheduledJobsRepository.deleteById(UUID.fromString(uuid));
--- a/src/main/java/com/otilm/core/service/impl/SchedulerServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/SchedulerServiceImpl.java
@@ -139,6 +139,13 @@ public class SchedulerServiceImpl implements SchedulerExternalService, Scheduler
                 throw new ValidationException(ValidationError.create("Unable to delete system job."));
             }
 
+            if (scheduledJobHistoryRepository.existsByScheduledJobUuidAndSchedulerExecutionStatusAndJobEndTimeIsNull(
+                    UUID.fromString(uuid), SchedulerJobExecutionStatus.STARTED)) {
+                logger.warn("Unable to delete scheduled job '{}' while it is executing.", scheduledJob.getJobName());
+                throw new ValidationException(ValidationError.create(
+                        "Unable to delete scheduled job while it is executing. Wait for the current run to finish."));
+            }
+
             try {
                 schedulerApiClient.deleteScheduledJob(scheduledJob.getJobName());
                 scheduledJobsRepository.deleteById(UUID.fromString(uuid));
--- a/src/main/resources/db/migration/V202607271000__scheduled_job_history_cascade_delete.sql
+++ b/src/main/resources/db/migration/V202607271000__scheduled_job_history_cascade_delete.sql
@@ -4,4 +4,4 @@ ALTER TABLE scheduled_job_history
     ADD CONSTRAINT fk_scheduled_job_history_scheduled_job
         FOREIGN KEY (scheduled_job_uuid) REFERENCES scheduled_job (uuid) ON DELETE CASCADE;
 
-CREATE INDEX idx_scheduled_job_history_scheduled_job_uuid ON scheduled_job_history (scheduled_job_uuid);
+CREATE INDEX IF NOT EXISTS idx_scheduled_job_history_scheduled_job_uuid ON scheduled_job_history (scheduled_job_uuid);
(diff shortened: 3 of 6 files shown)


src/main/java/com/otilm/core/service/impl/SchedulerServiceImpl.java
```diff
@@ -139,11 +139,6 @@ public void deleteScheduledJob(final String uuid) {
                 throw new ValidationException(ValidationError.create("Unable to delete system job."));
             }
 
-            if (scheduledJobHistoryRepository.existsByScheduledJobUuid(UUID.fromString(uuid))) {
-                logger.warn("Unable to delete job with existing history.");
-                throw new ValidationException(ValidationError.create("Unable to delete job with existing history."));
-            }
-
             try {
                 schedulerApiClient.deleteScheduledJob(scheduledJob.getJobName());
```
lubomirw, line 150: **[Medium]** | Correctness

`registerJobHistory(scheduledJob, STARTED, null)` (line 256) writes a history row *before* the task body runs, so the guard removed in this diff incidentally made an executing job undeletable. With it gone, a job can be deleted mid-run and the new DB cascade removes the live `STARTED` row out from under the running execution:

- Synchronously, the `finally` block calls `finalizeFinishedScheduledJob`, which does `scheduledJobHistoryRepository.save(...)` on a row that no longer exists — the merge re-INSERTs and hits the FK (the parent `scheduled_job` is gone), aborting finalization including the one-time-job deregistration branch.
- Asynchronously, `handleScheduledJobFinishedEvent` (line 281) does `findByUuid(...).orElseThrow(NotFoundException)` for both the job and the history, and now throws.

The window is minutes for long-running tasks such as scheduled Discovery. This path was unreachable before the guard was removed, so it is a genuine regression rather than a pre-existing gap.

_Suggested:_ Either make the completion paths tolerant — have `handleScheduledJobFinishedEvent` and `finalizeFinishedScheduledJob` log-and-return when the job or its history row has disappeared instead of throwing/re-inserting — or keep a narrow guard that rejects deletion only while an execution is actually in flight (a history row with status `STARTED` and a null `job_end_time`). The narrow guard still fixes #1801, since completed history would no longer block anything.

src/test/java/com/otilm/core/integration/service/SchedulerServiceITest.java
```diff
@@ -366,10 +367,17 @@ void testRegisterScheduledJobAndOperations() throws SchedulerException, NotFound
         Assertions.assertEquals(jobName, jobDetailDto.getJobName());
         Assertions.assertEquals(cronExpressionUpdate, jobDetailDto.getCronExpression());
 
+        ScheduledJobHistory jobHistory = new ScheduledJobHistory();
+        jobHistory.setScheduledJobUuid(jobDetailDto.getUuid());
+        jobHistory.setJobExecution(new Date());
+        jobHistory.setSchedulerExecutionStatus(SchedulerJobExecutionStatus.SUCCESS);
+        scheduledJobHistoryRepository.save(jobHistory);
```
lubomirw, line 381: **[Medium]** | Testing

The test saves a history row and then asserts `assertFalse(existsByScheduledJobUuid(...))` after the delete, but never asserts the row existed beforehand. If the save were ineffective or the UUID wiring drifted, the assertion would pass for the wrong reason — and after the removal of `testDeleteScheduledJob_WhenHistoryExists_ThrowsValidationException` this is the only test guarding cascade behaviour.

_Suggested:_ Add `Assertions.assertTrue(scheduledJobHistoryRepository.existsByScheduledJobUuid(jobDetailDto.getUuid()));` immediately after this save and before the `deleteScheduledJob(...)` call on line 376.

src/main/java/com/otilm/core/dao/entity/ScheduledJobHistory.java
```diff
@@ -31,6 +33,7 @@ public class ScheduledJobHistory extends UniquelyIdentified {
     private SchedulerJobExecutionStatus schedulerExecutionStatus;
 
     @ManyToOne(fetch = FetchType.LAZY)
+    @OnDelete(action = OnDeleteAction.CASCADE)
```
lubomirw, line 36: **[Low]** | Maintainability

`OnDeleteAction.CASCADE` appears exactly once in all of `src/main/java`, while the 60+ other cascades in this schema are declared only in migrations. It has no runtime effect in production (`ddl-auto` is unset there, so the cascade comes solely from `V202607271000`), but it is silently what makes the *test* schema cascade, since tests run with `flyway.enabled: false` and `ddl-auto: create-drop`. A future change to either side gives no signal to change the other. Worth being explicit about, not least because it means green CI on the new cascade assertion validates this annotation rather than the migration.

_Suggested:_ Keep the annotation, but add a comment naming the migration, e.g. `// Mirrors ON DELETE CASCADE from V202607271000__scheduled_job_history_cascade_delete.sql; production gets the cascade from that migration, tests get it from here (ddl-auto=create-drop).`

src/main/resources/db/migration/V202607271000__scheduled_job_history_cascade_delete.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE scheduled_job_history
+    DROP CONSTRAINT scheduled_job_history_scheduled_job_uuid_fkey;
+ALTER TABLE scheduled_job_history
+    ADD CONSTRAINT fk_scheduled_job_history_scheduled_job
+        FOREIGN KEY (scheduled_job_uuid) REFERENCES scheduled_job (uuid) ON DELETE CASCADE;
+
+CREATE INDEX idx_scheduled_job_history_scheduled_job_uuid ON scheduled_job_history (scheduled_job_uuid);
```
lubomirw: **[Nit]** | Maintainability

Sibling migrations in this tree use the guarded form (`idx_connector_interface_connector_uuid`, `idx_certificate_issuer_certificate_uuid`), so as written this fails on any environment where an operator hand-created the index as a hotfix. The index itself is correct and needed — confirmed it is not a duplicate (`V202306121100` is named "scheduler_indexes" but only adds primary keys) and PostgreSQL does not auto-index FK columns, so without it every job delete sequential-scans the history table.

_Suggested:_ `CREATE INDEX IF NOT EXISTS idx_scheduled_job_history_scheduled_job_uuid ON scheduled_job_history (scheduled_job_uuid);`

lubomirw (changes requested):
Automated multi-reviewer triage — 4 findings (REQUEST_CHANGES).

Reviewed against `main` from four independent angles (Copilot CLI, guided PR review, superpowers code review, hygiene scan). The hygiene scan came back clean. The one blocking item is the mid-run delete regression in `SchedulerServiceImpl`; the rest is migration hardening and small cleanups.

**Not anchorable to a diff line:**

- `src/main/java/com/otilm/core/dao/repository/ScheduledJobHistoryRepository.java:15` — **[Low]** | Maintainability — `existsByScheduledJobUuid` no longer has any caller in `src/main` now that the guard is gone; a repo-wide grep finds only the declaration and three test assertions (`SchedulerServiceITest:380`, `SchedulerListenerITest:60` and `:72`). It was the sole expression of the "history blocks deletion" business rule, so leaving it on the repository interface invites someone to reintroduce that rule. _Suggested:_ either delete it and have the tests assert via `findTopByScheduledJobUuidOrderByJobExecutionDesc(...) == null`, or keep it with a short comment marking it a test-only assertion helper that no longer encodes a business rule.

src/test/java/com/otilm/core/integration/service/SchedulerServiceITest.java
klaraf755: Added the assertion.

src/main/java/com/otilm/core/service/impl/SchedulerServiceImpl.java
klaraf755: Added guard checking if there is a job in progress before removing the job.

src/main/java/com/otilm/core/dao/entity/ScheduledJobHistory.java
klaraf755: Added a comment to the annotation.

src/main/resources/db/migration/V202607271000__scheduled_job_history_cascade_delete.sql
klaraf755: Added `IF NOT EXISTS`.

lubomirw (approved):

PR #1893, opened 2026-07-27
Reject token profile deletion with dependent keys or signing profiles

### LINKED ISSUE #1882: Deleting token profile with dependent keys returns 500 instead of 422 listing the blocking keys
### Description

Attempting to delete a token profile that still has dependent keys fails with an HTTP 500 internal server error. The API should instead reject the operation with a 422 response that lists the keys blocking the deletion, so the operator knows what to remove or reassign first.

### Steps to Reproduce

1. Create a token profile
2. Create one or more keys under that token profile
3. Attempt to delete the token profile
4. Observe the response

### Expected Behavior

The request is rejected with HTTP 422 and the response lists the dependent keys blocking the deletion.

### Actual Behavior

The request fails with HTTP 500 internal server error and no indication of which keys block the operation.

### Severity

Major

### Module

Keys

klaraf755:
Fixes #1882

## Problem

Deleting a token profile that still has dependent objects failed with an HTTP 500 —
the delete hit the FK constraint on `cryptographic_key.token_profile_uuid` and the
raw `DataIntegrityViolationException` propagated. The operator got no indication of
what blocks the deletion.

While fixing this, a second blocker was found: `signing_profile_version` also
references `token_profile` with `ON DELETE RESTRICT`, so signing profiles bound to
the token profile caused the same 500.

## Change

`deleteProfileInternal` now validates dependents before deleting (all three delete
overloads — single, token-instance-scoped, and bulk — funnel through it) and rejects
with HTTP 422: `Cannot delete Token Profile <name>: 12 dependent Keys, 2 dependent
Signing Profiles`

The message reports **counts only, not names**. A token profile can have thousands
of dependent keys, so listing them in an error message doesn't scale; operators can
see the full set by filtering the Key inventory (or Signing Profiles) by token
profile in the UI. Signing profiles are counted by distinct profile
(`COUNT(DISTINCT signing_profile_uuid)`), not by version, so the count matches what
the operator sees in the profile list.

New repository methods: `CryptographicKeyRepository.countByTokenProfileUuid`,
`SigningProfileVersionRepository.countDistinctSigningProfilesByTokenProfileUuid`.

## Notes

- **Bulk delete stays all-or-nothing**: one blocked profile rolls back the whole
  batch (class-level `@Transactional`). This is pre-existing behavior — before this
  fix the FK violation caused the same zero-progress rollback, just as a 500.
  Per-item reporting (the `BulkActionMessageDto` pattern used by token instance bulk
  delete) would change the endpoint contract and is left out of scope.
- The check-then-delete pattern has a narrow race window (a key created between the
  check and commit still trips the FK), matching the existing SCEP/ACME profile
  delete convention.

## Testing

Two new integration tests in `TokenProfileServiceITest` (existing Spring context
signature, no new context): deletion blocked by dependent keys, and by a signing
profile version referencing the token profile.

### DIFF
--- a/src/main/java/com/otilm/core/dao/repository/CryptographicKeyRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/CryptographicKeyRepository.java
@@ -20,4 +20,6 @@ public interface CryptographicKeyRepository extends SecurityFilterRepository<Cry
 
     @EntityGraph(attributePaths = {"tokenProfile", "items"})
     List<CryptographicKey> findByUuidIn(List<UUID> uuids);
+
+    long countByTokenProfileUuid(UUID tokenProfileUuid);
 }
--- a/src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
@@ -227,6 +227,8 @@ public class TokenProfileServiceImpl implements TokenProfileExternalService, Tok
                 deleteProfileInternal(uuid, false);
             } catch (NotFoundException e) {
                 logger.warn("Unable to find Token Profile with uuid {}. It may have already been deleted", uuid);
+            } catch (ValidationException e) {
+                logger.warn(e.getMessage());
             }
         }
     }
--- a/src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
@@ -201,7 +201,6 @@ public class TokenProfileServiceImpl implements TokenProfileExternalService, Tok
     public void deleteTokenProfile(SecuredUUID tokenProfileUuid) throws NotFoundException {
         logger.info("Deleting token profile with uuid: {}", tokenProfileUuid);
         deleteProfileInternal(tokenProfileUuid, true);
-
     }
 
     @Override
@@ -353,8 +352,8 @@ public class TokenProfileServiceImpl implements TokenProfileExternalService, Tok
 
     private void deleteProfileInternal(SecuredUUID uuid, boolean throwWhenAssociated) throws NotFoundException {
         TokenProfile tokenProfile = getTokenProfileEntity(uuid);
-        if (throwWhenAssociated && tokenProfile.getTokenInstanceReference() == null) {
-            throw new ValidationException(ValidationError.create("Token Profile has associated Token Instance. Use other API"));
+        if (throwWhenAssociated && tokenProfile.getTokenInstanceReference() != null) {
+            throw new ValidationException(ValidationError.create("Token Profile has associated Token Instance. Use the token instance scoped API to delete the Token Profile."));
         }
         validateNoDependentObjects(tokenProfile);
         attributeEngine.deleteObjectAttributeContent(Resource.TOKEN_PROFILE, tokenProfile.getUuid());
--- a/src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
@@ -365,17 +365,28 @@ public class TokenProfileServiceImpl implements TokenProfileExternalService, Tok
 
         long keyCount = cryptographicKeyRepository.countByTokenProfileUuid(tokenProfile.getUuid());
         if (keyCount > 0) {
-            blockers.add(keyCount + " dependent Keys");
+            blockers.add(keyCount + " dependent Key(s)");
         }
 
-        long signingProfileCount = signingProfileVersionRepository.countDistinctSigningProfilesByTokenProfileUuid(tokenProfile.getUuid());
-        if (signingProfileCount > 0) {
-            blockers.add(signingProfileCount + " dependent Signing Profiles");
+        List<String> latestVersionNames = signingProfileVersionRepository.findSigningProfileNamesUsingTokenProfileInLatestVersion(tokenProfile.getUuid());
+        if (!latestVersionNames.isEmpty()) {
+            blockers.add("dependent Signing Profile(s): " + String.join(", ", latestVersionNames));
+        }
+
+        // Superseded versions are retained for audit and cannot be edited, so these references
+        // can only be released by deleting the Signing Profile itself.
+        List<String> supersededOnlyNames = new ArrayList<>(signingProfileVersionRepository.findDistinctSigningProfileNamesByTokenProfileUuid(tokenProfile.getUuid()));
+        supersededOnlyNames.removeAll(latestVersionNames);
+        if (!supersededOnlyNames.isEmpty()) {
+            blockers.add("Signing Profile(s) referencing it only in superseded versions (released only by deleting the Signing Profile): "
+                    + String.join(", ", supersededOnlyNames));
         }
 
         if (!blockers.isEmpty()) {
+            // Single placeholder: sequential {} substitution would garble the message when the
+            // profile name itself contains a literal "{}".
             throw new ValidationException(ValidationError.create(
-                    "Cannot delete Token Profile {}: {}", tokenProfile.getName(), String.join(", ", blockers)));
+                    "Cannot delete Token Profile {}", tokenProfile.getName() + ": " + String.join("; ", blockers)));
         }
     }
 
--- a/src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
@@ -36,6 +36,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
@@ -357,7 +358,17 @@ public class TokenProfileServiceImpl implements TokenProfileExternalService, Tok
         }
         validateNoDependentObjects(tokenProfile);
         attributeEngine.deleteObjectAttributeContent(Resource.TOKEN_PROFILE, tokenProfile.getUuid());
-        tokenProfileRepository.delete(tokenProfile);
+        try {
+            tokenProfileRepository.delete(tokenProfile);
+            // Force the DELETE to execute here; without the flush it runs at commit,
+            // outside this try, and a concurrent FK violation would surface as HTTP 500.
+            tokenProfileRepository.flush();
+        } catch (DataIntegrityViolationException e) {
+            // The aborted transaction cannot be re-queried for the dependency counts.
+            throw new ValidationException(ValidationError.create(
+                    "Cannot delete Token Profile {}: dependent Key(s) or Signing Profile(s) were created concurrently. Retry to see current dependencies.",
+                    tokenProfile.getName()));
+        }
     }
 
     private void validateNoDependentObjects(TokenProfile tokenProfile) {
--- a/src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
@@ -19,8 +19,10 @@ import com.otilm.core.attribute.engine.records.ObjectAttributeContentInfo;
 import com.otilm.core.dao.entity.TokenInstanceReference;
 import com.otilm.core.dao.entity.TokenProfile;
 import com.otilm.core.dao.entity.TokenProfile_;
+import com.otilm.core.dao.repository.CryptographicKeyRepository;
 import com.otilm.core.dao.repository.TokenInstanceReferenceRepository;
 import com.otilm.core.dao.repository.TokenProfileRepository;
+import com.otilm.core.dao.repository.signing.SigningProfileVersionRepository;
 import com.otilm.core.model.auth.ResourceAction;
 import com.otilm.core.security.authz.AuthorizationEnforcer;
 import com.otilm.core.security.authz.ExternalAuthorization;
@@ -37,6 +39,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
@@ -61,6 +64,18 @@ public class TokenProfileServiceImpl implements TokenProfileExternalService, Tok
     // --------------------------------------------------------------------------------
     private TokenProfileRepository tokenProfileRepository;
     private TokenInstanceReferenceRepository tokenInstanceReferenceRepository;
+    private CryptographicKeyRepository cryptographicKeyRepository;
+    private SigningProfileVersionRepository signingProfileVersionRepository;
+
+    @Autowired
+    public void setSigningProfileVersionRepository(SigningProfileVersionRepository signingProfileVersionRepository) {
+        this.signingProfileVersionRepository = signingProfileVersionRepository;
+    }
+
+    @Autowired
+    public void setCryptographicKeyRepository(CryptographicKeyRepository cryptographicKeyRepository) {
+        this.cryptographicKeyRepository = cryptographicKeyRepository;
+    }
 
     @Autowired
     public void setAttributeEngine(AttributeEngine attributeEngine) {
@@ -341,8 +356,28 @@ public class TokenProfileServiceImpl implements TokenProfileExternalService, Tok
         if (throwWhenAssociated && tokenProfile.getTokenInstanceReference() == null) {
             throw new ValidationException(ValidationError.create("Token Profile has associated Token Instance. Use other API"));
         }
+        validateNoDependentObjects(tokenProfile);
         attributeEngine.deleteObjectAttributeContent(Resource.TOKEN_PROFILE, tokenProfile.getUuid());
         tokenProfileRepository.delete(tokenProfile);
+    }
+
+    private void validateNoDependentObjects(TokenProfile tokenProfile) {
+        List<String> blockers = new ArrayList<>();
+
+        long keyCount = cryptographicKeyRepository.countByTokenProfileUuid(tokenProfile.getUuid());
+        if (keyCount > 0) {
+            blockers.add(keyCount + " dependent Keys");
+        }
+
+        long signingProfileCount = signingProfileVersionRepository.countDistinctSigningProfilesByTokenProfileUuid(tokenProfile.getUuid());
+        if (signingProfileCount > 0) {
+            blockers.add(signingProfileCount + " dependent Signing Profiles");
+        }
+
+        if (!blockers.isEmpty()) {
+            throw new ValidationException(ValidationError.create(
+                    "Cannot delete Token Profile {}: {}", tokenProfile.getName(), String.join(", ", blockers)));
+        }
     }
 
     private void disableProfileInternal(SecuredUUID uuid) throws NotFoundException {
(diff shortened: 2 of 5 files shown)


src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
```diff
@@ -341,10 +356,30 @@ private void deleteProfileInternal(SecuredUUID uuid, boolean throwWhenAssociated
         if (throwWhenAssociated && tokenProfile.getTokenInstanceReference() == null) {
             throw new ValidationException(ValidationError.create("Token Profile has associated Token Instance. Use other API"));
         }
```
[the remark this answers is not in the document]
klaraf755: Fixed. However, in practice, this is likely dead code, since the migration has `NOT NULL` constraint on token instance FK. The method (and API) without token instance can be removed in some follow-up.

src/test/java/com/otilm/core/integration/service/TokenProfileServiceITest.java
```diff
@@ -216,6 +231,47 @@ void testRemoveTokenProfile() throws NotFoundException {
...
 
+    @Test
+    void testRemoveTokenProfile_withDependentKeys() {
+        createKey("testKey1");
+        createKey("testKey2");
+
+        ValidationException e = Assertions.assertThrows(
+                ValidationException.class,
+                () -> tokenProfileService.deleteTokenProfile(tokenProfile.getSecuredUuid())
+        );
```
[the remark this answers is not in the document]
klaraf755: Fixed.

src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
```diff
@@ -341,10 +356,30 @@ private void deleteProfileInternal(SecuredUUID uuid, boolean throwWhenAssociated
...
 
+    private void validateNoDependentObjects(TokenProfile tokenProfile) {
+        List<String> blockers = new ArrayList<>();
+
+        long keyCount = cryptographicKeyRepository.countByTokenProfileUuid(tokenProfile.getUuid());
+        if (keyCount > 0) {
+            blockers.add(keyCount + " dependent Keys");
+        }
+
+        long signingProfileCount = signingProfileVersionRepository.countDistinctSigningProfilesByTokenProfileUuid(tokenProfile.getUuid());
```
lubomirw: **[Medium]** | Correctness

`countDistinctSigningProfilesByTokenProfileUuid` counts every `signing_profile_version` row, including superseded ones. `SigningProfileServiceImpl` deliberately retains old versions when signing records exist (see the "Lenient version bump" comment around line 433), and `deleteSigningProfile` refuses a profile that has signing records (`Cannot delete Signing Profile '%s': it has signing records`). So a superseded version referencing this token profile blocks the deletion permanently — the operator cannot release it by editing the signing profile, and cannot delete the signing profile either. The 422 then names a signing profile whose current version no longer uses this token profile, with no path forward.

Related: the PR description says the count "matches what the operator sees in the profile list" — that does not hold for historical versions.

_Suggested:_ split the count into profiles whose latest version references the token profile vs. profiles where only a superseded version does, and word the message so the operator knows the second group can only be released by deleting the signing profile. Correct the claim in the PR description.

<sub>flagged by all three reviewers (guided-pr-review, superpowers, Copilot CLI); dead-end scenario verified against `SigningProfileServiceImpl`</sub>

src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
```diff
@@ -341,10 +356,30 @@ private void deleteProfileInternal(SecuredUUID uuid, boolean throwWhenAssociated
...
+        if (signingProfileCount > 0) {
+            blockers.add(signingProfileCount + " dependent Signing Profiles");
+        }
+
+        if (!blockers.isEmpty()) {
+            throw new ValidationException(ValidationError.create(
+                    "Cannot delete Token Profile {}: {}", tokenProfile.getName(), String.join(", ", blockers)));
```
lubomirw: **[Medium]** | Error handling

Issue #1882's Expected Behavior is "rejected with HTTP 422 and the response **lists the dependent keys** blocking the deletion". This message reports counts only, so QA verifying against the issue's acceptance criteria will find a mismatch. ~20 existing dependent-delete guards in core name the blockers (`TokenInstanceServiceImpl` — `Dependent Token Profiles: {}`, `AuthorityInstanceServiceImpl` — `Dependent RA profiles: {}`); `EntityInstanceServiceImpl` uses counts, so both conventions exist.

The scaling rationale in the PR description is reasonable for keys — but signing profiles are few, and naming them would be genuinely useful.

_Suggested:_ either list the first N names followed by "and X more" (at minimum for signing profiles), or update #1882's acceptance criteria to record the counts-only decision before QA picks it up.

src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
```diff
@@ -341,10 +356,30 @@ private void deleteProfileInternal(SecuredUUID uuid, boolean throwWhenAssociated
...
+            blockers.add(keyCount + " dependent Keys");
+        }
+
+        long signingProfileCount = signingProfileVersionRepository.countDistinctSigningProfilesByTokenProfileUuid(tokenProfile.getUuid());
+        if (signingProfileCount > 0) {
+            blockers.add(signingProfileCount + " dependent Signing Profiles");
+        }
+
+        if (!blockers.isEmpty()) {
+            throw new ValidationException(ValidationError.create(
```
lubomirw, line 401: **[Nit]** | Correctness

`ValidationError.formatDescription` substitutes placeholders with a sequential `replaceFirst("\\{\\}")` loop over the **already-substituted** string. Token profile names are only validated non-blank, so a name containing a literal `{}` swallows the second placeholder: the blocker list is spliced into the name and the real blocker text is dropped.

(Interpolation is otherwise correct here — `{}` works and `$`/`\` in names are safe via `Matcher.quoteReplacement`.)

_Suggested:_ use a single placeholder so the user-controlled value is substituted last:
```java
ValidationError.create("Cannot delete Token Profile {}",
        tokenProfile.getName() + ": " + String.join(", ", blockers))
```

<sub>flagged by superpowers + Copilot CLI</sub>

src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
```diff
@@ -341,10 +356,30 @@ private void deleteProfileInternal(SecuredUUID uuid, boolean throwWhenAssociated
...
         }
+        validateNoDependentObjects(tokenProfile);
         attributeEngine.deleteObjectAttributeContent(Resource.TOKEN_PROFILE, tokenProfile.getUuid());
         tokenProfileRepository.delete(tokenProfile);
     }
 
+    private void validateNoDependentObjects(TokenProfile tokenProfile) {
+        List<String> blockers = new ArrayList<>();
+
+        long keyCount = cryptographicKeyRepository.countByTokenProfileUuid(tokenProfile.getUuid());
```
lubomirw, line 379: **[Low]** | Error handling

The counts run here but `tokenProfileRepository.delete(...)` only flushes at commit, so a key or signing-profile version inserted by a concurrent transaction still trips the FK (`cryptographic_key_to_token_profile_key`, and `signing_profile_version.token_profile_uuid ON DELETE RESTRICT`) and surfaces as a `DataIntegrityViolationException` → HTTP 500 — the exact failure this PR removes, just in a narrower window. The PR description acknowledges this and points to the SCEP/ACME convention, so this is optional hardening rather than a blocker.

_Suggested:_ catch `DataIntegrityViolationException` around the delete and re-throw the same `ValidationException`, so the race also returns 422.

src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
```diff
@@ -341,10 +356,30 @@ private void deleteProfileInternal(SecuredUUID uuid, boolean throwWhenAssociated
...
+        if (keyCount > 0) {
+            blockers.add(keyCount + " dependent Keys");
```
lubomirw: **[Low]** | Maintainability

The count is always followed by a plural, so a single blocker renders as `1 dependent Keys` / `1 dependent Signing Profiles`. The test at line 272 asserts `"1 dependent Signing Profiles"` verbatim, which cements the grammar error — fixing the wording later breaks the test.

_Suggested:_ pluralise conditionally, or switch to a count-agnostic shape (`"dependent Keys: " + keyCount`), and relax the assertions to match the stable part (profile name + count) rather than the full English phrase.

<sub>flagged by all three reviewers</sub>

src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
```diff
@@ -341,10 +356,30 @@ private void deleteProfileInternal(SecuredUUID uuid, boolean throwWhenAssociated
         if (throwWhenAssociated && tokenProfile.getTokenInstanceReference() == null) {
```
lubomirw: **[Low]** | Correctness

Not introduced by this PR — flagging it because the new guard is inserted directly below it. The condition reads `throwWhenAssociated && tokenProfile.getTokenInstanceReference() == null` while the message says "Token Profile **has** associated Token Instance" — inverted. And since `token_profile.token_instance_ref_uuid` is `NOT NULL` (`V202228121130__cryptoraphic_provider.sql:24`), it can never fire, so `DELETE /v1/tokenProfiles/{uuid}` never redirects the caller to the token-scoped API as intended. The existing `testRemoveTokenProfile` encodes the current behaviour.

_Suggested:_ separate change — either flip to `!= null` and update the test, or drop the guard and the `throwWhenAssociated` parameter if the two delete endpoints are deliberately equivalent. Worth its own issue rather than folding into this fix.

<sub>flagged by guided-pr-review + superpowers; NOT NULL constraint verified</sub>

lubomirw (commented):
Automated multi-reviewer triage — 12 findings (COMMENT). Sources: Copilot CLI, `guided-pr-review`, superpowers code review, `pr-hygiene` (clean, 0 findings). Findings confirmed by multiple sources are marked as such. 9 are inline below; the 3 that have no line inside this diff are listed here.

**[Medium] | Error handling — bulk delete stays all-or-nothing, and the 422 names only one profile** (`TokenProfileServiceImpl.java:223-231`)
`deleteTokenProfile(List)` catches only `NotFoundException`, so the new `ValidationException` escapes the class-level `@Transactional` and rolls back profiles already deleted in the batch, while the message names just the first offender. Nine sibling services (`SigningProfileServiceImpl`, `ScepProfileServiceImpl`, `AcmeProfileServiceImpl`, `AuthorityInstanceServiceImpl`, `CmpProfileServiceImpl`, `TspProfileServiceImpl`, `TimeQualityConfigurationServiceImpl`, v2 `ConnectorServiceImpl`, v2 `ComplianceProfileServiceImpl`) instead catch `Exception` and accumulate `BulkActionMessageDto.failure(...)`.
Agreed the per-item contract change is out of scope as the PR description states — and the zero-progress rollback is genuinely pre-existing, since the FK violation propagated identically. One correction to a reviewer claim: fe-administrator does *not* render per-item results here — `token-profiles-epics.ts` maps failure to a single `error` string, and `bulkDeleteErrorMessages: BulkActionModel[]` in `ducks/token-profiles.ts` is vestigial state.
_Suggested minimum:_ aggregate blockers across the loop and throw one `ValidationException` listing every blocked profile, or add "no Token Profiles were deleted" to the message so the rollback is visible to the operator. Follow-up issue for the `List<BulkActionMessageDto>` contract.

**[Low] | Documentation — HTTP 422 not declared on the two single-delete operations** (`interfaces`: `api/interfaces/core/web/TokenProfileController.java:122,141`)
Both single-delete operations declare only 204 and 404, while the bulk endpoint (line 193) and create/edit (lines 82, 102) already declare 422 — so the new `ValidationException` → 422 is undocumented in the OpenAPI contract that drives the generated clients and docs site.
Sources disagree on whether this matters: `AuthorityInstanceController.deleteAuthorityInstance` and `EntityInstanceController.deleteEntityInstance` also omit 422 for the same dependent-object rejection, so the omission is consistent across controllers even though it is inconsistent *within* this one. Your call.
_Suggested (if pursued):_ companion `interfaces` PR copying the existing 422 `@ApiResponse` onto both single-delete operations.

**[Low] | Documentation — no "Remove Token Profile" process in the platform docs** (`documentation/docs/certificate-key/connectors/provider-interfaces/cryptography-provider.md`)
That page documents `Core -> Core : Check dependencies` for Remove `Token` instance, and `entity-provider.md` does the same for Remove `Entity`/`Location`, but there is no Token Profile process section at all — so the newly enforced "keys and signing profiles block deletion" rule is nowhere in the docs. Nothing became factually stale; this is a missing section rather than a correction.
_Suggested:_ add a `### Remove Token Profile` process under `## Processes` with the same `Check dependencies` step, naming Keys and Signing Profiles as blockers.

---
One further finding was triaged out as out-of-scope noise: `compliance_profile_association` rows with `resource = 'TOKEN_PROFILE'` are never cleaned up on delete. It has no FK so it never caused the 500 this PR fixes, and `RA_PROFILE` / `VAULT_PROFILE` share the identical gap — systemic, separate issue.

src/main/java/com/otilm/core/dao/repository/CryptographicKeyRepository.java
```diff
@@ -28,4 +28,6 @@ public interface CryptographicKeyRepository extends SecurityFilterRepository<Cry
 
     @EntityGraph(attributePaths = {"tokenProfile", "items"})
     List<CryptographicKey> findByUuidIn(List<UUID> uuids);
+
+    long countByTokenProfileUuid(UUID tokenProfileUuid);
```
lubomirw, line 32: **[Low]** | Performance

`cryptographic_key.token_profile_uuid` carries an FK (`V202228121130__cryptoraphic_provider.sql:102-104`) but no index, and no later migration adds one — unlike the signing side, which ships `idx_spv_token_profile_uuid` in `V202605051246__signing.sql:71`. Postgres does not index FK columns automatically, so this count sequentially scans the whole key table.

_Suggested:_ add a migration with `CREATE INDEX cryptographic_key_token_profile_uuid ON cryptographic_key (token_profile_uuid);`. It also speeds up the FK check Postgres runs on every `token_profile` delete, not just this new count.

<sub>flagged by Copilot CLI + guided-pr-review; absence of the index verified across all migrations</sub>

src/test/java/com/otilm/core/integration/service/TokenProfileServiceITest.java
```diff
@@ -216,6 +231,47 @@ void testRemoveTokenProfile() throws NotFoundException {
         );
     }
 
+    @Test
+    void testRemoveTokenProfile_withDependentKeys() {
```
lubomirw, line 236: **[Medium]** | Testing

Both new tests exercise only `deleteTokenProfile(SecuredUUID)` with a single blocker type. Two gaps:

1. The **bulk overload** — where the behaviour change is riskiest (see the note on the review body about the batch rollback) — is covered only by the pre-existing `testBulkRemove`, which deletes one dependency-free profile.
2. The comma-joined **two-blocker branch** of `String.join(", ", blockers)` never renders, so a regression there would pass CI.

_Suggested:_ add a test that bulk-deletes two profiles where only the second has dependents and asserts what happened to the first (this documents the all-or-nothing semantics and fails if they change), plus a test with both a dependent key and a dependent signing-profile version on one profile asserting the joined message.

src/test/java/com/otilm/core/integration/service/TokenProfileServiceITest.java
```diff
@@ -216,6 +231,47 @@ void testRemoveTokenProfile() throws NotFoundException {
...
+        version.setTokenProfile(tokenProfile);
+        signingProfileVersionRepository.save(version);
+
+        ValidationException e = Assertions.assertThrows(
+                ValidationException.class,
+                () -> tokenProfileService.deleteTokenProfile(tokenProfile.getSecuredUuid())
+        );
+
+        String errors = joinErrorDescriptions(e);
+        Assertions.assertTrue(errors.contains("Cannot delete Token Profile " + TOKEN_PROFILE_NAME + ": 1 dependent Signing Profiles"), errors);
```
lubomirw: **[Low]** | Testing

`testRemoveTokenProfile_withDependentKeys` ends with `Assertions.assertTrue(tokenProfileRepository.findByUuid(...).isPresent())`, but this test omits the equivalent check — so it verifies only that an exception was thrown and would still pass if the guard threw *after* deleting the row.

_Suggested:_ mirror the `findByUuid(...).isPresent()` assertion here, and optionally assert the `signing_profile_version` row survives too, so the test proves nothing was partially removed.

src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
klaraf755: Split the messages about signing profiles correctly - also replaced the counts with names for this purpose.

src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
klaraf755: Used names for Signing Profiles - also because of the split based on superseded versions mentioned in previous review comment.

src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
klaraf755: Changed to `key(s)`/`profile(s)` to allow both singular and plural.

src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
klaraf755: Addressed in https://github.com/OmniTrustILM/core/pull/1893#discussion_r3656214301.

src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
klaraf755: Used the placeholder.

src/test/java/com/otilm/core/integration/service/TokenProfileServiceITest.java
klaraf755: Added assertion for count of keys by Token Profile UUID.

src/main/java/com/otilm/core/service/impl/TokenProfileServiceImpl.java
klaraf755: Added the catch for `DataIntegrityViolationException`.

src/main/java/com/otilm/core/dao/repository/CryptographicKeyRepository.java
klaraf755: Added the migration, also for future speed up.

src/test/java/com/otilm/core/integration/service/TokenProfileServiceITest.java
klaraf755: Added tests.

klaraf755:
Reply to https://github.com/OmniTrustILM/core/pull/1893#pullrequestreview-4787194607:
**Bulk delete (all-or-nothing / single-offender message):** Addressed differently from the suggested minimum — the bulk loop now catches `ValidationException` the same way it already catches `NotFoundException`: the blocked profile is skipped with a warning log and the rest of the batch proceeds. This removes the batch rollback entirely (deletable profiles are actually deleted) and keeps the endpoint's existing `void`/204 contract, which — as you correctly note about fe-administrator mapping failures to a single error string — has no channel for per-item results anyway. Blocked profiles simply remain visible in the list after refresh, and the single-delete endpoint returns the detailed 422 naming their blockers. The proper per-item `List<BulkActionMessageDto>` contract (aligning with the nine sibling services) is a follow-up issue, since it changes the API surface. Covered by a new test (`testBulkRemove_skipsProfilesWithDependents`).

**OpenAPI 422 on single-delete operations:** Won't do in this PR. As noted, the omission is consistent with `AuthorityInstanceController` and `EntityInstanceController` for the same dependent-object rejection; declaring it would be an `interfaces` repo change and is better done once for all affected delete operations rather than piecemeal here.

**Docs "Remove Token Profile" process section:** Won't do in this PR — separate repo, and it's a missing section rather than something this PR made stale. Can be picked up together with the docs alignment for the follow-up above.

lubomirw (approved):

PR #1895, opened 2026-07-27
Register system certificate extensions and resolve RDN code conflicts

### LINKED ISSUE #1894: Wire system certificate-extension OIDs through the OID registry, with upgrade guards
### Description

Consume the `SystemOid` change from interfaces#791 so that a system certificate-extension OID behaves exactly like a registered custom one, and add the guards that promoting OIDs into the enum requires.

The registry plumbing is not optional. `CertificateRequestAttributeProjector.toRequestedExtension` reads `defaultCritical` and `valueEncoding` from the `CERTIFICATE_EXTENSION` cache; `CustomOidEntryServiceImpl.getOidToRecordMap` currently builds system-derived `OidRecord`s without those fields, so without this work every system extension would project as non-critical with no declared encoding.

Promoting OIDs into `SystemOid` also creates two upgrade hazards, both of which must be handled in the same release.

**Hazard 1 — an OID a deployment already registered as a Custom OID.** `getOidToRecordMap` merges the system entries *after* the DB rows, so the operator's `defaultCritical` / `valueEncoding` are silently replaced by the enum's values. Because `createCustomOidEntry` rejects any OID reserved by `SystemOid`, and the edit and delete paths skip the cache for system OIDs, the pre-existing row then becomes an orphan the operator can neither re-create nor meaningfully edit.

**Hazard 2 — a new system RDN code colliding with a custom code already registered on a _different_ OID.** `OidHandler.getCodeToOidMap` flattens every `code` and every `altCode` of every RDN entry into a single map with no collision check, then republishes it through a case-insensitive lookup. Two entries claiming one token means last-writer-wins, decided by map iteration order — and the winner can flip whenever any later custom-OID create, edit or delete rebuilds the cache.

Every consequence is silent, with no validation error anywhere:

- DN parsing resolves the token to the wrong OID (`PlatformX500NameStyle.attrNameToOID` is registry-first), so the issued certificate carries the wrong RDN type.
- Whitelist canonicalization in `CertificateRequestContentValidator` stops matching the mapping the operator authored, so STRICT rejects an RDN they believe they allowed.
- Request-attribute projection writes the value into the wrong RDN type.
- `AttributeEngine.validateRdnMappedField` resolves the code successfully, so the definition saves cleanly.
- EST `csrattrs` advertises the wrong OID to clients.

The existing duplicate-code guard cannot catch this. `getAllCodesInLowerCase` runs only when a *custom* entry is created or edited, and in this scenario the custom row already exists. A unit test over `SystemOid.values()` cannot catch it either, because the collision depends on deployment data. All five newly shipped codes (`SERIALNUMBER`, `STREET`, `PostalCode`, `BusinessCategory`, `UID`) are plausible operator choices.

**Resolution — deterministic precedence with the custom entry winning, plus a loud warning. No data is mutated and startup is never failed.**

The actual defect is not that collisions exist; it is that `getCodeToOidMap` resolves them by `HashMap` iteration order, so the winner is arbitrary and can flip whenever a later custom-OID create, edit or delete rebuilds the cache. Make the precedence explicit instead: **for a contested code token, the DB (custom) row wins over the system entry**, logged at WARN with the token and both OIDs. Note this is the opposite ordering from `getOidToRecordMap`'s `putAll(system)` — no conflict, since that map is keyed by OID and this one by code.

Rationale for custom-wins over the two alternatives that were considered:

- **It preserves pre-upgrade behaviour exactly.** Before the upgrade only one entry claimed the token — the custom one. Letting it keep winning means the upgrade changes nothing for that deployment and introduces no new misissuance risk. Failing startup or stripping the custom code both change behaviour; this does not.
- **Failing startup is disproportionate.** Core down means no issuance, no revocation, no protocol endpoints. Under SaaS, tenants cannot touch yml or env and so cannot self-remediate — the first they would learn of a naming collision is an outage.
- **Stripping the custom code causes the very corruption this guards against.** An operator who registered `UID` on their own OID would see `UID=x` in inbound DN strings silently start resolving to `0.9.2342.19200300.100.1.1`, and their rendered DNs would change too, dragging in an `updateCertificateDNs` pass.

The cost of custom-wins is small and bounded:

- **The mapping UI is unaffected.** `toOidSelectOptions` sets `value: e.oid`, so mappings authored through the dropdown carry dotted OIDs. A contested short code does not make the new system RDN unselectable or unmappable.
- **DN rendering is unaffected.** `oidToCodeMap` is OID → `code()`, unique per OID, so no collision is possible in that direction.
- The only surface that degrades is DN-string *parsing* of that one token — where custom-wins is exactly the behaviour the deployment already had.

**The collision set is closed and can only shrink.** `createCustomOidEntry` and the edit path already reject a custom code colliding with a system code, because `getAllCodesInLowerCase` reads the merged cache including system entries. So no new collisions can arise after the upgrade; the set is finite, fixed at upgrade time, and shrinks as operators rename. Remediation is already supported — renaming a custom code goes through the edit path, which calls `updateCertificateDNs`.

Blocked by the interfaces SNAPSHOT (interfaces#791).

### Definition of Done

- `CustomOidEntryMapper.toDetailDto(SystemOid)` emits `CertificateExtensionOidPropertiesDto{defaultCritical, valueEncoding}` for `CERTIFICATE_EXTENSION` entries. The comment stating that the enum has no fields to represent a certificate extension's typed properties is removed, since it is no longer true.
- `CustomOidEntryServiceImpl.getOidToRecordMap` carries `defaultCritical` and `valueEncoding` into the system-derived `OidRecord`s.
- `AttributeEngine` rejects an `ExtensionMappedField` on `2.5.29.17` with a message pointing the operator at the SAN mapping target, because the parser diverts SAN out of the extension list and such a mapping can never match. **Scope this guard to `2.5.29.17` only.** The criterion is that the OID is structurally unreachable through the extension path — the parser diverts it *and* a dedicated target exists. `2.5.29.17` is the only OID special-cased in the request pipeline; `2.5.29.37`, `2.5.29.15` and `2.5.29.19` survive parsing as ordinary requested extensions and `ExtensionMappedField` is the only way to map them, so they must stay mappable.
- Flyway migration deletes `custom_oid_entry` rows whose OID is now reserved by `SystemOid`. For RDN rows whose custom `code` differs from the enum's code, the migration also calls `updateCertificateDNs(oid, enumCode, customCode)` so display DNs stay consistent. Expect a partial match count rather than a full one — the original custom-create rewrite passed the dotted OID as `oldCode` and matched nothing wherever BouncyCastle already had a symbol for that OID, so the stored display state is mixed.
- `OidHandler.getCodeToOidMap` resolves a contested code token **deterministically, with the custom entry winning over the system entry**, rather than by `HashMap` iteration order. No data is mutated and startup is not failed.
- Each collision is logged at WARN naming the token and both OIDs, and surfaced to administrators (startup summary and/or an admin-visible warning) with the recommended action: rename the custom code. Silent tolerance is not acceptable — the point is that the condition is visible and finite, not that it is ignored.
- Release note documents the precedence rule and how an operator can check whether their deployment is affected.
- Tests:
  - Two RDN entries contesting one code token resolve to the **custom** OID, and resolve to it **stably across a cache rebuild** (add, edit and delete of an unrelated custom OID must not flip the winner). This is the regression test for the arbitrary-iteration-order defect.
  - BouncyCastle symbol parity — for every `SystemOid` RDN constant, `code` equals `BCStyle.INSTANCE.oidToDisplayName(oid)` case-insensitively wherever BC defines a symbol, with an explicit frozen-exception list of `EMAIL`, `SURNAME` and `DISTINGUISHED_NAME_QUALIFIER`. These three divergences shipped long ago and must not be "fixed" — `SN` for surname is the OpenSSL and RFC 4519 short name and is better than BC's `SURNAME`, BC's `DN` for dnQualifier collides with the universal meaning of "DN", and renaming any of them would require a DN rewrite across every affected row. Do not ship this test asserting a single exception; it will fail immediately.
  - Code and altCode uniqueness, case-insensitive, across all RDN constants. Compare *across* constants after deduping case-insensitively *within* a constant, otherwise a legitimate entry trips the test against itself.
  - No system code or altCode shadows a BouncyCastle keyword bound to a different OID: if `BCStyle.INSTANCE.attrNameToOID(token)` resolves, it must resolve to that constant's own OID. Registry-first parsing means our binding silently overrides BC's, which would misparse DN strings arriving from any BC- or OpenSSL-conventioned system.
  - `CustomOidEntryServiceITest` is updated — it currently asserts the system `CERTIFICATE_EXTENSION` list is empty and will fail by design.
- Release note enumerates the newly reserved OIDs, since operators lose the ability to create or edit them as Custom OIDs.

### Out of scope

This does **not** fix core#1883. The STRICT whitelist is evaluated against the resolved request-attribute set, not against the OID registry, so populating the registry adds no mappings. An RA profile with no mappings still falls back to the platform default set of six RDN-only attributes and still rejects Extended Key Usage. STRICT also remains unusable for ACME out of the box regardless of extension coverage, because that default set maps no SAN and every `dNSName` is therefore rejected. This work makes the manual workaround reachable; core#1883 stays open for the whitelist fix itself.

lubomirw:
Closes #1894

## Blocked by interfaces#792 — expect a red build until it merges

This compiles against `com.otilm:interfaces:2.18.1-SNAPSHOT` containing `getDefaultCritical()` and `getValueEncoding()` on `SystemOid`. Those ship in **interfaces#792**, still open, so CI here resolves the last published SNAPSHOT and fails to compile. Verified locally against a `mvn install` of that branch.

Merge order: interfaces#792 → its SNAPSHOT publishes → this.

## What changed

**Registry propagation.** `getOidToRecordMap` carries `defaultCritical` and `valueEncoding` into system-derived `OidRecord`s. Without it the projector reads them unset and emits Name Constraints as **non-critical**, contrary to RFC 5280 §4.2.1.10 — Key Usage and Basic Constraints only looked correct because `X509RequestContentRenderer.CRITICALITY_FORCED_OIDS` force-criticals them, masking the gap.

**API contract.** `toDetailDto(SystemOid)` gains a `CERTIFICATE_EXTENSION` branch, so `GET /v1/oids/system` stops returning entries whose required `defaultCritical` and `valueEncoding` are absent.

**SAN mapping guard.** `ExtensionMappedField` on `2.5.29.17` is rejected pointing at the SAN mapping target — the parser diverts SAN out of the extension list, so such a mapping could never match. Scoped to that one OID: it is the only OID special-cased in the request pipeline, and EKU, keyUsage and basicConstraints must stay mappable.

**Deterministic RDN code resolution.** `getCodeToOidMap` resolved a contested token by `HashMap` iteration order, so the winner was arbitrary and could flip on any unrelated cache rebuild, silently repointing a DN to another attribute type. It now resolves in favour of the operator-registered entry and is case-insensitive at the source, which also aligns the authoring gate with request-time resolution (authoring previously rejected `postalcode` while runtime accepted it).

**Shadowed rows are preserved, reported and manageable.** A `custom_oid_entry` row can predate its OID's promotion. System entries are merged with `putIfAbsent`, so the operator's record wins; the row is reported; it stays editable, with edits reaching the cache; and deleting it hands the OID back to the built-in.

**Conflicts are queryable state, not log noise.** `refreshCache` runs every 30 s by default, so a per-rebuild log line would emit ~2,880 identical lines a day per conflict. Both sets are published, logged only on change, and readable via `OidHandler.getRdnCodeConflicts()` and `CustomOidEntryServiceImpl.getShadowedCustomOidEntries()`.

## Decisions worth reviewing

**One consistent rule: the operator's record wins, and nothing is mutated.** This now applies to both hazards — a contested *code* across two OIDs, and a *same-OID* row shadowed by a promotion. Before the promotion only the operator's entry was in play, so letting it keep winning means the upgrade changes nothing. Failing startup was rejected as disproportionate (core down means no issuance or revocation, and SaaS tenants cannot self-remediate); deleting the row was rejected because it discards operator-chosen criticality and encoding irreversibly. Cost is bounded: DN *rendering* is OID→code, so no collision is possible in that direction. (An earlier version of this description claimed the mapping UI stores only dotted OIDs — it does not; `RdnMappedField.rdn` accepts a short code or a dotted OID.)

**The delete migration in the DoD is deliberately not implemented.** With shadowed rows now preserved and recoverable in place — editable, deletable, reported — the destructive migration has no remaining job. Recommend listing the newly reserved OIDs in the release note instead.

## Review round — what a multi-reviewer pass changed

Four independent reviewers (Copilot CLI, guided review, superpowers, hygiene) produced 41 findings, deduped to 24. Two clusters were substantive and both are fixed in `e9e5d47`:

**`caControlled` is reverted, here and in interfaces#792.** An earlier revision marked basicConstraints, nameConstraints and subjectKeyIdentifier as CA-owned and refused them as mapping targets. That was wrong: the STRICT whitelist admits an extension only when some request attribute maps its OID — `mappedExtensionOids` is populated solely from `ExtensionMappedField`, and the check is plain set membership with no exemption path. So refusing them as mapping targets also refused them **in incoming CSRs**, breaking the case core#1883 exists to fix, for an extension seeded precisely because `openssl req -addext` emits it. An earlier version of this PR body claimed they "remain admissible under the STRICT whitelist" — that claim was false. Blocking requester-driven CA-owned extensions properly needs a whitelist exemption plus a projection-time guard, designed together, which is separate work.

**Same-OID shadowing was destructive and silent.** `getOidToRecordMap` replaced the entire `OidRecord` for an OID a custom row already held, dropping the operator's code, alt codes, criticality and encoding before the contested-code logic ever ran — no contest, no conflict entry. A lost code leaves the registry entirely, so `resolveOid` fails and every DN carrying it breaks at request time; interfaces#792 promotes exactly the five RDN OIDs an operator previously had to hand-register. The reviewers were right that this made skipping the delete migration incoherent — one hazard was handled non-destructively while its twin was destructive.

Also from that pass: `getCodeToOidMap` no longer publishes state from unsynchronized read paths (it was called per CSR validation, per EST `csrattrs` call, and inside the Intune revocation loop, so a reader could overwrite the writer's fresh snapshot); the shadowed-row set is derived from one query and published in a single assignment, replacing a per-category read-modify-write on a `volatile` field; a mapper test that would pass vacuously against an interfaces build with no seeded extensions now asserts non-empty first.

## Two review predictions that did not hold

`PlatformX500NameStyleITest` was expected to break because its fixture registers code `UID` on custom OID `1.2.3.4.5.6`. It passes **8/8 unchanged** — that fixture is a *custom* entry, so operator-wins keeps it resolving exactly as the test already asserts. Core's own pre-existing test independently validates the precedence direction.

Only `CustomOidEntryServiceITest` needed updating, for its assertion that the system `CERTIFICATE_EXTENSION` list is empty.

## Tests

66 green locally across the six affected classes.

Both non-obvious behaviours were proved rather than assumed — the logic was written before its test in each case, so the test was checked against the old behaviour first:

- Reverting `claimToken` to last-writer-wins **and** the `TreeSet` iteration to raw `HashMap` order makes both precedence tests fail, resolving to the system OID instead of the operator's. Both reverts are needed: with the sorted iteration retained, the system OID happens to sort first, so last-writer-wins alone would still leave the custom entry winning.
- Restoring the destructive `put` makes the two shadowing tests fail with `expected: <0.9.2342.19200300.100.1.1> but was: <null>` (the code left the registry) and `expected: <UTF8_STRING> but was: <DER>`.

Conflict reporting is asserted through the published state rather than by capturing log output. `CustomOidEntryMapperTest` asserts every seeded extension surfaces both required fields, so a missing branch cannot ship entries that fail their own schema.

## Still outstanding

- **documentation PR #332** states "There are no built-in `Certificate Extension` entries. All extension OIDs are registered by you as **Custom OIDs**" and walks the operator through creating them — both become false on merge, since `createCustomOidEntry` now rejects the reserved OIDs. That page is the natural home for the release note listing them.
- No admin API serves the two conflict accessors, so the DoD's "admin-visible warning" is half met: the state exists, nothing exposes it.
- `CertificateRequestAttributeProjector.toRequestedExtension` falls back to non-critical with no encoding when the registry lacks an OID, rather than to `SystemOid`, leaving two sources of truth for built-in criticality.

### DIFF
--- a/src/main/java/com/otilm/core/oid/OidHandler.java
+++ b/src/main/java/com/otilm/core/oid/OidHandler.java
@@ -1,15 +1,19 @@
 package com.otilm.core.oid;
 
 import com.otilm.api.model.core.oid.OidCategory;
+import com.otilm.api.model.core.oid.SystemOid;
+import lombok.extern.slf4j.Slf4j;
 
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Pattern;
 
+@Slf4j
 public class OidHandler {
 
     private OidHandler() {
@@ -91,31 +95,54 @@ public class OidHandler {
         rdnCodeToOid.set(Collections.unmodifiableMap(lookup));
     }
 
+    /**
+     * RDN code/altCode → OID, matched case-insensitively so the authoring-time gate and request-time
+     * resolution agree. Codes and alt codes share one flat namespace, so two OIDs can claim the same
+     * token; see {@link #claimToken} for how that is resolved.
+     */
     public static Map<String, String> getCodeToOidMap() {
-        Map<String, String> reverseMap = new HashMap<>();
+        Map<String, String> reverseMap = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
         Map<String, OidRecord> rdnCache = oidCache.get(OidCategory.RDN_ATTRIBUTE_TYPE);
         if (rdnCache == null) {
             return reverseMap;
         }
-        for (Map.Entry<String, OidRecord> entry : rdnCache.entrySet()) {
-            String oidKey = entry.getKey();
-            OidRecord oidRecord = entry.getValue();
-
-            // map the code to the oidKey
+        // Sorted so a contested token resolves the same way on every rebuild, rather than by the
+        // iteration order of the backing HashMap.
+        for (String oid : new TreeSet<>(rdnCache.keySet())) {
+            OidRecord oidRecord = rdnCache.get(oid);
             if (oidRecord.code() != null) {
-                reverseMap.put(oidRecord.code(), oidKey);
+                claimToken(reverseMap, oidRecord.code(), oid);
             }
-
-            // map all altCodes to the oidKey
             if (oidRecord.altCodes() != null) {
                 for (String altCode : oidRecord.altCodes()) {
                     if (altCode != null) {
-                        reverseMap.put(altCode, oidKey);
+                        claimToken(reverseMap, altCode, oid);
                     }
                 }
             }
         }
         return reverseMap;
+    }
+
+    /**
+     * Assigns one code token to an OID, resolving a contest in favour of the operator-registered
+     * entry. A custom OID entry was already resolving that token before its OID became a system OID,
+     * so an upgrade must not silently repoint it; the built-in entry stays reachable by its dotted
+     * OID. Contests between two entries of the same kind keep the lowest OID, purely for determinism.
+     */
+    private static void claimToken(Map<String, String> reverseMap, String token, String oid) {
+        String incumbent = reverseMap.get(token);
+        if (incumbent == null || incumbent.equals(oid)) {
+            reverseMap.put(token, oid);
+            return;
+        }
+        boolean incumbentIsSystem = SystemOid.fromOID(incumbent) != null;
+        boolean candidateIsSystem = SystemOid.fromOID(oid) != null;
+        String winner = incumbentIsSystem && !candidateIsSystem ? oid : incumbent;
+        reverseMap.put(token, winner);
+        log.warn("RDN code '{}' is claimed by OID {} and OID {}; resolving it to {}. "
+                        + "Rename the custom OID entry's code or alt code to remove the ambiguity.",
+                token, incumbent, oid, winner);
     }
 
 
--- a/src/main/java/com/otilm/core/oid/OidHandler.java
+++ b/src/main/java/com/otilm/core/oid/OidHandler.java
@@ -7,6 +7,7 @@ import lombok.extern.slf4j.Slf4j;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
@@ -36,6 +37,10 @@ public class OidHandler {
      */
     private static final AtomicReference<Map<String, String>> rdnCodeToOid =
             new AtomicReference<>(Collections.emptyMap());
+
+    /** Code tokens claimed by more than one OID, republished on every rebuild. See {@link #getRdnCodeConflicts}. */
+    private static final AtomicReference<Map<String, Set<String>>> rdnCodeConflicts =
+            new AtomicReference<>(Map.of());
 
     /**
      * Serializes writers so that the read-copy-publish of a per-category map and the rebuild of the
@@ -102,8 +107,10 @@ public class OidHandler {
      */
     public static Map<String, String> getCodeToOidMap() {
         Map<String, String> reverseMap = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+        Map<String, Set<String>> conflicts = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
         Map<String, OidRecord> rdnCache = oidCache.get(OidCategory.RDN_ATTRIBUTE_TYPE);
         if (rdnCache == null) {
+            publishRdnCodeConflicts(conflicts, reverseMap);
             return reverseMap;
         }
         // Sorted so a contested token resolves the same way on every rebuild, rather than by the
@@ -111,17 +118,47 @@ public class OidHandler {
         for (String oid : new TreeSet<>(rdnCache.keySet())) {
             OidRecord oidRecord = rdnCache.get(oid);
             if (oidRecord.code() != null) {
-                claimToken(reverseMap, oidRecord.code(), oid);
+                claimToken(reverseMap, conflicts, oidRecord.code(), oid);
             }
             if (oidRecord.altCodes() != null) {
                 for (String altCode : oidRecord.altCodes()) {
                     if (altCode != null) {
-                        claimToken(reverseMap, altCode, oid);
+                        claimToken(reverseMap, conflicts, altCode, oid);
                     }
                 }
             }
         }
+        publishRdnCodeConflicts(conflicts, reverseMap);
         return reverseMap;
+    }
+
+    /**
+     * RDN code tokens claimed by more than one OID, mapped to every claimant. Empty when the registry
+     * is unambiguous. Exposed so the condition can be surfaced to an operator rather than living only
+     * in the log — the cache rebuilds every few seconds, so a per-rebuild log line is not a report.
+     */
+    public static Map<String, Set<String>> getRdnCodeConflicts() {
+        return rdnCodeConflicts.get();
+    }
+
+    /**
+     * Publishes the current conflict set, logging only when it differs from the last one. The cache is
+     * rebuilt on a short schedule, so logging per rebuild would emit the same warning thousands of
+     * times a day until an operator resolved it.
+     */
+    private static void publishRdnCodeConflicts(Map<String, Set<String>> conflicts, Map<String, String> resolved) {
+        Map<String, Set<String>> published = Collections.unmodifiableMap(new TreeMap<>(conflicts));
+        Map<String, Set<String>> previous = rdnCodeConflicts.getAndSet(published);
+        if (published.equals(previous)) {
+            return;
+        }
+        if (published.isEmpty()) {
+            log.info("RDN code conflicts resolved; every code and alt code now maps to a single OID.");
+            return;
+        }
+        published.forEach((token, claimants) -> log.warn(
+                "RDN code '{}' is claimed by OIDs {}; resolving it to {}. Rename the custom OID entry's code "
+                        + "or alt code to remove the ambiguity.", token, claimants, resolved.get(token)));
     }
 
     /**
@@ -130,7 +167,8 @@ public class OidHandler {
      * so an upgrade must not silently repoint it; the built-in entry stays reachable by its dotted
      * OID. Contests between two entries of the same kind keep the lowest OID, purely for determinism.
      */
-    private static void claimToken(Map<String, String> reverseMap, String token, String oid) {
+    private static void claimToken(Map<String, String> reverseMap, Map<String, Set<String>> conflicts,
+                                   String token, String oid) {
         String incumbent = reverseMap.get(token);
         if (incumbent == null || incumbent.equals(oid)) {
             reverseMap.put(token, oid);
@@ -138,11 +176,8 @@ public class OidHandler {
         }
         boolean incumbentIsSystem = SystemOid.fromOID(incumbent) != null;
         boolean candidateIsSystem = SystemOid.fromOID(oid) != null;
-        String winner = incumbentIsSystem && !candidateIsSystem ? oid : incumbent;
-        reverseMap.put(token, winner);
-        log.warn("RDN code '{}' is claimed by OID {} and OID {}; resolving it to {}. "
-                        + "Rename the custom OID entry's code or alt code to remove the ambiguity.",
-                token, incumbent, oid, winner);
+        reverseMap.put(token, incumbentIsSystem && !candidateIsSystem ? oid : incumbent);
+        conflicts.computeIfAbsent(token, t -> new TreeSet<>()).addAll(Set.of(incumbent, oid));
     }
 
 
--- a/src/main/java/com/otilm/core/oid/OidHandler.java
+++ b/src/main/java/com/otilm/core/oid/OidHandler.java
@@ -95,9 +95,12 @@ public class OidHandler {
         if (category != OidCategory.RDN_ATTRIBUTE_TYPE) {
             return;
         }
-        Map<String, String> lookup = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
-        lookup.putAll(getCodeToOidMap());
+        // The single writer of conflict state: this runs under WRITE_LOCK, so the published snapshot and
+        // the index it describes always come from the same rebuild. Readers never publish.
+        Map<String, Set<String>> conflicts = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+        Map<String, String> lookup = buildCodeToOid(conflicts);
         rdnCodeToOid.set(Collections.unmodifiableMap(lookup));
+        publishRdnCodeConflicts(conflicts, lookup);
     }
 
     /**
@@ -106,11 +109,17 @@ public class OidHandler {
      * token; see {@link #claimToken} for how that is resolved.
      */
     public static Map<String, String> getCodeToOidMap() {
+        return buildCodeToOid(new TreeMap<>(String.CASE_INSENSITIVE_ORDER));
+    }
+
+    /**
+     * Builds the code → OID index, recording any contested token into {@code conflicts}. Side-effect
+     * free: publication is the caller's job, so request-path readers cannot race the writer.
+     */
+    private static Map<String, String> buildCodeToOid(Map<String, Set<String>> conflicts) {
         Map<String, String> reverseMap = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
-        Map<String, Set<String>> conflicts = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
         Map<String, OidRecord> rdnCache = oidCache.get(OidCategory.RDN_ATTRIBUTE_TYPE);
         if (rdnCache == null) {
-            publishRdnCodeConflicts(conflicts, reverseMap);
             return reverseMap;
         }
         // Sorted so a contested token resolves the same way on every rebuild, rather than by the
@@ -128,23 +137,21 @@ public class OidHandler {
                 }
             }
         }
-        publishRdnCodeConflicts(conflicts, reverseMap);
         return reverseMap;
     }
 
     /**
      * RDN code tokens claimed by more than one OID, mapped to every claimant. Empty when the registry
-     * is unambiguous. Exposed so the condition can be surfaced to an operator rather than living only
-     * in the log — the cache rebuilds every few seconds, so a per-rebuild log line is not a report.
+     * is unambiguous.
      */
     public static Map<String, Set<String>> getRdnCodeConflicts() {
         return rdnCodeConflicts.get();
     }
 
     /**
-     * Publishes the current conflict set, logging only when it differs from the last one. The cache is
-     * rebuilt on a short schedule, so logging per rebuild would emit the same warning thousands of
-     * times a day until an operator resolved it.
+     * Publishes the current conflict set, logging only when it differs from the last one. The registry
+     * rebuilds on {@code settings.cache.refresh-interval} (30 s by default), so logging per rebuild
+     * would repeat the same warning thousands of times a day until an operator resolved it.
      */
     private static void publishRdnCodeConflicts(Map<String, Set<String>> conflicts, Map<String, String> resolved) {
         // Deep-immutable, and case-insensitive like every other code lookup here. Both matter because
@@ -158,7 +165,6 @@ public class OidHandler {
         if (published.equals(previous)) {
             return;
         }
-        // An empty set logs nothing: the state is queryable, and the log carries problems only.
         published.forEach((token, claimants) -> log.warn(
                 "RDN code '{}' is claimed by OIDs {}; resolving it to {}. Rename the custom OID entry's code "
                         + "or alt code to remove the ambiguity.", token, claimants, resolved.get(token)));
@@ -168,7 +174,8 @@ public class OidHandler {
      * Assigns one code token to an OID, resolving a contest in favour of the operator-registered
      * entry. A custom OID entry was already resolving that token before its OID became a system OID,
      * so an upgrade must not silently repoint it; the built-in entry stays reachable by its dotted
-     * OID. Contests between two entries of the same kind keep the lowest OID, purely for determinism.
+     * OID. Contests between two entries of the same kind keep the lexicographically first OID; determinism is
+     * the guarantee, not any numeric ordering of the arcs.
      */
     private static void claimToken(Map<String, String> reverseMap, Map<String, Set<String>> conflicts,
                                    String token, String oid) {
--- a/src/main/java/com/otilm/core/oid/OidHandler.java
+++ b/src/main/java/com/otilm/core/oid/OidHandler.java
@@ -147,7 +147,13 @@ public class OidHandler {
      * times a day until an operator resolved it.
      */
     private static void publishRdnCodeConflicts(Map<String, Set<String>> conflicts, Map<String, String> resolved) {
-        Map<String, Set<String>> published = Collections.unmodifiableMap(new TreeMap<>(conflicts));
+        // Deep-immutable, and case-insensitive like every other code lookup here. Both matter because
+        // this is process-wide static state handed out through a public accessor: a mutable claimant set
+        // would let a caller corrupt it, and note that TreeMap(Map) would silently drop the comparator.
+        Map<String, Set<String>> snapshot = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+        conflicts.forEach((token, claimants) ->
+                snapshot.put(token, Collections.unmodifiableSortedSet(new TreeSet<>(claimants))));
+        Map<String, Set<String>> published = Collections.unmodifiableMap(snapshot);
         Map<String, Set<String>> previous = rdnCodeConflicts.getAndSet(published);
         if (published.equals(previous)) {
             return;
--- a/src/main/java/com/otilm/core/oid/OidHandler.java
+++ b/src/main/java/com/otilm/core/oid/OidHandler.java
@@ -38,6 +38,13 @@ public class OidHandler {
     private static final AtomicReference<Map<String, String>> rdnCodeToOid =
             new AtomicReference<>(Collections.emptyMap());
 
+    /**
+     * Bumped under {@link #WRITE_LOCK} by every publication. Lets a caller that read source data without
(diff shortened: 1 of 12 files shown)


src/main/java/com/otilm/core/oid/OidHandler.java
```diff
@@ -91,33 +100,83 @@ private static void refreshRdnCodeLookup(OidCategory category) {
...
+    }
+
+    /**
+     * Publishes the current conflict set, logging only when it differs from the last one. The cache is
+     * rebuilt on a short schedule, so logging per rebuild would emit the same warning thousands of
+     * times a day until an operator resolved it.
+     */
+    private static void publishRdnCodeConflicts(Map<String, Set<String>> conflicts, Map<String, String> resolved) {
+        Map<String, Set<String>> published = Collections.unmodifiableMap(new TreeMap<>(conflicts));
+        Map<String, Set<String>> previous = rdnCodeConflicts.getAndSet(published);
```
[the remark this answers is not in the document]
lubomirw: Fixed in 5034aab55 — confirmed, and the check surfaced a second defect alongside it.

Your point stands exactly as written: `new TreeMap<>(conflicts)` is a shallow copy, so the `TreeSet` values published inside the unmodifiable map were the same instances built in `claimToken`. Since `getRdnCodeConflicts()` returns that map directly, a caller could reach through and mutate process-wide static state — and it would also corrupt the change-detection that keeps the warning from re-firing on every cache rebuild, which is the whole reason the state is published rather than just logged.

Each claimant set is now copied and wrapped before publication.

The second defect: that same copy silently dropped the comparator. `conflicts` is a `TreeMap` with `String.CASE_INSENSITIVE_ORDER`, but its static type is `Map`, so `new TreeMap<>(Map)` was selected over the `SortedMap` overload and the snapshot fell back to natural ordering. A lookup for `uid` returned null while `UID` resolved — inconsistent with every other code lookup in this class. The snapshot is now built with `CASE_INSENSITIVE_ORDER` explicitly.

Two tests were added and both were confirmed failing first: `publishedConflictState_isDeeplyImmutable` (mutation must throw) and `publishedConflictState_looksUpTokensCaseInsensitively`.

Worth noting the sibling accessor is unaffected — `getShadowedCustomOidEntries()` publishes a `Set<String>`, and `String` elements carry no deep-mutability risk.

src/main/java/com/otilm/core/service/impl/CustomOidEntryServiceImpl.java
```diff
@@ -270,6 +361,16 @@ public List<SearchFieldDataByGroupDto> getSearchableFieldInformation() {
         return searchFieldDataByGroupDtos;
     }
 
+    /**
+     * OIDs held by a {@code custom_oid_entry} row that a built-in system OID now shadows, so the row's
+     * configured properties no longer apply. Exposed so the condition can be surfaced to an operator;
+     * the resolution is to delete the row.
+     */
```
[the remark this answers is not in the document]
lubomirw: Valid, and the sharpest of the four — fixed in `59b37d649`.

You're right that it contradicts the behaviour, and worth recording why: that Javadoc was **true when written**. Shadowing used to be destructive — `getOidToRecordMap` did `putAll` of the built-in entries over the custom ones, so a shadowed row's configured code, alt codes, criticality and encoding really did stop applying. A later commit on this branch changed that to `putIfAbsent` so the operator's record wins, which inverted the behaviour and left this Javadoc asserting the opposite.

Rather than reword it, I deleted it. The interface declaration already documents the current behaviour correctly ("The custom entry wins, so the built-in defaults do not apply; deleting it falls back to them"), and `@Override` inherits it — so having two descriptions was exactly what let them drift apart. One source of truth now.

The runtime WARN text was already consistent with the new behaviour, so nothing operator-facing was wrong; this was documentation-only.

src/main/java/com/otilm/core/oid/OidHandler.java
```diff
@@ -1,15 +1,20 @@
 package com.otilm.core.oid;
 
 import com.otilm.api.model.core.oid.OidCategory;
+import com.otilm.api.model.core.oid.SystemOid;
+import lombok.extern.slf4j.Slf4j;
```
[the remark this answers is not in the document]
lubomirw: Correct — removed in `59b37d649`. It became unused when the code-token contest switched from asking `SystemOid.fromOID(oid)` to reading a provenance flag off `OidRecord`; the only remaining mention is inside a comment explaining why that switch was made, which needs no import.

src/main/java/com/otilm/core/oid/OidHandler.java
```diff
@@ -86,38 +95,114 @@ private static void refreshRdnCodeLookup(OidCategory category) {
...
+    /**
+     * Publishes the current conflict set, logging only when it differs from the last one. The registry
+     * rebuilds on {@code settings.cache.refresh-interval} (30 s by default), so logging per rebuild
+     * would repeat the same warning thousands of times a day until an operator resolved it.
+     */
+    private static void publishRdnCodeConflicts(Map<String, Set<String>> conflicts, Map<String, String> resolved) {
+        // Deep-immutable, and case-insensitive like every other code lookup here. Both matter because
+        // this is process-wide static state handed out through a public accessor: a mutable claimant set
+        // would let a caller corrupt it,; TreeMap(Map) would drop the comparator.
+        Map<String, Set<String>> snapshot = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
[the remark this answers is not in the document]
lubomirw: Fixed in `59b37d649`. Self-inflicted: an earlier edit trimmed a trailing clause off that sentence and replaced it with a semicolon without removing the comma. Now reads "would let a caller corrupt it, and TreeMap(Map) would drop the comparator."

src/main/java/com/otilm/core/service/CustomOidEntryExternalService.java
```diff
@@ -65,6 +66,12 @@ public interface CustomOidEntryExternalService {
      * @param category optional category filter; when {@code null}, all system OIDs are returned
      * @return list of system OID entries
      */
+    /**
+     * OIDs held by a custom entry that a built-in system OID now shares. The custom entry wins, so the
+     * built-in defaults do not apply; deleting it falls back to them. Empty when there is no conflict.
+     */
+    Set<String> getShadowedCustomOidEntries();
+
     List<CustomOidEntryDetailResponseDto> listSystemOidEntries(OidCategory category);
```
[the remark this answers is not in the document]
lubomirw: Valid — fixed in `59b37d649`, and it was slightly worse than described.

The new declaration was inserted between `listSystemOidEntries`' Javadoc and its signature, so all three of these were true at once: the "Returns the built-in system OID entries" block was left dangling, my block bound to `getShadowedCustomOidEntries()` (which happened to be correct), and `listSystemOidEntries` ended up undocumented.

`getShadowedCustomOidEntries` now sits below `listSystemOidEntries`, restoring the original Javadoc-to-method pairing. Caused by a scripted insertion anchored on the method signature rather than on its preceding doc comment.

src/main/java/com/otilm/core/service/impl/CustomOidEntryServiceImpl.java
```diff
@@ -68,9 +74,62 @@ public CustomOidEntryServiceImpl(CustomOidEntryRepository oidEntryRepository) {
...
+                .map(CustomOidEntry::getOid)
+                .collect(Collectors.toCollection(TreeSet::new));
+        if (shadowed.equals(shadowedCustomOidEntries)) {
+            return;
+        }
+        shadowedCustomOidEntries = Collections.unmodifiableSet(shadowed);
+        shadowed.forEach(oid -> log.warn(
+                "Custom OID entry {} shares its OID with a built-in system OID. The custom entry wins, so the "
+                        + "built-in defaults do not apply; delete the custom entry to fall back to them.", oid));
     }
```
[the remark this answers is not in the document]
lubomirw: Valid and adopted in `d4a207f03` — the scan is gone.

Your suggested derivation works exactly as described, and the piece that makes it clean already exists: `OidRecord.system`, added earlier in this PR so a contested RDN code could be decided on record provenance rather than on whether the dotted OID happens to be built-in. Because system entries are merged with `putIfAbsent`, a shadowed OID holds the **operator's** record (`system == false`) while an unshadowed one holds the built-in (`system == true`), so the set is a filter over `SystemOid.values()` against the published cache — no database access, and category scoping preserved by looking the OID up in its own category.

Both call sites are safe for this: the refresh calls it only after `cacheAllCategories` has committed its snapshot, and the delete path only after the built-in is back in the cache. So the cache is authoritative at each.

One thing worth adding to your rationale — it is not only cheaper, it is **more correct**. The accessor answers "which rows are currently shadowed in the registry", and a table scan answered a subtly different question about the database. Between a commit and its `afterCommit` publication those two disagree, and the registry-derived answer is the one that matches what request paths actually resolve.

Existing coverage validates it in both directions without change: `testShadowedRdnRowKeepsItsCodeResolving` asserts the set contains the seeded OID, `testDeletingAShadowedRowRestoresTheBuiltInEntry` asserts it clears once the row is gone. 72/72 green.

src/main/java/com/otilm/core/oid/OidHandler.java
```diff
@@ -86,38 +140,114 @@ private static void refreshRdnCodeLookup(OidCategory category) {
...
+        // would let a caller corrupt it, and TreeMap(Map) would drop the comparator.
+        Map<String, Set<String>> snapshot = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+        conflicts.forEach((token, claimants) ->
+                snapshot.put(token, Collections.unmodifiableSortedSet(new TreeSet<>(claimants))));
+        Map<String, Set<String>> published = Collections.unmodifiableMap(snapshot);
+        Map<String, Set<String>> previous = rdnCodeConflicts.getAndSet(published);
+        if (published.equals(previous)) {
+            return;
+        }
+        published.forEach((token, claimants) -> logger.warn(
```
klaraf755, line 220: The change-detection logs when a conflict *appears*, but the resolution transition is silent: when the set goes non-empty → empty, `published.equals(previous)` is false so we get past the guard, but `published.forEach` then iterates an empty map and logs nothing. An operator's log stream reads "one warning, then permanent silence" — and silence is what both "still broken, deduplicated" and "fixed" produce, which matters while nothing exposes `getRdnCodeConflicts()` over the API yet.

This branch already knows the transition happened — one `logger.info` when `previous` had entries that `published` no longer has ("RDN code conflict on '{}' resolved") would complete the story. Same applies to `publishShadowedCustomOidEntries`, which has the same one-sided logging.

src/main/java/com/otilm/core/service/impl/CustomOidEntryServiceImpl.java
```diff
@@ -187,15 +254,17 @@ else if (customOidEntry instanceof CertificateExtensionCustomOidEntry extensionE
...
-        if (SystemOid.fromOID(oid) == null) {
-            OidHandler.cacheOid(customOidEntry.getCategory(), oid, OidRecord.builder()
-                    .displayName(customOidEntry.getDisplayName())
-                    .code(code)
-                    .altCodes(altCodes)
-                    .defaultCritical(defaultCritical)
-                    .valueEncoding(valueEncoding)
-                    .build());
-        }
+        // Cached unconditionally: a custom row wins over a same-OID built-in (see getOidToRecordMap),
```
klaraf755, line 270: Adjacent, pre-existing, but this PR makes it more visible: the conflict warning tells the operator to *rename the custom OID entry's code*, and a case-only rename is impossible — a few lines above this hunk, `!oldCode.equals(code)` compares case-sensitively, so `FOO`→`Foo` proceeds into the uniqueness check, where `getAllCodesInLowerCase()` (which includes the row's own code) rejects it as "already used". With the authoring gate now case-insensitive-at-source, `equalsIgnoreCase` here (or excluding the row's own codes from the set) is a two-line fix that would fit this PR or an immediate follow-up.

src/main/java/com/otilm/core/service/CustomOidEntryExternalService.java
```diff
@@ -67,6 +68,12 @@ public interface CustomOidEntryExternalService {
      */
     List<CustomOidEntryDetailResponseDto> listSystemOidEntries(OidCategory category);
 
+    /**
+     * OIDs held by a custom entry that a built-in system OID now shares. The custom entry wins, so the
+     * built-in defaults do not apply; deleting it falls back to them. Empty when there is no conflict.
+     */
+    Set<String> getShadowedCustomOidEntries();
```
klaraf755, line 80: Worth stating *how* a shadowed row can exist: `createCustomOidEntry` rejects reserved OIDs, so the only source is a row that predates its OID's promotion to a system OID in an upgrade. That invariant tells the reader two useful things — the set can only ever shrink between releases (operators resolve entries, upgrades add them), and an empty set on a fresh install is expected, not a code path that never ran.

src/main/java/com/otilm/core/service/impl/CustomOidEntryServiceImpl.java
lubomirw: Valid and fixed in `4af481f84`. Verified the mechanism rather than taking it on description, and it is exactly as you say.

`rdnOidEntry` has code `RDN`; renaming to `Rdn` hits `!oldCode.equals(code)` (case-sensitive, so true), enters the uniqueness check, and `getAllCodesInLowerCase()` contains `rdn` from the row's own registry entry — `Code Rdn is already used`.

I took the second option you offered rather than `equalsIgnoreCase`. Ignoring case there would make the guard skip the branch entirely, so `setCode` and `updateCertificateDNs` never run and a requested case change silently becomes a no-op. Excluding the row's own code instead lets the rename actually happen, including the DN rewrite. The alt-code path is untouched, since it already excludes the row's own alt codes via `oldAltCodes`.

You are also right that this PR is what makes it matter: the contested-code warning I added tells the operator to rename the code, and a case-only collision is precisely the case where that was the only resolution and it was blocked.

Two tests, and the first one caught me out. `testRdnCodeCanBeRenamedToACaseVariantOfItself` initially **passed with the bug still in place** — `setUp` writes the row straight to the repository, and the `@AfterEach` refresh runs against a truncated table, so the row's own code was never in the registry and there was nothing to collide with. Added an explicit `refreshCache()` plus an assertion that `RDN` resolves, and it then failed with `Code Rdn is already used` before the fix. `testRdnCodeStillRejectedWhenAnotherRowOwnsIt` guards the other direction so excluding the row's own code does not weaken the check against other rows.

src/main/java/com/otilm/core/service/CustomOidEntryExternalService.java
lubomirw: Applied in `4af481f84` — good framing, and it answers a question a reader would otherwise have to reverse-engineer from `createCustomOidEntry`.

Added to the interface Javadoc: such a row can only arise from an upgrade, because creating one is rejected as a reserved OID; so the set grows only at an upgrade, shrinks only as operators resolve entries, and an empty set on a fresh install is the expected state rather than an unexercised code path.

Note this replaced a Javadoc that had gone stale in the opposite direction — it previously said a shadowed row's configured properties no longer apply, which was true while shadowing overwrote the row and became false once system entries moved to `putIfAbsent` and the custom record started winning. The interface is now the single description, so the impl cannot drift from it again.

src/main/java/com/otilm/core/oid/OidHandler.java
lubomirw: Not applying this one without a decision from the repo owner, because it reverses an explicit instruction rather than filling a gap.

An all-clear line existed on this branch and was removed deliberately in `cd62cebfd` ("Log only conflicts, never their resolution"), on the principle that the log should carry problems only and the conflict set is queryable state. So the silence you are seeing is intentional, not an oversight.

That said, your argument is stronger than the one I originally made for the all-clear, and it is specifically the part I did not consider: silence is ambiguous between "still broken, deduplicated" and "fixed", and that ambiguity only matters *because* nothing exposes `getRdnCodeConflicts()` or `getShadowedCustomOidEntries()` over the API yet. Those two facts together are the real problem — the log became the sole channel, and a one-sided log is a poor sole channel.

Which suggests the fix is the missing endpoint rather than a second log line. With the accessors served, resolution is observable by querying an empty set, and the log can stay problems-only without ambiguity. Both accessors are already in place for exactly that, and the endpoint is listed as outstanding on this PR.

Flagging both options to the owner: add the resolution transition at info on both publishers (two lines, reverses `cd62cebfd`), or land the diagnostics endpoint and leave the log one-sided. I have a mild preference for the endpoint, since it also removes the reason the warning had to be throttled in the first place. Will apply whichever is chosen.

lubomirw: Addressed in `80188dcc2`, though not the way you proposed — the owner chose the alternative, and I think it lands your point better.

Instead of an info line on the resolution transition, both publishers now warn on a change **and again hourly while the condition lasts**. Silence therefore means resolved, unambiguously, and any recent log window answers the question an operator actually has — does this hold right now — without needing to have caught the original warning. An info-on-resolution line only helps someone who observed the transition.

It also keeps the log carrying problems only. That mattered because the all-clear line you are asking for did exist on this branch and was removed deliberately in `cd62cebfd`; re-adding it would have reversed that decision, whereas re-warning respects it while fixing the ambiguity you identified.

Cost is 24 lines a day per lasting conflict, against 2 under change-only detection and 2,880 before any throttling.

The repeat policy is a small tested class, `PersistentWarningThrottle`, rather than a timestamp field duplicated across `OidHandler` and `CustomOidEntryServiceImpl`. Six unit tests, including the case I would otherwise have got wrong: a condition that resolves and later recurs reports immediately rather than waiting out the window started by the earlier occurrence.

On your framing that this matters *because* nothing exposes the accessors over the API — agreed, and that was priced separately. An endpoint means a new `interfaces` PR and release cycle (the contract lives there and #792 just merged), plus core wiring, plus a UI surface or the invisibility just moves up a level, plus docs: four repos for a condition that only arises at upgrade and is empty in most deployments. Judged overkill for now and left as an outstanding item on the PR.

klaraf755 (approved):

Issue #1903, opened 2026-07-28
Protocols - SCEP: tighten the renewal challenge-password waiver to proven issuance provenance


lubomirw:
### Description

`ScepServiceImpl.authenticateRenewal` waives the SCEP profile's challenge password for an authenticated renewal (added for #1887). The waiver requires the request to prove possession of the renewed certificate's key, the certificate to be in a renewable state, the subject DN to match, and the request to ask for no subject alternative name the renewed certificate does not already carry. Entitlement to *this* profile is then decided by one condition: `renewedCertificate.getRaProfileUuid()` equals the service's RA profile.

That association is weaker than issuance provenance. It is also set by `CertificateServiceImpl.switchRaProfile` and by `createRegistrationPlaceholder`, so a certificate the platform never issued — uploaded or discovered — earns the waiver once an operator associates it with the RA profile. A certificate issued under that RA profile through another channel (ACME, CMP, UI, API) equally earns it, and `scep_profile.ra_profile_uuid` carries no unique constraint, so several SCEP profiles with different challenge passwords can serve one RA profile.

The residual exposure is narrow: the holder of the private key of a valid, non-revoked, in-renewal-window certificate on this RA profile, with a matching subject DN and no additional names, obtains a certificate without the shared secret. It is documented in the method's Javadoc rather than fixed, because every tightening considered carries a functional-regression risk that needs QA against real deployments.

Raised by three independent reviewers during the multi-reviewer triage of the #1887 fix.

### Definition of Done

- A decision recorded on which provenance signal gates the waiver. The two candidates evaluated:
  - **Issued by this profile's CA** — verify the renewed certificate against `scepProfile.getCaCertificate()`'s public key. Real provenance, but withholds the waiver when the profile CA is a parent of the actual issuer, or after a CA key rotation.
  - **SCEP protocol association** — match the protocol-certificate association for this SCEP profile (`CertificateProtocolInfo.Scep`). Strongest, but the association table (`V202508051559__protocol_certificate_association.sql`) postdates existing certificates, whose renewals would lose the waiver.
- The chosen signal implemented in `authenticateRenewal` as an additional waiver condition, withholding the waiver rather than rejecting the renewal, so a client that supplies the challenge password is unaffected.
- Unit coverage for the withheld-waiver case, alongside the existing RA-profile and SAN cases in `ScepServiceImplRenewalAuthenticationTest`.
[...]

Issue #1902, opened 2026-07-28
Protocols - SCEP: verify EC renewal response delivery end-to-end


lubomirw:
### Description

A SUCCESS `CertRep` is enveloped to the client's own key. RSA client keys use CMS key transport; every other key type (EC) uses the RFC 8894 password recipient, which is keyed with the SCEP profile's challenge password — see `ScepResponse.setChallengePassword` / `ScepResponse.java:205-218` and `ScepServiceImpl.prepareMessage`.

It has never been confirmed end-to-end that the common SCEP clients can decrypt that response. sscep and jscep must support the password recipient and must use the same shared secret the profile holds. If they cannot, an EC enrollment or renewal issues a certificate the client is unable to retrieve — the platform commits a certificate the requester never receives.

`ScepServiceImpl.verifyResponseEnvelopable` already rejects an EC client key when the profile has no challenge password (added with #1834 / #1889), so the remaining unknown is purely client-side capability. Surfaced during the analysis of #1887.

### Test Approach

Manual

### Test Scope

SCEP enrollment and renewal against a SCEP profile with a challenge password configured, exercised with an EC client key. Both the profile-based and RA-profile-based SCEP endpoints. Relevant clients: sscep and jscep.

### Test Scenarios

1. Initial enrollment with an EC client key, profile with challenge password configured — confirm the client decrypts the SUCCESS CertRep and stores the issued certificate.
2. Renewal with an EC client key (PKCSReq signed with the existing certificate, no challengePassword in the CSR) — confirm the client decrypts the response. Note the challenge password is waived for the request but still keys the response envelope.
3. Same two scenarios with an RSA client key, as the key-transport control case.
4. Record for each client whether the RFC 8894 password recipient is supported, and the content encryption algorithm negotiated.
5. If a client cannot decrypt the password-enveloped response, capture the failure and raise a follow-up defect — the platform should reject such an enrollment before issuing rather than committing an unretrievable certificate.

### Related Feature

#1887

Issue #1901, opened 2026-07-28
Protocols - SCEP: implement RENEWAL_REQ (message type 17) enrollment


lubomirw:
### Description

SCEP messages with `messageType` 17 (`RENEWAL_REQ`, RFC 8894 §3.3.1) are not implemented. `ScepServiceImpl.processPkiOperation` recognizes the type in the challenge password / renewal validation branch, but the issuance dispatch below it only handles `PKCS_REQ` (19) and `CERT_POLL` (20), so a type 17 message falls through to the final `else` and is answered with `Unsupported Operation. The requested operation is not supported` and `failInfo=badRequest`.

Only the draft-nourse-scep-23 §3.1.1.2 renewal flow works today — a `PKCS_REQ` signed with the key of the certificate being replaced, which is what sscep and jscep send. `GetCACaps` already advertises `Renewal`, so a client that implements the RFC 8894 message type is told the capability exists and then rejected.

Found while fixing #1887 (renewal PKCSReq crashed on a missing challenge password); the fix there covers the PKCS_REQ renewal path only.

### Use Case

A SCEP client that implements RFC 8894 rather than the 2011 draft sends `RENEWAL_REQ` to renew its certificate. It should be served, not rejected, given the platform advertises `Renewal` in `GetCACaps`.

### Acceptance Criteria

- A `RENEWAL_REQ` (17) message carrying a valid PKCS#10 request and signed with the key of the certificate being replaced results in an issued certificate returned in a SUCCESS `CertRep`.
- Renewal authentication and policy for type 17 reuse the same path as the PKCS_REQ renewal flow (`ScepServiceImpl.authenticateRenewal`): signer certificate resolved from the inventory, CMS signature verified, archived / pending state rejected, subject DN matched, renewal timeframe enforced.
- Manual approval configured on the SCEP profile produces a CSR and a PENDING response for type 17, consistent with `PKCS_REQ`.
- A `RENEWAL_REQ` whose signer certificate is unknown to the platform is rejected with a SCEP FAILURE, not treated as an initial enrollment.
- `GetCACaps` continues to advertise `Renewal`, and the advertised capability is now truthful for both message types.
- Test coverage for the type 17 dispatch path.

PR #1904, opened 2026-07-28
Accept and authenticate SCEP renewal without a challenge password

### LINKED ISSUE #1887: Protocols - SCEP: renewal PKCSReq (signed with existing cert, without challengePassword) crashes with NullPointerException on String.equals; platform returns HTTP 500 instead of a SCEP-formatted error response
### Description

When a SCEP client sends a **renewal PKCSReq** — signed with an existing certificate's private key (proof of ownership) and **without a `challengePassword`** attribute in the CSR (renewal auth is by signature, not by shared secret) — the platform's SCEP handler crashes with a **NullPointerException**:

```
Cannot invoke "String.equals(Object)" because "challengePassword" is null
```

The handler unconditionally invokes `.equals()` on the `challengePassword` string extracted from the CSR, without first checking whether the client actually included one. On renewal PKCSReqs — where per RFC 8894 §3.3.1.2 the client authenticates by signing the SCEP message with the existing cert's key rather than by supplying a challenge password — the value is legitimately `null`, so the guard is missing.

The exception is not caught inside the SCEP endpoint, so it falls through to the REST error handler and the client receives **HTTP 500** with a JSON error body instead of a proper `application/x-pki-message` SCEP-formatted rejection. SCEP clients (`sscep`, device SCEP stacks) cannot parse HTTP 500 / JSON as SCEP — they fail at the transport layer with a generic "error while sending message" and have no way to know that the platform saw the request at all.

Consequences:

- **SCEP renewal is entirely non-functional** — the standard RFC 8894 renewal flow (existing-cert signature, no challengePassword) crashes the handler before any real logic runs
- Client sees only a transport-level failure, no actionable protocol-level error
- Platform audit log records the NPE, but the operator has to correlate manually to figure out that the client's "sending" error is actually a server-side crash
- Same class of defect as CMP KUR NPE (#1885) — protocol handlers assume one authentication mode and NPE when the other is used

### Steps to Reproduce

1. Configure a SCEP Profile (Enable Challenge Password + a valid password, Include CA Certificate Chain, Default RA Profile pointing at an EJBCA RA on Lenient validation)
2. Perform an initial enrollment via SCEP `PKCSReq` (with `challengePassword`) — cert appears in Certificate Inventory as `Issued/Valid` (note: due to #1834 the client itself may not receive the cert; download it from the UI)
3. Generate a new key pair + a new CSR **without a `challengePassword` attribute** (renewal-style; same Subject as the existing cert):

   ```
   openssl ecparam -name prime256v1 -genkey -noout -out qa-scep-new.key
   openssl req -new -key qa-scep-new.key -out qa-scep-new.csr \
     -subj "/CN=qa-scep.example.com" \
     -addext "subjectAltName = DNS:qa-scep.example.com"
   ```
4. Run `sscep enroll` in renewal mode — signing the SCEP message with the existing cert (`-O`) and key (`-K`):

   ```
   sscep enroll \
     -u "http://<local-proxy>/api/v1/protocols/scep/<scep-profile>/pkiclient.exe" \
     -c <ra-cert> \
     -k qa-scep-new.key \
     -r qa-scep-new.csr \
     -l qa-scep-renewed.crt \
     -K qa-scep.key \
     -O qa-scep.crt \
     -t 5 -n 12
   ```
5. Observe client:

   ```
   sscep: sending certificate request
   sscep: error while sending message
   ```

6. Observe proxy / server:

   ```
   POST /api/v1/protocols/scep/<scep-profile>/pkiclient.exe?operation=PKIOperation
       << 500  36b
   ```

7. Observe platform audit log for that operation — `operationResult: failure`, message:

   ```
   Cannot invoke "String.equals(Object)" because "challengePassword" is null
   ```

### Expected Behavior

- The SCEP handler recognises a renewal-style `PKCSReq` (existing-cert signature, no `challengePassword`) and processes it per RFC 8894 §3.3.1.2 — authenticates via the signer certificate and issues a new certificate for the same subject
- If `challengePassword` is `null`, the handler either treats it as a valid renewal path or returns a proper `application/x-pki-message` **SCEP-formatted** failure response (e.g. `pkiStatus: FAILURE` with `failInfo: badRequest` / `badMessageCheck`) explaining the missing / disallowed attribute
- No NullPointerException; no fallback to REST HTTP 500 / JSON error handler; client can parse the response as SCEP and surface a meaningful error to the operator

### Actual Behavior

- SCEP handler dereferences the null `challengePassword` unconditionally → NPE
- Response falls through to REST error handler, returned as HTTP 500 with a JSON body — outside the SCEP protocol
- Client (`sscep`) reports only `error while sending message`; no SCEP-level diagnostic
- SCEP renewal is completely non-functional on the platform
- Audit log records the NPE; operator has no visible protocol-level failure to correlate

### Severity

Major

### Module

Protocols

### Attachments

_No response_

### Logs

_No response_

### Environment

ILM Core: 2.18.1-SNAPSHOT (22 Jul 2026, 23:14)
Commit: 5f76a66
Branch: main
Build time: 2026-07-22T21:14:38Z
ILM Frontend: 2.18.0 (23 Jul 2026, 08:15)
Client:
- macOS Tahoe 26.5.1

lubomirw:
Closes #1887.

## Problem

A renewal PKCSReq is signed with the key of the certificate it replaces and carries no `challengePassword` attribute (RFC 8894 §3.3.1.2). The challenge-password gate dereferenced the request-side value without a null check, so every renewal against a profile with a password configured threw `NullPointerException`:

```
Cannot invoke "String.equals(Object)" because "challengePassword" is null
```

Being unchecked, it bypassed the `ScepException` → SCEP-response shaping and landed in the generic handler, so the client got a 36-byte JSON 500 it cannot parse. Broken since `21e6d0ccf` (2023) — renewal was never exercised until QA hit it on 2.18.x.

## What changed

**Classify before enforcing the shared secret.** `authenticateRenewal` resolves the signer certificate from the inventory, verifies the CMS signature against it, then applies the renewal checks. The challenge password is waived only for a positively authenticated renewal. A password that is *present* must still match, so a wrong shared secret is never silently accepted; a blank attribute counts as absent, which is what clients commonly send on renewal.

Signature verification now runs **ahead** of the certificate-state checks, so a client replaying a certificate it does not own cannot probe inventory state — it only ever sees the signature failure.

**Waiver conditions.** Proven key possession, plus a certificate in a renewable state whose subject DN matches the request. On top of that, the waiver itself requires the certificate to be associated with this profile's RA profile, its validation to be conclusive, and the request to ask for no subject alternative name the certificate does not already carry — those three only *withhold* the waiver, so a client that does supply the password is unaffected. A subject DN mismatch, an archived, pending, revoked, expired or invalid certificate, and a request outside the renewal window all *reject* the renewal, which is the behaviour that predates the waiver.

Subject alternative names are compared as DER structures rather than as rendered strings, because the two sides render an iPAddress differently (decoded text from the certificate, a hex octet string from the request) — comparing the renderings would silently withhold the waiver from a device renewing on an address it already holds.

**Renewable-state gate.** `checkRenewalTimeframe` only tests for revoked and expired certificates on the branch where a renewal threshold is configured, so with the default (unset) threshold that test never ran. Combined with the waiver, the key holder of a revoked certificate would have enrolled without the shared secret. The state and validation-status check now runs for every configuration, before the timeframe policy.

**Protocol-format responses.** `PKIOperation` answers unexpected runtime failures with a SCEP FAILURE carrying a generic `failInfoText`, mirroring the CMP fix in #1888. Request parsing sits inside the guard (malformed DER surfaces as an unchecked exception), the recovery path is itself guarded (building the failure needs the CA key through the token connector, which may be what failed), and the transaction is rolled back before answering so a failed attempt commits nothing. A checked `ScepException` from `decryptData` is now answered in the SCEP format too, instead of reaching the generic handler as JSON.

**Also fixed:** a latent `NullPointerException` in `ScepRequest.decryptData` when a password-enveloped request meets a profile with no challenge password; a no-op `assert` that let a recipient-less `EnvelopedData` reach the PKCS#10 parse with no content; a BouncyCastle exception message forwarded to the wire in the signature-failure text; missing `failInfo` values on `validateRenewal` rejections. The shared secret is compared with `MessageDigest.isEqual`. `verifyRequest` became private `verifyProofOfPossession` — renewal validation no longer runs there.

## Testing

41 SCEP unit tests and the SCEP integration test pass. New coverage: the reported crash, the waiver and its withholding conditions, revoked and expired certificates under the default null threshold, pending states, signature-before-state ordering, and an `ArgumentCaptor` assertion that the signature is verified against the signer certificate's key rather than any key.

The defect was reproduced against the pre-fix code before the fix was written, yielding the issue's exact message.

**Not covered:** the `pkiOperation` RuntimeException-to-FAILURE conversion. A unit test cannot reach it — `buildResponse` needs the token connector and the CA key — and the existing integration test has no provisioned SCEP profile.

## Known residuals

- If a nested `@Transactional` collaborator has already marked the transaction *globally* rollback-only, the commit still fails and the client gets a 500. Reproduced while writing tests: an issuance failure inside `generateCsr` triggers it, so this is reachable rather than theoretical.
- The guard covers PKIOperation once the message has parsed. A message that does not parse, and the profile-level errors from `init` / `validateProfile` (profile not found, disabled, no CA certificate), still answer as the platform's JSON error — there is no transaction id or nonce to echo, so no SCEP response can be formed. A full guarantee needs the error boundary outside the transaction, which means changes to the controller and the `ScepExternalService` contract in `interfaces`.
- Every runtime fault now answers `200` with `badRequest`, which sscep reads as permanent, and the endpoint no longer emits 5xx for genuine server faults. Deliberate, for CMP parity.
- The waiver trusts the RA-profile association, which an operator can set on a certificate the platform never issued — tracked in #1903.

## Follow-ups filed

- #1901 — `RENEWAL_REQ` (message type 17) is validated but then dispatched to "Unsupported Operation"; `GetCACaps` advertises `Renewal`.
- #1902 — QA: verify end-to-end that sscep and jscep can decrypt a password-enveloped `CertRep` to an EC signer.
- #1903 — tighten the waiver to proven issuance provenance.
- OmniTrustILM/documentation#337 — the SCEP pages still describe the pre-fix challenge-password rules.

### DIFF
--- a/src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
@@ -58,7 +58,6 @@ import org.bouncycastle.cms.CMSProcessableByteArray;
 import org.bouncycastle.cms.CMSSignedDataGenerator;
 import org.bouncycastle.operator.OperatorCreationException;
 import org.bouncycastle.pkcs.PKCSException;
-import org.bouncycastle.pkcs.jcajce.JcaPKCS10CertificationRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -324,11 +323,25 @@ public class ScepServiceImpl implements ScepExternalService {
     }
 
     private ResponseEntity<Object> pkiOperation(byte[] body) throws ScepException {
-        ScepRequest scepRequest;
+        ScepRequest scepRequest = new ScepRequest(body);
+        try {
+            return processPkiOperation(scepRequest);
+        } catch (RuntimeException e) {
+            // An unexpected runtime failure would otherwise escape to the generic error handler and leave
+            // the client with an HTTP 500 JSON body it cannot parse. Answer with a SCEP FAILURE instead, so
+            // the endpoint always responds in application/x-pki-message. The failInfoText is generic on
+            // purpose: internal exception messages are not shaped for the wire.
+            logger.error("Unexpected failure while processing SCEP PKIOperation: transactionId={}",
+                    scepRequest.getTransactionId(), e);
+            return buildResponse(scepRequest, buildFailedResponse(
+                    new ScepException("SCEP request processing failed", FailInfo.BAD_REQUEST),
+                    scepRequest.getTransactionId()));
+        }
+    }
+
+    private ResponseEntity<Object> processPkiOperation(ScepRequest scepRequest) throws ScepException {
         ScepResponse scepResponse;
         IntuneScepServiceClient intuneClient = null;
-
-        scepRequest = new ScepRequest(body);
 
         logger.debug("Processing SCEP request: transactionId={}", scepRequest.getTransactionId());
 
@@ -363,13 +376,14 @@ public class ScepServiceImpl implements ScepExternalService {
             intuneClient = buildIntuneClient(properties);
         }
 
-        // validate challenge password, if configured
         if (scepRequest.getMessageType().equals(MessageType.PKCS_REQ) || scepRequest.getMessageType().equals(MessageType.RENEWAL_REQ)) {
-            if (!validateScepChallengePassword(scepRequest.getChallengePassword())) {
-                return buildResponse(scepRequest, buildFailedResponse(new ScepException("Challenge password validation failed.", FailInfo.BAD_MESSAGE_CHECK), scepRequest.getTransactionId()));
-            }
-            // validate the request POP
             try {
+                // Classify the request before enforcing the shared secret: a renewal authenticates with the
+                // key of the certificate it replaces and legitimately carries no challenge password.
+                boolean authenticatedRenewal = authenticateRenewal(scepRequest);
+                // validate challenge password, if configured
+                validateChallengePassword(scepRequest.getChallengePassword(), authenticatedRenewal);
+                // validate the request POP
                 verifyRequest(scepRequest);
             } catch (ScepException e) {
                 return buildResponse(scepRequest, buildFailedResponse(e, scepRequest.getTransactionId()));
@@ -763,11 +777,32 @@ public class ScepServiceImpl implements ScepExternalService {
         return scepTransactionRepository.findByTransactionId(transactionId).orElse(null);
     }
 
-    private boolean validateScepChallengePassword(String challengePassword) {
-        if (scepProfile.getChallengePassword() == null || scepProfile.getChallengePassword().isEmpty()) {
-            return true;
+    /**
+     * Enforces the challenge password configured on the SCEP profile. A renewal request is signed with
+     * the key of the certificate being replaced and carries no challengePassword attribute
+     * (RFC 8894 §3.3.1.2), so an absent password is accepted once {@link #authenticateRenewal} has
+     * proven possession of that certificate's key. The waiver covers only an absent password: one that
+     * is present must still match, so a wrong shared secret is never silently accepted.
+     *
+     * @throws ScepException with {@link FailInfo#BAD_MESSAGE_CHECK} when the password is missing where
+     *                       required, or does not match the one configured on the profile
+     */
+    void validateChallengePassword(String requestChallengePassword, boolean authenticatedRenewal) throws ScepException {
+        String profileChallengePassword = scepProfile.getChallengePassword();
+        if (profileChallengePassword == null || profileChallengePassword.isEmpty()) {
+            return;
         }
-        return challengePassword.equals(scepProfile.getChallengePassword());
+        if (requestChallengePassword == null) {
+            if (authenticatedRenewal) {
+                logger.debug("Challenge password waived for an authenticated renewal request");
+                return;
+            }
+            throw new ScepException("The SCEP profile requires a challenge password but the request does not contain one.",
+                    FailInfo.BAD_MESSAGE_CHECK);
+        }
+        if (!profileChallengePassword.equals(requestChallengePassword)) {
+            throw new ScepException("Challenge password validation failed.", FailInfo.BAD_MESSAGE_CHECK);
+        }
     }
 
     public void verifyRequest(ScepRequest scepRequest) throws ScepException {
@@ -777,12 +812,9 @@ public class ScepServiceImpl implements ScepExternalService {
             throw new ScepException("Unsupported Operation", FailInfo.BAD_REQUEST);
         }
 
+        // The renewal classification and its validation run in authenticateRenewal, before the challenge
+        // password gate, because a renewal is authenticated by the signer certificate instead of the password.
         if (scepRequest.getMessageType().equals(MessageType.PKCS_REQ)) {
-            // Renewal check must be done for PKCS Request also. According to the RFC Version 3.1.1.2
-            // (https://datatracker.ietf.org/doc/id/draft-nourse-scep-23.txt), RENEWAL_REQ is not part of the message type
-            // Commonly used SCEP clients like JSCEP and SSCEP uses this version of RFC and
-            // may use PKCS_REQ for renewal
-            renewalValidation(scepRequest);
             try {
                 if (!scepRequest.verifyRequest()) {
                     throw new ScepException("Failed to verify PKCS#10 request POP, invalid signature", FailInfo.BAD_REQUEST);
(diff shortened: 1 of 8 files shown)


src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -354,23 +408,27 @@ private ResponseEntity<Object> pkiOperation(byte[] body) throws ScepException {
...
-        } catch (CMSException e) {
-            return buildResponse(scepRequest, buildFailedResponse(new ScepException("Unable to decrypt the data. " + e.getMessage(), FailInfo.BAD_REQUEST), scepRequest.getTransactionId()));
+        } catch (ScepException | CMSException e) {
+            // ScepException covers the undecryptable cases decryptData rejects itself (e.g. a password
+            // recipient with no challenge password configured); answering here keeps the response in the
+            // SCEP format instead of letting it reach the generic JSON error handler.
+            ScepException failure = e instanceof ScepException scepException
+                    ? scepException
+                    : new ScepException("Unable to decrypt the data. " + e.getMessage(), FailInfo.BAD_REQUEST);
+            return buildResponse(scepRequest, buildFailedResponse(failure, scepRequest.getTransactionId()));
```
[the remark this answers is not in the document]
lubomirw: Agreed — and it contradicted the policy stated a few lines up in the runtime-failure guard. Fixed in 11205872d: the CMS detail goes to `logger.error` with the transaction id, and the wire gets a generic "Unable to decrypt the request data" with `badRequest`. The `ScepException` arm needed no change; every message `decryptData` raises is already generic.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -763,69 +822,178 @@ private ScepTransaction getTransaction(String transactionId) {
...
+                    .formatted(renewedCertificate.getUuid(), renewedCertificate.getState(), renewedCertificate.getValidationStatus()),
+                    FailInfo.BAD_REQUEST);
+        }
+        if (!(new X500Name(renewedCertificate.getSubjectDn())).equals(scepRequest.getPkcs10Request().getSubject())) {
+            throw new ScepException("Subject DN for the renewal request does not match the original certificate",
+                    FailInfo.BAD_REQUEST);
         }
         // No need to verify the same key pair used in request since it is already handled by the rekey method in client operations
-        checkRenewalTimeframe(extCertificate);
+        checkRenewalTimeframe(renewedCertificate);
```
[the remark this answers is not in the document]
lubomirw: Correct, and the branch was also unreachable. `validateRenewal` now rejects `validationStatus` in {REVOKED, EXPIRED, INVALID, FAILED} and any `state != ISSUED` before `checkRenewalTimeframe` runs, so `else if (validationStatus == EXPIRED || state == REVOKED)` could never fire — while still dereferencing the nullable status to evaluate itself.

Fixed in 11205872d by removing that branch rather than null-guarding it, since the state gate covers revoked and expired for every threshold configuration (which was the point of adding it). Its Javadoc now records the precondition callers must establish. Regression test added: `certificateWithoutValidationStatus_isAnAuthenticatedRenewal` drives a configured threshold with a null validation status.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -763,71 +837,187 @@ private ScepTransaction getTransaction(String transactionId) {
...
+     * content is needed. Names that cannot be read withhold the waiver rather than guess.
+     */
+    private boolean requestedNamesAreAlreadyHeld(ScepRequest scepRequest) {
         try {
-            extCertificate = certificateService.getCertificateEntityByFingerprint(CertificateUtil.getThumbprint(scepRequest.getSignerCertificate()));
+            Map<String, List<String>> heldNames = CertificateUtil.getSAN(scepRequest.getSignerCertificate());
+            Map<String, List<String>> requestedNames = new Pkcs10CertificateRequest(
+                    scepRequest.getPkcs10Request().getEncoded()).getSubjectAlternativeNames();
+            for (Map.Entry<String, List<String>> requested : requestedNames.entrySet()) {
+                List<String> held = heldNames.getOrDefault(requested.getKey(), List.of());
```
klaraf755: **IP-address SANs never match here, so the waiver is silently withheld for them.**

The two sides of this comparison render iPAddress values differently:

- Certificate side: JCA's `X509Certificate.getSubjectAlternativeNames()` returns the IP already decoded to a string (`"192.168.0.1"`), which `getSanValueString` passes through.
- Request side: BC's `GeneralName.getName()` for iPAddress is a `DEROctetString`, whose `toString()` is `#c0a80001`-style hex.

So a renewal CSR carrying the same IP SAN as the certificate it renews never satisfies `held.containsAll(requested)` — the waiver is withheld, and since renewal clients by definition don't send the challenge password, they're rejected. That's the #1887 scenario persisting (as a clean SCEP failure now, not a 500) for network-device certificates, which are a core SCEP population. `dNSName`/`rfc822Name`/`URI` are `DERIA5String` on the BC side and match; iPAddress (and possibly otherName, where JCA hands back DER `byte[]`) do not.

It fails closed, so no security issue — but I'd add a test with an IP SAN and normalize the comparison (compare `GeneralNames` structurally from both sides, or decode octet strings to canonical text) before merge. The existing `requestAskingForHeldSans_isAnAuthenticatedRenewal` only covers a DNS name.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -763,71 +837,187 @@ private ScepTransaction getTransaction(String transactionId) {
...
+        // test only on the branch where a renewal threshold is configured, so enforce it for every
+        // configuration here, before the timeframe policy runs.
+        if (renewedCertificate.getState() != CertificateState.ISSUED
+                || NON_RENEWABLE_VALIDATION_STATUSES.contains(renewedCertificate.getValidationStatus())) {
+            throw new ScepException("Cannot renew certificate with UUID %s: it is not in a renewable state (state=%s, validation status=%s)"
+                    .formatted(renewedCertificate.getUuid(), renewedCertificate.getState(), renewedCertificate.getValidationStatus()),
+                    FailInfo.BAD_REQUEST);
+        }
+        if (!(new X500Name(renewedCertificate.getSubjectDn())).equals(scepRequest.getPkcs10Request().getSubject())) {
+            throw new ScepException("Subject DN for the renewal request does not match the original certificate",
```
klaraf755, line 1061: **PR body and code disagree on subject-DN mismatch: withhold vs. reject.**

The PR description says the "last three" waiver conditions — RA-profile association, **subject DN matching**, SANs — *withhold the waiver rather than rejecting the renewal, so a client that does supply the password is unaffected*. But this check throws, so a client that proves possession of an inventory certificate yet requests a different subject is rejected even with a correct challenge password.

That preserves the pre-existing behavior, but it contradicts the stated design. Worth an explicit decision: if a password-bearing client should be able to enroll for a new identity while signing with its old certificate, this check belongs with the withhold conditions in `authenticateRenewal`; if not, the PR body (and any spec doc) should be corrected to list DN mismatch as a rejection.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -100,6 +104,14 @@ public class ScepServiceImpl implements ScepExternalService {
             "Renewal",
             "SCEPStandard"
     );
+    // Validation statuses that make a certificate unusable, and therefore not renewable. EnumSet is used
+    // deliberately: unlike Set.of it tolerates a null argument to contains().
+    private static final Set<CertificateValidationStatus> NON_RENEWABLE_VALIDATION_STATUSES = EnumSet.of(
```
klaraf755, line 116: **`FAILED` here hard-blocks renewal even with a correct challenge password.**

`FAILED` means the validation *process* failed (e.g. CRL/OCSP endpoint unreachable), not that the certificate is bad. Since the gate in `validateRenewal` throws regardless of the password, a transient platform-side validation outage locks legitimate clients out of renewal entirely — and renewal is precisely the time-critical operation near expiry. Meanwhile `NOT_CHECKED` and `INACTIVE` pass, so the list doesn't follow a single principle.

Suggestion: let `REVOKED`/`EXPIRED`/`INVALID` reject, but have `FAILED` merely withhold the waiver (fall back to requiring the password). At minimum, a comment stating the intended principle would help the next reader.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -324,14 +336,91 @@ private ResponseEntity<Object> getResponseEntity(Object body, String contentType
...
-        ScepRequest scepRequest;
-        ScepResponse scepResponse;
-        IntuneScepServiceClient intuneClient = null;
+        ScepRequest scepRequest = null;
+        try {
+            // Parsing the request belongs inside the guard: the body is attacker-supplied DER and
+            // BouncyCastle reports malformed structures with unchecked exceptions.
+            scepRequest = new ScepRequest(body);
+            return processPkiOperation(scepRequest);
+        } catch (RuntimeException e) {
```
klaraf755: **Checked `ScepException` from `buildResponse` still escapes to the JSON handler.**

This guard catches `RuntimeException` only, but `buildResponse` throws checked `ScepException` (signing failure, token-connector trouble, `getSignedResponseData` I/O) and every `processPkiOperation` return path goes through it — those still propagate to the generic handler as the JSON 500 this PR set out to eliminate.

The inner recovery already handles "can't build a failure response" with a bare 500, so broadening this to `catch (Exception e)` completes the contract at no cost. (The `init`/`validateProfile` ScepExceptions — profile not found, disabled — would still answer as JSON; if that's deliberate, a note in the PR body would make it explicit.)

klaraf755 (commented):
Reviewed the full diff — the classification-before-enforcement design, the signature-before-state ordering, and the test coverage all look solid. Four inline findings below: one functional gap I'd resolve before merge (IP-address SANs), two decisions to make explicit (DN mismatch withhold-vs-reject, `FAILED` validation status), and one cheap completeness win on the error boundary.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
lubomirw: Confirmed and fixed in bc0e8671e — thank you, this was a real gap I introduced.

Verified the mechanism before fixing: both sides land in `CertificateUtil.getSanValueString`, which falls through to `value.toString()`. The certificate side arrives as JCA's already-decoded `String` ("192.168.0.1"), the request side as BC's `DEROctetString` ("#c0a80001"). Wrote the test first and it failed as you predicted: `expected: <true> but was: <false>` for an identical IP SAN on both sides.

Rather than normalizing the strings, `requestedNamesAreAlreadyHeld` now compares `GeneralName` DER structures read from the certificate's SAN extension and the request's extensionRequest attribute. That removes the whole class of rendering mismatch, including the `otherName` case you flagged. Two tests added: matching IP SAN earns the waiver, differing IP SAN withholds it.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
lubomirw: Fair catch — the PR description was wrong, not the code. Copilot flagged the same contradiction twice at low confidence.

Decision: DN mismatch keeps rejecting. Signing with one certificate and asking for a different subject is a malformed renewal rather than an enrollment the shared secret should authorize, and rejecting preserves the pre-existing behaviour, which I did not want to loosen inside a bugfix.

So I corrected the PR body instead: it now separates the three withhold conditions (RA-profile association, conclusive validation, SAN subset) from the rejections (DN mismatch, archived/pending/revoked/expired/invalid, outside the renewal window). Added a code comment at the check recording the reasoning. bc0e8671e.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
lubomirw: Agreed, and fixed in bc0e8671e. The list did conflate two different things.

`NON_RENEWABLE_VALIDATION_STATUSES` is now REVOKED / EXPIRED / INVALID and still rejects. FAILED moved to a separate `INCONCLUSIVE_VALIDATION_STATUSES` that only withholds the waiver, so the shared secret is required rather than the renewal being refused. Both sets carry the principle: reject only on evidence the certificate itself is unusable; an unreachable CRL or OCSP responder is not that.

Worth noting for the record that this does not rescue a renewal client during an outage — it sends no password, so it is still refused — but it stops a platform-side validation failure from locking out password-bearing clients, and it is the coherent rule. NOT_CHECKED and INACTIVE continue to pass. Test added for the withhold.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
lubomirw: Right, and broadened to `catch (Exception e)` in bc0e8671e.

One carve-out: when the message never parsed, `scepRequest` is null, so there is no transaction id or nonce to echo and no SCEP response can be formed — attempting one would end at the inner bare-500 fallback anyway. That case now surfaces as a protocol-level `ScepException`, which also keeps `ScepServiceITest`'s existing assertion that a garbage PKIOperation body throws.

The `init` / `validateProfile` behaviour is deliberate and was undocumented — you were right to ask. Those answer as JSON for the same reason: nothing has been parsed. Both points are now stated in the PR body under known residuals, alongside the global-rollback-only case, which I reproduced while writing tests (an issuance failure inside `generateCsr` triggers it).

klaraf755 (approved):

Issue #1900, opened 2026-07-28
Tighten the discovery failure-reason contract and place its types by layer


lubomirw:
### Description

Two follow-ups from the review of #1897, both deliberately left out of that change.

**The shaped-reason contract is too permissive.** `DiscoveryFailureReason.shape` passes through the message of any `ValidationException` on the assumption that the platform authors those messages. That assumption does not hold across all reachable callers: `CryptographicKeyServiceImpl` throws `ValidationException("Failed to calculate fingerprint from key content: " + keyItem.getKeyData())`, which is reachable from discovery key association and therefore lands in `processedError` — a field returned to API clients. The leaked value is a public key rather than a secret, so the impact is low, but it is exactly the runtime detail the repo's "don't leak runtime details to the wire" rule exists to keep off the wire.

The fix is a decision rather than a patch, which is why it is deferred: pass-through should be gated on an explicit shaped-reason marker type instead of on `ValidationException` as a class. `DiscoveryImportRollbackException` already plays that role for one path and could generalise. Changing the gate affects every current caller, so it wants its own review.

**Discovery types sit on the wrong side of a layer boundary.** `CertificateInternalService` returns `DiscoveredCertificateImport` and `CertificateHandler` takes `DiscoverySource`, both from `com.otilm.core.events.handlers.discovery`. The events package consumes services, so a service interface's own types should not live inside it. Moving them to a service-visible model package resolves it with no behaviour change.

While moving them: `DiscoveredCertificateImport` wraps a single `Certificate` and carries nothing the bare entity does not, and `DiscoveryCertificateContentWriter.insertCertificate`'s documented 1/0 return is discarded by its only caller. Either drop the wrapper or let it carry the one fact worth having — whether the row was inserted or adopted from a concurrent import.

### Definition of Done

- Reason pass-through is gated on an explicit marker type, and no `ValidationException` message reaches `processedError` by virtue of its class alone.
- A test covers the key-content case specifically, asserting the raw key data does not appear in the persisted reason.
- `DiscoverySource` and `DiscoveredCertificateImport` live outside `events.handlers`, with the wrapper either removed or carrying the inserted-vs-adopted flag.

lubomirw:
The leak half of this is closed in #1906 — Copilot raised it on the PR independently, which made three reviewers on the same throw site, and that settled the decision.

[...]

Issue #1899, opened 2026-07-28
Attribute content-item registration is an unguarded find-then-insert


lubomirw:
### Description

`AttributeEngine.registerAttributeContentItems` resolves each content item with `findByJsonAndAttributeDefinitionUuid` and inserts when the lookup returns null, with nothing guarding the gap between the two. This is the same check-then-insert shape as the discovery-import defect fixed in #1897, one layer down.

It is reachable concurrently: discovery post-processing runs several certificate imports in parallel, each applying per-row metadata through the attribute engine, so two of them can register the same not-yet-registered content item at once. The window is narrow because the discovery download phase pre-registers most content items, but it is not closed.

The outcome depends on whether the table carries a unique constraint on `(json, attribute_definition_uuid)`:

- With a constraint, the loser raises a unique violation, which marks its surrounding transaction rollback-only and surfaces as a generic "the import transaction was rolled back" — losing the certificate and the reason, exactly the failure mode #1897 exists to remove.
- Without one, both transactions succeed and the definition ends up with duplicate content items.

Confirming which applies is the first step.

This is pre-existing rather than introduced by #1897; that work only made the pattern easier to see.

### Severity

Minor

### Steps to Reproduce

1. Run a discovery whose certificates carry a metadata attribute value not yet registered for its definition.
2. Ensure several discovered certificates share that value, so parallel imports register it concurrently.
3. Observe either a duplicate `attribute_content_item` row for one definition, or a discovery certificate rolled back with a generic transaction-rollback reason.

### Expected

Concurrent registration of the same content item resolves to one row, and no caller's transaction is poisoned by losing the race.

### Actual

The gap between the lookup and the insert is unguarded, so concurrent callers either duplicate the row or one of them poisons its own transaction.

### Environment

Reachable from discovery post-processing on any release carrying the parallel import path.

PR #1906, opened 2026-07-28
Import discovered certificates per content group and report what was lost

### LINKED ISSUE #1879: Discovery: thread-unsafe shared-entity progress save (parallelism loss; possible certificate loss under load)
### Description

Discovery loads the `DiscoveryHistory discovery` entity in a `@Transactional(NOT_SUPPORTED)` context (no ambient transaction), so the instance is **detached**, then reports progress by mutating and saving that **single shared detached instance from parallel worker threads**: `discovery.setMessage(...); discoveryRepository.save(discovery)`. Hibernate entities are not thread-safe, and the save participates in a per-unit `REQUIRES_NEW` transaction.

Two call sites share this pattern:

1. **Download** — `CertificateHandler.createDiscoveredCertificate` (`@Transactional(REQUIRES_NEW)`, line 148; save at ~line 178), fanned out from `DiscoveryServiceImpl.downloadDiscoveredCertificatesBatchAsync` across up to **10 concurrent batches** (`downloadCertSemaphore`, virtual threads). Every batch ends with a save of the shared `discovery`.
2. **Post-processing** — `CertificateDiscoveredEventHandler.processDiscoveredCertificate` (~lines 344-346), one `REQUIRES_NEW` tx per certificate across 5 parallel threads; every MAXIMUM_PARALLELISM-th certificate saves the shared `discovery`.

This is the root cause behind #1874's NPE symptom (orphaned certificate UUID → null certificate in key upload). Affects all discovery types (network, IP, CT-logs, cryptosense).

**Scope — this issue owns three linked work items** (the first two folded in from a review of PR #1880, which shipped only the minimal NPE guard):

1. **Decouple progress reporting** from the per-unit transaction and from the shared entity (the core defect above).
2. **WARNING-status accounting for lost/failed key uploads.** PR #1880 surfaces skipped/failed key association at the *discovery* level (WARNING) via a `keyAssociationIncomplete` flag, but the underlying lost-certificate accounting belongs here: when the rollback is fixed, the certificates that were dropped must count toward the final status rather than depending on `countByDiscoveryAndProcessedErrorNotNull` alone (whose `processedError` marking rolled back with them).
3. **Per-certificate `processedError` for key-association failures.** Today `setProcessedError` is written only when certificate-entity creation fails (`CertificateDiscoveredEventHandler` ~line 293); the post-parallel key-upload loops only flip the discovery-level flag. Record a per-certificate error so the UI shows *which* certificates lack a key association — for the skip case (certs lost to the rollback) via the rollback fix, and for the throw case (key upload fails after the certificate committed) via a Certificate-UUID → DiscoveryCertificate mapping the `keyToCertificates` map does not currently carry.

### Steps to Reproduce

1. Run a Network Discovery against host(s) presenting many certificates so both the download and post-processing phases fan out across parallel threads.
2. Drive concurrency on the shared discovery progress row (high certificate count, up to 10 concurrent download batches) so a concurrent `discoveryRepository.save(discovery)` on the shared detached instance participates in a per-unit `REQUIRES_NEW` transaction under contention.
3. Compare the connector-reported discovered count against the certificates actually persisted and the discovery-certificate `processed` / `processedError` bookkeeping.
4. Inspect the Core log for rolled-back per-unit transactions around the progress-save (see Logs).

### Expected Behavior

- Progress reporting is decoupled from the per-unit transaction and from the shared entity — e.g. a single `@Modifying @Query` updating only the message column by discovery UUID (`UPDATE DiscoveryHistory d SET d.message = :message, d.updated = CURRENT_TIMESTAMP WHERE d.uuid = :uuid`) on a Writer bean, reused by both call sites, invoked in its own short transaction **after** `runInNewTransaction(...)` returns so a `@Transactional(REQUIRED)` writer cannot rejoin and roll back the unit's committed work. `emitDiscoveryFinished`'s failure-path read of the shared `discovery` is folded in too, so no writer of the shared instance remains.
- A progress-reporting save conflict never rolls back committed certificate work; the parallel download and post-processing are not serialized on one row.
- Every discovered certificate is either persisted or recorded with a `processedError`; lost/failed key associations count toward the discovery WARNING status and carry a per-certificate error.

### Actual Behavior

- Concurrent `UPDATE discovery_history SET message=... WHERE uuid=?` on the same row serialize on the PostgreSQL row write-lock (last-write-wins), throttling the parallel hot path; the reported percentage can transiently regress or jump.
- One mutable Hibernate entity instance is `setMessage`-mutated and `merge`-read concurrently across sessions (`DiscoveryHistory` has no `@Version`; its `@OneToMany`/`@ManyToMany` associations are LAZY, no cascade — so no `OptimisticLockException` and no cascade `ConcurrentModificationException`, but the sharing is incorrect regardless).
- **Possible under load (unconfirmed):** if a per-unit transaction throws (DB-level conflict under load, or an unrelated failure in the same tx), the rollback discards committed work — an entire download batch, or a single post-processed certificate plus its `processed`/`processedError` bookkeeping — leaving no trace and orphaning its UUID in the shared key-upload map (the #1874 NPE path, now guarded). No deterministic throw from the shared save alone has been identified, so the data-loss tail is plausible under load, not a proven routine outcome.

### Severity

**Minor**, pending evidence. Certain impact is parallelism/cosmetic; the data-loss tail is unconfirmed. **Re-escalate to Major if** production logs show non-trivial frequency of the signatures below during discovery, or a load repro demonstrates certificate loss.

### Logs

Query the observability backend (OpenTelemetry/Loki — core logs to stdout, no file appender) for discovery-window occurrences and correlate hit rate with discovery certificate counts:

```
Downloading batch .* failed
Unable to process certificate
UnexpectedRollbackException
did not complete
```

Or reproduce under load: a large multi-batch discovery with instrumentation on the per-unit rollback path.

### LINKED ISSUE #1897: Discovery loses certificates to rolled-back per-certificate transactions and cannot report which
### Description

Discovery post-processing runs one `REQUIRES_NEW` transaction per discovered row. When a single discovery contains several rows carrying the same certificate — the same certificate presented on multiple hosts or ports — those rows race on an unguarded find-then-insert in `checkAddCertificateContent`, whose deferred INSERT violates the fingerprint constraint at a later auto-flush inside an inner `REQUIRED` method. That marks the physical transaction rollback-only; the broad catch swallows it and the unit keeps writing, so the commit throws `UnexpectedRollbackException`.

Everything the transaction held is discarded together: the certificate, its `processed` flag, and its `processedError`. The one field meant to explain the failure is destroyed by the failure it was meant to describe, which is why `countByDiscoveryAndProcessedErrorNotNull` reports zero and the run falls through to a message that names nothing.

The user-visible symptom is the status *"Discovery completed with warnings. Some discovered certificate keys could not be associated during discovery."* — one sentence covering several unrelated causes, with no counts and no way to tell which certificates are affected. Whether it appears at all is a coin flip on where the auto-flush landed relative to the in-memory key queueing; landing earlier loses the certificate **silently**.

Scope of the fix: import one certificate per content group so the race cannot occur; reuse the existing conflict-resolving inserts so a losing caller re-reads instead of failing; move `processed` / `processedError` and progress reporting out of the per-certificate transaction so the record survives a rollback; and derive discovery status from counted outcomes so both an inventory gap and a key gap are reported with counts and attributed per certificate.

Design and implementation plan: `docs/superpowers/2026-07-27-discovery-import-integrity/` in the planning workspace.

This delivers items 1, 2 and 3 of #1879, which should be closed by it. It also replaces the status decision and message introduced by #1880 — that change was reporting-only and introduced no regression; it is superseded here rather than reverted, and its per-entry isolation of the key-upload loops is retained.

### Steps to Reproduce

1. Run a discovery against hosts presenting the same certificate several times, so post-processing fans out across parallel threads holding one certificate content.
2. Watch the core log during post-processing for `UnexpectedRollbackException` from `CertificateDiscoveredEventHandler`, and for `No committed certificate found for key with fingerprint`.
3. Compare the connector-reported discovered count against the certificates actually persisted.
4. Inspect the affected `discovery_certificate` rows.

Retrospective check for past runs:

```sql
SELECT dc.discovery_uuid, count(*) AS lost
FROM core.discovery_certificate dc
LEFT JOIN core.certificate c ON c.certificate_content_id = dc.certificate_content_id
WHERE dc.newly_discovered
  AND NOT dc.processed
  AND dc.processed_error IS NULL
  AND c.uuid IS NULL
GROUP BY 1
ORDER BY 2 DESC;
```

This also matches runs that crashed mid-processing, so cross-check against discovery end times.

### Expected Behavior

Every discovered certificate is either imported or recorded with a reason that survives the failure. The discovery status names what failed and how many, reports an inventory gap and a key gap together rather than letting one mask the other, and every affected row carries a per-certificate reason so the certificate list shows which ones.

### Actual Behavior

Affected certificates are absent from the inventory. `processed` stays false and `processedError` is null, because both writes rolled back with the certificate. The run reports either a clean completion or a warning that names no certificate, carries no count, and describes several distinct causes with one sentence.

### Severity

Major

### Environment

ILM Core 2.19.0-SNAPSHOT, PostgreSQL

lubomirw:
Closes #1897
Closes #1879

## The defect

A certificate discovered on several hosts produced several `discovery_certificate` rows sharing one `certificate_content`. Each was imported independently through an unguarded find-then-insert, so the copies raced. The losing transactions violated the fingerprint constraint at a later auto-flush inside an inner `REQUIRED` method, which marked them rollback-only; a broad catch swallowed that, the unit kept writing, and the commit threw. The certificate, its `processed` flag and its `processedError` were all discarded together — so the run knew something had gone wrong but had thrown away the evidence of what.

That is why every affected discovery ended on *"Discovery completed with warnings. Some discovered certificate keys could not be associated during discovery."* — a message naming no certificate and no count, because by the time it was emitted there was nothing left to name.

## The fix

Rows are grouped by `certificate_content_id`, so one certificate is one group is one transaction and two threads can no longer hold the same certificate. Each group imports through the existing conflict-resolving inserts and always re-reads the surviving row, so a group that loses a race against another *discovery* adopts the committed certificate instead of failing. Results are consumed on the orchestrator thread, bookkeeping is written in isolated transactions that survive the rollback of the work they describe, and the final status is derived from four counts rather than a single flag.

Reasons reaching `processedError` are classified rather than forwarded — that field is returned to API clients, so a raw `getMessage()` would put SQL fragments and column names on the wire.

## Outcome taxonomy

Every discovered row now resolves to exactly one outcome, and each contributes its own sentence to the status so two simultaneous failures are both visible:

| Outcome | `processed` | `processedError` | Counts toward |
|---|---|---|---|
| `IMPORTED` | true | null | — |
| `IGNORED` | true | null | — |
| `ENTITY_CREATION_FAILED` | true | shaped reason | inventory gap |
| `IMPORT_ROLLED_BACK` | true | shaped reason | inventory gap |
| `KEY_ASSOCIATION_FAILED` | true | shaped reason | key gap |
| `NOT_ATTEMPTED` | false | shaped reason | unattempted |

A clean run reports `PROCESSING`, which the finish handler rolls up to `COMPLETED`. Anything else reports `WARNING` and says how many certificates and why.

## Also fixed

- **Trigger scope.** Action-trigger failures are contained per trigger. Left to propagate they roll the group back, and since every group evaluates the same trigger set, one misconfigured trigger would lose every certificate in the discovery. Ignore-trigger failures still stop the import — there the ignore decision is unknown.
- **Ignore triggers see the certificate.** The in-memory candidate is stamped with fingerprint and content id. Without them a rule conditioned on the fingerprint read null, the evaluator recorded a failed condition and swallowed it, and the certificate was imported despite the rule.
- **Progress reporting cannot destroy the work it reports on** — neither the write nor a failure of it reaches the import (#1879 item 1).
- **Audit author** is stamped on discovered certificates, which the native insert otherwise skips along with `@PrePersist`.
- **Event histories already `FINISHED` are never re-marked `FAILED`**, so a messaging failure after the import committed no longer points the record at work that succeeded.

## Behaviour change worth noting

Grouping makes trigger evaluation **per unique certificate** rather than per discovered row. A certificate found on ten hosts now yields one evaluation and one trigger-history entry instead of ten, recorded against one representative row. This follows unavoidably from grouping and is arguably the better semantics, but it is user-visible and wants a release note. Tracked in OmniTrustILM/documentation#336 along with the status-table and `processedError` documentation.

## Verification

Full `mvn verify` on the pre-merge tree: 4,215 tests, 0 failures, 0 errors. Since then, 38 unit tests and 21 integration tests covering the changed paths, re-run after every fix round. All architecture gates green; `ContextSignatureGuardTest` unchanged (the new test classes are mock-only, so no new Spring context).

New tests cover the duplicate-row group, a group whose import rolls back (asserting the exact reason, not merely a non-null one), a key association that fails, the losing side of the insert race, an ignore trigger conditioned on the fingerprint, and the containment of progress and bookkeeping failures.

## Review

Four independent reviewers over three rounds, plus an adversarial adjudication pass. Findings deferred deliberately: #1899 (an unguarded find-then-insert in attribute content-item registration — pre-existing, same shape one layer down) and #1900 (the reason-shaping contract, and moving the discovery records out of `events.handlers`).

One known cost, unmeasured: bookkeeping runs as one isolated transaction per group on the consumer thread, so a large discovery pays that serially. It is the price of consuming on one thread without synchronisation, and it would disappear entirely under a per-certificate event model — a separate design question, not a change here.


src/main/java/com/otilm/core/events/handlers/discovery/DiscoveryFailureReason.java
```diff
@@ -0,0 +1,103 @@
...
+            String classified = classify(cause);
+            if (classified != null) {
+                return classified;
+            }
+        }
+        // Blank-checked, not just null-checked: ValidationException stringifies a null message, so a
+        // message-less one yields the literal "null", which would read as a bug in the certificate list.
+        if (throwable instanceof ValidationException && isUsable(throwable.getMessage())) {
+            return throwable.getMessage();
+        }
```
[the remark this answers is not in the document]
lubomirw: Fixed in be8c471f1. Only `DiscoveryImportRollbackException` is forwarded now — the one type whose message this class shaped itself. Everything else is classified, `ValidationException` included.

Worth noting the marker you suggested does not work here: `ValidationException implements PlatformException`, as do twenty other types, so gating on `PlatformException.safeMessage` would have widened the pass-through rather than narrowed it. It marks a message the platform authored, not one free of entity data — and the `CryptographicKeyServiceImpl` site you cited is itself a `ValidationException`.

src/test/java/com/otilm/core/events/handlers/discovery/DiscoveryFailureReasonTest.java
```diff
@@ -0,0 +1,153 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+class DiscoveryFailureReasonTest {
+
+    @Test
+    void passesThroughAControlledDomainMessage() {
+        assertThat(DiscoveryFailureReason.shape(
+                new ValidationException("the certificate has no subject alternative names")))
+                .isEqualTo("the certificate has no subject alternative names");
+    }
```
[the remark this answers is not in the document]
lubomirw: Updated in be8c471f1. The test now asserts the tightened contract, using your `keyItem.getKeyData()` example as the input and asserting it is classified rather than forwarded.

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
```diff
@@ -165,215 +190,512 @@ public void handleEvent(EventMessage eventMessage) throws EventException {
...
+        group.rows().forEach(row -> certificateHandler.updateDiscoveredCertificate(DiscoverySource.of(context), certificate, row.getMeta()));
 
-        if (logger.isDebugEnabled()) {
-            logger.debug("Finalize processing discovered certificate: {}", certificate.toStringShort());
+        for (TriggerAssociation triggerAssociation : context.triggers()) {
+            EventHistory eventHistory = triggerAssociation.getResource() == null ? platformEventHistory : discoveryEventHistory;
+            try {
+                context.eventContext().getTriggerEvaluator().evaluateTrigger(triggerAssociation.getTrigger(), triggerAssociation,
+                        certificate, referenceRowUuid, eventData, eventHistory);
+            } catch (RuleException e) {
```
vyskocilm, line 696: **Must-fix: containment gap — unchecked exceptions from trigger evaluation roll back the whole group, deterministically for every group.**

`catch (RuleException)` here (and at line 569 for ignore triggers) doesn't cover unchecked exceptions, and there is a concrete, operator-reachable one: any PROPERTY rule on RA Profile (including plain `RA Profile EMPTY`) resolves the path `raProfile.name` via `PropertyUtils.getProperty`, which throws commons-beanutils `NestedNullException` (extends `IllegalArgumentException` — verified against beanutils 1.11.0) because every freshly discovered certificate has `raProfile == null`. Value retrieval happens before the operator is applied, so even the EMPTY check throws. It escapes both `TriggerEvaluator` catches (217-223 covers only checked reflection exceptions; 468-481 only `RuleException`).

The subtle part: **widening this catch would not fix it.** `TriggerEvaluator`/`CertificateTriggerEvaluator` are class-level `@Transactional`, so the evaluator call crosses a Spring proxy that joins the group transaction — the moment the unchecked exception crosses that boundary, the tx is marked rollback-only (`globalRollbackOnParticipationFailure=true` default, verified against spring-tx 6.2.19), and the commit throws `UnexpectedRollbackException` regardless of what the handler catches. Since every group evaluates the same trigger list, one such rule rolls back every group: the discovery imports zero certificates.

(For the record, `origin/main` lost everything on this path too via the same rollback-only mechanics — but it reported COMPLETED/clean, whereas this PR at least reports the loss honestly. The comment above this catch promises containment it can't deliver.)

**Fix:** contain inside `TriggerEvaluator` — in `evaluatePropertyConditionItem` (or `getConditionEvaluationResult`), catch `NestedNullException`/unchecked exceptions and convert to `RuleException` / a failed-condition record, so nothing unchecked exits the evaluator proxy.

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
```diff
@@ -165,215 +190,512 @@ public void handleEvent(EventMessage eventMessage) throws EventException {
...
-                if (triggerHistory.isActionsPerformed()) {
-                    isIgnored = true;
-                    break;
-                }
+        EventHistory discoveryEventHistory = eventHistoryRepository.getReferenceById(context.discoveryEventHistoryUuid());
+        EventHistory platformEventHistory = eventHistoryRepository.getReferenceById(context.platformEventHistoryUuid());
+        UUID referenceRowUuid = rowUuids.getFirst();
+
+        List<TriggerHistory> ignoreHistories = new ArrayList<>();
+        for (TriggerAssociation triggerAssociation : context.ignoreTriggers()) {
```
vyskocilm, line 644: **Must-fix: adopt-existing path bypasses operator ignore rules conditioned on CUSTOM/META attributes.**

Ignore triggers are evaluated against the bare in-memory `candidate` whose uuid is null (uuid is generated only at `@PrePersist`), and the 6-arg evaluator overload passes `pendingCustomAttributes = null` — so `TriggerEvaluator` (180-189) returns `false` for every CUSTOM and META condition. When the certificate already exists in the inventory (it entered between the download-phase `newlyDiscovered` snapshot and this group's import: concurrent discovery of an overlapping estate, upload, issuance), `createDiscoveredCertificateAtomic` adopts the existing row anyway — which then receives discovery metadata, DISCOVERY event history, and all action triggers despite a matching ignore rule.

On `origin/main`, `createCertificateEntity` early-returned the persisted entity before ignore evaluation, so rules like "ignore where custom attribute environment=test" matched. This is a regression on the adopt path specifically (genuinely new certificates had the same uuid-null blind spot on main).

**Fix:** evaluate ignore triggers against `certificateRepository.findByFingerprint(...).orElse(candidate)`; for the insert race, carry the writer's existing 0/1 insert result through `DiscoveredCertificateImport` and re-evaluate against the adopted survivor before metadata/action side effects.

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
```diff
@@ -165,215 +190,512 @@ public void handleEvent(EventMessage eventMessage) throws EventException {
...
+     * reports a missing certificate that is sitting in the inventory.
+     */
+    void accountForUnconsumedGroups(DiscoveryRunAccumulator accumulator,
+                                    List<DiscoveryContentGroup> groups, Set<Long> consumedContentIds) {
+        for (DiscoveryContentGroup group : groups) {
+            if (consumedContentIds.contains(group.certificateContentId())) {
+                continue;
+            }
+            List<UUID> rowUuids = group.rows().stream().map(DiscoveryCertificate::getUuid).toList();
+            boolean committed = certificateRepository.existsByCertificateContentId(group.certificateContentId());
```
vyskocilm: **Should-fix: this probe is the only uncontained phase between the consumption loop and `associateKeys`.**

A transient `DataAccessException` from `existsByCertificateContentId` exits `importAndReport` before the key-association calls: the in-memory key maps are discarded and `Certificate.keyUuid` is never set for any consumed group — whose rows were already marked `processed=true`/`processedError=null` by `writeBookkeeping` during the loop. Result: a whole-run silent key gap with no per-row record. Every sibling phase (`importGroupSafely`, `writeBookkeeping`, `associateKeys`, `reportProgressSafely`, validation send) is individually contained; this one should be too.

**Fix:** catch failures per unconsumed group, record a bookkeeping failure, and continue so consumed key maps always get associated. `handleEvent` runs `NOT_SUPPORTED`, so the catch isn't undermined by an ambient rollback-only transaction.

vyskocilm (commented):
Independent review pass (multi-model, findings adversarially re-verified against dependency bytecode). Three line comments below cover the items worth addressing in this PR (two must-fix, one cheap should-fix). The rest are deferrable — listed here so they don't get lost, none of them blocks merge:

**Deferrable / follow-up candidates**

- **Batched `markProcessed`/`updateProcessedError` UUID lists are unchunked** (`DiscoveryCertificateRepository`): a single reason-group above the pgjdbc extended-mode cap of 65,535 bind parameters fails wholesale. The failure is contained and reported as a bookkeeping warning, so impact is bounded; chunking the UUID lists is a cheap hardening.
- **Validation-request failure is mislabeled**: a broker failure at `validationProducer.produceMessage` is recorded as a bookkeeping failure, so the WARNING says "per-certificate detail could not be recorded" while the real consequence — validation never queued for any imported certificate — is only in the log. Deserves its own counter/sentence.
- **`reportDownloadProgress` (REQUIRES_NEW inside the open batch tx)** transiently needs a second pooled connection per concurrent batch. The tradeoff is documented in the code; cheapest improvement is hoisting the progress call to after the batch method returns.
- **Authorization denial attribution**: OPA denial now aborts before any row is touched, leaving rows `processed=false` with no `processedError` and a generic WARNING that never names authorization. Consider naming authorization in the message or writing a NOT_ATTEMPTED reason.
- **Unconsumed-group accounting labels** (`accountForUnconsumedGroups`): after an abnormal early exit, `existsByCertificateContentId` can't distinguish this run's import from another actor's, nor detect a committed IGNORED verdict — conservative mislabeling in a rare path, worth a Javadoc correction at minimum.
- **Interrupt-flag hygiene**: after an interrupt-driven consumption break, the epilogue runs with the flag set. Verified this does not strand the run (Qpid clears the flag on its first sync and the producer RetryTemplate recovers), but capture-and-clear at the break with re-assert at method exit is cleaner.
- **`createDiscoveredCertificateAtomic` duplicates `createCertificateAtomic`'s insert-or-adopt skeleton**, and the copies already diverged: the new one stamps `author`, the sibling still inserts `i_author` NULL (pre-existing defect, now fixed in only one copy — deserves its own follow-up issue). The writer's documented 0/1 insert return is unused by its only caller.
- **`prepareIssuedCertificate` fingerprint stamping**: all 7 call sites hand-stamp fingerprint around the utility; folding it in would remove the footgun (counterpoint: `getThumbprint` throws checked exceptions, so it changes 7 signatures).
- **`GroupImportResult.committed` is derivable** from `rowResults` (all rows IMPORTED/IGNORED) at every construction site; deriving it makes inconsistent states unrepresentable.
- **`DiscoveryRunContext` duplicates `DiscoverySource` field-for-field** with a copy factory; composing would delete the converter.
- **Test nit**: `EventHandlersITest.generateSelfSignedCertificate` hand-rolls a BC generator while the same file already uses `CertificateGeneratorHelper.generateCACertificate`.

For context, two initially-reported severe findings were investigated and **refuted** during verification: (a) interrupt-driven stranding of the discovery in PROCESSING (Qpid's `InterruptedException` handler clears the flag before wrapping, and the producer retries on the clean thread); (b) post-commit `AFTER_COMMIT` notification failure being misclassified as a rollback (Spring dispatches these listeners from `afterCompletion`, whose invoker catches `Throwable` — the exception never reaches `runInNewTransaction`).

src/main/java/com/otilm/core/dao/repository/DiscoveryCertificateRepository.java
```diff
@@ -24,8 +28,20 @@ public interface DiscoveryCertificateRepository extends SecurityFilterRepository
...
     List<DiscoveryCertificate> findByCertificateContent(CertificateContent certificateContent);
 
+    /**
+     * Batched deliberately: every row of a content group shares one outcome and one reason, so a group collapses
+     * to a single statement rather than one transaction per row.
+     */
+    @Modifying
+    @Query("UPDATE DiscoveryCertificate dc SET dc.processed = true, dc.processedError = :processedError, " +
+            "dc.updated = CURRENT_TIMESTAMP WHERE dc.uuid IN :uuids")
+    void markProcessed(@Param("uuids") Collection<UUID> uuids, @Param("processedError") String processedError);
```
klaraf755, line 40: The `IN :uuids` here (and in `updateProcessedError` below) is unbounded: every row of a content group goes into one statement, one bind parameter per row. The PostgreSQL JDBC driver caps binds at 32,767 per statement, so a certificate discovered on more than ~32k hosts fails the bookkeeping write for its whole group. Unlikely, but a discovery against a large fleet with one shared wildcard cert is exactly the shape this PR handles. Chunking the collection (or passing an array + `unnest`) would close it cheaply — fine as a follow-up.

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
```diff
@@ -165,215 +190,512 @@ public void handleEvent(EventMessage eventMessage) throws EventException {
...
             try {
-                keyAssociationIncomplete |= !certificateHandler.uploadDiscoveredCertificateKey(entry.getKey(), entry.getValue());
+                if (entry.getKey().outcome() == DiscoveryCertificateOutcome.NOT_ATTEMPTED) {
+                    // processed stays false — the truthful record for a row the platform never reached — but the
+                    // reason is still written, because the status message points the operator at this list for it.
+                    transactionHandler.runInNewTransaction(() ->
+                            discoveryWriter.recordProcessedError(rowUuids, detail));
+                } else {
+                    transactionHandler.runInNewTransaction(() ->
+                            discoveryWriter.markProcessed(rowUuids, detail));
```
klaraf755: Small observability note: a committed group's rows are written `IMPORTED`/`processedError = null` here at consume time, then re-written to `KEY_ASSOCIATION_FAILED` with a reason if the key phase later fails. Between the two writes an API client sees a clean row for a certificate that will end up flagged. Harmless in practice since the discovery is still `IN_PROGRESS` at that point — just noting it's a deliberate transient state, in case anyone builds tooling that polls the row list mid-run.

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
```diff
@@ -165,215 +190,512 @@ public void handleEvent(EventMessage eventMessage) throws EventException {
...
+                    .formatted(counts.keyGaps()));
+        }
+        if (counts.notAttempted() > 0) {
+            sentences.add("%d certificate(s) could not be processed to a result."
+                    .formatted(counts.notAttempted()));
+        }
+        if (counts.bookkeepingFailures() > 0) {
+            sentences.add("Some per-certificate detail could not be recorded.");
+        }
+        sentences.add("See the discovery certificate list for per-certificate detail.");
```
klaraf755: Wording nit: when the *only* gap is a bookkeeping failure, the message reads "Some per-certificate detail could not be recorded. See the discovery certificate list for per-certificate detail." — i.e. it points the operator at the list whose detail just failed to write. Making this pointer sentence conditional on `inventoryGaps + keyGaps + notAttempted > 0` (or softening it to "may be incomplete" when only bookkeeping failed) would fix the contradiction.

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
```diff
@@ -165,215 +190,512 @@ public void handleEvent(EventMessage eventMessage) throws EventException {
...
+     */
+    GroupImportResult importGroupSafely(DiscoveryRunContext context, DiscoveryContentGroup group) {
+        List<UUID> rowUuids = group.rows().stream().map(DiscoveryCertificate::getUuid).toList();
+        try {
+            processCertSemaphore.acquire();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            return new GroupImportResult(group.certificateContentId(),
+                    resultsFor(rowUuids, DiscoveryCertificateOutcome.NOT_ATTEMPTED,
+                            "the import was interrupted before it began"), List.of(), false);
```
klaraf755, line 354: Capitalization nit: `processedError` texts are inconsistently cased — "the import was interrupted before it began" / "the certificate was imported, but…" (lowercase) vs "Import rolled back: …" / "Unable to create certificate entity: …" (capitalized). Cosmetic, but these strings are operator-facing and now load-bearing; worth picking one convention while the strings are fresh.

klaraf755 (commented):
Review findings posted inline — none blocking. Only the first one is worth a code change (can be a follow-up).

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
lubomirw: Fixed in 5aeff6d0f, contained inside `TriggerEvaluator` as you suggested — the nested-path failure becomes a `RuleException`, plus a guard on the condition boundary so nothing unchecked leaves the proxy.

Your point that widening the handler catch would not help was the decisive one; I would have tried that first. Verified the whole chain: `@Transactional` at `TriggerEvaluator:53`, the catch at `:220` covering only the three checked reflection exceptions, and `RA_PROFILE_NAME` declared nested through `Certificate_.raProfile`.

Added `testCertificateDiscoveredImportsDespiteAConditionOnAnAbsentAssociation`, and confirmed it earns its place by removing both guards and re-running: fails with "an unevaluable condition must not cost the certificate its import — expected: true but was: false". So the zero-import scenario reproduces from a plain `RA Profile EMPTY` rule.

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
lubomirw: Fixed in 5aeff6d0f: ignore triggers now evaluate against `certificateRepository.findByFingerprint(...).orElse(candidate)`.

Confirmed the mechanism — `TriggerEvaluator:180-189` gates CUSTOM on `objectUuid != null || pendingCustomAttributes != null` and META on `objectUuid != null` — and confirmed the regression by reading `createCertificateEntity:1147-1149`, which early-returned the persisted row. Genuinely new certificates keep the blind spot, exactly as you noted.

I did not carry the writer's 0/1 insert result through `DiscoveredCertificateImport` for the insert-race half; that is folded into #1900 with the rest of the wrapper work.

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
lubomirw: Fixed in 5aeff6d0f — contained per group, recording a bookkeeping failure and continuing, so the consumed key maps always reach `associateKeys`. Unit test added.

Also corrected the javadoc, which claimed more than the probe can know: as you note elsewhere, a present certificate may have been committed by another actor, and a group that committed an IGNORED verdict leaves none and so reads as never attempted.

lubomirw:
Thanks — this is the review that found the things the automated rounds didn't. All three line comments were verified against the code and fixed in `5aeff6d0f`; three of the deferrables in `c5ab09906`; the rest folded into #1900.

**Fixed in `5aeff6d0f`** (replies on the threads):

- Unchecked exceptions from trigger evaluation, contained inside `TriggerEvaluator`. Your observation that widening the handler's catch would not help was the decisive part — I'd have tried that first and shipped something that still lost every certificate. Regression test verified by removing the guards and watching it fail.
- Ignore triggers evaluated against the committed row on the adopt path.
- The unconsumed-group probe contained per group, and its javadoc corrected — it claimed more than the probe can know.

**Fixed in `c5ab09906`:**

- **Validation-request failure** now has its own flag and sentence. You were right that recording it as a bookkeeping failure was actively misleading: the operator read "per-certificate detail could not be recorded" while the real consequence was a whole run of unvalidated certificates.
- **`i_author` divergence** — good catch, and the worst state to leave it in. `createCertificateAtomic` stamps the author too. The duplicated insert-or-adopt skeleton itself is in #1900.
- **`reportDownloadProgress`** hoisted to after `createDiscoveredCertificate` returns, per your suggestion — no nested transaction, no second pooled connection per batch, no lock held on the discovery row for the batch's duration.

**Folded into #1900** with severity and reasoning: UUID chunking against the 65,535 bind-parameter cap (the one I'd least like forgotten), authorization-denial attribution, interrupt-flag hygiene, deriving `GroupImportResult.committed`, the `DiscoveryRunContext`/`DiscoverySource` duplication, `prepareIssuedCertificate` fingerprint stamping — with your counterpoint about the 7 changed signatures recorded — and the `generateSelfSignedCertificate` test nit.

Two notes on framing, since they affect how the PR should be read:

The trigger-evaluation finding is **not a regression from this work**, as you point out yourself — `main` lost the same certificates by the same mechanics and reported COMPLETED. I've recorded that in the spec, because the honest summary is that this PR made a pre-existing landmine visible and then defused it, rather than introducing anything.

And I've recorded the two findings you **refuted** during verification, in the plan doc alongside the confirmed ones. They bound how fragile this area actually is, which is worth as much to the next reader as the defects.

CI is green on `c5ab09906` apart from the SonarCloud items already triaged: two CRITICALs marked with rationale (one a genuine false positive — the class-level `@Transactional` the rule points at doesn't govern a private self-invoked method), and the S5778 pair whose structure can't change without changing what the test tests.

src/main/java/com/otilm/core/dao/repository/DiscoveryCertificateRepository.java
lubomirw: Agreed, and tracked in #1900 as a follow-up.

Your number sent me to check, because another reviewer had said 65,535 on the same line. In pgjdbc 42.7.8 the guard is `parameterCount > maxSupportedParameters`, and that is a **connection property** rather than a constant — checked only when `preferQueryMode != SIMPLE`. So the cap is version- and configuration-dependent: 32,767 is the signed-Int16 reading of the protocol field, 65,535 the unsigned one. I've noted on the issue that nobody should code against either figure.

I also recorded your `unnest` variant as the preferred fix over chunking — one bind regardless of row count removes the limit instead of staying under it, and it's what the driver's own error message suggests.

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
lubomirw: Good catch, and worth writing down rather than leaving implicit — I've added it to the design doc.

It's a direct consequence of writing bookkeeping during the consumption loop instead of at the end of the run, which is what lets a reason survive its own group's rollback. So the transient clean row is the price of the property this PR exists to deliver. Safe as you say, since the discovery stays `IN_PROGRESS` throughout, but now stated as a design property for exactly the polling case you describe.

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
lubomirw: Fixed in 00951d61a — the pointer is now conditional on `inventoryGaps + keyGaps + notAttempted > 0`.

It was worse than the version you reviewed: I'd since added a "Validation of the discovered certificates could not be requested." sentence, which says nothing per-certificate and was also getting the pointer appended. Both cases now state what happened and stop. Two existing tests asserted the old trailer and were updated deliberately; a new one pins the rule in both directions.

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
lubomirw: Fixed in 00951d61a.

There was a rule underneath the inconsistency that isn't visible from the diff: capitalised strings are *prefixes* (`"Import rolled back: "`) and lowercase ones are *fragments* meant to follow them, which is what lets reasons compose. But several fragments are persisted standalone, so they reached the operator lowercase — your read was right as displayed.

Rather than edit the literals, reasons are now cased once at the point of persistence. Composition stays free, and it can't drift again as strings are added.

lubomirw:
Thanks — you called only the first one worth a code change, but I've taken two of the four, because both turned out to affect what an operator actually reads.

**Fixed in `00951d61a`:**

- **The contradictory pointer.** Worse than the version you reviewed: I'd since added a "Validation of the discovered certificates could not be requested." sentence, which says nothing per-certificate and was also getting the pointer appended. Now conditional on `inventoryGaps + keyGaps + notAttempted > 0`.
- **Capitalisation.** Cased once at the point of persistence rather than editing the literals, so the prefix/fragment split that lets reasons compose survives and can't drift as strings are added.

**Tracked in #1900** — the unbounded `IN :uuids`, with your `unnest` variant recorded as preferable to chunking.

**Design doc** now states the two-write row state you spotted, as a design property rather than an accident.

Two things your review changed beyond the fixes, worth saying explicitly:

Your bind-parameter figure disagreed with another reviewer's on the same line (32,767 vs 65,535), so I checked pgjdbc 42.7.8 instead of picking one. The guard is `parameterCount > maxSupportedParameters`, a **connection property** rather than a constant, checked only when `preferQueryMode != SIMPLE`. Neither number is safe to code against, and **I'd already recorded 65,535 as fact on #1900** — that's now corrected there. The disagreement was more useful than either answer.

And the transient-clean-row note is the kind of finding I'd not have written down on my own. It's the price of writing bookkeeping during the consumption loop rather than at the end, which is exactly the property this PR exists to deliver — so it belongs in the spec next to the guarantee it pays for.

klaraf755 (approved):

PR #1908, opened 2026-07-29
Reject key creation for disabled Token Profiles

### LINKED ISSUE #1395: Keys: Key or key pair creation allowed for disabled Token Profile
## Description
ILM Core permits creation of a key or key pair even when the target Token Profile is in a **disabled** state. No validation of the Token Profile's enabled status is performed before processing the key creation request. This is a business logic violation — keys should only be creatable on active, enabled Token Profiles, as a disabled Token Profile signals that the underlying token is unavailable or intentionally taken out of service.

## Steps to Reproduce
1. Disable a Token Profile via **Profiles** → **Token Profiles** (or via API).
2. Send a `POST` request to the key creation endpoint, referencing the disabled Token Profile's UUID.
3. Observe that the key (or key pair) is created successfully.

## Expected Result
- The API rejects the key creation request with an appropriate error (e.g., `422 Unprocessable Entity`) when the specified Token Profile is disabled.
- An informative error message is returned indicating that the Token Profile is not in an enabled state.

## Actual Result
- The key or key pair is created successfully despite the referenced Token Profile being disabled.
- No validation error or rejection is returned by the API.

## Environment

ILM Core:
Commit: 2.17.0
Branch:
Build time: 2026-04-05 10:41

Client:
- OS:
- Browser:

lubomirw:
## Summary
- reject key creation when the Token Profile is disabled
- return a controlled validation error before connector processing
- cover the behavior with an integration regression test that verifies zero connector traffic

## Test plan
- `mvn --quiet '-Dtest=CryptographicKeyServiceITest#testAddKey_disabledTokenProfile' test`

Closes #1395

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
@@ -352,7 +352,7 @@ public class CryptographicKeyServiceImpl implements CryptographicKeyExternalServ
                         tokenProfileUuid)
                 .orElseThrow(
                         () -> new NotFoundException(
-                                TokenInstanceReference.class,
+                                TokenProfile.class,
                                 tokenProfileUuid
                         )
                 );
--- a/src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
@@ -356,6 +356,9 @@ public class CryptographicKeyServiceImpl implements CryptographicKeyExternalServ
                                 tokenProfileUuid
                         )
                 );
+        if (!Boolean.TRUE.equals(tokenProfile.getEnabled())) {
+            throw new ValidationException(ValidationError.create("Token Profile is disabled"));
+        }
 
         attributeEngine.validateCustomAttributesContent(Resource.CRYPTOGRAPHIC_KEY, request.getCustomAttributes());
         mergeAndValidateAttributes(type, tokenInstanceReference, request.getAttributes());
--- a/src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
@@ -356,9 +356,7 @@ public class CryptographicKeyServiceImpl implements CryptographicKeyExternalServ
                                 tokenProfileUuid
                         )
                 );
-        if (!Boolean.TRUE.equals(tokenProfile.getEnabled())) {
-            throw new ValidationException(ValidationError.create("Token Profile is disabled"));
-        }
+        validateTokenProfileEnabled(tokenProfile);
 
         attributeEngine.validateCustomAttributesContent(Resource.CRYPTOGRAPHIC_KEY, request.getCustomAttributes());
         mergeAndValidateAttributes(type, tokenInstanceReference, request.getAttributes());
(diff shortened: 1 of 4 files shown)


src/test/java/com/otilm/core/integration/service/CryptographicKeyServiceITest.java
```diff
@@ -292,6 +292,45 @@ void testAddKey() throws ConnectorException, AlreadyExistException, AttributeExc
...
+                .post(WireMock.urlPathMatching("/v1/cryptographyProvider/tokens/[^/]+/keys/secret/attributes/validate"))
+                .willReturn(WireMock.ok()));
+        mockServer.stubFor(WireMock
+                .post(WireMock.urlPathMatching("/v1/cryptographyProvider/tokens/[^/]+/keys/secret"))
+                .willReturn(WireMock.okJson("{\"name\":\"disabledProfileKey\", \"uuid\":\"249db149-8c51-11ed-a1eb-0242ac120003\", \"keyData\":{\"type\":\"Secret\", \"algorithm\":\"RSA\", \"format\":\"Raw\", \"value\":\"secret\"}}")));
+
+        tokenProfile.setEnabled(false);
+        tokenProfileRepository.saveAndFlush(tokenProfile);
+
+        KeyRequestDto request = new KeyRequestDto();
```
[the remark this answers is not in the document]
lubomirw: Fixed in 24b82235d. The test now resets WireMock request history immediately before invoking createKey, so the zero-traffic assertion is isolated from fixture activity.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -356,6 +356,9 @@ public KeyDetailDto createKey(UUID tokenInstanceUuid, SecuredParentUUID tokenPro
                                 tokenProfileUuid
                         )
                 );
+        if (!Boolean.TRUE.equals(tokenProfile.getEnabled())) {
```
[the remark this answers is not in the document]
lubomirw: Fixed in 24b82235d. The missing-profile path now constructs NotFoundException with TokenProfile.class, with a regression test verifying the reported entity type.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -352,10 +352,13 @@ public KeyDetailDto createKey(UUID tokenInstanceUuid, SecuredParentUUID tokenPro
                         tokenProfileUuid)
                 .orElseThrow(
                         () -> new NotFoundException(
-                                TokenInstanceReference.class,
+                                TokenProfile.class,
```
klaraf755, line 355: Good fix. `editKey:423` still has the identical `NotFoundException(TokenInstanceReference.class, request.getTokenProfileUuid())` on a `tokenProfileRepository.findByUuid` lookup — after this change it's the sole remaining occurrence in the file (`listCreateKeyAttributes:759` already uses `TokenProfile.class`). Right now `PUT /v1/keys/{uuid}` with a bad UUID says "TokenInstanceReference … not found" while `createKey` correctly says TokenProfile — two endpoints, one file, opposite labels for the same mistake. One-word fix while you're here.

src/test/java/com/otilm/core/integration/service/CryptographicKeyServiceITest.java
```diff
@@ -292,6 +292,68 @@ void testAddKey() throws ConnectorException, AlreadyExistException, AttributeExc
         Assertions.assertEquals(1, dto.getItems().size());
     }
 
+    @Test
+    void testAddKey_disabledTokenProfile() {
+        mockServer.stubFor(WireMock
```
klaraf755: All five stubs are dead by construction — the guard throws at `CryptographicKeyServiceImpl:359`, before `mergeAndValidateAttributes` (365), which is the first code touching `connectorApiFactory`. Line 334 asserts precisely that none of them is hit, so the test's own assertion proves these 15 lines unreachable.

`mockServer.resetRequests()` at line 321 is also a no-op — `@BeforeEach` builds a fresh `WireMockServer` (93) and setup persists via JPA only. Worse, it sits *after* all setup, so if connector-touching setup is ever added it silently erases that traffic and line 334 keeps passing while claiming zero traffic. Suggest deleting 297-311 and 321; `verify(0, …)` alone is the stronger test.

klaraf755 (commented):
Core change is sound — guard is placed before every connector call, the `NotFoundException` class swap breaks no contract (`NotFoundException(Class, Object)` is just `getSimpleName()` in a message, no error code or i18n key), and both new tests pin origin rather than just exception type. No new Spring context signature, so leaving `ContextSignatureGuardTest.BASELINE` alone is correct.

The gap is coverage. Three paths reach a Token Profile and skip the check:

- **`CryptographicKeyServiceImpl:434`** — `editKey` re-parents a key onto any profile with no enabled check. Create on an enabled profile, then `PUT /v1/keys/{uuid}` onto the disabled one. The guard is circumventable in two calls.
- **`CryptographicOperationServiceImpl:468`** — `getPublicAndPrivateKey` runs `verifyActive` on key *items*, never the profile. `generateCsr` (440) flows straight through, and `ClientOperationServiceImpl:2454` reaches it from issue (381) and renew/rekey (1907). Sign, CSR, and issuance on a disabled profile all still succeed.
- **`CryptographicKeyServiceImpl:756`** — `listCreateKeyAttributes` serves the create-key form *and hits the connector* for a disabled profile. This one contradicts the PR's own `verify(0, anyRequestedFor(anyUrl()))` assertion.

src/test/java/com/otilm/core/integration/service/CryptographicKeyServiceITest.java
lubomirw: Fixed in 5b044fbde. Removed all five unreachable connector stubs and resetRequests; the whole-test zero-request assertion now detects connector traffic from any phase.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
lubomirw: Fixed in 5b044fbde. editKey now reports TokenProfile for a failed profile lookup, with regression coverage for the entity label.

lubomirw:
Addressed the remaining enabled-state paths in 5b044fbde:

- reject reassignment to a disabled Token Profile before mutating key metadata
- reject create-attribute preflight before connector traffic
- reject CSR generation for disabled default or alternate key profiles before signing
- add regression coverage for all paths, including unchanged persisted metadata after rejected reassignment

klaraf755 (approved):

PR #1913, opened 2026-07-30
Import discovered certificates when an action trigger fails

### LINKED ISSUE #1897: Discovery loses certificates to rolled-back per-certificate transactions and cannot report which

lubomirw:
## Problem

Running a discovery with action triggers imported **zero** certificates. Every group logged `UnexpectedRollbackException` from `CertificateDiscoveredEventHandler`, each preceded by:

```
Execution 'test-action' of action 'test-execution-2' has not been performed. Reason:
Cannot invoke "AuthorityInstanceReference.getConnectorInterface()" because "authority" is null
```

## Root cause

`CertificateTriggerEvaluator` calls `CertificateServiceImpl.switchRaProfile`, which is class-level `@Transactional` and therefore joined the import transaction as `REQUIRED`. Selecting an adapter for an RA profile whose `AuthorityInstanceReference` is null threw an NPE, and an unchecked exception crossing a *participating* proxy marks the shared physical transaction rollback-only (`globalRollbackOnParticipationFailure` defaults to true).

`TriggerEvaluator.performActions` did catch the NPE — but a catch cannot unmark a rollback-only transaction, so the group's commit threw `UnexpectedRollbackException` afterwards. Every group evaluates the same trigger list, so one misconfigured execution cost the discovery every certificate it found.

## Changes

**Containment.** Action triggers now run after the import has committed, one transaction per trigger. Not inside the import transaction, because an execution reaches services that are class-level `@Transactional`; and one transaction each rather than one for the phase, because a poisoned transaction discards every write made in it — sharing one would lose the history, applied changes and notifications of the triggers that succeeded. A trigger failure is recorded in its own fresh transaction so it does not exist only in the log.

**A missing authority is a stated precondition.** `AuthorityProviderAdapterFactory.forAuthority` uses `Objects.requireNonNull`, and callers that can legitimately hold a null reference check it themselves and report it in their own terms. Selecting a catchable type instead routes the condition into whatever another class's handler means by it — which is how a missing authority could be reported as the authority refusing to cancel, stranding certificates mid-operation.

**`UnsupportedAuthorityVersionException` answers 400, not 500.** An unrecognised connector interface version is caller-fixable configuration. The response body is fixed rather than the exception's message, which names the authority and a version string the connector reported into an unvalidated column.

**Notification dispatch failures stay in the listener.** Spring propagates an exception thrown from an `AFTER_COMMIT` synchronization to whoever called commit, by which point the transaction has committed and cannot be undone — so an unreachable broker reported committed work as failed, here as a second trigger history contradicting the committed one. The retry template has already exhausted its attempts by then, so the log is the only useful disposition. A caller dispatching directly still sees the failure.

**Operator-facing text names the right transaction.** The failure classifier reported every `UnexpectedRollbackException` as the *import* transaction rolling back. On the trigger path the import has already committed, so only the trigger's own transaction can have — and that is the dominant failure shape there.

## Testing

- `EventHandlersITest` — 25 tests, including a failing action execution, a succeeding trigger applied across the transaction boundary, and a succeeding trigger kept when another fails.
- `CertificateDiscoveredEventHandlerContainmentTest` — 10 tests, including one transaction per trigger and a failure record surviving a lost trigger transaction.
- `DiscoveryFailureReasonTest` 16, `NotificationProducerTest` 3 (new), `AuthorityProviderAdapterFactoryTest` 6, architecture gates 20, `ContextSignatureGuardTest` 2. No new Spring context signature.

Two of the three new integration tests are honest about their limits: they pass with the containment removed, because no execution an integration test can configure fails *unchecked*. Their Javadoc says so. `aLostTriggerTransactionStillLeavesAFailureRecord` genuinely fails without its fix.

## Notes and follow-ups

- A notification is released as its own trigger commits, so it no longer implies that the triggers ordered after it have run. Documented at both the publish site and the trigger phase; accepted rather than deferred, since collecting messages until the phase ended would decouple each notification from its own trigger's success.
- The `AFTER_COMMIT` dispatch guard was applied to notifications only. Six other `@TransactionalEventListener` methods share the same latent flaw — worth a sweep.
- A per-trigger transaction spans the connector's `identify` HTTP call when a SET_FIELD execution switches the
  RA profile, so a DB transaction is held for the connector's full response time. Strictly narrower than
  before -- the whole import transaction used to span the same call -- but hoisting external calls out of
  `TriggerEvaluator` is its own change.
- Stale connector-scoped certificate META rows survive an authority force-delete: the force path detaches RA profiles and cleans only `object_type='AUTHORITY'` rows. Pre-existing and unrelated to this fix (that code path previously threw an NPE), and not expressible with today's `AttributeEngine` API, which has no connector-independent delete and no column marking `RA_PROFILE` as a row's source.

Related to #1897 (the preceding per-certificate import-integrity work) and #1900.

### DIFF
--- a/src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
+++ b/src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
@@ -32,6 +32,7 @@ import com.otilm.core.messaging.jms.producers.ValidationProducer;
 import com.otilm.core.messaging.model.EventMessage;
 import com.otilm.core.messaging.model.ValidationMessage;
 import com.otilm.core.service.CertificateInternalService;
+import com.otilm.core.service.TriggerInternalService;
 import com.otilm.core.service.handler.CertificateHandler;
 import com.otilm.core.tasks.ScheduledJobInfo;
 import com.otilm.core.util.CertificateUtil;
@@ -83,6 +84,7 @@ public class CertificateDiscoveredEventHandler extends EventHandler<Certificate>
     private DiscoveryCertificateRepository discoveryCertificateRepository;
     private DiscoveryWriter discoveryWriter;
     private AuthorizationEnforcer authorizationEnforcer;
+    private TriggerInternalService triggerService;
 
     // Kept alongside the inherited, more loosely typed handle, for the finders SecurityFilterRepository does not
     // expose.
@@ -132,6 +134,11 @@ public class CertificateDiscoveredEventHandler extends EventHandler<Certificate>
     @Autowired
     public void setAuthorizationEnforcer(AuthorizationEnforcer authorizationEnforcer) {
         this.authorizationEnforcer = authorizationEnforcer;
+    }
+
+    @Autowired
+    public void setTriggerService(TriggerInternalService triggerService) {
+        this.triggerService = triggerService;
     }
 
     @Override
@@ -702,7 +709,8 @@ public class CertificateDiscoveredEventHandler extends EventHandler<Certificate>
     }
 
     /**
-     * A committed group's result, plus what its action triggers need once the import transaction has closed.
+     * A group's import result, plus what its action triggers need once the import transaction has closed --
+     * absent when the group imported nothing.
      *
      * @param certificateUuid   the imported certificate, or null when the group imported nothing
      * @param eventData         the trigger payload; non-null exactly when {@code certificateUuid} is
@@ -745,11 +753,38 @@ public class CertificateDiscoveredEventHandler extends EventHandler<Certificate>
             try {
                 transactionHandler.runInNewTransaction(() -> runActionTrigger(context, imported, triggerAssociation));
             } catch (Exception e) {
-                // Nothing is swallowed inside the transaction, so it rolls back cleanly and the next trigger starts
-                // from a sound one.
+                // Only a failure that escaped the trigger's own handling reaches here, and its transaction is
+                // already gone -- taking the history the evaluator wrote in it. The next trigger starts in a fresh
+                // one, and the failure is recorded in a fresh one too, so it does not exist only in the log.
                 logger.error("Action trigger {} failed for discovered certificate {}: {}",
                         triggerAssociation.getTrigger().getUuid(), imported.certificateUuid(), e.getMessage(), e);
+                recordActionTriggerFailure(context, imported, triggerAssociation, e);
             }
+        }
+    }
+
+    /**
+     * Writes the failure where an operator looks for it. The evaluator records a failed execution itself, but into
+     * the transaction the failure poisoned, so that record dies with it — leaving the trigger looking as though it
+     * never ran. This one is written in its own transaction and therefore survives.
+     */
+    private void recordActionTriggerFailure(DiscoveryRunContext context, ImportedGroup imported,
+                                            TriggerAssociation triggerAssociation, Exception failure) {
+        try {
+            transactionHandler.runInNewTransaction(() -> {
+                TriggerHistory history = triggerService.createTriggerHistory(
+                        triggerAssociation.getTrigger().getUuid(), triggerAssociation, imported.certificateUuid(),
+                        imported.referenceRowUuid(), eventHistoryFor(context, triggerAssociation),
+                        Resource.CERTIFICATE);
+                history.setConditionsMatched(true);
+                history.setActionsPerformed(false);
+                triggerService.createTriggerHistoryRecord(history.getUuid(), null, null,
+                        "The trigger's actions could not be applied: " + DiscoveryFailureReason.shape(failure));
+            });
+        } catch (Exception e) {
+            // The last place the failure could have been recorded, so the log is all that is left.
+            logger.error("Could not record the failure of action trigger {} for certificate {}: {}",
+                    triggerAssociation.getTrigger().getUuid(), imported.certificateUuid(), e.getMessage(), e);
         }
     }
 
--- a/src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
+++ b/src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
@@ -84,8 +84,8 @@ public class CertificateDiscoveredEventHandler extends EventHandler<Certificate>
     private DiscoveryWriter discoveryWriter;
     private AuthorizationEnforcer authorizationEnforcer;
 
-    // Kept alongside the inherited, more loosely typed handle: the unconsumed-group accounting needs a
-    // certificate-specific finder that SecurityFilterRepository does not expose.
+    // Kept alongside the inherited, more loosely typed handle, for the finders SecurityFilterRepository does not
+    // expose.
     private final CertificateRepository certificateRepository;
 
     @Autowired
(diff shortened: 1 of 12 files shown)


src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
```diff
@@ -688,24 +700,124 @@ private GroupImportResult importContentGroupInternal(DiscoveryRunContext context
...
+            transactionHandler.runInNewTransaction(() -> {
+                TriggerHistory history = triggerService.createTriggerHistory(
+                        triggerAssociation.getTrigger().getUuid(), triggerAssociation, imported.certificateUuid(),
+                        imported.referenceRowUuid(), eventHistoryFor(context, triggerAssociation),
+                        Resource.CERTIFICATE);
+                history.setConditionsMatched(true);
+                history.setActionsPerformed(false);
+                triggerService.createTriggerHistoryRecord(history.getUuid(), null, null,
+                        "The trigger's actions could not be applied: "
+                                + DiscoveryFailureReason.shapeTriggerFailure(failure));
```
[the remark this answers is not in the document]
lubomirw: `conditionsMatched` is a primitive `boolean` with `@Column(nullable = false)`, so there is no "not determined" value to write here — only true or false.

`false` renders as *"rules did not match"*, which an operator reads as the trigger legitimately skipping this certificate. That hides the very failure this record exists to report, so it is the worse of the two.

The pairing kept here — `conditionsMatched=true`, `actionsPerformed=false`, plus a `TriggerHistoryRecord` carrying the reason — is exactly how `TriggerEvaluator.evaluateTrigger` already reports actions that did not complete (`setActionsPerformed(triggerHistory.getRecords().isEmpty())`), so the fallback matches the existing representation rather than inventing one.

You are right that for an unchecked failure escaping *rule* evaluation the flag asserts something unverified; the two guards on that path convert unchecked failures into a checked `RuleException`, which is handled separately and never reaches here, but the residual case exists. Rationale added in 273b1e044.

3keyroman (dismissed):

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
```diff
@@ -688,24 +700,118 @@ private GroupImportResult importContentGroupInternal(DiscoveryRunContext context
...
-                context.eventContext().getTriggerEvaluator().evaluateTrigger(triggerAssociation.getTrigger(), triggerAssociation,
-                        certificate, referenceRowUuid, eventData, eventHistory);
-            } catch (RuleException e) {
-                // Contained per trigger, and deliberately not an import failure: the certificate is already stored
-                // and trigger history is where an action failure is reported. Letting it out would roll this group
-                // back — and since every group evaluates the same triggers, one misconfigured action trigger would
-                // then lose every certificate in the discovery.
-                logger.error("Action trigger {} failed for certificate {}: {}",
-                        triggerAssociation.getTrigger().getUuid(), certificate.getUuid(), e.getMessage(), e);
+                transactionHandler.runInNewTransaction(() -> runActionTrigger(context, imported, triggerAssociation));
```
klaraf755, line 748: Each per-trigger transaction wraps the whole `evaluateTrigger`, and a SET_FIELD execution that switches the RA profile makes the connector's `identify` HTTP call inside it — so a DB transaction is held across an external call for the connector's full response time. This is strictly narrower than before (the entire import transaction used to span the same call), and hoisting external calls out of `TriggerEvaluator` is clearly out of scope here — but worth adding to the follow-up list alongside the `@TransactionalEventListener` sweep.

src/main/java/com/otilm/core/messaging/jms/producers/NotificationProducer.java
```diff
@@ -41,9 +41,20 @@ private void sendMessage(final NotificationMessage notificationMessage) {
...
+     * Spring propagates what an {@code AFTER_COMMIT} synchronization throws to whoever called commit, by which point
+     * the transaction has committed -- so letting a dispatch failure out reports committed work as failed. Retries
+     * are exhausted by here, so the log is all that is left.
+     */
     @TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT, fallbackExecution = true)
     public void onNotificationMessage(NotificationMessage notificationMessage) {
-        produceMessage(notificationMessage);
+        try {
+            produceMessage(notificationMessage);
+        } catch (Exception e) {
```
klaraf755, line 54: The AFTER_COMMIT reasoning in the Javadoc doesn't hold for the `fallbackExecution = true` branch: with no ambient transaction the listener runs synchronously inside `performActions`, whose per-execution catch used to turn a dispatch failure into a `TriggerHistoryRecord`. With this swallow, on the no-transaction path the failure lands only in the log while trigger history implies the notification went out. If no production caller publishes outside a transaction this is theoretical, but a one-line Javadoc caveat would keep the next reader from assuming the guard is loss-free there.

lubomirw: Applied in 77bac0748 — the Javadoc now says the guard is not loss-free there:

> Not loss-free on the `fallbackExecution` path though: with no transaction the listener runs inline, where the publisher could still have acted on the failure.

Checked how reachable it is: all nine publishers of a `NotificationMessage` through the event bus sit under a class-level `@Transactional` (`TriggerEvaluator` plus the eight event handlers), and `sendFollowUpEventsNotifications` is only ever called from `handleEvent` on those beans, so every current publish happens with a transaction active and takes the `AFTER_COMMIT` path. Theoretical today, as you said — the caveat is there so it stays visible if a caller without a transaction shows up.

One refinement on the mechanism: the `performActions` per-execution catch turning it into a `TriggerHistoryRecord` applies to the trigger-evaluator path; for the eight event handlers a dispatch failure would simply have propagated into the handler. Same conclusion either way.

src/main/java/com/otilm/core/events/handlers/CertificateDiscoveredEventHandler.java
lubomirw: Agreed, and confirmed: `runActionTrigger` wraps `evaluateTrigger` in `runInNewTransaction`, and a SET_FIELD execution that switches the RA profile reaches `switchRaProfile` -> `identifyByNewAuthority` -> `adapter.identify(...)`, which is the connector HTTP call. So the transaction is held for the connector's full response time, against the repo's own rule about not holding a transaction across an external call.

Also agreed it is not a regression — the whole import transaction spanned the same call before, so this narrows it — and that hoisting external calls out of `TriggerEvaluator` is its own change: the fix is a split (read/validate, call, then open a transaction for the writes) in shared trigger code used by every resource, not something to land in a rollback fix.

Added to the PR's follow-up list alongside the `@TransactionalEventListener` sweep, and it will go on core#1900 with them.

klaraf755 (approved):

PR #1914, opened 2026-07-30
Attribute certificate history events to the correct user

### LINKED ISSUE #1910: Certificate history events are attributed to an unrelated user because SecurityContext leaks across JMS event messages
### Description

Certificate history events — and every other audited write performed during event handling — are attributed to the wrong user. A user logged in through the internal OAuth2 provider uploads a certificate, and the resulting "Certificate uploaded" history row names an unrelated real user.

**Root cause — SecurityContext leaks across JMS event messages:**

- `POST /api/v1/certificates/upload/async` hands the work to the events queue. `CertificateUploadedEventHandler.constructEventMessage` (`CertificateUploadedEventHandler.java:85`) uses the 4-arg `EventMessage` constructor, leaving `userUuid` null (`EventMessage.java:25-30`).
- `EventListener.processMessage` (`EventListener.java:28-39`) authenticates only when `userUuid != null`, and never clears the context in a `finally`.
- The events queue is a `DefaultMessageListenerContainer` (`JmsConfig.java:129-143`, `events: 5` → concurrency range 1–5), so one long-lived consumer thread serves essentially all event messages at normal load.
- The only events populating `userUuid` are the two discovery events (`DiscoveryFinishedEventHandler.java:83`, `CertificateDiscoveredEventHandler.java:734`). Consuming one calls `authHelper.authenticateAsUser(...)`, installing the discovery initiator's identity on that consumer thread permanently.
- Every later upload inherits it. `CertificateEventHistory` and `Certificate` both extend `UniquelyIdentifiedAndAudited`, whose `author` column is `@CreatedBy`/`@LastModifiedBy` (`Audited.java:23-26`), resolved by `CustomAuditAware.getCurrentAuditor()` (`CustomAuditAware.java:13-28`) from `SecurityContextHolder` — yielding the stale user's username, or `system` when the context is empty.

**Scope is wider than certificate upload.** Nine of eleven `new EventMessage(...)` sites use the 4-arg constructor and inherit whatever identity is on the thread: `CERTIFICATE_UPLOADED`, `APPROVAL_REQUESTED`, `APPROVAL_CLOSED`, `CERTIFICATE_ACTION_PERFORMED`, `CERTIFICATE_STATUS_CHANGED`, `CERTIFICATE_NOT_COMPLIANT`, `CERTIFICATE_EXPIRING`, `CERTIFICATE_REGISTERED`, `SCHEDULED_JOB_FINISHED`. The uploaded certificate's own `i_author` is wrong too, not just the history row.

**Two further consequences of the same defect:**

1. **Custom attributes are dropped on async upload when the consumer thread has no authentication.** Verified by trace. `loadCustomAttributesSecurityResourceFilter` (`AttributeEngine.java:2038-2050`) falls through to `AuthHelper.loadObjectPermissions`, which returns a deny-all filter for a null `Authentication` (`AuthHelper.java:210-213`). `validateCustomAttributesContent` at `AttributeEngine.java:1168` then passes vacuously — with `areOnlySpecificObjectsAllowed == true`, lines `:1576-1577` filter both the definition map and its local attribute copy down to empty, hitting the early return at `:1587-1589`. Control therefore reaches the else branch at `:1179`, and the loop at `:1183` iterates the **original, unfiltered** `requestAttributes`, so `checkCustomAttributeUpdatePermissions` (`:1194-1204`) throws `AttributeException` on the first attribute. That is caught and swallowed at `CertificateUploadedEventHandler.java:158-160` — logged at ERROR (`Error updating custom attributes for certificate {}: {}`) but invisible to the operator, whose async upload already returned success. The certificate persists without its custom attributes, even though the API thread validated them as the uploader (`CertificateUploadServiceImpl.java:68-70`). Note the `ValidationException` path does *not* escape here; had it thrown, it would have bypassed `EventListener`'s `catch (EventException)` at `EventListener.java:36` and lost the certificate entirely.

   On a thread that *does* carry a stale identity, OPA is instead queried with that user's `rawData` (`AuthHelper.java:217`); if the stale user holds broad attribute permissions the write succeeds normally via `AttributeEngine.java:1171-1178`. The loss window is therefore a freshly restarted pod before any discovery has run, or a stale user lacking `ATTRIBUTE`/`MEMBERS` permissions — which is why the attribution symptom can be observed without this consequence appearing alongside it.
2. **`EventHandler.handleUser`** (`EventHandler.java:240-257`) re-authenticates as a trigger association's creator and never restores. Since `addEventHistory` runs after trigger evaluation (`CertificateUploadedEventHandler.java:146` then `:163`), any deployment with a Certificate-uploaded trigger attributes the history row to the trigger creator. Its guard at `:241` compares `triggeredBy` against `context.getCurrentUserUuid()` (seeded from the message), not the live SecurityContext, so a null/null comparison short-circuits and a poisoned context is never cleared.

**Proposed fix, in priority order:**

1. Save and restore the SecurityContext in a `try/finally` around `EventListener.processMessage` (`EventListener.java:28-39`) and the `ActionsListener` equivalent, mirroring `AuthHelper.runAsSystem` (`AuthHelper.java:119-139`). This alone ends cross-message identity leakage for every event type.
2. Propagate the acting user: `CertificateUploadServiceImpl.java:76` + `CertificateUploadedEventHandler.java:84-86` switch to the 8-arg `EventMessage` constructor. Guard the `authenticateAsUser` call so a deleted or disabled uploader degrades to unauthenticated rather than throwing out of `processMessage` into the JMS retry and losing the certificate.
3. Scope the trigger impersonation: snapshot/restore around `evaluateTriggers` / `evaluateIgnoreTriggers` (`EventHandler.java:204-238`), and compare against the live SecurityContext at `:241`.
4. Make the `"system"` fallback at `CustomAuditAware.java:15-16` log a WARN so silent misattribution is visible.

**Authorization impact was checked.** The no-trigger upload path contains no `@ExternalAuthorization`/`@PreAuthorize` calls, so propagating identity cannot newly deny an upload. `@ExternalAuthorization` fails closed on a null `Authentication` (`MethodSecurityConfig.java:30-42` → Spring Security's `AuthorizationManagerBeforeMethodInterceptor`), so nothing is silently bypassed by the aspect. Propagating identity also makes the custom-attribute filter consistent with request-time validation. Trigger processing keeps the trigger creator's permissions per #1063, so this does not conflict with that model.

Related: #1684 (event-history ownership for approval-driven transitions) touches two of the handlers that omit `userUuid`.

### Steps to Reproduce

1. Restart core, so all JMS consumer threads start with a clean SecurityContext.
2. Log in as user A and run any discovery. This consumes a `DISCOVERY_FINISHED` / `CERTIFICATE_DISCOVERED` event, which carries `userUuid` and authenticates the events-queue consumer thread as user A.
3. Log in as user B via the internal OAuth2 provider and upload a certificate (`POST /api/v1/certificates/upload/async`).
4. Open the uploaded certificate and view its Event History.

To confirm the other half: repeat step 3 on a freshly restarted pod *before* running any discovery — the author comes out as `system`.

### Expected Behavior

The "Certificate uploaded" history event is attributed to user B, who actually performed the upload, and the certificate's `i_author` is likewise user B. No audited write performed during event handling ever carries an identity belonging to an unrelated user.

### Actual Behavior

The history event is attributed to user A, the initiator of the last discovery, who had no involvement in the upload. On a freshly restarted pod that has not yet run a discovery, the same row is attributed to `system`. Reproduces on every upload.

The audit log records the correct actor, because it is captured on the HTTP request thread — only the JPA-audited `i_author` columns are wrong, which makes the discrepancy visible side by side:

```
[14:10:05.527] DEBUG [tomcat-handler-164646] [OAuth2LoginFilter - 145]: Session of user 'user_jv@example.com' logged using OAuth2 Provider 'internal' has been successfully validated.

[14:10:05.564] DEBUG [eventListener-1] [AbstractJmsEndpointConfig - 94]: Message JSON in endpoint eventListener: {"event":"certificate_uploaded","resource":"certificates","objectUuid":null,...,"userUuid":null,"scheduledJobInfo":null}

[14:10:05.569] INFO [auditLogsListener-1] [AuditLogServiceImpl - 51]: {..."actor":{"type":"user","authMethod":"session","uuid":"a0f1d004-...","name":"user_jv@example.com"},"source":{"method":"POST","path":"/api/v1/certificates/upload/async"},"operation":"upload","operationResult":"success"...}
```

`userUuid: null` on the wire is the identity being dropped at the async boundary; the history row is then written on `eventListener-1`, a thread the upload never authenticated.

### Severity

Critical

### Module

Core

### Environment

Reproduced on Core 2.18.0 and on main (HEAD `fb2e874f2`). Authentication via the internal OAuth2 provider (Keycloak). Deterministic — every upload. Log evidence above captured from a 2.18.0 deployment; the same defect is present at HEAD — `CertificateUploadedEventHandler.java:85`, `EventListener.java:28-39`, and `CustomAuditAware.java:13-28` are identical in substance.

lubomirw:
Fixes #1910.

A certificate's "Certificate uploaded" history event named an unrelated real user — deterministically, on every upload.

## Root cause

JMS listener container threads are long-lived and nothing cleared `SecurityContextHolder` between messages. Discovery events are the only ones that carried a user, so consuming one authenticated the consumer thread; every later certificate upload — whose event message carried no user — inherited that identity, and JPA auditing (`@CreatedBy` → `CustomAuditAware`) stamped it onto the certificate and its history rows. On a freshly restarted pod, before any discovery had run, the same path attributed them to `system` instead.

Reproduced on 2.18.0 and on main. Confirmed by restart test: `system` before any discovery, the discovery initiator's name after.

## What changed

- **`AbstractJmsEndpointConfig`** — reset the SecurityContext and actor MDC per message in the shared dispatch wrapper, so no identity outlives the message that installed it. Covers every endpoint, not just the events queue.
- **Acting user carried in the event message** for the events whose handlers write audited rows on the consuming thread: certificate upload, approval requested, approval closed. Resolved on the producing thread via `AuthHelper.getActingUserUuidOrNull()`.
  - Approval-requested takes the requester as an argument instead: its history row is written only for the first step, produced from `createApproval` on the actions-listener thread *before* that listener authenticates, so the thread cannot supply it. The approval's creator is passed explicitly.
  - An approval closed as `EXPIRED` is left unattributed — the expiry sweep runs as the scheduled job's user, who took no action on it.
- **`EventListener`** — authenticating as the carried user is best-effort, because losing attribution beats losing the event. Both failure shapes are reduced to no identity: an unreachable auth service throws, while a user that no longer resolves comes back as an anonymous principal that auditing would stamp as `anonymousUser` and authorization would accept as genuine.
- **`EventHandler`** — the trigger-association impersonation from #1063 is confined to the trigger loop. Actions still run with the association owner's permissions, but the audited writes afterwards belong to the acting user. `EventContext.currentUserUuid` is no longer trusted as proof that an identity is installed.
- **`TriggerHistory.message` / `TriggerHistoryRecord.message`** mapped as `TEXT`, matching the unbounded `VARCHAR` the migrations declare; the Hibernate-generated test schema defaulted them to `varchar(255)`.

## Authorization impact

Carrying the user widens the authorization context: the listener authenticates for the whole `handleEvent` call, so trigger evaluation and the attribute engine's object-permission filters now run as the acting user rather than unauthenticated. Neither the upload nor the approval path contains any `@ExternalAuthorization`-gated call, so an operation cannot newly be denied. Trigger actions keep the association owner's permissions.

Side effect worth knowing: custom attributes on an async upload were previously dropped on an unauthenticated thread (deny-all object filter → swallowed `AttributeException`). They now evaluate against the uploader, consistent with the validation the request thread already performed.

## Notes for review

- `throwsNotUploaded_whenIgnoreTriggerMatches` was previously green for the wrong reason: `authenticateAsUser` always failed in that ITest, marking the event FAILED and producing the asserted exception without the ignore trigger mattering. It now exercises the path it names. Other ITests may lean on the same accident.
- `jmsListenerContainerFactory` must stay without a `transactionManager` — the JPA transaction commits inside `processMessage`, so the `AFTER_COMMIT` listeners writing audited history rows still run under the message's identity, before the dispatch wrapper clears it. A comment records this; adding one later would silently re-break the bug with no test failing.
- A deleted trigger-association owner is deliberately *not* degraded the way a per-message failure is: that costs only attribution, whereas a trigger without its owner's identity would run actions under other permissions.

## Testing

Six new test classes, 32 unit tests, each watched fail before the fix. Verified green on this tree: the 32 unit tests, `EventHandlersITest` (22), `CertificateServiceITest` (161), `ApprovalServiceITest` (15), `RevocationApprovalRejectedITest` (2). A full-suite run (4381 tests) was green on an earlier revision of this branch; the final tree relies on CI for full coverage.

## Follow-up not included

A SET_FIELD trigger action mutating the certificate inside the trigger loop can auto-flush and stamp `Certificate.author` as the trigger owner, leaving the entity clean so the post-restore commit never re-stamps it. Confirming it needs a new fixture (only `seedIgnoreTrigger` exists today), and it only affects deployments with field-setting upload triggers.

### DIFF
--- a/src/main/java/com/otilm/core/events/EventHandler.java
+++ b/src/main/java/com/otilm/core/events/EventHandler.java
@@ -290,9 +290,12 @@ public abstract class EventHandler<T extends UniquelyIdentifiedObject> implement
      * FAILED.
      */
     protected void handleUser(EventContext<T> context, UUID triggeredBy) {
-        if (!Objects.equals(installedUserUuid(context), triggeredBy)) {
+        // Read from the installed principal, never from EventContext.currentUserUuid: that memo is seeded from the
+        // message and can name a user the thread does not hold, which would skip a needed impersonation.
+        UUID installedUserUuid = AuthHelper.getActingUserUuidOrNull();
+        if (!Objects.equals(installedUserUuid, triggeredBy)) {
             try {
-                logger.debug("Changing user from {} to {}", context.getCurrentUserUuid(), triggeredBy);
+                logger.debug("Changing user from {} to {}", installedUserUuid, triggeredBy);
                 if (triggeredBy == null) {
                     SecurityContextHolder.clearContext();
                 } else {
@@ -308,11 +311,4 @@ public abstract class EventHandler<T extends UniquelyIdentifiedObject> implement
         }
     }
 
-    /**
-     * The user the thread actually holds. {@link EventContext#getCurrentUserUuid()} is seeded from the message, so it
-     * can name a user that was never installed, and trusting that memo alone would skip a needed impersonation.
-     */
-    private UUID installedUserUuid(EventContext<T> context) {
-        return SecurityContextHolder.getContext().getAuthentication() == null ? null : context.getCurrentUserUuid();
-    }
 }
--- a/src/main/java/com/otilm/core/events/EventHandler.java
+++ b/src/main/java/com/otilm/core/events/EventHandler.java
@@ -4,6 +4,7 @@ import com.otilm.api.exception.EventException;
 import com.otilm.api.exception.ValidationException;
 import com.otilm.api.model.client.attribute.RequestAttribute;
 import com.otilm.api.model.core.auth.Resource;
+import com.otilm.api.model.core.logging.enums.ActorType;
 import com.otilm.api.model.core.other.ResourceEvent;
 import com.otilm.api.model.core.workflows.EventStatus;
 import com.otilm.core.dao.entity.UniquelyIdentifiedObject;
@@ -293,22 +294,33 @@ public abstract class EventHandler<T extends UniquelyIdentifiedObject> implement
         // Read from the installed principal, never from EventContext.currentUserUuid: that memo is seeded from the
         // message and can name a user the thread does not hold, which would skip a needed impersonation.
         UUID installedUserUuid = AuthHelper.getActingUserUuidOrNull();
-        if (!Objects.equals(installedUserUuid, triggeredBy)) {
-            try {
-                logger.debug("Changing user from {} to {}", installedUserUuid, triggeredBy);
-                if (triggeredBy == null) {
-                    SecurityContextHolder.clearContext();
-                } else {
-                    authHelper.authenticateAsUser(triggeredBy);
-                    requireImpersonated(triggeredBy);
-                }
-
-                context.setCurrentUserUuid(triggeredBy);
-            } catch (ValidationException e) {
-                // anonymous user
-                SecurityContextHolder.clearContext();
-                context.setCurrentUserUuid(null);
+        if (Objects.equals(installedUserUuid, triggeredBy)) {
+            // No switch needed, but the actor MDC still has to name the identity in effect: the loop clears it on
+            // entry, so without this the first trigger owned by the acting user would be audited with no actor at all.
+            if (triggeredBy != null) {
+                LoggingHelper.putActorInfoWhenNull(ActorType.USER, triggeredBy.toString(), null);
             }
+            return;
+        }
+
+        try {
+            logger.debug("Changing user from {} to {}", installedUserUuid, triggeredBy);
+            // Drop the previous owner's attribution before installing the next: authenticateAsUser only overwrites the
+            // actor uuid, and an ownerless trigger writes no actor at all, so a stale one would otherwise survive.
+            LoggingHelper.clearActorInfo();
+            if (triggeredBy == null) {
+                SecurityContextHolder.clearContext();
+            } else {
+                authHelper.authenticateAsUser(triggeredBy);
+                requireImpersonated(triggeredBy);
+            }
+
+            context.setCurrentUserUuid(triggeredBy);
+        } catch (ValidationException e) {
+            // anonymous user
+            SecurityContextHolder.clearContext();
+            LoggingHelper.clearActorInfo();
+            context.setCurrentUserUuid(null);
         }
     }
 
@@ -320,6 +332,9 @@ public abstract class EventHandler<T extends UniquelyIdentifiedObject> implement
      */
     private void requireImpersonated(UUID triggeredBy) {
         if (!triggeredBy.equals(AuthHelper.getActingUserUuidOrNull())) {
+            // Leave nothing installed for the caller to run under, independently of the loop's restore.
+            SecurityContextHolder.clearContext();
+            LoggingHelper.clearActorInfo();
             throw new PlatformAuthenticationException(
                     "User %s that associated the trigger could not be authenticated".formatted(triggeredBy));
         }
--- a/src/main/java/com/otilm/core/events/EventHandler.java
+++ b/src/main/java/com/otilm/core/events/EventHandler.java
@@ -18,6 +18,7 @@ import com.otilm.core.evaluator.TriggerEvaluator;
 import com.otilm.core.logging.LoggingHelper;
 import com.otilm.core.messaging.jms.producers.EventProducer;
 import com.otilm.core.messaging.model.EventMessage;
+import com.otilm.core.security.authn.PlatformAuthenticationException;
 import com.otilm.core.security.authz.SecuredUUID;
 import com.otilm.core.util.AuthHelper;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -284,10 +285,9 @@ public abstract class EventHandler<T extends UniquelyIdentifiedObject> implement
     }
 
     /**
-     * A deleted association owner or an unreachable auth service is deliberately not degraded here, unlike the
-     * per-message failure in {@code EventListener}: that costs only attribution, whereas a trigger evaluated without
-     * its owner's identity would run its actions under other permissions. The exception escapes, marking the event
-     * FAILED.
+     * A trigger owner that cannot be impersonated is deliberately not degraded here, unlike the per-message failure in
+     * {@code EventListener}: that costs only attribution, whereas a trigger evaluated without its owner's identity
+     * would run its actions under other permissions. The exception escapes, marking the event FAILED.
      */
     protected void handleUser(EventContext<T> context, UUID triggeredBy) {
         // Read from the installed principal, never from EventContext.currentUserUuid: that memo is seeded from the
@@ -300,6 +300,7 @@ public abstract class EventHandler<T extends UniquelyIdentifiedObject> implement
                     SecurityContextHolder.clearContext();
                 } else {
                     authHelper.authenticateAsUser(triggeredBy);
+                    requireImpersonated(triggeredBy);
                 }
 
                 context.setCurrentUserUuid(triggeredBy);
@@ -311,4 +312,16 @@ public abstract class EventHandler<T extends UniquelyIdentifiedObject> implement
         }
     }
 
+
+    /**
+     * Only an unreachable auth service makes {@code authenticateAsUser} throw. A deleted or disabled owner instead has
+     * the auth service answer "not authenticated", which the client turns into an anonymous principal - so the call
+     * returns normally and the trigger would otherwise evaluate under an identity that is neither its owner nor absent.
+     */
+    private void requireImpersonated(UUID triggeredBy) {
+        if (!triggeredBy.equals(AuthHelper.getActingUserUuidOrNull())) {
+            throw new PlatformAuthenticationException(
+                    "User %s that associated the trigger could not be authenticated".formatted(triggeredBy));
+        }
+    }
 }
--- a/src/main/java/com/otilm/core/events/EventHandler.java
+++ b/src/main/java/com/otilm/core/events/EventHandler.java
@@ -15,6 +15,7 @@ import com.otilm.core.dao.repository.SecurityFilterRepository;
 import com.otilm.core.dao.repository.workflows.EventHistoryRepository;
 import com.otilm.core.dao.repository.workflows.TriggerAssociationRepository;
 import com.otilm.core.evaluator.TriggerEvaluator;
+import com.otilm.core.logging.LoggingHelper;
 import com.otilm.core.messaging.jms.producers.EventProducer;
 import com.otilm.core.messaging.model.EventMessage;
 import com.otilm.core.security.authz.SecuredUUID;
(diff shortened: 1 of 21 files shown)


src/main/java/com/otilm/core/events/EventHandler.java
```diff
@@ -255,4 +307,12 @@ protected void handleUser(EventContext<T> context, UUID triggeredBy) {
...
         }
     }
+
+    /**
+     * The user the thread actually holds. {@link EventContext#getCurrentUserUuid()} is seeded from the message, so it
+     * can name a user that was never installed, and trusting that memo alone would skip a needed impersonation.
+     */
+    private UUID installedUserUuid(EventContext<T> context) {
+        return SecurityContextHolder.getContext().getAuthentication() == null ? null : context.getCurrentUserUuid();
+    }
```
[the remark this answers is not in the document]
lubomirw: Agreed, and confirmed reproducible — fixed in 6381c52a9.

You were right that the helper only used the SecurityContext as a boolean gate and then returned `context.getCurrentUserUuid()`, so a divergence between memo and holder still skipped impersonation. I added a test pinning exactly that case (thread holds the uploader, the memo claims the trigger creator, association `triggeredBy` is the trigger creator); it failed with zero interactions on `authenticateAsUser` before the change.

`handleUser` now derives the uuid from the installed principal via `AuthHelper.getActingUserUuidOrNull()`, which already treats a missing authentication, a non-platform principal and an unparseable uuid as absent. The memo no longer participates in the decision, and the private helper is gone.

src/main/java/com/otilm/core/events/EventHandler.java
```diff
@@ -237,10 +253,49 @@ protected boolean evaluateIgnoreTriggers(EventContext<T> context, EventContextTr
...
+            LoggingHelper.restoreActorInfo(actingUserActor);
+            context.setCurrentUserUuid(actingUserUuid);
+        }
+    }
+
+    /**
+     * A deleted association owner or an unreachable auth service is deliberately not degraded here, unlike the
+     * per-message failure in {@code EventListener}: that costs only attribution, whereas a trigger evaluated without
+     * its owner's identity would run its actions under other permissions. The exception escapes, marking the event
+     * FAILED.
```
klaraf755: This Javadoc claims a protection the code does not provide. By this PR's own analysis in `EventListener` ("For a deleted user the auth service answers \"not authenticated\" rather than failing"), `authenticateAsUser` for a deleted owner returns **normally**, installing an anonymous principal that auditing would stamp as `anonymousUser` and authorization would accept as genuine. No exception escapes and the event is not marked FAILED — the trigger's actions run under that anonymous principal, which is exactly the "run its actions under other permissions" outcome described here as prevented. Only the unreachable-auth-service shape actually throws. The `resolvesToRealUser` guard exists in `EventListener` but not here, where per this comment's own reasoning the right disposition is the opposite one: throw and fail the event rather than degrade. (The `catch (ValidationException)` below contradicts the "deliberately not degraded" claim the same way.) The behavior is pre-existing, but documenting it as handled is worse than silence — either add the guard with throw semantics, or correct the Javadoc.

src/main/java/com/otilm/core/events/EventHandler.java
```diff
@@ -237,10 +253,49 @@ protected boolean evaluateIgnoreTriggers(EventContext<T> context, EventContextTr
...
+        SecurityContext actingUserContext = SecurityContextHolder.getContext();
+        boolean actingUserWasAuthenticated = actingUserContext.getAuthentication() != null;
+        UUID actingUserUuid = context.getCurrentUserUuid();
+        Map<String, String> actingUserActor = LoggingHelper.snapshotActorInfo();
+        try {
+            SecurityContext detached = SecurityContextHolder.createEmptyContext();
+            detached.setAuthentication(actingUserContext.getAuthentication());
+            SecurityContextHolder.setContext(detached);
+            // Stops the acting user's actor name mixing with the trigger creator's uuid; restored in finally.
+            LoggingHelper.clearActorInfo();
```
klaraf755, line 274: This clears the actor MDC once at loop entry, but `authenticateAsUser` writes it with `putActorInfoWhenNull` — so with two associations owned by different creators, creator A's actor uuid is installed by the first iteration and survives into the second: creator B's trigger executes with A's actor attribution in the audit logs. The acting-user↔creator mixing is fixed here; creator↔creator mixing remains — the same bug class one level down. Clearing the actor MDC in `handleUser` whenever it switches identity (mirroring what it does to the SecurityContext) would close it.

src/main/java/com/otilm/core/events/EventHandler.java
lubomirw: Valid, and confirmed by test — fixed in ad044729d.

You're right that the javadoc claimed a protection the code didn't provide, and it contradicted this PR's own analysis one file over. I wrote a test that installs an anonymous principal from `authenticateAsUser` (the deleted-owner shape) and asserted the trigger must not be evaluated: it failed with `Expecting code to raise a throwable`, and `evaluateTrigger` was reached under the anonymous principal exactly as you described. Only the unreachable-auth-service shape threw.

`handleUser` now verifies after impersonating that the installed principal really is the intended owner, and raises `PlatformAuthenticationException` when it is not — so both failure shapes reach the same disposition, and it is the fail-the-event one you identified as correct here rather than `EventListener`'s degrade. The javadoc is reworded to describe what the code now does.

I kept `catch (ValidationException)` as-is: it predates this change and covers a different path, so narrowing it belongs in its own change.

src/main/java/com/otilm/core/events/EventHandler.java
lubomirw: I tested this one and it doesn't reproduce — the mechanism doesn't hold, so I've left the code as-is.

`putActorInfoWhenNull`'s name is misleading: the 3-arg overload `authenticateAsUser` calls writes each field unconditionally when the *argument* is non-null, rather than only when the MDC value is null.

```java
public static void putActorInfoWhenNull(ActorType actorType, String actorUuid, String actorName) {
    if (actorType != null) MDC.put(LOG_ACTOR_TYPE, actorType.name());
    if (actorUuid != null) MDC.put(LOG_ACTOR_UUID, actorUuid);
    if (actorName != null) MDC.put(LOG_ACTOR_NAME, actorName);
}
```

So the second iteration overwrites the actor uuid with creator B's rather than inheriting A's. I added `actorAttributionDoesNotCarryBetweenAssociationOwners` (two associations, different owners, recording the actor uuid seen inside each iteration) and it asserts `containsExactly(firstOwner, secondOwner)` — passing.

Worth noting it only passes for the right reason after I made the test double faithful: my stub was installing the principal without writing the actor MDC, unlike production. With the naive stub the test failed for a reason that had nothing to do with your point, which is what sent me to check the helper's body in the first place — so the comment was still worth raising.

One real gap it leaves: the actor *name* is only ever cleared at loop entry, and `authenticateAsUser` passes a null name, so no name is set during impersonation at all. That's fine today but relies on the null-name argument. Happy to add an explicit `clearActorInfo()` in `handleUser` alongside the SecurityContext switch if you'd rather not depend on that.

PR #1915, opened 2026-07-30
Add auditor system role and harden write and role-assignment authorization

### LINKED ISSUE #1911: Add auditor system role with automatically maintained read-only permissions
### Description

Depends on OmniTrustILM/interfaces#794 (`AccessType` classification on `ResourceAction` + the new `CERTIFICATE:REGISTER` action).

The platform has no read-only role. Anyone who needs full visibility without change rights is given `admin` or `superadmin` — both seeded with `allowAllResources = true` — or a hand-maintained custom role that silently falls behind. Add a system role `auditor` that grants every read action on every resource and no mutating action. It has no system user and is assignable to human users.

**Mechanism today.** Roles, users and permissions live in the auth service (schema `auth`), not in core's database. Core creates them through Flyway *Java* migrations that call the auth REST API — `createAcmeRole()`, `createScepRole()`, most recently `V202607031200__CreateAttributeContentResolverUserAndPermissions.java` on top of `DatabaseAuthMigration`. `system_role = true` blocks edits in the auth service (`RoleService.cs:39`, `:47`), in core (`RoleManagementServiceImpl.checkSystemRole()`) and on the frontend role detail page.

**Why not the acme/scep pattern.** The resource/action *catalog* already syncs itself from the `@ExternalAuthorization` annotation scan on every startup (`ContextRefreshListener` → `AuthResourceSynchronizer` → auth `ResourceService.SyncResourcesAsync`), but stored permission *rows* do not. A statically seeded auditor would lose coverage with every new resource or read action, and each gap would need a top-up migration — `V202308050825__UpdateAcmeScepRolesPermissions.java` exists for exactly that drift.

**Approach.** Bootstrap the role with a Flyway Java migration modelled on `V202607031200__…` (`systemRole = true`, no system user, checksum registered in `DatabaseMigration.JavaMigrationChecksums`, name added to `AuthHelper`). Then extend `AuthResourceSynchronizer.register()` so that, immediately after `syncResources`, it rebuilds the auditor permission payload from `ContextRefreshListener.getResources()` filtered with `ResourceAction.isGrantableToReadOnlyRole()` and full-replaces it through the auth API. Idempotent, safe with multiple replicas, and new resources or read actions are picked up on the next startup with no migration. Use the predicate rather than comparing `AccessType` directly — `!= WRITE` would admit both sensitive reads and the `NOT_GRANTABLE` sentinels, and a sentinel row makes the `savePermissions` call fail with `Unknown action` (`PermissionService.cs:126`). `SENSITIVE_READ` is therefore excluded: the auditor sees settings and metadata, but neither secret payloads (`GET_SECRET_CONTENT`) nor proxy installation instructions (`GET_PROXY_INSTALLATION`), which embed a client secret, a broker SAS key and a non-expiring token. No OPA policy change and no frontend change — the role ends up with ordinary permission rows and renders read-only like every other system role.

**Authorization is two-layer, and the derivation must respect it.** Most mutating operations are gated by a read action on the entry method *plus* a mutating action enforced through a proxy-routed nested call: the empty probe methods `CertificateServiceImpl.checkIssuePermissions` (:1603), `checkRenewPermissions` (:1608), `checkRevokePermissions` (:1613), and `submitCertificateRequest` (:1716, `CERTIFICATE:CREATE`). The v2 certificate lifecycle is therefore **not** exposed to a reads-only principal: `issueCertificate` (:431) reaches `CERTIFICATE:CREATE` via `certificateService.submitCertificateRequest` at :389, and async execution re-authenticates as the original requester (`ActionsListener:81`), so `CERTIFICATE:ISSUE` (:1095) binds as well. The same holds for `renewCertificate` (:1667), `rekeyCertificate` (:1823), `revokeCertificate` (:2020) and the manual paths at :2648, :2789 and :2903–:2907. `AuthHelper.runAsSystem` has one production call site (`OperationAttributeResolver:57`) and neutralizes none of these.

`CallbackServiceImpl.callback` (:167, :175) stays gated on `CONNECTOR:ANY`; the auditor's read rows satisfy it through rego's `AtLeastOneActionOnResource` (`method_policy.rego:94`), which is accepted. The approval path also stays as it is: the probes run only inside `if (!isApproved)`, so on approval-profiled RA profiles the auditor can initiate a revoke or an issue-of-existing request, which creates an attributable Approval object that executes only when a different privileged human approves it (`ApprovalServiceImpl:215-217` excludes the creator). The auditor deny test therefore covers the non-approval paths.

### Use Case

Four gates need fixing before the role is correct, since each one lets a read permission reach a write.

**G1 — `registerCertificate` (v2 `ClientOperationServiceImpl:481`)** creates placeholder certificate rows (`CertificateServiceImpl.createRegistrationPlaceholder:1149`, unannotated), performs state transitions, calls the connector register operation and writes encrypted challenge rows, all behind `RA_PROFILE:DETAIL` + `AUTHORITY:DETAIL`. Add a fifth probe `checkRegisterPermissions()` (`CERTIFICATE:REGISTER`) alongside the four at `CertificateServiceImpl:1595-1614` and call it as the first statement. Its only caller is `api/v2/client/ClientOperationControllerImpl:172`; no protocol flow uses it — ACME (`AcmeServiceImpl:1207`, `:727`), SCEP (`ScepServiceImpl:591`, `:657`), CMP (`CrmfIrCrMessageHandler:86`, `CrmfKurMessageHandler:111`, `RevocationMessageHandler:251`) and Locations (`LocationServiceImpl:831`, `:847`) all go through issue/renew/revoke.

**G2 — `issueExistingCertificate` (:1578)** commits a CSR attach plus completion attributes via `addCertificateRequestToExisting` (`CertificateServiceImpl:1170`, unannotated, read-scoped RA-profile re-check only) *before* the async ISSUE probe denies. Add a conditional caller-side probe inside `if (registered)` immediately before :1617: when a valid ACTIVE registration challenge was presented (`:1608-1612`) the verified challenge is the authorization; otherwise call `certificateService.checkCreatePermissions()`. An unconditional gate would break challenge-based self-service completion for low-privilege holders, so the condition matters.

**G3 — `CryptographicKeyServiceImpl.editKeyItem` (:1050)** performs a `cryptographicKeyItemRepository.save()` rename behind `CRYPTOGRAPHIC_KEY:DETAIL` + `TOKEN:DETAIL`. Re-gate to `CRYPTOGRAPHIC_KEY:UPDATE`.

**G4 — `TriggerServiceImpl.createTriggerAssociations` (:225)** deletes and inserts association rows behind `TRIGGER:DETAIL`. Re-gate to `TRIGGER:UPDATE`.

Additionally, `checkCreatePermissions()` currently has zero call sites. Wire it as the first statement of v2 `submitCertificateRequest` (:359). The body :359-404 is linear and every pre-:389 exit is a throw, so this denies nobody that :389 does not already deny — it only moves the denial ahead of `generateBase64EncodedCsr`, which today signs a CSR with a platform key before the nested CREATE gate refuses. `CertificateControllerImpl:184` and `ScepServiceImpl:657` already pass the :358 annotation first and are unaffected.

**System-role assignment guard.** System-role *membership* is currently unprotected in both directions. `RoleManagementServiceImpl.updateUsers` (:158) has no `checkSystemRole` call, and the auth service's `AssignUsersAsync` (`RoleService.cs:58`) has none either, so any `ROLE:UPDATE` holder can attach a human to `acme`, `scep` or `cmp`. In the other direction `UserManagementServiceImpl.updateRoles` (:229) and `updateRole` (:237) are gated on `USER:UPDATE` alone with no check on *which* role is being assigned, which makes `USER:UPDATE` equivalent to full platform admin. Since `auditor` must stay human-assignable, a blanket block on system roles is not an option. Add one shared guard used by all three methods:

- a role paired with a system user (`acme`, `scep`, `cmp`, `localhost`, `attribute-content-resolver`) is never assignable to a human user;
- a role with `allowAllResources` (`superadmin`, `admin`) requires the actor to already hold that role;
- every other role, including `auditor`, is assignable as today.

**No top-up migrations.** No seeded system-role grant set is affected by G1, G3 or G4 — `acme`/`scep`/`cmp` never held `certificates:register` (`V202308050825:35-42`, `V202404021100:40-47`) and no seeded grant contains `CRYPTOGRAPHIC_KEY:DETAIL` or `TRIGGER:DETAIL`; `admin`/`superadmin` match rego's `AllResourcesAllowed` before any action check. Custom roles that relied on the old `DETAIL` gates do lose those capabilities, which is covered by a release note rather than a migration: registration now requires `certificates:register`, key item rename requires `cryptographicKeys:update`, and trigger associations require `triggers:update`.

**Upgrade hazard to document.** `SyncResourcesAsync` deletes unreferenced actions (`ResourceService.cs:133-154`). A rolled-back or mixed-version core strips `register` and its grants; re-upgrading recreates the action but not the permission rows. Never run mixed core versions against one auth service. The bootstrap migration must also seed the action before granting it, since `PermissionService.cs:126` rejects unknown action names.

### Acceptance Criteria

- [ ] Flyway Java migration creates the `auditor` role with `systemRole = true` and no system user; checksum registered in `DatabaseMigration.JavaMigrationChecksums` and the name added to `AuthHelper`.
- [ ] `AuthResourceSynchronizer` rebuilds and full-replaces the auditor permission set on every startup, after resource sync, from the annotation scan filtered with `ResourceAction.isGrantableToReadOnlyRole()` — not by comparing `AccessType` directly; `SENSITIVE_READ`, `WRITE` and `NOT_GRANTABLE` are excluded.
- [ ] The derivation is extracted as a pure mapping function so it can be unit-tested without a running context, and it logs a summary when the derived set changes.
- [ ] `auditor` is not editable or deletable through the API or the frontend, and its effective permissions are visible read-only on the role detail page.
- [ ] G1: `checkRegisterPermissions()` added to `CertificateServiceImpl` and called as the first statement of `registerCertificate`.
- [ ] G2: conditional probe added inside `if (registered)` before the `addCertificateRequestToExisting` call, skipped when a verified ACTIVE registration challenge is present.
- [ ] G3: `editKeyItem` re-gated to `CRYPTOGRAPHIC_KEY:UPDATE`.
- [ ] G4: `createTriggerAssociations` re-gated to `TRIGGER:UPDATE`.
- [ ] `checkCreatePermissions()` called as the first statement of v2 `submitCertificateRequest`.
- [ ] Shared system-role assignment guard applied to `RoleManagementServiceImpl.updateUsers`, `UserManagementServiceImpl.updateRoles` and `UserManagementServiceImpl.updateRole`, implementing the three rules above.
- [ ] No changes required in `auth-opa-policies` or `fe-administrator`.
- [ ] Auditor deny integration test: a principal holding exactly the derived read set gets 403 on `registerCertificate`, `editKeyItem`, `createTriggerAssociations`, role permission update and every two-layer entry point (sync entries and direct `*Action` invocations), and 200 on certificate list, audit-log list, role detail and settings detail.
- [ ] Test asserts the auditor holds neither `GET_SECRET_CONTENT` nor `GET_PROXY_INSTALLATION`, and gets 403 on both secret-content and proxy-installation reads.
- [ ] Unit test on the derivation function asserts the payload is exactly the `READ` subset of the scanned catalog.
- [ ] Deny-ordering test for v2 `submitCertificateRequest` confirming the permission failure precedes CSR generation.
- [ ] Assignment-guard tests: attaching a human to `acme` is rejected, assigning `superadmin` without holding it is rejected, assigning `auditor` succeeds.
- [ ] Release note covers the three re-gated actions and the mixed-version upgrade hazard.

lubomirw:
Closes #1911. Part of OmniTrustILM/ilm#307.

> **Blocked on OmniTrustILM/interfaces#796.** This branch compiles only against the `AccessType` / `isGrantableToReadOnlyRole()` / `CERTIFICATE:REGISTER` additions in that PR. CI here fails until it merges and a new `interfaces` `2.18.1-SNAPSHOT` publishes.

Adds a read-only system role `auditor`, and closes the paths that would have let that role — or in some cases any operator — change state through a read permission.

## The auditor role

`AuthResourceSynchronizer` creates the role when it is absent and rebuilds its grants from the `@ExternalAuthorization` annotation scan on every startup, right after the resource-catalogue sync it already performs. `systemRole = true`, **no system user** — unlike acme/scep/cmp it is meant to be assigned to people.

Deriving the grants rather than listing them is the point. The catalogue already tracks the code automatically; stored permissions did not, which is why the acme/scep pattern of a hand-written grant list needs a top-up migration whenever a resource appears — `V202308050825__UpdateAcmeScepRolesPermissions` exists for exactly that drift. Membership of the read set is decided solely by `ResourceAction.isGrantableToReadOnlyRole()`, so classifying a new action on the enum is the only step required.

`SENSITIVE_READ` is excluded, so the auditor sees settings and metadata but neither secret payloads (`secrets:getSecretContent`) nor proxy installation instructions (`proxies:getProxyInstallation`) — the latter embed a client secret, a broker SAS key and a non-expiring token.

**Creation happens at startup rather than in a migration.** That changed during review. A migration runs once: on a name collision it would warn, return, be recorded as applied and never run again, leaving the feature permanently absent with no recovery. Reconciliation retries every boot, so the same collision heals once the name is freed. It also removes an auth-calling Flyway migration, and those abort the entire migration chain when the auth service is unreachable, leaving the schema half-built. A migration is only genuinely needed to create a paired system user, and this role has none.

Three details in the derivation are load-bearing rather than cosmetic:

- A resource whose actions are all excluded is **dropped** from the payload, not emitted with an empty action list. An empty list conventionally means allow-all (`DatabaseAuthMigration.createRole`), so emitting it would grant *every* action on that resource — the exact inverse of the feature. The kernel test for this failed with `allowAllActions=true` before the fix.
- The `ANY` sentinel arrives as scanned data (`ContextRefreshListener.registerResourceActions` records every annotation's code). Emitting it would make `savePermissions` reject the whole call with `Unknown action`, since the auth service never syncs `ANY` into its catalogue.
- A role named `auditor` that is **not** a system role belongs to the deployment and is left completely untouched. The lookup matches by name across all roles precisely so that is distinguishable from the role being absent, rather than creating a duplicate name.

Replicas start together, so several find the role missing and all try to create it. Role names are unique in the auth service, so a creation refused because the name now exists is treated as success and that role is reconciled instead.

The mapping is a pure static function over the scanned catalogue, free of Spring and of the auth client, so the boundary is pinned by unit tests rather than by a running context. Reconciliation failures never prevent startup.

## Writes reachable through a read permission

Each of these let a read permission reach a write. They were found by tracing full call chains rather than entry annotations — the v2 certificate lifecycle *looks* exposed but is genuinely gated one bean deeper, so it is untouched here.

| Method | Was | Now |
|---|---|---|
| `registerCertificate` | `RA_PROFILE:DETAIL` | `+ CERTIFICATE:REGISTER` via a new `checkRegisterPermissions` probe |
| `issueExistingCertificate` CSR attach | ungated | `CERTIFICATE:CREATE`, **only** when no registration challenge verified |
| `editKeyItem` | `CRYPTOGRAPHIC_KEY:DETAIL` | `CRYPTOGRAPHIC_KEY:UPDATE` |
| `createTriggerAssociations` | `TRIGGER:DETAIL` | `TRIGGER:UPDATE` |
| v2 `submitCertificateRequest` | annotation bypassed by self-invocation | previously-unused `checkCreatePermissions` probe wired first |

The CSR-attach gate is conditional deliberately: a verified challenge is itself the holder's authorization, and demanding the operator permission on top would make challenge-based self-service unusable for the low-privilege identities it exists for. Implementing it unconditionally first made the challenge test fail with `AuthorizationDenied`, which is what established the condition is needed.

`submitCertificateRequest`'s probe denies nobody the nested gate would admit — the body is linear and every earlier exit throws. It only moves the denial ahead of `generateBase64EncodedCsr`, which was signing a CSR with a platform key before the nested gate refused.

Note `createTriggerAssociations` is also proxy-routed from discovery creation and the event-settings update, so those paths now require `triggers:update` where `triggers:detail` sufficed.

## Role and user administration

Membership was editable from both directions with no check on the role involved. Any `ROLE:UPDATE` holder could attach a human to the acme/scep/cmp roles, and — the significant one — **`USER:UPDATE` alone was equivalent to full platform administration**, because a single `PUT /v1/users/{self}/roles/{superadmin}` granted it. One guard now serves both directions:

- a role paired with a system user takes no human members, **and** its system user cannot be detached. Membership updates replace the whole list (`AssignUsersAsync` clears then re-adds), so an omitted system user would unpair the role and leave the first rule nothing to recognise on a second call
- a system user holds its own role and no other
- a system user cannot be detached from its role by clearing the user's roles, or by `removeRole` — neither assigns anything, so neither was previously inspected
- a role granting all resources may only be **granted** by someone who already holds that exact role
- a system user cannot be enabled or disabled. The auth service refuses to update or delete a system user but not to disable one, and core checked neither, so `USER:ENABLE` alone could stop ACME, SCEP or CMP
- everything else, `auditor` included, stays assignable

Pairing is read from the auth service rather than from the `systemRole` flag, because `auditor` carries that flag too and must remain assignable.

The all-resources rule applies to **grants only**. A replacement resends every role the user keeps, so re-listing a role the user already holds is not a grant; nor is removing a member from a role. Refusing either would break a role-picker on an everyday edit while preventing no escalation.

The localhost bootstrap that creates the first administrator runs as the `localhost` system user and never holds `superadmin`, so the guard would have refused it on **every fresh install**. Nothing in the suite exercised `LocalAdmin`, so this would have shipped silently. It uses an internal path that skips the guard; the role it assigns is fixed, so the caller cannot steer it.

### Scope of this hardening

This narrows `USER:UPDATE` from admin-equivalent back to managing users. It does **not** solve privilege escalation generally, and does not try to.

Role administration is admin-equivalent by design: anyone holding `ROLE:CREATE` + `ROLE:UPDATE` can author a permission set from nothing, so they can grant themselves any authority — by enumerating resources and actions explicitly if not by the `allowAllResources` flag. No assignment-time rule can change that. Closing it would require a full no-escalation rule (a caller may only grant permissions they already hold), which would also remove the delegated administrator who assigns roles they do not personally hold. That trade was considered and deliberately not taken.

Every rule here is also **core-side only**. The auth API is independently reachable, so membership edits made directly against it bypass all of this. The authoritative guards belong beside the auth service's existing system-user checks — tracked in OmniTrustILM/auth#138.

## Upgrade impact

Permissions that must be granted after upgrade — existing roles get 403 until they are:

| Action | Previously sufficed | Affected operations |
|---|---|---|
| `certificates:register` *(new)* | `raProfiles:detail` + `authorities:detail` | certificate pre-registration |
| `cryptographicKeys:update` | `cryptographicKeys:detail` | renaming a key item |
| `triggers:update` | `triggers:detail` | trigger associations, creating a discovery with triggers, event settings |
| `certificates:create` | `raProfiles:detail` + `authorities:detail` | completing a REGISTERED placeholder without a challenge |

No seeded system role is affected — acme/scep/cmp never held `certificates:register`, and no seeded grant contains `CRYPTOGRAPHIC_KEY:DETAIL` or `TRIGGER:DETAIL`; admin/superadmin match `AllResourcesAllowed` before any action check. Only custom roles need attention, which is why there is no top-up migration.

**Do not run mixed core versions against one auth service.** `SyncResourcesAsync` deletes unreferenced actions, so an older instance strips `certificates:register` and its grants; re-upgrading recreates the action but not the permission rows.

Platform documentation for all of the above is tracked in OmniTrustILM/documentation#341.

## Tests

Every behavioural change was developed test-first against a real failing test. The deny tests use `BaseSpringBootTest.denyResourceAccess`, which stubs OPA per resource/action and runs the real Spring AOP stack, so they exercise proxy-routed nested gates rather than asserting annotations.

- 20 kernel unit tests on the derivation, including drop-empty-resource, `ANY` exclusion and an absent stored resource list
- 12 wiring tests on the reconciliation, covering create-when-absent, leave-a-customer's-role-alone and the concurrent-creation race
- 20 guard tests across both directions, every rule, the permit cases, and the bootstrap bypass
- deny tests for each re-gated method, plus a deny-**ordering** test asserting `generateCsr` is never reached when `CERTIFICATE:CREATE` is denied
- `JavaMigrationChecksumTest` recomputes each registered checksum from source, so editing a Java migration fails the build instead of breaking Flyway validation at startup

`ContextSignatureGuardTest` is unchanged — no new Spring context was purchased.

The full suite last ran clean at 4413 tests, 0 failures, before the final review fixes; the classes touched since have been re-run and CI covers the rest.

## Known limits

- **`getAuthorityCertificateChain` persists on a read path.** It is gated `RA_PROFILE:DETAIL` + `AUTHORITY:DETAIL` and saves `Certificate` and `CertificateContent` rows, so an auditor reading a CA chain causes writes. Raised in review and deliberately left: the persistence is the platform importing the chain in the background as part of the read, not the caller mutating state.
- **The role lookup assumes one page.** The auth service returns every role in one response today — the endpoint ignores paging parameters and asks for 1000 — so beyond 1000 roles the auditor role would look absent on every boot. Core cannot influence it; tracked in OmniTrustILM/auth#139.
- **No live auth service in tests.** No test stubs `/auth/resources/sync`, so `reconcileAuditorRole` is never reached in integration. The kernel is unit-tested hard and the wiring is Mockito, but the actual `savePermissions` acceptance is untested end to end.
- **`CallbackServiceITest` is intermittently red in full-suite ordering**, unrelated to this PR: tests run with `flyway.enabled: false` and `ddl-auto: create-drop` against a shared Testcontainers database, so one context closing drops the schema out from under another cached context. Nothing here fixes it.


src/main/java/com/otilm/core/security/authz/RoleAssignmentGuard.java
```diff
@@ -0,0 +1,153 @@
...
+
+    /** Guards the user -&gt; roles direction: the roles that would end up held by {@code userUuid}. */
+    public void checkRolesAssignableToUser(String userUuid, List<String> roleUuids) {
+        if (roleUuids == null) {
+            return;
+        }
+        for (String roleUuid : roleUuids) {
+            checkAssignment(roleManagementApiClient.getRoleDetail(roleUuid), List.of(userUuid));
+        }
+    }
```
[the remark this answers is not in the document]
lubomirw: Fixed in 7c57486eb — good catch, and it was worse than described.

`updateRoles` now also calls `RoleAssignmentGuard.checkRolesRetainedForUser`, which refuses any replacement that drops a role a system user currently holds; an empty list is just the extreme case of that.

The same gap existed on a path the comment did not cover: `removeRole` had no guard at all, so it could detach the paired role one call at a time rather than by clearing. That now goes through `checkRoleRemovableFromUser`. Both share one core, so the two directions cannot drift.

While checking the surrounding surface I also found that `USER:ENABLE` alone could disable `acme`/`scep`/`cmp` — the auth service refuses to update or delete a system user but not to disable one, and core checked neither. `enableUser`/`disableUser` now refuse a system user; the authoritative guard belongs in the auth service and is filed as OmniTrustILM/auth#138.

Covered by four new tests in `RoleAssignmentGuardITest`: clearing a system user's roles, removing its role, disabling it, and the permit case that a human user is unaffected.

src/main/java/com/otilm/core/security/authz/RoleAssignmentGuard.java
```diff
@@ -0,0 +1,187 @@
...
+                throw new ValidationException(
+                        "Role '%s' belongs to system user '%s' and cannot be removed from it."
+                                .formatted(held.getName(), user.getUsername()));
+            }
+        }
+    }
+
+    private void checkAssignment(RoleDetailDto role, List<String> userUuids) {
+        checkMembers(role, userUuids);
+        if (grantsAllResources(role.getUuid())) {
```
klaraf755: The all-resources rule is bypassable by widening after assigning, because it only evaluates `grantsAllResources` at assignment time. `RoleManagementServiceImpl.addPermissions` guards system roles (`checkSystemRole`) but not permission content, so a caller holding `ROLE:CREATE` + `ROLE:UPDATE` can: create role X with empty permissions, `updateUsers(X, [self])` — passes here, since X is neither allow-all nor paired — then `addPermissions(X, allowAllResources=true)`. Full platform administration, and this rule never fires.

A symmetric check on `addPermissions` would close it: a permission save that sets `allowAllResources=true` on a role that has members requires the caller to hold that role (or the role to have no members yet). Given the PR's stated invariant — an all-resources role can only be granted by a holder — this seems worth either fixing here or filing explicitly as an accepted residual.

klaraf755 (commented):
Reviewed the full diff. The design is sound and the read-gated-write fixes are all genuine; the inline comments cover one residual escalation path, one usability regression, a small robustness fix, and two caveats worth recording.

src/main/java/com/otilm/core/service/impl/UserManagementServiceImpl.java
```diff
@@ -227,6 +235,8 @@ private void clearAuthenticationData(String userUuid, String actionName) {
     @Override
     @ExternalAuthorization(resource = Resource.USER, action = ResourceAction.UPDATE)
     public UserDetailDto updateRoles(String userUuid, List<String> roleUuids) {
+        roleAssignmentGuard.checkRolesAssignableToUser(userUuid, roleUuids);
```
klaraf755, line 238: This treats every UUID in the replacement list as a new assignment, so it refuses legitimate edits on users who already hold an all-resources role. If user U already holds `superadmin`, an operator with `USER:UPDATE` but without superadmin cannot call `updateRoles(U, [superadmin, newRole])` to add a role while retaining the existing ones — yet retaining is not escalation. Distinguishing retained from newly-added roles needs one `getUserDetail` call, which the guard already makes on other paths.

`updateRole` (single add) is a workaround, but the batch endpoint is what a UI role-picker typically uses, so this will surface as a 4xx on an everyday operation.

Related asymmetry: removing a member from `superadmin` via `updateUsers` requires holding it, while removing via `removeRole` doesn't. Removal isn't escalation, so the role-direction check is stricter than its own rationale.

src/main/java/com/otilm/core/auth/ReadOnlyRolePermissions.java
```diff
@@ -0,0 +1,117 @@
...
+            return false;
+        }
+
+        Map<String, String> derivedGrants = new TreeMap<>();
+        for (ResourcePermissionsRequestDto resource : derived.getResources()) {
+            derivedGrants.put(resource.getName(),
+                    grantSignature(resource.getAllowAllActions(), resource.getActions(), resource.getObjects()));
+        }
+        Map<String, String> storedGrants = new TreeMap<>();
+        for (ResourcePermissionsDto resource : stored.getResources()) {
```
klaraf755: NPE if the auth service returns a non-null `SubjectPermissionsDto` with a null `resources` list — the null-`stored` case above is handled and tested, but not this one. It's caught by `reconcileAuditorRole`'s catch-all, so the failure mode is only "role not reconciled this boot + error log", but the DTO comes from a remote service and the fix is a one-line null-coalesce.

src/main/java/com/otilm/core/security/authz/RoleAssignmentGuard.java
```diff
@@ -0,0 +1,187 @@
...
+
+/**
+ * Decides which roles may be attached to which users. Membership is editable from both directions, so both enforce
+ * the same rules — otherwise the weaker endpoint is the way around the stronger one. A role paired with a system
+ * user is refused, as is a role granting all resources unless the caller already holds it; without the latter,
+ * {@code USER:UPDATE} alone would grant full platform administration. Everything else, {@code auditor} included,
+ * stays assignable.
+ */
+@Component
+public class RoleAssignmentGuard {
```
klaraf755, line 29: Same caveat the PR body records for enable/disable (OmniTrustILM/auth#138): every rule in this guard is core-side only, and the auth API is independently reachable, so membership edits made directly against the auth service bypass all of it. Worth folding into auth#138 or filing alongside it so the authoritative checks eventually live next to the auth service's existing system-user guards.

src/main/java/com/otilm/core/auth/AuthResourceSynchronizer.java
```diff
@@ -40,8 +51,107 @@ public void register() {
...
+                AuthHelper.AUDITOR_ROLE_NAME, derived.getResources().size(),
+                ReadOnlyRolePermissions.countActions(derived));
+    }
+
+    /**
+     * Matched by name across every role rather than among the system ones, so a name already taken by a role the
+     * platform did not create is told apart from the role being absent - which would create a duplicate of it.
+     */
+    private RoleDto findRoleNamedAuditor() {
+        RoleWithPaginationDto roles = roleManagementApiClient.getRoles();
```
klaraf755, line 152: `getRoles()` sends no paging parameters and the lookup filters the returned list by name. `RoleWithPaginationDto` hints the endpoint may page; if it ever does, an auditor role beyond page 1 would look absent — triggering a doomed create-then-reconcile cycle on every boot (create refused, race path re-reads the same truncated page, error logged). `LocalAdminServiceImpl.getSuperadminRoleUuid` already relies on the same full-list assumption, so this is precedent-consistent — just noting the inherited risk since this code, unlike that one, runs on every startup.

src/main/java/com/otilm/core/security/authz/RoleAssignmentGuard.java
lubomirw: Valid, and not fixed here — recorded as an accepted residual instead. Reasoning, since the decision matters more than the finding:

The suggested symmetric check on `addPermissions` would close the exact sequence you wrote, but not the capability. The same caller can enumerate every resource and action explicitly rather than setting `allowAllResources`, and arrive at the same authority. `ROLE:CREATE` + `ROLE:UPDATE` means authoring permission sets from nothing, so it was already admin-equivalent before this PR and remains so after. Adding the check would make the guard look like it enforces a boundary it does not — worse than leaving it visible.

What this PR does close is `USER:UPDATE` on its own, which previously reached administration via a single `PUT /v1/users/{self}/roles/{superadmin}`. That is a strictly lower privilege and it is not reachable through your bypass, since a `USER:UPDATE`-only holder can neither create a role nor write permissions into one.

Genuinely closing it needs the no-escalation rule — a caller may only grant permissions they already hold. That was weighed and declined: it removes the delegated administrator who assigns roles they do not personally hold, which is a common pattern. The PR body now states the scope explicitly under "Scope of this hardening" so a later reader does not assume more than is there.

src/main/java/com/otilm/core/service/impl/UserManagementServiceImpl.java
lubomirw: Fixed in 09f44b4de — both halves, and you were right that this would surface on an everyday operation.

The all-resources rule now applies to grants only. `checkRolesAssignableToUser` reads the target user's current roles once and skips any role already held, so `updateRoles(U, [superadmin, newRole])` succeeds for a caller without superadmin when U already had it. Nothing is granted by re-listing a role the user keeps, and a role dropped from the list never entered the loop anyway, so removal was already permitted.

The asymmetry you spotted is fixed too. `checkUsersAssignableToRole` now applies the rule only when the change adds a member the role does not already have, so removing a member no longer demands holding the role — consistent with `removeRole`, and consistent with the rule's own rationale.

Both went red first with `Role 'superadmin' grants access to all resources and can only be assigned by a user who already holds it`, which is exactly the wrong refusal. Four tests pin the boundary: retain-and-add succeeds, grant-to-a-user-without-it still 422s, remove-a-member succeeds, add-a-member still 422s.

One correction to what I said earlier: this costs a second `getUserDetail` on `updateRoles` rather than sharing the one the retention check already makes. Merging them would have meant a fourth public entry point on the guard, and an extra call on an administrative operation was not worth the added surface.

src/main/java/com/otilm/core/auth/ReadOnlyRolePermissions.java
lubomirw: Fixed in 09f44b4de. `matches` now treats an absent stored resource list as no grants, the same as an empty one.

Test-driven rather than patched: the test reproduces your exact failure before the fix —

```
java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()" because the return value of
  "com.otilm.api.model.core.auth.SubjectPermissionsDto.getResources()" is null
```

You were right that the blast radius was small — the catch-all in `reconcileAuditorRole` meant a skipped reconciliation and an error log rather than a failed startup. Still worth closing, since the DTO comes off the wire and the null case is indistinguishable from a role that genuinely holds nothing.

src/main/java/com/otilm/core/security/authz/RoleAssignmentGuard.java
lubomirw: Agreed, and folded into OmniTrustILM/auth#138 rather than left as a PR aside — I commented there widening its scope from enable/disable to role membership generally.

The caveat is broader than the enable/disable case: every rule in this guard is core-side. `RolesController.AssignUsersAsync` and the users/roles endpoints carry no equivalent check, and role sync from JWT claims in `UserService` assigns roles by name from token claims (creating unknown roles when enabled), so there are at least two paths that bypass all of this. The authoritative versions belong beside the existing "Cannot update system user" / "Cannot delete system role" guards.

The PR body now says so explicitly under "Scope of this hardening" instead of only mentioning it for enable/disable.

src/main/java/com/otilm/core/auth/AuthResourceSynchronizer.java
lubomirw: Confirmed and filed as OmniTrustILM/auth#139; the assumption is now documented at the lookup.

Checked the auth side rather than inferring from `RoleWithPaginationDto`: `RolesController.GetRolesAsync` takes no query parameters at all and constructs `new QueryRequestDto()`, which defaults to `Page = 1, PageSize = 1000`. So the endpoint does not page today, and core cannot make it — there is no parameter to send. Truncation begins past 1000 roles, and your description of the failure is right: the role looks absent, every boot attempts a creation the auth service refuses, and an error is logged.

Your read on the precedent is also right, and the difference you drew is the one that matters — `LocalAdminServiceImpl.getSuperadminRoleUuid` runs once at bootstrap, this runs on every startup. #139 covers accepting paging parameters (or dropping the paged contract if all-in-one-response is intended) and reviewing the other list endpoints for the same shape.

3keyroman (approved):

PR #1916, opened 2026-07-30
Fill certificate request entity identity from CSR data

### LINKED ISSUE #1853: Certificates - Details - Request tab: Common Name stays displayed as "<empty>" for a pre-registered certificate after it has been issued
### Description

For a certificate created through the **Pre-register** flow, after the registration is completed and the certificate reaches `State: Issued`, the **Request** tab of the certificate detail page still shows the **Common Name** as `<empty>` — the initial pre-registration placeholder value — instead of the actual CN of the now-issued certificate.

The certificate itself carries the correct CN in its Subject (visible on the Details tab, in the downloaded PEM), but the *Request tab* preserves the pre-registration snapshot rather than reflecting the request that completed the issuance.

### Steps to Reproduce

1. Navigate to **Certificates**
2. Click **Add Certificate → Pre-register**
3. Fill in the necessary info and click **Create** — this creates a pre-registered placeholder certificate
4. Complete the registration for that pre-registered certificate (attach CSR / issue / whatever finishes the flow) — the cert reaches `State: Issued`
5. Open the resulting certificate's detail page → switch to the **Request** tab
6. Observe: **Common Name = `<empty>`** even though the issued cert has a proper CN visible on the Details tab

### Expected Behavior

- Once the pre-registered cert is fully issued, the Request tab reflects the CN of the completed request / issued cert (or at least does not show `<empty>` in a way that suggests missing data)
- Request tab data stays consistent with the Details tab / actual cert contents after issuance completes

### Actual Behavior

- Request tab shows `Common Name: <empty>` indefinitely for pre-registered certs, even after they have been issued and carry a valid CN elsewhere
- User sees conflicting information across tabs on the same certificate detail page

### Severity

Minor

### Module

Certificates

### Attachments

Image

### Logs

_No response_

### Environment

ILM Core: 2.18.1-SNAPSHOT (21 Jul 2026, 10:31)
Commit: 6fd2f2c
Branch: main
Build time: 2026-07-21T08:31:05Z
ILM Frontend: 2.18.0 (20 Jul 2026, 23:47)
Client:
- macOS Tahoe 26.5.1

klaraf755:
Fixes #1853

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
@@ -1243,8 +1243,10 @@ public class CertificateServiceImpl implements CertificateExternalService, Certi
         final String fingerprint = CertificateUtil.getThumbprint(normalizedDer);
         CertificateRequestEntity certificateRequestEntity = certificateRequestRepository.findByFingerprint(fingerprint).orElse(null);
         if (certificateRequestEntity == null) {
-            certificateRequestEntity = certificate.prepareCertificateRequest(issueRequest.getFormat());
+            certificateRequestEntity = new CertificateRequestEntity();
+            CertificateUtil.prepareCertificateRequestEntityFromCsr(certificateRequestEntity, request);
             certificateRequestEntity.setCertificateType(CertificateType.X509);
+            certificateRequestEntity.setCertificateRequestFormat(issueRequest.getFormat());
             certificateRequestEntity.setFingerprint(fingerprint);
             certificateRequestEntity.setContent(Base64.getEncoder().encodeToString(normalizedDer));
             setCertificateRequestEntitySignatureAlgorithms(request, certificateRequestEntity);
@@ -1810,7 +1812,7 @@ public class CertificateServiceImpl implements CertificateExternalService, Certi
 
         Certificate certificate = new Certificate();
         // prepare certificate request data for certificate
-        CertificateUtil.prepareCsrObject(certificate, request);
+        CertificateUtil.prepareCertificateFromCsr(certificate, request);
 
         certificate.setState(CertificateState.REQUESTED);
         certificate.setComplianceStatus(ComplianceStatus.NOT_CHECKED);
--- a/src/main/java/com/otilm/core/util/CertificateUtil.java
+++ b/src/main/java/com/otilm/core/util/CertificateUtil.java
@@ -577,9 +577,13 @@ public class CertificateUtil {
                     )
             );
         }
+        // CertTemplate.subject is OPTIONAL in CRMF, so the parsed name can be null; both columns are
+        // nullable and stay empty for a subject-less request (identity carried in the SAN).
         X500Name subjectDN = X500Name.getInstance(new PlatformX500NameStyle(false), certificateRequest.getSubject());
-        modal.setSubjectDn(subjectDN.toString());
-        modal.setCommonName(getCommonNameFromSubjectDn(subjectDN));
+        if (subjectDN != null) {
+            modal.setSubjectDn(subjectDN.toString());
+            modal.setCommonName(getCommonNameFromSubjectDn(subjectDN));
+        }
         if (getKeyAlgorithmEnumFromProviderName(certificateRequest.getPublicKey().getAlgorithm()) != null) {
             modal.setPublicKeyAlgorithm(getKeyAlgorithmStringFromProviderName(certificateRequest.getPublicKey().getAlgorithm()));
         }
--- a/src/main/java/com/otilm/core/util/CertificateUtil.java
+++ b/src/main/java/com/otilm/core/util/CertificateUtil.java
@@ -12,6 +12,7 @@ import com.otilm.api.model.core.settings.CertificateValidationSettingsDto;
 import com.otilm.api.model.core.settings.PlatformSettingsDto;
 import com.otilm.api.model.core.settings.SettingsSection;
 import com.otilm.core.dao.entity.Certificate;
+import com.otilm.core.dao.entity.CertificateRequestEntity;
 import com.otilm.core.dao.entity.DiscoveryCertificate;
 import com.otilm.core.model.request.CertificateRequest;
 import com.otilm.core.model.request.CrmfCertificateRequest;
@@ -33,18 +34,11 @@ import org.bouncycastle.asn1.x500.X500Name;
 import org.bouncycastle.asn1.x509.*;
 import org.bouncycastle.asn1.x509.qualified.ETSIQCObjectIdentifiers;
 import org.bouncycastle.asn1.x509.qualified.QCStatement;
-import org.bouncycastle.cert.X509CertificateHolder;
-import org.bouncycastle.cert.X509v3CertificateBuilder;
-import org.bouncycastle.cert.jcajce.JcaX509CertificateConverter;
 import org.bouncycastle.cert.jcajce.JcaX509CertificateHolder;
-import org.bouncycastle.cert.jcajce.JcaX509v3CertificateBuilder;
 import org.bouncycastle.jcajce.provider.asymmetric.x509.CertificateFactory;
 import org.bouncycastle.jce.provider.BouncyCastleProvider;
 import org.bouncycastle.openssl.PEMParser;
-import org.bouncycastle.operator.ContentSigner;
 import org.bouncycastle.operator.DefaultAlgorithmNameFinder;
-import org.bouncycastle.operator.OperatorCreationException;
-import org.bouncycastle.operator.jcajce.JcaContentSignerBuilder;
 import org.bouncycastle.util.io.pem.PemObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -52,7 +46,6 @@ import org.slf4j.LoggerFactory;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.StringReader;
-import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
 import java.security.*;
 import java.security.cert.CertificateEncodingException;
@@ -62,7 +55,6 @@ import java.security.cert.X509Certificate;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.X509EncodedKeySpec;
 import java.util.*;
-import java.util.concurrent.ThreadLocalRandom;
 import java.util.stream.Collectors;
 
 public class CertificateUtil {
@@ -556,7 +548,7 @@ public class CertificateUtil {
     }
 
 
-    public static void prepareCsrObject(Certificate modal, CertificateRequest certificateRequest) throws NoSuchAlgorithmException, CertificateRequestException {
+    public static void prepareCertificateFromCsr(Certificate modal, CertificateRequest certificateRequest) throws NoSuchAlgorithmException, CertificateRequestException {
         if (certificateRequest.getPublicKey() == null) {
             throw new ValidationException(
                     ValidationError.create(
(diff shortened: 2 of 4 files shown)


src/main/java/com/otilm/core/util/CertificateUtil.java
```diff
@@ -643,16 +652,20 @@ private static String formatOtherNameSan(String oid, String value) {
         return "%s=%s".formatted(oid, value);
     }
 
-    private static void setSubjectDNParams(Certificate modal, X500Name subjectDN) {
-        modal.setSubjectDn(subjectDN.toString());
-
-        for (RDN i : subjectDN.getRDNs()) {
+    private static String getCommonNameFromSubjectDn(X500Name subjectDn) {
```
lubomirw: **[Medium]** | Correctness

This extraction changes the multi-CN tie-break for every certificate the platform parses, not just the request entity this PR targets. The loop it replaced had no `break`, so it kept overwriting and effectively took the **last** matching CN in `getRDNs()` order; this helper returns on the **first** match. Because `PlatformX500NameStyle(false).toString` reverses the RDN array before rendering, the stored `commonName` flips from the CN that appears *first* in the persisted `subjectDn` to the one that appears *last*.

Against the three-CN fixture already in the repo (`PlatformX500NameStyleITest:34`, which renders `..., CN=Xyzzy, CN=Bcd, CN=Abc`) the value changes from `Xyzzy` to `Abc`. `setSubjectDNParams` is on the main ingestion path — call sites at `CertificateUtil:359` (parsed X509), `:420`, `:559` (`prepareCertificateFromCsr`) and `:612` — so this silently alters every uploaded, discovered, issued and registered certificate with more than one CN. Nothing asserts it today.

_Suggested:_ restore last-wins (keep scanning and assign, returning the final match) so `commonName` still matches the leading CN of the rendered `subjectDn`. If first-wins is the deliberate new semantics, state that in the method Javadoc and add a test covering the multi-CN fixture.

<sub>flagged by superpowers code review; confirmed against the repo fixture during triage</sub>

src/main/java/com/otilm/core/util/CertificateUtil.java
```diff
@@ -577,6 +569,23 @@ public static void prepareCsrObject(Certificate modal, CertificateRequest certif
...
+    public static void prepareCertificateRequestEntityFromCsr(CertificateRequestEntity modal, CertificateRequest certificateRequest) throws NoSuchAlgorithmException, CertificateRequestException {
+        if (certificateRequest.getPublicKey() == null) {
+            throw new ValidationException(
+                    ValidationError.create(
+                            "Invalid public key in certificate request"
+                    )
+            );
+        }
+        X500Name subjectDN = X500Name.getInstance(new PlatformX500NameStyle(false), certificateRequest.getSubject());
+        modal.setSubjectDn(subjectDN.toString());
```
lubomirw: **[Medium]** | Correctness

`CertTemplate.subject` is OPTIONAL in CRMF and `CrmfCertificateRequest.getSubject()` returns it unguarded (`CrmfCertificateRequest.java:49`), while BouncyCastle's `X500Name.getInstance(style, null)` returns `null` — so `subjectDN.toString()` throws NPE. `addCertificateRequestToExisting` accepts `CertificateRequestFormat.CRMF` (`CertificateRequestUtils:118`) with no upstream subject validation, and the previous path (`certificate.prepareCertificateRequest`) never touched the subject at all — it copied the placeholder's identity. A subject-less CRMF attached to a REGISTERED placeholder therefore surfaces as an unhandled 500 on a path that previously succeeded.

_Suggested:_ null-guard the parsed name — both columns are nullable:

```java
X500Name subjectDN = X500Name.getInstance(new PlatformX500NameStyle(false), certificateRequest.getSubject());
if (subjectDN != null) {
    modal.setSubjectDn(subjectDN.toString());
    modal.setCommonName(getCommonNameFromSubjectDn(subjectDN));
}
```

Or reject it with a `ValidationException` naming the missing subject, mirroring the public-key check just above. Either way, add a CSR-attach test covering CRMF — the new tests exercise PKCS10 only.

<sub>flagged independently by Copilot CLI, guided-pr-review and superpowers code review</sub>

src/main/java/com/otilm/core/util/CertificateUtil.java
```diff
@@ -577,6 +569,23 @@ public static void prepareCsrObject(Certificate modal, CertificateRequest certif
         modal.setSubjectAlternativeNames(CertificateUtil.serializeSans(certificateRequest.getSubjectAlternativeNames()));
     }
 
+    public static void prepareCertificateRequestEntityFromCsr(CertificateRequestEntity modal, CertificateRequest certificateRequest) throws NoSuchAlgorithmException, CertificateRequestException {
```
lubomirw, line 565: **[Low]** | Maintainability

This repeats `prepareCertificateFromCsr` almost verbatim — same public-key null check and message, same `PlatformX500NameStyle` subject parse, same CN extraction, same key-algorithm guard, same SAN serialization — and inlines the subject handling instead of going through `setSubjectDNParams`. The two identity-population paths can now drift silently: a null-subject or DN-normalization fix applied to one will miss the other, and any subject-derived field added to `setSubjectDNParams` later will not reach the request entity.

_Suggested:_ extract one private helper that derives the shared values (subject DN string, common name, public-key algorithm, serialized SANs) and have both public methods apply that result to their own target, leaving only the `Certificate`-specific `publicKeyFingerprint` and `keySize` in `prepareCertificateFromCsr`.

src/main/java/com/otilm/core/util/CertificateUtil.java
```diff
@@ -577,6 +569,23 @@ public static void prepareCsrObject(Certificate modal, CertificateRequest certif
...
+        X500Name subjectDN = X500Name.getInstance(new PlatformX500NameStyle(false), certificateRequest.getSubject());
+        modal.setSubjectDn(subjectDN.toString());
+        modal.setCommonName(getCommonNameFromSubjectDn(subjectDN));
+        if (getKeyAlgorithmEnumFromProviderName(certificateRequest.getPublicKey().getAlgorithm()) != null) {
+            modal.setPublicKeyAlgorithm(getKeyAlgorithmStringFromProviderName(certificateRequest.getPublicKey().getAlgorithm()));
+        }
+        modal.setSubjectAlternativeNames(CertificateUtil.serializeSans(certificateRequest.getSubjectAlternativeNames()));
+    }
+
     private static void setIssuerDNParams(Certificate modal, X500Name issuerDN) {
```
lubomirw, line 601: **[Low]** | Maintainability

The refactor extracted the CN scan for the subject but left the byte-identical loop here in `setIssuerDNParams` untouched, so after this PR `issuerCommonName` still resolves last-CN-wins while `commonName` resolves first-CN-wins — two copies of the same logic that no longer agree.

_Suggested:_ rename the new helper to `getCommonNameFromDn(X500Name)` and call it from here too, so both sides share one definition of "the" CN. This disappears on its own if the tie-break comment above is resolved by sharing the helper.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1243,8 +1243,10 @@ public UUID addCertificateRequestToExisting(UUID certificateUuid, ClientCertific
         final String fingerprint = CertificateUtil.getThumbprint(normalizedDer);
         CertificateRequestEntity certificateRequestEntity = certificateRequestRepository.findByFingerprint(fingerprint).orElse(null);
         if (certificateRequestEntity == null) {
-            certificateRequestEntity = certificate.prepareCertificateRequest(issueRequest.getFormat());
+            certificateRequestEntity = new CertificateRequestEntity();
```
lubomirw, line 1246: **[Low]** | Correctness

Minor, raised for awareness rather than as a defect. The old `certificate.prepareCertificateRequest(...)` also set `keyUsage` from the placeholder (`Certificate.java:319`); the new construction never sets it, so it stays at the primitive default `0`. That persists fine (`@Column(nullable = false) int`), no reader of `CertificateRequestEntity.getKeyUsage()` exists in `src/main/java`, and copying a placeholder's key usage was arguably meaningless anyway — so this looks harmless.

_Suggested:_ no change needed unless the field is meant to carry the CSR's key-usage extension, in which case derive it from the request rather than leaving it at `0`. (`complianceStatus` is fine — the field initializer already defaults it to `NOT_CHECKED`.)

<sub>raised during triage while diffing the new construction against `Certificate.prepareCertificateRequest`</sub>

src/test/java/com/otilm/core/integration/service/CertificateServiceITest.java
```diff
@@ -2081,6 +2081,30 @@ void attachesDoubleBase64EncodedCsr_normalizesToPlainDerBeforeStorage() throws E
             CertificateRequestEntity stored = certificateRequestRepository.findByFingerprint(expectedFingerprint).orElseThrow();
             assertThat(stored.getContent()).isEqualTo(Base64.getEncoder().encodeToString(expectedDer));
         }
+
+        @Test
+        void attachesCsr_requestEntityIdentityComesFromCsrNotFromPlaceholder() throws Exception {
```
lubomirw, line 2086: **[Nit]** | Testing

The test proves the request entity now carries the CSR identity, but never asserts the other half of the invariant the code under test documents ("the registration identity already on the row is intentionally left untouched") — that the certificate row still holds `CN=placeholder-cn`.

_Suggested:_ reload the certificate after the call and assert `getCommonName()` / `getSubjectDn()` are still the placeholder values, so a future change that copies the CSR identity onto the placeholder row is caught here.

src/test/java/com/otilm/core/integration/service/CertificateServiceITest.java
```diff
@@ -2081,6 +2081,30 @@ void attachesDoubleBase64EncodedCsr_normalizesToPlainDerBeforeStorage() throws E
...
+            // given — a REGISTERED placeholder whose pre-registration identity differs from the CSR
```
lubomirw, line 2087: **[Low]** | Maintainability

The `// given` block narrates change history ("before the fix the request entity copied the placeholder's identity, so the Request tab kept showing the pre-registration snapshot") rather than describing the scenario. Once merged, "the fix" has no referent and the comment reads as a changelog entry inside the test.

_Suggested:_ keep the invariant, drop the history clause — e.g. `// given — a REGISTERED placeholder whose pre-registration identity differs from the CSR subject; copying the placeholder identity would leave the Request tab showing a stale snapshot`. The "before the fix" framing belongs in the commit message or PR body.

src/test/java/com/otilm/core/integration/service/CertificateServiceITest.java
```diff
@@ -2081,6 +2081,30 @@ void attachesDoubleBase64EncodedCsr_normalizesToPlainDerBeforeStorage() throws E
...
+                    aCertificate().withRaProfile(raProfile).withState(CertificateState.REGISTERED)
+                            .withCommonName("placeholder-cn").withSubjectDn("CN=placeholder-cn").build());
+            ClientCertificateIssueRequestDto issueRequest = new ClientCertificateIssueRequestDto();
+            issueRequest.setRequest(SAMPLE_PKCS10);
+            issueRequest.setFormat(CertificateRequestFormat.PKCS10);
+
+            // when
+            certificateService.addCertificateRequestToExisting(registered.getUuid(), issueRequest);
+
+            // then — the stored request carries the CSR's subject (CN=new_cert), not the placeholder
```
lubomirw: **[Nit]** | Maintainability

The `// then` comment restates the three assertions directly below it and repeats the expected literal `CN=new_cert`, so the prose duplicates the asserts and will drift if the fixture subject changes.

_Suggested:_ shorten to intent only — `// then — identity and format come from the CSR, not the placeholder`.

lubomirw (commented):
Automated multi-reviewer triage — 8 findings (COMMENT). Sources: Copilot CLI, guided-pr-review, superpowers code review, hygiene scan; each finding was re-verified against the working tree before posting. The missing-backfill point (existing `certificate_request` rows keep the pre-registration identity) was deliberately left out of this review.

src/main/java/com/otilm/core/util/CertificateUtil.java
klaraf755: Restored the original last-wins order and added a test asserting this order.

src/main/java/com/otilm/core/util/CertificateUtil.java
klaraf755: Added  a null guard and a test for CRMF.

src/main/java/com/otilm/core/util/CertificateUtil.java
klaraf755: Used shared helper.

src/main/java/com/otilm/core/util/CertificateUtil.java
klaraf755: Added a shared record and common method.

src/test/java/com/otilm/core/integration/service/CertificateServiceITest.java
klaraf755: Dropped the history clause.

src/test/java/com/otilm/core/integration/service/CertificateServiceITest.java
klaraf755: Updated the comment.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
klaraf755: Agreed, no change. The column is effectively dead on this path: nothing in `src/main` reads `CertificateRequestEntity.getKeyUsage()` (the detail mapper only maps the certificate's own key usage, not the request's), and the old code never stored a meaningful value either — it copied `keyUsage` from the certificate row, which is always 0 at that point on both creation paths (the registration placeholder never sets it, and the submit flow copies from a freshly constructed `Certificate`). So the persisted value is unchanged (0 before, 0 now). Deriving the CSR's requested key-usage extension would be a new feature (parse the `extensionRequest` attribute + expose it in the DTO), not a restoration of anything this PR dropped.

src/test/java/com/otilm/core/integration/service/CertificateServiceITest.java
klaraf755: Added an assertion.

lubomirw (approved):

PR #1920, opened 2026-07-30
Harden notification delivery and recipient resolution

lubomirw:
Discovery runs with triggers were logging `UnexpectedRollbackException` and a connector rejection on every notification. Three distinct causes.

## Rollback-only poisoning

`NotificationListener` is class-level `@Transactional`. `NotificationServiceImpl` is too, so it joins that transaction rather than starting one. A group or role resolving to zero users threw `ValidationException`, the inner interceptor marked the shared transaction rollback-only, and the listener's own catch could not clear it — commit then failed with `UnexpectedRollbackException`.

Everything written in that transaction went with it: notifications already created for earlier recipients of the same event (owner is added before groups), `PendingNotification` repetition counters, and the trigger history record written by the error handler to explain the failure.

- A group or role with no members is ordinary configuration, so `createNotificationForUsers` returns `null` instead of throwing.
- Internal notifications and trigger-history writes run in their own transactions, so one recipient cannot take the rest down.

Isolation is scoped at the call site via the existing `TransactionHandler`, matching the pattern established in #1913. Making `TriggerServiceImpl.createTriggerHistoryRecord` itself `REQUIRES_NEW` was rejected: `TriggerEvaluator` writes records while their parent `TriggerHistory` is still uncommitted, so a suspended transaction would fail the foreign key.

## Recipient type NONE stopped delivering

NONE is recipient-less by design — a webhook instance posts to the URL configured on the instance. The empty-recipients guard added in #1917 could not tell that apart from a profile whose recipients genuinely failed to resolve, so NONE profiles stopped reaching their connector entirely.

The suppression is now keyed on the provider kind rather than the recipient type, so NONE (and any recipient-less profile) still reaches a provider that can deliver without recipients.

## Connector called with nothing to deliver

An OBJECT recipient whose mapped attributes resolve to nothing is skipped per-recipient, and a NONE recipient has no delivery details at all. Whether an empty recipient list can still be delivered is a property of the provider, not something this listener can infer from the instance kind: a webhook posts to the URL on its own instance and ignores recipients, while an e-mail provider rejects a request with no address and says so.

The notification is therefore handed to the connector either way, and the provider decides — guessing "cannot deliver" from the kind string would lose a webhook notification silently. Recipients that genuinely failed to resolve are reported as a gap (on trigger history for the external path, counted and reported for the internal path), so a recipient that never reaches anyone is visible rather than dropped without trace. A NONE recipient is counted as prepared-by-design and does not raise that gap report.

Two integration tests assert the connector is still called with an empty recipient list (webhook NONE and webhook OBJECT); they pin this deliberate behaviour.

## Review follow-ups (klaraf755)

- Internal-notification success no longer overwrites external success when deciding whether the repetition counter advances (`notificationSent = notificationSent || outcome.notified() > 0`). Previously a monitoring profile with a memberless group but a working e-mail instance flipped `notificationSent` back to `false`, so `PendingNotification.repetitions` never advanced and the e-mail re-fired on every subsequent event past its configured limit.
- A NONE profile that still carries recipient UUIDs (a type edited from USER/GROUP/ROLE with the list kept) resolves to NONE-typed entries that produce no delivery DTO. Those are now counted as prepared-by-design, so the profile no longer writes a "could not prepare all of its recipients" failure to trigger history on every event.
- Reverted the class annotation to plain `@Transactional`. `processMessage` — the only proxied entry point — catches every exception in both branches, so nothing crosses the proxy for `rollbackFor = Exception.class` to act on; the `notificationSent` flag is the real guard. The Javadoc that asserted the rollback protection was dropped rather than left to rot.
- Save-time validation of recipient configuration (naming types require recipients, NONE must carry none, NONE + e-mail rejected) was considered and deferred: the connector-decides delivery model plus the gap reporting above handle misconfiguration at delivery time without adding a new validation surface. An earlier draft of this description claimed that validation existed — it does not, and the description has been corrected to match the code.

## Verification

`NotificationObjectRecipientITest`, `NotificationProfileServiceITest`, `NotificationServiceITest`, `NotificationListenerTest`, `NotificationServiceImplEmptyRecipientsTest`, plus `EventHandlersITest`, `TriggerEvaluatorITest`, `EventServiceITest`, `TriggerServiceITest`, `NotificationInstanceServiceITest`, `ContextSignatureGuardTest`, `CertificateDiscoveredEventHandler*Test`, `NotificationProducerTest`, `NotificationProfileServiceImplPersistEditTest`.

Review-fix reverification: `NotificationObjectRecipientITest` (6), `NotificationInternalNotificationITest` (3), `EventHandlersITest#testEventDataNotifications` (1) — all passing after the changes above.

The NONE regression test was confirmed to fail against the pre-fix guard before the fix was restored.


src/main/java/com/otilm/core/messaging/jms/listeners/NotificationListener.java
```diff
@@ -443,22 +483,21 @@ private void sendInternalNotifications(List<NotificationRecipient> recipients, I
...
-                                recipient.getRecipientUuid().toString(),
-                                resource, objectUuid != null ? objectUuid.toString() : null);
-                default ->
-                        throw new ValidationException("Unhandled recipient type for internal notification: " + recipient.getRecipientType());
-            }
+            // Each recipient commits on its own. Anything unchecked raised while creating one would otherwise mark
+            // the shared transaction rollback-only past the caller's catch, discarding the notifications already
+            // created for the recipients before it.
+            transactionHandler.runInNewTransaction(() -> createInternalNotification(recipient, notificationData, resource, objectUuid));
+        }
```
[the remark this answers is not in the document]
lubomirw: Fixed in 01dd9cde6. The loop body now sits in a per-recipient `try`/`catch` that counts the failure, logs the recipient type and UUID, and continues, so recipients after a failing one are still attempted.

Two follow-ups since: the failure count is reported on trigger history rather than only in the log (71f9016a3), matching how the external path reports its equivalent gap; and `NotificationInternalNotificationITest.recipientFailingInsideATransactionalCollaboratorIsIsolatedFromTheOthers` now covers this — a ROLE whose auth response carries a malformed user UUID sits between two USER recipients, and the test asserts both users keep their notification. Verified it fails with `UnexpectedRollbackException` when the isolation is removed.

src/main/java/com/otilm/core/messaging/jms/listeners/NotificationListener.java
```diff
@@ -138,8 +149,9 @@ private void sendByNotificationProfile(UUID notificationProfileUuid, Notificatio
         // send internal notification when not Default recipient type. Default internal notifications for events are sent in corresponding event handlers
         if (sendInternalNotifications) {
             try {
-                sendInternalNotifications(recipients, getInternalNotificationData(message), message.getResource(), message.getObjectUuid());
-                notificationSent = true;
+                InternalNotificationOutcome outcome = sendInternalNotifications(recipients, getInternalNotificationData(message), message.getResource(), message.getObjectUuid());
+                notificationSent = outcome.notified() > 0;
```
klaraf755: This overwrites the external result instead of combining with it. A monitoring profile with both an e-mail instance and internal notifications enabled, whose group has an e-mail address but no members: external delivery succeeds, internal notifies zero → `notificationSent` flips to `false` → `PendingNotification.repetitions` never increments → the external e-mail re-fires on every subsequent monitoring event, past its configured repetition limit.

```suggestion
                notificationSent = notificationSent || outcome.notified() > 0;
```

src/main/java/com/otilm/core/messaging/jms/listeners/NotificationListener.java
```diff
@@ -353,6 +421,7 @@ private void sendExternalNotifications(UUID notificationInstanceUUID, List<Notif
             logger.error("Cannot send notification to connector: {}", e.getMessage());
             throw e;
         }
+        return recipientsDto.size() == recipients.size();
```
klaraf755: A NONE profile that still carries recipient UUIDs (e.g. edited USER→NONE with the list kept — nothing at save time prevents this) gets NONE-typed entries from `explicitRecipients`; `constructNotificationRecipientDto` returns `null` for each, so this comparison fails and every event writes a "could not prepare all of its recipients" failure to trigger history — recreating the noise this PR removes. Either count the NONE skips as prepared-by-design here, or ship the save-time validation the PR body describes.

src/main/java/com/otilm/core/messaging/jms/listeners/NotificationListener.java
```diff
@@ -42,9 +44,16 @@
...
+ * Rolls back on checked exceptions too, rather than leaving Spring's default to be inferred: several methods here
+ * declare {@code NotFoundException} and {@code ConnectorException}, and under the default rules one escaping would
+ * commit whatever the message had accumulated -- a repetition counter advanced for a notification that failed.
+ * Nothing the recipients depend on rides on this transaction: internal notifications and trigger history each
+ * commit in one of their own, so they survive a rollback here by design.
+ */
 @Component
 @AllArgsConstructor
-@Transactional
+@Transactional(rollbackFor = Exception.class)
```
klaraf755: `processMessage` — the only proxied entry point — catches `Exception` in both branches, so no exception can escape to trigger this rollback. The methods declaring `NotFoundException`/`ConnectorException` that the Javadoc cites are private and never cross the proxy. Harmless as defense-in-depth, but the Javadoc asserts a protection the control flow can't exercise; the actual guard is the `notificationSent` flag. Soften the comment or drop the change.

src/main/java/com/otilm/core/messaging/jms/listeners/NotificationListener.java
```diff
@@ -295,7 +354,16 @@ private List<NotificationRecipient> getDefaultRecipients(ResourceEvent event, Ob
         return recipients;
     }
 
-    private void sendExternalNotifications(UUID notificationInstanceUUID, List<NotificationRecipient> recipients, Object notificationData, ResourceEvent
+    /**
+     * Whether every named recipient was prepared for delivery. The notification is handed to the connector either
```
klaraf755, line 351: This (and the updated tests asserting the connector is still called) directly contradicts the PR body's "It is no longer sent." The code's position is defensible — the description is what needs fixing when reconciling.

klaraf755 (changes requested):
The transaction-isolation fix is well-diagnosed and the new integration tests genuinely exercise the rollback-only path — nice work confirming the regression test failed pre-fix.

**The PR description doesn't match the diff.** The "Configuration validated at save time" section describes validation (recipient-naming types need recipients, NONE must carry none, NONE+e-mail rejected) that doesn't exist anywhere on this branch — `NotificationProfileServiceImpl` only validates instance existence. Either a commit is unpushed or the description needs rewriting. Details on the second mismatch are inline.

src/main/java/com/otilm/core/messaging/jms/listeners/NotificationListener.java
lubomirw: Fixed. Combined instead of overwriting: `notificationSent = notificationSent || outcome.notified() > 0`. Confirmed EventHandlersITest#testEventDataNotifications and NotificationInternalNotificationITest still pass. 0e6e6845b.

src/main/java/com/otilm/core/messaging/jms/listeners/NotificationListener.java
lubomirw: Fixed via option (a): NONE resolves to a null delivery DTO by design, so sendExternalNotifications now counts those skips (`skippedByDesign`) and returns `recipientsDto.size() + skippedByDesign == recipients.size()`. A NONE profile carrying UUIDs no longer records a per-event gap. Save-time validation deferred — see the updated PR description. NotificationObjectRecipientITest green. 0e6e6845b.

src/main/java/com/otilm/core/messaging/jms/listeners/NotificationListener.java
lubomirw: Dropped the change — reverted to plain `@Transactional` and removed the Javadoc. processMessage catches everything in both branches, so nothing crosses the proxy for rollbackFor to act on; the notificationSent flag is the guard. 0e6e6845b.

src/main/java/com/otilm/core/messaging/jms/listeners/NotificationListener.java
lubomirw: Corrected the description rather than the code. The connector-decides position stays; the 'It is no longer sent' claim and the 'Configuration validated at save time' section (which described validation NotificationProfileServiceImpl does not implement) are removed. The two integration tests asserting the connector is still called pin the intended behaviour.

klaraf755 (approved):

PR #1931, opened 2026-08-02
Exclude the real JMS producers from the component scan in ProducerMocks

### LINKED ISSUE #1932: Pair ProducerMocks with a component-scan exclusion for the real JMS producers
### Description

The `ProducerMocks` test module mocked the JMS producers with `@Bean @Primary`, which **shadows** the real beans but does not **replace** them. The real `NotificationProducer` stayed `@Component`-scanned and instantiated, so its `@TransactionalEventListener(phase = AFTER_COMMIT, fallbackExecution = true)` remained registered and could dispatch a real `jmsTemplate.convertAndSend` after a commit performed by an importing test.

No test importing the module publishes a `NotificationMessage` today, so this was latent rather than a live failure. Flagged by the review on #1836; fixed in #1931 by excluding the four producer classes from the component scan via `@TypeExcludeFilters` and guarding the annotation pairing.

### Definition of Done

- [ ] `ProducerMocks` declares a `TypeExcludeFilter` that removes `NotificationProducer`, `ActionProducer`, `EventProducer` and `ValidationProducer` from the component scan
- [ ] Every test class importing `ProducerMocks` declares the matching `@TypeExcludeFilters`
- [ ] An architecture test fails the build when the two annotations are half-applied
- [ ] An integration test asserts the mocks are the only beans of each producer type in the context
- [ ] `ContextSignatureGuardTest.BASELINE` is unchanged (no new context combination)
- [ ] Full test suite passes

_Retrospective best guess, reconstructed 2026-08-11 from PR #1931 (+276/-6, 9 files, 5 review rounds, merged within 12 hours). No contemporaneous record of the actual cost survives._

ivosh:
**TL;DR: `ProducerMocks` only *shadowed* the real JMS producers, so the real `NotificationProducer` stayed alive and could still send real JMS messages; this excludes those classes from the component scan instead.**

Closes #1932. Follow-up to the review finding on #1836 ([`ProducerMocks.java`](https://github.com/OmniTrustILM/core/pull/1836#pullrequestreview-4750901814)).

## What was wrong

`@Bean @Primary` **shadows** a bean; it does not **replace** it.

The real `NotificationProducer` is still `@Component`-scanned and instantiated. Its `@TransactionalEventListener(phase = AFTER_COMMIT, fallbackExecution = true)` therefore stays registered. Any test that imports the module, commits, and publishes a `NotificationMessage` would get a real `jmsTemplate.convertAndSend` — the exact I/O the mock exists to suppress.

**Not a live break today.** No test importing `ProducerMocks` publishes that event, so the listener never fires. It was latent, and the fix is to remove the definition rather than hide it.

## The fix

`MockedProducersTypeExcludeFilter` drops all four producer classes from the component scan. Each importing test class declares it:

```java
@Import(ProducerMocks.class)
@TypeExcludeFilters(ProducerMocks.MockedProducersTypeExcludeFilter.class)
```

Two details worth knowing:

- **A `@Bean TypeExcludeFilter` inside the module does not work.** Component scanning happens in the configuration-class *parse* phase; `@Bean` definitions are loaded *after* it, so the filter would never be consulted. `@TypeExcludeFilters` registers it as a singleton before the context refreshes, which is the supported hook.
- **`equals`/`hashCode` on the filter are stateless on purpose.** They keep every declaring test on one context-cache key, so no new context is booted. Spring Boot also rejects a `TypeExcludeFilter` that leaves them at identity.

`@Primary` stays as a fallback.

## Guarding it

The two annotations are separate declarations, so a new test can apply one and forget the other. Two tests cover that:

- `ProducerMocksExclusionArchTest` — fails the build if a context-loading test imports `ProducerMocks` without the exclusion.
- `MockBeanModuleWiringITest.producerModuleLeavesNoRealProducerBeanBehind` — asserts the mock is the *only* bean of each producer type, so the exclusion cannot silently lapse back to shadowing.

## Verification

- Full test suite passes.
- `ContextSignatureGuardTest.BASELINE` is unchanged at 58 — no new context combination is purchased.
- Removing the annotation from one test class was confirmed to fail the new arch guard.

### DIFF
--- a/src/test/java/com/otilm/core/architecture/ContextSignature.java
+++ b/src/test/java/com/otilm/core/architecture/ContextSignature.java
@@ -19,7 +19,7 @@ import java.util.stream.Stream;
  * - nested @TestConfiguration class simple-names,
  * - verbatim @SpringBootTest arguments,
  * - @AutoConfigure* annotation names, and
- * - @TypeExcludeFilters filter-class names.
+ * - @TypeExcludeFilters filter-class names of the NEAREST declaration in the chain.
  *
  * Two classes with equal signatures are expected to share one cached context. Import order is not a signature axis
  */
@@ -64,7 +64,10 @@ final class ContextSignature {
                 configs.addAll(t.configs());
                 springBootTest.addAll(t.springBootTest());
                 autoconfig.addAll(t.autoconfig());
-                typeExcludeFilters.addAll(t.typeExcludeFilters());
+                // Spring resolves @TypeExcludeFilters to the NEAREST declaration only.
+                if (typeExcludeFilters.isEmpty()) {
+                    typeExcludeFilters.addAll(t.typeExcludeFilters());
+                }
             }
             simple = extendsGraph.get(simple);
         }
--- a/src/test/java/com/otilm/core/architecture/ContextSignature.java
+++ b/src/test/java/com/otilm/core/architecture/ContextSignature.java
@@ -17,8 +17,9 @@ import java.util.stream.Stream;
  * - @TestPropertySource args,
  * - @DirtiesContext mode,
  * - nested @TestConfiguration class simple-names,
- * - verbatim @SpringBootTest arguments, and
- * - @AutoConfigure* annotation names.
+ * - verbatim @SpringBootTest arguments,
+ * - @AutoConfigure* annotation names, and
+ * - @TypeExcludeFilters filter-class names.
  *
  * Two classes with equal signatures are expected to share one cached context. Import order is not a signature axis
  */
@@ -47,6 +48,7 @@ final class ContextSignature {
         TreeSet<String> configs = new TreeSet<>();
         TreeSet<String> springBootTest = new TreeSet<>();
         TreeSet<String> autoconfig = new TreeSet<>();
+        TreeSet<String> typeExcludeFilters = new TreeSet<>();
 
         String simple = startSimpleName;
         int hops = 0;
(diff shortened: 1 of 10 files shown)


src/test/java/com/otilm/core/architecture/ContextSignature.java
```diff
@@ -62,13 +64,15 @@ static String of(String startSimpleName, Map<String, String> extendsGraph, Map<S
                 configs.addAll(t.configs());
                 springBootTest.addAll(t.springBootTest());
                 autoconfig.addAll(t.autoconfig());
+                typeExcludeFilters.addAll(t.typeExcludeFilters());
```
lubomirw: **[Low]** | Testing

`ContextSignature` unions `@TypeExcludeFilters` from the subclass and every ancestor, but Spring Boot 3.5.16 resolves only the nearest declaration; a subclass declaration shadows inherited filters. This can miscount future context-cache signatures.

_Suggested:_ Capture only the first non-empty filter declaration while walking subclass to ancestors. Add a regression where a subclass declares filter B over a base declaring filter A and assert that only B contributes.

<sub>flagged by Copilot CLI and Superpowers; independently verified against Spring Boot 3.5.16</sub>

lubomirw (commented):

src/test/java/com/otilm/core/architecture/ContextSignature.java
ivosh: Fixed in 0bee4a61.

Verified against the Spring Boot 3.5.16 sources: `TypeExcludeFiltersContextCustomizerFactory` resolves the annotation with `TestContextAnnotationUtils.findAnnotationDescriptor` and uses only `descriptor.getAnnotation().value()` — nearest declaration, no union. `ContextSignature.of` now keeps only the first non-empty declaration while walking subclass to ancestors.

Added `aSubclassTypeExcludeFiltersDeclarationShadowsTheAncestors`: base declares filter A, subclass declares B — the signature equals a standalone class declaring only B, and differs from one declaring both. The existing inheritance test is renamed to `typeExcludeFiltersAreInheritedWhenTheSubclassDeclaresNone`, which is the case it actually pins.

`ContextSignatureGuardTest` passes unchanged, so `BASELINE` needed no adjustment — no live test class declares `@TypeExcludeFilters` at two levels of one hierarchy.

lubomirw (approved):

PR #1935, opened 2026-08-03
Share one Spring test context across the management-API mock tests

### LINKED ISSUE #1934: Finish the Spring test-context consolidation: absorb the remaining single-class context forks
### Description

`#1704` (PR `#1836`) built the consolidation framework and collapsed the largest cluster. `#1935` and `#1946` continued it. This issue tracks what remains.

Current state, measured with the repo's own `ContextSignature` kernel at `a24efcb0`: **207 context-loading test classes resolve to 61 distinct context signatures, and 46 of those signatures serve exactly one test class.** 110 classes share the dominant signature — the two `@MockitoBean` fields on `BaseSpringBootTest`. `ContextSignatureGuardTest.BASELINE` holds the line at 61.

The kernel counts `@DynamicPropertySource` methods as a signature axis, because Spring's `DynamicPropertiesContextCustomizer` keys the context cache per declaring class. A per-class `@DynamicPropertySource` therefore forks a context even when every other axis matches; a `@TestPropertySource` whose source text is identical across classes does not. `#1946` converted the WireMock-port registrations from the former to the latter, against fixed ports held in `WireMockPorts`.

**A constraint that shapes the work:** moving a one-off local mock into a module does not by itself reduce the context count. The fork only changes shape, from `mocks=[X]` to `imports=[SomeModule]`, and remains a distinct signature. The count falls only when **several** classes land on one shared module combination — or when a class sheds its distinguishing axis entirely and joins an existing group.

Each round below is one pull request, sized for review, lowering `BASELINE` in lock-step. The per-round figures are verified by simulating the change against the kernel, not estimated. Stacking rounds A, 5, 2 and 3 takes the tree to **52 distinct signatures and 36 singletons**.

### Definition of Done

- [x] **Round 1 — Management-API cluster.** `RoleManagementServiceITest`, `RoleAssignmentGuardITest` and `UserManagementServiceCacheEvictionITest` share `ManagementApiMocks`. `AuthenticationCache` deliberately stays out of the module: `AuthenticationCacheITest` autowires the real cache as its subject, so a `@Primary` mock would replace what that test verifies. `UserManagementServiceITest` stays out — its `CertificateUploadService` mock is unshared. Landed in #1935.
- [x] **Round 2 (props leg) — WireMock ports.** Per-class `@DynamicPropertySource` registrations replaced by `@TestPropertySource` against fixed ports in `WireMockPorts`, so classes that spell the property identically share one cached context. Landed in #1946.
- [ ] **Round A — Globalize the pinned service URLs. −4 signatures (61 → 57).** `src/test/resources/application.yml` currently carries unreachable placeholders (`auth-service.base-url: http://authservice.dev`, `scheduler.base-url` and `provisioning.api.url` both `http://localhost:8080`), which 16 test classes override per-class to a `WireMockPorts` constant. Point the test-global yml at the pinned ports and delete those `@TestPropertySource` annotations. Global yml is not a signature axis, so the 9-class `AUTH_SERVICE` group, the 3-class `PROVISIONING_API` group and `SchedulerServiceITest` all fold into the dominant signature, and `CertificateServiceITest` lands on `CryptographicKeyServiceITest`. Confirm no test depends on the *manner* of failure when no stub is listening — the change swaps a DNS failure or a wrong-port response for a connection refusal. `SecurityConfigITest` and `TspSecurityChainITest` keep their annotation for `server.servlet.context-path=`.
- [ ] **Round 5 — No-op forks. −3 signatures.** `ConnectorRepositoryITest` forks on `@SpringBootTest(classes = Application.class)`, which restates the default resolution of `@SpringBootConfiguration`; dropping it lands the class on the 5-class empty-signature group. `AuditLogAspectITest.Config` and `UpdateTrustedCaMarkITest.EventCaptorConfig` each fork a whole context to register one test-only bean — a `@AuditLogged` demo bean and a `CertificateValidationEvent` captor. Hoist both into a component-scanned test `@Configuration` (a top-level `@TestConfiguration` would need `@Import`, which is itself a signature axis) so both classes join the dominant signature.
- [ ] **Round 2 — MockMvc security pair. −1 signature.** `SecurityConfigITest` and `TspSecurityChainITest` now differ by exactly one mock, `PlatformJwtDecoder`, which `SecurityConfigITest` declares but never stubs or verifies. Align the mock set so both share one context. `SessionConfigITest` cannot join the pair: its `JdbcIndexedSessionRepository` is a `@MockitoSpyBean`, not a `@MockitoBean`.
- [ ] **Round 3 — OAuth2 pair. −1 signature.** `OAuth2LoginServiceImplITest` and `OAuth2AuthenticationSuccessHandlerITest` collapse onto a shared mock set covering `AuditLogExternalService`, `AuditLogInternalService` and `OAuth2AuthorizedClientService`. `OAuth2LoginControllerITest` stays separate: it genuinely needs `webEnvironment = RANDOM_PORT`, driving a real `HttpClient` against `@LocalServerPort` to assert redirect behaviour. That question is settled; do not reopen it.
- [ ] **Round 6 — Context eviction.** No signature change; this reclaims re-boots. The 10 migration `ITest`s share one signature *and* carry `@DirtiesContext(AFTER_CLASS)` from `BaseMigrationTest`, so the shared context is evicted after every class — 10 boots where 1 would do. `BaseMigrationTest.resetSchema()` already drops and recreates the schema in `@AfterAll`, so establish by experiment whether the eviction is still load-bearing. Separately, 7 messaging `ITest`s across 5 signatures carry `@DirtiesContext(BEFORE_CLASS)`; audit those too.
- [ ] **Round 8 — Final tighten.** Lower `BASELINE` to its end state and refresh the "cost of a new test-context combination" guidance in `CLAUDE.md` with the resulting module list, plus the `WireMockPorts` / `@TestPropertySource` convention from `#1946` and the test-global-yml rule from Round A.
- [ ] Per-leg CI wall-clock recorded before and after. Local full-suite timing is not admissible evidence — the shared Docker stack makes it unreliable.

### Closed as won't-do

- **Registration path.** `ClientOperationServiceRegisterCustomAttributeITest`'s mock set is a strict subset of `ClientOperationServiceRegisterITest`'s, so a union module would collapse the two. But the two extra types are `@MockitoSpyBean`s and load-bearing — the in-code comment records that `RegistrationChallengeStore` is spied rather than mocked so the binding really persists, with individual tests stubbing failures. Putting verified spies in a shared module contradicts `CLAUDE.md`; −1 signature is not worth it.
- **Spy audit.** The singleton count falls to 36 through the rounds above, and the survivors are legitimately spy-driven. Document the convention in Round 8 rather than converting them.

ivosh:
**TL;DR: three integration tests each booted their own Spring context just to mock the same two auth API clients; they now share one. `BASELINE` 58 → 57.**

Round 1 of #1934.

## What was wrong

`RoleManagementServiceITest`, `RoleAssignmentGuardITest` and `UserManagementServiceCacheEvictionITest` each declared their own `@MockitoBean` for `UserManagementApiClient` / `RoleManagementApiClient`.

Every distinct mock-set is a distinct Spring context-cache key. So these three classes sat on **two** separate signatures and booted a context each, to mock beans a shared module already provides.

## The fix

Each class now imports the existing module and autowires the mock handle:

```java
@Import(ManagementApiMocks.class)
class RoleManagementServiceITest extends BaseSpringBootTest {

    @Autowired
    private RoleManagementApiClient roleManagementApiClient;
```

All three converge on one signature — `imports=[ManagementApiMocks]`, `mocks=[AuditLogsProducer, AuthenticationCache, OpaClient]` — so they share a single cached context.

`MockBeanResetListener` already resets module-declared mocks before each test method, so stubbings do not leak between the three classes now that they share a context.

## Two things worth knowing

**`AuthenticationCache` deliberately stays a per-class `@MockitoBean`.** It looks like the obvious next thing to fold into the module, and that would break `AuthenticationCacheITest` — that test imports the same module and autowires the *real* cache as its subject, so a `@Primary` mock would replace exactly what it verifies. This PR records that in the module's Javadoc so the next round does not have to rediscover it.

**The module also supplies a mock `PlatformAuthenticationClient`**, which these three classes now inherit. None of the paths they exercise reaches it, so it is inert here.

## Scope notes

- `UserManagementServiceITest` is **not** included. Its `CertificateUploadService` mock is unshared, so importing the module would leave it on a unique signature anyway — churn with no collapse.
- The saving is **one** context, not two: `RoleAssignmentGuardITest` and `UserManagementServiceCacheEvictionITest` already shared a signature before this change.

## Verification

- Full suite green on the baseline before the change (4517 tests) and the same 4517 tests after, with one known pre-existing flake (`SigningProfileServiceImplITest`, connector communication on an ephemeral port; passes in isolation).
- `ContextSignatureGuardTest` passes at `BASELINE = 57`; the count was independently recomputed against both SHAs with the repo's own `ContextSignature` kernel.
- The five classes touching `ManagementApiMocks` plus both architecture guards: 63 tests, 0 failures.

### DIFF
--- a/src/test/java/com/otilm/core/util/mockbeans/ManagementApiMocks.java
+++ b/src/test/java/com/otilm/core/util/mockbeans/ManagementApiMocks.java
@@ -11,6 +11,10 @@ import static org.mockito.Mockito.mock;
 
 /**
  * Mocks the platform authentication API clients (HTTP boundary to the auth service).
+ * <p>
+ * {@code AuthenticationCache} must NOT move here. {@code AuthenticationCacheITest} imports this module and
+ * autowires the real cache as its subject, so a {@code @Primary} mock would replace what that test verifies;
+ * the tests that need it mocked declare a per-class {@code @MockitoBean}.
  */
 @TestConfiguration
 public class ManagementApiMocks {
--- a/src/test/java/com/otilm/core/architecture/ContextSignatureGuardTest.java
+++ b/src/test/java/com/otilm/core/architecture/ContextSignatureGuardTest.java
@@ -12,7 +12,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 class ContextSignatureGuardTest {
 
     /** Committed baseline: the exact current distinct count. Update in lock-step with any change. */
-    static final int BASELINE = 58;
+    static final int BASELINE = 57;
 
     private static final Path TEST_ROOT = Path.of("src/test/java");
 
--- a/src/test/java/com/otilm/core/architecture/MockBeanModuleResetArchTest.java
+++ b/src/test/java/com/otilm/core/architecture/MockBeanModuleResetArchTest.java
@@ -1,0 +1,95 @@
+package com.otilm.core.architecture;
+
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * The mock modules in {@code com.otilm.core.util.mockbeans} declare their mocks as {@code @Bean @Primary} context
+ * singletons, which Spring's reset lifecycle does not touch — it only discovers {@code @MockitoBean} fields on the
+ * test class, its superclasses and its enclosing classes.
+ *
+ * {@code MockBeanResetListener} resets them instead, and {@code BaseSpringBootTest} is the only class registering it.
+ * An importer that does not inherit from there keeps its stubbings and interactions across tests sharing the cached
+ * context. This pins the two together.
+ */
+class MockBeanModuleResetArchTest {
+
+    private static final Path TEST_ROOT = Path.of("src/test/java");
+
+    private static final Path MODULE_ROOT = TEST_ROOT.resolve("com/otilm/core/util/mockbeans");
+
+    private static final String LISTENER_ROOT = "BaseSpringBootTest";
+
+    @Test
+    void everyMockModuleImporterInheritsTheMockResetListener() {
+        Set<String> modules = declaredModules();
+        Map<String, String> inheritance = TestClassTaxonomy.parseExtends(TEST_ROOT);
+
+        List<Path> importers = testSources()
+                // Only a context-loading class can hold module mocks; that also keeps meta-test fixture sources out.
+                .filter(file -> TestClassTaxonomy.loadsContext(file, inheritance))
+                .filter(file -> TestClassTaxonomy.annotationTokens(file).imports().stream().anyMatch(modules::contains))
+                .toList();
+
+        assertThat(importers)
+                .describedAs("the mock modules are expected to be in use; this guard is vacuous otherwise")
+                .isNotEmpty()
+                .describedAs("every test class importing a com.otilm.core.util.mockbeans module must extend "
+                        + LISTENER_ROOT + " — it is the only class registering MockBeanResetListener, and without "
+                        + "that listener the modules' @Bean @Primary singleton mocks keep their stubbings and "
+                        + "interactions across every test sharing the cached context")
+                .allMatch(file -> inheritsListenerRoot(file, inheritance));
+    }
+
+    /** Every module in the package counts, so a module added later is covered without touching this test. */
+    private static Set<String> declaredModules() {
+        try (Stream<Path> files = Files.list(MODULE_ROOT)) {
+            Set<String> modules = files
+                    .map(file -> file.getFileName().toString())
+                    .filter(name -> name.endsWith(".java"))
+                    .filter(name -> !name.equals("package-info.java"))
+                    .map(name -> name.substring(0, name.length() - ".java".length()))
+                    .collect(Collectors.toSet());
+            assertThat(modules)
+                    .describedAs("no mock module found under %s — the path is stale", MODULE_ROOT)
+                    .isNotEmpty();
+            return modules;
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    private static boolean inheritsListenerRoot(Path file, Map<String, String> inheritance) {
+        String current = TestClassTaxonomy.primaryClassName(file);
+        int guard = 0;
+        while (current != null && guard++ < 50) {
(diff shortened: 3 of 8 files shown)


lubomirw (dismissed):

src/test/java/com/otilm/core/util/mockbeans/ManagementApiMocks.java
```diff
@@ -11,6 +11,10 @@
 
 /**
  * Mocks the platform authentication API clients (HTTP boundary to the auth service).
+ * <p>
```
lubomirw, line 14: **[Nit]** | Documentation

`ManagementApiMocks`' shared singleton mocks depend on `MockBeanResetListener` for per-test isolation; importing this configuration outside a `BaseSpringBootTest` subclass can leak stubbings and interactions.

_Suggested:_ Document the `MockBeanResetListener` dependency and warn against importing this configuration outside `BaseSpringBootTest` subclasses.

<sub>flagged by Superpowers; multi-review triage marked this likely noise because current consumers register the listener</sub>

lubomirw (commented):

src/test/java/com/otilm/core/util/mockbeans/ManagementApiMocks.java
ivosh: Fixed in 8d0b0a1c — the hazard is real, though I moved it to package scope and made it enforced rather than documented in this one class.

Two things pushed the fix away from a warning paragraph here:

**The constraint is not specific to `ManagementApiMocks`.** It holds for `PollMocks` and `ProducerMocks` too — every module in the package declares `@Bean @Primary` singletons that Spring will not reset. So the contract now lives in a new `package-info.java` for `com.otilm.core.util.mockbeans`, stated once.

**The trap is reachable today, so prose alone was not enough.** `BaseSpringBootTest` is the only class registering `MockBeanResetListener`, and `BaseSpringBootTestNoAuth` is a *separate* root that does not extend it. A future test extending that one and importing a module would silently lose the reset. Following the `ProducerMocksExclusionArchTest` precedent, the new `MockBeanModuleResetArchTest` fails the build when a context-loading test imports a module without `BaseSpringBootTest` in its `extends` chain. Modules are discovered by listing the package directory, so a module added later is covered without touching the test.

All 12 current importers already extend `BaseSpringBootTest`, so nothing was broken — the guard is against drift. I checked it actually bites: a throwaway `extends BaseSpringBootTestNoAuth` + `@Import(PollMocks.class)` probe fails the test and names the offending file, and it returns to green once removed.

The existing `AuthenticationCache` Javadoc stays where it is — that one *is* module-specific.

lubomirw (approved):

PR #1936, opened 2026-08-03
Challenge-gate renew/rekey and protect certificate successors

### LINKED ISSUE #1774: Copy registration authorization to renew/rekey successors and register a successor certificate
### Description

Carry a certificate's registration authorization across renew/rekey, and support pre-registering a successor of an existing certificate. Part of the certificate registration flow (epic ilm#130).

### Scope

- When a certificate with an active `CertificateRegistrationAuthorization` is renewed/rekeyed, copy the authorization to the successor so the same credential keeps protecting the lineage.
- Verify the `authorizationSecret` on the renew **and rekey** paths (replacing the fail-closed guard from #1791), with the same lockout/window semantics as issue; a missing secret is indistinguishable from a wrong one.
- Support `register` with `sourceCertificateUuid`: create a placeholder linked to the source certificate (`PENDING` relation) so completing it becomes a renew/rekey of the source. The successor registration carries its own independent challenge — nothing is inherited from the source.

### Dependencies

Builds on the registration-authorization work in core#1570/#1791 and the contract additions in interfaces#767 (PR interfaces#804).

klaraf755:
Closes #1774. Builds on the registration-authorization work from #1570/#1791 and consumes the contract additions in OmniTrustILM/interfaces#804 (epic ilm#130).

**Depends on interfaces#804** — the pom points at `interfaces:2.19.1-SNAPSHOT`, so CI here can only go green after that PR merges and publishes the snapshot.

## What changed

**1. Renew/rekey of a challenge-protected certificate is now challenge-gated instead of fail-closed.** `rejectRenewOrRekeyOfLiveRegistration` (the temporary guard from #1791, whose Javadoc anticipated exactly this change) is replaced by the same `verifyRegistrationChallenge` gate the issue path uses: correct secret → proceeds, wrong or missing secret → denied with the failed-attempt increment committed under the row lock before the rejection (missing is deliberately indistinguishable from wrong — no oracle), `LOCKED`/`EXPIRED` deny, `CLOSED`/no row passes untouched (a no-challenge registration renews exactly as before). The challenge machinery's event label is parameterized so history records `RENEW`/`REKEY` rather than `ISSUE`.

**2. The authorization follows the lineage.** On a challenge-verified renew/rekey, the authorization is copied to the successor via the new `CertificateRegistrationAuthorizationWriter.copyToSuccessor` (same challenge ciphertext — no re-encryption, state `ACTIVE`, counter reset, `expiresAt` carried over). The existing fate-coupling already retires it if the successor terminates `FAILED`/`REJECTED`.

**3. Register with `sourceCertificateUuid`.** A registration naming a source certificate creates the `REGISTERED` placeholder as its successor (`PENDING` relation via `associateCertificates`, which enforces source `ISSUED`/`REVOKED`, no self-association, no duplicates); completing the placeholder later resolves the relation to `RENEWAL`/`REKEY` by key comparison, i.e. completion is the renewal of the source. Validated up front (source exists) so a bad UUID leaves no orphan row; the relation write sits inside the try block on both the connector-backed and platform-level paths so a pre-acceptance failure fails the placeholder. The successor's challenge is independent of the source's — staging a successor is operator-only (register + `CERTIFICATE UPDATE`), the challenge only substitutes for permission at completion.

## Behavior notes for review

- Platform automation renewing without a secret (location renew, CMP KUR, scheduled/workflow renew) was previously rejected cleanly by the fail-closed guard; it is still denied, but attempts now count toward lockout (default 5). Fail-closed either way; flagging the new side effect.
- Permission substitution (challenge standing in for `checkRenewPermissions`, as it does for `checkCreatePermissions` on issue) is deliberately **not** included — the secret must not ride the `ActionMessage`, and #1774 doesn't ask for it. Follow-up if the self-service portal needs it.

## Tests

Extended `ClientOperationServiceRegisterITest` (existing context signature — no new Spring context): the two `not supported yet` tests rewritten to the gated semantics, plus verified-renew/rekey copy assertions, missing-secret counting, lockout, `CLOSED`-passes, no-challenge regression, and register-with-source coverage (relation created, bad source leaves no row, platform-level path). New `CertificateRegistrationAuthorizationWriterITest` for `copyToSuccessor`. 


src/main/java/com/otilm/core/service/writer/registration/CertificateRegistrationAuthorizationWriter.java
```diff
@@ -45,4 +46,27 @@ public void close(UUID certificateUuid) {
...
+                .ifPresent(predecessor -> {
+                    CertificateRegistrationAuthorization successor = new CertificateRegistrationAuthorization();
+                    successor.setCertificateUuid(successorCertificateUuid);
+                    successor.setChallenge(predecessor.getChallenge());
+                    successor.setState(RegistrationState.ACTIVE);
+                    successor.setFailedAttempts(0);
+                    successor.setExpiresAt(predecessor.getExpiresAt());
+                    authorizationRepository.save(successor);
+                });
+    }
```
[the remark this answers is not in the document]
klaraf755: Not reachable: the only two call sites — the renew and rekey entry methods — invoke this immediately after `submitCertificateRequest` creates a fresh successor certificate in the same invocation, so the same successor UUID can never receive two copies, and a just-created certificate cannot already carry an authorization row (rows are created only at register time or by this copy).

Insert-only is also deliberate rather than an oversight: an upsert would let a future misuse silently overwrite an operator-staged successor registration's own challenge, which is independent of the predecessor's by design. The unique constraint failing loudly is the guard for that invariant. Documented in the Javadoc in 53604db94.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -869,6 +873,52 @@ private void maybeCreateRegistrationAuthorization(Certificate certificate, Clien
...
+     * rule {@code associateCertificates} enforces (ISSUED or REVOKED) is asserted here for the no-orphan
+     * guarantee; the relation itself is written atomically with the placeholder in
+     * {@code createRegistrationPlaceholder}.
+     */
+    private Certificate resolveRegistrationSource(ClientCertificateRegistrationDto request) throws NotFoundException {
+        UUID sourceUuid = request.getSourceCertificateUuid();
+        if (sourceUuid == null) {
+            return null;
+        }
+        Certificate source = certificateRepository.findByUuid(sourceUuid)
```
lubomirw, line 893: **[High]** | Security

A caller with register permission can resolve arbitrary source certificate UUIDs and create successor relations without source-specific CERTIFICATE UPDATE authorization; the bare repository lookup also exposes source existence and state through detailed errors.

_Suggested:_ Run a proxied UPDATE authorization check for sourceCertificateUuid before detailed validation, then reload/lock the source and reassert its state inside the placeholder/relation transaction.

<sub>flagged by Copilot CLI, Guided review, and Superpowers</sub>

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1738,6 +1773,13 @@ public ClientCertificateDataResponseDto renewCertificate(SecuredParentUUID autho
             return response;
         }
 
+        if (challengeAuthorized) {
+            // The verified authorization follows the successor, so the same challenge protects the renewed
+            // certificate's own follow-up operations. A later terminal failure of the successor closes the copy
+            // (fate-coupling in handleFailedOrRejectedEvent).
+            registrationAuthorizationWriter.copyToSuccessor(oldCertificate.getUuid(), UUID.fromString(newCertificate.getUuid()));
```
lubomirw: **[High]** | Security

Challenge verification releases its row lock, successor creation commits, and authorization copying runs in a separate transaction. A concurrent lockout can make the copy silently no-op, while a copy failure leaves a committed successor and PENDING relation without enqueueing the operation.

_Suggested:_ Keep external/CSR work outside transactions, then re-lock and revalidate the verified authorization while creating the successor and its authorization copy atomically; roll back the successor if transfer fails.

<sub>flagged by Copilot CLI, Guided review, and Superpowers</sub>

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
```diff
@@ -1551,29 +1567,136 @@ void issueWithExpiredWindowIsDeniedAndFlipsToExpired() throws Exception {
...
+    @Test
+    void renewWithCorrectChallengeCopiesAuthorizationToSuccessor() throws Exception {
+        KeyPair keyPair = generateKeyPair();
+        Certificate issued = seedIssuedCertWithContent(keyPair);
+        activeAuthorizationFor(issued.getUuid());
+        ClientCertificateRenewRequestDto request = new ClientCertificateRenewRequestDto();
+        request.setRequest(csrBase64(keyPair, RENEWABLE_SUBJECT_DN));
+        request.setAuthorizationSecret(CHALLENGE);
+
+        ClientCertificateDataResponseDto response = clientOperationService.renewCertificate(
```
lubomirw, line 1653: **[High]** | Testing

Three new renew/rekey success-path tests error before reaching copy logic because the mocked dapterFactory returns null and issue-attribute validation dereferences it. The feature's success behavior is therefore unverified.

_Suggested:_ Stub a non-null AuthorityProviderAdapter for these tests or in the shared fixture, then assert renew, rekey, authorization copy, enqueue, and event behavior execute.

<sub>flagged by Guided review; reproduced locally</sub>

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1163,7 +1163,7 @@ public Certificate createCertificateEntity(X509Certificate certificate) {
 
     @Override
     @Transactional
-    public Certificate createRegistrationPlaceholder(RaProfile raProfile, String effectiveSubjectDn, X509RequestContent registrationContent) {
+    public Certificate createRegistrationPlaceholder(RaProfile raProfile, String effectiveSubjectDn, X509RequestContent registrationContent, Certificate sourceCertificate) throws NotFoundException {
```
lubomirw, line 1166: **[Medium]** | Correctness

createRegistrationPlaceholder can throw checked NotFoundException during relation creation, but default @Transactional rollback rules commit the already-saved placeholder on checked exceptions.

_Suggested:_ Add an explicit rollback rule for checked failures and reload/lock the source within the transaction so relation failure cannot commit an orphan placeholder.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1895,6 +1938,11 @@ public ClientCertificateDataResponseDto rekeyCertificate(SecuredParentUUID autho
             return response;
         }
 
+        if (challengeAuthorized) {
+            // The verified authorization follows the successor, as on the renew path.
+            registrationAuthorizationWriter.copyToSuccessor(oldCertificate.getUuid(), UUID.fromString(newCertificate.getUuid()));
```
lubomirw: **[Medium]** | Correctness

Renew/rekey create a new successor registration authorization but never emit CERTIFICATE_REGISTERED, so configured external credential delivery and registration triggers do not receive the successor UUID.

_Suggested:_ Emit exactly one CERTIFICATE_REGISTERED event post-commit after each successful successor copy, and add renew/rekey event assertions.

<sub>flagged by Guided review and Superpowers</sub>

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -1178,7 +1178,14 @@ public Certificate createRegistrationPlaceholder(RaProfile raProfile, String eff
         certificate.setValidationStatus(CertificateValidationStatus.NOT_CHECKED);
         certificate.setCertificateType(CertificateType.X509);
         certificate.setRaProfile(raProfile);
-        return certificateRepository.save(certificate);
+        certificate = certificateRepository.save(certificate);
+        if (sourceCertificate != null) {
+            // Same-transaction PENDING relation, self-invoked like the submitCertificateRequest path so the
+            // caller's register permission is not additionally burdened with CERTIFICATE UPDATE. A relation
+            // failure rolls the placeholder back with it — no half-linked registration can exist.
+            associateCertificates(certificate.getUuid(), sourceCertificate.getUuid());
```
lubomirw: **[Medium]** | Correctness

This PENDING source relation is cleaned up only for synchronous setup failures. Async REGISTER failure or timeout transitions the successor to FAILED and closes its authorization, but leaves the source linked to the failed successor.

_Suggested:_ Delete predecessor relations in the locked terminal REGISTER-failure and timeout transaction, with connector-backed regression coverage.

src/test/java/com/otilm/core/integration/service/writer/registration/CertificateRegistrationAuthorizationWriterITest.java
```diff
@@ -0,0 +1,104 @@
...
+        persistPredecessorAuthorization(RegistrationState.ACTIVE, 3, window);
+
+        authorizationWriter.copyToSuccessor(predecessorUuid, successorUuid);
+
+        CertificateRegistrationAuthorization copied =
+                authorizationRepository.findByCertificateUuid(successorUuid).orElseThrow();
+        assertThat(copied.getChallenge()).isEqualTo(CIPHERTEXT);
+        assertThat(copied.getState()).isEqualTo(RegistrationState.ACTIVE);
+        assertThat(copied.getFailedAttempts()).isZero();
+        assertThat(copied.getExpiresAt().toInstant()).isEqualTo(window.toInstant());
```
lubomirw: **[Medium]** | Testing

This exact Instant assertion fails because PostgreSQL persists microsecond precision while OffsetDateTime.now() retains nanoseconds.

_Suggested:_ Compare values normalized to database precision; if the successor deadline is corrected to null, replace this assertion with the no-deadline invariant instead.

<sub>flagged by Guided review; reproduced locally</sub>

src/main/java/com/otilm/core/service/writer/registration/CertificateRegistrationAuthorizationWriter.java
```diff
@@ -45,4 +46,34 @@ public void close(UUID certificateUuid) {
...
+    public void copyToSuccessor(UUID predecessorCertificateUuid, UUID successorCertificateUuid) {
+        authorizationRepository.findByCertificateUuid(predecessorCertificateUuid)
+                .filter(predecessor -> predecessor.getState() == RegistrationState.ACTIVE)
+                .ifPresent(predecessor -> {
+                    CertificateRegistrationAuthorization successor = new CertificateRegistrationAuthorization();
+                    successor.setCertificateUuid(successorCertificateUuid);
+                    successor.setChallenge(predecessor.getChallenge());
+                    successor.setState(RegistrationState.ACTIVE);
+                    successor.setFailedAttempts(0);
+                    successor.setExpiresAt(predecessor.getExpiresAt());
```
lubomirw: **[Low]** | Correctness

The successor copy carries the predecessor's initial issuance deadline, but the registration-flow design requires copied authorizations to start with xpiresAt = null; a stale deadline can expire credential continuity.

_Suggested:_ Set successor xpiresAt to null unconditionally and update the writer test to enforce that invariant.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1703,9 +1735,12 @@ public void issueCertificateRejectedAction(final UUID certificateUuid) throws No
     public ClientCertificateDataResponseDto renewCertificate(SecuredParentUUID authorityUuid, SecuredUUID raProfileUuid, String certificateUuid, ClientCertificateRenewRequestDto request) throws NotFoundException, CertificateOperationException, CertificateRequestException {
         Certificate oldCertificate = validateOldCertificateForOperation(certificateUuid, raProfileUuid.toString(), ResourceAction.RENEW);
 
-        // Fail-closed: renew is not challenge-gated yet, so refuse to renew a certificate whose registration is
-        // still ACTIVE rather than let a secretless renew silently exit the challenge regime.
-        rejectRenewOrRekeyOfLiveRegistration(oldCertificate.getUuid());
+        // Self-service gate: a certificate with a live registration authorization renews only against its
+        // challenge; a wrong or missing secret is denied (and counted) before any successor exists. A verified
+        // challenge lets the authorization follow the successor (copied below). A certificate with no
+        // authorization row, or a CLOSED one, renews ungated as before.
```
lubomirw: **[Low]** | Maintainability

The phrase s before records change history rather than describing current behavior.

_Suggested:_ Replace with uthorization row, or a CLOSED one, renews without challenge verification. to retain the current invariant without historical narration.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
```diff
@@ -1551,29 +1567,136 @@ void issueWithExpiredWindowIsDeniedAndFlipsToExpired() throws Exception {
...
+                "the challenge gate must reject, not some downstream step");
+        Assertions.assertEquals(1,
+                authorizationRepository.findByCertificateUuid(issued.getUuid()).orElseThrow().getFailedAttempts(),
+                "a secretless rekey of a challenge-protected certificate is a failed attempt and must be counted");
+        verify(actionProducer, never()).produceMessage(Mockito.any());
+    }
+
+    @Test
+    void renewOfCertWithoutAuthorizationProceedsUngated() throws Exception {
+        // A certificate that was never challenge-protected renews exactly as before the gate existed.
```
lubomirw: **[Low]** | Maintainability

This comment narrates behavior relative to when the gate was introduced rather than stating the invariant.

_Suggested:_ Replace with A certificate without registration authorization renews without challenge verification..

<sub>flagged by Hygiene</sub>

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1895,6 +1938,11 @@ public ClientCertificateDataResponseDto rekeyCertificate(SecuredParentUUID autho
             return response;
         }
 
+        if (challengeAuthorized) {
+            // The verified authorization follows the successor, as on the renew path.
```
lubomirw: **[Nit]** | Maintainability

This comment only restates the immediately following copyToSuccessor call.

_Suggested:_ Remove the comment.

<sub>flagged by Hygiene</sub>

lubomirw (changes requested):
Automated multi-reviewer triage — 11 selected findings (REQUEST_CHANGES).

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Removed.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
klaraf755: Replaced.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Fixed in 601920ec2: `resolveRegistrationSource` now runs `certificateService.evaluatePermissionChain(...)` (proxied, `@ExternalAuthorization(CERTIFICATE, UPDATE)`, object-scoped to the source) **before** the repository lookup, so a caller without source-specific UPDATE is denied without learning whether the UUID exists or what state it is in. Regression test included (`registerWithSourceRequiresUpdatePermissionOnTheSource`).

On the reload/lock half: the relation is not written from this up-front validation — it is written by `associateCertificates` inside `createRegistrationPlaceholder`'s transaction, and `associateCertificates` re-reads the source and re-asserts the ISSUED/REVOKED predecessor rule there. The up-front check exists only for the no-orphan guarantee (reject before any placeholder row). The remaining delta from the suggestion is the pessimistic lock on that in-transaction re-read, which is the same unlocked pattern every existing lineage write uses (renew/rekey via `submitCertificateRequest` → `associateCertificates`); the residual window's worst case is a PENDING relation to a concurrently-archived certificate. If we want lock-and-reassert, it belongs in `associateCertificates` for all callers — I'd file that separately rather than have register diverge.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Replaced the comment.

src/main/java/com/otilm/core/service/writer/registration/CertificateRegistrationAuthorizationWriter.java
klaraf755: Set to null.

src/test/java/com/otilm/core/integration/service/writer/registration/CertificateRegistrationAuthorizationWriterITest.java
klaraf755: From this comment https://github.com/OmniTrustILM/core/pull/1936#discussion_r3707905055, the comparison is no longer needed, since expiresAt is set to null.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: On *“a copy failure leaves a committed successor and PENDING relation without enqueueing the operation”* — addressed in 7277b51c6: a copy failure now compensates. The successor is failed through the same path the action's own failures use (`handleFailedOrRejectedEvent`: FAILED transition, PENDING relation removed, failure recorded on the predecessor's history) and the caller gets an explicit abort; the predecessor's authorization is untouched, so the operation is simply retried. Full successor+copy atomicity was deliberately not used: the compliance gate sits between the two writes and can reach compliance connectors, and a transaction must not span an external call. Covered by `renewCopyFailureFailsTheSuccessorInsteadOfOrphaningIt` / `rekeyCopyFailureFailsTheSuccessorInsteadOfOrphaningIt`.

On *“a concurrent lockout can make the copy silently no-op”* — the verified path commits a failed-attempt counter **reset under the same row lock it verifies under**, so a concurrent lockout can only affect the copy if a complete zero-to-max burst of wrong-secret attempts lands in the milliseconds between that commit and the copy. Even then the outcome is a successor without the optional inherited credential — still fully permission-gated — on a renewal that was itself legitimately verified. The suggested re-lock-and-revalidate-or-abort alternative would let any wrong-secret caller deny legitimate renewals at will, which is the worse trade. The copy running only after the compliance early-return is likewise deliberate: a non-compliant successor is an operator-resolution dead end, and the holder's compliant retry against the untouched predecessor gets a fresh copy.

src/test/java/com/otilm/core/integration/service/ClientOperationServiceRegisterITest.java
klaraf755: Confirmed — same root cause as the local failure we hit: the merge of main brought #1890, which made `mergeAndValidateIssueAttributes` unconditional in `submitCertificateRequest`, so the renew/rekey entry paths now dereference the mocked `AuthorityProviderAdapterFactory` (null without a stub). These tests predate that merge. Fixed in 7277b51c6 by stubbing the adapter (`registeringAdapter()`) in all five renew/rekey tests that reach the submit path — the three success-path tests plus the two new copy-failure compensation tests — so renew, rekey, authorization copy, enqueue and compensation behavior are now actually exercised.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
klaraf755: Fixed in 26b6630b1, structurally: `createRegistrationPlaceholder` now re-reads the source inside the transaction **before** the placeholder is written — a vanished source aborts with nothing to roll back — and the relation is created through an entity-based variant of `associateCertificates` that throws only runtime exceptions, so no checked exception can occur after the save. Pinned by `placeholderRollsBackWhenSourceVanishesDuringLinking` (source deleted between validation and linking → `NotFoundException`, no row commits).

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Won't fix — reusing CERTIFICATE_REGISTERED here would misstate what happened, and the delivery need it targets is already met:

1. Nothing is registered on this path. The successor is a REQUESTED certificate in an automatic renewal pipeline — it never enters REGISTERED state and carries no register→issue binding. Triggers and audit consumers keyed on CERTIFICATE_REGISTERED as "a pre-registration was created" would start firing on ordinary renewals.
2. The event's payload and notification say "here are the credentials to complete issuance, by this deadline" — a renewal successor completes automatically (the RENEW/REKEY action is already enqueued), there is nothing for the recipient to do, and the deadline on a copied authorization is always null.
3. The party that needs the new UUID+credential pairing is the credential holder, who just performed the renewal: they presented the secret and receive the successor UUID in-band in the renew/rekey response. Out-of-band credential delivery exists for the operator-staged registration flow, which this is not; general renewal visibility for triggers/notifications is already provided by the RENEW/REKEY action-performed events.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
klaraf755: Fixed in 3037fa825: the poll listener's terminal-failure handling now retires the certificate's links as one step — the helper called at all three terminal-FAILED sites (`retireFailedCertificateLinks`) closes the registration authorization and deletes the predecessor relations inside the same locked transaction, matching the invariant `handleFailedOrRejectedEvent` maintains on the synchronous paths. Because it sits in the shared helper, this also covers async renew/rekey successor failures, which had the same dangling relation. Regression tests added to `CertificateStatusPollListenerTest`: a terminal REGISTER failure deletes the relation, a COMPLETED register keeps it.

lubomirw (approved):

PR #1938, opened 2026-08-03
Add notification profile event data categories with authorization gate

### LINKED ISSUE #1859: Add notification profile event data categories with authorization gates
### Description

Part of Epic OmniTrustILM/ilm#301 (Notification Event Data Enrichment).

Add the `eventDataCategories` configuration to notification profiles in core:

- Flyway migration: nullable `event_data_categories` (text array) on the **parent** `notification_profile` table (not the version table — the field is a data-exposure switch that must take immediate effect, including for monitoring streams pinned to older profile versions). Null/empty = no enrichment; existing profiles unchanged.
- Profile API: presence-aware update semantics (absent/null = keep current value, empty list = disable). Category-only edits create no new profile version.
- **Parent-field update ordering fix**: `editNotificationProfile` currently early-returns on unchanged version-carried fields BEFORE applying parent `description` changes, making description-only edits silent no-ops (pre-existing bug). Parent-field updates (description, categories) must be applied before that early return, passing through the parent's `@Version` optimistic lock.
- **Config-time authorization gates**: a create/update whose resulting state has a gated category enabled is refused unless the configuring user has the required unrestricted access — `CUSTOM_ATTRIBUTES` → ATTRIBUTE MEMBERS; `OBJECT_CONTENT` → CERTIFICATE DETAIL **and** RA_PROFILE MEMBERS (union of registered exporter requirements). The gate fires when a request (a) adds a gated category or (b) changes `notificationInstanceUuid` while a gated category remains enabled. Actionable error messages name the missing permission.

### User Story

As an ILM operator, I want to opt specific notification profiles into enriched event data, with the platform enforcing that only appropriately-privileged users can route object data to connectors.

### Use Case

An operator with full attribute access enables Custom attributes + Metadata on the "ops-team email" profile; a user with restricted ATTRIBUTE MEMBERS permission attempting the same is refused with a message naming the missing permission; swapping the profile's notification instance while a gated category is enabled re-triggers the gate.

### Acceptance Criteria

- [ ] Migration applies cleanly on an existing schema; existing profiles behave identically (no enrichment).
- [ ] Presence-aware updates verified: absent field preserves value (old-client safety), empty list clears, populated list sets; category-only and description-only edits persist and create no new version.
- [ ] Gate behavior verified for both gated categories, including destination-change re-gating and the negative (restricted-user) cases.
- [ ] No breaking changes to the profile API for existing clients.
- [ ] Sonar quality gate passes; new-lines coverage ≥ 80%.

3keyroman:
Closes #1859. Part of epic OmniTrustILM/ilm#301. Depends on OmniTrustILM/interfaces#805 publishing `2.20.0-SNAPSHOT` — CI cannot resolve the dependency until that merges.

Adds the opt-in `eventDataCategories` configuration to notification profiles:

- Nullable `event_data_categories` text array on the parent `notification_profile` table (migration `V202608031100`); null or empty means no enrichment, so existing profiles behave identically. Parent-level, not versioned: the current value applies at send time, including to monitoring streams pinned to older profile versions.
- Presence-aware updates: an absent field keeps the current value so older API clients cannot silently clear it, an empty list disables enrichment, and category-only edits update the parent in place without creating a profile version. Requested lists are normalized: null entries are a validation error, duplicates collapse, storage order is canonical.
- Config-time authorization gate: enabling `CUSTOM_ATTRIBUTES` requires unrestricted ATTRIBUTE MEMBERS access; `OBJECT_CONTENT` requires the certificate content exporter's unrestricted CERTIFICATE DETAIL and RA_PROFILE MEMBERS access. The gate also re-fires when the profile's notification instance changes while a gated category stays enabled, so a destination swap cannot become an ungated export route.
- The gate calls the authorization service outside any transaction against an unlocked read; the locked write refreshes the entities and re-asserts the decision, failing closed with a retry error when gate-relevant state changed concurrently.
- Bumps the interfaces dependency to `2.20.0-SNAPSHOT`.

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
@@ -6,6 +6,7 @@ import com.otilm.api.exception.ValidationException;
 import com.otilm.api.model.client.notification.*;
 import com.otilm.api.model.common.NameAndUuidDto;
 import com.otilm.api.model.core.auth.Resource;
+import com.otilm.api.model.core.notification.NotificationDataCategory;
 import com.otilm.api.model.core.notification.RecipientType;
 import com.otilm.api.model.core.scheduler.PaginationRequestDto;
 import com.otilm.core.dao.entity.notifications.NotificationInstanceReference;
@@ -20,8 +21,11 @@ import com.otilm.core.model.auth.ResourceAction;
 import com.otilm.core.security.authz.ExternalAuthorization;
 import com.otilm.core.security.authz.SecuredUUID;
 import com.otilm.core.security.authz.SecurityFilter;
+import jakarta.persistence.EntityManager;
+import jakarta.persistence.PersistenceContext;
 import com.otilm.core.service.NotificationProfileExternalService;
 import com.otilm.core.service.ResourceObjectAssociationService;
+import com.otilm.core.service.notifications.NotificationDataCategoryGate;
 import com.otilm.core.util.RequestValidatorHelper;
 import org.hibernate.exception.ConstraintViolationException;
 import org.slf4j.Logger;
@@ -36,8 +40,11 @@ import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 
 import java.util.ArrayList;
+import java.util.EnumSet;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
@@ -47,6 +54,8 @@ public class NotificationProfileServiceImpl implements NotificationProfileExtern
 
     private static final Logger logger = LoggerFactory.getLogger(NotificationProfileServiceImpl.class);
 
+    private EntityManager entityManager;
+
     private NotificationProfileServiceImpl self;
     private NotificationProfileRepository notificationProfileRepository;
     private NotificationProfileVersionRepository notificationProfileVersionRepository;
@@ -54,11 +63,17 @@ public class NotificationProfileServiceImpl implements NotificationProfileExtern
 
     private ExecutionRepository executionRepository;
     private ResourceObjectAssociationService resourceObjectAssociationService;
+    private NotificationDataCategoryGate notificationDataCategoryGate;
 
     @Lazy
     @Autowired
     public void setSelf(NotificationProfileServiceImpl self) {
         this.self = self;
+    }
+
+    @PersistenceContext
+    public void setEntityManager(EntityManager entityManager) {
+        this.entityManager = entityManager;
     }
 
     @Autowired
@@ -84,6 +99,11 @@ public class NotificationProfileServiceImpl implements NotificationProfileExtern
     @Autowired
     public void setResourceObjectAssociationService(ResourceObjectAssociationService resourceObjectAssociationService) {
         this.resourceObjectAssociationService = resourceObjectAssociationService;
+    }
+
+    @Autowired
+    public void setNotificationDataCategoryGate(NotificationDataCategoryGate notificationDataCategoryGate) {
+        this.notificationDataCategoryGate = notificationDataCategoryGate;
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
```diff
@@ -219,6 +290,55 @@ NotificationProfileDetailDto persistEditedVersion(SecuredUUID uuid, Notification
...
+    private static Set<NotificationDataCategory> gatedCategoriesRequiringAuthorization(
+            List<NotificationDataCategory> currentCategories, UUID currentDestination,
+            List<NotificationDataCategory> requestedCategories, UUID requestedDestination) {
+        List<NotificationDataCategory> resultingCategories = requestedCategories != null ? requestedCategories : currentCategories;
+
+        Set<NotificationDataCategory> resultingGated = resultingCategories.stream()
+                .filter(NotificationDataCategoryGate::isGated)
+                .collect(Collectors.toSet());
+        boolean addsGatedCategory = !new HashSet<>(currentCategories).containsAll(resultingGated);
+        boolean destinationChanges = !Objects.equals(currentDestination, requestedDestination);
```
lubomirw: **[High]** | Security

The category gate re-authorizes when gated categories are added or the notification instance UUID changes, but not when recipients change or when an already-referenced notification instance/provider configuration changes; if eventDataCategories controls enriched object data sent externally, update rights can redirect already-enabled enriched notifications.

_Suggested:_ Define the trust boundary explicitly and either re-gate recipient changes plus notification-instance/provider updates for profiles with enabled gated categories, or document why NOTIFICATION_PROFILE UPDATE / NOTIFICATION_INSTANCE UPDATE is sufficient authority for those routes; add regression tests for the chosen boundary.

src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
```diff
@@ -171,17 +211,48 @@ public NotificationProfileDetailDto editNotificationProfile(SecuredUUID uuid, No
...
         NotificationProfileVersion currentVersion = notificationProfileVersionRepository.findTopByNotificationProfileUuidOrderByVersionDesc(uuid.getValue()).orElseThrow(() -> new NotFoundException(NotificationProfileVersion.class, uuid));
+        entityManager.refresh(currentVersion);
+
+        // The gate ran outside this transaction against an unlocked read. If a concurrent edit changed
+        // the gate-relevant state in between, the pre-lock authorization no longer covers this request:
+        // fail closed with a retry rather than calling the authorization service under the row lock.
+        Set<NotificationDataCategory> gatedCategoriesUnderLock = gatedCategoriesRequiringAuthorization(
+                notificationProfile.getEventDataCategories(), currentVersion.getNotificationInstanceRefUuid(),
+                updateRequestDto.getEventDataCategories(), updateRequestDto.getNotificationInstanceUuid());
+        if (!authorizedGatedCategories.containsAll(gatedCategoriesUnderLock)) {
```
lubomirw: **[Low]** | Testing

persistEditedVersion has a fail-closed guard for the case where the unlocked pre-check authorized fewer gated categories than the refreshed locked state requires, but no test currently reaches that TOCTOU path.

_Suggested:_ Add a focused unit or integration test where unlocked state requires no/fewer gated categories, refreshed locked state requires more, and persistEditedVersion throws the retry ValidationException before persistence.

<sub>flagged by Guided</sub>

src/test/java/com/otilm/core/integration/service/NotificationProfileServiceITest.java
```diff
@@ -200,6 +202,161 @@ void testCreateNotificationProfile() throws NotFoundException, AlreadyExistExcep
...
+        requestDto.setEventDataCategories(List.of(
+                NotificationDataCategory.ASSOCIATIONS, NotificationDataCategory.METADATA, NotificationDataCategory.METADATA));
+        NotificationProfileDetailDto created = notificationProfileService.createNotificationProfile(requestDto);
+        Assertions.assertEquals(List.of(NotificationDataCategory.METADATA, NotificationDataCategory.ASSOCIATIONS),
+                created.getEventDataCategories());
+    }
+
+    @Test
+    void testEventDataCategories_gateRefusesRestrictedAttributeAccess() {
+        restrictObjectAccess(Resource.ATTRIBUTE, com.otilm.core.model.auth.ResourceAction.MEMBERS);
```
lubomirw: **[Nit]** | Maintainability

The added integration test repeats the fully qualified com.otilm.core.model.auth.ResourceAction type inline where a simple import appears collision-free.

_Suggested:_ Add import com.otilm.core.model.auth.ResourceAction and use ResourceAction.MEMBERS / ResourceAction.DETAIL.

<sub>flagged by Hygiene</sub>

lubomirw (changes requested):
Automated multi-reviewer triage - 3 selected findings (REQUEST_CHANGES).

src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
3keyroman: Resolved via the documentation route in 61a343d20, matching the settled epic design (OmniTrustILM/ilm#301 records recipient edits as non-triggering). The trust boundary is now stated explicitly in the gate's Javadoc: the gate fires on the two edges that change the export decision itself — enabling a gated category (what leaves) and swapping the notification instance while one stays enabled (where it leaves to). Recipient edits only choose who the provider addresses, an authority NOTIFICATION_PROFILE UPDATE holders already exercise over all notification content; instance configuration changes are governed by the instance's own update authority, which already controls delivery of every notification through that instance, enriched or not. A regression test (testEventDataCategories_recipientChangesDoNotRegate) pins the boundary: a user who lost the gated permission can still change recipients on a gated profile, while the destination-swap and category-add tests pin the re-gating edges.

src/main/java/com/otilm/core/service/impl/NotificationProfileServiceImpl.java
3keyroman: Added in 61a343d20 — lockedStateDemandingUnauthorizedGatingFailsClosedBeforePersisting: the unlocked pre-check authorized nothing, the locked state (concurrently gaining a gated category plus a destination swap) demands gating, and persistEditedVersion throws the retry ValidationException with both repositories verified untouched.

src/test/java/com/otilm/core/integration/service/NotificationProfileServiceITest.java
3keyroman: Fixed in 61a343d20 — ResourceAction imported, the four fully qualified usages cleaned up.

lubomirw (approved):

PR #1945, opened 2026-08-04
Remove dead build/config settings ahead of Spring Boot 4

### LINKED ISSUE #1942: Dead build/config settings: obsolete tomcat.version override, vestigial war-plugin block, executable-jar launch script, includeSystemScope, hypersistence banner key
**TL;DR: five build/config settings in `core` are dead — they configure nothing today and get in the way of the Spring Boot 4 upgrade. Delete them while we are still on Boot 3.5.**

Second step in the pre-Boot-4 cleanup series, after the QueryDSL removal (#1939). All five verified on `main` (496fb52c, 2026-08-04).

## What to remove

1. **`tomcat.version` = 10.1.55 override** — `pom.xml` lines 26–29.
   The override's own comment says to delete it once the parent ships Boot ≥ 3.5.15. The parent `com.otilm:dependencies` is on Boot **3.5.16** now. Boot 3.5.16 manages exactly Tomcat 10.1.55, so the effective version does not change. Left in place, the pin would silently hold Tomcat back when the parent moves on.
2. **`maven-war-plugin` configuration block** — `pom.xml` ~line 330.
   The project has no `<packaging>` element, so it packages as `jar`. The war plugin never runs. The whole block is dead.
3. **`<executable>true</executable>`** on `spring-boot-maven-plugin` — `pom.xml` line 354.
   This prepends a shell launch script to the jar. **Spring Boot 4 removes the feature entirely.** Nothing consumes it: the container entrypoint runs `exec java $JAVA_OPTS -jar ./app.jar` (`docker/opt/otilm/entry.sh`), and the Dockerfile only does `jar xf`.
4. **`<includeSystemScope>true</includeSystemScope>`** — `pom.xml` line 355.
   There are zero `system`-scope dependencies. The flag does nothing.
5. **`hibernate.types.print.banner: false`** — `application.yml` lines 99–102.
   The key belongs to hypersistence-utils, which is not a dependency of this project. Nothing reads it.

## Risk

Items 1, 2, 4, 5 are configuration no-ops; removing them cannot change behavior. Item 3 changes the artifact: the jar loses its shell-script prefix, so it can no longer be executed directly as `./app.jar`. No such usage exists in the repo, CI, or the Docker pipeline.

## Verification plan

- Full test suite before and after — results must be identical.
- `mvn help:evaluate -Dexpression=tomcat.version` after the removal must still print `10.1.55`.
- Jar content comparison — the only difference must be the removed launch-script prefix.

ivosh:
Five build/config settings configure nothing today and get in the way of the Spring Boot 4 upgrade. Remove them while still on Boot 3.5.

- tomcat.version=10.1.55 override: the parent com.otilm:dependencies is on Boot 3.5.16, which manages exactly Tomcat 10.1.55, so the effective version is unchanged. Left in place, the pin would silently hold Tomcat back once the parent moves on. Verified: tomcat-embed-core/-el/-websocket all still resolve to 10.1.55, so CVE-2026-41293 and CVE-2026-43512 stay fixed.
- maven-war-plugin block: the project declares no <packaging>, so it builds as a jar and the war plugin never runs.
- <executable>true</executable>: prepends a shell launch script to the jar. Spring Boot 4 removes the feature entirely. Nothing consumes it -- the container entrypoint runs "java -jar ./app.jar" and the Dockerfile only does "jar xf".
- <includeSystemScope>true</includeSystemScope>: there are no system-scope dependencies, so the flag is inert.
- hibernate.types.print.banner: the key belongs to hypersistence-utils, which is not a dependency here. Nothing reads it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,10 +23,6 @@
         <testcontainers.version>1.21.4</testcontainers.version>
         <git-commit-id-maven.version>9.0.2</git-commit-id-maven.version>
         <java.version>21</java.version>
-        <!-- Temporary override for CVE-2026-41293 and CVE-2026-43512.
-          Remove after the parent com.otilm:dependencies upgrades to a Spring Boot release that ships Tomcat 10.1.55+
-          (expected with Spring Boot 3.5.15). Then delete this override. -->
-        <tomcat.version>10.1.55</tomcat.version>
 
         <sonar.host.url>https://sonarcloud.io</sonar.host.url>
         <sonar.organization>ilm</sonar.organization>
@@ -327,21 +323,6 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-war-plugin</artifactId>
-                <configuration>
-                    <packagingExcludes>
-                        **/META-INF/maven/**,
-                        **/WEB-INF/classes/*.p12,
-                        **/WEB-INF/classes/*.jks,
-                        **/WEB-INF/classes/application-*.properties
-                    </packagingExcludes>
-                    <archive>
-                        <addMavenDescriptor>false</addMavenDescriptor>
-                    </archive>
-                </configuration>
-            </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <argLine>@{argLine} -Dspring.test.context.cache.maxSize=128</argLine>
@@ -350,10 +331,6 @@
             <plugin>
                 <groupId>org.springframework.boot</groupId>
                 <artifactId>spring-boot-maven-plugin</artifactId>
-                <configuration>
-                    <executable>true</executable>
-                    <includeSystemScope>true</includeSystemScope>
-                </configuration>
             </plugin>
             <plugin>
                 <groupId>org.hibernate.orm.tooling</groupId>
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -96,11 +96,6 @@ provider:
           - PT5M
         max-attempts: 50
 
-hibernate:
-  types:
-    print:
-      banner: false
-
 info:
   app:
     name: ILM Core
--- a/src/test/resources/application.yml
+++ b/src/test/resources/application.yml
@@ -46,11 +46,6 @@ cmp:
         timeout: 10
   verbose: true
 
-hibernate:
-  types:
-    print:
-      banner: false
-
 logging:
   level:
     org:


lubomirw (commented):
Automated multi-reviewer triage — 2 findings, neither blocking.

The diff itself verifies clean. All five inertness claims check out empirically: released parent `com.otilm:dependencies:1.4.2` is Boot 3.5.16 with no Tomcat override and `spring-boot-dependencies-3.5.16` manages `tomcat.version=10.1.55` exactly (so the effective version is unchanged and both CVEs stay fixed); there is no `<packaging>` element, so the war plugin never ran; there are no `<scope>system</scope>` dependencies; `hypersistence-utils` is not a dependency, directly or transitively; and `docker/opt/otilm/entry.sh` runs `exec java $JAVA_OPTS -jar ./app.jar`, so nothing consumed the executable-jar launch script.

Both findings below are **off-diff** — this PR is 33 deletions and 0 additions, so there is no line to anchor an inline comment to. Both are follow-ups rather than changes requested here.

**[Low] | Security — `src/main/resources/trustStore.jks`**

The removed `<packagingExcludes>` was the only place in the build expressing intent to keep `*.jks` / `*.p12` / `application-*.properties` out of the artifact. Because it was scoped to `maven-war-plugin`, which never runs under jar packaging, that intent never took effect: `src/main/resources/trustStore.jks` already ships at `BOOT-INF/classes/` today and still does after this PR, so behaviour is unchanged. The file is referenced nowhere in `src/`, `.github/`, `Dockerfile*` or `docker/` — `TrustedCertificatesConfig` builds its trust store from the JDK's `cacerts` plus DB-held certificates, and the other `trustStore` matches are unrelated local variables. No `application-*.properties` exists in the repo at all.

_Suggested:_ the removal here is right — don't re-add a war-plugin config that can never run under jar packaging. As a follow-up, confirm `trustStore.jks` is dead and delete it, or document why it ships. If keystores must be permanently barred from the artifact, express it via `<resources><excludes>` or `maven-jar-plugin`, which do run for jar packaging.

**[Low] | Maintainability — `config/trivy.yaml`**

In the same spirit as this cleanup: `config/trivy.yaml` is itself dead config. The org reusable workflow only consults it when `allow-trivy-config-override` is true, and that input defaults to `false`; neither `test_docker_image.yaml` nor `publish_docker.yaml` passes it, and nothing else in the repo references the file. So the enforced policy is the org default (`CRITICAL` + `HIGH`, `os,library`), not this file's `severity: [CRITICAL]`.

_Suggested:_ delete `config/trivy.yaml`. It is strictly weaker than the org default it does not replace, so keeping it around only invites the false impression that core runs a narrower gate. If a repo-local policy is genuinely wanted, it has to opt in via `allow-trivy-config-override: true` explicitly.

<sub>flagged by guided-pr-review + superpowers code review (trustStore.jks) and by triage verification (trivy.yaml). Copilot CLI and the hygiene scan returned no findings.</sub>

ivosh:
> Automated multi-reviewer triage — 2 findings, neither blocking.
> 
> Both findings below are **off-diff** — this PR is 33 deletions and 0 additions, so there is no line to anchor an inline comment to. Both are follow-ups rather than changes requested here.

Created a follow-up task https://github.com/OmniTrustILM/core/issues/1948.

lubomirw (approved):

PR #1946, opened 2026-08-04
Pin WireMock ports so mock integration tests share cached Spring contexts

### LINKED ISSUE #1934: Finish the Spring test-context consolidation: absorb the remaining single-class context forks

ivosh:


src/test/java/com/otilm/core/integration/service/NotificationServiceITest.java
```diff
@@ -91,7 +86,7 @@ void testNotificationsOperations() throws NotFoundException {
     }
 
     private void setupAuthServiceMock() {
-        mockServer = new WireMockServer(AUTH_SERVICE_MOCK_PORT);
+        mockServer = new WireMockServer(WireMockPorts.AUTH_SERVICE);
```
lubomirw, line 96: **[Medium]** | Testing

`testNotificationsOperations()` binds the shared fixed port `WireMockPorts.AUTH_SERVICE` (10001) but stops the server only via an inline `mockServer.stop()` at the end of the test body — no `@AfterEach`, no `try/finally`. A failure before that line leaks 10001 for the rest of the JVM fork, cascading `Address already in use` onto every other class in the AUTH_SERVICE cohort on the same CI shard. The leak is pre-existing; this PR widens the blast radius by consolidating ~10 classes onto 10001.

_Suggested:_ Move the stop into a null-guarded `@AfterEach` (or wrap the body in try/finally), matching the teardown this PR already added to `TriggerEvaluatorITest`/`NotificationProfileServiceITest` (@AfterEach) and `FilterPredicatesBuilderITest`/`SecretServiceITest` (try/finally); drop the inline stop.

<sub>flagged by Copilot CLI, guided-pr-review, superpowers</sub>

src/test/java/com/otilm/core/integration/auth/oauth2/OAuth2LoginControllerITest.java
```diff
@@ -41,16 +41,12 @@
 
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@TestPropertySource(properties = {"auth-service.base-url=http://localhost:" + WireMockPorts.AUTH_SERVICE_SECURITY_CHAIN,
```
lubomirw: **[Medium]** | Testing

This `@TestPropertySource` points `auth-service.base-url` at the fixed cohort port `AUTH_SERVICE_SECURITY_CHAIN` (10003), but `setUp()` starts the mock with `new WireMockServer(0)` (a random port) and never stubs it — the two halves the `WireMockPorts` contract requires to agree do not. Currently latent: these OAuth2/session tests never exercise auth-service, so the mock is effectively dead. Pre-existing, not a regression from this PR.

_Suggested:_ Either bind to `new WireMockServer(WireMockPorts.AUTH_SERVICE_SECURITY_CHAIN)` and add the stubs if the path is meant to run, or drop the unused property and vestigial `mockServer` so the class no longer claims the 10003 cohort.

src/test/java/com/otilm/core/util/WireMockPorts.java
```diff
@@ -0,0 +1,40 @@
...
+ * serve: a type-level annotation sits outside the class body's scope and cannot read it.
+ * <p>
+ * A port is shared by a cohort, not global. Classes that share one cached Spring context must spell the property
+ * identically — {@link com.otilm.core.architecture.ContextSignature} compares the annotation's source text — so they
+ * share a constant by necessity. Classes that share a context with nobody get their own port instead: it buys them
+ * nothing to join a crowded one, and it widens the blast radius when a stub outlives its test.
+ */
+public final class WireMockPorts {
+
+    /** auth-service stub for the nine-class shared context group, plus {@code EventHandlersITest}. */
```
lubomirw: **[Low]** | Documentation

The `AUTH_SERVICE` and `PROVISIONING_API` Javadocs enumerate hand-counted cohorts ("nine-class shared context group, plus `EventHandlersITest`", "three-class shared context group"). Nothing verifies these counts and they rot as tests are added or removed — the same drift this repo's own "avoid stale enumerated lists in docs" principle warns against.

_Suggested:_ Describe each port by which service it mocks and why the group shares it, without the numeric class count.

src/test/java/com/otilm/core/util/WireMockPorts.java
```diff
@@ -0,0 +1,40 @@
...
+ * Both halves — the {@code @TestPropertySource} URL and the {@code new WireMockServer(...)} call — must read the port
+ * from here. Spelling the number twice per class lets one copy drift, and the failure mode is the application quietly
+ * calling a port nothing is listening on: a timeout, not a clear error. The constants are {@code public static final}
+ * so a concatenation such as {@code "auth-service.base-url=http://localhost:" + WireMockPorts.AUTH_SERVICE} is a
+ * compile-time constant expression, which an annotation value must be. A class's own {@code private} constant cannot
+ * serve: a type-level annotation sits outside the class body's scope and cannot read it.
+ * <p>
+ * A port is shared by a cohort, not global. Classes that share one cached Spring context must spell the property
+ * identically — {@link com.otilm.core.architecture.ContextSignature} compares the annotation's source text — so they
+ * share a constant by necessity. Classes that share a context with nobody get their own port instead: it buys them
```
lubomirw: **[Nit]** | Documentation

The class Javadoc says context-sharing classes share a constant "by necessity" and classes that "share a context with nobody get their own port instead" — but `AUTH_SERVICE_STANDALONE` (10000, documented "need no shared port") is used by both `CertificateServiceITest` and `SecretServiceITest`, and `AUTH_SERVICE_SECURITY_CHAIN` (10003) is shared by three classes that each load their own context. The stated rule contradicts the actual grouping.

_Suggested:_ Reword so the rule matches reality — ports are grouped by service/theme and safe to reuse because Surefire runs classes sequentially and each tears its server down; context identity is not the sole grouping criterion.

<sub>flagged by guided-pr-review, superpowers</sub>

lubomirw (commented):
Automated multi-reviewer triage — 4 findings (Comment). Sources: Copilot CLI, guided-pr-review, superpowers, hygiene. The substantive new code — the string-literal-aware comment scan, the `@DynamicPropertySource` regex + loud-fail marker cross-check, and the signature axis — was verified correct; these are the residual gaps.

src/test/java/com/otilm/core/integration/service/NotificationServiceITest.java
ivosh: Fixed in be7efea2 — the inline `mockServer.stop()` is gone; teardown is now a null-guarded `@AfterEach` that also nulls the field, so a failure anywhere in the test body still releases 10001. Verified by running `NotificationServiceITest`, `NotificationProfileServiceITest` and `AuthServiceITest` in one fork: 13/13, no `Address already in use`.

src/test/java/com/otilm/core/integration/auth/oauth2/OAuth2LoginControllerITest.java
ivosh: Fixed in be7efea2 — took the second option: the vestigial `mockServer` and the `auth-service.base-url` property are both gone, so the class no longer claims the 10003 cohort (which is now just `SecurityConfigITest` and `TspSecurityChainITest`, both of which bind the constant and stub it). Confirmed the mock really was dead: all 13 tests pass without it, and `ContextSignatureGuardTest` still reports 61 — dropping the property neither merged nor forked a context.

src/test/java/com/otilm/core/util/WireMockPorts.java
ivosh: Fixed in be7efea2 — the counts are out of the Javadocs. Each constant is now described by the service it mocks and the theme of the tests that use it. `AUTH_SERVICE_STANDALONE` also lost its "share a context with no other class, and so need no shared port" claim, which was the same drift in another form: `CertificateServiceITest` and `SecretServiceITest` both bind 10000 and (differing by `@RecordApplicationEvents`) do not share a context.

src/test/java/com/otilm/core/util/WireMockPorts.java
ivosh: Fixed.  "by necessity" / "share a context with nobody" wording is gone. The residual contradiction did survive in the `AUTH_SERVICE_STANDALONE` constant Javadoc, and that is fixed in be7efea2 along with the sibling finding. The class Javadoc now states the rule as you suggested: ports are grouped by mocked service and test theme, sharing a context forces sharing a constant but not the reverse, and reuse is safe because Surefire runs classes sequentially and each stops its own server.

lubomirw (approved):

PR #1952, opened 2026-08-04
Move the stubbed service URLs into the test-global application.yml

ivosh:
Round A of #1934.

`#1946` pinned the WireMock ports but named them **per test class**, in a `@TestPropertySource`. That annotation's source text is a `ContextSignature` axis, so every distinct spelling still forked a context. Naming the URLs once in the test-global `src/test/resources/application.yml` keeps them off the signature entirely — the yml is identical for every context, so it forks nothing.

### Effect

The 9-class `auth-service` group, the 3-class `provisioning-api` group and `SchedulerServiceITest` fold into the dominant signature, and `CertificateServiceITest` lands on `CryptographicKeyServiceITest`.

| | before | after |
|---|---|---|
| distinct context signatures | 61 | **57** |
| single-class signatures | 46 | **43** |

`BASELINE` lowered to 57 in lock-step.

### The three auth-service constants collapse into one

`AUTH_SERVICE_SECURITY_CHAIN` (10003) and `AUTH_SERVICE_STANDALONE` (10000) existed only so classes that could not share a context would not fight over a port. With the URL named globally there is one port to bind, and sharing it stays safe because Surefire runs test classes sequentially in one fork and every class stops its server when it finishes.

`SecurityConfigITest` and `TspSecurityChainITest` keep a `@TestPropertySource` for `server.servlet.context-path=`, so they keep their own contexts either way; they now bind 10001 like everyone else.

### Why this is safe

The previous defaults were unreachable placeholders — `http://authservice.dev` (DNS-unresolvable) and `http://localhost:8080` for both scheduler and provisioning-api — so no test could have been relying on a response from them. A test that never stubs the service now gets a connection refusal instead of a DNS failure. Moving scheduler and provisioning off `localhost:8080` also removes the hazard of a developer's own app listening there.

The Flyway Java migrations that also read `auth-service.base-url` are unaffected: they resolve it from the `AUTH_SERVICE_BASE_URL` environment variable or a classpath `application.properties`, never from `application.yml`.

### Verification

- `ContextSignatureGuardTest` + `ContextSignatureTest` green at `BASELINE = 57`.
- All 18 touched ITests green (387 tests).
- Full local suite: **4544 tests, 0 failures, 2 errors** — both `ConnectorCommunicationException` against ephemeral ports (60097, 52492) drawn by `new WireMockServer(0)` in files this PR does not touch, and both pass when re-run in isolation. That is the `dynamicPort()` collision class the `WireMockPorts` Javadoc already describes; converting those connector stubs is out of scope here.
- Per-leg CI wall-clock to be read off this PR's run, per the issue's evidence requirement.


src/test/resources/application.yml
```diff
@@ -34,8 +34,9 @@ connector:
     max-connections: 20
     pending-acquire-timeout: 10s
 
+# Points at the port com.otilm.core.util.WireMockPorts.AUTH_SERVICE names.
 auth-service:
-  base-url: http://authservice.dev
+  base-url: http://localhost:10001
```
lubomirw, line 39: **[Medium]** | Testing

`@TestPropertySource` inline properties sit **above** OS environment variables in Spring Boot's property precedence; test `application.yml` sits **below** them. Moving these three URLs out of per-class annotations and into the yml drops them under `AUTH_SERVICE_BASE_URL`, `SCHEDULER_BASE_URL` and `PROVISIONING_API_URL`.

Those variables are not hypothetical: README marks `AUTH_SERVICE_BASE_URL` and `SCHEDULER_BASE_URL` as **required** (lines 92, 97), and `src/main/resources/application.yml` references all three with no default — `${AUTH_SERVICE_BASE_URL}` at :47, `${SCHEDULER_BASE_URL}` at :156, `${PROVISIONING_API_URL}` at :310. A developer who has them exported in order to run core locally will now have these ITests bypass WireMock and talk to the real services; the scheduler and provisioning tests issue POST/DELETE. CI stays green because the variables are unset on the runners, so the failure mode is developer-machine-only.

_Suggested:_ declare the three URLs once on `BaseSpringBootTest` and `BaseSpringBootTestNoAuth` via `@TestPropertySource` rather than in the yml. The inherited annotation text is identical for every subclass, so the context-sharing win this PR is after is preserved — `ContextSignature.distinctCount` still returns 57 — while the highest-precedence pin is restored. It also removes the constant/yml duplication raised separately below.

<sub>flagged by Copilot CLI; precedence rule and env-var wiring verified against README and src/main/resources/application.yml</sub>

src/test/java/com/otilm/core/util/WireMockPorts.java
```diff
@@ -1,28 +1,20 @@
...
- * machine happens to hold the port it drew.
+ * Surefire runs test classes sequentially in one fork, so a fixed port is safe; a fixed port below the 49152 ephemeral
+ * floor is also immune to the collisions {@code dynamicPort()} suffers when another process asks for an ephemeral port.
  * <p>
- * Ports are grouped by the service they mock and by the theme of the tests that use them. Classes that share one cached
- * Spring context must spell the property identically — {@link com.otilm.core.architecture.ContextSignature} compares the
- * annotation's source text — so they necessarily share a constant; but a constant may also be reused by classes that
- * each load their own context. That reuse is safe because Surefire runs test classes sequentially and every class stops
- * its server when it finishes.
+ * These constants and the URLs in the test {@code application.yml} must stay in sync — the yml cannot reference them.
```
lubomirw: **[Low]** | Maintainability

This line states the constants and the yml URLs "must stay in sync", but 10001 / 10010 / 10011 now exist as literals in both places, joined only by prose. A future edit to either side drifts silently and surfaces as connection-refused or never-matched-stub failures spread across many ITest classes, nowhere near the line that caused it.

_Suggested:_ add a guard test alongside `ContextSignatureGuardTest` — the repo already uses committed-invariant guard tests for exactly this purpose — that reads `src/test/resources/application.yml` and asserts the port in `auth-service.base-url`, `scheduler.base-url` and `provisioning.api.url` equals `WireMockPorts.AUTH_SERVICE`, `SCHEDULER` and `PROVISIONING_API` respectively. Roughly 15 lines, and with no Spring context it leaves `BASELINE` untouched. Alternatively, the base-class `@TestPropertySource` approach suggested on `application.yml` removes the duplication altogether.

<sub>flagged independently by guided-pr-review, superpowers code-review and Copilot CLI</sub>

lubomirw (commented):
Automated multi-reviewer triage — 4 findings (COMMENT).

Sources: Copilot CLI, `guided-pr-review`, superpowers code-review, and a comment/log hygiene pass — run independently, then deduped and re-verified against the diff. Three of the four findings were raised by more than one reviewer. Nothing here restates an existing thread; the PR had none open.

**Not inline** — `pom.xml` is not part of this diff:

- **`pom.xml:326`** — **[Low]** | Testing — Fixed shared ports rest entirely on "Surefire runs test classes sequentially in one fork". That is true today, but by default only: the surefire block sets just `argLine`, and there is no `forkCount` / `reuseForks` / `<parallel>`, no `src/test/resources/junit-platform.properties`, and no `@Execution` anywhere in the tree. This PR increases the number of classes leaning on that default by collapsing 10000 and 10003 onto 10001 — 13 classes now bind it. Enabling parallel classes later would not always fail loudly: a live stub on 10001 answers another class's calls, so `verify()` counts foreign requests and assertions can pass for the wrong reason. _Suggested:_ pin the invariant where it can regress — add `junit.jupiter.execution.parallel.enabled=false` to the surefire `<configurationParameters>`, or at minimum a comment there pointing at `com.otilm.core.util.WireMockPorts`. <sub>flagged by superpowers code-review; pom and test-tree state verified</sub>

Two further items were triaged and deliberately left out: the class Javadoc narrating the migration in before/after terms (the constraint itself is worth keeping), and the fully-qualified class names in the new `application.yml` comments (the file has no imports, so the FQN is arguably the clearer choice).

src/test/java/com/otilm/core/util/WireMockPorts.java
```diff
@@ -1,28 +1,20 @@
...
+ * Naming the ports globally rather than per class is what keeps them off the context signature: a per-class
+ * {@code @TestPropertySource} would fork a context (see {@link com.otilm.core.architecture.ContextSignature}), whereas
+ * the test-global yml is identical for every context and so forks nothing. Classes stubbing the same service therefore
```
lubomirw: **[Low]** | Documentation

"Classes stubbing the same service therefore share one port and one cached context" does not hold. 13 classes bind `AUTH_SERVICE`: `SecurityConfigITest` and `TspSecurityChainITest` extend `BaseSpringBootTestNoAuth` and keep their own `@TestPropertySource("server.servlet.context-path=")`, and among the remaining 11, `@Transactional`, `@RecordApplicationEvents` and `@MockitoBean` fork contexts further. Sharing a port is a runtime-lifecycle property; sharing a context depends on every other signature axis — and this Javadoc is the only place the invariant is explained to the next maintainer.

_Suggested:_ state the two facts separately, e.g. "Classes stubbing the same service share one port. Whether they also share a cached context depends on their other context axes — the port is not one of them, which is the point." Worth also restoring a one-line note on when a new constant is warranted, since that per-group rationale left with `AUTH_SERVICE_SECURITY_CHAIN` and `AUTH_SERVICE_STANDALONE`.

<sub>flagged independently by hygiene, superpowers code-review and Copilot CLI; class count verified against the head tree</sub>

src/test/resources/application.yml
ivosh: Fixed in 0bb13898 — and confirmed empirically before fixing. With the three variables exported to bogus ports, `AuthServiceITest` fails 4/5 with `Connection refused: localhost/127.0.0.1:19999`; with the fix it passes 5/5. The precedence inversion was real, not theoretical.

Took the suggested route: one `@TestPropertySource` on `BaseSpringBootTest` and `BaseSpringBootTestNoAuth`. All 17 affected classes extend one of the two, and `ContextSignature` unions the `props` axis up the `extends` chain, so the inherited text is identical everywhere — I simulated it against the kernel before writing any code, and the count stays **57/43**.

One deviation from the suggestion: the yml values stay rather than move. 37 context-loading classes extend neither base class, so the yml is still the floor for them; the base-class annotation is what restores precedence for the ones that matter. That does leave the duplication, so I closed it with a guard instead — see the thread below.

src/test/java/com/otilm/core/util/WireMockPorts.java
ivosh: Fixed in 0bb13898. The claim was wrong as written — sharing a port is a runtime-lifecycle property, sharing a context depends on every other signature axis. The Javadoc now states the two separately, close to your wording:

> Classes stubbing the same service therefore share one port; whether they also share a cached context depends on their other context axes — the port is not one of them, which is the point.

Also restored the per-group rationale that left with the deleted constants ("a new constant is warranted only when two services must be stubbed with conflicting behaviour at the same time; two classes stubbing the *same* service never need separate ports, because they never run concurrently"), and rewrote the paragraph on the binding mechanism, since the properties now live on the base classes rather than only in the yml.

src/test/java/com/otilm/core/util/WireMockPorts.java
ivosh: Fixed in 0bb13898 with the suggested guard, `WireMockPortsGuardTest`. It parses the test `application.yml` and asserts each of the three URLs resolves to the port its `WireMockPorts` constant names. No Spring context, so `BASELINE` is untouched.

The alternative — dropping the yml values and relying solely on the base-class annotation — does not work: 37 context-loading classes extend neither base class and still need the yml. The duplication is structural, which is what makes the guard the right answer rather than the fallback.

The guard also carries a second assertion the original suggestion did not cover, and it turned out to be the more important one. See the note in the summary comment.

ivosh:
Thanks — all four findings were valid and are fixed in 0bb13898. Nothing marked won't-fix.

**The precedence finding was the important one, and it was a live bug.** Rather than reason from the precedence table, I reproduced it: with `AUTH_SERVICE_BASE_URL`, `SCHEDULER_BASE_URL` and `PROVISIONING_API_URL` exported to bogus ports, `AuthServiceITest` fails 4/5 with `Connection refused: localhost/127.0.0.1:19999` — WireMock bypassed, real host dialled. With the fix, 5/5. Exactly the developer-machine-only failure described.

**The `pom.xml` finding** is fixed as suggested — `junit.jupiter.execution.parallel.enabled=false` in `<configurationParameters>`, with a comment stating *why* rather than just *what*: the danger is not the collision but that a live stub on a shared port answers a foreign class's calls, so `verify()` counts the wrong requests and a test passes for the wrong reason. Confirmed the four surefire profiles only add `<includes>`/`<excludes>` and deep-merge with the base `<configuration>`, so the parameter propagates to all of them.

I mutation-tested both halves rather than trusting the green:

| Mutation | Result |
| --- | --- |
| `@TestPropertySource` removed from `BaseSpringBootTestNoAuth` | ❌ fails — *"must carry @TestPropertySource pinning the stub URLs"* |
| yml `scheduler.base-url` drifted 10011 → 19011 | ❌ fails — *"must point at the port WireMockPorts names"* |

**Verification:** distinct signatures still **57/43** (measured with the repo's own kernel before and after — the fix is signature-neutral by design, since the kernel unions the `props` axis up the `extends` chain); 402 tests green across all 17 affected ITests plus the guards.

Two items from the triage I agree should stay out: the before/after Javadoc narration (the constraint itself is worth keeping) and the FQNs in the yml comments (no imports there, so the FQN is clearer).

lubomirw (approved):

PR #1955, opened 2026-08-04
Share hardened RestClient defaults across OAuth2 and provisioning

### LINKED ISSUE #1949: Migrate OAuth2Util from RestTemplate to RestClient
**TL;DR: `OAuth2Util` is the last `RestTemplate` user in core — migrate its two call sites to `RestClient` so the codebase is on one HTTP-client idiom before the Spring Boot 4.1 flip.**

Part of the Spring Boot 4.1 pre-work under the migration epic. Verified on `main` (2026-08-04).

## Current state

`src/main/java/com/otilm/core/util/OAuth2Util.java` creates a throwaway `new RestTemplate()` at two sites, both wired to `OAuth2ErrorResponseErrorHandler`:

1. **RP-initiated logout** (line 69). Sends `GET` to the provider's end-session endpoint with `id_token_hint`. The response body is ignored. Failures are caught and logged — logout is best-effort.
2. **`getUserInfo`** (line 86). Sends `GET` with `Accept: application/json` and a bearer token. Deserializes the body to `Map<String, Object>`.

These are the **only `RestTemplate` references in the repo**, main and test. The provisioning package already uses `RestClient` (`ProvisioningApiConfig` plus two API clients), so this migration closes the inconsistency.

## Why before Boot 4.1

- `RestTemplate` is in maintenance mode; `RestClient` is its Framework 6.1+ replacement and the API Boot 4.1 centers on.
- The change is framework-neutral: `RestClient` behaves identically on Boot 3.5, so this is zero-risk front-loading.
- With one client idiom, the flip PR's `spring-boot-restclient` starter addition touches one pattern instead of two.

## Suggested shape

One shared `private static final RestClient`, built once with `.defaultStatusHandler(new OAuth2ErrorResponseErrorHandler())` — `RestClient` is immutable and thread-safe, so per-call construction is unnecessary. Logout becomes `.get().uri(uri).retrieve().toBodilessEntity()`; userinfo becomes `.get().uri(uri).accept(APPLICATION_JSON).headers(h -> h.setBearerAuth(token)).retrieve().body(new ParameterizedTypeReference<>() {})`.

## Verification

- `OAuth2UtilTest` must pass unchanged; extend it if the two HTTP paths lack coverage.
- Full test suite before and after — identical results expected.
- Regression surface is the OIDC login flow (userinfo feeds claim/username resolution) and provider-side logout; an optional smoke test is a login → logout round trip against a test provider.

ivosh:
Closes #1949.

`OAuth2Util` was the last `RestTemplate` user in core. Its two call sites — RP-initiated logout and the userinfo lookup — now share one immutable `RestClient` instead of constructing a throwaway `RestTemplate` per call.

Swapping the client also swaps the transport, and `ClientHttpRequestFactoryBuilder.detect()` passes Apache's own defaults straight through. Those defaults are written for a short-lived client, not one held for the lifetime of the process. So they are set explicitly, in a new `PlatformHttpClients` that **both** platform `RestClient`s — OAuth2 and provisioning — now build from.

Every item below was found by an iterated review pass and is pinned by a mutation-verified test.

## Shared defaults (`PlatformHttpClients`)

| Setting | `detect()` gave | Now | Reason |
|---|---|---|---|
| Pool size | `maxPerRoute=5`, `maxTotal=25` | 100 / 100 | 5 per route would throttle the userinfo call, which runs on **every** JWT-authenticated request (`PlatformJwtAuthenticationConverter.convert`). Measured 5 → 40 concurrent. |
| Cookies | `BasicCookieStore`, confirmed replaying between calls | disabled | The client is process-wide; one cookie jar would replay one user's provider session cookie onto everyone else's calls. |
| Retries | `HttpRequestRetryExec`, honours `Retry-After` verbatim | idempotent I/O-error retry only | `Retry-After` is unbounded, letting a degraded peer decide how long our threads block. The I/O-error retry is kept — that is what recovers a concurrently closed pooled connection. |
| Connection lease | 3 minutes | 5 seconds | The pool now caps concurrency, so a peer that accepts connections and then stalls would park every *later* caller for three minutes behind an exhausted pool. Shed load instead, matching `connector.api-client.pending-acquire-timeout`. |
| Proxy | `SystemDefaultRoutePlanner` + `SystemDefaultCredentialsProvider` | unchanged | Already correct — the builder resolves through the JVM `ProxySelector` and default `Authenticator`, both populated by `ProxyConfiguration`. |

**Connect and read timeouts are deliberately not set here.** Boot applies the connection-config customizer *after* mapping the caller's `ClientHttpRequestFactorySettings`, so setting them in the shared builder would silently overwrite what a consumer passed to `build(settings)` — which is exactly how `ProvisioningApiConfig` supplies its `provisioning.api.connect-timeout` / `read-timeout`. A test pins that ordering.

## OAuth2-specific

| Setting | Reason |
|---|---|
| Connect 5s, read 10s | Apache leaves connect at 3 minutes and the read timeout **unbounded**. A provider that accepts a connection and then stalls would hold pooled connections indefinitely and, once the pool is exhausted, stall authentication platform-wide. |
| Apache, not the JDK client | The JDK client ignores the JVM default `Authenticator` that `ProxyConfiguration` installs, and silently **strips the bearer token** once an authenticator is registered on it. Both verified empirically. |
| 3xx status handler | Apache refuses cross-authority redirects on `Authorization`-bearing requests and returns the raw 3xx, which made `retrieve()` yield a `null` body and drop the userinfo claims with no exception and no log. The query string is stripped from the message so the logout URI's `id_token_hint` cannot reach the logs. |

## Behavioural deltas worth naming

- **Provisioning's concurrency ceiling toward its backend goes 5 → 100.** That 5 was Apache's default, not a deliberate bulkhead — flagging it in case that backend is small.
- **Provisioning loses the 429/503 transport retry, and is not weakened by it.** Its API clients already carry `@Retryable(retryFor = RestClientException.class)`, which both statuses satisfy, so retries now use the configured backoff instead of a peer-dictated stall.

## Tests

`getUserInfo` becomes package-private so the error mapping of `OAuth2ErrorResponseErrorHandler` is observable — both callers swallow every exception, so nothing above it can assert on the HTTP contract.

`OAuth2UtilUserInfoTest` covers the success path and bearer/`Accept` headers, `400` → `OAuth2AuthorizationException`, `500` → `HttpServerErrorException`, an end-to-end authenticating forward proxy (asserting both `Proxy-Authorization` and `Authorization`), no cookie replay, no retry on `503`, a stalled provider failing within the read timeout rather than hanging, an unfollowed redirect surfaced as an error, that redirect message being free of the query string, and one retry after a connection reset. `OAuth2UtilTest` gains coverage of the userinfo claim merge and its fallback, with request verification so the assertions cannot pass on a swallowed exception.

`PlatformHttpClientsTest` guards the shared defaults: 40 simultaneous requests (a literal, not derived from the production constant, so lowering it back towards 5 breaks the test), no cookie replay, no retry on `503`, one retry after a connection reset, fail-fast once the pool is exhausted, and the consumer's own read timeout surviving the shared customizer.

Each guard was checked by reverting its fix — timings are asserted, not just exception types, because an unbounded timeout still throws eventually and would otherwise pass.

Full suite: **4633 run, 0 failures, 2 skipped**.


ivosh:
## Follow-up: shared the hardened defaults across both `RestClient`s (de442d76)

The migration commit hand-rolled an Apache `HttpClient` inside `OAuth2Util`. Reviewing that raised a fair question — does the other `RestClient` consumer need the same treatment, and can it be shared? Both answers are yes.

I probed what `ClientHttpRequestFactoryBuilder.detect()` (used by `ProvisioningApiConfig`) actually produces:

| Default | `detect()` before | Now |
|---|---|---|
| Pool | `maxPerRoute=5`, `maxTotal=25` | 100 / 100 |
| Cookies | `BasicCookieStore`, confirmed replaying between calls | disabled |
| Retries | `HttpRequestRetryExec`, honours `Retry-After` verbatim | idempotent I/O-error retry only |
| Proxy | `SystemDefaultRoutePlanner` + `SystemDefaultCredentialsProvider` | unchanged — already correct |

New `PlatformHttpClients.requestFactoryBuilder()` carries these; `OAuth2Util` and `ProvisioningApiConfig` both build from it. Since Boot's builder already applies `useSystemProperties()` semantics, the explicit call was redundant — `OAuth2Util` drops 48 lines and keeps only its response mapping.

**Provisioning is not weakened by losing the 429/503 transport retry**: its API clients already carry `@Retryable(retryFor = RestClientException.class)`, which both statuses satisfy, so retries now use the configured backoff instead of a peer-dictated stall. The one behavioural delta worth naming: its concurrency ceiling toward the provisioning backend goes 5 to 100. That 5 was Apache's default, not a deliberate bulkhead — flagging it in case that backend is small.

`PlatformHttpClientsTest` pins all three guards; mutation check: reverting `requestFactoryBuilder()` to a bare builder fails 3 of its 4 tests (and takes 35s instead of 0.05s, from the `Retry-After` sleep).

Full suite: **4553 passed, 0 failures**.

src/test/java/com/otilm/core/util/OAuth2UtilUserInfoTest.java
```diff
@@ -0,0 +1,210 @@
...
+     */
+    @Test
+    void testGetUserInfo_AuthenticatesAgainstConfiguredForwardProxy() {
+        String previousHost = System.getProperty("http.proxyHost");
+        String previousPort = System.getProperty("http.proxyPort");
+        MultiServerAuthenticator proxyAuthenticator = new MultiServerAuthenticator();
+        proxyAuthenticator.add("localhost:" + mockServer.port(), "proxy-user", "proxy-password");
+        Authenticator.setDefault(proxyAuthenticator);
+        System.setProperty("http.proxyHost", "localhost");
+        System.setProperty("http.proxyPort", String.valueOf(mockServer.port()));
```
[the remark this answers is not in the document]
ivosh: Fixed in eb87edfe.

src/main/java/com/otilm/core/util/OAuth2Util.java
```diff
@@ -28,6 +32,43 @@ public class OAuth2Util {
...
+     */
+    private static final RestClient restClient = buildRestClient();
+
+    /**
+     * Pool sizing, cookie handling, proxy wiring and retry policy come from
+     * {@link PlatformHttpClients}; what is specific here is how the provider's responses are mapped.
+     */
+    private static RestClient buildRestClient() {
+        return RestClient.builder()
+                .requestFactory(PlatformHttpClients.requestFactoryBuilder().build())
```
lubomirw: **[Medium]** | Error handling

OAuth2 client inherits a three-minute connect timeout and no read timeout, so a stalled identity provider can consume pooled connections and block authentication requests indefinitely.

_Suggested:_ Build the OAuth2 request factory with explicit bounded connect/read settings and add slow-provider timeout tests. If configurable, expose it through Settings UI.

<sub>flagged by Superpowers; validated by multi-review triage</sub>

src/test/java/com/otilm/core/config/http/PlatformHttpClientsTest.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Guards the request-factory defaults every platform {@code RestClient} inherits. A bare
+ * {@code ClientHttpRequestFactoryBuilder.detect()} fails each of these.
+ */
+class PlatformHttpClientsTest {
+
+    private static final int CONCURRENT_REQUESTS = 10;
```
lubomirw: **[Low]** | Testing

The concurrency test uses 10 requests, so it catches Apache's default limit of 5 but does not pin the PR's measured 40-request requirement or 100-connection configuration.

_Suggested:_ Exercise at least 40 simultaneous requests independently of the production constant and verify the configured maximum remains 100.

<sub>flagged by Guided PR review; validated by multi-review triage</sub>

src/test/java/com/otilm/core/util/OAuth2UtilUserInfoTest.java
```diff
@@ -0,0 +1,211 @@
...
+
+    @Test
+    void testGetUserInfo_DeserializesBodyAndSendsBearerToken() {
+        mockServer.stubFor(WireMock.get(WireMock.urlPathEqualTo("/userinfo"))
+                .willReturn(WireMock.okJson("{\"sub\":\"abc\",\"groups\":[\"a\",\"b\"]}")));
+
+        Map<String, Object> userInfo = OAuth2Util.getUserInfo(userInfoUrl, "the-token");
+
+        Assertions.assertEquals("abc", userInfo.get("sub"));
+        Assertions.assertEquals(java.util.List.of("a", "b"), userInfo.get("groups"));
```
lubomirw: **[Nit]** | Maintainability

This assertion uses the fully qualified java.util.List even though no simple-name collision exists.

_Suggested:_ Import java.util.List and use List.of(...).

<sub>flagged by Hygiene; validated by multi-review triage</sub>

lubomirw (commented):
Automated multi-reviewer triage - 3 findings (COMMENT).

src/test/java/com/otilm/core/util/OAuth2UtilUserInfoTest.java
ivosh: Fixed.

src/main/java/com/otilm/core/util/OAuth2Util.java
ivosh: Confirmed and fixed.

On the Settings UI suggestion — declining, and want to flag the reasoning in case you disagree.  OAuth2Util  and  PlatformHttpClients  are static utilities initialised at class-load, with no Spring context to read a settings store from. More to the point, the codebase already treats HTTP transport tuning as infrastructure:  validation.crl  and  validation.ocsp  timeouts are plain yml with no env var at all, and  connector.api-client  sets connect/response/pending-acquire the same way. These are failure-mode bounds, not something an operator should be tuning per tenant. Happy to revisit if you see a concrete case for making them operator-visible.

src/test/java/com/otilm/core/config/http/PlatformHttpClientsTest.java
ivosh: Raised to 40, kept as a literal rather than deriving it from  PlatformHttpClients.MAX_CONNECTIONS  — your point about independence is the right one; derived, the test couldn't catch the constant being lowered back towards Apache's 5.

Declining the second half.  assertEquals(100, MAX_CONNECTIONS)  reads the very constant it asserts on, so it can only ever pass — it pins the literal, not the behaviour.
What it needs instead is a test that actually exercises the full 100, and there now is one:  `testFailsFastWhenPoolIsExhausted` saturates the pool to  MAX_CONNECTIONS , then asserts the next caller is shed rather than parked.

lubomirw (approved):

PR #1968, opened 2026-08-05
Complete pre-registered certificates over SCEP enrolment

### LINKED ISSUE #1775: Persist SAN on pre-registered certificates and verify SCEP requests against pre-registrations
### Description

Two related pieces of the certificate registration flow (epic ilm#130): persist the pre-registration SAN on the placeholder, and verify SCEP enrollments against pre-registrations.

### Scope

- **SAN persistence:** extend `createRegistrationPlaceholder` to store the requested SAN — a converter between the register request's OpenSSL SAN string (`RegisterWireBuilder`, `SAN_PREFIXES`/`GeneralNameType`) and the certificate entity's serialized SAN map (`CertificateUtil`, `SAN_TYPE_MAP`). Needed for inventory visibility and SCEP subject/SAN matching.
- **SCEP:** honor `ScepProfile.challengeSource`; on a SCEP request, identity-tuple match (normalized subject DN, or exact SAN-set for subject-less) against `REGISTERED` placeholders under the profile's RA profile; verify the PKCS#9 `challengePassword`; uniform failure shape (no oracle); binding step in `ScepServiceImpl`. RSA-CA scope.

### Dependencies

Builds on core#1570 and the SCEP contract addition (corresponding interfaces issue).

klaraf755:
Closes #1775
Part of OmniTrustILM/ilm#311 (SCEP, the issue's designated first protocol; CMP/ACME/EST follow separately). Requires OmniTrustILM/interfaces#813 (`ScepChallengeSource` contract).

A SCEP profile can now gate enrolment on certificate pre-registrations instead of a shared challenge password. With `challengeSource = CERTIFICATE_REGISTRATION`, an initial enrolment is matched to a `REGISTERED` placeholder under the profile's RA profile — subject compared by normalized rendering, SAN set as tiebreak and then exact-equality check — and completed through `issueExistingCertificate` with the CSR's `challengePassword` as the authorization secret, so the existing registration challenge gate (row lock, failed-attempt lockout, issuance window) authorizes issuance of exactly the pre-registered identity. Renewals keep the possession-based waiver.

Security properties, each pinned by a test:
- Every rejection (no match, ambiguous, SAN mismatch, wrong/locked/expired challenge) is wire-identical — `badMessageCheck` with one generic message — so an unauthenticated prober cannot enumerate registrations; reasons go to logs (with the CSR identity) and certificate event history.
- Only a uniquely matched registration ever has its challenge checked, and exactly once per attempt.
- Hardening on the shared completion path: a presented `authorizationSecret` that no ACTIVE authorization verifies is rejected — the operator-permission fallback applies only when no secret was presented. This closes the match-then-gate race and stops unverifiable secrets being silently ignored.
- Registration mode requires an RSA CA certificate (a non-RSA CA key forces the RFC 8894 password recipient, undecryptable without a shared password); profile validation also rejects a configured password and Intune, and switching a profile into registration mode clears any stored password.

Known limitation: EC-keyed clients cannot renew in registration mode (a renewal carries no challenge to envelope the response with) — rejected with `badAlg`, covered by a test.

Testing: pure matching kernel unit tests, SCEP-layer unit tests, and an end-to-end ITest (RSA CA, real CMS messages, per-message connector-decrypt stub) reusing an existing context signature — `ContextSignatureGuardTest` baseline unchanged. Manually verified with sscep against a live authority, including SAN matching and the negative paths.

The FE selector for the new profile field is tracked separately in `fe-administrator`.


src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -849,14 +907,160 @@ private void prepareMessage(ScepRequest scepRequest, ScepResponse scepResponse)
...
+     * Completes the matched pre-registration through the standard completion operation, presenting the
+     * CSR's challenge as the authorization secret — the challenge gate, CSR attach and ISSUE enqueue are
+     * the same as for an operator completion. Registration completion is always asynchronous: the client
+     * receives PENDING and polls the transaction.
+     */
+    private ScepResponse completeRegistration(ScepRequest scepRequest) throws ScepException {
+        ClientCertificateIssueRequestDto requestDto = new ClientCertificateIssueRequestDto();
+        try {
+            requestDto.setRequest(new String(Base64.getEncoder().encode(scepRequest.getPkcs10Request().getEncoded())));
+        } catch (IOException e) {
```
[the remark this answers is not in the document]
klaraf755: Changed to `endoceToString`.

src/main/java/com/otilm/core/dao/repository/CertificateRepository.java
```diff
@@ -193,6 +193,22 @@ AND NOT EXISTS (SELECT 1 FROM CertificateStatusPoll p WHERE p.certificateUuid =
...
+     * whose registration authorization is ACTIVE. Registrations without a challenge have no authorization
+     * row and are excluded — they cannot authenticate an enrolment.
+     */
+    @Query("""
+            SELECT c FROM Certificate c
+                WHERE c.raProfileUuid = :raProfileUuid
+                    AND c.state = ?#{T(com.otilm.api.model.core.certificate.CertificateState).REGISTERED}
+                    AND EXISTS (SELECT 1 FROM CertificateRegistrationAuthorization a
+                        WHERE a.certificateUuid = c.uuid
+                            AND a.state = ?#{T(com.otilm.core.dao.entity.RegistrationState).ACTIVE})
```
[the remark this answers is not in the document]
klaraf755: Added the filter.

src/test/java/com/otilm/core/service/scep/impl/ScepServiceImplRegistrationModeTest.java
```diff
@@ -0,0 +1,327 @@
...
+        when(request.getPkcs10Request()).thenReturn(csr);
+        when(request.getChallengePassword()).thenReturn(challengePassword);
+        return request;
+    }
+
+    private static JcaPKCS10CertificationRequest csr(String subjectDn, String... dnsSans) throws Exception {
+        JcaPKCS10CertificationRequestBuilder builder =
+                new JcaPKCS10CertificationRequestBuilder(new X500Name(subjectDn), keyPair.getPublic());
+        builder.addAttribute(PKCSObjectIdentifiers.pkcs_9_at_challengePassword, new DERPrintableString(CHALLENGE));
+        if (dnsSans.length > 0) {
```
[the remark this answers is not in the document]
klaraf755: The unconditional `challengePassword` attribute in `csr(...)` was indeed inconsistent with the scenarios that stub a missing/incorrect presented challenge — the mock severed the CSR→accessor link, so no assertion was checking the wrong thing, but the fixture was misleading. Merged the CSR builder and the `ScepRequest` stub into one helper whose single challenge argument drives both, so the CSR content and the presented challenge cannot diverge (a missing challenge now yields a CSR without the attribute; a wrong one is embedded as that wrong value). Fixed in 2e273f52c.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -133,10 +143,17 @@ public class ScepServiceImpl implements ScepExternalService {
     private RaProfile raProfile;
     private List<RequestAttribute> issueAttributes;
     private ScepProfile scepProfile;
+    /** The pre-registered certificate a registration-mode enrolment was matched to; set per request. */
+    private Certificate matchedRegistration;
```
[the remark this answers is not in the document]
klaraf755: The singleton-with-per-request-fields pattern is pre-existing in `ScepServiceImpl` (`scepProfile`, `raProfile`, `caCertificateChain`, … are all populated per request by `init()`) and deserves a separate refactor — but this PR's new field didn't need to participate in it: `matchedRegistration` is now a local threaded as a parameter from the match to the completion, so the PR adds no shared mutable state and the overwrite scenario cannot involve the registration match. Fixed in edf868ce8.

src/test/java/com/otilm/core/integration/service/scep/ScepRegistrationEnrolmentITest.java
```diff
@@ -0,0 +1,515 @@
...
+
+    /**
+     * The connector decrypt must return the real content-encryption key or the CSR can never be read: the
+     * key is extracted from the message's KeyTransRecipientInfo and decrypted locally with the test-held CA
+     * private key, and the stub returns exactly that value. Later stubs override earlier ones, so each
+     * message gets its own.
+     */
+    private void stubConnectorDecrypt(byte[] scepMessage) throws Exception {
+        byte[] encryptedKey = extractEncryptedContentEncryptionKey(scepMessage);
+        Cipher cipher = Cipher.getInstance("RSA/ECB/PKCS1Padding");
```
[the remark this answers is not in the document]
klaraf755: KCS#1 v1.5 is the CMS key-transport scheme SCEP clients produce (rsaEncryption); the test must unwrap the envelope it built with the matching scheme, using an ephemeral throwaway key. OAEP would not decrypt the message under test.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -849,14 +904,160 @@ private void prepareMessage(ScepRequest scepRequest, ScepResponse scepResponse)
...
+        try {
+            requestDto.setRequest(new String(Base64.getEncoder().encodeToString(scepRequest.getPkcs10Request().getEncoded())));
+        } catch (IOException e) {
+            throw new ScepException("Unable to decode PKCS#10 request", e, FailInfo.BAD_REQUEST);
+        }
+        requestDto.setFormat(CertificateRequestFormat.PKCS10);
+        requestDto.setAttributes(issueAttributes);
+        requestDto.setAuthorizationSecret(scepRequest.getChallengePassword());
+        try {
+            clientOperationExternalService.issueExistingCertificate(
```
lubomirw, line 1044: **[High]** | Security

A replayed or concurrent registration completion can replace the CSR because challenge verification and certificate claiming are separate, and CSR attachment leaves the certificate REGISTERED.

_Suggested:_ Atomically claim the certificate under its row lock, reject an existing CSR, and add a concurrent different-key replay test.

<sub>flagged by Copilot CLI and Superpowers</sub>

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -849,14 +904,160 @@ private void prepareMessage(ScepRequest scepRequest, ScepResponse scepResponse)
...
+                    matchedRegistration.getUuid().toString(),
+                    requestDto);
+        } catch (ValidationException | NotFoundException e) {
+            // Denial detail (locked authorization, expired window, wrong challenge) stays in the log and
+            // the certificate event history; the wire carries the anti-enumeration text so challenge
+            // failures are indistinguishable from lookup misses.
+            logger.info("SCEP registration completion rejected: {}", e.getMessage());
+            throw new ScepException(REGISTRATION_REJECTION, FailInfo.BAD_MESSAGE_CHECK);
+        }
+        applyProtocolAssociationBestEffort(matchedRegistration);
```
lubomirw, line 1063: **[High]** | Correctness

ISSUE is already enqueued before protocol association and SCEP poll tracking become durable; a joined runtime failure or later outer-transaction failure can roll back tracking while issuance continues.

_Suggested:_ Persist the completion claim, protocol attribution, poll mapping, and durable outbox before publishing; isolate genuinely best-effort writes behind separate transaction proxies.

<sub>flagged by Copilot CLI, Guided PR review, and Superpowers</sub>

src/test/java/com/otilm/core/integration/service/scep/ScepRegistrationEnrolmentITest.java
```diff
@@ -0,0 +1,515 @@
...
+
+        ResponseEntity<Object> response = postPkiOperation(enrolment(SUBJECT_DN, List.of("device-1.example"), CHALLENGE));
+
+        assertScepFormatted(response);
+        assertEquals(String.valueOf(PkiStatus.PENDING.getValue()), attribute(response, ScepConstants.id_pkiStatus));
+        Certificate completed = certificateRepository.findByUuid(placeholder.getUuid()).orElseThrow();
+        assertEquals(CertificateState.REGISTERED, completed.getState(),
+                "the placeholder stays REGISTERED until the async ISSUE action completes");
+        assertNotNull(completed.getCertificateRequestUuid(), "the enrolment CSR is attached");
+        assertNotNull(completed.getProtocolAssociation(), "the completion is attributed to SCEP");
```
lubomirw: **[High]** | Testing

This added E2E test dereferences a detached lazy protocolAssociation and fails with LazyInitializationException.

_Suggested:_ Reload through findWithAssociationsByUuid or query the association repository before asserting.

<sub>flagged by Guided PR review; independently reproduced</sub>

src/main/java/com/otilm/core/service/registration/RegistrationIdentityMatcher.java
```diff
@@ -0,0 +1,127 @@
...
+                ? MatchResult.matched(sanMatches.getFirst().certificateUuid())
+                : MatchResult.ambiguous();
+    }
+
+    private static String normalize(X500Name subject) {
+        return X500Name.getInstance(PlatformX500NameStyle.NORMALIZED, subject).toString();
+    }
+
+    private static boolean sansEqual(Map<String, List<String>> normalizedCsrSans, Candidate candidate) {
+        return normalizedCsrSans.equals(normalizeSans(CertificateUtil.deserializeSans(candidate.serializedSans())));
```
lubomirw, line 131: **[Medium]** | Correctness

String-based SAN comparison rejects equivalent identities: CSR IP addresses render as hex while registration values are decoded text, and DNS case, equivalent IP forms, and duplicates remain unnormalized.

_Suggested:_ Compare canonical per-type or DER GeneralName sets and add IP, DNS-case, and duplicate-value tests.

<sub>flagged by Guided PR review and Superpowers</sub>

src/main/java/com/otilm/core/service/registration/RegistrationIdentityMatcher.java
```diff
@@ -0,0 +1,127 @@
...
+     */
+    public static MatchResult match(X500Name csrSubject, Map<String, List<String>> csrSans, List<Candidate> candidates) {
+        String normalizedCsrSubject = normalize(csrSubject);
+        Map<String, List<String>> normalizedCsrSans = normalizeSans(csrSans);
+
+        List<Candidate> subjectMatches = new ArrayList<>();
+        for (Candidate candidate : candidates) {
+            String normalizedCandidateSubject;
+            try {
+                normalizedCandidateSubject = normalize(new X500Name(PlatformX500NameStyle.NORMALIZED, candidate.subjectDn()));
```
lubomirw: **[Medium]** | Correctness

A valid SAN-only registration has a null or blank candidate subject, which throws during parsing and is skipped before SAN comparison.

_Suggested:_ Normalize absent subjects to the empty name, match through SAN equality, and add subjectless unit and enrolment tests.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -849,14 +904,160 @@ private void prepareMessage(ScepRequest scepRequest, ScepResponse scepResponse)
...
+            String presented = scepRequest.getChallengePassword();
+            if (presented != null && !presented.isEmpty()) {
+                return presented;
+            }
+        }
+        return resolvePollEnvelopePassword(scepRequest);
+    }
+
+    private String resolvePollEnvelopePassword(ScepRequest scepRequest) {
+        ScepTransaction transaction = getTransaction(scepRequest.getTransactionId());
```
lubomirw, line 942: **[Medium]** | Correctness

Registration-mode poll password recovery uses a global transaction-ID lookup even though IDs can collide across SCEP profiles.

_Suggested:_ Use findByTransactionIdAndScepProfile(transactionId, scepProfile) and add a cross-profile collision test.

src/main/java/com/otilm/core/dao/repository/CertificateRepository.java
```diff
@@ -193,6 +193,23 @@ AND NOT EXISTS (SELECT 1 FROM CertificateStatusPoll p WHERE p.certificateUuid =
...
+            SELECT c FROM Certificate c
+                WHERE c.raProfileUuid = :raProfileUuid
+                    AND c.state = ?#{T(com.otilm.api.model.core.certificate.CertificateState).REGISTERED}
+                    AND c.archived = false
+                    AND EXISTS (SELECT 1 FROM CertificateRegistrationAuthorization a
+                        WHERE a.certificateUuid = c.uuid
+                            AND a.state = ?#{T(com.otilm.core.dao.entity.RegistrationState).ACTIVE})
+            """
+    )
+    List<Certificate> findRegisteredWithActiveRegistrationAuthorizationByRaProfileUuid(@Param("raProfileUuid") UUID raProfileUuid);
```
lubomirw, line 211: **[Medium]** | Performance

Every unauthenticated initial enrolment loads and parses all active registrations under the RA profile, making request cost grow linearly with inventory.

_Suggested:_ Prefilter by normalized subject in SQL, add a suitable composite or partial index, and SAN-disambiguate only the narrowed candidates.

src/test/java/com/otilm/core/integration/service/scep/ScepRegistrationEnrolmentITest.java
```diff
@@ -0,0 +1,515 @@
...
+        registeredPlaceholder(ScepMessageTestData.SUBJECT_DN, null, CHALLENGE);
+
+        // No decrypt stub: the profile has no shared password, so a password-recipient request is
+        // undecryptable by construction and must be answered as a SCEP failure.
+        ResponseEntity<Object> response = scepService.handlePost(
+                SCEP_PROFILE_NAME, ScepServiceImpl.SCEP_OPERATION_PKI_OPERATION,
+                ScepMessageTestData.passwordEnvelopedPkcsReq(CHALLENGE));
+
+        assertScepFormatted(response);
+        assertEquals(String.valueOf(PkiStatus.FAILURE.getValue()), attribute(response, ScepConstants.id_pkiStatus));
```
lubomirw, line 422: **[Low]** | Testing

This test asserts only generic FAILURE, so broad error handling can make it pass for an unrelated setup, connector, or response-building failure.

_Suggested:_ Assert the intended BAD_REQUEST and failure text path, verify no state mutation, and verify the expected connector interaction.

src/test/java/com/otilm/core/service/registration/RegistrationIdentityMatcherTest.java
```diff
@@ -0,0 +1,163 @@
...
+class RegistrationIdentityMatcherTest {
+
+    private static final UUID CANDIDATE_A = UUID.randomUUID();
+    private static final UUID CANDIDATE_B = UUID.randomUUID();
+
+    private static Map<String, OidRecord> savedRdnCache;
+
+    @BeforeAll
+    static void snapshotAndSeedRdnCache() {
+        // Snapshot the original global cache BEFORE seeding.
```
lubomirw: **[Nit]** | Maintainability

The cache snapshot and seeding comments restate the setup method and following calls.

_Suggested:_ Remove the two redundant comments if this test should differ from the established sibling-test pattern.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/service/scep/impl/ScepServiceImplRegistrationModeTest.java
```diff
@@ -0,0 +1,329 @@
...
+    private CertificateInternalService certificateService;
+
+    @BeforeAll
+    static void setUpClass() throws Exception {
+        if (Security.getProvider(BouncyCastleProvider.PROVIDER_NAME) == null) {
+            Security.addProvider(new BouncyCastleProvider());
+        }
+        keyPair = KeyPairGenerator.getInstance("RSA").generateKeyPair();
+
+        // Snapshot the original global cache BEFORE seeding.
```
lubomirw: **[Nit]** | Maintainability

The cache snapshot and seeding comments restate the setup method and following calls.

_Suggested:_ Remove the two redundant comments if the ordering rationale is not needed.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/integration/service/scep/ScepRegistrationEnrolmentITest.java
```diff
@@ -0,0 +1,515 @@
...
+    }
+
+    private void assertScepFormatted(ResponseEntity<Object> response) throws Exception {
+        assertEquals("application/x-pki-message", response.getHeaders().getFirst("Content-Type"),
+                "a SCEP client can only parse application/x-pki-message");
+        assertNotNull(response.getBody());
+        assertNotNull(new CMSSignedData((byte[]) response.getBody()).getSignerInfos().getSigners());
+    }
+
+    /** Reads a SCEP attribute out of the signed response. */
```
lubomirw: **[Nit]** | Maintainability

Private-helper Javadoc restates that attribute reads a SCEP response attribute.

_Suggested:_ Remove the Javadoc if consistency with the existing SCEP integration test is not required.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/integration/service/scep/ScepRegistrationEnrolmentITest.java
```diff
@@ -0,0 +1,515 @@
...
+                .setProvider(BouncyCastleProvider.PROVIDER_NAME).getCertificate(builder.build(signer));
+    }
+
+    /**
+     * The token connector signs the response. The value is passed through without verification here, so a
+     * fixed well-formed signature is enough to exercise the real response-building path.
+     */
+    private void stubTokenSigning() throws Exception {
+        KeyPair throwaway = KeyPairGenerator.getInstance("EC").generateKeyPair();
+        java.security.Signature signature = java.security.Signature.getInstance("SHA256withECDSA");
```
lubomirw: **[Nit]** | Maintainability

java.security.Signature is used as a fully qualified inline type despite no simple-name collision.

_Suggested:_ Import java.security.Signature and use the simple name for both references.

<sub>flagged by Hygiene</sub>

lubomirw (changes requested):
Automated multi-reviewer triage — 13 selected findings (REQUEST_CHANGES).

Overflow findings not anchorable in the Core diff:
- #9 documentation/docs/certificate-key/protocols/scep/scep-profile.md:13 — SCEP documentation omits challenge source, default, identity matching, RSA/password/Intune constraints, and the EC-renewal limitation. Suggested: document registration mode before release. (flagged by Docs check via all three code reviewers)

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
klaraf755: The overwrite was possible because completion checked the challenge and attached the CSR as two separately-locked steps and left the cert REGISTERED, so a second completion (a fresh-transaction-id replay or a concurrent one) with the same identity/challenge but a different key overwrote the first CSR before the async ISSUE action claimed the row.

Rather than transition state — the register-bound and plain ISSUE actions require the placeholder to still be REGISTERED — `addCertificateRequestToExisting` now rejects a second attach while holding `SELECT … FOR UPDATE`, so only the first binds. To keep the "restored to REGISTERED so the holder can retry" path working after an approval rejection, `issueCertificateRejectedAction` clears the attached request/key on restore.

Test `differentKeyReplayCannotReplaceTheAttachedCsr`: a second enrolment with the same subject/SAN and challenge but a fresh key and transaction id is rejected and the first CSR is unchanged. The row lock serializes concurrent completions into that same attach, so the sequential test covers the race.

Fixed in 0150ffab6.

src/test/java/com/otilm/core/integration/service/scep/ScepRegistrationEnrolmentITest.java
klaraf755: Fixed in 0150ffab6.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
klaraf755: Fixed. The poll mapping and attribution joined the request's outer transaction, so they only became durable *after* `issueExistingCertificate` published the ISSUE message (a non-transactional JMS send) — a later commit failure rolled back the tracking while issuance proceeded, stranding the client's transactionId.

A new `ScepRegistrationTrackingWriter` records the poll mapping in its own `REQUIRES_NEW` transaction before the publish (a separate bean is needed — `REQUIRES_NEW` is skipped on self-invocation). Nothing throws after the enqueue, so any failure before it discards the staged mapping and the client can retry; success keeps it. Attribution commits in its own transaction after success (best-effort).

Fixed in 3d6ef95fd.

src/main/java/com/otilm/core/service/registration/RegistrationIdentityMatcher.java
klaraf755: Fixed. `normalizeSans` now canonicalizes each value per type before comparison, and it runs on both the CSR and registration sides so they converge:
- **IP** → octet hex via BouncyCastle (`#c0a80101`), the same form `getSAN` renders from a CSR, so a registration's decoded text and equivalent IPv6 forms collapse to the same value. Parsing is purely lexical — no DNS lookup. An unparseable value is left untouched so a genuine mismatch still mismatches.
- **DNS** → lowercased (case-insensitive per RFC 5280).
- Duplicates dropped, order removed.

Tests: IP hex-vs-text, equivalent IPv6 forms, different-IP mismatch, DNS case-insensitivity, and duplicate collapse in `RegistrationIdentityMatcherTest`, plus `CertificateUtilTest.getSanRendersCsrIpAddressAsOctetHex` grounding that a CSR IP SAN really reads back as octet hex.

Fixed in 5a251da73.

src/main/java/com/otilm/core/service/registration/RegistrationIdentityMatcher.java
klaraf755: Fixed. A null or blank candidate subject now normalizes to the empty name instead of throwing in `new X500Name(...)` and being skipped, so a SAN-only registration matches a SAN-only enrolment through SAN equality. Both sides go through the same `normalize`, so an empty CSR subject and an absent registration subject converge to the same empty rendering; a null `csrSubject` is treated as empty too.

A present-but-malformed DN still throws and is skipped (the malformed-row protection is unchanged), and a non-empty subject cannot collide with the empty rendering in either direction — so this does not widen matching for certs that do carry a subject.

Tests: `subjectlessRegistrationMatchesBySans`, `blankSubjectRegistrationIsTreatedAsSubjectless`, and `subjectlessRegistrationWithDifferentSansIsSanMismatch` in `RegistrationIdentityMatcherTest`, plus `subjectlessRegistrationCompletesBySanOnly` end-to-end in `ScepRegistrationEnrolmentITest`.

Fixed in ca7614b30.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
klaraf755: Fixed. `getTransaction` now uses `findByTransactionIdAndScepProfile(transactionId, scepProfile)`, so the poll fetch is scoped to the polling profile — matching the exists-check in `resolveResponse`. A colliding transaction id can no longer resolve another profile's transaction or recover its challenge. Test `transactionLookupIsScopedByScepProfile` asserts same-id transactions under two profiles resolve independently.

Fixed in fa1d390cd.

src/main/java/com/otilm/core/dao/repository/CertificateRepository.java
klaraf755: The matcher is deliberately correctness-scoped here. A subject prefilter would key on `subjectDnNormalized`, which is only populated when an issued X509 certificate is parsed — registration placeholders do not set it. Adding the prefilter therefore requires populating it (matcher-consistently) on registration creation *and* a backfill migration for existing placeholders, or they would stop matching. That is a worthwhile inventory-scale optimization but out of scope for this change; tracking it as a follow-up rather than blocking.

src/test/java/com/otilm/core/service/registration/RegistrationIdentityMatcherTest.java
klaraf755: Done in e1592aab8 — removed the redundant comments.

src/test/java/com/otilm/core/service/scep/impl/ScepServiceImplRegistrationModeTest.java
klaraf755: Done in e1592aab8 — removed the redundant comments.

src/test/java/com/otilm/core/integration/service/scep/ScepRegistrationEnrolmentITest.java
klaraf755: Done in e1592aab8 — removed the redundant comments.

src/test/java/com/otilm/core/integration/service/scep/ScepRegistrationEnrolmentITest.java
klaraf755: Done in e1592aab8 — imported `java.security.Signature` and use the simple name.

src/test/java/com/otilm/core/integration/service/scep/ScepRegistrationEnrolmentITest.java
klaraf755: Strengthened. The test now asserts `FailInfo.BAD_REQUEST` and, crucially, that the request was rejected before any completion — the placeholder stays REGISTERED with no CSR attached and no ISSUE is enqueued (`verify(actionProducer, never())`). So it can no longer pass on an unrelated setup, connector, or response-building failure that merely ends in FAILURE.

Fixed in c8c44a050.

src/main/java/com/otilm/core/dao/repository/CertificateRepository.java
lubomirw: Please link follow up issue here

lubomirw (approved):

src/main/java/com/otilm/core/dao/repository/CertificateRepository.java
klaraf755: Follow up issue: https://github.com/OmniTrustILM/core/issues/1988.

PR #1989, opened 2026-08-07
Complete pre-registered certificates over CMP enrolment

### LINKED ISSUE #1933: Verify ACME, EST and CMP enrolments against pre-registrations
### Description

core#1775 honours `ScepProfile.challengeSource` and verifies SCEP enrolments against a pre-registration. The same capability is missing for the other enrolment protocols, so a pre-registered certificate can only be completed over SCEP.

### User Story

As an operator, I want an ACME, EST or CMP client to complete a pre-registered certificate, so the registration authorization gates issuance regardless of which protocol the endpoint speaks.

### Acceptance Criteria

- [ ] ACME, EST and CMP honour the profile's challenge source.
- [ ] When set to certificate registration, the enrolment is matched against an active pre-registration (subject and SAN) instead of the shared profile secret.
- [ ] A failed match is rejected with the protocol's native error, not a generic failure.
- [ ] Per-protocol behaviour documented where the semantics differ (ACME account binding, CMP protection, EST client auth).

### Dependencies

Builds on core#1775 (SCEP) and the contract additions for the remaining protocol profiles.

klaraf755:
Closes #1933 
Part of OmniTrustILM/ilm#311 (CMP), building on the SCEP delivery #1968 (now merged to `main`) — it reuses the shared pre-registration core (`ProtocolChallengeSource`, `RegistrationIdentityMatcher`, the registration authorization + challenge store, and the `issueExistingCertificate` completion path). Uses the `ProtocolChallengeSource` contract from OmniTrustILM/interfaces#813.

A CMP profile (V2, shared-secret request protection) can now gate enrolment on certificate pre-registrations instead of a stored shared secret. With `challengeSource = CERTIFICATE_REGISTRATION`, the RFC 4210 `senderKID` carries the pre-registered certificate's UUID and **the registration challenge is the PBM MAC key** — there is no password field; producing a verifying MAC *is* the authentication. `ir`/`cr` completes a `REGISTERED` placeholder, MAC-protected `kur` rekeys the already-issued certificate, and the asynchronous `pollReq`/`certConf` follow-ups authenticate against the same challenge.

An enrolment passes two independent checks: the MAC verifies under the registration's challenge — resolved from the `senderKID` and run through the shared `RegistrationChallengeGate` (per-row lock, failed-attempt lockout, issuance window) — **and** the CRMF template's subject + SAN set exactly equal the registered identity (`RegistrationIdentityMatcher`). Completion runs through `issueExistingCertificate` with the challenge as the authorization secret; the response is MAC-keyed with the same challenge so the client can verify it.

Security properties, each pinned by a test:
- Every rejection (no match, wrong/locked/expired challenge, identity mismatch, unsupported body) is wire-identical — `badMessageCheck` with one generic message — so a prober cannot enumerate registrations; reasons go to logs (with the `senderKID`) and certificate event history.
- Only a uniquely resolved registration ever has its challenge checked, counted once per attempt (lockout preserved).
- Registration mode stores no shared secret, so a MAC body that resolves no challenge (e.g. a forged revocation) is rejected rather than allowed to verify against an empty key.
- The async follow-ups (`pollReq`/`certConf`) authenticate against the challenge that survives issuance, without the strict enrolment cert-state check but still gate-verified — so the client can collect an asynchronously-issued certificate without a bypass.
- Profile validation requires V2 + shared-secret request protection and forbids a stored shared secret; rejects `V2_3GPP`/`V3`; the `challenge_source` column is NOT NULL, default `PROTOCOL_DEFAULT`.

`PollFeature` treats a still-`REGISTERED` placeholder at poll-budget exhaustion as async-acceptance (`waiting`), so slow asynchronous issuance yields a poll response rather than a hard `systemFailure` timeout.

Accepted trade-off: a forged follow-up counts toward lockout like any wrong MAC, so an observer of the cleartext `senderKID` can lock the authorization (a DoS on the self-service credential, operator reset required) — pre-existing, not widened here.

Known follow-up (pre-existing, tracked separately): revocation does not close the registration authorization, so a challenge holder could `kur`-rekey a revoked certificate.

Testing: gate predicate unit tests, CMP resolver + MAC-computation unit tests, `PollFeatureTest`, and an end-to-end ITest building real PBM-protected `PKIMessage`s (no connector crypto needed) on an existing context signature — `ContextSignatureGuardTest` baseline unchanged. Manually verified with `openssl cmp` against a live authority: `ir → waiting → pollReq → certificate → certConf → pkiConf`, plus `kur` and the negative paths.

The FE selector for the new CMP profile field is tracked separately in `fe-administrator`.


src/main/java/com/otilm/core/service/impl/ScepProfileServiceImpl.java
```diff
@@ -367,6 +367,33 @@ private static boolean hasRsaDecryptionKey(Certificate caCertificate) {
...
+        }
+        if (intuneEnabled) {
+            throw new ValidationException(ValidationError.create(
+                    "Intune requires the profile challenge password as the challenge source"));
+        }
+        if (!hasRsaDecryptionKey(caCertificate)) {
+            throw new ValidationException(ValidationError.create(
+                    "Certificate registration challenge source requires an RSA CA certificate; requests enveloped to a non-RSA CA key need a shared challenge password to decrypt"));
+        }
+    }
```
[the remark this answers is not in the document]
klaraf755: Already removed.

src/main/java/com/otilm/core/service/impl/CmpProfileServiceImpl.java
```diff
@@ -450,6 +482,27 @@ private void validateAndSetProtectionMethods(CmpProfile cmpProfile, CmpProfileRe
      * not prefill the stored secret, so a blank or omitted value keeps the stored one; a non-blank value replaces it.
      * Rejected only when there is nothing to keep — on create, or on edit of a profile without a stored secret.
      */
+    /**
+     * Rules of the certificate-registration challenge source: the per-registration challenge is the MAC key,
+     * so the profile stores no shared secret; MAC/shared-secret request protection exists only on the V2
+     * variant (V2_3GPP forces signature both ways, V3 is unsupported), so registration mode requires V2 and
```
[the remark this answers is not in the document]
klaraf755: Fixed — moved the `applySharedSecret` Javadoc back onto `applySharedSecret` so it documents the right method, leaving the registration-source Javadoc as the only one on `validateRegistrationChallengeSource`. Fixed in 3abc174ac.

src/main/java/com/otilm/core/service/cmp/message/handler/CrmfIrCrMessageHandler.java
```diff
@@ -97,6 +131,77 @@ public ClientCertificateDataResponseDto handle(PKIMessage request, Configuration
...
+                    List.of(new RegistrationIdentityMatcher.Candidate(
+                            matched.getUuid(), matched.getSubjectDn(), matched.getSubjectAlternativeNames())));
+            if (result.outcome() == RegistrationIdentityMatcher.Outcome.SAN_MISMATCH) {
+                certificateEventHistoryService.addEventHistory(matched.getUuid(), CertificateEvent.ISSUE,
+                        CertificateEventStatus.FAILED,
+                        "CMP enrolment subject alternative names do not match the registered ones", "");
+            }
+            if (result.outcome() != RegistrationIdentityMatcher.Outcome.MATCHED) {
+                throw new CmpProcessingException(tid, PKIFailureInfo.badMessageCheck, CmpRegistrationResolver.REGISTRATION_REJECTION);
+            }
```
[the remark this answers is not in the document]
klaraf755: Good catch. `verifyRegistrationIdentity` now records a FAILED event on the resolved registration for a subject mismatch (`NO_MATCH`) too, not just SAN mismatches — the candidate is already resolved from the senderKID, so any identity mismatch is attributable to it; the wire still carries the single generic rejection. Covered by a new `subjectMismatchIsRejectedAndRecordedOnTheRegistration` test. Fixed in 3abc174ac.

src/main/java/com/otilm/core/service/cmp/registration/CmpRegistrationResolver.java
```diff
@@ -0,0 +1,160 @@
...
+     * As {@link #resolveAndVerify} but for a CMP follow-up (pollReq / certConf) of a registration exchange:
+     * by the time the client polls, the placeholder has moved past {@code REGISTERED}, so the certificate
+     * state is not constrained — only that the senderKID references a certificate of {@code raProfile} whose
+     * authorization still verifies the MAC through the gate (a forged follow-up is still counted and locks
+     * out). Issuance leaves the authorization ACTIVE, so the challenge remains available to key the response.
+     */
+    public RegistrationMacResolution resolveAndVerifyFollowup(RaProfile raProfile, ASN1OctetString senderKID,
+                                                              Predicate<byte[]> macMatches,
+                                                              ASN1OctetString tid) throws CmpBaseException {
+        Certificate certificate = resolveEligibleCertificate(raProfile, senderKID, null, tid);
```
lubomirw, line 91: **[High]** | Security

A pollReq or certConf can authenticate with any active registration challenge in the same RA profile because the senderKID certificate is never bound to the transaction selected by transactionID.

_Suggested:_ Persist the authenticated registration UUID on each CMP transaction and require follow-up senderKID resolution to match it before polling or confirming, including KUR successor handling.

<sub>flagged by Copilot CLI, Guided PR review, Superpowers</sub>

src/main/java/com/otilm/core/service/cmp/message/validator/impl/ProtectionMacValidator.java
```diff
@@ -83,4 +81,48 @@ public Void validate(PKIMessage message, ConfigurationContext configuration) thr
...
+     * Bodies whose MAC registration mode authenticates against a pre-registration challenge: the enrolment and
+     * rekey requests (ir/cr/kur) and the async follow-ups of a registration exchange (pollReq/certConf). Every
+     * other MAC body (e.g. a revocation) has no registration challenge to key it and is rejected below.
+     */
+    private static boolean isRegistrationMacBody(PKIMessage message) {
+        int bodyType = message.getBody().getType();
+        return bodyType == PKIBody.TYPE_INIT_REQ
+                || bodyType == PKIBody.TYPE_CERT_REQ
+                || bodyType == PKIBody.TYPE_KEY_UPDATE_REQ
+                || bodyType == PKIBody.TYPE_POLL_REQ
```
lubomirw: **[High]** | Security

ProtectionValidator returns early for an unprotected request, so registration-mode pollReq and certConf messages can reach their handlers without challenge authentication.

_Suggested:_ Reject missing protection in registration mode before the early return and defensively require a matched registration in both follow-up handlers.

src/main/java/com/otilm/core/service/cmp/message/validator/impl/ProtectionMacValidator.java
```diff
@@ -83,4 +81,48 @@ public Void validate(PKIMessage message, ConfigurationContext configuration) thr
...
+        byte[] salt = pbmParameter.getSalt().getOctets();        // --    salt (octetstring)
+        AlgorithmIdentifier owf = pbmParameter.getOwf();         // --    owf  (algIdentifier)
+        // The salted secret is the input to the first iteration; the output of the final iteration
+        // (called "BASEKEY", size "H") forms the symmetric key.
+        byte[] basekey = new byte[passwordAsBytes.length + salt.length];
+        System.arraycopy(passwordAsBytes, 0, basekey, 0, passwordAsBytes.length);
+        System.arraycopy(salt, 0, basekey, passwordAsBytes.length, salt.length);
+        MessageDigest dig = MessageDigest.getInstance(owf.getAlgorithm().getId(),
+                BouncyCastleProvider.PROVIDER_NAME);
+        for (int i = 0; i < pbmParameter.getIterationCount().getValue().intValue(); i++) {
```
lubomirw: **[High]** | Performance

The client-controlled PBM iteration count is unbounded and evaluated while the registration gate holds a pessimistic row lock and database transaction.

_Suggested:_ Validate an algorithm allowlist, salt size, and a positive server-defined maximum iteration count before entering the registration gate.

<sub>flagged by Guided PR review, Superpowers</sub>

src/main/java/com/otilm/core/service/cmp/message/handler/CrmfKurMessageHandler.java
```diff
@@ -99,20 +101,37 @@ public ClientCertificateDataResponseDto handle(PKIMessage request, Configuration
                     "re-key operation failed: both public key are the same; must be different");
         }
 
+        // In registration mode the senderKID (resolved and MAC-verified at the protection layer) must name
+        // the very certificate this kur rekeys: a challenge authorizes rekeying only its own certificate.
+        if (configuration.isRegistrationMode()
```
lubomirw: **[High]** | Security

Registration-mode KUR verifies that senderKID names the target certificate, but the downstream rekey path checks only subject DN and permits the CRMF request to introduce different SAN identities.

_Suggested:_ Apply RegistrationIdentityMatcher to the KUR CRMF subject and SAN set before rekeying, and add negative SAN-mismatch coverage.

src/main/java/com/otilm/core/service/cmp/message/handler/CrmfIrCrMessageHandler.java
```diff
@@ -49,13 +60,33 @@ public class CrmfIrCrMessageHandler implements MessageHandler<ClientCertificateD
             PKIBody.TYPE_INIT_REQ,          // ir       [0]  CertReqMessages,       --Initialization Req
             PKIBody.TYPE_CERT_REQ);         // cr       [2]  CertReqMessages,       --Certification Req
 
+    private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(CrmfIrCrMessageHandler.class);
```
lubomirw: **[Nit]** | Maintainability

The logger declaration uses fully qualified SLF4J type names inline even though the simple names are collision-free.

_Suggested:_ Import org.slf4j.Logger and org.slf4j.LoggerFactory, then use Logger and LoggerFactory in the declaration.

<sub>flagged by Hygiene</sub>

src/main/java/com/otilm/core/service/cmp/message/validator/impl/ProtectionMacValidator.java
```diff
@@ -83,4 +81,48 @@ public Void validate(PKIMessage message, ConfigurationContext configuration) thr
...
+
+    /**
+     * Whether the message's Password-Based MAC verifies under {@code passwordAsBytes}, computed per
+     * RFC 4210 §5.1.3.1 from the message's own PBMParameter (salt, owf, iteration count, mac algorithm).
+     * The pure computation, reusable both with the profile shared secret and with a registration challenge.
+     */
+    static boolean matchesMac(PKIMessage message, byte[] passwordAsBytes) throws NoSuchAlgorithmException, NoSuchProviderException, IOException, InvalidKeyException {
+        PKIHeader header = message.getHeader();
+        PBMParameter pbmParameter = PBMParameter.getInstance(
+                header.getProtectionAlg().getParameters());      // -- PBMParameter
```
lubomirw: **[Nit]** | Maintainability

The PBMParameter, salt, OWF, and MAC-comparison comments restate the adjacent expressions.

_Suggested:_ Remove those restatement comments while retaining the useful BASEKEY rationale.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/integration/service/cmp/message/handler/CmpRegistrationEnrolmentITest.java
```diff
@@ -0,0 +1,326 @@
...
+        KeyPair keyPair = CmpTestUtil.generateKeyPairEC();
+        PKIBody body = CmpTestUtil.createRegistrationCrmfBody(keyPair, 0L, PKIBody.TYPE_INIT_REQ, subjectDn, dnsSans, null);
+        return CmpTestUtil.createMacBasedMessageWithSenderKid(
+                "0102030405060708", challenge, body,
+                senderKid.toString().getBytes(StandardCharsets.UTF_8)).toASN1Structure();
+    }
+
+    private String failText(byte[] responseBytes) {
+        PKIMessage response = PKIMessage.getInstance(responseBytes);
+        return ((org.bouncycastle.asn1.cmp.ErrorMsgContent) response.getBody().getContent())
```
lubomirw, line 325: **[Nit]** | Maintainability

Two casts use the fully qualified ErrorMsgContent type inline.

_Suggested:_ Import org.bouncycastle.asn1.cmp.ErrorMsgContent and use ErrorMsgContent in both casts.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/integration/service/cmp/message/handler/CmpRegistrationEnrolmentITest.java
```diff
@@ -0,0 +1,326 @@
...
+    /**
+     * The issued certificate the poll returns once issuance finishes. Persists a backing row so the
+     * dispatcher's by-UUID chain lookup resolves, but returns a detached twin carrying a plain (non-proxy)
+     * CertificateContent, so response-building can read the content without the seeding session (the test is
+     * not transactional).
+     */
+    private Certificate seedIssuedCertificate() throws Exception {
+        KeyPair keyPair = CmpTestUtil.generateKeyPairEC();
+        var holder = CmpTestUtil.makeV3Certificate(BigInteger.valueOf(System.nanoTime()), keyPair, "CN=issued", keyPair, "CN=issued");
+        String content = java.util.Base64.getEncoder().encodeToString(holder.getEncoded());
```
lubomirw: **[Nit]** | Maintainability

The test uses java.util.Base64 inline instead of importing the collision-free type.

_Suggested:_ Import java.util.Base64 and use Base64.getEncoder().

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/integration/service/CmpProfileServiceITest.java
```diff
@@ -149,6 +150,112 @@ void testGetCmpProfileByUuid() throws NotFoundException {
...
+
+    @Test
+    void omittedChallengeSourceKeepsStoredSource() throws ConnectorException, AlreadyExistException, AttributeException, NotFoundException {
+        CmpProfileDto created = cmpProfileService.createCmpProfile(registrationCreateRequest("KeepReg"));
+
+        CmpProfileEditRequestDto edit = new CmpProfileEditRequestDto();
+        edit.setVariant(CmpProfileVariant.V2);
+        edit.setRequestProtectionMethod(ProtectionMethod.SHARED_SECRET);
+        edit.setResponseProtectionMethod(ProtectionMethod.SHARED_SECRET);
+        // challengeSource omitted -> keep the stored CERTIFICATE_REGISTRATION
```
lubomirw: **[Nit]** | Maintainability

The comment restates the test name and the visibly omitted challengeSource setter.

_Suggested:_ Remove the comment line.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/service/cmp/registration/CmpRegistrationResolverTest.java
```diff
@@ -0,0 +1,228 @@
...
+
+    private Certificate registeredCertificate() {
+        Certificate certificate = new Certificate();
+        certificate.setUuid(CERT_UUID);
+        certificate.setState(CertificateState.REGISTERED);
+        certificate.setRaProfileUuid(RA_PROFILE_UUID);
+        return certificate;
+    }
+
+    /** Makes the gate invoke the predicate with the stored plaintext and return its result. */
```
lubomirw: **[Nit]** | Maintainability

The private helper Javadoc restates its name and implementation.

_Suggested:_ Remove the Javadoc line.

<sub>flagged by Hygiene</sub>

lubomirw (changes requested):
Automated multi-reviewer triage - 12 findings (REQUEST_CHANGES).

**Findings that could not be anchored to an added line:**
- **[Medium] Security** `src/main/java/com/otilm/core/service/cmp/impl/CmpServiceImpl.java:352` - A registration rejection raised before challenge matching still builds shared-secret response protection with an empty key, making the error MAC publicly reproducible. _Suggested:_ Return an unprotected CMP error when registration mode has no matched challenge, or reject an empty response key so the existing unprotected fallback is used. _(flagged by Copilot CLI)_
- **[Low] Documentation** `documentation/docs/certificate-key/protocols/cmp/cmp-profile.md:9` - The CMP profile documentation omits challengeSource, its default and restrictions, UUID senderKID semantics, and challenge-as-PBM-key behavior. _Suggested:_ Document both challenge-source values, the certificate-registration workflow, supported operations, configuration constraints, and follow-up authentication. _(flagged by Docs check (Copilot CLI, Guided PR review, Superpowers))_

src/main/java/com/otilm/core/service/cmp/message/handler/CrmfKurMessageHandler.java
klaraf755: Fixed.

src/main/java/com/otilm/core/service/cmp/message/handler/CrmfIrCrMessageHandler.java
klaraf755: Fixed.

src/main/java/com/otilm/core/service/cmp/message/validator/impl/ProtectionMacValidator.java
klaraf755: Fixed.

src/test/java/com/otilm/core/integration/service/cmp/message/handler/CmpRegistrationEnrolmentITest.java
klaraf755: Fixed.

src/test/java/com/otilm/core/integration/service/cmp/message/handler/CmpRegistrationEnrolmentITest.java
klaraf755: Fixed.

src/test/java/com/otilm/core/integration/service/CmpProfileServiceITest.java
klaraf755: Fixed.

src/test/java/com/otilm/core/service/cmp/registration/CmpRegistrationResolverTest.java
klaraf755: Fixed.

src/main/java/com/otilm/core/service/cmp/registration/CmpRegistrationResolver.java
klaraf755: Fixed in e620c58aa. The transaction already persisted the certificate UUID ("persist the authenticated registration UUID on each CMP transaction" was in place for ir/cr — a kur transaction stores the rekeyed successor instead), so the fix adds the missing check rather than new persistence: in registration mode, pollReq and certConf now require the senderKID-matched registration to equal the transaction's certificate, or — for kur follow-ups — to be its recorded predecessor via the certificate relation ("including KUR successor handling"). Mismatches surface as the generic registration rejection, and a rejected unbound follow-up no longer marks the transaction FAILED, so a foreign registration can neither read nor poison another registration's in-flight transaction. Covered by new resolver unit tests and end-to-end ITests (cross-registration pollReq/certConf rejected with state preserved, own-transaction poll, kur successor poll).

src/main/java/com/otilm/core/service/cmp/message/validator/impl/ProtectionMacValidator.java
klaraf755: Fixed in abf73bebe. ProtectionValidator.validateIn now rejects a message with neither protection bits nor a protectionAlg when the profile is in registration mode (notAuthorized / CMPVALPRO530), instead of returning success — so an unprotected pollReq or certConf no longer reaches a handler unauthenticated. Non-registration profiles keep the RFC 4210-permitted unprotected path unchanged. The defensive matched-registration requirement is already in place in every follow-up handler: ir/cr and kur reject a null/mismatched matched registration, and pollReq/certConf reject through requireTransactionBinding (e620c58aa). Covered by new ProtectionValidatorTest cases (unprotected rejected in registration mode, allowed otherwise).

src/main/java/com/otilm/core/service/cmp/message/validator/impl/ProtectionMacValidator.java
klaraf755: Fixed in 91c97226f. ProtectionMacValidator now validates the client-supplied PBMParameter up front, before the MAC computation and — in registration mode — before the challenge gate takes its row lock: iteration count must be positive and within a server-defined maximum (100000), the salt within a maximum size (1024 bytes), and the one-way-function and MAC algorithms within an allowlist (SHA-1/256/384/512 and HMAC-SHA1/256/384/512). Rejections are badMessageCheck (bounds) / badAlg (algorithm). This closes the unbounded digest loop that would otherwise run while holding the pessimistic authorization lock.

Note on scope: the unbounded loop itself predates this PR (the shared-secret path always ran it); this PR newly runs it under the registration row lock, which is what escalates the impact. The bound sits at the top of validate(), so it hardens both paths. The limits are protocol-hardening constants rather than Settings entries for now — say the word if you would prefer them operator-configurable. Covered by new ProtectionMacParameterValidationTest cases.

src/main/java/com/otilm/core/service/cmp/message/handler/CrmfKurMessageHandler.java
klaraf755: Fixed in e0b54884c. Extracted the ir/cr subject-and-SAN identity check into a shared CmpRegistrationIdentityVerifier, and the kur rekey path now applies it: in registration mode, after confirming senderKID names the target certificate, the kur CRMF's subject and SAN set must equal the matched registration's via RegistrationIdentityMatcher before rekeyCertificate is called. A SAN (or subject) that differs from the registered identity is rejected with the generic registration rejection and recorded against the certificate. ir/cr and kur now share one implementation, so they cannot diverge again.

Coverage: new CmpRegistrationIdentityVerifierTest asserts the negative SAN-mismatch and subject-mismatch cases (rejection + event history) plus the matching case; the existing ir/cr end-to-end SAN/subject-mismatch ITests exercise the same shared verifier. A full end-to-end kur ITest needs the rekey harness (an issued certificate with a surviving ACTIVE authorization plus the rekey connector wiring) — happy to add it as a follow-up if you want the integration-level kur coverage too.

klaraf755:
**[Medium] Security — empty-key error MAC** (CmpServiceImpl, unanchored review finding): fixed in eff254ee4.

buildProcessingErrorResponse now returns an unprotected CMP error when the profile is in registration mode and no challenge has matched (getMatchedChallenge() == null) — the pre-match rejections (unresolved senderKID, wrong state, wrong MAC). Previously these were MAC-protected via the shared-secret response strategy, which in registration mode keys on the matched challenge and falls back to an empty key when nothing matched, yielding a MAC anyone could reproduce. A post-match rejection (e.g. a SAN mismatch, where the challenge did match) still carries a challenge-keyed MAC the legitimate client can verify, so that path is unchanged. RFC 4210 permits unprotected error messages. Covered by a new ITest asserting a pre-match rejection returns no protection bits.

lubomirw (approved):

Issue #1941, opened 2026-08-03
[core] Spring Boot 4.1 migration


lubomirw:
### Description

Umbrella for the `core` half of the Spring Boot 4.1 migration (OmniTrustILM/ilm#322).

Re-planned 2026-08-19 on the pilot-spike findings (posted as a comment on the epic). The
organising fact: `core` compiles on 4.1 after 22 mechanical file edits and the suite then
runs 4924 tests at 97.5% green. The work is therefore not "get it to compile" but "find
and fix what compiles clean and behaves differently" - two Hibernate 7 production breaks
(#2017) and the silently substituted Jackson wire (#2111) own nearly all of it. The spike
branch `core@spike/sb41-poc` is the reference implementation for the baseline.

**Preparation is complete as of 2026-08-24**, all of it delivered on Spring Boot 3.5.
(#2015 landed 2026-08-21 in PR #2119; #2088, the last item, landed 2026-08-24 in PR #2121.)
The goldens (#1992) and the ObjectMapper consolidation (#2014) are load-bearing for
#2104's staged Jackson 3 route. #2013 closed complete: the classification found nothing
to narrow (47 genuine IO sites; the one Jackson-motivated site legitimately stays
broad). #2015 landed in PR #2119: its regression suite is what makes #2018 verifiable,
including the new RC#6 work, and its deprecation inventory is recorded in that PR's
description - including the one deprecation that must stay
(`AuthorizationManager.check` is abstract AND deprecated on 6.5.11, so it is kept and
delegates to `authorize`) and the `Supplier<? extends Authentication>` widening that
will still break at 7.1.0. Note for #2018: fifteen test call sites moved from `check`
to `authorize` by hand, none of which produced a deprecation warning - warning counts
are not a complete inventory.

**Migration runs after OmniTrustILM/interfaces#806 publishes**, as sub-PRs on the shared
`feature/spring-boot-4.1` integration branch. The branch exists because the suite is red
between children and `main` must stay green - NOT because the code does not compile (the
original premise, falsified by the spike). Landing order is load-bearing; it encodes
"fix masking failures first, then measure":

1. #2016 - baseline: compiles, boots, suite measurable (RC#2 + RC#5 inside)
2. #2111 - pin the wire, fix HttpStatus, discovery sweep, re-measure
3. #2017 and #2018 in parallel - essentially all remaining red
4. #2020 - any time after the baseline
[...]

PR #1994, opened 2026-08-09
Baseline Jackson 2 JSON output with golden-file tests

### LINKED ISSUE #1992: Golden-file JSON serialization tests to baseline Jackson 2 behaviour before the Jackson 3 migration
### Description

Add golden-file serialization tests that pin **current Jackson 2 JSON output** before the Spring Boot 4.1 / Jackson 3 migration (#1941). These are the highest-value insurance in the prep list: Jackson 3 changes defaults, and without a byte-level baseline a shape regression is invisible until it reaches a connector, a client, or — worst case — a persisted JSONB column.

Golden files must be captured **now, on 3.5**, and **before** the ObjectMapper-centralization prep item. If centralization lands first, the goldens bake in whatever drift consolidation introduced, and the baseline is worthless.

#### Mapper parity

Goldens for API/DTO surfaces must be produced with the **production** mapper, not a bare `new ObjectMapper()`. `WebAppConfig#jsonObjectMapper` is a pure factory method, so `new WebAppConfig().jsonObjectMapper()` yields exact parity with **no Spring context**. Its customizations all affect on-wire bytes:

- `WRITE_DATES_AS_TIMESTAMPS` disabled → ISO-8601 date strings
- `JavaTimeModule` registered
- `JsonInclude.Include.NON_NULL` → nulls omitted
- `FAIL_ON_MISSING_EXTERNAL_TYPE_ID_PROPERTY` disabled → relevant to `KeyData`'s `EXTERNAL_PROPERTY` typing

`AcmeJsonProcessor` uses its own bespoke `new ObjectMapper()`, so the ACME protocol surface must be pinned against *that* mapper — the divergence is itself worth recording.

#### Surfaces to cover

**1. Polymorphic `@JsonTypeInfo` hierarchies** (highest risk — discriminator name/value/placement all change shape). Annotated base types live in `interfaces`, not `core`:

| Base type | Discriminator | Subtypes |
|---|---|---|
| `BaseAttributeV2` / `BaseAttributeV3` | `type` (defaultImpl `Data*`) | Data / Group / Info / Metadata / Custom |
| `BaseAttributeContentV3` | `contentType` | Boolean, CodeBlock, Date, DateTime, File, Float, Integer, Object, String, Text, Time, `ResourceObjectContent` |
| `ResourceObjectContentData` | `resource` | `ResourceSimpleContentData`, `ResourceCertificateContentData`, **`ResourceSecretContentData`** |
| `BaseAttributeConstraint` | `type` | Regexp / Range / DateTime |
| `MappedField` | `fieldType` | Rdn / San / Extension |
| `ConnectInfo` | `version` (EXISTING_PROPERTY) | V1, V2 |
| `KeyData` | `format` (EXTERNAL_PROPERTY) | Raw / Spki / Prki |

**2. `@JsonTypeInfo` secret containment.** `OutboundSecretContainment` is fail-closed and depends on the serialized graph: it scans string leaves via the Spring mapper and structurally rejects populated `ResourceSecretContentData.content`, `CredentialAttributeContentV2`, `SecretAttributeContentV2`, `SecretAttributeContentData.secret`. A Jackson 3 shape change that renames, relocates, or suppresses one of these fields silently converts a leak-blocking check into a no-op. Pin both the positive shape and the containment behaviour.

**3. ACME protocol payloads** (RFC 8555 wire contract, ~7 classes): `Directory`, `DirectoryMeta`, `Account`, `Order`, `Authorization`, `Challenge`, `ProblemDocument`. SCEP / CMP / TSP are ASN.1 / PKCS#7 binary and are **out of scope** — no Jackson surface.

**4. JSON/JSONB DB columns** (a shape change here silently corrupts persisted data): `AttributeContentItem.json`, `Certificate.complianceResult`, `Certificate.pendingRevokeAttributes`, `AuditLog.logRecord`, `Secret.complianceResult`, `Secret2SyncVaultProfile.secretAttributes`, `CryptographicKeyItem.complianceResult`, `CertificateRequestEntity.complianceResult`, `DiscoveryCertificate.meta`, `ComplianceProfileRule.attributes`, `ProtocolCertificateAssociations.customAttributes`, `AttributeDefinition.definition`, plus `ObjectToJsonConverter`-backed `Approval.objectData`, `ScheduledJob.objectData`, `ConditionItem.value`, `ExecutionItem.data`. `ObjectToJsonConverter` is the sharpest edge — a generic `AttributeConverter<Object, String>` that reads back through `Serializable.class`.

**5. Representative API DTOs.** ~67 REST controllers, but core defines essentially no DTOs of its own — the contract layer is entirely `com.otilm.api.model.*` from `interfaces`. Cover a representative set (certificate detail/list, connector, key, auth/user) rather than exhaustively; the polymorphic attribute types in (1) are embedded in most of them and carry the real risk.

#### Test-context budget

The suite must be **plain JUnit 5 with no Spring annotations at all** — no `@SpringBootTest`, no `CONTEXT_ROOTS` base class, no new `@Import`/`@MockitoBean`/`@ActiveProfiles` combination. `ContextSignatureGuardTest.BASELINE` (currently 57) **must not change**. Building the mapper via `new WebAppConfig().jsonObjectMapper()` keeps the suite outside the taxonomy entirely.

There is no existing golden-file convention — `src/test/resources` currently holds exactly one `.json` fixture (`rabbitMQ-definitions.json`). This issue establishes the pattern.

**Estimate (agent execution - review/verify/supervise):** 1.75 md
**Human baseline:** 5 md

### Acceptance Criteria

- [ ] Golden `.json` resources are committed under a dedicated test-resources directory and are byte-compared by the tests.
- [ ] Fixtures are deterministic: fixed timestamps, fixed UUIDs, sorted map keys — no clock, no randomness, no environment dependence.
- [ ] API/DTO and DB-column goldens are produced through `new WebAppConfig().jsonObjectMapper()`; ACME goldens through the `AcmeJsonProcessor` mapper. The two mappers' differing configuration is recorded.
- [ ] Every `@JsonTypeInfo` hierarchy in the table above has at least one golden per subtype, asserting discriminator property **name, value and placement**.
- [ ] `OutboundSecretContainment` has regression coverage proving it still rejects each of the four containment-critical shapes, and that the check is driven by serialized output rather than passing vacuously.
- [ ] All ACME wire types have goldens; the deliberate exclusion of SCEP / CMP / TSP is noted.
- [ ] Every JSON/JSONB column payload type has a round-trip golden (serialize → compare → deserialize → re-serialize → compare), including the `ObjectToJsonConverter` `Serializable` read path.
- [ ] Tests are plain JUnit 5; `ContextSignatureGuardTest.BASELINE` is unchanged at 57.
- [ ] `mvn -B verify` passes; Spotless / Checkstyle gates pass.
- [ ] A short note in the repo explains how to regenerate goldens intentionally and that an unexplained diff during #1941 is a migration finding, not a test to update.

ivosh:
Closes #1992. Prep for #1941.

Adds golden-file JSON serialization tests that pin the platform's **current Jackson 2 output**, so the Spring Boot 4.1 / Jackson 3 migration can be diffed against a byte-level baseline instead of discovered from the field.

Jackson 3 changes defaults in ways that have no compiler signature — a renamed key, a dropped null, a date rendered as a number, a discriminator that stops being emitted. Tests that assert on objects keep passing through all of it. These goldens do not.

Captured **now, before** the ObjectMapper-centralization prep item: if consolidation lands first, the goldens bake in whatever drift it introduced and the baseline is worthless.

### Coverage — 55 tests, 49 golden files

| Surface | Why it's ordered here |
|---|---|
| `jsonb` columns | Worst blast radius. The JSON mapping *is* the schema — no migration guards it, so drift splits a table into old-shape and new-shape rows. Every case round-trips, since the read-back leg is what a stored row experiences on each load-and-save. |
| `ResponseAttributeSerializer` redaction | The platform's **primary** secret redaction. Had zero coverage. |
| `OutboundSecretContainment` shapes | Both its checks are shape-coupled; drift degrades the guard into a no-op that still passes. |
| `@JsonTypeInfo` hierarchies | Discriminator name, value and placement — shared with ~40 connector repos that won't upgrade in lockstep. |
| REST + ACME wire contracts | Third-party ACME clients parse strictly against the RFC. |

### Mapper parity

Goldens use the mappers production actually uses. The web mapper comes from `WebAppConfig#jsonObjectMapper` called directly — it takes no collaborators, so the suite gets exact parity with **no Spring context**, leaving `ContextSignatureGuardTest.BASELINE` untouched at 57. ACME keeps its own bare mapper, and that divergence is asserted as a test rather than hidden, since consolidating the 27 bespoke mapper sites is the next prep item.

### Things this turned up

Writing the read-side tests surfaced facts that weren't documented anywhere:

- **`ResponseAttributeSerializer` is live and untested.** It nulls secret content before it reaches an API response. Because it's declared on the `ResponseAttributeV2.content` *field*, it isn't cancelled the way the class-level serializer on `BaseAttribute` is. It's hand-written against Jackson 2 internals the upgrade must rewrite — had it silently stopped firing, secrets would reach the wire with every test green.
- **`BaseAttributeV2`'s `@JsonSubTypes` registrations are vestigial.** None of the classes it names actually extend it (`DataAttributeV2 extends DataAttribute extends BaseAttribute`), so deserializing through that type fails with an unresolvable type id. The live path is `BaseAttribute`, resolved by the hand-written `BaseAttributeDeserializer` switching on `version` and `type`.
- **A secret arriving as JSON is invisible to the containment guard's structural check.** It reads back as a plain v2 content with an untyped map, because it carries no `contentType` for `AttributeContentDeserializer` to branch on — so every `instanceof` misses and only the value-echo scan catches it. Existing behaviour, not a regression, but it means the value check is the *sole* net for anything arriving over the wire. Worth knowing before anyone changes serialization.
- **`BaseAttributeSerializer` is dormant** (every concrete subclass cancels it), and **`MetadataAttributeDeserializer` is dead code** — never referenced. The real custom-ser/deser risk surface is 3 live classes, not 5. Which are live is recorded in the package docs, since it isn't apparent from the annotations.

### Validation

- 55 tests green; mutation-checked (corrupting a golden fails the suite).
- Deterministic under `Asia/Tokyo` + Turkish locale. Jackson renders `Date` in UTC via its own default, independent of the JVM timezone.
- Runs in the `test-non-integration` CI group; `ContextSignatureGuardTest` still passes at 57.
- `-Dgolden.regenerate=true` is **refused when `CI` is set** — a regenerating CI run would rewrite the baseline while reporting success, the one failure mode this suite can't tolerate because it looks like a pass.

### For reviewers

During the migration a golden diff is a **finding to explain, not a test to update**. Trace it to a documented Jackson 3 behaviour change, record it on #1941, then regenerate. `package-info.java` covers the workflow.

🤖 Generated with [Copilot CLI](https://github.com/features/copilot/cli)


src/test/java/com/otilm/core/serialization/golden/SecretContainmentGoldenTest.java
```diff
@@ -0,0 +1,270 @@
...
+ * A test that only asserted "the guard throws" would keep passing through all of that. These tests therefore pin the
+ * serialized shape <i>and</i> assert the guard's behaviour against it.
+ */
+class SecretContainmentGoldenTest {
+
+    private static final String SECRET_VALUE = "s3cr3t-material-expanded-server-side";
+
+    private final ObjectMapper mapper = GoldenMappers.web();
+
+    private final OutboundSecretContainment containment = new OutboundSecretContainment(GoldenMappers.web());
```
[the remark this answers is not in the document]
ivosh: Fixed in 8ca0b744.

src/test/java/com/otilm/core/serialization/golden/AttributeTypeInfoGoldenTest.java
```diff
@@ -0,0 +1,297 @@
...
+                Arguments.of("certificate", "certificates", certificateResource()),
+                // Deliberately left with a null content: a populated one is a leak shape that
+                // OutboundSecretContainment refuses outright, and is covered by SecretContainmentGoldenTest.
+                Arguments.of("secret", "secrets", new ResourceSecretContentData(
+                        "6f1a4d3c-0000-4000-8000-000000000006", "Vault Secret", null)));
+    }
+
+    private static ResourceSimpleContentData simpleResource(AttributeResource resource, String name) {
+        return new ResourceSimpleContentData(resource, "6f1a4d3c-0000-4000-8000-00000000000" + resource.ordinal(), name, null);
+    }
```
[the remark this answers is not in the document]
ivosh: Fixed in 8ca0b744.

src/test/java/com/otilm/core/serialization/golden/AcmeProtocolGoldenTest.java
```diff
@@ -0,0 +1,166 @@
...
+        JsonNode webTree = GoldenMappers.web().valueToTree(new DirectoryMeta());
+
+        assertThat(acmeTree.has("type"))
+                .describedAs("the ACME mapper has no NON_NULL inclusion, so populated and absent fields alike are keys")
+                .isTrue();
+        assertThat(webTree.isEmpty())
+                .describedAs("the wire mapper omits null fields entirely — the two mappers genuinely differ, and the "
+                        + "ACME goldens are baselined against the ACME mapper for that reason")
+                .isTrue();
+    }
```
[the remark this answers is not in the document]
ivosh: Already addressed: this test was rewritten in 59fcac65 (after the review ran) as `inboundJwsEnvelopeMapperKeepsItsDefaultsAndRejectsUnknownMembers`. It now serializes an all-null `DirectoryMeta` and asserts the envelope mapper still renders its keys, while the wire mapper omits them — i.e. exactly the absent-field-as-explicit-null behaviour you flagged as unasserted.

src/test/java/com/otilm/core/serialization/golden/AttributeTypeInfoGoldenTest.java
```diff
@@ -0,0 +1,592 @@
...
+        assertThat(tree.path("format").asText())
+                .describedAs("the format discriminator must stay a sibling of 'value'")
+                .isEqualTo(formatCode);
+        assertThat(tree.path("value").has("format"))
+                .describedAs("an EXTERNAL_PROPERTY discriminator must not appear inside the typed value; moving it "
+                        + "there changes the wire contract for every connector exchanging keys")
+                .isFalse();
+    }
+
+    private static Stream<Arguments> keyDataFormats() {
```
lubomirw, line 519: **[Low]** | Testing

`KeyData` registers EPRKI and CUSTOM subtypes, but the golden matrix covers only RAW/SPKI/PRKI. Those formats have active org consumers, so their Jackson output can regress without detection.

_Suggested:_ Add deterministic `EprkiKeyValue` and `CustomKeyValue` cases with corresponding golden fixtures and round-trip assertions.

src/test/java/com/otilm/core/serialization/golden/JsonColumnGoldenTest.java
```diff
@@ -0,0 +1,278 @@
...
+import java.util.Map;
+import java.util.UUID;
+import org.hibernate.type.format.jackson.JacksonJsonFormatMapper;
+import org.junit.jupiter.api.Test;
+import org.springframework.test.util.ReflectionTestUtils;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Pins the JSON written into the platform's {@code jsonb} columns. No schema migration guards them, so the JSON mapping
```
lubomirw, line 53: **[Low]** | Testing

The JSONB suite omits the String payload used by `SigningRecord.requestMetadataJson` and `SigningRecordOutbox.requestMetadataJson`, despite Hibernate normalizing that JSON during persistence.

_Suggested:_ Add a representative raw-JSON String golden that exercises Hibernate serialization and round-tripping for both signing-record columns.

src/test/java/com/otilm/core/serialization/golden/SecretContainmentGoldenTest.java
```diff
@@ -0,0 +1,257 @@
...
+                .describedAs("a response with no secret must pass, otherwise the echo test proves nothing")
+                .doesNotThrowAnyException();
+    }
+
+    /**
+     * The recorder strips {@code type}, {@code username} and {@code keyStoreType} by exact name. A rename would record
+     * the discriminator value as a secret and refuse every later response mentioning it.
+     */
+    @Test
+    void recordingCapturesSecretLeavesButNotTheLowEntropyDiscriminatorAndUsername() {
```
lubomirw, line 141: **[Low]** | Testing

This test says it pins exact-name removal of `keyStoreType`, but it only constructs `BasicAuthSecretContent`, so the `KeyStoreSecretContent` branch remains untested.

_Suggested:_ Add a `KeyStoreSecretContent` case asserting that secret content and password are recorded while `type` and `keyStoreType` are excluded.

lubomirw (commented):
Automated multi-reviewer triage — 3 findings (COMMENT).

src/test/java/com/otilm/core/serialization/golden/AttributeTypeInfoGoldenTest.java
ivosh: Fixed in b3939289. Added `eprki` and `custom` rows to the `keyDataFormats()` matrix, so all five registered `KeyData` subtypes are now pinned, with goldens `key-data-eprki.json` and `key-data-custom.json`. `CustomKeyValue` is the only subtype with a non-scalar payload, so it also pins how an `EXTERNAL_PROPERTY` discriminator behaves when the typed member is not a single-property object.

src/test/java/com/otilm/core/serialization/golden/SecretContainmentGoldenTest.java
ivosh: Fixed in b3939289. Added `recordingCapturesBothKeyStoreSecretLeavesButNotItsTypeTag`, which records a `KeyStoreSecretContent` and asserts the key-store bytes and password are captured while `keyStore`, `KEY_STORE` and `PKCS12` stay out. The class javadoc's `keyStoreType` claim is now actually backed by a test.

src/test/java/com/otilm/core/serialization/golden/JsonColumnGoldenTest.java
ivosh: Fixed in b3939289, though the stated mechanism turned out to be the opposite of what happens.

Hibernate does **not** normalize this column. `AbstractJsonFormatMapper` (6.6.53) short-circuits the `String` java type in both directions — `toString` and `fromString` return the value verbatim, with no Jackson step at all. So the persistence layer is a pass-through.

The gap is real for a different reason: the payload is built by `TspSigningRecordFactory.buildRequestMetadataJson` through the **wire** mapper, so the column inherits its `NON_NULL` inclusion and an absent policy or nonce leaves no key at all. That shape was unbaselined.

Two tests added to `JsonColumnGoldenTest`:
- `signingRecordMetadataColumnKeepsTheShapeTheWireMapperGivesItAndIsStoredVerbatim` — golden `column-signing-record-metadata.json` produced via the production assembly path, plus assertions that `NON_NULL` drops absent policy/nonce and that `hashAlgorithm` is written as `SHA_256` (`Enum.name()`), not the `SHA-256` code every wire payload carries.
- `stringJsonColumnsAreBoundAndReadBackWithoutAnyJacksonStep` — pins the pass-through itself, so reintroducing a Jackson round trip here (which would double-encode the payload into a JSON string literal) fails loudly.

lubomirw (approved):

PR #2001, opened 2026-08-10
Stop nested integration tests re-running in the non-integration worker

ivosh:
## Summary

- stop nested integration tests from running again in the non-integration worker
- preserve the integration shard assignment derived from the 20-run experiment in #1997
- guard all profile boundaries and explicit transfers against gaps or duplicate execution

## Established effects

- removes 125 duplicate nested integration-test executions from the non-integration worker
- post-change 20-run means were 9:42, 9:19, and 9:14 for the three integration workers, a 28-second (5.1%) spread; the non-integration worker is a separate ~3-minute leg (mean 2:56) and is not part of that spread
- leaves the existing four-worker workflow and coverage merge unchanged

Advances OmniTrustILM/ilm#308.

## Validation

- `mvn -B -ntp test -Dmaven.compiler.proc=full -Dtest=CiTestSplitTest,TestClassTaxonomyTest,ContextSignatureTest`
- `mvn -B -ntp test -Ptest-non-integration -Dmaven.compiler.skip=true` (2,111 tests; zero integration reports)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -451,8 +451,7 @@
     </repositories>
 
     <profiles>
-        <!-- Worker 1: everything outside the integration package.
-             The regex also excludes nested classes declared inside ITests. -->
+        <!-- Worker 1: every runnable test except the integration *ITests (and their nested classes). -->
         <profile>
             <id>test-non-integration</id>
             <build>
--- a/pom.xml
+++ b/pom.xml
@@ -451,7 +451,9 @@
     </repositories>
 
     <profiles>
-        <!-- Worker 1: everything except integration ITests (service unit tests + all the rest). -->
+        <!-- Worker 1: everything outside the integration package.
+             The regex also excludes nested classes declared inside ITests; an Ant *ITest.java pattern
+             excludes only the outer class and causes its @Nested tests to run again on this worker. -->
         <profile>
             <id>test-non-integration</id>
             <build>
@@ -461,7 +463,7 @@
                         <artifactId>maven-surefire-plugin</artifactId>
                         <configuration>
                             <excludes>
-                                <exclude>com/otilm/core/integration/**/*ITest.java</exclude>
+                                <exclude>%regex[.*/integration/.*ITest.*]</exclude>
                             </excludes>
                         </configuration>
                     </plugin>
@@ -469,7 +471,7 @@
             </build>
         </profile>
 
-        <!-- Worker 2: integration ITests outside the service package. -->
+        <!-- Worker 2: integration ITests outside the service package, except transfers to worker 4. -->
         <profile>
             <id>test-integration-core</id>
             <build>
@@ -483,6 +485,8 @@
                             </includes>
                             <excludes>
                                 <exclude>com/otilm/core/integration/service/**/*ITest.java</exclude>
+                                <exclude>com/otilm/core/integration/cryptography/PQCITest.java</exclude>
+                                <exclude>com/otilm/core/integration/search/TimeQualityConfigurationSearchITest.java</exclude>
                             </excludes>
                         </configuration>
                     </plugin>
@@ -491,8 +495,9 @@
         </profile>
 
         <!-- Workers 3 & 4 split integration.service by the shared %regex boundary below.
-             The split is currently unbalanced, and rebalancing is manual as the suite grows. -->
-        <!-- Worker 3: integration.service nested subpackages + flat classes A-C. -->
+             Four heavy A-C classes are transferred to worker 4 to balance wall-clock cost.
+             CertificateServiceITest and CryptographicKeyServiceITest stay together because they share a context. -->
+        <!-- Worker 3: integration.service nested subpackages + flat classes A-C, except transfers. -->
         <profile>
             <id>test-integration-service-1</id>
             <build>
@@ -506,6 +511,10 @@
                             </includes>
                             <excludes>
                                 <exclude>%regex[.*/integration/service/[D-Z][^/]*ITest.*]</exclude>
+                                <exclude>com/otilm/core/integration/service/AcmeProfileServiceITest.java</exclude>
+                                <exclude>com/otilm/core/integration/service/AcmeServiceITest.java</exclude>
+                                <exclude>com/otilm/core/integration/service/CertificateServiceITest.java</exclude>
+                                <exclude>com/otilm/core/integration/service/CryptographicKeyServiceITest.java</exclude>
                             </excludes>
                         </configuration>
                     </plugin>
@@ -513,7 +522,7 @@
             </build>
         </profile>
 
-        <!-- Worker 4: integration.service flat classes D-Z. -->
+        <!-- Worker 4: integration.service flat classes D-Z plus transfers from workers 2 and 3. -->
         <profile>
             <id>test-integration-service-2</id>
             <build>
@@ -524,6 +533,12 @@
                         <configuration>
                             <includes>
                                 <include>%regex[.*/integration/service/[D-Z][^/]*ITest.*]</include>
+                                <include>com/otilm/core/integration/service/AcmeProfileServiceITest.java</include>
+                                <include>com/otilm/core/integration/service/AcmeServiceITest.java</include>
+                                <include>com/otilm/core/integration/service/CertificateServiceITest.java</include>
+                                <include>com/otilm/core/integration/service/CryptographicKeyServiceITest.java</include>
+                                <include>com/otilm/core/integration/cryptography/PQCITest.java</include>
+                                <include>com/otilm/core/integration/search/TimeQualityConfigurationSearchITest.java</include>
                             </includes>
                         </configuration>
                     </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -461,7 +461,7 @@
                         <artifactId>maven-surefire-plugin</artifactId>
                         <configuration>
                             <excludes>
-                                <exclude>%regex[.*/integration/.*ITest.*]</exclude>
+                                <exclude>%regex[com/otilm/core/integration/.*ITest.*]</exclude>
                             </excludes>
                         </configuration>
                     </plugin>
@@ -507,7 +507,7 @@
                                 <include>com/otilm/core/integration/service/**/*ITest.java</include>
                             </includes>
                             <excludes>
-                                <exclude>%regex[.*/integration/service/[D-Z][^/]*ITest.*]</exclude>
+                                <exclude>%regex[com/otilm/core/integration/service/[D-Z][^/]*ITest.*]</exclude>
                                 <exclude>com/otilm/core/integration/service/AcmeProfileServiceITest.java</exclude>
                                 <exclude>com/otilm/core/integration/service/AcmeServiceITest.java</exclude>
                                 <exclude>com/otilm/core/integration/service/CertificateServiceITest.java</exclude>
@@ -529,7 +529,7 @@
                         <artifactId>maven-surefire-plugin</artifactId>
                         <configuration>
                             <includes>
-                                <include>%regex[.*/integration/service/[D-Z][^/]*ITest.*]</include>
+                                <include>%regex[com/otilm/core/integration/service/[D-Z][^/]*ITest.*]</include>
                                 <include>com/otilm/core/integration/service/AcmeProfileServiceITest.java</include>
                                 <include>com/otilm/core/integration/service/AcmeServiceITest.java</include>
                                 <include>com/otilm/core/integration/service/CertificateServiceITest.java</include>
--- a/pom.xml
+++ b/pom.xml
@@ -452,8 +452,7 @@
 
     <profiles>
         <!-- Worker 1: everything outside the integration package.
-             The regex also excludes nested classes declared inside ITests; an Ant *ITest.java pattern
-             excludes only the outer class and causes its @Nested tests to run again on this worker. -->
+             The regex also excludes nested classes declared inside ITests. -->
         <profile>
             <id>test-non-integration</id>
             <build>
@@ -495,8 +494,7 @@
         </profile>
 
         <!-- Workers 3 & 4 split integration.service by the shared %regex boundary below.
-             Four heavy A-C classes are transferred to worker 4 to balance wall-clock cost.
-             CertificateServiceITest and CryptographicKeyServiceITest stay together because they share a context. -->
+             Some heavy A-C classes are transferred to worker 4 to balance wall-clock cost. -->
         <!-- Worker 3: integration.service nested subpackages + flat classes A-C, except transfers. -->
         <profile>
             <id>test-integration-service-1</id>
--- a/src/test/java/com/otilm/core/architecture/CiTestSplitTest.java
+++ b/src/test/java/com/otilm/core/architecture/CiTestSplitTest.java
@@ -145,9 +145,22 @@ class CiTestSplitTest {
     @Test
     void nonIntegrationProfileExcludesNestedIntegrationTests() throws Exception {
         List<String> excludes = profilePatterns("test-non-integration", "exclude");
+        String outerClass = "com/otilm/core/integration/service/SampleITest.java";
+        String nestedClass = "com/otilm/core/integration/service/SampleITest$SomethingTest.java";
 
-        assertThat(matchesAny(excludes, "com/otilm/core/integration/service/SampleITest.java")).isTrue();
-        assertThat(matchesAny(excludes, "com/otilm/core/integration/service/SampleITest$Nested.java")).isTrue();
+        assertThat(matchesAny(SUREFIRE_DEFAULT_INCLUDES, nestedClass))
+                .describedAs("""
+                        The nested class this guard uses must be one surefire's default <includes> actually
+                        claim — otherwise the exclusion assertion below passes vacuously and the guard proves
+                        nothing about the duplicate-execution regression it exists to catch.""")
+                .isTrue();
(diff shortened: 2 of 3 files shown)


src/test/java/com/otilm/core/architecture/CiTestSplitTest.java
```diff
@@ -99,19 +115,44 @@ void everyRunnableTestIsRunByExactlyOneCiProfile() throws Exception {
...
+        assertThat(coreExcludes).containsExactlyInAnyOrderElementsOf(expectedCoreExcludes);
+        assertThat(shard2Includes).containsAll(CORE_SHARD_TRANSFERS);
+    }
+
+    @Test
+    void nonIntegrationProfileExcludesNestedIntegrationTests() throws Exception {
+        List<String> excludes = profilePatterns("test-non-integration", "exclude");
+
+        assertThat(matchesAny(excludes, "com/otilm/core/integration/service/SampleITest.java")).isTrue();
+        assertThat(matchesAny(excludes, "com/otilm/core/integration/service/SampleITest$Nested.java")).isTrue();
```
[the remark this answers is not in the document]
ivosh: Fixed in 43ec5d32.

pom.xml
```diff
@@ -427,7 +427,9 @@
     </repositories>
 
     <profiles>
-        <!-- Worker 1: everything except integration ITests (service unit tests + all the rest). -->
+        <!-- Worker 1: everything outside the integration package.
+             The regex also excludes nested classes declared inside ITests; an Ant *ITest.java pattern
+             excludes only the outer class and causes its @Nested tests to run again on this worker. -->
```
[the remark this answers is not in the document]
ivosh: Fixed in 43ec5d32.

lubomirw (commented):
Automated multi-reviewer triage — 7 findings (COMMENT).

Sources: Copilot CLI (clean, zero findings), `guided-pr-review`, `superpowers:requesting-code-review`, `pr-hygiene`. 14 raw findings deduped to 10; the 7 posted here are the actionable set plus one judgment call. Two rows judged non-defects were dropped, and one claim (shard 2 "double-selecting" 14 nested classes) was verified harmless — JUnit Platform dedups by unique ID within one discovery request.

**Baseline confirmed independently by both deep reviewers:** replaying all four profiles over the real tree at `b582b17b` gives **zero gaps and zero double-runs** (230/84/78/48 across 440 runnable files), and the nested-dedup fix works — 16 nested classes (84 `@Test` methods, ~125 executions via `@ParameterizedTest`) were claimed by worker 1 before and by none now, corroborating the 125 figure in the description. The change is fundamentally sound; everything below is hardening.

---

**Not anchorable to a diff line** (the code it concerns is untouched by this PR, but this PR is what makes worker 1's correctness depend on it):

- `src/test/java/com/otilm/core/architecture/CiTestSplitTest.java:218` — **[Medium]** | Testing — `matches()` applies `%regex` patterns to `.java`-suffixed paths, but surefire substitutes `.java`→`.class` only for Ant globs, never for `%regex`. Both regexes currently in the pom end in `.*`, so both spaces agree and there is no live defect. But a plausible copy-paste of the neighbouring `.java` glob style into a regex (`%regex[...ITest.java]`) would replay here as *matching* while surefire matched *nothing* — leaving `test-non-integration` excluding nothing, restoring every duplicate execution, with all assertions in this class still green. _Suggested:_ evaluate `%regex` patterns against the `.class` form of the path while continuing to translate globs as today; optionally assert that no `%regex` in pom.xml ends in a literal `.java`.

src/test/java/com/otilm/core/architecture/CiTestSplitTest.java
```diff
@@ -94,19 +108,53 @@ void everyRunnableTestIsRunByExactlyOneCiProfile() throws Exception {
...
+        List<String> expectedCoreExcludes = new ArrayList<>();
+        expectedCoreExcludes.add(SERVICE_PACKAGE);
+        expectedCoreExcludes.addAll(CORE_SHARD_TRANSFERS);
+
+        assertThat(coreExcludes).containsExactlyInAnyOrderElementsOf(expectedCoreExcludes);
+        assertThat(shard2Includes).containsAll(CORE_SHARD_TRANSFERS);
+    }
+
+    @Test
+    void nonIntegrationProfileExcludesNestedIntegrationTests() throws Exception {
```
lubomirw: **[Medium]** | Testing

The partition guard is structurally blind to the entity this PR exists to fix. `everyRunnableTestIsRunByExactlyOneCiProfile` filters `.java` files, so it can never observe `Outer$Nested.class` entries. This new guard substitutes two hand-written paths (`SampleITest`, `SampleITest$SomethingTest`) that do not exist in the tree, and it only proves worker 1 *excludes* them — never that any worker *claims* them.

The self-check against `SUREFIRE_DEFAULT_INCLUDES` is a good anti-vacuity touch, but it still only validates the literal, not the real tree.

A related latent hole: a **static, non-`@Nested`** `*Test` class inside an integration ITest would now be claimed by no profile at all (it ran on worker 1 before this change, and JUnit does not execute static nested classes via their enclosing class). Verified not live today — 7 static nested classes exist under `integration/`, none named `*Test`/`*Tests`/`*TestCase`.

_Suggested:_ derive nested paths from the real tree instead of the synthetic literal (16 such classes already exist under `SigningProfileServiceImplITest`, `SigningRecordServiceITest`, `NativeArrayFilterSearchITest`) and extend the claim check: for each nested test class, assert it is either claimed by exactly one profile, or is an `@Nested` inner class whose enclosing class is claimed by exactly one profile — failing on static nested test classes that no profile claims.

<sub>flagged by guided-pr-review and superpowers (2 sources); normalized from High to Medium after verifying nothing currently breaks</sub>

src/test/java/com/otilm/core/architecture/CiTestSplitTest.java
```diff
@@ -43,6 +43,20 @@ class CiTestSplitTest {
             .of("test-non-integration", "test-integration-core", "test-integration-service-1",
                     "test-integration-service-2");
 
+    private static final String SERVICE_SPLIT_BOUNDARY = "%regex[.*/integration/service/[D-Z][^/]*ITest.*]";
+
+    private static final String SERVICE_PACKAGE = "com/otilm/core/integration/service/**/*ITest.java";
+
+    private static final List<String> SERVICE_SHARD_TRANSFERS = List
```
lubomirw, line 51: **[Medium]** | Testing

Nothing asserts that these six hand-curated paths resolve to real classes, and surefire silently ignores include/exclude patterns that match nothing.

So if one of the six is renamed or moved, the pom entry and this constant still agree, every guard stays green — and the class falls back to its letter-based shard (an A–C rename reverts to shard 1; a renamed `PQCITest`/`TimeQualityConfigurationSearchITest` reverts to `test-integration-core`). Even `everyRunnableTestIsRunByExactlyOneCiProfile` passes, because the class is still claimed by exactly one profile. The measured balance this PR bought degrades with no CI signal at all.

_Suggested:_ assert every entry in `SERVICE_SHARD_TRANSFERS` and `CORE_SHARD_TRANSFERS` resolves to an existing file under `TEST_ROOT` (e.g. `assertThat(TEST_ROOT.resolve(path)).exists()`), with a `describedAs` explaining that a stale path is dead surefire config — the same anti-vacuity spirit as `integrationRootContainsOnlyContextTests`.

<sub>flagged by guided-pr-review and superpowers (2 sources)</sub>

pom.xml
```diff
@@ -461,15 +461,15 @@
                         <artifactId>maven-surefire-plugin</artifactId>
                         <configuration>
                             <excludes>
-                                <exclude>com/otilm/core/integration/**/*ITest.java</exclude>
+                                <exclude>%regex[.*/integration/.*ITest.*]</exclude>
```
lubomirw: **[Low]** | Correctness

This regex drops the `com/otilm/core/` anchor the previous glob had, and no longer requires `ITest` to be a suffix — while workers 2–4 include only `com/otilm/core/integration/**`. A future `com/otilm/other/integration/FooITest.java` would therefore be excluded from worker 1 and claimed by no profile at all; `com/otilm/plugin/integration/service/ZebraITest.java` would be silently pulled into shard 2.

Verified latent, not live: all 213 `/integration/` paths sit under `com/otilm/core/integration`, and the file-level partition guard would catch the gap case for real `.java` files. This is drift-proofing.

_Suggested:_ anchor the exclude to the same root as the includes — `%regex[com/otilm/core/integration/.*ITest.*]` — so the exclude cannot outgrow the includes when a new top-level package or module appears.

<sub>flagged by guided-pr-review and superpowers (2 sources)</sub>

pom.xml
```diff
@@ -491,8 +493,8 @@
         </profile>
 
         <!-- Workers 3 & 4 split integration.service by the shared %regex boundary below.
-             The split is currently unbalanced, and rebalancing is manual as the suite grows. -->
-        <!-- Worker 3: integration.service nested subpackages + flat classes A-C. -->
+             Some heavy A-C classes are transferred to worker 4 to balance wall-clock cost. -->
```
lubomirw, line 496: **[Low]** | Documentation

The replaced comment warned that "the split is currently unbalanced, and rebalancing is manual as the suite grows". The new text states only what was done, so the pom no longer tells the next reader that this hand-curated transfer list is a point-in-time artifact, nor where the balance evidence came from.

_Suggested:_ restore a staleness note alongside the provenance — that the transfer list comes from the 20-run measurement, will drift as the suite grows, and should be re-measured when worker spread exceeds a stated threshold.

Related, for the description rather than the code: it reports three worker means (9:42, 9:19, 9:14) for a four-worker matrix, omitting the non-integration worker whose load this change alters most — and that worker is the basis for the 5.1% spread claim.

src/test/java/com/otilm/core/architecture/CiTestSplitTest.java
```diff
@@ -28,10 +28,10 @@
  * proves the partition by replaying surefire's include/exclude matching over the real test tree, so it catches drift in
  * any pattern.
  * <p>
- * The heavy {@code integration.service} package is split by leading class letter (A-C vs D-Z) via a shared
- * {@code %regex} boundary — excluded from {@code test-integration-service-1}, included by
- * {@code test-integration-service-2}. {@link #flatServiceSplitBoundaryMustBeConsistent} keeps the two sides identical
- * so the flat classes cannot silently gap or double-run.
+ * The heavy {@code integration.service} package is split by leading class letter (A-C vs D-Z), with heavy classes
```
lubomirw: **[Low]** | Documentation

This javadoc contradicts the invariant the tests actually assert. It says "the same patterns must be excluded from `test-integration-service-1` and included by `test-integration-service-2`", but shard 1 excludes 5 patterns while shard 2 includes 7 — shard 2 additionally claims the two `test-integration-core` transfers (`PQCITest`, `TimeQualityConfigurationSearchITest`), which shard 1 never excludes.

It also omits those core→shard-2 transfers entirely, even though this PR adds `coreTransfersMustBeExcludedFromCoreAndIncludedByServiceShard2` specifically to guard them; and it no longer names any guard method (the removed text linked `flatServiceSplitBoundaryMustBeConsistent`), so a reader cannot see which test enforces which half of the invariant.

_Suggested:_ scope the sentence to the service subset — e.g. "The service-package patterns (the A–C/D–Z boundary plus the service transfers) must be excluded from `test-integration-service-1` and included by `test-integration-service-2`; shard 2 additionally claims transfers from `test-integration-core`" — and `{@link}` both `serviceSplitPatternsMustBeConsistent` and `coreTransfersMustBeExcludedFromCoreAndIncludedByServiceShard2`.

<sub>flagged by pr-hygiene and guided-pr-review (2 sources)</sub>

src/test/java/com/otilm/core/architecture/CiTestSplitTest.java
```diff
@@ -43,6 +43,20 @@ class CiTestSplitTest {
             .of("test-non-integration", "test-integration-core", "test-integration-service-1",
                     "test-integration-service-2");
 
+    private static final String SERVICE_SPLIT_BOUNDARY = "%regex[.*/integration/service/[D-Z][^/]*ITest.*]";
+
+    private static final String SERVICE_PACKAGE = "com/otilm/core/integration/service/**/*ITest.java";
```
lubomirw: **[Nit]** | Maintainability

`SERVICE_PACKAGE` holds an Ant glob (`com/otilm/core/integration/service/**/*ITest.java`) and is used solely as `test-integration-core`'s expected exclude, so the name suggests a package constant rather than the one pattern it stands for.

_Suggested:_ rename to `CORE_SERVICE_EXCLUDE` or `SERVICE_PACKAGE_PATTERN`, matching the naming of `SERVICE_SPLIT_BOUNDARY` right above it.

src/test/java/com/otilm/core/architecture/CiTestSplitTest.java
ivosh: Fixed in 087445ce.

Replaced by `nestedTestClassesInsideIntegrationTestsRunOnExactlyOneWorker`, derived from the real tree via a new `TestClassTaxonomy.nestedClasses` parser. For every nested class inside a runnable integration test it asserts: never claimed by `test-non-integration`; a `@Nested` one claimed only by profiles that also claim its enclosing class (within one worker both selectors resolve to the same JUnit unique id, so that is not a duplicate — across workers it is); a non-`@Nested` one claimed by exactly one profile, which is the static-nested gap you flagged. An anti-vacuity assertion requires the derived population to be non-empty.

Mutation-checked: dropping the trailing `.*` from worker 1's exclude makes it fail on 16 real classes (`NativeArrayFilterSearchITest$OidEntryAltCodesFilterTest`, the `SigningProfileServiceImplITest`/`SigningRecordServiceITest` nests), not on a literal.

src/test/java/com/otilm/core/architecture/CiTestSplitTest.java
ivosh: Fixed in 087445ce.

Added `transferredClassesMustExist`: every entry of `SERVICE_SHARD_TRANSFERS` and `CORE_SHARD_TRANSFERS` must resolve to a regular file under `TEST_ROOT`, with a `describedAs` spelling out that surefire ignores a pattern matching nothing so the class silently reverts to its letter-based shard. Mutation-checked by renaming `PQCITest.java` in the constant — the guard fails.

src/test/java/com/otilm/core/architecture/CiTestSplitTest.java
ivosh: Fixed in 087445ce. Renamed to `CORE_SERVICE_EXCLUDE`, with a javadoc line saying it is `test-integration-core`'s exclude for the service package.

pom.xml
ivosh: Fixed in 087445ce, with one deliberate deviation.

Anchored as suggested: worker 1's exclude is now `%regex[com/otilm/core/integration/.*ITest.*]`, and the A-C/D-Z boundary is `%regex[com/otilm/core/integration/service/[D-Z][^/]*ITest.*]` on both sides.

I verified the semantics against surefire's own matcher rather than assuming — `ResolvedTest` 3.5.6 applies the regex to the class-file path relative to the test-classes root, so the anchored form matches `com/otilm/core/integration/service/SampleITest.class` and `...SampleITest$SomethingTest.class` while rejecting `com/otilm/plugin/integration/service/ZebraITest.class`. Exactly the drift you described.

I did not make `ITest` a suffix. The trailing `.*` is what covers the `Outer$Nested` entries, which is the regression this PR exists to fix — requiring a suffix would reopen it. The new nested guard now depends on that and says so in its failure message.

pom.xml
ivosh: Not fixing the pom comment; fixing the description half.

On the staleness note: I drafted one (provenance in #1997, re-measure when the integration-worker spread exceeds a threshold) and then dropped it. A comment stating that a hand-curated list is point-in-time tends to rot in exactly the way it warns about — the threshold becomes stale, the referenced run becomes unfindable. `transferredClassesMustExist` (added in 087445ce for your other comment) is the durable half: it fails when a transfer stops naming a real class. The provenance stays in #1997, which is linked from this PR. If you would still like the note, I will add it.

On the description: correct, and now amended. It reported only the three integration workers. The non-integration worker's post-change 20-run mean was 2:56 (min 2:25, p50 3:02, p90 3:08, max 3:10, sd 0:14), from the same commit 4bcbabd5 as the other figures, and the description now says explicitly that the 28-second / 5.1% spread is across the three integration workers only — the non-integration worker is a ~3-minute leg and is not part of that spread.

src/test/java/com/otilm/core/architecture/CiTestSplitTest.java
ivosh: Partly fixed in 087445ce.

The contradiction is gone: I deleted the sentence claiming the same patterns are excluded from shard 1 and included by shard 2, rather than rewriting it to enumerate both halves. The javadoc now states only the split rule, and the invariant lives where it is actually enforced — `serviceSplitPatternsMustBeConsistent` and `coreTransfersMustBeExcludedFromCoreAndIncludedByServiceShard2`, whose names and `describedAs` messages carry it.

I deliberately did not add the `{@link}`s back. The class javadoc had grown into a second, drifting description of the guards below it — that drift is what produced the contradiction you found. A prose index of method names would be the same liability, so I would rather the tests be the only statement of the invariant. Happy to reconsider if you would still prefer the links.

lubomirw (approved):

PR #2006, opened 2026-08-11
Prefilter the SCEP registration enrolment lookup by normalized subject

### LINKED ISSUE #1988: Prefilter SCEP registration lookup by subject in SQL
### Description

Flagged in PR #1968 review (performance, medium): `CertificateRepository.findRegisteredWithActiveRegistrationAuthorizationByRaProfileUuid` loads every REGISTERED certificate with an ACTIVE registration authorization under the RA profile, and the caller parses all of them to match the enrolment request. Since this runs on every unauthenticated initial enrolment, request cost grows linearly with the registration inventory.

Narrow the candidate set in SQL by prefiltering on the normalized subject, add a suitable composite or partial index, and apply SAN disambiguation only to the narrowed candidates. Existing registrations need a data migration so their stored normalized subject matches the form the prefilter compares against — otherwise pre-existing REGISTERED placeholders would no longer be found by enrolment.

Ref: https://github.com/OmniTrustILM/core/pull/1968#discussion_r3726692425

### Definition of Done

- Registration lookup prefilters by normalized subject in SQL instead of loading all active registrations
- Composite or partial index supports the new query
- Data migration normalizes/backfills the subject of existing registrations so they remain matchable
- SAN disambiguation runs only on the narrowed candidates
- Unit tests cover the scoped lookup

klaraf755:
Closes #1988.

## Problem
Every unauthenticated SCEP initial enrolment in registration mode loaded **all** REGISTERED certificates with an ACTIVE registration authorization under the RA profile, then normalized each candidate's subject DN in Java to find the match. Request cost grew linearly with the registration inventory, on an endpoint an unauthenticated client can drive (flagged in PR #1968 review).

## Change
- **Reuse `subject_dn_normalized`**: registration placeholders now populate it at write time (`CertificateUtil.applyRegistrationSubject`), `''` for SAN-only registrations. The canonical normalization lives in two `CertificateUtil` helpers that the matcher, the write path, the backfill, and the query parameter all share — one implementation, no drift.
- **Narrowed lookup**: the repository finder gains a `subjectDnNormalized`
  equality predicate; `ScepServiceImpl` computes the CSR's normalized subject once
  and passes it. The matcher keeps its Java-side subject re-check on the narrowed
  candidates (defense against stored-column drift; false matches impossible,
  drift misses fail closed).
- **Partial composite index** `(ra_profile_uuid, subject_dn_normalized)
  WHERE state = 'REGISTERED' AND archived = false`.
- **Backfill migration** (`V202608071000`, Java Flyway) for existing
  REGISTERED / PENDING_REGISTRATION / PENDING_APPROVAL placeholders. It seeds the
  RDN OID registry from `custom_oid_entry` merged with the built-in `SystemOid`
  codes and publishes it once — Flyway runs before Spring populates the registry,
  so without the seed a stored DN using a registry RDN code would fail to parse
  and be left permanently unmatchable.

## Tests
Unit + integration coverage for the helpers, the narrowed finder (including the
`''` SAN-only case), an RDN-order-equivalence enrolment test proving the SQL
prefilter and the Java matcher agree, and a migration ITest covering the
backfill matrix, the system-RDN-code seed with zero custom rows, and
custom-over-system precedence. Full suite green; no new Spring test-context
combination (`ContextSignatureGuardTest` unchanged).


src/test/java/com/otilm/core/service/scep/impl/ScepServiceImplRegistrationModeTest.java
```diff
@@ -300,7 +300,9 @@ private static ScepException scepRejection(Executable invocation) {
     }
 
     private void stubCandidates(Certificate... candidates) {
-        when(certificateRepository.findRegisteredWithActiveRegistrationAuthorizationByRaProfileUuid(RA_PROFILE_UUID))
+        when(certificateRepository
+                .findRegisteredWithActiveRegistrationAuthorizationByRaProfileUuidAndSubjectDnNormalized(
+                        eq(RA_PROFILE_UUID), anyString()))
                 .thenReturn(Arrays.asList(candidates));
     }
```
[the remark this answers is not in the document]
klaraf755: Fixed in 9cd2712c3. stubCandidates now matches eq(CertificateUtil.normalizeSubjectDn(new X500Name("CN=device-1"))) — the exact normalized subject the caller must derive from the CSR — instead of anyString(). If the production code stops normalizing the CSR subject or passes the wrong value, the stub no longer matches, the finder returns no candidates, and the matched/ambiguous/SAN-mismatch tests fail instead of passing on an over-broad matcher.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -1026,7 +1026,9 @@ private Certificate matchRegistration(ScepRequest scepRequest) throws ScepExcept
             throw new ScepException(REGISTRATION_REJECTION, FailInfo.BAD_MESSAGE_CHECK);
         }
         List<Certificate> candidates = certificateRepository
-                .findRegisteredWithActiveRegistrationAuthorizationByRaProfileUuid(raProfile.getUuid());
+                .findRegisteredWithActiveRegistrationAuthorizationByRaProfileUuidAndSubjectDnNormalized(
+                        raProfile.getUuid(),
+                        CertificateUtil.normalizeSubjectDn(scepRequest.getPkcs10Request().getSubject()));
         RegistrationIdentityMatcher.MatchResult result = RegistrationIdentityMatcher
```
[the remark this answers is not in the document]
klaraf755: Fixed in 9cd2712c3. The CSR subject is now captured once and rendered null-safe: `X500Name csrSubjectDn = scepRequest.getPkcs10Request().getSubject(); String csrSubject = csrSubjectDn == null ? "" : csrSubjectDn.toString();`, used by all the diagnostic log lines. This aligns the diagnostic rendering with the null-handling already applied to the same value by RegistrationIdentityMatcher.match and CertificateUtil.normalizeSubjectDn, so a SAN-only enrolment (absent subject) renders as the empty string instead of throwing.

src/main/java/com/otilm/core/dao/repository/CertificateRepository.java
```diff
@@ -200,21 +200,23 @@ AND NOT EXISTS (SELECT 1 FROM CertificateStatusPoll p WHERE p.certificateUuid =
...
-     * excluded — they cannot authenticate an enrolment.
+     * The pre-registrations a protocol enrolment presenting the given normalized subject can complete under the RA
+     * profile: REGISTERED placeholders whose registration authorization is ACTIVE, prefiltered by the stored normalized
+     * subject so the identity match receives only same-subject candidates. Registrations without a challenge have no
+     * authorization row and are excluded — they cannot authenticate an enrolment.
      */
     @Query("""
             SELECT c FROM Certificate c
                 WHERE c.raProfileUuid = :raProfileUuid
+                    AND c.subjectDnNormalized = :subjectDnNormalized
```
lubomirw: **[High]** | Correctness

During a rolling deployment, old pods can create registrations with a null normalized subject after the one-shot backfill, and this equality-only lookup makes those registrations permanently invisible to new pods.

_Suggested:_ Temporarily include rows whose normalized subject is null and retain Java-side matching for them; remove the fallback only after a later final backfill once old writers are gone.

src/main/resources/db/migration/V202608071100__certificate_registration_lookup_index.sql
```diff
@@ -0,0 +1,7 @@
+-- Backs the registration-mode enrolment lookup (RA profile + normalized subject over REGISTERED,
+-- non-archived placeholders), which runs on every unauthenticated initial enrolment. Partial index:
+-- REGISTERED is a small subset, so the index stays tiny and the lookup does an index scan instead of
+-- a sequential scan of the (potentially very large) certificate table.
+CREATE INDEX "idx_certificate_registration_lookup"
+    ON "certificate" ("ra_profile_uuid", "subject_dn_normalized")
```
lubomirw, line 6: **[High]** | Correctness

The API and database allow an unbounded normalized subject DN, so a sufficiently large value can exceed PostgreSQL's B-tree tuple limit and abort this migration or a later registration write.

_Suggested:_ Index a fixed-width digest with 
a_profile_uuid, while retaining exact normalized-subject equality in the query to reject hash collisions.

<sub>flagged by Copilot CLI and Superpowers</sub>

src/main/resources/db/migration/V202608071100__certificate_registration_lookup_index.sql
```diff
@@ -0,0 +1,7 @@
+-- Backs the registration-mode enrolment lookup (RA profile + normalized subject over REGISTERED,
+-- non-archived placeholders), which runs on every unauthenticated initial enrolment. Partial index:
+-- REGISTERED is a small subset, so the index stays tiny and the lookup does an index scan instead of
+-- a sequential scan of the (potentially very large) certificate table.
+CREATE INDEX "idx_certificate_registration_lookup"
```
lubomirw, line 5: **[Medium]** | Performance

Plain CREATE INDEX scans the certificate table while blocking inserts, updates, and deletes, which risks disrupting a rolling rollout on a large inventory.

_Suggested:_ Use CREATE INDEX CONCURRENTLY in a Flyway migration configured to run outside a transaction, including retry handling for an invalid residual index.

<sub>flagged by Copilot CLI and Superpowers</sub>

src/main/java/db/migration/V202608071000__RegistrationSubjectDnNormalizedMigration.java
```diff
@@ -0,0 +1,119 @@
...
+                } catch (RuntimeException e) {
+                    logger.warn("Skipping normalized subject backfill of certificate {}: the stored subject DN"
+                            + " does not re-parse, so registration-mode enrolment cannot match this row. Cause: {}",
+                            certificateUuid, e.getMessage());
+                    skipped++;
+                    continue;
+                }
+                update.setString(1, normalized);
+                update.setObject(2, UUID.fromString(certificateUuid));
+                update.addBatch();
```
lubomirw: **[Medium]** | Performance

The migration retains the full default JDBC result set and every pending update batch entry until the final xecuteBatch(), so startup memory grows with the entire registration inventory.

_Suggested:_ Process rows in bounded keyset pages and execute and clear update batches at a fixed size such as 1,000 rows.

lubomirw (changes requested):
Automated multi-reviewer triage — 4 findings (REQUEST_CHANGES).

src/main/java/com/otilm/core/dao/repository/CertificateRepository.java
klaraf755: Fixed in c0b60812b. The finder now includes rows whose normalized subject is NULL — `AND (c.subjectDnNormalized = :subjectDnNormalized OR c.subjectDnNormalized IS NULL)` — and the caller's RegistrationIdentityMatcher still does the authoritative exact subject check on the returned candidates, so a placeholder written by an old instance during a rolling upgrade (no normalized subject) stays matchable instead of permanently invisible. NULL rows are only produced transiently by such writers, so the prefilter still narrows in steady state; kept as a permanent fallback rather than a two-phase remove-later, so there is no follow-up migration to track. Covered by a new repository test (registrationLookupIncludesRowsWithNoNormalizedSubject).

src/main/java/db/migration/V202608071000__RegistrationSubjectDnNormalizedMigration.java
klaraf755: Fixed in c0b60812b. The backfill no longer materializes the whole inventory or one giant batch: the source query streams via a server-side cursor (setFetchSize, valid because the migration runs inside Flyway's transaction), and the update batch is flushed every 500 rows (executeBatch + clearBatch). Both the result set and the pending updates stay bounded regardless of registration count.

src/main/resources/db/migration/V202608071100__certificate_registration_lookup_index.sql
klaraf755: Accepting this for now rather than indexing a digest. Rationale: normalized subject DNs are short in practice, registration writes are authenticated (operator/connector), and the SQL prefilter is only a coarse narrowing step — RegistrationIdentityMatcher does the authoritative exact subject check on the returned candidates, so the index need not be exact. A column-type bound (varchar(n)) was considered and rejected: ALTER COLUMN TYPE on the certificate table takes an ACCESS EXCLUSIVE lock and scans the table, which is the very write-blocking a rolling upgrade should avoid. If it ever proves necessary, the cheapest mitigation is an app-level length bound on the normalized subject at the registration write path (no schema change).

src/main/resources/db/migration/V202608071100__certificate_registration_lookup_index.sql
klaraf755: Kept as a plain CREATE INDEX for consistency. All 48 existing index migrations in the repo are plain (non-concurrent), including on the certificate table, and there is no executeInTransaction/CONCURRENTLY precedent. This index is also partial (the REGISTERED subset only), so its build is cheaper than most of those. Introducing CONCURRENTLY here would pioneer a first-of-its-kind non-transactional-migration pattern that the test suite cannot exercise — Flyway is disabled in tests and the schema is Hibernate-built — so it would ship unvalidated. If the platform wants CONCURRENTLY index builds generally, that is better handled as its own cross-cutting convention change than pioneered unvalidated in this PR.

lubomirw (approved):

PR #2011, opened 2026-08-11
Return the required flag when listing custom attributes

klaraf755:
Core part of fix for https://github.com/OmniTrustILM/fe-administrator/issues/1993


src/main/java/com/otilm/core/dao/entity/AttributeDefinition.java
```diff
@@ -168,6 +168,7 @@ public CustomAttributeDefinitionDto mapToCustomAttributeDefinitionDto() {
         dto.setEnabled(enabled);
         dto.setVersion(AttributeVersion.fromIntVersion(attribute.getVersion()));
         dto.setResources(this.relations.stream().map(AttributeRelation::getResource).toList());
+        dto.setRequired(required);
```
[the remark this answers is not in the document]
klaraf755: Done.

lubomirw (commented):

src/test/java/com/otilm/core/integration/service/CustomAttributeServiceITest.java
```diff
@@ -114,6 +114,7 @@ void testListAttributes() {
         Assertions.assertEquals(attribute.getUuid(), attributes.getFirst().getUuid());
         Assertions.assertEquals(attribute.getName(), attributes.getFirst().getName());
         Assertions.assertEquals(attribute.getProperties().getLabel(), attributes.getFirst().getLabel());
+        Assertions.assertEquals(attribute.getProperties().isRequired(), attributes.getFirst().isRequired());
```
lubomirw, line 117: **[Low]** | Testing

The `Boolean.TRUE.equals(required)` guard added in this PR is load-bearing but untested. `V202402171510__UpdateAndOptimizeAttributesModelMigration.java:51` adds the column as `required BOOLEAN NULL`, and :77 backfills it from `(definition->'properties'->>'required')::bool`, which yields NULL for any legacy CUSTOM row whose definition JSON has no `properties` object. Because the DTO setter takes a primitive `boolean`, reverting to a direct `dto.setRequired(required)` would still compile via auto-unboxing and pass CI, then NPE on the list endpoint in production.

_Suggested:_ Add a second `AttributeDefinition` to the fixture with `definition.setRequired(null)` and assert the listed DTO reports `false`, so the null coalescing is pinned by a test.

src/main/java/com/otilm/core/dao/entity/AttributeDefinition.java
```diff
@@ -168,6 +168,7 @@ public CustomAttributeDefinitionDto mapToCustomAttributeDefinitionDto() {
         dto.setEnabled(enabled);
         dto.setVersion(AttributeVersion.fromIntVersion(attribute.getVersion()));
         dto.setResources(this.relations.stream().map(AttributeRelation::getResource).toList());
+        dto.setRequired(Boolean.TRUE.equals(required));
```
lubomirw, line 171: **[Low]** | Correctness

While this mapper is open: `mapToCustomAttributeDefinitionDto` still never sets `protectionLevel`, even though the field exists on the response DTO (`interfaces` `AttributeDefinitionDto:59`), the entity carries it as a column, and `mapToCustomAttributeDefinitionDetailDto` does set it (line 222). So `GET /v1/attributes/custom` returns `protectionLevel: null` for every row while the detail endpoint returns the real value — the same class of omission this PR fixes for `required`, one line away. It is pre-existing rather than introduced here, and fe-administrator reads `protectionLevel` only on the detail/form pages, so deferring is defensible.

_Suggested:_ Add `dto.setProtectionLevel(protectionLevel);` next to the new `setRequired` call while the file is open, which also avoids a second trip through the interfaces → core → fe release chain. If it is deliberately out of scope, raise it as a follow-up issue.

src/test/java/com/otilm/core/integration/service/CustomAttributeServiceITest.java
klaraf755: Added `testListAttributesWithNullRequired` in e867aa235 — saves a definition with `required = null` and asserts the listed DTO reports `false`. Added as a standalone test rather than in the shared fixture, since `testListAttributes` asserts on list size.

src/main/java/com/otilm/core/dao/entity/AttributeDefinition.java
klaraf755: Added `dto.setProtectionLevel(protectionLevel);` to `mapToCustomAttributeDefinitionDto` in e867aa235. The entity column is written on every create/update via `AttributeEngine` and backfilled to `NONE` by V202601191554 for existing rows, so mapping from the entity field (consistent with `required` in the same mapper) is safe. Covered by an assertion in the new test.

lubomirw (approved):

PR #2007, opened 2026-08-11
Break enum constants one per line

ivosh:
> Parent change [OmniTrustILM/dependencies#146](https://github.com/OmniTrustILM/dependencies/pull/146) is merged and `build-tools:1.4.3-SNAPSHOT` republished (`1.4.3-20260811.134904-4`), so CI here is green against the new profile.

## What

Reformat under the parent's new `alignment_for_enum_constants=49`. Enum tables were packed several constants to a line and staircased off the right margin, so adding a constant reflowed the whole block.

18 files changed, **−212 net lines**. The `.editorconfig` line keeps IntelliJ in parity so the IDE does not undo it.

## Two source changes the formatter cannot make

**Dangling comma.** A constant list ending in `,` collapses its terminating semicolon onto the last constant as `,;`. JDT has no setting for it, so the dangling comma is dropped in the four files that had one.

**Shredded trailing comments.** Re-indenting a trailing line comment leaves it shredded into a narrow column rather than rejoined — `join_lines_in_comments` does not merge a trailing comment's continuations. `CertificateStateTransition` had comments running one word per line:

```java
PENDING_APPROVAL_TO_REVOKED(PENDING_APPROVAL, REVOKED, CertificateEvent.REVOKE, ...), // approved
                                                                                      // revoke
                                                                                      // completed
                                                                                      // synchronously
                                                                                      // (connector
                                                                                      // 200)
```

Those move above their constant, which is also where they read as documentation of the transition. Net effect across the repo: shredded comment lines drop from 249 to 236.

## Also: one-line Javadoc

The `.editorconfig` picks up `ij_java_doc_do_not_wrap_if_one_line = true`, which stops IntelliJ exploding a one-line `/** ... */` into three on every reformat. That was pure diff churn rather than a gate failure — Eclipse preserves either form, so Spotless calls both clean and never corrected it. No source changes result.

## Verification

- `mvn spotless:check checkstyle:check`: clean across **1665 files**, 0 Checkstyle violations. Full compile passes.
- **No behaviour change**: all 17 reformatted files verified **token-identical** to `origin/main` with a comment- and string-aware lexer.
- Commit 2 records the reformat SHA in `.git-blame-ignore-revs`, and labels both entries — a bare list of SHAs gives no clue what either one was.

Issue #1964, opened 2026-08-05
Discovery v2 adapter lifecycle and core API


lubomirw:
### Description

The Core-side API surface for discovery v2, plus the v2 provider adapter's lifecycle operations and the access-control wiring behind them.

The FE-facing read surface is delivered. The adapter lifecycle and the write-side gaps below are not.

### Delivered

**Run message log** — `GET /v1/discoveries/{uuid}/messages`, paged oldest-first, `DiscoveryMessageDto {severity, code, message, occurrences, firstSeenAt, lastSeenAt}`. The detail carries `runMessageCount` rather than the log itself, so a client polling a live run does not re-fetch a bounded-but-large log on every poll. Contract: interfaces#895. Aggregated storage: #2127.

**Mapper extracted** — `com.otilm.core.mapper.discovery.DiscoveryDtoMapper` holds the detail, list, message and item mappings; `Discovery` no longer implements `DtoMapper` and carries no wire-mapping logic. Named for the resource, not for one DTO, because it produces four.

**Detail DTO completed** — `resources`, `stoppable`, `progress`, `runMessageCount`. A v1 run stores none of the first three, so the mapper synthesizes them exactly rather than defaulting: it targets certificates by definition, cannot be stopped at all, and has no connector reporting progress. `stoppable` is a new nullable column, snapshotted from the initiate response.

**Connector relays** — `GET /{connectorUuid}/resources`, `/attributes`, `/{resource}/attributes`. A v1-only connector is answered with a synthesized `certificates` and never called; both attribute relays refuse it 422, since it publishes no schema and an empty list reads as "no configuration needed". The `{resource}` segment is refused 422 twice over: against the contract's discoverable set before any call, and against the connector's own live `listSupportedResources`. Gated `@ExternalAuthorization(resource = Resource.CONNECTOR, action = ANY)` — never `DISCOVERY`, whose `hasObjectAccess = false` would skip per-connector ACLs.

**Items listing** — `GET /v1/discoveries/{uuid}/items`, one native `UNION ALL` over `discovery_item` and `discovery_certificate`. Filters apply inside each branch so a single-resource query reads one table; the count is two indexed counts summed. Certificate rows synthesize what a v1 provider never supplied — payload built at read time from the deduplicated `certificate_content`, number from staging order, reference from the fingerprint — and the number is assigned over the whole run then filtered, so it does not shift when a caller filters. Adds nullable `sequence` and `discovered_at` to `discovery_certificate`.

[...]

Issue #1959, opened 2026-08-05
Discovery v2 schema groundwork


lubomirw:
### Description

Schema groundwork for discovery v2 in Core: the mechanical `DiscoveryHistory` → `Discovery` rename ("History" is misleading — it is the live run) plus the v2 columns and the two new tables.

### Scope

- Migration 1 (separate mechanical commit): `ALTER TABLE discovery_history RENAME TO discovery` + constraint/index renames; IDE-rename the entity and references (repository is already `DiscoveryRepository`). No behavior change — existing discovery ITests pass unchanged.
- Migration 2: `discovery` gains `connector_interface_uuid` (NULL = v1 legacy), `run_meta jsonb`, `resources jsonb`, `last_applied_sequence`, progress columns, `run_messages jsonb`, `stopped_at`, `connector_state`.
- New `discovery_work` agenda table (`work_type STATUS|DRAIN|PROCESS`, `attempt`, `next_due_at`, unique per run+type) and generic staging table `discovery_item` (dense `sequence`, `unique_ref` dedupe constraint, `payload jsonb`, `processed_at`, `processed_error`).
- Entities + repositories; `DiscoveryWorkRepository` mirrors `CertificateStatusPollRepository` (no `FOR UPDATE` — advisory-lock exclusion; `ON CONFLICT` upsert; `OffsetDateTime`).

### Acceptance Criteria

- [ ] Full existing suite green after the rename commit (behavior-frozen)
- [ ] Repository ITests: upsert idempotency, due-ordering query, cascade delete
- [ ] Sonar quality gate passes with >= 80% coverage on new code

### References

- Epic: OmniTrustILM/ilm#267 (core plan Tasks 1–2)

PR #2012, opened 2026-08-11
Adopt the discovery v2 web API contract

### LINKED ISSUE #1991: Adopt the discovery v2 interfaces release in Core
### Description

The discovery v2 contract (OmniTrustILM/interfaces#811, #815, #830) lands on the `2.20.0-SNAPSHOT` line Core already tracks, so it arrives with **no opt-in and no version gate**. Two consequences hit Core the moment interfaces#830 merges, neither of them discovery feature work:

- **Two client-facing DTOs are renamed**, and Core references them in seven files: `DiscoveryHistoryDetailDto` → `DiscoveryDetailDto`, `DiscoveryHistoryDto` → `DiscoveryListDto`. The rename fixes the misnomer the platform is already correcting on the entity side in #1959 — the class named "History" is the live run, and the second is the list projection the entity already maps with `mapToListDto()`. The create request keeps the name `DiscoveryDto`, so no reference to it changes. (It must not be renamed to `DiscoveryRequestDto`: that simple name is taken by the connector-facing request, and `DiscoveryServiceImpl` needs both types in one file — `DiscoveryDtoNameCollisionTest` in interfaces guards this.)
- **Seven new abstract methods** appear on `interfaces/core/web/DiscoveryController`, so `DiscoveryControllerImpl` stops satisfying its interface until every one is implemented.

**This is the repair, not an optional slice.** Every Core build fails from the moment the interfaces snapshot publishes until this merges, including work with no connection to discovery. It carries no discovery behaviour and depends on none of #1959–#1963, so it lands first — reviewed and merge-ready before interfaces#830 merges, since the ordering cannot be reversed (the stubs reference types that do not exist until then).

### Scope

- Apply the two renames across every Core reference.
- Implement all seven new methods in `DiscoveryControllerImpl` as stubs answering 501, so an unimplemented endpoint says so rather than failing obscurely, and greps as a TODO. `ConnectorControllerImpl` needs no change.

| Endpoint | Method |
|---|---|
| `PATCH /v1/discoveries/{uuid}/stop` | `stopDiscovery` |
| `PATCH /v1/discoveries/{uuid}/resume` | `resumeDiscovery` |
| `PATCH /v1/discoveries/{uuid}/cancel` | `cancelDiscovery` |
| `GET /v1/discoveries/{uuid}/items` | `getDiscoveryItems` |
| `GET /v1/discoveries/{connectorUuid}/resources` | `listDiscoveryResources` |
| `GET /v1/discoveries/{connectorUuid}/attributes` | `getDiscoveryAttributes` |
| `GET /v1/discoveries/{connectorUuid}/{resource}/attributes` | `getDiscoveryResourceAttributes` |

The last three are keyed by **connector** UUID, not run UUID — same path shape, different entity, following the split `AuthorityInstanceController` already uses. The `{resource}` path segment binds by wire code and needs `IPlatformEnumConverterFactory` registered in Core's web config (#1964 owns that wiring); until then that relay does not bind even as a stub.

- Nothing else: no behaviour, no schema, no client wiring. Real implementations belong to #1961 and #1964.

### Acceptance Criteria

- [ ] Core compiles and the full suite passes against the interfaces release
- [ ] All seven endpoints answer 501 from Core's own stub
- [ ] No existing discovery behaviour changes; existing discovery ITests pass unchanged
- [ ] Sonar quality gate passes

### References

- Epic: OmniTrustILM/ilm#267
- Repairs the break introduced by OmniTrustILM/interfaces#830; blocks all Core work after it
- Real controller implementation: #1964. Client wiring and capability gate: #1961.

lubomirw:
Closes #1991. Repairs Core against the discovery v2 web API contract in OmniTrustILM/interfaces#830. No discovery behaviour — this only makes Core build again.

## Why this exists

interfaces#830 adds seven **abstract** methods to `interfaces/core/web/DiscoveryController` and renames two client-facing DTOs. Core depends on that module at the same `2.20.0-SNAPSHOT` line it publishes, so there is **no version gate**: the contract arrives unannounced and Core stops compiling the moment a snapshot publishes, including builds of work with no connection to discovery.

**This PR cannot go green before interfaces#830 merges** — it references types that do not exist in the published snapshot until then. It was developed against a locally installed build of that branch. The intended sequence:

1. Review this diff (reviewable while red).
2. Merge interfaces#830 → snapshot publishes.
3. **Merge #2024 first** (or rebase this branch over it): the same snapshot that delivers the discovery contract now also carries the interfaces#847 signing rename (`CONTENT_SIGNING` → `DOCUMENT_SIGNING`), whose core adoption is #2024. Without it, CI here fails on signing symbols that have nothing to do with discovery — verified locally against a build of the interfaces branch.
4. Re-run CI here → green → merge immediately.

That holds Core's broken window to one CI cycle. The order cannot be reversed.

## The renames

| Before | After |
|---|---|
| `DiscoveryHistoryDetailDto` | `DiscoveryDetailDto` |
| `DiscoveryHistoryDto` | `DiscoveryListDto` |

The class named "History" is the live run, not an archive of finished ones — the same misnomer #1959 corrects on the entity side. The second is the list projection the entity already maps with `mapToListDto()`, so `DiscoveryListDto` matches both the method and five existing `*ListDto` precedents in `interfaces`.

Two things deliberately unchanged:

- **The `DiscoveryHistory` entity keeps its name.** Renaming it is #1959's job, and mixing it in here would turn an unblocking fix into a schema discussion.
- **The client create request keeps the name `DiscoveryDto`.** An earlier revision of interfaces#830 renamed it to `DiscoveryRequestDto`, which collides with the long-standing connector-facing `com.otilm.api.model.connector.discovery.DiscoveryRequestDto`. `DiscoveryServiceImpl` needs both — it imports the client request for `createDiscovery` and constructs the connector request in `discoverCertificatesByProvider` — and Java has no import aliasing, so one would have to be fully qualified at every use. Worse than the verbosity: two same-named types where one is built from the other invites picking the wrong one, and the compiler only notices when the shapes differ. Reverted before merge, so no reference to the create request changes here.

## The seven stubs

All on `DiscoveryControllerImpl`. `ConnectorControllerImpl` needs no change — the three connector-scoped attribute relays live on `DiscoveryController`, because `ConnectorController`'s own attribute endpoints are generic across function groups and a provider-interface-specific relay does not belong beside them.

| Endpoint | Method |
|---|---|
| `PATCH /v1/discoveries/{uuid}/stop` | `stopDiscovery` |
| `PATCH /v1/discoveries/{uuid}/resume` | `resumeDiscovery` |
| `PATCH /v1/discoveries/{uuid}/cancel` | `cancelDiscovery` |
| `GET /v1/discoveries/{uuid}/items` | `getDiscoveryItems` |
| `GET /v1/discoveries/{connectorUuid}/resources` | `listDiscoveryResources` |
| `GET /v1/discoveries/{connectorUuid}/attributes` | `getDiscoveryAttributes` |
| `GET /v1/discoveries/{connectorUuid}/{resource}/attributes` | `getDiscoveryResourceAttributes` |

Note the last three are keyed by **connector** UUID, not run UUID — same path shape, different entity, following the split `AuthorityInstanceController` already uses.

Each answers **501**, not an empty result: a caller cannot distinguish an empty list from "nothing found", and silence here would read as a working endpoint returning no data.

They carry **neither `@AuditLogged` nor authorization annotations**, deliberately. There is no operation to audit and no resource to check while the body does nothing. Authentication still applies — `SecurityConfig` is `anyRequest().authenticated()` — but resource-level authorization, and the action catalogue `AuthResourceSynchronizer` pushes for these actions, belong to #1964 with the behaviour. The checked exceptions stay on the signatures so filling them in later does not change any signature.

`{resource}` binds by wire code via the `@InitBinder` `ResourceCodeConverter` registration on the controller (landed during review): a valid code reaches the 501 stub, an unknown code answers the platform's 422 `Unknown Resource Name ...`.

## Verification

- `mvn -B clean compile` — SUCCESS
- `mvn -B test -Ptest-non-integration` — **2250 tests, 0 failures**, including `ExternalServiceAuthorizationArchTest` and the other architecture tests
- `mvn -B spotless:apply` — clean, no changes to the touched files under the new enum-formatting style (#2007)

**ITests were not run locally** — they need database infrastructure. `DiscoveryServiceITest` is in the renamed set (compile-verified via `test-compile`), so CI's integration profiles are the real check on it.

One diagnostic worth recording, because it cost time and will recur: the first build against the renamed contract failed with ~400 errors — a wrong `AuditLogMessage` constructor and every JPA static metamodel class (`Approval_`, `Certificate_`, …) missing. All of it was **cascade**. A single duplicate-import error aborted the annotation-processing round, so Lombok and the Hibernate metamodel generator never ran. Only one error was real. When a Core build reports missing `*_` metamodel classes, look past them for the first genuine error rather than chasing the count.

Note: the lifecycle verbs changed to PATCH and `getDiscoveryItems` gained a `newlyDiscovered` filter during interfaces#830's review; the stubs here inherit both from the interface annotations, so no code change was needed beyond the signature realignment already pushed.

A later review round flipped the run detail's `resources` and `effectiveCapabilities` to always-present, so `DiscoveryHistory.mapToDto` now synthesizes both (`["certificates"]` / `[]` — exact for every pre-v2 run); the discovery v2 implementation (#1964) replaces the constants with stored targets and synced capabilities. The `{resource}` binding also moved from a global converter factory to the platform's per-controller `@InitBinder` editor pattern during review.


src/main/java/com/otilm/core/api/web/DiscoveryControllerImpl.java
```diff
@@ -142,4 +148,59 @@ public List<SearchFieldDataByGroupDto> getSearchableFieldInformation() {
...
+     * resource to check while the body does nothing. Authentication still applies -- SecurityConfig authenticates every
+     * request -- but the resource-level authorization these need, and the action catalogue AuthResourceSynchronizer
+     * pushes for them, belong to #1964 along with the behaviour. The checked exceptions stay on the signatures so
+     * filling them in later does not change the contract.
+     */
+
+    @Override
+    public PaginationResponseDto<DiscoveryItemDto> getDiscoveryItems(String uuid, Resource resource,
+            Boolean newlyDiscovered, int itemsPerPage, int pageNumber) throws NotFoundException {
+        throw notImplemented();
```
ivosh, line 179: **˜ Minor ** | Error handling

DiscoveryControllerImpl.java:166 throws `ResponseStatusException`, but the global `ExceptionHandlingAdvice.java:613` catches every `Exception` and maps it to HTTP 500 before Spring's response-status resolver can process it. All seven new endpoints therefore return `500 Internal server error` rather than the promised 501. Throw the established `NotSupportedException`, which has a specific 501 handler at `ExceptionHandlingAdvice.java:589`, or add a dedicated `ResponseStatusException` handler.

ivosh (dismissed):

src/main/java/com/otilm/core/api/web/DiscoveryControllerImpl.java
```diff
@@ -142,4 +147,59 @@ public List<SearchFieldDataByGroupDto> getSearchableFieldInformation() {
...
+    @Override
+    public List<BaseAttribute> getDiscoveryAttributes(String connectorUuid)
+            throws NotFoundException, ConnectorException {
+        throw notImplemented();
+    }
+
+    @Override
+    public List<BaseAttribute> getDiscoveryResourceAttributes(String connectorUuid, Resource resource)
+            throws NotFoundException, ConnectorException {
+        throw notImplemented();
```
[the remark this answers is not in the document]
lubomirw: Registered in `WebAppConfig.addFormatters` as of 981d03703 — `{resource}` now binds by wire code, so the stub answers rather than a 400 from failed enum conversion, and the real implementations inherit the binding.

src/test/java/com/otilm/core/integration/service/DiscoveryServiceITest.java
```diff
@@ -128,7 +128,7 @@ void tearDown() {
     void testListDiscoveries() {
         final DiscoveryResponseDto discoveryHistoryDTO = discoveryService
                 .listDiscoveries(SecurityFilter.create(), new SearchRequestDto());
-        final List<DiscoveryHistoryDto> discoveries = discoveryHistoryDTO.getDiscoveries();
+        final List<DiscoveryListDto> discoveries = discoveryHistoryDTO.getDiscoveries();
```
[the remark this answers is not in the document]
lubomirw: Renamed to `discoveryResponseDto` in 981d03703.

src/main/java/com/otilm/core/api/web/DiscoveryControllerImpl.java
lubomirw: Confirmed and fixed in 981d03703: the stubs throw `NotSupportedException`, whose dedicated handler maps to 501 with the standard `ErrorMessageDto` body. The reasoning is recorded on the helper so the next transitional stub does not repeat the mistake.

ivosh (approved):

src/main/java/com/otilm/core/config/WebAppConfig.java
```diff
@@ -73,6 +74,9 @@ public void addArgumentResolvers(List<HandlerMethodArgumentResolver> argumentRes
 
     @Override
     public void addFormatters(FormatterRegistry registry) {
+        // Binds IPlatformEnum-typed request/path parameters by wire code (e.g. "certificates", "keys"),
+        // never by Java enum member name -- the discovery controller's {resource} segment relies on it.
+        registry.addConverterFactory(new IPlatformEnumConverterFactory());
```
ivosh: **~ Minor** | Error handling

WebAppConfig.java:79 introduces a third, global mechanism for binding `IPlatformEnum` params, alongside the two Core already uses. The established neighbour pattern is a per-controller `@InitBinder` `PropertyEditor`: 13 editors in `src/main/java/com/otilm/core/util/converter/` are registered at 23 sites across 18 controllers — `ResourceControllerImpl.java:35`, `EventControllerImpl.java:30-31`, `TriggerControllerImpl.java:41`, `RuleControllerImpl.java:38`, `ActionControllerImpl.java:38`, `CallbackControllerImpl.java:35`, `AuthControllerImpl.java:44`, `ApprovalProfileControllerImpl.java:41`, `CustomAttributeControllerImpl.java:56-57`, `v2/ComplianceControllerImpl.java:35`, `v2/ComplianceProfileControllerImpl.java:48`, plus `CertificateControllerImpl.java:80-81`, `CryptographicKeyControllerImpl.java:57`, `VaultProfileControllerImpl.java:41`, `SigningProfileControllerImpl.java:47`, `ProxyControllerImpl.java:44`, `SettingControllerImpl.java:39`, `ConnectorControllerImpl.java:58-60`; the second mechanism is the three hand-written `Converter<String, …>` at WebAppConfig.java:107-121 (`KeyAlgorithm`, `SigningRecordStatisticsPeriod`, `OidCategory`).

I checked the blast radius and the registration is *not* a regression: `TypeConverterDelegate` consults a registered custom editor before the `ConversionService`, and a specific `Converter<String,X>` outranks a factory keyed on the `IPlatformEnum` interface, so every existing enum path/param stays on its current binder. Its only live effect today is the new discovery `{resource}`.

The visible consequence is an inconsistent failure mode. `ResourceCodeConverter` calls `Resource.findByCode`, which throws `ValidationException` → `ExceptionHandlingAdvice.java:231` → **422**; the factory throws `IllegalArgumentException`, surfacing as `MethodArgumentTypeMismatchException` → `ExceptionHandlingAdvice.java:158` → **400**. So `GET /v1/discoveries/{connectorUuid}/notAResource/attributes` answers 400 while `GET /v1/resources/notAResource/…` answers 422 for the same class of mistake.

Two readings: if the factory is meant to *replace* the per-controller editors platform-wide, say so in the comment at WebAppConfig.java:77 and open a follow-up to retire the 13 editors and the three now-redundant converters at 107-121, so the two do not drift; if the intent is just to make `{resource}` bind, the smaller, pattern-matching move is `@InitBinder registerCustomEditor(Resource.class, new ResourceCodeConverter())` on `DiscoveryControllerImpl`, matching the eleven siblings above and giving the same 422 they do. Either way this need not block the PR.

src/main/java/com/otilm/core/api/web/DiscoveryControllerImpl.java
```diff
@@ -145,4 +150,60 @@ public List<SearchFieldDataByGroupDto> getSearchableFieldInformation() {
...
+    }
+
+    @Override
+    public List<BaseAttribute> getDiscoveryAttributes(String connectorUuid)
+            throws NotFoundException, ConnectorException {
+        throw notImplemented();
+    }
+
+    @Override
+    public List<BaseAttribute> getDiscoveryResourceAttributes(String connectorUuid, Resource resource)
```
ivosh, line 209: **~ Minor** | Security

`DiscoveryControllerImpl.java:197` (and `:165`) bind a `Resource` from the request, but this controller registers no `@InitBinder`. That leaves the new global `IPlatformEnumConverterFactory` (`WebAppConfig.java:79`) as the binder, and it throws `IllegalArgumentException` for an unknown code. Spring's `TypeConverterSupport.convertIfNecessary` catches `IllegalArgumentException` and rewraps it as `TypeMismatchException` -> `MethodArgumentTypeMismatchException`, which `ExceptionHandlingAdvice.java:158-163` answers with `ErrorMessageDto.getInstance(ex.getMessage())`. That message is `Failed to convert value of type 'java.lang.String' to required type 'com.otilm.api.model.core.auth.Resource'` — so `GET /v1/discoveries/{connectorUuid}/bogus/attributes` puts an internal fully-qualified Java class name on the wire. CLAUDE.md's "Don't leak runtime details to the wire" names stack-frame class names specifically.

Compare the neighbour that binds the same enum from the same `{resource}` segment: `ResourceControllerImpl.java:33-36` registers `@InitBinder` -> `ResourceCodeConverter` -> `Resource.findByCode`, which throws `ValidationException` (`RuntimeException`, not `IllegalArgumentException`, so it is *not* rewrapped) and reaches the advice's `ValidationException` handler as a clean 422 `Unknown Resource Name bogus`. All 18 controllers that bind a platform enum today follow that `@InitBinder` pattern; this is the first one that does not.

Suggested fix, matching the siblings: add `@InitBinder` on `DiscoveryControllerImpl` registering `ResourceCodeConverter` for `Resource.class`. That both removes the class-name leak and makes the two `{resource}` segments answer the same status and message. If the global factory is meant to be the new house style instead, the fix belongs in `ExceptionHandlingAdvice.handleMethodArgumentTypeMismatchException` — return a shaped message rather than `ex.getMessage()`.

Need not block the PR: the leak is a class name, not data, and the endpoint answers 501 once binding succeeds.

ivosh (changes requested):

src/main/java/com/otilm/core/config/WebAppConfig.java
lubomirw: Fixed in 542ce7ee5 — your second reading was the intended one: the registration existed only to make {resource} bind, not to introduce a new house style. Gone from WebAppConfig; DiscoveryControllerImpl registers ResourceCodeConverter via @InitBinder like the eighteen siblings, so both discovery {resource} surfaces (path segment and the items listing's query param) answer the same 422 the rest of the platform gives an unknown wire code. With the editor in place the factory had no live consumer left, so removing it also removes the drift hazard.

src/main/java/com/otilm/core/api/web/DiscoveryControllerImpl.java
lubomirw: Fixed in 542ce7ee5 — @InitBinder + ResourceCodeConverter on DiscoveryControllerImpl, matching ResourceControllerImpl: an unknown code now takes findByCode's ValidationException to a clean 422 "Unknown Resource Name bogus", and the class-name-bearing 400 route is unreachable (the global factory registration is removed outright, per the sibling thread). The interfaces#830 items-listing Javadoc that named the factory as Core's mechanism is being reworded to stay mechanism-neutral.

src/main/java/com/otilm/core/api/web/DiscoveryControllerImpl.java
```diff
@@ -56,6 +64,14 @@ public void setDiscoveryService(DiscoveryExternalService discoveryService) {
         this.discoveryService = discoveryService;
     }
 
+    // The house pattern for binding a platform enum by wire code.
+    // ResourceCodeConverter goes through Resource.findByCode, whose ValidationException answers an unknown
+    // code with a clean 422 — the global ConversionService route would 400 with Spring's class-name message.
+    @InitBinder
```
ivosh, line 69: **~ Minor** | Documentation

DiscoveryControllerImpl.java:70 now registers ResourceCodeConverter, so a valid `{resource}` binds and reaches the 501 stub, while an invalid code follows the converter's 422 path. The PR description still says binding is deferred to #1964 and requests fail before reaching the stub. Update that acceptance criterion to describe the current 501/422 behavior; this need not block the PR.

src/main/java/com/otilm/core/api/web/DiscoveryControllerImpl.java
```diff
@@ -145,4 +161,60 @@ public List<SearchFieldDataByGroupDto> getSearchableFieldInformation() {
         return discoveryService.getSearchableFieldInformationByGroup();
     }
 
+    /*
```
ivosh, line 162: **~ Minor** | Documentation

DiscoveryControllerImpl.java:164 compresses stub behavior, audit and authorization policy, authentication, and signature compatibility into two dense paragraphs, burying the security-relevant invariant mid-block and re-describing neighboring classes. Restructure it as one summary sentence followed by `<p><b>Stub response:</b> ...`, `<p><b>Authorization:</b> ...`, and `<p><b>Compatibility:</b> ...`; keep SecurityConfig and interface details in their own documentation.

src/main/java/com/otilm/core/api/web/DiscoveryControllerImpl.java
```diff
@@ -56,6 +64,14 @@ public void setDiscoveryService(DiscoveryExternalService discoveryService) {
         this.discoveryService = discoveryService;
     }
 
+    // The house pattern for binding a platform enum by wire code.
```
ivosh: **. Nitpick** | Maintainability

DiscoveryControllerImpl.java:67 only labels the immediately visible `ResourceCodeConverter` registration as the "house pattern"; sibling controllers use the same binder without this narration. Remove this line while retaining lines 68–69, which explain the non-obvious 422-versus-400 behavior.

ivosh (dismissed):

src/main/java/com/otilm/core/api/web/DiscoveryControllerImpl.java
lubomirw: Fixed — the PR body paragraph predated the binder landing and contradicted the code (and the body's own later appendix). It now describes the current behavior: valid code -> 501 stub, unknown code -> 422 Unknown Resource Name. The #1964 AC was already reworded when the binding landed early.

src/main/java/com/otilm/core/api/web/DiscoveryControllerImpl.java
lubomirw: Fixed in e796a393c — one summary sentence plus labeled Stub response / Authorization / Compatibility paragraphs; the SecurityConfig re-description reduced to "Authentication is unaffected."

src/main/java/com/otilm/core/api/web/DiscoveryControllerImpl.java
lubomirw: Fixed in e796a393c — the narration line is gone, the 422-versus-400 rationale stays.

ivosh (dismissed):

ivosh (approved):

Issue #1974, opened 2026-08-06
Content-signing workflow + level orchestrator


ivosh:
### Description

The content-signing runtime at SIGNED and TIMESTAMPED level. `SigningWorkflowType.CONTENT_SIGNING` is already admin-configurable end to end - create, update, read, connector binding, formatting-attribute storage - so what is missing is execution.

**Profile surface.** `ManagedContentSigningWorkflow` (exactly `{signatureFormattingConnectorUuid, signatureFormattingConnectorAttributes}`) widens with `family` (validated at save: the connector must advertise the profile's family code. The existing save-time check accepts a connector carrying `FeatureFlag.CONTENT_SIGNING` on any of the four family interfaces; this narrows it to the interface the profile's family names), `maxLevel` (`SignatureLevel`, DP-15; additionally capped at save time by the connector's declared per-rung capability flags for the profile's family - DP-16), `timestampSource` (a reference to a TIMESTAMPING Signing Profile, required when maxLevel is `TIMESTAMPED` or higher; v1 save-time validation requires the referenced profile to be ILM-managed, D4, and to satisfy the record floor: recording on, DEFERRED_DURABLE minimum, DP-6 - this child writes that floor predicate and OmniTrustILM/core#2002 reuses it for the content- and raw-signing profiles themselves) and `documentSizeCap`. The matching wire DTOs widen in `interfaces` (OmniTrustILM/interfaces#832). The profile's `allowedRequestParameterGroups` and `defaultSignatureParameters` (OmniTrustILM/interfaces#889) are validated at save by #2117, which also permission-checks and merges request-provided parameters; step 1 dispatch carries the resulting effective object as `signatureParameters` on the PAdES computeDtbs request. Core mints `signingTime` at acceptance for determinism, and a content-signing profile may reference a time-quality configuration exactly as a timestamping profile does - the unversioned `time_quality_config_uuid` already sits on the profile header - in which case acceptance gates on `TimeQualityRegister` status. Content-signing profile models enter the same `SIGNING_PROFILE_CACHE` through the per-workflow mapper dispatch, so the hot path is a cache hit plus live resolution; resolved profiles stay never-cached, so certificate, key and connector changes take effect immediately.

[...]

Issue #1972, opened 2026-08-06
Extract common Signing Engine from signing/tsa


ivosh:
### Description

Extract profile resolution, the formatting-connector client, the Signer seam and signing records from `signing/tsa/` into a common Signing Engine package (`signing/engine/`), shared by three v1 consumers: the existing Timestamping Engine (stays in `signing/tsa/`), the new content-signing engine (`signing/contentsigning/`), and the raw-hash path - CSC `signHash`, which is the degenerate run of the same spine (resolve, sign, record; zero connector steps) and ships before any formatting connector exists. The existing timestamping tests pin behavior during the extraction.

**Scope boundary for the raw path (DP-8, 2026-08-11).** This child exposes the **seam** - a workflow-agnostic engine whose resolver return type and Signer entry point admit a raw-hash consumer. The `RAW_SIGNING` workflow that consumes it (the widened `ManagedRawSigningWorkflow`, `ResolvedManagedRawSigningProfile` with its resolver, the digest policy and the run itself) is built in epic OmniTrustILM/ilm#189, child OmniTrustILM/core#2004. The extraction work is identical either way, so this estimate is unchanged - only the ownership line moved.

**Lifts as-is** - already workflow-agnostic, mechanical repackage: the Signer seam (`Signer` / `SignerFactory` / `SignerCreator` / `CryptographicOperationServiceSigner`, which calls `signDataWithoutEventHistory` - the method that exists precisely for engine callers managing their own audit trail); the whole `signing/record/` strategy stack, whose `SigningRecordInput` already carries the `signature` / `signedDocument` / `dtbs` fields the TSP path leaves empty; and formatting-attribute persistence and replay (`AttributeOperation.WORKFLOW_FORMATTING`) - the storage and replay machinery lifts as-is, but (corrected 2026-08-13) the save-time definition fetch rides the legacy flat `SIGNATURE_FORMATTING` route (`SignatureFormattingApiClient.listFormattingAttributes`) for the content-signing branch too, not the per-operation contract. Switching that branch to aggregate the seven per-operation schemas is OmniTrustILM/core#1974's; this child does no route work.

**Widens** - the four deliberate narrowings:

- `SigningProfileResolver.resolve(...)` returns the sealed `ResolvedSigningProfile` instead of `ResolvedManagedTimestampingProfile`; `SigningProfileResolverFactory`'s return type widens with it, and its selection rule (first supporting resolver wins) is unchanged.
[...]

PR #2024, opened 2026-08-12
Rename content-signing workflows and stored values to document signing

### LINKED ISSUE #2023: Rename CONTENT_SIGNING to DOCUMENT_SIGNING, with the workflow_type migration
### Description

The Core half of DP-13, following OmniTrustILM/interfaces#846. Rename only - no
profile-surface widening and no orchestrator work, both of which stay in #1974.

**Touches.** `ContentSigningWorkflow` / `ManagedContentSigningWorkflow` /
`DelegatedContentSigningWorkflow` -> the `DocumentSigning` equivalents;
`SigningWorkflow`'s `permits` clause; `SigningProfileMapper`;
`SigningProfileServiceImpl`; the profile test fixtures and builders
(`ContentSigningWorkflowRequestDtoBuilder`); `ContentSigningFormattingMock`.

**One Flyway migration** rewrites the persisted vocabulary.

Corrected 2026-08-12, during implementation. This issue originally specified a
rewrite of `content_signing` to `document_signing`, on the grounds that both
columns are "read through `SigningWorkflowType.findByCode`". They are not.
`SigningProfile.workflowType` and `SigningProfileVersion.workflowType` are both
mapped `@Enumerated(EnumType.STRING)`, so Hibernate persists and reads
`Enum.name()`: the stored value is the uppercase `CONTENT_SIGNING`, never the
`content_signing` wire code. The original schema comment in
`V202605051246__signing.sql` says so directly - `-- cache: CONTENT_SIGNING |
RAW_SIGNING | TIMESTAMPING`. A lowercase `UPDATE` would have matched zero rows
and silently left every existing signing profile unreadable, since `Enum.valueOf`
throws on the stale name. The migration rewrites the uppercase constant in:

- `signing_profile_version.workflow_type`
- `signing_profile.workflow_type`, a denormalized cache of the former
- `connector_interface.features`, added to scope for the same reason:
  `ConnectorInterfaceEntity.features` is a `List<FeatureFlag>` mapped
  `@Enumerated(EnumType.STRING)` over a `TEXT[]`, so it stores `FeatureFlag`
  constant names, and OmniTrustILM/interfaces#846 renames
  `FeatureFlag.CONTENT_SIGNING`. Rewritten with `array_replace`. No connector
  currently advertises the flag, so this is a safety net for existing databases
  rather than an observed break.

There is no lenient path on any of the three, so the migration and the renamed
enum land in the same PR.

**Not changed: the historical migration comment.** `V202605051246__signing.sql`
still reads `-- cache: CONTENT_SIGNING | RAW_SIGNING | TIMESTAMPING`. It is an
already-applied migration; editing it changes its Flyway checksum and fails
validation on every existing deployment. The new migration documents the rename
instead. This is a deliberate deviation from the "no `CONTENT_SIGNING` in schema"
criterion below.

**Sequencing (revised 2026-08-12).** This train lands **first, ahead of all other
AdES work** - before #1972, #1974 and OmniTrustILM/fe-administrator#1997. Merges
the same day as OmniTrustILM/interfaces#846, prepared against a locally-installed
snapshot of that branch - the technique #1972's Task 0 already uses.

This reverses the original ordering, which listed #1972 as a blocker and pulled
the `model/signing/resolved/` artifacts it mints
(`ResolvedManagedContentSigningProfile`, `StaticKeyManagedContentSigningResolver`
and `SigningProfileMapper.toManagedContentSigningModel`) into this issue's scope
for renaming after the fact. With the rename going first, that inherited scope
disappears: none of those three exist on `main` today, and #1972 mints them under
the `DocumentSigning` vocabulary from the start. **#1972's plan needs a naming
edit** - it is currently written against `CONTENT_SIGNING` and cites the rename as
living in OmniTrustILM/interfaces#832 and #1974, both stale.

### Definition of Done

- [ ] Workflow classes, mapper, service, fixtures, builders and mocks renamed
- [ ] Flyway migration rewrites both `workflow_type` columns and the `connector_interface.features` array, using the uppercase enum constant, verified against a database carrying existing pre-rename rows
- [ ] No `content_signing`, `CONTENT_SIGNING` or `ContentSigning` remains in code or fixtures, and none in schema apart from the historical migration comment noted above
- [ ] Full suite green, unmodified apart from the renames

### Implementation

PR #2024. Compiles clean; `SigningProfileMapperTest` and
`SigningWorkflowModelsTest` green. The migration was applied through Flyway
against PostgreSQL 16 on a fixture carrying pre-rename rows: both `workflow_type`
columns and the `features` array were rewritten, with `TIMESTAMPING`,
`RAW_SIGNING` and `SECRET_ROTATION` siblings and `NULL` arrays left intact, and
zero residual `CONTENT_SIGNING` in any column.

ivosh:
Closes #2023

The Core half of DP-13, following OmniTrustILM/interfaces#846 (PR OmniTrustILM/interfaces#847). Rename only — no profile-surface widening and no orchestrator work, both of which stay in #1974.

Prepared against a locally-installed `2.20.0-SNAPSHOT` snapshot of the interfaces branch.

## What changed

- `ContentSigningWorkflow` / `ManagedContentSigningWorkflow` / `DelegatedContentSigningWorkflow` → the `DocumentSigning` equivalents, plus `SigningWorkflow`'s `permits` clause
- `SigningProfileMapper` (import, switch arm, `buildDocumentSigningWorkflowDto`) and `SigningProfileServiceImpl` (three switch arms, the feature-flag check, the error message)
- Profile test fixtures, builders (`ContentSigningWorkflowRequestDtoBuilder` → `DocumentSigningWorkflowRequestDtoBuilder`) and `ContentSigningFormattingMock` → `DocumentSigningFormattingMock`

## Correction — the migration rewrites the **enum constant name**, not the wire code

The issue specifies a migration rewriting `content_signing` → `document_signing`, on the basis that both columns are "read through `SigningWorkflowType.findByCode`". They are not.

`SigningProfile.workflowType` and `SigningProfileVersion.workflowType` are both mapped `@Enumerated(EnumType.STRING)`, so Hibernate persists and reads `Enum.name()` — the stored value is the uppercase `CONTENT_SIGNING`, never the `content_signing` code. The original schema comment in `V202605051246__signing.sql` confirms it: `-- cache: CONTENT_SIGNING | RAW_SIGNING | TIMESTAMPING`.

A lowercase `UPDATE` would have matched zero rows and silently left every existing signing profile unreadable, since `Enum.valueOf` throws on the stale name. The migration therefore rewrites the uppercase constant.

## Added scope — `connector_interface.features`

`ConnectorInterfaceEntity.features` is a `List<FeatureFlag>` mapped `@Enumerated(EnumType.STRING)` over `connector_interface.features TEXT[]`, so it stores `FeatureFlag` constant names. The interfaces PR renames `FeatureFlag.CONTENT_SIGNING`, which would orphan any array element still holding the old name. The same migration rewrites it with `array_replace`.

No connector in the platform currently advertises the flag, so this is a safety net for existing dev and production databases rather than an observed break.

## Not changed — the historical migration comment

`V202605051246__signing.sql` line 10 still reads `-- cache: CONTENT_SIGNING | RAW_SIGNING | TIMESTAMPING`. It is an already-applied migration; editing it changes its Flyway checksum and fails validation on every existing deployment. The new migration documents the rename instead. This is a deliberate deviation from the issue's "no `CONTENT_SIGNING` remains in schema" wording.

## Verification

- `mvn compile` and `mvn test-compile` green
- `SigningProfileMapperTest` and `SigningWorkflowModelsTest`: 19 tests, 0 failures
- Migration applied via Flyway against PostgreSQL 16 on a fixture carrying pre-rename rows — both `workflow_type` columns and the `features` array rewritten, with `TIMESTAMPING` / `RAW_SIGNING` / `SECRET_ROTATION` siblings and `NULL` arrays left intact, and zero residual `CONTENT_SIGNING` in any column

## Sequencing

Merges the same day as OmniTrustILM/interfaces#846, ahead of OmniTrustILM/fe-administrator#2011.

The issue lists #1972 as a blocker and puts its `model/signing/resolved/` artifacts in scope. #1972 is still open and none of `ResolvedManagedContentSigningProfile`, `StaticKeyManagedContentSigningResolver` or `SigningProfileMapper.toManagedContentSigningModel` exist on `main`, so that part of the scope is currently a no-op. If #1972 merges first, those three names need renaming on top of this branch.


src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -811,15 +811,15 @@ private void applyWorkflow(SigningProfile p, SigningProfileVersion version, Work
...
-                            "Signature formatting connector is required for content signing workflow");
+                            "Signature formatting connector is required for document signing workflow");
                 }
                 Connector contentConnector = connectorService
                         .getConnectorEntity(SecuredUUID.fromUUID(w.getSignatureFormattingConnectorUuid()));
-                validateFormattingConnectorFeature(contentConnector, FeatureFlag.CONTENT_SIGNING,
-                        SigningWorkflowType.CONTENT_SIGNING);
+                validateFormattingConnectorFeature(contentConnector, FeatureFlag.DOCUMENT_SIGNING,
+                        SigningWorkflowType.DOCUMENT_SIGNING);
                 version.setSignatureFormattingConnector(contentConnector);
```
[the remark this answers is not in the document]
ivosh: Fixed.

ivosh (commented):
Superseded — this automated pre-review was published by mistake on my own PR. Its inline comments have been removed; the findings were either fixed on this branch or carried forward outside this PR.

src/main/java/com/otilm/core/model/signing/workflow/DocumentSigningWorkflow.java
```diff
@@ -0,0 +1,18 @@
+package com.otilm.core.model.signing.workflow;
+
+import com.otilm.api.model.client.signing.profile.workflow.SigningWorkflowType;
+
+/**
+ * Sealed interface for the {@code DOCUMENT_SIGNING} workflow model.
+ *
+ * <p>
+ * Document signing has no common fields shared between managed and delegated signing.
+ */
```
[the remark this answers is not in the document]
ivosh: Fixed in dbce0902.

src/main/resources/db/migration/V202608121000__rename_content_signing_to_document_signing.sql
```diff
@@ -0,0 +1,19 @@
+-- Rename the CONTENT_SIGNING signing workflow type to DOCUMENT_SIGNING (DP-13).
```
lubomirw: **[Low]** | Maintainability

The migration comment includes the opaque internal planning reference `DP-13`, which is not useful to future readers without that context.

_Suggested:_ Remove ` (DP-13)` from the comment.

<sub>flagged by Hygiene</sub>

src/main/resources/db/migration/V202608121000__rename_content_signing_to_document_signing.sql
```diff
@@ -0,0 +1,19 @@
...
+UPDATE "signing_profile_version"
+   SET "workflow_type" = 'DOCUMENT_SIGNING'
+ WHERE "workflow_type" = 'CONTENT_SIGNING';
+
+-- Denormalized cache of the latest signing_profile_version row.
+UPDATE "signing_profile"
+   SET "workflow_type" = 'DOCUMENT_SIGNING'
+ WHERE "workflow_type" = 'CONTENT_SIGNING';
+
+-- FeatureFlag.CONTENT_SIGNING was renamed in the same contract change; the flag
```
lubomirw: **[Low]** | Maintainability

This comment narrates the feature flag's rename history rather than preserving only the durable persistence constraint.

_Suggested:_ Replace lines 15-16 with: `-- connector_interface.features stores enum names in a text array, so the persisted feature value requires the same replacement.`

<sub>flagged by Hygiene</sub>

lubomirw (commented):
Automated multi-reviewer triage - 2 findings (COMMENT).

lubomirw (dismissed):
Automated multi-reviewer triage - 1 finding (REQUEST_CHANGES).

src/main/resources/db/migration/V202608121000__rename_content_signing_to_document_signing.sql
```diff
@@ -0,0 +1,19 @@
+-- Rename the CONTENT_SIGNING signing workflow type to DOCUMENT_SIGNING (DP-13).
+-- Both workflow_type columns and connector_interface.features are mapped with
+-- @Enumerated(EnumType.STRING), so they persist the Java constant name rather
+-- than the wire code. Hibernate resolves them via Enum.valueOf and throws on an
+-- unknown value, so every stored occurrence must be rewritten here.
+UPDATE "signing_profile_version"
+   SET "workflow_type" = 'DOCUMENT_SIGNING'
```
lubomirw, line 3: **Deployment question** | Correctness

Is Core guaranteed to use a stop-the-world or `Recreate` deployment for this release? The default Helm `Deployment` uses rolling replacement, so an old pod may remain live after Flyway rewrites `CONTENT_SIGNING` to `DOCUMENT_SIGNING`; that pod could then fail to read migrated rows or write the old value back.

If rolling deployment is possible, could this use an expand-migrate-contract rollout that temporarily supports both names?

<sub>flagged by Copilot CLI; validated against the Core Flyway startup and Helm Deployment defaults</sub>

lubomirw (commented):
Additional finding.

src/main/resources/db/migration/V202608121000__rename_content_signing_to_document_signing.sql
ivosh: Fixed in fa820517.

src/main/resources/db/migration/V202608121000__rename_content_signing_to_document_signing.sql
ivosh: Fixed in fa820517.

src/main/resources/db/migration/V202608121000__rename_content_signing_to_document_signing.sql
ivosh: Good catch on the premise — I checked, and you're right that nothing forces a stop-the-world rollout: `charts/ilm/templates/core-deployment.yaml` sets no `strategy:` block, so Kubernetes defaults to `RollingUpdate`. Even at the default `replicas: 1`, `maxSurge` starts the new pod (which runs Flyway) before the old one terminates, so the overlap window genuinely exists.

Where I land differently is on the exposure inside that window: for this migration it is empty.

- **No `CONTENT_SIGNING` profile can exist.** The profile form filters the selectable workflow types down to Timestamping alone (`.filter((wt) => wt === SigningWorkflowType.Timestamping)`) and disables the rest, so this workflow has never been creatable through the UI.
- **No connector advertises the feature flag.** Confirmed by an org-wide code search across `OmniTrustILM` and `3KeyCompany`; the `go-sdk` handler base also deliberately tolerates platform-known flags that are absent from its generated vocabulary.

So the surviving old pod has no migrated row to misread and no path to write the old value back — all three `UPDATE`s are no-ops against real data today.

On expand-migrate-contract specifically: it isn't retrofittable from this PR. Tolerating both names requires the **old** code to accept the new constant, but these columns are `@Enumerated(EnumType.STRING)` and `Enum.valueOf` throws on an unknown name — so a dual-read window would need a lenient reader shipped in a *prior* release across three repos. OmniTrustILM/interfaces#846 weighed exactly this and rejected it ("no lenient path and no dual-read window"), which is why the rename is a same-day three-repo train rather than a phased rollout.

Worth noting the pattern is also well established here: `V202302071415__rename_cert_algorithms.sql` and `V202310131425__split_certificate_status.sql` both rewrite `certificate` values in place, on a far hotter table, with no dual-read window.

That said, your underlying question outlives this PR — the platform has no stated policy for enum-name migrations under rolling deployment, and the precedents above carry the same theoretical exposure. Happy to raise that as its own issue if you think it's worth pinning down; I'd rather not settle a platform-wide deployment policy inside a rename PR.

lubomirw: Totally agree, please let's create separate issue where we can address it, not in scope of any epic, and link it here.

lubomirw (approved):

src/main/resources/db/migration/V202608121000__rename_content_signing_to_document_signing.sql
ivosh: Raised as #2028 — standalone, not attached to any epic.

While writing it up I found one thing that strengthens your point beyond what I said above: there is no `strategy:` block anywhere in the ILM chart, not just in `core-deployment.yaml`. And the workload is templated (`kind: {{ $workloadType }}`, Deployment or StatefulSet) with `autoscaling.enabled` support, so the overlap isn't confined to the single-replica surge case — multiple Core replicas can be live at once.

The issue records the verified facts, the blast radius (110 `@Enumerated(EnumType.STRING)` mappings across 59 files, several on hot tables), and five options. My leaning is the combination of a lenient deserialization seam plus a startup guard that refuses to run code older than its schema — that buys a dual-read window for every future rename without a three-release dance, and turns silent corruption into a loud failure.

Left #2024 out of scope there, per the reasoning above.

PR #2035, opened 2026-08-12
Restore ASCII diagrams collapsed by comment reflow

### LINKED ISSUE #1661: Apply centralized Java formatting (core)
Part of the centralized Java formatting epic (OmniTrustILM/ilm#255).

Apply the canonical Java formatting to `core`.

### Work
- Scaffolding (editorconfig, gitattributes, hooks, `.idea/codeStyles`); parent → `dependencies:1.4.1-SNAPSHOT`.
- One-shot `spotless:apply` + braces (NeedBraces) + wildcard expansion (IntelliJ Optimize Imports) + 52 unused imports removed.
- Formatting commit recorded in `.git-blame-ignore-revs`. CI `build_pr.yml` already runs a `spotless:check` job.

### Status
Complete on branch `feat/java-formatting`: compiles clean, `spotless:check` 0, `checkstyle:check` 0 across 1,395 files. PR pending.

ivosh:
Restores three query-graph diagrams and one wire-format spec that were flattened into running paragraphs by an earlier mechanical reformat.

## What happened

The diagrams live in plain `/* */` block comments. The shared profile sets `comment.format_block_comments=true` with `join_lines_in_comments=true`, so `4e8d54ee` ("Apply Spotless formatting and Checkstyle fixes") joined every tree line into one paragraph and re-wrapped it at 120 columns. All indentation and tree structure was lost, leaving text like:

```
* Certificate (root) |-- NOT archived |-- state == ISSUED |-- validationStatus IN (VALID, EXPIRING) |-- keyUuid IS
* NOT NULL |-- ALL items must have valid algorithm ...
```

## Changes

**`CertificateEligibilityUtil`** — the SCEP CA, CMP Signing and Digital Signing query graphs, recovered from `8800a5f7` (the commit immediately before the reflow) and wrapped in `// @formatter:off` / `// @formatter:on`. No profile change was needed: `use_on_off_tags` is already enabled with these exact tags.

The SCEP and CMP diagrams are byte-identical to the originals. The Digital Signing diagram differs in two respects, both deliberate:

- its ETSI line was 128 characters, so the trailing parenthetical is wrapped onto an indented continuation line to stay within the 120-column margin;
- the child under `ALL private keys must meet criteria` was indented with bare spaces even though a sibling follows it, which loses the vertical rail and can read as closing the tree. It now uses `|   |--`, matching the convention the SCEP diagram uses for non-final children. The CMP diagram keeps bare-space indentation there, because in that tree it genuinely is the last child.

**`SecretsUtil`** — the `v1|...` encoding spec had been absorbed into the sentence above it. It is now `{@code ...}` in its own paragraph, which survives the formatter without needing `@formatter:off`.

Promoting the spec out of prose into `{@code ...}` turns a loose sentence into an exact literal, so it was corrected to match what the code actually emits: the version token is lowercase (`SecretEncodingVersion.V1("v1")`, and `isSecretStringV1` gates on `parts[0].equals("v1")`), and the secret and salt segments are Base64 while the count is plain decimal. The spec now reads `v1|base64(encryptedSecret)|base64(salt)|count` — the segment holds the PBE ciphertext, not the plaintext, since the only caller passes `encryptedSecret`.

Note the same comment text is duplicated in `software-cryptography-provider` (`com/czertainly/cp/soft/util/SecretsUtil.java`), which this PR does not touch; that copy remains drifted and is worth a follow-up.

## Why `@formatter:off` and not `<pre>`

`<pre>` is not reliable protection here. `comment.format_source` is not set in the profile, so Eclipse's default of `true` applies and it formats Java source *inside* `<pre>` blocks. `V1|secret|salt|count` parses as a bitwise-or expression, and a `<pre>`-only version of it was rewritten to `V1 | secret | salt | count`.

The existing `<pre>` tables in `SnowflakeSerialNumberGenerator` and `InstanceIdResolver` survive only because their content does not parse as Java. That is luck rather than protection, so the diagrams use the explicit on/off tags instead.

The diagrams also stay in plain `/* */` rather than being promoted to Javadoc. A Javadoc `<pre>` block nested inside `@formatter:off` does survive Spotless byte-identical, but the diagrams contain six bare `&` characters (`usage & SIGN`, `usage & (DECRYPT | SIGN)`, ...) and `javadoc -Xdoclint:all` rejects those as `bad HTML entity`. Escaping them to `&amp;` would defeat the verbatim restoration, so the plain block-comment form is deliberate.

## Verification

- `spotless:apply` over the restored files is a no-op — output is byte-identical, so the diagrams are stable under the gate.
- `spotless:check` and `compile` pass on top of the merged `main`.
- A repository-wide scan for art still outside both `<pre>` and `@formatter:off` returns no remaining hits. The equivalent scan over `interfaces` and `scheduler` found no damage in either.

`ProtectionMacValidator` was examined and deliberately left alone: its inline ASN.1 block was replaced with an `@see` link to RFC 4210 §5.1.3.1 by #1989, which is an intentional improvement rather than reflow damage.

Comments only — no behaviour change.


src/main/java/com/otilm/core/util/SecretsUtil.java
```diff
@@ -190,7 +190,10 @@ public static String calculateSecretContentFingerprint(SecretContent secretConte
     }
 
     /**
-     * Encoded the secret value into string V1|secret|salt|count
+     * Encodes the secret value into a string.
+     *
+     * <p>
+     * {@code v1|base64(secret)|base64(salt)|count}
```
lubomirw: **[Low]** | Documentation

The exact V1 wire-format literal says `base64(secret)`, but the only caller passes `encryptedSecret`; readers can reasonably infer that plaintext is merely Base64-encoded.

_Suggested:_ Document the segment as `base64(ciphertext)` or `base64(encryptedSecret)` and align the parameter name/description.

<sub>flagged by Guided, Superpowers</sub>

lubomirw (commented):

src/main/java/com/otilm/core/util/SecretsUtil.java
ivosh: Fixed in ee5615a3.

Agreed — the only caller passes `encryptedSecret` (the PBE ciphertext), so `base64(secret)` was misleading. The literal is now `v1|base64(encryptedSecret)|base64(salt)|count`, and the parameter plus its `@param` description were renamed to match.

lubomirw (approved):

Issue #2005, opened 2026-08-10
[core] RAW_SIGNING workflow runtime


ivosh:
### Description

Core umbrella under epic OmniTrustILM/ilm#189 (SM: Raw Signing). This is the
epic's substance: the `RAW_SIGNING` workflow, its resolver, and the degenerate
run of the common Signing Engine that signs a client-supplied digest.

DP-8 (2026-08-10) settled that CSC `signHash` resolves a `RAW_SIGNING` Signing
Profile through the CSC protocol profile's `rawSigningProfileRef`, rather than
riding the credential's `DOCUMENT_SIGNING` profile with formatting skipped. The
workflow type is a property of the profile, not of the request, so that
decision is what gives this umbrella something to build.

**The starting point is real but empty.** The enum value, the sealed
`RawSigningWorkflow` (permitting two empty records), the request DTO and
profile-save handling all exist. Missing: any resolver - the sealed
`ResolvedSigningProfile` hierarchy has no raw-signing member - any engine run,
and any digest policy.

**The run adds nothing to the engine.** Step 2 of the section 3.2 machine and
nothing else: no DTBS construction, no `formattingContext`, no timestamps, no
validation material, no level state machine, no connector call. That is why
this epic ships no connector work and no helm work.

**Estimate (agent execution - review/verify/supervise):** 2.75 md committed,
1.5 md gated on NC14
**Human baseline:** 9 md committed, 4 md gated on NC14

### Acceptance Criteria

- [ ] A `RAW_SIGNING` Signing Profile can be onboarded with a scheme, a
      certificate, allowed digest algorithms and a max-digests cap, and
      profile save refuses a formatting connector
- [ ] A client-supplied digest is signed through the common Signing Engine
      with zero connector calls and no state machine, and leaves a signing
      record
- [ ] A digest algorithm outside the profile list, and a request above the
      cap, are both refused explicitly
- [ ] All children closed; Sonar clean and coverage >= 80 % on each

### Children

- [ ] #2004 - RAW_SIGNING workflow, resolver and engine run - 2 / 6.5 md
- [ ] #2078 - Signer seam prehashed-digest dispatch (DP-20) - 0.75 / 2.5 md

**Total: 2.75 / 9 md committed**
**Gated: 1.5 / 4 md**

### Gated (not filed yet)

- Standalone internal raw-signing API endpoint - the client-facing non-CSC
[...]

Issue #1970, opened 2026-08-06
[core] AdES signing engine


ivosh:
### Description

The AdES signing engine in Core: the common Signing Engine extracted from the timestamping code, the synchronous content-signing workflow with its level state machine, document transport, the timestamp source bridge, the record-policy floor, the native document-signing controller and the certificate-eligibility rules. The asynchronous lifecycle and the LT extension step (`extendToLevel` orchestration) live under epic OmniTrustILM/ilm#354, umbrella OmniTrustILM/core#2115.

Every seam this umbrella touches is an actual class in today's `core`, verified against the checkouts, not an invention. Four deliberate narrowings widen (the profile-model cache loader literally says "for now"), and the timestamping path's behavior is pinned by its existing test suite throughout.

Epic boundaries (settled; validation split out 2026-08-09): **SM: AdES owns the engine.** SM: CSC API (OmniTrustILM/ilm#188) consumes it. Trust configuration and validation orchestration moved to SM: Signature Validation (OmniTrustILM/ilm#186), umbrella OmniTrustILM/core#1998. One-time-key schemes and sessions moved to SM: One-time key and session-key Signing (OmniTrustILM/ilm#191).

**Estimate (agent execution - review/verify/supervise):** 15 md
**Human baseline:** 45.5 md

### Acceptance Criteria

- [ ] A content-signing Signing Profile signs at SIGNED and TIMESTAMPED through a formatting connector; private keys never leave the Core-managed token
- [ ] Every operation leaves a signing record, and a TIMESTAMPED signature traces to its timestamp record by token serial
- [ ] All children closed; Sonar clean and coverage >= 80 % on each

### Children

- [x] #1972 - Extract common Signing Engine from signing/tsa - 2.5 / 8 md
- [x] #1974 - Content-signing workflow + level orchestrator - 3 / 10 md
- [ ] #1975 - Document transport (digest XOR document) - 1 / 3 md
- [x] #1976 - Timestamp source bridge - 1 / 3 md
- [ ] #2002 - AdES record-policy floor: mandatory recording, DEFERRED_DURABLE minimum, rebind onto #253 templates - 1 / 3 md

- [ ] #2010 - api/web native document-signing controller - 1.75 / 4.5 md
- [x] #2023 - Rename CONTENT_SIGNING to DOCUMENT_SIGNING, with the workflow_type migration - 0.25 / 1 md
[...]

Issue #844, opened 2024-09-03
CT logs discovery provider not visible


doloban:
Hello,
I have installed 0.0.0 version of CZERTAINLY, in custom .yaml values I have also enabled ct logs provider with the following code:

ctLogsDiscoveryProvider:  
  enabled: true 
  image:
    tag: develop-latest
    pullPolicy: Always 
  logging:
    level: "DEBUG"

I can not see the connector in connectors list though, even after opening CZERTAINLY in anonymous windows or after pressing CTRL + F5. But CT logs pod is running, can be seen on screenshot.
![image](https://github.com/user-attachments/assets/3b4889b6-39db-444a-941f-28da67833de2)

Best Regards,
Denys Doloban

3keyroman:
Did you install or upgrade? There is a `post-install` hook in the Helm chart that will register the connector only after install, see here: https://github.com/CZERTAINLY/CZERTAINLY-Helm-Charts/blob/develop/charts/czertainly/templates/hooks/register-connectors-job.yaml.

If you do the upgrade, it will not be executed. The connector will be deployed, but you should add it manually, for example using the administrator web interface. Go to Connectors, add new, and provide URL `http://ct-logs-discovery-provider-service:8080` with no auth and connect.

doloban:
I completely forgot to manually add the Connector as I thought it would me added automatically. I have now successfully registered Connector, thank you!

PR #2036, opened 2026-08-12
Extract the shared signing engine and restore MQ formatting routing

### LINKED ISSUE #1972: Extract common Signing Engine from signing/tsa

ivosh:
**TL;DR — the signer, the certificate validator and the scheme resolver move out of `signing/tsa` into a new `signing/engine`, so content signing and raw signing can reuse them without dragging RFC 3161 along.**

Closes #1972.

## Why

This machinery was written for timestamping. It lived under `signing/tsa`. Its error type was `TspException`, which carries an RFC 3161 failure-info bit.

That shape is fine for TSP and wrong for everything else. Any other workflow wanting a signer had to duplicate the code or import the TSP vocabulary. Neither works for #1970 (AdES) or #2005 (raw signing).

## What moved

| Was | Is now |
| --- | --- |
| `signing/tsa/signer/*` | `signing/engine/signer/*` |
| `signing/tsa/certificate/*` | `signing/engine/certificate/*` |
| `signing/tsa/resolver/*` | `signing/engine/resolver/*` |
| `signing/tsa/CertificateChain` | `signing/engine/CertificateChain` |

The moves are pure relocations plus the error-type swap below. No signing logic changed.

The content-signing workflow's own artifacts — the formatting client and the profile resolver — land in a new `signing/contentsigning` package. It is named for the workflow type, not the standards family: an `ades` package would re-acquire exactly the format vocabulary the engine is meant to stay clear of, and content signing admits non-AdES document formats later. `interfaces` publishes the same machinery under `signatures.contentsigning`.

## The new error vocabulary

`SigningEngineException` replaces `TspException` on the shared seams. It carries a `SigningEngineFailure` classification instead of a protocol bit.

`TspErrorMapper` translates back at the TSP boundary. **Every failure-info bit on the wire is unchanged.** `TspErrorMapperTest` pins each mapping.

Each exception carries two messages, deliberately. `operatorMessage()` holds diagnostic detail and reaches the log only. `getMessage()` holds the wire-safe text, so `PlatformException.safeMessage()` cannot leak internals to a client.

A connector fault raised inside a named step keeps `CONNECTOR_FAULT`. It does not collapse to `STEP_FAILED`. An operator can therefore still tell "the connector was down" from "our step broke".

## A real bug this fixes

`ProxyClientConfig` never declared MQ beans for signature formatting. Those beans simply did not exist.

**A formatting connector with a proxy set silently fell back to the REST client.** The call bypassed the proxy altogether. Nothing logged a warning.

Both formatting clients now route through `ConnectorApiFactory`, and the MQ beans are declared. `ConnectorApiFactorySignatureFormattingITest` covers the REST-vs-MQ routing.

## The guardrail

`SigningEngineIsolationArchTest` fails the build if anything under `signing/engine` depends on:

- a workflow package (`signing.tsa`, `signing.contentsigning`),
- RFC 3161's error vocabulary (`api.interfaces.core.tsp`),
- the content-signing formatting DTOs.

A fourth rule keeps the resolved-model hierarchy out of `signing.tsa`, since every workflow shares it.

## Rebased onto the CONTENT_SIGNING revert

This branch was written during the DP-13 window, when the workflow type was briefly `DOCUMENT_SIGNING`. #2065 and OmniTrustILM/interfaces#867 reverted that. The three artifacts this branch *mints* carried the only `DOCUMENT_SIGNING` references left anywhere, so the revert never touched them and the rebase alone did not compile.

The last commit absorbs it: `ResolvedManagedDocumentSigningProfile`, `StaticKeyManagedDocumentSigningResolver` and `SigningProfileMapper.toManagedDocumentSigningModel` take their `ContentSigning` forms.

**One contract change rides along, and it is not a rename.** `embedValidationMaterial` no longer exists. `extendToLevel` fuses the fetch and the embed into the single egress-bearing operation, so the client exposes that instead — same seven operations, one of them reshaped.

Two things the new contract offers are deliberately **not** exposed here: the per-operation `listFormattingAttributes(operation)` route and `extendToLevel`'s status/cancel companions. Profile save still fetches attribute definitions through the legacy flat `SignatureFormattingApiClient` for the content-signing branch. Switching that route is #1974's, and this child does no route work.

## Before this can merge

**This branch does not compile against the current `main` of interfaces.** It needs the content-signing formatting API clients from OmniTrustILM/interfaces#844, which is still open.

Merge that one first. Then re-run CI here. Until then a red build here means the missing dependency, not a defect in this diff.

## Testing

Against a locally-installed snapshot of the interfaces branch: **4096 unit tests and 105 integration tests, 0 failures**, 2 pre-existing skips. `SigningEngineIsolationArchTest` and `ContextSignatureGuardTest` both green.


src/main/java/com/otilm/core/signing/contentsigning/formatting/ContentSigningFormattingClient.java
```diff
@@ -0,0 +1,106 @@
...
+    public SignedDocumentResponseDto embedArchiveTimestamp(ApiClientConnectorInfo connector,
+            EmbedTimestampRequestDto request) throws SigningEngineException {
+        return call("embedArchiveTimestamp", connector, client -> client.embedArchiveTimestamp(connector, request));
+    }
+
+    private <T> T call(String step, ApiClientConnectorInfo connector, ConnectorCall<T> operation)
+            throws SigningEngineException {
+        try {
+            return operation.execute(connectorApiFactory.getContentSigningFormattingApiClient(connector));
+        } catch (ConnectorException e) {
```
lubomirw: **[Medium]** | Error handling

The interfaces REST client can throw an unchecked `IllegalStateException` when a successful response has no body, while this boundary catches only `ConnectorException`. That bypasses the intended `SigningEngineException` / `CONNECTOR_FAULT` mapping; the MQ implementation already reports the same condition as `ConnectorException`.

_Suggested:_ Prefer normalizing interfaces PR #844 so its REST client throws `ConnectorException`, or catch both exception types here, and add a concrete-client empty-body test.

<sub>flagged by Guided and Copilot CLI</sub>

lubomirw (changes requested):
Automated multi-reviewer triage — 3 findings (`REQUEST_CHANGES`).

The following findings could not be anchored to a right-side diff line:

- `src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java:908` — **[High] Correctness:** Interfaces PR #843 allows `CONTENT_SIGNING` only on the PAdES, XAdES, CAdES, and JAdES family interfaces, but Core validates it only on `SIGNATURE_FORMATTING`; the test mock masks this by advertising the invalid pairing. **Suggested:** inspect the family interfaces for content signing (or land #1974's family-aware validation first), keep timestamping on `SIGNATURE_FORMATTING`, and correct the mock. <sub>flagged by Guided and Copilot CLI</sub>
- `src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java:1048` — **[High] Correctness — question:** Interfaces PR #844 exposes operation-scoped `listFormattingAttributes(operation)`, while profile save still calls the legacy flat `SignatureFormattingApiClient`. The PR description assigns aggregation to #1974. **What is the intended sequencing here:** must #1974 land before content-signing profile save is expected to work with conforming connectors, or should this PR switch to and aggregate the operation-scoped schemas now? **Suggested:** either make that prerequisite explicit/enforced or port the reachable-operation aggregation here. <sub>flagged by Copilot CLI</sub>

src/main/java/com/otilm/core/signing/contentsigning/formatting/ContentSigningFormattingClient.java
ivosh: Fixed in 7728be4c.

`call` now catches `IllegalStateException` alongside `ConnectorException`, so both land on a step-named `CONNECTOR_FAULT`. The API client is resolved outside the guard so only the connector operation is covered.

Added `anEmptyBodyFromTheRestTransportBecomesAStepNamedEngineFailure` to `ContentSigningFormattingClientTest`, pinning the failure, the step name and the wire-safe client message.

No leak risk on the new path: `e.getMessage()` goes to `operatorMessage`, never to `getMessage()`.

I did not normalize interfaces #844 — that stays open as the tidier fix if you want it. Note the same gap exists on `TimestampingConnectorSignatureFormattingClient`, where the flat `SignatureFormattingApiClient` can instead return a null body and NPE at `.getDtbs()`. Different failure mode, left alone here.

src/test/java/com/otilm/core/util/mocks/SignerConnectorMock.java
```diff
@@ -5,14 +5,14 @@
 import java.util.List;
 
 /**
- * Mock of a V2 delegated-signer connector — stubs {@code GET /v2/info} advertising {@link ConnectorInterface#SIGNING}.
- * Used exclusively to back delegated signing profiles.
+ * Mock of a V2 signature-formatting connector — stubs {@code GET /v2/info} advertising
+ * {@link ConnectorInterface#SIGNATURE_FORMATTING}.
  */
 public class SignerConnectorMock extends BaseConnectorMock {
```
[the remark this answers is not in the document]
ivosh: Javadoc corrected in 7728be4c; the advertised interface is unchanged.

`ConnectorInterface.SIGNING` is removed by OmniTrustILM/interfaces#843, which merges before this PR, so restoring it would not compile.

The concern does not apply either: a delegated-signing scheme binds a connector without inspecting its interfaces (`SigningProfileServiceImpl`, `DelegatedSigningRequestDto` branch). Registration does demand at least one functional interface (`ConnectorV2Adapter:114`), so the mock has to advertise something — `SIGNATURE_FORMATTING` serves.

What was fair is that the Javadoc described it as a formatting-connector mock. It now says what it backs.

lubomirw (approved):

Issue #2000, opened 2026-08-09
JSON columns are serialized by a mapper nobody states, and it writes dates as numbers


ivosh:
### Description

**Hibernate serializes every `@JdbcTypeCode(SqlTypes.JSON)` column through a `FormatMapper` that nothing in this codebase configures, and its default writes `java.time` values as numbers and arrays instead of text.**

`MappingSettings.JSON_FORMAT_MAPPER` is unset, so Hibernate builds its own mapper. That mapper leaves `SerializationFeature.WRITE_DATES_AS_TIMESTAMPS` on.

A date written through it is stored as `1768469400.000000000`, `[18, 0, 5]` or `[2026, 3, 29]`.

A customizer that sets the right mapper already exists in the codebase. It is registered for tests only. Production has never used it, so the golden tests assert a shape production does not write.

### When it breaks

`AttributeContentItem.json` is declared as the abstract `AttributeContent`. Jackson writes a root value using its declared type, so the `@JsonFormat(shape = STRING)` carried by `DateTimeAttributeContentV3` and its siblings is bypassed and the mapper's date default applies.

Every custom attribute of content type DATE, TIME or DATETIME is therefore stored in an unusable form:

- The search SQL casts the extracted text to `time` or `date`, and the cast fails.
  `ERROR: invalid input syntax for type time: "[18, 0, 5]"`
- A DATETIME value additionally fails to deserialize (`DateTimeParseException: Text '1.545836313213E9' could not be parsed at index 0`), which makes the object carrying the attribute unopenable rather than merely unfilterable.

### When it is fine

Every other attribute column nests its content below the root. A nested value is written using its runtime class, so `@JsonFormat` does apply and those columns were always stored as text. `AttributeDefinition.definition`, both `meta` columns and the four `List<RequestAttribute>` columns are unaffected.

`audit_log.log_record.additionalData` is an open `Map<String, Object>` filled by `AuditLogAspect` with audited method parameters, so a date-typed parameter is stored as an epoch number. Nothing casts that column in SQL, so those rows are cosmetically wrong rather than broken.

Four columns never reach this mapper at all. `Approval.objectData`, `ScheduledJob.objectData`, `ConditionItem.value` and `ExecutionItem.data` also carry `ObjectToJsonConverter`. Hibernate resolves the JPA converter, which makes the relational type `String`, and a `String` reaches the driver verbatim.

### Scope

[...]

PR #2038, opened 2026-08-13
Centralize application ObjectMapper recipes in one factory

### LINKED ISSUE #2014: Centralize the 26 new ObjectMapper() sites behind one injected mapper
### Description

`src/main/java` constructs `new ObjectMapper()` in **26 places**. Each bare instance carries
Jackson's own defaults rather than the configuration the application actually uses, so Jackson 3
will change the behaviour of all 26 independently and invisibly. Consolidating them behind one
injected mapper (or one factory) turns 26 uncontrolled surfaces into one reviewed one.

The production wire mapper is built by `WebAppConfig#jsonObjectMapper`, whose customizations all
affect on-wire bytes: `WRITE_DATES_AS_TIMESTAMPS` disabled, `JavaTimeModule` registered,
`JsonInclude.Include.NON_NULL`, and `FAIL_ON_MISSING_EXTERNAL_TYPE_ID_PROPERTY` disabled. A bare
`new ObjectMapper()` agrees with none of that. Consolidation therefore **changes serialized
output** at some of the 26 sites - which is the point, but it has to be visible.

**Blocked by #1992.** The golden-file baseline must be captured first. #1992 states the
constraint directly: if centralization lands first, the goldens bake in whatever drift
consolidation introduced and the baseline is worthless. With the goldens in place, every byte
this issue changes shows up as a failing golden that must be explained.

A further 48 `new ObjectMapper()` sites exist in `src/test/java`. Those are out of scope here;
the test harness is handled during the test-infrastructure migration.

Part of the preparation list on #1941.

**Estimate (agent execution - review/verify/supervise):** 1.0 md
**Human baseline:** 3 md

### Definition of Done

- [ ] All 26 `new ObjectMapper()` sites in `src/main/java` are either replaced by the injected
      mapper or documented in this issue as deliberately independent, with the reason.
- [ ] `AcmeJsonProcessor`'s bespoke mapper is treated as a deliberate exception, or folded in
      with the wire-format impact assessed against its RFC 8555 goldens.
- [ ] Every golden-file diff produced by this change is explained and accepted in the PR, not
      silently regenerated.
- [ ] The golden suite from #1992 passes after regeneration, and the regeneration is a separate,
      reviewable commit.
- [ ] `mvn -B verify` passes; Spotless / Checkstyle gates pass; Sonar clean and coverage
      >= 80 % on changed lines.

ivosh:
Replaces every bare `new ObjectMapper()` in `src/main/java` with a recipe from one reviewed factory, so the Jackson 3 migration has a single place to change instead of 26 independent ones.

### 26 sites, 6 configurations, 4 recipes

| Recipe | Sites | Why |
|---|---|---|
| `wire()` | HTTP bodies, plus the 4 converged below | the existing `WebAppConfig` recipe |
| `auditLog()` | `LoggerWrapper` | the wire recipe without null suppression — an audit line records that a member was absent |
| `storage()` | persisted JSON | Jackson's defaults, preserved |
| `lenientStorage()` | payloads produced elsewhere | `storage()` + the two leniency flags those sites already set |

`WebAppConfig#jsonObjectMapper` now delegates to `wire()`, so the bean and every static caller are the same recipe by construction rather than by agreement.

### What changed behaviour, and what deliberately did not

Converged onto the wire mapper — four sites whose JSON never reaches the database:

- **`AuditLogServiceImpl`** already built the wire recipe by hand; it injects the bean now. Byte-identical.
- **`ActionsListener`**, **`DiscoveryCertificateTask`** convert in-process payloads and *gain the time module they were missing*, turning a latent failure on any `java.time` member into a working one.
- **`AuthHelper`** reads a user profile from a `static` method, so it takes the recipe rather than the bean.

Persisted JSON stayed on `storage()` deliberately. No migration guards those columns, so a shape change there splits a table into rows written before it and rows written after — the failure mode #2000 documents. That risk is not worth uniformity.

### Two deliberate exceptions, per the DoD

- **`AcmeJsonProcessor`** stays independent. `AcmeProtocolGoldenTest#inboundJwsEnvelopeParserRejectsUnknownMembers` pins it as strict on unknown members; the wire mapper is lenient. Folding it in would silently relax an RFC 8555 parser. This answers the DoD's open question.
- **`CertificateUtil`** keeps `ORDER_MAP_ENTRIES_BY_KEYS` so a re-written subject-alternative-name column stays byte-stable.

Both now source their base from the factory with the delta stated in Javadoc, so they are one line to review rather than a hidden divergence.

### Three things this turned up

- **`SerializationUtil` reconfigured its shared static mapper on every `deserialize` and `convertValue` call**, mutating it under concurrent callers. A data race, independent of Jackson 3. The flag belongs to the recipe now.
- **`SettingServiceImpl` held an injected mapper *and* a bespoke one, and wrote settings into the same column with both.** Renamed to `wireMapper` / `storageMapper` so the divergence is legible. Not folded together — that would change stored settings.
- **The two Flyway migrations were safe to edit** only because both are already `isAltered` in `DatabaseMigration.JavaMigrationChecksums`. Had they not been, editing the source would have failed `JavaMigrationChecksumTest` and refused startup on every deployment that already ran them.

### Golden files

**No golden was regenerated, because none changed.** All 78 pass untouched. That is the evidence that no pinned surface moved, and it is why there is no separate regeneration commit — the DoD asks for one only if there are diffs to explain.

### The guard

`ObjectMapperCentralizationArchTest` forbids `new ObjectMapper()` outside the factory. Without it the 26 sites regrow one commit at a time and the next one is invisible in review. Its import is scoped to this module's `target/classes`, since `interfaces` shares the package prefix.

### DIFF
--- a/src/main/java/com/otilm/core/serialization/ObjectMapperFactory.java
+++ b/src/main/java/com/otilm/core/serialization/ObjectMapperFactory.java
@@ -1,0 +1,121 @@
+package com.otilm.core.serialization;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.MapperFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.databind.json.JsonMapper;
+import com.fasterxml.jackson.dataformat.csv.CsvGenerator;
+import com.fasterxml.jackson.dataformat.csv.CsvMapper;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder;
+
+/**
+ * Factory for every kind of {@link ObjectMapper} in the platform, centralized at one place.
+ */
+public final class ObjectMapperFactory {
+
+    private ObjectMapperFactory() {
+    }
+
+    /**
+     * The wire recipe: HTTP request and response bodies, and anything shaped to match them. Exposed as the
+     * {@code jacksonObjectMapper} bean, which callers inside the container should inject rather than call this.
+     */
+    public static ObjectMapper wire() {
+        return Jackson2ObjectMapperBuilder
+                .json()
+                .featuresToDisable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS,
+                        DeserializationFeature.FAIL_ON_MISSING_EXTERNAL_TYPE_ID_PROPERTY)
+                .modules(new JavaTimeModule())
+                .serializationInclusion(JsonInclude.Include.NON_NULL)
+                .build();
+    }
+
+    /**
+     * The audit-log recipe, pinned as the audit subsystem has always built it rather than derived from {@link #wire()}.
+     * Null members survive, because an audit line records that a member was absent.
+     */
+    public static ObjectMapper auditLog() {
+        ObjectMapper mapper = new ObjectMapper();
+        mapper.findAndRegisterModules();
+        mapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
+        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        return mapper;
+    }
+
+    /**
+     * Jackson's own defaults, for JSON that is persisted. No migration guards those columns, so a shape change here
+     * splits a table into rows written before it and rows written after.
+     */
+    public static ObjectMapper storage() {
+        return new ObjectMapper();
+    }
+
+    /**
+     * {@link #storage()} for payloads produced elsewhere, where an unrecognized member is expected rather than
+     * exceptional.
+     */
+    public static ObjectMapper lenientStorage() {
+        return storage()
+                .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+                .configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
+    }
+
+    /**
+     * The ACME-request recipe. An unrecognised member is rejected, because RFC 8555 request bodies are exact.
+     */
+    public static ObjectMapper acmeRequest() {
+        return JsonMapper.builder().enable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES).build();
+    }
+
+    /**
+     * The attribute-content recipe. Unknown members are tolerated, because a definition may be written by an older
+     * connector version than the one reading it.
+     */
+    public static ObjectMapper attributeContent() {
+        return JsonMapper
+                .builder()
+                .findAndAddModules()
+                .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+                .build();
+    }
+
+    /**
+     * The secret-thumbprint recipe. The ordering settings are load-bearing: the serialized bytes are hashed, so a
+     * different key order would rewrite every stored thumbprint.
+     */
+    public static ObjectMapper secretThumbprint() {
+        return JsonMapper
+                .builder()
+                .findAndAddModules()
+                .configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true)
+                .configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true)
+                .build();
+    }
+
+    /**
+     * The JMS-message recipe. It takes the container's builder as a parameter, which a static context cannot reach, so
+     * that Spring Boot's Jackson autoconfiguration still applies.
+     */
+    public static ObjectMapper jmsMessage(Jackson2ObjectMapperBuilder builder) {
+        return builder
+                .createXmlMapper(false)
+                .featuresToDisable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
+                .build()
+                .findAndRegisterModules();
+    }
+
+    /**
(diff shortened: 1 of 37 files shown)


src/main/java/com/otilm/core/serialization/ObjectMapperFactory.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * {@link #storage()} for payloads produced elsewhere, where an unrecognized member is expected rather than
+     * exceptional.
+     */
+    public static ObjectMapper lenientStorage() {
+        return storage()
+                .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+                .configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
```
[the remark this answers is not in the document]
ivosh: Fixed in 9769a84b. Split the recipe: `lenientStorage()` is now unknown-members-only and `emptyBeanTolerantStorage()` adds `FAIL_ON_EMPTY_BEANS`. Only `SerializationUtil` and `CryptographicHelper` had that leniency on main, so they use the new recipe; `DatabaseAttributeMigration`, `ExecutionItem`, `CallbackServiceImpl` and `AuthenticationServiceException` keep unknown-only as before. Note `convertValue` serializes internally, so `ExecutionItem` and `CallbackServiceImpl` were affected too.

src/main/java/com/otilm/core/util/DatabaseAttributeMigration.java
```diff
@@ -25,8 +25,7 @@
 public class DatabaseAttributeMigration {
 
     private static final Logger logger = LoggerFactory.getLogger(DatabaseAttributeMigration.class);
-    private static final ObjectMapper ATTRIBUTES_OBJECT_MAPPER = new ObjectMapper()
-            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+    private static final ObjectMapper ATTRIBUTES_OBJECT_MAPPER = ObjectMapperFactory.lenientStorage();
```
lubomirw: **[High]** | Correctness

Could we keep this historical migration's mapper pinned instead of delegating to `ObjectMapperFactory.lenientStorage()`? The architecture rule says migrations must preserve the recipe they originally ran with; if the factory changes for Jackson 3, fresh installs could serialize this migration differently from upgraded installs. If that drift is intentionally acceptable, this may be safe as-is with the constraint documented; otherwise the migration should retain its local recipe.

_Suggested:_ Restore the local mapper recipe and exempt `DatabaseAttributeMigration` from the centralization rule, or document and test why future factory changes cannot affect this migration.

src/main/java/com/otilm/core/service/impl/AuditLogServiceImpl.java
```diff
@@ -151,13 +148,13 @@ public ExportResultDto exportAuditLogs(final List<SearchFilterRequestDto> filter
             builder.message(a.getMessage());
 
             try {
-                builder.operationData(MAPPER.writeValueAsString(a.getLogRecord().operationData()));
+                builder.operationData(objectMapper.writeValueAsString(a.getLogRecord().operationData()));
```
lubomirw: **[Medium]** | Correctness

Could we confirm that audit export no longer needs the discovered JDK8 modules before switching to the wire mapper? The prior mapper registered `Jdk8Module`, `ParameterNamesModule`, and the Java-time module, while the wire mapper registers only Java time; an `Optional` in operation or additional data now falls into `ERROR_SERIALIZATION`. Current producers appear to unwrap or avoid `Optional`, so this can stay as-is if those types are deliberately outside the supported audit-data contract.

_Suggested:_ If the broader behavior must remain supported, add an audit-export factory recipe preserving module discovery, date formatting, and `NON_NULL`, with a regression test for JDK8 data.

<sub>flagged by Superpowers code review</sub>

lubomirw (commented):
Automated multi-reviewer triage - 2 findings (COMMENT).

src/main/java/com/otilm/core/util/DatabaseAttributeMigration.java
ivosh: Fixed in 72693eda. Restored the local pinned recipe (`new ObjectMapper().configure(FAIL_ON_UNKNOWN_PROPERTIES, false)`) and made the exemption explicit in `ObjectMapperCentralizationArchTest`. You're right that this is migration code: its only caller is `V202402171510__UpdateAndOptimizeAttributesModelMigration`, so it sits outside `db.migration` purely for size and the package-based exemption missed it. It now carries the same javadoc rationale as the three migrations pinned earlier in this PR.

src/main/java/com/otilm/core/service/impl/AuditLogServiceImpl.java
ivosh: Fixed in 72693eda. Confirmed the concern is real rather than theoretical, so I preserved the old behaviour instead of accepting the narrowing: added `ObjectMapperFactory.auditLogExport()` pinning `findAndRegisterModules()` + `WRITE_DATES_AS_TIMESTAMPS` off + `NON_NULL`, and `AuditLogServiceImpl` uses it again as a static field.

Two things decided it. `jackson-datatype-jdk8` is a compile dependency, and `Jackson2ObjectMapperBuilder.modules(...)` *replaces* well-known-module discovery, so `wire()` genuinely loses it. And the payload types are open — `LogRecord.operationData()` is a bare `Serializable`, `additionalData()` a `Map<String, Object>` — so any producer can put a JDK8 value there.

Couldn't reuse `auditLog()` — it has `FAIL_ON_UNKNOWN_PROPERTIES` where the export needs `NON_NULL`. Added `ObjectMapperFactoryTest` per your suggestion, covering `Optional`/`OptionalInt`, date-as-text, null-dropping, and a guard asserting `wire()` still fails on `Optional` so the reason this recipe exists can't silently disappear. Plain JUnit, no new Spring context.

lubomirw (approved):

PR #2064, opened 2026-08-14
Discovery v2 schema groundwork

### LINKED ISSUE #1959: Discovery v2 schema groundwork

lubomirw:
Closes #1959. Epic OmniTrustILM/ilm#267 — core plan Tasks 1–2, the schema everything after it builds on.

## Commit 1 — mechanical rename, behavior-frozen

`discovery_history` -> `discovery`: the table holds the live run, not an archive. The migration renames the only three live named objects — the table, `discovery_history_pkey`, and `discovery_certificate_to_discovery_history_key` — enumerated across every migration since init: the old `discovery_id_seq` died in 2022 and `discovery_uuid_unique` carries no stale name. The two frozen Flyway Java migrations keep operating on `discovery_history` deliberately — that is the table'''s name at their point in history. Entity, metamodel references, JPQL entity names in `DiscoveryRepository` and camel-case members follow in code; the repository was already `DiscoveryRepository`. Full suite green on this commit alone.

## Commit 2 — v2 columns, discovery_work, discovery_item

- `discovery` gains the run-level v2 columns: connector-interface link (NULL = v1 legacy), `run_meta`, `resources`, the `last_applied_sequence` drain cursor, progress columns, `run_messages`, `stopped_at`, `connector_state`.
- `discovery_work`: the per-run agenda the sweep ticks — one row per run and work type (`STATUS | DRAIN | PROCESS`), `ON CONFLICT` upsert that moves the due time, due-ordered index. `DiscoveryWorkRepository` mirrors `CertificateStatusPollRepository`: ordered due query with no row locking (the sweep'''s advisory lock provides exclusion), native upsert, `reschedule`, per-run delete.
- `discovery_item`: generic staging with dense `sequence`, `(discovery, resource, unique_ref)` dedupe, jsonb payload kept deliberately untyped — staging must never fail on a payload shape this Core does not know yet — and the four columns mirroring `discovery_certificate` (`processed_error`, `inventory_uuid`, `newly_discovered`, `meta`) so the union listing later reads one shape from both stores. `DiscoveryItemRepository` stays deliberately bare; the tasks that need queries bring them.

Both new entities carry read-only `@ManyToOne` associations with `@OnDelete(CASCADE)` mirroring the migration'''s foreign keys: the test schema is generated from entities (`ddl-auto: create-drop`), not Flyway, so a SQL-only cascade would be untestable — the `ScheduledJobHistory` precedent, same comment.

## Verification

- Unit profile: **4846 tests, 0 failures** — including the six new repository ITests, which ran locally against Testcontainers Postgres: upsert idempotency, due ordering and cascade delete for the agenda; jsonb round-trip (persistence context cleared so the columns are actually read), dedupe violation and cascade delete for staging.
- CI integration profiles re-prove the ITests against the Flyway-built schema.

### DIFF
--- a/src/main/java/com/otilm/core/dao/entity/DiscoveryItem.java
+++ b/src/main/java/com/otilm/core/dao/entity/DiscoveryItem.java
@@ -44,7 +44,7 @@ public class DiscoveryItem extends UniquelyIdentified {
     private UUID discoveryUuid;
 
     // Mirrors ON DELETE CASCADE from the migration for the test environment, which generates its schema from
-    // the entities (same note as ScheduledJobHistory); the writable column stays the scalar discoveryUuid above.
+    // the entities; the writable column stays the scalar discoveryUuid above.
     @OnDelete(action = OnDeleteAction.CASCADE)
     @ManyToOne(fetch = FetchType.LAZY)
     @JoinColumn(name = "discovery_uuid", insertable = false, updatable = false)
@@ -71,16 +71,14 @@ public class DiscoveryItem extends UniquelyIdentified {
     @Column(name = "processed_at")
     private OffsetDateTime processedAt;
 
-    // Per-item processing failure (parity: discovery_certificate.processed_error).
     @Column(name = "processed_error")
     private String processedError;
 
-    // Object this item became (parity: discovery_certificate.inventory_uuid);
-    // null until processed, permanently null if processing failed.
+    // Object this item became; null until processed, permanently null if processing failed.
     @Column(name = "inventory_uuid")
     private UUID inventoryUuid;
 
-    // Not already in inventory when staged, matched by fingerprint (parity: discovery_certificate.newly_discovered).
+    // Not already in inventory when staged, matched by fingerprint.
     @Column(name = "newly_discovered", nullable = false)
     private boolean newlyDiscovered;
 
--- a/src/main/java/com/otilm/core/dao/entity/DiscoveryItem.java
+++ b/src/main/java/com/otilm/core/dao/entity/DiscoveryItem.java
@@ -1,0 +1,104 @@
+package com.otilm.core.dao.entity;
+
+import com.otilm.api.model.common.attribute.common.MetadataAttribute;
+import com.otilm.api.model.core.auth.Resource;
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.EnumType;
+import jakarta.persistence.Enumerated;
+import jakarta.persistence.FetchType;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.ManyToOne;
+import jakarta.persistence.Table;
+import jakarta.persistence.UniqueConstraint;
+import java.time.OffsetDateTime;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import lombok.Getter;
+import lombok.Setter;
+import org.hibernate.annotations.JdbcTypeCode;
+import org.hibernate.annotations.OnDelete;
+import org.hibernate.annotations.OnDeleteAction;
+import org.hibernate.type.SqlTypes;
+
+/**
+ * One staged discovery v2 item, whatever its resource type — the resource-agnostic sibling of
+ * {@link DiscoveryCertificate}, whose v1 rows keep living beside these until the evidence-gated unification
+ * (core#2027). The parity columns ({@code processedError}, {@code inventoryUuid}, {@code newlyDiscovered},
+ * {@code meta}) deliberately mirror that entity so the union listing reads one shape from both stores.
+ *
+ * <p>
+ * {@code payload} keeps the connector-reported union untyped ({@code Map}): staging must never fail on a payload shape
+ * this Core version does not know yet, and the typed view is synthesized at read time where a failure can be answered,
+ * not dropped.
+ */
+@Getter
+@Setter
+@Entity
+@Table(name = "discovery_item", uniqueConstraints = @UniqueConstraint(name = "uq_discovery_item_ref",
+        columnNames = {"discovery_uuid", "resource", "unique_ref"}))
+public class DiscoveryItem extends UniquelyIdentified {
+
+    @Column(name = "discovery_uuid", nullable = false)
+    private UUID discoveryUuid;
+
+    // Mirrors ON DELETE CASCADE from the migration for the test environment, which generates its schema from
+    // the entities (same note as ScheduledJobHistory); the writable column stays the scalar discoveryUuid above.
+    @OnDelete(action = OnDeleteAction.CASCADE)
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "discovery_uuid", insertable = false, updatable = false)
+    private Discovery discovery;
+
+    @Enumerated(EnumType.STRING)
+    @Column(name = "resource", nullable = false)
+    private Resource resource;
+
+    // The run-wide dense cursor assigned by the connector, starting at 1 — never page-scoped.
+    @Column(name = "sequence", nullable = false)
+    private long sequence;
+
+    @Column(name = "unique_ref", nullable = false)
+    private String uniqueRef;
+
+    @Column(name = "payload", columnDefinition = "jsonb", nullable = false)
+    @JdbcTypeCode(SqlTypes.JSON)
+    private Map<String, Object> payload;
+
+    @Column(name = "discovered_at")
+    private OffsetDateTime discoveredAt;
+
+    @Column(name = "processed_at")
+    private OffsetDateTime processedAt;
+
+    // Per-item processing failure (parity: discovery_certificate.processed_error).
+    @Column(name = "processed_error")
+    private String processedError;
+
+    // Object this item became (parity: discovery_certificate.inventory_uuid);
+    // null until processed, permanently null if processing failed.
+    @Column(name = "inventory_uuid")
+    private UUID inventoryUuid;
+
+    // Not already in inventory when staged, matched by fingerprint (parity: discovery_certificate.newly_discovered).
+    @Column(name = "newly_discovered", nullable = false)
+    private boolean newlyDiscovered;
+
+    // Provider-reported location context; unrecoverable if dropped at staging — key location lives here.
+    @Column(name = "meta", columnDefinition = "jsonb")
+    @JdbcTypeCode(SqlTypes.JSON)
+    private List<MetadataAttribute> meta;
+
+    // No-op override required by Sonar S2160 (a field-adding subclass of UniquelyIdentified must override
+    // equals): identity stays the UUID and the added columns deliberately do not affect equality. Matches the
+    // convention used by the other field-adding entities; dropping it just re-raises the finding.
+    @Override
+    public boolean equals(Object o) {
+        return super.equals(o);
+    }
+
+    @Override
+    public int hashCode() {
+        return super.hashCode();
+    }
+}
--- a/src/main/java/com/otilm/core/dao/entity/DiscoveryItem.java
+++ b/src/main/java/com/otilm/core/dao/entity/DiscoveryItem.java
@@ -23,10 +23,8 @@ import org.hibernate.annotations.OnDeleteAction;
 import org.hibernate.type.SqlTypes;
 
 /**
- * One staged discovery v2 item, whatever its resource type — the resource-agnostic sibling of
- * {@link DiscoveryCertificate}, whose v1 rows keep living beside these until the evidence-gated unification
- * (core#2027). The parity columns ({@code processedError}, {@code inventoryUuid}, {@code newlyDiscovered},
- * {@code meta}) deliberately mirror that entity so the union listing reads one shape from both stores.
+ * One staged discovery item, whatever its resource type: the run-scoped record of something a Discovery Provider
+ * reported, held until processing turns it into an inventory object — or records, on the row itself, why it could not.
  *
  * <p>
  * {@code payload} keeps the connector-reported union untyped ({@code Map}): staging must never fail on a payload shape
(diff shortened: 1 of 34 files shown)


src/main/java/com/otilm/core/dao/repository/DiscoveryWorkRepository.java
```diff
@@ -0,0 +1,46 @@
...
+     * {@code (discovery_uuid, work_type)} — unlike an exists-check-then-insert, a concurrent loser is a clean due-time
+     * update (no constraint violation, no aborted transaction).
+     */
+    @Modifying
+    @Query(value = """
+            INSERT INTO {h-schema}discovery_work (uuid, discovery_uuid, work_type, attempt, next_due_at)
+            VALUES (:uuid, :discoveryUuid, :workType, 0, :nextDueAt)
+            ON CONFLICT (discovery_uuid, work_type) DO UPDATE SET next_due_at = EXCLUDED.next_due_at
+            """, nativeQuery = true)
+    void schedule(@Param("uuid") UUID uuid, @Param("discoveryUuid") UUID discoveryUuid,
```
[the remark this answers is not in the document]
lubomirw: Fixed in eec8f8609 — the sharper of your two options: reschedule() now targets the natural key (discoveryUuid, workType), so a caller-retained row uuid can never silently address nothing; that also matches how CertificateStatusPollRepository addresses rows. The schedule() javadoc now states the losing insert's uuid is discarded on conflict, and the conflict arm additionally resets the backoff counter (re-arm = fresh start), with new ITest cases pinning both behaviors.

src/test/java/com/otilm/core/integration/repository/DiscoveryRepositoryITest.java
```diff
@@ -0,0 +1,74 @@
...
+        keyProgress.setProcessed(3L);
+
+        Discovery run = new Discovery();
+        run.setName("nightly-scan");
+        run.setKind("IP-HostName");
+        run.setStatus(DiscoveryStatus.IN_PROGRESS);
+        run.setConnectorStatus(DiscoveryStatus.IN_PROGRESS);
+        run.setConnectorUuid(UUID.randomUUID());
+        run.setConnectorName("network-discovery");
+        run.setConnectorInterfaceUuid(UUID.randomUUID());
```
ivosh: **~ Minor** | Testing

DiscoveryRepositoryITest.java:49 and :53-56 populate `connectorInterfaceUuid`, `progressProcessed`, `progressTotalEstimate`, and `progressPhase`, but assertions at :65-72 never verify them; `stoppedAt` is only checked for non-nullness. A wrong entity or migration mapping for these columns could leave `v2RunColumnsRoundTrip` green. Assert each exact value after clearing the persistence context. This need not block the PR.

src/main/java/com/otilm/core/dao/repository/DiscoveryWorkRepository.java
```diff
@@ -0,0 +1,52 @@
...
+public interface DiscoveryWorkRepository extends JpaRepository<DiscoveryWork, UUID> {
+
+    /**
+     * Returns rows due for a tick ({@code next_due_at} reached), soonest-due first. No row locking: cross-node mutual
+     * exclusion is provided by the cluster-wide advisory lock held by the discovery sweep, so a plain ordered read is
+     * enough.
+     */
+    List<DiscoveryWork> findByNextDueAtLessThanEqualOrderByNextDueAt(OffsetDateTime cutoff, Pageable pageable);
+
+    /**
```
ivosh, line 25: **~ Minor** | Documentation

DiscoveryWorkRepository.java:25-30 combines scheduling semantics, backoff ownership, concurrency behavior, and UUID identity in one paragraph; DiscoveryWorkRepositoryITest.java:24-27 and DiscoveryRepositoryITest.java:24-27 use similarly dense one-sentence inventories. Use a one-line summary followed by short `<p><b>Concurrency:</b>` and `<p><b>Identity:</b>` sections, leaving neighbour details on their own declarations. This need not block the PR.

src/main/resources/db/migration/V202608141000__rename_discovery_history_to_discovery.sql
```diff
@@ -0,0 +1,9 @@
+-- Mechanical rename: the table holds the live discovery run, not an archive of finished ones,
```
ivosh: **. Nitpick** | Documentation

V202608141000__rename_discovery_history_to_discovery.sql:1 labels the edit “Mechanical” and ends with “No behavior change,” while the ALTER statements already show that history. Keep only the durable rationale that the table represents a live run rather than an archive; this need not block the PR.

src/test/java/com/otilm/core/integration/search/DiscoverySearchITest.java
```diff
@@ -73,12 +73,15 @@ void setAttributeEngine(AttributeEngine attributeEngine) {
         this.attributeEngine = attributeEngine;
     }
 
-    private static final DateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss");
+    /** Parsed in the system zone, preserving the instants the former SimpleDateFormat fixture produced. */
```
ivosh: **. Nitpick** | Documentation

DiscoverySearchITest.java:76 explains the helper by referring to the former SimpleDateFormat fixture, leaving change-history narration in the test. Rewrite it as a durable reason, such as “Interpret local fixture timestamps in the system zone to preserve their intended instants”; this is non-blocking.

src/main/java/com/otilm/core/events/handlers/DiscoveryFinishedEventHandler.java
```diff
@@ -54,26 +54,26 @@ protected EventContext<DiscoveryHistory> prepareContext(EventMessage eventMessag
                     ? DiscoveryStatus.COMPLETED
                     : reportedStatus;
             discovery.setStatus(finalStatus);
-            discovery.setEndTime(new Date());
+            discovery.setEndTime(OffsetDateTime.now(ZoneOffset.UTC));
```
ivosh, line 59: **~ Minor** | Correctness

src/main/java/com/otilm/core/events/handlers/DiscoveryFinishedEventHandler.java:52 reads and conditionally updates the discovery without a pessimistic lock. The event listener runs with concurrency 5, so concurrent PROCESSING/WARNING deliveries can both observe a non-terminal state and the last commit can overwrite the first terminal outcome. Add a `PESSIMISTIC_WRITE` finder to `DiscoveryRepository` and use it here.

Note that this race predates the timestamp rename.

src/main/java/com/otilm/core/dao/entity/DiscoveryItem.java
```diff
@@ -0,0 +1,105 @@
...
+ * {@link DiscoveryCertificate}, whose v1 rows keep living beside these until the evidence-gated unification
+ * (core#2027). The parity columns ({@code processedError}, {@code inventoryUuid}, {@code newlyDiscovered},
+ * {@code meta}) deliberately mirror that entity so the union listing reads one shape from both stores.
+ *
+ * <p>
+ * {@code payload} keeps the connector-reported union untyped ({@code Map}): staging must never fail on a payload shape
+ * this Core version does not know yet, and the typed view is synthesized at read time where a failure can be answered,
+ * not dropped.
+ */
+@Getter
```
ivosh, line 34: **. Nitpick** | Documentation

The class Javadoc is a dense implementation narrative that mixes staging history, sibling-entity comparison, issue tracking, persistence shape, and error handling, making the actual domain contract difficult to scan. The three-line comment above `equals` has the same problem: it explains the Sonar rule, repository history, and what happens if the override is removed instead of documenting the equality contract. Reduce the Javadoc to the enduring purpose and payload invariant, and replace the equality comment with a concise statement such as `Identity remains UUID-based`; this is non-blocking.

src/main/java/com/otilm/core/dao/entity/DiscoveryItem.java
```diff
@@ -0,0 +1,105 @@
...
+import org.hibernate.annotations.OnDelete;
+import org.hibernate.annotations.OnDeleteAction;
+import org.hibernate.type.SqlTypes;
+
+/**
+ * One staged discovery v2 item, whatever its resource type — the resource-agnostic sibling of
+ * {@link DiscoveryCertificate}, whose v1 rows keep living beside these until the evidence-gated unification
+ * (core#2027). The parity columns ({@code processedError}, {@code inventoryUuid}, {@code newlyDiscovered},
+ * {@code meta}) deliberately mirror that entity so the union listing reads one shape from both stores.
+ *
```
ivosh, line 28: **. Nitpick** | Documentation

This Javadoc defines `DiscoveryItem` through `DiscoveryCertificate`’s current storage model and a list of fields that presently mirror it. A later change to that sibling can silently make this accurate-looking contract false without touching this class. Document `DiscoveryItem`’s own persisted shape and invariants directly, and keep the cross-model migration rationale in core#2027; this is non-blocking.

src/main/java/com/otilm/core/dao/entity/DiscoveryItem.java
```diff
@@ -0,0 +1,105 @@
...
+
+    // Provider-reported location context; unrecoverable if dropped at staging — key location lives here.
+    @Column(name = "meta", columnDefinition = "jsonb")
+    @JdbcTypeCode(SqlTypes.JSON)
+    private List<MetadataAttribute> meta;
+
+    // No-op override required by Sonar S2160 (a field-adding subclass of UniquelyIdentified must override
+    // equals): identity stays the UUID and the added columns deliberately do not affect equality. Matches the
+    // convention used by the other field-adding entities; dropping it just re-raises the finding.
+    @Override
```
ivosh: **. Nitpick** | Documentation

The equality comment justifies this class’s behavior through the convention currently used by “other field-adding entities.” If those entities change their equality strategy, this comment rots even though `DiscoveryItem` has not changed; “dropping it just re-raises the finding” also records tooling history rather than a code invariant. State only the local contract—that equality and hashing remain UUID-based despite the added fields; this is non-blocking.

ivosh (changes requested):

src/test/java/com/otilm/core/integration/repository/DiscoveryRepositoryITest.java
lubomirw: Fixed in 2ad88c298 — connectorInterfaceUuid, progressProcessed, progressTotalEstimate and progressPhase asserted by exact value, stoppedAt by exact instant (post-clear, so the columns are genuinely read).

src/main/java/com/otilm/core/dao/repository/DiscoveryWorkRepository.java
lubomirw: Fixed in 2ad88c298 — schedule() reads as a one-line summary plus Concurrency and Identity sections, and both ITest class Javadocs split into one sentence per claim.

src/main/resources/db/migration/V202608141000__rename_discovery_history_to_discovery.sql
lubomirw: Fixed in 2ad88c298 — only the live-run rationale and the constraint-name note remain.

src/test/java/com/otilm/core/integration/search/DiscoverySearchITest.java
lubomirw: Fixed in 2ad88c298 — "Interprets local fixture timestamps in the system zone, preserving their intended instants."

src/main/java/com/otilm/core/events/handlers/DiscoveryFinishedEventHandler.java
lubomirw: Fixed in 2ad88c298 — DiscoveryRepository gains a PESSIMISTIC_WRITE finder (the CertificateRepository precedent, transaction note included) and the handler reads through it, so concurrent deliveries serialize on the row and the non-terminal guard holds. All 25 EventHandlersITest cases green through the locked path. Thanks for flagging it even though it predates the PR.

src/main/java/com/otilm/core/dao/entity/DiscoveryItem.java
lubomirw: Fixed in 2ad88c298 — the class Javadoc states the entity's own purpose and the payload invariant; the equality comment shrinks to the local contract (UUID-based identity, S2160-required no-op).

src/main/java/com/otilm/core/dao/entity/DiscoveryItem.java
lubomirw: Fixed in 2ad88c298 — the sibling-derived definition is gone; the migration comments keep the SQL-side parity notes and core#2027 owns the unification rationale.

src/main/java/com/otilm/core/dao/entity/DiscoveryItem.java
lubomirw: Fixed in 2ad88c298 — both entities now state only the local contract: identity and hashing stay UUID-based, added columns never affect equality.

ivosh (approved):

Issue #2013, opened 2026-08-11
Narrow Jackson-motivated catch (IOException) to catch (JsonProcessingException)


ivosh:
### Description

Jackson 3 makes its exceptions unchecked. `JsonProcessingException` no longer extends
`IOException` in the way Jackson 2 arranged it, so a broad `catch (IOException)` written to
handle a Jackson failure silently stops catching what it was written for. The compiler will not
tell us: the catch block still compiles, it just never runs, and the exception propagates past
the handler that was supposed to deal with it.

`src/main/java` currently has 47 `catch (IOException` sites. Not all of them are Jackson-
motivated - some are genuine stream or file IO. Each one needs to be classified, and the
Jackson-motivated ones narrowed to `catch (JsonProcessingException)` (or the Jackson 3
equivalent, per the delta assessment in OmniTrustILM/dependencies#139).

This is preparation work and must land on Spring Boot 3.5, before the framework bump. Narrowing
the catch on 3.5 is behaviour-preserving and reviewable in isolation. Doing it during the
migration mixes a semantic change into a diff that is already large.

Part of the preparation list on #1941.

### Definition of Done

- [ ] Every `catch (IOException)` in `src/main/java` is classified as Jackson-motivated or
      genuine IO; the classification is recorded in the PR description.
- [ ] Jackson-motivated catches are narrowed; genuine IO catches are left alone.
- [ ] Where a single catch legitimately covers both, it stays broad and a comment says why.
- [ ] No behaviour change on 3.5: the suite is green and the narrowed handlers still fire for
      the inputs they were written for.
- [ ] `mvn -B verify` passes; Spotless / Checkstyle gates pass; Sonar clean and coverage
      >= 80 % on changed lines.

ivosh:
## Classification complete — nothing to narrow

All `catch (IOException)` sites in `core`'s `src/main/java` were classified. The count has drifted
to **48** since this issue was filed (it quotes 47). **47 are genuine IO and one is Jackson-motivated,
but that one legitimately covers both concerns and cannot be split on Jackson 2.**

No code change is warranted, so no PR: the finding is the deliverable.

### The one Jackson-motivated site

`service/impl/ExportProcessor.java:83`. The `try` covers `ObjectWriter.writeValue` (Jackson CSV)
**and** the `OutputStreamWriter` / `ByteArrayOutputStream` / `ZipOutputStream` IO around it.

[...]

PR #2065, opened 2026-08-14
Revert the document signing rename back to content signing

### LINKED ISSUE #2023: Rename CONTENT_SIGNING to DOCUMENT_SIGNING, with the workflow_type migration

ivosh:
Reverts #2024. `CONTENT_SIGNING` stays as the signing workflow type in the domain model, mappers and persisted values.

### How
The rename is undone **semantically on top of current `main`** rather than by reverting the merge commit, so the work merged afterwards is preserved.

Restored names:
- `DocumentSigningWorkflow`, `ManagedDocumentSigningWorkflow`, `DelegatedDocumentSigningWorkflow` -> `ContentSigningWorkflow`, `ManagedContentSigningWorkflow`, `DelegatedContentSigningWorkflow`
- `SigningProfileMapper` / `SigningProfileServiceImpl` workflow handling
- Test builders and mocks (`ContentSigningWorkflowRequestDtoBuilder`, `ContentSigningFormattingMock`)

### Database
`V202608121000__rename_content_signing_to_document_signing.sql` is **kept in place** because it may already have been applied in existing environments; removing it would break Flyway history. A new forward migration `V202608141000__revert_document_signing_to_content_signing.sql` renames the persisted values back in `signing_profile_version.workflow_type`, `signing_profile.workflow_type` and the `connector_interface.features` array.

### Verification
`mvn verify -DskipITs` - 4840 tests pass. One unrelated flake in `NotificationInstanceServiceITest` (connector mock returned `text/html`) passed on re-run in isolation.

### Coordination
Depends on the interfaces revert of #847 (contract change) - merge that first. Pairs with the fe-administrator revert of #2013.

Draft on purpose - please confirm merge ordering.


src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -814,15 +814,15 @@ private void applyWorkflow(SigningProfile p, SigningProfileVersion version, Work
...
-                            "Signature formatting connector is required for document signing workflow");
+                            "Signature formatting connector is required for content signing workflow");
                 }
                 Connector documentConnector = connectorService
                         .getConnectorEntity(SecuredUUID.fromUUID(w.getSignatureFormattingConnectorUuid()));
-                validateFormattingConnectorFeature(documentConnector, FeatureFlag.DOCUMENT_SIGNING,
-                        SigningWorkflowType.DOCUMENT_SIGNING);
+                validateFormattingConnectorFeature(documentConnector, FeatureFlag.CONTENT_SIGNING,
+                        SigningWorkflowType.CONTENT_SIGNING);
                 version.setSignatureFormattingConnector(documentConnector);
```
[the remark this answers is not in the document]
ivosh: Fixed in c061b97c.

src/test/java/com/otilm/core/service/signingprofile/SigningProfileTestBase.java
```diff
@@ -374,14 +374,14 @@ protected SigningProfileRequestDto buildManagedOneTimeKeyRawRequest(String name)
     }
 
     /**
-     * Builds a request using a DELEGATED scheme and DOCUMENT_SIGNING workflow.
+     * Builds a request using a DELEGATED scheme and CONTENT_SIGNING workflow.
      */
     protected SigningProfileRequestDto buildDelegatedDocumentRequest(String name) {
```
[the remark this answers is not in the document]
ivosh: Fixed in c061b97c.

src/test/java/com/otilm/core/integration/service/SigningProfileServiceImplITest.java
```diff
@@ -1924,7 +1923,7 @@ void create_timestamping_requiredAttributeMissing_throwsValidationException() {
                             .withDelegatedSigning(signerConnector.getUuid())
                             .withTimestamping(aTimestampingWorkflow()
                                     .withSignatureFormattingConnector(
-                                            UUID.fromString(documentSigningFormattingConnector.getUuid()))
+                                            UUID.fromString(contentSigningFormattingConnector.getUuid()))
```
lubomirw: **[Low]** | Testing

This timestamping-required-attribute test stubs `timestampingFormattingMock`, but submits `contentSigningFormattingConnector`. It therefore throws during `TIMESTAMPING` feature validation and can pass without exercising the missing-required-attribute path.

_Suggested:_ Use `timestampingFormattingConnector.getUuid()` and assert that the `ValidationException` identifies the missing required attribute.

src/main/resources/db/migration/V202608141000__revert_document_signing_to_content_signing.sql
```diff
@@ -0,0 +1,17 @@
+-- Revert the DOCUMENT_SIGNING signing workflow type back to CONTENT_SIGNING.
```
lubomirw, line 1: **[Nit]** | Maintainability

Could we reevaluate whether this header adds useful context beyond the migration filename and the compensation rationale immediately below it? Would retaining lines 2–3 alone make the migration's purpose equally clear with less repetition?

_Suggested:_ Remove line 1 if the compensation comment is sufficient; otherwise keep it if the descriptive header is an intentional migration convention.

<sub>flagged by Hygiene review</sub>

lubomirw (commented):

src/test/java/com/otilm/core/integration/service/SigningProfileServiceImplITest.java
ivosh: Fixed in 940b6879.

Good catch — the test was passing vacuously. Confirmed by running it: with `timestampingFormattingConnector` submitted it now reaches the `AttributeEngine` required-attribute check instead of failing TIMESTAMPING feature validation. Also renamed the stubbed attribute `req_content_attr` → `req_timestamping_attr` (copy-paste leftover from the sibling content-signing test), and added the assertion you suggested so the exception's origin is pinned rather than just its type.

src/main/resources/db/migration/V202608141000__revert_document_signing_to_content_signing.sql
ivosh: Keeping the header — it's the existing migration convention rather than incidental repetition. The migration this one compensates, `V202608121000__rename_content_signing_to_document_signing.sql`, opens with the identical one-line form, as do other recent migrations (`V202608031100`, `V202608071100`). Keeping the rename/revert pair symmetric helps anyone diffing the two side by side.

lubomirw (approved):

PR #2077, opened 2026-08-14
Clear only populated tables in the per-test database reset

### LINKED ISSUE #2003: Per-test database reset truncates all ~100 tables, costing 20% of integration-worker time
### Description

Every integration test method pays a full database reset. `BaseSpringBootTest.setupAuth()` (`src/test/java/com/otilm/core/util/BaseSpringBootTest.java:60-98`) is a `@BeforeEach` that enumerates every table in schema `core` through JDBC `DatabaseMetaData.getTables`, issues one multi-table `TRUNCATE`, then calls `settingService.refreshCache()`.

The reset is unconditional. There is no dirty tracking, no write detection, and no read-only skip, so it runs in full even when the test wrote nothing. `src/main/java/com/otilm/core/dao/entity` holds 102 `@Entity` classes, so 100+ tables are truncated per test method, and `SettingServiceImpl.refreshCache()` (`src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java:99-105`) then issues at least five queries against the just-emptied schema. Roughly 191 of the 212 `*ITest` classes inherit this path, spanning about 2434 test methods. Only `BaseMessagingIntTest` overrides `setupAuth()` and skips truncation.

Measurements published by #1997 put database reset at **5:54 per workflow, 20.9% of aggregate integration-worker time**. Within each worker, reset time correlates with wall clock almost perfectly:

| Worker | Mean reset share | Wall/reset correlation |
|---|---:|---:|
| integration-core | 15.9% | 0.957 |
| integration-service-1 | 21.9% | 0.965 |
| integration-service-2 | 25.1% | 0.984 |

Shard rebalancing is exhausted as a lever — the integration-worker means now differ by only 5.1% — so the reset path is the largest remaining single cost in the integration workers.

The instrumentation that produced these numbers is **not on `main`**: #1997 was closed unmerged, and `TestCostRecorder`, `TestCostTestExecutionListener`, `src/test/resources/META-INF/spring.factories`, `.github/scripts/summarize_test_costs.py` and `.github/scripts/summarize_test_workers.py` exist only on the `test/balance-ci-workers` branch. Those numbers are the baseline this issue is measured against; landing the instrumentation itself is out of scope here.

Candidate directions, not a prescription:

- Cache the table list per Spring context instead of re-reading JDBC metadata on every test method.
- Skip the `TRUNCATE` when the test wrote nothing.
- Roll back instead of truncating where the test is already transactional.
- Restore from a template schema or snapshot rather than truncating 100+ tables.
- Seed the settings cache once per context instead of re-querying it per test method.

Follows on from #1693.

### Definition of Done

- [ ] Table list resolved from the catalog on every reset, not cached per JVM — `BaseMigrationTest` drops and recreates the schema, so a cached list goes stale mid-run
- [ ] Reset no longer unconditional per table — a table holding no rows is not cleared (per-test skipping applies to under 5% of methods; per-table skipping removes ~95% of the work)
- [ ] Measured database reset share of aggregate integration-worker time reduced from the 20.9% baseline to below 10%
- [ ] Test isolation preserved — no cross-test data leakage; full suite green across repeated runs
- [ ] `CiTestSplitTest`, `ContextSignatureTest`, `ContextSignatureGuardTest` and `TestClassTaxonomyTest` remain green
- [ ] Before/after reset timings recorded in the PR description

ivosh:
**TL;DR: the per-test database reset drops from 104s to 15s across the three integration workers (-85%), by deleting only the tables that actually hold rows instead of truncating all ~100 every time.**

Refs #2003.

## What was slow

Every integration test method ran `TRUNCATE` over every table in the test schema. That is ~100 tables per test method, and there are ~2,500 test methods across the three integration workers.

`TRUNCATE` allocates a fresh relfilenode for each table it names. It does that whether or not the table held a single row. So the reset paid the full price of the schema even though a typical test writes a handful of tables.

Measured on this branch: 95% of the individual table clears were clearing an already-empty table.

## What changed

The scan and the deletes move into a server-side plpgsql function. For each table it probes for rows first, and only issues `DELETE` when the table is populated. Foreign-key triggers are disabled for the transaction (`session_replication_role = replica`) so the tables can be emptied in catalog order without deriving a dependency order.

Two deliberate choices worth calling out:

- **The table list is read from the catalog on every reset, not cached.** `BaseMigrationTest` runs `DROP SCHEMA core CASCADE` in `@AfterAll`, and the 12 migration ITests run the schema forward from older revisions. The table set is not stable within a run, so a cached list goes stale mid-run. Enumeration is only 0.3% of the reset cost anyway.
- **The function lives in `public`, not in `core`.** Same reason: anything stored in `core` is dropped by those migration tests.

## Results

Reset cost measured directly with a counter, identical reset counts in both variants:

| Worker | Resets | Before | After | Change |
|---|---:|---:|---:|---:|
| `integration-core` | 681 | 39.60 s | 5.45 s | -86% |
| `integration-service-1` | 886 | 29.36 s | 5.17 s | -82% |
| `integration-service-2` | 949 | 35.29 s | 4.74 s | -87% |
| **total** | **2516** | **104.25 s** | **15.36 s** | **-85%** |

Wall clock on the measuring machine was too noisy to use as a metric (the same profile varied between 221s and 1583s in one session), so the reset path was instrumented and measured directly.

## When it could break

Two risks were checked and are not real here:

- **Dead-tuple bloat from repeated `DELETE`.** Per-quintile cost over a 681-reset run is flat at 3-5 ms. The old `TRUNCATE` path is the one that degrades over a run: 46.1 ms to 65.9 ms first-to-last quintile.
- **Different end state.** The same tables end up empty. Sequences are left alone, exactly as the previous `TRUNCATE` (without `RESTART IDENTITY`) left them.

The function requires a superuser connection for `session_replication_role`. Testcontainers' `test` user is superuser, so this holds for the test path it is used on.


src/test/java/com/otilm/core/util/TestDatabaseCleaner.java
```diff
@@ -0,0 +1,99 @@
...
+    private static final String CLEAR_CALL = "SELECT public.otilm_clear_test_schema(?)";
+
+    private static final Set<String> INSTALLED_JDBC_URLS = ConcurrentHashMap.newKeySet();
+
+    private TestDatabaseCleaner() {
+    }
+
+    static void clear(DataSource dataSource, String schema) throws SQLException {
+        try (Connection connection = dataSource.getConnection()) {
+            if (INSTALLED_JDBC_URLS.add(Objects.requireNonNullElse(connection.getMetaData().getURL(), UNKNOWN_URL))) {
```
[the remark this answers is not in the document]
ivosh: Fixed in 6b4dccdd.

src/test/java/com/otilm/core/util/BaseSpringBootTest.java
```diff
@@ -55,41 +55,27 @@ public class BaseSpringBootTest {
...
         MDC.clear();
     }
 
-    private void truncateTables() throws SQLException {
-        if (jdbcTemplate.getDataSource() == null) {
+    private void clearTables() throws SQLException {
+        DataSource dataSource = jdbcTemplate.getDataSource();
+        if (dataSource == null) {
             throw new SQLException("JDBCTemplate does not have initialized data source");
         }
```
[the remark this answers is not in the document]
ivosh: Fixed in 6b4dccdd.

lubomirw (commented):
Automated multi-reviewer triage — 6 findings (COMMENT). Sources: Copilot CLI, guided-pr-review, superpowers code review, pr-hygiene; findings deduped and independently verified before posting. All Low/Nit — nothing blocking.

src/test/java/com/otilm/core/util/TestDatabaseCleaner.java
```diff
@@ -0,0 +1,106 @@
...
+ * <p>
+ * Uses {@code DELETE} rather than {@code TRUNCATE} because truncating costs a fresh relfilenode per named table even
+ * when that table is already empty. The table list is read from the catalog on every call, and the function itself
+ * lives in {@code public}, because {@code BaseMigrationTest} drops and recreates the schema mid-run.
+ */
+final class TestDatabaseCleaner {
+
+    private static final String UNDEFINED_FUNCTION_SQL_STATE = "42883";
+
+    /** A schema that does not match this matches no tables, which would silently skip clearing. */
```
lubomirw: **[Low]** | Error handling

The `SCHEMA_NAME` guard exists (per its own javadoc) to stop a schema that matches no tables from silently skipping the clear — but any well-formed wrong value (mixed-case `default_schema` override, typo, dropped schema) passes the regex, matches zero `pg_class` rows, and every test then runs with no per-test cleanup and no error.

_Suggested:_ `RAISE EXCEPTION` in `otilm_clear_test_schema` when `target_schema` has no row in `pg_namespace` (e.g. `IF NOT EXISTS (SELECT 1 FROM pg_namespace WHERE nspname = target_schema) THEN RAISE EXCEPTION 'schema % not found', target_schema; END IF;`), so a misconfigured schema fails the first test loudly.

<sub>flagged by guided-pr-review + superpowers code review</sub>

src/test/java/com/otilm/core/util/TestDatabaseCleaner.java
```diff
@@ -0,0 +1,106 @@
...
+    private static final Set<String> INSTALLED_JDBC_URLS = ConcurrentHashMap.newKeySet();
+
+    private TestDatabaseCleaner() {
+    }
+
+    static void clear(DataSource dataSource, String schema) throws SQLException {
+        if (schema == null || !SCHEMA_NAME.matcher(schema).matches()) {
+            throw new SQLException("Invalid schema name: " + schema);
+        }
+        try (Connection connection = dataSource.getConnection()) {
```
lubomirw, line 81: **[Low]** | Correctness

`clear()` silently assumes the pooled connection has auto-commit enabled (true today by the Hikari default): with `spring.datasource.hikari.auto-commit=false` the function's DELETEs would be rolled back when the pool reclaims the connection, silently disabling the per-test reset — and the 42883 recovery path would fail with 25P02 (aborted transaction) instead of reinstalling.

_Suggested:_ assert or enforce the expectation at the top of `clear()` — e.g. `if (!connection.getAutoCommit()) { connection.setAutoCommit(true); }` — so a future pool-config change fails loudly instead of leaking state between tests.

src/test/java/com/otilm/core/util/TestDatabaseCleaner.java
```diff
@@ -0,0 +1,106 @@
...
+            if (INSTALLED_JDBC_URLS.add(Objects.requireNonNullElse(connection.getMetaData().getURL(), UNKNOWN_URL))) {
```
lubomirw: **[Low]** | Correctness

`INSTALLED_JDBC_URLS.add()` marks the URL installed before `install()` completes; if JUnit parallel execution is ever enabled, a second thread skips the install and races into the function call, and a concurrent `CREATE OR REPLACE FUNCTION` from the retry path can fail with PostgreSQL's "tuple concurrently updated" — not SQLState 42883, so the retry does not heal it. Harmless today (`junit.jupiter.execution.parallel.enabled=false`), but the cleaner will outlive that assumption silently.

_Suggested:_ synchronize the install per URL (e.g. `computeIfAbsent` on a lock map, marking installed only on success), or note in the class javadoc that single-threaded test execution is assumed.

src/test/java/com/otilm/core/util/TestDatabaseCleaner.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private static final String CLEAR_FUNCTION = """
+            CREATE OR REPLACE FUNCTION public.otilm_clear_test_schema(target_schema text) RETURNS integer AS $fn$
+            DECLARE
+                relation text;
+                populated boolean;
+                cleared integer := 0;
+            BEGIN
+                -- Tables are emptied in catalog order rather than dependency order, so disable constraint
+                -- enforcement instead of deriving an order. Requires a superuser connection.
```
lubomirw: **[Nit]** | Documentation

`SET LOCAL session_replication_role = replica` raises the suite's database privilege floor from table ownership (all the old TRUNCATE needed) to superuser, but that is noted only in a comment inside the plpgsql text — invisible to anyone adopting the class or pointing the suite at a non-Testcontainers database, where it surfaces as an unexplained 42501 in every test.

_Suggested:_ state the superuser requirement (and why: FK enforcement is disabled instead of deriving a dependency order) in the `TestDatabaseCleaner` class javadoc — the surface other repos will read if they copy this utility.

src/test/java/com/otilm/core/integration/migration/BaseMigrationTest.java
```diff
@@ -47,9 +46,6 @@ static void migrationDatasource(DynamicPropertyRegistry registry) {
     @Autowired
     private DataSource dataSource;
 
-    @Value("${spring.jpa.properties.hibernate.default_schema:core}")
-    private String dbSchema;
-
     @AfterAll
     void resetSchema() throws SQLException {
         if (!dbSchema.matches("^[a-zA-Z0-9_]+$")) {
```
lubomirw: **[Nit]** | Maintainability

`resetSchema()` keeps its own inline `^[a-zA-Z0-9_]+$` regex to validate the same `dbSchema` value that `TestDatabaseCleaner.SCHEMA_NAME` now validates (with a different exception type), so the two validators of one shared field can drift apart.

_Suggested:_ expose a single package-visible validator (e.g. `TestDatabaseCleaner.requireValidSchemaName(String)`) and use it in both call sites.

<sub>flagged by guided-pr-review + superpowers code review</sub>

src/test/java/com/otilm/core/util/BaseSpringBootTest.java
```diff
@@ -55,41 +55,27 @@ public class BaseSpringBootTest {
...
-                }
-            }
-            if (!tableNames.isEmpty()) {
-                String truncateSql = "TRUNCATE " + String.join(", ", tableNames);
-                try (var statement = connection.prepareStatement(truncateSql)) {
-                    statement.execute();
-                }
-            }
+    private void clearTables() throws SQLException {
+        DataSource dataSource = jdbcTemplate.getDataSource();
```
lubomirw: **[Nit]** | Maintainability

After this change `jdbcTemplate` is used only as a nullable `DataSource` holder, keeping the `getDataSource()` null-check and `SQLException` wrapper alive, while `BaseMigrationTest` already injects `DataSource` directly.

_Suggested:_ replace the `JdbcTemplate` field with `@Autowired private DataSource dataSource;` and call `TestDatabaseCleaner.clear(dataSource, dbSchema)` directly, deleting the `clearTables()` null-check branch.

src/test/java/com/otilm/core/util/TestDatabaseCleaner.java
ivosh: Fixed in 2f88a61a.

src/test/java/com/otilm/core/util/TestDatabaseCleaner.java
ivosh: Fixed in 2f88a61a.

src/test/java/com/otilm/core/util/TestDatabaseCleaner.java
ivosh: Fixed in 2f88a61a.

src/test/java/com/otilm/core/util/TestDatabaseCleaner.java
ivosh: Fixed in 2f88a61a.

src/test/java/com/otilm/core/integration/migration/BaseMigrationTest.java
ivosh: Fixed in 2f88a61a.

src/test/java/com/otilm/core/util/BaseSpringBootTest.java
ivosh: Fixed in 2f88a61a.

lubomirw (approved):

PR #2081, opened 2026-08-15
Persist and validate branding settings under the PLATFORM section

### LINKED ISSUE #2060: Persist and validate branding settings under the PLATFORM section
### Description

Persist, validate and serve UI branding through the existing settings subsystem, as a new category under the PLATFORM section.

Follows the established shape in `SettingServiceImpl` (`getPlatformSettings` / `updatePlatformSettings`), which already maps `PLATFORM_UTILS` and `PLATFORM_CERTIFICATES` by category. Branding is stored as `Setting` rows in the same table; no migration is required because `setting.value` is already mapped `length = Integer.MAX_VALUE`.

**Persistence**

- Read and write the branding category in `SettingServiceImpl`, including the operator `defaultTheme`.
- Wire `getBrandingSettings` / `updateBrandingSettings` through `SettingControllerImpl` with `@AuditLogged` consistent with the neighbouring settings endpoints.
- Authorization: `@ExternalAuthorization(resource = Resource.SETTINGS, action = ResourceAction.LIST)` on the read and `ResourceAction.UPDATE_BRANDING` on the write.
- Reset to default: clearing a branding field removes its `Setting` row so the platform default applies again, per field rather than all-or-nothing.
- `SettingsCache` continues to refresh on the existing `${settings.cache.refresh-interval}` schedule; no change to the cache mechanism.

**Validation**, applied on the update path so nothing malformed reaches the settings table and, from there, every page render:

- Each of the five colours must be a valid hex colour string.
- Logos, per slot: accepted formats PNG and SVG determined by content sniffing rather than filename or declared MIME type; at most 1 MB on the decoded bytes; aspect ratio between 1:1 and 3:1 inclusive.
- Rejections raise `ValidationException` naming the specific rule that failed, so the Appearance form can surface it directly.

SVG sanitization is deliberately a separate issue, so it gets a focused security review.

### User Story

As a platform administrator, I want my branding choices stored globally on the instance and rejected clearly when unusable, so that every user sees them without a redeploy and I can correct mistakes without guessing.

### Use Case

An administrator saves branding. Valid values persist to the `setting` table, survive restart, and are picked up by other replicas within the settings-cache refresh interval. An oversized logo, one at 5:1, or an invalid hex value is refused with a message naming the failing rule.

### Acceptance Criteria

1. [ ] Branding is readable and writable through the platform settings service as a category under the PLATFORM section.
2. [ ] Values persist to the `setting` table and survive a restart, with no database migration introduced.
3. [ ] The read is gated by `Resource.SETTINGS` + `LIST`; the update is gated by `Resource.SETTINGS` + `UPDATE_BRANDING`. A user holding `SETTINGS` + `UPDATE` but not `UPDATE_BRANDING` is refused the update.
4. [ ] Both operations emit audit log entries consistent with the existing settings endpoints.
5. [ ] Clearing an individual branding field removes its stored row so that field falls back to its platform default independently.
6. [ ] Each of the five colours is validated as a hex colour; an invalid value is rejected naming the field.
7. [ ] Logo format is determined by content sniffing; a file whose content is neither PNG nor SVG is rejected regardless of its filename or declared MIME type.
8. [ ] A logo whose decoded size exceeds 1 MB is rejected.
9. [ ] A logo whose aspect ratio falls outside 1:1 to 3:1 inclusive is rejected. Ratio is determined for PNG from the image header and for SVG from its declared dimensions or viewBox.
10. [ ] Each rejection names the specific rule that failed, and unit tests cover every rule at its boundary, including an SVG with neither width/height nor viewBox.
11. [ ] The existing `/v1/settings` platform GET and PUT keep their current shape and behaviour; existing settings tests pass unchanged.
12. [ ] Sonar quality gate passes and new code meets at least 80% coverage.

Jakub-Bartal-01:
Closes #2060

### DIFF
--- a/src/main/java/com/otilm/core/api/web/SettingControllerImpl.java
+++ b/src/main/java/com/otilm/core/api/web/SettingControllerImpl.java
@@ -3,8 +3,11 @@ package com.otilm.core.api.web;
 import com.otilm.api.exception.NotFoundException;
 import com.otilm.api.interfaces.core.web.SettingController;
 import com.otilm.api.model.core.auth.Resource;
+import com.otilm.api.model.core.logging.Sensitive;
 import com.otilm.api.model.core.logging.enums.Module;
 import com.otilm.api.model.core.logging.enums.Operation;
+import com.otilm.api.model.core.settings.BrandingSettingsDto;
+import com.otilm.api.model.core.settings.BrandingSettingsUpdateDto;
 import com.otilm.api.model.core.settings.EventSettingsDto;
 import com.otilm.api.model.core.settings.EventsSettingsDto;
 import com.otilm.api.model.core.settings.PlatformSettingsDto;
@@ -49,6 +52,23 @@ public class SettingControllerImpl implements SettingController {
     @AuditLogged(module = Module.CORE, resource = Resource.SETTINGS, operation = Operation.UPDATE, name = "platform")
     public void updatePlatformSettings(PlatformSettingsUpdateDto request) {
         settingService.updatePlatformSettings(request);
+    }
+
+    @Override
+    @AuditLogged(module = Module.CORE, resource = Resource.SETTINGS, operation = Operation.DETAIL, name = "branding")
+    public BrandingSettingsDto getBrandingSettings() {
+        return settingService.getBrandingSettings();
+    }
+
+    /**
+     * The body is kept out of the verbose audit record. Two logos at their limit are close to 2.7 MiB of base64, and
+     * the audit record travels over JMS: recording them would put that on the queue for every branding save, to say
+     * only what the operation name already says.
+     */
+    @Override
+    @AuditLogged(module = Module.CORE, resource = Resource.SETTINGS, operation = Operation.UPDATE, name = "branding")
+    public void updateBrandingSettings(@Sensitive BrandingSettingsUpdateDto brandingSettingsDto) {
+        settingService.updateBrandingSettings(brandingSettingsDto);
     }
 
     @Override
--- a/src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
@@ -10,6 +10,9 @@ import com.otilm.api.exception.ValidationException;
 import com.otilm.api.model.core.auth.Resource;
 import com.otilm.api.model.core.logging.enums.AuditLogOutput;
 import com.otilm.api.model.core.other.ResourceEvent;
+import com.otilm.api.model.core.settings.BrandingSettingsDto;
+import com.otilm.api.model.core.settings.BrandingSettingsUpdateDto;
+import com.otilm.api.model.core.settings.BrandingTheme;
 import com.otilm.api.model.core.settings.CertificateRegistrationSettingsDto;
 import com.otilm.api.model.core.settings.CertificateRegistrationSettingsUpdateDto;
 import com.otilm.api.model.core.settings.CertificateRequestAttributesSettingsDto;
@@ -46,6 +49,7 @@ import com.otilm.core.service.TriggerExternalService;
 import com.otilm.core.service.TriggerInternalService;
 import com.otilm.core.service.registration.CertificateRegistrationDefaults;
 import com.otilm.core.settings.SettingsCache;
+import com.otilm.core.settings.branding.BrandingSettingsValidator;
 import com.otilm.core.util.AttributeDefinitionUtils;
 import com.otilm.core.util.SecretEncodingVersion;
 import com.otilm.core.util.SecretsUtil;
@@ -64,10 +68,13 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -92,6 +99,12 @@ public class SettingServiceImpl implements SettingExternalService, SettingIntern
 
     private static final String DESERIALIZATION_ERROR_MESSAGE = "Cannot deserialize OAuth2 Provider Settings for provider '%s'.";
     private static final Logger logger = LoggerFactory.getLogger(SettingServiceImpl.class);
+
+    /**
+     * Branding is stored as one row per field rather than one serialized blob, so that clearing a single colour or logo
+     * removes only its own row and that field falls back to its platform default independently of the rest.
+     */
+    private static final Map<String, BrandingField> BRANDING_FIELDS = brandingFields();
 
     private final ObjectMapper wireMapper;
     private final SettingsCache settingsCache;
@@ -213,7 +226,124 @@ public class SettingServiceImpl implements SettingExternalService, SettingIntern
 
         platformSettings.setCertificates(certificateSettingsDto);
 
+        // Branding
+        platformSettings
+                .setBranding(
+                        readBrandingSettings(mappedSettings.get(SettingsSectionCategory.PLATFORM_BRANDING.getCode())));
+
         return platformSettings;
+    }
+
+    private record BrandingField(Function<BrandingSettingsUpdateDto, String> fromUpdate,
+            BiConsumer<BrandingSettingsDto, String> intoResponse) {
+    }
+
+    private static Map<String, BrandingField> brandingFields() {
+        Map<String, BrandingField> fields = new LinkedHashMap<>();
+        fields
+                .put("primaryColor", new BrandingField(BrandingSettingsUpdateDto::getPrimaryColor,
+                        BrandingSettingsDto::setPrimaryColor));
+        fields
+                .put("secondaryColor", new BrandingField(BrandingSettingsUpdateDto::getSecondaryColor,
+                        BrandingSettingsDto::setSecondaryColor));
+        fields
+                .put("tertiaryColor", new BrandingField(BrandingSettingsUpdateDto::getTertiaryColor,
+                        BrandingSettingsDto::setTertiaryColor));
+        fields
+                .put("backgroundColor", new BrandingField(BrandingSettingsUpdateDto::getBackgroundColor,
+                        BrandingSettingsDto::setBackgroundColor));
+        fields
+                .put("textColor",
+                        new BrandingField(BrandingSettingsUpdateDto::getTextColor, BrandingSettingsDto::setTextColor));
+        fields
+                .put("lightLogo",
+                        new BrandingField(BrandingSettingsUpdateDto::getLightLogo, BrandingSettingsDto::setLightLogo));
+        fields
+                .put("darkLogo",
+                        new BrandingField(BrandingSettingsUpdateDto::getDarkLogo, BrandingSettingsDto::setDarkLogo));
+        fields
+                .put("defaultTheme",
+                        new BrandingField(
+                                update -> update.getDefaultTheme() == null ? null : update.getDefaultTheme().getCode(),
+                                SettingServiceImpl::applyDefaultTheme));
+        return Map.copyOf(fields);
+    }
+
+    /**
+     * A stored theme code that no longer maps to anything is dropped rather than thrown: branding is read on every page
+     * render, and one unrecognised value must not take the whole platform settings read down with it.
+     */
+    private static void applyDefaultTheme(BrandingSettingsDto branding, String code) {
+        try {
+            branding.setDefaultTheme(BrandingTheme.findByCode(code));
+        } catch (ValidationException e) {
+            logger.warn("Ignoring unknown stored branding default theme '{}'.", code);
+        }
+    }
+
+    private BrandingSettingsDto readBrandingSettings(Map<String, Setting> brandingSettings) {
+        BrandingSettingsDto branding = new BrandingSettingsDto();
+        if (brandingSettings == null) {
+            return branding;
+        }
+        BRANDING_FIELDS.forEach((name, field) -> {
+            Setting setting = brandingSettings.get(name);
+            if (setting != null && setting.getValue() != null && !setting.getValue().isBlank()) {
+                field.intoResponse().accept(branding, setting.getValue());
+            }
+        });
+        return branding;
+    }
+
+    private Map<String, Setting> storedBrandingSettings() {
+        Map<String, Setting> stored = new HashMap<>();
+        settingRepository
+                .findBySectionAndCategory(SettingsSection.PLATFORM, SettingsSectionCategory.PLATFORM_BRANDING.getCode())
+                .forEach(setting -> stored.put(setting.getName(), setting));
+        return stored;
+    }
+
+    @Override
+    @ExternalAuthorization(resource = Resource.SETTINGS, action = ResourceAction.LIST)
+    public BrandingSettingsDto getBrandingSettings() {
+        return readBrandingSettings(storedBrandingSettings());
+    }
+
+    @Override
+    @ExternalAuthorization(resource = Resource.SETTINGS, action = ResourceAction.UPDATE_BRANDING)
+    public void updateBrandingSettings(BrandingSettingsUpdateDto brandingSettings) {
+        BrandingSettingsValidator.validate(brandingSettings);
+
+        // Held before the rows are read, because a field is written by looking for its row and inserting when there is
+        // none: two concurrent first updates would otherwise both find nothing and insert the same field twice.
+        settingRepository.lockBrandingWrites();
+
+        Map<String, Setting> stored = storedBrandingSettings();
+        BRANDING_FIELDS
+                .forEach((name, field) -> writeBrandingField(name, field.fromUpdate().apply(brandingSettings),
+                        stored.get(name)));
+
+        cacheAfterCommit(() -> settingsCache.cacheSettings(SettingsSection.PLATFORM, getPlatformSettingsInternal()));
+    }
+
+    private void writeBrandingField(String name, String value, Setting stored) {
+        if (value == null) {
+            // Reset to default is per field: with no row, the read leaves the field unset and the UI uses its own.
+            if (stored != null) {
+                settingRepository.delete(stored);
+            }
+            return;
+        }
+
+        Setting setting = stored;
+        if (setting == null) {
+            setting = new Setting();
+            setting.setSection(SettingsSection.PLATFORM);
+            setting.setCategory(SettingsSectionCategory.PLATFORM_BRANDING.getCode());
+            setting.setName(name);
+        }
+        setting.setValue(value);
+        settingRepository.save(setting);
     }
 
     private Setting certificateSetting(Map<String, Setting> certificateSettings, String name) {
--- a/src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
+++ b/src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
@@ -1,0 +1,324 @@
+package com.otilm.core.settings.branding;
+
+import com.otilm.api.exception.ValidationException;
+import com.otilm.api.model.core.settings.BrandingSettingsUpdateDto;
+import java.io.ByteArrayInputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.Locale;
+import java.util.Objects;
+import java.util.OptionalDouble;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.zip.CRC32;
+import javax.xml.XMLConstants;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+
+/**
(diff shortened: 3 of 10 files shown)


src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
```diff
@@ -213,9 +226,123 @@ private PlatformSettingsDto buildPlatformSettings() {
...
+    public BrandingSettingsDto getBrandingSettings() {
+        return readBrandingSettings(storedBrandingSettings());
+    }
+
+    @Override
+    @ExternalAuthorization(resource = Resource.SETTINGS, action = ResourceAction.UPDATE_BRANDING)
+    public void updateBrandingSettings(BrandingSettingsUpdateDto brandingSettings) {
+        BrandingSettingsValidator.validate(brandingSettings);
+
+        Map<String, Setting> stored = storedBrandingSettings();
```
lubomirw, line 321: **[High]** | Correctness

Concurrent first branding updates can both observe no row and insert the same field; without a business-key constraint, duplicate rows make later reads and clears nondeterministic.

_Suggested:_ Acquire a transaction-scoped branding advisory lock before reading the rows, enforce uniqueness for section/category/name, and add a concurrent-update integration test.

<sub>flagged by Copilot CLI, Guided PR review, Superpowers review</sub>

src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
```diff
@@ -0,0 +1,229 @@
...
+    private static byte[] decode(String field, String payload) {
+        try {
+            return Base64.getDecoder().decode(payload);
+        } catch (IllegalArgumentException e) {
+            throw reject(field, "is not valid base64");
+        }
+    }
+
+    private static void validatePng(String field, byte[] content) {
+        if (content.length < PNG_HEADER_LENGTH || !startsWithPngSignature(content) || !hasIhdrTag(content)) {
```
lubomirw: **[Medium]** | Correctness

This accepts the 24-byte test fixture as a PNG without verifying a complete IHDR, CRCs, image data, or IEND, so an invalid logo can be persisted.

_Suggested:_ Use a bounded structural PNG parser that validates the IHDR length/CRC and required chunks, and replace header-only positive fixtures with complete PNGs.

<sub>flagged by Copilot CLI, Guided PR review, Superpowers review</sub>

src/main/java/com/otilm/core/api/web/SettingControllerImpl.java
```diff
@@ -51,6 +53,18 @@ public void updatePlatformSettings(PlatformSettingsUpdateDto request) {
...
     }
 
+    @Override
+    @AuditLogged(module = Module.CORE, resource = Resource.SETTINGS, operation = Operation.DETAIL, name = "branding")
+    public BrandingSettingsDto getBrandingSettings() {
+        return settingService.getBrandingSettings();
+    }
+
+    @Override
+    @AuditLogged(module = Module.CORE, resource = Resource.SETTINGS, operation = Operation.UPDATE, name = "branding")
```
lubomirw, line 69: **[Medium]** | Performance

With verbose auditing enabled, the full DTO puts both encoded logos into the audit message (roughly 2.7 MiB at their limits); the JMS send also occurs after the settings transaction has committed.

_Suggested:_ Exclude logo payloads from additional audit data with a sensitive parameter or custom audit projection while retaining useful color/theme fields.

src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
```diff
@@ -0,0 +1,229 @@
...
+            return;
+        }
+
+        Matcher matcher = DATA_URI.matcher(dataUri);
+        if (!matcher.matches()) {
+            throw reject(field, MALFORMED_REASON);
+        }
+
+        String declaredMediaType = matcher.group(1).toLowerCase(Locale.ROOT);
+        byte[] content = decode(field, matcher.group(2));
```
lubomirw, line 90: **[Low]** | Performance

Direct service callers bypass Bean Validation, and this service-level validator decodes the entire payload before enforcing its one-MiB bound.

_Suggested:_ Reject data URIs longer than BrandingSettingsUpdateDto.LOGO_MAX_LENGTH before regex/decode, while retaining the exact decoded-size check.

<sub>flagged by Copilot CLI, Guided PR review</sub>

src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
```diff
@@ -0,0 +1,229 @@
...
+        return ratioOf(Double.parseDouble(width.group(1)), Double.parseDouble(height.group(1)));
+    }
+
+    /** {@code viewBox="minX minY width height"} — the last two values are the intrinsic proportions. */
+    private static OptionalDouble ratioFromViewBox(Element root) {
+        String[] parts = root.getAttribute("viewBox").trim().split("[\\s,]+");
+        if (parts.length != 4) {
+            return OptionalDouble.empty();
+        }
+        try {
```
lubomirw, line 283: **[Low]** | Correctness

Double.parseDouble accepts NaN; ratioOf and validateRatio then let a NaN ratio pass because every comparison with NaN is false.

_Suggested:_ Require finite, positive viewBox width and height and a finite ratio with Double.isFinite; add NaN and infinity regression tests.

<sub>flagged by Copilot CLI, Guided PR review, Superpowers review</sub>

src/main/java/com/otilm/core/api/web/SettingControllerImpl.java
```diff
@@ -51,6 +53,18 @@ public void updatePlatformSettings(PlatformSettingsUpdateDto request) {
...
+    public void updateBrandingSettings(BrandingSettingsUpdateDto brandingSettingsDto) {
```
lubomirw: **[Low]** | Correctness

The dependent interface documents HTTP 204 with @ApiResponse, but that annotation affects OpenAPI only; this void Spring handler returns 200 at runtime.

_Suggested:_ Declare @ResponseStatus(HttpStatus.NO_CONTENT) on the interface endpoint (or return ResponseEntity<Void>) and add an HTTP-level status test.

<sub>flagged by Superpowers review</sub>

src/test/java/com/otilm/core/integration/service/BrandingSettingsITest.java
```diff
@@ -0,0 +1,185 @@
...
+    /** The same caller still reads branding, which is gated by the broader LIST the settings pages already need. */
+    @Test
+    void anOperatorWithoutUpdateBrandingCanStillReadBranding() {
+        settingService.updateBrandingSettings(update(PRIMARY, null, null));
+        denyResourceAccess(Resource.SETTINGS, ResourceAction.UPDATE_BRANDING);
+
+        Assertions.assertEquals(PRIMARY, settingService.getBrandingSettings().getPrimaryColor());
+    }
+
+    /** The existing platform settings surface keeps its behaviour; branding only adds a category alongside it. */
```
lubomirw: **[Low]** | Maintainability

This comment restates the compatibility test name and narrates the feature change rather than a hidden invariant.

_Suggested:_ Remove the Javadoc.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/settings/branding/BrandingLogoValidatorTest.java
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    @Test
+    void anSvgOutsideTheAllowedAspectRatioIsRejected() {
+        String wide = "<svg xmlns='http://www.w3.org/2000/svg' width='500' height='100'/>";
+
+        Assertions.assertTrue(rejectionMessage(svgLogo(wide)).contains("aspect ratio"));
+    }
+
+    /** Named in the acceptance criteria: with neither dimensions nor a viewBox there is nothing to measure. */
```
lubomirw: **[Low]** | Maintainability

The acceptance-criteria reference is planning residue, and the remaining text repeats the test name.

_Suggested:_ Remove the Javadoc.

<sub>flagged by Hygiene</sub>

src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
```diff
@@ -213,9 +226,123 @@ private PlatformSettingsDto buildPlatformSettings() {
...
 
+        // Branding
+        platformSettings
+                .setBranding(
+                        readBrandingSettings(mappedSettings.get(SettingsSectionCategory.PLATFORM_BRANDING.getCode())));
+
         return platformSettings;
     }
 
+    /** One accessor pair per stored branding field: read the value out of an update, put it back into a response. */
```
lubomirw: **[Nit]** | Maintainability

The Javadoc expands the record component names fromUpdate and intoResponse without adding a non-obvious constraint.

_Suggested:_ Remove the Javadoc.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/integration/service/BrandingSettingsITest.java
```diff
@@ -0,0 +1,185 @@
...
+
+    @Test
+    void anInvalidColorIsRejectedAndNothingIsStored() {
+        BrandingSettingsUpdateDto branding = update("not-a-color", SECONDARY, null);
+
+        Assertions.assertThrows(ValidationException.class, () -> settingService.updateBrandingSettings(branding));
+        Assertions.assertTrue(storedBranding().isEmpty());
+    }
+
+    /** Validation runs before the first write, so a bad logo cannot leave half the update applied. */
```
lubomirw: **[Nit]** | Maintainability

The comment duplicates anInvalidLogoIsRejectedBeforeAnyFieldIsWritten.

_Suggested:_ Remove the Javadoc.

<sub>flagged by Hygiene</sub>

lubomirw (changes requested):
Automated multi-reviewer triage - 11 actionable findings (REQUEST_CHANGES).

Overflow (cannot anchor outside this PR's diff):
- **[Low] | Documentation** `OmniTrustILM/interfaces:src/main/java/com/otilm/api/model/core/settings/BrandingSettingsUpdateDto.java:75` (also `PublicBrandingDto.java:59`) - The published PNG example is only a signature, and the SVG example is `<svg/>` without dimensions or a viewBox, so Core rejects both. _Suggested:_ replace them with complete valid PNG and dimensioned SVG data URIs in both schemas. <sub>flagged by Docs check (Superpowers review)</sub>

src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
```diff
@@ -0,0 +1,229 @@
...
+            throw reject(field, MALFORMED_REASON);
+        }
+
+        String declaredMediaType = matcher.group(1).toLowerCase(Locale.ROOT);
+        byte[] content = decode(field, matcher.group(2));
+
+        if (content.length > BrandingSettingsUpdateDto.LOGO_MAX_DECODED_BYTES) {
+            throw reject(field, "is %d bytes; the maximum is %d bytes"
+                    .formatted(content.length, BrandingSettingsUpdateDto.LOGO_MAX_DECODED_BYTES));
+        }
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed. The encoded string is now bounded before the regex runs and before anything is decoded:

```java
if (dataUri.length() > BrandingSettingsUpdateDto.LOGO_MAX_LENGTH) {
    throw reject(field, "is %d encoded characters; the maximum is %d, being %d decoded bytes"...);
}
```

`LOGO_MAX_LENGTH` is derived from the 1 MiB decoded bound, so anything longer could never have decoded to a permitted size. The exact decoded-size check is unchanged and still the one that decides at the boundary — `aLogoExactlyAtTheAllowedSizeClearsTheSizeCheck` asserts a payload of exactly the maximum clears it and is judged on its content rather than its size, and `aLogoTooLongToBeWithinTheSizeLimitIsRejectedWithoutBeingDecoded` covers the new pre-check.

This closes the same gap as the [Low] finding on line 73, so both are covered by one change.

src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
```diff
@@ -0,0 +1,229 @@
...
+
+        switch (declaredMediaType) {
+            case PNG_MEDIA_TYPE -> validatePng(field, content);
+            case SVG_MEDIA_TYPE -> validateSvg(field, content);
+            default -> throw reject(field, BrandingSettingsUpdateDto.LOGO_VALIDATION_MESSAGE);
+        }
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed. `validateAndSanitize` no longer switches on the declared type. The decoded bytes decide the format, and the declaration only has to agree with what they turn out to be:

```java
if (startsWithPngSignature(content)) {
    requireDeclared(field, declaredMediaType, PNG_MEDIA_TYPE, "a PNG image");
    validatePng(field, content);
    return dataUri;
}

Element svgRoot = svgRootOrNull(content);
if (svgRoot != null) {
    requireDeclared(field, declaredMediaType, SVG_MEDIA_TYPE, "an SVG document");
    return sanitizedSvg(field, svgRoot);
}

throw reject(field, "is neither a PNG image nor a parseable SVG document");
```

Content that is neither now gets one consistent rejection whatever it declares — `contentThatIsNeitherAPngNorAnSvgIsRejectedWhateverItDeclares` runs the same JPEG bytes under `image/png`, `image/svg+xml` and `image/jpeg` and asserts the same message for all three. A mislabelled logo is refused rather than resolved in favour of the content, which is what the class Javadoc already claimed and the code did not do.

src/main/java/com/otilm/core/settings/branding/BrandingSettingsValidator.java
```diff
@@ -0,0 +1,53 @@
...
+        accessors.put("tertiaryColor", BrandingSettingsUpdateDto::getTertiaryColor);
+        accessors.put("backgroundColor", BrandingSettingsUpdateDto::getBackgroundColor);
+        accessors.put("textColor", BrandingSettingsUpdateDto::getTextColor);
+        return Map.copyOf(accessors);
+    }
+
+    public static void validate(BrandingSettingsUpdateDto branding) {
+        COLORS.forEach((field, accessor) -> validateColor(field, accessor.apply(branding)));
+        LOGOS.forEach((field, accessor) -> BrandingLogoValidator.validate(field, accessor.apply(branding)));
+    }
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed. `validated` now names the problem instead of dereferencing:

```java
if (branding == null) {
    throw new ValidationException("Branding settings must be provided.");
}
```

Covered by `aMissingUpdateIsRefusedRatherThanDereferenced`.

src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
Jakub-Bartal-01: Advisory lock added, with the concurrency test. Not the unique constraint — reasoning below.

`updateBrandingSettings` now takes a transaction-scoped lock before it reads the rows, following the `lockOAuth2ProviderWrites` precedent already in `SettingRepository` for the same class of race:

```java
@Query(value = "SELECT pg_advisory_xact_lock(hashtext(platform-branding-settings))", nativeQuery = true)
Object lockBrandingWrites();
```

`aBrandingUpdateWaitsForAConcurrentOneToRelease` holds that lock on a connection of its own and asserts the update does not complete while it is held, then completes once it is released. I checked it is a real regression test: with the `lockBrandingWrites()` call commented out it fails with *"the update did not wait for the branding lock"*.

On the unique constraint — I have left it out deliberately. It would mean a migration on the shared `setting` table, and issue #2060 AC#2 requires branding to persist *"with no database migration introduced"*. It would also be a table-wide constraint enforced on every other settings category rather than on branding, which is a wider change than this PRs scope. The lock fully serializes branding writes across nodes, which is the race reported here. Happy to raise it as a follow-up against the `setting` table if you want the constraint as belt-and-braces.

src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
Jakub-Bartal-01: Fixed. `pngDimensions` now walks the whole chunk sequence instead of reading dimensions off fixed offsets: IHDR must come first at its declared 13-byte length, every chunks CRC must match, IDAT must be present, and IEND must close the file exactly at its end with nothing after it. Every step is bounded by the already size-checked array, so a hostile chunk length terminates the walk rather than driving it.

The 24-byte fixture is gone. Positive fixtures are now complete PNGs written by `ImageIO`, and the old hand-built header survives only as `pngHeaderOnly`, used by the tests that assert it is *rejected*:

- `aPngThatIsNothingButAHeaderIsRejected` — the exact payload this finding named
- `aPngTruncatedBeforeItsEndChunkIsRejected`
- `aPngWithACorruptedChunkIsRejected` — CRC catches an edited chunk
- `aPngWhoseFirstChunkIsNotIhdrIsRejected`

`SvgSanitizerTest.aPngIsStoredUnchanged` on the follow-up branch carried the same 24-byte fixture and now uses a real PNG too.

src/main/java/com/otilm/core/api/web/SettingControllerImpl.java
Jakub-Bartal-01: Fixed. The body is annotated `@Sensitive`, which is what `AuditLogAspect` reads to keep a parameter out of the verbose `additionalData` map.

One thing to flag: `@Sensitive` is per parameter, not per field, so the colours and theme go with the logos rather than being retained. Doing better would mean a per-field audit projection in `AuditLogAspect`, which affects every audited endpoint and is more than this PR should carry. The record still identifies the operation as `SETTINGS`/`UPDATE`/`branding` with its actor, which is the part that matters for an audit trail; the alternative was ~2.7 MiB of base64 on the JMS queue per save to add the colour values. Say the word if you want the projection and I will raise it separately.

The post-commit JMS send is existing aspect behaviour, unchanged here.

src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
Jakub-Bartal-01: Fixed together with the [High] finding on line 78 — the encoded length is now bounded before the regex runs and before `Base64.getDecoder().decode` is called, while the exact decoded-size check stays as the one that decides at the boundary. Details and test names in [that thread](https://github.com/OmniTrustILM/core/pull/2081#discussion_r3789362207).

src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
Jakub-Bartal-01: Fixed. `ratioOf` now requires both dimensions and the resulting ratio to be finite:

```java
private static OptionalDouble ratioOf(double width, double height) {
    if (!Double.isFinite(width) || !Double.isFinite(height) || width <= 0 || height <= 0) {
        return OptionalDouble.empty();
    }
    double ratio = width / height;
    return Double.isFinite(ratio) ? OptionalDouble.of(ratio) : OptionalDouble.empty();
}
```

An undetermined ratio falls back to the existing "neither usable width and height nor a viewBox" rejection rather than silently passing. `anSvgWhoseViewBoxIsNotAFiniteSizeIsRejected` covers `NaN`, `Infinity` and `-Infinity` in either position and in both.

Only the viewBox path was affected in practice — `SVG_LENGTH` never matches `NaN` on `width`/`height` — but the guard sits in the shared helper so both paths are covered.

src/main/java/com/otilm/core/api/web/SettingControllerImpl.java
Jakub-Bartal-01: Fixed, in the interfaces repo where the endpoint is declared — `@ResponseStatus(HttpStatus.NO_CONTENT)` on `SettingController.updateBrandingSettings`, pushed to OmniTrustILM/interfaces#868 as `48e9f966`. That matches the convention the rest of the interfaces already follow (62 controllers declare `@ResponseStatus`).

Asserted over HTTP rather than off the annotation, in the new `SettingControllerBrandingITest`:

```java
mockMvc.perform(put("/v1/settings/platform/branding")...)
       .andExpect(status().isNoContent());
```

The same class also covers the read and a 422 on an invalid colour. Note the neighbouring `/v1/settings` writes (`platform`, `events`, `authentication`) have the same 204-documented-but-200-at-runtime gap; I have left those alone rather than widen this PR, but they are worth a sweep.

src/test/java/com/otilm/core/integration/service/BrandingSettingsITest.java
Jakub-Bartal-01: Removed.

src/test/java/com/otilm/core/settings/branding/BrandingLogoValidatorTest.java
Jakub-Bartal-01: Removed. The test name already says what the case is; the acceptance-criteria reference belonged in the issue, not the source.

src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
Jakub-Bartal-01: Removed.

src/test/java/com/otilm/core/integration/service/BrandingSettingsITest.java
Jakub-Bartal-01: Removed.

lubomirw (approved):

PR #2082, opened 2026-08-15
Sanitize uploaded SVG logos before they are stored

### LINKED ISSUE #2062: Sanitize uploaded SVG logos
### Description

Sanitize uploaded SVG logos before they are stored, so that accepting SVG does not turn the logo slots into a stored-XSS vector.

An SVG is an XML document that can carry `<script>` elements, event-handler attributes (`onload`, `onclick`), `javascript:` URLs, `<foreignObject>`, and external references via `xlink:href`, `<use>`, `<image>` or CSS `@import`. Branding logos are served same-origin and render on the login page for unauthenticated visitors, which makes this the highest-exposure surface in the Epic.

Scope:

- Parse and sanitize on upload, storing only the sanitized document. The original is not retained.
- Strip script elements, all event-handler attributes, `javascript:` and `data:` URLs, `<foreignObject>`, and external resource references.
- Configure XML parsing to disable DTDs and external entity resolution, so sanitization itself cannot be used for XXE.
- Reject rather than silently repair a document that fails to parse.
- Force the stored data URI's media type to `image/svg+xml`, regardless of what the caller declared, so the label can never disagree with the content.

The frontend counterpart is a hard rule that SVG logos are rendered through an `img` element and never inlined into the DOM. That rule is enforced in the logo-rendering issue; this issue is the server side.

### User Story

As a platform operator, I want uploaded SVG logos stripped of anything executable, so that branding cannot be used to run script against my users, including on the unauthenticated login page.

### Use Case

An administrator uploads an SVG containing an `onload` handler and an external image reference. The stored document has both removed; the rendered logo is visually unchanged.

### Acceptance Criteria

1. [ ] Uploaded SVGs are sanitized before storage and only the sanitized document is persisted.
2. [ ] Script elements, event-handler attributes, `javascript:` and `data:` URLs, `<foreignObject>` and external resource references are all removed.
3. [ ] XML parsing has DTD processing and external entity resolution disabled.
4. [ ] An SVG that cannot be parsed is rejected with a clear message rather than stored partially sanitized.
5. [x] A stored SVG logo carries `image/svg+xml` in its data URI, set from the sanitized content rather than from the caller's declaration. A response-level `Content-Security-Policy` is **not** part of this issue: per Epic decision 3 logos are delivered as base64 data URIs inside a JSON DTO, so there is no per-logo HTTP response to carry headers, and an application-wide CSP is tracked separately in #2112.
6. [ ] Tests cover a representative set of hostile payloads (inline script, event handler, `javascript:` URL, external entity, external `xlink:href`) and assert each is neutralized, plus a benign SVG that survives sanitization visually intact.
7. [ ] Sonar quality gate passes and new code meets at least 80% coverage.

Jakub-Bartal-01:
Closes #2062

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
@@ -312,7 +312,7 @@ public class SettingServiceImpl implements SettingExternalService, SettingIntern
     @Override
     @ExternalAuthorization(resource = Resource.SETTINGS, action = ResourceAction.UPDATE_BRANDING)
     public void updateBrandingSettings(BrandingSettingsUpdateDto brandingSettings) {
-        // Only the validated form is written: SVG logos come back sanitized, so the submitted document never lands.
+        // SVG logos come back sanitized, so the submitted document never reaches the settings table.
         BrandingSettingsUpdateDto sanitized = BrandingSettingsValidator.validated(brandingSettings);
 
         // Held before the rows are read, because a field is written by looking for its row and inserting when there is
--- a/src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
@@ -312,7 +312,8 @@ public class SettingServiceImpl implements SettingExternalService, SettingIntern
     @Override
     @ExternalAuthorization(resource = Resource.SETTINGS, action = ResourceAction.UPDATE_BRANDING)
     public void updateBrandingSettings(BrandingSettingsUpdateDto brandingSettings) {
-        BrandingSettingsValidator.validate(brandingSettings);
+        // Only the validated form is written: SVG logos come back sanitized, so the submitted document never lands.
+        BrandingSettingsUpdateDto sanitized = BrandingSettingsValidator.validated(brandingSettings);
 
         // Held before the rows are read, because a field is written by looking for its row and inserting when there is
         // none: two concurrent first updates would otherwise both find nothing and insert the same field twice.
@@ -320,7 +321,7 @@ public class SettingServiceImpl implements SettingExternalService, SettingIntern
 
         Map<String, Setting> stored = storedBrandingSettings();
         BRANDING_FIELDS
-                .forEach((name, field) -> writeBrandingField(name, field.fromUpdate().apply(brandingSettings),
+                .forEach((name, field) -> writeBrandingField(name, field.fromUpdate().apply(sanitized),
                         stored.get(name)));
 
         cacheAfterCommit(() -> settingsCache.cacheSettings(SettingsSection.PLATFORM, getPlatformSettingsInternal()));
--- a/src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
+++ b/src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
@@ -14,6 +14,7 @@ import java.util.zip.CRC32;
 import javax.xml.XMLConstants;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.TransformerException;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 
@@ -63,12 +64,17 @@ public final class BrandingLogoValidator {
     }
 
     /**
+     * Validates a submitted logo and returns the form of it that may be stored. An SVG comes back sanitized, so the
+     * original is never what reaches the settings table; a PNG comes back unchanged, having no executable content to
+     * remove.
+     *
      * @param field the branding field the logo was submitted under, so a rejection names the slot the operator sees
      * @param dataUri the submitted value; {@code null} means the slot is being cleared and is always accepted
+     * @return the value to store, or {@code null} when the slot is being cleared
      */
-    public static void validate(String field, String dataUri) {
+    public static String validateAndSanitize(String field, String dataUri) {
         if (dataUri == null) {
-            return;
+            return null;
         }
 
         // Bounded before the regex runs and before anything is decoded: a payload this long cannot decode to a
@@ -97,14 +103,13 @@ public final class BrandingLogoValidator {
         if (startsWithPngSignature(content)) {
             requireDeclared(field, declaredMediaType, PNG_MEDIA_TYPE, "a PNG image");
             validatePng(field, content);
-            return;
+            return dataUri;
         }
 
         Element svgRoot = svgRootOrNull(content);
         if (svgRoot != null) {
             requireDeclared(field, declaredMediaType, SVG_MEDIA_TYPE, "an SVG document");
-            validateSvg(field, svgRoot);
-            return;
+            return sanitizedSvg(field, svgRoot);
         }
 
         throw reject(field, UNRECOGNISED_REASON);
@@ -209,7 +214,11 @@ public final class BrandingLogoValidator {
                 | ((long) (content[offset + 2] & 0xFF) << 8) | (content[offset + 3] & 0xFF);
     }
 
-    private static void validateSvg(String field, Element root) {
+    /**
+     * The stored value is built from the sanitized document rather than from the submitted bytes, so there is no path
+     * by which the original — scripts, event handlers, external references and all — reaches the settings table.
+     */
+    private static String sanitizedSvg(String field, Element root) {
         OptionalDouble ratio = ratioFromWidthAndHeight(root);
         if (ratio.isEmpty()) {
             ratio = ratioFromViewBox(root);
@@ -218,8 +227,19 @@ public final class BrandingLogoValidator {
             throw reject(field, "is an SVG with neither usable width and height nor a viewBox, "
                     + "so its aspect ratio cannot be determined");
         }
-
         validateRatio(field, ratio.getAsDouble());
+
+        Document document = root.getOwnerDocument();
+        SvgSanitizer.sanitize(document);
+        try {
+            return "data:%s;base64,%s"
+                    .formatted(SVG_MEDIA_TYPE,
+                            Base64
+                                    .getEncoder()
+                                    .encodeToString(SvgSanitizer.serialize(document).getBytes(StandardCharsets.UTF_8)));
+        } catch (TransformerException e) {
+            throw reject(field, "is an SVG that could not be rewritten after sanitization");
+        }
     }
 
     /** {@code null} rather than a rejection: content that is not an SVG may still be a PNG, and the caller decides. */
--- a/src/main/java/com/otilm/core/settings/branding/SvgSanitizer.java
+++ b/src/main/java/com/otilm/core/settings/branding/SvgSanitizer.java
@@ -1,0 +1,176 @@
+package com.otilm.core.settings.branding;
+
+import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+import java.util.regex.Pattern;
+import javax.xml.XMLConstants;
+import javax.xml.transform.OutputKeys;
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerException;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
+import org.w3c.dom.Attr;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NamedNodeMap;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+/**
+ * Strips everything executable or externally referencing out of an operator-supplied SVG logo.
+ *
+ * <p>
+ * Branding logos are stored by the platform and rendered on the login page to visitors who have not authenticated, so
+ * an SVG logo is the highest-exposure input the branding feature accepts. The frontend renders logos through an
+ * {@code img} element, where browsers already refuse to run script or fetch external resources; this sanitizer is the
+ * server-side half, so that a logo is safe even if something later inlines it into the page.
+ *
+ * <p>
+ * Elements are handled by allow-list rather than by removing the dangerous ones. A deny-list only ever covers the
+ * attacks known when it was written — {@code script} and {@code foreignObject} are the obvious two, but {@code
+ * animate} can rewrite an attribute after load and {@code handler} exists in SVG 1.2 — whereas an allow-list of drawing
+ * constructs is safe against constructs nobody here has heard of yet.
+ */
+public final class SvgSanitizer {
+
+    /** Drawing constructs and their supporting definitions. Anything not on this list is dropped with its subtree. */
+    private static final Set<String> ALLOWED_ELEMENTS = Set
+            .of("svg", "g", "defs", "symbol", "use", "switch", "a", "title", "desc", "metadata", "style", "path",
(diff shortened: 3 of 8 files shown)


lubomirw (commented):
Automated multi-reviewer triage - 6 findings (COMMENT).

src/main/java/com/otilm/core/settings/branding/SvgSanitizer.java
```diff
@@ -0,0 +1,176 @@
...
+    private static void sanitizeAttributes(Element element) {
+        NamedNodeMap attributes = element.getAttributes();
+        List<Attr> declared = new ArrayList<>(attributes.getLength());
+        for (int i = 0; i < attributes.getLength(); i++) {
+            declared.add((Attr) attributes.item(i));
+        }
+
+        for (Attr attribute : declared) {
+            String name = attribute.getName().toLowerCase(Locale.ROOT);
+            if (name.startsWith("on") || isRejectedReference(name, attribute.getValue())) {
```
lubomirw: **[Medium]** | Security

Attributes such as fill, filter, mask and clip-path can retain external url(...) references because only href/src and style are inspected.

_Suggested:_ Validate every SVG attribute that accepts a functional IRI, removing values unless each URL is a same-document fragment, and add hostile tests for fill, filter, mask and clip-path.

<sub>flagged by Guided PR review, Copilot CLI</sub>

src/main/java/com/otilm/core/settings/branding/SvgSanitizer.java
```diff
@@ -0,0 +1,176 @@
...
+    private static boolean isRejectedReference(String name, String value) {
+        return REFERENCE_ATTRIBUTES.contains(name) && !LOCAL_REFERENCE.matcher(value.trim()).matches();
+    }
+
+    /** CSS can fetch too, through {@code @import} and {@code url()}, so both are removed from style declarations. */
+    private static String sanitizeCss(String css) {
+        if (css == null || css.isBlank()) {
+            return css;
+        }
+        if (CSS_IMPORT.matcher(css).find() || EXTERNAL_CSS_REFERENCE.matcher(css).find()) {
```
lubomirw: **[Medium]** | Security

CSS escapes such as \\75rl(https://attacker.example/x) or escaped @import tokens bypass the literal regular expressions and survive sanitization.

_Suggested:_ Use a CSS parser with an allow-list, or remove style elements and attributes entirely; add tests covering escaped URL and import tokens.

src/main/java/com/otilm/core/settings/branding/SvgSanitizer.java
```diff
@@ -0,0 +1,176 @@
...
+
+    private static void sanitizeAttributes(Element element) {
+        NamedNodeMap attributes = element.getAttributes();
+        List<Attr> declared = new ArrayList<>(attributes.getLength());
+        for (int i = 0; i < attributes.getLength(); i++) {
+            declared.add((Attr) attributes.item(i));
+        }
+
+        for (Attr attribute : declared) {
+            String name = attribute.getName().toLowerCase(Locale.ROOT);
```
lubomirw: **[Medium]** | Security

Checking qualified attribute names and accepting every raw # fragment allows alternate XLink prefixes and xml:base to turn retained references into external URLs.

_Suggested:_ Match reference attributes by namespace URI and local name, strip xml:base, and validate resolved URIs as same-document references.

<sub>flagged by Guided PR review, Copilot CLI</sub>

src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
```diff
@@ -132,19 +147,30 @@ private static void validateSvg(String field, byte[] content) {
             throw reject(field, "is an SVG with neither usable width and height nor a viewBox, "
                     + "so its aspect ratio cannot be determined");
         }
-
         validateRatio(field, ratio.getAsDouble());
+
+        SvgSanitizer.sanitize(document);
+        try {
+            return "data:%s;base64,%s"
```
lubomirw, line 235: **[Medium]** | Security

Issue #2062 requires image/svg+xml and restrictive Content-Security-Policy response headers, but this PR stores a data URI and does not add those response-layer controls.

_Suggested:_ Serve logos through a dedicated hardened asset response with the required headers, or explicitly split that acceptance criterion before closing issue #2062.

src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
```diff
@@ -313,11 +313,12 @@ public BrandingSettingsDto getBrandingSettings() {
     @Override
     @ExternalAuthorization(resource = Resource.SETTINGS, action = ResourceAction.UPDATE_BRANDING)
     public void updateBrandingSettings(BrandingSettingsUpdateDto brandingSettings) {
-        BrandingSettingsValidator.validate(brandingSettings);
+        // Only the validated form is written: SVG logos come back sanitized, so the submitted document never lands.
```
lubomirw: **[Nit]** | Maintainability

The comment above the sanitized assignment may restate the validated method call, sanitized variable name, and subsequent write operation.

_Suggested:_ If you agree it is redundant, remove or shorten the comment; otherwise keep it as a security-invariant note.

<sub>flagged by Hygiene</sub>

src/main/java/com/otilm/core/settings/branding/SvgSanitizer.java
```diff
@@ -0,0 +1,176 @@
...
+import javax.xml.transform.stream.StreamResult;
+import org.w3c.dom.Attr;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NamedNodeMap;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+/**
+ * Strips everything executable or externally referencing out of an operator-supplied SVG logo.
```
lubomirw, line 25: **[Nit]** | Maintainability

The class Javadoc expands into implementation and threat-model narration beyond a concise class summary.

_Suggested:_ If you want tighter hygiene, shorten it to a concise sanitizer summary and keep only non-obvious threat-model constraints.

<sub>flagged by Hygiene</sub>

src/main/java/com/otilm/core/settings/branding/SvgSanitizer.java
Jakub-Bartal-01: Fixed in de4b78c3. Rather than enumerating the attributes that accept a functional IRI, the `url()` rule now runs against **every** attribute that survives the other checks — `fill`, `stroke`, `filter`, `mask`, `clip-path` and the `marker-*` family included — so an attribute that gains a functional IRI in a later SVG revision is covered without the list being updated.

`anExternalFunctionalIriInAPresentationAttributeDoesNotSurvive` is parameterized over `fill`, `stroke`, `filter`, `mask`, `clip-path` and `marker-start`.

src/main/java/com/otilm/core/settings/branding/SvgSanitizer.java
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        for (Attr attribute : declared) {
+            String name = attribute.getName().toLowerCase(Locale.ROOT);
+            if (name.startsWith("on") || isRejectedReference(name, attribute.getValue())) {
+                element.removeAttributeNode(attribute);
+            } else if ("style".equals(name)) {
+                attribute.setValue(sanitizeCss(attribute.getValue()));
+            }
+        }
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed in de4b78c3 — see the reply on the sibling thread. The `url()` check moved off the `href`/`xlink:href`/`src` list and now applies to every attribute, with `anExternalFunctionalIriInAPresentationAttributeDoesNotSurvive` covering `fill`, `stroke`, `filter`, `mask`, `clip-path` and `marker-start`.

src/main/java/com/otilm/core/settings/branding/SvgSanitizer.java
Jakub-Bartal-01: Fixed in de4b78c3, by resolving the escapes rather than by removing `style` altogether — SVG exporters routinely emit a `<style>` block of `.cls-1{fill:#fff}` rules, so dropping it would break ordinary logos.

`decodeCssEscapes` implements the CSS escape grammar (a backslash and up to six hex digits closed by one optional whitespace, or a backslash and one literal character, with an unaddressable code point resolving to U+FFFD as CSS specifies), and the `@import` / `url()` / `image-set()` comparisons run against the decoded copy. Only the *decision* uses the decoded text; what survives is the original, unmodified, so the unescaping cannot itself corrupt a stylesheet.

Escapes are legal wherever a CSS value is parsed, which includes SVG presentation attributes, so the same decoding is applied there — `fill="\75 rl(…)"` is caught as well as `style="fill:\75 rl(…)"`. `anEscapedFetchDoesNotSurvive` is parameterized over the escaped `url(` and escaped `@import` in a stylesheet, and the escaped `url(` in both a presentation attribute and a style attribute.

Also added `image-set()` to the rejected set while here: it accepts a bare string, so it fetches without ever writing `url(`.

src/main/java/com/otilm/core/settings/branding/SvgSanitizer.java
Jakub-Bartal-01: Fixed in de4b78c3, on all three points.

**Prefix.** Reference attributes are now matched on **local name** rather than on the qualified name, which is strictly broader than matching the XLink namespace URI: `href` under any prefix bound to any namespace is treated as a reference, so `xl:href` is caught along with `xlink:href`. `anExternalReferenceUnderAnAlternateXlinkPrefixDoesNotSurvive` covers it. Namespace URIs are still used where the identity matters — `xmlns`/`xmlns:*` declarations are matched by `XMLNS_ATTRIBUTE_NS_URI` and kept, since dropping one would unbind a prefix the document still uses.

**`xml:base`.** Stripped, matched by `XML_NS_URI` + local name `base`, with `anXmlBaseDoesNotSurvive` asserting the attribute is gone and the `#g` reference beneath it survives.

**Resolution.** With `xml:base` gone there is no base to resolve against, so a `#…`-only value is same-document by construction rather than by assumption. That is now stated on `LOCAL_REFERENCE` and on the `isRejected` Javadoc, since the two rules only hold together.

One thing your review surfaced indirectly: the `url()` pattern was backtracking over consumed whitespace and an opening quote to satisfy its `(?!#)` lookahead, so `url( #g)` and `url('#g')` were being reported as **external** and the declaration blanked — a legitimate gradient silently dropped. The quantifiers are now possessive, and `aQuotedOrPaddedLocalReferenceSurvives` pins the four spellings.

src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
Jakub-Bartal-01: Correct that criterion 5 is unmet, and it will stay that way — pushing back on the suggested remedy rather than implementing it. AC5 was written assuming a logo asset endpoint that the Epic explicitly rejected.

Epic ilm#273, ratified decision 3, is *"Logo storage and formats. Base64 in `setting.value`, matching the platform's existing convention for binary content and requiring no migration"*, and decision 2 chose a read-only `GET /v1/branding` returning a DTO *"rather than a generated static asset"*. Under that shape a logo has no HTTP response of its own, so there is nowhere for a per-logo `Content-Type` or `Content-Security-Policy` to go. Serving each logo from a dedicated hardened endpoint would meet AC5, but it reverses both decisions and reworks the frontend contract and #2083 along with it — too large to fold into a sanitization PR, and not this issue's call to make.

What that delivery shape *can* enforce, and this PR does:

- The stored data URI's media type is set from the sanitized content, never from the caller's declaration — `requireDeclared` refuses a mislabelled upload outright, and `sanitizedSvg` rebuilds the URI with `image/svg+xml` regardless.
- The document is stripped of everything executable before storage, so there is no script left for a CSP to have to block.
- The frontend renders logos only through an `img` element and never inlines them, which is the rule the Epic pairs with this issue.

So AC5 has been split rather than quietly left open: #2062's criterion 5 is amended to state the data-URI equivalent, and the application-wide CSP — which core does not set on *any* response today, branding or otherwise — is now core#2112. That is the more honest home for it: a CSP scoped to one logo response would have been a narrow fix to a gap that spans every endpoint.

src/main/java/com/otilm/core/settings/branding/BrandingLogoValidator.java
```diff
@@ -218,8 +227,19 @@ private static void validateSvg(String field, Element root) {
...
+
+        Document document = root.getOwnerDocument();
+        SvgSanitizer.sanitize(document);
+        try {
+            return "data:%s;base64,%s"
+                    .formatted(SVG_MEDIA_TYPE,
+                            Base64
+                                    .getEncoder()
+                                    .encodeToString(SvgSanitizer.serialize(document).getBytes(StandardCharsets.UTF_8)));
+        } catch (TransformerException e) {
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Correct that criterion 5 is unmet, and it will stay that way — pushing back on the suggested remedy rather than implementing it. AC5 was written assuming a logo asset endpoint that the Epic explicitly rejected.

Epic ilm#273, ratified decision 3, is *"Logo storage and formats. Base64 in `setting.value`, matching the platform's existing convention for binary content and requiring no migration"*, and decision 2 chose a read-only `GET /v1/branding` returning a DTO *"rather than a generated static asset"*. Under that shape a logo has no HTTP response of its own, so there is nowhere for a per-logo `Content-Type` or `Content-Security-Policy` to go. Serving each logo from a dedicated hardened endpoint would meet AC5, but it reverses both decisions and reworks the frontend contract and #2083 along with it — too large to fold into a sanitization PR, and not this issue's call to make.

What that delivery shape *can* enforce, and this PR does:

- The stored data URI's media type is set from the sanitized content, never from the caller's declaration — `requireDeclared` refuses a mislabelled upload outright, and `sanitizedSvg` rebuilds the URI with `image/svg+xml` regardless.
- The document is stripped of everything executable before storage, so there is no script left for a CSP to have to block.
- The frontend renders logos only through an `img` element and never inlines them, which is the rule the Epic pairs with this issue.

So AC5 has been split rather than quietly left open: #2062's criterion 5 is amended to state the data-URI equivalent, and the application-wide CSP — which core does not set on *any* response today, branding or otherwise — is now core#2112. That is the more honest home for it: a CSP scoped to one logo response would have been a narrow fix to a gap that spans every endpoint.

src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
Jakub-Bartal-01: Agreed on the redundant half and shortened it. `validated(...)` already says the value is validated, and `sanitized` already says it is the sanitized one, so that much was restating the line below it. What was not obvious from either name is that SVG sanitization happens inside a method called `validated`, so the note is kept down to that:

```java
// SVG logos come back sanitized, so the submitted document never reaches the settings table.
```

src/main/java/com/otilm/core/settings/branding/SvgSanitizer.java
Jakub-Bartal-01: Trimmed in de4b78c3. The paragraph about the frontend rendering through `img` went — that constraint is enforced in the frontend issue and stating it here only invited it to drift out of date.

Kept the allow-list rationale, since it is the one thing a reader would otherwise be likely to undo: without the `animate` and `handler` examples, replacing the allow-list with a deny-list of the obvious dangerous elements looks like a simplification rather than a regression. The paragraph now also covers why attributes go the *other* way — kept unless they execute, rebase or fetch — which the class needed anyway after this round of changes.

lubomirw (approved):

PR #2083, opened 2026-08-15
Serve branding to unauthenticated clients

### LINKED ISSUE #2063: Serve branding to unauthenticated clients
### Description

Serve branding to unauthenticated clients so the login page can render the customer's identity before anyone signs in.

No endpoint serves operator-configured application data anonymously today: `AuthHelper.PERMITTED_ENDPOINTS` covers only `/v?/health/**` and `/v?/connector/register`, with `OAUTH2_ENDPOINTS` covering the login flow. This is the first anonymous endpoint to expose customer content, so the response shape is deliberately narrow.

Scope:

- `BrandingControllerImpl` implementing `GET /v1/branding`, read-only, returning `PublicBrandingDto`.
- Register the path in `AuthHelper.PERMITTED_ENDPOINTS`.
- The response is built by explicit field mapping from the branding settings; it never returns a settings DTO directly, so a future settings field cannot leak by inheritance.
- Reads go through the existing `SettingsCache`, so the anonymous path adds no database load per request.

Anonymous disclosure of the customer's logo and brand colours is an accepted, stated trade-off recorded in the Epic.

### User Story

As an anonymous visitor, I want the instance to show its owner's identity on the login screen, so that I can tell I am on the right system before entering credentials.

### Use Case

An unauthenticated browser requests `GET /v1/branding` and receives the configured colours, logos and default theme, and nothing else.

### Acceptance Criteria

1. [ ] `GET /v1/branding` returns branding without authentication and is registered in the permit-all list.
2. [ ] The endpoint is read-only; any other HTTP method against it is refused.
3. [ ] The response type contains branding fields only and is populated by explicit mapping, not by returning or extending a settings DTO.
4. [ ] A test asserts the serialized anonymous response contains exactly the expected branding keys, and fails if any additional settings field appears.
5. [ ] The endpoint reads through `SettingsCache` and issues no per-request database query.
6. [ ] With branding unset, the endpoint returns platform defaults rather than an error.
7. [ ] Existing authenticated settings endpoints and the OAuth2 login flow are unaffected; their tests pass unchanged.
8. [ ] Sonar quality gate passes and new code meets at least 80% coverage.

Jakub-Bartal-01:
Closes #2063


src/main/java/com/otilm/core/service/impl/BrandingServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Reads through the settings cache, which is populated at startup and on the configured refresh interval, so a page
+     * load costs no database query. A cache that has not been populated yet reports unconfigured branding rather than
+     * failing: the login page then renders the platform's own identity, which is what it did before branding existed.
+     */
+    private static PublicBrandingDto toPublicBranding(BrandingSettingsDto branding) {
+        PublicBrandingDto publicBranding = new PublicBrandingDto();
+        if (branding == null) {
+            return publicBranding;
+        }
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Checked this against the DTO, and `configured` is safe on that path — but the check turned up a real defect one field over, so this thread was worth having.

**On `configured` itself.** `PublicBrandingDto.configured` is a primitive `boolean`, so it cannot be null, and the class carries `@JsonInclude(JsonInclude.Include.ALWAYS)` specifically to override Core's web `ObjectMapper` default of `NON_NULL`. The early return therefore serializes `"configured": false` — present, boolean, false. That was already covered by `anUnbrandedInstanceReportsUnconfiguredBrandingRatherThanFailing`, which exercises exactly the `branding == null` path (`clearBranding()` caches a `PlatformSettingsDto` with no branding), though only by reading the value. It now asserts the key is a present boolean before reading it, so the property you were asking about is stated rather than implied:

```java
Assertions.assertTrue(response.path("configured").isBoolean(), response.toString());
Assertions.assertFalse(response.get("configured").asBoolean());
```

**What the check did find.** `defaultTheme` is the field that behaves the way you suspected `configured` might — and deliberately so. It resolves to a `$ref` on the shared `BrandingTheme` schema, and a `$ref` cannot be declared nullable without emitting a sibling `type` that contradicts the referenced enum, so it is annotated `NON_NULL` and **omitted** when no theme is configured. Absence is representable in every generated client; `null` on a required `$ref` is representable in none.

Two tests were wrong about that, and `e4b2dbd8` fixes both:

- `theResponseCarriesExactlyTheBrandingKeysAndNothingElse` configured only `primaryColor` and expected `defaultTheme` in the key set anyway, so it **failed**. The key set is now split into the eight keys the response always carries plus `defaultTheme` on top, and the exact-key assertion runs against a response *with* a theme — the fullest shape the endpoint can produce — so it still fails the build if a settings field leaks into the DTO. The omission is pinned by `defaultThemeIsAbsentRatherThanNullWhenNoThemeIsConfigured`.
- `theResponseKeepsTheSameShapeWhetherOrNotBrandingIsConfigured` claimed "unset fields stay in the response as explicit nulls" — untrue of `defaultTheme`, and it passed only because neither side of the comparison configured a theme. It now configures one and excludes the documented exception; removing that exclusion makes it fail, which I verified rather than assumed.

So: no change to `BrandingServiceImpl` — the mapping was right — and the response-shape guarantee it relies on is now asserted honestly instead of by coincidence.

src/test/java/com/otilm/core/integration/api/web/PublicBrandingControllerITest.java
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * An unbranded instance is a normal state, not an error: the client applies the platform's own look. {@code
+     * configured} is asserted to be a present boolean rather than merely falsy, because a primitive field behind
+     * {@code @JsonInclude(ALWAYS)} is the only reason the key cannot go missing on the path where there is no branding
+     * to map at all.
+     */
+    @Test
+    void anUnbrandedInstanceReportsUnconfiguredBrandingRatherThanFailing() throws Exception {
+        JsonNode response = fetchBranding();
+
```
lubomirw, line 142: **[High]** | Testing

The exact-key assertion requires `defaultTheme`, but `PublicBrandingDto` intentionally omits that field when it is null, so this primary-color-only case fails deterministically.

_Suggested:_ Configure a non-null theme for the full-shape assertion and add a separate assertion that `defaultTheme` is absent when unset.

<sub>flagged by Guided PR review, Copilot CLI, and Superpowers</sub>

src/test/java/com/otilm/core/integration/api/web/PublicBrandingControllerITest.java
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * The keys the response carries whatever is configured. {@code defaultTheme} is deliberately not among them: it
+     * resolves to a {@code $ref} on the shared {@code BrandingTheme} schema, which cannot be declared nullable, so
+     * {@code PublicBrandingDto} omits it when unset rather than sending it as null.
+     */
+    private static final Set<String> ALWAYS_PRESENT_KEYS = Set
+            .of("configured", "primaryColor", "secondaryColor", "tertiaryColor", "backgroundColor", "textColor",
+                    "lightLogo", "darkLogo");
+
+    /** Everything the anonymous response is allowed to carry. A key outside this set is a leak, not a feature. */
```
lubomirw, line 67: **[High]** | Testing

This new `@MockitoSpyBean` context signature raises the measured Spring context count from the committed baseline of 57 to 58, so `ContextSignatureGuardTest` fails.

_Suggested:_ Reuse an existing context signature, or deliberately update `ContextSignatureGuardTest.BASELINE` to 58 if the additional context boot is justified.

<sub>flagged by Guided PR review and Copilot CLI</sub>

src/main/java/com/otilm/core/service/impl/BrandingServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+
+        publicBranding.setPrimaryColor(branding.getPrimaryColor());
+        publicBranding.setSecondaryColor(branding.getSecondaryColor());
+        publicBranding.setTertiaryColor(branding.getTertiaryColor());
+        publicBranding.setBackgroundColor(branding.getBackgroundColor());
+        publicBranding.setTextColor(branding.getTextColor());
+        publicBranding.setLightLogo(branding.getLightLogo());
```
lubomirw, line 48: **[Medium]** | Security

An anonymous request can return both allowed 1 MiB logos inline, producing roughly 2.7 MiB of JSON and repeatable response allocation/bandwidth cost.

_Suggested:_ Make the exposure decision explicit and add appropriate protection, such as cache headers plus gateway rate limiting, or separately cacheable logo URLs.

<sub>flagged by Copilot CLI; needs product/architecture judgment</sub>

src/main/java/com/otilm/core/util/AuthHelper.java
```diff
@@ -71,7 +71,13 @@ public class AuthHelper {
...
     public static final String AUDITOR_ROLE_NAME = "auditor";
 
-    public static final List<String> PERMITTED_ENDPOINTS = List.of("/v?/health/**", "/v?/connector/register");
+    /**
+     * {@code /v?/branding} is the first entry here that serves operator-configured content rather than platform
+     * mechanics: the login page renders the customer's identity before anyone signs in. It is read-only and returns a
+     * purpose-built DTO carrying branding fields and nothing else.
+     */
+    public static final List<String> PERMITTED_ENDPOINTS = List
+            .of("/v?/health/**", "/v?/connector/register", "/v?/branding");
```
lubomirw: **[Low]** | Security

The permit-all path bypasses authentication for every HTTP method, while read-only enforcement currently depends on Spring having only a GET handler.

_Suggested:_ For defense in depth, make both the Spring Security matcher and the authentication-filter bypass specific to GET requests for this path.

<sub>flagged by Copilot CLI; current non-GET requests are rejected by handler mapping</sub>

src/main/java/com/otilm/core/service/impl/BrandingServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+import com.otilm.core.settings.SettingsCache;
+import java.util.Objects;
+import java.util.stream.Stream;
+import org.springframework.stereotype.Service;
+
+/**
+ * Serves the branding an unauthenticated caller may read.
+ *
+ * <p>
+ * This is the platform's first anonymous endpoint carrying operator-configured content, so the response is assembled
```
lubomirw: **[Low]** | Maintainability

Calling this the platform's "first" anonymous operator-content endpoint records a milestone that will become stale as the platform evolves.

_Suggested:_ Keep only the durable rationale that the response is mapped field by field to prevent accidental anonymous exposure.

<sub>flagged by Hygiene</sub>

src/main/java/com/otilm/core/service/impl/BrandingServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    @UnauthenticatedEndpoint
+    public PublicBrandingDto getPublicBranding() {
+        PlatformSettingsDto platform = SettingsCache.getSettings(SettingsSection.PLATFORM);
+        return toPublicBranding(platform == null ? null : platform.getBranding());
+    }
+
+    /**
+     * Reads through the settings cache, which is populated at startup and on the configured refresh interval, so a page
+     * load costs no database query. A cache that has not been populated yet reports unconfigured branding rather than
+     * failing: the login page then renders the platform's own identity, which is what it did before branding existed.
```
lubomirw: **[Low]** | Maintainability

The phrase "which is what it did before branding existed" records change history rather than explaining the current behavior.

_Suggested:_ End the sentence after stating that unpopulated cache state reports unconfigured branding and the login page uses the platform identity.

<sub>flagged by Hygiene</sub>

src/main/java/com/otilm/core/util/AuthHelper.java
```diff
@@ -71,7 +71,13 @@ public class AuthHelper {
     /** {@code AuthResourceSynchronizer} rederives this role's grants from the resource catalogue on every startup. */
     public static final String AUDITOR_ROLE_NAME = "auditor";
 
-    public static final List<String> PERMITTED_ENDPOINTS = List.of("/v?/health/**", "/v?/connector/register");
+    /**
+     * {@code /v?/branding} is the first entry here that serves operator-configured content rather than platform
```
lubomirw: **[Low]** | Maintainability

The "first entry here" wording is historical decision narration that will become stale.

_Suggested:_ Document only the current reason for anonymous access: the login page needs a read-only, purpose-built branding DTO before sign-in.

<sub>flagged by Hygiene</sub>

src/main/java/com/otilm/core/api/web/BrandingControllerImpl.java
```diff
@@ -0,0 +1,32 @@
+package com.otilm.core.api.web;
+
+import com.otilm.api.interfaces.core.web.BrandingController;
+import com.otilm.api.model.core.branding.PublicBrandingDto;
+import com.otilm.core.service.BrandingExternalService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * The anonymous branding surface. Read-only by construction: the controller interface declares a single {@code GET}, so
```
lubomirw: **[Nit]** | Maintainability

The class Javadoc is longer than necessary, although its read-only and audit-spam rationale is useful.

_Suggested:_ Tighten it to the two durable points: the interface permits only GET, and anonymous requests are not audit-logged to prevent log spam.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/integration/api/web/PublicBrandingControllerITest.java
```diff
@@ -0,0 +1,197 @@
...
+        settingsCache.cacheSettings(SettingsSection.PLATFORM, platform);
+    }
+
+    private JsonNode fetchBranding() throws Exception {
+        String body = mvc.perform(get(PATH)).andExpect(status().isOk()).andReturn().getResponse().getContentAsString();
+        return mapper.readTree(body);
+    }
+
+    @Test
+    void brandingIsServedWithoutAuthentication() throws Exception {
```
lubomirw, line 121: **[Nit]** | Maintainability

The comment partly restates the descriptive test name.

_Suggested:_ Keep only the additional rationale: `The client applies the platform's own look when branding is unconfigured.`

<sub>flagged by Hygiene</sub>

lubomirw (changes requested):
Automated multi-reviewer triage - 9 selected findings (REQUEST_CHANGES).

src/test/java/com/otilm/core/integration/api/web/PublicBrandingControllerITest.java
Jakub-Bartal-01: Already fixed, in `e4b2dbd8` — the triage and that commit crossed. `theResponseCarriesExactlyTheBrandingKeysAndNothingElse` now configures `BrandingTheme.LIGHT` and asserts against `PERMITTED_KEYS` (the eight always-present keys plus `defaultTheme`), so the exact-key assertion runs against the fullest response the endpoint can produce. The omission has its own test, `defaultThemeIsAbsentRatherThanNullWhenNoThemeIsConfigured`, which asserts `!response.has("defaultTheme")` and holds the rest to `ALWAYS_PRESENT_KEYS`.

`theResponseKeepsTheSameShapeWhetherOrNotBrandingIsConfigured` had the same latent bug the other way round — it claimed unset fields stay as explicit nulls and passed only because neither side configured a theme. It now configures one and excludes the documented exception.

src/test/java/com/otilm/core/integration/api/web/PublicBrandingControllerITest.java
Jakub-Bartal-01: Confirmed, and fixed by reusing an existing signature rather than raising the baseline. `mvn test -Dtest=ContextSignatureGuardTest` reported 58 against a `BASELINE` of 57.

Your diagnosis pointed at the `@MockitoSpyBean`, but that was not the whole cause: stripping the spy and re-running the census still gave 58. The class had no mock overrides of its own beyond the base, and no committed combination pairs `BaseSpringBootTestNoAuth` + `@AutoConfigureMockMvc` + the cleared context path with an empty mock set.

What made reuse possible was the GET-only change on the thread below. The write assertions now take the authenticated branch, where the JDBC session filter tries to persist a session into a `spring_session` table the test schema does not have — so the test needs exactly the three session mocks `TspSecurityChainITest` already declares. Adopting that set verbatim makes the two signatures identical and the context shared: the census is back to 57 and the class now runs in 0.1s instead of booting a context.

The one casualty was `theEndpointIssuesNoDatabaseQuery`, since the spy is the remaining distinguishing axis. It moved to a plain `BrandingServiceImplTest` (no context, so no signature cost), which states the same property — the settings cache is the service the only source, there is no repository to reach for — and additionally covers the empty-cache path the integration test could not reach, because `clearBranding` always cached a `PlatformSettingsDto` and never left the section null.

One trap worth recording: the signature is read out of the source text, so `SessionRepositoryFilter<?>` and the raw `SessionRepositoryFilter` are different signatures. The declaration has to stay raw (with `@SuppressWarnings("rawtypes")`) or the context splits again. The class Javadoc says so.

src/main/java/com/otilm/core/service/impl/BrandingServiceImpl.java
Jakub-Bartal-01: Two of the three are now in the code; the third is deployment-side.

The payload figure is right: `LOGO_MAX_DECODED_BYTES` is 1 MiB per logo, so two base64 logos come to roughly 2.7 MB of JSON.

Added `BrandingCacheControlConfig`, which puts `Cache-Control: public, max-age=60` on `/v?/branding`. `public` is correct here — the response is identical for every caller and carries no identity — so a page reload, a second tab or a back-navigation is served from the browser cache instead of the server. 60s keeps an operator who changes the branding from waiting to see it on the login page. Asserted by `theResponseIsPubliclyCacheable`, against the literal header value rather than the constant.

The exposure decision is now stated where the decision lives, in that class Javadoc: two logos are carried inline, bounded by the upload cap, and caching bounds what honest traffic costs — not what a determined caller can ask for.

Rate limiting I have deliberately not put in the application. Deliberate repetition against an anonymous endpoint is a gateway concern, and it is where `/v?/health/**` and `/v?/connector/register` are already protected; a per-instance limiter in Core would not survive horizontal scaling. Separately cacheable logo URLs would be the answer if the login page needed more than one round trip, but it needs the colours and the logos together, so a second request would only add latency to every login.

src/main/java/com/otilm/core/util/AuthHelper.java
Jakub-Bartal-01: Done. `PERMITTED_ENDPOINTS` now splits in two: a method-agnostic list (`/v?/health/**`, `/v?/connector/register`, unchanged) and `GET_PERMITTED_ENDPOINTS`, holding `/v?/branding` alone.

Both halves of the bypass respect it. `SecurityConfig` adds a second `requestMatchers(HttpMethod.GET, AuthHelper.getGetOnlyPermitAllEndpoints()).permitAll()`, and `permitAllEndpointInRequest` takes the request method, so `PlatformAuthenticationFilter` only short-circuits to the anonymous token for a GET. A `POST`/`PUT`/`DELETE` against the path is now 401 from the security chain rather than 400 from the handler mapping — `theEndpointIsReadOnly` asserts that, and `AuthHelperBrandingEndpointTest` covers the matcher directly, including that the method-agnostic entries stay open for every method.

Two consequences worth naming. `HEAD` is no longer anonymous either; I left it out because the login page GETs and nothing HEADs this path, and every extra permitted method is surface to justify. There is no CORS configuration in Core, so no `OPTIONS` preflight depends on the path being open.

src/main/java/com/otilm/core/service/impl/BrandingServiceImpl.java
Jakub-Bartal-01: Trimmed. The milestone claim is gone; the Javadoc now opens on the durable reason, that the response is assembled field by field into a purpose-built DTO so a field later added to platform settings is not served to anonymous callers by default.

src/main/java/com/otilm/core/service/impl/BrandingServiceImpl.java
Jakub-Bartal-01: Trimmed — the sentence now ends at "the login page then renders the platform's own identity."

src/main/java/com/otilm/core/util/AuthHelper.java
Jakub-Bartal-01: Reworded, and the constant it sat on is now split (see the GET-only thread). The Javadoc on `GET_PERMITTED_ENDPOINTS` states the current reason only: the login page needs the customer's identity before anyone signs in, the response is a purpose-built DTO carrying branding fields and nothing else, and only GET is opened so a write is refused by the security chain rather than by the handler mapping happening to register no other method.

src/main/java/com/otilm/core/api/web/BrandingControllerImpl.java
Jakub-Bartal-01: Tightened to the two durable points, in three lines: the controller interface declares a single GET so the path is read-only by construction, and it is deliberately not audit-logged because an unauthenticated caller could otherwise fill the audit log at will.

src/test/java/com/otilm/core/integration/api/web/PublicBrandingControllerITest.java
Jakub-Bartal-01: Trimmed to your wording. The Javadoc now opens "The client applies the platform's own look when branding is unconfigured." and keeps only the part the test name does not carry — why `configured` is asserted to be a present boolean rather than merely falsy.

lubomirw (approved):

Issue #1982, opened 2026-08-06
SAD: minting, binding, budget, replay tracking


ivosh:
### Description

The security center of the CSC part. Core mints the SAD at `credentials/authorize` - **inside the RSSP**, which is the single design move that makes all four legacy policy gaps fixable (D6).

Mint and consume are **two independent client calls** (`credentials/authorize`, then `signatures/*`), an arbitrary delay apart and possibly on different Core replicas - not two halves of one request. In spec terms this is credential authorization mode `explicit` (CSC 8.1.4), for which `credentials/authorize` is the required method; the transport bearer is *service* authorization, a separate layer that stays OAuth2. The spec's other mode - `oauth2code`, an AS-minted SAD, which SHALL NOT use `credentials/authorize` - is the legacy Keycloak shape and is not v1. CSC 11.8 explicitly allows explicit-mode authorize with no authentication objects, so no PIN or OTP machinery is needed.

**Form: an opaque handle referencing Core-persisted rows** - not a signed token. The budget and replay state must exist server-side anyway, so a self-contained JWT would add key-management and revocation questions while removing nothing; and an opaque handle keeps the SAD modeled as provider-agnostic signature-activation evidence (the SAM constraint), so a future SAM-minted SAD is a different evidence kind behind the same seam.

**The budget IS the occurrence set (decided 2026-08-10).** A SAD header row (user, credential, expiry) plus **one uniquely-indexed row per authorized hash occurrence**: SAD id, digest algorithm, hash, occurrence index. Each occurrence row carries its own state - `AUTHORIZED` -> `RESERVED` -> `CONSUMED` or `RELEASED` - and, once reserved, the owning `(operationId, itemId)`.

Why not a membership set plus a counter: authorize `{h1, h2}` with `numSignatures` 2 would let a client sign `h1` twice and never sign `h2`, because a set membership test plus a decrement cannot tell those two runs apart. `numSignatures` therefore stops being independent state - it is the number of occurrence rows minted, and the cap the profile's `multisign` bounds.

[...]

PR #2089, opened 2026-08-17
Verify the data-to-be-signed is bound to the authorized document

### LINKED ISSUE #2037: DTBS digest binding verifier
### Description

The common contract (OmniTrustILM/interfaces#818) **specifies** the mandatory `documentDigest` echo on the `computeDtbs` response; this child **implements** the Core-side check - one uniform rule for every family: equality of the echoed digest against the digest the SAD authorized, run between steps 1 and 2 of the level state machine, before key release (design specification 3.2, 4.4).

The core#1974 reconcile found this component unowned - the contract defines the echo, OmniTrustILM/core#1974 runs the gate, OmniTrustILM/core#1982 runs the same check inside the SAD reservation transaction, but nothing owned the verification code between them.

**One service, one rule.** `DtbsBindingVerifier`: input the authorized digest(s) and the echoed `documentDigest`; output the binding verdict. Core never parses the DTBS and never recomputes the digest from the document - per-family independent extraction was considered and dropped: it pulls CMS/XML/JOSE structure parsing into Core and leaves the transforming shapes (enveloped PAdES - the pilot - and enveloped/enveloping XAdES) on the echo anyway, so the system-level trust statement would not change (document-format understanding stays connector-side by design). The check catches bugs and document mix-ups, while envelope construction sits inside the signing trust boundary; SDR authorization stays the recorded malice-proof upgrade (design spec 4.4).

- The rule is family-blind: PAdES, XAdES, CAdES and JAdES all satisfy DP-7 the same way; `signHash` needs no echo (the client-submitted digest is itself the authorized value).
- Multi-document operations check each echoed digest against its own authorized occurrence.

Failure currency: a missing echo or a failed equality check fails in engine vocabulary (invalid input / step failure naming the step), mapped by each protocol edge. The verifier is pure and stateless - no I/O, no persistence.

Placement: the verifier lives with the content-signing workflow (`signing/contentsigning/`) - no AdES vocabulary enters `signing/engine/` (the core#1972 isolation rule).

### Definition of Done

- [ ] `DtbsBindingVerifier` refuses a missing, unusable or mismatched documentDigest echo with a checked engine exception, so no caller can reach key release past a failed check
- [ ] Behavior matches the echo contract the common skeleton specifies (interfaces#818), with positive and negative vectors
- [ ] A missing or mismatched echo fails in engine vocabulary, naming the step

Enforcement in the signing flow is out of this child's scope: the verifier ships ahead of its consumers. The OmniTrustILM/core#1974 gate (between steps 1 and 2) and the OmniTrustILM/core#1982 reservation-time check call this one implementation without duplicating the rule; that key release is actually refused is verified by their acceptance criteria. (DoD re-scoped 2026-08-17.)

ivosh:
## What this adds

The Core-side half of DP-7. The formatting contract (interfaces#818) specifies the mandatory `documentDigest` echo on the `computeDtbs` response; nothing owned the code that checks it. #1974 runs the gate between steps 1 and 2 of the level state machine, #1982 runs the same check inside the SAD reservation transaction — this is the one implementation both consume.

`DtbsBindingVerifier` compares the echoed digest against the digest the operation authorized, before the signing key is released. Pure and stateless: no I/O, no persistence, no Spring bean.

`DocumentDigest` carries a digest together with the algorithm that produced it, because a digest cannot be interpreted without one.

## Decisions worth a look

**A broken echo is not a mismatch.** An echo that is absent, of another algorithm, or of a length its algorithm never produces proves the connector broke the contract — it does not prove the document differs. Those get the neutral "internal error" client text; only a genuine digest inequality tells the client its document was not the authorized one. Operator detail (both digests, hex) stays in `operatorMessage()` and never reaches the wire.

**Algorithm disagreement is refused, not resolved.** Core never re-derives a digest from the document, so two digests under different algorithms are simply incomparable and the step fails.

**`DocumentDigest` copies in and copies out.** The authorized digest is a security boundary, so the record clones the array on construction and returns a clone from `value()`. A caller that reuses its buffer cannot move what was authorized after the fact, and `equals`/`hashCode` stay stable. This follows `CertificateChain`, which snapshots its chain in the same way.

**Multi-document pairs positionally.** `verifyAll` checks each response against the authorized digest at the same position and refuses lists of different lengths. A set of echoes that matches as a *set* is not a binding — the swapped-echo test pins this. The size-mismatch refusal is an `IllegalArgumentException` on purpose: mis-pairing the lists is a Core caller bug, and a generic 500 is the right answer for a bug.

**Failure classification.** A genuine digest mismatch is its own failure value, `BINDING_VIOLATION` — it is the security event DP-7 exists for and must stay distinguishable from an ordinary fault for records and alerting, and the exhaustive `TspErrorMapper` switch forces every protocol edge to map it consciously (the TSP edge never calls the verifier, so it maps to the same server-fault answer). The broken-echo cases stay `CONNECTOR_FAULT`: those really are a connector not delivering. All are named to step `computeDtbs`.

## No caller yet

Like `TransitionGuard` in #2036, this ships ahead of its consumers (#1974, #1982) — that is the epic's build order, not dead code. Please don't file it as unused. #2037's DoD is being re-scoped so the enforcement proof lands with #1974/#1982, which already carry the gate acceptance criteria.

## Verification

- 23 unit vectors, positive and negative; 100 % line and branch coverage on both new classes (JaCoCo).
- `spotless:check` clean.

## Out of scope

Raw signing needs no echo — the client submits the digest itself — so `signHash` never reaches this verifier.

### DIFF
--- a/src/main/java/com/otilm/core/signing/contentsigning/DtbsBindingVerifier.java
+++ b/src/main/java/com/otilm/core/signing/contentsigning/DtbsBindingVerifier.java
@@ -33,11 +33,14 @@ public final class DtbsBindingVerifier {
      * Verifies the echo a {@code computeDtbs} response carries against the authorized digest.
      *
      * @throws IllegalArgumentException if the authorized digest is not as long as its algorithm produces
-     * @throws SigningEngineException if the echo is missing, unusable or bound to a different document
+     * @throws SigningEngineException if the response or its echo is missing, unusable or bound to a different document
      */
     public static void verify(DocumentDigest authorized, ComputeDtbsResponseDto response)
             throws SigningEngineException {
-        Objects.requireNonNull(response, "response");
+        requireWellFormedAuthorizedDigest(authorized);
+        if (response == null) {
+            throw brokenEcho("connector returned no computeDtbs response");
+        }
         verifyEcho(authorized, echoOf(response));
     }
 
--- a/src/main/java/com/otilm/core/signing/contentsigning/DtbsBindingVerifier.java
+++ b/src/main/java/com/otilm/core/signing/contentsigning/DtbsBindingVerifier.java
@@ -1,0 +1,110 @@
+package com.otilm.core.signing.contentsigning;
+
+import com.otilm.api.model.connector.signatures.contentsigning.common.ComputeDtbsResponseDto;
+import com.otilm.core.signing.engine.error.SigningEngineException;
+import com.otilm.core.signing.engine.error.SigningEngineFailure;
+import java.security.MessageDigest;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Verifies that the data-to-be-signed a formatting connector produced is bound to the document the caller authorized,
+ * by comparing the connector's echoed {@code documentDigest} against the authorized digest. The engine runs this before
+ * it releases the signing key, so a connector bug or a document mix-up cannot get an unauthorized document signed.
+ *
+ * <p>
+ * The rule is family-blind: PAdES, XAdES, CAdES and JAdES all satisfy it the same way. The platform never parses the
+ * data-to-be-signed and never re-derives the digest from the document — the connector must echo the digest it actually
+ * committed to, and the authorized digest is the one computed once when the operation was accepted. Raw signing needs
+ * no echo at all, because there the client submits the digest itself.
+ * </p>
+ */
+public final class DtbsBindingVerifier {
+
+    private static final String STEP = "computeDtbs";
+
+    private static final String MISMATCH_CLIENT_MESSAGE = "The document the signature would cover is not the document that was authorized";
+
+    private static final String BROKEN_ECHO_CLIENT_MESSAGE = "Internal error during signature formatting";
+
+    private DtbsBindingVerifier() {
+    }
+
+    /**
+     * Verifies the echo a {@code computeDtbs} response carries against the authorized digest.
+     *
+     * @throws SigningEngineException if the echo is missing, unusable or bound to a different document
+     */
+    public static void verify(DocumentDigest authorized, ComputeDtbsResponseDto response)
+            throws SigningEngineException {
+        Objects.requireNonNull(response, "response");
+        verify(authorized, echoOf(response));
+    }
+
+    /**
+     * Verifies an echoed digest against the authorized digest.
+     *
+     * @param echoed the digest the connector committed to, or {@code null} when it echoed none
+     * @throws SigningEngineException if the echo is missing, unusable or bound to a different document
+     */
+    public static void verify(DocumentDigest authorized, DocumentDigest echoed) throws SigningEngineException {
+        Objects.requireNonNull(authorized, "authorized");
+        if (echoed == null) {
+            throw brokenEcho("connector echoed no documentDigest");
+        }
+        requireUsableEcho(authorized, echoed);
+        if (!MessageDigest.isEqual(authorized.value(), echoed.value())) {
+            throw SigningEngineException
+                    .stepFailed(SigningEngineFailure.BINDING_VIOLATION, STEP,
+                            "connector committed to %s but %s was authorized".formatted(echoed, authorized), null,
+                            MISMATCH_CLIENT_MESSAGE);
+        }
+    }
+
+    /**
+     * Verifies a multi-document operation, pairing each response with the authorized digest at the same position. Every
+     * document is checked against its own authorized occurrence, so one document's echo can never vouch for another.
+     *
+     * @throws IllegalArgumentException if the two lists differ in size, which means the caller paired them wrong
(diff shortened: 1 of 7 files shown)


src/main/java/com/otilm/core/signing/contentsigning/DtbsBindingVerifier.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Verifies the echo a {@code computeDtbs} response carries against the authorized digest.
+     *
+     * @throws SigningEngineException if the echo is missing, unusable or bound to a different document
+     */
+    public static void verify(DocumentDigest authorized, ComputeDtbsResponseDto response)
+            throws SigningEngineException {
+        Objects.requireNonNull(response, "response");
+        verify(authorized, echoOf(response));
+    }
```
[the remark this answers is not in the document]
ivosh: Fixed in 0a16f647.

klaraf755 (commented):
Automated code review (high effort): 2 minor findings, neither blocking. Everything that usually goes wrong in this kind of code checks out — constant-time comparison, defensive copies in both directions, consistent equals/hashCode, no digest material in client messages, and the `@EnumSource` rewrite of the TSP mapper test is a nice touch.

src/main/java/com/otilm/core/signing/contentsigning/DtbsBindingVerifier.java
```diff
@@ -0,0 +1,130 @@
...
+     */
+    private static void requireWellFormedAuthorizedDigest(DocumentDigest authorized) {
+        Objects.requireNonNull(authorized, "authorized");
+        if (!authorized.hasLengthOfItsAlgorithm()) {
+            throw new IllegalArgumentException("Authorized digest is %d bytes, which %s never produces"
+                    .formatted(authorized.length(), authorized.algorithm().getCode()));
+        }
+    }
+
+    private static DocumentDigest echoOf(ComputeDtbsResponseDto response) {
```
klaraf755: **[diagnostics — minor]** A half-present echo collapses into "connector echoed no documentDigest", which misstates what the connector actually sent. A connector that supplies `documentDigest` but omits `documentDigestAlgorithm` gets an operator log line saying it echoed no digest — sending whoever debugs the connector to look at the field that is actually populated. The `CONNECTOR_FAULT` classification is right, and `treatsAResponseMissingEitherHalfOfTheEchoAsNoEcho` pins the current message deliberately — but distinguishing the three cases (no digest / no algorithm / neither) in the operator message costs two lines and names the actual defect.

src/main/java/com/otilm/core/signing/contentsigning/DtbsBindingVerifier.java
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Verifies the echo a {@code computeDtbs} response carries against the authorized digest.
+     *
+     * @throws IllegalArgumentException if the authorized digest is not as long as its algorithm produces
+     * @throws SigningEngineException if the response or its echo is missing, unusable or bound to a different document
+     */
+    public static void verify(DocumentDigest authorized, ComputeDtbsResponseDto response)
+            throws SigningEngineException {
+        requireWellFormedAuthorizedDigest(authorized);
```
klaraf755: **[simplification — minor]** `verify()` validates the authorized digest here, then delegates to `verifyEcho`, which runs `requireWellFormedAuthorizedDigest` again — so every `verify()` call validates twice. Harmless at runtime, but the next editor changing the well-formedness rule has to notice both sites. Dropping this call keeps one authority; the only observable change is that a malformed authorized digest combined with a null response reports the null response first (both still failures).

src/main/java/com/otilm/core/signing/contentsigning/DtbsBindingVerifier.java
ivosh: Fixed in d35651e2. `requireCompleteEcho` now distinguishes the three cases: neither half present keeps "connector echoed no documentDigest", a missing digest reports "connector echoed a SHA-256 documentDigestAlgorithm but no documentDigest", and a missing algorithm reports "connector echoed a 32-byte documentDigest but no documentDigestAlgorithm". All three stay CONNECTOR_FAULT with the same client message. `treatsAResponseMissingEitherHalfOfTheEchoAsNoEcho` is split into three tests, one per case.

src/main/java/com/otilm/core/signing/contentsigning/DtbsBindingVerifier.java
ivosh: Fixed in d35651e2. Dropped the call from `verify()`, leaving `verifyEcho` as the single authority. As you noted, the one observable change is ordering: a malformed authorized digest combined with a null response now reports the null response first.

klaraf755 (approved):

PR #2090, opened 2026-08-17
Bind the shared connector mocks to the loopback address

### LINKED ISSUE #1958: Test mock servers bind the wildcard address, so a dynamic port can collide with another local process
**TL;DR: 63 test-local mock servers still bind a dynamic port on the wildcard address. The OS can hand out a port another local process already holds on loopback, and the test's traffic then reaches that process. `WireMockPorts` already established the remedy — finish applying it.**

## Observed failure

During a full-suite run on `chore/rest-template` (PR #1955), `OAuth2UtilUserInfoTest.testGetUserInfo_DoesNotRetryOnServiceUnavailable` failed. The test stubs `503` on its own WireMock server and asserts `HttpServerErrorException`. What it got:

```
Unexpected exception type thrown,
  expected: <org.springframework.web.client.HttpServerErrorException>
  but was:  <org.springframework.web.client.HttpClientErrorException$Unauthorized>

Caused by: HttpClientErrorException$Unauthorized: 401 Unauthorized:
  "MCP server is running in restricted mode. Please, provide valid authorization token"
```

The request never reached WireMock. It reached an unrelated MCP server running on the same machine. The same test passed in four other full-suite runs — it is a genuine flake, not a logic error.

## Mechanism

`new WireMockServer(0)` binds `0.0.0.0`, the IPv4 wildcard. When the OS picks an ephemeral port it does not consider ports already bound on the *more specific* `127.0.0.1`, and Jetty sets `SO_REUSEADDR`, so the bind succeeds anyway. A subsequent connect to that port is routed to the more specific bind — the other process — and the mock server silently never sees the request.

Developer machines are full of loopback listeners in the ephemeral range. On the machine that produced the failure:

```
$ lsof -nP -iTCP -sTCP:LISTEN
idea       127.0.0.1:52364, 127.0.0.1:52404, 127.0.0.1:52418, 127.0.0.1:52434, ...
github     127.0.0.1:52776, 127.0.0.1:52777
embedding  127.0.0.1:52412
cef_serve  127.0.0.1:52315
```

The failure mode is silent and misleading: the assertion fails against a *plausible-looking* HTTP response from a foreign process, so it reads as a product bug rather than an environment collision.

## Scope

- **58** `new WireMockServer(0)`
- **5** `options().port(0)` / `dynamicPort()`
- **19** files already converted to `WireMockPorts`

## Suggested fix

`WireMockPorts` (added to `main` in the `NotificationObjectData` work) already names the remedy in its own Javadoc:

> a fixed port below the 49152 ephemeral floor is also immune to the collisions `dynamicPort()` suffers when another process on the machine happens to hold the port it drew

So the direction is settled — this is about finishing the job:

1. Extend `WireMockPorts` with constants for the remaining stub roles.
2. Convert the 63 dynamic-port sites, keeping every port below 49152.
3. Consider an ArchUnit rule (the repo already has `src/test/java/com/otilm/core/architecture/`) failing any `new WireMockServer` that is not given a `WireMockPorts` constant, so the pattern cannot come back.

Note that the constants rely on Surefire running test classes sequentially in one fork, which `WireMockPorts` documents; that assumption should be stated wherever the ArchUnit rule lives too.

## Why a separate PR

Found while working PR #1955. A loopback-binding fix was drafted there and deliberately reverted: it touched only 3 files, used a different remedy than the `WireMockPorts` convention `main` had meanwhile adopted, and a partial fix to a systemic problem is worse than none.

ivosh:
`new WireMockServer(0)` binds the IPv4 wildcard. The OS picks an ephemeral port without considering ports already bound on the more specific loopback address, and Jetty sets SO_REUSEADDR, so the bind succeeds anyway — a later connect is then routed to the foreign process holding 127.0.0.1 on that port, and the stub never sees the request. The failure surfaces as a plausible-looking HTTP response from an unrelated local listener, so it reads as a product bug rather than an environment collision.

Naming the bind address fixes it. `options().bindAddress("127.0.0.1").dynamicPort()` makes the OS skip every port already taken on loopback, so the port stays dynamic and the stub still gets the request. `LoopbackWireMock` centralises that configuration and the rationale; the connector mocks and the three test bases that build their own stub reach it through `start()` and `url()`.

Stub URLs use the literal `127.0.0.1` rather than `localhost`. That name resolves to both `127.0.0.1` and `::1`, and a client picking the IPv6 form cannot reach an IPv4-bound server.

`WireMockPorts` keeps the three infrastructure stubs whose URL has to exist before the Spring context does — auth-service, scheduler and provisioning-api, each injected as a base URL — and its doc now states that as the test for adding a constant. Every stub converted here hands its URL to the code under test at runtime, so none of them needs one.

`SigningProfileServiceImplITest` guards its `@AfterEach` stops against a null mock, so a setup that aborts part-way surfaces its own failure instead of an NPE.

No `@TestPropertySource` or `@DynamicPropertySource` changes, so the context signature count is untouched and `ContextSignatureGuardTest.BASELINE` stays at 57.

Verified against a live collision: with another local process holding `127.0.0.1:63342`, a wildcard bind on that port succeeded and the probe was answered by that process, while the loopback bind on it was refused. Twenty-five consecutive loopback-bound dynamic starts each landed on a free port and were reached. `SigningProfileServiceImplITest`, `ContentSigningEngineITest`, the TSA/TSP controllers, the CMP handlers, the compliance and signing-profile bases, `WireMockPortsGuardTest` and `ContextSignatureGuardTest` all pass.

Refs OmniTrustILM/core#1958


src/test/java/com/otilm/core/util/mocks/BaseConnectorMock.java
```diff
@@ -29,17 +29,17 @@ public abstract class BaseConnectorMock {
 
     protected final WireMockServer server;
 
-    protected BaseConnectorMock() {
-        this.server = new WireMockServer(0);
+    protected BaseConnectorMock(int port) {
+        this.server = new WireMockServer(port);
         this.server.start();
     }
```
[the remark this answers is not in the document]
ivosh: Fixed in 1e4a53eb.

src/test/java/com/otilm/core/service/cmp/CmpTestUtil.java
```diff
@@ -95,9 +96,13 @@
...
 public class CmpTestUtil {
 
+    /**
+     * Starts a cryptography-provider stub that signs with the platform key, on {@link WireMockPorts#CONNECTOR}. The
+     * caller must stop the returned server before either factory runs again, since both share that port.
+     */
     public static WireMockServer createSigningPlatform() {
         // prepare mock server - for cryptographic provider
-        WireMockServer mockServer = new WireMockServer(0);
+        WireMockServer mockServer = new WireMockServer(WireMockPorts.CONNECTOR);
```
vyskocilm: Hi,

first of all I am not a Java developer, so I know very little about a local customs here. So take my comment with a grain of salt. While I understand the problem, the solution of hardcoding the ports breaks a parallel execution of unit tests for `core`. Which seems to be unwanted side-effect.

The following was recommended by GPT 5.6 Sol model, however I can't really judge it myself if this is correct. But two independent review agreed.

Key refiments:

   •  bindAddress("127.0.0.1").dynamicPort()  is valid WireMock 3.13.2 API and addresses the wildcard-vs-loopback issue.
   •  getUrl()  should return  http://127.0.0.1:<port> , not  localhost , to avoid IPv6/IPv4 mismatch.
   • A singleton port number allocator is risky; a singleton running server can be valid for Spring context-wide resources, but these 
     connector mocks pass URLs at runtime, so separate dynamic mock instances are simpler.
   • Fixed ports are only a fallback for services whose URL truly must be known before Spring context creation; even there, 
      @DynamicPropertySource  can often avoid fixed ports.

vyskocilm: BTW: suggestions talks about singletons because that was what I suggested as a solution first.

ivosh: Fixed in 2f28d96e — you were right, and the grain of salt wasn't needed.

I checked each point against the code and the runtime:

- **`bindAddress("127.0.0.1").dynamicPort()` is valid in 3.13.2** — confirmed both methods on `WireMockConfiguration` in the resolved jar.
- **It addresses the root cause.** The wildcard bind was the actual defect, not the dynamic port. I reproduced the collision live: with IntelliJ holding `127.0.0.1:63342`, a wildcard bind on that same port succeeded and the probe was answered by `Server: IntelliJ IDEA 2026.2.0.1` — precisely the "plausible-looking response from an unrelated listener" the PR described. The loopback bind on that port was refused instead, so the OS can never hand it to us. 25 consecutive loopback-bound dynamic starts each landed free and were reached.
- **`getUrl()` now returns `http://127.0.0.1:<port>`.** `localhost` resolves to both `127.0.0.1` and `::1` here; it happens to work because IPv4 sorts first, which is resolution-order luck rather than a guarantee against an IPv4-bound server.
- **"Fixed ports are only a fallback for URLs needed before context creation"** — this was the decisive one. Every mock this PR touched passes its URL at runtime (`withUrl(mock.getUrl())`, `connector.setUrl(...)`), so none qualified. `WireMockPorts` now keeps only the three infrastructure stubs that are injected as a base URL before the context exists, and its doc states that as the test.
- **No singleton allocator** — separate dynamic instances, as you suggested.

One correction in the other direction: parallel execution was already disabled on `main` before this PR (`junit.jupiter.execution.parallel.enabled=false`, with a comment citing `WireMockPorts`), and CI's four workers are separate runner VMs. So the fixed ports deepened an existing constraint rather than introducing one. That doesn't change the conclusion — your approach removes the coupling instead of adding to it.

It also deleted complexity that only existed to serve fixed ports: `CONTENT_SIGNING_FORMATTING_SECONDARY` and `startSecondContentSigningFormatting()` were there solely because two same-kind mocks couldn't share a port. Net −6 lines of constants and one fewer factory method.

vyskocilm (approved):

PR #2092, opened 2026-08-18
Serve request, renew and identify attribute schemas from the authority connector

### LINKED ISSUE #1943: Serve request, renew and identify attribute schemas from the authority connector
### Description

Implement the three new attribute schemas in Core, defensively.

**Adapter layer.** Add `listCertificateRequestAttributes`, `listRenewAttributes` and `listIdentifyAttributes` to `AuthorityProviderAdapter`. `AuthorityProviderV3Adapter` calls the new connector endpoints and passes each response through the existing `contained(...)` secret-echo guard, exactly as `listIssueAttributes` / `listRevokeAttributes` / `listRegisterAttributes` do. `AuthorityProviderV2Adapter` returns an empty list — the v2 contract has no such endpoints — so no call site needs to branch on interface version.

**Defensive resolution in one place.** One helper turns "connector does not offer this schema" into an empty list: HTTP 404, `ErrorCode.OPERATION_NOT_SUPPORTED` (501) and an empty body all resolve the same way, logged at debug. It must not swallow real failures — auth errors, other 5xx, connector communication failures, timeouts and malformed bodies keep surfacing as connector errors, because a silently empty schema on a connector that *does* have one would drop attributes the CA requires.

**Core API.** Implement the renew and identify counterparts added to the contract — client API `GET /v2/operations/authorities/{authorityUuid}/raProfiles/{raProfileUuid}/attributes/{renew,identify}` and web API `GET /v1/authorities/{authorityUuid}/raProfiles/{raProfileUuid}/attributes/{renew,identify}` — returning the resolved schema.

**Operation flows — simple wiring only.** Renew and identify accept their attribute values, validate them structurally against the listed definitions, and put them on the wire: `CertificateRenewRequestDtoV3.attributes` (now sourced from the renew schema, not the issue schema) and the new `CertificateIdentificationRequestDtoV3.attributes`. Values persist where the existing operations persist theirs.

`listCertificateRequestAttributes` is implemented on the adapter but **not consumed by any flow in this issue** — `GET /v1/certificates/csr/attributes` keeps its current platform-only resolution. Wiring it in, merge modes and value-source bindings are the container under OmniTrustILM/ilm#278.

### Acceptance Criteria

- [ ] Three adapter methods added and implemented for v3; v2 returns empty for all three.
- [ ] One shared not-supported-to-empty helper; unit-tested against 404, `OPERATION_NOT_SUPPORTED` (501), an empty array, and a populated array.
- [ ] Real connector failures (401/403, other 5xx, timeout, malformed body) still propagate as connector errors — covered by test.
- [ ] `contained(...)` applied to all three new responses.
- [ ] Core API endpoints implemented for renew and identify (client + web), returning the resolved schema.
- [ ] Renew and identify send their attribute values on the wire; round-trip covered by test.
- [ ] `listCertificateRequestAttributes` implemented and unit-tested on both adapters, with no behaviour change to `GET /v1/certificates/csr/attributes` (regression test).
- [ ] `mvn -B verify` passes including integration tests; new-lines coverage meets the Sonar gate.

klaraf755:
Closes https://github.com/OmniTrustILM/core/issues/1943

Implements the three new Authority v3 attribute schemas defensively. The adapter contract gains `listCertificateRequestAttributes`, `listRenewAttributes` and `listIdentifyAttributes`; the v3 adapter resolves "connector does not offer this schema" — 404, `OPERATION_NOT_SUPPORTED`, or a bare 501 — to an empty schema in one shared helper, while genuine failures (auth, other 5xx, timeouts, malformed bodies) propagate. All three responses pass the `contained(...)` secret-echo guard. The v2 adapter returns empty for all three, so no call site branches on interface version.

Renew and identify get their client-API and web-API listing endpoints. Values travel the register pattern: validated against the connector schema, persisted on the certificate (renew/rekey under `CERTIFICATE_RENEW` on the successor; identify under `CERTIFICATE_IDENTIFY`, replaced on every identification), and read back by the adapter onto the wire — round-trip proven by ITest. Both value sets surface on the certificate detail.

`listCertificateRequestAttributes` is implemented but consumed by no flow; `GET /v1/certificates/csr/attributes` keeps platform-only resolution, pinned by a regression test. With the new endpoints unstubbed (404 everywhere), the full existing suite is the 2.19.x-parity proof: `mvn verify` — 4861 tests, 0 failures. `ContextSignatureGuardTest` baseline unchanged.

Requires the paired interfaces PR https://github.com/OmniTrustILM/interfaces/pull/880 to merge and republish the 2.20.0-SNAPSHOT first — this branch compiles against it.


src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -2317,6 +2351,17 @@ public void rekeyCertificateAction(final UUID certificateUuid, ClientCertificate
...
+            // the renew schema and persist under CERTIFICATE_RENEW on the successor, where the adapter reads them.
+            extendedAttributeService.mergeAndValidateRenewAttributes(raProfile, request.getAttributes());
+            if (request.getAttributes() != null && !request.getAttributes().isEmpty()) {
+                attributeEngine
+                        .updateObjectDataAttributesContent(ObjectAttributeContentInfo
+                                .builder(Resource.CERTIFICATE, certificate.getUuid())
+                                .connector(raProfile.getAuthorityInstanceReference().getConnectorUuid())
+                                .operation(AttributeOperation.CERTIFICATE_RENEW)
+                                .build(), request.getAttributes());
+            }
```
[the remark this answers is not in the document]
klaraf755: There is no retry path for renew/rekey: the actions admit only REQUESTED/PENDING_APPROVAL successors and the first execution moves the certificate out of both, their sole caller is the actions listener replaying an immutable queued message, and every new renewal attempt creates a fresh successor with an empty attribute slot. The unconditional persist on identify exists because identify writes to a long-lived certificate through repeatable flows — that situation does not arise here.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -217,6 +224,76 @@ public List<BaseAttribute> listRevokeAttributes(AuthorityInstanceReference autho
...
+     * Resolves "connector does not offer this schema" to an empty list, keyed on status codes because connectors are
+     * polyglot and a problem body is not guaranteed: HTTP 404 (endpoint absent), ErrorCode.OPERATION_NOT_SUPPORTED, and
+     * a bare 501 all mean the optional attribute-schema endpoint is not provided. Everything else — auth failures,
+     * other 5xx, communication failures, malformed bodies — propagates: silently emptying a schema the connector does
+     * have would drop attributes the CA requires.
+     */
+    private List<BaseAttribute> schemaOrEmpty(SchemaCall call, String operation) throws ConnectorException {
+        try {
+            return call.get();
+        } catch (ConnectorEntityNotFoundException e) {
```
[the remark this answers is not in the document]
klaraf755: Added the handling for null response.

src/test/java/com/otilm/core/service/handler/authority/AuthorityProviderV3AdapterTest.java
```diff
@@ -886,6 +923,95 @@ void listRegisterAttributesDelegates() throws ConnectorException {
...
+        when(certClientV3.listRequestAttributes(eq(connectorInfo), any())).thenReturn(expected);
+        assertEquals(expected, adapter.listCertificateRequestAttributes(authority, raProfile));
+    }
+
+    @Test
+    void listRenewAttributesDelegates() throws ConnectorException {
+        List<BaseAttribute> expected = List.of(mock(BaseAttribute.class));
+        when(certClientV3.listRenewAttributes(eq(connectorInfo), any())).thenReturn(expected);
+        assertEquals(expected, adapter.listRenewAttributes(authority, raProfile));
+    }
```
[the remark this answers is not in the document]
klaraf755: Added the test case along with null response resolution.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -2032,6 +2054,18 @@ public void renewCertificateAction(final UUID certificateUuid, ClientCertificate
...
+            // successor (renew + connector), where the adapter reads them onto the wire.
+            extendedAttributeService.mergeAndValidateRenewAttributes(raProfile, request.getAttributes());
+            if (request.getAttributes() != null && !request.getAttributes().isEmpty()) {
+                attributeEngine
+                        .updateObjectDataAttributesContent(ObjectAttributeContentInfo
+                                .builder(Resource.CERTIFICATE, certificate.getUuid())
+                                .connector(raProfile.getAuthorityInstanceReference().getConnectorUuid())
+                                .operation(AttributeOperation.CERTIFICATE_RENEW)
+                                .build(), request.getAttributes());
+            }
```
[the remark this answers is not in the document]
klaraf755: Same as https://github.com/OmniTrustILM/core/pull/2092#discussion_r3801705328.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -3129,6 +3172,18 @@ private List<MetadataAttribute> identifyByNewAuthority(Certificate certificate,
             throw new CertificateOperationException(String
                     .format("Cannot switch RA profile for certificate. Certificate not identified by authority of new RA profile %s. Certificate: %s",
                             newRaProfile.getName(), certificate.toStringShort()));
+        } catch (AttributeException | NotFoundException e) {
```
ivosh: **~ Minor** | Error handling

CertificateServiceImpl.java:3175 — the new `catch (AttributeException | NotFoundException e)` branch is described as "The supplied identify attributes failed structural validation against the connector's schema", but the two commonest schema failures never reach it. `mergeAndValidateIdentifyAttributes` (CertificateServiceImpl.java:3152) delegates to `AttributeEngine.validateUpdateDataAttributes` → `validateDataAttributesContent`, which throws **`ValidationException`** at AttributeEngine.java:781 both for a missing required attribute ("Content for required … attribute … is not provided.") and for a supplied attribute whose definition is unknown ("Content for attribute {} is provided but definition is not found"). `ValidationException extends RuntimeException` and is unrelated to `AttributeException`, so those failures fall through to the *next* branch, `catch (ValidationException e)` at CertificateServiceImpl.java:3187, and are written to certificate event history and surfaced to the operator as "Identification by authority of new RA profile %s **rejected the certificate**" — attributing a platform-side input rejection to a connector that was never called. `mergeAndValidateIdentifyAttributes` also raises `ValidationException` when the authority has no connector (ExtendedAttributeServiceImpl.java:188), landing in the same misleading branch.

 The same shape exists on the manual-issue path: ClientOperationServiceImpl.java:3288 catches `ValidationException` *before* the new `AttributeException` catch at :3293, so a missing required identify attribute is reported as "Manual upload rejected by connector identify". The new branches are not dead (a definition-update failure genuinely throws `AttributeException`), they just miss the main case. 

Fix: either run `mergeAndValidate*` outside the try that wraps the connector call, or catch `ValidationException` around the validate step separately so schema failures keep the "rejected by schema" wording.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -2032,6 +2054,18 @@ public void renewCertificateAction(final UUID certificateUuid, ClientCertificate
         additionalInformation.put("New Certificate UUID", certificate.getUuid());
         boolean connectorAccepted = false;
         try {
+            // Validate the connector's renew-operation attributes against the schema — always, so a required renew
+            // attribute is enforced even when the client sends none — then persist the supplied ones on the
+            // successor (renew + connector), where the adapter reads them onto the wire.
+            extendedAttributeService.mergeAndValidateRenewAttributes(raProfile, request.getAttributes());
```
ivosh: **! Major** | Correctness

ClientOperationServiceImpl.java:2060 (and the rekey twin at :2356) call `mergeAndValidateRenewAttributes` inside `renewCertificateAction` / `rekeyCertificateAction`. Those are the **asynchronous** action bodies, consumed from `ActionsListener` (ActionsListener.java:173) after the synchronous entry point has already returned.

The synchronous entry `renewCertificate` (ClientOperationServiceImpl.java:1951) never touches `request.getAttributes()`: it builds the CSR, calls `submitAndShapeFailure` to create the successor certificate, and returns a 200 with the successor UUID, then produces the ActionMessage. So a client that posts a renew attribute the connector schema does not declare gets **HTTP 200 plus a real successor certificate row**, and only later a `ValidationException` ("Content for attribute X is provided but definition is not found", AttributeEngine.java:2376) that strands the successor in a failed state. The caller never sees a 4xx.

The two established neighbours both validate *before* enqueuing the action:
- revoke: `mergeAndValidateRevokeAttributes` at ClientOperationServiceImpl.java:2483, on the synchronous `revokeCertificate`, immediately before the `ActionMessage` is produced — identical approval/ActionMessage shape to renew.
- issue: `mergeAndValidateIssueAttributes` at ClientOperationServiceImpl.java:482, inside the synchronous `submitCertificateRequest`, before the persistence transaction opens.

If synchronous validation is intended and this is an oversight, move (or duplicate) the `mergeAndValidateRenewAttributes` call into `renewCertificate` / `rekeyCertificate` before `actionProducer.produceMessage(...)`, mirroring `revokeCertificate`. If validation is deliberately deferred to the action — e.g. because the schema must be re-read after approval — then explain the rationale why the new `attributes` field on `ClientCertificateRenewRequestDto`/`ClientCertificateRekeyRequestDto` then has a different error contract from every other attribute-bearing operation on this API.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -3115,9 +3145,22 @@ private List<MetadataAttribute> identifyByNewAuthority(Certificate certificate,
                             newRaProfile.getName(), certificate.toStringShort()));
         }
         try {
+            // Validate the connector's identify-operation attributes against the schema — always, so a required
+            // identify attribute is enforced even when the operator sends none — then persist them unconditionally:
+            // the write replaces this (certificate, connector) slot, so re-identifying against a connector the
+            // certificate was assigned to before cannot resurface a previous stint's values.
+            extendedAttributeService.mergeAndValidateIdentifyAttributes(newRaProfile, identifyAttributes);
+            attributeEngine
```
ivosh: **~ Minor** | Correctness

CertificateServiceImpl.java:3153 persists the identify values *before* the connector call at :3161. Every failure from there raises the **checked** `CertificateOperationException`; the class is annotated plain `@Transactional` (CertificateServiceImpl.java:242, no `rollbackFor`), and the method's own Javadoc at :3129 states the intent — "the caller's transaction survives". Spring's default rules do not roll back on a checked exception, so on the single-certificate path (`updateCertificateObjects` → `switchRaProfile`, :896) the `CERTIFICATE_IDENTIFY` content is committed against the **new** RA profile's connector while `certificate.setRaProfile(newRaProfile)` at :3091 never runs. The bulk path behaves the opposite way: `bulkUpdateCertificatesObjects` wraps each certificate in its own explicit transaction and rolls it back on any `Exception` (:1017-1022), so the same failure discards the write there.

Two readings: 
(a) leaving the values is acceptable — they are invisible (the detail read at :718-724 is keyed on the certificate's *current* connector) and are replaced wholesale by the next successful switch to that connector, since `updateObjectDataAttributesContent` deletes the (object, connector, operation) slot first — in which case add a one-line comment saying the write deliberately survives a failed switch, and note the divergence between the single and bulk paths; 
(b) otherwise move the persist to after `identify` returns. Nothing on this path needs it earlier: unlike renew, the adapter receives the values as a parameter (:3162) rather than reading them back from the engine.

Related, same lines: the write now sits inside the ambient transaction that spans two connector round-trips (`listIdentifyAttributes` inside the merge, then `identify`), so its row locks are held for the full connector latency — the case CLAUDE.md's "Transactions and external calls" warns about.

src/main/java/com/otilm/core/evaluator/CertificateTriggerEvaluator.java
```diff
@@ -77,7 +77,8 @@ protected void performSetFieldPropertyExecution(String fieldIdentifier, Object a
                 ? null
                 : SecuredUUID.fromUUID(propertyUuid != null ? propertyUuid : propertyUuids.getFirst());
         switch (searchableField) {
-            case RA_PROFILE_NAME -> certificateService.switchRaProfile(certificateUuid, newPropertyUuid);
+            // Trigger-driven switches carry no operator input, so no identify attributes are supplied.
+            case RA_PROFILE_NAME -> certificateService.switchRaProfile(certificateUuid, newPropertyUuid, List.of());
```
ivosh, line 81: **~ Minor** | Documentation

CertificateTriggerEvaluator.java:81 passes `List.of()` into `switchRaProfile`, and `mergeAndValidateIdentifyAttributes` (CertificateServiceImpl.java:3152) now runs on **every** RA-profile switch, not only operator-driven ones. The same is true on the renew side: `renewCertificateAction` validates the renew schema unconditionally (ClientOperationServiceImpl.java:2060), and two of its callers cannot supply attributes at all — `LocationServiceImpl.java:1154` builds a bare `ClientCertificateRenewRequestDto` for location-driven renewal, and `ActionsListener.java:170` replays whatever the trigger stored.

The validator enforces required definitions even for an empty submission (`AttributeEngine.validateAttributesContent`, AttributeEngine.java:2390: "Content for required {} attribute {} is not provided."). So the first v3 connector that marks a renew or identify attribute `required` silently converts these automated paths from working to permanently failing — trigger-driven RA-profile switches raise `CertificateOperationException`, location renewals raise `LocationException`.

Please confirm that enforcing required attributes on unattended flows is the intent. If it is not the intent, the alternative is to skip (or only warn on) the required check when the caller supplied no attributes at all — e.g. an explicit `unattended` flag on the merge call.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
ivosh: **~ Minor** | Correctness

ClientOperationServiceImpl.java:2060 (renew), :2356 (rekey), :3277 (manual issue / identify) and CertificateServiceImpl.java:3152 (RA-profile switch) run merge-and-validate and then persist the submitted values under `CERTIFICATE_RENEW` / `CERTIFICATE_IDENTIFY`, and CertificateServiceImpl.java:707/717 reads them back into the new `CertificateDetailDto.renewAttributes` / `identifyAttributes` fields. On a **v2** authority those values can never be delivered: `AuthorityProviderV2Adapter.renew` (AuthorityProviderV2Adapter.java:80-95) builds a `CertificateRenewRequestDto` that has no `attributes` property at all, and `identify` (AuthorityProviderV2Adapter.java:121-135) documents in-line that it drops them. So the API can report back, on `GET /v1/certificates/{uuid}`, a set of renew/identify attributes that the CA never saw — the detail response says they were applied, the wire says otherwise.

It is a narrow path, not a common one: `AuthorityProviderV2Adapter.listRenewAttributes` / `listIdentifyAttributes` return `List.of()`, so the published schema is empty and `AttributeEngine.validateDataAttributesContent` (AttributeEngine.java:2373-2378) rejects any value whose definition it cannot find. It is still reachable, because that method first falls back to `findByTypeAndConnectorUuidAndAttributeUuidAndName` (AttributeEngine.java:756-762) and will accept — and silently re-stamp the definition's `operation` — any value whose uuid+name matches an *existing* DATA definition of the same connector, e.g. one of that authority's issue attributes.

Note the neighbour pattern the register flow already established for exactly this situation: `ExtendedAttributeServiceImpl.mergeAndValidateRegisterAttributes` (ExtendedAttributeServiceImpl.java:205-212) short-circuits with `if (!(adapter instanceof RegisterCapability ...)) return;` before it ever validates or persists, so a non-capable authority cannot accumulate values it cannot send. The new `mergeAndValidateRenewAttributes` (:167) and `mergeAndValidateIdentifyAttributes` (:185) have no equivalent guard.

If accepting-and-storing on v2 is intended (harmless because the schema is empty), a one-line comment on the two new merge methods saying so would settle it. Otherwise, mirror the register precedent: skip the persist — or reject a non-empty `attributes` — when the resolved adapter cannot carry the values on the wire, so the detail response never claims an attribute the CA never received.

src/test/java/com/otilm/core/integration/service/CsrGenerationAttributesITest.java
```diff
@@ -113,6 +114,18 @@ void withProfileFallsBackToPlatformDefaultSetWhenNothingConfigured() throws Exce
...
+    @Test
+    void csrResolutionStaysPlatformOnly_regressionForConnectorRequestSchema() throws Exception {
+        // Regression pin: the connector-provided certificate-request schema exists on the adapter
+        // (listCertificateRequestAttributes) but is deliberately NOT consumed here — wiring it into this resolution
+        // is a separate planned change. CSR-attribute resolution must not gain any connector consultation.
+        RaProfile raProfile = newEnabledRaProfile();
+
+        certificateService.getCsrGenerationAttributes(SecuredUUID.fromUUID(raProfile.getUuid()));
+
+        verifyNoInteractions(extendedAttributeService);
```
ivosh: **~ Minor** | Testing

CsrGenerationAttributesITest.java:118-127 — this test is the PR description's "pinned by a regression test" for `GET /v1/certificates/csr/attributes` staying platform-only, but as written it cannot fail for the reason it claims.

The fixture is `newEnabledRaProfile()` (CsrGenerationAttributesITest.java:77-82), which saves an RaProfile with **no** `authorityInstanceReference`. The resolution path short-circuits on exactly that: `RaProfileCertificateRequestAttributeServiceImpl.listConnectorIssueAttributes` returns `List.of()` when the authority or its connector is null (RaProfileCertificateRequestAttributeServiceImpl.java:102-108) before `extendedAttributeService` is ever touched. So `verifyNoInteractions(extendedAttributeService)` at line 126 is satisfied by the absence of an authority, not by the absence of a connector-schema consultation.

Consequence: if a later change did wire `listCertificateRequestAttributes` into `IssuanceDefinitionResolver.resolve` / `resolveIssueAttributeSet`, it would almost certainly sit behind the same authority-null guard the sibling call sits behind — and this test would stay green. It also duplicates coverage the neighbouring `withProfileReturnsTheResolvedSetNotTheSeed` (line 98, `verify(extendedAttributeService, never()).listIssueCertificateAttributes(any())`) already provides for the no-authority shape.

Fix: build the pin on the fixture that *can* reach the connector — the authority+connector profile from `withProfileMergesPersistedConnectorSet` (CsrGenerationAttributesITest.java:130-140) — then assert `verify(extendedAttributeService, never()).listCertificateRequestAttributes(...)` (or `verifyNoMoreInteractions` after the issue-set call). Then a regression that adds connector-request-schema consultation actually turns the test red.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -697,6 +704,26 @@ public CertificateDetailDto getCertificate(SecuredUUID uuid)
...
+                            .getObjectDataAttributesContent(ObjectAttributeContentInfo
+                                    .builder(Resource.CERTIFICATE, certificate.getUuid())
+                                    .connector(certificate
+                                            .getRaProfile()
+                                            .getAuthorityInstanceReference()
+                                            .getConnectorUuid())
+                                    .operation(AttributeOperation.CERTIFICATE_RENEW)
+                                    .build()));
+            dto
+                    .setIdentifyAttributes(attributeEngine
```
ivosh, line 718: **~ Minor** | Testing

CertificateServiceImpl.java:718 adds `dto.setIdentifyAttributes(...)` under `CERTIFICATE_IDENTIFY`, and CertificateServiceImpl.java:3153 / ClientOperationServiceImpl.java:3278 add the writes that populate that slot. The PR description states the values are "read back by the adapter onto the wire — round-trip proven by ITest" and "Both value sets surface on the certificate detail."

Only the renew half has that proof. `V3RenewRevokeITest.renew_carriesRenewAttributeValuesOnTheWire` (V3RenewRevokeITest.java:135-187) covers persist → wire → `detail.getRenewAttributes()`. Grepping the whole test tree for `getIdentifyAttributes`, `CERTIFICATE_IDENTIFY` or `mergeAndValidateIdentifyAttributes` returns only `ExtendedAttributeServiceImplTest.java:281-289` — a pure mock-delegation unit test. Nothing exercises the identify persistence, and nothing reads `CertificateDetailDto.getIdentifyAttributes()` anywhere. The adapter-side wire test (`AuthorityProviderV3AdapterTest.identifyPutsAttributesOnTheWire`, line ~477) proves only that whatever the caller hands in is forwarded — it does not prove the service persists to, or reads back from, the `CERTIFICATE_IDENTIFY` slot.

So the untested half is precisely the part the description calls proven. Either add an identify counterpart to `V3RenewRevokeITest.renew_carriesRenewAttributeValuesOnTheWire` (upload/identify with a value, verify it on the identify request body and on `detail.getIdentifyAttributes()`), or soften the description to say the round-trip is proven for renew and unit-covered for identify.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -3129,6 +3172,18 @@ private List<MetadataAttribute> identifyByNewAuthority(Certificate certificate,
             throw new CertificateOperationException(String
                     .format("Cannot switch RA profile for certificate. Certificate not identified by authority of new RA profile %s. Certificate: %s",
                             newRaProfile.getName(), certificate.toStringShort()));
+        } catch (AttributeException | NotFoundException e) {
+            // The supplied identify attributes failed structural validation against the connector's schema.
```
ivosh: **~ Minor** | Documentation

CertificateServiceImpl.java:3175-3186 adds `catch (AttributeException | NotFoundException e)` with the comment "The supplied identify attributes failed structural validation against the connector's schema." That is not where structural validation failures actually land.

`ExtendedAttributeServiceImpl.mergeAndValidateIdentifyAttributes` (ExtendedAttributeServiceImpl.java:197-199) calls `AttributeEngine.validateUpdateDataAttributes`, which declares only `throws AttributeException` (AttributeEngine.java:721-723); the content check it delegates to, `validateDataAttributesContent` (AttributeEngine.java:728-730), throws the **unchecked** `ValidationException`. A missing required identify attribute — the exact case the new code says it enforces ("always, so a required identify attribute is enforced even when the operator sends none", line 3148-3149) — therefore raises `ValidationException` and is caught 12 lines lower at CertificateServiceImpl.java:3187.

That handler still carries the pre-diff comment "A connector may reject identification for any policy it implements … Forward its own reason so the operator sees the cause" (lines 3188-3189), and reports "Identification by authority of new RA profile %s **rejected the certificate**". After this diff that sentence is false for one path: the connector was never contacted, and the operator is told the CA rejected their certificate when Core rejected their attribute values. The same class already treats this distinction as load-bearing — see line 3135, "The connector is never contacted, so this is not a rejection and must not be reported as one." The mirror site has the same shape: ClientOperationServiceImpl.java:3288-3292 reports "Manual upload rejected by connector identify" for what is now also a Core-side schema failure, and its `catch (AttributeException e)` at line 3293 is reachable only for definition-update errors.

Two readings. If routing engine `ValidationException`s into the connector-rejection message is acceptable, say so — update the comments at 3176/3188 (and ClientOperationServiceImpl.java:3260-3263, whose Javadoc still describes the body as only calling identify and surfacing the 422) to state that Core-side attribute validation shares that handler. Otherwise catch `ValidationException` from the `mergeAndValidate…` call specifically and report it with the attribute-rejection wording the new branch already has. Note neither new catch branch has a test. Non-blocking, but the comments should not stay as they are.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -2032,6 +2054,18 @@ public void renewCertificateAction(final UUID certificateUuid, ClientCertificate
         additionalInformation.put("New Certificate UUID", certificate.getUuid());
         boolean connectorAccepted = false;
         try {
+            // Validate the connector's renew-operation attributes against the schema — always, so a required renew
```
ivosh: **. Nitpick** | Maintainability

ClientOperationServiceImpl.java:2057-2068 and ClientOperationServiceImpl.java:2354-2365 are the same eleven lines twice — same `mergeAndValidateRenewAttributes` call, same `if (attributes != null && !isEmpty())` guard, same `updateObjectDataAttributesContent` builder chain with the same `CERTIFICATE_RENEW` operation — differing only in the explanatory comment above them.

It lands in methods that are already long: `renewCertificateAction` spans lines 2033-2187 (~155 lines) and `rekeyCertificateAction` starts at 2330 and runs comparably. CLAUDE.md's "AI-written code: refactor before review" asks for >80-100-line methods with heavy comment overhead to be split into named helpers before human review, and this diff grows both of them.

Fix: extract one private helper next to the other renew plumbing, e.g. `private void validateAndPersistRenewAttributes(RaProfile raProfile, Certificate successor, List<RequestAttribute> attributes)`, holding the block and the shared rationale comment; each call site then reads as a single named line. That also removes the risk of the two copies drifting — note they must stay identical for the adapter read at `AuthorityProviderV3Adapter.renewAttributesFor` (AuthorityProviderV3Adapter.java:531) to see rekey values at all.

src/test/java/com/otilm/core/service/handler/authority/AuthorityProviderV3AdapterTest.java
```diff
@@ -886,6 +923,97 @@ void listRegisterAttributesDelegates() throws ConnectorException {
...
+        when(certClientV3.listRequestAttributes(eq(connectorInfo), any())).thenReturn(expected);
+        assertEquals(expected, adapter.listCertificateRequestAttributes(authority, raProfile));
+    }
+
+    @Test
+    void listRenewAttributesDelegates() throws ConnectorException {
+        List<BaseAttribute> expected = List.of(mock(BaseAttribute.class));
+        when(certClientV3.listRenewAttributes(eq(connectorInfo), any())).thenReturn(expected);
+        assertEquals(expected, adapter.listRenewAttributes(authority, raProfile));
+        when(certClientV3.listRenewAttributes(eq(connectorInfo), any())).thenReturn(null);
```
ivosh, line 950: **. Nitpick** | Testing

Four of the new tests pack multiple independent behaviours into a single method, against sibling tests in the same files that keep one behaviour each.

1. `AuthorityProviderV3AdapterTest.listRenewAttributesDelegates` (lines 936-942) does two act/assert cycles with a re-stub in between: delegation, then null-body → empty. Every neighbouring delegation test — `listIssueAttributesDelegates` (906), `listRevokeAttributesDelegates` (913), `listRegisterAttributesDelegates` (920), `listCertificateRequestAttributesDelegates` (929), `listIdentifyAttributesDelegates` (945) — is a single arrange/act/assert. Split the null case into its own `nullResponseBodyResolvesEmpty`, which also makes it obvious that the same `schemaOrEmpty` null branch is untested for the request and identify entry points (only renew exercises it today).

2. `AuthorityProviderV2AdapterTest.newAttributeSchemas_returnEmptyOnV2_withoutTouchingTheConnector` (lines 439-446) asserts three separate methods. A failure names none of them. Also, `new…` in a test name rots — these will not be new next release; prefer `requestRenewIdentifySchemas_returnEmptyOnV2`.

3. `testListRenewAndIdentifyCertificateAttributes_resolveEmptyOnV2Authority` — the `verbAndOtherVerb` shape — appears twice, at ClientOperationServiceV2ITest.java:310 and RaProfileServiceITest.java:668, each asserting two endpoints.

4. `V3RenewRevokeITest.renew_carriesRenewAttributeValuesOnTheWire` (lines 135-187, ~53 lines) proves the wire carriage and then, at lines 181-186, a second behaviour the name does not mention — that the values surface on the successor's certificate detail. Worth its own test (`renewAttributeValuesSurfaceOnSuccessorDetail`) sharing the fixture.

All take-it-or-leave-it.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -217,6 +224,83 @@ public List<BaseAttribute> listRevokeAttributes(AuthorityInstanceReference autho
...
+        return contained(response, request.getAuthorityAttributes(), request.getRaProfileAttributes());
+    }
+
+    @FunctionalInterface
+    private interface SchemaCall {
+        List<BaseAttribute> get() throws ConnectorException;
+    }
+
+    /**
+     * Resolves "connector does not offer this schema" to an empty list, keyed on status codes because connectors are
```
ivosh: **. Nitpick** | Documentation

AuthorityProviderV3Adapter.java:265-271 — five lines, one paragraph, no lead-ins, fusing four concerns (what it resolves, why the decision is keyed on status codes, which three signals count, what deliberately propagates and why). The opening sentence runs ~44 words and chains a colon, a parenthetical and a three-item list before it lands, so the safety-critical half — "silently emptying a schema the connector does have would drop attributes the CA requires" — arrives last, buried.

The class already shows the better shape: the `contained(...)` Javadoc at lines 357-368 and the field-resolution note at 540-545 use `<p>` breaks and `<b>` lead-ins. Suggested restructure, nothing factual dropped:

```
/**
 * Resolves "connector does not offer this schema" to an empty list.
 * <p>
 * <b>Not-offered signals:</b> HTTP 404 (endpoint absent), {@code OPERATION_NOT_SUPPORTED}, and a bare 501.
 * Keyed on status because connectors are polyglot and a problem body is not guaranteed.
 * <p>
 * <b>Everything else propagates</b> — auth failures, other 5xx, communication failures, malformed bodies.
 * Silently emptying a schema the connector does have would drop attributes the CA requires.
 */
```

The interface-side block at AuthorityProviderAdapter.java:102-107 has the same density: what the schema is, its optionality, the v3 resolution rules, the v2 behaviour and the not-yet-consumed status, all in one paragraph. Same one-concern-per-`<p>` treatment applies.

src/main/java/com/otilm/core/api/web/RAProfileManagementControllerImpl.java
```diff
@@ -345,4 +345,26 @@ public List<BaseAttribute> listIssueCertificateAttributes(String authorityUuid,
                         SecuredUUID.fromString(raProfileUuid));
     }
 
+    @Override
+    @AuditLogged(module = Module.CERTIFICATES, resource = Resource.ATTRIBUTE, name = "renew",
+            affiliatedResource = Resource.RA_PROFILE, operation = Operation.LIST_ATTRIBUTES)
+    public List<BaseAttribute> listRenewCertificateAttributes(String authorityUuid,
```
ivosh: **. Nitpick** | Documentation

`RAProfileManagementControllerImpl.java:351/362` add web-API listings for the renew and identify schemas, and `CertificateServiceImpl.java:707/717` add `renewAttributes` / `identifyAttributes` to the certificate detail — the surfaces an operator UI would need in order to collect these values.

Org-wide `gh search code --owner OmniTrustILM --owner 3KeyCompany` finds no consumer: `renewAttributes` returns zero hits anywhere, and fe-administrator's checked-in generated client (`fe-administrator:src/types/openapi/apis/ClientOperationsV2Api.ts`, `CertificateInventoryApi.ts`) plus its call sites (`src/ducks/certificates-epics.ts`, which drives `updateCertificateObjects` for the RA-profile switch) still know only `listIssueCertificateAttributes` and the pre-change DTOs. Nothing breaks — the additions are additive on the wire — but until fe-administrator regenerates its openapi types and adds the attribute dialogs, renew and identify values can only be supplied by an API client, and the RA-profile switch dialog will keep sending no `attributes`.

The PR body scopes itself to the backend and does not claim any FE work, so this is a follow-up rather than a gap in the disclosure: worth linking a fe-administrator issue from #1943 so the endpoints do not sit unconsumed. Does not block the PR.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapter.java
```diff
@@ -96,6 +99,30 @@ List<BaseAttribute> listIssueAttributes(AuthorityInstanceReference authority, Ra
     List<BaseAttribute> listRevokeAttributes(AuthorityInstanceReference authority, RaProfile raProfile)
             throws ConnectorException;
 
+    /**
+     * Lists the connector's certificate-request attribute schema (subject RDNs, SANs, extensions the CA asks for).
+     * Optional on the connector: v3 resolves 404, {@code OPERATION_NOT_SUPPORTED} (501) and an empty array to an empty
+     * list; v2 has no such endpoint and always returns empty. Implemented but not consumed by any flow yet — wiring it
+     * into {@code GET /v1/certificates/csr/attributes} is a separate planned change.
```
ivosh: **. Nitpick** | Documentation

AuthorityProviderAdapter.java:105-106 ends the `listCertificateRequestAttributes` contract doc with "Implemented but not consumed by any flow yet — wiring it into `GET /v1/certificates/csr/attributes` is a separate planned change." The same forward-looking claim is repeated at CsrGenerationAttributesITest.java:119-121 ("...wiring it into this resolution is a separate planned change").

The first half is a genuine WHY and should stay — it tells a future reader why an interface method has no callers. The second half is roadmap/planning narration: it states an intention that lives in the PR body and the tracking issue, and it goes stale the moment the wiring lands (nothing in the build will fail to remind anyone to delete it).

Suggested trim, keeping the rationale and dropping the plan:

before: `... v2 has no such endpoint and always returns empty. Implemented but not consumed by any flow yet — wiring it into {@code GET /v1/certificates/csr/attributes} is a separate planned change.`
after:  `... v2 has no such endpoint and always returns empty. Not consumed by any flow today; {@code GET /v1/certificates/csr/attributes} resolves CSR attributes platform-side.`

Same trim applies to the test comment: keep "deliberately NOT consumed here" and the pin's purpose, drop "is a separate planned change".

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -3115,9 +3145,22 @@ private List<MetadataAttribute> identifyByNewAuthority(Certificate certificate,
...
+            // Validate the connector's identify-operation attributes against the schema — always, so a required
+            // identify attribute is enforced even when the operator sends none — then persist them unconditionally:
+            // the write replaces this (certificate, connector) slot, so re-identifying against a connector the
+            // certificate was assigned to before cannot resurface a previous stint's values.
+            extendedAttributeService.mergeAndValidateIdentifyAttributes(newRaProfile, identifyAttributes);
+            attributeEngine
+                    .updateObjectDataAttributesContent(ObjectAttributeContentInfo
+                            .builder(Resource.CERTIFICATE, certificate.getUuid())
+                            .connector(newRaProfile.getAuthorityInstanceReference().getConnectorUuid())
+                            .operation(AttributeOperation.CERTIFICATE_IDENTIFY)
```
ivosh: **. Nitpick** | Security

CertificateServiceImpl.java:3153-3160 persists the operator-supplied identify values under `(certificate, newConnector, CERTIFICATE_IDENTIFY)`. Twelve lines later, `switchRaProfile` deliberately purges the *previous* authority's data — but only metadata: CertificateServiceImpl.java:3094-3103 calls `attributeEngine.deleteObjectAttributesContent(AttributeType.META, ...connector(currentRaProfile...getConnectorUuid())...)`. Nothing deletes the previous connector's `AttributeType.DATA` / `CERTIFICATE_IDENTIFY` slot.

Why this matters: the identify schema is connector-defined and arbitrary — a CA may declare an enrolment password, challenge phrase, or agent token as an identify attribute. Those values are stored verbatim by `AttributeEngine.updateObjectDataAttributesContent` (no SECRET redaction in `getResponseAttributes`, AttributeEngine.java:1345). After the certificate moves from authority A to authority B, A's supplied values stay on the certificate row indefinitely, for an authority that no longer governs it. The adjacent META purge shows the author already treats "previous authority's data should not linger on this certificate" as the rule; DATA is the one that can hold operator-typed secrets, and it is the one left behind.

Note the comment at CertificateServiceImpl.java:3150-3151 addresses a *different* risk (resurfacing a previous stint's values on re-identify), which the unconditional write does solve. Retention across a switch-away is not covered by it.

Recommended to extend the purge at :3096 to also drop the old connector's `CERTIFICATE_IDENTIFY` (and `CERTIFICATE_RENEW`) DATA slot alongside META.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -217,6 +224,83 @@ public List<BaseAttribute> listRevokeAttributes(AuthorityInstanceReference autho
...
+     * polyglot and a problem body is not guaranteed: HTTP 404 (endpoint absent), ErrorCode.OPERATION_NOT_SUPPORTED, and
+     * a bare 501 all mean the optional attribute-schema endpoint is not provided. Everything else — auth failures,
+     * other 5xx, communication failures, malformed bodies — propagates: silently emptying a schema the connector does
+     * have would drop attributes the CA requires.
+     */
+    private List<BaseAttribute> schemaOrEmpty(SchemaCall call, String operation) throws ConnectorException {
```
ivosh: **~ Minor** | Security

AuthorityProviderV3Adapter.java:272-300 (`schemaOrEmpty`) is reached from two callers with very different consequences, and the Javadoc only reasons about one of them.

**Read-only listing** — `listRenewAttributes` / `listIdentifyAttributes` called from `RaProfileServiceImpl` and `ClientOperationServiceImpl`: an empty result just means the operator UI shows no fields. Benign, and exactly what the design intends.

**Validation gate** — `ExtendedAttributeServiceImpl.mergeAndValidateRenewAttributes` (ExtendedAttributeServiceImpl.java:162-177) and `mergeAndValidateIdentifyAttributes` (:180-196) source the `definitions` list handed to `attributeEngine.validateUpdateDataAttributes` from the *same* `schemaOrEmpty`-wrapped call. When it resolves empty, the engine validates against an empty definition set, so a **required** renew/identify attribute is not enforced and the operation proceeds to the connector without it. That directly contradicts the intent stated in the call-site comments: "Validate ... always, so a required renew attribute is enforced even when the client sends none" (ClientOperationServiceImpl.java:2057-2059 and :2354-2356; CertificateServiceImpl.java:3148-3149). The claim holds only while the schema call succeeds.

The Javadoc already names this hazard — "silently emptying a schema the connector does have would drop attributes the CA requires" — but excludes only the *other* error classes. HTTP 404 is the case where a connector that does have the schema can still look like it does not: a reverse proxy or gateway rewriting the path, a wrong connector base URL, or a connector rolled back to a build without the route all surface as `ConnectorEntityNotFoundException` at AuthorityProviderV3Adapter.java:282, indistinguishable from "endpoint absent". The only signal an operator gets is `log.debug` (:283-285, :289-292, :296).

No code change is strictly required. 
Suggestions: raise the resolution message on the 404 branch from `debug` to `info`/`warn` so a silently degraded validation gate is visible in ops logs. Add one line to the `schemaOrEmpty` Javadoc stating that an empty resolution also means required-attribute enforcement is off for that operation.

If the fail-open is a deliberate availability choice (never let an optional schema block a renew), saying that explicitly in the same Javadoc is enough.

src/main/java/com/otilm/core/service/v2/impl/ExtendedAttributeServiceImpl.java
```diff
@@ -141,6 +141,64 @@ public List<BaseAttribute> listRegisterCertificateAttributes(RaProfile raProfile
         return List.of();
     }
 
+    @Override
+    public List<BaseAttribute> listRenewCertificateAttributes(RaProfile raProfile)
```
ivosh, line 145: **. Nitpick** | Documentation

ExtendedAttributeServiceImpl.java:145 (`listRenewCertificateAttributes`) and :155 (`listIdentifyCertificateAttributes`) null-check the connector and then go straight to the adapter. The two closest siblings do one more thing first: `listIssueCertificateAttributes` and `listRevokeCertificateAttributes` (:112-122) call `validateLegacyConnector(connector)`, which throws `NotFoundException("Legacy Authority. V2 Implementation not found on the connector")` for a LegacyEjbca-backed connector (:265-273).

So on a legacy authority the two new endpoints — `GET /v1/authorities/{a}/raProfiles/{r}/attributes/renew` and `.../identify`, and their `/v2/operations/...` twins — answer `200 []`, while `.../attributes/issue` and `.../attributes/revoke` on the very same RA profile answer `404`. That is very likely deliberate and even preferable: the published OpenAPI text for both new operations says "An authority that offers no renew schema yields an empty list — this is not an error".

The reason it is worth a line: the third sibling, `listRegisterCertificateAttributes` (:127-140), skips the same guard and carries an explicit comment explaining the skip ("No legacy-connector check here (unlike issue/revoke)..."). The new pair skips it silently, so the next reader cannot tell an intentional divergence from an omission. If intended, copy the register sibling's one-line rationale onto the two new methods; if not, add `validateLegacyConnector(connector)` to match issue/revoke.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -3115,9 +3145,22 @@ private List<MetadataAttribute> identifyByNewAuthority(Certificate certificate,
                             newRaProfile.getName(), certificate.toStringShort()));
         }
         try {
+            // Validate the connector's identify-operation attributes against the schema — always, so a required
+            // identify attribute is enforced even when the operator sends none — then persist them unconditionally:
+            // the write replaces this (certificate, connector) slot, so re-identifying against a connector the
+            // certificate was assigned to before cannot resurface a previous stint's values.
+            extendedAttributeService.mergeAndValidateIdentifyAttributes(newRaProfile, identifyAttributes);
```
ivosh, line 3163: **~ Minor** | Error handling

CertificateServiceImpl.java:3152 calls `mergeAndValidateIdentifyAttributes` inside the same `try` whose new `catch (AttributeException | NotFoundException e)` at :3175 is commented "The supplied identify attributes failed structural validation against the connector's schema."

That catch mostly will not fire for the case it names. The structural content check is `AttributeEngine.validateDataAttributesContent`, which throws `new ValidationException(errors)` (AttributeEngine.java:780) — and `com.otilm.api.exception.ValidationException` extends `RuntimeException`, not `AttributeException` (which is a checked `Exception`). Only the definition-upsert half (`updateDataAttributeDefinitions`) raises `AttributeException`.

So the common failure — the operator submits an identify attribute the connector schema does not declare, or omits a required one — lands in `catch (ValidationException e)` at CertificateServiceImpl.java:3187, whose own comment reads "A connector may reject identification for any policy it implements", and is reported to the operator as "Identification by authority of new RA profile %s rejected the certificate: ...". The connector was never contacted at that point.

Same shape in the manual-issue path: ClientOperationServiceImpl.java:3277 validates, and the `catch (ValidationException e)` at :3288 rewrites the message to "Manual upload rejected by connector identify: ...", while the new `catch (AttributeException e)` at :3293 carries the accurate wording.

If the attribution matters (it is what the operator reads in certificate event history), catch `ValidationException` from the merge call separately — e.g. wrap just the `mergeAndValidateIdentifyAttributes` + persist pair in its own try, or add `ValidationException` to the schema-rejection catch and re-throw the connector-rejection case from inside the adapter call. If the merged wording is acceptable, drop the misleading "failed structural validation" comment at :3175 so the next reader is not misled.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -3226,14 +3271,30 @@ private List<MetadataAttribute> identifyUploadedCertificateOrReject(Certificate
...
                     .forAuthority(raProfile.getAuthorityInstanceReference())
-                    .identify(raProfile, request.getCertificate());
+                    .identify(raProfile, request.getCertificate(),
+                            request.getAttributes() != null ? request.getAttributes() : List.of());
         } catch (ValidationException e) {
             certificateEventHistoryService
                     .addEventHistory(certificate.getUuid(), CertificateEvent.ISSUE, CertificateEventStatus.FAILED,
                             "Manual upload rejected by connector identify: " + e.getMessage(), "");
             throw new ValidationException("Manual upload rejected by connector identify: " + e.getMessage());
+        } catch (AttributeException e) {
```
ivosh, line 3319: **~ Minor** | Error handling

ClientOperationServiceImpl.java:3293 adds `catch (AttributeException e)` -> "Identify attributes rejected by connector schema", and CertificateServiceImpl.java:3175 adds `catch (AttributeException | NotFoundException e)` -> "Identify attributes rejected by schema of new RA profile %s". Both are documented as handling "the supplied identify attributes failed structural validation against the connector's schema".

That is not where structural validation failures land. `AttributeEngine.validateUpdateDataAttributes` (AttributeEngine.java:721) is two calls: `updateDataAttributeDefinitions` — the only source of `AttributeException`, raised for a malformed *definition* coming back from the connector — and `validateDataAttributesContent` (AttributeEngine.java:728), which is what checks the operator's values and ends at line 780 with `throw new ValidationException(errors)`. `ValidationException` is an unchecked `RuntimeException` in `com.otilm.api.exception`, unrelated to `AttributeException`.

So the actual case these blocks were written for — a missing required identify attribute, or a value of the wrong content type — is caught by the pre-existing `catch (ValidationException e)` that sits *above* them: ClientOperationServiceImpl.java:3288 reports it as "Manual upload rejected by connector identify", and CertificateServiceImpl.java:3187 reports it as "Identification by authority of new RA profile %s rejected the certificate". Both blame the connector for a client-input error, and both write that wrong attribution into certificate event history. The same misattribution applies to `mergeAndValidateIdentifyAttributes`'s connector-unavailable path, which throws `ValidationException(CONNECTOR_UNAVAILABLE_MESSAGE)` (ExtendedAttributeServiceImpl.java:180).

Two readings. If the intent really was to distinguish schema rejections, the validate call needs to be in its own try (or the `ValidationException` block needs to tell the two apart) so the new message can fire. If the `AttributeException` block is only meant to cover malformed connector-supplied definitions, reword its message and comment to say that — as written the comment describes behaviour the code does not have. Either way the flow is correct, only the diagnostics are wrong.

Separately, both new blocks rethrow as `new ValidationException(...)` / `new CertificateOperationException(...)` with `e.getMessage()` only, dropping the cause (*Effective Java* Item 73/75). That matches the surrounding pre-existing style, so it is only worth changing if you touch these lines anyway.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -217,6 +224,83 @@ public List<BaseAttribute> listRevokeAttributes(AuthorityInstanceReference autho
...
+        ApiClientConnectorInfo connectorDto = connectorForApiClient(authority);
+        CertificateAttributeListRequestDtoV3 request = attributeListRequest(authority, raProfile);
+        List<BaseAttribute> response = schemaOrEmpty(() -> connectorApiFactory
+                .getCertificateApiClientV3(connectorDto)
+                .listIdentifyAttributes(connectorDto, request), "identify");
+        return contained(response, request.getAuthorityAttributes(), request.getRaProfileAttributes());
+    }
+
+    @FunctionalInterface
+    private interface SchemaCall {
```
ivosh: **. Nitpick** | Maintainability

AuthorityProviderV3Adapter.java:261 declares `private interface SchemaCall` and :272 `private List<BaseAttribute> schemaOrEmpty(...)`, wedged between the `listIdentifyAttributes` override (ends :258) and the `validateIssueAttributes` override (:305).

The file already has an established home for exactly this: the `// ---- private helpers ----` marker at line 486, whose first member is the analogous `@FunctionalInterface private interface ConnectorCall` (:489) followed by its runner `executeWithAcceptanceGuard`. `SchemaCall`/`schemaOrEmpty` are the same construct — a functional interface plus the helper that runs it — and belong next to their twin.

Moving them keeps the override block scannable as a contiguous list of the interface's methods (*Google Java Style* §3.4.2, logical member ordering). Pure placement, no behaviour change; take it or leave it.

src/test/java/com/otilm/core/integration/service/v3/V3RenewRevokeITest.java
```diff
@@ -120,6 +131,61 @@ void renew_async202_parksSuccessorPendingIssue_viaV3_notV2() throws Exception {
...
+                        postRequestedFor(urlEqualTo(V3_RENEW_PATH))
+                                .withRequestBody(matchingJsonPath("$.attributes[0].name", equalTo("validityOverride")))
+                                .withRequestBody(matchingJsonPath("$.attributes[0].content[0].data", equalTo("P90D"))));
+
+        // The submitted values also surface on the successor's detail, like issue/revoke/register values do.
+        CertificateDetailDto detail = certificateExternalService.getCertificate(SecuredUUID.fromUUID(successorUuid));
+        Assertions
+                .assertEquals(1, detail.getRenewAttributes().size(),
+                        "renew attributes must surface on the successor certificate detail");
+        Assertions.assertEquals("validityOverride", detail.getRenewAttributes().get(0).getName());
```
ivosh: **. Nitpick** | Maintainability

V3RenewRevokeITest.java:186 uses `detail.getRenewAttributes().get(0).getName()`. It is the only `.get(0)` in this file, and the repo is on Java 21 (`<java.version>21</java.version>` in pom.xml) with `.getFirst()` the dominant idiom in tests — 472 occurrences against 127 of `.get(0)`. `getFirst()` also reads better beside the `assertEquals(1, ...size())` on the line above. Cosmetic only.

ivosh (changes requested):

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Moved to the synchronous entries, mirroring revoke: an invalid submission now fails 422 before any successor exists (the operation contracts gained ConnectorException/AttributeException, matching revoke). The values persist on the successor right after its creation and the actions only read them back.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
klaraf755: Restructured: the merge/validate call sits in its own try, so ValidationException from the engine reports as a schema rejection; the connector-rejection wording now fires only for the identify call itself.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
klaraf755: Restructured: the merge/validate call sits in its own try, so ValidationException from the engine reports as a schema rejection; the connector-rejection wording now fires only for the identify call itself.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Restructured: the merge/validate call sits in its own try, so ValidationException from the engine reports as a schema rejection; the connector-rejection wording now fires only for the identify call itself. The renew half no longer has catches at all — its validation moved to the synchronous entry and propagates.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
klaraf755: Moved the persist after the identify call — nothing needed it earlier since the adapter takes the values as a parameter; a rejected identification now leaves no write behind and no row lock spans the connector round trips.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Added the rejection: non-empty values against an empty resolved schema now fail validation, so the detail can never claim attributes the CA never received. The uuid+name fallback is the engine's generic callback-reconciliation behaviour (validateDataAttributesContent's missing-definition loop re-stamps operations for every caller — issue↔revoke included) and predates this PR; we'll raise the engine-level tightening separately.

src/test/java/com/otilm/core/integration/service/CsrGenerationAttributesITest.java
klaraf755: Rebuilt on the authority+connector fixture; the pin is now that the issue listing is the resolution's only consultation (verifyNoMoreInteractions).

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
klaraf755: Added V3IdentifyITest — an RA-profile switch with a value, asserted on the identify wire body and on the certificate detail.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Resolved by the synchronous move — validation happens once in each sync entry, and a single persistRenewAttributes helper is shared by renew and rekey.

src/main/java/com/otilm/core/evaluator/CertificateTriggerEvaluator.java
klaraf755: Intended for now, and discussed: required-attribute enforcement stays uniform — an operation the CA would reject is better failed early with a named attribute than at the connector. We are aware this leaves a gap for the unattended paths: trigger-driven RA-profile switches and location renewals have no way to supply identify/renew attribute values, so an authority marking such an attribute required will fail those flows until the trigger/location configuration can carry values. Accepted as a known limitation for this release.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
klaraf755: Purging the old connector's identify slot alongside its metadata now. Renew values deliberately stay: they are creation provenance of the successor, exactly like issue values, which also survive switches.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
klaraf755: Documented the consequence on schemaOrEmpty and added the empty-schema rejection in the merge methods, so supplied values can never pass a degraded gate unvalidated. Kept the 404 log at debug — it is the common benign case for optional endpoints, and warn would fire on every listing against every connector that has not implemented the routes.

src/main/java/com/otilm/core/service/v2/impl/ExtendedAttributeServiceImpl.java
klaraf755: Restored the register sibling's rationale line on both methods — the skip is deliberate.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
klaraf755: Applied to both.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderAdapter.java
klaraf755: Applied to both — the rationale stays, the roadmap sentence is gone.

src/test/java/com/otilm/core/service/handler/authority/AuthorityProviderV3AdapterTest.java
klaraf755: Applied all four: null-body split per entry point (request/renew/identify), the V2 test renamed, both combined ITests split, and the detail assertion is its own test.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
klaraf755: Moved next to ConnectorCall in the private-helpers section.

src/test/java/com/otilm/core/integration/service/v3/V3RenewRevokeITest.java
klaraf755: Applied (in the split-out detail test).

ivosh (dismissed):

lubomirw (approved):

PR #2099, opened 2026-08-18
Fix newly created secrets being left inactive and unusable

### LINKED ISSUE #2098: Secret create message published before commit leaves secret INACTIVE
### Description

Creating a secret through Core leaves it permanently in `INACTIVE` state. The secret appears in inventory but is unusable, because the action message never reaches the connector and the caller still receives a success response.

Two defects stack.

**1. The action message is published before the creating transaction commits.**

`SecretServiceImpl` carries a class-level `@Transactional` (`SecretServiceImpl.java:121`). `createSecret` persists the `Secret` and then calls `produceActionMessage` at `SecretServiceImpl.java:381` while that transaction is still open. `ActionProducer.produceMessage` sends via `jmsTemplate.convertAndSend` (`ActionProducer.java:23`), which is not enlisted in the JPA transaction, so the broker receives the message immediately. `actionsListener` consumes it on another thread in its own READ COMMITTED transaction, reaches `getSecretEntity` (`SecretServiceImpl.java:797-800`), cannot see the uncommitted row, and throws `NotFoundException`.

**2. The resulting failure is swallowed, which makes a transient race permanent.**

`AbstractJmsEndpointConfig.java:109` catches generic `Exception`, logs at line 110, and does not rethrow. `RetryTemplate` therefore records the attempt as successful, so the message is ACKed and dropped — no redelivery, no dead-letter. `RetryConfig.java:52-54` marks only `JmsException`, `JMSException` and `IOException` retryable, so `MessageHandlingException` would be discarded even if it did propagate.

`createSecret` sets `INACTIVE` at `SecretServiceImpl.java:334`; only `createSecretAction` promotes the secret to `ACTIVE` (line 412) or `FAILED` (line 405, and only when the vault call itself throws). Because the message is gone, neither transition ever happens and the object is stuck with no recovery path.

The failure is connector-agnostic — everything fails inside Core before `createSecretInVault` is reached. Observed with common-credential-provider, reported also with hashicorp-vault-connector.

**Scope.** All four Secret action sites route through the same helper and share the defect: CREATE (`:381`), UPDATE (`:478`), DELETE (`:585`), UPDATE_SOURCE_VAULT_PROFILE (`:928`).

**The correct pattern already exists in this codebase.** `ApprovalServiceImpl.java:414-417` wraps the same `actionProducer.produceMessage` call in `TransactionSynchronizationManager.registerSynchronization(... afterCommit())`. The certificate flows avoid the race a different way — `issueCertificate`, `issueExistingCertificate`, `renewCertificate` and `rekeyCertificate` all run under `@Transactional(propagation = Propagation.NOT_SUPPORTED)`, so no transaction is open at send time. `ClientOperationServiceImpl:2449` (`revokeCertificate`) has the same latent shape under the class-level transaction and should be reviewed alongside the fix.

**Suggested fix.** Defer the publish in `produceActionMessage` until after commit — either the `registerSynchronization` pattern above, or a `@TransactionalEventListener(AFTER_COMMIT, fallbackExecution = true)` consumer as used by `NotificationProducer.java:48`. One change covers all four Secret actions. Separately, stop swallowing handler failures at `AbstractJmsEndpointConfig.java:109` so unhandleable messages are redelivered or dead-lettered rather than vanishing silently.

**Related logging defect.** `MessageHandlingException` is `@Data @ToString` with an `Object messageData` field (`interfaces` — `api/exception/MessageHandlingException.java:7-13`), and `ActionMessage` is `@Data` with `private Object data` (`core/.../messaging/model/ActionMessage.java:9,19`). Because `AbstractJmsEndpointConfig.java:110` passes the exception to the logger, the full message payload — including its `data` map — is written to ERROR level on every failed action. This defeats the intent documented eleven lines earlier at `AbstractJmsEndpointConfig.java:95-96` ("Log only metadata, never the body: message payloads can carry secrets"). The payload field must be kept out of the logged representation.

### Steps to Reproduce

1. Configure a Vault Instance with a vault connector (observed with common-credential-provider) and a Vault Profile with no approval profile attached.
2. Create a secret through Core against that vault profile.
3. Watch the Core log for the `actionsListener` entry that follows within roughly 25 ms of the request.
4. Open the secret in inventory.

### Expected Behavior

The action message is published only after the creating transaction commits. `actionsListener` finds the `Secret`, calls the connector, stores the vault version and metadata, and promotes the secret to `ACTIVE`. If the connector call fails, the secret moves to `FAILED`. If the message genuinely cannot be handled, it is redelivered or dead-lettered rather than silently dropped, and the payload is not written to the log.

### Actual Behavior

The listener fails roughly 23 ms after the send, while the creating transaction is still open. Identifiers and the payload are replaced with placeholders below:

```
[2026-08-18 12:03:58.168] WARN  [tomcat-handler-38722] [org.hibernate.engine.spi.ActionQueue - 1319]: The batch containing 2 statements could not be sorted. This might indicate a circular entity relationship.
[2026-08-18 12:03:58.197] INFO  [tomcat-handler-38722] [com.otilm.core.messaging.jms.configuration.PlatformJmsPoolConnectionFactory - 32]: Creating pooled connection wrapping: org.apache.qpid.jms.JmsConnection@92d3c40
[2026-08-18 12:03:58.220] ERROR [actionsListener-1] [com.otilm.core.messaging.jms.listeners.ActionsListener - 85]: Failed to perform Secret create action!: Object of type 'Secret' identified by <secret-uuid> is not found.
[2026-08-18 12:03:58.237] ERROR [actionsListener-1] [com.otilm.core.messaging.jms.listeners.AbstractJmsEndpointConfig - 109]: Unexpected error in endpoint 'actionsListener'
MessageHandlingException(queueName=action, messageData=ActionMessage(resource=SECRET, resourceAction=CREATE, resourceUuid=<secret-uuid>, userUuid=<user-uuid>, data={encryptedContent=<redacted>, name=<redacted>, updatedSourceVaultProfileUuid=null, attributes=[], deleteInVault=false, originalState=null}, approvalProfileResource=VAULT_PROFILE, approvalProfileResourceUuid=<vault-profile-uuid>, approvalUuid=null, approvalStatus=null))
        at com.otilm.core.messaging.jms.listeners.ActionsListener.processMessage(ActionsListener.java:88)
        at com.otilm.core.messaging.jms.listeners.ActionsListener.processMessage(ActionsListener.java:34)
        at com.otilm.core.messaging.jms.listeners.AbstractJmsEndpointConfig.lambda$listenerEndpointInternal$0(AbstractJmsEndpointConfig.java:100)
        at org.springframework.retry.support.RetryTemplate.doExecute(RetryTemplate.java:367)
        at org.springframework.retry.support.RetryTemplate.execute(RetryTemplate.java:240)
        at com.otilm.core.messaging.jms.listeners.AbstractJmsEndpointConfig.lambda$listenerEndpointInternal$1(AbstractJmsEndpointConfig.java:89)
```

Note that the Hibernate flush and the producer connection both occur on `tomcat-handler-38722` — the request thread, inside the still-open transaction. The transaction commits afterwards, which is why the secret does appear in inventory, at `INACTIVE`. No retry follows, nothing is dead-lettered, the caller receives a success response, and the connector is never invoked.

### Severity

Critical

### Module

Secrets

### Environment

Observed 2026-08-18 on Core with RabbitMQ (Qpid JMS over AMQP). Vault connectors: common-credential-provider (observed), hashicorp-vault-connector (reported). All code references verified against `core` `origin/main`.

lubomirw:
## Problem

Creating a secret through Core left it permanently `INACTIVE` (#2098). `SecretServiceImpl` carries a class-level `@Transactional`, and `produceActionMessage` published the JMS action message via `actionProducer.produceMessage` **while that transaction was still open**. `actionsListener` consumed it on another thread in its own READ COMMITTED transaction, could not see the uncommitted `Secret` row, and threw `NotFoundException`. The secret committed `INACTIVE`, the connector was never invoked, and the caller still got a success response.

All four Secret action sites — CREATE, UPDATE, DELETE, UPDATE_SOURCE_VAULT_PROFILE — route through the same `produceActionMessage` helper and shared the defect.

## Fix

Defer the publish to an `afterCommit` synchronization via `TransactionSynchronizationManager.registerSynchronization(...)`, so the message reaches the broker only once the creating transaction has committed and the listener's transaction can load the row. One change covers all four action sites.

This mirrors the pattern already used in `ApprovalServiceImpl` (afterCommit `actionProducer.produceMessage`) and `NotificationProducer` (`@TransactionalEventListener(AFTER_COMMIT)`).

## Tests

- New `SecretServiceITest#createSecretPublishesActionMessageOnlyAfterCommit` reads the secret from a separate `REQUIRES_NEW` transaction at publish time; fails before the fix (row invisible), passes after.
- The existing broker shim in `SecretServiceITest` previously ran the listener **inline in the producer's transaction** and so only passed by riding this bug. It now runs the handler in a `REQUIRES_NEW` transaction, faithfully modelling the real separate-transaction consumer. Full class: 22/22 green.

## Scope

This PR fixes only the **pre-commit publish race** (defect 1 of #2098). Deliberately deferred to follow-ups:

- **Durable dispatch / redelivery / DLQ** — an `afterCommit` send that exhausts retries (e.g. broker down) commits the DB but loses the message. This is the same tradeoff the platform already accepts in `ApprovalServiceImpl` / `NotificationProducer`; closing it needs a transactional outbox or redelivery, tracked separately.
- **Handler-failure swallow** in `AbstractJmsEndpointConfig` and the **action-payload leak** into ERROR logs — separate follow-ups.

An independent `copilot` review pass confirmed no no-transaction path (all producers enter through the class-level `@Transactional`) and raised only the durability gap above, which is out of scope by design.

Refs #2098

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/SecretServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/SecretServiceImpl.java
@@ -116,6 +116,8 @@ import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Pageable;
 import org.springframework.security.concurrent.DelegatingSecurityContextExecutorService;
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
 
 @Service(value = Resource.Codes.SECRET)
 @Transactional
@@ -492,7 +494,14 @@ public class SecretServiceImpl implements SecretExternalService, SecretInternalS
         actionMessage.setResource(Resource.SECRET);
         actionMessage.setResourceAction(resourceAction);
         actionMessage.setResourceUuid(secret.getUuid());
-        actionProducer.produceMessage(actionMessage);
+        // actionsListener consumes this on another thread in its own transaction; publishing before the creating
+        // transaction commits leaves that transaction unable to load the secret, so defer the send until after commit.
+        TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
+            @Override
+            public void afterCommit() {
+                actionProducer.produceMessage(actionMessage);
+            }
+        });
     }
 
     @ExternalAuthorization(resource = Resource.SECRET, action = ResourceAction.UPDATE)
--- a/src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
+++ b/src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
@@ -151,12 +151,8 @@ class SecretServiceITest extends BaseSpringBootTest {
     @BeforeEach
     void setUp() throws AlreadyExistException, AttributeException, NoSuchAlgorithmException, JsonProcessingException {
 
-        // Stand in for the broker so the listener runs the way it does in production.
-        // Broker model: process the action message inline instead of queueing it, approved to bypass approval.
-        // Transaction boundary: the real listener runs on another thread in its own transaction, only after the
-        // producing transaction has committed — so run the handler in a fresh REQUIRES_NEW transaction.
-        // Why not inline: at afterCommit the producing transaction has already committed; an inline handler would
-        // join that committed transaction, so its writes would never commit and would be discarded at cleanup.
+        // REQUIRES_NEW so the handler runs in its own transaction like the real broker: at afterCommit the producing
+        // transaction has already committed, so an inline handler would join it and its writes would be discarded.
         final TransactionTemplate listenerTransaction = new TransactionTemplate(transactionManager);
         listenerTransaction.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
         Mockito.doAnswer(invocation -> {
@@ -949,11 +945,8 @@ class SecretServiceITest extends BaseSpringBootTest {
     @Test
     void createSecretPublishesActionMessageOnlyAfterCommit()
             throws NotFoundException, AttributeException, AlreadyExistException, ConnectorException {
-        // Assert the action message is published only after the creating transaction commits.
-        // Broker model: the real listener loads the secret on another thread in its own READ COMMITTED transaction.
-        // Failure guarded: publishing while the creating transaction is still open leaves the row invisible to that
-        // separate transaction, so the listener throws NotFoundException and the secret stays INACTIVE.
-        // Check: read the secret from a REQUIRES_NEW transaction at the moment the message is published.
+        // Read from a separate REQUIRES_NEW transaction: a publish before commit leaves the secret invisible there,
+        // exactly as it is to the real listener's own transaction.
         final AtomicBoolean secretVisibleFromSeparateTransaction = new AtomicBoolean();
         final TransactionTemplate separateTransaction = new TransactionTemplate(transactionManager);
         separateTransaction.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
--- a/src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
+++ b/src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
@@ -151,10 +151,12 @@ class SecretServiceITest extends BaseSpringBootTest {
     @BeforeEach
     void setUp() throws AlreadyExistException, AttributeException, NoSuchAlgorithmException, JsonProcessingException {
 
-        // Stand in for the broker: process the action message instead of queueing it, and approve it to bypass
-        // approval. The real listener consumes on another thread in its own transaction after the producing
-        // transaction has committed, so run the handler in a fresh REQUIRES_NEW transaction rather than inline in the
-        // producer's — inline would run in the producer's already-committed transaction and its writes would be lost.
+        // Stand in for the broker so the listener runs the way it does in production.
+        // Broker model: process the action message inline instead of queueing it, approved to bypass approval.
+        // Transaction boundary: the real listener runs on another thread in its own transaction, only after the
+        // producing transaction has committed — so run the handler in a fresh REQUIRES_NEW transaction.
+        // Why not inline: at afterCommit the producing transaction has already committed; an inline handler would
+        // join that committed transaction, so its writes would never commit and would be discarded at cleanup.
         final TransactionTemplate listenerTransaction = new TransactionTemplate(transactionManager);
         listenerTransaction.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
         Mockito.doAnswer(invocation -> {
@@ -947,10 +949,11 @@ class SecretServiceITest extends BaseSpringBootTest {
     @Test
     void createSecretPublishesActionMessageOnlyAfterCommit()
             throws NotFoundException, AttributeException, AlreadyExistException, ConnectorException {
-        // The listener consumes the action message on another thread in its own READ COMMITTED transaction, so it can
-        // only load the secret once the creating transaction has committed. Reproduce that constraint by reading the
-        // secret from a separate transaction at the moment the message is published: if the publish still happens
-        // inside the open creating transaction, the row is invisible and the listener would throw NotFoundException.
+        // Assert the action message is published only after the creating transaction commits.
+        // Broker model: the real listener loads the secret on another thread in its own READ COMMITTED transaction.
+        // Failure guarded: publishing while the creating transaction is still open leaves the row invisible to that
+        // separate transaction, so the listener throws NotFoundException and the secret stays INACTIVE.
+        // Check: read the secret from a REQUIRES_NEW transaction at the moment the message is published.
         final AtomicBoolean secretVisibleFromSeparateTransaction = new AtomicBoolean();
         final TransactionTemplate separateTransaction = new TransactionTemplate(transactionManager);
         separateTransaction.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
--- a/src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
+++ b/src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
@@ -7,6 +7,7 @@ import com.github.tomakehurst.wiremock.client.WireMock;
 import com.otilm.api.exception.AlreadyExistException;
 import com.otilm.api.exception.AttributeException;
 import com.otilm.api.exception.ConnectorException;
+import com.otilm.api.exception.MessageHandlingException;
 import com.otilm.api.exception.NotFoundException;
 import com.otilm.api.exception.SecretOperationException;
 import com.otilm.api.exception.ValidationException;
@@ -81,6 +82,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
@@ -94,6 +96,9 @@ import org.springframework.http.MediaType;
 import org.springframework.test.context.bean.override.mockito.MockitoBean;
 import org.springframework.test.context.event.ApplicationEvents;
 import org.springframework.test.context.event.RecordApplicationEvents;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.TransactionDefinition;
+import org.springframework.transaction.support.TransactionTemplate;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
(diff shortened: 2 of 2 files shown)


src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
```diff
@@ -144,13 +151,23 @@ class SecretServiceITest extends BaseSpringBootTest {
     @BeforeEach
     void setUp() throws AlreadyExistException, AttributeException, NoSuchAlgorithmException, JsonProcessingException {
 
-        // Process message instead of sending it to the queue and set approval status to approved to bypass approval in
-        // tests
+        // Stand in for the broker: process the action message instead of queueing it, and approve it to bypass
+        // approval. The real listener consumes on another thread in its own transaction after the producing
+        // transaction has committed, so run the handler in a fresh REQUIRES_NEW transaction rather than inline in the
+        // producer's — inline would run in the producer's already-committed transaction and its writes would be lost.
```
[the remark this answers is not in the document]
lubomirw: Comment is accurate. `isActualTransactionActive()` stays true through `afterCommit`, so an inline handler participates in the just-committed transaction and its writes are lost — the 5 SecretServiceITest failures on the first fix pass proved exactly that. "already-committed transaction" names that state; the REQUIRES_NEW shim is what makes the writes durable. Happy to reword for clarity if it still reads ambiguously.

src/main/java/com/otilm/core/service/impl/SecretServiceImpl.java
```diff
@@ -492,7 +494,14 @@ private void produceActionMessage(SecretActionData secretActionData, VaultProfil
         actionMessage.setResource(Resource.SECRET);
         actionMessage.setResourceAction(resourceAction);
         actionMessage.setResourceUuid(secret.getUuid());
-        actionProducer.produceMessage(actionMessage);
+        // actionsListener consumes this on another thread in its own transaction; publishing before the creating
+        // transaction commits leaves that transaction unable to load the secret, so defer the send until after commit.
+        TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
+            @Override
+            public void afterCommit() {
+                actionProducer.produceMessage(actionMessage);
```
lubomirw, line 502: **[High]** | Error handling

If the broker send exhausts retries (or the process dies after commit), the DB is committed but no action message is queued — a durable-dispatch gap. Deferred by design and tracked in #2100 (transactional outbox / redelivery). This is the same afterCommit-publish tradeoff already accepted in `ApprovalServiceImpl` and documented in `NotificationProducer`, so fixing it in this one file alone would diverge from the convention.

_Suggested:_ resolve in #2100 across all afterCommit-publish sites (outbox, or at minimum try/catch+log).

<sub>flagged by Copilot CLI + superpowers review</sub>

lubomirw (commented):
Automated multi-reviewer triage (Copilot CLI, guided-pr-review, superpowers, hygiene). Event: Comment — 0 in-scope must-fix; 1 High deferred, rest optional.

All four reviewers confirm the fix itself is **correct** and consistent with the established afterCommit-publish idiom in `ApprovalServiceImpl` / `NotificationProducer`.

Optional / out-of-scope (no change requested here):
- **[Low] `registerSynchronization` throws if no active tx** (`SecretServiceImpl:499`) — verified safe: every producing entry point is `@Transactional` with no propagation override or self-invocation. `ApprovalServiceImpl` does not guard either.
- **[Low] afterCommit-publish idiom duplicated** across services — extract a shared helper / `@TransactionalEventListener`; a future refactor that would also centralize the #2100 durability handling.
- **[Nit] deferred send is silent (no log)** — matches the `ApprovalServiceImpl` actionProducer afterCommit, which is also silent.

Title and `bug` label are accurate as-is.

src/main/java/com/otilm/core/service/impl/SecretServiceImpl.java
ivosh: **! Major** | Error handling

SecretServiceImpl.java:502 lets `ActionProducer.produceMessage` exceptions escape from `afterCommit`. `ActionProducer.java:22` rethrows after retries, so CREATE, UPDATE, DELETE, and UPDATE_SOURCE_VAULT_PROFILE can commit their database work and then return an HTTP failure, encouraging a retry against already-committed state; the analogous `NotificationProducer.java:42-57` explicitly contains this failure because Spring propagates after-commit exceptions.

_Suggested:_ Catch and log exhausted dispatch failures inside `afterCommit`, as `NotificationProducer.onNotificationMessage` does, and test that a throwing `ActionProducer` cannot fail the committed service call.

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
```diff
@@ -144,13 +151,23 @@ class SecretServiceITest extends BaseSpringBootTest {
...
+        final TransactionTemplate listenerTransaction = new TransactionTemplate(transactionManager);
+        listenerTransaction.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
         Mockito.doAnswer(invocation -> {
             ActionMessage msg = invocation.getArgument(0);
             msg.setApprovalStatus(ApprovalStatusEnum.APPROVED);
             msg.setApprovalUuid(UUID.randomUUID());
-            actionListener.processMessage(msg);
+            listenerTransaction.executeWithoutResult(status -> {
```
ivosh: **! Major** | Performance

SecretServiceITest.java:164 now models the listener with a `REQUIRES_NEW` transaction, matching production where `ActionsListener.java:137` invokes `SecretServiceImpl.processSecretAction`. Because `SecretServiceImpl` has class-level `@Transactional`, SecretServiceImpl.java:1068 keeps that transaction open while CREATE, UPDATE, DELETE, and source-vault migration perform connector HTTP calls (for example SecretServiceImpl.java:644-646 and 1180-1182), holding database resources throughout potentially slow external requests.

_Suggested:_ Move `processSecretAction` orchestration to a `NOT_SUPPORTED` service and perform only local mutations in short transactional writer methods around the connector calls.

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
```diff
@@ -144,13 +151,23 @@ class SecretServiceITest extends BaseSpringBootTest {
     @BeforeEach
     void setUp() throws AlreadyExistException, AttributeException, NoSuchAlgorithmException, JsonProcessingException {
 
-        // Process message instead of sending it to the queue and set approval status to approved to bypass approval in
-        // tests
+        // Stand in for the broker: process the action message instead of queueing it, and approve it to bypass
```
ivosh: **. Nitpick** | Documentation

SecretServiceITest.java:154-157 and SecretServiceITest.java:950-953 each combine the broker model, approval bypass, transaction boundary, and guarded failure into long linear paragraphs. The critical reason for REQUIRES_NEW is buried inside a 35-plus-word sentence, making this PR's regression rationale harder to scan.

_Suggested:_ Split each comment into a one-line summary followed by short labeled paragraphs for the broker model, transaction boundary, and failure being guarded.

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
ivosh: **! Major** | Performance

SecretServiceITest.java:164 now models the listener with a `REQUIRES_NEW` transaction, matching production where `ActionsListener.java:137` invokes `SecretServiceImpl.processSecretAction`. Because `SecretServiceImpl` has class-level `@Transactional`, SecretServiceImpl.java:1068 keeps that transaction open while CREATE, UPDATE, DELETE, and source-vault migration perform connector HTTP calls (for example SecretServiceImpl.java:644-646 and 1180-1182), holding database resources throughout potentially slow external requests.

_Suggested:_ Move `processSecretAction` orchestration to a `NOT_SUPPORTED` service and perform only local mutations in short transactional writer methods around the connector calls.

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
ivosh: **. Nitpick** | Documentation

SecretServiceITest.java:154-157 and SecretServiceITest.java:950-953 each combine the broker model, approval bypass, transaction boundary, and guarded failure into long linear paragraphs. The critical reason for REQUIRES_NEW is buried inside a 35-plus-word sentence, making this PR's regression rationale harder to scan.

_Suggested:_ Split each comment into a one-line summary followed by short labeled paragraphs for the broker model, transaction boundary, and failure being guarded.

ivosh (dismissed):

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
lubomirw: Fixed in bfd8c589d — both comments now lead with a one-line summary followed by short labeled points (broker model / transaction boundary / failure guarded).

lubomirw: Went further in d405da8bb — trimmed both to a two-line WHY (just the REQUIRES_NEW rationale). The rest was restating the test name and assertion message.

ivosh (approved):

PR #2102, opened 2026-08-18
Extract v1 discovery flow behind provider adapter

### LINKED ISSUE #1960: Extract v1 discovery flow behind provider adapter
### Description

Introduce the discovery provider adapter seam (authority-v3 factory pattern) and extract the existing v1 flow behind it, behavior-frozen. This gives discovery its missing versioning seam: today `ConnectorApiFactory` builds the v1 discovery client unconditionally and `DiscoveryServiceImpl` (768 lines) is version-blind.

### Scope

- `DiscoveryProviderAdapter` (`start`/`stop`/`resume`/`cancel`) + `DiscoveryProviderAdapterFactory` switching on the connector-interface association. NULL interface row = v1 (deliberate inverse of the authority factory's NULL rule — discovery's legacy is v1); unknown version = error.
- `V1DiscoveryProviderAdapter` = pure extraction of the current flow (`runDiscoveryAsync` body, wait loop, page download, v1 attribute path) — method bodies unchanged, imports only.
- `DiscoveryServiceImpl` shrinks to orchestration and CRUD; v1 `stop`/`resume` map to a 422 "not supported by connector version".

### Acceptance Criteria

- [ ] All existing discovery ITests pass unchanged (extraction commit allows no behavior change)
- [ ] Factory unit tests: NULL row -> v1, "v2" -> v2, unknown -> error
- [ ] ArchUnit fences stay green (`TransactionalBoundaryArchTest`, `ConnectorApiClientArchTest`)

### References

- Epic: OmniTrustILM/ilm#267 (core plan Task 3); precedent `AuthorityProviderAdapterFactory`

lubomirw:
Closes #1960 (epic OmniTrustILM/ilm#267).

## What this does

Introduces the provider-adapter seam the discovery v2 flow will plug into, and moves the entire v1 run flow behind it:

- **`DiscoveryProviderAdapter`** — lifecycle contract (`start`, `stop`, `resume`, `cancel`).
- **`V1DiscoveryProviderAdapter`** — the legacy flow extracted from `DiscoveryServiceImpl` behavior-frozen: method bodies moved verbatim, no logic changes. The full discovery ITest suite passes unchanged.
- **`V2DiscoveryProviderAdapter`** — placeholder; every operation fails with `IllegalStateException` until the real implementation lands (core#1964).
- **`DiscoveryProviderAdapterFactory`** — routing: a run with no `connector_interface_uuid` association (or an unloadable run) goes to v1, `"v2"` goes to the v2 adapter, anything else — missing interface row, null version, unknown version — refuses with `UnsupportedDiscoveryVersionException`.
- **Refusal handling** — REST surfaces the refusal as 400 with a fixed body; a scheduled/async run ends as a terminal `FAILED` run with a curated message and a `DISCOVERY_FINISHED` event, since both callers expect a result rather than an exception.

## Schema amendments (migration `V202608151300`)

The four `progress_*` columns collapse into a single `progress` jsonb snapshot (`DiscoveryProgressDto` whole — torn snapshots unrepresentable under the two concurrent writers), and `resources` becomes `TEXT[]` of `Resource` enum names, the `connector_interface.features` shape.

**Compatibility note:** the migration drops and re-adds columns rather than converting them. Both columns are unwritten by any code path so no data is at risk; the only window is a rolling deploy where a pod running an older main snapshot overlaps the new schema. No released version carries these columns.

## Plan deviation

`start` takes `(UUID, ScheduledJobInfo)` rather than `(Discovery, ConnectorDto)` as planned: the v1 flow reloads its own context across transaction boundaries, so handing it a pre-loaded entity would be misleading.

## Review already folded in

Pre-opening review rounds are in the branch as separate commits: an internal multi-review pass and an external review of the refusal path (detached-entity mapping after the terminal save, missing `DISCOVERY_FINISHED` emission, raw exception text persisted as the run message — all fixed, with a regression ITest).

## Deliberately deferred

`V1DiscoveryProviderAdapter.start` keeps its 97-line moved-verbatim body: extracting the commented phases into named methods is pure structure, but it would widen this PR's review surface and dilute the "bodies moved unchanged" guarantee the extraction is built on. The split lands with the v2-era rework in core#1964.

### DIFF
--- a/src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
+++ b/src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
@@ -26,6 +26,7 @@ import com.otilm.api.model.common.ErrorMessageDto;
 import com.otilm.api.model.core.acme.ProblemDocument;
 import com.otilm.api.model.core.auth.Resource;
 import com.otilm.core.exception.UnsupportedAuthorityVersionException;
+import com.otilm.core.exception.UnsupportedDiscoveryVersionException;
 import com.otilm.core.security.authn.PlatformAuthenticationException;
 import com.otilm.core.security.exception.AuthenticationServiceException;
 import com.otilm.core.util.AuthHelper;
@@ -465,6 +466,20 @@ public class ExceptionHandlingAdvice {
     public ErrorMessageDto handleUnsupportedAuthorityVersionException(UnsupportedAuthorityVersionException ex) {
         LOG.warn("HTTP 400: {}", ex.getMessage(), ex);
         return ErrorMessageDto.getInstance("The authority's connector interface version is not supported.");
+    }
+
+    /**
+     * Handler for {@link UnsupportedDiscoveryVersionException} — the discovery twin of the authority handler above,
+     * same reasoning: caller-fixable configuration, fixed body because the message carries an unvalidated
+     * connector-reported version string.
+     *
+     * @return a fixed message that names neither the discovery nor the version
+     */
+    @ExceptionHandler(UnsupportedDiscoveryVersionException.class)
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    public ErrorMessageDto handleUnsupportedDiscoveryVersionException(UnsupportedDiscoveryVersionException ex) {
+        LOG.warn("HTTP 400: {}", ex.getMessage(), ex);
+        return ErrorMessageDto.getInstance("The discovery's connector interface version is not supported.");
     }
 
     /**
--- a/src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
+++ b/src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
@@ -469,9 +469,12 @@ public class ExceptionHandlingAdvice {
     }
 
     /**
-     * Handler for {@link UnsupportedDiscoveryVersionException} — the discovery twin of the authority handler above,
-     * same reasoning: caller-fixable configuration, fixed body because the message carries an unvalidated
-     * connector-reported version string.
+     * Handler for {@link UnsupportedDiscoveryVersionException}.
+     *
+     * <p>
+     * The discovery twin of {@link #handleUnsupportedAuthorityVersionException}, for the same reasons: an unrecognised
+     * connector interface version is caller-fixable configuration, and the body is fixed because the exception's
+     * message carries an unvalidated connector-reported version string.
      *
      * @return a fixed message that names neither the discovery nor the version
      */
--- a/src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderAdapter.java
+++ b/src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderAdapter.java
@@ -2,6 +2,7 @@ package com.otilm.core.service.handler.discovery;
 
 import com.otilm.api.model.client.discovery.DiscoveryDetailDto;
 import com.otilm.core.dao.entity.Discovery;
+import com.otilm.core.exception.UnsupportedDiscoveryVersionException;
 import com.otilm.core.tasks.ScheduledJobInfo;
 import java.util.UUID;
 
@@ -16,24 +17,33 @@ import java.util.UUID;
  */
 public interface DiscoveryProviderAdapter {
 
-    /** Runs the whole provider-side discovery for the run, returning the run detail at handoff or completion. */
+    /**
+     * Runs the whole provider-side discovery for the run, returning the run detail at handoff or completion.
+     *
+     * @throws UnsupportedDiscoveryVersionException when the run cannot be dispatched to this adapter after all — the
+     * routing-refusal signal {@code runDiscovery} converts into a terminal FAILED run rather than an escaping
+     * exception.
+     */
     DiscoveryDetailDto start(UUID discoveryUuid, ScheduledJobInfo scheduledJobInfo);
 
     /**
-     * @throws UnsupportedOperationException when the connector version cannot suspend a run (v1 cannot); callers wiring
-     * the lifecycle endpoints must map this to the contract's 422.
+     * @throws UnsupportedOperationException when the connector version can never suspend a run (v1 cannot); callers
+     * wiring the lifecycle endpoints must map this to the contract's 422. An adapter whose implementation is merely
+     * pending signals that with {@link IllegalStateException} instead — a defect, not part of this contract.
      */
     void stop(Discovery discovery);
 
     /**
-     * @throws UnsupportedOperationException when the connector version cannot resume a run (v1 cannot); callers wiring
-     * the lifecycle endpoints must map this to the contract's 422.
+     * @throws UnsupportedOperationException when the connector version can never resume a run (v1 cannot); callers
+     * wiring the lifecycle endpoints must map this to the contract's 422. An adapter whose implementation is merely
+     * pending signals that with {@link IllegalStateException} instead — a defect, not part of this contract.
      */
     void resume(Discovery discovery);
 
     /**
-     * @throws UnsupportedOperationException when the connector version cannot cancel a run (v1 cannot); callers wiring
-     * the lifecycle endpoints must map this to the contract's 422.
+     * @throws UnsupportedOperationException when the connector version can never cancel a run (v1 cannot); callers
+     * wiring the lifecycle endpoints must map this to the contract's 422. An adapter whose implementation is merely
+     * pending signals that with {@link IllegalStateException} instead — a defect, not part of this contract.
      */
     void cancel(Discovery discovery);
 }
--- a/src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderAdapter.java
+++ b/src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderAdapter.java
@@ -14,6 +14,11 @@ import java.util.UUID;
  * {@code start} takes the run's uuid rather than the entity: the v1 flow reloads the run with its associations inside
  * its own transaction boundaries, because it executes asynchronously in a different session than the one that created
  * the run.
+ *
+ * <p>
+ * Lifecycle refusals carry meaning: {@code UnsupportedOperationException} means the connector generation can never
+ * perform the operation (callers map it to the contract's 422); an adapter whose implementation is merely pending
+ * signals that with {@link IllegalStateException} instead — a defect, not part of this contract.
  */
 public interface DiscoveryProviderAdapter {
 
@@ -27,23 +32,17 @@ public interface DiscoveryProviderAdapter {
     DiscoveryDetailDto start(UUID discoveryUuid, ScheduledJobInfo scheduledJobInfo);
 
     /**
-     * @throws UnsupportedOperationException when the connector version can never suspend a run (v1 cannot); callers
-     * wiring the lifecycle endpoints must map this to the contract's 422. An adapter whose implementation is merely
-     * pending signals that with {@link IllegalStateException} instead — a defect, not part of this contract.
+     * @throws UnsupportedOperationException when the connector version can never suspend a run (v1 cannot)
      */
     void stop(Discovery discovery);
 
     /**
-     * @throws UnsupportedOperationException when the connector version can never resume a run (v1 cannot); callers
-     * wiring the lifecycle endpoints must map this to the contract's 422. An adapter whose implementation is merely
-     * pending signals that with {@link IllegalStateException} instead — a defect, not part of this contract.
+     * @throws UnsupportedOperationException when the connector version can never resume a run (v1 cannot)
      */
     void resume(Discovery discovery);
 
     /**
-     * @throws UnsupportedOperationException when the connector version can never cancel a run (v1 cannot); callers
-     * wiring the lifecycle endpoints must map this to the contract's 422. An adapter whose implementation is merely
-     * pending signals that with {@link IllegalStateException} instead — a defect, not part of this contract.
+     * @throws UnsupportedOperationException when the connector version can never cancel a run (v1 cannot)
      */
     void cancel(Discovery discovery);
 }
--- a/src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderAdapter.java
+++ b/src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderAdapter.java
@@ -19,12 +19,21 @@ public interface DiscoveryProviderAdapter {
     /** Runs the whole provider-side discovery for the run, returning the run detail at handoff or completion. */
     DiscoveryDetailDto start(UUID discoveryUuid, ScheduledJobInfo scheduledJobInfo);
 
-    /** @throws UnsupportedOperationException when the connector version cannot suspend a run (v1 cannot). */
+    /**
+     * @throws UnsupportedOperationException when the connector version cannot suspend a run (v1 cannot); callers wiring
+     * the lifecycle endpoints must map this to the contract's 422.
+     */
     void stop(Discovery discovery);
 
-    /** @throws UnsupportedOperationException when the connector version cannot resume a run (v1 cannot). */
+    /**
+     * @throws UnsupportedOperationException when the connector version cannot resume a run (v1 cannot); callers wiring
+     * the lifecycle endpoints must map this to the contract's 422.
+     */
     void resume(Discovery discovery);
 
-    /** @throws UnsupportedOperationException when the connector version cannot cancel a run (v1 cannot). */
+    /**
+     * @throws UnsupportedOperationException when the connector version cannot cancel a run (v1 cannot); callers wiring
+     * the lifecycle endpoints must map this to the contract's 422.
+     */
     void cancel(Discovery discovery);
 }
--- a/src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderAdapter.java
+++ b/src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderAdapter.java
@@ -1,0 +1,30 @@
+package com.otilm.core.service.handler.discovery;
+
+import com.otilm.api.model.client.discovery.DiscoveryDetailDto;
+import com.otilm.core.dao.entity.Discovery;
+import com.otilm.core.tasks.ScheduledJobInfo;
+import java.util.UUID;
+
+/**
+ * The version seam for discovery provider interaction: one implementation per connector discovery interface generation,
+ * selected by {@link DiscoveryProviderAdapterFactory}.
+ *
+ * <p>
+ * {@code start} takes the run's uuid rather than the entity: the v1 flow reloads the run with its associations inside
+ * its own transaction boundaries, because it executes asynchronously in a different session than the one that created
+ * the run.
+ */
+public interface DiscoveryProviderAdapter {
+
+    /** Runs the whole provider-side discovery for the run, returning the run detail at handoff or completion. */
+    DiscoveryDetailDto start(UUID discoveryUuid, ScheduledJobInfo scheduledJobInfo);
+
+    /** @throws UnsupportedOperationException when the connector version cannot suspend a run (v1 cannot). */
+    void stop(Discovery discovery);
+
+    /** @throws UnsupportedOperationException when the connector version cannot resume a run (v1 cannot). */
+    void resume(Discovery discovery);
+
+    /** @throws UnsupportedOperationException when the connector version cannot cancel a run (v1 cannot). */
+    void cancel(Discovery discovery);
+}
--- a/src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderV1Adapter.java
+++ b/src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderV1Adapter.java
@@ -71,9 +71,9 @@ import org.springframework.transaction.support.DefaultTransactionDefinition;
  * {@link UnsupportedOperationException}; callers wiring the lifecycle endpoints must map that to the contract's 422.
  */
 @Component
-public class V1DiscoveryProviderAdapter implements DiscoveryProviderAdapter {
+public class DiscoveryProviderV1Adapter implements DiscoveryProviderAdapter {
 
-    private static final Logger logger = LoggerFactory.getLogger(V1DiscoveryProviderAdapter.class);
+    private static final Logger logger = LoggerFactory.getLogger(DiscoveryProviderV1Adapter.class);
 
     private static final Semaphore downloadCertSemaphore = new Semaphore(10);
 
@@ -90,7 +90,7 @@ public class V1DiscoveryProviderAdapter implements DiscoveryProviderAdapter {
     private final ConnectorApiFactory connectorApiFactory;
     private final EventProducer eventProducer;
 
-    public V1DiscoveryProviderAdapter(DiscoveryProperties discoveryProperties,
+    public DiscoveryProviderV1Adapter(DiscoveryProperties discoveryProperties,
             PlatformTransactionManager transactionManager, DiscoveryRepository discoveryRepository,
             ConnectorRepository connectorRepository, CertificateRepository certificateRepository,
             DiscoveryCertificateRepository discoveryCertificateRepository, AttributeEngine attributeEngine,
--- a/src/main/java/com/otilm/core/service/handler/discovery/V1DiscoveryProviderAdapter.java
+++ b/src/main/java/com/otilm/core/service/handler/discovery/V1DiscoveryProviderAdapter.java
@@ -1,0 +1,572 @@
+package com.otilm.core.service.handler.discovery;
+
+import com.otilm.api.exception.AttributeException;
+import com.otilm.api.exception.ConnectorException;
+import com.otilm.api.exception.DiscoveryException;
+import com.otilm.api.exception.NotFoundException;
+import com.otilm.api.interfaces.client.v1.DiscoverySyncApiClient;
+import com.otilm.api.model.client.discovery.DiscoveryDetailDto;
+import com.otilm.api.model.common.attribute.common.AttributeContent;
+import com.otilm.api.model.common.attribute.common.AttributeType;
+import com.otilm.api.model.common.attribute.common.DataAttribute;
+import com.otilm.api.model.common.attribute.common.MetadataAttribute;
+import com.otilm.api.model.connector.discovery.DiscoveryDataRequestDto;
+import com.otilm.api.model.connector.discovery.DiscoveryProviderCertificateDataDto;
+import com.otilm.api.model.connector.discovery.DiscoveryProviderDto;
+import com.otilm.api.model.connector.discovery.DiscoveryRequestDto;
+import com.otilm.api.model.core.auth.Resource;
+import com.otilm.api.model.core.connector.ConnectorDto;
+import com.otilm.api.model.core.discovery.DiscoveryStatus;
+import com.otilm.core.attribute.engine.AttributeEngine;
+import com.otilm.core.attribute.engine.records.ObjectAttributeContentInfo;
+import com.otilm.core.client.ConnectorApiFactory;
+import com.otilm.core.dao.entity.Certificate;
+import com.otilm.core.dao.entity.Connector;
+import com.otilm.core.dao.entity.Discovery;
+import com.otilm.core.dao.repository.CertificateRepository;
+import com.otilm.core.dao.repository.ConnectorRepository;
+import com.otilm.core.dao.repository.DiscoveryCertificateRepository;
+import com.otilm.core.dao.repository.DiscoveryRepository;
+import com.otilm.core.events.data.DiscoveryResult;
+import com.otilm.core.events.handlers.CertificateDiscoveredEventHandler;
+import com.otilm.core.events.handlers.DiscoveryFinishedEventHandler;
+import com.otilm.core.messaging.jms.producers.EventProducer;
+import com.otilm.core.model.discovery.DiscoveryContext;
+import com.otilm.core.service.CredentialInternalService;
+import com.otilm.core.service.DiscoveryProperties;
+import com.otilm.core.service.ResourceInternalService;
+import com.otilm.core.service.handler.CertificateHandler;
+import com.otilm.core.tasks.ScheduledJobInfo;
+import com.otilm.core.util.AttributeDefinitionUtils;
+import com.otilm.core.util.AuthHelper;
+import com.otilm.core.util.CertificateUtil;
+import java.security.cert.X509Certificate;
+import java.time.Duration;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.Semaphore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.TransactionStatus;
+import org.springframework.transaction.support.DefaultTransactionDefinition;
+
+/**
+ * The v1 discovery flow, extracted behavior-frozen from {@code DiscoveryServiceImpl}: start the run at the provider,
+ * wait for completion, download the certificate pages, and hand the staged results to processing. Method bodies moved
+ * unchanged; only the wiring is new.
+ *
+ * <p>
+ * The v1 discovery interface has no run lifecycle, so {@link #stop}, {@link #resume} and {@link #cancel} refuse with
+ * {@link UnsupportedOperationException} — callers map that to the contract's 422.
+ */
+@Component
+public class V1DiscoveryProviderAdapter implements DiscoveryProviderAdapter {
+
+    private static final Logger logger = LoggerFactory.getLogger(V1DiscoveryProviderAdapter.class);
+
+    private static final Semaphore downloadCertSemaphore = new Semaphore(10);
+
+    private final DiscoveryProperties discoveryProperties;
+    private final PlatformTransactionManager transactionManager;
+    private final DiscoveryRepository discoveryRepository;
+    private final ConnectorRepository connectorRepository;
+    private final CertificateRepository certificateRepository;
+    private final DiscoveryCertificateRepository discoveryCertificateRepository;
+    private final AttributeEngine attributeEngine;
+    private final CertificateHandler certificateHandler;
+    private final CredentialInternalService credentialService;
+    private final ResourceInternalService resourceService;
+    private final ConnectorApiFactory connectorApiFactory;
+    private final EventProducer eventProducer;
+
+    public V1DiscoveryProviderAdapter(DiscoveryProperties discoveryProperties,
+            PlatformTransactionManager transactionManager, DiscoveryRepository discoveryRepository,
+            ConnectorRepository connectorRepository, CertificateRepository certificateRepository,
+            DiscoveryCertificateRepository discoveryCertificateRepository, AttributeEngine attributeEngine,
+            CertificateHandler certificateHandler, CredentialInternalService credentialService,
+            ResourceInternalService resourceService, ConnectorApiFactory connectorApiFactory,
+            EventProducer eventProducer) {
+        this.discoveryProperties = discoveryProperties;
+        this.transactionManager = transactionManager;
+        this.discoveryRepository = discoveryRepository;
+        this.connectorRepository = connectorRepository;
+        this.certificateRepository = certificateRepository;
+        this.discoveryCertificateRepository = discoveryCertificateRepository;
+        this.attributeEngine = attributeEngine;
+        this.certificateHandler = certificateHandler;
+        this.credentialService = credentialService;
+        this.resourceService = resourceService;
+        this.connectorApiFactory = connectorApiFactory;
+        this.eventProducer = eventProducer;
+    }
+
+    @Override
+    public DiscoveryDetailDto start(UUID discoveryUuid, ScheduledJobInfo scheduledJobInfo) {
+        // reload discovery modal with all association since it could be in separate transaction/session due to async
+        DiscoveryContext context = loadDiscoveryContext(discoveryUuid);
+        Discovery discovery = context.getDiscovery();
+        if (context.getDiscoveryStatus() == DiscoveryStatus.FAILED) {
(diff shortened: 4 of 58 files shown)


src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderV1Adapter.java
```diff
@@ -0,0 +1,571 @@
...
+        // reload discovery modal with all association since it could be in separate transaction/session due to async
+        String message = null;
+        Connector connector = null;
+        List<DataAttribute> dataAttributes = null;
+        Discovery discovery = discoveryRepository.findWithTriggersByUuid(discoveryUuid);
+        try {
+            logger.info("Loading discovery context: name={}, uuid={}", discovery.getName(), discovery.getUuid());
+            connector = connectorRepository
+                    .findByUuid(discovery.getConnectorUuid())
+                    .orElseThrow(() -> new NotFoundException(Connector.class, discovery.getConnectorUuid()));
```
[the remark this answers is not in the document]
lubomirw: This line moved verbatim from `DiscoveryServiceImpl` — the extraction is deliberately behavior-frozen, and the missing-run NPE is the pre-existing legacy behavior that the factory's null-tolerance knowingly preserves (see `DiscoveryProviderAdapterFactory`'s javadoc). The controlled not-found outcome arrives with the v2 lifecycle rework in #1964, which rewrites this path.

src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
```diff
@@ -420,445 +358,29 @@ public void runDiscoveryAsync(UUID discoveryUuid) {
...
-            return finalizeDiscoveryInTx(context, true, null);
-        } catch (Exception e) {
-            logger.error("Error in discovery '{}' at provider: {}", discovery.getName(), e.getMessage());
-            context.setDiscoveryFailed("Error in provider during discovery: " + e.getMessage());
-            if (e instanceof InterruptedException) {
-                Thread.currentThread().interrupt();
-            }
-            return finalizeDiscoveryInTx(context, true, null);
+            return discoveryProviderAdapterFactory.forDiscovery(discovery).start(discoveryUuid, scheduledJobInfo);
+        } catch (UnsupportedDiscoveryVersionException e) {
```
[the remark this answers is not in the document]
lubomirw: Fixed in 74ddf971d: the placeholder's `start` now throws `UnsupportedDiscoveryVersionException`, so a run that gains a v2 association anyway ends terminal FAILED (with a `DISCOVERY_FINISHED` event) through the same refusal handling as routing refusals, instead of being left non-terminal by the async caller. Covered end-to-end by `runDiscoveryWithV2AssociationEndsTerminalWhileTheAdapterIsAPlaceholder`.

src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
```diff
@@ -420,445 +358,29 @@ public void runDiscoveryAsync(UUID discoveryUuid) {
...
+            // A routing refusal must still end as a terminal, user-visible run state: the async caller swallows
+            // whatever escapes here, and the scheduler expects a result rather than an exception.
+            logger.warn("Discovery {} cannot be dispatched: {}", discoveryUuid, e.getMessage());
+            discovery.setStatus(DiscoveryStatus.FAILED);
+            discovery.setConnectorStatus(DiscoveryStatus.FAILED);
+            // The curated text, not e.getMessage(): the raw message carries the connector-reported version
+            // string, which is unvalidated input — it stays in the log, like the REST handler's fixed body.
+            discovery.setMessage(UNSUPPORTED_VERSION_MESSAGE);
+            discovery.setEndTime(OffsetDateTime.now(ZoneOffset.UTC));
+            discoveryRepository.save(discovery);
```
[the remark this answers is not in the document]
lubomirw: Rule A stays green here — in CI and locally — because `CrudRepository.save` runs in the repository proxy's own transaction, so neither of the rule's stated failure modes applies to this call. The Writer convention is the project direction though, so the terminal transition moved into `DiscoveryWriter.markDispatchRefused` (74ddf971d).

Doing that surfaced a real bug the direct save was masking: a `NOT_SUPPORTED` method runs all its reads in one transaction-less synchronization scope sharing a single `EntityManager`, so re-reading the run there after the write resolves to the pre-refusal entity in the first-level cache. The writer therefore maps and returns the terminal detail from inside its own transaction.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
```diff
@@ -467,6 +468,20 @@ public ErrorMessageDto handleUnsupportedAuthorityVersionException(UnsupportedAut
...
+    /**
+     * Handler for {@link UnsupportedDiscoveryVersionException} — the discovery twin of the authority handler above,
+     * same reasoning: caller-fixable configuration, fixed body because the message carries an unvalidated
+     * connector-reported version string.
+     *
+     * @return a fixed message that names neither the discovery nor the version
+     */
+    @ExceptionHandler(UnsupportedDiscoveryVersionException.class)
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    public ErrorMessageDto handleUnsupportedDiscoveryVersionException(UnsupportedDiscoveryVersionException ex) {
```
ivosh: **~ Minor** | Maintainability

ExceptionHandlingAdvice.java:480 adds an `@ExceptionHandler(UnsupportedDiscoveryVersionException.class)` returning 400, but this PR leaves no request path that can reach it.

The exception is constructed in exactly two places: `DiscoveryProviderAdapterFactory.forDiscovery` (DiscoveryProviderAdapterFactory.java:192-209) and `V2DiscoveryProviderAdapter.start` (V2DiscoveryProviderAdapter.java:22). Grepping the diff for `forDiscovery(` / `discoveryProviderAdapterFactory` gives one production call site — `DiscoveryServiceImpl.runDiscovery` (DiscoveryServiceImpl.java:422) — and that method catches `UnsupportedDiscoveryVersionException` itself and converts it into a terminal FAILED run.

`grep -rn "runDiscovery" src/main/java` gives two entry points into it: `DiscoveryServiceImpl.runDiscoveryAsync` (`@Async`, so nothing it throws can reach the HTTP response for `DiscoveryControllerImpl.createDiscovery:100`) and `DiscoveryCertificateTask:106` (scheduler, no request). The only escape from the catch block is `.orElseThrow(() -&gt; e)` when `markDispatchRefused` finds no row — a delete racing the refusal — and even that fires on a thread with no request attached.

Contrast the neighbour this handler is modelled on: `handleUnsupportedAuthorityVersionException` (line 465) is genuinely reachable, because `AuthorityProviderAdapterFactory.forAuthority` is called from ~20 synchronous service methods (AuthorityInstanceServiceImpl, RaProfileServiceImpl, ExtendedAttributeServiceImpl, ClientOperationServiceImpl) that run inside the request thread.

The PR description states "REST surfaces the refusal as 400 with a fixed body; a scheduled/async run ends as a terminal FAILED run". The second half is implemented and tested; the first half is not true of this PR, which is why this is being raised rather than treated as disclosed deferral.

_Suggested:_ Either say in the PR description and the handler Javadoc that the 400 handler is staged for the v2 lifecycle endpoints (`stopDiscovery`/`resumeDiscovery`/`cancelDiscovery`) that will call the adapter synchronously, or drop the handler until a synchronous caller exists.

<sub>flagged by Correctness & races + Reference integrity + API contract & compat</sub>

src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
```diff
@@ -420,445 +365,27 @@ public void runDiscoveryAsync(UUID discoveryUuid) {
...
+            // string, which is unvalidated input — it stays in the log, like the REST handler's fixed body.
+            // The writer also maps the terminal detail; re-reading it here would resolve to this scope's stale
+            // first-level-cached entity (see the writer's javadoc).
+            DiscoveryDetailDto failedDetail = discoveryWriter
+                    .markDispatchRefused(discoveryUuid, UNSUPPORTED_VERSION_MESSAGE)
+                    .orElseThrow(() -> e);
+            eventProducer
+                    .produceMessage(DiscoveryFinishedEventHandler
+                            .constructEventMessage(discoveryUuid,
+                                    UUID.fromString(AuthHelper.getUserIdentification().getUuid()), null,
```
ivosh: **. Nitpick** | Error handling

**In this PR's refusal handler the acting user is resolved *after* the run has already been committed as FAILED, so a lookup failure loses the `DISCOVERY_FINISHED` event while leaving the row terminal.**

`DiscoveryServiceImpl.java:379-381` commits the terminal state through `discoveryWriter.markDispatchRefused(...)`. Only then, at line 385, does `UUID.fromString(AuthHelper.getUserIdentification().getUuid())` run. `AuthHelper.java:203-213` throws `ValidationException` when the thread carries no authentication, and `AuthHelper.java:220-222` documents that an anonymous caller's principal is real but has a **null** `userUuid` — which `UUID.fromString` turns into an NPE. Either way the run is already FAILED in the database, `eventProducer.produceMessage` never runs, and an exception escapes the very method whose comment at line 372-373 says "the scheduler expects a result rather than an exception".

This is narrow, not urgent: `runDiscovery` is `@ExternalAuthorization`-guarded, and `ContextAwarePoolExecutor` propagates the `SecurityContext` onto the `@Async` thread, so in practice the lookup succeeds. And the raw `UUID.fromString(AuthHelper.getUserIdentification().getUuid())` form is the established service-layer idiom here (thirteen sites in `src/main`).

What makes the refusal path different is the ordering. The v1 flow resolves the user uuid as the first statement of `loadDiscoveryContext` — before any state change — and carries it in the context. The refusal path resolves it last. `AuthHelper.getActingUserUuidOrNull()` exists for exactly this shape; its javadoc says it is "resolved on the producing thread so attribution can travel with an async message", and `DiscoveryFinishedEventHandler.constructEventMessage` passes `userUuid` straight into `EventMessage` with no null check, so a null is harmless.

_Suggested:_ Use `AuthHelper.getActingUserUuidOrNull()` for the event's user uuid, or resolve the uuid before calling `markDispatchRefused`, so the terminal write and its event cannot come apart.

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderAdapter.java
```diff
@@ -0,0 +1,39 @@
...
+ * its own transaction boundaries, because it executes asynchronously in a different session than the one that created
+ * the run.
+ */
+public interface DiscoveryProviderAdapter {
+
+    /** Runs the whole provider-side discovery for the run, returning the run detail at handoff or completion. */
+    DiscoveryDetailDto start(UUID discoveryUuid, ScheduledJobInfo scheduledJobInfo);
+
+    /**
+     * @throws UnsupportedOperationException when the connector version cannot suspend a run (v1 cannot); callers wiring
```
ivosh: **? Question** | Documentation — Minor if confirmed

**The interface this PR introduces documents one exception contract; the only other implementation in the PR throws different types.**

`DiscoveryProviderAdapter.java:22-38` says for `stop`, `resume` and `cancel`: `@throws UnsupportedOperationException ... callers wiring the lifecycle endpoints must map this to the contract's 422`. `V1DiscoveryProviderAdapter` honours that. `V2DiscoveryProviderAdapter.java:28,33,38` throws `IllegalStateException` instead, so a caller written against the Javadoc would map v1 to 422 and let v2 fall through to 500.

Same gap on `start`: `DiscoveryProviderAdapter.java:19-20` documents no exception, yet `V2DiscoveryProviderAdapter.java:23` throws `UnsupportedDiscoveryVersionException` — and `DiscoveryServiceImpl.java:371` catches exactly that type as the routing-refusal signal. The seam's most load-bearing behaviour is documented only on the implementation, not on the interface future callers will code against.

Two readings. If the split is deliberate — `UnsupportedOperationException` means "this generation cannot", `IllegalStateException` means "not built yet" — the interface should say so, because nothing in the tree distinguishes them today. If it is not deliberate, `V2DiscoveryProviderAdapter` should throw `UnsupportedOperationException` for the three lifecycle methods.

_Suggested:_ Document both exception types on the `DiscoveryProviderAdapter` methods — including `UnsupportedDiscoveryVersionException` on `start` — or make V2's lifecycle methods throw `UnsupportedOperationException` like V1.

src/test/java/com/otilm/core/service/handler/discovery/V1DiscoveryProviderAdapterTest.java
```diff
@@ -0,0 +1,23 @@
...
+import com.otilm.core.dao.entity.Discovery;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** Pins that the v1 interface rejects the lifecycle operations it never supported. */
+class V1DiscoveryProviderAdapterTest {
+
+    // The rejecting operations touch no collaborator, so the adapter is constructed without any.
+    private final V1DiscoveryProviderAdapter adapter = new V1DiscoveryProviderAdapter(null, null, null, null, null,
```
ivosh: **. Nitpick** | Testing

**A twelve-null constructor call is fragile in a way the compiler cannot help with.**

`V1DiscoveryProviderAdapterTest.java:12-13` constructs the adapter as `new V1DiscoveryProviderAdapter(null, null, null, null, null, null, null, null, null, null, null, null)`. Adding, removing or reordering a collaborator on the twelve-argument constructor silently keeps compiling here, and a reader cannot tell which argument is which.

Second point, same file: `lifecycleOperationsAreUnsupported` at `V1DiscoveryProviderAdapterTest.java:16-22` asserts three independent behaviours in one method — `stop`, `resume` and `cancel` — so the first failure hides the other two. `V2DiscoveryProviderAdapterTest.java:23-29` has the same shape.

Both are cheap to fix and neither is load-bearing, so take it or leave it.

_Suggested:_ Build the adapter with Mockito mocks (or a small named helper) instead of positional nulls, and split each lifecycle rejection into its own test method.

src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
```diff
@@ -420,445 +365,27 @@ public void runDiscoveryAsync(UUID discoveryUuid) {
...
-            context.setDiscoveryFailed("Failed to run discovery at the provider: " + e.getMessage());
-            throw new DiscoveryException(discovery.getName(), context.getMessage());
+            return discoveryProviderAdapterFactory.forDiscovery(discovery).start(discoveryUuid, scheduledJobInfo);
+        } catch (UnsupportedDiscoveryVersionException e) {
+            // A routing refusal must still end as a terminal, user-visible run state: the async caller swallows
+            // whatever escapes here, and the scheduler expects a result rather than an exception.
+            logger.warn("Discovery {} cannot be dispatched: {}", discoveryUuid, e.getMessage());
+            // The curated text, not e.getMessage(): the raw message carries the connector-reported version
+            // string, which is unvalidated input — it stays in the log, like the REST handler's fixed body.
+            // The writer also maps the terminal detail; re-reading it here would resolve to this scope's stale
```
ivosh: **. Nitpick** | Documentation

**The same rationale is written twice, and the second copy points at the first.**

`DiscoveryServiceImpl.java:377-378` says: "The writer also maps the terminal detail; re-reading it here would resolve to this scope's stale first-level-cached entity (see the writer's javadoc)." That fact already has an authoritative home in `DiscoveryWriter.java:70-74`, which explains the `NOT_SUPPORTED` / shared-`EntityManager` mechanism in full — and the comment explicitly refers the reader there.

Two copies of one invariant drift apart the moment either side changes. Keeping the pointer and dropping the restatement loses nothing.

_Suggested:_ Trim `DiscoveryServiceImpl.java:377-378` to the cross-reference alone ("the writer maps the detail — see its Javadoc for why this scope cannot re-read it").

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderV1Adapter.java
```diff
@@ -0,0 +1,571 @@
...
+        this.attributeEngine = attributeEngine;
+        this.certificateHandler = certificateHandler;
+        this.credentialService = credentialService;
+        this.resourceService = resourceService;
+        this.connectorApiFactory = connectorApiFactory;
+        this.eventProducer = eventProducer;
+    }
+
+    @Override
+    public DiscoveryDetailDto start(UUID discoveryUuid, ScheduledJobInfo scheduledJobInfo) {
```
ivosh, line 115: **. Nitpick** | Maintainability

**The verbatim move is the cheapest moment this method will ever have to get shorter.**

`V1DiscoveryProviderAdapter.java:115-211` is 97 lines, with block comments marking each phase — `// discover certificates by provider` (line 123), `// download and create discovered certificates` (line 144), `// process duplicates` (line 155). Those are exactly the extraction points.

The repository's CLAUDE.md is direct about the shape: "Long methods (&gt;80–100 lines) with heavy comment overhead are a smell. Refactor into named helpers before requesting human review" and "Replace block comments explaining a multi-step sequence with extracted methods whose names describe the step." The neighbouring private methods in the same class — `discoverCertificatesByProvider`, `downloadDiscoveredCertificates` — already follow that pattern.

The counter-argument is real and this PR states it: the flow was moved verbatim so the existing discovery ITest suite keeps its meaning. Extracting the three phases into named private methods is pure structure with no behaviour change, so it would not weaken that guarantee — but it does widen the review surface, which is a fair reason to defer.

_Suggested:_ Extract the three commented phases of `start` into named private methods, or note in the PR that the split is deliberately deferred to the v2 work.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
```diff
@@ -467,6 +468,20 @@ public ErrorMessageDto handleUnsupportedAuthorityVersionException(UnsupportedAut
         return ErrorMessageDto.getInstance("The authority's connector interface version is not supported.");
     }
 
+    /**
+     * Handler for {@link UnsupportedDiscoveryVersionException} — the discovery twin of the authority handler above,
```
ivosh: **. Nitpick** | Documentation

**The new Javadoc packs four facts into one 40-word sentence, right below a twin that separates them.**

`ExceptionHandlingAdvice.java:472-474` fuses what it handles, that it mirrors the authority handler, why 400 rather than 500, and why the body is fixed — all into a single em-dash-and-colon sentence with no structural break.

Its model is twelve lines above. `ExceptionHandlingAdvice.java:453-462` opens with a one-line summary ("Handler for `{@link UnsupportedAuthorityVersionException}`"), then puts the rationale in its own `&lt;p&gt;` with short sentences. Same facts, readable in one pass.

Separately, "the authority handler above" is a positional reference: it breaks silently if the handlers are ever reordered, and CLAUDE.md's "Writing for future AI sessions" section asks for docs that stand on their own. `{@link #handleUnsupportedAuthorityVersionException}` survives a move.

_Suggested:_ Restructure the Javadoc as a one-line summary plus a `&lt;p&gt;` rationale mirroring `handleUnsupportedAuthorityVersionException`, and replace "the authority handler above" with an `{@link}` to it.

src/test/java/com/otilm/core/integration/repository/DiscoveryRepositoryITest.java
```diff
@@ -1,5 +1,6 @@
 package com.otilm.core.integration.repository;
 
+import com.otilm.api.model.connector.discovery.v2.DiscoveryProgressDto;
```
ivosh, line 3: **. Nitpick** | Documentation

DiscoveryRepositoryITest.java:25 still reads "The critical case is {@code progressByResource}, the first enum-keyed map under {@code @JdbcTypeCode(SqlTypes.JSON)} in any core entity". This PR removes that field from `Discovery` (Discovery.java, the four `progress_*` columns collapse into `progress`), so no `progressByResource` symbol exists after the change — the enum-keyed map now lives one level down as `DiscoveryProgressDto.byResource`.

The test body was updated to match (`run.setProgress(progress)`, `back.getProgress().getByResource()`), but the Javadoc above it was not. The reasoning it records is still correct; only the symbol name is stale. `gh search code --owner OmniTrustILM --owner 3KeyCompany progressByResource` confirms this Javadoc line is the only surviving reference to the old name outside the lines this PR rewrites.

_Suggested:_ Rename the Javadoc reference to `progress.byResource` (or `DiscoveryProgressDto#getByResource`) so it points at a symbol that still exists.

src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
```diff
@@ -420,445 +365,27 @@ public void runDiscoveryAsync(UUID discoveryUuid) {
...
-        dtoRequest.setName(discovery.getName());
-        dtoRequest.setKind(discovery.getKind());
-        dtoRequest.setAttributes(AttributeDefinitionUtils.getClientAttributes(context.getDataAttributes()));
-
-        DiscoverySyncApiClient discoveryApiClient = connectorApiFactory
-                .getDiscoveryApiClient(context.getConnectorDto());
-
-        // start discovery at provider
-        DiscoveryProviderDto response;
+        Discovery discovery = discoveryRepository.findByUuid(discoveryUuid).orElse(null);
```
ivosh, line 368: **. Nitpick** | Maintainability

`DiscoveryServiceImpl.java:368` loads the full `Discovery` entity purely to route, and the only field the routing actually reads is `connectorInterfaceUuid` (`DiscoveryProviderAdapterFactory.forDiscovery`). Two costs follow from making it an entity load rather than a scalar read.

First, it is what creates the stale-cache hazard this PR then has to document its way around. `runDiscovery` is `@Transactional(propagation = Propagation.NOT_SUPPORTED)`, and by the reasoning in the new `DiscoveryWriter.markDispatchRefused` javadoc (`DiscoveryWriter.java:70-74`) that scope shares one `EntityManager` across all its reads — so this line is precisely what puts the pre-refusal run row into that first-level cache and forces the DTO to be mapped inside the writer. A scalar projection of `connector_interface_uuid` would route identically while seeding nothing into the persistence context, and the workaround would stop being load-bearing.

Second, on the v1 path the row is immediately read again: `V1DiscoveryProviderAdapter.loadDiscoveryContext` reloads it with `findWithTriggersByUuid`, so every v1 run now pays two reads of the same row where it previously paid one.

_Suggested:_ Route on a scalar projection of `connector_interface_uuid` (or move that projection read into `DiscoveryProviderAdapterFactory`) so the NOT_SUPPORTED scope's persistence context never holds the run row.

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderV1Adapter.java
```diff
@@ -0,0 +1,571 @@
...
+            logger.info("Loading discovery context: name={}, uuid={}", discovery.getName(), discovery.getUuid());
+            connector = connectorRepository
+                    .findByUuid(discovery.getConnectorUuid())
+                    .orElseThrow(() -> new NotFoundException(Connector.class, discovery.getConnectorUuid()));
+            dataAttributes = attributeEngine
+                    .getDefinitionObjectAttributeContent(AttributeType.DATA, connector.getUuid(), null,
+                            Resource.DISCOVERY, discovery.getUuid());
+
+            credentialService.loadFullCredentialData(dataAttributes);
+            resourceService.loadResourceObjectContentData(dataAttributes);
```
ivosh, line 250: **~ Minor** | Performance

`V1DiscoveryProviderAdapter.loadDiscoveryContext` opens an explicit read-only transaction at V1DiscoveryProviderAdapter.java:233 and does not commit it until line 266. Line 250, `resourceService.loadResourceObjectContentData(dataAttributes)`, sits inside that window.

That call is not always DB-only. `ResourceServiceImpl.loadResourceObjectContentData` reaches `getResourceObjectContentData` (ResourceServiceImpl.java:373-384), whose else-branch dispatches to `attributeResourceServices.get(resource.getCode()).getResourceObjectContent(uuid)`. For a SECRET-typed resource attribute that lands in `SecretServiceImpl.getResourceObjectContent` (SecretServiceImpl.java:1113) → `getSecretContent` → `connectorApiFactory.getSecretApiClient(...).getSecretContent(...)` at SecretServiceImpl.java:856-858 — an HTTP call to the vault connector. A discovery whose data attributes reference a stored secret therefore holds a database connection for the vault's whole response time.

CLAUDE.md's "Transactions and external calls" rule is explicit that a transaction must not be held across an HTTP call to an external system, and the rest of this flow follows the rule well — the provider calls all run between commit points.

The body was moved verbatim from `DiscoveryServiceImpl`, so this PR does not introduce the span; the extraction is just the first time it is visible as this class's own contract. If the author prefers to keep the freeze intact, the honest option is a follow-up issue rather than an edit here.

_Suggested:_ Split loadDiscoveryContext so the read-only transaction covers only the run/connector/attribute-definition reads and commits before `credentialService.loadFullCredentialData` and `resourceService.loadResourceObjectContentData` run, or file a follow-up if the verbatim freeze must hold.

src/main/java/com/otilm/core/service/handler/discovery/V1DiscoveryProviderAdapter.java
```diff
@@ -0,0 +1,571 @@
...
+                logger.warn("Failed to serialize discovery metadata");
+            }
+        }
+
+        return discovery;
+    }
+
+    private DiscoveryDetailDto finalizeDiscoveryInTx(DiscoveryContext discoveryContext, boolean updateMetadata,
+            String preProcessingMessage) {
+        TransactionStatus transaction = transactionManager.getTransaction(new DefaultTransactionDefinition());
```
ivosh: **~ Minor** | Error handling

All four explicit transactions in `V1DiscoveryProviderAdapter` are opened and committed with nothing in between to close them on failure: 233/266 in `loadDiscoveryContext`, 154/190 in `start`, 518/521 in `updateDiscoveryStateInTx`, and 550/562 in `finalizeDiscoveryInTx`. CLAUDE.md's "Transactions and external calls" pattern spells out the shape these should have — `try { … commit(tx); } catch (RuntimeException e) { rollback(tx); throw e; }`. A throw between `getTransaction` and `commit` leaves the transaction and its connection bound to the thread, and because `DiscoveryServiceImpl.runDiscovery` is `NOT_SUPPORTED` there is no outer boundary that will clean it up; on the `@Async` executor or the scheduler thread that is a leaked pooled connection.

One such path is reachable through routing this PR newly documents. `DiscoveryServiceImpl.java:368` loads the run with `findByUuid(...).orElse(null)`, and `DiscoveryProviderAdapterFactory.forDiscovery` sends `null` to v1 on the stated grounds that an unloadable run "belongs to the v1 flow's established handling". Following that path: `loadDiscoveryContext` calls `discoveryRepository.findWithTriggersByUuid` at line 239, which returns a bare `Discovery` and can be `null`; the NPE at line 241 is swallowed by the catch at 252; `DiscoveryContext` stores the null discovery without a guard (DiscoveryContext.java:29-34); `start` sees FAILED at line 119 and calls `finalizeDiscoveryInTx` at 120; that opens a transaction at line 550 and then NPEs inside `updateDiscoveryState` on `discovery.setStatus(...)` at line 530 — with the transaction from 550 never committed nor rolled back.

The behaviour is carried over verbatim from `DiscoveryServiceImpl`, so this PR does not introduce it. If the freeze is meant to cover it, the factory javadoc's claim that this failure is "handled" by v1 is the part worth correcting.

_Suggested:_ Wrap each explicit transaction in try/commit/catch-rollback-rethrow as CLAUDE.md's pattern shows, or, if the behaviour freeze must hold, drop the factory javadoc's claim that the v1 flow handles an unloadable run and file a follow-up.

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderV1Adapter.java
```diff
@@ -0,0 +1,571 @@
...
+
+        } catch (Exception e) {
+            message = e.getMessage();
```
ivosh, line 253: **~ Minor** | Security

`V1DiscoveryProviderAdapter.java:253` assigns a bare `e.getMessage()` as the run's message, and `V1DiscoveryProviderAdapter.java:137` and `:292` concatenate a raw `e.getMessage()` onto a prefix the same way. All three land in `DiscoveryContext.message`, which `V1DiscoveryProviderAdapter.java:529` writes to the entity via `discovery.setMessage(...)`, and `Discovery.mapToDto()` at `Discovery.java:167` publishes as `DiscoveryDetailDto.message` — so an unshaped runtime message reaches a REST response body. Line 253 is the sharpest instance: the `try` it guards wraps `credentialService.loadFullCredentialData(...)` and `resourceService.loadResourceObjectContentData(...)` (`V1DiscoveryProviderAdapter.java:249-250`), so a secret-store, attribute-engine or Hibernate failure is copied through verbatim with not even a prefix. CLAUDE.md's "Don't leak runtime details to the wire" asks for a controlled domain-exception gate on exactly this surface. Raising it as a follow-up rather than a change to this PR: the bodies are disclosed as moved verbatim and behavior-frozen, and this PR's own new refusal path at `DiscoveryServiceImpl.java:417` deliberately does the opposite by persisting the constant `UNSUPPORTED_VERSION_MESSAGE`. The extraction is the moment the two conventions become visible side by side.

_Suggested:_ Open a follow-up issue to shape the v1 adapter's failure messages behind a controlled domain-exception check before they are persisted as the run message, matching the curated text this PR's refusal path already uses.

src/main/java/com/otilm/core/service/handler/discovery/V2DiscoveryProviderAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+ * synchronous operations fail loud with {@link IllegalStateException}.
+ */
+@Component
+public class V2DiscoveryProviderAdapter implements DiscoveryProviderAdapter {
+
+    private static final String NOT_IMPLEMENTED = "The discovery v2 provider adapter is not implemented yet";
+
+    @Override
+    public DiscoveryDetailDto start(UUID discoveryUuid, ScheduledJobInfo scheduledJobInfo) {
+        throw new UnsupportedDiscoveryVersionException(NOT_IMPLEMENTED);
```
ivosh: **. Nitpick** | Documentation

The PR description says `V2DiscoveryProviderAdapter` is a "placeholder; every operation fails with `IllegalStateException`". V2DiscoveryProviderAdapter.java:23 throws `UnsupportedDiscoveryVersionException` from `start`, and only stop/resume/cancel (lines 28, 33, 38) use `IllegalStateException`. The class Javadoc gets this right, and the distinction is load-bearing: it is precisely because `start` throws the routing exception that a v2-associated run is caught by `DiscoveryServiceImpl.runDiscovery` and ends terminal FAILED instead of leaking out of the async caller. A reviewer who trusts the PR body will read the v2 start path as a leak.

_Suggested:_ Correct the PR body's v2 bullet to say start refuses with UnsupportedDiscoveryVersionException and only the lifecycle operations throw IllegalStateException.

src/main/java/com/otilm/core/service/handler/discovery/V2DiscoveryProviderAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @Override
+    public void stop(Discovery discovery) {
+        throw new IllegalStateException(NOT_IMPLEMENTED);
```
ivosh: **~ Minor** | Error handling

V2DiscoveryProviderAdapter.java:28, :33 and :38 throw `IllegalStateException` for `stop`/`resume`/`cancel`, but the interface this PR adds documents a different contract for exactly those three methods: `DiscoveryProviderAdapter.java:22`, `:28` and `:34` each declare `@throws UnsupportedOperationException ... callers wiring the lifecycle endpoints must map this to the contract's 422`. The v1 sibling honours that contract (`V1DiscoveryProviderAdapter` stop/resume/cancel all throw `UnsupportedOperationException`); only the v2 placeholder diverges, so a future caller that maps the documented type gets a 422 for v1 and an unmapped 500 for v2.

The repository's own idiom points the same way: `CmpServiceImpl.java:244` (`case V3 -&gt; throw new UnsupportedOperationException("not implemented")`) and `ProtectionPBMac1Validator.java:17` use `UnsupportedOperationException` for not-yet-implemented operations, while the neighbouring adapter package reserves `IllegalStateException` for genuinely unexpected state (`AuthorityProviderV3Adapter.java:493`, `:510` — "Unexpected v3 status"). *Effective Java* Item 72 (favour standard exceptions) puts "object does not support the requested operation" under `UnsupportedOperationException`; nothing about the placeholder's *state* is illegal.

If the split is deliberate — v1 can never do this, v2 cannot do it yet, and you want the two to surface as 422 vs 500 — that reasoning is not in the interface, which currently promises one type for all implementations.

_Suggested:_ Throw `UnsupportedOperationException` from the v2 placeholder's `stop`/`resume`/`cancel` to match the interface's `@throws` contract, or amend the three `@throws` clauses in `DiscoveryProviderAdapter` to state that a not-yet-implemented version throws `IllegalStateException` instead.

<sub>flagged by Java engineering & idioms + API contract & compat</sub>

src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
```diff
@@ -105,61 +78,63 @@
...
     private static final Logger logger = LoggerFactory.getLogger(DiscoveryServiceImpl.class);
 
-    private static final Semaphore downloadCertSemaphore = new Semaphore(10);
-
-    private final DiscoveryProperties discoveryProperties;
-
-    private EventProducer eventProducer;
-    private NotificationProducer notificationProducer;
-    private PlatformTransactionManager transactionManager;
+    private static final String UNSUPPORTED_VERSION_MESSAGE = "The discovery's connector interface version is not supported.";
```
ivosh: **. Nitpick** | Maintainability

`DiscoveryServiceImpl.java:90` defines `UNSUPPORTED_VERSION_MESSAGE = "The discovery's connector interface version is not supported."`, and `ExceptionHandlingAdvice.java:482` hard-codes the byte-identical literal for the 400 body. The comment at `DiscoveryServiceImpl.java:375-376` explicitly ties the two together ("it stays in the log, like the REST handler's fixed body"), so this PR is treating them as one user-visible string held in two places — a wording change to either surface silently makes the async run message and the REST body disagree. This is the one thing the authority twin does not have to worry about: `UnsupportedAuthorityVersionException` has a single copy, in the advice.

_Suggested:_ Hoist the fixed refusal text to one shared constant (e.g. on `UnsupportedDiscoveryVersionException`) and reference it from both `DiscoveryServiceImpl` and `ExceptionHandlingAdvice`.

src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
lubomirw: Fixed in 880cad371: the acting user now resolves via `AuthHelper.getActingUserUuidOrNull()` as the first statement of the catch, before `markDispatchRefused`, so the terminal write and its event cannot come apart.

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderAdapter.java
lubomirw: The split is deliberate, and 880cad371 makes the interface say so: `start` now documents `UnsupportedDiscoveryVersionException` as the routing-refusal signal, and the lifecycle methods state that `UnsupportedOperationException` means "this generation can never" (contract, maps to 422) while a pending implementation's `IllegalStateException` is a defect outside the contract. V2 *will* support stop/resume/cancel, so having the placeholder throw `UnsupportedOperationException` would give future wiring the wrong 422 signal.

src/test/java/com/otilm/core/service/handler/discovery/V1DiscoveryProviderAdapterTest.java
lubomirw: Both taken in 880cad371: the adapter builds with named Mockito mocks, and each lifecycle rejection asserts in its own test method — in both adapter test classes.

src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
lubomirw: Trimmed in 880cad371 to the pointer alone; the writer javadoc keeps the one authoritative copy.

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderV1Adapter.java
lubomirw: Taking the offered alternative: deferred deliberately, and the PR body now says so. The verbatim freeze is this PR's core review guarantee, and even a pure-structure split widens the surface reviewers must re-verify. The extraction happens with the v2-era rework in #1964.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
lubomirw: Restructured in 880cad371: one-line summary plus a `<p>` rationale mirroring the authority handler, and `{@link #handleUnsupportedAuthorityVersionException}` instead of the positional reference.

src/test/java/com/otilm/core/integration/repository/DiscoveryRepositoryITest.java
lubomirw: Fixed in 880cad371 — the javadoc now points at `DiscoveryProgressDto#byResource`.

src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
lubomirw: Agreed on the mechanics — the entity load is what seeds the stale L1 cache and double-reads on v1. Deferring the projection to #1964 though: it reshapes the factory contract this PR just built and tested, and #1964 rewrites dispatch anyway. The issue now carries the note ("Routing goes scalar here") with this thread's reasoning, so the writer's in-transaction mapping stops being load-bearing there.

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderV1Adapter.java
lubomirw: Follow-up filed as #2107 ("Harden the extracted v1 discovery flow") — first AC splits `loadDiscoveryContext` so no transaction spans the credential/resource loaders. Keeping the freeze intact here, as suggested.

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProviderV1Adapter.java
lubomirw: Same follow-up, #2107 — second AC shapes the v1 flow's failure messages behind a controlled domain-exception gate, matching the curated text the refusal path uses. Deferred rather than edited here to keep the verbatim freeze, as the comment itself suggests.

ivosh (approved):

Issue #2026, opened 2026-08-12
Comment event handlers and notification delivery


klaraf755:
### Description

Delivers the event side of the comment subsystem (Epic: Adding comments to objects, OmniTrustILM/ilm#246): handlers for the two comment events and their notification paths. The events themselves (`COMMENT_CREATED`, `COMMENT_RESOLVED`, shared `CommentEventData`) are defined in the interfaces issue; the after-commit publishing calls live in the core subsystem issue. This issue makes the events do something.

**Handlers** - package `com.otilm.core.events.handlers`, registered by bean name equal to the event code (existing convention):

- `@Component(ResourceEvent.Codes.COMMENT_CREATED)` extending `EventHandler<Comment>`, and a sibling `@Component(ResourceEvent.Codes.COMMENT_RESOLVED)` sharing a common base - one bean per event code; they differ in event data population and follow-up notifications.
- `getEventData(...)` maps the `Comment` entity to `CommentEventData` (verbatim Markdown body, host resource, object UUID, author, `parentUuid`; resolution fields populated only by the resolved handler).

**Group-scoped trigger resolution** - the base `EventHandler` fetches only platform-level trigger associations; overriding-resource resolution is per-handler (`getOverridingTriggers()` defaults to empty - see `CertificateEventsHandler` for the precedent, which walks the certificate's groups and RA profile). The comment handlers override it **generically**: resolve the host object's group memberships from the polymorphic `GroupAssociation` table by (`hostResource`, `hostObjectUuid`) - one query, no per-resource code - and call `fetchEventTriggers(context, Resource.GROUP, groupUuid)` for each. Effective for the `hasGroups` resources (certificates, keys, secrets); a no-op otherwise. This is what makes a trigger or notification profile associated at a group fire for comments on that group's objects.

**Internal (in-app) notifications** - direct, configuration-free delivery via `sendFollowUpEventsNotifications()` (precedent: `ApprovalRequestedEventHandler`). Thread participants = the root author plus the distinct authors of its replies, minus the acting user (never a self-notification):

- `COMMENT_CREATED` for a **root** comment notifies the host object's owner via the built-in `OWNER` recipient type, which resolves through `OwnerAssociation`. For this to work the `NotificationMessage` must carry the **host** resource and object UUID (not `Resource.COMMENT` / the comment UUID). Effective for `hasOwner` resources; resolves to nobody and is skipped elsewhere.
[...]

PR #2105, opened 2026-08-19
Add comment threads across supported resources

### LINKED ISSUE #2025: Polymorphic comment subsystem: entity, migration, service and REST implementation
### Description

Implements the comment subsystem in core against the API contract defined in the `interfaces` issue (Epic: Adding comments to objects, OmniTrustILM/ilm#246). Comments form one-level threads: a top-level comment is the thread root and the resolvable unit; replies attach to a root and are never resolvable.

**Persistence** - mirror the `Approval` attach pattern (`dao/entity/Approval.java`):

- Flyway migration `V<yyyyMMddHHmm>__comment.sql`: table `comment` with `uuid`, `object_uuid UUID NOT NULL`, `resource TEXT NOT NULL` (`@Enumerated(STRING)` on the entity), `parent_uuid UUID NULL` (self-referencing FK; NULL = thread root), `author_uuid`, `author_username`, `created_at`, `body TEXT` (verbatim Markdown source), `resolved_at`, `resolved_by_uuid`, `resolved_by_username`. A CHECK constraint keeps resolution columns NULL on replies (`parent_uuid IS NOT NULL`).
- **Index on `(resource, object_uuid)`** from the first migration - every listing filters on that pair. (The `approval` table lacks this; do not copy that omission.) A second index on `parent_uuid` serves reply loading and participant queries.
- `Comment` entity in `dao/entity/`, repository extending `SecurityFilterRepository` in `dao/repository/`.

**Service** - one generic service, no per-resource branching. Reference implementations: `EventControllerImpl` / its service for a generic resource+uuid sub-collection, `ApprovalServiceImpl` for the service layout.

- Requests are validated against `Resource.commentableResources`; any other resource is rejected with a validation error, not silently accepted.
- Body validation: required, non-blank, capped at the contract maximum (65536 characters); oversized input is rejected with a validation error, not truncated.
- Threading: depth is exactly one - a `parentUuid` pointing at a reply, or at a comment on a different object, is rejected.
- Author identity and timestamp come from the authenticated principal (`AuthHelper`), never from the payload.
- Write path follows the core orchestrator + writer convention (`service/writer/`, short REQUIRED-transaction writer bean; no `@Transactional` on repositories).
- Mutability (ratified on the Epic): append-only - create, resolve, unresolve, delete; no edit.
  - **Resolve/unresolve**: thread roots only; permitted for the root author or any user holding `COMMENT` on the host object. Records actor and timestamp.
  - **Delete a reply**: permitted for the reply author, the host object's owner (via `OwnerAssociation`, where the association exists), or a holder of `UPDATE` on the host object.
  - **Delete a root**: while the root has no replies, same rule as a reply. Once replies exist, only the host object's owner or an `UPDATE` holder may delete it, and the delete cascades to the replies - a root author must not be able to erase other users' words. Audit-logged with the full text, so nothing is silently lost.
- The list path is a plain paged query over thread roots carrying per-root reply counts; replies page through their own endpoint - no filtering in round one.

**Authorization** - read gate + capability model (supersedes the Epic's original two-branch table, decided 2026-08-18), enforced with the existing machinery, no OPA/auth-repo change:

- **Read** a thread = read access to the host object, delivered by the resource's own authorized lookup `ResourceExtensionService.getResourceObjectExternal(SecuredUUID)` (`DETAIL`-level, the same granularity the platform already applies to the object detail, including group/owner fallback). The one call also covers object existence and supplies the object name for events. Enforced server-side on both detail and list paths. Coverage gaps to close first: `VAULT` and `NOTIFICATION_PROFILE` do not implement `ResourceExtensionService` yet.
- **Post** (root or reply) = the read gate above **plus** `ResourceAction.COMMENT` on the host resource via imperative `AuthorizationEnforcer.enforce`:
  - host resource with object access (15 of the 22): `enforce(resource, COMMENT, objectUuid)` - object-level or resource-level grant on that specific object;
  - host resource without object access (`CERTIFICATE`, `CRYPTOGRAPHIC_KEY`, `SECRET`, `DISCOVERY`, `APPROVAL`, `APPROVAL_PROFILE`, `NOTIFICATION_PROFILE`): `enforce(resource, COMMENT)` resource-level.
  - No comment-specific parent resolution: parent/owner/group scoping arrives through the read gate exactly as the platform applies it to the object's detail - can read plus holds `COMMENT` means can post.
- Do not hold a DB lock across enforcement or the read gate (blocking OPA HTTP request).
- **Registration**: `ContextRefreshListener` scans only the static and programmatic authorization annotations, so a `COMMENT` action enforced only dynamically would never reach the auth-service catalogue - and the sync deletes unregistered actions together with their grants. Extend the registration path so `COMMENT` is pushed for every resource in `Resource.commentableResources` (this is what makes the EnumSet the single source of truth).

**Controller and audit** - `CommentControllerImpl` in `api/web/` implementing the `CommentController` interface. Audit pattern confirmed against `EventControllerImpl.getObjectEventHistory`: `@AuditLogged(module = Module.CORE, resource = Resource.COMMENT, operation = ...)` with the dynamic host captured per-parameter - `@LogResource(affiliated = true)` on the `Resource` parameter and `@LogResource(uuid = true, affiliated = true)` on the object UUID parameter. The comment body must reach the audit log content - the failure mode of OmniTrustILM/core#1453 (approval comment lost between API and audit log) is covered by test here.

**Host-object deletion** - deleting a commented object must not leave orphaned rows. Cascade delete of threads (roots and replies) via the generic `(resource, object_uuid)` key, exercised per resource in the parameterized suite.

**Events** - creating a comment (root or reply) publishes `COMMENT_CREATED`, and resolving or reopening a thread publishes `COMMENT_RESOLVED`, after commit (`TransactionSynchronizationManager`, `EventProducer`), with `CommentEventData` fully populated (`parentUuid` distinguishes replies; the `resolved` flag distinguishes resolve from reopen). Delete publishes nothing. The handlers are a separate core issue.

**Authorization test matrix** - integration tests **parameterized over `Resource.commentableResources`** - one test definition, executed for all 22 resources, so a later addition to the set extends coverage automatically. For each resource, both the detail and list paths, four principal profiles:

1. no read access on the host object -> sees no comments, cannot post;
2. read + `COMMENT` (object-level where supported, resource-level otherwise) -> sees, can post; cannot update the host object;
3. read without `COMMENT` -> sees, cannot post;
4. auditor role -> sees, cannot post (auto-derived role must never acquire `COMMENT`).

Plus the read-gate scoping for certificate/key/secret (resource-level `COMMENT` with no read path denies both list and post; opening any read path - e.g. group `MEMBERS` or an owner association - together with resource-level `COMMENT` enables both), and the resolve/delete permission branches (non-author `COMMENT` holder resolves a thread; a reply author deletes their reply; the object owner and an update holder delete a root with replies, cascading; a root author cannot delete their root once it has replies; a plain reader can do none of these).

### User Story

As a platform user who lacks the permission to make a change myself, I want to raise the request as a comment thread on the object concerned, so that the request, the reasoning and the outcome stay attached to the object and leave an auditable trail.

### Use Case

Epic use cases 1-4 (one-level threads; the thread root is the resolvable unit).

### Acceptance Criteria

- [ ] A comment is stored against (`resource`, `objectUuid`) with no per-object schema change; the table carries an index on that pair; a resource outside `Resource.commentableResources` is rejected.
- [ ] Author and creation timestamp come from the authenticated principal, never from the client payload.
- [ ] Body is required, non-blank and capped at the contract maximum; oversized input is rejected, not truncated.
- [ ] Thread depth is exactly one: a reply to a reply, or a `parentUuid` on a different object, is rejected.
- [ ] Only thread roots can be resolved/unresolved; resolution records actor and timestamp; a CHECK constraint keeps resolution columns NULL on replies.
- [ ] Resolve is permitted for the root author or a `COMMENT` holder on the host object; reply deletion for the author, the host object's owner, or an `UPDATE` holder; root deletion follows the no-replies/with-replies rule with cascade - each branch covered by test.
- [ ] Comment visibility equals read access to the host object, enforced server-side on both detail and list paths.
- [ ] Posting requires the host-object read gate (`getResourceObjectExternal`) plus `ResourceAction.COMMENT`: object-level where the resource has object access, resource-level where it does not.
- [ ] `COMMENT` is registered with the auth service for exactly the resources in `Resource.commentableResources`; the auditor role never contains it.
- [ ] The authorization suite is parameterized over `Resource.commentableResources` and covers the four principal profiles on both paths for all 22 resources.
- [ ] All mutations are `@AuditLogged` with the host as affiliated resource/object via `@LogResource` parameter marking; the audit record carries the comment body (core#1453 failure mode covered by test).
- [ ] Deleting a host object removes its threads on every supported resource, leaving no orphaned rows - exercised per resource.
- [ ] Creating a comment or reply publishes `COMMENT_CREATED`, and resolving or reopening a thread publishes `COMMENT_RESOLVED`, after commit with `CommentEventData` populated (verbatim body, host resource, object, `parentUuid`; on the resolved event the `resolved` flag and the acting user of the state change).
- [ ] The Markdown source is persisted verbatim; rendered HTML is never stored.
- [ ] Sonar quality gate passes; new code has >=80% test coverage.

klaraf755:
Closes #2025

Implementation of the polymorphic comment subsystem against the contract in OmniTrustILM/interfaces#848 (Epic: OmniTrustILM/ilm#246). **Depends on the interfaces PR — merge that first.**

## What's here

- **Persistence** — `comment` table (Flyway `V202608191000`) keyed by `(resource, object_uuid)` with an index on that pair, self-referencing `parent_uuid` FK `ON DELETE CASCADE`, and a CHECK keeping resolution columns NULL on replies. `Comment` entity mirrors both constraints for the entity-generated test schema (the `DiscoveryItem` precedent); `CommentRepository` on `SecurityFilterRepository`; DTO mapping in `com.otilm.core.mapper.comment.CommentMapper`.
- **Authorization — read gate + capability model** (supersedes the Epic's original two-branch table; issue updated): reading a thread = the host resource's own `ResourceExtensionService.getResourceObjectExternal` (its `DETAIL` chain, including owner/group fallback), posting = read gate + `ResourceAction.COMMENT`. No comment-specific parent resolution. Gap-fill: VAULT, NOTIFICATION_PROFILE and APPROVAL_PROFILE now implement `ResourceExtensionService` under code-named beans.
- **Registration** — `ContextRefreshListener` pushes `COMMENT` for every resource in `Resource.commentableResources`, so the auth-service sync never drops the dynamically-enforced action; the derived auditor role provably never contains it.
- **Service** — orchestrator + `CommentWriter` pair (NOT_SUPPORTED orchestrator, short REQUIRED writer transactions); append-only mutability with the ratified resolve/delete rules; author identity from the principal; events published after commit.
- **Audit** — `CommentControllerImpl` `@AuditLogged` with the host as affiliated resource. New request-scoped `AuditOperationDataOverride` (mirrors `AuditResultOverride`) so void-response mutations (delete, resolve) carry the full comment text into the audit record — the historical drop-the-text failure mode is covered by test, including delete of a root with replies.
- **Host-object deletion** — every commentable resource's delete path removes the object's threads; the parameterized suite drives a real object of each resource through its actual service delete. APPROVAL is the documented exception (the platform has no approval delete path).
- **Events** — `COMMENT_CREATED` (roots and replies), `COMMENT_RESOLVED` (resolve and reopen, `resolved` flag); nothing on delete. Internal-notification rendering for both events in `NotificationListener` (bodies truncated at 500 chars since notification text is persisted).
- **Tests** — ~130 new: repository/constraints, validation, the authorization matrix parameterized over all 22 resources (profiles: no-read, read+COMMENT, read-without-COMMENT, owner fallback) plus all resolve/delete permission branches, host-deletion cascade, event publication, audit content.

## Notes for reviewers

- `@ExternalAuthorizationDynamic(DETAIL/COMMENT)` on list/create; `@AnyPrincipalEndpoint` + imperative enforcement on the uuid-keyed methods (host known only after the lookup).
- Known pre-existing issue found on the way: `@LogResource(affiliated = true)` without `resource = true` never records the affiliated resource type — `EventControllerImpl` is affected; fix tracked separately.
- Full suite: 4975 tests green.


src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    @Override
+    @ExternalAuthorizationDynamic(action = ResourceAction.DETAIL)
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    public CommentResponseDto listComments(SecuredResource resource, SecuredUUID objectUuid,
+            PaginationRequestDto pagination) throws NotFoundException {
+        Resource hostResource = validateCommentable(resource);
+        resourceExtensionService(hostResource).getResourceObjectInternal(objectUuid.getValue());
+        RequestValidatorHelper.revalidatePaginationRequestDto(pagination);
```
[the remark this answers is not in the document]
klaraf755: Changed to external.

src/main/java/com/otilm/core/api/web/CommentControllerImpl.java
```diff
@@ -0,0 +1,74 @@
...
+            throws NotFoundException {
+        return commentService
+                .listComments(SecuredResource.fromResource(resource), SecuredUUID.fromUUID(objectUuid), pagination);
+    }
+
+    @Override
+    @AuditLogged(module = Module.CORE, resource = Resource.COMMENT, operation = Operation.CREATE)
+    public CommentDto createComment(@LogResource(resource = true, affiliated = true) Resource resource,
+            @LogResource(uuid = true, affiliated = true) UUID objectUuid, CommentCreateRequestDto request)
+            throws NotFoundException {
```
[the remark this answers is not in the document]
klaraf755: `@Valid` annotation is present on controller in the interfaces, not needed here.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
```diff
@@ -0,0 +1,310 @@
...
+        Resource hostResource = resource.getResource();
+        if (!hostResource.commentable()) {
+            throw new ValidationException("Resource %s does not support comments".formatted(hostResource.getLabel()));
+        }
+        return hostResource;
+    }
+
+    private ResourceExtensionService resourceExtensionService(Resource resource) {
+        return resourceExtensionServices.get(resource.getCode());
+    }
```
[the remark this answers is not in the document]
klaraf755: Switched to `ResourceService` instead of `ResourceExtensionService`, so this comment is no longer relevant.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
```diff
@@ -0,0 +1,306 @@
...
+
+    @Override
+    @AnyPrincipalEndpoint
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    public void deleteComment(UUID uuid) throws NotFoundException {
+        Comment comment = getComment(uuid);
+        ResourceObjectDto hostObject = readGate(comment);
+        NameAndUuidDto actor = AuthHelper.getUserIdentification();
+        boolean isAuthor = actor.getUuid().equals(comment.getAuthorUuid().toString());
+        boolean rootWithReplies = comment.getParentUuid() == null && commentRepository.existsByParentUuid(uuid);
```
ivosh: **! Major** | Correctness

CommentServiceImpl.java:175 checks for replies outside the transaction that deletes at line 187. If a reply commits between those operations, the root author follows the no-replies branch and `ON DELETE CASCADE` erases the newly committed reply, violating the rule at lines 177-181 that only an owner or update holder may erase a replied-to root.

_Suggested:_ Move the reply-existence check and delete into one CommentWriter transaction that locks the root row before checking and atomically applies the evaluated capabilities.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
```diff
@@ -0,0 +1,306 @@
...
+        NameAndUuidDto actor = AuthHelper.getUserIdentification();
+
+        Comment comment = new Comment();
+        comment.setResource(hostResource);
+        comment.setObjectUuid(objectUuid.getValue());
+        comment.setParentUuid(request.getParentUuid());
+        comment.setAuthorUuid(UUID.fromString(actor.getUuid()));
+        comment.setAuthorUsername(actor.getName());
+        comment.setBody(request.getBody());
+        Comment saved = commentWriter.create(comment);
```
ivosh, line 180: **! Major** | Correctness

CommentServiceImpl.java:131 validates the polymorphic host before the independent comment transaction at CommentServiceImpl.java:144. Because the comment table has no foreign key to host tables, a host deletion can remove existing comments and the host between those operations, after which this insert commits an orphan; the sequential host-deletion test does not exercise this interleaving.

_Suggested:_ Serialize comment creation and every host's local deletion phase with a transaction-scoped lock keyed by resource and object UUID, then revalidate the host under that lock before inserting.
Alternatively consider adding a FK, taking into account the target performance.

<sub>flagged by Correctness & races + Maintainability & tests + Transaction & concurrency</sub>

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
```diff
@@ -0,0 +1,306 @@
...
+
+        CommentEventData eventData = baseEventData(saved, hostObject.getName());
+        recordAuditData(eventData);
+        publishAfterCommit(new EventMessage(ResourceEvent.COMMENT_CREATED, Resource.COMMENT, saved.getUuid(), null,
```
ivosh, line 184: **! Major** | Correctness

CommentServiceImpl.java:148 and :221 enqueue COMMENT_CREATED and COMMENT_RESOLVED, but no IEventHandler bean is registered under either event code. EventListener.java:37 therefore resolves a null handler and fails at line 39, so triggers and the new NotificationListener rendering are never reached. CommentEventITest only verifies calls to the mocked producer, not broker consumption. 

_Suggested:_ Confirm this is intended (future handlers plug in). Otherwise register handlers for both comment event codes and add an integration test that drives each message through EventListener into NotificationListener.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
ivosh: **! Major** | Correctness

CommentServiceImpl.java:175 checks for replies before the separate writer transaction deletes at line 187. A root author can observe no replies, a concurrent user can then insert one, and the migration's `ON DELETE CASCADE` will erase that reply, violating this PR's stated invariant that authors cannot erase other users' words.

_Suggested:_ Lock the root and recheck replies in the short delete transaction, and lock the parent while inserting a reply so creation and deletion serialize.

ivosh: **! Major** | Correctness

CommentServiceImpl.java:171-187 checks for replies before `CommentWriter.delete` opens its transaction. A reply can commit after `existsByParentUuid` returns false but before the root deletion, causing the cascade to erase another user's reply despite the deletion rule. The same split at CommentServiceImpl.java:195-222 lets resolve/unresolve update zero rows after a concurrent deletion and still publish a success event.

_Suggested:_ Add a pessimistic `findAndLockByUuid` path and transactional writer operations that lock and reassert comment state before delete or resolution, with reply creation acquiring the same parent lock.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
```diff
@@ -0,0 +1,306 @@
...
+        NameAndUuidDto actor = AuthHelper.getUserIdentification();
+        if (!actor.getUuid().equals(comment.getAuthorUuid().toString())) {
+            authorizationEnforcer
+                    .enforce(comment.getResource(), ResourceAction.COMMENT,
+                            SecuredUUID.fromUUID(comment.getObjectUuid()));
+        }
+
+        OffsetDateTime changedAt = OffsetDateTime.now();
+        if (resolved) {
+            commentWriter.resolve(uuid, changedAt, UUID.fromString(actor.getUuid()), actor.getName());
```
ivosh: **~ Minor** | Correctness

CommentServiceImpl.java:210 and :212 ignore the affected-row count returned by CommentWriter. If deletion commits after `getComment` and `readGate`, the update returns zero, but line 221 still publishes `COMMENT_RESOLVED` and the audit path reports a mutation of a nonexistent thread.

_Suggested:_ Require each resolution update to affect exactly one row and return a not-found or conflict error before auditing or publishing when it affects zero.

src/main/java/com/otilm/core/messaging/jms/listeners/NotificationListener.java
```diff
@@ -888,6 +889,24 @@ yield new InternalNotificationEventData("Request %s for %s from %s is %s"
                 yield new InternalNotificationEventData("%s scheduled task has finished for %s with result %s"
                         .formatted(data.getJobType(), data.getJobName(), data.getStatus()), null);
             }
+            case COMMENT_CREATED -> {
```
ivosh, line 897: **~ Minor** | Testing

NotificationListener.java:892 adds root/reply and resolve/reopen rendering, while NotificationListener.java:926 adds a persisted-text truncation boundary, but no comment test exercises this listener. The neighboring NotificationListenerTest.java:67 directly verifies certificate rendering, so these branches lack the established regression coverage.

_Suggested:_ Add focused NotificationListenerTest cases for root, reply, resolve, reopen, and the 500/501-character truncation boundary.

src/main/java/com/otilm/core/mapper/comment/CommentMapper.java
```diff
@@ -0,0 +1,46 @@
...
+        CommentDto dto = new CommentDto();
+        dto.setUuid(comment.getUuid());
+        dto.setResource(comment.getResource());
+        dto.setObjectUuid(comment.getObjectUuid());
+        dto.setAuthor(new NameAndUuidDto(comment.getAuthorUuid().toString(), comment.getAuthorUsername()));
+        dto.setCreatedAt(comment.getCreatedAt());
+        dto.setBody(comment.getBody());
+        dto.setParentUuid(comment.getParentUuid());
+        if (comment.getParentUuid() == null) {
+            dto.setResolved(comment.getResolvedAt() != null);
```
ivosh: **. Nitpick** | Maintainability

CommentMapper.java:25 is public, but its signature does not reveal that resolution and reply fields are populated only for roots, that null replies become an empty list only for roots, or that reply DTOs omit those fields. Callers must reverse-engineer this non-obvious contract from the implementation.

_Suggested:_ Add method-level Javadoc documenting the root-only fields and null-replies convention.

src/main/java/com/otilm/core/dao/repository/CommentRepository.java
```diff
@@ -0,0 +1,46 @@
...
+public interface CommentRepository extends SecurityFilterRepository<Comment, UUID> {
+
+    Page<Comment> findByResourceAndObjectUuidAndParentUuidIsNullOrderByCreatedAtAsc(Resource resource, UUID objectUuid,
+            Pageable pageable);
+
+    List<Comment> findByParentUuidInOrderByCreatedAtAsc(Collection<UUID> parentUuids);
+
+    boolean existsByParentUuid(UUID parentUuid);
+
+    boolean existsByResourceAndObjectUuid(Resource resource, UUID objectUuid);
```
ivosh: **. Nitpick** | Maintainability

CommentRepository.java:26 adds existsByResourceAndObjectUuid, but the sole local caller is CommentHostDeletionITest.java:85; no production path uses it. Org-wide gh search was unavailable due API rate limiting, so external callers could not be checked, although this core repository is an internal surface.

_Suggested:_ Remove the production repository method and perform the integration-test assertion through a test-only database query.

src/main/java/com/otilm/core/dao/repository/CommentRepository.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.data.repository.query.Param;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface CommentRepository extends SecurityFilterRepository<Comment, UUID> {
+
+    Page<Comment> findByResourceAndObjectUuidAndParentUuidIsNullOrderByCreatedAtAsc(Resource resource, UUID objectUuid,
+            Pageable pageable);
+
+    List<Comment> findByParentUuidInOrderByCreatedAtAsc(Collection<UUID> parentUuids);
```
ivosh: **~ Minor** | Performance

CommentRepository.java:22 returns an unbounded `List` of replies, and CommentServiceImpl.java:113-116 loads all of them for every root on the requested page. A single long-lived thread can therefore produce an arbitrarily large query, heap allocation, and REST response despite root pagination.

_Suggested:_ Paginate replies or enforce an explicit per-thread limit while exposing the total reply count.

src/main/java/com/otilm/core/api/web/CommentControllerImpl.java
```diff
@@ -0,0 +1,74 @@
...
+            @LogResource(uuid = true, affiliated = true) UUID objectUuid, PaginationRequestDto pagination)
+            throws NotFoundException {
+        return commentService
+                .listComments(SecuredResource.fromResource(resource), SecuredUUID.fromUUID(objectUuid), pagination);
+    }
+
+    @Override
+    @AuditLogged(module = Module.CORE, resource = Resource.COMMENT, operation = Operation.CREATE)
+    public CommentDto createComment(@LogResource(resource = true, affiliated = true) Resource resource,
+            @LogResource(uuid = true, affiliated = true) UUID objectUuid, CommentCreateRequestDto request)
```
ivosh, line 58: **~ Minor** | Documentation

src/main/java/com/otilm/core/api/web/CommentControllerImpl.java:51 exposes the validated create request, while CommentServiceImpl.java:103-105, 132-134, and 197-199 adds validation failures to list, create, resolve, and unresolve. ExceptionHandlingAdvice.java:170-181 and 231-237 returns these failures as HTTP 422, but the dependent CommentController contract documents only success and 404 responses, hiding a deliberate wire response from generated clients.

_Suggested:_ Add a 422 ApiResponse with the List<String> error schema to listComments, createComment, resolveComment, and unresolveComment in the dependent CommentController contract.

src/main/java/com/otilm/core/api/web/CommentControllerImpl.java
```diff
@@ -0,0 +1,74 @@
+package com.otilm.core.api.web;
+
+import com.otilm.api.exception.NotFoundException;
+import com.otilm.api.interfaces.core.web.CommentController;
```
ivosh: **~ Minor** | Documentation

CommentControllerImpl.java:4 imports the contract introduced by OmniTrustILM/interfaces#882, but this PR identifies interfaces#848 as the prerequisite; #848 is the issue closed by that interfaces PR. Following the stated merge order could therefore leave this PR uncompilable.

_Suggested:_ Replace the interfaces#848 references in this PR description with OmniTrustILM/interfaces#882.

ivosh (changes requested):

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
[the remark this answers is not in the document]
klaraf755: Fixed in 964330371 — host lookups now go through ResourceInternalService.getResourceObject, the authorized per-resource chain (including owner/group fallback).

src/main/java/com/otilm/core/api/web/CommentControllerImpl.java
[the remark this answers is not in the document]
klaraf755: No change needed: the @Valid (and @RequestBody) annotations live on the CommentController interface method, and Spring MVC resolves handler parameter annotations from implemented interfaces — the convention 51 of 53 controller impls in this codebase follow; the TSP pair is the outlier. Bean validation is exercised by test (blank and oversized bodies rejected, exact-cap body accepted unchanged).

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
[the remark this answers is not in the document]
klaraf755: Fixed in 964330371 — the lookup now goes through ResourceInternalService.getResourceObject, which fails fast with NotSupportedException naming the resource when no extension service is registered; CommentableResourceExtensionITest additionally pins that every commentable resource has one.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
klaraf755: Mitigated in 9133b3adc: CommentWriter.create re-verifies the host inside the insert's own transaction (plain DB read, no OPA under the transaction), so a host deletion that committed meanwhile rolls the comment back instead of committing an orphan. A residual window is inherent to the polymorphic key: no FK can exist across 22 host tables, so full closure would require either share-locking the host row (several host delete flows make connector HTTP calls mid-transaction, so comment inserts would block behind remote round-trips) or a per-object lock-row registry maintained by all 22 create and delete paths. Approvals and the owner/group/trigger associations share the identical window today, mostly without even the pre-check this path does — if we close this class of race it should be a platform decision covering all five tables. The worst-case outcome here is one API-unreachable row (the host 404s every path) in a table wiped by key on every host delete.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
klaraf755: Intentional layering: the handlers implement #2026 on the stacked comment-notifications branch (complete — both beans registered by event code, group-scoped trigger resolution, internal follow-ups; full-suite verify green). That PR opens on this one's merge and the two ship in the same release, exactly because of the unguarded handler lookup this comment describes.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
klaraf755: Fixed in 9133b3adc — root deletion moved into CommentWriter.deleteRoot(uuid, mayCascade), which locks the root row inside the delete transaction (FOR UPDATE conflicts with the key-share lock a reply insert takes on its parent, so the re-check cannot be invalidated before the delete commits) and rejects a non-cascading delete when replies raced in. Authorization still runs strictly before the transaction. Resolution updates and reply deletions now require an affected row (see the sibling thread).

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
klaraf755: Fixed in 9133b3adc — resolve and unresolve require exactly one affected row; zero rows raises NotFoundException before anything is audited or published. Reply deletion checks its count the same way.

src/main/java/com/otilm/core/messaging/jms/listeners/NotificationListener.java
klaraf755: Fixed in 9133b3adc — six focused NotificationListenerTest cases: root, reply, resolved and reopened rendering texts, plus the 500/501 truncation boundary (500 survives unchanged, 501 abbreviates to 500 + ellipsis).

src/main/java/com/otilm/core/dao/repository/CommentRepository.java
klaraf755: Fixed in b526d923e — the listing no longer embeds replies: roots carry a replyCount (single GROUP BY per page) and replies page through GET /v1/comments/{uuid}/replies with the same pagination wrapper. Contract change in interfaces c7ba81c6 (PR #882).

src/main/java/com/otilm/core/api/web/CommentControllerImpl.java
klaraf755: Fixed in interfaces c7ba81c6 (PR #882) — 422 with the platform's string-array body is documented on list, create, resolve, unresolve and the new replies endpoint, matching the ValidationException mapping; issue #848's 400 wording was corrected as well.

ivosh (dismissed):

src/main/java/com/otilm/core/service/writer/CommentWriter.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * The host lookup shares the insert's transaction: no foreign key can hold on the polymorphic key, so a host
+     * deletion committed between the orchestrator's read gate and this transaction rolls the comment back here instead
+     * of committing an orphan.
+     */
+    @Transactional
+    public Comment create(Comment comment) throws NotFoundException {
+        Comment saved = commentRepository.save(comment);
+        resourceService.getResourceObjectInternal(comment.getResource(), comment.getObjectUuid());
```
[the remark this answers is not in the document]
klaraf755: Confirmed — the plain SELECT only closed the wide gap between the read gate and the writer transaction, not the concurrent-transaction window. Fixed in f452143dd with a transaction-scoped advisory lock via the existing `ClusterOperationSynchronizer`, keyed `comment-host:<resource>:<objectUuid>` and acquired in both `CommentWriter.create` and `CommentWriter.deleteAllForObject`; every host-delete path already funnels through the latter inside the host's own transaction. Both interleavings are covered by `CommentWriterConcurrencyITest`, which observes the blocked waiter in `pg_locks` before letting the first transaction commit.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
```diff
@@ -0,0 +1,346 @@
...
+        comment.setObjectUuid(objectUuid.getValue());
+        comment.setParentUuid(request.getParentUuid());
+        comment.setAuthorUuid(UUID.fromString(actor.getUuid()));
+        comment.setAuthorUsername(actor.getName());
+        comment.setBody(request.getBody());
+        Comment saved = commentWriter.create(comment);
+
+        CommentEventData eventData = baseEventData(saved, hostObject.getName());
+        recordAuditData(eventData);
+        publishAfterCommit(new EventMessage(ResourceEvent.COMMENT_CREATED, Resource.COMMENT, saved.getUuid(), null,
```
lubomirw, line 184: **[High]** | Correctness

Is there a follow-up Core PR that registers handlers for `COMMENT_CREATED` and `COMMENT_RESOLVED`? In this PR, `EventListener` looks handlers up by event code, but neither new code has an `IEventHandler` bean, so these events reach a null handler and triggers/notifications do not run.

_Suggested:_ If a follow-up exists, please link it and confirm deployment ordering; otherwise register both handlers here and add an end-to-end listener test before merging.

<sub>flagged by Guided PR Review</sub>

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
```diff
@@ -0,0 +1,346 @@
...
+    private void publishAfterCommit(EventMessage eventMessage) {
+        if (TransactionSynchronizationManager.isSynchronizationActive()) {
+            TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
+                @Override
+                public void afterCommit() {
+                    eventProducer.produceMessage(eventMessage);
+                }
+            });
+        } else {
+            eventProducer.produceMessage(eventMessage);
```
lubomirw, line 376: **[Medium]** | Error handling

The writer transaction has already committed when synchronous event publication runs here. If broker retries are exhausted, the request reports failure for a durable mutation, and a client retry can create a duplicate comment.

_Suggested:_ Persist an outbox record in the writer transaction and dispatch it asynchronously, so broker availability cannot turn a committed write into an apparent request failure.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
```diff
@@ -0,0 +1,346 @@
...
+                .toList();
+        return CommentMapper.toResponseDto(roots, threads);
+    }
+
+    @Override
+    @AnyPrincipalEndpoint
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    public CommentResponseDto listReplies(UUID uuid, PaginationRequestDto pagination) throws NotFoundException {
+        Comment root = getComment(uuid);
+        if (root.getParentUuid() != null) {
```
lubomirw, line 145: **[Low]** | Security

`listReplies` checks whether the comment is a root before enforcing host read access; `changeResolution` does the same around line 236. An unauthorized caller with a comment UUID can therefore distinguish roots from replies through 422 versus 403 responses.

_Suggested:_ Call `readGate` immediately after loading the comment in both methods, before validating root/reply shape.

<sub>flagged by Guided PR Review and Superpowers</sub>

src/main/java/com/otilm/core/api/web/CommentControllerImpl.java
```diff
@@ -0,0 +1,81 @@
...
+    @AuditLogged(module = Module.CORE, resource = Resource.COMMENT, operation = Operation.CREATE)
+    public CommentDto createComment(@LogResource(resource = true, affiliated = true) Resource resource,
+            @LogResource(uuid = true, affiliated = true) UUID objectUuid, CommentCreateRequestDto request)
+            throws NotFoundException {
+        return commentService
+                .createComment(SecuredResource.fromResource(resource), SecuredUUID.fromUUID(objectUuid), request);
+    }
+
+    @Override
+    @AuditLogged(module = Module.CORE, resource = Resource.COMMENT, operation = Operation.RESOLVE)
```
lubomirw, line 65: **[Low]** | Correctness

Resolve, unresolve, and delete carry the host only inside operation data; their structured `affiliatedResource` fields remain empty. Host-based audit filtering therefore finds create/list records but misses these mutations.

_Suggested:_ Add a request-scoped affiliation override populated from the loaded comment, and assert host type and UUID for every UUID-keyed mutation.

<sub>flagged by Guided PR Review</sub>

src/test/java/com/otilm/core/integration/service/CommentAuthorizationITest.java
```diff
@@ -0,0 +1,253 @@
...
+import org.springframework.context.annotation.Import;
+import org.springframework.security.access.AccessDeniedException;
+import org.springframework.security.core.context.SecurityContextHolder;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * The authorization matrix from the issue, parameterized over every commentable resource: visibility equals read access
```
lubomirw: **[Low]** | Maintainability

This Javadoc refers to "the issue," while the comments around lines 118 and 146 use unexplained `Profile 1` and `Profile 3` matrix labels. Those references depend on planning context that is not available in the test itself.

_Suggested:_ Rewrite all three as standalone behavioral descriptions without issue or profile numbering.

<sub>flagged by Hygiene</sub>

lubomirw (changes requested):
Automated multi-reviewer triage — 5 findings (REQUEST_CHANGES).

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
klaraf755: The handlers deliberately live in the follow-up for #2026 — this PR carries the API and storage, the event-delivery PR opens once this one merges (branch is ready). The deployment-ordering trap you describe was real, though: `EventListener` NPE'd on any event without a registered handler, and the NPE escaped the `EventException` catch into the JMS retry template — a poison message per comment. Fixed in 944cc23d6: a missing handler is now logged and the event skipped, covered by a listener test.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
klaraf755: A transactional outbox is platform infrastructure — every existing event flow has the same commit-then-publish property, so I'd rather track that as a platform issue than introduce it in a comments PR. The user-visible defect is fixed in 7c9a36587: the post-commit publish is now best-effort, so a broker failure logs the lost event instead of failing a request whose write already committed and inviting a duplicating retry. The comment row is the source of truth; the event only drives notifications, which are best-effort throughout this feature. Covered by a test stubbing a broker failure.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
klaraf755: Fixed in 05f61b07d — the read gate now answers before shape validation in both `listReplies` and `changeResolution` (`deleteComment` already gated first). A new test pins that an unauthorized caller gets `AccessDeniedException` for roots and replies alike.

src/main/java/com/otilm/core/api/web/CommentControllerImpl.java
klaraf755: Fixed in 05f61b07d with the suggested request-scoped override: `AuditAffiliationOverride` mirrors `AuditOperationDataOverride`, the service records the host from the loaded comment right after the read gate, and the aspect fills the structured affiliated fields whenever no parameter provided them (parameter-declared affiliation still wins). The audit test now asserts host type and UUID on resolve, unresolve and delete. `listReplies` got the same treatment — it is UUID-keyed too, so its LIST records had the same gap.

src/test/java/com/otilm/core/integration/service/CommentAuthorizationITest.java
klaraf755: Fixed in 05f61b07d — the class Javadoc and both profile-numbered comments are rewritten as standalone behavioral descriptions with no planning-context references.

lubomirw (dismissed):

lubomirw (approved):

Issue #2101, opened 2026-08-18
Action message payload written to ERROR log on handler failure


lubomirw:
### Description

When an action handler fails, `AbstractJmsEndpointConfig` logs the caught exception object at ERROR level. That exception is a `MessageHandlingException` (`@Data`/`@ToString`, carrying an `Object messageData` field) wrapping the full `ActionMessage` (`@Data`, with a `private Object data` field), so the entire message payload — including its `data` map, which can hold a registration credential — is written to the log on every failed action.

This contradicts the intent stated a few lines earlier in the same class: "Log only metadata, never the body: message payloads can carry secrets."

The payload must be kept out of the logged representation. The log call site should emit only routing metadata, and `MessageHandlingException.toString()` should exclude the payload field so no future call site can leak it either.

### Steps to Reproduce

1. Create a secret so a CREATE action is queued with encrypted content in its `data` map.
2. Make the handler fail — for example point the vault profile at an unreachable connector.
3. Read the ERROR entry that `AbstractJmsEndpointConfig` logs for the `actionsListener` endpoint.

### Expected Behavior

The failure logs only routing metadata: queue name, resource, action and resource UUID. No message payload reaches the log at any level.

### Actual Behavior

The full `ActionMessage` is rendered into the ERROR line, including its `data` map, so secret material can be written to the log in a form an operator or log shipper can read.

### Severity

Major

### Module

Core

### Environment

Core with RabbitMQ (Qpid JMS over AMQP), verified against `core` `origin/main`. `MessageHandlingException` lives in the `interfaces` repo (`api/exception/MessageHandlingException.java`), so a complete fix spans both repos. Split out of #2098.

Issue #2100, opened 2026-08-18
Action messages lost on dispatch failure: no redelivery, DLQ, or durable post-commit send


lubomirw:
### Description

The action queue has no recovery path for a message that cannot be delivered or handled, so work Core has already committed can be lost silently. Two paths stack.

**Consumer side.** `AbstractJmsEndpointConfig` catches generic `Exception`, logs it, and does not rethrow, so `RetryTemplate` records the attempt as successful and the message is ACKed and dropped — no redelivery, no dead-letter. `RetryConfig` marks only `JmsException`, `JMSException` and `IOException` retryable, so a `MessageHandlingException` would be discarded even if it did propagate.

**Producer side.** Since #2099 the publish runs in an `afterCommit` synchronization. If the broker is unreachable and `producerRetryTemplate` exhausts its retries, the database work is already committed but the action message is never queued — for example a secret committed `INACTIVE` that a retry cannot recreate because the name already exists. This is the same trade-off already accepted in `ApprovalServiceImpl` and documented in `NotificationProducer`.

Applies to every resource that publishes through `ActionProducer`, not only secrets.

A transactional outbox (persist the action in the creating transaction, dispatch from a poller, mark sent on success) closes both paths. A lighter step is to stop swallowing unhandleable messages so they redeliver or dead-letter, and to mark the resource `FAILED` rather than leaving it silently in its pre-action state when post-commit dispatch fails.

### Steps to Reproduce

1. Configure a vault profile with no approval profile attached and create a secret, so a CREATE action is queued.
2. Make the handler fail — for example point the vault instance at an unreachable connector.
3. Watch the `actionsListener` ERROR entry, then check the broker for redelivery or a dead-letter message.
4. For the producer path: stop the broker, then create a secret and observe that the row commits while no message is ever queued.

### Expected Behavior

An action message that cannot be handled is redelivered according to the retry policy and then dead-lettered, so it stays recoverable and visible to an operator. A dispatch failure after commit either persists the intent for a later attempt, or leaves the resource in a state that shows the action did not run.

### Actual Behavior

The failure is logged once and the message is dropped: no redelivery, no dead-letter queue, and the caller has already received a success response. The resource keeps its pre-action state — a secret stays `INACTIVE` — with no automatic recovery path.

### Severity

Major

### Module

Core

### Environment

[...]

PR #2106, opened 2026-08-19
Fix newly created secrets being left inactive and unusable

### LINKED ISSUE #2098: Secret create message published before commit leaves secret INACTIVE

lubomirw:
Backport of #2099 onto the `2.19.0` tag for the **2.19.1 hotfix**, for a customer hit by #2098 (secrets stuck `INACTIVE`).

## What

`SecretServiceImpl.produceActionMessage` published the JMS action message inside the class-level `@Transactional`, before commit, so `actionsListener` (separate thread, own transaction) could not load the secret and failed — leaving it `INACTIVE`. The fix defers the publish to a `TransactionSynchronizationManager.registerSynchronization(...afterCommit())` callback, mirroring `ApprovalServiceImpl`. Covers all four Secret action sites via the shared helper.

## Backport notes

- **Re-applied by hand**, not cherry-picked: the `2.19.0` base predates the mainline Spotless/Checkstyle reformatting (48 commits), so the merged commit `488ea840f` conflicts on the reformatted `SecretServiceITest` import block.
- The production change is byte-for-byte equivalent to what merged on `main`.
- The test broker shim is adapted to the `2.19.0` structure and now runs the listener in a `REQUIRES_NEW` transaction (the old inline shim only passed by riding the bug). Added the focused regression test `createSecretPublishesActionMessageOnlyAfterCommit`.

## Merge order

Pure code fix — **no version bump here**. Depends on #2108 (httpcore5 CVE pins + `2.19.1-SNAPSHOT` dev version), which must merge first: this branch's image build fails Trivy until the pins land. After #2108 merges, this branch rebases onto `hotfix/2.19.1` (picking up the pins) and goes green. The `2.19.1-SNAPSHOT → 2.19.1` release commit happens at tag time.

## Verification

`SecretServiceITest` — 22/22 green (`REQUIRES_NEW` shim + new regression test).

## Scope

Defect 1 (the pre-commit race) only, matching #2099. Durability (#2100) and log-leak (#2101) are not backported — not customer-blocking for 2.19.1.

Refs #2098, #2099

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/SecretServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/SecretServiceImpl.java
@@ -62,6 +62,8 @@ import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Pageable;
 import org.springframework.security.concurrent.DelegatingSecurityContextExecutorService;
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
 
 import java.security.NoSuchAlgorithmException;
 import java.util.*;
@@ -370,7 +372,14 @@ public class SecretServiceImpl implements SecretExternalService, SecretInternalS
         actionMessage.setResource(Resource.SECRET);
         actionMessage.setResourceAction(resourceAction);
         actionMessage.setResourceUuid(secret.getUuid());
-        actionProducer.produceMessage(actionMessage);
+        // actionsListener consumes this on another thread in its own transaction; publishing before the creating
+        // transaction commits leaves that transaction unable to load the secret, so defer the send until after commit.
+        TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
+            @Override
+            public void afterCommit() {
+                actionProducer.produceMessage(actionMessage);
+            }
+        });
     }
 
     @ExternalAuthorization(resource = Resource.SECRET, action = ResourceAction.UPDATE)
--- a/src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
+++ b/src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
@@ -65,6 +65,9 @@ import org.springframework.test.context.DynamicPropertySource;
 import org.springframework.test.context.bean.override.mockito.MockitoBean;
 import org.springframework.test.context.event.ApplicationEvents;
 import org.springframework.test.context.event.RecordApplicationEvents;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.TransactionDefinition;
+import org.springframework.transaction.support.TransactionTemplate;
 import com.otilm.core.events.SecretContentUpdatedEvent;
 
 import java.io.Serializable;
@@ -73,6 +76,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
@@ -115,6 +119,8 @@ class SecretServiceITest extends BaseSpringBootTest {
     private ConnectorRepository connectorRepository;
     @Autowired
     private ActionsListener actionListener;
+    @Autowired
+    private PlatformTransactionManager transactionManager;
     @MockitoBean
     private ActionProducer actionProducer;
     @MockitoBean
@@ -129,12 +135,21 @@ class SecretServiceITest extends BaseSpringBootTest {
     @BeforeEach
     void setUp() throws AlreadyExistException, AttributeException, NoSuchAlgorithmException, JsonProcessingException {
 
-        // Process message instead of sending it to the queue and set approval status to approved to bypass approval in tests
+        // REQUIRES_NEW so the handler runs in its own transaction like the real broker: at afterCommit the producing
+        // transaction has already committed, so an inline handler would join it and its writes would be discarded.
+        final TransactionTemplate listenerTransaction = new TransactionTemplate(transactionManager);
+        listenerTransaction.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
         Mockito.doAnswer(invocation -> {
                     ActionMessage msg = invocation.getArgument(0);
                     msg.setApprovalStatus(ApprovalStatusEnum.APPROVED);
                     msg.setApprovalUuid(UUID.randomUUID());
-                    actionListener.processMessage(msg);
+                    listenerTransaction.executeWithoutResult(status -> {
+                        try {
+                            actionListener.processMessage(msg);
+                        } catch (MessageHandlingException e) {
+                            throw new IllegalStateException(e);
+                        }
+                    });
                     return null; // because produceMessage returns void
                 }
         ).when(actionProducer).produceMessage(any());
@@ -772,6 +787,31 @@ class SecretServiceITest extends BaseSpringBootTest {
         } finally {
             authServiceMock.stop();
         }
+    }
+
+    @Test
+    void createSecretPublishesActionMessageOnlyAfterCommit() throws NotFoundException, AttributeException, AlreadyExistException, ConnectorException {
+        // Read from a separate REQUIRES_NEW transaction: a publish before commit leaves the secret invisible there,
+        // exactly as it is to the real listener's own transaction.
+        final AtomicBoolean secretVisibleFromSeparateTransaction = new AtomicBoolean();
+        final TransactionTemplate separateTransaction = new TransactionTemplate(transactionManager);
+        separateTransaction.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
+
+        Mockito.doAnswer(invocation -> {
+            ActionMessage msg = invocation.getArgument(0);
+            secretVisibleFromSeparateTransaction.set(separateTransaction.execute(status -> secretRepository.findByUuid(msg.getResourceUuid()).isPresent()));
+            return null;
+        }).when(actionProducer).produceMessage(any());
+
+        SecretRequestDto request = new SecretRequestDto();
+        request.setName("afterCommitSecret");
+        request.setSecret(new BasicAuthSecretContent());
+
+        secretService.createSecret(request, vaultProfile.getSecuredParentUuid(), vaultInstance.getSecuredUuid());
+
+        assertThat(secretVisibleFromSeparateTransaction.get())
+                .as("action message must be published only after the creating transaction commits")
+                .isTrue();
     }
 
 }
--- a/src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
+++ b/src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
@@ -65,10 +65,8 @@ import org.springframework.test.context.DynamicPropertySource;
 import org.springframework.test.context.bean.override.mockito.MockitoBean;
 import org.springframework.test.context.event.ApplicationEvents;
 import org.springframework.test.context.event.RecordApplicationEvents;
-import org.springframework.transaction.PlatformTransactionManager;
-import org.springframework.transaction.TransactionDefinition;
-import org.springframework.transaction.support.TransactionTemplate;
 import com.otilm.core.events.SecretContentUpdatedEvent;
+import com.otilm.core.events.transaction.TransactionHandler;
 
 import java.io.Serializable;
 import java.security.NoSuchAlgorithmException;
@@ -120,7 +118,7 @@ class SecretServiceITest extends BaseSpringBootTest {
     @Autowired
     private ActionsListener actionListener;
     @Autowired
-    private PlatformTransactionManager transactionManager;
+    private TransactionHandler transactionHandler;
     @MockitoBean
     private ActionProducer actionProducer;
     @MockitoBean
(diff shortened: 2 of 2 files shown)


klaraf755 (commented):
**Verdict: approve — the fix is sound, nothing blocking.**

The production change matches mainline and the `ApprovalServiceImpl` pattern, all four `produceActionMessage` call sites run under the class-level `@Transactional`, and since `BaseSpringBootTest` isn't `@Transactional`, the new regression test crosses a real commit boundary. Inline notes below are non-blocking; two things can't be anchored inline because the files aren't in the diff:

1. **The fix un-masks a CREATE/DELETE race on the TSP credential path** (`TspProfileBasicCredentialServiceImpl.create`). `create()` is `NOT_SUPPORTED`, so `createSecret` commits and publishes CREATE; a duplicate username then makes `deleteVaultSecretQuietly` publish DELETE moments later. With `messaging.concurrency.actions: 10` and no per-resource ordering, the DELETE can be consumed while the CREATE's vault write is still in flight — the row is gone, the vault delete no-ops as idempotent, then the CREATE lands the credential material in the vault with no core record left to clean it up. Pre-fix this was unreachable because the CREATE action always failed to load the secret. Same on `main` — I'd open a follow-up issue rather than touch this backport.

2. **`revokeCertificate` (`ClientOperationServiceImpl:2071`) still publishes in-transaction** — it has no propagation override, so the class-level `@Transactional` applies; the other four `actionProducer` sites in that class are `NOT_SUPPORTED`. Same defect class as #2098, just for certificates. Also follow-up-issue material, mainly so the next audit doesn't have to re-derive all five sites.

src/main/java/com/otilm/core/service/impl/SecretServiceImpl.java
```diff
@@ -370,7 +372,14 @@ private void produceActionMessage(SecretActionData secretActionData, VaultProfil
         actionMessage.setResource(Resource.SECRET);
         actionMessage.setResourceAction(resourceAction);
         actionMessage.setResourceUuid(secret.getUuid());
-        actionProducer.produceMessage(actionMessage);
+        // actionsListener consumes this on another thread in its own transaction; publishing before the creating
+        // transaction commits leaves that transaction unable to load the secret, so defer the send until after commit.
+        TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
```
klaraf755, line 377: Non-blocking, and matching mainline byte-for-byte is the right call for the hotfix — flagging for the mainline follow-up: this registration has no `isSynchronizationActive()` guard, unlike `SettingServiceImpl.cacheAfterCommit`, `CustomOidEntryServiceImpl.publishAfterCommit`, and `CacheEvictor`, which all fall back to an inline send. It's safe today only because of the class-level `@Transactional`; the first self-invocation or `NOT_SUPPORTED` override (a live idiom on `TspProfileBasicCredentialServiceImpl` and `ClientOperationServiceImpl`) turns this into a bare `IllegalStateException` after all the encrypt/DB work.

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
```diff
@@ -129,12 +135,21 @@ static void authServiceProperties(DynamicPropertyRegistry registry) {
...
+        // REQUIRES_NEW so the handler runs in its own transaction like the real broker: at afterCommit the producing
+        // transaction has already committed, so an inline handler would join it and its writes would be discarded.
+        final TransactionTemplate listenerTransaction = new TransactionTemplate(transactionManager);
+        listenerTransaction.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
         Mockito.doAnswer(invocation -> {
                     ActionMessage msg = invocation.getArgument(0);
                     msg.setApprovalStatus(ApprovalStatusEnum.APPROVED);
                     msg.setApprovalUuid(UUID.randomUUID());
-                    actionListener.processMessage(msg);
+                    listenerTransaction.executeWithoutResult(status -> {
```
klaraf755: Worth a sentence in the comment above: this shim turns a consumer-side handler failure into an `IllegalStateException` thrown out of `commit()` — a state production can't produce (real handler failures stay on the listener thread and only NACK), and the inner transaction's rollback discards any `FAILED`-state writes. The suite currently avoids this only because `testSecretOperationFailed` re-stubs `produceMessage` and calls `processSecretAction` directly; the next person writing a failing-action test through `createSecret` will hit it blind.

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
```diff
@@ -129,12 +135,21 @@ static void authServiceProperties(DynamicPropertyRegistry registry) {
     @BeforeEach
     void setUp() throws AlreadyExistException, AttributeException, NoSuchAlgorithmException, JsonProcessingException {
 
-        // Process message instead of sending it to the queue and set approval status to approved to bypass approval in tests
+        // REQUIRES_NEW so the handler runs in its own transaction like the real broker: at afterCommit the producing
+        // transaction has already committed, so an inline handler would join it and its writes would be discarded.
+        final TransactionTemplate listenerTransaction = new TransactionTemplate(transactionManager);
```
klaraf755: `TransactionHandler.runInNewTransaction(Runnable/Supplier)` is an existing `@Component` doing exactly this (already autowired in `DiscoveryWriterITest`). Using it here and in the new test would drop the `PlatformTransactionManager` field, both `TransactionTemplate` constructions, and — via the `Supplier` overload — the `AtomicBoolean` too. As it stands the REQUIRES_NEW setting, the load-bearing detail of the PR, is written twice in this class.

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
```diff
@@ -129,12 +135,21 @@ static void authServiceProperties(DynamicPropertyRegistry registry) {
     @BeforeEach
     void setUp() throws AlreadyExistException, AttributeException, NoSuchAlgorithmException, JsonProcessingException {
 
-        // Process message instead of sending it to the queue and set approval status to approved to bypass approval in tests
+        // REQUIRES_NEW so the handler runs in its own transaction like the real broker: at afterCommit the producing
```
klaraf755: The replaced comment lost its second half: the lines below still fabricate `APPROVED` + a random `approvalUuid`, but the "bypass approval in tests" rationale is gone. Since `ActionsListener.processMessage` branches on `getApprovalUuid() != null`, someone adding approval coverage will either chase a phantom bug or delete those lines and strand every action at `PENDING_APPROVAL`. One restored sentence covers it.

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
```diff
@@ -774,4 +789,29 @@ void testGetSearchableFieldInformation() {
...
+            return null;
+        }).when(actionProducer).produceMessage(any());
+
+        SecretRequestDto request = new SecretRequestDto();
+        request.setName("afterCommitSecret");
+        request.setSecret(new BasicAuthSecretContent());
+
+        secretService.createSecret(request, vaultProfile.getSecuredParentUuid(), vaultInstance.getSecuredUuid());
+
+        assertThat(secretVisibleFromSeparateTransaction.get())
```
klaraf755, line 809: The flag is `false` both when the publish happened pre-commit and when it never happened at all, and the assertion message claims the former in both cases. Adding
```java
Mockito.verify(actionProducer, times(1)).produceMessage(any());
```
alongside the flag assertion separates "not published" from "published too early" and also catches double-registration (duplicate messages).

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
lubomirw: Done in 9059b3509 — both sites now use `transactionHandler.runInNewTransaction`, dropping the `PlatformTransactionManager` field, both `TransactionTemplate` constructions and three imports. One correction: the `AtomicBoolean` has to stay — the value is computed inside the Mockito answer and has to escape it, so the `Supplier` overload replaces `execute(...)` but not the holder.

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
lubomirw: Fixed in 9059b3509. Confirmed the trap: `ActionsListener` creates an Approval and returns when `approvalUuid` is null, so the comment now states the pre-approval is what bypasses it.

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
lubomirw: Added in 9059b3509: `Mockito.verify(actionProducer, times(1)).produceMessage(any())` ahead of the flag assertion.

src/main/java/com/otilm/core/service/impl/SecretServiceImpl.java
lubomirw: Verified and folded into #2100 (see comment there) — all three precedents confirmed. Leaving the backport byte-identical to mainline as you suggest.

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
lubomirw: Verified. The shim comment was reworked in 9059b3509 but kept to the REQUIRES_NEW and approval rationale; the `IllegalStateException`-out-of-commit divergence is real but felt like comment bloat for a hotfix, so it's noted in #2114's context rather than in the file. Happy to add the sentence if you'd rather have it at the call site.

PR #2109, opened 2026-08-19
Issue the timestamps an AdES signature needs in-process

### LINKED ISSUE #1976: Timestamp source bridge
### Description

The D4 bridge. `timestampSource` on the content-signing workflow is `INTERNAL {signingProfileRef}`, pointing at an ILM-managed TIMESTAMPING Signing Profile. An external TSA arrives post-v1 as a TIMESTAMPING Signing Profile with the delegated signing scheme - its connector holds the URL and credentials - so a TSA URL never appears on a signing profile.

`InternalTimestampSource` deliberately does **not** go through the TSP protocol edge - no HTTP, no TSP profile, no protocol authentication. It resolves the referenced TIMESTAMPING profile with the same resolver factory, synthesizes a `TspRequest` directly from the connector's imprint (message imprint plus digest algorithm; policy = the profile's default), and calls the timestamp engine in-process. **Corrected 2026-08-13: the engine is *not* unchanged.** `ManagedTimestampEngine.process(...)` returns the wire-shaped `TspResponse` and even its internal form reports time-quality and certificate-validation failures as `TspResponse.rejected(...)` - there is no token-shaped entry point. This child adds a protocol-neutral sibling entry (timestamp token out, failures in engine vocabulary), with `process` staying the thin TSP wrapper over it. The pipeline is unchanged: time-quality gate, serial generation, token generation and recording all run exactly as they do for TSP.

Three deliberate consequences:

- A degraded time source blocks the TIMESTAMPED and ARCHIVAL steps through the `TimeQualityRegister` gate. That is correct - an AdES timestamp with untrustworthy time is worse than a failed operation - and it surfaces as an explicit step failure naming the step.
- The TIMESTAMPING profile's `enabledProtocols` gate does not apply to in-process invocation: being referenced by a content-signing profile is its own explicit, admin-controlled authorization.
- **Recording (DP-6, decided 2026-08-08):** token issuance is recorded under the TIMESTAMPING profile's own record policy, using a **new `SigningProtocol` value** that distinguishes in-process issuance from TSP (the enum value ships in OmniTrustILM/interfaces#832). The AdES operation's record carries the token serials in its metadata, so an ARCHIVAL signature traces to its two timestamp records. The referenced TIMESTAMPING profile must itself satisfy the record floor (recording on, DEFERRED_DURABLE minimum) - a save-time rule beside the ILM-managed check, enforced by the profile surface in OmniTrustILM/core#1974 (which owns the entity column), because the traceability is only as good as its record policy. The rejected alternative - logging only on the AdES operation - would have weakened TSA-side audit.

### Definition of Done

- [ ] `InternalTimestampSource` takes an imprint (message imprint + digest algorithm) and the TIMESTAMPING profile reference and returns a timestamp token, below the TSP protocol edge - proven by tests driving it directly; the two workflow steps that call it ship with OmniTrustILM/core#1974
- [ ] ManagedTimestampEngine exposes a protocol-neutral entry point (token out, engine-vocabulary failures); process stays the TSP wrapper, its wire behavior pinned by the existing suite
- [ ] An external RFC 3161 source stays additive without a contract change: post-v1 it is a TIMESTAMPING Signing Profile with the delegated signing scheme, reached through the same `timestampSource` reference
- [ ] A degraded time source fails the step explicitly, naming it
- [ ] Issuance recorded under the TIMESTAMPING profile's record policy with the new in-process SigningProtocol value, and the issued token's serial returned to the caller; putting that serial on the AdES record ships with OmniTrustILM/core#1974, which owns the record
- [ ] `SUPPORTED_PROTOCOLS` asserted to contain only gate-eligible protocols, so an in-process value can never be offered as enableable

ivosh:
**Adds a protocol-neutral way for the platform to issue itself an RFC 3161 timestamp, so an AdES signature can get the tokens it needs without going out over the wire.** Closes #1976.

## Why

An AdES signature needs timestamps. The TIMESTAMPED step needs one over the signature; the ARCHIVAL step needs another.

Today the only way to get a token out of the TSA engine is `ManagedTimestampEngine.process(...)`. That returns a wire-shaped `TspResponse`, which is the wrong currency for an in-process caller. A time-quality failure came back as `TspResponse.rejected(...)`, so a calling workflow step had nothing to throw and no token-shaped success to unwrap.

## What this adds

**`ManagedTimestampEngine.issue(...)` is the new protocol-neutral entry point.** It returns an `IssuedTimestamp` — the encoded token, its serial number, its generation time. Failures come back as `SigningEngineException` in the engine's own vocabulary. `process(...)` is now a thin wrapper over it.

**`InternalTimestampSource` is the bridge.** It takes an imprint and the name of a TIMESTAMPING Signing Profile. It resolves that profile through the same resolver factory the protocol path uses, synthesizes a `TspRequest`, and calls `issue(...)` directly. No HTTP, no TSP profile, no protocol authentication.

**`SigningProtocol.INTERNAL_TSA` records the difference.** Issuance is recorded under the TIMESTAMPING profile's own record policy, exactly as it is for TSP. The protocol value on the record says whether a client asked over RFC 3161 or the platform issued the token to itself.

The pipeline is otherwise unchanged: time-quality gate, serial generation, token generation, recording.

## Three deliberate consequences

**A degraded time source stops the step.** The `TimeQualityRegister` gate runs for in-process issuance exactly as it does for TSP. An AdES timestamp carrying untrustworthy time is worse than a failed operation. The failure names the calling step, so an operator sees which step stopped and why.

**`enabledProtocols` is not consulted in-process.** Being referenced by a content-signing profile *is* the authorization, and that reference is admin-controlled.

**`INTERNAL_TSA` can never be offered as enableable.** `SUPPORTED_PROTOCOLS` is the list an operator may switch on per workflow type. Offering a protocol with no client edge would advertise a route no caller can take. `SigningProfileSupportedProtocolsTest` asserts that everything offered is gate-eligible and that `INTERNAL_TSA` is not among it.

## The engine now checks the connector's echo

The connector assembles the `TSTInfo`. The token it hands back is only as trustworthy as its echo of what was asked for.

On the RFC 3161 path the client checks that itself, against its own request. **In-process there is no such counterparty, and the token goes straight into a signature.** So `issue(...)` verifies three things before anything is recorded or returned:

- the stamped message imprint, compared in constant time;
- its digest algorithm;
- the serial number the platform generated.

A mismatch is a `CONNECTOR_FAULT` and nothing is recorded.

`genTime` is deliberately **not** compared. Its encoded precision is the connector's choice, so an exact match would be brittle without adding assurance the imprint and the serial do not already give. The reason is in the method's Javadoc so the next reader does not "fix" the omission.

## Wire behaviour

**The failure info a client sees is unchanged.** The engine used to build `TspResponse.rejected(...)` inline at each failure site; it now throws in its own vocabulary and `TspErrorMapper` maps that back. `TIME_UNAVAILABLE` is a new `SigningEngineFailure` value that exists so the protocol-neutral path does not lose the distinction, and it maps to RFC 3161's `timeNotAvailable` — the same answer the inline code gave.

**One observable change.** `process(...)` no longer declares `throws TspException`; it renders engine failures as rejections itself. A formatting-connector failure therefore comes back as a rejection carrying the wire-safe `clientMessage` ("Internal error during DTBS formatting") instead of a thrown `TspException` whose message carried the connector's own text. The `failureInfo` is the same `SYSTEM_FAILURE` as before, and the connector's detail is still in the operator log. `TspErrorMapper.toTspException` stays for failures `TsaServiceImpl` raises outside the engine.

## No caller yet

Like #2089, this ships ahead of its consumers. The two workflow steps that call `InternalTimestampSource` land with #1974, which also owns putting the token serial on the AdES operation's record. That is the epic's build order, not dead code — please don't file it as unused.

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
@@ -123,7 +123,8 @@ import org.springframework.transaction.annotation.Transactional;
 public class SigningProfileServiceImpl implements SigningProfileExternalService, SigningProfileInternalService {
 
     /**
-     * Defines which signing protocols are allowed for each workflow type.
+     * Defines which signing protocols an operator may enable per workflow type. Only protocols a client can actually
+     * reach belong here; {@code SigningProfileSupportedProtocolsTest} holds that line.
      */
     private static final Map<SigningWorkflowType, Set<SigningProtocol>> SUPPORTED_PROTOCOLS = Map
             .of(SigningWorkflowType.TIMESTAMPING, EnumSet.of(SigningProtocol.TSP));
--- a/src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
@@ -300,7 +300,7 @@ public class SigningProfileServiceImpl implements SigningProfileExternalService,
     /** Cache loader. No authorization annotation by design. */
     @Override
     @Cacheable(value = CacheConfig.SIGNING_PROFILE_CACHE, key = "#name", sync = true)
-    @Transactional(readOnly = true)
+    @Transactional(readOnly = true, noRollbackFor = NotFoundException.class)
     public SigningProfileModel<?, ?> loadSigningProfileModel(String name) throws NotFoundException {
         SigningProfile profile = signingProfileRepository
                 .findByName(name)
--- a/src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
@@ -124,7 +124,7 @@ public class SigningProfileServiceImpl implements SigningProfileExternalService,
 
     /**
      * Defines which signing protocols an operator may enable per workflow type. Only protocols a client can actually
-     * reach belong here; {@code SigningProfileSupportedProtocolsTest} holds that line.
+     * reach belong here.
      */
     private static final Map<SigningWorkflowType, Set<SigningProtocol>> SUPPORTED_PROTOCOLS = Map
             .of(SigningWorkflowType.TIMESTAMPING, EnumSet.of(SigningProtocol.TSP));
@@ -297,17 +297,11 @@ public class SigningProfileServiceImpl implements SigningProfileExternalService,
         return self.loadSigningProfileModel(name);
     }
 
-    /**
-     * Package-private internal cache loader, self-invoked.
-     *
-     * @throws IllegalStateException if the profile has no version row matching its {@code latestVersion}, or the
-     * version declares a managed scheme but its {@code managedSigningType} is {@code null} (DB integrity violations)
-     * @throws IllegalArgumentException if the profile's workflow type has no model mapper, or the version declares a
-     * scheme the mapper does not support
-     */
+    /** Cache loader. No authorization annotation by design. */
+    @Override
     @Cacheable(value = CacheConfig.SIGNING_PROFILE_CACHE, key = "#name", sync = true)
     @Transactional(readOnly = true)
-    SigningProfileModel<?, ?> loadSigningProfileModel(String name) throws NotFoundException {
+    public SigningProfileModel<?, ?> loadSigningProfileModel(String name) throws NotFoundException {
         SigningProfile profile = signingProfileRepository
                 .findByName(name)
                 .orElseThrow(() -> new NotFoundException(SigningProfile.class, name));
--- a/src/main/java/com/otilm/core/signing/tsa/InternalTimestampSource.java
+++ b/src/main/java/com/otilm/core/signing/tsa/InternalTimestampSource.java
@@ -1,0 +1,127 @@
+package com.otilm.core.signing.tsa;
+
+import com.otilm.api.exception.NotFoundException;
+import com.otilm.api.model.common.enums.cryptography.DigestAlgorithm;
+import com.otilm.api.model.core.signing.SigningProtocol;
+import com.otilm.core.model.signing.SigningProfileModel;
+import com.otilm.core.model.signing.resolved.ResolvedManagedTimestampingProfile;
+import com.otilm.core.model.signing.resolved.ResolvedSigningProfile;
+import com.otilm.core.service.SigningProfileInternalService;
+import com.otilm.core.signing.engine.error.SigningEngineException;
+import com.otilm.core.signing.engine.error.SigningEngineFailure;
+import com.otilm.core.signing.engine.resolver.SigningProfileResolverFactory;
+import com.otilm.core.signing.tsa.messages.IssuedTimestamp;
+import com.otilm.core.signing.tsa.messages.TimestampImprint;
+import com.otilm.core.signing.tsa.messages.TspRequest;
+import java.util.List;
+import java.util.Optional;
+import org.springframework.stereotype.Component;
+
+/**
+ * Issues the timestamps an AdES signature needs from an ILM-managed TIMESTAMPING Signing Profile, in-process.
+ *
+ * <p>
+ * The call goes below the TSP protocol edge: no HTTP, no TSP profile, no protocol authentication. Being referenced by a
+ * content-signing profile is the authorization, so the referenced profile's {@code enabledProtocols} are not consulted
+ * — an in-process protocol can never be listed there anyway.
+ * </p>
+ */
+@Component
+public class InternalTimestampSource {
+
+    private static final String CLIENT_MESSAGE = "Internal error while timestamping the signature";
+
+    private final SigningProfileInternalService signingProfileService;
+    private final SigningProfileResolverFactory signingProfileResolverFactory;
+    private final ManagedTimestampEngine engine;
+
+    public InternalTimestampSource(SigningProfileInternalService signingProfileService,
+            SigningProfileResolverFactory signingProfileResolverFactory, ManagedTimestampEngine engine) {
+        this.signingProfileService = signingProfileService;
+        this.signingProfileResolverFactory = signingProfileResolverFactory;
+        this.engine = engine;
+    }
+
+    /**
+     * Issues a timestamp over {@code imprint} from the named TIMESTAMPING Signing Profile.
+     *
+     * @param timestampingProfileName the profile's name, which the caller resolves from the {@code signingProfileUuid}
+     * its workflow configuration stores — the name is this bridge's key because the profile-model cache is name-keyed
+     * @param step the calling workflow step, which names the failure when issuance does not succeed
+     * @throws SigningEngineException if the profile cannot be used or the timestamp cannot be issued
+     */
+    public IssuedTimestamp timestamp(TimestampImprint imprint, String timestampingProfileName, String step)
+            throws SigningEngineException {
+        try {
+            SigningProfileModel<?, ?> profile = loadProfile(timestampingProfileName);
+            ResolvedManagedTimestampingProfile resolved = resolveTimestampingProfile(profile);
+            requireUsableImprint(imprint, resolved);
+            return engine.issue(requestFor(imprint), profile, resolved, SigningProtocol.INTERNAL_TSA);
+        } catch (SigningEngineException e) {
+            throw SigningEngineException.stepFailed(e.failure(), step, e.operatorMessage(), e, e.clientMessage());
+        }
+    }
+
+    private SigningProfileModel<?, ?> loadProfile(String timestampingProfileName) throws SigningEngineException {
+        SigningProfileModel<?, ?> profile;
+        try {
+            profile = signingProfileService.getSigningProfileModel(timestampingProfileName);
+        } catch (NotFoundException e) {
+            throw misconfigured("timestamping Signing Profile '%s' does not exist".formatted(timestampingProfileName),
+                    e);
+        }
+        if (!profile.enabled()) {
+            throw misconfigured("timestamping Signing Profile '%s' is disabled".formatted(timestampingProfileName),
+                    null);
+        }
+        return profile;
+    }
+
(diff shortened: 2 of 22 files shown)


src/main/java/com/otilm/core/signing/tsa/ManagedTimestampEngine.java
```diff
@@ -22,141 +22,228 @@
...
+        if (isPlatformFault(e.failure())) {
+            logger
+                    .error("Timestamp issuance failed for signing profile '{}': {}", timestampingProfile.name(),
+                            e.operatorMessage(), e);
+        } else {
+            logger
+                    .warn("Rejecting timestamp request for signing profile '{}': {}", timestampingProfile.name(),
+                            e.operatorMessage(), e);
+        }
+    }
```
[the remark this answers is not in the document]
ivosh: Fixed in aea67c1e.

src/main/java/com/otilm/core/signing/tsa/ManagedTimestampEngine.java
```diff
@@ -22,141 +22,231 @@
...
             var certificateChain = signingScheme.chain();
 
-            var token = tokenGenerator.generate(request, timestampingProfile, certificateChain, serialNumber, genTime);
-
+            TimeStampToken token = tokenGenerator
+                    .generate(request, timestampingProfile, certificateChain, serialNumber, genTime);
             if (Boolean.TRUE.equals(timestampingProfile.validateTokenSignature())) {
-                var verifier = new JcaSimpleSignerInfoVerifierBuilder().build(certificateChain.signingCertificate());
-                token.validate(verifier);
+                token.validate(new JcaSimpleSignerInfoVerifierBuilder().build(certificateChain.signingCertificate()));
```
lubomirw, line 119: **[High]** | Correctness

For `INTERNAL_TSA`, this still honors `validateTokenSignature=false`, so the in-process caller can embed and record a token whose CMS signature was never verified.

_Suggested:_ Force signature validation when `protocol == SigningProtocol.INTERNAL_TSA` (while preserving the setting for TSP), and cover an invalid-signature token on the internal path.

src/main/java/com/otilm/core/signing/tsa/ManagedTimestampEngine.java
```diff
@@ -22,141 +22,231 @@
...
             }
 
-            byte[] encodedToken = token.getEncoded();
-            recordSigning(signingProfile, request, serialNumber, genTime, encodedToken);
+            requireTokenBoundToRequest(token, request, serialNumber);
```
lubomirw, line 122: **[High]** | Security

The formatting connector builds the signed `TSTInfo`, but this binding check omits `genTime`; the connector can return a token asserting a different time while the signing record and `IssuedTimestamp` retain the unrelated local value.

_Suggested:_ Compare the token's generation time with the trusted clock value using precision-aware normalization, reject mismatches, and record/return the verified token time.

src/main/java/com/otilm/core/signing/tsa/InternalTimestampSource.java
```diff
@@ -0,0 +1,127 @@
...
+            return engine.issue(requestFor(imprint), profile, resolved, SigningProtocol.INTERNAL_TSA);
+        } catch (SigningEngineException e) {
+            throw SigningEngineException.stepFailed(e.failure(), step, e.operatorMessage(), e, e.clientMessage());
+        }
+    }
+
+    private SigningProfileModel<?, ?> loadProfile(String timestampingProfileName) throws SigningEngineException {
+        SigningProfileModel<?, ?> profile;
+        try {
+            profile = signingProfileService.getSigningProfileModel(timestampingProfileName);
```
lubomirw: **[Medium]** | Correctness

This bridge calls `getSigningProfileModel`, whose implementation carries `@ExternalAuthorization(SIGNING_PROFILE, DETAIL)`. That adds an unrelated permission gate to internal issuance even though the content-profile reference is documented as the authorization.

_Suggested:_ Expose an authorization-free internal cached loader and retain the DETAIL check only on external entry points.

lubomirw: **[Medium]** | Error handling

`getSigningProfileModel` documents `IllegalStateException` and `IllegalArgumentException` for malformed profile rows or mappings, but only `NotFoundException` is translated here, so those failures escape the step-scoped `SigningEngineException` contract.

_Suggested:_ Translate both documented runtime failures to `MISCONFIGURED` and add coverage for each case.

src/main/java/com/otilm/core/signing/tsa/ManagedTimestampEngine.java
```diff
@@ -22,141 +22,231 @@
...
+    private static void logFailure(ResolvedManagedTimestampingProfile timestampingProfile, SigningProtocol protocol,
+            SigningEngineException e) {
+        if (isPlatformFault(e.failure())) {
+            logger
+                    .error("Timestamp issuance failed for signing profile '{}' via {}: {}", timestampingProfile.name(),
+                            protocol, e.operatorMessage(), e);
+        } else {
+            logger
+                    .warn("Refusing to issue a timestamp for signing profile '{}' via {}: {}",
+                            timestampingProfile.name(), protocol, e.operatorMessage(), e);
```
lubomirw: **[Low]** | Maintainability

Would a full stack trace be useful for the non-platform WARN cases here? These are classified as requester/operator rejections and can occur per request, so passing `e` as the trailing throwable may add noisy stacks without extra triage value.

_Suggested:_ If not, omit the throwable from the WARN branch and retain stack traces for the platform-fault ERROR branch.

src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -123,7 +123,8 @@
 public class SigningProfileServiceImpl implements SigningProfileExternalService, SigningProfileInternalService {
 
     /**
-     * Defines which signing protocols are allowed for each workflow type.
+     * Defines which signing protocols an operator may enable per workflow type. Only protocols a client can actually
+     * reach belong here; {@code SigningProfileSupportedProtocolsTest} holds that line.
```
lubomirw: **[Low]** | Maintainability

Production Javadoc names `SigningProfileSupportedProtocolsTest`, coupling the explanation to an internal test/process detail that can drift independently.

_Suggested:_ End the comment after: "Only protocols a client can actually reach belong here."

<sub>flagged by Hygiene</sub>

src/main/java/com/otilm/core/signing/tsa/InternalTimestampSource.java
```diff
@@ -0,0 +1,127 @@
...
+import com.otilm.core.signing.tsa.messages.TspRequest;
+import java.util.List;
+import java.util.Optional;
+import org.springframework.stereotype.Component;
+
+/**
+ * Issues the timestamps an AdES signature needs from an ILM-managed TIMESTAMPING Signing Profile, in-process.
+ *
+ * <p>
+ * The call goes below the TSP protocol edge: no HTTP, no TSP profile, no protocol authentication. Being referenced by a
```
lubomirw: **[Nit]** | Maintainability

This class Javadoc expands into protocol-plumbing detail beyond the component's concise contract.

_Suggested:_ Consider tightening it while preserving the non-obvious authorization invariant and why `enabledProtocols` is not consulted.

<sub>flagged by Hygiene</sub>

lubomirw (changes requested):
Automated multi-reviewer triage - 7 selected findings (REQUEST_CHANGES).

src/main/java/com/otilm/core/signing/tsa/InternalTimestampSource.java
ivosh: Fixed in cf33d84f.

ivosh: Fixed in cf33d84f.

src/main/java/com/otilm/core/signing/tsa/ManagedTimestampEngine.java
ivosh: Fixed in cf33d84f.

src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
ivosh: Fixed in cf33d84f.

src/main/java/com/otilm/core/signing/tsa/InternalTimestampSource.java
ivosh: Fixed in cf33d84f.

src/main/java/com/otilm/core/signing/tsa/ManagedTimestampEngine.java
ivosh: No change. The conditional is pre-existing — `main` carries the same `if (Boolean.TRUE.equals(timestampingProfile.validateTokenSignature()))`, and this PR did not branch around it: `process` and `InternalTimestampSource` both funnel through the same `issueToken`, so the internal path verifies the token signature exactly when the profile says to. Whether the check runs is therefore the operator's configuration choice on the timestamping Signing Profile, the same as it already is for RFC 3161 callers.

src/main/java/com/otilm/core/signing/tsa/ManagedTimestampEngine.java
ivosh: No change. The engine has never verified the connector's `genTime` echo — `main` had no binding check at all — and this PR added checks only for the two fields that are exactly verifiable, the imprint and the serial. `genTime` is not: RFC 3161 encodes it as a `GeneralizedTime` whose fractional-second precision is the TSA's choice, so a conforming connector may legally stamp whole seconds, and BouncyCastle's `TimeStampTokenInfo.getGenTime()` returns a `java.util.Date`, which floors any comparison at milliseconds regardless of what the connector encoded. Any check would therefore need a tolerance rather than an equality, which is why the method's Javadoc documents the omission instead.

lubomirw (approved):

Issue #883, opened 2024-11-18
Required field commonName missing in the certificateRequest of Certificate detail


3keyroman:
**Describe the bug**
According to the [Get Certificate Details API](https://docs.czertainly.com/api/core-certificate/#tag/Certificate-Inventory/operation/getCertificate) the `certificateRequest` field, when present, has a required field `commonName`. However, when we submit CSR with empty Common Name, it does not return `commonName` field.

This breaks REST API clients that relies on the OpenAPI specification.

**To Reproduce**
1. Create CSR with empty CN
2. Request certificate using [Issue Certificate API](https://docs.czertainly.com/api/core-client-operations/#tag/Client-Operations-v2/operation/issueCertificate) and obtain its UUID from response
3. Request details about the certificate using UUID and [Get Certificate Details API](https://docs.czertainly.com/api/core-certificate/#tag/Certificate-Inventory/operation/getCertificate)
4. See the response body and `certificateRequest` field that contains no `commonName` field

**Expected behavior**
The `commonName` field in `certificateRequest` field in response from the [Get Certificate Details API](https://docs.czertainly.com/api/core-certificate/#tag/Certificate-Inventory/operation/getCertificate) should be defined with `<empty>` as it is in certificate object Common Name, or it should not be mandatory by the OpenAPI specification.

**Additional context**
Tested on version 2.13.0

PR #2110, opened 2026-08-19
Wire discovery v2 clients and capability gate

### LINKED ISSUE #1961: Wire discovery v2 clients and capability gate
### Description

Wire the discovery v2 clients from the interfaces release through `ConnectorApiFactory`, and generalize the ENFORCED feature-flag gate.

### Scope

- **No pom bump.** Core already depends on `interfaces` at `2.20.0-SNAPSHOT`, the same line the discovery v2 work publishes to — the contract arrives the moment a snapshot publishes, which is what #1991 exists to absorb and why it lands first.
- `getDiscoveryV2ApiClient(ApiClientConnectorInfo)` accessor routing REST vs MQ via `shouldUseMq` (existing chokepoint); REST bean in `ApplicationConfig`, MQ bean in `ProxyClientConfig`.
- **The drain response size bound is `ClientTuning.maxInMemorySize` on the single shared WebClient — never a per-client cap.** A per-client cap is unsafe here: `buildCertificateWebClient` derives the cached mTLS client from the instance `webClient` and stores it in a static cache keyed by connector UUID alone, so for a connector implementing two interfaces whichever client warmed the cache would win — either discovery's cap never applies on the mTLS path, or it leaks onto the authority client and breaks `getCrl`, whose CRL bodies reach tens of megabytes.
- Expose `maxInMemorySize` as configuration: add it to `ConnectorApiClientProperties` and the `connector.api-client` block in `application.yml`, so `CONNECTOR_API_CLIENT_MAX_IN_MEMORY_SIZE` binds like the other four knobs. Currently fixed at the 16 MB default with no override.
- Property surface for `DiscoveryMqTimeouts` (`status`/`drain`/`control`), so a deployment whose connectors return large batches can raise the drain budget. Defaults 30s each.
- Override `ProxyClientImpl.sendRequestForEntity(..., Duration)` — the new overload the discovery MQ client calls. `ProxyClientImpl` overrides only the five-argument form, so the six-argument one falls through to an interface default that wraps in `ResponseEntity.ok`, and MQ `cancel` reports 200 where the contract declares 204. `sendRequestForEntityAsync` already takes the timeout, so this is a delegation, not new machinery.
- `ConnectorCapabilityService` generalization keyed on the bound connector-interface row (not connector+code — avoids the multi-version ambiguity its javadoc warns about); authority-typed overload kept.
- ~~Supported-resource sync at connector reconcile~~ **Dropped 2026-08-20** with the per-resource capability model (adversarial review; contract change in OmniTrustILM/interfaces#883). Nothing is persisted: the resource listing relays live and create-time validation calls `listSupportedResources` inline — both in core#1964. Decision record: `docs/superpowers/2026-08-04-discovery-provider-api-v2/per-run-stoppability-decision.md` (planning workspace).

### Acceptance Criteria

- [ ] `CONNECTOR_API_CLIENT_MAX_IN_MEMORY_SIZE` overrides the cap end to end; a response above it fails as a connector error rather than an OOM
- [ ] MQ `cancel` returns 204, not 200, once the six-argument override lands
- [ ] The bodiless-error change that arrives with the snapshot is safe: a bodiless 4xx/5xx from a connector now arrives as a mapped `ConnectorException` where it previously escaped as `IllegalStateException`, so every existing `catch (ConnectorException)` around a client call still behaves correctly — `CertificateStatusPollListener` in particular, which sits on the JMS retry/DLQ boundary
- [ ] Factory tests: proxied → MQ, direct → REST, missing MQ bean → REST fallback
- [ ] Gate tests per interface row (~~reconcile persistence~~ dropped per the scope note above)

### References

- Epic: OmniTrustILM/ilm#267 (core plan Tasks 4–5)
- Snapshot absorption: #1991

lubomirw:
Closes #1961 (epic OmniTrustILM/ilm#267, core plan Tasks 4–5).

> **Merge order: OmniTrustILM/interfaces#883 first.** The last commit adopts its contract (`DiscoveryDetailDto.stoppable` replaces `effectiveCapabilities`), so this branch compiles only against the snapshot carrying that change. Conversely, once #883 merges, core main stops compiling until this lands — the two should merge back-to-back.

## Client wiring (Task 4)

- REST and MQ discovery v2 clients become beans and route through `getDiscoveryApiClientV2` via the existing `shouldUseMq` chokepoint, with REST fallback when the MQ bean is absent.
- The drain response size bound lands as `ClientTuning.maxInMemorySize` on the single shared WebClient, surfaced as `connector.api-client.max-in-memory-size` (DataSize, validated < 2 GB) — deliberately never a per-client cap, which would collide with `BaseApiClient`'s connector-keyed mTLS client cache.
- The MQ client carries per-operation timeout budgets from `discovery.mq-timeouts.*` (status/drain/control, 30s defaults), and `ProxyClientImpl` overrides the six-argument `sendRequestForEntity` so an explicit timeout keeps the status-preserving path — without it, MQ `cancel`'s 204 arrives as the interface default's `ResponseEntity.ok`.

## Capability gate (Task 5, slimmed)

`ConnectorCapabilityService.supports(ConnectorInterfaceEntity, FeatureFlag)` is the new primary entry point, keyed on the bound interface row so it stays correct across multiple versions of one interface; the authority-typed overload delegates to it.

**Plan deviation:** the plan's supported-resources persistence (column, migration, reconcile sync) is deliberately absent. An independent adversarial review of the per-resource capability model led to replacing it with per-run stoppability declared at initiate (interfaces#883); the resource listing relays live in core#1964. Decision record: `docs/superpowers/2026-08-04-discovery-provider-api-v2/per-run-stoppability-decision.md` in the planning workspace.

## Review already folded in

A four-source multi-review ran pre-opening; its round is the last commit and includes two finds worth a reviewer's eye: a duplicate top-level `discovery:` YAML key that SnakeYAML 2 rejects at boot but no test could see (the test-classpath yml shadows the main one), and the context-signature baseline bump recording the deliberately forked proxy-disabled test context.

Deployment note: the four new env vars need chart enumeration — folded into OmniTrustILM/helm-charts#321.

### DIFF
--- a/src/main/java/com/otilm/core/messaging/proxy/DiscoveryMqTimeoutsProperties.java
+++ b/src/main/java/com/otilm/core/messaging/proxy/DiscoveryMqTimeoutsProperties.java
@@ -1,0 +1,15 @@
+package com.otilm.core.messaging.proxy;
+
+import java.time.Duration;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+/**
+ * Per-operation timeout budgets for the discovery v2 MQ client — bound from {@code discovery.mq-timeouts.*} so a
+ * deployment whose connectors return large drain batches can raise the drain budget without touching the proxy-wide
+ * request timeout. Positivity is enforced by {@code DiscoveryMqTimeouts} itself at bean construction.
+ */
+@ConfigurationProperties(prefix = "discovery.mq-timeouts")
+@Validated
+public record DiscoveryMqTimeoutsProperties(Duration status, Duration drain, Duration control) {
+}
--- a/src/main/java/com/otilm/core/messaging/proxy/DiscoveryMqTimeoutsProperties.java
+++ b/src/main/java/com/otilm/core/messaging/proxy/DiscoveryMqTimeoutsProperties.java
@@ -1,5 +1,6 @@
 package com.otilm.core.messaging.proxy;
 
+import jakarta.validation.constraints.NotNull;
 import java.time.Duration;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.validation.annotation.Validated;
@@ -7,9 +8,11 @@ import org.springframework.validation.annotation.Validated;
 /**
  * Per-operation timeout budgets for the discovery v2 MQ client — bound from {@code discovery.mq-timeouts.*} so a
  * deployment whose connectors return large drain batches can raise the drain budget without touching the proxy-wide
- * request timeout. Positivity is enforced by {@code DiscoveryMqTimeouts} itself at bean construction.
+ * request timeout. A missing key fails at binding with the full property path; positivity is enforced by
+ * {@code DiscoveryMqTimeouts} itself at bean construction.
  */
 @ConfigurationProperties(prefix = "discovery.mq-timeouts")
 @Validated
-public record DiscoveryMqTimeoutsProperties(Duration status, Duration drain, Duration control) {
+public record DiscoveryMqTimeoutsProperties(@NotNull Duration status, @NotNull Duration drain,
+        @NotNull Duration control) {
 }
--- a/src/main/java/com/otilm/core/messaging/proxy/DiscoveryMqTimeoutsProperties.java
+++ b/src/main/java/com/otilm/core/messaging/proxy/DiscoveryMqTimeoutsProperties.java
@@ -6,9 +6,14 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.validation.annotation.Validated;
 
 /**
- * Per-operation timeout budgets for the discovery v2 MQ client — bound from {@code discovery.mq-timeouts.*} so a
- * deployment whose connectors return large drain batches can raise the drain budget without touching the proxy-wide
- * request timeout. A missing key fails at binding with the full property path; positivity is enforced by
+ * Per-operation timeout budgets for the discovery v2 MQ client, bound from {@code discovery.mq-timeouts.*}.
+ *
+ * <p>
+ * <b>Customization:</b> a deployment whose connectors return large drain batches raises the drain budget here, without
+ * touching the proxy-wide request timeout.
+ *
+ * <p>
+ * <b>Validation:</b> a missing key fails at binding, naming the full property path. Positivity is enforced by
  * {@code DiscoveryMqTimeouts} itself at bean construction.
  */
 @ConfigurationProperties(prefix = "discovery.mq-timeouts")
--- a/src/main/java/com/otilm/core/messaging/proxy/ProxyClientConfig.java
+++ b/src/main/java/com/otilm/core/messaging/proxy/ProxyClientConfig.java
@@ -15,6 +15,7 @@ import com.otilm.api.clients.mq.LocationApiClient;
 import com.otilm.api.clients.mq.NotificationInstanceApiClient;
 import com.otilm.api.clients.mq.ProxyClient;
 import com.otilm.api.clients.mq.TokenInstanceApiClient;
+import com.otilm.api.clients.mq.discovery.v2.DiscoveryMqTimeouts;
 import com.otilm.api.clients.mq.signing.SignatureFormattingApiClient;
 import com.otilm.api.clients.mq.signing.contentsigning.ContentSigningFormattingApiClient;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
@@ -27,7 +28,7 @@ import org.springframework.context.annotation.Configuration;
  */
 @Configuration
 @ConditionalOnProperty(name = "proxy.enabled", havingValue = "true")
-@EnableConfigurationProperties(ProxyProperties.class)
+@EnableConfigurationProperties({ProxyProperties.class, DiscoveryMqTimeoutsProperties.class})
 public class ProxyClientConfig {
 
     /**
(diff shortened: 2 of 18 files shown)


src/main/java/com/otilm/core/service/handler/ConnectorCapabilityService.java
```diff
@@ -20,21 +20,29 @@
...
-        return authority != null && advertises(authority.getConnectorInterface(), flag);
+        return advertises(iface, flag);
+    }
+
+    /**
+     * Per-authority check — reads the authority's own bound interface entity directly.
+     */
+    public boolean supports(AuthorityInstanceReference authority, FeatureFlag flag) {
+        return supports(authority != null ? authority.getConnectorInterface() : null, flag);
     }
```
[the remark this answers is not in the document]
lubomirw: Done — core#1961's description and acceptance criteria now carry the scope change: the supported-resource sync bullet is struck with a dated "Dropped" note pointing at the contract change (interfaces#883), the live-relay/inline-validation successors in core#1964, and the decision record in the planning workspace. The reconcile-persistence AC is struck the same way, so release notes and reviewers see exactly what shipped here.

src/test/java/com/otilm/core/integration/client/ConnectorApiFactoryDiscoveryV2ITest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Autowired
+    private ConnectorApiFactory connectorApiFactory;
+    @Autowired
+    private ConnectorRepository connectorRepository;
+
+    @Test
+    void directConnectorGetsTheRestClient() {
+        DiscoverySyncApiClient client = connectorApiFactory.getDiscoveryApiClientV2(savedConnectorDto(null));
+
+        Assertions.assertInstanceOf(com.otilm.api.clients.discovery.v2.DiscoveryApiClient.class, client);
```
ivosh: **. Nitpick** | Testing

ConnectorApiFactoryDiscoveryV2ITest.java:28 and :35 and ConnectorApiFactoryDiscoveryV2NoProxyITest.java:41 use JUnit `Assertions.assertInstanceOf`. The analogous REST/MQ selection tests in ConnectorApiFactorySignatureFormattingITest.java:38 and :50 use AssertJ `assertThat(...).isInstanceOf(...)`, matching the convention used by this PR's other new tests.

_Suggested:_ Replace all three `Assertions.assertInstanceOf` calls with AssertJ `assertThat(client).isInstanceOf(...)`.

src/main/java/com/otilm/core/messaging/proxy/DiscoveryMqTimeoutsProperties.java
```diff
@@ -0,0 +1,18 @@
+package com.otilm.core.messaging.proxy;
+
+import jakarta.validation.constraints.NotNull;
+import java.time.Duration;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+/**
+ * Per-operation timeout budgets for the discovery v2 MQ client — bound from {@code discovery.mq-timeouts.*} so a
```
ivosh: **. Nitpick** | Documentation

DiscoveryMqTimeoutsProperties.java:9-12 packs the binding source, customization rationale, missing-key behavior, and delegated positivity validation into one unstructured paragraph. This PR makes the load-bearing validation detail easy to miss among neighboring concerns.

_Suggested:_ Rewrite it as a one-line summary followed by `<p><b>Customization:</b> ...` and `<p><b>Validation:</b> ...` paragraphs with short sentences.

src/main/java/com/otilm/core/messaging/proxy/ProxyClientConfig.java
```diff
@@ -206,6 +207,17 @@ public com.otilm.api.clients.mq.v3.AuthorityApiClient mqAuthorityApiClientV3(Pro
         return new com.otilm.api.clients.mq.v3.AuthorityApiClient(proxyClient);
     }
 
+    /**
+     * Create MQ-based discovery v2 client bean. Unlike its siblings it carries per-operation timeout budgets, because a
```
ivosh: **. Nitpick** | Documentation

ProxyClientConfig.java:211 says the method creates an MQ discovery-v2 client bean, which is already explicit in `@Bean`, the method name, and the return type. For this PR, the useful documentation is only the non-obvious reason this client receives per-operation timeout budgets.

_Suggested:_ Remove the declarative first sentence and retain only the timeout-budget rationale.

ivosh (dismissed):

src/test/java/com/otilm/core/integration/client/ConnectorApiFactoryDiscoveryV2ITest.java
lubomirw: Fixed in 5aed17cc3 — all three assertions switched to AssertJ `assertThat(client).isInstanceOf(...)`, matching the signature-formatting sibling.

src/main/java/com/otilm/core/messaging/proxy/DiscoveryMqTimeoutsProperties.java
lubomirw: Restructured in 5aed17cc3 — one-line summary plus **Customization** and **Validation** paragraphs, same shape as the interfaces#883 counterpart.

src/main/java/com/otilm/core/messaging/proxy/ProxyClientConfig.java
lubomirw: Trimmed in 5aed17cc3 — the declarative sentence is gone; the javadoc keeps only the timeout-budget rationale.

klaraf755 (approved):

PR #2116, opened 2026-08-20
Handle comment events and deliver notifications

### LINKED ISSUE #2026: Comment event handlers and notification delivery

klaraf755:
Closes #2026

Event handlers and notification delivery for the comment subsystem merged in #2105 (Epic: OmniTrustILM/ilm#246). #2105 already publishes `COMMENT_CREATED`/`COMMENT_RESOLVED`; this PR makes them do something.

## What's here

- **Handlers** — `CommentCreatedEventHandler` and `CommentResolvedEventHandler` (bean names = event codes) on a shared `CommentEventsHandler` base. A root comment notifies the host object's owner — resolved once, at publish time, and published as a plain user recipient, so the self-exclusion decision and the delivered recipient can never disagree. A reply notifies the thread's participants (distinct authors, minus the actor); resolve/reopen notifies participants with the `resolved` flag in the payload. Group-scoped trigger associations are honored via the host object's group memberships.
- **Subject resolution** — `NotificationSubjectResolver` treats a comment event's subject as the **host object**: owner resolution, enrichment, persisted deep links and DEFAULT recipients all key on the host, matching handler-driven follow-ups. Profile-driven internal notifications persist the resolved subject as their target (a `COMMENT` target would deep-link nowhere).
- **DEFAULT recipients** — comment events join the owner+groups default arm, resolved against the subject; previously a DEFAULT profile on a comment event silently resolved to nobody.
- **Trigger conditions** — `FilterField` gains COMMENT properties (host resource, author as platform-user list, body, timestamps) including a new presence-only `SearchFieldTypeEnum.PRESENCE` for the parent reference: a raw UUID means nothing to a user, so only empty/not-empty is offered — 'Parent Comment is empty' is the roots-only condition. `ResourceToClass` maps COMMENT for trigger evaluation.
- **Internal notifications carry no comment body** — the notification points at the thread; user-authored Markdown never mirrors into persisted notification text. The event payload keeps the body verbatim for operator-bound templates.
- **Tests** — handler unit tests, the end-to-end handler ITest (group scoping, participants, owner, condition-scoped roots-only trigger, bodiless notifications), evaluator coverage for every new filter field, subject-resolver and DEFAULT-recipient units, thread-participant repository queries. Shared `AuthServiceWireMockStubs` dedupes the auth-service stub JSON; `ProducerMocks` gains a synchronous `NotificationMessage` recorder.

## Notes for reviewers

- OBJECT-recipient profiles on comment events are **deliberately reachable**: an admin who binds one is routing comment notifications to the contact stored on the host object's attributes. Pinned by `commentEventsResolveToTheHostObject`.
- The missing-handler guard in `EventListener` landed in #2105, so deployment ordering between the two PRs was already safe.
- Full suite: 5166 tests green.

### DIFF
--- a/src/main/java/com/otilm/core/enums/FilterField.java
+++ b/src/main/java/com/otilm/core/enums/FilterField.java
@@ -57,6 +57,7 @@ import com.otilm.core.dao.entity.Group_;
 import com.otilm.core.dao.entity.Location_;
 import com.otilm.core.dao.entity.OwnerAssociation_;
 import com.otilm.core.dao.entity.RaProfile_;
+import com.otilm.core.dao.entity.ResourceObjectAssociation_;
 import com.otilm.core.dao.entity.ScheduledJob_;
 import com.otilm.core.dao.entity.Secret2SyncVaultProfile_;
 import com.otilm.core.dao.entity.Secret_;
@@ -310,17 +311,19 @@ public enum FilterField {
             ResourceAction.class),
     APPROVAL_STATUS(Resource.APPROVAL, null, null, Approval_.status, Constants.STATUS, SearchFieldTypeEnum.LIST,
             ApprovalStatusEnum.class),
-    APPROVAL_CREATED_AT(Resource.APPROVAL, null, null, Approval_.createdAt, "Created At", SearchFieldTypeEnum.DATETIME),
+    APPROVAL_CREATED_AT(Resource.APPROVAL, null, null, Approval_.createdAt, Constants.CREATED_AT,
+            SearchFieldTypeEnum.DATETIME),
     APPROVAL_EXPIRY_AT(Resource.APPROVAL, null, null, Approval_.expiryAt, "Expiry At", SearchFieldTypeEnum.DATETIME),
     APPROVAL_CLOSED_AT(Resource.APPROVAL, null, null, Approval_.closedAt, "Closed At", SearchFieldTypeEnum.DATETIME),
 
     // Comment
-    COMMENT_HOST_RESOURCE(Resource.COMMENT, null, null, Comment_.resource, "Host Resource", SearchFieldTypeEnum.LIST,
-            Resource.class),
+    COMMENT_HOST_RESOURCE(Resource.COMMENT, null, null, ResourceObjectAssociation_.resource, "Host Resource",
+            SearchFieldTypeEnum.LIST, Resource.class),
     COMMENT_PARENT(Resource.COMMENT, null, null, Comment_.parentUuid, "Parent Comment", SearchFieldTypeEnum.PRESENCE),
     COMMENT_AUTHOR(Resource.COMMENT, Resource.USER, null, Comment_.authorUsername, "Author", SearchFieldTypeEnum.LIST),
     COMMENT_BODY(Resource.COMMENT, null, null, Comment_.body, "Body", SearchFieldTypeEnum.STRING),
-    COMMENT_CREATED_AT(Resource.COMMENT, null, null, Comment_.createdAt, "Created At", SearchFieldTypeEnum.DATETIME),
+    COMMENT_CREATED_AT(Resource.COMMENT, null, null, Comment_.createdAt, Constants.CREATED_AT,
+            SearchFieldTypeEnum.DATETIME),
     COMMENT_RESOLVED_AT(Resource.COMMENT, null, null, Comment_.resolvedAt, "Resolved At", SearchFieldTypeEnum.DATETIME),
 
     // OID Entry
@@ -424,7 +427,7 @@ public enum FilterField {
             SearchFieldTypeEnum.DATETIME),
     SIGNING_RECORD_SIGNED_DOCUMENT_RETRIEVED_AT(Resource.SIGNING_RECORD, null, null,
             SigningRecord_.signedDocumentRetrievedAt, "Signed Document Retrieved At", SearchFieldTypeEnum.DATETIME),
-    SIGNING_RECORD_CREATED(Resource.SIGNING_RECORD, null, null, Audited_.created, "Created At",
+    SIGNING_RECORD_CREATED(Resource.SIGNING_RECORD, null, null, Audited_.created, Constants.CREATED_AT,
             SearchFieldTypeEnum.DATETIME);
 
     private static final FilterField[] VALUES;
@@ -489,6 +492,7 @@ public enum FilterField {
         public static final String STATE = "State";
         public static final String ENABLED = "Enabled";
         public static final String STATUS = "Status";
+        public static final String CREATED_AT = "Created At";
     }
 
 }
--- a/src/main/java/com/otilm/core/enums/FilterField.java
+++ b/src/main/java/com/otilm/core/enums/FilterField.java
@@ -44,6 +44,7 @@ import com.otilm.core.dao.entity.CertificateProtocolAssociation_;
 import com.otilm.core.dao.entity.CertificateRelation_;
 import com.otilm.core.dao.entity.CertificateRequestEntity_;
 import com.otilm.core.dao.entity.Certificate_;
+import com.otilm.core.dao.entity.Comment_;
 import com.otilm.core.dao.entity.Connector2FunctionGroup_;
 import com.otilm.core.dao.entity.ConnectorInterfaceEntity_;
 import com.otilm.core.dao.entity.Connector_;
@@ -312,6 +313,15 @@ public enum FilterField {
     APPROVAL_CREATED_AT(Resource.APPROVAL, null, null, Approval_.createdAt, "Created At", SearchFieldTypeEnum.DATETIME),
     APPROVAL_EXPIRY_AT(Resource.APPROVAL, null, null, Approval_.expiryAt, "Expiry At", SearchFieldTypeEnum.DATETIME),
     APPROVAL_CLOSED_AT(Resource.APPROVAL, null, null, Approval_.closedAt, "Closed At", SearchFieldTypeEnum.DATETIME),
+
+    // Comment
+    COMMENT_HOST_RESOURCE(Resource.COMMENT, null, null, Comment_.resource, "Host Resource", SearchFieldTypeEnum.LIST,
+            Resource.class),
+    COMMENT_PARENT(Resource.COMMENT, null, null, Comment_.parentUuid, "Parent Comment", SearchFieldTypeEnum.PRESENCE),
+    COMMENT_AUTHOR(Resource.COMMENT, Resource.USER, null, Comment_.authorUsername, "Author", SearchFieldTypeEnum.LIST),
+    COMMENT_BODY(Resource.COMMENT, null, null, Comment_.body, "Body", SearchFieldTypeEnum.STRING),
+    COMMENT_CREATED_AT(Resource.COMMENT, null, null, Comment_.createdAt, "Created At", SearchFieldTypeEnum.DATETIME),
+    COMMENT_RESOLVED_AT(Resource.COMMENT, null, null, Comment_.resolvedAt, "Resolved At", SearchFieldTypeEnum.DATETIME),
 
     // OID Entry
     OID_ENTRY_OID(Resource.OID, null, null, CustomOidEntry_.oid, "OID", SearchFieldTypeEnum.STRING),
--- a/src/main/java/com/otilm/core/events/handlers/CommentEventsHandler.java
+++ b/src/main/java/com/otilm/core/events/handlers/CommentEventsHandler.java
@@ -12,6 +12,9 @@ import com.otilm.core.events.EventContextTriggers;
 import com.otilm.core.events.EventHandler;
 import com.otilm.core.messaging.model.NotificationMessage;
 import com.otilm.core.messaging.model.NotificationRecipient;
+import com.otilm.core.model.auth.ResourceAction;
+import com.otilm.core.security.authz.AuthorizationEnforcer;
+import com.otilm.core.security.authz.SecuredUUID;
 import com.otilm.core.service.ResourceObjectAssociationService;
 import java.util.ArrayList;
 import java.util.List;
@@ -28,6 +31,8 @@ public abstract class CommentEventsHandler extends EventHandler<Comment> {
 
     protected ResourceObjectAssociationService resourceObjectAssociationService;
 
+    private AuthorizationEnforcer authorizationEnforcer;
+
     protected CommentEventsHandler(CommentRepository repository, TriggerEvaluator<Comment> triggerEvaluator) {
         super(repository, triggerEvaluator);
         this.commentRepository = repository;
@@ -36,6 +41,11 @@ public abstract class CommentEventsHandler extends EventHandler<Comment> {
     @Autowired
     public void setResourceObjectAssociationService(ResourceObjectAssociationService resourceObjectAssociationService) {
         this.resourceObjectAssociationService = resourceObjectAssociationService;
+    }
+
+    @Autowired
+    public void setAuthorizationEnforcer(AuthorizationEnforcer authorizationEnforcer) {
+        this.authorizationEnforcer = authorizationEnforcer;
     }
 
     @Override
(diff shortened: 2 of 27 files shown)


src/main/java/com/otilm/core/service/notifications/NotificationSubjectResolver.java
```diff
@@ -24,6 +25,10 @@ public static SubjectRef resolveSubject(Resource resource, UUID objectUuid, Even
                 && approval.getObjectUuid() != null) {
             return new SubjectRef(approval.getResource(), approval.getObjectUuid());
         }
+        if (eventData instanceof CommentEventData comment && comment.getResource() != null
+                && comment.getObjectUuid() != null) {
+            return new SubjectRef(comment.getResource(), comment.getObjectUuid());
```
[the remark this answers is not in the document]
klaraf755: The whitelist is per-resource on purpose: OBJECT recipients read raw attribute content, so each resource on the list is a deliberate decision about exposing its attributes to notification connectors. Approvals work the same way — polymorphic targets, but OBJECT delivery still resolves only for certificates. Comments inherit that rule; widening the list just because comments can sit on other hosts would expose e.g. secret attributes as a side effect. If a real use case for another host shows up, adding it is a one-line decision then.

src/main/java/com/otilm/core/enums/FilterField.java
```diff
@@ -309,10 +311,21 @@ public enum FilterField {
...
 
+    // Comment
+    COMMENT_HOST_RESOURCE(Resource.COMMENT, null, null, ResourceObjectAssociation_.resource, "Host Resource",
+            SearchFieldTypeEnum.LIST, Resource.class),
+    COMMENT_PARENT(Resource.COMMENT, null, null, Comment_.parentUuid, "Parent Comment", SearchFieldTypeEnum.PRESENCE),
+    COMMENT_AUTHOR(Resource.COMMENT, Resource.USER, null, Comment_.authorUsername, "Author", SearchFieldTypeEnum.LIST),
+    COMMENT_BODY(Resource.COMMENT, null, null, Comment_.body, "Body", SearchFieldTypeEnum.STRING),
+    COMMENT_CREATED_AT(Resource.COMMENT, null, null, Comment_.createdAt, Constants.CREATED_AT,
+            SearchFieldTypeEnum.DATETIME),
+    COMMENT_RESOLVED_AT(Resource.COMMENT, null, null, Comment_.resolvedAt, "Resolved At", SearchFieldTypeEnum.DATETIME),
```
ivosh, line 327: **! Major** | Correctness

FilterField.java:327 adds `COMMENT_RESOLVED_AT` as a trigger discriminator, but CommentEventsHandler.java:25 inherits `EventHandler.prepareContext()`, which reloads the current Comment rather than its event-time state. With five concurrent event consumers, a quick resolve followed by reopen can make the earlier resolve observe a null `resolvedAt`, causing `NOT_EMPTY` rules to be skipped even though its payload says `resolved=true`.

_Suggested:_ Evaluate mutable comment trigger fields from an event-time snapshot using the payload's `resolved` flag, and eventually add a resolve-then-reopen concurrency test.

<sub>flagged by Correctness & races + Transaction & concurrency</sub>

src/main/java/com/otilm/core/events/handlers/CommentResolvedEventHandler.java
```diff
@@ -0,0 +1,27 @@
...
+
+    @Autowired
+    protected CommentResolvedEventHandler(CommentRepository repository, TriggerEvaluator<Comment> triggerEvaluator) {
+        super(repository, triggerEvaluator);
+    }
+
+    @Override
+    protected void sendFollowUpEventsNotifications(EventContext<Comment> eventContext) {
+        Comment comment = eventContext.getResourceObjects().getFirst();
+        publishFollowUpNotification(eventContext, comment,
```
ivosh, line 25: **! Major** | Correctness

CommentServiceImpl.java:248 reads the comment outside a transaction, while CommentWriter.java:45 and CommentWriter.java:50 perform unconditional updates in separate transactions without checking the prior resolved state. Repeated or racing resolve/reopen requests can each publish an event even when no corresponding transition occurred; this PR's CommentResolvedEventHandler.java:24 then delivers duplicate or contradictory notifications.

_Suggested:_ Move the resolution-state check and update into one transactional writer method that locks with `findWithLockByUuid` and publishes only after a real transition commits.

src/main/java/com/otilm/core/events/handlers/CommentEventsHandler.java
```diff
@@ -0,0 +1,78 @@
...
+        for (UUID groupUuid : resourceObjectAssociationService
+                .getGroupUuids(comment.getResource(), comment.getObjectUuid())) {
+            eventContextTriggers.add(fetchEventTriggers(eventContext, Resource.GROUP, groupUuid));
+        }
+        return eventContextTriggers;
+    }
+
+    protected List<NotificationRecipient> threadParticipantsExcept(UUID rootUuid, UUID actingUser) {
+        return commentRepository
+                .findThreadParticipantUuids(rootUuid)
```
ivosh, line 81: **! Major** | Security

src/main/java/com/otilm/core/events/handlers/CommentEventsHandler.java:59 converts every historical thread author into a recipient without checking current DETAIL access to the host object. A user removed from the host's authorized group can therefore continue receiving notifications that disclose the actor and host type/name via NotificationListener.java:900-914.

_Suggested:_ Filter thread participants through current host-object DETAIL authorization before publishing their notifications.

ivosh (changes requested):

src/main/java/com/otilm/core/enums/FilterField.java
klaraf755: Working as intended, and uniformly with every other handler: `prepareContext` loads the current object because conditions describe the object as it is *now*, not as it was when the event was queued. A "Resolved At is not empty" rule declining to fire for a resolve that has already been reopened isn't a missed rule — it's the rule correctly refusing to act on a state that no longer exists, and that's the behaviour we want. Evaluating from the payload instead would make comment triggers behave unlike certificate or approval triggers, which have the same property.

Note the follow-up notifications to thread participants are payload-driven and unconditional, so both the resolve and the reopen are still reported — both actions genuinely happened.

src/main/java/com/otilm/core/events/handlers/CommentResolvedEventHandler.java
klaraf755: Fixed in f990b517e. Both updates now carry the state predicate (`AND c.resolvedAt IS NULL` / `IS NOT NULL`), so the state check and the write are a single atomic statement — no read-then-write window. A request that changes nothing updates no rows and publishes nothing, so repeated or racing calls can't produce duplicate or contradictory notifications; a genuinely missing comment still fails with not found, from the read that precedes the update. Covered by two tests: a repeated resolve leaves `resolvedAt` unchanged without throwing, and publishes no second event.

src/main/java/com/otilm/core/events/handlers/CommentEventsHandler.java
klaraf755: Fixed in 4a5935e45 — each participant is now re-authorized against the host before being notified. Rather than a partial check, this goes through a new enforcer entry point that decides for a named user instead of the current principal: it resolves that user to a principal and hands it to `ExternalAuthorizationCore.decide`, which already accepts the `Authentication` explicitly, so the complete decision applies — resource permission, object permission, owner and group-member fallbacks. That matters: an "is the participant the owner or in the host's groups" approximation would have dropped everyone whose access comes from a role grant, which is most admins.

No impersonation is involved, so the security context is untouched; the one real hazard was that resolving a user replays their actor MDC even on a cache hit, which would misattribute the caller's audited records — the caller's actor info is snapshotted and restored around the resolution, with a test for it. A user the auth service no longer resolves, or cannot be reached for, is denied rather than notified.

One consistency note this surfaces: comments are now the first authorization-filtered recipient set in the platform — a notification profile configured with a `USER` recipient still delivers with no access check. Happy to raise that as a platform issue if you want the two aligned.

ivosh (dismissed):

src/main/java/com/otilm/core/events/handlers/CommentEventsHandler.java
```diff
@@ -0,0 +1,78 @@
...
+import com.otilm.core.messaging.model.NotificationRecipient;
+import com.otilm.core.service.ResourceObjectAssociationService;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.Transactional;
+
+@Component
```
lubomirw, line 32: **[High]** | Performance

CommentEventsHandler is class-level @Transactional, so the inherited trigger evaluation path can impersonate the trigger owner while the handler transaction is open. EventHandler.handleUser calls AuthHelper.authenticateAsUser, which can perform auth-service HTTP through PlatformAuthenticationClient.authenticateByUserUuid, holding DB transaction/connection resources across external I/O.

_Suggested:_ Remove the ambient transaction from comment handler orchestration, or mark the outer handler flow Propagation.NOT_SUPPORTED and keep only the repository writes in short transactional operations, so trigger-owner auth-service calls happen outside a DB transaction.

src/main/java/com/otilm/core/messaging/jms/listeners/NotificationListener.java
```diff
@@ -203,8 +203,10 @@ private void sendByNotificationProfile(UUID notificationProfileUuid, Notificatio
         // sent in corresponding event handlers
         if (sendInternalNotifications) {
             try {
+                // The persisted notification targets the resolved subject, not the event record: for comment
+                // events the deep link has to point at the host object, matching handler-driven follow-ups.
                 InternalNotificationOutcome outcome = sendInternalNotifications(recipients,
-                        getInternalNotificationData(message), message.getResource(), message.getObjectUuid());
+                        getInternalNotificationData(message), subject.resource(), subject.objectUuid());
```
lubomirw: **[Medium]** | Correctness

sendByNotificationProfile now persists internal notifications against the resolved subject for every profile-driven event. That is needed for comment events, but approval subjects are also resolved to their target object, so approval notifications can deep-link to the approved object instead of the approval record/details.

_Suggested:_ Limit resolved-subject persistence to comment events, or otherwise preserve Resource.APPROVAL plus the approval UUID for approval internal notifications while still using the resolved subject for comment-host notifications.

src/main/java/com/otilm/core/events/handlers/CommentEventsHandler.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    protected List<EventContextTriggers> getOverridingTriggers(EventContext<Comment> eventContext, Comment comment)
+            throws EventException {
+        List<EventContextTriggers> eventContextTriggers = new ArrayList<>();
+        for (UUID groupUuid : resourceObjectAssociationService
+                .getGroupUuids(comment.getResource(), comment.getObjectUuid())) {
+            eventContextTriggers.add(fetchEventTriggers(eventContext, Resource.GROUP, groupUuid));
+        }
+        return eventContextTriggers;
```
lubomirw, line 70: **[Medium]** | Correctness

	hreadParticipantsExcept queries indThreadParticipantUuids(rootUuid) when the event is handled, so follow-up recipients come from the thread's current participant set rather than the participant set at the time of the comment event. If event processing is delayed and another user replies before the older event is handled, that later participant can receive a notification for an earlier event they were not part of.

_Suggested:_ Capture the participant UUIDs when publishing the comment event, or query participants using an event-time cutoff so recipient selection reflects the thread state at the event time.

lubomirw (changes requested):
Automated multi-reviewer triage — 3 findings (REQUEST_CHANGES).

lubomirw (dismissed):

src/main/java/com/otilm/core/events/handlers/CommentEventsHandler.java
klaraf755: Fixed in cc1ddf7f6 — comment orchestration is `NOT_SUPPORTED` now, so the whole flow runs without an ambient transaction.

The sharpest instance was actually this PR's own participant authorization rather than the trigger path: it makes a blocking OPA call per participant, so a pooled connection was held across N external round-trips. Taking the transaction off the orchestration covers that and the trigger-owner impersonation you point at. Nothing in comment handling writes — the event-history rows the base class keeps go through their own short transactions, which `CommentEventHandlersITest` asserts (it checks the history row for a group-scoped trigger, so a lost write fails the build).

This deliberately deviates from the six sibling handlers, since none of them authorizes anybody while handling; the class javadoc says so, to stop the inconsistency being "fixed" later. The base handler's transaction is unchanged for every other event.

src/main/java/com/otilm/core/messaging/jms/listeners/NotificationListener.java
klaraf755: Good catch — that was a regression this PR introduced, not an intended change. Fixed in 337a45793: only the two comment events point at the resolved subject, and every other event keeps the coordinates its message carries, so an approval notification opens the approval again rather than the object awaiting it.

The decision is a static kernel (`internalNotificationTarget`) with its own test asserting both directions: `APPROVAL_REQUESTED` keeps `Resource.APPROVAL` plus the approval UUID, the comment events resolve to the host. Only the persisted target was affected — OWNER resolution and enrichment were already subject-based before this PR, deliberately so for approvals.

src/main/java/com/otilm/core/events/handlers/CommentEventsHandler.java
klaraf755: Leaving this as it is, deliberately. Recipients are the thread's participants as they are when the event is handled — the same current-state semantics the trigger evaluation uses, which was the conclusion on the `COMMENT_RESOLVED_AT` thread in the previous review round.

The consequence is over-notification, not disclosure: somebody who joined the thread just before an older event is handled may be told about a slightly earlier comment in that same thread. Since 4a5935e45 every participant is authorized against the host before being notified, so that person demonstrably still has read access to the object — they simply learn about one comment earlier than their own.

A publish-time snapshot would be stale in the other direction (a participant who joins between publish and delivery gets nothing, even though the thread is now theirs) and would put a recipient list in the event payload, which the handler-driven path deliberately keeps out of it.

lubomirw (approved):

PR #2119, opened 2026-08-20
Pin @ExternalAuthorization, clear deprecations, and fix denial auditing

### LINKED ISSUE #2015: Spring Security 6.5 deprecation burn-down and @ExternalAuthorization regression suite
### Description

Burn down the Spring Security 6.5 deprecations while `core` is still on Spring Boot 3.5, and
build a regression suite around the `@ExternalAuthorization` interceptor stack.

Doing this first turns the Spring Security 7 step in the migration into a version bump instead
of a bump plus a redesign. Deprecated APIs on 6.5 are the ones removed in 7, so every one left
standing becomes a compile error inside an already-large migration diff, at the exact moment
there is no green baseline to compare against.

51 files in `src/main/java` import `org.springframework.security`. `SecurityConfig` is 229 lines
and holds the only `SecurityFilterChain`.

The regression suite is the more important half. Authentication and authorization failures do
not surface in unit tests, and the migration issue for Security 7 has no way to prove it did not
break anything without a suite that exercises the interceptor stack end to end. Building it here,
against known-good 3.5 behaviour, is what makes the later step verifiable at all.

Part of the preparation list on #1941.

### Definition of Done

- [ ] Every Spring Security deprecation warning in the `core` build is resolved or recorded in
      this issue with the reason it must stay.
- [ ] A regression suite covers the `@ExternalAuthorization` interceptor stack: permitted access,
      denied access, missing identity, and the OPA-resolution path.
- [ ] The suite reuses an existing test-context combination from `com/otilm/core/util/mockbeans/`;
      `ContextSignatureGuardTest.BASELINE` is unchanged, or raised in this PR with justification.
- [ ] Authorization behaviour on 3.5 is demonstrably unchanged by the burn-down.
- [ ] `mvn -B verify` passes; Spotless / Checkstyle gates pass; Sonar clean and coverage
      >= 80 % on changed lines.

ivosh:
**Closes #2015. This pins the `@ExternalAuthorization` stack with a regression suite, clears every Spring Security deprecation in that stack, and fixes one real bug the suite found: a denial with no recorded permission destroyed the audit record and the denial with it.**

Three strands of work, all in one commit because the tests are what made the other two safe.

## 1. The bug

`AuditLogAspect` names the denied permission in its audit message. To do that it reads a resource/action pair that the authorization layer records on the request.

**Not every denial records that pair.** Only `ExternalAuthorizationCore` writes it. An object-level OPA failure and a rejected token type both arrive with nothing recorded.

**When it's fine:** a resource-level denial. `ExternalAuthorizationCore.decideBasedOnOpaResult` records the pair before it returns one — including when the OPA call itself failed, because `checkAccess` swallows the exception and returns `unauthorized()`. Every one of these paths has a pair to read.

**When it breaks:** an object-level OPA failure. `AuthHelper.loadObjectPermissions` has no try/catch, so the `AccessDeniedException` propagates with nothing recorded. `Resource.findByCode(null)` then throws from inside the aspect's own catch block — the block that was about to re-raise the denial. The denial is destroyed on its way out. The `finally` block still runs and writes an audit record with no operation result, so the insert fails too.

The fix is a null guard, in an extracted `appendDeniedPermission` helper.

## 2. The deprecations

| API | On 6.5.11 (today) | On 7.1.0 (target) | What we did |
|---|---|---|---|
| `AuthorizationManager.check` | abstract **and** deprecated | gone | kept, delegates to `authorize` |
| `AuthorizationManager.authorize` | default method | the only method | now holds the decision |
| `ConfigAttribute` | deprecated | gone | removed |
| `security.access.AccessDeniedException` | not deprecated | still there | left alone |

**`check` is both abstract and deprecated on 6.5.11.** That is why this cannot be finished today. We are required to implement it and warned for doing so. The decision logic moved to `authorize`, which survives; `check` is now `@Deprecated(forRemoval = true)` and does nothing but delegate.

**One thing will still break at 7.1.0.** The supplier widens from `Supplier<Authentication>` to `Supplier<? extends Authentication>`, so our override stops being an override. It is one parameter type in one class, and the note is in the Javadoc on `check` so the upgrade finds both changes together.

`AccessDeniedException` is **not** deprecated and still exists in 7.1.0. We import it in 22 places. It is easy to assume the whole `security.access` package dies with `ConfigAttribute`; it does not.

## 3. The regression suite

The suite pins what the annotation stack actually sends to OPA: resource, action, object UUIDs, and parent-before-child ordering. A denied parent must short-circuit the child check, because evaluating the child leaks whether it would have permitted the action.

The HTTP tests double as a guard on request scope. The test schema has no Spring Session tables, so if the denied pair ever moves back to session scope, recording it creates a session and those tests fail on the `core.spring_session` insert. Add the DDL to make that failure go away and the guard is gone with it.

## Verification

- Full suite: **5208 tests, 0 failures**.
- New-lines coverage: **94.1%**, against the 80% gate.
- Fix 1 is mutation-checked. With the null guard removed the new test fails, so it discriminates rather than just passing.
- `ContextSignatureGuardTest` BASELINE is unchanged at 58 — no new Spring context was purchased.

Two full runs each hit a flake that is not from this branch. `DiscoveryServiceMessagingITest` timed out connecting to the broker; `SigningProfileServiceImplITest` failed on the connector-mock port collision, and **the failing test moved between two otherwise identical runs**. Both classes pass in isolation on this commit.

## Worth knowing for the rest of the migration

Fifteen test call sites moved from `check` to `authorize` by hand. None of them produced a deprecation warning, because calling a non-deprecated override does not warn. `-Xlint:deprecation` cannot see them, so a compiler-driven sweep would have missed every one. **Warning counts are not a complete inventory.**

### DIFF
--- a/src/main/java/com/otilm/core/aop/AuditLogAspect.java
+++ b/src/main/java/com/otilm/core/aop/AuditLogAspect.java
@@ -163,14 +163,8 @@ public class AuditLogAspect {
     }
 
     /**
-     * Names the denied permission in the audit message, when the authorization layer recorded one.
-     *
-     * <p>
-     * Not every {@link AccessDeniedException} carries a denied resource/action pair — only
-     * {@code ExternalAuthorizationCore} records it, so an object-level OPA failure or a rejected token type arrives
-     * with nothing recorded. {@code Resource.findByCode(null)} throws, and this runs inside the catch block that
-     * re-raises the denial: without the guard the denial is destroyed on its way out and the audit record is built with
-     * no operation result at all.
+     * Appends the recorded denied permission to the audit message, or returns the original message when no permission
+     * was recorded.
      */
     private static String appendDeniedPermission(String message) {
         String resourceName = AuthHelper.getDeniedPermissionResource();
--- a/src/main/java/com/otilm/core/aop/AuditLogAspect.java
+++ b/src/main/java/com/otilm/core/aop/AuditLogAspect.java
@@ -139,11 +139,7 @@ public class AuditLogAspect {
         } catch (Exception e) {
             String message = e.getMessage();
             if (e instanceof AccessDeniedException) {
-                String resourceNameAccessDenied = AuthHelper.getDeniedPermissionResource();
-                String resourceActionName = AuthHelper.getDeniedPermissionResourceAction();
-                message = "%s. Required '%s' action permission for resource '%s'"
-                        .formatted(message, BeautificationUtil.camelToHumanForm(resourceActionName),
-                                Resource.findByCode(resourceNameAccessDenied).getLabel());
+                message = appendDeniedPermission(message);
             }
 
             logBuilder.operationResult(OperationResult.FAILURE);
@@ -164,6 +160,27 @@ public class AuditLogAspect {
             logBuilder.timestamp(OffsetDateTime.now());
             auditLogsProducer.produceMessage(new AuditLogMessage(logBuilder.build(), output));
         }
+    }
+
+    /**
+     * Names the denied permission in the audit message, when the authorization layer recorded one.
+     *
+     * <p>
+     * Not every {@link AccessDeniedException} carries a denied resource/action pair — only
+     * {@code ExternalAuthorizationCore} records it, so an object-level OPA failure or a rejected token type arrives
+     * with nothing recorded. {@code Resource.findByCode(null)} throws, and this runs inside the catch block that
+     * re-raises the denial: without the guard the denial is destroyed on its way out and the audit record is built with
+     * no operation result at all.
+     */
+    private static String appendDeniedPermission(String message) {
+        String resourceName = AuthHelper.getDeniedPermissionResource();
+        String resourceActionName = AuthHelper.getDeniedPermissionResourceAction();
+        if (resourceName == null || resourceActionName == null) {
+            return message;
+        }
+        return "%s. Required '%s' action permission for resource '%s'"
+                .formatted(message, BeautificationUtil.camelToHumanForm(resourceActionName),
+                        Resource.findByCode(resourceName).getLabel());
     }
 
     /**
--- a/src/main/java/com/otilm/core/security/authz/ExternalAuthorizationConfigAttribute.java
+++ b/src/main/java/com/otilm/core/security/authz/ExternalAuthorizationConfigAttribute.java
@@ -1,16 +1,23 @@
 package com.otilm.core.security.authz;
 
-import org.springframework.security.access.ConfigAttribute;
+/**
+ * One resolved property of an {@code @ExternalAuthorization} annotation on its way to the OPA request.
+ *
+ * <p>
+ * Deliberately not a Spring Security {@code ConfigAttribute}: nothing in the framework consumes these, and that
+ * interface is removed in Spring Security 7.
+ */
+public record ExternalAuthorizationConfigAttribute(String attributeName, Object attributeValue) {
 
-public record ExternalAuthorizationConfigAttribute(String attributeName,
-        Object attributeValue) implements ConfigAttribute {
-
-    @Override
-    public String getAttribute() {
+    /**
+     * Renders the attribute for the trace logs in {@code OpaSecuredAnnotationMetadataExtractor}, which reference it as
+     * a method reference rather than calling it directly.
+     */
+    public String describe() {
         return "%s=%s".formatted(this.attributeName, this.attributeValue);
     }
 
-    public String getAttributeValueAsString() {
+    public String attributeValueAsString() {
         return attributeValue.toString();
     }
 }
--- a/src/main/java/com/otilm/core/security/authz/ExternalAuthorizationConfigAttribute.java
+++ b/src/main/java/com/otilm/core/security/authz/ExternalAuthorizationConfigAttribute.java
@@ -9,10 +9,7 @@ package com.otilm.core.security.authz;
  */
 public record ExternalAuthorizationConfigAttribute(String attributeName, Object attributeValue) {
 
-    /**
-     * Renders the attribute for the trace logs in {@code OpaSecuredAnnotationMetadataExtractor}, which reference it as
-     * a method reference rather than calling it directly.
-     */
+    /** Renders the attribute for trace logging. */
     public String describe() {
         return "%s=%s".formatted(this.attributeName, this.attributeValue);
     }
--- a/src/test/java/com/otilm/core/integration/aop/AuditLogAspectITest.java
+++ b/src/test/java/com/otilm/core/integration/aop/AuditLogAspectITest.java
@@ -23,7 +23,9 @@ import com.otilm.core.dao.repository.AuditLogRepository;
 import com.otilm.core.logging.LoggingHelper;
 import com.otilm.core.messaging.jms.listeners.AuditLogsListener;
 import com.otilm.core.messaging.model.AuditLogMessage;
+import com.otilm.core.model.auth.ResourceAction;
 import com.otilm.core.service.SettingExternalService;
+import com.otilm.core.util.AuthHelper;
 import com.otilm.core.util.BaseSpringBootTest;
 import java.util.List;
 import java.util.Optional;
@@ -35,6 +37,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.TestConfiguration;
 import org.springframework.context.annotation.Bean;
 import org.springframework.mock.web.MockHttpServletRequest;
+import org.springframework.security.access.AccessDeniedException;
 import org.springframework.web.context.request.RequestContextHolder;
 import org.springframework.web.context.request.ServletRequestAttributes;
 
@@ -64,6 +67,27 @@ class AuditLogAspectITest extends BaseSpringBootTest {
         @AuditLogged(module = Module.PROTOCOLS, resource = Resource.SIGNING_RECORD, operation = Operation.SIGN)
         public void performAndSucceed() {
             // no override — aspect must record SUCCESS
+        }
+
+        /**
+         * Denies without recording a denied resource/action pair, the way an OPA transport failure or a token type the
+         * authorization manager does not recognise does.
+         */
+        @AuditLogged(module = Module.PROTOCOLS, resource = Resource.SIGNING_RECORD, operation = Operation.SIGN)
+        public void denyWithoutRecordingPermission() {
+            throw new AccessDeniedException("An error occurred when calling OPA.");
+        }
+
+        /**
+         * Denies after recording a pair, the way {@code ExternalAuthorizationCore} does for every resource-level
+         * decision it returns.
+         */
+        @AuditLogged(module = Module.PROTOCOLS, resource = Resource.SIGNING_RECORD, operation = Operation.SIGN)
+        public void denyRecordingPermission() {
+            AuthHelper
+                    .setDeniedPermissionResourceAction(Resource.SIGNING_RECORD.getCode(),
+                            ResourceAction.LIST.getCode());
+            throw new AccessDeniedException("Access Denied");
         }
     }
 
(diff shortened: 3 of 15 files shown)


src/test/java/com/otilm/core/integration/security/authz/ExternalAuthorizationStaticITest.java
```diff
@@ -0,0 +1,177 @@
...
+     */
+    @Test
+    void deniesWhenAuthenticationIsNotAPlatformToken() {
+        SecurityContextHolder
+                .getContext()
+                .setAuthentication(new UsernamePasswordAuthenticationToken("someone", "secret", List.of()));
+
+        assertThatThrownBy(() -> signingRecordService.getSearchableFieldInformation())
+                .isInstanceOf(AuthorizationDeniedException.class);
+        verify(opaClient, never()).checkResourceAccess(any(), any(), any(), any());
```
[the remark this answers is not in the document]
ivosh: Fixed in 19e880f8. Added `deniesWhenThereIsNoAuthentication`, which clears the security context rather than substituting a wrong token. The behaviour it pins is `AuthenticationCredentialsNotFoundException` from Spring's method interceptor — the denial happens before `AbstractExternalAuthorizationManager` is consulted at all — and OPA is never called. That closes the "missing identity" item in #2015's definition of done.

src/main/java/com/otilm/core/aop/AuditLogAspect.java
```diff
@@ -166,6 +162,27 @@ public Object log(ProceedingJoinPoint joinPoint) throws Throwable {
...
+     * with nothing recorded. {@code Resource.findByCode(null)} throws, and this runs inside the catch block that
+     * re-raises the denial: without the guard the denial is destroyed on its way out and the audit record is built with
+     * no operation result at all.
+     */
+    private static String appendDeniedPermission(String message) {
+        String resourceName = AuthHelper.getDeniedPermissionResource();
+        String resourceActionName = AuthHelper.getDeniedPermissionResourceAction();
+        if (resourceName == null || resourceActionName == null) {
+            return message;
+        }
```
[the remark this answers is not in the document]
ivosh: Not fixing here: the staleness is pre-existing and unchanged by this branch. On `origin/main`, `decide()` is identical — `check()` records the denied pair before `checkGroupOwnerAssociations` runs, that fallback can grant, and nothing clears the pair; main's `AuthHelper` has no clear method either. What this branch changed is the pair's scope, session to request, which narrows the staleness window rather than opening it: previously a stale pair survived across every request on the session.

The mechanism you describe is real, and the fix you name is the right one — clear the pair whenever the final decision grants. It belongs in `ExternalAuthorizationCore`, which this PR does not touch, so it is being tracked separately rather than widened into this diff.

src/test/java/com/otilm/core/integration/api/web/ExternalAuthorizationHttpITest.java
```diff
@@ -0,0 +1,110 @@
...
+     */
+    @Test
+    void returnsForbiddenWhenTheDenialRecordedNoPermissionPair() throws Exception {
+        when(opaClient.checkObjectAccess(any(), any(), any(), any()))
+                .thenThrow(new AccessDeniedException("An error occurred when calling OPA."));
+
+        mockMvc
+                .perform(get(STATISTICS_ENDPOINT))
+                .andExpectAll(status().isForbidden(), jsonPath("$.code").value("ACCESS_DENIED"),
+                        jsonPath("$.message").value(startsWith("Access denied for the specified operation")));
```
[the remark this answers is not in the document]
ivosh: Not fixing here: pre-existing. `ExceptionHandlingAdvice.java:406` is byte-identical on `origin/main` and this branch does not touch that file, so tightening the assertion to an exact match would pin behaviour this PR neither introduced nor owns.

The underlying point stands — appending a raw framework `getMessage()` to a 403 body can expose runtime detail, and our own guidance is to gate that on a controlled domain-exception type. Tracked separately, together with the exact-match assertion that should land with it.

lubomirw (commented):
Automated multi-reviewer triage - 9 findings (COMMENT). Copilot CLI, Guided PR review, and Superpowers found no correctness/security issues; these comments are hygiene maintainability items.

src/test/java/com/otilm/core/util/AuthHelperDeniedPermissionTest.java
```diff
@@ -0,0 +1,88 @@
...
+    void recordingADeniedPairCreatesNoSession() {
+        MockHttpServletRequest request = bindRequest();
+
+        AuthHelper.setDeniedPermissionResourceAction("certificates", "detail");
+
+        assertThat(request.getSession(false)).as("recording a denial must not create a session").isNull();
+    }
+
+    /**
+     * The regression this pins: with the pair held at session scope, a second request on the same session reads the
```
lubomirw: **[Low]** | Maintainability

The phrase "The regression this pins" records development history rather than the current constraint.

_Suggested:_ Consider rewriting the sentence to begin with the behavior itself, for example: "With the pair held at session scope, a second request..."

<sub>flagged by Hygiene</sub>

src/main/java/com/otilm/core/aop/AuditLogAspect.java
```diff
@@ -166,6 +162,27 @@ public Object log(ProceedingJoinPoint joinPoint) throws Throwable {
         }
     }
 
+    /**
```
lubomirw, line 165: **[Nit]** | Maintainability

The private helper Javadoc is detailed implementation narrative.

_Suggested:_ Author judgment: if you want stricter hygiene here, shorten it to the behavior contract, e.g. "Appends the recorded denied permission to the audit message, or returns the original message when no permission was recorded."

<sub>flagged by Hygiene</sub>

src/main/java/com/otilm/core/security/authz/ExternalAuthorizationConfigAttribute.java
```diff
@@ -1,16 +1,23 @@
...
+ * interface is removed in Spring Security 7.
+ */
+public record ExternalAuthorizationConfigAttribute(String attributeName, Object attributeValue) {
 
-public record ExternalAuthorizationConfigAttribute(String attributeName,
-        Object attributeValue) implements ConfigAttribute {
-
-    @Override
-    public String getAttribute() {
+    /**
```
lubomirw: **[Nit]** | Maintainability

The Javadoc explains the current method-reference call site rather than the method behavior.

_Suggested:_ Replace it with a concise behavior note such as: "/** Renders the attribute for trace logging. */"

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/integration/aop/AuditLogAspectITest.java
```diff
@@ -209,10 +233,49 @@ void auditResultOverride_auditsOperationAsSuccess_whenNoOverrideSet() {
         Assertions.assertEquals(OperationResult.SUCCESS, auditLogs.getFirst().getOperationResult());
     }
 
-    /**
-     * Binds a request scope around the action so the request-scoped {@link AuditResultOverride} resolves — production
-     * audited methods always run inside a DispatcherServlet request, but the test invokes the bean directly.
-     */
+    @Test
+    void deniedWithoutRecordedPermission_auditsFailureAndRethrowsTheDenial() {
+        // given
```
lubomirw, line 238: **[Nit]** | Maintainability

The added given/when/then labels around self-explanatory test code add noise.

_Suggested:_ Remove the labels at the flagged setup/assertion blocks if you want the tests to match the surrounding style.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/integration/api/web/ExternalAuthorizationHttpITest.java
```diff
@@ -0,0 +1,91 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.hamcrest.Matchers.startsWith;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+/**
```
lubomirw, line 21: **[Nit]** | Maintainability

The class-level Javadoc is long and mixes test-path explanation with why this endpoint was chosen.

_Suggested:_ Author judgment: tighten it to two sentences covering the real HTTP/method-security path, request-scope denial guard, and selected endpoint.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/integration/security/authz/ExternalAuthorizationStaticITest.java
```diff
@@ -0,0 +1,164 @@
...
+import org.springframework.security.authorization.AuthorizationDeniedException;
+import org.springframework.security.core.context.SecurityContextHolder;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+/**
```
lubomirw, line 27: **[Nit]** | Maintainability

The class-level Javadoc gives a detailed explanation of neighboring coverage and resource choice.

_Suggested:_ Author judgment: shorten it to the suite contract: static annotation requests, parent/child OPA ordering, and resources selected to avoid fallback authorization.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/integration/security/authz/ExternalAuthorizationStaticITest.java
```diff
@@ -0,0 +1,164 @@
...
+        SecurityContextHolder
+                .getContext()
+                .setAuthentication(new UsernamePasswordAuthenticationToken("someone", "secret", List.of()));
+
+        assertThatThrownBy(() -> signingRecordService.getSearchableFieldInformation())
+                .isInstanceOf(AuthorizationDeniedException.class);
+        verify(opaClient, never()).checkResourceAccess(any(), any(), any(), any());
+    }
+
+    /** Missing identity, as distinct from the wrong identity above: no authentication at all. */
```
lubomirw: **[Nit]** | Maintainability

The one-line Javadoc restates the test name and depends on comparison with the previous test.

_Suggested:_ Remove the redundant note.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/security/authz/AbstractExternalAuthorizationManagerTest.java
```diff
@@ -22,25 +25,50 @@ class AbstractExternalAuthorizationManagerTest {
 
     Authentication authentication = createPlatformAuthentication();
 
+    /**
```
lubomirw: **[Nit]** | Maintainability

The test Javadoc is long for behavior already expressed by the method name and parameterization.

_Suggested:_ Author judgment: shorten it to "Verifies deprecated check delegates to authorize for both grant and denial decisions," or keep it if the Spring Security 6.5 migration context is worth preserving.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/security/authz/AbstractExternalAuthorizationManagerTest.java
```diff
@@ -22,25 +25,50 @@ class AbstractExternalAuthorizationManagerTest {
...
+     *
+     * <p>
+     * Both outcomes are exercised because denial is what every other exit of {@code authorize} returns as well: pinned
+     * to deny alone, a {@code check} that stopped delegating and returned a constant denial would still pass.
+     */
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    @SuppressWarnings("removal")
+    void deprecatedCheckReturnsTheSameDecisionAsAuthorize(boolean granted) {
+        // given
```
lubomirw, line 32: **[Nit]** | Maintainability

The given/when/then labels do not add information.

_Suggested:_ Remove the labels.

<sub>flagged by Hygiene</sub>

src/test/java/com/otilm/core/util/AuthHelperDeniedPermissionTest.java
ivosh: Fixed in 7f08b3c0. The sentence now opens with the behaviour: "At session scope, a second request on the same session reads the first request's denial, so a 403 can name a permission that was never checked."

src/main/java/com/otilm/core/security/authz/ExternalAuthorizationConfigAttribute.java
ivosh: Fixed in 7f08b3c0. Now `/** Renders the attribute for trace logging. */`, verbatim from your suggestion — the method-reference clause is gone.

src/test/java/com/otilm/core/integration/security/authz/ExternalAuthorizationStaticITest.java
ivosh: Fixed in 7f08b3c0. Removed. It also depended on positional comparison with the test above it, which breaks the moment the tests are reordered.

src/main/java/com/otilm/core/aop/AuditLogAspect.java
ivosh: Fixed in 7f08b3c0, adopting your wording verbatim: "Appends the recorded denied permission to the audit message, or returns the original message when no permission was recorded."

The mechanism the old paragraph described is what makes the null guard undeletable, so it needed somewhere to live. It is pinned by a test rather than a comment now — `AuditLogAspectITest.deniedWithoutRecordedPermission_auditsFailureAndRethrowsTheDenial` fails with a `DataIntegrityViolationException` if the guard is removed, which is a stronger record than the prose was.

src/test/java/com/otilm/core/integration/api/web/ExternalAuthorizationHttpITest.java
ivosh: Fixed in 7f08b3c0. Down to two sentences covering the three things you named — the real HTTP/method-security path, the request-scope denial guard, and why this endpoint.

The one sentence I was reluctant to lose is the guard's failure mode, since it is the only place it was written down. It is now in the PR description: the denial tests fail on the `core.spring_session` insert if the pair returns to session scope, so adding the missing DDL to quiet that failure would silently remove the guard.

src/test/java/com/otilm/core/integration/security/authz/ExternalAuthorizationStaticITest.java
ivosh: Fixed in 7f08b3c0. Two sentences: the suite contract with parent-before-child ordering, then the resource selection that keeps the group/owner fallback out of the way. The sibling-suite paragraph is gone.

src/test/java/com/otilm/core/security/authz/AbstractExternalAuthorizationManagerTest.java
ivosh: **Fixed in 7f08b3c0, but removed rather than shortened.**

I applied your one-liner first. A hygiene pass over the result then flagged it as redundant against the signature, and it was right: the method name carries "check returns the same decision as authorize", `@ValueSource(booleans = {true, false})` carries "both grant and denial", and `@Deprecated` on the production `check` carries "deprecated". Nothing in the sentence was unavailable from the code, so the Javadoc is gone entirely.

The Spring Security 6.5 migration context you offered to preserve is in the PR description instead — `check` is deprecated *and* still abstract on 6.5, which is why a deprecated method has a test at all.

Leaving this thread open since that is not the change you asked for. Say the word and I will put your one-liner back.

src/test/java/com/otilm/core/integration/aop/AuditLogAspectITest.java
ivosh: **Not fixing: the surrounding style already has these labels.**

Your suggestion is conditional on matching the file's style, and the premise runs the other way here. `origin/main` already carries 6 given/when/then labels in this class, and every pre-existing test uses them. Removing them from the new tests would make those the only unlabelled tests in the file.

Happy to strip them class-wide in a separate change if you would rather the convention went the other way — that is a bigger call than this PR.

src/test/java/com/otilm/core/security/authz/AbstractExternalAuthorizationManagerTest.java
ivosh: **Not fixing: same as the labels thread on `AuditLogAspectITest`.**

`origin/main` carries 12 given/when/then labels in this class and every pre-existing test uses them, so dropping them only from the new test would make it the odd one out.

lubomirw (approved):

PR #2120, opened 2026-08-21
Configure and drive content signing to SIGNED and TIMESTAMPED

### LINKED ISSUE #1974: Content-signing workflow + level orchestrator

ivosh:
**TL;DR — a CONTENT_SIGNING profile can now be configured with a level ladder, and the engine drives a content signature to SIGNED or TIMESTAMPED through it, recording every step.**

Closes #1974, scoped to **B and B+T only**. `LONG_TERM` and `ARCHIVAL` are out of scope here.

## What this builds

A CONTENT_SIGNING Signing Profile gains four settings: `family`, `maxLevel`, `timestampSource` and `documentSizeCap`. Profile save validates them. The first three are mandatory for ILM-managed signing and must be absent for delegated signing. `documentSizeCap` is optional under either scheme.

The engine then levels a signature against that configuration. It calls the formatting connector to build the data to be signed, releases the key, and — when the ceiling is `TIMESTAMPED` — issues a signature timestamp in-process from the referenced TIMESTAMPING profile. Every step lands in a signing record with its timestamp serials.

Time quality rides on the timestamp source. A `maxLevel` of `SIGNED` must omit a timestamp source; `TIMESTAMPED` or higher must name one. Where a timestamp is issued, its time is gated by the referenced TIMESTAMPING profile's own Time Quality Configuration, which already runs inside issuance. **A content-signing profile never names a Time Quality Configuration directly, so the workflow carries no such field and the engine runs no gate of its own.**

Non-idempotent steps enter through an **acquisition seam** (`ContentSigningAcquisitions`). A randomized signature and a freshly issued timestamp token both go through it. That is the hook core#1973 needs to persist and replay them; the synchronous implementation here does neither yet.

## What this does NOT prove

The formatting connector is a mock, by necessity.
So this branch proves orchestration, key-release gating, timestamp bridging and recording. **It does not prove that a well-formed PAdES file comes out the other end.** That verification arrives with the real connector.

## Delegated content signing was unreachable before this branch

Worth knowing before reading the delegated changes. The `interfaces-ades` bean validator rejects a delegated content-signing profile that carries a formatting connector, and core has always *required* one. Every such request therefore hit a `400` at the HTTP boundary before core ever saw it.

That means **this branch reshapes a surface that never worked**, and the delegated path has never had end-to-end exercise in the field. Core now agrees with the validator: a delegated profile carries no formatting connector and no level ladder, because the signer connector does both itself.

## Accepted residuals

Each of these is a deliberate call, not an oversight.

- The delegated formatting-connector refusal runs **inside** `persistCreate`/`persistUpdate`, so the connector is contacted before the request is refused. Only a direct service caller reaches that ordering — the bean validator rejects it at the HTTP boundary first.
- No integration test pins `documentSizeCap` persistence on the **managed** path.
- `documentSizeCap` is **not enforced** on either scheme. The check belongs with the handler that accepts a signing request, which is not this branch. The interfaces wording is corrected on OmniTrustILM/interfaces#893 to promise enforcement only when an ILM-managed signing request arrives.
- A narrow save/delete race on a referenced timestamp-source profile yields a generic `500` rather than a shaped message.
- The delete-refusal message names the blocking profile without ACL scoping.
- A **failed run after the timestamp token is issued leaves an orphan timestamp record.** This is correct per the no-rollback-after-external-commitment rule — the TSA did the work — but it is worth stating plainly.
- `augment(...)` is deliberate, fully-tested dead code. Nothing calls it until `LONG_TERM` lands.
- `sign(...)` is unreached from `src/main` as well: nothing resolves a content-signing profile until the request handler lands (core#2117), so the whole ladder is exercised by tests only. That is what keeps several review findings latent rather than live.
- `computeDtbs` carries no `signatureParameters`. The wire DTO has the field; assembling and populating the object is core#2117's, and the acceptance criterion now lives there.
- No keyUsage/EKU purpose rule exists for `CONTENT_SIGNING`. That gap predates this branch and the code says so.
- Some schema drift predates this branch and is untouched by it.


src/main/java/com/otilm/core/signing/contentsigning/ManagedContentSigningEngine.java
```diff
@@ -0,0 +1,404 @@
...
+            throws SigningEngineException {
+        requireReachableLevel(request.targetLevel(), profile);
+        requireTrustworthyTime(profile);
+        requireAcceptableSigningCertificate(profile);
+        Instant signingTime = clockSource.wallTimeInstant();
+
+        ComputeDtbsResponseDto dtbs = computeDtbs(request, profile, signingTime);
+        // The key is released only once the connector has committed to the authorized document.
+        DtbsBindingVerifier.verify(request.authorizedDigest(), dtbs);
+        byte[] signatureValue = acquisitions.signatureValue(profile, dtbs.getDtbs());
```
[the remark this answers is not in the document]
ivosh: Fixed in 47950b2a.

src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -1032,8 +1115,12 @@ private List<ResponseAttribute> persistSignatureFormattingConnectorAttributes(Si
...
             throws ConnectorException, NotFoundException {
         return switch (workflow) {
-            case ContentSigningWorkflowRequestDto w ->
-                fetchFormattingAttributeDefinitions(w.getSignatureFormattingConnectorUuid());
+            case ContentSigningWorkflowRequestDto w -> w.getSignatureFormattingConnectorUuid() == null
+                    ? List.of()
+                    : contentSigningFormattingAttributes
+                            .aggregate(
+                                    connectorService.getConnectorForApiClient(w.getSignatureFormattingConnectorUuid()),
+                                    w.getMaxLevel());
```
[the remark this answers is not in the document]
ivosh: Fixed in 47950b2a.

src/test/java/com/otilm/core/integration/signing/contentsigning/ContentSigningEngineITest.java
```diff
@@ -0,0 +1,395 @@
...
+        assertThat(recordsFor(tsaProfile.getName()))
+                .anySatisfy(record -> assertThat(record.getName()).contains(serialHex));
+        assertThat(recordsFor(contentProfile.getName())).isNotEmpty();
+    }
+
+    @Test
+    void signsAPadesDocumentToSignedLevelAndLeavesARecord() throws Exception {
+        // given: a PAdES profile whose ceiling is SIGNED, which names no timestamp source
+        SigningProfileDto contentProfile = createContentSigningProfile("signed-run-pades", SignatureFamily.PADES,
+                SignatureLevel.SIGNED, null);
```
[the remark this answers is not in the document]
ivosh: Fixed in 47950b2a.

src/test/java/com/otilm/core/integration/signing/contentsigning/ContentSigningEngineITest.java
```diff
@@ -0,0 +1,395 @@
...
+            contentSigningFormattingMock.stop();
+        }
+    }
+
+    @Test
+    void signsAPadesDocumentToTimestampedLevelAndTracesToItsTimestampRecord() throws Exception {
+        // given: a PAdES profile whose ceiling is TIMESTAMPED, pointing at an ILM-managed TSA profile
+        SigningProfileDto tsaProfile = createTimestampingProfile("timestamped-run-tsa");
+        SigningProfileDto contentProfile = createContentSigningProfile("timestamped-run-pades", SignatureFamily.PADES,
+                SignatureLevel.TIMESTAMPED, UUID.fromString(tsaProfile.getUuid()));
```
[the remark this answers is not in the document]
ivosh: Fixed in 47950b2a.

src/main/java/com/otilm/core/signing/contentsigning/ManagedContentSigningEngine.java
```diff
@@ -0,0 +1,428 @@
...
+    private SignedContent augmentToTargetLevel(AugmentationRequest request, SigningProfileModel<?, ?> signingProfile,
+            ResolvedManagedContentSigningProfile profile, SigningProtocol protocol, List<BigInteger> serials)
+            throws SigningEngineException {
+        requireReachableLevel(request.targetLevel(), profile);
+        if (request.targetLevel() == SignatureLevel.SIGNED) {
+            throw new SigningEngineException(SigningEngineFailure.INVALID_INPUT,
+                    "augmentation to SIGNED was requested, but the document already carries a signature",
+                    "The document is already signed; augmentation needs a level above SIGNED");
+        }
+        requireTrustworthyTime(profile);
```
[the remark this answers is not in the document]
ivosh: Fixed in 8b81019c.

The gate is gone from the augmentation path; `sign` keeps it. Your reading of the split is the one the design intends: the local clock gates where Core mints its own `signingTime` claim, and augmentation mints none, so the only authoritative time on that path is the referenced TIMESTAMPING profile's — which `ManagedTimestampEngine` gates itself.

One correction to the scenario, for the record: a profile with no explicit Time Quality Configuration resolves to `LocalClockTimeQualityConfiguration`, which `TimeQualityRegisterImpl` answers `OK` unconditionally. So the refusal you describe could only ever have hit an operator who had attached an explicit configuration. Real, but narrower than it looks.

Added `augmentsWhileTheLocalClockIsDegradedBecauseTheTimestampsTimeIsTheTsasOwn` so the decision is pinned rather than resting on the absence of a line.

src/main/java/com/otilm/core/signing/contentsigning/ContentSigningRecordFactory.java
```diff
@@ -0,0 +1,81 @@
...
+
+    /** The serials are what make an operation traceable to the timestamp records issued for it. */
+    private String buildRequestMetadataJson(SigningProfileModel<?, ?> signingProfile,
+            ResolvedManagedContentSigningProfile profile, SignedContent result) {
+        Map<String, Object> metadata = new LinkedHashMap<>();
+        metadata.put("signingProfileName", signingProfile.name());
+        metadata.put("signingProfileVersion", signingProfile.version());
+        metadata.put("family", profile.family().name());
+        metadata.put("signatureLevel", result.level().name());
+        metadata.put("timestampTokenSerials", hexSerials(result.timestampSerials()));
```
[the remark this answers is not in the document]
ivosh: Fixed in 4238364f.

Confirmed on every point — `SigningRecordInputMapper` writes `request_metadata_json` only when `recordRequestMetadata` is on, and `SigningProfileVersion` defaults it off, so the trace that #1974's criterion mandates was absent on a default profile. The ITest only passed because it configures `recordingEverything()`, and its assertion read the TSA record's name rather than the content record.

Fixed with a dedicated column rather than by defeating the toggle: `timestamp_token_serials TEXT[]` on `signing_record` and `signing_record_outbox` (migration `V202608221000`), written unconditionally next to the requester identity, which is ungated for the same reason. GIN-indexed, so a serial lookup is a real query rather than a string parse.

`TimestampSigningRecordFactory` populates the same column with its own serial, so the join works from both ends — otherwise you would be joining the content record's array against a display name or against metadata that is itself toggle-gated.

Not on `SigningRecordDto` / `SigningRecordListDto` yet; those live in the `interfaces` repo and need their own change. The ITest asserts the join on the entity in the meantime.

src/main/java/com/otilm/core/signing/contentsigning/profile/ContentSigningWorkflowValidator.java
```diff
@@ -0,0 +1,156 @@
...
+        requireUsableTimestampingProfile(referenced);
+    }
+
+    /**
+     * Only an ILM-managed TIMESTAMPING profile can issue these timestamps, and it must record what it issues --
+     * otherwise a signature's timestamp traces to nothing.
+     */
+    private void requireUsableTimestampingProfile(UUID referenced) {
+        SigningProfileVersion version = versionRepository
+                .findLatestByProfileUuid(referenced)
```
lubomirw, line 155: **[High]** | Correctness

`validate(w, connector)` takes no profile UUID, so it structurally cannot notice that a profile names itself. `requireUsableTimestampingProfile` reads `findLatestByProfileUuid`, which during an edit is still the *stale* version — TIMESTAMPING, MANAGED, above the floor — so editing a TIMESTAMPING profile into a CONTENT_SIGNING one that names itself passes.

The damage lands on the second edit: once that self-referencing version is superseded, `timestampSourceBlockers` reports it under the superseded branch, whose stated remedy is "released only by deleting the Signing Profile" — the profile now blocks its own deletion, permanently.

Two steps and an odd one, so unlikely — but the end state has no way out.

_Suggested:_ Pass the target profile UUID into `validate` (null on create) and reject `referenced.equals(targetUuid)` outright. Validating against the version being written rather than the stored latest version would close the stale-read half too.

<sub>flagged by Copilot CLI; validator signature and blocker branch verified at 2430572</sub>

src/main/java/com/otilm/core/signing/contentsigning/ManagedContentSigningEngine.java
```diff
@@ -0,0 +1,408 @@
...
+            return signToTargetLevel(request, signingProfile, profile, protocol, serials);
+        } catch (Exception e) {
+            throw loggedFailure(profile, request.targetLevel(), serials, e);
+        }
+    }
+
+    private SignedContent signToTargetLevel(ContentSigningRequest request, SigningProfileModel<?, ?> signingProfile,
+            ResolvedManagedContentSigningProfile profile, SigningProtocol protocol, List<BigInteger> serials)
+            throws SigningEngineException {
+        requireReachableLevel(request.targetLevel(), profile);
```
lubomirw, line 92: **[Medium]** | Correctness

`documentSizeCap` is plumbed through the migration, entity, workflow model, mapper and `ResolvedManagedContentSigningProfile`, but `documentSizeCap()` is read by nothing outside two tests.

The residuals list covers the non-enforcement. What it does not cover is that the published contract already asserts the opposite: `ContentSigningWorkflowDto:54` and `ContentSigningWorkflowRequestDto:56` on interfaces **main** both read "Largest document accepted for signing, in bytes. Enforced when a signing request arrives, under both ILM-managed and delegated signing." So an operator sets a cap, the API says it is enforced, and nothing enforces it.

_Suggested:_ Either correct the interfaces wording now (say the cap is recorded and will be enforced at request admission), or gate on it here beside `requireReachableLevel`. Worth deciding which, since the cap is currently the only bound on document size and the engine holds several full copies of the document.

<sub>flagged by superpowers review and Copilot CLI; schema text confirmed on interfaces origin/main</sub>

src/main/java/com/otilm/core/signing/contentsigning/profile/ContentSigningWorkflowValidator.java
```diff
@@ -0,0 +1,156 @@
...
+                .orElseThrow(() -> new ValidationException(
+                        "The Signing Profile named as the timestamp source does not exist: " + referenced));
+        if (version.getWorkflowType() != SigningWorkflowType.TIMESTAMPING) {
+            throw new ValidationException(
+                    "The Signing Profile named as the timestamp source uses the %s workflow; a TIMESTAMPING profile is required"
+                            .formatted(version.getWorkflowType()));
+        }
+        if (version.getSigningScheme() != SigningScheme.MANAGED) {
```
lubomirw, line 163: **[Medium]** | Correctness

`requireUsableTimestampingProfile` checks workflow type, `SigningScheme.MANAGED` and the record floor, but never `managedSigningType`. `ManagedSigningType` has both `STATIC_KEY` and `ONE_TIME_KEY`, and `ManagedSchemeResolver:46` handles only `StaticKeyManagedSigning` — anything else throws "uses an unsupported signing scheme".

So a MANAGED/ONE_TIME_KEY TIMESTAMPING profile is accepted as a timestamp source and then cannot be resolved, which is exactly what this class's own javadoc promises cannot happen: "so a profile can never advertise a family, level or timestamp source that signing would later refuse".

_Suggested:_ Add a `managedSigningType != STATIC_KEY` refusal after the `MANAGED` check, naming static-key as the requirement, and cover it alongside the existing not-timestamping / not-managed / below-floor cases in `ContentSigningWorkflowValidatorTest`.

<sub>flagged by Copilot CLI; ManagedSigningType constants and ManagedSchemeResolver verified</sub>

src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -866,6 +906,77 @@ private void applyWorkflow(SigningProfile p, SigningProfileVersion version, Work
...
 
+    /**
+     * Runs outside the persistence transaction, so a request the contract already refuses never reaches a connector
+     * round-trip: {@code fetchFormattingAttributeDefinitions} calls the named provider straight after this.
+     */
+    private void validateContentSigningWorkflow(SigningProfileRequestDto request) throws NotFoundException {
+        if (!(request.getWorkflow() instanceof ContentSigningWorkflowRequestDto w)) {
+            return;
+        }
+        if (request.getSigningScheme().getSigningScheme() != SigningScheme.MANAGED) {
```
lubomirw, line 919: **[Medium]** | Correctness

Same root cause on the profile's own scheme: this branches only on `SigningScheme.MANAGED`, and `StaticKeyManagedContentSigningResolver.supports()` keys on the workflow type alone. A MANAGED/ONE_TIME_KEY CONTENT_SIGNING profile therefore saves cleanly and then fails in `ManagedSchemeResolver` on every request. Latent while no request path reaches the engine, but the profile is savable today.

_Suggested:_ Refuse a non-`STATIC_KEY` `managedSigningType` for managed content signing here, or narrow `StaticKeyManagedContentSigningResolver.supports()` to require `StaticKeyManagedSigning` as well.

src/main/java/com/otilm/core/signing/contentsigning/resolver/StaticKeyManagedContentSigningResolver.java
```diff
@@ -40,10 +44,14 @@ public ResolvedManagedContentSigningProfile resolve(SigningProfileModel<?, ?> mo
         ManagedContentSigningWorkflow workflow = (ManagedContentSigningWorkflow) model.workflow();
         ResolvedManagedScheme resolvedScheme = schemeResolver.resolve(model.name(), model.signingScheme());
         ApiClientConnectorInfo connector = resolveFormattingConnector(workflow.signatureFormattingConnectorUuid());
+        String timestampSourceProfileName = workflow.maxLevel() == SignatureLevel.SIGNED
```
lubomirw, line 47: **[Medium]** | Correctness

Managed CONTENT_SIGNING profiles were creatable before this branch (`SigningProfileServiceImpl` on main required only a formatting connector), and the migration adds `signature_family` / `max_signature_level` nullable with no backfill.

`maxLevel() == SIGNED ? null : profileNameFor(...)` treats a **null** maxLevel as "above SIGNED", so such a row lands in the else branch and fails with `TimestampSourceResolver:26` — "the Signing Profile names no timestamp source" — which names the wrong defect entirely. The same profile also can't be edited back into shape: a GET returns `family: null` and `requireFamily` then rejects the PUT.

Copilot adds the delegated side: those rows keep a stale `signature_formatting_connector_uuid`, which now 422s when a client re-sends it.

_Suggested:_ Backfill `max_signature_level = 'SIGNED'` for existing managed rows in the migration, or refuse a null maxLevel explicitly here with a message that names the real cause ("predates the level ladder; re-save with a family and maxLevel"). Either way the upgrade path needs a decision.

<sub>flagged by guided review, superpowers review and Copilot CLI; failure message verified</sub>

src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -1032,8 +1153,12 @@ private List<ResponseAttribute> persistSignatureFormattingConnectorAttributes(Si
     private List<BaseAttribute> fetchFormattingAttributeDefinitions(WorkflowRequestDto workflow)
             throws ConnectorException, NotFoundException {
         return switch (workflow) {
-            case ContentSigningWorkflowRequestDto w ->
-                fetchFormattingAttributeDefinitions(w.getSignatureFormattingConnectorUuid());
+            case ContentSigningWorkflowRequestDto w -> w.getSignatureFormattingConnectorUuid() == null
+                    ? List.of()
+                    : contentSigningFormattingAttributes
```
lubomirw, line 1173: **[Medium]** | Correctness

This is the fix that makes the new interfaces wording true — content signing now aggregates per-operation attributes instead of calling `listFormattingAttributes`. The problem is that the client-facing endpoint did not move with it: `listSignatureFormattingConnectorAttributes` (line 278) still calls `fetchFormattingAttributeDefinitions(connectorUuid)`, i.e. the old `SignatureFormatting` route.

Before this branch both sides used the same route, so they agreed by construction. Now a client is served one schema while save validates against another, and `fe-administrator` does call it (`SigningProfileForm.tsx:452`). Distinct from the missing-route gap tracked as core#2124 — this endpoint returns an actively wrong answer rather than no answer.

_Suggested:_ Route the listing through `ContentSigningFormattingAttributes.aggregate` for CONTENT_SIGNING (which means taking a maxLevel on the endpoint), or make it fail loudly rather than serve a schema save will reject.

<sub>flagged by guided review; divergence and the FE call site verified</sub>

src/main/java/com/otilm/core/signing/contentsigning/formatting/ContentSigningFormattingAttributes.java
```diff
@@ -0,0 +1,104 @@
...
+            ContentSigningFormattingOperation operation) {
+        BaseAttribute existing = merged.putIfAbsent(declared.getName(), declared);
+        if (existing != null && !sameDefinition(existing, declared)) {
+            throw new ValidationException(
+                    "Signature Formatting Provider declares formatting attribute '%s' differently for operation '%s' than for an earlier operation; a shared name must carry one shared definition"
+                            .formatted(declared.getName(), operation.getCode()));
+        }
+    }
+
+    private static boolean sameDefinition(BaseAttribute existing, BaseAttribute declared) {
```
lubomirw, line 83: **[Low]** | Error handling

`CANONICAL_FORM` is `ObjectMapperFactory.storage()`, a bare `new ObjectMapper()` with no JSR-310 modules — and `jsonColumn()`'s own javadoc in that same factory says "without them Jackson rejects every Java 8 date outright". `DateTimeAttributeContentV2` holds a `ZonedDateTime`, so a formatting attribute carrying date/time default content can fail canonicalization here.

The catch then returns `false`, and `mergeOne` reports the connector "declares formatting attribute X differently for operation Y" — blaming the connector for our own serialization fault, and blocking profile save with a diagnosis that sends the operator to the wrong vendor.

_Suggested:_ Register the discovered modules on `CANONICAL_FORM` the way `jsonColumn()` does, and on `JsonProcessingException` throw a platform error instead of returning `false`.

<sub>flagged by guided review; DateTimeAttributeContentV2 confirmed to carry ZonedDateTime</sub>

src/main/java/com/otilm/core/signing/contentsigning/ContentSigningRecordFactory.java
```diff
@@ -0,0 +1,80 @@
...
+        try {
+            return objectMapper.writeValueAsString(metadata);
+        } catch (JsonProcessingException e) {
+            throw new IllegalStateException("Failed to serialize content-signing record request metadata", e);
+        }
+    }
+
+    private static List<String> hexSerials(List<BigInteger> serials) {
```
lubomirw: **[Low]** | Maintainability

This hex rendering is the only join between a content-signing record's `timestamp_token_serials` and the timestamp records it points at — the join interfaces#893 publishes to API consumers. It is written independently in three places: here, `ManagedContentSigningEngine:375`, and inline at `TimestampSigningRecordFactory:55`.

Zero-padding, upper-casing or fixing the width in one of the three breaks traceability silently, with nothing to compile against and no test to fail.

_Suggested:_ Extract one formatter beside the record model (e.g. in `com.otilm.core.signing.record`) and have all three call sites use it, so the two sides of the join cannot drift.

<sub>flagged by superpowers review and Copilot CLI; all three copies verified</sub>

src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -554,6 +561,31 @@ void deleteInOwnTransaction(SigningProfile profile) throws ValidationException {
...
+        supersededOnlyNames.removeAll(latestVersionNames);
+
+        List<String> blockers = new ArrayList<>();
+        if (!latestVersionNames.isEmpty()) {
+            blockers
+                    .add("it is named as the timestamp source by Signing Profile(s) (repoint them, or lower their maxLevel to SIGNED, to release it): "
+                            + String.join(", ", latestVersionNames));
+        }
+        if (!supersededOnlyNames.isEmpty()) {
+            blockers
```
lubomirw, line 583: **[Low]** | Documentation

The superseded-only branch tells the operator the hold is "released only by deleting the Signing Profile", but `deleteSigningProfile` also refuses any profile that has signing records — so following the instruction hits a second wall the message never mentions. It also doesn't say *which* profile it means, referencing or referenced.

The latest-version branch just above is the good example by contrast: it names the two real remedies ("repoint them, or lower their maxLevel to SIGNED").

_Suggested:_ Spell the full remedy out and name the referencing profile, e.g. "released only by deleting the referencing Signing Profile '<name>', which first requires deleting its signing records".

<sub>flagged by superpowers review; the second refusal verified in deleteSigningProfile</sub>

src/main/java/com/otilm/core/signing/contentsigning/ManagedContentSigningEngine.java
```diff
@@ -0,0 +1,408 @@
...
+        this.signingRecordStrategyFactory = signingRecordStrategyFactory;
+        this.recordFactory = recordFactory;
+    }
+
+    /**
+     * Signs {@code request}'s document to its target level.
+     *
+     * @param protocol how the run was initiated, which is what its signing record will carry
+     */
+    public SignedContent sign(ContentSigningRequest request, SigningProfileModel<?, ?> signingProfile,
```
lubomirw, line 79: **[Low]** | Documentation

Nothing under `src/main` references `ManagedContentSigningEngine` — `sign()` is reached only from `ManagedContentSigningEngineTest` and `ContentSigningEngineITest` — and no production path resolves a content-signing profile either: `signingProfileResolverFactory.resolve` is called only from `TsaServiceImpl` and `InternalTimestampSource`.

The residuals list flags `augment(...)` as deliberate dead code but says nothing about `sign()`, while the TL;DR reads "the engine drives a content signature to SIGNED or TIMESTAMPED". That difference matters to a reviewer: it is why none of the ladder behaviour is exercised outside tests, and it is what makes several other findings latent rather than live.

_Suggested:_ One line in the PR body saying `sign()` is also unreached pending the request handler (core#2117).

<sub>flagged by guided review; verified — no src/main caller</sub>

src/main/java/com/otilm/core/signing/contentsigning/ManagedContentSigningEngine.java
```diff
@@ -0,0 +1,408 @@
...
+        }
+
+        SignedContent result = new SignedContent(signedDocument, reached, serials);
+        recordSigning(signingProfile, profile, result, dtbs.getDtbs(), signatureValue, signingTime, protocol);
+        return result;
+    }
+
+    /**
+     * Raises a signature made elsewhere, entering the machine at the signature-timestamp imprint with {@code SIGNED} as
+     * the given cursor. Deliberately has no caller: the machine is augmentation-ready, and the door is not opened in
```
lubomirw: **[Low]** | Maintainability

Three spots where a comment argues a decision rather than stating a constraint, and each rots on a different trigger:

- **here** — "Deliberately has no caller: the machine is augmentation-ready, and the door is not opened in v1" goes stale the moment a caller lands or the next version ships.
- **`requireReachableLevel` (~line 152)** — "the second one goes away once the higher rungs gain steps of their own"; "the second one" is a positional reference into the body that misdirects as soon as a third refusal appears or the two are reordered.
- **`ContentSigningWorkflowValidatorTest` (~line 218)** — four lines explaining why a test was *not* written, including which mocking tools refuse to fabricate the instance.

Worth noting what the scan did **not** find, across 5091 added lines: no internal refs, no log noise, three log statements total, all `warn`/`error` on failure paths.

_Suggested:_ Keep the behavioural sentence in each and drop the argument — this reasoning belongs in the PR discussion. For the second, hang the temporariness off `ContentSigningTransitions.HIGHEST_EXECUTABLE_LEVEL`, which already documents that raising the ceiling needs new steps.

lubomirw (changes requested):
Automated multi-reviewer triage — 14 findings (REQUEST_CHANGES).

Four reviewers ran independently (Copilot CLI, guided review, superpowers code review, hygiene) — 48 raw findings, 27 after dedupe. These are the 14 I could verify against `2430572`; the judgment calls and three rows I judged likely noise are held back rather than posted.

Your **Accepted residuals** list did a lot of work here: several findings landed on ground you had already claimed, and I dropped those. What survives are the cases where a reviewer found something the residual does not cover — most notably the `documentSizeCap` contract wording, which asserts enforcement on interfaces `main` today.

Three corrections I made to reviewer output, for the record:
- All three reviewers claimed the edit path silently invalidates a referenced timestamp source. It does not — `InternalTimestampSource.requireRecordFloor` says "**no longer** meets the signing-record floor" and `resolveTimestampingProfile` has an `instanceof` guard with a specific message. Deliberate request-time re-validation, so I dropped it to a usability preference and left it out.
- The ungated `signingTime` at `SIGNED` was raised twice as a defect. A claimed signing time at B level is unverified by definition — that is what B+T is for. Dropped.
- The GIN index without `CONCURRENTLY` was raised twice. 0 of 18 core migrations use it, so that is house convention, not a defect here. Dropped.

**Findings with no line in this diff** — all three in the `documentation` repo, which this PR does not touch:

- **`docs/signing/signing-profile.md:54`** and **`docs/signing/concepts.md:65`** (Medium) — both still state Timestamping is the only available workflow, and the Workflow Properties section links only Timestamping. This PR adds four operator-facing settings (family, maxLevel, timestampSource, documentSizeCap) and the SIGNED/TIMESTAMPED ladder. Worth saying: these pages were *already* stale, since managed CONTENT_SIGNING was creatable before this branch — but this is the change that makes fixing them unavoidable. Suggested: a `docs/signing/contentsigning/configuration.md` covering the four settings and the ladder (including that a maxLevel above TIMESTAMPED is refused in this version), linked from signing-profile.md, plus the concepts.md Workflows section and its Available configuration table.
- **`docs/signing/signing-profile.md:100`** (Medium) — the Deletion section lists only signing records and TSP-Profile references; the new timestamp-source blocker is undocumented, as is the fact that a reference held only by a superseded version is permanent for that profile's lifetime. `docs/signing/timestamping/limitations.md:31` needs the same addition.
- **`docs/signing/signing-records.md:58`** (Low) — presents Recording Enabled and the three persistence modes as freely chosen, but `SigningRecordFloor` now makes recording mandatory and BEST_EFFORT forbidden for any TIMESTAMPING profile named as a timestamp source, refused both at save and at issue time.

src/main/java/com/otilm/core/signing/contentsigning/profile/ContentSigningWorkflowValidator.java
ivosh: Fixed in 4c683a22. `validate` now takes the target profile UUID (null on create, the profile's UUID on update) and `requireDistinctTimestampSource` rejects a self-reference outright, so the end state with no way out is unreachable. New test: `refusesAProfileThatNamesItselfAsItsTimestampSource`.

src/main/java/com/otilm/core/signing/contentsigning/profile/ContentSigningWorkflowValidator.java
ivosh: Fixed in 4c683a22. `requireUsableTimestampingProfile` now refuses a `managedSigningType` other than `STATIC_KEY` after the `MANAGED` check, naming both the found and the required value. Covered by `refusesAReferencedProfileThatIsNotStaticKeyManaged` alongside the existing not-timestamping / not-managed / below-floor cases.

src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
ivosh: Fixed in 4c683a22. `validateContentSigningWorkflow` now calls `requireStaticKeyManagedSigning` on the profile's own scheme before the connector is fetched, so a MANAGED/ONE_TIME_KEY content-signing profile is refused at save. Covered by `managedContentSigning_withAOneTimeKeyScheme_throwsValidationException`.

src/main/java/com/otilm/core/signing/contentsigning/formatting/ContentSigningFormattingAttributes.java
ivosh: Fixed in 4c683a22. `CANONICAL_FORM` is now `ObjectMapperFactory.jsonColumn()`, so the JSR-310 modules are registered and a `ZonedDateTime` default renders. `JsonProcessingException` now throws `IllegalStateException` instead of returning `false`, so our own serialization fault can no longer surface as the connector-disagreement message.

src/main/java/com/otilm/core/signing/contentsigning/ContentSigningRecordFactory.java
ivosh: Fixed in 4c683a22. Extracted `com.otilm.core.signing.record.TimestampTokenSerialNumbers` with `hex(BigInteger)` and `hex(List<BigInteger>)`; all three call sites route through it and the two duplicate private helpers are gone.

src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
ivosh: Fixed in 4c683a22. The superseded-only branch now spells the full remedy: the hold is released only by deleting the referencing Signing Profile, which first requires deleting its signing records. The referencing profile names are already appended to the message.

src/main/java/com/otilm/core/signing/contentsigning/ManagedContentSigningEngine.java
ivosh: Fixed in 4c683a22. All three trimmed to their behavioural content: the `augment` javadoc drops the no-caller argument, `requireReachableLevel` now hangs the temporariness off `ContentSigningTransitions.HIGHEST_EXECUTABLE_LEVEL` with no positional reference, and the four-line absent-test note in `ContentSigningWorkflowValidatorTest` is gone.

src/main/java/com/otilm/core/signing/contentsigning/ManagedContentSigningEngine.java
ivosh: Fixed in 4c683a22 — added to the accepted residuals in the PR body: `sign(...)` is unreached from `src/main` as well, nothing resolves a content-signing profile until the request handler lands (core#2117), so the whole ladder is exercised by tests only, and that is what keeps several findings latent rather than live.

src/main/java/com/otilm/core/signing/contentsigning/ManagedContentSigningEngine.java
ivosh: Taking the first of your two options: correct the wording, keep the check where it belongs. On OmniTrustILM/interfaces#893, `documentSizeCap` on both DTOs now reads "Largest document accepted for signing, in bytes. Enforced when an ILM-managed signing request arrives." The check itself arrives with core#2010/core#2117, and the PR body's residual now says so.

Not gating here on purpose. The engine receives a `DocumentTransferDto` that may be a `DigestOnlyDocumentTransferDto` with no document to measure, and the delegated path never reaches this engine at all, so an engine-side gate could only ever bound half the surface while the wording still needed a carve-out. The handler that accepts the request is the layer that holds the upload and knows its size.

Leaving this open — the interfaces change is not merged yet.

src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
ivosh: No change here, deliberately. The interim is the answer until core#2124 lands.

You are right that the two sides now disagree. Routing the listing through `ContentSigningFormattingAttributes.aggregate` is not available to this endpoint: it needs a `maxLevel` the route has no parameter for, and deriving one from `signingProfileUuid` would contradict the line interfaces#893 just shipped, which states that parameter is authorization-only. Making it fail loudly would break the FE's attribute panel before core#2124 has anywhere for it to go.

So core is knowingly left as-is and the state is now recorded on core#2124 as a DoD item: the legacy route is left with no content-signing caller only once that endpoint plus OmniTrustILM/fe-administrator#1997 exist. Until then a content-signing caller of the legacy route gets a timestamping-shaped descriptor set from a connector declaring both interfaces, or an opaque connector error from a content-signing-only one.

Leaving open for you.

src/main/java/com/otilm/core/signing/contentsigning/resolver/StaticKeyManagedContentSigningResolver.java
ivosh: No change, by decision rather than disagreement with the analysis.

No customer uses content signing today: content-signing profiles have been configurable since 2.19.1, but the formatting connector that would make them functional does not exist yet, so a profile row could exist while never having been able to sign anything. On that basis both a backfill migration and an upgrade release note were explicitly ruled out.

A backfill is also not safely writable: `signature_family` cannot be inferred from existing data, so any value would silently mis-configure the profile — and `max_signature_level = 'SIGNED'` alone still leaves such a row un-editable, because a GET returns `family: null` and `requireFamily` rejects the PUT.

Leaving open — happy to add the explicit null-maxLevel refusal with a message naming the real cause if you would rather have the better error than none.

lubomirw: No change is okay approach.

lubomirw (approved):

PR #2121, opened 2026-08-22
Support Java time values without exposing Jackson internals

### LINKED ISSUE #2088: ObjectMapperFactory.storage() has no Jackson modules, so any java.time value it touches throws
### Description

**`ObjectMapperFactory.storage()` is a bare `new ObjectMapper()`, and under Jackson 2.22 a bare mapper refuses to serialize a `java.time` value rather than degrading.**

`MapperFeature.REQUIRE_HANDLERS_FOR_JAVA8_TIMES` is on by default in jackson-databind 2.22.1. Without `jackson-datatype-jsr310` registered, Jackson throws instead of writing anything.

`@JsonFormat` does not help. It configures a serializer; it does not supply the missing handler.

Three factory recipes are affected, because two are built from the first:

- `storage()` — `new ObjectMapper()`
- `lenientStorage()` — `storage()` plus `FAIL_ON_UNKNOWN_PROPERTIES = false`
- `emptyBeanTolerantStorage()` — `lenientStorage()` plus `FAIL_ON_EMPTY_BEANS = false`

### Steps to reproduce

1. Settings → Custom Attributes → create `repro-datetime`, content type **Datetime**, resource **Certificates**.
2. Workflows → Actions → **Executions** tab → Add.
3. Execution Type **Set Field**, Resource **Certificates**, field source *Custom Attribute*, field `repro-datetime`, any value → Save.

Or without a running platform, against the test classpath:

```java
ExecutionItem item = new ExecutionItem();
item.setFieldSource(FilterFieldSource.CUSTOM);
item.setFieldIdentifier("repro-datetime|DATETIME");
item.setData(new ArrayList<>(List.of(new DateTimeAttributeContentV3(ZonedDateTime.now()))));

item.mapToDto();   // throws
```

### Actual result

HTTP 400:

```
Java 8 date/time type `java.time.ZonedDateTime` not supported by default: add Module
"com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling
(or disable `MapperFeature.REQUIRE_HANDLERS_FOR_JAVA8_TIMES`)
(through reference chain: java.util.ArrayList[0]->DateTimeAttributeContentV3["data"])
```

Observed on develop-02 running `core:develop-latest`, audit log `POST /api/v1/workflows/executions`, `operationResult: failure`. The local snippet produces the identical message and reference chain.

### Expected result

The execution is created and the response carries the datetime value.

### Where it comes from

`ExecutionItem.mapToDto` (`ExecutionItem.java:85`) calls `ObjectMapperFactory.lenientStorage()` and `convertValue` on the attribute content it holds. `ActionServiceImpl.createExecution` calls it at line 118 to build the response, after `save` at line 116 and before the flush. **Hibernate never sees the value, so no column mapper is involved.** `convertValue` throws `IllegalArgumentException`, and `ExceptionHandlingAdvice.java:148` maps that to 400.

`ActionServiceImpl.updateExecution` has the same pair at lines 144 and 146, so editing such an execution fails the same way.

This is not environment-specific. It reproduces on a flat classpath, with no fat jar and no database.

No test covers execution creation with a DATETIME custom attribute, which is why CI is green.

### When it breaks

Any path that hands one of these three mappers an object graph containing a `java.time` value. `ExecutionItem.mapToDto` is the one reachable from the UI today. Thirteen other call sites share the recipe family:

| File | Recipe |
| --- | --- |
| `util/MetaDefinitions.java:15` | `storage()` |
| `util/SerializationUtil.java:11` | `emptyBeanTolerantStorage()` |
| `util/CryptographicHelper.java:22` | `emptyBeanTolerantStorage()` |
| `service/impl/SecretServiceImpl.java:125` | `storage()` |
| `service/impl/SettingServiceImpl.java:104` | `storage()` |
| `service/impl/CallbackServiceImpl.java:615` | `lenientStorage()` |
| `validation/certificate/X509CertificateValidator.java:48` | `storage()` |
| `security/exception/AuthenticationServiceException.java:19` | `lenientStorage()` |
| `dao/entity/CertificateEventHistory.java:71` | `storage()` (read) |
| `dao/entity/CryptographicKeyEventHistory.java:68` | `storage()` (read) |
| `intune/scepvalidation/IntuneClient.java:457` | `storage()` |
| `intune/scepvalidation/IntuneScepServiceClient.java:54` | `storage()` |
| `intune/scepvalidation/IntuneRevocationClient.java:54` | `storage()` |

### When it is fine

A payload with no `java.time` value serializes exactly as before. The REST attribute path is unaffected: there `data` stays a `String`.

### Why now

The feature that turns this into a hard failure arrived with Jackson 2.20. Before it, a bare mapper wrote a `java.time` value as a bean-shaped object — wrong, but silent. Several of these call sites write to storage, so a change of recipe changes the shape of what they write and needs to be stated deliberately.

### Secondary finding

The raw Jackson message reached the HTTP client. It carries class names and a Jackson reference chain, which is what the "don't leak runtime details to the wire" rule in `CLAUDE.md` exists to prevent. `ExceptionHandlingAdvice.handleIllegalArgumentException` forwards `getMessage()` for an exception type that is not one of ours.

### Acceptance Criteria

- [ ] The three storage recipes handle `java.time` values, or each call site that can see one is moved to a recipe that does.
- [ ] Creating a set-field execution for a DATETIME custom attribute succeeds, covered by a test.
- [ ] For each of the 14 call sites, it is recorded whether its payload can carry a date.
- [ ] Any change of recipe states what it does to the serialized shape.
- [ ] `handleIllegalArgumentException` no longer forwards a raw exception message to the client.

### Related

Not fixed by #2000 / #2084. Those state the Hibernate `FormatMapper` for `@JdbcTypeCode(SqlTypes.JSON)` columns; this failure happens before Hibernate is reached.

ivosh:
**Closes #2088. `ObjectMapperFactory.storage()` was a bare `new ObjectMapper()`, and on jackson-databind 2.22.1 a bare mapper refuses to serialize a `java.time` value instead of degrading — so creating a set-field execution for a DATETIME custom attribute answered HTTP 400, with the raw Jackson message on the wire.**

## 1. The bug

`ExecutionItem.mapToDto` builds the response by converting the stored attribute content through `lenientStorage()`, which is derived from `storage()`. `DateTimeAttributeContentV3` holds a `ZonedDateTime`.

`MapperFeature.REQUIRE_HANDLERS_FOR_JAVA8_TIMES` is on by default in jackson-databind 2.22.1. Without `jackson-datatype-jsr310` registered, Jackson throws rather than writing anything.

**This is live on `main` today.** `jackson-bom` resolves to 2.22.1 on the current 3.5 build. The feature arrived in Jackson 2.20, so it came from ordinary BOM drift, not from the Spring Boot 4.1 work.

**When it's fine:** any payload with no `java.time` value serializes exactly as before. The REST attribute path is unaffected, because there `data` stays a `String`.

**When it breaks:** any path handing one of these three recipes an object graph containing a `java.time` value. `ExecutionItem.mapToDto` is the one reachable from the UI, on create (`ActionServiceImpl:118`) and on update (`:146`).

**The database was never at risk.** `ExecutionItem.data` is annotated `@Convert(converter = ObjectToJsonConverter.class)`, and that converter serializes through the injected `ObjectMapper` bean — the `wire()` recipe, which already registers `JavaTimeModule`. The column held ISO text before this change and holds ISO text after it. No data migration is implied.

## 2. A second site, silent

`CallbackServiceImpl.processGroupAttributes:615` performs the same conversion into `List<BaseAttribute>`, wrapped in a `catch (Exception e)` that only logs at debug level. A connector returning a group attribute with a DATETIME default silently loses its attribute definitions. Same root cause, no error surfaced, no test. The recipe fix closes it.

## 3. The fix, line by line

The three added lines do not carry equal weight, and it is worth being explicit about which one fixes the report.

| Line | Fixes the reported 400? | What it buys |
| --- | --- | --- |
| `registerModule(new JavaTimeModule())` | **Yes — this is the fix** | necessary and sufficient for the reported endpoint |
| `disable(WRITE_DATES_AS_TIMESTAMPS)` | No | text form for every *other* date the recipe touches |
| `disable(ADJUST_DATES_TO_CONTEXT_TIME_ZONE)` | No | preserves the caller's offset for every other caller |

The two zone/format lines have no effect on the reported endpoint, because `DateTimeAttributeContentV3.data` carries its own `@JsonFormat(shape = STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSSXXX")` and `@JsonDeserialize(using = ZonedDateTimeDeserializer.class)`. That field bypasses both global settings. They are pinned by unit tests instead, and anyone reading only the integration test would draw the wrong conclusion about which line does the work.

**Only the time module is registered, deliberately.** `findAndRegisterModules()` would also install the JDK8 module and unwrap `Optional` — a shape change for exactly the callers `emptyBeanTolerantStorage()` exists for. A test pins that `Optional` handling is unchanged.

## 4. Serialized shape

A date now writes as ISO-8601 text keeping its offset, matching what `wire()` and `jsonColumn()` already chose. Non-date payloads serialize byte-identically.

**Nothing that previously succeeded changes shape.** Writing a date through these recipes has thrown since Jackson 2.20, so no stored row can contain a recipe-written date, and there is no legacy shape for this change to break.

## 5. Call-site audit

All 14 sites, and whether each payload can carry a date:

| Verdict | Sites |
| --- | --- |
| **Date-capable, broken before this change** | `ExecutionItem:85` (loud 400), `CallbackServiceImpl:615` (silent) |
| **Safe — tree API only** | `IntuneClient:457`, `IntuneScepServiceClient:54`, `IntuneRevocationClient:54` use only `createObjectNode()` / `readTree()`, never POJO serialization |
| **Safe — payload has no date field** | `X509CertificateValidator:48` and `MetaDefinitions.serializeValidation` (`CertificateValidationCheckDto` is enum + enum + String), `AuthenticationServiceException:19`, `SecretServiceImpl:128`, `SettingServiceImpl:104`, `CryptographicHelper:22`, `SerializationUtil:11` (ACME contacts, identifiers, JWS payload maps) |
| **Latent, no date caller today** | `MetaDefinitions.serialize(Map<String, Object>)` — every caller puts UUIDs, strings and serial numbers into `additionalInformation`; the signature permits a date |
| **Unaffected either way** | `CertificateEventHistory:71`, `CryptographicKeyEventHistory:68` read into an untyped `Map<String, Object>`, where a date module has no target type to bind |

## 6. The leaked message

`handleIllegalArgumentException` forwarded `ex.getMessage()`, so the client received the Jackson class names and reference chain. That is what the "don't leak runtime details to the wire" rule in `CLAUDE.md` exists to prevent.

The handler now logs the real message and returns a fixed one. It only ever sees what the JDK and third-party libraries throw — `ValidationException` and `ConstraintViolationException` have their own handlers and keep their messages.

## Verification

- Full suite: **5241 tests, 0 failures, 0 errors**, 2 skipped.
- New-lines coverage: **100%** against the 80% gate.
- Both fixes are mutation-checked. Reverting the recipe makes the integration test fail with the issue's exact reference chain; reverting the handler makes the new advice test fail with the leaked message.
- The zone decision is pinned on its own: without that one flag the unit test reports `15:00Z` instead of `17:00+02:00`.
- `ContextSignatureGuardTest` BASELINE unchanged at 58 — `ActionServiceITest` already extends `BaseSpringBootTest`, so no new Spring context was purchased.

Note that `ExceptionHandlingAdvice` sits under `sonar.coverage.exclusions`, so its changed line contributes nothing to the coverage figure. The new advice test is what covers it.

## Follow-ups, deliberately not folded in

- #2122 - the debug-swallowing `catch (Exception e)` in `processGroupAttributes` stops hiding *this* bug, but will hide the next one.
- #2123 - our own deliberate `IllegalArgumentException` throws (`DtbsBindingVerifier`, `CertificateChain`) now lose their curated messages at the HTTP layer. The fix is converting them to `ValidationException`.


lubomirw (commented):
Automated multi-reviewer triage - 4 findings (COMMENT).

Overflow findings (not inline-able on the current diff):
- src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java:153 - [Low Error handling] The generic IllegalArgumentException handler now correctly avoids leaking runtime detail, but the new comment assumes deliberate request rejections always use ValidationException even though the codebase still has many first-party IllegalArgumentException throw sites that may reach REST responses and lose actionable client detail. Suggested: Keep the generic fallback for unexpected IllegalArgumentException, but audit request-facing IllegalArgumentException throw sites and convert intentionally user-safe validation failures to ValidationException or another controlled domain exception.
- src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java:143 - [Nit Documentation] The new Javadoc is somewhat verbose for the IllegalArgumentException handler. Suggested: Optionally tighten it to a shorter contract such as: Handles IllegalArgumentException without returning potentially internal messages; deliberate validation messages should use ValidationException.
- src/main/java/com/otilm/core/serialization/ObjectMapperFactory.java:61 - [Nit Documentation] The storage mapper Javadoc is long, but most of it records the non-obvious Optional-shape migration constraint. Suggested: Optionally condense it to two sentences covering offset-preserving date serialization and the deliberate exclusion of module discovery to avoid changing persisted Optional shapes.
- src/test/java/com/otilm/core/integration/service/ActionServiceITest.java:69 - [Nit Documentation] The DATETIME regression test Javadoc narrates more of the path than needed. Suggested: Optionally shorten it to: Guards that DATETIME values pass through ExecutionItem.mapToDto with their submitted offset; ObjectMapperFactoryTest separately pins mapper zone behavior.

lubomirw (approved):

PR #2132, opened 2026-08-24
Clear the Sonar findings on the branding settings code

Jakub-Bartal-01:
Clears the five SonarCloud findings that #2081 and #2082 left on `main`. No behaviour changes.

### `BrandingLogoValidator`

- **java:S8786** — `SVG_LENGTH` matched `\s*(unit)?\s*$`, so with no unit present both whitespace runs could claim the same spaces and the engine backtracked over every split. The runs are now possessive, which makes the match linear; a 40 000-space input goes from quadratic to 1 ms. The accepted language is unchanged.
- **java:S127** — the PNG chunk walk advanced `offset` from inside the loop body, because the step is the chunk length read from that chunk. It is a `while` loop now, so nothing assigns to a `for` counter.

### `SvgSanitizer`

- **java:S1192** — `"style"` appeared three times, for the `<style>` element and the `style` attribute. It is a named constant.

### Tests

- **java:S8491** — the Javadoc explaining why an unrecognised stored theme code is dropped sat above `aBrandingUpdateWaitsForAConcurrentOneToRelease`, which it does not describe. Moved onto `anUnknownStoredThemeIsIgnoredRatherThanFailingTheRead`.
- **java:S5778** — `everyRejectionNamesTheFieldItCameFrom` built its PNG fixture inside the `assertThrows` lambda. The fixture is built first, so only the call under test can throw from inside the lambda.


src/test/java/com/otilm/core/integration/service/BrandingSettingsITest.java
```diff
@@ -181,6 +177,10 @@ void aBrandingUpdateWaitsForAConcurrentOneToRelease() throws Exception {
         Assertions.assertEquals(PRIMARY, settingService.getBrandingSettings().getPrimaryColor());
     }
 
+    /**
+     * The platform settings read is on the hot path for every page render, so a theme code that no longer maps to
+     * anything is dropped rather than allowed to fail the whole read.
+     */
```
ivosh: **. Nitpick** | Documentation

The test name `anUnknownStoredThemeIsIgnoredRatherThanFailingTheRead` already states the behaviour, and the three lines below the comment show it directly. The Javadoc restates that in prose, so it costs a maintainer a read without telling them anything the signature and body do not.

_Suggested:_ Drop the Javadoc block; the S8491 fix is equally satisfied by removing the comment as by relocating it.

ivosh (changes requested):

src/test/java/com/otilm/core/integration/service/BrandingSettingsITest.java
Jakub-Bartal-01: Kept one sentence, dropped the restatement. The test name already says the read is not failed; what it cannot carry is *why* that matters — the platform settings read is on the hot path for every page render — so that clause stays and the rest is gone.

Now: `/** The platform settings read is on the hot path for every page render, so it must not fail over one bad value. */`

ivosh (approved):

Issue #1883, opened 2026-07-24
STRICT request-attribute validation rejects standard CSR extensions (EKU), making ACME/protocol issuance fail on RA profiles with no extension mappings


lubomirw:
### Description

When issuing a certificate via ACME finalize against RA profile `EJBCA-Empty` (no custom request attributes, no extension mappings, request-attribute validation set to **STRICT**), the request is rejected with `acme:error:badCsr` because the CSR carries the Extended Key Usage extension (OID `2.5.29.37`). Switching the profile to **LENIENT** lets issuance succeed.

**Root cause** (confirmed against a local `core` checkout): STRICT validation runs a presence-whitelist over CSR extensions — every extension OID in the CSR must be explicitly mapped by an `ExtensionMappedField` in the resolved request-attribute set, or the request is rejected (`CertificateRequestContentValidator.java:260-265`). A profile with no mappings does **not** get an empty set: resolution falls back to the platform default set, which is 6 RDN-only attributes with zero SAN/extension mappings (`RaProfileCertificateRequestAttributeServiceImpl.java:85-89`, `CsrAttributes.java:45-54`). So under STRICT, any CSR carrying any non-SAN standard extension (EKU, keyUsage, basicConstraints) fails. Only SAN (`2.5.29.17`) is exempt (`X509RequestContentParser.java:243-245`). LENIENT skips the whitelist entirely (`ProtocolRequestAttributeValidator.java:78-80`).

**Why this is a bug, not merely strict config:** nearly every real CSR generator — and all common ACME clients — emit EKU/keyUsage. Platform-generated CSRs are built *from* the mapped definitions so they always pass; only externally-supplied CSRs (ACME/SCEP/CMP + REST upload) are exposed — exactly the CSRs the operator cannot control. The whitelist is presence-only (extension values are opaque DER, not constrained), so rejecting benign standard extensions adds no real security. Net effect: STRICT is unusable with real protocol clients out of the box.

**Recommended fixes:**

*Primary:*

1. **Default-allow standard extensions** — a built-in, configurable allowlist of benign OIDs (`2.5.29.37` EKU, `2.5.29.15` keyUsage, `2.5.29.19` basicConstraints) consulted in the whitelist check (`CertificateRequestContentValidator.java:260`), OR seed non-required extension-mapped attributes for those OIDs into the platform default set (`CsrAttributes.java:45` / `DefaultRequestAttributeSet.seed()`).
[...]

PR #2142, opened 2026-08-25
Project and validate structured key usage and extended key usage values

### LINKED ISSUE #1950: Project and validate structured key usage and extended key usage values
### Description

Backend implementation of the structured extension inputs. Parent: OmniTrustILM/ilm#325. Contract:
OmniTrustILM/interfaces#808. The JSON encoding and `JsonSchemaConstraint` enforcement moved to
OmniTrustILM/core#2131.

Key Usage and Extended Key Usage now travel as typed lists on `X509RequestContent`, so ASN.1
happens at exactly the two boundaries that already translate `subjectAltNames`: the renderer encodes on the way
out, the parser decodes on the way in. The projector and the validator never touch bytes.

**`StructuredExtensionCodec`.** One new pure kernel in `com.otilm.core.util`, beside
`X509RequestContentRenderer`, the single place that knows the two ASN.1 shapes. It sits there rather than in
`com.otilm.core.certificate.request` because that package already depends on `core.util`, and the reverse edge
would close a package cycle. Encode and decode live together so they cannot drift.

| Extension | OID | ASN.1 |
|---|---|---|
| Key Usage | 2.5.29.15 | `BIT STRING` at X.509 bit positions |
| Extended Key Usage | 2.5.29.37 | `SEQUENCE OF OBJECT IDENTIFIER` |

Both directions speak the extension's inner value DER, base64-encoded - what `getExtnValue().getOctets()`
yields and what `ExtensionsGenerator.addExtension` consumes.

**Bit layout - the trap.** `CertificateKeyUsage.getIndex()` is the X.509 BIT STRING bit position (0 =
digitalSignature), which is what `CertificateUtil.isKeyUsagePresent` already relies on.
`CertificateKeyUsage.getBit()` returns `1 << index`, which is the platform's **database bitmask** layout
(`V202509041555__CertificateRequestEntityBitmask` via `BitMaskEnum.convertSetToBitMask`) and is **not** the DER
layout - nor is BouncyCastle's `KeyUsage` constant layout, which uses `1 << 7` for digitalSignature and
`1 << 15` for decipherOnly. Three plausible-looking integers, two wrong. Encode bit positions directly: for
position `i`, set `bytes[i / 8] |= 0x80 >>> (i % 8)`, then strip trailing zero bits per DER minimal encoding.
Golden vectors pin it, cross-checked against BouncyCastle's own reader and against OpenSSL.

**Projection.** `CertificateRequestAttributeProjector.applyMappedField` gains a structured path ahead of its
switch, keyed on `StructuredExtensionCodec.oidFor(field)`, which returns the target's OID or `null`. It appends
the attribute's values to the matching typed list - no ASN.1 - mirroring how SAN entries are appended. The OID
is still claimed in `seenExtensionOids`, so a duplicate structured target across two attributes, or a collision
with an explicit `ExtensionMappedField` on the same OID, is rejected with the existing RFC 5280
one-extension-once message.

**Rendering.** `X509RequestContentRenderer.toExtensions` builds the two extensions from the typed fields,
next to the SAN block it already has: OID from the codec, criticality from the OID registry with
`CRITICALITY_FORCED_OIDS` applied, value from the codec. An empty list emits nothing - RFC 5280 forbids an
empty keyUsage bit string and an empty EKU sequence. The existing duplicate-OID guard covers them.

**Parsing.** `X509RequestContentParser` diverts the two OIDs out of `parseExtensions` into the typed fields,
exactly as `Extension.subjectAlternativeName` is diverted today.

`extendedKeyUsage` is lossless - any OID round-trips. `keyUsage` is not: X.509 closes it at nine named bits
and `CertificateKeyUsage.fromIndex(9)` throws.

So the single unrepresentable case is a key usage bit beyond the named list. ASN.1 permits one and no CA sets
one, so it is exotic rather than realistic - but it must not be dropped, because strict validation promises
"only what the resolved set maps" and the CSR is forwarded to the CA verbatim. `ParsedRequestContent` therefore
gains `unrepresentableExtensionValues` alongside `unsupportedSans`, carrying entries like `Key Usage bit 9`,
surfaced by the same whitelist pass. An undecodable value is reported the same way rather than thrown.

**Registration-time validation.** `AttributeEngine.validateMappedField` gains a structured branch: the
attribute must be a `list`; it must declare the items it permits or set `extensibleList`; `readOnly` with
`extensibleList` is contradictory and rejected; and every declared item must be a `CertificateKeyUsage` code or a
purpose registered under `OidCategory.EXTENDED_KEY_USAGE` (`SystemOid` plus `ExtendedKeyUsageCustomOidEntry`).

The "declare the items or opt out" rule is not bureaucracy: `AttributeEngine`'s existing predefined-list check
rejects any content item absent from the definition's content, so a structured target with empty content would
reject every CSR value at request time - a silent footgun.

**SAN treatment.** `validateExtensionMappedField` additionally rejects an `ExtensionMappedField` whose OID is
either of the two, naming the structured target - the same rule, place and wording SAN already uses ("use the
SAN mapping target instead").

**Backwards compatibility.** Both `validateFieldMapping` call sites are write paths - an operator saving a set
and a connector registering definitions. Projection never re-validates, so a definition already stored with a
base64-DER mapping keeps projecting unchanged, with no migration. Re-saving that set fails until the author
switches to the structured target, and so does a connector re-registering such a definition, which happens on
reconnect and on connector upgrade. This is deliberate and must be release-noted.

**Uploaded-CSR enforcement.** Because the parser produces typed lists, the validator compares strings and never
touches ASN.1. `CertificateRequestContentValidator` gains a structured branch that compares the CSR's values
against the definition per the semantics agreed on ilm#325:

| Definition | Meaning for the CSR's values |
|---|---|
| `list`, predefined content | permitted set - every value must be in it |
| `+ required` | must also be non-empty |
| `readOnly` | pinned - the set must equal the content exactly |
| `extensibleList` | unrestricted, except an item the platform cannot name, which fails closed |

Violations route through the existing `recordViolation`, honouring `RequestAttributePolicy` (strict errors,
lenient warns).

**The whitelist needs an explicit pass.** Because the parser diverts these OIDs out of `extensions`, the
generic extension-OID whitelist pass can no longer see them - so a CSR requesting a key usage that *no*
definition maps would otherwise sail through a strict policy and reach the CA unfiltered. The validator tracks
which structured targets the set maps and flags a typed field the CSR carries that none of them claims.

The existing `ExtensionMappedField` path is unchanged and stays presence-only; core#2131 addresses it.

### Acceptance Criteria

- [ ] `StructuredExtensionCodec` encodes and decodes both shapes, pinned by golden DER vectors: `digitalSignature` = `03 02 07 80`, `keyCertSign` = `03 02 02 04`, `digitalSignature`+`keyEncipherment` = `03 02 05 A0`, `decipherOnly` = `03 03 07 00 80`, EKU `serverAuth` = `30 0A 06 08 2B 06 01 05 05 07 03 01`, TLS feature 5 = `30 03 02 01 05`.
- [ ] Encode/decode round-trip per shape, plus a cross-check against BouncyCastle's own `KeyUsage` reader and against a CSR generated by OpenSSL.
- [ ] Projection populates the typed fields and emits no `RequestedExtension` for either; a duplicate structured target, or a collision with an explicit extension mapping on the same OID, is rejected.
- [ ] Rendering produces the two extensions with registry criticality, keyUsage forced critical, and nothing at all for an empty list.
- [ ] Parsing diverts the two OIDs into the typed fields; a key usage bit the platform cannot name lands in `unrepresentableExtensionValues` rather than being dropped.
- [ ] Platform round-trip: project, render into a real CSR, parse back, and the typed values are identical.
- [ ] EKU purpose OIDs validated against the `OidCategory.EXTENDED_KEY_USAGE` registry; an unregistered OID is rejected with an actionable message.
- [ ] A structured target that is not a list, declares no permitted items without `extensibleList`, or is both `readOnly` and `extensibleList`, is rejected at registration.
- [ ] An `ExtensionMappedField` on 2.5.29.15 or 2.5.29.37 is rejected at registration naming the structured target, while a stored definition carrying one still projects.
- [ ] Uploaded-CSR validation compares the CSR's values against the mapping per the table above, under both lenient (warn) and strict (error).
- [ ] A structured extension the CSR carries that no definition maps is flagged under whitelist.
- [ ] `mvn -B verify` passes including integration tests.
- [ ] Sonar quality gate passes and new-code line coverage is at least 80%.

### Notes

TLS Feature was in scope and was dropped: its vocabulary is an open IANA registry with nothing authoritative to
validate against, and the feature is effectively abandoned in the field. It moves to core#2131 via the
structural JSON tree.

Implementation plan:
`request-attributes-design-and-plans/core/2026-08-21-structured-extension-mapping-targets-core.md`

Populating `ClientCertificateDataResponseDto.requestAttributeWarnings` belongs to OmniTrustILM/core#1951. This
issue produces the warnings; that one carries them to the client.

klaraf755:
Closes #1950

**Requires OmniTrustILM/interfaces#907.** 

Key usage and extended key usage now travel as typed values, so ASN.1 happens at exactly the two boundaries
that already translate `subjectAltNames`: the renderer encodes on the way out, the parser decodes on the way
in. `StructuredExtensionCodec` holds both directions so they cannot drift; the projector and validator never
touch bytes.

**New capability.** An uploaded CSR's key usage and extended key usage are checked *by value*, not only for
presence. A predefined list is a permitted set, `required` demands a non-empty one, `extensibleList` lifts the
restriction - the attribute semantics the platform already has, no new field. This is what
OmniTrustILM/core#1883 was closed pending.

There is no pinned-set case. `readOnly` would express "exactly these", but a read-only attribute cannot be a
list and a set-valued target must be one; a test pins that so the absence is not mistaken for an oversight.

**Two things worth a reviewer's attention.**

*Bit layout.* `CertificateKeyUsage` exposes three plausible integers and only one is right for DER:
`getIndex()` is the X.509 bit position, `getBit()` is the database bitmask layout
(`V202509041555__CertificateRequestEntityBitmask`), and BouncyCastle's `KeyUsage` constants are a third,
first-octet-major packed int. `getBit()` is the trap - it is on the same enum and produces a valid-looking
extension with wrong bits. The codec sets bit positions directly and never converts to an int. Golden vectors
pin it, cross-checked against DER written by OpenSSL and read back through BouncyCastle's own `KeyUsage`.

*The whitelist needed its own pass.* The parser diverts these OIDs out of the extension list, so the generic
extension-OID whitelist pass can no longer see them - a CSR requesting a key usage that no definition maps
would have passed a strict policy and reached the CA unfiltered. `checkStructuredWhitelist` restores the
pre-change outcome; this was found reviewing the diff, not by a failing test, so it is the piece I would most
like a second pair of eyes on.

**Fails closed.** A key usage bit outside the nine X.509 names cannot be represented, so it is recorded in
`ParsedRequestContent.unrepresentableExtensionValues` alongside unsupported SANs rather than dropped. Exotic
(no CA sets one) but the CSR is forwarded to the CA verbatim, so silence would mean bypass.

**Breaking on edit.** Mapping either OID as opaque base64 DER is rejected at registration, naming the
structured target - the same treatment subjectAltName already gets. Projection never re-validates, so a stored
definition keeps projecting unchanged and no migration is needed; but re-saving such a set, or a connector
re-registering such a definition, requires switching to the structured target. Release-noted on
OmniTrustILM/ilm#325, and a test pins the still-projects half.

**Verified.** All four CI profiles locally: `test-non-integration` 2746, `test-integration-core` 810,
`test-integration-service-1` 1028, `test-integration-service-2` 1016. `ContextSignatureGuardTest.BASELINE`
unchanged at 58 - nothing here adds a Spring context. Also exercised by hand against a live EJBCA: the
generated CSR carries key usage critical with the right bits and extended key usage as requested, and strict
validation rejects both an out-of-set bit and an unmapped structured extension.

`StructuredExtensionCodec` lives in `com.otilm.core.util` beside the renderer, not in
`com.otilm.core.certificate.request`: that package already depends on `core.util`, and the reverse edge would
close a package cycle.


src/test/java/com/otilm/core/certificate/request/CsrAttrsEncoderExtensionRequestTest.java
```diff
@@ -40,22 +40,25 @@ void requestsSubjectAltNameUnderExtensionRequest_whenSanMapped() throws Exceptio
...
     @Test
     void listsExtensionOidUnderExtensionRequest_whenExtensionMapped() throws Exception {
-        // given — an EXTENSION-mapped attribute (extended key usage, 2.5.29.37)
-        var ekuOid = "2.5.29.37";
-        var definitions = List.of(aMappedDataAttribute().withName("eku").mappingExtension(ekuOid).build());
+        // given — an EXTENSION-mapped attribute (basic constraints, 2.5.29.19). Not key usage or extended
+        // key usage: those have structured mapping targets, so the opaque route to them is rejected.
+        var extensionOid = "2.5.29.19";
+        var definitions = List
+                .of(aMappedDataAttribute().withName("basicConstraints").mappingExtension(extensionOid).build());
```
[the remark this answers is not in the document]
klaraf755: Confirmed and fixed. `CsrAttrsEncoder` filtered `ExtensionMappedField` only, so a structurally mapped key usage contributed no OID to `extensionRequest` - an EST client would submit a CSR without it and then fail validation for a requirement it was never told about. Structured targets now contribute their implied OID the same way SAN does a few lines below, and a test covers both.

Retargeting that fixture to basic constraints was still right on its own terms: it was demonstrating an opaque mapping to 2.5.29.37, which registration now rejects. The missing coverage was a separate gap, which is the useful half of this finding.

src/main/java/com/otilm/core/util/StructuredExtensionCodec.java
```diff
@@ -0,0 +1,196 @@
...
+            if ((bytes[index / 8] & 0x80 >>> index % 8) == 0) {
+                continue;
+            }
+            try {
+                usages.add(CertificateKeyUsage.fromIndex(index));
+            } catch (IllegalArgumentException e) {
+                unrepresentable.add("bit " + index);
+            }
+        }
+        return new Decoded<>(usages, unrepresentable);
```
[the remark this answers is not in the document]
klaraf755: Confirmed and fixed, for both shapes. An empty BIT STRING decoded to an empty set, which made the extension indistinguishable from absent: `structuredValues` returned empty, the whitelist pass skipped it, and the CSR went to the CA unfiltered. Same for an empty extended key usage sequence on the other line you flagged.

Both are malformed anyway - RFC 5280 4.2.1.3 requires at least one bit, and 4.2.1.12 a sequence of one or more purposes - so the codec now rejects them and the parser records them in `unrepresentableExtensionValues`, failing closed like an unsupported SAN. Tests at the codec and parser level.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -1798,6 +1801,11 @@ private static void validateMappedField(DataAttributeV3 attribute, MappedField f
             throw new AttributeException("fieldMapping field is missing fieldType", attribute.getUuid(),
                     attribute.getName(), attribute.getType(), connectorUuidStr);
         }
+        String structuredOid = StructuredExtensionCodec.oidFor(field);
+        if (structuredOid != null) {
+            validateStructuredMappedField(attribute, connectorUuidStr, structuredOid);
+            return;
```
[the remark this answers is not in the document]
klaraf755: Confirmed and fixed. `rejectDuplicateExtensionOids` inspected `ExtensionMappedField` only, so a definition declaring the same structured target twice saved cleanly and then threw on every non-empty projection - savable but unusable, which is exactly what that check exists to prevent. It now resolves the OID through `StructuredExtensionCodec.oidFor` as well.

My projector test covered two *attributes* both mapping key usage, not one attribute with two key-usage fields, which is why this slipped through. Test added at registration level.

src/main/java/com/otilm/core/certificate/request/CertificateRequestContentValidator.java
```diff
@@ -116,6 +126,12 @@ public static RequestAttributeValidationResult validate(List<? extends BaseAttri
             boolean required = v3.getProperties() != null && v3.getProperties().isRequired();
 
             for (MappedField field : v3.getFieldMapping().getFields()) {
+                String structuredOid = StructuredExtensionCodec.oidFor(field);
+                if (structuredOid != null) {
+                    mappedStructuredOids.add(structuredOid);
+                    validateStructuredTarget(v3, structuredOid, content, required, policy, result);
+                    continue;
+                }
```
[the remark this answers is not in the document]
klaraf755: Confirmed and fixed - this was the most serious of the four.

My backwards-compatibility test covered projection only, which is what the design promised in those words. But validation regressed: the parser diverts these OIDs out of `content.extensions`, so a legacy required mapping reported as missing and the structured whitelist flagged the CSR's value as unmapped. A strict profile carrying a legacy mapping would have started rejecting requests it previously accepted - a functional regression on upgrade, which is worse than the documented "re-saving requires the structured target".

A stored `ExtensionMappedField` on either OID is now recognised as a presence-only claim on the typed field: it satisfies `required`, it claims the OID so the whitelist leaves it alone, and it constrains nothing further. That last part is deliberate - its content is a base64 DER blob, not a list of codes, so there is nothing meaningful to compare decoded values against, and presence is all it ever enforced. Three tests cover the accepted, missing and unconstrained-value cases.

src/main/java/com/otilm/core/attribute/CertificateRequestAttributeProjector.java
```diff
@@ -86,6 +88,8 @@ public static X509RequestContent project(List<DataAttributeV3> definitions,
         content.setCertificateType(CertificateType.X509);
         content.setSubject(sink.subject.isEmpty() ? null : sink.subject);
         content.setSubjectAltNames(sink.subjectAltNames.isEmpty() ? null : sink.subjectAltNames);
+        content.setKeyUsage(sink.keyUsage.isEmpty() ? null : sink.keyUsage);
```
lubomirw, line 91: **[High]** | Correctness

Projected `keyUsage` and `extendedKeyUsage` exist only in typed `requestContent`. When the connector does not advertise `CERTIFICATE_REQUEST_STRUCTURED`, `RegisterWireBuilder` sends neither that content nor equivalent flat extensions, silently registering an identity without the requested usages.

_Suggested:_ DER-encode both typed lists into flat `CertificateExtension` entries using `StructuredExtensionCodec` and the renderer's criticality rules, then add a regression test for a non-structured connector.

<sub>flagged by Guided PR review, Superpowers, and Copilot CLI</sub>

src/main/java/com/otilm/core/util/StructuredExtensionCodec.java
```diff
@@ -0,0 +1,208 @@
...
+            purposes.add(new ASN1ObjectIdentifier(purposeOid));
+        }
+        return encoded(new DERSequence(purposes));
+    }
+
+    public static Decoded<CertificateKeyUsage> decodeKeyUsage(String base64Value) {
+        ASN1BitString bits = ASN1BitString.getInstance(parse(base64Value, KEY_USAGE_OID));
+        byte[] bytes = bits.getBytes();
+        int significantBits = bytes.length * 8 - bits.getPadBits();
+        if (significantBits <= 0) {
```
lubomirw: **[High]** | Security

`decodeKeyUsage` checks the BIT STRING's nominal significant length rather than whether any bit is set. A padded all-zero value such as `03 02 07 00` returns empty `values` and `unrepresentable` lists, making a present extension invisible to strict whitelist enforcement.

_Suggested:_ Track whether any bit is set and reject the value when none are; add codec, parser, and validator regressions for padded all-zero encodings.

<sub>flagged by Guided PR review and Copilot CLI</sub>

src/main/java/com/otilm/core/util/StructuredExtensionCodec.java
```diff
@@ -0,0 +1,208 @@
...
+            } catch (IllegalArgumentException e) {
+                throw new ValidationException("'%s' is not a key usage".formatted(code));
+            }
+        }
+        return usages;
+    }
+
+    public static List<String> toPurposeOids(List<String> oids) {
+        for (String oid : oids) {
+            if (!OidHandler.isOid(oid)) {
```
lubomirw, line 114: **[Medium]** | Error handling

`OidHandler.isOid` accepts values such as `1.40`, but `ASN1ObjectIdentifier` rejects them later during rendering. An extensible EKU list can therefore pass request validation and fail with an unchecked encoder exception.

_Suggested:_ Validate purpose OIDs with the ASN.1 parser at the projection boundary and translate constructor failures into a controlled `ValidationException`.

src/main/java/com/otilm/core/certificate/request/CertificateRequestContentValidator.java
```diff
@@ -210,6 +234,109 @@ private static <K, E> List<String> collectMatched(K key, Set<K> mappedKeys, List
         return values;
     }
 
+    /**
```
lubomirw: **[Nit]** | Maintainability

The `validateStructuredTarget` contract Javadoc is immediately followed by the Javadoc for `legacyStructuredOid`, so the first block is attached to no declaration.

_Suggested:_ Move the contract block directly above `validateStructuredTarget`.

<sub>flagged by Hygiene</sub>

lubomirw (changes requested):
Automated multi-reviewer triage - 5 findings (REQUEST_CHANGES).

The documentation finding cannot be anchored to this repository's diff:

- `documentation/docs/certificate-key/concept-design/core-components/request-attribute.md:23` - The guide still prescribes opaque Base64 DER for Extended Key Usage and says extension values are not constraint-checked, while Core now uses typed KU/EKU targets and validates their values. Extend OmniTrustILM/documentation#376 with permitted-set versus `extensibleList` semantics, strict whitelist behavior, unsupported-bit rejection, and the migration rule for stored opaque mappings. (Docs check: Guided PR review and Copilot CLI.)

src/main/java/com/otilm/core/attribute/CertificateRequestAttributeProjector.java
klaraf755: Real, but register-only — the finding is broader than the defect.

Issue, renew and rekey build a certification request, and `X509RequestContentRenderer` already encodes both typed lists into it via `addStructuredExtension` (lines 92–95). The capability flag never enters that path, because the extensions travel inside the CSR. Verified against a live EJBCA authority without the flag: the CSR carried keyUsage critical and EKU codeSigning.

Register is the one operation with no CSR, so it renders the flat wire — and there `renderFlatExtensions` read only `content.getExtensions()`, which the typed targets sit outside of. So they were dropped, exactly as described, but only for register.

Fixed as suggested: both lists are now DER-encoded into flat `CertificateExtension` entries. The flat wire can represent them, and did before the typed targets existed, so this restores prior behaviour rather than failing closed. Criticality comes from one shared helper, `X509RequestContentRenderer.structuredExtensionCritical`, so the CSR and flat forms cannot drift apart. No duplicate-OID risk — `ProjectionSink.seenExtensionOids` already rejects a typed target and an opaque mapping claiming the same OID.

Two regressions added: a non-structured connector with both lists populated, and one with them empty (extensions must stay unset rather than become an empty list).

src/main/java/com/otilm/core/util/StructuredExtensionCodec.java
klaraf755: Confirmed. The guard was testing the wrong quantity.

`significantBits` was `bytes.length * 8 - getPadBits()`, rejected only when `<= 0`. For `03 02 07 00` that is 1, so the guard passed; the loop then found no set bit and returned empty `values` and `unrepresentable` lists — precisely the state the guard's own comment says is impossible.

The check now runs after the loop, on whether anything was actually decoded, so it catches a nominally non-empty but all-zero string at any width. Regressions added for the single-octet (`03 02 07 00`) and multi-octet (`03 03 01 00 00`) forms.

src/main/java/com/otilm/core/util/StructuredExtensionCodec.java
klaraf755: Confirmed, and probed against BouncyCastle 1.84 rather than taken on trust: `1.40`, `1.40.1` and `0.99` are rejected; `2.999.1` and `1.2.3` are accepted.

The regex `^[0-2](\.(0|[1-9]\d{0,38})){1,127}$` cannot express the rule that a first arc of 0 or 1 caps the second arc at 39, so an `extensibleList` EKU purpose could clear request validation and then throw `IllegalArgumentException` unchecked inside `encodeExtendedKeyUsage` — the encoder's `encoded()` helper catches only `IOException`.

`toPurposeOids` now validates with the same parser that will do the encoding, translating a failure into `ValidationException`. Regression added for `1.40`.

src/main/java/com/otilm/core/certificate/request/CertificateRequestContentValidator.java
klaraf755: Confirmed — two Javadoc blocks were stacked, so the contract block documented nothing and `legacyStructuredOid`'s own block won. Moved directly above `validateStructuredTarget`.

lubomirw (dismissed):

lubomirw (approved):

PR #2147, opened 2026-08-25
Serve the cryptographic asset inventory contract with a 501 stub

vyskocilm:
Depends on OmniTrustILM/interfaces#913 (and the contract from interfaces#909). CI here will fail to compile until that merges and the interfaces snapshot is published.

## What

- `CryptographicAssetControllerImpl` — implements `/v1/cryptoAssets` list, detail and searchable-fields.
- `CryptographicAssetExternalService` + `CryptographicAssetServiceImpl` — every operation throws `NotSupportedException` (HTTP 501 via `ExceptionHandlingAdvice`).
- `StatisticsControllerImpl.getCryptographicAssetStatistics()` — same refusal; core did not compile without it once #909 landed.
- `com.otilm.core.model.auth.Resource` — adds `CRYPTO_ASSET("cryptoAssets")`.

## Why a service layer rather than throwing in the controller

The authorization annotations are the part that actually works. `@ExternalAuthorization(resource = CRYPTO_ASSET, action = LIST/DETAIL)` registers the resource and its actions with the auth service on context refresh, so roles can be defined and the frontend wired against the real gates while the read model is built. An unpermitted caller gets 403 from the aspect ahead of the 501. A controller-only stub would leave the endpoints ungated and the resource actionless in the role editor.

The statistics operation shares the `LIST` action rather than carrying a gate of its own, per the contract's Javadoc.

## The auth-sync enum

Core carries a second resource enum (`com.otilm.core.model.auth.Resource`) that `ContextRefreshListener` resolves codes against. Without the `CRYPTO_ASSET` entry, context refresh fails at startup with `Unknown Resource Name cryptoAssets` — caught by booting the context in the new test, not by compilation.

## Not in scope

The inventory read model itself: the projection over stored CBOM documents, its filter fields and the dashboard aggregates. The new test's refusal assertions are what should fail first when that lands.

## Verification

- 2652 unit tests, 0 failures; all ArchUnit rules pass (including the external-service authorization rules)
- New `CryptographicAssetControllerStubITest`: boots the context, pins the 501 on all four operations and the `list`/`detail` registration
- `spotless:check` clean
- Built against interfaces #913 installed locally as 2.20.0-SNAPSHOT

### DIFF
--- a/src/main/java/com/otilm/core/service/CryptographicAssetExternalService.java
+++ b/src/main/java/com/otilm/core/service/CryptographicAssetExternalService.java
@@ -1,0 +1,61 @@
+package com.otilm.core.service;
+
+import com.otilm.api.exception.NotFoundException;
+import com.otilm.api.model.client.certificate.SearchRequestDto;
+import com.otilm.api.model.client.dashboard.CryptographicAssetStatisticsDto;
+import com.otilm.api.model.common.PaginationResponseDto;
+import com.otilm.api.model.core.cryptoasset.CryptographicAssetDetailDto;
+import com.otilm.api.model.core.cryptoasset.CryptographicAssetDto;
+import com.otilm.api.model.core.search.SearchFieldDataByGroupDto;
+import com.otilm.core.security.authz.SecuredUUID;
+import com.otilm.core.security.authz.SecurityFilter;
+import java.util.List;
+
+/**
+ * Read side of the cryptographic asset inventory — the deduplicated cross-CBOM asset view served under
+ * {@code /v1/cryptoAssets}. Assets enter the inventory through the CBOM document sync, so this service has no write
+ * operations; the sync itself stays on the CBOM services.
+ *
+ * <p>
+ * Every operation currently rejects with {@link com.otilm.api.exception.NotSupportedException}: the API contract is
+ * ratified (interfaces#909, interfaces#913) but the inventory projection and its persistence are not built yet. The
+ * authorization annotations on the implementation are live, so the resource and its actions reach the auth service
+ * before the data does and the endpoints answer 403 to an unpermitted caller rather than 501.
+ */
+public interface CryptographicAssetExternalService {
+
+    /**
+     * List one page of the deduplicated cross-CBOM asset inventory, narrowed by the filters in the request.
+     *
+     * @param filter security filter narrowing the result to the caller's permitted objects
+     * @param request search request carrying the filters, sort and paging
+     * @return one page of matching assets
+     */
+    PaginationResponseDto<CryptographicAssetDto> listCryptographicAssets(SecurityFilter filter,
+            SearchRequestDto request);
+
+    /**
+     * Retrieve one asset with its verdict provenance, normalized properties, per-source payloads and recorded object
+     * identifiers.
+     *
+     * @param uuid secured unique identifier of the asset
+     * @return the asset detail
+     * @throws NotFoundException if no asset with the given UUID is in the inventory
+     */
+    CryptographicAssetDetailDto getCryptographicAsset(SecuredUUID uuid) throws NotFoundException;
+
+    /**
+     * Fields the list operation accepts in its filters, grouped by field source.
+     *
+     * @return the searchable field definitions
+     */
+    List<SearchFieldDataByGroupDto> getSearchableFieldInformationByGroup();
+
+    /**
+     * Count badges, distribution maps and the sync-completeness block for the inventory dashboard. Shares the list
+     * permission with {@link #listCryptographicAssets} rather than carrying a gate of its own.
+     *
+     * @return the dashboard statistics
+     */
+    CryptographicAssetStatisticsDto getCryptographicAssetStatistics();
+}
--- a/src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
@@ -1,0 +1,61 @@
+package com.otilm.core.service.impl;
+
+import com.otilm.api.exception.NotFoundException;
+import com.otilm.api.exception.NotSupportedException;
+import com.otilm.api.model.client.certificate.SearchRequestDto;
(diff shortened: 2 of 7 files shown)


lubomirw (dismissed):

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -0,0 +1,61 @@
...
+import com.otilm.core.model.auth.ResourceAction;
+import com.otilm.core.security.authz.ExternalAuthorization;
+import com.otilm.core.security.authz.SecuredUUID;
+import com.otilm.core.security.authz.SecurityFilter;
+import com.otilm.core.service.CryptographicAssetExternalService;
+import java.util.List;
+import org.springframework.stereotype.Service;
+
+/**
+ * Serves the ratified cryptographic asset inventory contract with a refusal on every operation: the inventory
```
ivosh: **~ Minor** | Documentation

This PR independently explains the missing projection, live authorization, and 403-before-501 behavior in CryptographicAssetExternalService.java:20-23, CryptographicAssetServiceImpl.java:21-30, and CryptographicAssetControllerStubITest.java:20-26. When the projection lands, these differently worded copies must all change; the implementation is the authoritative home for this temporary behavior.

_Suggested:_ Keep the staging rationale on `CryptographicAssetServiceImpl`, reduce the interface documentation to its stable contract, and limit the test documentation to the invariant it verifies.

src/test/java/com/otilm/core/integration/api/web/CryptographicAssetControllerStubITest.java
```diff
@@ -0,0 +1,62 @@
...
+import com.otilm.core.util.BaseSpringBootTest;
+import java.util.UUID;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * The cryptographic asset inventory endpoints exist so the contract ratified in interfaces#909 is served by a real
```
ivosh: **~ Minor** | Documentation

CryptographicAssetControllerStubITest.java:20 and CryptographicAssetExternalService.java:21 cite `interfaces#909` and `interfaces#913`, but a repository-wide search finds no matching in-tree specification or plan. These references are the comments' evidence that the stub is deliberate, yet a future maintainer cannot resolve them from the repository.

_Suggested:_ Remove the issue-number references and state the relevant contract constraint directly, or add a resolvable in-repository design document.

<sub>flagged by Maintainability & tests + Noise & hygiene</sub>

src/test/java/com/otilm/core/integration/api/web/CryptographicAssetControllerStubITest.java
```diff
@@ -0,0 +1,62 @@
...
+    private CryptographicAssetController cryptographicAssetController;
+
+    @Autowired
+    private StatisticsController statisticsController;
+
+    @Autowired
+    private ContextRefreshListener contextRefreshListener;
+
+    @Test
+    void everyInventoryOperationRefusesAsNotImplemented() {
```
ivosh: **. Nitpick** | Testing

CryptographicAssetControllerStubITest.java:40 executes four independent endpoint operations in one test, so the first failure prevents the remaining operations from running and the report does not identify the affected endpoint. The sibling SigningRecordStatisticsControllerITest.java:19-38 keeps each endpoint behavior in a separate test method.

_Suggested:_ Split the four refusal checks into one test method per controller operation.

src/main/java/com/otilm/core/service/CryptographicAssetExternalService.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.List;
+
+/**
+ * Read side of the cryptographic asset inventory — the deduplicated cross-CBOM asset view served under
+ * {@code /v1/cryptoAssets}. Assets enter the inventory through the CBOM document sync, so this service has no write
+ * operations; the sync itself stays on the CBOM services.
+ *
+ * <p>
+ * Every operation currently rejects with {@link com.otilm.api.exception.NotSupportedException}: the API contract is
+ * ratified (interfaces#909, interfaces#913) but the inventory projection and its persistence are not built yet. The
```
ivosh: **. Nitpick** | Documentation

CryptographicAssetExternalService.java:21 records transient `interfaces#909` and `interfaces#913` references in the API documentation. Those issue IDs describe delivery history rather than a durable service constraint and will become stale once the projection is implemented.

_Suggested:_ Remove the issue references while retaining the durable authorization-before-data rationale.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -0,0 +1,61 @@
...
+@Service
+public class CryptographicAssetServiceImpl implements CryptographicAssetExternalService {
+
+    private static final String NOT_IMPLEMENTED = "Cryptographic asset inventory is not implemented yet";
+
+    @Override
+    @ExternalAuthorization(resource = Resource.CRYPTO_ASSET, action = ResourceAction.LIST)
+    public PaginationResponseDto<CryptographicAssetDto> listCryptographicAssets(SecurityFilter filter,
+            SearchRequestDto request) {
+        throw new NotSupportedException(NOT_IMPLEMENTED);
```
ivosh: **~ Minor** | Documentation

CryptographicAssetServiceImpl.java:41, :47, :53, and :59 always throw `NotSupportedException`, which `ExceptionHandlingAdvice` renders as HTTP 501 with `ErrorMessageDto`. The interfaces#913 contracts advertise only 200/422 for listing, 200/404 for detail, and 200 for search/statistics, so every permitted request currently receives an undocumented response.

_Suggested:_ Add a 501 `ApiResponse` with `ErrorMessageDto` to all four interfaces#913 contract methods for the stub period.

Issue #2130, opened 2026-08-23
Serialise discovery agenda claims and processing batches


lubomirw:
### Description

`DiscoveryProcessTickWorker` selects its batch with a plain read — `findByDiscoveryUuidAndNewlyDiscoveredTrueAndProcessedFalseAndProcessedErrorIsNullOrderByCreatedAsc` — which orders and bounds the rows but does not claim them. Two ticks against the same run therefore select the same rows and run the whole import pipeline over them twice.

OmniTrustILM/core#1963 narrowed the window rather than closing it. The agenda row is now parked one backstop interval out instead of due-now, so the sweep no longer publishes a competing tick alongside the direct continuation. Two paths remain:

- a batch that runs longer than `discovery.work.continuation-backstop` (default 1 minute), after which the sweep legitimately re-publishes while the first tick is still working
- duplicate JMS delivery, which the backstop does not address at all

The listener runs at `provider-discovery-work: 1-3` consumers per pod, so two ticks for one run can be in flight across pods simultaneously; the claim query takes no lock or lease, so both select the same rows.

A second bound is missing on the same path, and belongs here for the same reason. `DiscoveryProcessTickWorker` deliberately takes a content group whole even when its row count exceeds `discovery.processing.batch-size`, because splitting it would run that certificate's triggers, event histories and validation once per page. The batch bound is therefore exact for ordinary groups and best-effort for a single oversized one: a certificate reported at very many locations materialises its whole group in memory, limited only by how many locations it was found at. That trade-off is deliberate today and only resolvable alongside the claim — an explicit maximum group size needs bulk handling for groups above it, and bulk handling needs rows that can be claimed without repeating the group's side effects.

The same shape appears on the `discovery_work` agenda itself, and belongs with this fix rather than beside it. `DiscoveryWorkWriter.resetAttempt` refreshes a row's attempt counter without taking the sweep's advisory lock, so a claimer that has already read the old value can overwrite the reset — a healthy run then keeps climbing toward termination. And the workers trust the `attempt` carried by their message: a redelivered old tick with a high attempt can end a run that has since recovered, while one with a low attempt lowers the persisted counter so the budget never exhausts. Both violate the consecutive-unanswered-ticks rule the budget is defined by.

[...]

PR #2155, opened 2026-08-26
Store discovery run messages as aggregated rows

### LINKED ISSUE #2127: Store discovery run messages as aggregated rows
### Description

`discovery.run_messages` is a JSONB array of strings, appended by read-modify-write under the run row's write lock and capped at 200 entries by dropping the oldest. Three properties make that the wrong shape for a run that can last days:

- Every append rewrites the whole value. Past the TOAST threshold that is a full out-of-line rewrite per message, and the `PROCESS` worker appends after every batch — 5,000 appends on a large degraded run.
- Correctness rests on every caller remembering `findWithLockByUuid`. Forget it and a message is silently lost to a concurrent append; take it and message writes queue behind the drain, which holds the same lock across staging an entire page.
- Dropping the oldest inverts the artifact's purpose. An operator opening a WARNING finds 200 near-identical tail sentences and no trace of the root cause at the start of the run.
- A flat list of strings carries no severity, so the only question the terminal decision can ask it is whether it is empty. `DiscoveryProcessTickWorker.finish` ends a run `WARNING` on exactly that, which means one batch that failed transiently and was reimported cleanly on the next tick still downgrades a run whose every row succeeded.

Replace it with a child table holding one row per *distinct* message, with an occurrence counter so repeated problems aggregate instead of accumulating.

**Columns.** `id BIGINT GENERATED ALWAYS AS IDENTITY` (ordering only — never exposed), `discovery_uuid` FK `ON DELETE CASCADE`, `severity`, `code`, `message`, `occurrences`, `first_seen_at`, `last_seen_at`.

`code` identifies the *kind* of problem and is what dedup and the overflow rule group on. Connector `ERROR` events supply their own (`DiscoveryErrorEvent.code`); Core assigns codes to its own producers — the certificate staging failure, each per-batch gap kind, and the terminal reason. `severity` is stored rather than derived: Core has no registry of a connector's codes and cannot map one to a level.

Follows OmniTrustILM/core#1963, which added the producers. API exposure is OmniTrustILM/core#1964 and its interfaces contract change.

### Definition of Done

- [ ] Table created as above, unique on `(discovery_uuid, code, md5(message))`
- [ ] Appends are `INSERT … ON CONFLICT DO UPDATE SET occurrences = occurrences + :n, last_seen_at = …` — no run-row lock, so a lost update is impossible by construction
- [ ] Ordering is the identity column, never a timestamp: `now()` is transaction-start time, so every message written in one tick ties, and `@CreationTimestamp` inverts across pods under clock skew
- [ ] Bounds enforced in one writer bean: distinct messages per `(run, code)`, rows per run, characters per message; overflow folds into a per-code suppression row carrying the count. Fixed constants rather than configuration — they are safety limits against a producer or a connector that misbehaves, and no deployment holds information that would tell it to move one
- [ ] Overflow drops the newest and keeps the oldest, and the terminal reason is exempt so a run's ending always lands
- [ ] `DiscoveryRunCounts.describeGaps()` emits `(code, count)` pairs so per-batch summaries aggregate into one row per kind, instead of each batch minting a distinct string by embedding its own count; v1's status message keeps its rendered sentences
- [ ] Four call sites rewired; `run_messages` column, `DiscoveryRunLifecycle.append` and `DiscoveryWriter.appendRunMessages` removed
- [ ] The terminal status decision reads severity rather than non-emptiness, so a recovered problem no longer ends a clean run `WARNING`
- [ ] ITests: upsert aggregation across batches, each bound, ordering under same-transaction writes, cascade delete

lubomirw:
## TLDR

`discovery.run_messages` was a JSONB array on the run row: rewritten whole under that row's write lock on every append, and capped at 200 by dropping the oldest — which inverted its purpose, since an operator opening a `WARNING` found 200 near-identical tail sentences and no trace of what started the run.

This replaces it with a `discovery_message` child table holding one row per *distinct* problem plus an occurrence counter. Appends are an upsert that takes no run-row lock. The terminal status decision now reads message **severity** instead of whether the log is non-empty.

Closes #2127.

> **Does not build until OmniTrustILM/interfaces#912 merges.** `DiscoveryMessageSeverity` comes from there, and `DiscoveryController.getDiscoveryRunMessages` is declared there — served here by the existing 501 stub block, since the listing itself belongs to #1964.

## What changed

- **Schema.** `discovery_message`: `id BIGINT GENERATED ALWAYS AS IDENTITY`, FK `ON DELETE CASCADE`, unique on `(discovery_uuid, code, message_hash)`. `run_messages` is dropped without backfill — it was added in this same unreleased cycle.
- **Writes.** One writer bean. `INSERT … ON CONFLICT DO UPDATE SET occurrences = occurrences + n`, so two appenders add their counts rather than one overwriting the other, and no caller has to take the run row to be safe.
- **Producers.** Per-batch gap summaries carry their counts in `occurrences` rather than in the text, and staging failures are named by kind rather than by which certificate hit them, so repeats aggregate instead of accumulating. Per-certificate detail keeps its existing home on `discovery_certificate.processed_error`.
- **Ending.** A run ends `COMPLETED` unless a row carries a reason of its own or the run collected a `WARNING`/`ERROR` message. A batch that failed transiently and was reimported cleanly is recorded at `INFO` and no longer downgrades the run.

## Three things worth a reviewer's attention

1. **Ordering is the identity column, never a timestamp.** `now()` is transaction-start time, so every message one tick writes ties, and a wall-clock column inverts across pods. An ITest asserts the timestamps *do* tie, which is what makes that real rather than asserted.
2. **`message_hash` is a stored generated column** (`md5(message)`), not an expression index. ITests build their schema from the entities (`ddl-auto: create-drop`, Flyway off) and an expression index cannot be mapped, so `ON CONFLICT` would have no matching constraint in any ITest.
3. **The writer is `REQUIRED` throughout, never `REQUIRES_NEW`.** Most callers append while holding the run row's `SELECT … FOR UPDATE`, and the inserted row takes `FOR KEY SHARE` on that same row through its FK. From a separate transaction that wait can never be satisfied — the run would hang rather than fail.

## Bounds are constants, not configuration

The issue originally asked for these to be configurable. Two of the three are unreachable in practice and none has an operator tuning story, so they ship as constants in the writer. #2127's Definition of Done has been updated to match, rather than left disagreeing with the code. It is a decision rather than an omission: an estate with more than fifty kinds of problem has no operator remedy, but the log is advisory, a suppression row says data was dropped, and `MAX_PER_CODE <= MAX_PER_RUN` is an invariant a dial would let an operator break.

## Verification

- 824 tests green (`Discovery*`, `Certificate*`, `*ArchTest`, `ContextSignatureGuardTest`); checkstyle and spotless clean.
- The migration was applied against a throwaway PostgreSQL 17 to check the DDL, the upsert's aggregation, the generated hash column and the FK cascade.

### DIFF
--- a/src/main/java/com/otilm/core/dao/repository/DiscoveryMessageRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/DiscoveryMessageRepository.java
@@ -23,6 +23,17 @@ public interface DiscoveryMessageRepository extends JpaRepository<DiscoveryMessa
      * new rows only — a repeat always lands, however many rows the run already has, because aggregating onto an
      * existing row grows nothing.
      *
+     * <p>
+     * The window widens, never moves. {@code now()} is transaction-start time, so a transaction that began earlier can
+     * commit later and would otherwise rewind {@code last_seen_at} — past {@code first_seen_at} entirely, when it
+     * aggregates onto a row a shorter, later transaction created. {@code LEAST}/{@code GREATEST} make both ends
+     * monotonic, which is what lets the pair be read as the span a fault was active.
+     *
+     * <p>
+     * The bounds are soft ceilings, not guarantees: the counts and the insert are not atomic against each other, so
+     * appenders that each read a count one below the limit all pass and a run can exceed it by the number of concurrent
+     * appenders. Locking to close that would reintroduce the serialisation this design removes.
+     *
      * @return 1 when the message was recorded, 0 when a bound refused it a row of its own
      */
     @Modifying
@@ -41,7 +52,9 @@ public interface DiscoveryMessageRepository extends JpaRepository<DiscoveryMessa
                    AND (SELECT count(*) FROM {h-schema}discovery_message
                         WHERE discovery_uuid = CAST(:discoveryUuid AS UUID)) < :maxPerRun)
             ON CONFLICT (discovery_uuid, code, message_hash) DO UPDATE
-                SET occurrences = m.occurrences + EXCLUDED.occurrences, last_seen_at = EXCLUDED.last_seen_at
+                SET occurrences = m.occurrences + EXCLUDED.occurrences,
+                    first_seen_at = LEAST(m.first_seen_at, EXCLUDED.first_seen_at),
+                    last_seen_at = GREATEST(m.last_seen_at, EXCLUDED.last_seen_at)
             """, nativeQuery = true)
     int appendWithinBounds(@Param("discoveryUuid") UUID discoveryUuid, @Param("severity") String severity,
             @Param("code") String code, @Param("message") String message, @Param("occurrences") long occurrences,
@@ -58,14 +71,16 @@ public interface DiscoveryMessageRepository extends JpaRepository<DiscoveryMessa
             VALUES (CAST(:discoveryUuid AS UUID), CAST(:severity AS VARCHAR), CAST(:code AS VARCHAR),
                     CAST(:message AS VARCHAR), CAST(:occurrences AS BIGINT), now(), now())
             ON CONFLICT (discovery_uuid, code, message_hash) DO UPDATE
-                SET occurrences = m.occurrences + EXCLUDED.occurrences, last_seen_at = EXCLUDED.last_seen_at
+                SET occurrences = m.occurrences + EXCLUDED.occurrences,
+                    first_seen_at = LEAST(m.first_seen_at, EXCLUDED.first_seen_at),
+                    last_seen_at = GREATEST(m.last_seen_at, EXCLUDED.last_seen_at)
             """, nativeQuery = true)
     void append(@Param("discoveryUuid") UUID discoveryUuid, @Param("severity") String severity,
             @Param("code") String code, @Param("message") String message, @Param("occurrences") long occurrences);
 
     /**
-     * Whether the run collected anything at or above a given severity — what the terminal decision asks, in place of
-     * the old "is the log non-empty", which downgraded a run over a problem that had since been recovered.
+     * Whether the run collected anything at or above a given severity — what the terminal decision asks, so a run that
+     * recovered from a problem it recorded is not downgraded for it.
      */
     boolean existsByDiscoveryUuidAndSeverityIn(UUID discoveryUuid, Collection<DiscoveryMessageSeverity> severities);
 
--- a/src/main/java/com/otilm/core/dao/repository/DiscoveryMessageRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/DiscoveryMessageRepository.java
@@ -24,6 +24,11 @@ public interface DiscoveryMessageRepository extends JpaRepository<DiscoveryMessa
      * existing row grows nothing.
      *
      * <p>
+     * A row's severity is the highest ever recorded for it, which is what stops the run-level suppression row —
+     * covering every code at once — understating the worst it stands for. The {@code CASE} compares the member names
+     * {@code @Enumerated(STRING)} stores, so renaming a constant means changing this too.
+     *
+     * <p>
      * The window widens, never moves. {@code now()} is transaction-start time, so a transaction that began earlier can
      * commit later and would otherwise rewind {@code last_seen_at} — past {@code first_seen_at} entirely, when it
      * aggregates onto a row a shorter, later transaction created. {@code LEAST}/{@code GREATEST} make both ends
@@ -53,6 +58,10 @@ public interface DiscoveryMessageRepository extends JpaRepository<DiscoveryMessa
                         WHERE discovery_uuid = CAST(:discoveryUuid AS UUID)) < :maxPerRun)
             ON CONFLICT (discovery_uuid, code, message_hash) DO UPDATE
                 SET occurrences = m.occurrences + EXCLUDED.occurrences,
+                    severity = CASE
+                        WHEN 'ERROR' IN (m.severity, EXCLUDED.severity) THEN 'ERROR'
+                        WHEN 'WARNING' IN (m.severity, EXCLUDED.severity) THEN 'WARNING'
+                        ELSE m.severity END,
                     first_seen_at = LEAST(m.first_seen_at, EXCLUDED.first_seen_at),
                     last_seen_at = GREATEST(m.last_seen_at, EXCLUDED.last_seen_at)
             """, nativeQuery = true)
@@ -72,6 +81,10 @@ public interface DiscoveryMessageRepository extends JpaRepository<DiscoveryMessa
                     CAST(:message AS VARCHAR), CAST(:occurrences AS BIGINT), now(), now())
             ON CONFLICT (discovery_uuid, code, message_hash) DO UPDATE
                 SET occurrences = m.occurrences + EXCLUDED.occurrences,
+                    severity = CASE
+                        WHEN 'ERROR' IN (m.severity, EXCLUDED.severity) THEN 'ERROR'
+                        WHEN 'WARNING' IN (m.severity, EXCLUDED.severity) THEN 'WARNING'
+                        ELSE m.severity END,
                     first_seen_at = LEAST(m.first_seen_at, EXCLUDED.first_seen_at),
                     last_seen_at = GREATEST(m.last_seen_at, EXCLUDED.last_seen_at)
             """, nativeQuery = true)
--- a/src/main/java/com/otilm/core/dao/repository/DiscoveryMessageRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/DiscoveryMessageRepository.java
@@ -1,0 +1,76 @@
+package com.otilm.core.dao.repository;
+
+import com.otilm.api.model.core.discovery.DiscoveryMessageSeverity;
+import com.otilm.core.dao.entity.DiscoveryMessage;
+import java.util.Collection;
+import java.util.List;
+import java.util.UUID;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Modifying;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.query.Param;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface DiscoveryMessageRepository extends JpaRepository<DiscoveryMessage, Long> {
+
+    /**
+     * Records one occurrence group of a problem, aggregating onto the row already carrying it.
+     *
+     * <p>
+     * No lock is taken anywhere: a concurrent append of the same problem loses the race to the unique index and lands
+     * in the {@code DO UPDATE}, so the two counts add rather than one overwriting the other. The bound is a guard on
+     * new rows only — a repeat always lands, however many rows the run already has, because aggregating onto an
+     * existing row grows nothing.
+     *
+     * @return 1 when the message was recorded, 0 when a bound refused it a row of its own
+     */
+    @Modifying
+    @Query(value = """
+            INSERT INTO {h-schema}discovery_message AS m
+                (discovery_uuid, severity, code, message, occurrences, first_seen_at, last_seen_at)
+            SELECT CAST(:discoveryUuid AS UUID), CAST(:severity AS VARCHAR), CAST(:code AS VARCHAR),
+                   CAST(:message AS VARCHAR), CAST(:occurrences AS BIGINT), now(), now()
+            WHERE EXISTS (SELECT 1 FROM {h-schema}discovery_message
+                          WHERE discovery_uuid = CAST(:discoveryUuid AS UUID)
+                            AND code = CAST(:code AS VARCHAR)
+                            AND message_hash = md5(CAST(:message AS TEXT)))
+               OR ((SELECT count(*) FROM {h-schema}discovery_message
+                    WHERE discovery_uuid = CAST(:discoveryUuid AS UUID)
+                      AND code = CAST(:code AS VARCHAR)) < :maxPerCode
+                   AND (SELECT count(*) FROM {h-schema}discovery_message
+                        WHERE discovery_uuid = CAST(:discoveryUuid AS UUID)) < :maxPerRun)
+            ON CONFLICT (discovery_uuid, code, message_hash) DO UPDATE
+                SET occurrences = m.occurrences + EXCLUDED.occurrences, last_seen_at = EXCLUDED.last_seen_at
+            """, nativeQuery = true)
+    int appendWithinBounds(@Param("discoveryUuid") UUID discoveryUuid, @Param("severity") String severity,
+            @Param("code") String code, @Param("message") String message, @Param("occurrences") long occurrences,
+            @Param("maxPerCode") int maxPerCode, @Param("maxPerRun") int maxPerRun);
+
+    /**
+     * As {@link #appendWithinBounds}, for the two messages that must land whatever the run has already collected: the
+     * reason a run ended, and the row saying that other messages were dropped.
+     */
+    @Modifying
+    @Query(value = """
+            INSERT INTO {h-schema}discovery_message AS m
+                (discovery_uuid, severity, code, message, occurrences, first_seen_at, last_seen_at)
+            VALUES (CAST(:discoveryUuid AS UUID), CAST(:severity AS VARCHAR), CAST(:code AS VARCHAR),
+                    CAST(:message AS VARCHAR), CAST(:occurrences AS BIGINT), now(), now())
+            ON CONFLICT (discovery_uuid, code, message_hash) DO UPDATE
+                SET occurrences = m.occurrences + EXCLUDED.occurrences, last_seen_at = EXCLUDED.last_seen_at
+            """, nativeQuery = true)
+    void append(@Param("discoveryUuid") UUID discoveryUuid, @Param("severity") String severity,
+            @Param("code") String code, @Param("message") String message, @Param("occurrences") long occurrences);
+
+    /**
+     * Whether the run collected anything at or above a given severity — what the terminal decision asks, in place of
+     * the old "is the log non-empty", which downgraded a run over a problem that had since been recovered.
+     */
+    boolean existsByDiscoveryUuidAndSeverityIn(UUID discoveryUuid, Collection<DiscoveryMessageSeverity> severities);
+
+    /** The run's log, oldest first. Ordered by the identity column; see {@code DiscoveryMessage.id}. */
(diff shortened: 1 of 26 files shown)


src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProcessTickWorker.java
```diff
@@ -143,19 +158,49 @@ private long backlogOf(UUID discoveryUuid) {
...
                     .error("Processing batch {} of discovery {} did not complete: {}", attempt, run.getUuid(),
                             e.getMessage(), e);
-            discoveryWriter
-                    .appendRunMessages(run.getUuid(),
-                            List.of("A batch of discovered certificates could not be processed."));
+            // INFO only while the rows are still in the backlog for another tick to retry: rows that run out of
+            // attempts carry their own reason, and a stall that exhausts the budget ends the run itself. A batch
+            // that threw after stamping its last row leaves nothing to retry, so no later tick revisits the
+            // failure -- and an ending decided on severity would read that run as clean.
+            boolean retryable = backlogOf(run.getUuid()) > 0;
```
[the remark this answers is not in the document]
lubomirw: Right, and my fix only covered the single-batch case — which is exactly what its test exercised (one certificate at a batch size of one), so it passed while leaving the hole you describe.

Fixed in cde670714 by counting only the rows of the batch that failed, through a new `countByUuidInAndNewlyDiscoveredTrueAndProcessedFalseAndProcessedErrorIsNull`. Rows the pipeline stamped with a reason stay excluded on purpose: they carry their own, so `decideEnding` already ends the run `WARNING` for them.

The multi-batch case now has its own test — three rows at a batch size of one, the first batch stamping its row and then throwing — asserting `batchProcessingAbandoned` rather than `batchProcessingFailed`, and `WARNING` once the remaining batches import cleanly. It fails against the previous predicate.

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProcessTickWorker.java
```diff
@@ -232,9 +277,11 @@ private Ending decideEnding(Discovery run) {
             return null;
         }
         // Run-level gaps (a failed bookkeeping write, validation never requested) leave every row clean, so the
-        // message log is checked too.
+        // message log is checked too -- by severity, not by whether it holds anything. A run collects messages for
+        // things it recovered from, and ending a run whose every row imported on the strength of one of those
+        // would report a warning about nothing an operator can act on.
         boolean rowsFailed = certificateRepository.existsByDiscoveryUuidAndProcessedErrorIsNotNull(run.getUuid());
-        boolean runLevelGaps = run.getRunMessages() != null && !run.getRunMessages().isEmpty();
+        boolean runLevelGaps = messageRepository.existsByDiscoveryUuidAndSeverityIn(run.getUuid(), UNRECOVERED);
```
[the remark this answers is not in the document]
lubomirw: Agreed on the mechanism, and it predates this PR rather than arriving with it. On `main` the same window exists: `decideEnding` read `run.getRunMessages()` from the locked entity, and an appender had to take that same run lock — so a terminator that won the lock committed `COMPLETED` before the appender's write landed, and the message ended up on a terminal run. Moving the log to its own table changed what blocks (FK `KEY SHARE` against `FOR UPDATE`) but not the shape or the outcome.

Closing it means coordinating the terminal decision with in-flight batches, which is the claim design in OmniTrustILM/core#2130 — the same issue that owns two ticks selecting the same rows. It now carries this case explicitly: "The terminal decision refuses to end a run while a batch claim is outstanding, so a run cannot complete between another tick's row commit and its gap record."

Leaving it there rather than growing this PR into the claim work.

src/main/java/com/otilm/core/dao/repository/DiscoveryMessageRepository.java
```diff
@@ -0,0 +1,91 @@
...
+    @Modifying
+    @Query(value = """
+            INSERT INTO {h-schema}discovery_message AS m
+                (discovery_uuid, severity, code, message, occurrences, first_seen_at, last_seen_at)
+            VALUES (CAST(:discoveryUuid AS UUID), CAST(:severity AS VARCHAR), CAST(:code AS VARCHAR),
+                    CAST(:message AS VARCHAR), CAST(:occurrences AS BIGINT), now(), now())
+            ON CONFLICT (discovery_uuid, code, message_hash) DO UPDATE
+                SET occurrences = m.occurrences + EXCLUDED.occurrences,
+                    first_seen_at = LEAST(m.first_seen_at, EXCLUDED.first_seen_at),
+                    last_seen_at = GREATEST(m.last_seen_at, EXCLUDED.last_seen_at)
```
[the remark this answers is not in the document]
lubomirw: Fixed in cde670714, on both upserts rather than only the unbounded one, so the invariant is a single sentence — a row's severity is the highest ever recorded for it — instead of a rule with an exception a future producer could trip over.

```sql
severity = CASE
    WHEN 'ERROR' IN (m.severity, EXCLUDED.severity) THEN 'ERROR'
    WHEN 'WARNING' IN (m.severity, EXCLUDED.severity) THEN 'WARNING'
    ELSE m.severity END
```

The literals are the member names `@Enumerated(STRING)` stores, which the javadoc now flags, since renaming a constant would otherwise break this silently. `standInSeverity`'s "may understate an ERROR it covers" caveat is gone — it stopped being true.

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProcessTickWorker.java
```diff
@@ -139,23 +154,60 @@ private long backlogOf(UUID discoveryUuid) {
...
                     .error("Processing batch {} of discovery {} did not complete: {}", attempt, run.getUuid(),
                             e.getMessage(), e);
-            discoveryWriter
-                    .appendRunMessages(run.getUuid(),
-                            List.of("A batch of discovered certificates could not be processed."));
+            // INFO only while this batch's own rows are still pending for another tick to retry: rows that run out
+            // of attempts carry their own reason, and a stall that exhausts the budget ends the run itself. Judged
+            // on this batch and never the run, because another batch's pending rows would make a failure that
+            // stamped all of its own look retryable -- and nothing would ever revisit it.
+            boolean retryable = pendingIn(batch) > 0;
```
klaraf755: `retryable = pendingIn(batch) > 0` can be fooled by a concurrent tick — `selectPendingContents`'s own comment says selection "does not claim" rows. If a second tick finishes the same rows first, this tick sees 0 pending, logs `BATCH_PROCESSING_ABANDONED` at WARNING, and `decideEnding` downgrades a run that actually succeeded end-to-end. Worth deriving retryability from persisted row state at decision time rather than from this batch's own uuids.

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProcessTickWorker.java
```diff
@@ -139,23 +154,60 @@ private long backlogOf(UUID discoveryUuid) {
...
+            messageWriter
```
klaraf755: The `messageWriter.append`/`appendAll` calls inside these catch blocks aren't themselves guarded. If the same correlated failure that tripped the outer exception (DB outage, lock timeout) also hits this write, it escapes uncaught to `DiscoveryWorkListener`'s generic catch — exactly the path the surrounding comment says this code exists to avoid, bypassing the tick's own stall/backoff budget.

src/main/java/com/otilm/core/service/handler/CertificateHandler.java
```diff
@@ -219,9 +220,7 @@ public List<String> stageDiscoveredCertificates(String batch, Discovery discover
                 logger
                         .error("Unable to create discovery certificate {} in batch {} for discovery {}. Message: {}",
                                 identifier, batch, discovery.getName(), e.getMessage(), e);
-                failures
-                        .add("Certificate %s could not be staged: %s"
-                                .formatted(identifier, DiscoveryFailureReason.shape(e)));
+                failures.add(DiscoveryFailureReason.shape(e));
```
klaraf755, line 223: This drops the certificate identifier from the failure entry. The new Javadoc says "the certificate that hit it is already in the log above," but that's `logger.error` (application log, not user-facing), and in every exception path here the `DiscoveryCertificate` row is never persisted either. An operator looking at the new run-messages endpoint sees "3 certificates could not be staged: invalid PEM" with no way to identify which 3 — the old JSONB log named them.

src/main/resources/db/migration/V202608251000__discovery_run_message_rows.sql
```diff
@@ -0,0 +1,28 @@
...
+-- The old column is dropped without backfill: it was added in this same unreleased cycle (V202608141100), so
+-- only mainline and staging runs carry values, and those are bare strings with no code or severity to map onto
+-- the new shape.
+
+CREATE TABLE "discovery_message" (
+    -- Ordering only, never exposed; a timestamp cannot order these (see DiscoveryMessage.id).
+    "id" BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
+    "discovery_uuid" UUID NOT NULL,
+    "severity" VARCHAR NOT NULL,
+    "code" VARCHAR NOT NULL,           -- kind of problem: connector-supplied for connector errors, else Core's own
```
klaraf755, line 13: `code` and `message` are both unbounded `VARCHAR`, and nothing here backs either of two length assumptions made elsewhere: interfaces#912 bounds `DiscoveryErrorEvent.code` to 64 chars on the claim that storage requires it, and `DiscoveryMessageWriter`'s `MAX_PER_CODE=20`/`MAX_MESSAGE_LENGTH=500` are framed as storage-safety bounds with nothing here to require them — so a legitimately verbose connector with >20 distinct failure reasons gets messages silently suppressed with no way to raise the cap. Suggest a matching `VARCHAR(64)` (or `CHECK`) on `code`, and either a real constraint or a reworded rationale for `message`'s cap.

src/main/java/com/otilm/core/service/writer/discovery/DiscoveryMessageWriter.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Transactional
+    public void appendAll(UUID discoveryUuid, List<DiscoveryMessageDraft> drafts) {
+        if (drafts.isEmpty()) {
+            // The clean path for both hot callers, which would otherwise open and commit a transaction per batch
+            // and per page to write nothing.
+            return;
+        }
+        drafts.forEach(draft -> record(discoveryUuid, draft));
```
klaraf755: `appendAll` does one DB round-trip per distinct message in a loop rather than a single batched upsert — on exactly the "same fault tens of thousands of times" hot path this PR targets, while holding the run row's lock. Worth batching into one multi-row `INSERT ... ON CONFLICT`.

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProcessTickWorker.java
```diff
@@ -37,31 +43,40 @@ public class DiscoveryProcessTickWorker {
 
     private static final Logger logger = LoggerFactory.getLogger(DiscoveryProcessTickWorker.class);
 
+    /** The severities that mean the run did not get everything done, whatever its rows say. */
+    private static final Set<DiscoveryMessageSeverity> UNRECOVERED = EnumSet
```
klaraf755, line 50: `UNRECOVERED` here is a hand-picked `EnumSet` that duplicates `DiscoveryRunLifecycle.severityOf()`'s exhaustive switch (which safely defaults new values to `ERROR`). A future severity value would update `severityOf` automatically but silently fall outside `UNRECOVERED`, letting a run with only that severity finish `COMPLETED` wrongly. Consider owning this decision in one place, e.g. `DiscoveryRunLifecycle.countsAsUnrecovered(...)`.

klaraf755 (commented):
Great write-up in the description. A few correctness/coverage items worth a look before merge, mostly around the tick worker's retry classification and the message writer's error isolation.

src/test/java/com/otilm/core/messaging/jms/listeners/discovery/DiscoveryRunReaperUnitTest.java
```diff
@@ -122,9 +126,10 @@ void workLost_oneRunFailureDoesNotAbortTheBatch() {
         assertThatCode(() -> reaper.reap()).doesNotThrowAnyException();
```
klaraf755, line 126: This went from asserting the actual persisted `run_messages` content to only `verify(messageWriter).appendRunEnded(...))` — a mock-interaction check that can't catch a bug inside the writer itself anymore. Worth confirming `DiscoveryMessageWriterITest` independently exercises `appendRunEnded` end-to-end for a FAILED/ERROR run.

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProcessTickWorker.java
lubomirw: Real, and it's the concurrency defect OmniTrustILM/core#2130 owns rather than something this predicate introduced — the previous `backlogOf(run)` was equally claim-free.

Worth noting the racing case mostly resolves itself through row state before this message matters: as #2130 describes, two ticks race on the certificate insert and the loser's whole content group rolls back recorded as `processed_error`, so `decideEnding` ends the run WARNING from `rowsFailed` regardless of what I file here. The narrow case your comment describes — a concurrent tick completing the rows entirely cleanly while this one throws — needs the claim to distinguish, which is #2130's first DoD item.

I did fold the terminal-transition half of this into #2130 while reviewing: "The terminal decision refuses to end a run while a batch claim is outstanding, so a run cannot complete between another tick's row commit and its gap record."

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProcessTickWorker.java
lubomirw: Good catch, and worse than it looks: the escape skips `stall()`, so the attempt counter never climbs and the budget never ends the run — the opposite of what swallowing the original failure was for.

Fixed in b05ee15ee. Both catch bodies now record through a helper that logs and stops, and `pendingIn(batch)` moved inside it too since that is itself a query the same outage would take down.

src/main/java/com/otilm/core/service/handler/CertificateHandler.java
lubomirw: Accepted as a trade-off rather than fixed, and now documented on `DiscoveryMessageDraft` as such.

The reasoning: before this table existed those identifiers were only in the application log, and they still are — this log is a bounded summary above it rather than a replacement. Naming each occurrence is exactly what would make it too long to read, which is the growth the aggregation exists to stop. You are right that for staging failures there is no row to fall back on either, so the javadoc no longer claims per-certificate detail is always reachable; it names the three cases where it is not.

If identifying the specific certificates turns out to matter operationally, the fix is a bounded refs column on the message row, which is a schema change beyond what #2127 asks for.

src/main/resources/db/migration/V202608251000__discovery_run_message_rows.sql
lubomirw: You are right that the storage claim was unbacked, and I have corrected the claim rather than the schema.

Not adding `VARCHAR(64)`: a length change would then need a migration, the check costs something on every insert, and the shortening is a code-level limit anyway — nothing about the column requires it. So the interfaces javadoc that said an unbounded code "has to be shortened to be stored" was simply wrong, and it now says what actually happens (interfaces 6c2beed0): Core's writer shortens a longer code, and two long codes sharing a prefix would merge into one entry once shortened. That is the real reason for the bound, and it survives without a database constraint.

`message` needs nothing either — it is hashed for the unique index, so its length never threatens the index, and its constant already describes itself as a readability cap rather than a storage one.

On `MAX_PER_CODE=20`: that these are constants rather than configuration is a deliberate decision recorded in #2127's Definition of Done. A connector exceeding 20 distinct reasons for one code is not silent — overflow folds into a suppression row carrying the count of what was dropped, and the oldest entries are the ones kept.

src/main/java/com/otilm/core/service/writer/discovery/DiscoveryMessageWriter.java
lubomirw: Declining this one — I think the hot path is the other branch.

`appendAll` loops over *distinct kinds*, not occurrences: at most five gap kinds per batch or roughly seven staging reasons per page, hard-capped at `MAX_PER_RUN` of 50 for a whole run. The "same fault tens of thousands of times" case never enters the loop at all — it aggregates onto one existing row in a single upsert, which is the property the design is built around.

Batching would also defeat the bound: `appendWithinBounds` evaluates its EXISTS and two counts per row to decide whether that row may be created, and a multi-row `INSERT ... ON CONFLICT` cannot make that decision per element.

src/main/java/com/otilm/core/service/handler/discovery/DiscoveryProcessTickWorker.java
lubomirw: Fixed in b05ee15ee — it is now `EnumSet.complementOf(EnumSet.of(INFO))`, so anything above INFO counts and a severity added later counts by default, which is the same safe direction `severityOf`'s `default -> ERROR` takes.

Left in the worker rather than moved to `DiscoveryRunLifecycle`: it feeds a repository query and the worker is its only consumer, so a second home seemed like more indirection than the two lines are worth. Happy to move it if you would rather both severity decisions sat together.

src/test/java/com/otilm/core/messaging/jms/listeners/discovery/DiscoveryRunReaperUnitTest.java
lubomirw: Checked, and the gap was real: `DiscoveryStatusTickWorkerITest` has four tests ending a run FAILED and none of them asserted the message log, while `DiscoveryMessageWriterITest` exercised `appendRunEnded` only at WARNING. So nothing covered the status-to-severity mapping end-to-end once the reaper test moved to a mock.

Fixed in b05ee15ee by asserting the terminal message inside `failedAnswer_endsTheRunAndReleasesItsAgendaAndHandle`, which runs the real terminator: code `runEnded`, severity ERROR, and the message equal to the run's own reason.

klaraf755 (dismissed):

klaraf755 (approved):

PR #2156, opened 2026-08-26
Accept an ASN.1 JSON tree as a certificate extension value

### LINKED ISSUE #2131: Encode JSON certificate extension values to DER and enforce JSON-schema constraints
### Description

Backend for JSON-valued certificate extensions. Split out of OmniTrustILM/ilm#325; the structured targets are
OmniTrustILM/core#1950. Contract counterpart: OmniTrustILM/interfaces#897, which resolves both open questions
(registry-authoritative schemas plus a narrowing `JsonSchemaConstraint`; JSON input derived from
`valueEncoding` = `DER`).

**`AsnJsonCodec`.** New pure kernel in `com.otilm.core.util` beside the renderer (the same package-cycle
reasoning as `StructuredExtensionCodec`): parses a structural ASN.1 JSON tree and encodes DER. Encode-only -
uploaded-CSR validation of these extensions stays presence-only, so nothing decodes; tests round-trip through
BouncyCastle's readers. Node vocabulary (boolean, integer, oid, utf8String, ia5String, printableString,
octetString, bitString, generalizedTime, null, sequence, set, tagged) is defined in the design spec. Malformed
trees throw with the offending node's JSON path.

**Renderer hook.** `X509RequestContentRenderer.encodeExtensionValue`'s `DER` case: a value starting with `{`
is a JSON tree encoded via the codec; anything else takes the existing base64 path byte-for-byte unchanged.
Base64's alphabet cannot produce `{`, so the discriminator is unambiguous and stored values need no migration.

**Schema resolution and validation layers.** A custom OID entry's `valueSchema` rides the registry cache
(`OidRecord` gains the field); system extensions resolve from Core classpath resources keyed by OID - shipped
for Basic Constraints (2.5.29.19) and TLS Feature (1.3.6.1.5.5.7.1.24). A DER extension with no schema is
grammar-only: any valid tree is accepted.

Request-time layering for a value mapped to a DER extension: grammar tree first (codec parse), then the
registry shape schema when the OID has one, then the attribute's `JsonSchemaConstraint` (enforced by
`AttributeDefinitionUtils`, see interfaces#897). Each layer's failure is worded distinctly and names the
failing JSON path - "does not match the registered schema for extension 2.5.29.19 (at $.sequence[0])" versus
"violates the attribute's JSON Schema constraint (at ...)" - so an operator sees which schema rejected.
Whether two schemas are mutually satisfiable is not checkable; the mitigations are the layered attribution and
the existing default-content-at-registration hook, which catches an unsatisfiable constraint at save time when
the definition declares an example value.

**Registration-time validation** in `AttributeEngine`: a `JsonSchemaConstraint`'s data and an OID entry's
`valueSchema` must each be a well-formed JSON Schema (parse + meta-validate); default content of
extension-mapped attributes is additionally checked against the registry schema.

**Dependency:** none new - `com.networknt:json-schema-validator` arrives transitively from interfaces#897,
where enforcement lives so connectors keep enforcing their own constraints locally.

**No URL-sourced schemas** and no outbound fetch anywhere on any path - the URL variant was cut (additive to
restore later).

**Design spec:**
`request-attributes-design-and-plans/core-design/2026-08-25-json-extension-values-design.md`

### Acceptance Criteria

- [ ] `AsnJsonCodec` encodes every node type, pinned by golden DER vectors: Basic Constraints `{"sequence":[{"boolean":true},{"integer":0}]}` = `30 06 01 01 FF 02 01 00`, TLS Feature `{"sequence":[{"integer":5}]}` = `30 03 02 01 05`; round-trips verified through BouncyCastle's readers; explicit vs implicit tagging covered.
- [ ] Renderer accepts a JSON tree for a DER extension and still accepts legacy base64 byte-for-byte - regression test with a stored-style value.
- [ ] Registry shape validation enforced for OIDs carrying a schema; a schema-less DER OID accepts any valid tree.
- [ ] Layered error attribution: registry-schema violations and constraint violations worded distinctly, each naming the failing JSON path.
- [ ] Core ships schemas for 2.5.29.19 and 1.3.6.1.5.5.7.1.24 as classpath resources, covered by tests.
- [ ] Invalid JSON Schema documents (constraint data or `valueSchema`) rejected at registration; `valueSchema` on a non-DER encoding rejected.
- [ ] No outbound fetch on any attribute path.
- [ ] `mvn -B verify` passes including integration tests; Sonar gate and 80% new-code coverage.

klaraf755:
Closes #2131

Lets an operator supply a DER certificate extension's value as a structural JSON tree instead of hand-encoded base64, and validates that tree against a JSON Schema registered per OID.

Depends on OmniTrustILM/interfaces#910 (the `JsonSchemaAttributeConstraint` contract and the networknt dependency). CI here cannot pass until that merges.

## The value format

A DER extension value beginning with `{` is read as a JSON tree; anything else is base64 DER, exactly as before. Base64's alphabet cannot produce `{`, so no stored value changes meaning and no migration is needed.

Each node is an object with exactly one key naming its ASN.1 type:

```json
{"sequence": [{"boolean": true}, {"integer": 0}]}
```

Supported: `boolean`, `integer`, `oid`, `utf8String`, `ia5String`, `printableString`, `octetString` (base64), `bitString` (`{value, padBits}`), `generalizedTime`, `null`, `sequence`, `set`, `tagged` (`{tagNo, explicit, value}`).

The dialect is inspired by X.697 but is **not JER** — JER is schema-driven and has no Java runtime. It should not be documented under that name. `AsnJsonCodec` is the grammar: what it accepts is the definition, and a rejected value names the offending node's JSON path.

## Two validation layers

A value passes the **codec grammar** first — so an extension with no schema at all still cannot carry anything unencodable. Where a schema exists it narrows what is already well-formed ASN.1:

- **Registry schema**, per OID. A custom OID carries one in its new `value_schema` column; a system OID's ships with Core as a classpath resource, since an entry for a system OID cannot be created (it is reserved). Messages read `does not match the registered schema for extension <oid>`.
- **JSON Schema constraint**, per attribute definition (interfaces#910). An RA-profile policy narrowing what the registry already permits. Messages read `violates the attribute's JSON Schema constraint`.

The wordings differ deliberately so an operator can tell which layer rejected.

All eight system certificate extensions now ship a schema. Each is checked by encoding an accepted tree and handing the bytes to BouncyCastle's typed class — which is what proves the implicit tagging in Name Constraints and Private Key Usage Period is right; comparing bytes against the same construction would prove nothing. A completeness test fails the build if a system extension is added without a schema.

## Scope: externally supplied CSRs

Both schema layers apply to extension values the **platform composes** from request attributes. Neither is applied when a CSR is **uploaded** to an issuance request.

On the uploaded path, `CertificateRequestContentValidator` still enforces presence of a required mapped field, the strict whitelist, and the values of the structured Key Usage and Extended Key Usage targets — those are typed, so the parser hands over decoded codes and OIDs with no ambiguity to resolve. What is not checked is the **shape of an opaque extension's value**.

This is a deliberate boundary. Checking it would mean decoding the CSR's DER back into a JSON tree, and that decode is not well-defined: a context tag carries no record of whether it was IMPLICIT or EXPLICIT, so distinct trees share an encoding. Measured against BouncyCastle 1.84:

```
A0 05 3003020101  =  EXPLICIT [0] of SEQUENCE{INTEGER 1}
                  =  IMPLICIT [0] of SEQUENCE{SEQUENCE{INTEGER 1}}

A0 03 020101      =  IMPLICIT [0] of SEQUENCE{INTEGER 1}
                  =  EXPLICIT [0] of INTEGER 1
```

Only primitive implicit tags (`80 01 01`) are unambiguous. The interpretation lives in the ASN.1 type definition, which the bytes do not carry, so any decoder must guess — and it would guess precisely on the extensions that use tagging: Name Constraints (`[0]`/`[1]` subtree lists) and Private Key Usage Period (`[0]`/`[1]` times). A wrong guess rejects a valid CSR, which is the failure mode the validator already avoids by skipping constraints for extension-mapped fields.

The boundary is also coherent on its own terms: these schemas describe values authored in the JSON dialect, and an uploaded CSR was never authored that way — validating it would assert an interpretation the requester never expressed.

A schema-guided decoder could disambiguate where the registered schema determines the shape, reporting "cannot be checked" otherwise. That is a larger change and a separate decision; it is not needed for this feature to be correct.

## Notes for review

- **Validation ordering is load-bearing.** `validateAttributesContent` removes each matched definition from the mapping as it goes, using the leftovers to find missing required attributes. The JSON check must therefore run *before* it, or it sees no definition for any attribute that carried content — which made the feature silently inert in an earlier revision. `AttributeEngineITest` covers this through the real path and fails against the old ordering.
- **Schema loading never reaches the network.** A non-local `$ref` is refused when a schema is registered, and the loader is disabled outright so no route can fetch one — a `$ref` target is resolved lazily, so an unloadable one surfaces during validation rather than from `getSchema`, and is reported as unverifiable instead of escaping.
- Basic Constraints requires `cA` to be absent rather than explicitly `false`, per X.690 §11.5 — a DEFAULT value must not be encoded in DER. The empty sequence is the end-entity form.

## Testing

Unit and integration suites pass. Verified end to end against a running instance: a JSON tree submitted through an RA-profile request attribute reaches the generated CSR as correct DER (`30 07 0C 05 68656C6C6F` for a UTF8String member), Basic Constraints `{"sequence":[]}` encodes as `30 00` and reads back as `CA:FALSE`, and a constraint violation returns 400 naming the offending node.


src/main/java/com/otilm/core/util/ExtensionSchemas.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /** The schema governing {@code oid}'s value, or empty when neither the registry nor Core ships one. */
+    public static Optional<JsonSchema> resolve(String oid) {
+        Map<String, OidRecord> registry = OidHandler.getOidCache(OidCategory.CERTIFICATE_EXTENSION);
+        OidRecord record = registry == null ? null : registry.get(oid);
+        if (record != null && record.valueSchema() != null) {
+            return Optional.of(load(record.valueSchema()));
+        }
+        return shippedSchema(oid).map(ExtensionSchemas::load);
```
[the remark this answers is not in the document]
klaraf755: Fixed. Only a system entry, or no entry, falls back to the shipped schema now; a custom entry declaring none leaves the value unconstrained, matching how its criticality and encoding already win while it exists.

An existing test had encoded the old behaviour and failed on the change — split into three cases: built-in entry, no entry, and custom entry with no schema.

src/main/java/com/otilm/core/util/ExtensionSchemas.java
```diff
@@ -0,0 +1,157 @@
...
+            throw new ValidationException("Not a valid JSON Schema document: not JSON");
+        }
+        if (parsed == null || !(parsed.isObject() || parsed.isBoolean())) {
+            throw new ValidationException("Not a valid JSON Schema document: the root must be an object or a boolean");
+        }
+        rejectNonLocalRefs(parsed, "$");
+        try {
+            FACTORY.getSchema(parsed);
+        } catch (RuntimeException e) {
+            throw new ValidationException("Not a valid JSON Schema document: " + e.getMessage());
```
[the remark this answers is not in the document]
klaraf755: Confirmed and fixed. `getSchema` accepts `{"type":123}`, `{"minItems":"x"}` and `{"required":"notAnArray"}` without complaint, so such a document registered and then constrained nothing.

Now validated against the 2020-12 metaschema. That needed `ClasspathSchemaLoader` ahead of `DisallowSchemaLoader` — the blanket disallow was blocking the library's own bundled metaschema.

src/main/java/com/otilm/core/util/ExtensionSchemas.java
```diff
@@ -0,0 +1,157 @@
...
+     * Rejects a schema document networknt cannot load. Applied to a registration's {@code valueSchema} and to a JSON
+     * Schema constraint's data, so a broken document fails at save rather than at first use.
+     */
+    public static void requireValidSchema(String schemaDocument) {
+        JsonNode parsed;
+        try {
+            parsed = MAPPER.readTree(schemaDocument);
+        } catch (IOException e) {
+            throw new ValidationException("Not a valid JSON Schema document: not JSON");
+        }
```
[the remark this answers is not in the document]
klaraf755: Confirmed — `readTree(null)` throws `IllegalArgumentException`, which the `IOException` catch does not cover. A missing document is now rejected as a validation error before parsing.

src/main/java/com/otilm/core/util/ExtensionSchemas.java
```diff
@@ -0,0 +1,157 @@
...
+     * which the platform does not do, so accepting one would register a schema that silently constrains nothing.
+     */
+    private static void rejectNonLocalRefs(JsonNode node, String path) {
+        if (node.isObject()) {
+            for (Map.Entry<String, JsonNode> property : node.properties()) {
+                if ("$ref".equals(property.getKey()) && property.getValue().isTextual()
+                        && !property.getValue().textValue().startsWith("#")) {
+                    throw new ValidationException(
+                            "Not a valid JSON Schema document: $ref at %s points outside the document; only local references such as #/$defs/name are supported"
+                                    .formatted(path));
```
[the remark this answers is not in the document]
klaraf755: Right, `{"const":{"$ref":"https://example.invalid/literal"}}` is a valid schema and was being rejected. The walk now skips `const`, `enum`, `default` and `examples`, whose values are instance data. Disabled loader kept as defence in depth.

src/main/java/com/otilm/core/util/ExtensionSchemas.java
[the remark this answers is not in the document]
klaraf755: Fixed — the compiler's message is logged and the operator gets controlled text, matching the info-leak rule the renderer already follows.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -775,13 +782,134 @@ private void validateDataAttributesContent(UUID connectorUuid, String operation,
...
+            if (!(field instanceof ExtensionMappedField ext) || ext.getExtensionOid() == null) {
+                continue;
+            }
+            OidRecord record = registry == null ? null : registry.get(ext.getExtensionOid());
+            ExtensionValueEncoding encoding = record != null
+                    ? record.valueEncoding()
+                    : systemExtensionEncoding(ext.getExtensionOid());
+            if (encoding == ExtensionValueEncoding.DER) {
+                oids.add(ext.getExtensionOid());
+            }
```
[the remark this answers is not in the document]
klaraf755: Confirmed. Worth noting this was a fix I had already claimed to make earlier in this branch's history — the edit silently did not apply, so the check still excluded null. Now applied and covered.

src/main/java/com/otilm/core/util/AsnJsonCodec.java
```diff
@@ -0,0 +1,231 @@
...
+        try {
+            return toAsn1(tree, "$").toASN1Primitive().getEncoded(ASN1Encoding.DER);
+        } catch (IOException e) {
+            throw new ValidationException("Extension value could not be encoded to DER");
+        } catch (ValidationException e) {
+            throw e;
+        } catch (RuntimeException e) {
+            // BouncyCastle rejects some values with its own unchecked exceptions. Callers validate on request
+            // paths, so one escaping would be a 500 rather than a rejection.
+            throw new ValidationException("Extension value could not be encoded to DER: " + e.getMessage());
```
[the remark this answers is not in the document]
klaraf755: Fixed — fixed safe message. Every grammar failure already carries its own path-bearing message through the preceding `ValidationException` catch, so nothing informative is lost.

src/main/java/com/otilm/core/util/AsnJsonCodec.java
```diff
@@ -0,0 +1,231 @@
...
+        return vector;
+    }
+
+    private static ASN1Encodable parseTagged(JsonNode node, String path) {
+        if (!node.isObject() || !node.has("tagNo") || !node.has("value")) {
+            throw new ValidationException("Node at %s must carry tagNo and value".formatted(path));
+        }
+        int tagNo = requireBoundedInt(node.get("tagNo"), path + ".tagNo", 0, 30);
+        boolean explicit = !node.has("explicit") || requireBoolean(node.get("explicit"), path + ".explicit");
+        return new DERTaggedObject(explicit, tagNo, toAsn1(node.get("value"), path + ".value"));
```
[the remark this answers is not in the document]
klaraf755: Good catch — `explict` fell back to `explicit=true` and encoded different DER silently. Members other than `tagNo`, `explicit` and `value` are now rejected by name.

src/main/java/com/otilm/core/util/AsnJsonCodec.java
```diff
@@ -0,0 +1,231 @@
...
+            throw new ValidationException("Value at %s is not a valid OID".formatted(path));
+        }
+    }
+
+    private static ASN1Encodable parseBitString(JsonNode node, String path) {
+        if (!node.isObject() || !node.has("value")) {
+            throw new ValidationException("Node at %s must carry a base64 value and optional padBits".formatted(path));
+        }
+        byte[] bytes = requireBase64(node.get("value"), path + ".value");
+        int padBits = node.has("padBits") ? requireBoundedInt(node.get("padBits"), path + ".padBits", 0, 7) : 0;
```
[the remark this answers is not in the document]
klaraf755: Same fix — `padbits` defaulted `padBits` to 0 and changed the encoded value. Only `value` and `padBits` are accepted now.

src/main/resources/extension-schemas/1.3.6.1.5.5.7.1.24.json
```diff
@@ -0,0 +1,14 @@
+{
+  "type": "object",
+  "required": ["sequence"],
+  "properties": {
+    "sequence": {
+      "type": "array",
+      "items": {
```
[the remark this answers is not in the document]
klaraf755: Fixed, `minItems: 1`. I had dropped `minItems` deliberately for Basic Constraints, where the empty sequence is the legitimate end-entity form, and carried that over to TLS Feature where it is not.

src/main/resources/extension-schemas/2.5.29.16.json
```diff
@@ -0,0 +1,26 @@
+{
+  "type": "object",
+  "required": ["sequence"],
+  "properties": {
+    "sequence": {
+      "type": "array",
+      "minItems": 1,
+      "maxItems": 2,
+      "items": {
+        "type": "object",
```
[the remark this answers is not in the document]
klaraf755: Confirmed — the schema accepted `[1],[0]` and `[0],[0]`, both of which encode a malformed sequence. A homogeneous `items` rule cannot express ordered optional slots; replaced with `oneOf` over the three legal shapes, each with `items: false`.

src/main/resources/extension-schemas/2.5.29.30.json
```diff
@@ -0,0 +1,63 @@
...
+  "type": "object",
+  "required": ["sequence"],
+  "properties": {
+    "sequence": {
+      "type": "array",
+      "minItems": 1,
+      "maxItems": 2,
+      "items": {
+        "type": "object",
+        "required": ["tagged"],
```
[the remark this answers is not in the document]
klaraf755: Same problem and same fix: `oneOf` over `[0],[1]`, `[0]` alone and `[1]` alone, each closed with `items: false`.

src/main/resources/extension-schemas/2.5.29.30.json
```diff
@@ -0,0 +1,63 @@
...
+    "generalSubtree": {
+      "type": "object",
+      "required": ["sequence"],
+      "properties": {
+        "sequence": {
+          "type": "array",
+          "minItems": 1,
+          "maxItems": 3,
+          "prefixItems": [
+            {
```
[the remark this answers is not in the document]
klaraf755: Confirmed — `prefixItems` alone does not bound the tail, so `{"utf8String":"oops"}` passed in the `minimum` slot. `minimum` `[0]` and `maximum` `[1]` are now typed as non-negative integers in their ordered positions, with `items: false`.

src/main/resources/extension-schemas/2.5.29.30.json
```diff
@@ -0,0 +1,428 @@
...
+                  "required": [
+                    "tagged"
+                  ],
+                  "properties": {
+                    "tagged": {
+                      "type": "object",
+                      "required": [
+                        "tagNo"
+                      ],
+                      "properties": {
```
ivosh, line 184: **! Major** | Correctness

src/main/resources/extension-schemas/2.5.29.30.json:182 constrains a GeneralName only by tag number, so a DNS name using default explicit tagging or a non-IA5 value passes despite requiring IMPLICIT IA5String. The alternatives around src/main/resources/extension-schemas/2.5.29.30.json:220 also accept minimum/maximum distances, although RFC 5280 requires minimum to be zero (and therefore omitted in DER) and maximum absent. This can emit an invalid critical Name Constraints extension.

_Suggested:_ Model each GeneralName tag's required tagging and value type, and reject all GeneralSubtree minimum and maximum members.

src/main/java/com/otilm/core/util/AsnJsonCodec.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    /**
+     * Parses a tree value. This is the one place the dialect's text form is read, so a caller that needs both the tree
+     * and its encoding parses once and cannot end up judging a different tree than it encodes.
+     */
+    public static JsonNode parse(String json) {
+        JsonNode tree;
+        try {
+            tree = MAPPER.readTree(json);
```
ivosh: **~ Minor** | Correctness

src/main/java/com/otilm/core/util/AsnJsonCodec.java:51 uses readTree without trailing-token or duplicate-key detection. Inputs such as `{"integer":1} garbage` are accepted and encoded as the prefix, while duplicate type keys collapse to the last value, silently discarding operator-authored content instead of rejecting malformed JSON.

_Suggested:_ Parse with trailing-token and strict duplicate-field detection enabled, and add rejection tests for both cases.

src/main/java/com/otilm/core/util/ExtensionSchemas.java
```diff
@@ -0,0 +1,234 @@
...
+                return Optional.empty();
+            }
+            return Optional.of(new String(resource.readAllBytes(), StandardCharsets.UTF_8));
+        } catch (IOException e) {
+            throw new IllegalStateException("Core-shipped extension schema for " + oid + " is unreadable", e);
+        }
+    }
+
+    /**
+     * Rejects a schema document networknt cannot load. Applied to a registration's {@code valueSchema} and to a JSON
```
ivosh: **~ Minor** | Documentation

ExtensionSchemas.java:111 promises unloadable schemas fail at save, while ExtensionSchemas.java:220-223 says `$ref` failures surface lazily during value validation; ExtensionSchemas.java:40 also claims eager resolution. These conflicting guarantees obscure the no-network threat model.

_Suggested:_ Rewrite these comments to distinguish registration checks from lazy resolution failures handled by `validateShape`.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -778,13 +785,136 @@ private void validateDataAttributesContent(UUID connectorUuid, String operation,
...
+        // mapping as it goes (it uses the leftovers to find missing required attributes), so running after it
+        // would see no definition for any attribute that actually carried content.
+        List<ValidationError> errors = validateJsonExtensionValues(definitionsMapping, requestAttributes);
+        errors.addAll(validateAttributesContent(definitionsMapping, requestAttributes));
         if (!errors.isEmpty()) {
             throw new ValidationException(errors);
         }
     }
 
+    /**
```
ivosh, line 798: **. Nitpick** | Documentation

AttributeEngine.java:798-802 combines constraint validation, declared-content validation, and the unsatisfiable-pair caveat in one paragraph. ExtensionSchemas.java:87-90 similarly uses a 39-word sentence for return semantics, API rationale, and registry limitations, making load-bearing behavior difficult to scan.

_Suggested:_ Restructure each block as a one-line summary followed by short `<p>` sections with bold concern labels.

src/main/java/com/otilm/core/service/impl/CustomOidEntryServiceImpl.java
```diff
@@ -212,8 +214,13 @@ public CustomOidEntryDetailResponseDto createCustomOidEntry(CustomOidEntryReques
                 }
                 defaultCritical = additionalProperties.getDefaultCritical();
                 valueEncoding = additionalProperties.getValueEncoding();
+                valueSchema = additionalProperties.getValueSchema();
```
ivosh, line 217: **~ Minor** | Correctness

CustomOidEntryServiceImpl.java:217 accepts the new schema, but the current fe-administrator generated CertificateExtensionOidPropertiesDto and custom-OID form expose only defaultCritical and valueEncoding; the detail view also omits the schema. Operators using the normal UI cannot configure the feature this PR advertises, and the PR description coordinates interfaces#910 but not the frontend.

_Suggested:_ Coordinate a fe-administrator PR that regenerates the DTO and adds a DER-only valueSchema editor, submission mapping, and detail display.

src/test/java/com/otilm/core/attribute/engine/RequestAttributeDefinitionValidationTest.java
```diff
@@ -67,6 +69,38 @@ static void restoreRdnRegistry() {
...
+        DataAttributeV3 definition = validDefinition();
+        JsonSchemaAttributeConstraint constraint = new JsonSchemaAttributeConstraint();
+        constraint.setData("this is not json");
+        definition.setConstraints(List.of(constraint));
+        assertRejected(definition, "valid JSON Schema document");
+    }
+
+    @Test
+    void jsonSchemaConstraintCheckedEvenWithoutAFieldMapping() {
+        // The connector registration path used to run this check only for mapped attributes, so a connector
```
ivosh: **. Nitpick** | Documentation

RequestAttributeDefinitionValidationTest.java:83-84 records how the connector path behaved previously rather than explaining a current invariant; that history belongs in this PR or its commit message.

_Suggested:_ Remove the two-line history comment.

src/test/java/com/otilm/core/util/ExtensionSchemasTest.java
```diff
@@ -0,0 +1,254 @@
...
+        assertThat(ExtensionSchemas
+                .validateShape("1.3.6.1.5.5.7.1.24", MAPPER.readTree("{\"sequence\":[{\"integer\":5}]}"))).isEmpty();
+        assertThat(ExtensionSchemas
+                .validateShape("1.3.6.1.5.5.7.1.24", MAPPER.readTree("{\"sequence\":[{\"oid\":\"1.2\"}]}")))
+                .isNotEmpty();
+    }
+
+    @Test
+    void aRegistryEntrySchemaBeatsTheShippedResource() throws Exception {
+        // given — a custom entry for 2.5.29.19 demanding an entirely different shape
```
ivosh: **. Nitpick** | Maintainability

ExtensionSchemasTest.java:103 and ExtensionSchemasTest.java:112 restate the setup and assertions already expressed by `aRegistryEntrySchemaBeatsTheShippedResource` and the test body.

_Suggested:_ Remove both given/then comments.

src/test/java/com/otilm/core/util/X509RequestContentRendererTest.java
```diff
@@ -566,6 +566,35 @@ class ExtensionValueEncodingTest {
         // ExtendedKeyUsage — not in the forced-critical set, so criticality does not interfere
         private static final String EXT_OID = "2.5.29.37";
 
+        @Test
+        void encodesAJsonTree_whenDerValueStartsWithABrace() throws Exception {
+            // given — a structural ASN.1 JSON tree instead of base64 DER
```
ivosh: **. Nitpick** | Maintainability

X509RequestContentRendererTest.java:571 and X509RequestContentRendererTest.java:591 merely repeat their test names and immediately adjacent inputs; the method names and assertions already identify the scenarios.

_Suggested:_ Remove the redundant given comments at lines 571 and 591.

src/test/java/com/otilm/core/util/ShippedExtensionSchemaTest.java
```diff
@@ -0,0 +1,255 @@
...
+
+    @Test
+    void tlsFeatureRejectsAnEmptySequence() {
+        // RFC 7633 Features is a SEQUENCE OF INTEGER; an empty one asserts nothing.
+        assertRejects("1.3.6.1.5.5.7.1.24", "{\"sequence\":[]}");
+    }
+
+    @Test
+    void privateKeyUsagePeriodRejectsReversedAndDuplicateMembers() {
+        // notBefore [0] precedes notAfter [1] and each appears at most once; a homogeneous items rule let
```
ivosh: **. Nitpick** | Documentation

ShippedExtensionSchemaTest.java:214-215 and ShippedExtensionSchemaTest.java:240 describe what earlier schema rules allowed instead of documenting only the current ordering and cardinality invariants.

_Suggested:_ Keep the current invariants at lines 214-215 and remove the prior-implementation narration there and at line 240.

src/test/java/com/otilm/core/util/AsnJsonCodecTest.java
```diff
@@ -0,0 +1,226 @@
...
+    void rejectsANullNodeCarryingAValue() {
+        // ASN.1 NULL has no content, so a value here is a mistake; encoding 05 00 anyway would discard it.
+        assertThatThrownBy(() -> AsnJsonCodec.encodeFromString("{\"null\":123}"))
+                .isInstanceOf(ValidationException.class)
+                .hasMessageContaining("$.null");
+    }
+
+    @Test
+    void sortsSetMembersAsDerRequires() throws Exception {
+        // DER orders SET components by encoding, so the declared order is not preserved. Documented, not a bug.
```
ivosh: **. Nitpick** | Maintainability

AsnJsonCodecTest.java:196 already explains DER ordering; the trailing “Documented, not a bug” adds no technical information.

_Suggested:_ Delete the final “Documented, not a bug” sentence.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
```diff
@@ -197,6 +197,20 @@ private static byte[] encodeExtensionValue(String value, ExtensionValueEncoding
...
+     * Encodes a structural ASN.1 JSON tree value. Wrong input is reported naming both accepted forms, because a
+     * DER-encoded extension takes either a JSON tree or base64 DER and the author needs to know which failed.
+     */
+    private static byte[] encodeJsonTree(String value) throws IOException {
+        try {
+            return AsnJsonCodec.encodeFromString(value);
+        } catch (RuntimeException e) {
+            throw new IOException(
+                    "Invalid DER extension value; a value starting with '{' must be a valid ASN.1 JSON tree, "
+                            + "anything else base64-encoded DER: " + e.getMessage());
```
ivosh: **~ Minor** | Security

X509RequestContentRenderer.java:210 appends `getMessage()` from any `RuntimeException` to the new `IOException`. The production caller wraps that exception in `CertificateException`, whose REST handler returns the cause message, so an unexpected library exception could disclose internal details to the client.

_Suggested:_ Gate the appended detail on `e instanceof ValidationException` and use fixed safe text for every other runtime exception.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -778,13 +785,136 @@ private void validateDataAttributesContent(UUID connectorUuid, String operation,
...
+     */
+    static void validateJsonSchemaDeclarations(DataAttributeV3 attribute, String connectorUuidStr)
+            throws AttributeException {
+        if (attribute.getConstraints() != null) {
+            for (BaseAttributeConstraint<?> constraint : attribute.getConstraints()) {
+                if (constraint.getType() != AttributeConstraintType.JSON_SCHEMA) {
+                    continue;
+                }
+                try {
+                    ExtensionSchemas.requireValidSchema((String) constraint.getData());
```
ivosh, line 813: **~ Minor** | Documentation

AttributeEngine.java:812 rejects a JSON_SCHEMA constraint whose `data` is null, while interfaces#910's `JsonSchemaAttributeConstraint.java:20-21` does not mark `data` REQUIRED or `@NotNull`; the generated contract therefore permits a request this PR rejects during registration. 

_Suggested:_ Mark `JsonSchemaAttributeConstraint.data` as REQUIRED and add `@NotNull` in interfaces#910.

src/main/java/com/otilm/core/util/AsnJsonCodec.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    public static byte[] encode(JsonNode tree) {
+        try {
+            return toAsn1(tree, "$").toASN1Primitive().getEncoded(ASN1Encoding.DER);
+        } catch (IOException e) {
+            throw new ValidationException("Extension value could not be encoded to DER");
+        } catch (ValidationException e) {
+            throw e;
+        } catch (RuntimeException e) {
```
ivosh, line 94: **~ Minor** | Error handling

src/main/java/com/otilm/core/util/AsnJsonCodec.java:73 catches every RuntimeException and replaces it without logging or preserving the cause. Unlike ExtensionSchemas.requireValidSchema, which logs unexpected compilation failures, this makes codec defects indistinguishable from invalid input, contrary to Effective Java Items 73 and 77.

_Suggested:_ Catch only expected Bouncy Castle input exceptions and log any unavoidable fallback RuntimeException before throwing the controlled ValidationException.

src/main/java/com/otilm/core/util/ExtensionSchemas.java
```diff
@@ -0,0 +1,234 @@
...
+                return List.of();
+            }
+            List<String> messages = new ArrayList<>();
+            for (ValidationMessage violation : schema.get().validate(value)) {
+                messages
+                        .add("does not match the registered schema for extension %s (at %s): %s"
+                                .formatted(oid, violation.getInstanceLocation(), violation.getMessage()));
+            }
+            return messages;
+        } catch (RuntimeException e) {
```
ivosh, line 273: **~ Minor** | Error handling

src/main/java/com/otilm/core/util/ExtensionSchemas.java:219 converts every RuntimeException into an unverifiable-schema message without recording the cause. The compilation path at lines 133–137 logs the same class of failure, while this path leaves operators unable to distinguish malformed stored data from a programming defect, contrary to Effective Java Item 77.

_Suggested:_ Log the caught exception with the extension OID before returning the controlled validation message.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
```diff
@@ -197,6 +197,20 @@ private static byte[] encodeExtensionValue(String value, ExtensionValueEncoding
...
     }
 
+    /**
+     * Encodes a structural ASN.1 JSON tree value. Wrong input is reported naming both accepted forms, because a
+     * DER-encoded extension takes either a JSON tree or base64 DER and the author needs to know which failed.
+     */
+    private static byte[] encodeJsonTree(String value) throws IOException {
+        try {
+            return AsnJsonCodec.encodeFromString(value);
+        } catch (RuntimeException e) {
```
ivosh, line 214: **~ Minor** | Error handling

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java:207 catches every RuntimeException, embeds its message in the IOException, and drops its cause. AsnJsonCodec exposes malformed input as ValidationException, so this broader catch can mislabel an unexpected defect and propagate internal text; decodeBase64Der at line 223 instead preserves the cause behind controlled text. This conflicts with Effective Java Items 73 and 77.

_Suggested:_ Catch ValidationException specifically and pass it as the cause of an IOException with controlled text.

ivosh (changes requested):

src/main/resources/extension-schemas/2.5.29.30.json
klaraf755: Fixed, and the recheck against the RFCs turned up more than the finding named — worth recording because the two reviews contradicted each other here and only the RFC settled it.

**GeneralSubtree minimum and maximum are now rejected outright.** RFC 5280 4.2.1.10: *"Within this profile, the minimum and maximum fields are not used with any name forms, thus, the minimum MUST be zero, and maximum MUST be absent."* Since minimum is DEFAULT 0 and DER forbids encoding a default, both members are absent and the base is a GeneralSubtree's only member. The previous revision *added* those slots on the earlier review's advice, which made the schema permit exactly what the profile forbids.

**Each GeneralName tag now pins its tagging and value type**, taken from the CHOICE rather than tag number alone, so an explicitly tagged dNSName or one carrying a utf8String is refused. directoryName `[4]` is the exception and requires explicit, because Name is a CHOICE and cannot be tagged implicitly.

Two more from the same pass, neither flagged:

**TLS Feature no longer requires a non-empty sequence.** I fetched RFC 7633: `Features ::= SEQUENCE OF INTEGER`, with no SIZE constraint. The earlier review asserted `SIZE (1..MAX)`, I took it on trust, and the resulting `minItems` was rejecting a legal encoding. Reverted.

**Key Usage rejects an all-zero bit string**, per 4.2.1.3 requiring at least one bit set — the codec already refused to read such a value while the schema let us write one.

Re-derived and left unchanged: Subject Directory Attributes `SEQUENCE SIZE (1..MAX)`, Basic Constraints pathLen >= 0 with cA required true, EKU `SIZE (1..MAX)`, Subject Key Identifier, Private Key Usage Period.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
klaraf755: Fixed. The renderer forwards only the codec's own `ValidationException` message, which is controlled and names the offending node; anything else gets fixed text with the cause attached rather than dropped.

src/main/java/com/otilm/core/util/AsnJsonCodec.java
klaraf755: Fixed — both the codec's catch-all and `validateShape`'s degradation log the cause before returning the controlled message, so a defect is distinguishable from bad input.

src/main/java/com/otilm/core/util/ExtensionSchemas.java
klaraf755: Fixed. The comment claiming eager `$ref` resolution was simply wrong — resolution is lazy, which is why an unloadable reference surfaces during validation. `requireValidSchema`'s contract now separates what is checked at save from what cannot be.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
klaraf755: Already addressed in the contract: `JsonSchemaAttributeConstraint.data` carries `requiredMode = REQUIRED` and `@NotBlank` in interfaces#910, so the generated contract no longer permits the null this rejects.

src/main/java/com/otilm/core/service/impl/CustomOidEntryServiceImpl.java
klaraf755: Tracked in fe-administrator#1982, which covers the OID form ("registering a certificate-extension OID accepts its JSON Schema") with a matching acceptance criterion. Regenerating the client surfaces the field in the typed model; the form control and detail row are hand-written and belong to that issue.

I have corrected three things in it that this backend work invalidated: TLS Feature is no longer a typed mapping target, the `readOnly` pinned-set reading is unreachable, and a constraint schema is an inline document only. Also noted there that the OID form's schema field is DER-only, per the `@AssertTrue` on the DTO.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
klaraf755: Shortened, here and in two other blocks with the same problem — including one I had made longer in an earlier commit. Each now leads with what the method does and leaves the caveat to its own line.

I did not adopt the bold concern labels: nothing else in this codebase uses them, so three methods doing so would be the inconsistency. At this length the paragraph break already separates the concerns.

src/test/java/com/otilm/core/attribute/engine/RequestAttributeDefinitionValidationTest.java
klaraf755: Removed, along with the other comments in this PR that restated a test name or narrated how an earlier revision behaved.

src/test/java/com/otilm/core/util/ExtensionSchemasTest.java
klaraf755: Removed the redundant given/then comments here and in `X509RequestContentRendererTest`, keeping only the ones that state something the test name and inputs do not — the `{`-discriminator note on the "neither form" case, for instance.

src/test/java/com/otilm/core/util/X509RequestContentRendererTest.java
klaraf755: Removed the redundant given/then comments here and in `X509RequestContentRendererTest`, keeping only the ones that state something the test name and inputs do not — the `{`-discriminator note on the "neither form" case, for instance.

src/test/java/com/otilm/core/util/ShippedExtensionSchemaTest.java
klaraf755: Removed the redundant given/then comments here and in `X509RequestContentRendererTest`, keeping only the ones that state something the test name and inputs do not — the `{`-discriminator note on the "neither form" case, for instance.

src/test/java/com/otilm/core/util/AsnJsonCodecTest.java
klaraf755: Removed the redundant given/then comments here and in `X509RequestContentRendererTest`, keeping only the ones that state something the test name and inputs do not — the `{`-discriminator note on the "neither form" case, for instance.

src/main/java/com/otilm/core/util/ExtensionSchemas.java
klaraf755: Fixed. `validateShape` now logs the caught exception with the extension OID before returning the controlled message, so malformed stored data is distinguishable from a defect. The degradation itself stays — a schema written straight into the table must not turn every request for that extension into a 500.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
klaraf755: Fixed as suggested. `encodeJsonTree` catches `ValidationException` specifically and forwards its message, which is controlled and names the offending node; any other `RuntimeException` gets fixed text with the exception passed as the cause, matching what `decodeBase64Der` already did.

src/main/java/com/otilm/core/util/AsnJsonCodec.java
klaraf755: Confirmed both and fixed in `543f03a98`. I had tightened the *schema* parses earlier in this PR but missed the *value* parse, which is the one operators actually type into.

Measured before the fix:

```
{"integer":1} garbage      -> {"integer":1}          trailing text discarded
{"integer":1,"integer":2}  -> {"integer":2}  size=1  first value discarded
```

The duplicate is the nastier of the two, and worse than the finding says: the collapsed object still has exactly one key, so it satisfies this grammar's one-key-per-node rule and encodes 2 without complaint. Nothing downstream could catch it.

`parse` now uses a reader with `FAIL_ON_TRAILING_TOKENS` and `STRICT_DUPLICATE_DETECTION`, with a rejection test for each.

src/main/resources/extension-schemas/2.5.29.15.json
```diff
@@ -0,0 +1,31 @@
...
+  "properties": {
+    "bitString": {
+      "type": "object",
+      "required": [
+        "value"
+      ],
+      "properties": {
+        "value": {
+          "type": "string",
+          "not": {
```
ivosh: **! Major** | Correctness

src/main/resources/extension-schemas/2.5.29.15.json:15 rejects only three literal spellings. Values such as `AAAA` decode to all-zero octets, pass this schema and `AsnJsonCodec`, and produce a Key Usage with no set bit, violating RFC 5280's requirement and potentially yielding a rejected or non-conformant certificate.

_Suggested:_ Add a Key Usage-specific semantic check that rejects every decoded all-zero bit string, with tests for bypasses such as `AAAA` and unpadded `AA`.

src/test/java/com/otilm/core/attribute/engine/RequestAttributeDefinitionValidationTest.java
```diff
@@ -67,6 +69,37 @@ static void restoreRdnRegistry() {
...
+        DataAttributeV3 definition = validDefinition();
+        JsonSchemaAttributeConstraint constraint = new JsonSchemaAttributeConstraint();
+        constraint.setData("this is not json");
+        definition.setConstraints(List.of(constraint));
+        assertRejected(definition, "valid JSON Schema document");
+    }
+
+    @Test
+    void jsonSchemaConstraintCheckedEvenWithoutAFieldMapping() {
+        // A connector attribute needs the check whether or not it declares a field mapping.
```
ivosh: **. Nitpick** | Maintainability

RequestAttributeDefinitionValidationTest.java:83 merely restates `jsonSchemaConstraintCheckedEvenWithoutAFieldMapping` without explaining an additional constraint or rationale.

_Suggested:_ Remove the redundant comment.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
ivosh: **Re-check: partial.**

encodeJsonTree() at X509RequestContentRenderer.java:211-213 catches ValidationException and preserves it as the IOException cause, but lines 214-217 still catch every other RuntimeException and label it with WRONG_DER_FORM.

Cause preservation and controlled text were fixed, but unexpected defects are still translated as invalid DER input.

ivosh (changes requested):

src/main/resources/extension-schemas/2.5.29.15.json
klaraf755: You are right that this is a bypass, and the fix is to remove the schema rather than strengthen it — the whole approach was wrong.

A bit string's content cannot be constrained from JSON Schema. Enumerating three base64 spellings of all-zero was the best that could be done and, as you show, it misses `AAAA` and every other spelling.

The deeper point is that this OID should not accept a JSON tree at all. Key Usage and Extended Key Usage have typed mapping targets that draw on a closed vocabulary — `CertificateKeyUsage` and the EKU OID registry — so they cannot express a malformed value. Authoring a *new* opaque mapping for either OID is already refused with "use the keyUsage mapping target instead". A hand-written tree was a second, weaker way in, reachable only through a legacy definition, and it accepted values the target itself cannot produce.

So in `100f725cc`:

- Both shipped schemas are gone. `2.5.29.15` and `2.5.29.37` no longer have one.
- A JSON tree value for an OID with a typed target is refused outright, naming the target to use.
- The completeness sweep excludes those OIDs, so the build no longer demands schemas for them.

Base64 DER through a legacy mapping is unaffected — only the `{`-prefixed form is refused. Nothing existing can break, since the tree form is new in this PR.

This removes the all-zero problem rather than narrowing it, and it removes the need for an OID-specific semantic check in a codec that is otherwise OID-agnostic.

src/main/java/com/otilm/core/util/X509RequestContentRenderer.java
klaraf755: Fair — cause preservation was only half of it. An unexpected defect was still labelled as invalid DER input, which would send the author to inspect a value that is fine.

Separated: a `ValidationException` keeps the `WRONG_DER_FORM` text naming both accepted forms, and any other `RuntimeException` now reports "Extension value could not be encoded" with the exception as the cause.

src/test/java/com/otilm/core/attribute/engine/RequestAttributeDefinitionValidationTest.java
klaraf755: Removed.

src/main/resources/extension-schemas/2.5.29.30.json
```diff
@@ -0,0 +1,449 @@
...
+                        "tagNo": {
+                          "const": 0
+                        },
+                        "explicit": {
+                          "const": false
+                        },
+                        "value": {
+                          "type": "object",
+                          "required": [
+                            "sequence"
```
ivosh, line 194: **~ Minor** | Correctness

2.5.29.30.json:191-196 accepts any sequence for otherName, including an empty one, although OtherName requires an OID and a tagged value. Likewise, 2.5.29.30.json:401-405 accepts any octet-string length for iPAddress, while Name Constraints requires exactly 8 or 32 octets. This PR can therefore compose invalid Name Constraints despite passing its registry schema.

_Suggested:_ Fully constrain each GeneralName alternative's inner structure, or reject alternatives whose valid structure is not yet modeled.

src/main/java/com/otilm/core/util/ExtensionSchemas.java
```diff
@@ -0,0 +1,255 @@
...
+
+    /**
+     * Rejects a {@code $ref} that points outside the document. Such a reference cannot be resolved without fetching it,
+     * which the platform does not do, so accepting one would register a schema that silently constrains nothing.
+     */
+    private static void rejectNonLocalRefs(JsonNode node, String path) {
+        if (node == null || !node.isObject()) {
+            return;
+        }
+        JsonNode ref = node.get("$ref");
```
ivosh: **~ Minor** | Correctness

ExtensionSchemas.java:195 checks only `$ref`, but draft 2020-12 also defines `$dynamicRef`, which networknt resolves lazily. A remote `$dynamicRef` therefore passes this PR's registration validation and fails only when an extension value is first validated as an unloadable schema.

_Suggested:_ Reject non-local `$dynamicRef` values alongside `$ref` and add a registration-time test.

src/main/java/com/otilm/core/util/ExtensionSchemas.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    /**
+     * Rejects a schema document that cannot be trusted to constrain anything: it must parse as exactly one JSON value,
+     * declare no dialect but draft 2020-12, reference nothing outside itself, carry well-formed keywords, and compile.
+     *
+     * <p>
+     * A reference resolved only on use escapes all of that; {@link #validateShape} reports it as unverifiable.
+     */
+    public static void requireValidSchema(String schemaDocument) {
```
ivosh, line 132: **~ Minor** | Correctness

ExtensionSchemas.java:136 enables trailing-token checks but not duplicate-key detection, so a document such as `{"required":["a"],"required":["b"]}` silently loses its first constraint. This PR correctly prevents the identical loss in AsnJsonCodec, but registration still accepts a schema different from what the operator authored.

_Suggested:_ Parse schema documents with strict duplicate detection and reject repeated members at registration.

src/main/java/com/otilm/core/util/ExtensionSchemas.java
```diff
@@ -0,0 +1,255 @@
...
+                    "Not a valid JSON Schema document: $ref at %s points outside the document; only local references such as #/$defs/name are supported"
+                            .formatted(path));
+        }
+        for (Map.Entry<String, JsonNode> property : node.properties()) {
+            String childPath = path + "." + property.getKey();
+            subschemasOf(property.getKey(), property.getValue()).forEach(child -> rejectNonLocalRefs(child, childPath));
+        }
+    }
+
+    /** The subschemas a keyword's value holds, or nothing when the keyword does not hold subschemas. */
```
ivosh: **. Nitpick** | Maintainability

ExtensionSchemas.java:207 paraphrases `subschemasOf` without explaining a non-obvious constraint or rationale, so it adds no information beyond the method and implementation.

_Suggested:_ Remove this redundant Javadoc.

src/main/java/com/otilm/core/service/impl/CustomOidEntryServiceImpl.java
ivosh: **Re-check: not addressed.**

PR #2156 still references only OmniTrustILM/interfaces#910; the fix delta contains only Core backend files, and no corresponding fe-administrator PR was found.

The requested frontend DTO, editor, submission mapping, and detail display were not coordinated.

ivosh (dismissed):

src/main/resources/extension-schemas/2.5.29.30.json
klaraf755: Both fixed in `e2311fbb6`, verified against networknt.

**otherName** now requires the OID plus the explicitly tagged value, exactly two members and closed with `items: false`. An empty sequence and an OID on its own are both refused.

**iPAddress** is constrained to 8 or 32 octets, per RFC 5280 4.2.1.10: *"the iPAddress field of GeneralName MUST contain eight (8) octets, encoded in the style of RFC 4632 (CIDR)... For IPv6 addresses, the iPAddress field MUST contain 32 octets"* — address plus mask, not a bare address. Expressed as base64 lengths of 12 and 44, which is what those octet counts encode to.

Six cases checked, including that a 4-octet address is now refused and dNSName is unaffected.

src/main/java/com/otilm/core/util/ExtensionSchemas.java
klaraf755: Correct, and fixed in `e2311fbb6`. `$dynamicRef` is walked alongside `$ref` now, with a test that a remote one is refused at registration rather than at first use.

The same pass covered a related gap you did not raise: a `$ref` without a leading `#` is not always external. An empty reference means the document root, and an absolute one matching the document's own `$id` resolves inside it — both were being refused. A nested `$id` re-scopes the base URI, which the check still does not follow; such a reference falls through to `DisallowSchemaLoader`, so it fails later rather than being missed.

src/main/java/com/otilm/core/util/ExtensionSchemas.java
klaraf755: Fixed in `e2311fbb6` — the schema parse now uses `STRICT_DUPLICATE_DETECTION` as well as trailing-token detection.

You are right that this was the same loss in two places and I only closed one. I added the strict reader to `AsnJsonCodec` for *values* after an earlier finding, and did not carry it to the schema parse, so a document could still lose a constraint silently at registration.

ivosh (approved):

PR #2158, opened 2026-08-26
Give each user named views of an inventory listing

### LINKED ISSUE #2034: Per-user saved views: storage, API and orphan cleanup
### Description

A view is what a user sees when they open an inventory: a named set of columns, the filter that narrows the rows, and the ordering applied to them. Core owns the storage, because a view has to follow the user across sessions, browsers and platform restarts.

Scope:
- A new table keyed by the user UUID from the token, with no foreign key - Core has no user entity, users live behind `UserManagementApiClient`. Follow the existing FK-less `user_uuid` precedent in `scheduler`, `internal_notifications` and `approval_profiles`. Ships as a Flyway migration.
- Implement `ListViewController` (`/v1/listViews`): list filtered by resource, create, update, delete. Every operation is scoped to the authenticated user; a view is never visible or addressable by another user.
- Store the ordered column list, each column carrying its field source, field identifier and an optional label override. Column order is the stored array order and is display order.
- Store and return the view's filters and its ordering. A view restores columns, filters and ordering together.
- Resolve stored column identifiers against the live field catalogue on read, so a renamed or deleted attribute degrades to a skipped column instead of requiring stored views to be migrated.
- Enforce name uniqueness per user and resource, and at most one default view per user and resource - marking one default clears the flag on the previous one.
- Audit-log the endpoints against `Resource.LIST_VIEW`.
- Clean up rows belonging to users deleted from the identity service, by a defined mechanism.

The resource is settable only on create: a view's columns are resolved against one resource's catalogue for its whole life.

### Acceptance Criteria

1. [ ] A view survives logout, a different browser and a platform restart, and is visible only to the user who created it.
2. [ ] Views can be listed for a resource, created, renamed, edited and deleted; a second view of the same name for the same user and resource is rejected.
3. [ ] Column order round-trips exactly as sent, and a per-column label override round-trips including being cleared back to the catalogue label.
4. [ ] A view's filters and ordering are stored and returned, and applying a view applies all three together.
5. [ ] At most one view per user and resource is the default; marking a new one clears the previous.
6. [ ] A column naming an attribute that no longer exists is skipped on read without failing the request.
7. [ ] Rows for a user deleted from the identity service are cleaned up, with the mechanism documented.
8. [ ] Sonar quality gate passes and new code has at least 80% coverage.

Jakub-Bartal-01:
Closes #2034

### DIFF
--- a/src/main/java/com/otilm/core/dao/entity/ListView.java
+++ b/src/main/java/com/otilm/core/dao/entity/ListView.java
@@ -20,27 +20,24 @@ import org.hibernate.annotations.JdbcTypeCode;
 import org.hibernate.type.SqlTypes;
 
 /**
- * A named set of columns, filters and ordering that one user has saved for one listing.
+ * A named set of columns, filters and ordering that one user has saved for one listing. The owning user is a bare UUID
+ * with no foreign key - users live in the identity service - so the rows are removed where the platform learns of the
+ * deletion, in {@link com.otilm.core.service.UserManagementExternalService#deleteUser(String)}.
  *
  * <p>
- * The owning user is a bare UUID with no foreign key: users live in the identity service, not in this database, which
- * is the same arrangement {@code scheduled_job}, {@code notification_recipient} and {@code approval_recipient} already
- * use. Rows are removed when the user is deleted through
- * {@link com.otilm.core.service.UserManagementExternalService#deleteUser(String)}.
- *
- * <p>
- * Columns, filters and ordering are stored as the request shapes the API accepts, and column identifiers are resolved
- * against the resource's live field catalogue on read - a renamed or deleted attribute drops out of the view instead of
- * requiring stored rows to be migrated.
+ * Field identifiers are resolved against the resource's live field catalogue on read, so a renamed or deleted attribute
+ * drops out of the view instead of requiring stored rows to be migrated.
  */
 @Setter
 @Getter
 @ToString
 @RequiredArgsConstructor
 @Entity
-@Table(name = "list_view", uniqueConstraints = @UniqueConstraint(name = "uk_list_view_user_resource_name",
+@Table(name = "list_view", uniqueConstraints = @UniqueConstraint(name = ListView.UNIQUE_NAME_CONSTRAINT,
         columnNames = {"user_uuid", "resource", "name"}))
 public class ListView extends UniquelyIdentifiedAndAudited {
+
+    public static final String UNIQUE_NAME_CONSTRAINT = "uk_list_view_user_resource_name";
 
     @Column(name = "user_uuid", nullable = false, updatable = false)
     private UUID userUuid;
@@ -71,4 +68,16 @@ public class ListView extends UniquelyIdentifiedAndAudited {
     @Column(name = "sort", columnDefinition = "jsonb")
     @JdbcTypeCode(SqlTypes.JSON)
     private SearchSortRequestDto sort;
+
+    // No-op overrides required by S2160: identity and hashing stay UUID-based, and the added columns never
+    // affect equality.
+    @Override
+    public boolean equals(Object o) {
+        return super.equals(o);
+    }
+
+    @Override
+    public int hashCode() {
+        return super.hashCode();
+    }
 }
--- a/src/main/java/com/otilm/core/dao/entity/ListView.java
+++ b/src/main/java/com/otilm/core/dao/entity/ListView.java
@@ -1,0 +1,74 @@
+package com.otilm.core.dao.entity;
+
+import com.otilm.api.model.client.certificate.SearchFilterRequestDto;
+import com.otilm.api.model.client.certificate.SearchSortRequestDto;
+import com.otilm.api.model.core.auth.Resource;
+import com.otilm.api.model.core.listview.ListViewColumnDto;
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.EnumType;
+import jakarta.persistence.Enumerated;
+import jakarta.persistence.Table;
+import jakarta.persistence.UniqueConstraint;
+import java.util.List;
+import java.util.UUID;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+import org.hibernate.annotations.JdbcTypeCode;
+import org.hibernate.type.SqlTypes;
+
+/**
+ * A named set of columns, filters and ordering that one user has saved for one listing.
+ *
+ * <p>
+ * The owning user is a bare UUID with no foreign key: users live in the identity service, not in this database, which
+ * is the same arrangement {@code scheduled_job}, {@code notification_recipient} and {@code approval_recipient} already
+ * use. Rows are removed when the user is deleted through
+ * {@link com.otilm.core.service.UserManagementExternalService#deleteUser(String)}.
+ *
+ * <p>
+ * Columns, filters and ordering are stored as the request shapes the API accepts, and column identifiers are resolved
+ * against the resource's live field catalogue on read - a renamed or deleted attribute drops out of the view instead of
+ * requiring stored rows to be migrated.
+ */
+@Setter
+@Getter
+@ToString
+@RequiredArgsConstructor
+@Entity
+@Table(name = "list_view", uniqueConstraints = @UniqueConstraint(name = "uk_list_view_user_resource_name",
+        columnNames = {"user_uuid", "resource", "name"}))
+public class ListView extends UniquelyIdentifiedAndAudited {
+
+    @Column(name = "user_uuid", nullable = false, updatable = false)
+    private UUID userUuid;
+
+    @Column(name = "resource", nullable = false, updatable = false)
+    @Enumerated(EnumType.STRING)
+    private Resource resource;
+
+    @Column(name = "name", nullable = false, length = 255)
+    private String name;
+
+    @Column(name = "default_view", nullable = false)
+    private boolean defaultView;
+
+    // S1948: entities are Serializable through UniquelyIdentifiedObject, but nothing Java-serializes them - Jackson
+    // owns the persistence shape of these JSONB fields.
+    @SuppressWarnings("java:S1948")
+    @Column(name = "columns", nullable = false, columnDefinition = "jsonb")
+    @JdbcTypeCode(SqlTypes.JSON)
+    private List<ListViewColumnDto> columns;
+
+    @SuppressWarnings("java:S1948")
+    @Column(name = "filters", columnDefinition = "jsonb")
+    @JdbcTypeCode(SqlTypes.JSON)
+    private List<SearchFilterRequestDto> filters;
+
+    @SuppressWarnings("java:S1948")
+    @Column(name = "sort", columnDefinition = "jsonb")
+    @JdbcTypeCode(SqlTypes.JSON)
+    private SearchSortRequestDto sort;
+}
--- a/src/main/java/com/otilm/core/dao/repository/ListViewRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/ListViewRepository.java
@@ -1,0 +1,41 @@
+package com.otilm.core.dao.repository;
+
+import com.otilm.api.model.core.auth.Resource;
+import com.otilm.core.dao.entity.ListView;
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Modifying;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.query.Param;
+import org.springframework.stereotype.Repository;
+
+/**
+ * Views are private to the user who saved them, so every finder is keyed by the user UUID rather than routed through
+ * the secured-search repository: there is no object-level permission to evaluate, and no view is addressable by anyone
+ * else.
+ */
+@Repository
+public interface ListViewRepository extends JpaRepository<ListView, UUID> {
+
+    List<ListView> findByUserUuidOrderByNameAsc(UUID userUuid);
+
+    List<ListView> findByUserUuidAndResourceOrderByNameAsc(UUID userUuid, Resource resource);
+
+    Optional<ListView> findByUuidAndUserUuid(UUID uuid, UUID userUuid);
+
+    boolean existsByUserUuidAndResourceAndName(UUID userUuid, Resource resource, String name);
+
+    boolean existsByUserUuidAndResourceAndNameAndUuidNot(UUID userUuid, Resource resource, String name, UUID uuid);
+
+    @Modifying
+    @Query("UPDATE ListView v SET v.defaultView = FALSE WHERE v.userUuid = :userUuid AND v.resource = :resource "
+            + "AND v.defaultView = TRUE AND v.uuid <> :keptUuid")
+    int clearDefaultExcept(@Param("userUuid") UUID userUuid, @Param("resource") Resource resource,
+            @Param("keptUuid") UUID keptUuid);
+
+    @Modifying
+    @Query("DELETE FROM ListView v WHERE v.userUuid = :userUuid")
+    int deleteByUserUuid(@Param("userUuid") UUID userUuid);
+}
--- a/src/main/java/com/otilm/core/dao/repository/ListViewRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/ListViewRepository.java
@@ -30,8 +30,9 @@ public interface ListViewRepository extends JpaRepository<ListView, UUID> {
     boolean existsByUserUuidAndResourceAndNameAndUuidNot(UUID userUuid, Resource resource, String name, UUID uuid);
 
     @Modifying
-    @Query("UPDATE ListView v SET v.defaultView = FALSE WHERE v.userUuid = :userUuid AND v.resource = :resource "
-            + "AND v.defaultView = TRUE AND v.uuid <> :keptUuid")
+    @Query("UPDATE ListView v SET v.defaultView = FALSE, v.updated = CURRENT_TIMESTAMP "
+            + "WHERE v.userUuid = :userUuid AND v.resource = :resource AND v.defaultView = TRUE "
+            + "AND v.uuid <> :keptUuid")
     int clearDefaultExcept(@Param("userUuid") UUID userUuid, @Param("resource") Resource resource,
             @Param("keptUuid") UUID keptUuid);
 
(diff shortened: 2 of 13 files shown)


src/main/java/com/otilm/core/service/writer/ListViewWriter.java
```diff
@@ -0,0 +1,46 @@
...
+     * default row exists, so demoting afterwards would never be reached. The uuid is assigned here rather than at
+     * persist time so the statement that clears the others can name the row being kept.
+     */
+    @Transactional
+    public ListView save(ListView view) {
+        if (view.getUuid() == null) {
+            view.setUuid(UUID.randomUUID());
+        }
+        if (view.isDefaultView()) {
+            listViewRepository.clearDefaultExcept(view.getUserUuid(), view.getResource(), view.getUuid());
```
lubomirw, line 40: **[High]** | Correctness

When `editView` marks an OSIV-managed entity as default, the bulk demotion query can auto-flush that dirty entity first and violate `uk_list_view_single_default` before the previous default is cleared. The current test calls the service directly and uses Hibernate DDL without the partial index, so it does not exercise production behavior.

_Suggested:_ Demote and flush the previous default inside the writer before setting the edited entity's default flag (or detach/copy the incoming entity), and test against the Flyway-managed partial index.

src/main/java/com/otilm/core/service/impl/UserManagementServiceImpl.java
```diff
@@ -233,6 +240,7 @@ public void deleteUser(String userUuid) {
         UUID uuid = UUID.fromString(userUuid);
         certificateService.removeCertificateUser(uuid);
         objectAssociationService.removeOwnerAssociations(uuid);
+        listViewService.deleteViewsOfUser(uuid);
```
lubomirw, line 243: **[High]** | Correctness

The class-level transaction spans the external identity deletion and this local cleanup. If this or a later local step fails, the database transaction rolls back saved-view cleanup after the identity service has permanently deleted the user.

_Suggested:_ Run deletion orchestration without an ambient transaction, call the external identity service first, and perform local cleanup in short transactions after external success.

<sub>flagged by Guided PR Review</sub>

src/main/java/com/otilm/core/service/impl/ListViewServiceImpl.java
```diff
@@ -0,0 +1,221 @@
...
+    @Override
+    public int deleteViewsOfUser(UUID userUuid) {
+        return listViewWriter.deleteAllForUser(userUuid);
+    }
+
+    private static void applyRequest(ListView view, ListViewUpdateRequestDto request) {
+        view.setName(request.getName());
+        view.setColumns(request.getColumns());
+        view.setDefaultView(request.isDefaultView());
+        view.setFilters(request.getFilters() == null || request.getFilters().isEmpty() ? null : request.getFilters());
```
lubomirw, line 179: **[High]** | Correctness

Only columns are validated against the resource catalogue. Filters and sorting are persisted even when their fields belong to another resource, their conditions are unsupported, or the sort field is not sortable, leaving a saved view that its listing cannot apply.

_Suggested:_ Validate filter fields and conditions plus sort fields against the same resource catalogue and capability flags used by the listing search endpoint before saving.

<sub>flagged by Guided PR Review</sub>

src/main/java/com/otilm/core/service/impl/ListViewServiceImpl.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    @Override
+    @AnyPrincipalEndpoint
+    public ListViewDto createView(ListViewRequestDto request) throws AlreadyExistException {
+        UUID userUuid = loggedUserUuid();
+        Resource resource = request.getResource();
+        validateColumns(resource, request.getColumns());
+
+        if (listViewRepository.existsByUserUuidAndResourceAndName(userUuid, resource, request.getName())) {
```
lubomirw, line 95: **[Low]** | Error handling

The name existence check races with concurrent create or rename requests. The unique constraint rejects one request, but the resulting persistence exception escapes as HTTP 500 instead of the documented 409.

_Suggested:_ Flush the write inside the transaction and translate the named unique-constraint violation to `AlreadyExistException`, with a concurrent regression test.

<sub>flagged by Guided PR Review and Superpowers</sub>

src/main/java/com/otilm/core/service/writer/ListViewWriter.java
lubomirw, line 40: **[Low]** | Error handling

Concurrent requests that promote different defaults can both clear the prior default and then collide on the partial unique index, producing an uncontrolled HTTP 500.

_Suggested:_ Serialize default replacement per user and resource with a transaction-level lock, or retry/translate the constraint collision to a controlled conflict response.

<sub>flagged by Guided PR Review and Superpowers</sub>

src/main/java/com/otilm/core/dao/repository/ListViewRepository.java
```diff
@@ -0,0 +1,41 @@
...
+    List<ListView> findByUserUuidAndResourceOrderByNameAsc(UUID userUuid, Resource resource);
+
+    Optional<ListView> findByUuidAndUserUuid(UUID uuid, UUID userUuid);
+
+    boolean existsByUserUuidAndResourceAndName(UUID userUuid, Resource resource, String name);
+
+    boolean existsByUserUuidAndResourceAndNameAndUuidNot(UUID userUuid, Resource resource, String name, UUID uuid);
+
+    @Modifying
+    @Query("UPDATE ListView v SET v.defaultView = FALSE WHERE v.userUuid = :userUuid AND v.resource = :resource "
```
lubomirw: **[Low]** | Correctness

The bulk default-clearing update bypasses entity auditing and leaves the inherited `i_upd` timestamp unchanged on demoted views.

_Suggested:_ Set `v.updated = CURRENT_TIMESTAMP` in the JPQL update.

<sub>flagged by Guided PR Review</sub>

src/main/java/com/otilm/core/dao/entity/ListView.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.UUID;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+import org.hibernate.annotations.JdbcTypeCode;
+import org.hibernate.type.SqlTypes;
+
+/**
+ * A named set of columns, filters and ordering that one user has saved for one listing.
```
lubomirw: **[Nit]** | Maintainability

The class Javadoc expands into persistence, cleanup, and migration implementation narration instead of a concise entity contract.

_Suggested:_ Condense it to two sentences covering user ownership, the absent foreign key and deletion cleanup, and live field resolution.

<sub>flagged by Hygiene</sub>

src/main/java/com/otilm/core/service/writer/ListViewWriter.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private final ListViewRepository listViewRepository;
+
+    @Autowired
+    public ListViewWriter(ListViewRepository listViewRepository) {
+        this.listViewRepository = listViewRepository;
+    }
+
+    /**
+     * Writes the view, and where it claims to be the default clears the flag on the user's previous default for the
```
lubomirw: **[Nit]** | Maintainability

The save-method Javadoc narrates execution order and UUID mechanics across several sentences.

_Suggested:_ Replace it with two sentences stating that prior defaults are cleared before saving and that a missing UUID is assigned first so the retained row can be excluded.

<sub>flagged by Hygiene</sub>

lubomirw (changes requested):
Automated multi-reviewer triage - 8 findings (REQUEST_CHANGES).

src/main/java/com/otilm/core/service/impl/UserManagementServiceImpl.java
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed, but not by restructuring `deleteUser`.

`deleteViewsOfUser` now runs in its own `REQUIRES_NEW` transaction, so the rows stay removed once the identity service has acknowledged the deletion, whatever a later local step does. The propagation sits on `ListViewServiceImpl`, not on the writer: `TransactionalBoundaryArchTest` Rule D requires every public `..service.writer..` method to be `@Transactional(REQUIRED)`, and putting `REQUIRES_NEW` there fails that rule. `viewsOfADeletedUserStayRemovedWhenTheSurroundingTransactionRollsBack` covers it - it fails with plain `REQUIRED`.

Reshaping `deleteUser`'s own boundary is out of scope here. The class-level transaction spanning `removeUser` predates this PR and equally spans `certificateService.removeCertificateUser`, `objectAssociationService.removeOwnerAssociations` and `attributeEngine.deleteObjectAttributeContent`; this PR adds one line to that existing sequence. Splitting the orchestration is a change to user deletion as a whole and belongs in its own PR.

Jakub-Bartal-01: Fixed, but not by restructuring `deleteUser`.

`deleteViewsOfUser` now runs in its own `REQUIRES_NEW` transaction, so the rows stay removed once the identity service has acknowledged the deletion, whatever a later local step does. The propagation sits on `ListViewServiceImpl`, not on the writer: `TransactionalBoundaryArchTest` Rule D requires every public `..service.writer..` method to be `@Transactional(REQUIRED)`, and putting `REQUIRES_NEW` there fails that rule. `viewsOfADeletedUserStayRemovedWhenTheSurroundingTransactionRollsBack` covers it - it fails with plain `REQUIRED`.

Reshaping `deleteUser`'s own boundary is out of scope here. The class-level transaction spanning `removeUser` predates this PR and equally spans `certificateService.removeCertificateUser`, `objectAssociationService.removeOwnerAssociations` and `attributeEngine.deleteObjectAttributeContent`; this PR adds one line to that existing sequence. Splitting the orchestration is a change to user deletion as a whole and belongs in its own PR.

src/main/java/com/otilm/core/service/impl/ListViewServiceImpl.java
```diff
@@ -0,0 +1,221 @@
...
+            throw new AlreadyExistException(ListView.class, request.getName());
+        }
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed by serializing, not just by translating.

`createView` and `editView` are now `@Transactional`, and both take `ClusterOperationSynchronizer.lock("list-view:<user>:<resource>")` before the name check - the same transaction-scoped advisory lock `SigningProfileServiceImpl.persistUpdate` uses to serialize a read-decide-write section. The check and the insert therefore share one transaction that no second writer can interleave with.

The unique-name violation is still translated to `AlreadyExistException` as a backstop, in case a writer ever bypasses the lock; other integrity violations surface unchanged. `concurrentCreatesOfTheSameNameLeaveOneViewAndOneRejection` runs two threads at the same name and asserts one view and one `AlreadyExistException` - never a data-integrity failure.

src/main/java/com/otilm/core/service/impl/ListViewServiceImpl.java
Jakub-Bartal-01: Fixed by serializing, not just by translating.

`createView` and `editView` are now `@Transactional`, and both take `ClusterOperationSynchronizer.lock("list-view:<user>:<resource>")` before the name check - the same transaction-scoped advisory lock `SigningProfileServiceImpl.persistUpdate` uses to serialize a read-decide-write section. The check and the insert therefore share one transaction that no second writer can interleave with.

The unique-name violation is still translated to `AlreadyExistException` as a backstop, in case a writer ever bypasses the lock; other integrity violations surface unchanged. `concurrentCreatesOfTheSameNameLeaveOneViewAndOneRejection` runs two threads at the same name and asserts one view and one `AlreadyExistException` - never a data-integrity failure.

src/main/java/com/otilm/core/service/writer/ListViewWriter.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    @Transactional
+    public ListView save(ListView view) {
+        if (view.getUuid() == null) {
+            view.setUuid(UUID.randomUUID());
+        }
+        if (view.isDefaultView()) {
+            listViewRepository.clearDefaultExcept(view.getUserUuid(), view.getResource(), view.getUuid());
+        }
+        return listViewRepository.save(view);
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed by the same serialization as the name race: `createView` and `editView` hold `ClusterOperationSynchronizer.lock("list-view:<user>:<resource>")` for the rest of the transaction, so clear-then-save cannot interleave with another promotion for that user and listing. `uk_list_view_single_default` stays as the backstop it was described as.

The related ordering bug in this method was real and is fixed separately - see the thread on the auto-flush.

src/main/java/com/otilm/core/service/writer/ListViewWriter.java
Jakub-Bartal-01: Fixed by the same serialization as the name race: `createView` and `editView` hold `ClusterOperationSynchronizer.lock("list-view:<user>:<resource>")` for the rest of the transaction, so clear-then-save cannot interleave with another promotion for that user and listing. `uk_list_view_single_default` stays as the backstop it was described as.

The related ordering bug in this method was real and is fixed separately - see the thread on the auto-flush.

Jakub-Bartal-01: Confirmed and fixed.

I checked the auto-flush claim against Hibernate 6.6.53 rather than taking it as given: `StandardJdbcMutationExecutor.execute` calls `session.autoFlushIfRequired(getAffectedTableNames())` as its first action, so a bulk JPQL update on `list_view` does flush a dirty managed `ListView` first. `ListViewWriter.save` now detaches the incoming view before demoting, so the demotion runs first and the promotion is written by the merge that follows.

`ListViewServiceITest` now creates `uk_list_view_single_default` itself in `@BeforeEach` - the partial index has no entity counterpart, so `ddl-auto` cannot produce it - and `promotingAnAlreadyLoadedViewToDefaultKeepsOneDefault` promotes a view loaded in the same transaction, which is the shape a request has. Without the detach both that test and the existing `markingAViewDefaultClearsThePreviousDefault` fail with `duplicate key value violates unique constraint "uk_list_view_single_default"`.

src/main/java/com/otilm/core/service/impl/ListViewServiceImpl.java
```diff
@@ -0,0 +1,221 @@
...
+    public int deleteViewsOfUser(UUID userUuid) {
+        return listViewWriter.deleteAllForUser(userUuid);
+    }
+
+    private static void applyRequest(ListView view, ListViewUpdateRequestDto request) {
+        view.setName(request.getName());
+        view.setColumns(request.getColumns());
+        view.setDefaultView(request.isDefaultView());
+        view.setFilters(request.getFilters() == null || request.getFilters().isEmpty() ? null : request.getFilters());
+        view.setSort(request.getSort());
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed for fields and conditions; pushing back on sortability.

Filter field source and identifier, and the sort field, are now validated against the same catalogue as the columns, so a certificate view carrying a `CKI_NAME` filter or ordering is rejected on write. Filter conditions are validated too, against the operator set the listing advertises for that field: `SearchHelper.availableConditions(FilterField)` is extracted from `prepareSearch` so both callers read from one source, and attribute-sourced fields use the `conditions` already on their `SearchFieldDataDto`. Three tests cover it.

Sortability cannot be validated on this branch. `SearchFieldDataDto.sortable` and `displayable` exist on the DTO but nothing in core populates them yet - `grep -rn "setSortable\|setDisplayable" src/main/java` is empty - so every catalogue entry has `sortable == null` and a "must be sortable" check would reject every ordering. That check belongs with the change that starts populating the flags (core#2031/#2033), which is also what would let the column validation move from "the field exists" to "the field is displayable".

src/main/java/com/otilm/core/service/impl/ListViewServiceImpl.java
Jakub-Bartal-01: Fixed for fields and conditions; pushing back on sortability.

Filter field source and identifier, and the sort field, are now validated against the same catalogue as the columns, so a certificate view carrying a `CKI_NAME` filter or ordering is rejected on write. Filter conditions are validated too, against the operator set the listing advertises for that field: `SearchHelper.availableConditions(FilterField)` is extracted from `prepareSearch` so both callers read from one source, and attribute-sourced fields use the `conditions` already on their `SearchFieldDataDto`. Three tests cover it.

Sortability cannot be validated on this branch. `SearchFieldDataDto.sortable` and `displayable` exist on the DTO but nothing in core populates them yet - `grep -rn "setSortable\|setDisplayable" src/main/java` is empty - so every catalogue entry has `sortable == null` and a "must be sortable" check would reject every ordering. That check belongs with the change that starts populating the flags (core#2031/#2033), which is also what would let the column validation move from "the field exists" to "the field is displayable".

src/main/java/com/otilm/core/dao/repository/ListViewRepository.java
```diff
@@ -0,0 +1,41 @@
...
+            + "AND v.defaultView = TRUE AND v.uuid <> :keptUuid")
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed: `clearDefaultExcept` now sets `v.updated = CURRENT_TIMESTAMP`, matching the other bulk updates in `..dao.repository..`. `demotingAViewAdvancesItsAuditTimestamp` asserts the demoted view's `i_upd` moves.

src/main/java/com/otilm/core/dao/repository/ListViewRepository.java
Jakub-Bartal-01: Fixed: `clearDefaultExcept` now sets `v.updated = CURRENT_TIMESTAMP`, matching the other bulk updates in `..dao.repository..`. `demotingAViewAdvancesItsAuditTimestamp` asserts the demoted view's `i_upd` moves.

src/main/java/com/otilm/core/dao/entity/ListView.java
Jakub-Bartal-01: Condensed to user ownership, the absent foreign key and where the rows are cleaned up, then live field resolution.

src/main/java/com/otilm/core/service/writer/ListViewWriter.java
Jakub-Bartal-01: Condensed. It now states that the previous default is cleared before the write and why the uuid is assigned first, and the detach that the ordering depends on carries its own note.

lubomirw (approved):

Issue #2017, opened 2026-08-11
Spring Data JPA and Hibernate 7 behaviour migration


ivosh:
### Description

Migrate the persistence layer's BEHAVIOUR to Hibernate 7.4.5 / Spring Data 2026.0.0. The
mechanical half (processor rename, plugin relocation) lands with the #2016 baseline. What
this issue owns is the largest cluster of real work in the migration: on the measured
baseline of 5661 tests / 125 failing, roughly 82 failures are Hibernate, and two of the
three items are production breaks rather than test churn.

Each item below has a standalone reproducer that runs on both Hibernate 6.6.53.Final and
7.4.5.Final, so the behaviour change is demonstrated rather than asserted.

---

### RC#7 - Criteria no longer resolves subclass attributes (27 tests) - FIXED

`altCodes` and `code` are declared on the subclass `RdnAttributeTypeCustomOidEntry`;
`CustomOidEntryRepository` roots its criteria at the base `CustomOidEntry`, and
`FilterPredicatesBuilder` resolved them by name against that root. Hibernate 7 throws
`PathElementException: Could not resolve attribute 'altCodes' of ...CustomOidEntry`.

**This is intentional, not a regression**, and is documented in the Hibernate 7.0
migration guide, section *Criteria API and Implicit Treats*
(https://docs.hibernate.org/orm/7.0/migration-guide/):

> Since Hibernate 7.0, aligning with the JPA specification, the Criteria API will no longer
> allow retrieving subtype attributes this way, and it's going to require an explicit
> `jakarta.persistence.criteria.CriteriaBuilder#treat` to be called on the path first to
> downcast it to the subtype which defines the attribute. Implicit treats are still going to
> be applied when an HQL query dereferences a path belonging to an inheritance subtype.

So only the string-based Criteria `Path#get` / `From#join` are affected; HQL is not.

Production impact: filtering OID entries by code or alt-code is broken at runtime, not just
in tests.

**Fix, implemented:** a general guard in `FilterPredicatesBuilder` that downcasts the root
with `CriteriaBuilder.treat` when the filter field's declaring type is a strict subtype of
the root type. It is derived from the attribute's own declaring type, so any future subclass
filter field is covered without further change; joins and non-subclass fields are untouched.
Measured effect: 125 failing -> 98.

**Carved out to #2149 and moved to preparation.** `treat()` is not a Hibernate 7 API - the
[...]

PR #2160, opened 2026-08-27
Resolve subclass-declared OID filter fields through the metamodel attribute

### LINKED ISSUE #2149: Resolve subclass-declared OID filter fields through the metamodel attribute
### Description

**Two OID filter fields stop resolving under Hibernate 7, and the obvious fix — an explicit
Criteria `treat` — silently corrupts the result count, so the field is resolved through its
metamodel attribute instead.**

`FilterField.OID_ENTRY_CODE` and `OID_ENTRY_ALT_CODES` name attributes declared on the
subclass `RdnAttributeTypeCustomOidEntry`. `CustomOidEntryRepository` roots its criteria at
the base `CustomOidEntry`. `FilterPredicatesBuilder` resolves the filter field **by name**
against that root.

Hibernate 6 performs an implicit downcast. **This works today on `main`** — filtering OID
entries by code and alt-code is not broken in production. Hibernate 7 does not perform the
downcast, and fails with:

```
PathElementException: Could not resolve attribute 'altCodes'
  of com.otilm.core.dao.entity.oid.CustomOidEntry
```

This is **intentional**, documented in the Hibernate 7.0 migration guide under *Criteria API
and Implicit Treats* (https://docs.hibernate.org/orm/7.0/migration-guide/):

> Since Hibernate 7.0, aligning with the JPA specification, the Criteria API will no longer
> allow retrieving subtype attributes this way, and it's going to require an explicit
> `jakarta.persistence.criteria.CriteriaBuilder#treat` to be called on the path first to
> downcast it to the subtype which defines the attribute. Implicit treats are still going to
> be applied when an HQL query dereferences a path belonging to an inheritance subtype.

Only the **string-based** Criteria `Path#get` and `From#join` are affected. HQL is not. The
typed `Path#get(SingularAttribute)` overload is not either — and that is what this fix uses.

### Why an explicit `treat` is the wrong fix

`CriteriaBuilder.treat` does make the attribute resolve. It also **restricts the query to
the subtype**, and Hibernate applies that restriction to the count query while dropping it
from the matching list query.

Measured on mixed-category fixtures — 2 non-RDN entries, 3 RDN entries — on Hibernate 6.6,
before and after adding the treat:

| Filter | Returned page | `totalItems` |
|---|---|---|
| `OID_ENTRY_CODE` NOT_EQUALS `CN` | unchanged, 4 rows | **4 → 2** |
| `OID_ENTRY_CODE` NOT_CONTAINS `C` | unchanged, 4 rows | **4 → 2** |
| `OID_ENTRY_CODE` EMPTY | unchanged, 3 rows | **3 → 1** |

The page and the count disagree **inside a single response**. The generated SQL shows why:

```sql
-- list: treat dropped, plain column reference
select distinct ... from core.custom_oid_entry coe1_0
 where coe1_0.code is null

-- count: treat materialised as a subtype-restricted derived table
select count(distinct coe1_0.oid)
  from (select * from core.custom_oid_entry t where t.category='RDN_ATTRIBUTE_TYPE') coe1_0
 where coe1_0.code is null
```

The same experiment on **Hibernate 7.4 produces byte-identical output**. This is not version
drift. A `treat` is wrong on both lines.

### Approach

Resolve the field through the **metamodel attribute** rather than through its name.
`FilterField` already holds the `Attribute` object; the builder was discarding it and passing
`getName()`.

```java
private static Path<?> resolveFieldPath(From from, Attribute<?, ?> fieldAttribute) {
    if (isDeclaredOnStrictSubtypeOf(from, fieldAttribute)
            && fieldAttribute instanceof SingularAttribute<?, ?> singularAttribute) {
        return from.get((SingularAttribute) singularAttribute);
    }
    return from.get(fieldAttribute.getName());
}
```

Hibernate honours the attribute object directly instead of looking its name up on the path's
own type. **No downcast is needed, and no subtype restriction is added.**

The subtype test is derived from the attribute's own declaring type rather than from a list
of known fields, so any future subclass filter field is covered without further change. Every
filter-field resolution site in the builder routes through the helper: the main path, the four
JSON-path branches, the native-array `NOT EXISTS` subquery, and the group-join subquery.

**One caveat, stated plainly.** `Path.get(SingularAttribute<? super X, Y>)` strictly expects an
attribute declared on `X` or a supertype. Passing a subtype attribute is outside the JPA
contract; it works because of how Hibernate resolves it. That trades one Hibernate-specific
behaviour for another. This one is verified on 6.6 **and** 7.4 and is exactly
behaviour-preserving. `treat` is spec-correct but changes results.

A subtype-declared attribute that is **not singular** has no such route. Rather than fall back
to the name lookup that Hibernate 7 refuses, `resolveFieldPath` rejects it with a
`ValidationException`. No filter field is plural today - `code` and `altCodes` are both
generated as `SingularAttribute` - so the branch is unreachable, but the next person to add a
subclass filter field gets a clear error instead of a silent regression.

### The new tests are a forward guard, not current coverage

`main` builds on Hibernate 6.6, where the name-based lookup still resolves. That means
`SubclassFilterFieldSearchITest` passes with or without this change **today**; it only starts
failing without it once the Hibernate 7 bump lands.

What it does guard on the current line is the page-versus-total agreement. Those two tests go
red the moment anyone reintroduces the `treat`, on either Hibernate version.

### Evidence

| Check | Result |
|---|---|
| 6.6 before the fix vs 6.6 after it, 20 recorded filter cases | byte-identical — no-op confirmed |
| 7.4 after the fix vs the 6.6 baseline | byte-identical — resolves correctly, same semantics |
| `CustomOidEntryServiceITest` on 7.4 | 22/22 green, was red |
| `NativeArrayFilterSearchITest` on 7.4 | 42/42 green |
| New `SubclassFilterFieldSearchITest` under the `treat` variant | 2 of 6 fail on count agreement |
| Whole-suite reduction on the 4.1 baseline branch | 125 failing → 98, measured with the earlier `treat` variant |

The 8 `FilterPredicatesBuilderITest` failures on the 4.1 baseline are unrelated literal
quoting (`'test'` vs `test`) and reproduce identically without this change. They belong
to #2017.

### Definition of Done

- [ ] Subclass-declared filter fields resolve without a downcast, derived from the attribute's
      declaring type rather than hard-coded per field.
- [ ] Filtering OID entries by both `code` and `altCodes` is proven working at the API level,
      not only in the builder unit test.
- [ ] A regression test asserts that `totalItems` agrees with the returned page for a negated
      subclass filter.
- [ ] Non-subclass filter fields and joined associations are demonstrably unaffected.
- [ ] The suite is green on the current Spring Boot 3.5 / Hibernate 6.6 line - this must land
      as a no-op there.
- [ ] `mvn -B verify` passes; Spotless and Checkstyle gates pass.

### Follow-up

`def99bbd` on `refactor/2016-spring-boot-4.1-baseline` implements the `treat` version and
carries the count divergence. It must be **superseded** by this change, not rebased onto it.

Preparation for the Spring Boot 4.1 migration, carved out of #2017. Lands on `main`, not on
the `feature/spring-boot-4.1` integration branch. Part of the preparation list on #1941.

ivosh:
**Two OID filter fields stop resolving under Hibernate 7, and the obvious fix — an explicit Criteria `treat` — silently corrupts the result count, so the field is resolved through its metamodel attribute instead.**

Closes #2149.

## What breaks, and when

`FilterField.OID_ENTRY_CODE` and `OID_ENTRY_ALT_CODES` name attributes declared on the subclass `RdnAttributeTypeCustomOidEntry`. `CustomOidEntryRepository` roots its criteria at the base `CustomOidEntry`. `FilterPredicatesBuilder` resolved those filter fields **by name** against that root.

**When it's fine — today, on `main`.** Hibernate 6 performs an implicit downcast. Filtering OID entries by code and alt-code is not broken in production.

**When it breaks — after the Hibernate 7 bump.** Hibernate 7 does not perform the downcast. It fails with:

```
PathElementException: Could not resolve attribute 'altCodes'
  of com.otilm.core.dao.entity.oid.CustomOidEntry
```

This is **intentional**, documented in the [Hibernate 7.0 migration guide](https://docs.hibernate.org/orm/7.0/migration-guide/) under *Criteria API and Implicit Treats*:

> Since Hibernate 7.0, aligning with the JPA specification, the Criteria API will no longer allow retrieving subtype attributes this way, and it's going to require an explicit `jakarta.persistence.criteria.CriteriaBuilder#treat` to be called on the path first to downcast it to the subtype which defines the attribute. Implicit treats are still going to be applied when an HQL query dereferences a path belonging to an inheritance subtype.

Only the **string-based** Criteria `Path#get` is affected. HQL is not. The typed `Path#get(SingularAttribute)` overload is not either — and that is what this PR uses.

## Why an explicit `treat` is the wrong fix

`CriteriaBuilder.treat` does make the attribute resolve. It also **restricts the query to the subtype**. Hibernate applies that restriction to the count query while dropping it from the matching list query.

Measured on mixed-category fixtures — 2 non-RDN entries, 3 RDN entries — on Hibernate 6.6, before and after adding the treat:

| Filter | Returned page | `totalItems` |
|---|---|---|
| `OID_ENTRY_CODE` NOT_EQUALS `CN` | unchanged, 4 rows | **4 → 2** |
| `OID_ENTRY_CODE` NOT_CONTAINS `C` | unchanged, 4 rows | **4 → 2** |
| `OID_ENTRY_CODE` EMPTY | unchanged, 3 rows | **3 → 1** |

The page and the count disagree **inside a single response**. The generated SQL shows why:

```sql
-- list: treat dropped, plain column reference
select distinct ... from core.custom_oid_entry coe1_0
 where coe1_0.code is null

-- count: treat materialised as a subtype-restricted derived table
select count(distinct coe1_0.oid)
  from (select * from core.custom_oid_entry t where t.category='RDN_ATTRIBUTE_TYPE') coe1_0
 where coe1_0.code is null
```

The same experiment on **Hibernate 7.4 produces byte-identical output**. This is not version drift. A `treat` is wrong on both lines.

## The approach

Resolve the field through the **metamodel attribute** rather than through its name. `FilterField` already holds the `Attribute` object; the builder was discarding it and passing `getName()`.

Hibernate honours the attribute object directly instead of looking its name up on the path's own type. **No downcast is needed, and no subtype restriction is added.**

The subtype test is derived from the attribute's own declaring type rather than from a list of known fields, so a future subclass filter field is covered without further change. Every filter-field resolution site in the builder routes through the helper: the main path, the five JSON-path branches (the array arm and the one- through four-segment arms), the native-array `NOT EXISTS` subquery, and the group-join subquery.

`SortOrderBuilder.resolveExpression` routes through it too, so ordering by a subclass-declared field resolves the same way filtering does. It previously built one dotted string path and resolved the last segment by name; under Hibernate 7 that segment raises `PathElementException`, which extends `IllegalArgumentException` and so is caught by `resolveExpression`'s own handler and reported as *"Field OID_ENTRY_CODE cannot be sorted on this resource"* — the sort would have disappeared rather than failed. Not reachable today: `SortSpecification.from` has no production caller yet.

A subtype-declared attribute that is **not singular** has no such route. Rather than fall back to the name lookup, `resolveFieldPath` rejects it with a `ValidationException`. No filter field is plural today — `code` and `altCodes` are both generated as `SingularAttribute` — so the branch is unreachable. The next person to add a subclass filter field gets a clear error instead of a silent regression.

**One caveat, stated plainly.** `Path.get(SingularAttribute<? super X, Y>)` strictly expects an attribute declared on `X` or a supertype. Passing a subtype attribute is outside the JPA contract; it works because of how Hibernate resolves it. That trades one Hibernate-specific behaviour for another. This one is verified on 6.6 **and** 7.4 and is exactly behaviour-preserving. `treat` is spec-correct but changes results.

## The integration tests are a forward guard; the unit test is current coverage

`main` builds on Hibernate 6.6, where the name-based lookup still resolves. `SubclassFilterFieldSearchITest` therefore passes with or without this change **today**. It only starts failing without it once the Hibernate 7 bump lands.

What it does guard on the current line is the page-versus-total agreement. Those two tests go red the moment anyone reintroduces the `treat`, on either Hibernate version.

`FilterPredicatesBuilderFieldPathTest` is the exception: it drives `resolveFieldPath` against a stubbed metamodel and does cover the resolution branches today, including the plural-subtype guard that no filter field can reach.

## Evidence

| Check | Result |
|---|---|
| 6.6 before the fix vs 6.6 after it, 20 recorded filter cases | byte-identical — no-op confirmed |
| 7.4 after the fix vs the 6.6 baseline | byte-identical — resolves correctly, same semantics |
| `CustomOidEntryServiceITest` on 7.4 | 22/22 green, was red |
| `NativeArrayFilterSearchITest` on 7.4 | 42/42 green |
| New `SubclassFilterFieldSearchITest` under the `treat` variant | 2 of 6 fail on count agreement |
| `SubclassFilterFieldSearchITest` + 3 filter suites on 6.6 | 127/127 green |
| Spotless, Checkstyle | pass |
| SonarCloud quality gate | passed — 87.5% coverage on new code |

The 8 `FilterPredicatesBuilderITest` failures on the 4.1 baseline branch are unrelated literal quoting (`'test'` vs `test`) and reproduce identically without this change. They belong to #2017.

### DIFF
--- a/src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
+++ b/src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
@@ -309,7 +309,7 @@ public class FilterPredicatesBuilder {
      * @throws ValidationException if {@code fieldAttribute} is declared on a subtype but is not singular
      */
     @SuppressWarnings({"unchecked", "rawtypes"})
-    static Path<?> resolveFieldPath(final From from, final Attribute<?, ?> fieldAttribute) {
+    static <Y> Path<Y> resolveFieldPath(final From from, final Attribute<?, ?> fieldAttribute) {
         if (!isDeclaredOnStrictSubtypeOf(from, fieldAttribute)) {
             return from.get(fieldAttribute.getName());
         }
@@ -318,7 +318,7 @@ public class FilterPredicatesBuilder {
                     + fieldAttribute.getDeclaringType().getJavaType().getSimpleName()
                     + " and is not singular, which filter resolution does not support");
         }
-        return from.get((SingularAttribute) singularAttribute);
+        return from.get(singularAttribute);
     }
 
     private static boolean isDeclaredOnStrictSubtypeOf(final From<?, ?> from, final Attribute<?, ?> fieldAttribute) {
--- a/src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
+++ b/src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
@@ -37,6 +37,7 @@ import jakarta.persistence.criteria.Root;
 import jakarta.persistence.criteria.Subquery;
 import jakarta.persistence.metamodel.Attribute;
 import jakarta.persistence.metamodel.PluralAttribute;
+import jakarta.persistence.metamodel.SingularAttribute;
 import java.io.Serializable;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
@@ -294,6 +295,37 @@ public class FilterPredicatesBuilder {
         return preparedValue;
     }
 
+    /**
+     * Resolves {@code fieldAttribute} as a path on {@code from}.
+     *
+     * <p>
+     * An attribute declared on an inheritance subtype of what {@code from} resolves to is addressed through the
+     * metamodel attribute itself rather than by name. The resulting path carries no restriction to that subtype, so the
+     * query still spans the whole hierarchy. Every other attribute is addressed by name.
+     *
+     * @param from the root or join the filter field is resolved against
+     * @param fieldAttribute the attribute being filtered on
+     * @return the path to the attribute
+     * @throws ValidationException if {@code fieldAttribute} is declared on a subtype but is not singular
+     */
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    private static Path<?> resolveFieldPath(final From from, final Attribute<?, ?> fieldAttribute) {
+        if (!isDeclaredOnStrictSubtypeOf(from, fieldAttribute)) {
+            return from.get(fieldAttribute.getName());
+        }
+        if (!(fieldAttribute instanceof SingularAttribute<?, ?> singularAttribute)) {
+            throw new ValidationException("Filter field " + fieldAttribute.getName() + " is declared on subtype "
+                    + fieldAttribute.getDeclaringType().getJavaType().getSimpleName()
+                    + " and is not singular, which filter resolution does not support");
+        }
+        return from.get((SingularAttribute) singularAttribute);
+    }
+
+    private static boolean isDeclaredOnStrictSubtypeOf(final From<?, ?> from, final Attribute<?, ?> fieldAttribute) {
+        final Class<?> declaringType = fieldAttribute.getDeclaringType().getJavaType();
+        return !declaringType.equals(from.getJavaType()) && from.getJavaType().isAssignableFrom(declaringType);
+    }
+
     private static <T> Predicate getPropertyFilterPredicate(final CriteriaBuilder criteriaBuilder,
             final CommonAbstractCriteria query, final Root<T> root, SearchFilterRequestDto filterDto,
             Map<String, From> joinedAssociations) {
@@ -310,7 +342,7 @@ public class FilterPredicatesBuilder {
 
         Expression expression = null;
         if (filterField.getFieldAttribute() != null && !isCountOperator(filterDto.getCondition())) {
-            expression = from.get(filterField.getFieldAttribute().getName());
+            expression = resolveFieldPath(from, filterField.getFieldAttribute());
         }
 
         boolean isJsonArray = false;
@@ -319,28 +351,28 @@ public class FilterPredicatesBuilder {
             if (isJsonArray) {
                 expression = criteriaBuilder
                         .function("jsonb_path_query_array", String.class,
-                                from.get(filterField.getFieldAttribute().getName()),
+                                resolveFieldPath(from, filterField.getFieldAttribute()),
                                 criteriaBuilder.literal(getArrayJsonPath(filterField.getJsonPath())));
             } else {
                 expression = switch (filterField.getJsonPath().length) {
                     case 1 -> criteriaBuilder
                             .function(JSONB_EXTRACT_PATH_TEXT_FUNCTION_NAME, String.class,
-                                    from.get(filterField.getFieldAttribute().getName()),
+                                    resolveFieldPath(from, filterField.getFieldAttribute()),
                                     criteriaBuilder.literal(filterField.getJsonPath()[0]));
                     case 2 -> criteriaBuilder
                             .function(JSONB_EXTRACT_PATH_TEXT_FUNCTION_NAME, String.class,
-                                    from.get(filterField.getFieldAttribute().getName()),
+                                    resolveFieldPath(from, filterField.getFieldAttribute()),
                                     criteriaBuilder.literal(filterField.getJsonPath()[0]),
                                     criteriaBuilder.literal(filterField.getJsonPath()[1]));
                     case 3 -> criteriaBuilder
                             .function(JSONB_EXTRACT_PATH_TEXT_FUNCTION_NAME, String.class,
-                                    from.get(filterField.getFieldAttribute().getName()),
+                                    resolveFieldPath(from, filterField.getFieldAttribute()),
                                     criteriaBuilder.literal(filterField.getJsonPath()[0]),
                                     criteriaBuilder.literal(filterField.getJsonPath()[1]),
                                     criteriaBuilder.literal(filterField.getJsonPath()[2]));
                     case 4 -> criteriaBuilder
                             .function(JSONB_EXTRACT_PATH_TEXT_FUNCTION_NAME, String.class,
-                                    from.get(filterField.getFieldAttribute().getName()),
+                                    resolveFieldPath(from, filterField.getFieldAttribute()),
                                     criteriaBuilder.literal(filterField.getJsonPath()[0]),
                                     criteriaBuilder.literal(filterField.getJsonPath()[1]),
                                     criteriaBuilder.literal(filterField.getJsonPath()[2]),
@@ -660,7 +692,7 @@ public class FilterPredicatesBuilder {
                     .where(criteriaBuilder
                             .isTrue(criteriaBuilder
                                     .function(functionName, Boolean.class, criteriaBuilder.literal(value.toString()),
-                                            subFrom.get(filterField.getFieldAttribute().getName()))));
+                                            resolveFieldPath(subFrom, filterField.getFieldAttribute()))));
             return criteriaBuilder.not(criteriaBuilder.exists(subquery));
         }).toArray(Predicate[]::new);
         return notExistsPredicates.length == 1 ? notExistsPredicates[0] : criteriaBuilder.and(notExistsPredicates);
@@ -851,7 +883,7 @@ public class FilterPredicatesBuilder {
                                         resource == Resource.CRYPTOGRAPHIC_KEY
                                                 ? originalRoot.get(CryptographicKeyItem_.keyUuid)
                                                 : originalRoot.get(UniquelyIdentified_.uuid)),
-                        joinGroup.get(fieldAttribute.getName()).in(filterValues));
+                        resolveFieldPath(joinGroup, fieldAttribute).in(filterValues));
 
         return criteriaBuilder.not(criteriaBuilder.exists(subquery));
     }
--- a/src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
+++ b/src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
@@ -309,7 +309,7 @@ public class FilterPredicatesBuilder {
      * @throws ValidationException if {@code fieldAttribute} is declared on a subtype but is not singular
      */
     @SuppressWarnings({"unchecked", "rawtypes"})
-    private static Path<?> resolveFieldPath(final From from, final Attribute<?, ?> fieldAttribute) {
+    static Path<?> resolveFieldPath(final From from, final Attribute<?, ?> fieldAttribute) {
         if (!isDeclaredOnStrictSubtypeOf(from, fieldAttribute)) {
             return from.get(fieldAttribute.getName());
         }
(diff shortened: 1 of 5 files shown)


src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -294,6 +295,37 @@ private static Object prepareDurationValue(AttributeContentType contentType, Str
...
+     * metamodel attribute itself rather than by name. The resulting path carries no restriction to that subtype, so the
+     * query still spans the whole hierarchy. Every other attribute is addressed by name.
+     *
+     * @param from the root or join the filter field is resolved against
+     * @param fieldAttribute the attribute being filtered on
+     * @return the path to the attribute
+     * @throws ValidationException if {@code fieldAttribute} is declared on a subtype but is not singular
+     */
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    static <Y> Path<Y> resolveFieldPath(final From from, final Attribute<?, ?> fieldAttribute) {
```
Jakub-Bartal-01, line 312: **[Medium]** | Correctness

The helper covers every predicate site, but not the ordering one. prepareExpression (same file, line 931) walks the path with a string get, and SortOrderBuilder.resolveExpression feeds it buildPathToProperty(field.getJoinAttributes(), field.getFieldAttribute()) - so ordering a CustomOidEntry query by OID_ENTRY_CODE or OID_ENTRY_ALT_CODES takes exactly the implicit-downcast route this change replaces. Under Hibernate 7 that get("code") throws PathElementException, which extends IllegalArgumentException, so SortOrderBuilder's catch converts it into "Field OID_ENTRY_CODE cannot be sorted on this resource" - sorting quietly disappears instead of failing loudly. Not reachable today: nothing calls SortSpecification.from, so only the filter path is live.

_Suggested:_ Resolve the last path segment in SortOrderBuilder.resolveExpression through resolveFieldPath, or give prepareExpression an overload that takes the Attribute for its final token, so ordering resolves a subtype attribute the same way filtering now does. If it stays out of scope here, record it on #2149 so it lands before list-view sorting reaches the OID resource.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -294,6 +295,37 @@ private static Object prepareDurationValue(AttributeContentType contentType, Str
         return preparedValue;
     }
 
+    /**
+     * Resolves {@code fieldAttribute} as a path on {@code from}.
+     *
+     * <p>
+     * An attribute declared on an inheritance subtype of what {@code from} resolves to is addressed through the
+     * metamodel attribute itself rather than by name. The resulting path carries no restriction to that subtype, so the
```
Jakub-Bartal-01, line 303: **[Low]** | Maintainability

The javadoc states that the resolved path carries no subtype restriction, but not why the obvious alternative is wrong: CriteriaBuilder.treat also resolves the attribute, and Hibernate keeps its restriction in the count query while dropping it from the list query, so totalItems stops matching the returned page. from.get(singularAttribute) passes an attribute outside the SingularAttribute<? super X, Y> bound the JPA signature declares, which reads like an oversight to the next maintainer, and the reason it is deliberate lives only in the pull request body.

_Suggested:_ Add one sentence to the javadoc: an explicit treat resolves the attribute but restricts the count query while the list query drops the restriction, so the total disagrees with the page. Point at SubclassFilterFieldSearchITest as the guard that catches a reintroduction.

Jakub-Bartal-01 (commented):

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
ivosh: Not adding javadoc. This very rationale is already present in the PR body - refer to section "Why an explicit treat is the wrong fix".

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
ivosh: Fixed in 77e5cbc1.

Premise confirmed. `OID_ENTRY_CODE` and `OID_ENTRY_ALT_CODES` are `RdnAttributeTypeCustomOidEntry_` attributes resolved against a `CustomOidEntry` root, and `PathElementException` extends `IllegalArgumentException` — checked with `javap` against the 6.6.53 jar on the classpath, where the `@apiNote` says it does so "to satisfy a questionable requirement of the JPA criteria query API". So after the bump `resolveExpression`'s own catch would have reported it as "Field OID_ENTRY_CODE cannot be sorted on this resource" rather than failing.

`SortOrderBuilder.resolveExpression` now walks the join attributes by name and resolves the field attribute through `FilterPredicatesBuilder.resolveFieldPath`, so ordering and filtering resolve a subtype attribute the same way. Your first option rather than the `prepareExpression` overload: `resolveExpression` already holds the `FilterField`, so the `Attribute` is in hand, and `prepareExpression` keeps its single string-path contract. Not deferred to #2149.

Two things the rewrite depends on, checked rather than assumed. All 181 `FilterField` constants carry a non-null `fieldAttribute`, so moving the leaf out of `buildPathToProperty` changes nothing reachable; across the 29 fields that declare joins, the old split point was always the joins/leaf boundary, so the joins and the leaf `get` are identical either way. And `ValidationException` extends `RuntimeException`, not `IllegalArgumentException`, so `resolveFieldPath`'s plural-subtype guard message propagates instead of being flattened into the unsortable-field text.

On the guard, stated the way the PR body states it for the filter tests: `SubclassFilterFieldSearchITest.ordersBySubclassDeclaredCodeAcrossEveryCategory` is a forward guard, not a regression test. Reverting `SortOrderBuilder` and rerunning it leaves it green and the emitted SQL byte-identical — 6.6.53 resolves the name through `EntityTypeImpl.findSubtypeAttribute`, so nothing on this line can distinguish the two routes. What it does assert today is that a `SortSpecification` is honoured on the OID resource at all: descending by `OID_ENTRY_CODE` over the mixed-category fixtures, all five entries returned, the two carrying a code in code order. Descending because ascending would coincide with both the fixture insertion order and the oid order; with the sort argument dropped it fails at index 0.

Two things your comment's closing line points at that this change does not cover, both pre-existing. `CustomOidEntry` is keyed by `oid` and has no `uuid`, so `tieBreak` returns empty and a paged sort by `OID_ENTRY_CODE` leaves the equal-valued rows in undefined order across pages. And `ResourceToClass` has no `OID` entry, so `resourceOf` returns null and the "field belongs to this resource" guard is skipped for OID sorts. Both want recording on #2149 before list-view sorting reaches the resource.

Jakub-Bartal-01 (approved):

Issue #2033, opened 2026-08-12
Sorting on attribute-sourced columns


Jakub-Bartal-01:
### Description

Extend request-driven sorting to custom, meta and data columns. Attribute filtering today is an order-agnostic `EXISTS` subquery in `FilterPredicatesBuilder`; ordering needs the value itself as a sort key.

Scope:
- Build the sort key as a correlated scalar subquery over `attribute_content_2_object` joined to `attribute_content_item`, extracting with `jsonb_extract_path_text` and applying the same per-content-type cast used by the filter path (INTEGER, FLOAT, DATE, TIME, DATETIME cast; others compared as text).
- Resolve multi-valued ambiguity deterministically by taking the lowest `item_order`.
- Assess query cost on the certificate and key tables at production-like volume; add an expression index or a materialized projection if the plan requires it, delivered as a Flyway migration.
- Flip `sortable` to true for attribute fields once supported.

### Acceptance Criteria

1. [x] Sorting by a custom, meta or data column orders the whole result set.
2. [x] Rows with no value for the sorted attribute sort consistently (documented as first or last) and are not dropped.
3. [x] Multi-valued attributes sort on the lowest `item_order` value.
4. [ ] Query plan reviewed at production-like volume; any required index ships as a Flyway migration.
5. [x] Sonar quality gate passes and new code has at least 80% coverage.
6. [x] The published API prose no longer says attribute fields cannot be ordered on. `ConfigurableColumnsDocs.CATALOGUE_FLAGS` in `OmniTrustILM/interfaces` states that attribute-sourced fields report `sortable` as `false` because ordering by attribute value is not supported yet; that sentence ships amended in the same release as this change, or the document contradicts the behaviour.

Jakub-Bartal-01:
### Why this has no open PR

The code shipped, but not through a PR that targeted `main`. #2179 carried this change and was
opened against `feat/2031-request-driven-sorting` rather than `main`, so that it would show only the
attribute-sorting diff on top of #2178. It merged into that branch on 2026-09-01, and #2178 then
merged to `main` on 2026-09-02 as `c1657ebf2`, carrying both changes. Because this issue's own PR
never targeted the default branch, no closing keyword fired and nothing moved the Status off Review.

### What is done

Verified against `c1657ebf2` on `main`:

- **AC1** — `FilterPredicatesBuilder.getAttributeSortKey` builds the correlated scalar subquery over
[...]

Issue #2031, opened 2026-08-12
Server-side sorting on the configurable-column list endpoints


Jakub-Bartal-01:
### Description

`SearchRequestDto` now carries an optional sort, and the shared `SecurityFilterRepository` can build a request-driven ORDER BY. This ticket applies it to the list endpoints that ship configurable columns: certificates, cryptographic keys, discoveries, connectors (`POST /v2/connectors/list`), secrets, CBOMs and signing records.

Scope:
- Translate `sort` into an `Order` for each of the seven services, resolving the field through the `FilterField` metamodel paths (`joinAttributes`, `fieldAttribute`) that filtering already uses.
- Carry the ordering through the two-phase query. The listing selects a page of UUIDs and then fetches by `uuid IN (...)`, so an ORDER BY applied only to the first phase is silently discarded by the second.
- Remove ordering that is baked into derived method names where it now conflicts, starting with `CryptographicKeyItemRepository.findFullByUuidInOrderByCreatedAtDesc`.
- Reject a sort naming a field the catalogue reports as `sortable=false` with a clear validation error rather than ignoring it, so the frontend cannot silently show an unsorted table.
- Leave each endpoint's existing default ordering in place for requests that omit `sort`.

Attribute-sourced sorting is out of scope here and lands in the attribute-sorting ticket.

### Acceptance Criteria

1. [ ] Each of the seven endpoints orders by a requested property field, ascending and descending.
2. [ ] Ordering spans the whole result set: page 2 continues where page 1 ended, verified across a page boundary on a data set larger than one page.
3. [ ] A request omitting `sort` returns rows in exactly the order it did before this change, on every one of the seven endpoints.
4. [ ] A sort naming a non-sortable or unknown field is rejected with an actionable message.
5. [ ] Sorting composes with filters and with row-level authorization - a sorted listing never returns an object the caller could not otherwise see.
6. [ ] Sonar quality gate passes and new code has at least 80% coverage.

Issue #2020, opened 2026-08-11
Align third-party starters: OpenTelemetry and springdoc


ivosh:
### Description

Align the third-party starters that track the Spring Boot major rather than their own
schedule: `opentelemetry-spring-boot-starter` and springdoc.

These fail differently from the rest of the migration. A starter built against the
previous Spring Boot major usually still compiles - it breaks at context refresh or at
first use, as a missing bean, a failed auto-configuration, or silently absent telemetry.
Verifying that the build passes proves nothing here; the exporters and the generated
spec have to actually work.

Two facts from the pilot spike (findings on OmniTrustILM/ilm#322):

- **springdoc is UNVERIFIED by the spike** - treat it as this issue's open risk, not as
  covered.
- The OpenTelemetry incubator (`opentelemetry-sdk-extension-incubator`) pins
  `jackson-dataformat-yaml` 2.x at runtime scope, not optional - the sole non-migratable
  third-party Jackson 2 dependency. Jackson 2 therefore stays on the classpath
  regardless of #2104; record the resolved version here and coordinate with #2104's
  wire inventory.

Sub-PR on the `feature/spring-boot-4.1` integration branch. Blocked by #2016. Part of
the migration list on #1941.

### Definition of Done

- [ ] `opentelemetry-spring-boot-starter` is on a version supporting Spring Boot 4.1,
      and traces, metrics and logs still reach an OTLP collector when the exporters are
      enabled.
- [ ] The `otel.*` configuration keys in `application.yml` still bind; any rename is
      recorded for the release notes and for #2021.
- [ ] The incubator's Jackson 2 runtime pin is recorded, with the resolved version, in
      this issue and in #2104's wire inventory.
- [ ] springdoc is on its Spring Boot 4-compatible major and the generated OpenAPI
      specification is produced without error.
- [ ] The generated specification is diffed against the 3.5 output; every difference is
      eliminated or justified in this issue.
- [ ] Context refresh produces no auto-configuration warnings from either starter.
- [ ] `mvn -B verify` passes; Spotless / Checkstyle gates pass.

PR #2162, opened 2026-08-27
Project attribute values and publish column capability flags

### LINKED ISSUE #2032: Project attribute values into list responses and flag displayable fields
### Description

Attribute-sourced columns cannot render today: the listing DTOs carry no attribute or metadata values, and no bulk endpoint exists to fetch them per object. This ticket populates them on request and tells the frontend which fields are legitimate columns.

The contract exposes this through `AttributeProjectable`, implemented by all seven listing entries - `CertificateDto`, `KeyItemDto`, `DiscoveryListDto`, `core.connector.v2.ConnectorDto`, `SecretDto`, `CbomDto` and `SigningRecordListDto` - so the projection populates one type instead of branching per resource.

Scope:
- When a request carries `columns` naming custom, meta or data fields, resolve their values for the returned objects in a set-based query (not per row) and populate the projected-values map, keyed by field source and then by field identifier.
- Handle absent values and multi-valued attributes, ordering values by `attribute_content_2_object.item_order`.
- Emit `displayable` and `sortable` per field from `getSearchableFieldInformation` and `AttributeEngine.getResourceSearchableFields`, for all seven resources.
- Mark Secret-content and encrypted-protection attributes `displayable=false` so they never reach the column picker.
- Mark attribute-sourced fields `sortable=false` until attribute sorting lands.
- `Resource.CBOM` and `Resource.SIGNING_RECORD` report `hasCustomAttributes=false`, so their catalogues legitimately expose no custom-attribute fields. The catalogue must reflect that rather than inventing empty groups.

### Acceptance Criteria

1. [ ] Requesting a custom, meta or data field in `columns` returns its value for each row in one set-based query, on every resource that has such fields.
2. [ ] Objects without a value for a requested attribute omit that entry rather than being dropped from the page.
3. [ ] Multi-valued attributes return values in `item_order` sequence.
4. [ ] The catalogue reports `displayable` and `sortable` per field for all seven resources; Secret and encrypted attributes are not displayable.
5. [ ] A request without `columns` returns the response unchanged, byte for byte, on all seven endpoints.
6. [ ] Sonar quality gate passes and new code has at least 80% coverage.

Jakub-Bartal-01:
Closes #2032

Attribute-sourced columns could not render: the listing DTOs carried no attribute or metadata values, and nothing told the frontend which fields are legitimate columns in the first place. This does both.

## Projection

When a listing request names custom, meta or data fields as `columns`, their values are loaded **for the whole page in one query** and set on the entries. A page of twenty-five objects would otherwise issue twenty-five round trips before it could be serialized.

Every listing that supports configurable columns implements `AttributeProjectable`, so `AttributeColumnProjector` fills any of them rather than each resource growing its own branch. The seven inventories differ only in which uuid their attribute content hangs off, and the call site supplies that.

- Values are keyed by field source and then by field identifier, because an identifier is unique only within its source.
- A multi-valued attribute keeps the sequence it was stored in — the query orders by object, definition and `aco.order`, so grouping is a single pass.
- An object with no value for a requested column carries no entry for it, and an object with no values at all is left untouched rather than given an empty map. **Absence omits the entry; it never drops the row.**
- The query narrows by attribute *name* as well as type, so a resource with thirty attributes does not load all of them to show two. Name is not unique on its own, so the exact `(type, name, content type)` triple is still matched before a value is kept.
- A column naming an attribute that no longer exists, or an identifier carrying no content type, is skipped rather than rejected — one stale entry in a saved view must not fail the listing that carries it.

**Key items are the one listing whose entries do not own their attributes.** They carry the key's, and several items can share a key, so the projector applies one set of values to every entry sharing a uuid rather than to the first it sees.

## Catalogue flags

Every published field now answers `displayable` and `sortable`.

- A **property** field is sortable when a single scalar path leads to it. A field with no attribute behind it has nothing to order by; a native array holds many values per row; and a JSON-path field's attribute is the whole JSONB column, so ordering by it would order by the serialized document rather than by the value the field names. These are exactly the cases `SortOrderBuilder` rejects or would mis-order.
- **No attribute-sourced field is sortable yet.** Ordering by one needs a correlated scalar subquery over JSONB that no index supports, and a multi-valued attribute leaves the sort key ambiguous (#2033). The field says so at the point a column is chosen rather than when a header refuses.
- **Secret content, encrypted content, code blocks and hidden attributes are never offered as columns.** A secret is never rendered; encrypted content is ciphertext only its own decryption path can read, and a listing does not take that path; a code block is multi-line by construction and a table cell holds one line; and a definition marked `visible=false` is one the contract says to hide from the user, which rules out a column of its own. The projection drops all four again on the way out, because the flag is a hint on a response a caller is free to ignore.

## Projected values carry only what a cell renders

Four content types are reduced rather than serialized whole, each rebuilt as a fresh value so nothing stored is touched.

- **CREDENTIAL and OBJECT are reduced to their reference**, which is all the cell shows for either. A credential's data is a nested attribute list that can carry secret-bearing content; an object's data is arbitrary serializable data of no declared shape.
- **RESOURCE is reduced to its resource, name and uuid** — what the cell needs to render a link labelled by name. Its subtypes carry the referenced object's own content alongside that identity: a base64 certificate body, a `SecretContent`, or the resolved attributes of the referenced object. The reduction rebuilds the base `ResourceObjectContentData`, so it is a whitelist: a fourth subtype added later is reduced without an edit.
- **FILE keeps its name and media type and drops the base64 body**, which no cell shows and which a page of rows would otherwise serialize once per file.

Content stored under the version 2 contract arrives as a generic map rather than as its content class — the persisted json carries no `contentType` for the polymorphic reader to key on — so each reduction reads the fields it keeps by name when the typed class is absent. A shape neither branch recognises yields no data at all, so an unrecognised value degrades to a reference-only cell rather than passing through.

The code-block exclusion is the one addition the UX design asked for beyond the ticket text (fe-administrator#2022 handover): `getAttributeContent` returns a `<CodeBlock>` element for that content type, which would break the one-line row rule the moment someone picked such a field.

## Compatibility

A request that names no attribute-sourced column does no work here at all, and `attributeValues` is `@JsonInclude(NON_NULL)` on all seven DTOs, so a response without `columns` is byte for byte what it was before. Asserted directly for the no-columns, empty-columns and property-only-columns cases.

CBOMs and signing records report `hasCustomAttributes=false`; their catalogues already omit the group rather than publishing an empty one, and there is now a test holding that.

## One pre-existing bug found, deliberately not fixed

`DiscoveryServiceImpl.getSearchableFieldInformationByGroup` publishes `FilterField.CKI_NAME` — a *cryptographic key* field — as the discovery name field. Filtering works by accident, because the path `name` resolves on `Discovery` too, but sorting by it would be rejected: `SortOrderBuilder` checks that the field's root resource matches the queried one. So discovery's name column will advertise `sortable=true` and then fail validation.

Fixing it means changing the field identifier discoveries publish for their name filter, which is a contract change for existing API clients. That belongs in its own issue with its own decision, not slipped into this one. Flagging it here so it is not lost.

## Verification

- `mvn checkstyle:check` clean; `spotless:apply` applied.
- New: `AttributeColumnProjectionITest` (14), `ColumnCatalogueFlagsITest` (14), `KeyAttributeColumnProjectionITest` (3), `SearchHelperColumnFlagsTest` (33), `ProjectedContentShapeTest` (20).
- Regression: the `integration.search`, `integration.attribute` and filter suites (539) plus the certificate / discovery / secret / CBOM / signing-record / key / connector suites (**1532 tests, 0 failures**).


src/main/java/com/otilm/core/dao/repository/AttributeContent2ObjectRepository.java
```diff
@@ -84,6 +85,35 @@ List<ObjectAttributeContent> getObjectCustomAttributesContent(@Param("attributeT
...
+     * a view puts on screen. Name is not unique on its own - the same name may exist under two content types - so the
+     * caller still matches the exact field identifier.
+     */
+    @Query("""
+            SELECT new com.otilm.core.attribute.engine.records.ProjectedAttributeContent(
+                aco.objectUuid, ad.type, ad.name, ad.contentType, aci.json, aci.encryptedData)
+                FROM AttributeContent2Object aco
+                JOIN AttributeContentItem aci ON aci.uuid = aco.attributeContentItemUuid
+                JOIN AttributeDefinition ad ON ad.uuid = aci.attributeDefinitionUuid
+                WHERE ad.enabled = true AND ad.type IN (:attributeTypes) AND ad.name IN (:attributeNames)
```
lubomirw: **[High]** | Security

The bulk projection query omits the allowed/forbidden attribute-definition UUID filtering that `AttributeEngine` already applies on its own read paths. A caller holding only resource LIST access can name a CUSTOM attribute as a column and read back restricted plaintext values it is not entitled to see.

_Suggested:_ Thread the caller's attribute security filter (allowed/forbidden definition UUIDs) into the projection call and into this query, and add a test asserting a denied definition yields no values.

<sub>flagged by Copilot CLI, superpowers code review</sub>

lubomirw: **[High]** | Correctness

`ad.enabled = true` silently excludes DATA and META attributes: their definition-creation paths leave the nullable `enabled` column unset, so the predicate never matches and those two sources always project empty.

_Suggested:_ Apply the `enabled` predicate to CUSTOM only (or treat a null `enabled` as enabled for DATA/META), and extend coverage to all three sources rather than CUSTOM alone.

<sub>flagged by guided-pr-review, superpowers code review</sub>

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -317,6 +324,12 @@ public CryptographicKeyResponseDto listCryptographicKeys(SecurityFilter filter,
             return dto;
         }).toList();
 
+        // Attribute content of a key hangs off the key, not off its items, so several listed items may share one
+        // set of values.
+        attributeColumnProjector
+                .project(Resource.CRYPTOGRAPHIC_KEY, request.getColumns(), listedKeyDtos,
```
lubomirw, line 330: **[High]** | Correctness

The UUID extractor is fixed to `getKeyWrapperUuid()`, which is right for CUSTOM and DATA content but wrong for META: key META attributes hang off each individual key-item UUID, not off the wrapper. META columns will therefore come back empty for every listed key item.

_Suggested:_ Resolve the object UUID per source — wrapper UUID for CUSTOM/DATA, key-item UUID for META — and merge both result maps per listed item.

<sub>flagged by Copilot CLI, guided-pr-review, superpowers code review</sub>

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
```diff
@@ -0,0 +1,209 @@
...
+            RequestedColumn column = matchRequested(requested, content);
+            if (column == null) {
+                continue;
+            }
+            byObject
+                    .computeIfAbsent(content.objectUuid(), uuid -> new EnumMap<>(FilterFieldSource.class))
+                    .computeIfAbsent(column.source(), source -> new LinkedHashMap<>())
+                    .computeIfAbsent(column.fieldIdentifier(), identifier -> new ArrayList<>())
+                    .add(AttributeVersionHelper
+                            .convertAttributeContentToV3(content.contentItem(), content.contentType()));
```
lubomirw: **[High]** | Security

CREDENTIAL content is serialized straight through `convertAttributeContentToV3`, bypassing the existing response redactor. `CredentialAttributeContentData.attributes` can carry secret-bearing nested content, and this projection would ship it into a list response — while the frontend only ever reads `reference`.

_Suggested:_ Emit a reference-only projection shape for CREDENTIAL (or route it through the existing redactor), and assert in a test that nested `attributes` are absent from the projected payload.

lubomirw: **[High]** | Performance

FILE content is projected whole, including the base64 `FileAttributeContentData.content` body, even though a list cell only renders filename and MIME type. A page of rows carrying file attributes serializes full file bodies into the listing response.

_Suggested:_ Strip `content` from the projected FILE shape, keeping filename and MIME type, and add a payload-shape test pinning that.

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -48,6 +48,8 @@ public static SearchFieldDataDto prepareSearch(final FilterField filterField, Ob
             values = withoutNull;
         }
         fieldDataDto.setValue(values);
+        fieldDataDto.setDisplayable(true);
+        fieldDataDto.setSortable(isSortable(filterField));
```
lubomirw, line 86: **[High]** | Correctness

`sortable` is now advertised to clients, but no listing honours it: every affected service calls the no-sort repository overload, so a client sorting on a field the catalogue declared sortable gets unsorted results. Certificate and key listings additionally use a two-stage load and would need the UUID ranking restored after the second query.

_Suggested:_ Either wire `request.getSort()` end to end (including restoring rank order in the two-stage loads), or keep `sortable=false` until it is wired, so the flag never over-promises.

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -48,6 +48,8 @@ public static SearchFieldDataDto prepareSearch(final FilterField filterField, Ob
             values = withoutNull;
         }
         fieldDataDto.setValue(values);
+        fieldDataDto.setDisplayable(true);
```
lubomirw, line 85: **[Medium]** | Correctness

`displayable=true` is set unconditionally for every property filter, so all of them become column-picker options. Many certificate fields have neither a value in the list payload nor a renderer in the frontend registry, so picking them yields a permanently blank column.

_Suggested:_ Model displayability per listing field rather than defaulting to true — only fields the listing actually emits and the frontend can render should be offered.

src/test/java/com/otilm/core/integration/search/AttributeColumnProjectionITest.java
```diff
@@ -0,0 +1,238 @@
...
+    void anObjectWithoutAValueIsStillReturned() {
+        List<DiscoveryListDto> discoveries = list(List.of(customColumn(ENVIRONMENT)));
+
+        // The point of the assertion: absence omits the entry, it does not drop the row from the page.
+        Assertions.assertNull(entry(discoveries, withoutValues).getAttributeValues());
+    }
+
+    @Test
+    void aRequestedColumnTheObjectHasNoValueForIsAbsentRatherThanEmpty() {
+        Discovery partial = saveDiscovery("discovery-partial");
```
lubomirw, line 246: **[Low]** | Testing

The `partial` fixture owns neither requested attribute, so the assertion only re-tests the already-covered "no values at all" case. It cannot catch the regression it is named for — an empty sibling entry appearing beside a real value.

_Suggested:_ Give the fixture a value for one of the two requested attributes, then assert the present field is projected and only the missing one is absent.

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
```diff
@@ -0,0 +1,209 @@
...
+/**
+ * Populates the attribute-sourced columns a listing request asked for.
+ *
+ * <p>
+ * Every listing that supports configurable columns implements {@link AttributeProjectable}, so one component fills any
+ * of them rather than each resource growing its own branch. Values are loaded for the whole page in one query: a page
+ * of twenty-five objects would otherwise issue twenty-five round trips before it could be serialized.
+ *
+ * <p>
+ * A request that names no attribute-sourced column does nothing at all here, which is what keeps a response without
```
lubomirw: **[Low]** | Maintainability

"identical to what it was before this existed" dates the Javadoc to this PR — a reader a year from now has no "before" to compare against.

_Suggested:_ State the durable contract instead: a request that names no attribute-sourced column leaves the listing entries unchanged.

<sub>flagged by pr-hygiene</sub>

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -120,9 +122,40 @@ public static SearchFieldDataDto prepareSearchForJSON(final SearchFieldObject at
         fieldDataDto.setType(searchFieldTypeEnum.getFieldType());
         fieldDataDto.setValue(attributeSearchInfo.getContentItems());
         fieldDataDto.setAttributeContentType(attributeSearchInfo.getAttributeContentType());
+        fieldDataDto.setDisplayable(isDisplayable(attributeSearchInfo));
+        // Ordering by an attribute needs a correlated scalar subquery over JSONB that no index
+        // supports, and a multi-valued attribute leaves the sort key ambiguous. Until that lands,
```
lubomirw: **[Low]** | Maintainability

"Until that lands" narrates a roadmap that the code cannot keep in sync.

_Suggested:_ Keep the standing technical reason and drop the future tense — ordering by an attribute needs a correlated scalar subquery over JSONB that no index supports, and a multi-valued attribute leaves the sort key ambiguous.

<sub>flagged by pr-hygiene</sub>

src/test/java/com/otilm/core/integration/search/ColumnCatalogueFlagsITest.java
```diff
@@ -0,0 +1,147 @@
...
+        // Ordering across a join aggregates the joined values rather than refusing, so such a field is offered.
+        SearchFieldDataDto signingProfile = field(signingRecordService.getSearchableFieldInformation(),
+                FilterField.SIGNING_RECORD_SIGNING_PROFILE.name()).orElseThrow();
+
+        Assertions.assertEquals(true, signingProfile.getSortable());
+    }
+
+    @Test
+    void aNativeArrayPropertyFieldIsNotSortable() {
+        // It holds many values per row, so there is no single value to order the row by. No inventory in scope
```
lubomirw: **[Low]** | Maintainability

The comment carries a temporal claim ("today") plus speculation about future inventory growth — both go stale the moment an inventory publishes such a field, and neither helps a reader understand the assertion.

_Suggested:_ Keep only the stable reason: a multi-valued field has no single scalar key to order a row by, so the shared catalogue helper marks it non-sortable.

<sub>flagged by pr-hygiene</sub>

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
```diff
@@ -0,0 +1,209 @@
...
+ * {@code columns} identical to what it was before this existed.
+ */
+@Component
+@RequiredArgsConstructor
+public class AttributeColumnProjector {
+
+    /** Separates the attribute name from its content type in a field identifier, as the catalogue builds it. */
```
lubomirw: **[Nit]** | Maintainability

The Javadoc restates what `FIELD_IDENTIFIER_SEPARATOR` already says by name.

_Suggested:_ Drop the comment.

<sub>flagged by pr-hygiene</sub>

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
```diff
@@ -0,0 +1,209 @@
...
+     * Content that is never projected, whatever a request asks for. The catalogue already marks these fields
+     * undisplayable, so a well-behaved caller cannot reach them; this is the second lock, because the first one is a
+     * flag on a response the caller is free to ignore.
+     */
+    private static final Set<AttributeContentType> WITHHELD_CONTENT_TYPES = Set
+            .of(AttributeContentType.SECRET, AttributeContentType.CODEBLOCK);
+
+    private final AttributeContent2ObjectRepository attributeContent2ObjectRepository;
+
+    /** One requested attribute column, split into the parts the stored content is matched by. */
```
lubomirw: **[Nit]** | Maintainability

The Javadoc restates the record name and its components without adding anything.

_Suggested:_ Drop the comment.

<sub>flagged by pr-hygiene</sub>

src/test/java/com/otilm/core/integration/search/AttributeColumnProjectionITest.java
```diff
@@ -0,0 +1,238 @@
...
+import java.util.Set;
+import java.util.UUID;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+/**
+ * Projection of attribute-sourced columns into a listing, exercised end to end through the discovery listing - one of
+ * the seven inventories that carry configurable columns, and the cheapest of them to populate.
```
lubomirw: **[Nit]** | Documentation

The class Javadoc hard-codes an inventory count that will drift, and "the cheapest of them to populate" is a subjective claim a reader cannot verify.

_Suggested:_ State only that the discovery listing exercises attribute-column projection end to end.

<sub>flagged by pr-hygiene</sub>

src/test/java/com/otilm/core/integration/search/AttributeColumnProjectionITest.java
```diff
@@ -0,0 +1,238 @@
...
+    @Autowired
+    private DiscoveryRepository discoveryRepository;
+
+    @Autowired
+    private DiscoveryExternalService discoveryService;
+
+    @Autowired
+    private AttributeEngine attributeEngine;
+
+    /** Carries values for both attributes, one of them multi-valued. */
```
lubomirw: **[Nit]** | Maintainability

These two fixture comments (lines 51 and 54) repeat what `withValues` / `withoutValues` already communicate.

_Suggested:_ Drop both comments.

<sub>flagged by pr-hygiene</sub>

lubomirw (changes requested):
Automated multi-reviewer triage — 15 findings (REQUEST_CHANGES). Sources: Copilot CLI, guided-pr-review, superpowers code review, pr-hygiene.

src/main/java/com/otilm/core/dao/repository/AttributeContent2ObjectRepository.java
Jakub-Bartal-01: Fixed in 3ca286029. Confirmed against the per-object read directly above it: `getObjectCustomAttributesContent` applies the allowed/forbidden definition uuids and the projection query applied neither, so LIST access alone was enough to read a restricted custom attribute's plaintext by naming it as a column.

The derivation of those two lists was inline in `AttributeEngine.getObjectCustomAttributesContent`; it is now extracted into `CustomAttributeContentFilter` with a public `loadCustomAttributeContentFilter()`, and both the detail read and the projector go through it — so the listing applies the same rule rather than a second copy of it that can drift. Added an ITest that restricts `ATTRIBUTE`/`MEMBERS` via the existing `restrictObjectAccess` helper and asserts the projection comes back with no values.

Jakub-Bartal-01: Fixed in 3ca286029. Verified the mechanism rather than taking it on trust: `setEnabled(true)` is called in exactly one place (`AttributeEngine`, the new-custom-attribute branch), the DATA and META creation paths never call it, and the column is a nullable `Boolean` — so `ad.enabled = true` excluded both sources outright.

Taken the first option: the `enabled` predicate now applies to custom definitions only, guarded by the same `customType` parameter as the permission filter, rather than treating null as enabled — only custom definitions carry the flag, so that is where it means something. Coverage extended with an ITest that stores metadata on a discovery and asserts the metadata column projects; the security test above covers the custom side.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
Jakub-Bartal-01: Fixed in 3ca286029. Confirmed against the key detail read, which makes exactly this split: custom and data content are read against `key.getUuid()`, while metadata is read per item against `UUID.fromString(k.getUuid())`.

`project` now takes an optional second resolver for metadata, and the key listing passes the item uuid for it. Rather than branching on whether the two differ, the requested columns are grouped by the resolver they need — when the same reference is passed twice the two groups merge and the whole page is still one query, so no other listing gains a second round trip. Values are accumulated per entry in an `IdentityHashMap` and set once after every pass, so the second pass adds to the first instead of replacing it, and metadata belonging to one key item can no longer land on its siblings through a shared map.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -317,6 +324,12 @@ public CryptographicKeyResponseDto listCryptographicKeys(SecurityFilter filter,
...
+                        keyItem -> AttributeColumnProjector.parseUuid(keyItem.getKeyWrapperUuid()));
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed in 3ca286029 — see the reply on the triage thread for this line. The uuid is routed per source (wrapper for custom/data, key item for metadata) and the two result maps are merged per listed item.

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
Jakub-Bartal-01: Fixed in 3ca286029. Note on the wording: `OutboundSecretContainment` is a callback-response guard that refuses rather than strips, so there was no existing redactor on this path to route through — the fix is a projection shape instead.

`toProjectedValue` now emits reference-only content for CREDENTIAL, so the nested `attributes` list never reaches a listing response, which also matches what the frontend reads. Added `ProjectedContentShapeTest` asserting the projected value carries the reference and no data at all.

Jakub-Bartal-01: Fixed in 3ca286029. FILE is now projected with `fileName` and `mimeType` and no `content`, built as a new `FileAttributeContentData` rather than by mutating the deserialized value, so nothing stored is touched. The same test class pins the shape: filename and media type present, body null, and every other content type projected unchanged by identity.

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Fixed in 3ca286029, taking the second option. Confirmed the finding: no service reads `request.getSort()` — the only `getSort()` calls in `src/main` are in `ListViewServiceImpl`, which stores and validates a saved view's sort. PR #2157 landed the repository-side ordering (`SortSpecification`, `SortOrderBuilder`, the sort overloads), and its consumers are not wired.

Wiring seven listings plus the rank restoration in the two certificate/key two-stage loads is its own piece of work, and doing it here would both balloon this PR and pre-empt that one — so the catalogue no longer over-promises: `sortable` is reported false, and `isOrderableField` is the predicate it becomes once the services pass the sort through. The catalogue ITest now asserts no published field claims to be sortable, with the reason stated where the flag is set.

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -120,9 +122,40 @@ public static SearchFieldDataDto prepareSearchForJSON(final SearchFieldObject at
...
 
+    /**
+     * Whether a listing may be ordered by a property field. A sort resolves to one scalar path from the queried root,
+     * so a field with no attribute behind it has nothing to order by; a native array holds many values per row rather
+     * than one; and a JSON-path field's attribute is the whole JSONB column, which would order by the serialized
+     * document instead of by the value the field names.
+     */
+    private static boolean isSortable(final FilterField filterField) {
+        return filterField.getFieldAttribute() != null && !filterField.isNativeArrayField()
+                && filterField.getJsonPath() == null;
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed in 3ca286029. Both cases are now excluded, and the reasoning generalised so a new derived field is classified without an edit: a field with a non-null `expectedValue` displays a comparison rather than the value (`PRIVATE_KEY`), and a field sharing one attribute with another field of the same resource displays something drawn out of it (the three validation checks all read `certificateValidationResult`) — the latter derived from `FilterField.values()` rather than listed.

Separately, the flag is reported as false altogether for now, since no listing honours a sort yet — see the triage thread on `SearchHelper.java:52`. The predicate is asserted in `ColumnCatalogueFlagsITest`, not the plain unit test: `FilterField`'s class initializer reads the JPA static metamodel and fails without a persistence context.

src/test/java/com/otilm/core/integration/search/AttributeColumnProjectionITest.java
Jakub-Bartal-01: Fixed in 3ca286029, exactly as suggested. `partial` is now given a value for `environment` and none for `owning-team`, and the test asserts the projected key set is precisely `{environment|TEXT}` with the stored value on it — so an empty sibling entry appearing beside a real value now fails.

src/test/java/com/otilm/core/integration/search/AttributeColumnProjectionITest.java
```diff
@@ -0,0 +1,238 @@
...
+        Assertions.assertNull(entry(discoveries, withoutValues).getAttributeValues());
+    }
+
+    @Test
+    void aRequestedColumnTheObjectHasNoValueForIsAbsentRatherThanEmpty() {
+        Discovery partial = saveDiscovery("discovery-partial");
+        List<DiscoveryListDto> discoveries = list(List.of(customColumn(ENVIRONMENT), customColumn(OWNING_TEAM)));
+
+        Assertions.assertNull(entry(discoveries, partial).getAttributeValues());
+    }
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed in 3ca286029 — see the triage thread on this test. The fixture owns one of the two requested attributes and the assertion pins the exact key set, so it can no longer pass by repeating the no-values case.

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Not changed in 3ca286029, and I'd like to push back on doing it in this PR — the finding is fair, but the fix as scoped is not something I can derive.

Two things I checked. First, there is no mechanical narrowing available: every one of the 181 `FilterField` entries has a non-null `fieldAttribute`, so the obvious rule (a marker-only field has no value to render) excludes nothing. Second, the criterion in the finding is partly a frontend fact — "a renderer in the frontend registry" — which the catalogue cannot see, so the backend cannot compute it.

That leaves adding an explicit per-field displayability decision to `FilterField`, i.e. a judgement call on up to 181 entries across seven inventories about which fields the listing payload actually carries. That is a product decision I would be guessing at, and getting it wrong silently removes columns users can pick today. I'd rather it were its own issue with the inventory owners, than 181 guesses buried in this PR — happy to raise it if you agree, or to take a rule if you have one in mind.

Worth noting the blast radius is a blank column rather than an error, and the two flags that do carry a risk of over-promising or leaking are fixed here.

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
Jakub-Bartal-01: Done in 3ca286029 — replaced with the durable contract: a request that names no attribute-sourced column leaves every listing entry unchanged.

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Done in 3ca286029 — the future tense is gone and the standing technical reason kept: ordering by an attribute needs a correlated scalar subquery over JSONB that no index supports, and a multi-valued attribute leaves the sort key ambiguous.

src/test/java/com/otilm/core/integration/search/ColumnCatalogueFlagsITest.java
Jakub-Bartal-01: Done in 3ca286029. That test became `aNativeArrayPropertyFieldIsDisplayable` (sortability is now asserted for every field in one place), and its comment is just the stable reason: a multi-valued field has no single scalar key to order a row by, but it does have values to render.

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
Jakub-Bartal-01: Dropped in 3ca286029.

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
Jakub-Bartal-01: Dropped in 3ca286029.

src/test/java/com/otilm/core/integration/search/AttributeColumnProjectionITest.java
Jakub-Bartal-01: Done in 3ca286029 — the class Javadoc is now one line: projection of attribute-sourced columns into a listing, exercised end to end through the discovery listing. The count and the "cheapest to populate" claim are gone.

src/test/java/com/otilm/core/integration/search/AttributeColumnProjectionITest.java
Jakub-Bartal-01: Dropped in 3ca286029. The two fixture fields now stand on their names.

src/main/java/com/otilm/core/dao/repository/AttributeContent2ObjectRepository.java
```diff
@@ -84,6 +85,35 @@ List<ObjectAttributeContent> getObjectCustomAttributesContent(@Param("attributeT
...
+                    AND aco.objectType = :objectType AND aco.objectUuid IN (:objectUuids)
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed in 3ca286029. Confirmed the mechanism: `setEnabled(true)` appears in exactly one place in the attribute engine — the new-custom-attribute branch — the DATA and META creation paths never call it, and the column is a nullable `Boolean`, so the predicate excluded both sources entirely.

Predicated by attribute type, as suggested: the `enabled` check now applies to custom definitions only. Added an ITest storing metadata on a discovery and asserting the metadata column projects, so the source is covered rather than assumed.

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Follow-up: the wiring is already tracked as **core#2031** ("Server-side sorting on the configurable-column list endpoints"), still open, with core#2033 covering attribute-sourced sorting after it. So the split is deliberate — PR #2157 landed the repository-side ordering, #2031 wires the endpoints — and `sortable` becoming true is #2031's to flip, together with the rank restoration in the two-stage certificate and key loads.

That also makes `isOrderableField` the thing #2031 picks up rather than a dead branch: it is the predicate, already narrowed for derived fields, waiting on the one call site.

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
```diff
@@ -0,0 +1,286 @@
...
+        if (contentType == AttributeContentType.FILE && value.getData() instanceof FileAttributeContentData file) {
+            FileAttributeContentData withoutBody = new FileAttributeContentData();
+            withoutBody.setFileName(file.getFileName());
+            withoutBody.setMimeType(file.getMimeType());
+
+            BaseAttributeContentV3<Serializable> reduced = referenceOnly(value, contentType);
+            reduced.setData(withoutBody);
+            return reduced;
+        }
+        return value;
```
lubomirw: **[High]** | Security | Confidence: High

`OBJECT` is arbitrary serializable data, while `RESOURCE` subclasses can carry nested attributes, certificate bodies, or `SecretContent`. The list-cell renderer uses only the reference or resource/name/UUID, so projecting the complete value exposes data that the column never needs.

_Suggested:_ Reduce `OBJECT` to reference-only and reduce `RESOURCE` to its resource, name, and UUID identity fields. Add tests proving nested attributes, certificate bodies, and secret content are absent.

<sub>flagged by Superpowers review; verified against the interfaces content classes and frontend list-cell renderer</sub>

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -120,9 +156,62 @@ public static SearchFieldDataDto prepareSearchForJSON(final SearchFieldObject at
...
+    /**
+     * Whether an attribute field may be requested as a column.
+     *
+     * <p>
+     * Three kinds of content are withheld. A secret is never rendered anywhere. Encrypted content is stored as
+     * ciphertext that only its own decryption path can read, and a listing does not take that path. A code block is
+     * multi-line by construction - the frontend renders it as a block element - so a single-line table cell cannot hold
+     * one without breaking the row.
+     */
+    private static boolean isDisplayable(final SearchFieldObject attributeSearchInfo) {
```
lubomirw, line 214: **[Medium]** | Security | Confidence: High

Column eligibility ignores the attribute definition's `visible` flag, and the projection query does not carry it either. Attributes documented as helper-only can therefore appear in the picker and be returned when manually requested.

_Suggested:_ Carry definition visibility through `SearchFieldObject` and `ProjectedAttributeContent`, require `visible=true` both when setting `displayable` and when collecting projected rows, and cover hidden definitions with regression tests.

<sub>flagged by Superpowers review; verified against both attribute-property documentation pages and the catalogue/projection constructors</sub>

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -48,6 +82,8 @@ public static SearchFieldDataDto prepareSearch(final FilterField filterField, Ob
             values = withoutNull;
         }
         fieldDataDto.setValue(values);
+        fieldDataDto.setDisplayable(true);
```
lubomirw, line 85: **[Medium]** | Correctness | Confidence: High

Every property filter is advertised as displayable even when the listing DTO and frontend registry cannot supply it. For certificates, fields such as `KEY_USAGE`, `SUBJECT_ALTERNATIVE_NAMES`, and `CERTIFICATE_PROTOCOL` are offered but fall through to a permanently empty attribute cell.

_Suggested:_ Publish `displayable=true` only for property fields backed by each listing contract, or add the missing DTO projections and frontend renderers. Add a contract test for every property field marked displayable.

<sub>flagged by Superpowers review; verified against the checked-out `feat/2018-column-picker` frontend registry</sub>

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
```diff
@@ -0,0 +1,286 @@
...
+     * CREDENTIAL data is a nested attribute list that can carry secret-bearing content, and the cell shows the
+     * reference, so only the reference is projected. FILE data carries the base64 body, which no cell shows - a file
+     * renders as its name and media type - so a page of rows would otherwise serialize every file it lists.
+     */
+    static BaseAttributeContentV3<?> toProjectedValue(BaseAttributeContentV3<?> value,
+            AttributeContentType contentType) {
+        if (contentType == AttributeContentType.CREDENTIAL) {
+            return referenceOnly(value, contentType);
+        }
+        if (contentType == AttributeContentType.FILE && value.getData() instanceof FileAttributeContentData file) {
```
lubomirw: **[Medium]** | Performance | Confidence: High

Persisted V2 JSON is deserialized as generic `BaseAttributeContentV2`, so file `data` becomes a map. The `instanceof FileAttributeContentData` guard does not match, leaving the complete base64 body in every projected file value.

_Suggested:_ Convert V2 `FILE` data to its typed shape before reduction, or safely rebuild only file name and MIME type from the generic value. Add a persistence-level V2 file projection test.

<sub>flagged by Superpowers review; verified against `AttributeContentDeserializer`, `AttributeContentItem`, and `AttributeVersionHelper`</sub>

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -317,6 +324,13 @@ public CryptographicKeyResponseDto listCryptographicKeys(SecurityFilter filter,
             return dto;
         }).toList();
 
+        // A key's own attributes hang off the key, so several listed items may share one set of values, while its
+        // metadata hangs off each item - the same split the key detail read makes.
+        attributeColumnProjector
+                .project(Resource.CRYPTOGRAPHIC_KEY, request.getColumns(), listedKeyDtos,
```
lubomirw, line 330: **[Low]** | Testing | Confidence: High

The end-to-end projection suite exercises discoveries only, leaving the key listing's unique split untested: custom/data values resolve from the shared key while metadata resolves from each key item.

_Suggested:_ Add an integration test with multiple items sharing a key, key-level custom/data values, and distinct item-level metadata; assert every row receives values from the correct UUID source.

<sub>flagged by Guided review; verified against the projection integration suite and key listing path</sub>

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
```diff
@@ -0,0 +1,286 @@
...
+
+    private static List<RequestedColumn> metadataColumns(List<RequestedColumn> requested) {
+        return requested.stream().filter(column -> column.attributeType() == AttributeType.META).toList();
+    }
+
+    private static List<RequestedColumn> columnsOtherThanMetadata(List<RequestedColumn> requested) {
+        return requested.stream().filter(column -> column.attributeType() != AttributeType.META).toList();
+    }
+
+    /** Loads and collects the columns that share one uuid resolver, adding what it finds to {@code collected}. */
```
lubomirw: **[Nit]** | Maintainability | Confidence: High

The one-line private-method Javadoc restates `collectPass` and its parameters without adding a contract or rationale.

_Suggested:_ Remove the Javadoc line.

<sub>flagged by Hygiene; verified against the method signature and body</sub>

src/test/java/com/otilm/core/integration/search/AttributeColumnProjectionITest.java
```diff
@@ -0,0 +1,334 @@
...
+        Assertions
+                .assertEquals(List.of("production", "staging", "dr-site"),
+                        values.stream().map(value -> (Object) value.getData()).toList());
+    }
+
+    @Test
+    void anObjectWithoutAValueIsStillReturned() {
+        List<DiscoveryListDto> discoveries = list(List.of(customColumn(ENVIRONMENT)));
+
+        // The point of the assertion: absence omits the entry, it does not drop the row from the page.
```
lubomirw: **[Nit]** | Maintainability | Confidence: High

The comment repeats the test name and the immediately following null assertion.

_Suggested:_ Remove the comment line.

<sub>flagged by Hygiene; verified against the surrounding test</sub>

src/test/java/com/otilm/core/integration/search/ColumnCatalogueFlagsITest.java
```diff
@@ -0,0 +1,168 @@
...
+    void aNativeArrayPropertyFieldWouldNotBeOrderable() {
+        // It holds many values per row, so there is no single value to order the row by.
+        Assertions.assertFalse(SearchHelper.isOrderableField(FilterField.TIME_QUALITY_CONFIGURATION_NTP_SERVERS));
+    }
+
+    @ParameterizedTest
+    @EnumSource(value = FilterField.class,
+            names = {"OCSP_VALIDATION", "CRL_VALIDATION", "SIGNATURE_VALIDATION", "PRIVATE_KEY"})
+    void aDerivedPropertyFieldWouldNotBeOrderable(FilterField filterField) {
+        // What these display is not what their attribute holds: the three validation checks each name one check inside
```
lubomirw: **[Nit]** | Maintainability | Confidence: High

The explanation hard-codes the current count of validation-check fields, so adding another field can make the comment stale without changing the tested rule.

_Suggested:_ Replace "the three validation checks" with "the validation-check fields".

<sub>flagged by Hygiene; verified against the parameterized enum list</sub>

lubomirw (changes requested):
Automated multi-reviewer triage - 8 findings (REQUEST_CHANGES).

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
Jakub-Bartal-01: Fixed in 9ab36276b. Confirmed both halves against the content classes: `ObjectAttributeContentV3.data` is a bare `Serializable` with no declared shape, and `ResourceObjectContentData` has three subtypes — `ResourceCertificateContentData` carries a base64 `content`, `ResourceSecretContentData` carries a `SecretContent`, and `ResourceSimpleContentData` carries a `List<ResponseAttribute>`. Against the cell renderer, `toLabel` reads only `reference` for OBJECT, and `toResourceValue` reads only `reference`, `data.resource`, `data.uuid` and `data.name`.

OBJECT is now reference-only, the same shape CREDENTIAL already took. RESOURCE is reduced by rebuilding the base `ResourceObjectContentData` from the subtype rather than by clearing the offending fields on it, so it is a whitelist: a fourth subtype added later is reduced without an edit here, instead of leaking until someone remembers this branch.

`ProjectedContentShapeTest` pins the wire shape rather than the object: it serializes the projected value with the attribute mapper and asserts the certificate body, the secret and the `attributes` key are absent from the json, with a separate test that the three identity fields survive.

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Fixed in 9ab36276b, both halves.

One correction to the framing before the fix: this is not a case of the projection being laxer than an existing path. `isVisible()` is read nowhere in core — not on the detail read, not in the filter catalogue — so today a hidden attribute is already returned by the detail page and already published as a filterable field. So I have implemented it as what the flag documents ("visible and can be displayed, otherwise it should be hidden to the user"), not as a repair of a privilege gap, and the other read paths keep their current behaviour rather than being changed under this PR.

`visible` is not a column on `attribute_definition` — it lives on the properties inside the definition document, and `BaseAttribute` declares no accessor for it, since each of the three attribute shapes narrows the return type to its own properties class. So: a small `AttributeDefinitionProperties.isVisible(BaseAttribute)` resolves the shape, `SearchFieldObject` carries the flag (it already receives `ad.definition`, so no extra query), and `isDisplayable` requires it. The projection query now selects `ad.definition` alongside the content and the projector drops a hidden row on the way out — the same second-lock the withheld content types get, and the catalogue queries already select that document, so this is not a new kind of read. Regression tests on both sides: the catalogue does not offer a hidden attribute, and naming one as a column projects no values.

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
Jakub-Bartal-01: Fixed in 9ab36276b. Confirmed the mechanism: `AttributeContentDeserializer` keys on the presence of `contentType` in the persisted json, and version 2 content has none, so it binds `BaseAttributeContentV2<T>` with `T` unresolved and Jackson produces a `LinkedHashMap`. `convertAttributeContentToV3` copies that data across verbatim, so the `instanceof FileAttributeContentData` guard never matched and the whole body went out.

Taken the second option, and generalised: each reduction reads the fields it keeps by name when the typed class is absent, so FILE keeps `fileName`/`mimeType` and RESOURCE keeps `resource`/`uuid`/`name` under either contract. A shape neither branch recognises — a scalar, or null — now yields no data at all rather than being passed through, so the failure mode is a reference-only cell rather than a leak. That also replaces the previous `aFileValueThatCarriesNoFileDataIsLeftAlone` test, which asserted exactly the pass-through this finding is about. Version 2 file and resource values each have a test asserting the body is absent from the serialized projection.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
Jakub-Bartal-01: Added in 9ab36276b as `KeyAttributeColumnProjectionITest`: one key with two items, a key-level custom value and a distinct metadata value on each item, driven through `listCryptographicKeys`. Three assertions — the key-level value reaches both rows, each item gets its own metadata and not its sibling's, and both sources land on the same entry, which is the one the two-pass accumulation could get wrong by replacing rather than adding.

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
Jakub-Bartal-01: Dropped in 9ab36276b.

src/test/java/com/otilm/core/integration/search/AttributeColumnProjectionITest.java
Jakub-Bartal-01: Dropped in 9ab36276b.

src/test/java/com/otilm/core/integration/search/ColumnCatalogueFlagsITest.java
Jakub-Bartal-01: Done in 9ab36276b — "the validation-check fields", and the same in the `SearchHelper` Javadoc the test mirrors.

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Not changed, and I want to hold the position from the round-one thread on this line — but with better evidence than I gave then, because your three examples check out.

Verified the backend half: `CertificateDetailDtoMapper.toListDto` sets neither `keyUsage` nor `subjectAlternativeNames`, and the protocol association is not mapped either, so those three fields are offered and would render blank. That part of the finding is correct and I am not disputing it.

What I still cannot derive is the rule. Two obstacles, both checked:

1. There is no mechanical narrowing. All 181 `FilterField` entries have a non-null `fieldAttribute`, so "a marker-only field has nothing to render" excludes none of them. Nor does a structural property predict it: `CERTIFICATE_PROTOCOL` is reached through a join and is absent from the list DTO, but `raProfile` and `groups` are reached through joins and are present. A reflective contract test would need a `FilterField` → DTO-property mapping that does not exist — `CERTIFICATE_PROTOCOL` publishes `protocolAssociation` as its attribute, and nothing links that to a DTO getter — so every such test needs a hand-written exception list, which is the judgement call again with extra machinery around it.

2. The second criterion in the finding is a frontend fact. You verified it against `feat/2018-column-picker`, which is unmerged; the merged frontend has no per-property renderer registry to read. So the authority for half the rule does not exist yet in either repo, and a backend list written against an unmerged branch becomes a second source of truth that drifts from it.

So the actual work is an explicit per-field displayability decision over 181 entries across seven listing mappers, taken jointly with whoever owns the frontend registry. That is a real piece of work with a real answer — it is just not one this PR can take without guessing, and a wrong guess silently removes a column rather than leaving a blank one.

I would rather split it the way `sortable` was split. That flag over-promised too; it is reported false here and core#2031 owns turning it on. This is the same shape: I will raise it as its own issue against the column-picker epic, with your three certificate fields as the worked examples, and let it be decided with the inventory and frontend owners rather than inside this PR. Say the word and I will open it; if you would rather I take a rule you have in mind instead, I will implement that here.

lubomirw: Agreed, not in this PR — no structural predictor, and I don't want 181 guesses buried in here.

One correction: the frontend registry is merged. #2084 landed on main at 09:18 today, a few hours before your reply, so `registry.ts` and `columnPicker.ts` are there now.

Splitting the comment, since I folded two things into it. Fields the DTO carries but nothing renders (`FINGERPRINT`, `SUBJECTDN`, `KEY_SIZE`, ...) are already fe-administrator#2067's scope — but fe-administrator#2020 can pass all seven of its ACs and still ship 24 blank certificate columns, so I'll add an AC there. Your three examples, where nothing carries the value, have no owner. Please open that one against the column-picker epic — it's core#2163 in reverse.

One lead for it: `getProjectedContent` keys on `column.fieldSource` without caring which source, and `FilterFieldSource` has `Property`, so `attributeValues.property.KEY_USAGE` would render as-is. Only `AttributeColumnProjector:189` stops it. Might beat adding N DTO fields and N renderers — worth costing first.

Not blocking the merge.

lubomirw (approved):

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Opened as OmniTrustILM/core#2180, under the column-picker epic ilm#133. Version 2.20.0, Sprint 4, Estimate 3, Component Filters & Rules — same shape as core#2163, which the body points at as its mirror.

It carries your three certificate fields as the worked examples, the two candidate rules and why each fails, and the scope boundary you drew: fields the DTO carries but nothing renders belong to fe-administrator#2067 with the extra AC on #2020, while this one covers the case where nothing carries the value at all.

Your projection lead is in there as a step before any DTO work: `parseRequestedColumns` drops a column whose source has no attribute type, which is what excludes PROPERTY, and the frontend `getProjectedContent` does not care which source it reads — so `attributeValues.property.KEY_USAGE` would render as-is. The DoD asks for that to be costed against adding N DTO fields and N renderers, with the choice recorded, since it may change the answer for all seven inventories.

Also worth noting now that core#2031 has landed: `sortable` is off the unconditional list, so `displayable` is the last capability flag still set for every field regardless.

PR #2164, opened 2026-08-28
Resolve request attributes from the connector schema and re-enable merge modes

### LINKED ISSUE #1944: Resolve connector request attributes and re-enable merge modes
### Description

Consume the v3 `/request/attributes` connector endpoint and lift the hold that OmniTrustILM/core#1872 put on RA-profile request-attribute configuration.

**Connector-sourced resolution.** `RequestAttributeSetResolver` and `RaProfileCertificateRequestAttributeServiceImpl` gain the authority connector as a source: when an RA profile is in play, Core calls the adapter's `listCertificateRequestAttributes` and resolves the connector set together with the platform default set (`certificatesRequestAttributes` setting / `CsrAttributes` seed) and the per-RA-profile static override. `GET /v1/certificates/csr/attributes` returns that resolved set when `raProfileUuid` is supplied.

**Re-enable what was held back.** core#1872 rejects any merge mode other than `Static only` and rejects value-source bindings, throwing operation-not-supported, with the supporting code deliberately retained for exactly this moment. Remove those rejections and re-enable both: merge modes combine the connector set with the platform/static sets, and value-source bindings resolve during request building. The tests core#1872 changed are updated back to assert the working behaviour rather than deleted.

**The connector stays optional.** A 404 or `OPERATION_NOT_SUPPORTED` from `/request/attributes` must leave resolution exactly as it is today — platform default set, or the per-RA-profile static override. The defensive helper from the prerequisite issue already covers this; the point here is that the resolution path must have a no-connector branch that is regression-tested, not merely tolerated.

**Structured content.** The merged set is what feeds the structured `requestContent` projection for connectors advertising `CERTIFICATE_REQUEST_STRUCTURED`. Document the interaction so a merge-mode change cannot silently alter what rides the wire.

### Acceptance Criteria

- [ ] `GET /v1/certificates/csr/attributes` consults the authority connector when `raProfileUuid` is supplied and returns the resolved set.
- [ ] Merge modes beyond `Static only` accepted; the operation-not-supported rejection from OmniTrustILM/core#1872 removed and its tests restored to assert working behaviour.
- [ ] Value-source bindings accepted and applied during request building.
- [ ] Connector absent (404 / `OPERATION_NOT_SUPPORTED`): resolution byte-identical to 2.19.x — regression test.
- [ ] The merged set feeds the `CERTIFICATE_REQUEST_STRUCTURED` projection; interaction documented and covered by test.
- [ ] `mvn -B verify` passes including integration tests; new-lines coverage meets the Sonar gate.

klaraf755:
Closes #1944

Depends on OmniTrustILM/interfaces#921, which un-hides `valueSourceBindings` on the RA-profile request-attribute DTOs. CI here needs that snapshot for the bindings to arrive over the wire; the merge-mode half is independent of it.

## The connector set came from the wrong endpoint

Resolution treated `listIssueCertificateAttributes` — the issue *operation* schema — as the "connector set". That is a different axis: those attributes are values Core sends with an issue call, not the identity fields a CA asks a request to carry. It now consults `listCertificateRequestAttributes`, the schema the v3 `/request/attributes` endpoint serves, through a new `ExtendedAttributeService` method following the renew/identify pattern (no legacy-connector check: the schema is v3-only, so an older authority routes to an adapter that returns empty rather than a version-framed 404).

Nothing in the field changes behaviour: OmniTrustILM/core#1872 rejected every mode but `Static only`, so no stored configuration could reach the connector-consulting branch.

## Merge modes and value-source bindings accepted again

Both rejections are gone from `updateConfiguration`. The tests OmniTrustILM/core#1872 rewrote now assert the working behaviour instead — the round-trip test carries `CONNECTOR_ONLY` and a binding with `collectionRef` and params again, and the two rejection tests became a `@EnumSource` pass over all three modes plus a binding-replacement test.

`effectiveMode(null)` stays `STATIC_ONLY`, matching the interfaces default: merge modes are opt-in. OmniTrustILM/core#1872 left the `merge` javadoc still claiming `null → MERGE`; corrected.

## Connector absent

A connector that does not serve the endpoint resolves to an empty schema in the adapter (404, `OPERATION_NOT_SUPPORTED`, bare 501, or a null body), so resolution lands on the static set or the platform default exactly as before — regression-tested in both directions in `CsrGenerationAttributesITest`.

The fail-open counterpart is pinned too: a 500 from `/request/attributes` must *reject*, not fall back, since it is not one of the statuses the adapter resolves to empty. That is one line apart in `schemaOrEmpty`, so it gets its own test.

## Merge mode reaches the wire, not just the form

The issue path and the CSR-attributes endpoint both resolve through `IssuanceDefinitionResolver`, so a definition the merge mode admits is a definition the structured `requestContent` projection maps for connectors advertising `CERTIFICATE_REQUEST_STRUCTURED`. Documented there and covered by test.

## One test changed mechanism

`rejectsUploadedCsr_whenStrictProfileAndAttributeSetCannotBeResolved` produced its failure by stubbing a 500 on the v2 issue-attributes endpoint. Under the correct source that scenario is unreachable for a v2 authority — the v2 adapter returns empty by contract without a wire call — so it moved onto a v3 authority via `AuthorityFixtures.v3Authority`. Same assertion, real mechanism.

## Also here

A WARN when a non-`STATIC_ONLY` mode resolves an empty connector set. An operator picking `Connector only` against an authority that serves no request schema gets the platform default set — the connector contributes nothing and the mode skips the static set — and nothing said so. Every v2 authority is in that position by contract. `Static only` stays silent.

`mvn -B verify` passes: 5808 tests. `ContextSignatureGuardTest.BASELINE` unchanged.


src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
```diff
@@ -137,15 +151,6 @@ public RaProfileCertificateRequestAttributesDto getConfiguration(RaProfile raPro
 
     @Override
     public void updateConfiguration(RaProfile raProfile, RaProfileCertificateRequestAttributesUpdateDto request) {
-        AttributeSetMergeMode effectiveMode = RequestAttributeSetResolver.effectiveMode(request.getMergeMode());
-        if (effectiveMode != AttributeSetMergeMode.STATIC_ONLY) {
```
vyskocilm, line 141: 🔴 **bug — merged set can carry two definitions with the same `name`, which 500s the attribute engine.**

Removing this rejection makes `MERGE` reachable. `RequestAttributeSetResolver.mergeKey` (`:71`, untouched) keys on `"uuid:" + uuid` whenever a UUID is present and only falls back to `"name:"`. A connector `commonName` and the static/default `commonName` (fixed UUID `9abaeba0-…`) have different UUIDs, so `mergeUnion` keeps both.

`IssuanceDefinitionResolver.resolve` then hands both to `AttributeEngine.validateUpdateDataAttributes(null, null, …)`, which registers two `attribute_definition` rows and later does `…collect(Collectors.toMap(AttributeDefinition::getName, d -> d))` (`AttributeEngine.java:752`) → unchecked `IllegalStateException: Duplicate key`. That is a 500 on every structured issue/register for the profile.

Nothing covers it: every new test uses distinct names (`c1`/`connector-cn` vs `s1`/`static-only`).

Fix: dedupe by name as well as UUID in `mergeKey`/`mergeUnion`, or reject a merged set containing duplicate names.

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
```diff
@@ -137,15 +151,6 @@ public RaProfileCertificateRequestAttributesDto getConfiguration(RaProfile raPro
...
-            throw new ValidationException(
-                    String.format("Merge mode %s is not supported. Use `Static Only` mode.", effectiveMode));
-        }
-        if (request.getValueSourceBindings() != null && !request.getValueSourceBindings().isEmpty()) {
```
vyskocilm, line 145: 🔴 **bug — with this rejection gone, bindings are unreachable *and* destructive over REST until interfaces#921 merges.**

`pom.xml:66` pins `interfaces:2.20.0-SNAPSHOT`, and on interfaces **main** `RaProfileCertificateRequestAttributesUpdateDto.valueSourceBindings` still carries `@JsonIgnore @Schema(hidden = true)` (`:44-46`) — the re-expose lives on the unmerged #921 branch.

So over HTTP the field never deserializes: `request.getValueSourceBindings()` is always the default empty list. Bindings can never be created, and every `PATCH /authorities/{a}/raProfiles/{r}/requestAttributes` reaches `replaceValueSourceBindings(uuid, [])` — **deleting any binding rows that exist**. Any unrelated edit to the static set silently drops the bindings.

The new ITests set `valueSourceBindings` on the DTO object directly, bypassing Jackson, so they pass while the REST path does nothing.

The body already says this depends on interfaces#921; this is the concrete reason it is a hard ordering constraint rather than a soft one. Worth a guard that skips the replace when the field is absent, so the window is not destructive even if the merge order slips.

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
```diff
@@ -81,10 +85,18 @@ private List<BaseAttribute> resolve(RaProfile raProfile, RaProfileCertificateReq
...
                 : deserializeOrEmpty(stored.getRequestAttributes());
-        List<BaseAttribute> connectorSet = RequestAttributeSetResolver
-                .effectiveMode(mode) == AttributeSetMergeMode.STATIC_ONLY
-                        ? List.of()
-                        : listConnectorIssueAttributes(raProfile);
+        AttributeSetMergeMode effective = RequestAttributeSetResolver.effectiveMode(mode);
+        List<BaseAttribute> connectorSet = effective == AttributeSetMergeMode.STATIC_ONLY
+                ? List.of()
+                : listConnectorRequestAttributes(raProfile);
+        if (effective != AttributeSetMergeMode.STATIC_ONLY && connectorSet.isEmpty()) {
```
vyskocilm: 🟡 **risk — this WARN fires on the healthy case, on a hot path.**

The condition is evaluated before the merge, so under `MERGE` with a **non-empty static set** on any authority that serves no request schema — every v2 authority by contract (`AuthorityProviderV2Adapter:193`) — resolution lands correctly on the static set, which your own `connectorWithoutRequestSchemaFallsBackToTheStaticOverride` pins as the desired 2.19.x behaviour. Yet the log says the resolved set came from "the static set or the platform default only", implying something went wrong.

`resolveIssueAttributeSet` runs per certificate request, per EST `/csrattrs` (`EstCsrAttrsServiceImpl:26`) and per ACME/SCEP/CMP request (`ProtocolRequestAttributeValidator:55`), so one such profile emits this on every request, forever, with no operator action that would stop it.

Fix: warn on `merged.isEmpty()` — the case where the fallback actually bit — or restrict the condition to `CONNECTOR_ONLY`.

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
```diff
@@ -137,15 +151,6 @@ public RaProfileCertificateRequestAttributesDto getConfiguration(RaProfile raPro
...
-        AttributeSetMergeMode effectiveMode = RequestAttributeSetResolver.effectiveMode(request.getMergeMode());
-        if (effectiveMode != AttributeSetMergeMode.STATIC_ONLY) {
-            throw new ValidationException(
-                    String.format("Merge mode %s is not supported. Use `Static Only` mode.", effectiveMode));
-        }
-        if (request.getValueSourceBindings() != null && !request.getValueSourceBindings().isEmpty()) {
-            throw new ValidationException(
-                    "Value-source bindings are not supported in this version. Use the `Static Only` mode without bindings.");
-        }
         AttributeEngine.validateRequestAttributeDefinitions(request.getRequestAttributes());
```
vyskocilm, line 155: 🟡 **risk — `updateConfiguration` is no longer atomic in the sense that now matters.**

`saveStaticSet` and `replaceValueSourceBindings` are two independent `@Transactional(REQUIRED)` units, and this method carries no ambient transaction of its own. Called through `RaProfileServiceImpl` (class-level `@Transactional`) they join one tx — but called directly, as this PR's own tests do, they do not.

Previously the second call was always a no-op delete, because non-empty bindings were rejected right here. Now it carries real data: if it fails (`value_source_kind NOT NULL`, or a deadlock against a concurrent `deleteByRaProfileUuid`) the caller gets a 500 with the static set and merge mode already committed and the bindings unchanged — a half-applied configuration.

Separately, `toBindingEntities` (`:245`) calls `binding.getValueSourceType().name()` unguarded. The HTTP path is covered by `@Valid` + `@NotNull`, but this is a public service method and the `ValidationException` that used to short-circuit any non-empty binding list is gone. A null check with a domain-shaped message costs nothing.

src/main/java/com/otilm/core/certificate/request/RequestAttributeSetResolver.java
```diff
@@ -38,7 +38,7 @@ public static AttributeSetMergeMode effectiveMode(AttributeSetMergeMode mode) {
     /**
      * @param staticSet the RA-Profile static definitions (may be {@code null})
      * @param connectorSet the connector-supplied dynamic definitions (may be {@code null})
-     * @param mode the per-RA-Profile merge mode; {@code null} is treated as {@link AttributeSetMergeMode#MERGE}
+     * @param mode the per-RA-Profile merge mode; {@code null} is treated as {@link AttributeSetMergeMode#STATIC_ONLY}
```
vyskocilm, line 41: 🟡 **risk — the doc sweep is incomplete; three more copies of the old claim survive.**

This correction is right, and so is the read-view comment in the impl. But `null -> MERGE` is still asserted in two files the PR does not touch:

- `RaProfileCertificateRequestAttributeService.java:24` — `@param mode the merge mode; {@code null} defaults to … MERGE`
- `RaProfileCertificateRequestAttributeService.java:30-31` — `defaults to {@link AttributeSetMergeMode#MERGE} when unset`
- `RaProfileCertificateRequestAttribute.java:53` — "`null` is treated as `AttributeSetMergeMode#MERGE` by the resolver"

The first two are on the public service interface, the most authoritative doc surface for exactly the two `resolveIssueAttributeSet` overloads; the third is on the entity, so a reader expects a `merge_mode IS NULL` row to consult the connector. `effectiveMode(null)` returns `STATIC_ONLY`.

This matters more now than before the PR: with merge modes live and the connector genuinely consulted, a reader who trusts the interface will believe an unconfigured profile calls out to the authority.

src/test/java/com/otilm/core/integration/service/v2/UploadedCsrValidationITest.java
```diff
@@ -165,18 +178,21 @@ void rejectsMalformedUploadedCsr_asValidationError_notServerFault() throws Excep
...
-                                .urlPathMatching(
-                                        "/v2/authorityProvider/authorities/[^/]+/certificates/issue/attributes"))
+                        .post(WireMock.urlPathMatching("/v3/authorityProvider/certificates/request/attributes"))
                         .willReturn(WireMock.serverError().withBody("authority connector is unavailable")));
 
         ClientCertificateRequestDto request = uploadRequest(csrMissingCommonName);
+        request.setRaProfileUuid(v3.raProfile().getUuid());
 
         // when / then — strict policy fails closed: an unavailable attribute set blocks issuance
         assertThatThrownBy(() -> clientOperationService.submitCertificateRequest(request, null))
```
vyskocilm, line 200: 🟡 **risk — this assertion cannot tell the path it tests from the path that means the test reached nothing.**

`ProtocolRequestAttributeValidator.resolveDefinitions` catches `ConnectorException | NotFoundException` and wraps **both** into the same `CertificateException`, whose outer sentence is `"Request-attribute set is unavailable; strict RA profile '%s' cannot validate the request (%s)"`. Only the parenthesised reason differs — `the authority connector is unavailable` for the 500 you stub, versus `the request-attribute set is not configured on the authority connector` for a `NotFoundException`. `hasMessageContaining("unavailable")` matches both, because the word is in the shared wrapper.

There is also no `mockServer.verify`, so nothing proves the stub was hit. The test has just moved onto a brand-new v3 fixture with no other stubs — if that fixture failed to route to the v3 adapter, this would stay green while exercising none of the 500 boundary it exists for.

Fix: assert `"the authority connector is unavailable"`, and add `mockServer.verify(postRequestedFor(urlPathMatching("/v3/authorityProvider/certificates/request/attributes")))`.

(I could not run this — Testcontainers is down on my machine — so this is read off the code, not off a failure.)

vyskocilm (commented):
**Review of #2164** — architecture/convention pass plus a correctness pass, both against head `33858c71`.

**Two merge-blockers** (inline, on the two removed rejections), four smaller items, and one open question. No architecture rule is broken: transaction and writer boundaries are right on the paths that newly change meaning, and the authorization surface is untouched.

**What ran here**

- ArchUnit fences: **102 tests, 0 failures**.
- `spotless:check` + `checkstyle:check`: 0 violations.
- 149 non-container unit tests green, incl. `ExtendedAttributeServiceImplTest` (29, both new ones), `AuthorityProviderV3AdapterTest` (66), `RequestAttributeSetResolverTest` (16).
- `ContextSignatureGuardTest` **run**, not just diffed — 2 green, and since it asserts exact equality that is real proof `BASELINE` is unchanged.

**What did not run: the container runtime is down here, so 5 of your 6 changed test classes never executed** — `CertificateRequestIntegrationITest`, `CsrGenerationAttributesITest`, `RaProfileCertificateRequestAttributeServiceITest`, `RaProfileServiceITest`, `UploadedCsrValidationITest`. CI is green on all of them, but that is CI's word, not mine, and a green ITest is exactly what a green-but-empty ITest looks like — which is why two comments below are about assertions rather than production code. Sonar coverage parity also not computed locally.

**Claims I checked and confirmed**

- *"Nothing in the field changes behaviour."* Holds, and it can be dated: `saveStaticSet` is the only write path to `merge_mode`, and both the feature (`4120f69e`, #1689) and the rejection (`bd8a499b`, #1872) are ancestors of `2.19.0`. No release ever shipped the endpoint without the rejection, so no deployment can hold a non-`STATIC_ONLY` row except by direct DB edit.
- The `schemaOrEmpty` 500-vs-not-offered boundary is right — three narrow catches, everything else rethrown — and all branches are covered by `AuthorityProviderV3AdapterTest`, which I ran.
- The connector call is now live where it used to be dead code, and no transaction spans it: both entry points are `NOT_SUPPORTED` and every caller is a separate bean.
- `replaceValueSourceBindings` is a real write now, and `RaProfileServiceImpl` being class-level `@Transactional` does join both writer calls into one tx **on that path** — see the atomicity comment for the path where it does not.

Small wording note, not a finding: the body says connector-absent is regression-tested in `CsrGenerationAttributesITest`, but that test mocks `ExtendedAttributeService` to empty, so it pins resolution-given-empty rather than the adapter's status mapping. The mapping is genuinely covered — in `AuthorityProviderV3AdapterTest`.

**Not anchorable to a diffed line — `collectionRef` is stored and echoed but never reaches the resolved definition.** `RequestAttributeSetResolver.toValueSource` (`:126`, untouched) builds a `ValueSource` from `kind` and `params` only; `ValueSource` has no `collectionRef` field, and `CollectionValueSourceProvider.resolveOptions` has no implementation and no caller in `src/main`. With the guard gone an operator can save `{valueSourceType: STATIC_LIST, collectionRef: "cmdb.servers"}`, read it back from `getConfiguration`, and get a definition with no way to fetch options. `valueSourceBindingBindsToAConnectorSuppliedDefinition` sets `collectionRef` but asserts only `getKind()`, so it passes either way. Either wire it through or reject it on save.

**Open question, not a finding.** `CONNECTOR_ONLY` against an authority serving no request schema resolves empty, skips the static set by design, and falls back to the platform default — so with `externalCsrValidationStrict` the operator's policy is silently replaced by the platform default set, and the only trace is the new WARN. Every v2 authority is in that position by contract. The nearby precedent (`ExtendedAttributeServiceImpl.mergeAndValidateRenewAttributes`) rejects rather than proceeds when the schema is empty. Should `updateConfiguration` reject `CONNECTOR_ONLY` on a non-v3 authority, or is the WARN the intended contract? Your call — just flagging it was never stated as a decision.

src/test/java/com/otilm/core/integration/service/CertificateRequestIntegrationITest.java
```diff
@@ -345,8 +346,40 @@ void throwsValidation_whenRaProfileUuidIsUnknown() {
...
+                        .post(WireMock.urlPathMatching("/v3/authorityProvider/certificates/request/attributes"))
+                        .willReturn(WireMock.serverError().withBody("request-attributes endpoint is unavailable")));
+        stubSigning();
+
+        var request = baseRequest();
+        request.setCsrAttributes(List.of(commonNameAttribute("FailClosedOnRequestSchema")));
+
+        // when / then
+        assertThatThrownBy(() -> clientOperationService.submitCertificateRequest(request, null))
+                .isInstanceOf(ConnectorException.class);
```
vyskocilm: 🟡 **risk — asserts the exception type but never its origin.**

The stated purpose (your comment just above) is that a 500 is *not* one of the statuses `schemaOrEmpty` resolves to empty. But `ConnectorException` is satisfied by any of its subtypes anywhere in `submitCertificateRequest` — `ConnectorEntityNotFoundException`, `ConnectorProblemException`, `ConnectorCommunicationException` — and an unstubbed WireMock route yields a 404 that maps to one.

So it does establish fail-closed rather than fail-open (fail-open would throw nothing), but not that the `/request/attributes` 500 is what closed it.

Fix: narrow to `ConnectorServerException` — the type `schemaOrEmpty` rethrows for a non-501 5xx — or verify the stubbed POST was received.

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
klaraf755: Confirmed and fixed in `cd9b1bcf5`. I traced it end to end: `Collectors.toMap(AttributeDefinition::getName, d -> d)` at `AttributeEngine:755`, reached with the resolver output from `ClientOperationServiceImpl:889` and `:959`.

`mergeUnion` now claims a definition by name as well as UUID, connector first, and `mergeKey` is gone. The fix goes slightly further than connector-vs-static: two same-named definitions from a single connector schema also collapse. Dropping one is data loss, but it is data the engine cannot store, so it beats a 500 on every issue.

Worth recording why nothing covered it — I raised this exact case earlier, then withdrew it after checking the projector (UUID-keyed, tolerant of duplicates) and `validateRequestAttributeDefinitions` (no name-uniqueness rule) and treating their tolerance as system-wide. The registration path is the one that cannot represent it. Two tests now pin it: connector wins a name collision with differing UUIDs, and a distinct name still contributes.

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
klaraf755: Fixed in `c33aaf49e`, and it needed a contract change to be fixable at all — `adc3d7d7` on OmniTrustILM/interfaces#921.

Core could not implement the guard you suggested on its own: `valueSourceBindings` defaulted to `new ArrayList<>()`, so after deserialization an omitted field and an explicitly empty array are indistinguishable. The DTO field now defaults to null, and `updateConfiguration` skips the replace when it is null. Omit to keep the stored bindings, send `[]` to clear them.

One correction to the framing: this is not only the pre-merge window. The endpoint is a `PATCH`, so with the empty-list default it would have kept wiping bindings after #921 merged, on any partial body that did not mention them. Two tests cover it — absent preserves and still writes the static set, empty clears.

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
klaraf755: Fixed in `c33aaf49e` — the warn moved after the merge and is gated on `merged.isEmpty()`, your first suggestion.

You are right about which case it was firing on. The RED test reproduced it verbatim before the fix: `MERGE` with a non-empty static set and an empty connector schema logged `…the resolved set comes from the static set or the platform default only` while resolving correctly to the static set. That is the steady state of every v2-backed profile, on a per-request path.

`mergeWithAStaticSetLogsNoWarningWhenTheConnectorServesNoSchema` now pins the silence, alongside the existing `staticOnlyResolutionLogsNoWarning`.

src/main/java/com/otilm/core/service/impl/RaProfileCertificateRequestAttributeServiceImpl.java
klaraf755: Both fixed in `c33aaf49e`. `updateConfiguration` now carries `@Transactional`, so the static-set write and the binding replace commit as one unit regardless of the caller — the direct-call path this PR's own tests use included.

The unguarded `getValueSourceType().name()` is now a `ValidationException` naming the target attribute. The RED test produced exactly the NPE you predicted: `Cannot invoke "ValueSourceType.name()" because the return value of "ValueSourceBindingDto.getValueSourceType()" is null`.

src/main/java/com/otilm/core/certificate/request/RequestAttributeSetResolver.java
klaraf755: Fixed in `f4b110e35`, which predates the other three commits here — same three locations you list. A grep for `AttributeSetMergeMode#MERGE`, `-> MERGE` and `to MERGE` across `src` now returns nothing.

Agreed the two on the service interface are the sharpest case: that is the doc surface for the two `resolveIssueAttributeSet` overloads, so a reader who trusts it would expect an unconfigured profile to call the authority.

src/test/java/com/otilm/core/integration/service/v2/UploadedCsrValidationITest.java
klaraf755: Fixed in `a6607873c`. Asserts `"the authority connector is unavailable"` — the reason string rather than the shared wrapper — plus `mockServer.verify(postRequestedFor(urlPathMatching("/v3/authorityProvider/certificates/request/attributes")))`.

Your read off the code was right, and the wrapper is at `ProtocolRequestAttributeValidator:60-66`: the two reasons differ only inside the parentheses, and `unavailable` sits in the outer sentence both paths share.

src/test/java/com/otilm/core/integration/service/CertificateRequestIntegrationITest.java
klaraf755: Fixed in `a6607873c`. Narrowed to `ConnectorServerException` — the type `schemaOrEmpty` rethrows for a non-501 5xx — with a `mockServer.verify` on the stubbed POST.

I mutation-checked it rather than just re-running: replacing the 500 stub with `okJson("[]")` now fails with `Expecting code to raise a throwable`, where the old `isInstanceOf(ConnectorException.class)` would have been satisfied by the 404 from the unstubbed route. So it now distinguishes fail-closed-on-500 from reached-nothing.

vyskocilm (approved):
Approving. All six findings verified against `2c3b2e2c`, not just read off the replies.

- Name-collision 500 — `mergeUnion` claims by name and UUID, connector first; `connectorWinsANameCollisionEvenWhenTheUuidsDiffer` pins it.
- Destructive `PATCH` — `valueSourceBindings` null-defaulted on the contract side, replace skipped when null; absent-preserves and empty-clears both covered. Your correction stands: this outlived the pre-merge window.
- WARN — moved after the merge, gated on `merged.isEmpty()`, with the silence pinned for the v2-backed steady state.
- Atomicity + unguarded `getValueSourceType()` — `@Transactional` on `updateConfiguration`, `ValidationException` naming the target attribute.
- Doc sweep — all five sites; `AttributeSetMergeMode#MERGE` greps clean.
- Both fail-closed assertions now name their path and verify the stub.

`collectionRef` removal (`2c3b2e2c`) is more than was asked but is the better answer than wiring or rejecting. Column drop is safe by the dating you gave: `V202606271400` shipped in 2.19.0/2.19.1, and that same release rejected every update carrying a non-empty binding list, so no row could exist. `V202608281000` collides with nothing open. No `collectionRef` reference survives in `src`.

`CONNECTOR_ONLY` against a v2 authority stays a WARN — agreed, on your reasoning: a save-time gate would enforce a v2/v3 distinction the adapter contract deliberately erases, and could not detect the v3 case that matters anyway.

One residual, not blocking. The dedupe lives in `mergeUnion`, so it only covers `MERGE`. `CONNECTOR_ONLY` returns `connectorDefs` unfiltered, so a connector schema serving two same-named definitions still reaches `Collectors.toMap(AttributeDefinition::getName, …)` (`AttributeEngine:756`) and 500s — that path is newly reachable, since 2.19.x rejected the mode at save time. `STATIC_ONLY` has the same shape but is pre-existing, and `validateRequestAttributeDefinitions` has no name-uniqueness rule to stop it. Both need malformed input to fire. Worth a follow-up issue rather than a change here.

Issue #2166, opened 2026-08-28
Offer curve class representatives in searchable fields once ingest canonicalization lands


karolhrdina:
### Description

core#2074's searchable-fields endpoint serves curve filter values as `SELECT DISTINCT` over the
normalized `curve` column. The acceptance criterion behind it — "the curve filter's values are the
197 curve class representatives, not the 246 raw producer tokens, so the UI cannot offer three
spellings of one curve" — is therefore delegated: today the column receives only case/whitespace/
Unicode-folded spellings, so two producer spellings of one physical curve (`p-256`, `secp256r1`)
are still offered as two values. The class-level canonicalization (union-find over the registry,
one `secg/*` representative per class, `P-256` as display alias only) is the #2072 ingest
pipeline's obligation.

When #2072's canonicalization lands, close the loop on the search surface:

- [ ] Verify end to end that the curve value list offers exactly one `secg/*` representative per
      equivalence class present in the data — one value per class, never one per producer spelling.
- [ ] Update `CryptoAssetListQueryITest.distinctValueQueriesReturnCanonicalSortedValuesWithoutNullsOrDuplicates`
      — it currently pins today's behavior (`p-256` and `secp256r1` as two distinct stored values)
      and must start failing the moment ingest folds them; rewrite the expectation to the folded
      representative.
- [ ] Move fixture vocabulary in the crypto-asset search/service/searchable-fields ITests to the
      ratified `secg/*` spellings so tests document the ratified vocabulary rather than producer
      aliases.
- [ ] Re-check the searchable-fields javadoc in `CryptoAssetRepository`/`CryptographicAssetServiceImpl`
      (this delegation is documented there) and drop the caveat once it no longer applies.

Blocked by #2072. Surfaced by the core#2074 pre-PR review.

---

## Added 2026-08-28 — two gaps found reviewing core#2167 and core#2161

- [ ] **Migrate or invalidate saved list-view filter values over the canonicalized columns.** `ListView`
      persists `filters` as JSONB, values included, and nothing validates a stored value against the
      catalogue. A view saved today with `CBOM_ASSET_CURVE EQUALS "p-256"` returns **zero rows forever**
      once ingest folds that spelling into a `secg/*` representative — no error, no signal to the operator
      who saved it. None of the checkboxes above covers a persisted artefact.
- [ ] **Name the layer the folding must land in.** The checkbox above assigns class canonicalization to
[...]

PR #2146, opened 2026-08-25
Persist the cryptographic asset inventory, with the identity key fenced

vyskocilm:
Closes #2071.

Persistence layer for the cryptographic asset inventory: Flyway migration, entities, repositories,
writer beans, search fields, and the identity-key leak fence. Out of scope and unchanged here: CBOM
parsing/ingest (the caller that fills these tables), the read surface (#2074), and the PQC rule set
itself. Rebased on `main@4e3eac4f` and built against interfaces `2.20.0-SNAPSHOT`, which now carries
the ratified contract.

## What lands

**Schema** — `V202608251000__crypto_asset_inventory.sql`

- `crypto_asset`: typed identity/filter columns (asset type, name, oid, algorithm family, primitive,
  parameter set, curve, mode, padding, variant), `identity_key` with a plain `UNIQUE` so it can serve
  as an `ON CONFLICT` arbiter, `ruleset_version`, merge bookkeeping (`properties_leaf_count`,
  `properties_hash`, `properties_source_uuid`, `source_count`), unindexed JSONB
  `merged_crypto_properties`, and the PQC verdict columns including the evaluated-fields snapshot.
- `crypto_asset_source`: the per-CBOM link, `UNIQUE (asset_uuid, cbom_uuid)`, retained
  `original_crypto_properties`, capped `evidence`, `occurrence_count`, first/last seen. FK CASCADE to
  the asset, RESTRICT to the cbom row, plus `idx_crypto_asset_source_cbom` — the composite unique
  index cannot serve the RESTRICT check or per-CBOM removal, both of which lead with `cbom_uuid`.
- `crypto_asset_alias`: `(absorbed_key, canonical_key, reason, decided_by, decided_at)`.
  `absorbed_key` carries no FK on purpose (the absorbed row is gone by the time the alias exists);
  `canonical_key` does, against the `identity_key` arbiter.
- `cbom_tombstone`, plus `asset_sync_state` / `asset_sync_error` / `assets_synced_at` on `cbom`.
- Btree indexes for every filter column plus `lower(name)` and `lower(oid)`. No GIN, no
  `CREATE EXTENSION`, no partitioning in v1.

**Persistence** — repositories extend `SecurityFilterRepository` and carry no `@Transactional`; every
modifying path goes through a `*Writer` bean, per the CLAUDE.md transactional-boundary rule. Identity
is computed by `CryptoAssetIdentityCalculator` from the typed fields only — it never reads aliases.
[...]

Issue #2145, opened 2026-08-25
Implement cryptographic asset detail, statistics and sync visibility endpoints


karolhrdina:
### Description

Controller implementations for the inventory detail, statistics and sync
visibility surface (ilm#299, design doc section 8) — second half of the
former single endpoints story. The list and searchable-fields half is
core#2074; this story builds on its controller, security-filter and fixture
plumbing, and its statistics must reconcile against that story's list.

- Detail: merged payload plus per-source originals, verdict block (verdict,
  rule id, reason, rule-set version, evaluated fields), source CBOM list
  linking to existing CBOM detail, capped evidence with true counts,
  document-internal references labeled as opaque. Detail shows the elected
  representative payload AND every source's original payload beside it —
  worth a test rather than a restatement: the elected view is one source's
  payload verbatim, so a sparse but crucial observation can lose the
  election to a richer generic one. A refuted OID surfaced under the search
  opt-in renders as refuted, per the flag on the detail DTO
  (interfaces#874).
- Statistics endpoint: count badges, statByAssetType /
  statByAlgorithmFamily (top-N plus overflow count) / statByPqcVerdict,
  and the sync-completeness block (synced / total / failed CBOMs) — built
  on countGroupedUsingSecurityFilter, virtual-thread parallel group-bys,
  pure calculator class. Gated by the CRYPTO_ASSET resource's LIST action,
  not AnyPrincipal.
- Sync-state fields exposed on the existing CBOM list/detail DTOs.
- The identity key is not part of the read surface: it appears in no DTO
  and no response body. The CI exposure check covers named fields,
  accessors and quoted literals; it CANNOT see a key travelling inside a
  map-typed payload. Four such payloads reach a client on this surface and
  each needs checking by hand: `electedPayload` and every
  `sources[].payload` on the asset detail, `verdict.evaluatedFields`, and
  `CbomDetailDto.content`, the existing case. `verdict.evaluatedFields` is
  written by core#2151, not here -- see the note on that issue: an
  allowlist on its write side closes the channel at the source, and this
  hand-check is the backstop, not the fix.
- `@ExternalAuthorization(resource = Resource.CRYPTO_ASSET, ...)` on the
  operations; audit logging per house rules. core#2075 was cancelled on
  2026-08-26 -- the catalogue sync and auditor-role classification happen
  automatically at startup, and the one piece of it that is real code, the
[...]

PR #2167, opened 2026-08-28
Serve the cryptographic asset inventory list and searchable fields

### LINKED ISSUE #2074: Implement cryptographic asset inventory search endpoints
### Description

Controller implementation for the inventory search surface (ilm#299, design
doc section 8) — the list and searchable-fields operations of the contract
defined in interfaces#874. First half of the former single endpoints story;
the detail, statistics and sync-visibility half is core#2145, which builds
on this story's controller, security-filter and fixture plumbing.

- List: every filter from epic AC5 (asset type, algorithm family,
  primitive, parameter set / key size, elliptic curve, PQC verdict, source
  CBOM), combinable, plus free-text as a case-insensitive OR predicate
  over lower(name) / lower(oid) — small new FilterPredicatesBuilder work,
  no platform precedent. Fixed deterministic ordering (name ASC, uuid
  ASC). A REAL runtime page-size clamp (the schema's 1000 maximum is
  documentation-only platform-wide today).
- Refuted OIDs are excluded from the DEFAULT free-text and OID predicates
  and reachable only through an explicit opt-in (a facet, or
  includeRefuted=true). The value stays stored and auditable — it is the
  only record that a producer is unreliable — but a default query must not
  return it: rendering a refuted flag after retrieval is too late, the
  answer is already wrong, and a compliance or remediation caller cannot
  afford one. Measured: 4 of the 84 distinct corpus OIDs are refuted, on 11
  rows, one of them answering an "AES" query with a post-quantum KEM. The
  `oid` column itself stays the verbatim producer value. (Rendering a
  refuted OID as refuted on the detail view is the companion story's scope,
  per the flag on the detail DTO in interfaces#874.)
- Searchable-fields endpoint with server-known enum values. The curve
  filter's values are the 197 curve class representatives, not the 246 raw
  producer tokens, so the UI cannot offer three spellings of one curve.
- Normalized identity and filter fields answer filter queries for every
  asset type wherever derivable — certificates, protocols and material
  included — not only for algorithms.
- The identity key is not part of the read surface: it appears in no DTO,
  no response body, no path parameter and not in the searchable-fields
  enum. The redaction ruling in core#2070 depends on this.
- `@ExternalAuthorization(resource = Resource.CRYPTO_ASSET, action =
  ResourceAction.LIST)` on the operations, and `@AuthEndpoint(resourceName =
  Resource.CRYPTO_ASSET)` on the list operation so the synced catalogue
  entry carries a listing endpoint; audit logging per house rules.
- Absorbed from the cancelled core#2075, and not optional: the core-local
  mirror entry `CRYPTO_ASSET("cryptoAssets")` in
  `com.otilm.core.model.auth.Resource`. On ContextRefreshedEvent,
  `ContextRefreshListener` maps every annotated resource back through
  `com.otilm.core.model.auth.Resource.findByCode`, which throws a
  ValidationException on an unknown code -- so the FIRST
  `@ExternalAuthorization` naming CRYPTO_ASSET anywhere in core stops the
  application from starting until that entry exists. core#2071 (PR
  core#2146) adds `Resource.CRYPTO_ASSET` uses in `FilterField` and
  `ResourceToClass`, but not that entry. Everything else core#2075 held --
  the auth-service catalogue sync and the auditor-role read classification
  -- is derived automatically at startup by `AuthResourceSynchronizer` from
  these annotations, which is why closing it on 2026-08-26 was sound.

### Acceptance Criteria

- [ ] Every AC5 filter works individually and combined; responses paged
      and clamped; verifiable standalone with ordinary fixtures (volume
      gates live in the volume-validation task).
- [ ] A default search whose only match is a refuted OID returns NOTHING
      for it; the same search with the opt-in returns the row.
- [ ] Filtering by a normalized field returns non-algorithm assets that
      carry it.
- [ ] Searchable-fields offers curve class representatives, not raw
      producer tokens; the identity key is not offered.
- [ ] No response on any operation contains the identity key.
- [ ] 403 without permission on all operations, server-side enforced, and
      the application starts with the annotations in place -- the resource
      sync resolves `cryptoAssets` instead of throwing.

karolhrdina:
Implements OmniTrustILM/core#2074: the list and searchable-fields operations of the cryptographic asset inventory contract (OmniTrustILM/interfaces#874), on top of the OmniTrustILM/core#2071 / OmniTrustILM/core#2146 data model. The last three commits answer the review on this PR; every inline thread carries its resolution.

## What it does

- **List** (`POST /v1/cryptoAssets`): every AC5 filter (asset type, algorithm family, primitive, parameter set, curve, PQC verdict, source CBOM), combinable, plus free-text as a case-insensitive OR over `lower(name)`/`lower(oid)` — LIKE wildcards escaped, the pattern bound (never inlined into statement text), multi-value payloads refused. Deterministic order by the **served label** (name, else unrefuted OID; the contract's "ordered by name ascending" names the served field) with the repository's uuid tie-break (OmniTrustILM/core#2157's `SortOrderBuilder`). Real runtime clamp at the schema's 1000; page offset guarded against int overflow (422, never a wrapped page).
- **Refuted OIDs** answer no default value predicate and no free-text query — the row stays findable by name, the value stays stored and auditable (EMPTY/NOT_EMPTY untouched), an explicit facet filter opts the request in, and the facet refuses unadvertised conditions and non-scalar/non-boolean values (422, never an inverted predicate). A refuted OID is also never served as a display name — not in the list, the object picker, or the sort key.
- **`quarantined`** fires only for `REFUTED_CERTIFICATE_DIGEST` — the one guard the contract sentence ("sources make contradicting claims … quarantined pending reconciliation") describes. The other guards are permanent-by-design splits; flagging them would advertise a queue that can never drain.
- **Searchable fields** (`GET /v1/cryptoAssets/search`): 17 fields, PROPERTY group only, value lists as recursive-CTE loose index scans over the existing btrees (milliseconds instead of eight parallel seq scans per filter-panel open); the source-CBOM list is scoped by the caller's own `cboms:list` object access — the platform's first cross-resource value list is not served platform-wide. Identity key in no field, no value list, no response: fence arch tests plus wire-level assertions including a recomputed exact-key check.
- **Counting**: `totalItems` uses a plain-count variant — `count(DISTINCT)` is unparallelizable and measured seconds-vs-milliseconds at 5M rows; a test pins plain == distinct through the only predicate shape that could differ.
- **Auth**: `@ExternalAuthorization(CRYPTO_ASSET, LIST)` on both operations (403 tested over HTTP), `@AuthEndpoint` + `ResourceExtensionService` bean for the role editor's object picker — which honours its filters. Detail/statistics stay 501 (OmniTrustILM/core#2145's scope).

## Review resolution map

- 🔴 quarantined contract mismatch, sort-vs-served-name, offset overflow, refuted-OID-as-name, picker filters, fence-test hardening → commit `eadc62c2`.
- 🔴 facet array-value inversion (+ carve-out disarm compound), free-text multi-value, literal-vs-bound pattern, migration-premise javadoc, `%` positive control, untested carve-out branches → commit `ef9177a6`.
- 🟡 `count(DISTINCT)` and the eight unbounded `SELECT DISTINCT` value scans → commit `da770563`.
- 🟡 cross-resource CBOM serial list → scoped in `eadc62c2`; **size** remains unbounded for a permitted caller — a cap is a contract change, sign-off still requested.
- 🟡 `hasObjectAccess` → OmniTrustILM/interfaces#924 (lock-step release is its AC). `@AuthEndpoint` stays: the ticket text mandates the synced listing endpoint; until the flag ships, `/v1/resources` errs on the hidden side.
- Both-null / refuted-nameless `name` residual → OmniTrustILM/interfaces#925 (nullable name, guaranteed fallback, or per-row refuted flag — one to ratify).
- Deep-offset paging → OmniTrustILM/core#2169 (keyset story, with the review's measurements).
- Curve class representatives → deferral adjudicated on the thread; OmniTrustILM/core#2166 extended (folding landing spot, saved-view JSONB rot, re-key consequence). **Asking the ticket owner to move that AC checkbox from OmniTrustILM/core#2074 to OmniTrustILM/core#2166 so #2074 can close honestly.**
- OmniTrustILM/core#2165 occurrence-triple coupling → noted on that ticket (`occurrenceCount` is now client-facing).
- **Pushback, one item**: withholding `CBOM_ASSET_RULESET_VERSION` until the stored-int-vs-string ruling — OmniTrustILM/core#2071's AC reads "ruleset_version is written on every insert and update; **a row keyed under an older ruleset is findable by query**", so the field stays served. The R3 ripple (NUMBER→STRING would change the advertised operator set and rot saved numeric conditions) stands as this PR's known coupling.
- Sonar's clustered `S5778` lambdas restructured (single throwing invocation per assertion) across the flagged tests.

## Review notes that remain decisions

1. **Migration comment drift:** `V202608271000`'s "never emits lower()" premise is superseded — correction now lives in the free-text predicate's javadoc (with the review's prefix-LIKE finding: a future index story needs trigram GIN for infix AND `text_pattern_ops` for prefix). The applied file stays untouched (Flyway checksum).
2. **Sort refused with 422 by ratified v1 design**; OmniTrustILM/core#2157 landed the mechanism with no service wired yet. Related platform trap (unchanged here): OmniTrustILM/core#2158 saved views validate sort by catalogue membership only, so a view can persist a sort this endpoint refuses — needs a platform decision (view-save learns sortability, or listings accept sort).
3. **Serial value list size** (see resolution map) — sign-off requested.
4. **`ruleset_version` NUMBER filter** — kept per OmniTrustILM/core#2071 AC; R3 ripple documented.

## Validation

- All six changed/new integration suites green at final HEAD, including the review-driven regression tests (facet inversion compound, served-name ordering, per-guard quarantined, offset wrap, scoped serial list over HTTP both ways, exact-key wire fence).
- Full `-Ptest-non-integration` group at final HEAD: green apart from the two known environmental `ProxyPropertiesTest` hostname failures on the dev machine (`mezeq` unresolvable).
- spotless + checkstyle clean; ContextSignatureGuard baseline unchanged; NativeQuerySchemaArchTest green over the new native value queries.
- CI on the PR: all checks green including Sonar (96.6% new-code coverage before this round; gate re-runs on push).

Link: OmniTrustILM/core#2074

### DIFF
--- a/src/main/java/com/otilm/core/api/web/CryptographicAssetControllerImpl.java
+++ b/src/main/java/com/otilm/core/api/web/CryptographicAssetControllerImpl.java
@@ -11,6 +11,7 @@ import com.otilm.api.model.core.logging.enums.Module;
 import com.otilm.api.model.core.logging.enums.Operation;
 import com.otilm.api.model.core.search.SearchFieldDataByGroupDto;
 import com.otilm.core.aop.AuditLogged;
+import com.otilm.core.auth.AuthEndpoint;
 import com.otilm.core.logging.LogResource;
 import com.otilm.core.security.authz.SecuredUUID;
 import com.otilm.core.security.authz.SecurityFilter;
@@ -32,6 +33,7 @@ public class CryptographicAssetControllerImpl implements CryptographicAssetContr
 
     @Override
     @AuditLogged(module = Module.CORE, resource = Resource.CRYPTO_ASSET, operation = Operation.LIST)
+    @AuthEndpoint(resourceName = Resource.CRYPTO_ASSET)
     public PaginationResponseDto<CryptographicAssetDto> listCryptographicAssets(SearchRequestDto request) {
         return cryptographicAssetService.listCryptographicAssets(SecurityFilter.create(), request);
     }
--- a/src/main/java/com/otilm/core/dao/repository/CbomRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/CbomRepository.java
@@ -59,4 +59,8 @@ public interface CbomRepository extends SecurityFilterRepository<Cbom, UUID> {
             """)
     int updateAssetSyncState(@Param("uuid") UUID uuid, @Param("state") CbomAssetSyncState state,
             @Param("error") String error, @Param("syncedAt") OffsetDateTime syncedAt);
+
+    /** Distinct CBOM serial numbers, the value list of the inventory's source-CBOM filter. */
+    @Query("SELECT DISTINCT c.serialNumber FROM Cbom c ORDER BY c.serialNumber")
+    List<String> findDistinctSerialNumber();
 }
--- a/src/main/java/com/otilm/core/dao/repository/CbomRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/CbomRepository.java
@@ -59,8 +59,4 @@ public interface CbomRepository extends SecurityFilterRepository<Cbom, UUID> {
             """)
     int updateAssetSyncState(@Param("uuid") UUID uuid, @Param("state") CbomAssetSyncState state,
             @Param("error") String error, @Param("syncedAt") OffsetDateTime syncedAt);
-
-    /** Distinct CBOM serial numbers, the value list of the inventory's source-CBOM filter. */
-    @Query("SELECT DISTINCT c.serialNumber FROM Cbom c ORDER BY c.serialNumber")
-    List<String> findDistinctSerialNumber();
 }
--- a/src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
@@ -182,27 +182,89 @@ public interface CryptoAssetRepository extends SecurityFilterRepository<CryptoAs
      * token. Class-level canonicalization -- folding P-256 and secp256r1 into one secg/* representative -- is the
      * ingest pipeline's obligation (core#2072): these lists offer exactly what that pipeline stores, one entry per
      * stored spelling, and become the ratified class representatives the moment it writes them.
+     *
+     * <p>
+     * Native recursive CTEs -- a loose index scan. Postgres has no btree skip scan, so {@code SELECT DISTINCT} walks
+     * the whole table (a parallel seq scan pinning workers): hundreds of milliseconds per column at millions of rows,
+     * for eight columns on every filter-panel open. The CTE instead hops index-min to index-min over the per-column
+     * btrees the migration already ships -- O(distinct values x log rows), reliably milliseconds. {@code min()} ignores
+     * NULLs, and the strictly-greater walk makes the values distinct and sorted by construction.
      */
-    @Query("SELECT DISTINCT a.algorithmFamily FROM CryptoAsset a WHERE a.algorithmFamily IS NOT NULL "
-            + "ORDER BY a.algorithmFamily")
+    @Query(value = """
+            WITH RECURSIVE vals AS (
+                SELECT min(algorithm_family) AS v FROM {h-schema}crypto_asset
+                UNION ALL
+                SELECT (SELECT min(algorithm_family) FROM {h-schema}crypto_asset WHERE algorithm_family > vals.v)
+                FROM vals WHERE vals.v IS NOT NULL
+            )
+            SELECT v FROM vals WHERE v IS NOT NULL ORDER BY v
+            """, nativeQuery = true)
     List<String> findDistinctAlgorithmFamily();
 
-    @Query("SELECT DISTINCT a.primitive FROM CryptoAsset a WHERE a.primitive IS NOT NULL ORDER BY a.primitive")
+    @Query(value = """
+            WITH RECURSIVE vals AS (
+                SELECT min(primitive) AS v FROM {h-schema}crypto_asset
+                UNION ALL
+                SELECT (SELECT min(primitive) FROM {h-schema}crypto_asset WHERE primitive > vals.v)
+                FROM vals WHERE vals.v IS NOT NULL
+            )
+            SELECT v FROM vals WHERE v IS NOT NULL ORDER BY v
+            """, nativeQuery = true)
     List<String> findDistinctPrimitive();
 
-    @Query("SELECT DISTINCT a.parameterSet FROM CryptoAsset a WHERE a.parameterSet IS NOT NULL ORDER BY a.parameterSet")
+    @Query(value = """
+            WITH RECURSIVE vals AS (
+                SELECT min(parameter_set) AS v FROM {h-schema}crypto_asset
+                UNION ALL
+                SELECT (SELECT min(parameter_set) FROM {h-schema}crypto_asset WHERE parameter_set > vals.v)
+                FROM vals WHERE vals.v IS NOT NULL
+            )
+            SELECT v FROM vals WHERE v IS NOT NULL ORDER BY v
+            """, nativeQuery = true)
     List<String> findDistinctParameterSet();
 
-    @Query("SELECT DISTINCT a.curve FROM CryptoAsset a WHERE a.curve IS NOT NULL ORDER BY a.curve")
+    @Query(value = """
+            WITH RECURSIVE vals AS (
+                SELECT min(curve) AS v FROM {h-schema}crypto_asset
+                UNION ALL
+                SELECT (SELECT min(curve) FROM {h-schema}crypto_asset WHERE curve > vals.v)
+                FROM vals WHERE vals.v IS NOT NULL
+            )
+            SELECT v FROM vals WHERE v IS NOT NULL ORDER BY v
+            """, nativeQuery = true)
     List<String> findDistinctCurve();
 
-    @Query("SELECT DISTINCT a.mode FROM CryptoAsset a WHERE a.mode IS NOT NULL ORDER BY a.mode")
+    @Query(value = """
+            WITH RECURSIVE vals AS (
+                SELECT min(mode) AS v FROM {h-schema}crypto_asset
+                UNION ALL
+                SELECT (SELECT min(mode) FROM {h-schema}crypto_asset WHERE mode > vals.v)
+                FROM vals WHERE vals.v IS NOT NULL
+            )
+            SELECT v FROM vals WHERE v IS NOT NULL ORDER BY v
+            """, nativeQuery = true)
     List<String> findDistinctMode();
 
-    @Query("SELECT DISTINCT a.padding FROM CryptoAsset a WHERE a.padding IS NOT NULL ORDER BY a.padding")
+    @Query(value = """
+            WITH RECURSIVE vals AS (
+                SELECT min(padding) AS v FROM {h-schema}crypto_asset
+                UNION ALL
+                SELECT (SELECT min(padding) FROM {h-schema}crypto_asset WHERE padding > vals.v)
+                FROM vals WHERE vals.v IS NOT NULL
+            )
+            SELECT v FROM vals WHERE v IS NOT NULL ORDER BY v
+            """, nativeQuery = true)
     List<String> findDistinctPadding();
 
-    @Query("SELECT DISTINCT a.variant FROM CryptoAsset a WHERE a.variant IS NOT NULL ORDER BY a.variant")
+    @Query(value = """
+            WITH RECURSIVE vals AS (
+                SELECT min(variant) AS v FROM {h-schema}crypto_asset
+                UNION ALL
+                SELECT (SELECT min(variant) FROM {h-schema}crypto_asset WHERE variant > vals.v)
+                FROM vals WHERE vals.v IS NOT NULL
+            )
+            SELECT v FROM vals WHERE v IS NOT NULL ORDER BY v
+            """, nativeQuery = true)
     List<String> findDistinctVariant();
 
     /**
--- a/src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
@@ -2,6 +2,8 @@ package com.otilm.core.dao.repository.cbom;
 
 import com.otilm.core.dao.entity.cbom.CryptoAsset;
 import com.otilm.core.dao.repository.SecurityFilterRepository;
+import com.otilm.core.model.cbom.CryptoAssetListRow;
+import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
@@ -173,4 +175,46 @@ public interface CryptoAssetRepository extends SecurityFilterRepository<CryptoAs
     @Modifying
     @Query("DELETE FROM CryptoAsset a WHERE a.uuid = :uuid")
     int deleteAsset(@Param("uuid") UUID uuid);
+
+    /**
+     * Distinct stored values of one normalized filter column, for the searchable-fields value lists. The columns hold
+     * the canonical normalized spelling, so these lists can never offer two producer spellings of one value.
+     */
+    @Query("SELECT DISTINCT a.algorithmFamily FROM CryptoAsset a WHERE a.algorithmFamily IS NOT NULL "
+            + "ORDER BY a.algorithmFamily")
+    List<String> findDistinctAlgorithmFamily();
+
+    @Query("SELECT DISTINCT a.primitive FROM CryptoAsset a WHERE a.primitive IS NOT NULL ORDER BY a.primitive")
+    List<String> findDistinctPrimitive();
+
+    @Query("SELECT DISTINCT a.parameterSet FROM CryptoAsset a WHERE a.parameterSet IS NOT NULL ORDER BY a.parameterSet")
+    List<String> findDistinctParameterSet();
+
+    @Query("SELECT DISTINCT a.curve FROM CryptoAsset a WHERE a.curve IS NOT NULL ORDER BY a.curve")
+    List<String> findDistinctCurve();
+
+    @Query("SELECT DISTINCT a.mode FROM CryptoAsset a WHERE a.mode IS NOT NULL ORDER BY a.mode")
+    List<String> findDistinctMode();
+
+    @Query("SELECT DISTINCT a.padding FROM CryptoAsset a WHERE a.padding IS NOT NULL ORDER BY a.padding")
+    List<String> findDistinctPadding();
+
+    @Query("SELECT DISTINCT a.variant FROM CryptoAsset a WHERE a.variant IS NOT NULL ORDER BY a.variant")
+    List<String> findDistinctVariant();
+
+    /**
+     * List-page rows for the given assets. A projection rather than the entity: the list serves none of the JSONB
+     * payload columns, and a page can be 1000 rows. The occurrence total is summed over the per-source rows, whose
+     * count is deliberately uncapped (capping drops evidence payloads, never the count). Rows come back in no
+     * particular order -- IN provides none -- so the caller restores its page order.
+     */
+    @Query("""
+            SELECT new com.otilm.core.model.cbom.CryptoAssetListRow(a.uuid, a.name, a.assetType,
+                    a.pqcVerdict, a.sourceCount, a.identityGuard, COALESCE(SUM(s.occurrenceCount), 0L))
+            FROM CryptoAsset a
+            LEFT JOIN CryptoAssetSource s ON s.assetUuid = a.uuid
+            WHERE a.uuid IN :uuids
+            GROUP BY a.uuid, a.name, a.assetType, a.pqcVerdict, a.sourceCount, a.identityGuard
+            """)
+    List<CryptoAssetListRow> findListRowsByUuids(@Param("uuids") Collection<UUID> uuids);
 }
--- a/src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
@@ -178,7 +178,10 @@ public interface CryptoAssetRepository extends SecurityFilterRepository<CryptoAs
 
     /**
      * Distinct stored values of one normalized filter column, for the searchable-fields value lists. The columns hold
-     * the canonical normalized spelling, so these lists can never offer two producer spellings of one value.
+     * the stored normalized spelling (case/whitespace/Unicode-folded), so the lists collapse casing variants of one
+     * token. Class-level canonicalization -- folding P-256 and secp256r1 into one secg/* representative -- is the
+     * ingest pipeline's obligation (core#2072): these lists offer exactly what that pipeline stores, one entry per
+     * stored spelling, and become the ratified class representatives the moment it writes them.
      */
     @Query("SELECT DISTINCT a.algorithmFamily FROM CryptoAsset a WHERE a.algorithmFamily IS NOT NULL "
             + "ORDER BY a.algorithmFamily")
@@ -209,12 +212,12 @@ public interface CryptoAssetRepository extends SecurityFilterRepository<CryptoAs
      * particular order -- IN provides none -- so the caller restores its page order.
      */
     @Query("""
-            SELECT new com.otilm.core.model.cbom.CryptoAssetListRow(a.uuid, a.name, a.assetType,
+            SELECT new com.otilm.core.model.cbom.CryptoAssetListRow(a.uuid, a.name, a.oid, a.assetType,
                     a.pqcVerdict, a.sourceCount, a.identityGuard, COALESCE(SUM(s.occurrenceCount), 0L))
             FROM CryptoAsset a
             LEFT JOIN CryptoAssetSource s ON s.assetUuid = a.uuid
             WHERE a.uuid IN :uuids
-            GROUP BY a.uuid, a.name, a.assetType, a.pqcVerdict, a.sourceCount, a.identityGuard
+            GROUP BY a.uuid, a.name, a.oid, a.assetType, a.pqcVerdict, a.sourceCount, a.identityGuard
             """)
     List<CryptoAssetListRow> findListRowsByUuids(@Param("uuids") Collection<UUID> uuids);
 }
--- a/src/main/java/com/otilm/core/enums/FilterField.java
+++ b/src/main/java/com/otilm/core/enums/FilterField.java
@@ -80,6 +80,7 @@ import com.otilm.core.dao.entity.signing.SigningRecord_;
 import com.otilm.core.dao.entity.signing.TimeQualityConfiguration_;
 import com.otilm.core.dao.entity.signing.TspProfile_;
 import com.otilm.core.model.auth.ResourceAction;
+import com.otilm.core.model.cbom.CryptoAssetIdentityGuard;
 import jakarta.persistence.metamodel.Attribute;
 import java.util.Arrays;
 import java.util.List;
@@ -414,6 +415,17 @@ public enum FilterField {
             "Identity Rule Set Version", SearchFieldTypeEnum.NUMBER),
     CBOM_ASSET_SOURCE_COUNT(Resource.CRYPTO_ASSET, null, null, CryptoAsset_.sourceCount, "Source CBOMs",
             SearchFieldTypeEnum.NUMBER),
+
+    // Free text spans lower(name)/lower(oid) and the refuted-OID opt-in couples with the OID
+    // predicates, so all three are built by dedicated branches in FilterPredicatesBuilder rather
+    // than the generic attribute path. CBOM_ASSET_SOURCE_CBOM matches through an EXISTS subquery,
+    // never a join: the uuid page query has no DISTINCT, and a row with several sources would
+    // otherwise repeat inside one page.
+    CBOM_ASSET_FREE_TEXT(Resource.CRYPTO_ASSET, null, null, null, "Text Search", SearchFieldTypeEnum.FREE_TEXT),
+    CBOM_ASSET_OID_REFUTED(Resource.CRYPTO_ASSET, null, null, CryptoAsset_.identityGuard, "OID Refuted",
+            SearchFieldTypeEnum.BOOLEAN, null, CryptoAssetIdentityGuard.REFUTED_OID, false, null),
+    CBOM_ASSET_SOURCE_CBOM(Resource.CRYPTO_ASSET, Resource.CBOM, null, Cbom_.serialNumber, "Source CBOM",
+            SearchFieldTypeEnum.LIST),
 
     // Signing Profile
     SIGNING_PROFILE_NAME(Resource.SIGNING_PROFILE, null, null, SigningProfile_.name, "Name",
--- a/src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
@@ -116,9 +116,10 @@ public class CryptographicAssetServiceImpl implements CryptographicAssetExternal
 
     /**
      * The group is PROPERTY only: crypto assets carry no attribute-engine attributes, so offering CUSTOM/META groups
-     * would advertise filter sources this resource does not serve. Curve values are canonical class representatives by
-     * construction -- the column stores the normalized spelling. The enum-backed and boolean fields get their values
-     * from {@link SearchHelper} (enum codes, or none).
+     * would advertise filter sources this resource does not serve. Curve values are the distinct stored spellings of
+     * the normalized column -- one entry per stored token; they become the ratified class representatives once
+     * core#2072's ingest canonicalization writes them (see the value-list queries' javadoc). The enum-backed and
+     * boolean fields get their values from {@link SearchHelper} (enum codes, or none).
      */
     @Override
     @ExternalAuthorization(resource = Resource.CRYPTO_ASSET, action = ResourceAction.LIST)
@@ -239,7 +240,10 @@ public class CryptographicAssetServiceImpl implements CryptographicAssetExternal
     private static CryptographicAssetDto toDto(CryptoAssetListRow row) {
         CryptographicAssetDto dto = new CryptographicAssetDto();
         dto.setUuid(row.uuid());
-        dto.setName(row.name());
+        // The contract marks name REQUIRED and the wire mapper drops nulls, so a nameless producer row serves its
+        // recorded OID -- the same next-best stable label displayLabel gives the object picker. A row with neither
+        // still serializes without the field; that residual is interfaces' contract friction, raised on the PR.
+        dto.setName(row.name() != null ? row.name() : row.oid());
         dto.setType(row.assetType());
         dto.setSourceCbomCount(row.sourceCount());
         dto.setOccurrenceCount(row.occurrenceCount());
--- a/src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
@@ -2,43 +2,104 @@ package com.otilm.core.service.impl;
 
 import com.otilm.api.exception.NotFoundException;
 import com.otilm.api.exception.NotSupportedException;
+import com.otilm.api.exception.ValidationException;
 import com.otilm.api.model.client.certificate.SearchRequestDto;
 import com.otilm.api.model.client.dashboard.CryptographicAssetStatisticsDto;
 import com.otilm.api.model.common.PaginationResponseDto;
 import com.otilm.api.model.core.auth.Resource;
 import com.otilm.api.model.core.cryptoasset.CryptographicAssetDetailDto;
 import com.otilm.api.model.core.cryptoasset.CryptographicAssetDto;
+import com.otilm.api.model.core.cryptoasset.PqcVerdict;
+import com.otilm.api.model.core.search.FilterFieldSource;
 import com.otilm.api.model.core.search.SearchFieldDataByGroupDto;
+import com.otilm.api.model.core.search.SearchFieldDataDto;
+import com.otilm.core.comparator.SearchFieldDataComparator;
+import com.otilm.core.dao.entity.cbom.CryptoAsset;
+import com.otilm.core.dao.entity.cbom.CryptoAsset_;
+import com.otilm.core.dao.repository.CbomRepository;
+import com.otilm.core.dao.repository.cbom.CryptoAssetRepository;
+import com.otilm.core.enums.FilterField;
 import com.otilm.core.model.auth.ResourceAction;
+import com.otilm.core.model.cbom.CryptoAssetListRow;
 import com.otilm.core.security.authz.ExternalAuthorization;
 import com.otilm.core.security.authz.SecuredUUID;
 import com.otilm.core.security.authz.SecurityFilter;
 import com.otilm.core.service.CryptographicAssetExternalService;
+import com.otilm.core.util.FilterPredicatesBuilder;
+import com.otilm.core.util.RequestValidatorHelper;
+import com.otilm.core.util.SearchHelper;
+import jakarta.persistence.criteria.CriteriaBuilder;
+import jakarta.persistence.criteria.CriteriaQuery;
+import jakarta.persistence.criteria.Predicate;
+import jakarta.persistence.criteria.Root;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.UUID;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import org.apache.commons.lang3.function.TriFunction;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.data.domain.PageRequest;
+import org.springframework.data.domain.Pageable;
 import org.springframework.stereotype.Service;
 
 /**
- * Serves the ratified cryptographic asset inventory contract with a refusal on every operation: the inventory
- * projection over stored CBOM documents does not exist yet, so there is nothing to read. The authorization annotations
- * are the working part — they register {@code cryptoAssets:list} and {@code cryptoAssets:detail} with the auth service
- * on context refresh, which is what lets roles be defined and the frontend's generated client be wired against the real
- * gates while the read model is built.
+ * Serves the ratified cryptographic asset inventory contract. List and searchable-fields are real reads over the
+ * deduplicated cross-CBOM asset projection; detail, statistics and sync visibility remain {@link NotSupportedException}
+ * refusals until core#2145 builds them.
  *
  * <p>
- * {@link NotSupportedException} maps to HTTP 501 in {@code ExceptionHandlingAdvice}, so a permitted caller learns the
- * operation is not implemented while an unpermitted one is still refused with 403 by the authorization aspect ahead of
- * this body.
+ * {@link NotSupportedException} maps to HTTP 501 in {@code ExceptionHandlingAdvice}, so a permitted caller learns a
+ * still-refused operation is not implemented while an unpermitted caller is refused with 403 by the authorization
+ * aspect ahead of this body.
  */
 @Service
 public class CryptographicAssetServiceImpl implements CryptographicAssetExternalService {
 
     private static final String NOT_IMPLEMENTED = "Cryptographic asset inventory is not implemented yet";
 
+    // The OpenAPI schema documents 1000 as the maximum items per page; this is where it is actually enforced.
+    private static final int MAX_ITEMS_PER_PAGE = 1000;
+
+    private CryptoAssetRepository cryptoAssetRepository;
+
+    private CbomRepository cbomRepository;
+
+    @Autowired
+    public void setCryptoAssetRepository(CryptoAssetRepository cryptoAssetRepository) {
+        this.cryptoAssetRepository = cryptoAssetRepository;
+    }
+
+    @Autowired
+    public void setCbomRepository(CbomRepository cbomRepository) {
+        this.cbomRepository = cbomRepository;
+    }
+
     @Override
     @ExternalAuthorization(resource = Resource.CRYPTO_ASSET, action = ResourceAction.LIST)
     public PaginationResponseDto<CryptographicAssetDto> listCryptographicAssets(SecurityFilter filter,
             SearchRequestDto request) {
-        throw new NotSupportedException(NOT_IMPLEMENTED);
+        RequestValidatorHelper.revalidateSearchRequestDto(request);
+        validatePaging(request);
+        TriFunction<Root<CryptoAsset>, CriteriaBuilder, CriteriaQuery<?>, Predicate> where = (root, cb,
+                criteriaQuery) -> FilterPredicatesBuilder
+                        .getFiltersPredicate(cb, criteriaQuery, root, request.getFilters());
+        Pageable page = PageRequest.of(request.getPageNumber() - 1, request.getItemsPerPage());
+        // The documented order is name ASC, uuid ASC. Only the name term is spelled here: the repository appends
+        // the ascending-uuid tiebreak to every paged secured query (SortOrderBuilder), which is what keeps page
+        // boundaries deterministic inside a group of equal names.
+        List<UUID> pageUuids = cryptoAssetRepository
+                .findUuidsUsingSecurityFilter(filter, where, page, (root, cb) -> cb.asc(root.get(CryptoAsset_.name)));
+        long totalItems = cryptoAssetRepository.countUsingSecurityFilter(filter, where);
+        PaginationResponseDto<CryptographicAssetDto> response = new PaginationResponseDto<>();
+        response.setItems(loadPage(pageUuids));
+        response.setItemsPerPage(request.getItemsPerPage());
+        response.setPageNumber(request.getPageNumber());
+        response.setTotalItems(totalItems);
+        response.setTotalPages((int) Math.ceil((double) totalItems / request.getItemsPerPage()));
+        return response;
     }
 
     @Override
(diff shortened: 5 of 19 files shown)


src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -47,15 +114,145 @@ public CryptographicAssetDetailDto getCryptographicAsset(SecuredUUID uuid) throw
...
+        dto.setName(row.name() != null ? row.name() : row.oid());
+        dto.setType(row.assetType());
+        dto.setSourceCbomCount(row.sourceCount());
+        dto.setOccurrenceCount(row.occurrenceCount());
+        // A never-evaluated asset is served as UNKNOWN, the ratified "the rules cannot classify this asset"
+        // verdict, because the contract marks the field required.
+        dto.setPqcVerdict(row.pqcVerdict() == null ? PqcVerdict.UNKNOWN : row.pqcVerdict());
+        // The guard records that a safety rule kept contradicting or ambiguous producer claims apart, which is
+        // exactly the contract's "contradicting claims quarantined pending reconciliation".
+        dto.setQuarantined(row.identityGuard() != null);
```
vyskocilm: L255: 🔴 bug: `quarantined` fires for all three guard values, but the contract promises one of them.

The contract text is `CryptographicAssetDto.quarantined` — "True when sources make contradicting claims about this asset that are quarantined pending reconciliation". Only `REFUTED_CERTIFICATE_DIGEST` is that: contradicting claims, with a reconciliation queue behind it.

`BARE_CN_SUBJECT` is not. It is a precautionary DN-scope split with no contradiction and, by the ratified addendum, no reconciliation path at all — a CN-only row may never be hand-merged into a full-DN row, so the duplicates are permanent by design and explicitly unbounded in an estate mixing a bare-CN producer with a full-DN one. A customer running one bare-CN scanner would see every certificate from it badged "quarantined pending reconciliation": an operator queue that can never drain.

`REFUTED_OID` is one component's OID contradicting its own other properties — not "sources" plural — and has no reconciliation path either.

Fix: `row.identityGuard() == CryptoAssetIdentityGuard.REFUTED_CERTIFICATE_DIGEST`. Anything wider needs the flag re-worded in interfaces ("a safety rule keeps this asset a separate row"), not widened silently here. The comment on L253-254 asserts the equivalence and should go with it.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -2,43 +2,110 @@
...
+        validatePaging(request);
+        TriFunction<Root<CryptoAsset>, CriteriaBuilder, CriteriaQuery<?>, Predicate> where = (root, cb,
+                criteriaQuery) -> FilterPredicatesBuilder
+                        .getFiltersPredicate(cb, criteriaQuery, root, request.getFilters());
+        Pageable page = PageRequest.of(request.getPageNumber() - 1, request.getItemsPerPage());
+        // The documented order is name ASC, uuid ASC. Only the name term is spelled here: the repository appends
+        // the ascending-uuid tiebreak to every paged secured query (SortOrderBuilder), which is what keeps page
+        // boundaries deterministic inside a group of equal names.
+        List<UUID> pageUuids = cryptoAssetRepository
+                .findUuidsUsingSecurityFilter(filter, where, page, (root, cb) -> cb.asc(root.get(CryptoAsset_.name)));
```
vyskocilm: L100: 🔴 bug: the list sorts on the `name` column but serves `coalesce(name, oid)`. The two disagree.

The contract says rows are ordered by name ascending then UUID ascending (`interfaces` `CryptographicAssetController` L48), and the only `name` the contract defines is the served DTO field. Postgres puts NULL `name` last under ASC, so a nameless row — whose served `name` is its OID — lands after every named row regardless of that OID's value. Wire-visible result: `["aes", "1.2.840.10045.4.3.2"]`, i.e. descending.

Fix: sort by the same coalesce `displayLabel` already builds at L236-238; the uuid tiebreak still gets appended. Current tests miss this because both fixture OIDs happen to sort after the fixture names.

vyskocilm (commented):
Independent review of this branch — four passes: query-layer correctness and security, contract conformance against the merged interfaces artifact, an architecture/convention gate, and an adversarial pass that tried to refute each of your eight review notes and to break the new tests.

## What ran

- Full `-Ptest-non-integration` at head `6f865de7`: **2933 tests, 0 failures, 0 errors** — your count exactly. `ProxyPropertiesTest` passed 5/5 here, which corroborates your 2 errors as environmental.
- All six changed/new integration suites green (43 tests), against a live PostgreSQL 16.14.
- 40 arch/guard tests green, including `ContextSignatureGuardTest` (baseline holds at 61 — unchanged, and correctly so: the new HTTP test reuses the `BaseSpringBootTest` + bare `@AutoConfigureMockMvc` signature three suites already have) and `IdentityKeyExposureFenceArchTest` plus its self-test.
- spotless and checkstyle clean.
- Query plans measured on synthetic `crypto_asset` tables at 2M and 5M rows carrying this migration's exact indexes.

## What did not run

Testcontainers is broken on this machine, so the suites above ran against an external Postgres rather than the pinned 17-alpine. No version-sensitive behaviour is in play in these suites, but it is a parity caveat. Sonar's new-code coverage gate was not computed.

## Checked and clean

Identity-key non-exposure is airtight on both operations, and not only via the fence: the projection enumerates its eight columns explicitly, `CryptoAsset` carries no `@ToString`, the service logs nothing, and `PaginationResponseDto` is not `Loggable`, so the audit aspect cannot reach it. The refuted-OID carve-out classification is complete and correct against real NULL semantics — including `NOT_MATCHES`, where `textregexeq(NULL, p) = false` is never true, so its placement in the "never satisfies" set is right. The EXISTS-over-join reasoning holds: the uuid page query really does emit no DISTINCT. LIKE escaping was executed against real Postgres for `_`, `%`, a lone backslash and the empty string, and behaves exactly as your tests claim. Count, page and projection cannot structurally disagree. No raw DB exception can reach the wire. Authorization is right on both operations, and no policy bundle, permission seed or auth migration is present — correct: the annotations are the whole registration.

Your eight notes: **2, 4, 5, 7 confirmed** (7 reproduced against the project's own wire mapper). **1, 3, 6 confirmed but incomplete.** **8 is the one I would push back on** — see the comment on `FilterPredicatesBuilder` L103.

**16 inline comments below** — three I would fix before merge, six worth your call, the rest small.

## Findings with no line in this diff to sit on

- **`CryptographicAssetControllerITest` L58-67.** `theResourceIsRegisteredWithItsReadActions` asserts actions only. The one auth-sync behaviour this PR actually adds is the `@AuthEndpoint` listing endpoint, and nothing asserts it. One line pins it: `assertThat(synced.getListObjectsEndpoint()).isEqualTo("/v1/cryptoAssets")`.
- **`FilterField` L414-415 (note 6).** Offering the identity ruleset stamp as a NUMBER is a *new* wire commitment, not a restatement — the merged contract commits `int` only for the PQC stamp. If the stored-int-vs-string question resolves to string it costs both a Flyway migration and a wire change: NUMBER→STRING flips the advertised operator set from GREATER/LESSER to CONTAINS/STARTS_WITH, and rots any saved view holding a numeric condition on the field. Your note names only the contract ripple. Cheap while the tables are empty — or the field could simply be withheld until the ruling lands, since no AC asks for it.
- **Deep offset paging.** Page 4000 (`OFFSET 3,999,000 LIMIT 1000`) measured 3,602 ms at 5M rows. Platform-wide trait rather than this PR's doing, but this is the resource carrying the millions-of-rows mandate; keyset pagination on `(name, uuid)` is the eventual answer.
- **core#2166 is missing a checklist item.** Saved list views persist filter *values* as JSONB. A view saved today with `CBOM_ASSET_CURVE EQUALS "p-256"` returns zero rows forever once ingest folds it to a `secg/*` representative — no error, no signal. Worth recording there too that curve feeds the identity preimage, so folding after any real ingest is a re-key event, not just a test rewrite.
- **core#2165's cross-lane section** says no work is required in #2074. Slightly off now: this PR serves `occurrenceCount` as `COALESCE(SUM(s.occurrenceCount), 0)`, so #2165's occurrence-triple multiplicity item moves a client-facing field. Ticket text, not this diff.
- **Untested branches worth knowing about:** the OID carve-outs for `STARTS_WITH`/`ENDS_WITH`/`MATCHES`/`NOT_MATCHES`/`NOT_CONTAINS` against a refuted row; the source-CBOM predicate's default-branch refusal; free text with no value; `loadPage`'s vanished-row skip; the both-null name branch; and the 422 status itself over the wire (only the exception type is asserted, at service level).

## Retracted

Two things a mechanical pass flagged that are not real: the `@AutoConfigureMockMvc` context-signature concern (baseline verified unchanged, and correctly unchanged), and the identity-key mentions in the new test (they are the assertion, not a leak).

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -75,19 +81,37 @@ private FilterPredicatesBuilder() {
...
+
     public static <T> Predicate getFiltersPredicate(final CriteriaBuilder criteriaBuilder,
             final CommonAbstractCriteria query, final Root<T> root, final List<SearchFilterRequestDto> filterDtos) {
         Map<String, From> joinedAssociations = new HashMap<>();
 
+        // An explicit filter on the refuted-OID facet is the caller opting into matching refuted OID
+        // values, so the carve-outs below switch off for the whole request; the facet's own predicate
+        // then selects rows by refutedness. Without this, combining the facet with an OID filter would
+        // be self-contradictory and always empty.
+        boolean refutedOidsOptedIn = filterDtos != null && filterDtos
```
vyskocilm, line 106: L103: 🔴 bug: presence-keyed opt-in + the array-value operator flip lets refuted rows into a request whose intent excludes them.

L420-427 computes `Boolean.parseBoolean(filterDto.getValue().toString())`, and `List.of("true").toString()` is `"[true]"` → `false` → the operator flips. This block keys the opt-in on field *presence* alone, so a single array-valued facet filter does two things at once: inverts the facet, and disarms the refuted carve-outs (L613-618, L645-647) for the whole request.

The leaking shape is `[FREE_TEXT CONTAINS x, OID_REFUTED NOT_EQUALS ["true"]]`. Caller intent: "matching rows, excluding refuted". What happens: `NOT_EQUALS` flips to `EQUALS REFUTED_OID` (refuted rows only) while the free-text OID side loses its guard — so the response is exclusively refuted rows, matched through their refuted OIDs. Same permission scope, so not a confidentiality breach, but it inverts the refuted-OID acceptance criterion for exactly the compliance caller it was written for. It is also persistable in a saved list view: `ListViewServiceImpl.validateFilters` checks field and condition, never value shape.

Note 8 has the mechanism but calls it "no leak" and defers the guard to a platform follow-up. This combination is the leak-shaped case, and it exists because this PR adds both the second `expectedValue` field and the presence-keyed switch. The `value instanceof List → 422` guard belongs here.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -2,43 +2,110 @@
...
+        long totalItems = cryptoAssetRepository.countUsingSecurityFilter(filter, where);
```
vyskocilm: L101: 🟡 risk: `countUsingSecurityFilter` emits `count(DISTINCT uuid)`, which Postgres cannot parallelize. Measured warm on a 5M-row `crypto_asset` with this migration's exact indexes: **9,348 ms**, against **182 ms** for `count(*)`. That is the default first page of the inventory.

The DISTINCT is provably unnecessary on this path — `FilterField` L419-423 justifies the EXISTS subquery precisely so no predicate ever duplicates a root row, and `Resource.CRYPTO_ASSET` is declared without groups or owner, so the security filter adds only `uuid IN`. Fix: a non-distinct count variant for this endpoint.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -47,15 +114,145 @@ public CryptographicAssetDetailDto getCryptographicAsset(SecuredUUID uuid) throw
...
+        List<SearchFieldDataDto> fields = List
+                .of(SearchHelper.prepareSearch(FilterField.CBOM_ASSET_FREE_TEXT),
+                        SearchHelper.prepareSearch(FilterField.CBOM_ASSET_TYPE),
+                        SearchHelper.prepareSearch(FilterField.CBOM_ASSET_NAME),
+                        SearchHelper.prepareSearch(FilterField.CBOM_ASSET_OID),
+                        SearchHelper.prepareSearch(FilterField.CBOM_ASSET_OID_REFUTED),
+                        SearchHelper
+                                .prepareSearch(FilterField.CBOM_ASSET_ALGORITHM_FAMILY,
+                                        cryptoAssetRepository.findDistinctAlgorithmFamily()),
+                        SearchHelper
```
vyskocilm, line 150: L136: 🟡 risk: eight unbounded `SELECT DISTINCT … ORDER BY` per searchable-fields call, with no caching anywhere in the path. Postgres ≤16 has no skip scan, so the per-column btrees from `V202608271000` are never used for these — the planner picks Parallel Seq Scan. Measured at 5M rows: 254–432 ms each, ~2.2 s summed, on every filter-panel open. It is also a cheap lever against the DB, since each request pins three parallel workers for the duration.

Fix: a recursive-CTE loose index scan per column returns ~1 ms against the indexes that already exist; a short-TTL cache would also do.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -47,15 +114,145 @@ public CryptographicAssetDetailDto getCryptographicAsset(SecuredUUID uuid) throw
...
+        // PageRequest would otherwise turn an out-of-range value into a 500 rather than a shaped 422.
+        if (request.getPageNumber() < 1) {
+            throw new ValidationException("Page number must be at least 1, but was " + request.getPageNumber());
+        }
+        if (request.getItemsPerPage() < 1) {
+            throw new ValidationException("Items per page must be at least 1, but was " + request.getItemsPerPage());
+        }
+        if (request.getItemsPerPage() > MAX_ITEMS_PER_PAGE) {
+            request.setItemsPerPage(MAX_ITEMS_PER_PAGE);
+        }
```
vyskocilm, line 234: L204-212: 🟡 risk: `pageNumber` is bounded below but not above, so `(int) Pageable.getOffset()` wraps — defeating the purpose stated two lines up.

`pageNumber=2147485, itemsPerPage=1000` → offset 2,147,484,000 → `(int)` = −2,147,483,296 → `setFirstResult` throws `IllegalArgumentException` → generic 400, not the shaped 422 this method promises.

Worse: `pageNumber=4294969, itemsPerPage=1000` → offset 4,294,968,000 wraps to **+704**, and the response silently serves rows 704–1703 while echoing the requested page number and the true `totalItems`. No error at all.

Fix: refuse when `(long) (pageNumber - 1) * itemsPerPage > Integer.MAX_VALUE`.

src/main/java/com/otilm/core/api/web/CryptographicAssetControllerImpl.java
```diff
@@ -32,6 +33,7 @@ public void setCryptographicAssetService(CryptographicAssetExternalService crypt
 
     @Override
     @AuditLogged(module = Module.CORE, resource = Resource.CRYPTO_ASSET, operation = Operation.LIST)
+    @AuthEndpoint(resourceName = Resource.CRYPTO_ASSET)
```
vyskocilm, line 36: L36: 🟡 risk: `@AuthEndpoint` makes `ContextRefreshListener` sync a list-objects endpoint, but interfaces builds `Resource.CRYPTO_ASSET` with the two-arg constructor, so `hasObjectAccess` is false and `ResourceServiceImpl.listResources` serves that verbatim on `/v1/resources`.

Two platform surfaces then disagree about the same resource: the auth service advertises object access, `/v1/resources` denies it. A role editor gating the object picker on `/v1/resources` hides exactly what this PR wires up, leaving `listResourceObjects`, `getResourceObjectExternal` and `evaluatePermissionChain` unreachable.

Of the 19 resources carrying `@AuthEndpoint`, 18 pass `objectAccess=true`; CRYPTO_ASSET is the only exception. Note 2 describes this but names no interfaces issue. Either file that flag change and merge it in lock-step, or drop `@AuthEndpoint` and the extension bean — the shipped state is internally inconsistent either way.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -47,15 +114,145 @@ public CryptographicAssetDetailDto getCryptographicAsset(SecuredUUID uuid) throw
...
+        return cb.coalesce(root.get(CryptoAsset_.name), root.get(CryptoAsset_.oid));
+    }
+
+    private static CryptographicAssetDto toDto(CryptoAssetListRow row) {
+        CryptographicAssetDto dto = new CryptographicAssetDto();
+        dto.setUuid(row.uuid());
+        // The contract marks name REQUIRED and the wire mapper drops nulls, so a nameless producer row serves its
+        // recorded OID -- the same next-best stable label displayLabel gives the object picker. A row with neither
+        // still serializes without the field; that residual is interfaces' contract friction, raised on the PR.
+        dto.setName(row.name() != null ? row.name() : row.oid());
```
vyskocilm: L246: 🟡 risk: two separate things ride on this fallback.

**Both-null is real production data, not a corner case.** The migration makes `name` and `oid` nullable deliberately, for components carrying no cryptoProperties — the `UNROUTABLE` backstop type the contract enum declares. Reproduced against this project's own wire mapper: `{"uuid":"…","type":"certificate","pqcVerdict":"unknown","sourceCbomCount":0,"occurrenceCount":0,"quarantined":false}` — no `name` member at all, against `requiredMode = REQUIRED`. Strict generated clients (Kotlin non-null constructor params, validating Go/Java, TS with a runtime validator) fail the whole page on one such row. Note 7 disposes of it correctly, but no interfaces issue exists and no ticket owns it.

**A refuted OID is served here as the display `name` with nothing marking it.** The contract's own principle for refuted identifiers is stated on `CryptographicAssetOidDto`: the flag "is what lets a client label such an entry instead of presenting it as fact". This DTO has no such flag, so a UI resolving OID-shaped names against a registry renders precisely the false match the refuted ruling exists to prevent. The round trip fails too: copy the displayed name, paste it into free text, get zero rows — because the carve-out at L646 correctly refuses it. Cheapest coherent fix is to exclude `REFUTED_OID` rows from the OID fallback.

src/main/java/com/otilm/core/dao/repository/CbomRepository.java
```diff
@@ -59,4 +59,8 @@ public interface CbomRepository extends SecurityFilterRepository<Cbom, UUID> {
             """)
     int updateAssetSyncState(@Param("uuid") UUID uuid, @Param("state") CbomAssetSyncState state,
             @Param("error") String error, @Param("syncedAt") OffsetDateTime syncedAt);
+
+    /** Distinct CBOM serial numbers, the value list of the inventory's source-CBOM filter. */
+    @Query("SELECT DISTINCT c.serialNumber FROM Cbom c ORDER BY c.serialNumber")
+    List<String> findDistinctSerialNumber();
```
vyskocilm: L65: 🟡 risk: this is the platform's first *cross-resource* value list. The precedents are unfiltered too (`CertificateRepository` L172-188, `DiscoveryRepository` L38-39, `EntityInstanceReferenceRepository` L17-21), but each reads its own resource's table. This one hands every CBOM serial number on the platform to a caller holding `cryptoAssets:list` and no `cboms:list` — the bypassed gate is precisely the `cboms:list` resource gate. Note 5 asks for sign-off and is right to.

Undisclosed and separate: the list is **unbounded**. Every precedent is a bounded vocabulary (signature algorithms, connector names); this one grows one entry per CBOM document, forever. At tens of thousands of documents the searchable-fields response becomes a multi-MB payload, rebuilt on every filter-panel open. No cap, no paging. `CbomRepository` already extends `SecurityFilterRepository`, so filtering it is available if the sign-off goes the other way.

src/test/java/com/otilm/core/integration/api/web/CryptographicAssetAuthorizationHttpITest.java
```diff
@@ -0,0 +1,86 @@
...
+        assertThat(listResponse).doesNotContainPattern(IDENTITY_KEY);
+
+        String searchableFieldsResponse = mockMvc
+                .perform(get(SEARCHABLE_FIELDS_ENDPOINT))
+                .andExpect(status().isOk())
+                .andReturn()
+                .getResponse()
+                .getContentAsString();
+        assertThat(searchableFieldsResponse).doesNotContainPattern(IDENTITY_KEY);
+    }
```
vyskocilm, line 144: L69: 🟡 risk: this test can pass on an empty page. It asserts only `doesNotContainPattern` and never that the seeded row reached `items`, so any filter or security regression that empties the response satisfies both assertions. `jsonPath("$.totalItems").value(1)` or `$.items[0].name` = `ecdsa` closes it.

Second, the pattern catches field *names*, not values. Executed by hand: it matches `identityKey`, `identity_key`, `identity-key`, `identity key`, `identitykey` in any case. It does **not** match a field named `"identity"` carrying the key, a bare `CRT|H|v1|…` value landing in `name` or `oid`, a raw hex digest, or base64. On a wire payload the realistic leak is a value, not a name — the fence was built to scan source for identifiers, which is a different job.

The seed goes through `upsertIdentity`, so `CryptoAssetIdentityCalculator.calculate(...)` can recompute this row's exact key from the same fixture input and assert `doesNotContain(key)` on both raw responses. Strictly stronger, and worth keeping alongside the name pattern rather than instead of it.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -560,9 +605,95 @@ private static <T> Predicate getPropertyFilterPredicate(final CriteriaBuilder cr
...
+        }
+        if (filterDto.getCondition() != FilterConditionOperator.CONTAINS) {
+            throw new ValidationException(
+                    "Free-text filter for field " + filterField + " supports only the CONTAINS operator.");
+        }
+
+        Expression<String> pattern = criteriaBuilder
+                .lower(criteriaBuilder.literal("%" + escapeLikeWildcards(filterValues.getFirst().toString()) + "%"));
+        Predicate nameMatches = criteriaBuilder
+                .like(criteriaBuilder.lower(root.get(CryptoAsset_.NAME)), pattern, LIKE_ESCAPE_CHAR);
```
vyskocilm, line 706: L642: 🔵 nit: `V202608271000` L167-169 states "No lower(name) / lower(oid) expression index: FilterPredicatesBuilder never emits lower()". These lines now emit exactly that, so the migration's premise is false for the next reader of the checkout. The applied file is checksum-frozen and a PR description is not part of the tree — so put the correction in this method's javadoc: free text lowercases both columns, the migration's premise is superseded, the conclusion still holds because infix LIKE is btree-unservable and a trigram GIN is the upgrade path.

Measured while confirming this: prefix LIKE is equally unservable under the `en_US.utf8` collation — `STARTS_WITH` is advertised on both STRING fields and emitted raw at L492, and at 2M rows the plan shows `Filter: (name ~~ 'ML-KEM-768-19%')` with 1.37M rows removed, never an Index Cond. So a trigram-GIN upgrade would need `text_pattern_ops` coverage for prefix too, not just infix.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -47,15 +114,145 @@ public CryptographicAssetDetailDto getCryptographicAsset(SecuredUUID uuid) throw
...
+    @ExternalAuthorization(resource = Resource.CRYPTO_ASSET, action = ResourceAction.LIST)
+    public List<NameAndUuidDto> listResourceObjects(SecurityFilter filter, List<SearchFilterRequestDto> filters,
+            PaginationRequestDto pagination) {
+        return cryptoAssetRepository
+                .listResourceObjects(filter, CryptographicAssetServiceImpl::displayLabel, null, pagination);
+    }
+
+    // DETAIL is this resource's own object-read gate, and it has no parent resource to chain through.
+    @Override
+    @ExternalAuthorization(resource = Resource.CRYPTO_ASSET, action = ResourceAction.DETAIL)
```
vyskocilm: L192: 🔵 nit: every other extension service gates `evaluatePermissionChain` with `UPDATE` — its one generic caller is `ResourceServiceImpl.updateAttributeContentForObject`, which is a write. Unreachable today because `Resource.CRYPTO_ASSET` has `hasCustomAttributes=false`, and DETAIL does avoid syncing a spurious `update` action onto a read-only resource, which is a good reason. Worth stating the latent condition in the comment above — "must become UPDATE if hasCustomAttributes is ever enabled" — or it gets copied as precedent.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -560,9 +605,95 @@ private static <T> Predicate getPropertyFilterPredicate(final CriteriaBuilder cr
...
+        if (filterField != FilterField.CBOM_ASSET_FREE_TEXT) {
+            throw new ValidationException("Free-text filter is not defined for field " + filterField + ".");
+        }
+        if (filterDto.getCondition() != FilterConditionOperator.CONTAINS) {
+            throw new ValidationException(
+                    "Free-text filter for field " + filterField + " supports only the CONTAINS operator.");
+        }
+
+        Expression<String> pattern = criteriaBuilder
+                .lower(criteriaBuilder.literal("%" + escapeLikeWildcards(filterValues.getFirst().toString()) + "%"));
```
vyskocilm: L639-640: 🔵 nit: `cb.literal` inlines the pattern into the SQL statement text; the generic CONTAINS at L497 binds it instead. Traced through hibernate-core 6.6.53: `SqmLiteral` → `AbstractSqlAstTranslator.visitQueryLiteral` → `Dialect.appendLiteral`, which `PostgreSQLDialect` does not override.

Not injectable — quote-doubling plus `standard_conforming_strings=on` closes that, and `MATCHES`/`GREATER` already inline literals as platform precedent. The residual is that search terms (which people do paste secrets into) land verbatim in Postgres server logs under `log_min_duration_statement`, where a bound parameter would not, and a multi-MB search string becomes a multi-MB statement. `cb.lower(((HibernateCriteriaBuilder) cb).value(pattern))` binds and keeps the SQL-side lowering.

Same line: `filterValues.getFirst()` silently drops the rest of a multi-value payload on a field advertised `multiValue=false`. The `value instanceof List → 422` guard would cover both this and the facet inversion at L103.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -47,15 +114,145 @@ public CryptographicAssetDetailDto getCryptographicAsset(SecuredUUID uuid) throw
...
+        return cryptoAssetRepository
+                .findResourceObject(objectUuid.getValue(), CryptographicAssetServiceImpl::displayLabel);
+    }
+
+    @Override
+    @ExternalAuthorization(resource = Resource.CRYPTO_ASSET, action = ResourceAction.LIST)
+    public List<NameAndUuidDto> listResourceObjects(SecurityFilter filter, List<SearchFilterRequestDto> filters,
+            PaginationRequestDto pagination) {
+        return cryptoAssetRepository
+                .listResourceObjects(filter, CryptographicAssetServiceImpl::displayLabel, null, pagination);
```
vyskocilm: L187: 🔵 nit: `filters` is discarded (`null` passed through), so a search typed into the role editor's object picker is ignored and the caller gets unfiltered pages. `EntityInstanceServiceImpl` L393-398 is the filter-honouring precedent. Several resources ignore it too, so this is consistent-but-inconsistent — worth a decision rather than inheritance, since this is new surface.

src/test/java/com/otilm/core/integration/search/CryptoAssetSearchITest.java
```diff
@@ -177,6 +186,196 @@ void aFilterCombinesWithAnother() {
...
+        assetWriter
+                .upsertIdentity(new CryptoAssetIdentityFields(CryptographicAssetType.ALGORITHM, "AESX128",
+                        "oid-lookalike", null, null, null, null, null, null, null), null);
+
+        assertThat(search(aPropertyFilter(FilterField.CBOM_ASSET_FREE_TEXT, FilterConditionOperator.CONTAINS, "s_1")))
+                .describedAs("an underscore matches a literal underscore, not any character")
+                .containsExactly(underscored);
+        assertThat(search(aPropertyFilter(FilterField.CBOM_ASSET_FREE_TEXT, FilterConditionOperator.CONTAINS, "%")))
+                .describedAs("a lone percent matches rows containing a literal percent -- here, none")
+                .isEmpty();
```
vyskocilm: L321-323: 🔵 nit: no positive control. With no seeded row containing a literal `%`, this assertion cannot distinguish correct escaping from a query that matches nothing at all. The `_` case above has its pair (`AES_128` matches, `AESX128` does not) — seed a row with a literal `%` in the name and assert it comes back.

src/test/java/com/otilm/core/integration/cbom/CryptoAssetListQueryITest.java
```diff
@@ -0,0 +1,230 @@
...
+        assertThat(page0).describedAs("page 0 is the first two of the full order").containsExactly(aesFirst, aesSecond);
+        assertThat(page1)
+                .describedAs("page 1's boundary falls inside the tied 'aes' name group -- the tiebreak makes it "
+                        + "deterministic")
+                .containsExactly(aesThird, ecdsa);
+        assertThat(page2).describedAs("page 2 is the null-name row, still last").containsExactly(nullNamed);
+    }
+
+    @Test
+    void distinctValueQueriesReturnCanonicalSortedValuesWithoutNullsOrDuplicates() {
```
vyskocilm, line 115: L106: 🔵 nit: core#2166 says this test "must start failing the moment ingest folds them". It only does if folding lands at or below `upsertIdentity` — the writer path this fixture uses. core#2166 assigns class folding to the ingest pipeline; if it lands in the extractor, upstream of the writer, every fixture here bypasses it and the tripwire never fires. Worth naming the required landing spot in core#2166, since the AC then rests on that ticket's rewrite checkbox rather than on this test.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
vyskocilm: Second gate pass adds a harder fact than the one above, and narrows the fix.

`grep -ri quarantin src/main` returns **only this line and its comment**. There is no reconciliation queue anywhere in core — so the second half of the contract sentence, "quarantined pending reconciliation", currently has no mechanism behind it for *any* of the three guard values, not just the two that do not belong.

For the record on the ratified text I cited: it is R14 addendum, ratified 2026-08-20, verbatim — "a CN-only row may not be hand-merged into a full-DN row — the refusal list keeps the DN-scope split, so these duplicates are permanent by design: 7 rows on the current corpus, unbounded in an estate mixing a bare-CN producer with a full-DN one. The ruling is that no path may exist — not even a manual one." Item 11, ratified the same day, restates it. So `BARE_CN_SUBJECT` is definitionally not "pending" anything.

`REFUTED_CERTIFICATE_DIGEST` is the one value the wording does fit — R13 specifies "a row-level flag plus a reconciliation-queue entry showing both contradicting claims". That queue is the piece core does not have yet, which is worth knowing before this flag reaches a UI that offers an action against it.

src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
```diff
@@ -173,4 +175,49 @@ void applyPqcVerdict(@Param("uuid") UUID uuid, @Param("verdict") String verdict,
...
+    List<String> findDistinctAlgorithmFamily();
+
+    @Query("SELECT DISTINCT a.primitive FROM CryptoAsset a WHERE a.primitive IS NOT NULL ORDER BY a.primitive")
+    List<String> findDistinctPrimitive();
+
+    @Query("SELECT DISTINCT a.parameterSet FROM CryptoAsset a WHERE a.parameterSet IS NOT NULL ORDER BY a.parameterSet")
+    List<String> findDistinctParameterSet();
+
+    @Query("SELECT DISTINCT a.curve FROM CryptoAsset a WHERE a.curve IS NOT NULL ORDER BY a.curve")
+    List<String> findDistinctCurve();
```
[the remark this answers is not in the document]
vyskocilm: Adjudicating this one, because it and the deferral to core#2166 are two different claims and only one of them holds.

**The acceptance criterion is genuinely unmet** — you are right, and this was the one substantive point the other review passed over. `findDistinctCurve()` returns raw normalized tokens, and `CryptoAssetListQueryITest:106-118` pins exactly that by seeding `SECP256R1`, `secp256r1` and `P-256` as three fixtures.

**But "must land before this endpoint closes #2074" is a ticket-closure question, not a merge blocker.** Class-level union-find canonicalization is the ingest pipeline's obligation, it is blocked on core#2072, and core#2166 already exists, is linked, and names the exact test to rewrite. Nothing in the conventions is violated by deferring; the follow-up requirement is satisfied. The right resolution is to **move that AC checkbox from #2074 to #2166**, so #2074 can close honestly, rather than to hold this PR for work that cannot be done in it.

One thing to add to core#2166 while it is open: its checkbox says folding lands in "the #2072 ingest pipeline" without naming the layer. The pinning test seeds through `assetWriter.upsertIdentity(...)`, so it only trips if folding lands **at or below the writer**. If it lands in the extractor upstream, every fixture bypasses it and the tripwire never fires — the safety net becomes the checklist item, not the test.

src/main/java/com/otilm/core/enums/FilterField.java
```diff
@@ -415,6 +416,17 @@ public enum FilterField {
...
             SearchFieldTypeEnum.NUMBER),
 
+    // Free text spans lower(name)/lower(oid) and the refuted-OID opt-in couples with the OID
+    // predicates, so all three are built by dedicated branches in FilterPredicatesBuilder rather
+    // than the generic attribute path. CBOM_ASSET_SOURCE_CBOM matches through an EXISTS subquery,
+    // never a join: the uuid page query has no DISTINCT, and a row with several sources would
+    // otherwise repeat inside one page.
+    CBOM_ASSET_FREE_TEXT(Resource.CRYPTO_ASSET, null, null, null, "Text Search", SearchFieldTypeEnum.FREE_TEXT),
+    CBOM_ASSET_OID_REFUTED(Resource.CRYPTO_ASSET, null, null, CryptoAsset_.identityGuard, "OID Refuted",
+            SearchFieldTypeEnum.BOOLEAN, null, CryptoAssetIdentityGuard.REFUTED_OID, false, null),
```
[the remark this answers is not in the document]
vyskocilm: Confirmed, and the two anchors are two halves of one defect — worth reading together with the thread on `FilterPredicatesBuilder` L103.

This line introduces the *reachability*: `CBOM_ASSET_OID_REFUTED` is the second `expectedValue` field on the platform, so the pre-existing `Boolean.parseBoolean(filterDto.getValue().toString())` at `FilterPredicatesBuilder:421` now has a new way to be reached. L103 introduces the other half, which is new in this PR and has no analogue on `PRIVATE_KEY`: the opt-in is keyed on field *presence*, so the same array-valued term that inverts the facet also disarms the refuted carve-outs for the whole request.

The compound is what matters. `[FREE_TEXT CONTAINS "aes", OID_REFUTED NOT_EQUALS ["true"]]` → `parseBoolean("[true]")` is `false` → `NOT_EQUALS` flips to `EQUALS REFUTED_OID`, while the free-text OID side has already lost its guard. Result: exclusively refuted rows, matched through their refuted OIDs — the inversion of the acceptance criterion that says a default search whose only match is a refuted OID returns nothing.

So your ask (strict scalar-boolean validation before exposing the field) is the right fix, and it belongs in this PR rather than a platform follow-up: PR note 8 disposes of the flip as inherited, which is true of the flip alone and not of the compound. `ListViewServiceImpl.validateFilters` checks field and condition but never value shape, so the bad payload is also persistable in a saved list view.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -296,7 +320,7 @@ private static Object prepareDurationValue(AttributeContentType contentType, Str
```
vyskocilm, line 323: L320: 🔵 nit: `getPropertyFilterPredicate` now runs to L617 — roughly 297 lines against the ~80–100 the codebase aims at. This PR did not create that (it was already far over) and it does extract four helpers, so this is not an ask to refactor it here.

Recording it because the four new branches added at L332-354 sit at the top of a method that is already past the point where a reader can hold it, and the next PR to touch this method is the one that should split it. The natural seams are visible: the special-field branches, the expression resolution, and the operator switch.

vyskocilm (commented):
Second review pass — the architecture and convention gate, re-run with the feedback already on this PR as an input rather than in isolation. **Verdict: fail, nine findings.** Almost all of them already have a thread above; this comment is the delta, the adjudication of where two reviewers said the same thing, and an honest note on what is *not* backed by a test run.

## Where the existing comments landed

Every point on this PR was re-verified independently rather than inherited. **None of the sixteen inline comments from the earlier review is wrong**, and **four of the six Copilot comments are confirmed**. Two Copilot points were overstated rather than wrong:

- `CryptographicAssetServiceImpl:187` (discarded `filters`) — the fact is right, "bug" is not. I counted nine implementations of `listResourceObjects`: the platform does it both ways. It is a decision to make, not a rule broken.
- `CryptoAssetRepository:197` (curve tokens) — the acceptance criterion really is unmet, and this was the one substantive thing the other review passed over. But the deferral to core#2166 is properly ticketed and blocked on core#2072. Adjudicated in that thread: move the AC checkbox to core#2166 rather than hold this PR.

Where both reviewers hit the same line, the fuller statement is generally the later one — except on the boolean-inversion defect, where Copilot's anchor (`FilterField:425-426`, which introduces the reachability) and the other's (`FilterPredicatesBuilder:103`, which introduces the disarm) are two halves of one problem. Replied on both threads.

**Contradiction with the PR body, twice.** Note 8 concludes "no leak": true of the operator flip alone, which is inherited, and not true of the compound this PR creates. Note 5 describes the CBOM value list as bypassing row-level security "per platform precedent": the precedents bypass scope *within their own resource*, while this crosses a resource gate — the thing being signed off is larger than the note describes.

## What the gate itself missed the first time

Worth saying plainly: an earlier pass of this same gate returned two findings. Re-run against the contract in `interfaces` and the checklist, it should have caught five more — the refuted-OID inversion, the cross-resource value list, `quarantined` firing on a superset of its contract meaning, the sort-column/served-field mismatch, and the omissible `REQUIRED` field. All five were reachable by reading the diff; none needed a database. The first pass checked that the annotations were *present* and stopped there.

## Sonar

Quality gate **passed** — 96.6% coverage on new code (against the ticket's own 80% bar), 0.0% duplication, 0 security hotspots. That closes the one criterion no local run could verify, since containers will not start on this machine.

The **12 new issues are all code smells — zero bug-typed, zero vulnerability-typed**, so none of them touches the reliability rating and none is a gate finding. Not restating them as review comments; the bots have that covered. One observation worth acting on: four of them are the same "one invocation possibly throwing" lambda rule clustered in `CryptoAssetSearchITest` 230–300 and `CryptographicAssetControllerITest:51`. That is the shape where a structural fix clears the set in one go, rather than four separate replies as line numbers shift on each push. Also `0 accepted` means none has been triaged yet.

## Honest gaps

- **No integration test ran.** The container runtime is broken here, so all six changed suites, the 403 assertions and the wire-level assertions were **not executed** by this pass. An earlier pass did run them against an external PostgreSQL 16.14 rather than the pinned 17-alpine — worth treating as a parity caveat, not as a clean run.
- **Three findings are reasoned from source, not observed.** The refuted-OID inversion, the sort mismatch and the page-number overflow are all expression-tree or arithmetic reasoning traced by hand. Each is checkable, none has a failing test behind it. The inversion in particular deserves a written regression test *before* the fix is believed — it is the one that silently returns the opposite of what was asked.
- **Performance numbers quoted earlier** (the count-distinct timing, the value-list scans) have their premises confirmed in source but their magnitudes are not reproducible without a database.

## Clean, and worth stating

No policy bundle, no permission seed, no auth migration anywhere in the diff — registration is the annotations alone, which is correct and is the single most commonly gotten-wrong thing in this area. No transaction opened, no lock taken, no external call: the surface is read-only. No raw exception message can reach a client. The identity key is named in no filter field, and the build-time fence plus its self-test are green. Context-signature baseline correctly unchanged. spotless and checkstyle clean across 1921 files.

## Small, no thread of its own

The PR description uses bare `#2074`, `#2157`, `#2158`. In-code references are correctly repo-prefixed; the description is the part that outlives the context, so it is worth prefixing there too.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Fixed in eadc62c2: quarantined == (identityGuard == REFUTED_CERTIFICATE_DIGEST), with the rule and the widen-only-via-interfaces-rewording condition stated at the mapping. mapsListRowsToDtos now pins all three guard values (digest true, bare-CN false, refuted-OID false). Your follow-up's harder fact — no reconciliation mechanism exists in core for ANY guard yet — is visible in the interfaces#925 ticket text as well; the flag now claims only what R13 actually promises for the digest case.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Fixed in eadc62c2: the list orders by the served label — the same guarded coalesce displayLabel builds (name, else oid unless refuted) — and SortOrderBuilder still appends the ascending-uuid tiebreak. Regression test ordersByTheServedNameNotTheBareColumn seeds an OID ('0.aaa') that sorts before every fixture name, exactly the shape the old fixtures missed.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
karolhrdina: Fixed in ef9177a6, in this PR as argued: an expectedValue field refuses any non-scalar or non-boolean value with a shaped 422 before any predicate is built, so the array shape can neither invert the facet nor ride the disarmed carve-outs. The regression test (theRefutedFacetRefusesNonScalarAndNonBooleanValues) was written first and failed on the pre-fix code with your exact compound — [FREE_TEXT CONTAINS x, OID_REFUTED NOT_EQUALS ["true"]]. The presence-keyed opt-in stays, which is now safe: only a scalar true/false ever reaches the flip.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Fixed in da770563: countRowsUsingSecurityFilter — a plain count(root) variant on the repository — with the justification in its javadoc, and a test pinning plain == distinct == 1 through the one predicate shape that would differ (an asset matched through two sources).

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Fixed in da770563: the seven per-column value lists are recursive-CTE loose index scans (index-min hops over the btrees V202608271000 already ships; sorted and distinct by construction, min() skips NULLs). The eighth list — CBOM serials — went the other way for a different reason: see the CbomRepository thread (scoped by the caller's CBOM access).

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Fixed in eadc62c2: validatePaging refuses (long)(pageNumber-1)*itemsPerPage > Integer.MAX_VALUE with the shaped 422. Both of your shapes are tests now — the wrap-to-+704 page and the negative-offset IllegalArgumentException page.

src/main/java/com/otilm/core/api/web/CryptographicAssetControllerImpl.java
karolhrdina: Filed as interfaces#924 with lock-step release as its acceptance criterion. Kept @AuthEndpoint rather than dropping it: the #2074 ticket text itself mandates the synced listing endpoint ('so the synced catalogue entry carries a listing endpoint'), so removal would regress this ticket's own AC. Until interfaces ships the flag, /v1/resources stays the conservative surface — the picker is hidden, never wrongly shown. The stale AuthServiceITest stub is named in interfaces#924 as part of the flip.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Both halves addressed. eadc62c2 excludes REFUTED_OID from the OID fallback everywhere a label is served — the list's name, the picker's label, and the sort key are one expression now — so the false-fact rendering and the copy-paste-zero-rows round trip are both gone; such a row serves no name at all. That makes both-null AND refuted-nameless the shapes with no servable label, and interfaces#925 now owns the ratification (nullable name, a guaranteed fallback, or a per-row refuted flag like the detail DTO's).

src/main/java/com/otilm/core/dao/repository/CbomRepository.java
karolhrdina: Fixed in eadc62c2: the serial list is scoped by the caller's own cboms:list object access — populateSecurityFilter (the rule-filter-fields mechanism) plus a grouped count over the filtered rows — so a caller without CBOM access gets an empty value list, tested over HTTP in both directions. The unbounded-size half is still real for a permitted caller (N documents = N serials, no cap or paging) and stays named in the PR notes: a cap would be a contract change, and CbomRepository's SecurityFilterRepository plumbing is now the natural place if the sign-off asks for one.

src/test/java/com/otilm/core/integration/api/web/CryptographicAssetAuthorizationHttpITest.java
karolhrdina: Fixed in eadc62c2: positive control (totalItems 1 + items[0].name ecdsa) so an emptied page can no longer satisfy the assertions vacuously, plus the strongest version of your suggestion — CryptoAssetIdentityCalculator recomputes the seeded row's exact key from the same fixture input and both raw responses assert doesNotContain(key), alongside the existing name-pattern check.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
karolhrdina: Fixed in ef9177a6: the free-text predicate's javadoc now carries the correction (method emits lower(), migration premise superseded, no-expression-index conclusion still holds) plus your prefix-LIKE measurement — the upgrade path is recorded as trigram GIN for infix AND text_pattern_ops btrees for prefix, as a new migration, never an edit to the applied file.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Fixed in eadc62c2: the comment now states the latent condition — DETAIL is deliberate while hasCustomAttributes is false, and must become UPDATE if custom attributes are ever enabled — so the precedent copies with its condition attached.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
karolhrdina: Fixed in ef9177a6: the pattern is bound via HibernateCriteriaBuilder.value(...) with lower() kept on the SQL side, so search terms no longer reach statement text (server logs, pg_stat_statements); and a multi-value free-text payload is refused with a 422 instead of silently dropping the tail. The facet's ["true"] shape is covered by the scalar-boolean guard on the same commit.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Implemented in eadc62c2 — the decision went to honouring the filters on this new surface, per the entity-instance precedent: the picker builds the same predicate the inventory list does. Tested (listResourceObjectsHonoursPickerFilters).

src/test/java/com/otilm/core/integration/search/CryptoAssetSearchITest.java
karolhrdina: Fixed in ef9177a6: a row with a literal percent in its name is seeded, and the lone-% search returns exactly that row (plus a percent-inside-a-longer-literal case), so correct escaping is now distinguishable from matching nothing.

src/test/java/com/otilm/core/integration/cbom/CryptoAssetListQueryITest.java
karolhrdina: core#2166 updated with all three: a checkbox naming the required landing spot (at or below the writer, or the tripwire never fires), the saved-list-view JSONB value rot on folding, and the curve-in-preimage consequence (folding after real ingest is a re-key event, not a test rewrite).

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
karolhrdina: Recorded, and deliberately not refactored here per your own framing. The three new guard branches this PR adds at the top are the smallest they can be; the split along the seams you name (special-field branches / expression resolution / operator switch) belongs to the next PR that touches the method.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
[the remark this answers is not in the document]
karolhrdina: Implemented in eadc62c2: the picker's filters now build the same predicate the inventory list uses and narrow the object listing; tested.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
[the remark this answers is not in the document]
karolhrdina: Addressed in eadc62c2 plus interfaces#925: a nameless row serves its recorded OID unless that OID is refuted (never presented as fact); the residual shapes with no servable label — both-null, refuted-nameless — are now ratification-owned in interfaces#925 (nullable name, guaranteed fallback, or a per-row refuted flag).

src/main/java/com/otilm/core/dao/repository/CbomRepository.java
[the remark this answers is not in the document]
karolhrdina: Fixed in eadc62c2 per the adjudication in the sibling thread: the value list is scoped by the caller's own cboms:list object access instead of enforcing-or-serving-all; empty for a caller without CBOM access, tested over HTTP.

src/main/java/com/otilm/core/api/web/CryptographicAssetControllerImpl.java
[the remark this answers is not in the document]
karolhrdina: Filed as interfaces#924 (flag flip, lock-step release as its AC). @AuthEndpoint stays because the #2074 ticket text mandates the synced listing endpoint; until the flag ships, /v1/resources errs on the hidden side.

src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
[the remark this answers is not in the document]
karolhrdina: Deferral executed per this adjudication: core#2166 now carries the landing-spot checkbox (at or below the writer), the saved-view JSONB rot item, and the re-key consequence. Moving the AC checkbox from #2074's text to #2166 is a ticket-owner edit — flagged in the PR description rather than done unilaterally.

vyskocilm (approved):
Approving. All 23 threads re-derived against `da770563` rather than read off the replies — including the ones whose disposal was a deferral, since a filed ticket is a claim like any other.

**Fixed in the tree, verified:**

- `quarantined` is now `identityGuard == REFUTED_CERTIFICATE_DIGEST` alone, with the widen-only-via-interfaces condition stated at the mapping and all three guard values pinned.
- The list orders by `displayLabel` — the same guarded coalesce it serves — with `SortOrderBuilder`'s uuid tie-break, and `ordersByTheServedNameNotTheBareColumn` seeds the OID shape the old fixtures missed.
- The expectedValue guard refuses non-scalar and non-boolean values before any predicate is built, so the `["true"]` compound cannot invert the facet or ride the disarmed carve-outs. Presence-keyed opt-in staying is fine now: only a scalar reaches the flip, and the 422 aborts the whole request regardless of the order predicates are built in.
- `validatePaging` refuses `(long)(pageNumber-1)*itemsPerPage > Integer.MAX_VALUE`; both wrap shapes are tests.
- `REFUTED_OID` is excluded from the label in one expression, used by the list, the picker and the sort key alike.
- Free text binds the pattern through `HibernateCriteriaBuilder.value(...)` with `lower()` kept SQL-side, and refuses a multi-value payload. The migration-premise correction and the prefix-LIKE measurement live in the method javadoc, with a new migration named as the upgrade path.
- `countRowsUsingSecurityFilter` is a `distinct=false` variant of the existing builder — existing callers untouched. Seven value lists are recursive-CTE index hops; the CBOM serials are scoped by the caller's own `cboms:list` through `populateSecurityFilter`.
- The fence test has its positive control and the recomputed exact-key assertion on both raw responses; the lone-`%` search has a seeded literal-percent row.
- Picker filters honoured, per the entity-instance precedent. `DETAIL` carries its latent `UPDATE` condition.

No new migration, no new Spring context signature.

**Open, and not blocking the merge:**

1. **The curve AC is still on core#2074, unchecked.** My adjudication was that the checkbox moves to core#2166 — that ticket now carries the landing-spot, JSONB-rot and re-key items, but #2074 line 63 still reads "Searchable-fields offers curve class representatives, not raw". Flagging it in the PR description was the right call rather than editing someone's ticket; it does mean **#2074 must not be closed on this merge**. The body says "Implements", not "Closes", so nothing auto-closes — keep it that way.
2. **CBOM serial list size** is unbounded for a permitted caller. Sign-off still requested; `CbomRepository`'s `SecurityFilterRepository` plumbing is in place if it goes the other way.
3. **interfaces#924** must ship in lock-step or `/v1/resources` keeps contradicting the auth service. Erring hidden is the right side to err on meanwhile.
4. **CLA is unsigned** on this PR — that blocks the merge button, not the review.

vyskocilm (approved):

PR #2172, opened 2026-08-30
Send the resolved signer's algorithm to the formatting connector and pin the binding digest

### LINKED ISSUE #2159: Populate signatureAlgorithm on the formatting requests from the resolved Signer
### Description

Core-side half of DP-23 (design spec 2.7): populate the `signatureAlgorithm` field that OmniTrustILM/interfaces#915 adds to `ComputeDtbsRequestDto` and `EmbedSignatureValueRequestDto`. The value already exists in Core - `Signer.getSignatureAlgorithm()`, resolved from the key plus the cryptographic provider's attributes - but the Signer is created only inside `LiveContentSigningAcquisitions.signatureValue`, after the DTBS is computed.

- Surface the algorithm before DTBS: a lookup on the `ContentSigningAcquisitions` seam that resolves the algorithm for a resolved scheme without exercising the key, called once at the top of `signToTargetLevel` and used for both requests. Resolution is profile-deterministic, so a resume re-derives the same value; drift is caught by the connector's `formattingContext` cross-check, which answers `CONTEXT_MISMATCH` (DP-23).
- Set the field in `ManagedContentSigningEngine.computeDtbs` and `embedSignatureValue`.

This is engine plumbing, not a new capability - it would have been a core#1974 DoD line if that issue were still open.

**Scope addition 2026-08-27 (spec 2.4).** The same pre-computeDtbs site enforces the digest coherence: the authorized document digest algorithm (and a digestOnly transfer's) must equal the resolved signature algorithm's digest half, implicit-digest algorithms comparing via their recorded digest identifier. A mismatch is refused as INVALID_INPUT naming both algorithms (the caller can re-hash; the CSC credential surface already names the key algorithm); an algorithm whose digest has no `DigestAlgorithm` member (today Ed448, SHAKE256) is refused as MISCONFIGURED. This removes the misfiled-CONNECTOR_FAULT path for this cause; `DtbsBindingVerifier` stays as the backstop, no longer the first line.

**Correction 2026-08-27 (seam).** This issue originally specified the lookup on the `SignerFactory` seam. That is not available to the caller. `SigningEngineIsolationArchTest` forbids any class in `com.otilm.core.signing.contentsigning` from depending on `com.otilm.core.signing.engine.signer..`, because the orchestrator must reach non-replayable steps only through `ContentSigningAcquisitions` or a run can never be persisted and replayed. The lookup therefore lives on `ContentSigningAcquisitions`, and `LiveContentSigningAcquisitions` delegates to `SignerFactory.signatureAlgorithm`. That is also the better arrangement: the algorithm announced to the connector and the algorithm signed with are now one implementation's answer, not two independent resolutions that happen to agree.

**DP-24 amendment 2026-08-28 (spec 2.7).** **The coherence check is v1 profile policy under DP-7, not an amendment of it.** DP-7 never fixed the echo's algorithm, and no request field carries another one to a connector. So the pin, not CMS arithmetic, is what makes a mismatched pair unsatisfiable, and `SignatureDigestCoherence` says that in its own words rather than presenting the coupling as structural. Two consequences ride with it. The pin is CMS-shaped: detached XAdES and JAdES submitted digests track their Reference or sigD digest parameter, so the check takes the `SignatureFamily` when those families arrive (OmniTrustILM/ilm#354). And the engine is only today's enforcement point: when acceptance-time digest validation lands (core#1975) the pin applies there first, and this check remains the invariant.

### Definition of Done

- [ ] `signatureAlgorithm` set on both formatting requests from the resolved Signer's algorithm, resolved once per run before computeDtbs through the `ContentSigningAcquisitions` seam
- [ ] Digest coherence enforced before computeDtbs: mismatch -> INVALID_INPUT naming both algorithms; unmappable digest (Ed448) -> MISCONFIGURED; DP-7 binding check remains the backstop, and the check records the pin as v1 policy under DP-7 (DP-24)
- [ ] Engine and formatting-client tests cover the field and the coherence check (inline and digestOnly shapes), including that compute and embed carry the same value

ivosh:
**TL;DR: Core now tells the formatting connector which signature algorithm it will sign with, and refuses a run whose authorized digest is not the digest that algorithm signs.**

Closes #2159.

> **Merge order: this PR must not merge before OmniTrustILM/interfaces#927.** It compiles only against that change; a build resolving today's published `2.20.0-SNAPSHOT` cannot compile.

## What changes

The engine resolves the signature algorithm once and passes that same value to `computeDtbs` and `embedSignatureValue`. It comes from the acquisitions seam, so the orchestrator stays off the signer package. One announced value carried into both requests is what makes them unable to disagree. `SignerCreator.create` runs again to produce the signature, and the determinism the SPI requires is what keeps that run in step with the announced value.

`SignatureDigestCoherence` is new. It runs before `computeDtbs` and refuses a run whose authorized digest, or whose `digestOnly` transfer digest, is not the digest the resolved signer's algorithm commits to.

## Why the algorithm is resolved here (DP-23)

The algorithm is a function of the key plus the cryptographic provider's attributes. Core is the only party that sees both. An RSA public key does not reveal whether PKCS#1 v1.5 or PSS will be applied, so a connector cannot derive it and must be told.

`SignerFactory.signatureAlgorithm` answers it without exercising the key, because the connector is told before there is anything to sign. That is safe only if every `SignerCreator` derives the algorithm purely from the resolved scheme. That constraint is now stated on `SignerCreator.create`, where an implementer of the SPI will read it.

## Why the digest is pinned (DP-24)

v1 policy pins the binding chain — the authorized digest, a `digestOnly` transfer's digest, and the connector's echo — to the digest half of the signature algorithm, for every family. An inline transfer carries no document-digest algorithm field at all. A `digestOnly` transfer carries one explicitly, and policy pins it to the digest the signature algorithm commits to rather than letting it name another.

The check runs before `computeDtbs`, which is the engine's only point ahead of key release. That placement is the point of it. Without it, the run reaches the connector, the connector echoes the only digest the request lets it produce, and `DtbsBindingVerifier` reports an operator's misconfiguration as a connector fault. The binding gate stays as the backstop.

## Failure classification

- **INVALID_INPUT** — the authorized or submitted digest is not the one the algorithm signs, including a `digestOnly` transfer that names no algorithm at all.
- **MISCONFIGURED** — the algorithm commits to a digest the platform cannot name. Ed448 signs SHAKE256, which is no `DigestAlgorithm` value, so it can never fill the mandatory echo and is unusable for content signing whatever the caller sends.
- **MISCONFIGURED** — the key algorithm plus its signing attributes resolve to no `SignatureAlgorithm` the platform knows. That is a Signing Profile an operator can fix, so it is translated rather than left to escape as an unchecked throw a caller would log as a platform fault.

Client messages carry algorithm codes and nothing more. OIDs, the key algorithm, the resolved algorithm name and the cause stay in the operator message.

## Known gap, deliberately not fixed here

`ContentSigningFormattingClient` classifies every connector refusal as `CONNECTOR_FAULT`, including operator-fixable ones such as `PARAMETER_UNSUPPORTED`. `ConnectorException` carries the connector's message but not its `errorCode`, so there is nothing to classify on. Reading the code back means changing a shared exception every connector on the platform throws, which is a transport change well outside this issue. The decision is now recorded in the client rather than left implicit.

## Testing

`SignatureDigestCoherenceTest` covers each refusal and the accept path, and sweeps every `SignatureAlgorithm` the platform knows to prove Ed448 is the only pre-flight refusal. `ManagedContentSigningEngineTest` pins that both requests carry the same algorithm and that the run stops before the connector is asked. `ContentSigningEngineITest` proves it end to end, where a real key and a real cryptographic provider resolve the algorithm rather than a mock.

Affected units plus the content-signing ITest: 99 tests, 0 failures.


src/main/java/com/otilm/core/signing/contentsigning/SignatureDigestCoherence.java
```diff
@@ -0,0 +1,74 @@
...
+            throws SigningEngineException {
+        String oid = signatureAlgorithm.getDigestAlgorithmIdentifier().getAlgorithm().getId();
+        try {
+            return DigestAlgorithm.findByOid(oid);
+        } catch (RuntimeException e) {
+            throw new SigningEngineException(SigningEngineFailure.MISCONFIGURED,
+                    "signature algorithm '%s' commits to the digest identified by OID %s, which the platform has no digest algorithm for"
+                            .formatted(signatureAlgorithm.getCode(), oid),
+                    e, "Signing key algorithm %s cannot be used for content signing"
+                            .formatted(signatureAlgorithm.getCode()));
```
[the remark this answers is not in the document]
ivosh: Not changed. Premise confirmed: `SigningEngineException:36` passes `clientMessage` to `super(...)`, so `getMessage()` is the wire text, and the PR description sentence was wrong. The description is corrected; the code stands.

The design's rule (§3.1) is that `operatorMessage()` holds diagnostic detail, "including connector-reported text". A `SignatureAlgorithm` code is contract vocabulary, not diagnostic detail — the client supplies it on `computeDtbs.signatureAlgorithm` and reads it back on the response. It carries no OID, no key UUID, no cause text; those stay in the operator message.

Generic-ing the message was not taken because the same client text is what makes the two `INVALID_INPUT` messages actionable: "The document digest is SHA-256, but this Signing Profile signs a SHA-512 digest" names the operator's fix. Those two also interpolate algorithm codes and were not flagged, so the suggestion would have left the file inconsistent with itself.

src/main/java/com/otilm/core/signing/engine/signer/StaticManagedKeySignerCreator.java
```diff
@@ -61,14 +62,47 @@ public Signer create(ResolvedManagedScheme signingSchemeModel) throws SigningEng
...
+     * -- a SHA-1 digest, or a PQC parameter set outside the enum. That is a Signing Profile the operator can fix, so it
+     * is refused as MISCONFIGURED rather than escaping as the unchecked throw a caller would log as a platform fault.
+     */
+    private static SignatureAlgorithm resolveSignatureAlgorithm(CryptographicKeyItemModel privateKeyItem,
+            CryptographicKeyItemModel publicKeyItem, List<RequestAttribute> requestAttributes)
+            throws SigningEngineException {
+        String algorithmName = resolveSignatureAlgorithmName(privateKeyItem, publicKeyItem, requestAttributes);
+        try {
+            return SignatureAlgorithm.findByCode(algorithmName);
+        } catch (ValidationException e) {
```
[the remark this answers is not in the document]
ivosh: Fixed in 7e0b26dc. Confirmed: `CryptographyUtil:65` reads `data_rsaSigScheme` out of the attribute list first, so an empty list can never reach `findByCode`.

`throwsMisconfigured_whenTheAttributesNameAnUnsupportedSignatureAlgorithm` builds the SHA-1 RSA set you suggested via `RsaSignatureAttributesBuilder`, which resolves to `SHA1WITHRSA` — no `SignatureAlgorithm` entry — and asserts this catch's classification, its operator message and its client message.

The two catches produced the same `MISCONFIGURED` verdict, which is why the old test looked green. Both tests now assert the operator message as well, so neither can silently take the other's path. The existing test is renamed to `throwsMisconfigured_whenTheSigningAttributesNameNoSignatureAlgorithm`, after the path it covers.

src/main/java/com/otilm/core/signing/engine/signer/SignerFactory.java
```diff
@@ -29,4 +30,13 @@ public Signer create(ResolvedManagedScheme signingScheme) throws SigningEngineEx
...
                 .create(signingScheme);
     }
+
+    /**
+     * The signature algorithm {@link #create} would sign with, answered without exercising the key. It resolves a
+     * throwaway {@link Signer} that the signing call resolves again, and the two agree by the determinism
+     * {@link SignerCreator#create} requires.
+     */
+    public SignatureAlgorithm signatureAlgorithm(ResolvedManagedScheme signingScheme) throws SigningEngineException {
+        return create(signingScheme).getSignatureAlgorithm();
```
Jakub-Bartal-01, line 40: **[Low]** | Documentation

This makes create() a call any run performs twice — once to read the algorithm, once to sign — with the first Signer thrown away unused. SignerCreator.create gains the determinism obligation right below it but says nothing about the discard, so an SPI author has no reason to expect it. StaticManagedKeySignerCreator is a pure function of the scheme and CryptographicOperationServiceSigner is a value holder, so nothing leaks today; a creator that opened a session or leased a handle in create() would leak one per run, and would look correct against the contract as written.

_Suggested:_ Say on SignerCreator.create that the returned Signer may be created only to read its algorithm and then discarded, so create() must acquire nothing that needs releasing. Adding SignatureAlgorithm signatureAlgorithm(ResolvedManagedScheme) to the SPI would remove the obligation instead of documenting it.

src/main/java/com/otilm/core/signing/contentsigning/SignatureDigestCoherence.java
```diff
@@ -0,0 +1,74 @@
...
+     * identifier the platform has no {@link DigestAlgorithm} for can never fill the {@code documentDigestAlgorithm}
+     * echo the binding gate needs, so the algorithm cannot be used for content signing at all -- which is a key or
+     * profile choice, not a caller's mistake.
+     */
+    private static DigestAlgorithm committedDigest(SignatureAlgorithm signatureAlgorithm)
+            throws SigningEngineException {
+        String oid = signatureAlgorithm.getDigestAlgorithmIdentifier().getAlgorithm().getId();
+        try {
+            return DigestAlgorithm.findByOid(oid);
+        } catch (RuntimeException e) {
```
Jakub-Bartal-01: **[Low]** | Error handling

The try body is one call, and DigestAlgorithm.findByOid throws exactly ValidationException when no entry matches. Catching RuntimeException means any other unchecked failure from that lookup is reported as MISCONFIGURED with the operator message 'the platform has no digest algorithm for' this OID, sending someone to fix a Signing Profile that is fine. The sibling site added in this PR, StaticManagedKeySignerCreator.resolveSignatureAlgorithm, catches ValidationException around the equivalent SignatureAlgorithm.findByCode.

_Suggested:_ Catch ValidationException here too, matching the sibling site, and let anything else escape as the platform fault it is.

Jakub-Bartal-01 (commented):

src/main/java/com/otilm/core/signing/contentsigning/SignatureDigestCoherence.java
ivosh: Fixed in ffaad42f. Premise confirmed: `DigestAlgorithm.findByOid` `orElseThrow`s exactly `ValidationException`, and the `getDigestAlgorithmIdentifier()` chain sits on the line above the `try`, so nothing else unchecked can reach that catch.

Narrowed to `ValidationException`, matching `StaticManagedKeySignerCreator.resolveSignatureAlgorithm`. Anything else now escapes as the platform fault it is.

`SignatureDigestCoherenceTest` (8) and `ContentSigningEngineITest` (12) green.

src/main/java/com/otilm/core/signing/engine/signer/SignerFactory.java
ivosh: Fixed in ffaad42f. `SignerCreator.create` now states it: "The returned `Signer` may be created only so its algorithm can be read and then discarded, so acquire nothing that needs releasing."

Worded as a permission rather than "a run calls this twice", because the discard is not universal: `StaticKeyManagedTimestampTokenGenerator:51` creates one `Signer`, reads its algorithm and signs with that same instance. Content signing is the path that throws one away.

Not taking the SPI `signatureAlgorithm(ResolvedManagedScheme)`. A second entry point would resolve the key items and the signing attributes independently of `create()`, so the announced algorithm and the signing one could disagree — the drift the determinism paragraph directly above it exists to forbid. Documenting the obligation keeps one resolution path; today's only implementor, `StaticManagedKeySignerCreator`, acquires nothing.

Leaving this open in case you would still take the SPI method over the sentence.

Jakub-Bartal-01 (approved):

Jakub-Bartal-01 (approved):

Issue #1688, opened 2026-06-30
Complete Mode A deferred work: re-key extension projection and multi-valued RDN/SAN projection


ivosh:
## Description

Two items were consciously deferred during PR #1673 (Mode A — platform-built PKCS#10, `CertificateRequestAttributeProjector` / `X509RequestContentRenderer`). Both were confirmed against the Request Attributes design spec and Phase-1 plan during review and are tracked here so they are not lost.

**1. Re-key-from-key extension projection.** `ClientOperationServiceImpl.createRequestFromKeys` passes `null` for the projected `Extensions`, so a re-keyed certificate whose CSR is generated server-side from a key carries no SAN / extension-mapped attributes. This is pre-existing behavior (the base low-level overload never had an extensions parameter), and renew is tracked under #1706. Per the design plan, renew/re-key content **seeding** is Phase 4 scope (§9 renew row, §13 edge-case 12, INDEX "renew-seeding").

**2. Multi-valued RDN / SAN projection.** `CertificateRequestAttributeProjector.extractStringValues` reads only `content.getFirst()` and only when the data is a `String`, so a multi-value attribute (e.g. several SANs or repeated OUs under one mapping) keeps only its first entry and non-string content is dropped. This matches the legacy `buildSubject` baseline (single value per RDN via `getSingleItemAttributeContentValue`), so it is **not a regression** — but the design spec explicitly calls for multi-value (§8 "ordered / multi-valued RDNs, e.g. two OUs"; §13 edge-case 4 "1→many FQDN → CN + dNSName and many→1 repeated OU / multiple SANs", ordered via `MappedField.order`). The single-value limitation is documented in code as of PR #1673.

Also worth folding in at merge time (spec/plan reconciliation, not blocking): v3 attribute routing shipped via `authorityProviderAdapterFactory.forAuthority(...)` rather than the plan's `getCertificateApiClientV3(...)`, and the projector/renderer class names/packages drifted from the plan (`CertificateRequestAttributeProjector`/`CsrAttributes` under `attribute`/`util` vs planned `RequestAttributeProjector`/`CsrRequestAttributes` under `certificate/request`).

## Definition of Done

[...]

PR #2174, opened 2026-08-31
Carry the predecessor's identity into renew and rekey

### LINKED ISSUE #1706: Seed v3 renew request content (subject + SAN) from the existing certificate
## Description

On a v3 renew, populate structured `requestContent` on the connector wire with the renewed certificate's identity **derived from the existing certificate**, so a structured-capable connector receives a typed subject/SAN instead of having to re-parse the CSR. The renewed certificate keeps the same identity as the one it replaces. This is the renew analogue of the register dual-wire (#1633). Part of #1572 (epic OmniTrustILM/ilm#131).

## Scope — what ships

- **Pure seeder kernel** (`X509Certificate` → `X509RequestContent`, no Spring/Docker): extract the subject RDNs and SANs from the existing certificate into structured `X509RequestContent`. **Subject + SAN only — no extensions are seeded** (issued-certificate extensions are largely CA-populated and must not be echoed back as requested extensions; this narrows Task 5b's plan, which listed subject/SAN/**extensions**). Reuse `CertificateUtil` / the Phase-1 extraction helpers — do **not** re-implement BouncyCastle parsing. Exhaustive unit tests.
- **Validation at the orchestrator boundary.** The seeded content is validated by the canonical `CertificateRequestContentValidator` (→ `CertificateRequestValidationException`) at the renew orchestrator, **not** inside the pure seeder (per Task 5b).
- **Route `renewCertificateAction` through the adapter factory.** `ClientOperationServiceImpl.renewCertificateAction` currently bypasses the adapter and calls the v2 connector client directly. Convert it to dispatch via `adapterFactory.forAuthority(authority)` → `AdapterOperationResult`, mirroring the issuance conversion in #1712 / PR #1751 (`finalizeIssuance` / `failClaimedCertificate` / state-divergence guard). The v2 adapter path renews exactly as today.
- **Populate `requestContent` in `AuthorityProviderV3Adapter.renew`**, gated on `FeatureFlag.CERTIFICATE_REQUEST_STRUCTURED` — mirroring `AuthorityProviderV3Adapter.register()`, which already applies the same gate and sets the field. When the connector advertises structured content, seed via the kernel and set `wire.setRequestContent(...)`; otherwise leave it unset and rely on the CSR the wire already carries. Renew currently sets neither the content nor the gate.

## Design note — extension seeding (resolved 2026-07-11: deferred)

**Resolution (2026-07-11).** Direction from the CEO: extension seeding is **deferred entirely** for now — renew seeds **subject DN + SAN from the existing certificate**, exactly as this issue's scope already states. This supersedes the 2026-07-10 feedback ("seed the same subset of attributes, including extensions, as the original"). The analysis below stays as the reference for whenever extension seeding is picked up again. One implementation edge to decide explicitly: the renew wire carries the stored original CSR next to the seeded `requestContent`; if the CA rewrote the DN at issuance the two differ, and structured content is authoritative per the design — accept that, or assert equality and warn on mismatch.

**What the plan states.** Task 5b of the register dual-wire plan (`request-attributes-design-and-plans/core/2026-06-01-request-attributes-phase4-core-register-dualwire.md`, §7 renew content-seeding) seeds subject **and SAN and extensions** from the existing certificate. Its extension step is a blanket copy — the kernel skeleton's `seedExtensions(X509Certificate)` carries the comment *"copy non-SAN extensions -> RequestedExtension (encoding=DER)"* — with no allow-list and no exclusion logic. The only extension "blocklist" anywhere in the plan set is unrelated: Task 3b (phase-1 plan, `X509RequestContentRenderer.effectiveCritical`) forces `KeyUsage (2.5.29.15)` and `BasicConstraints (2.5.29.19)` to stay **critical**; it overrides criticality *upward only*, decides nothing about whether an extension is carried, and lives on the PKCS#10 (Mode A) build path, not the renew seeder.

**What is wrong.** A blanket copy of an *issued* certificate's extensions echoes back extensions that are the CA's to populate, not the requester's: Authority Key Identifier and Subject Key Identifier, CRL Distribution Points, Authority Info Access, Signed Certificate Timestamps (CT), and the CA-finalized Basic Constraints. These are outputs of issuance, not inputs to a request. The plan never addressed which extensions are safe to carry on renew.

**Consequences if implemented naively.** Requesting the old AKI/SKI pins key/issuer identifiers that no longer apply (SKI is derived from the public key and is wrong under rekey; AKI is the issuer's). CRL DP and AIA would pin the renewed certificate to the *old* CA's distribution/OCSP URLs, overriding the new issuer's defaults. SCTs are per-issuance artifacts and are meaningless — or invalid — as a request input. A copied Basic Constraints could assert CA-ness incorrectly. Depending on the connector/CA, the outcome ranges from silent identity drift (renewed certificate diverges from what the operator intended) to outright issuance rejection.

**Recommendations.**
1. **This slice: seed subject + SAN only, no extensions** — sidesteps the defect entirely (SAN is the one request-controlled identity dimension beyond the subject).
2. **If extensions are wanted later, use a positive allow-list of request-controlled OIDs** (realistically `KeyUsage 2.5.29.15` and `ExtendedKeyUsage 2.5.29.37`; SAN is already handled on its own path) — never a blanket copy.
3. **Never carry key- or issuer-derived extensions** (SKI/AKI); let the CA recompute them at issuance.
4. **Fix Task 5b in the plan** to replace the blanket non-SAN copy with the allow-list approach before any future extension-seeding work is scheduled, so the plan and the shipped behavior do not diverge.

## User Story

As a certificate manager, I want a renewed certificate to keep the identity (subject and SAN) of the certificate it replaces, delivered as typed structured content to connectors that support it, so structured-capable authorities can renew without re-parsing a CSR and the renewed certificate cannot silently change its subject or SAN.

## Use Case

An operator renews a certificate on a v3 authority whose connector advertises structured request content. The platform seeds subject RDNs and SANs from the existing certificate and sends them as `requestContent` on the v3 renew wire. Against a connector without the structured capability, the same renew proceeds on the CSR/flat path. Non-v3 (v2 adapter) authorities renew exactly as today.

## Acceptance Criteria

- [ ] Pure seeder kernel with exhaustive unit tests: subject RDNs + SANs extracted from an existing certificate into `X509RequestContent`; no extensions seeded; edge cases (empty SAN, multi-valued RDNs, IP/DNS/email SAN types).
- [ ] `renewCertificateAction` dispatches through `adapterFactory.forAuthority(...)`; v2 adapter renew path regression-green.
- [ ] The v3 renew branch sets `requestContent` iff the connector advertises `CERTIFICATE_REQUEST_STRUCTURED`, using the same gate as register; otherwise leaves it unset.
- [ ] Seeded content is validated by `CertificateRequestContentValidator` at the orchestrator boundary (not inside the seeder).
- [ ] Integration test asserts a v3 renew against a structured-capable connector sends a **populated** `requestContent` (subject + SAN) on `CertificateRenewRequestDtoV3` (the wire input the otpki renew leg consumes), and that a non-capable connector receives no `requestContent`.
- [ ] New-code coverage ≥80% (line+branch, SonarCloud parity); Sonar gate clean.

## Dependencies

- #1633 — register dual-wire (`RegisterWireBuilder`, capability gate, renderer call sites); the v3 renew branch mirrors its capability-gate + set-`requestContent` pattern.
- #1712 (PR #1751) — issuance adapter-dispatch unification; the `renewCertificateAction` → `adapterFactory` conversion follows the same template. If #1751 has not merged, this slice performs the equivalent renew-side conversion itself.
- interfaces#732 — `requestContent` on `CertificateRenewRequestDtoV3` (merged).
- Design reference: Task 5b in `request-attributes-design-and-plans/core/2026-06-01-request-attributes-phase4-core-register-dualwire.md` (§7 renew content-seeding). This issue narrows that task per the shipped-code reality above.

## Downstream — sequencing note (not an acceptance criterion)

OmniTrustILM/otpki-connector#3, child *"Route structured `requestContent` (Mode C/E, flat fallback)"*, contains a **renew leg**: when `CertificateRenewRequestDtoV3.requestContent` is present, the connector sets the typed subject/SAN on the OTPKI `EndEntity` renewal template. That plan's JIT note instructs the executor to **drop the renew leg** if nothing populates `requestContent` on renew — and this issue is what populates it. Sequence: land this issue before that otpki child starts, or expect the leg to be dropped there and re-added afterwards. Whoever picks up that child should check this issue's state first.

klaraf755:
Closes #1706
Refs #1688 — its first Definition-of-Done item; the issue stays open for the spec/plan reconciliation item.

Carries the identity of the certificate being replaced — subject DN and SAN — into renew and rekey: as typed `requestContent` on the v3 renew wire when the connector advertises structured content, and as a real SAN extension block in the rekey CSR the platform builds itself.

No `interfaces` change: `CertificateRenewRequestDtoV3.requestContent` and the `FeatureFlag` values already ship.

## What was missing

`AuthorityProviderV3Adapter.renew` set the CSR, the existing certificate, meta and the three attribute sets, and never touched `requestContent` or the capability gate — so a structured-capable connector had to re-parse the CSR on renew, and the otpki renew leg had nothing to consume.

Separately, `createRequestFromKeys` passed `null` for the projected `Extensions` when regenerating a rekey CSR from a key. SAN is itself an X.509 extension (`2.5.29.17`) and can only ride in the CSR's extension block, so every re-keyed certificate lost its SAN outright — v2 and v3 alike, structured or not.

## The kernel is an entry point, not a new class

`X509RequestContentParser` already decodes RDNs, recovers an `otherName`'s OID and encoding, decodes IP octets, and reports SAN kinds `GeneralNameType` cannot model. It gains `parse(X509Certificate)`; only the two inputs are new — the subject from `getSubjectX500Principal()`, and the SAN extension read as raw DER — and the per-name loop is now shared by both sources. Reading raw DER rather than `getSubjectAlternativeNames()` is deliberate: an exotic kind then reaches the loop that reports it, instead of being rejected before parsing.

Subject and SAN only. `keyUsage` and `extendedKeyUsage` stay unseeded even though the content model now carries them as typed fields: in an issued certificate both are commonly set by the CA's certificate profile, so echoing them back as a request input can contradict the profile that will issue the successor. Extension seeding stays deferred per direction of 2026-07-11; whenever it is picked up it needs a positive allow-list of request-controlled OIDs, never a blanket copy of an issued certificate's extensions — that would echo back CA-owned AKI/SKI, CRL DP, AIA, SCTs and a CA-finalized Basic Constraints.

## Seeding lives in the adapter, and that is what covers rekey

`rekeyCertificateAction` reaches the connector through `adapter.renew(oldCert, newCert, null)` — rekey is a renew at the authority. So one call site inside the v3 adapter covers both legs, behind the same `CERTIFICATE_REQUEST_STRUCTURED` gate `register` uses, with no change to `AuthorityProviderAdapter.renew`'s signature and none to the v2 adapter. Orchestrator-side seeding, as the phase-4 plan sketched it, would have had to be wired twice — once with a null request DTO — and would compute an identity v2 authorities cannot use.

## Where the seeded identity comes from

The predecessor certificate, except when the operator supplied a CSR on renew — then that CSR.

Renew validates only the public key against the predecessor (`validatePublicKeyForCsrAndCertificate(..., true)`); unlike rekey it does not call `validateSubjectDnForCertificate`, so a supplied renew CSR may legitimately carry a different DN or SAN. Seeding from the old certificate there would let structured content silently override what the operator asked for on any structured-capable authority. A reuse-CSR renew and a rekey both take the certificate instead — which is where a CA-rewritten DN travels.

## An identity that cannot be represented in full is never sent reduced

`x400Address` and `ediPartyName` have no `GeneralNameType` counterpart — both are structured ASN.1 with no canonical string form, and `GeneralNameEntry` carries a single string — and a modelled kind can still fail to decode (an `iPAddress` that is neither 4 nor 16 octets). Either way the parser reports the kind rather than dropping it.

- **Renew** then sends no `requestContent` at all, and the connector parses the CSR — today's behaviour exactly, so no regression, and never a structured identity narrower than the certificate it replaces.
- **Rekey** fails closed with a `ValidationException`. The platform builds that CSR itself, so nothing downstream could recover a SAN dropped there.

## No separate validation pass over seeded content

#1706 named `CertificateRequestContentValidator`. That validator checks content against resolved request-attribute definitions under a whitelist policy — it is the gate for externally supplied requests (ACME/SCEP/CMP and uploaded CSRs), not a well-formedness check. Running it here would whitelist-check a *historical* identity against today's profile, so renewing a certificate whose DN predates or exceeds the current request-attribute set would start failing.

Operator input is already validated at its own boundary: `submitCertificateRequest` → `generateBase64EncodedCsr(request.getRequest(), …)` → `validateUploadedRequestAttributes`, which renew and rekey both reach through `submitAndShapeFailure` before any action runs. What remains is platform-derived from an issued certificate, and the model's invariants hold by construction — blank RDNs are skipped, an `otherName` always carries OID and encoding.

Task 5b of the phase-4 plan also had the seeder default every field to `readOnly` and un-set it per `FieldKind`. The shipped model carries no `readOnly` field anywhere, so that dimension is retired rather than deferred.

## Tests

- **Certificate source:** RDN order, repeated OUs, every typed SAN kind, a UPN `otherName` recovering OID + `UTF8_STRING`, a certificate with no SAN, `x400Address` reported rather than dropped, and an EKU-bearing certificate asserting nothing lands in `extensions` / `keyUsage` / `extendedKeyUsage`.
- **Round-trip:** seeded content → `X509RequestContentRenderer.toExtensions` → re-parse equals the original. The seeder is the renderer's inverse and the two must not drift.
- **Seeder:** each source case including rekey's null DTO, drop-on-unrepresentable, and an unparseable predecessor.
- **Adapter:** content populated when the capability is advertised and null when it is not; seeded from the supplied CSR; seeded from the predecessor on rekey.
- **Integration:** the renew wire carries subject + SAN against a structured-capable connector and nothing against one without the flag; the rekey CSR carries the predecessor's SAN. Both were first checked against a deliberately broken baseline — the rekey one with the call site reverted to `null`, the wire one with the capability dropped — so neither passes vacuously.

New-code coverage: `RenewContentSeeder` 88% line / 100% branch, `X509RequestContentParser` 96% / 90%. All four CI profiles green locally, `ContextSignatureGuardTest.BASELINE` unchanged at 61.

## Unblocks

OmniTrustILM/otpki-connector#5's renew leg, which was instructed to drop that leg for as long as nothing populated `requestContent` on renew.

### DIFF
--- a/src/main/java/com/otilm/core/certificate/request/RenewContentSeeder.java
+++ b/src/main/java/com/otilm/core/certificate/request/RenewContentSeeder.java
@@ -1,0 +1,86 @@
+package com.otilm.core.certificate.request;
+
+import com.otilm.api.exception.CertificateRequestException;
+import com.otilm.api.exception.ValidationException;
+import com.otilm.api.model.connector.v3.certificate.X509RequestContent;
+import com.otilm.api.model.core.v2.ClientCertificateRenewRequestDto;
+import com.otilm.core.dao.entity.Certificate;
+import com.otilm.core.dao.entity.CertificateRequestEntity;
+import com.otilm.core.model.request.CertificateRequest;
+import com.otilm.core.util.CertificateRequestUtils;
+import com.otilm.core.util.CertificateUtil;
+import com.otilm.core.util.X509RequestContentRenderer;
+import java.io.IOException;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
+import org.bouncycastle.asn1.x509.Extensions;
+
+/**
+ * Seeds the identity a successor inherits from the certificate it replaces — subject DN and SAN, no extensions — for
+ * the structured renew wire and for the rekey CSR the platform builds itself.
+ */
+@Slf4j
+public final class RenewContentSeeder {
+
+    private RenewContentSeeder() {
+    }
+
+    /**
+     * The typed identity for the structured renew wire: the CSR the operator supplied when there is one, and otherwise
+     * the predecessor certificate. Empty when that identity cannot be carried in full, in which case the caller leaves
+     * {@code requestContent} unset and the connector parses the CSR the wire already carries.
+     */
+    public static Optional<X509RequestContent> seed(Certificate oldCertificate, Certificate newCertificate,
+            ClientCertificateRenewRequestDto request) {
+        try {
+            ParsedRequestContent parsed = operatorSuppliedCsr(newCertificate, request)
+                    ? X509RequestContentParser.parse(storedCsr(newCertificate))
+                    : X509RequestContentParser.parse(storedCertificate(oldCertificate));
+            if (!parsed.unsupportedSans().isEmpty()) {
+                log
+                        .warn("Not seeding structured renew content: subject alternative name {} has no typed representation",
+                                parsed.unsupportedSans());
+                return Optional.empty();
+            }
+            return Optional.of(parsed.content());
+        } catch (CertificateException | CertificateRequestException | RuntimeException e) {
+            log.warn("Not seeding structured renew content: the identity to carry forward could not be read", e);
+            return Optional.empty();
+        }
+    }
+
+    /**
+     * The SAN extension block a rekey CSR inherits, or null when the predecessor carries no SAN. Fails closed: the
+     * platform builds this CSR itself, so nothing downstream can recover a SAN dropped here.
+     */
+    public static Extensions rekeySanExtensions(X509Certificate oldCertificate) {
+        ParsedRequestContent parsed = X509RequestContentParser.parse(oldCertificate);
+        if (!parsed.unsupportedSans().isEmpty()) {
+            throw new ValidationException(
+                    "The certificate carries a subject alternative name the platform cannot re-request: %s"
+                            .formatted(String.join(", ", parsed.unsupportedSans())));
+        }
+        try {
+            return X509RequestContentRenderer.toExtensions(parsed.content());
+        } catch (IOException e) {
+            throw new ValidationException(
+                    "Failed to build the subject alternative name of the re-keyed request. Error: " + e.getMessage());
+        }
+    }
+
+    private static boolean operatorSuppliedCsr(Certificate newCertificate, ClientCertificateRenewRequestDto request) {
+        return request != null && request.getRequest() != null && newCertificate.getCertificateRequest() != null;
+    }
+
+    private static CertificateRequest storedCsr(Certificate certificate) throws CertificateRequestException {
+        CertificateRequestEntity stored = certificate.getCertificateRequest();
+        return CertificateRequestUtils
+                .createCertificateRequest(stored.getContent(), stored.getCertificateRequestFormat());
+    }
+
+    private static X509Certificate storedCertificate(Certificate certificate) throws CertificateException {
+        return CertificateUtil.parseCertificate(certificate.getCertificateContent().getContent());
+    }
+}
--- a/src/main/java/com/otilm/core/certificate/request/RenewContentSeeder.java
+++ b/src/main/java/com/otilm/core/certificate/request/RenewContentSeeder.java
@@ -13,16 +13,18 @@ import com.otilm.core.util.X509RequestContentRenderer;
 import java.io.IOException;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
-import java.util.List;
 import java.util.Optional;
+import java.util.stream.Stream;
 import lombok.extern.slf4j.Slf4j;
 import org.bouncycastle.asn1.x500.RDN;
 import org.bouncycastle.asn1.x500.X500Name;
 import org.bouncycastle.asn1.x509.Extensions;
 
 /**
- * Seeds the identity a successor inherits from the certificate it replaces — subject DN and SAN, no extensions — for
- * the structured renew wire and for the rekey CSR the platform builds itself.
+ * Seeds what a successor inherits, for the structured renew wire and for the rekey CSR the platform builds itself. A
+ * predecessor certificate yields its subject DN and SAN; the extensions of an issued certificate are the CA's and are
+ * never seeded. A CSR source yields what the operator requested in it, extensions included, because the content is
+ * authoritative for a structured connector.
  */
 @Slf4j
 public final class RenewContentSeeder {
@@ -54,15 +56,20 @@ public final class RenewContentSeeder {
                         .warn("Not seeding structured renew content: the subject packs a multi-valued RDN, which typed content cannot express");
                 return Optional.empty();
             }
-            if (!parsed.unsupportedSans().isEmpty()) {
+            if (!parsed.unsupportedSans().isEmpty() || !parsed.unrepresentableExtensionValues().isEmpty()) {
+                // The content is authoritative for the connector, so a partial one narrows what was requested.
+                // Both lists hold what the typed model could not carry; either one means send nothing.
                 log
-                        .warn("Not seeding structured renew content: subject alternative name {} has no typed representation",
-                                parsed.unsupportedSans());
+                        .warn("Not seeding structured renew content: {} has no typed representation",
+                                Stream
+                                        .concat(parsed.unsupportedSans().stream(),
+                                                parsed.unrepresentableExtensionValues().stream())
+                                        .toList());
                 return Optional.empty();
             }
-            X509RequestContent content = identityOnly(parsed.content());
-            if (isEmpty(content.getSubject()) && isEmpty(content.getSubjectAltNames())) {
-                log.warn("Not seeding structured renew content: no subject or subject alternative name to carry");
+            X509RequestContent content = parsed.content();
+            if (!content.isRequestContentProvided()) {
+                log.warn("Not seeding structured renew content: the source carries no identity to send");
                 return Optional.empty();
             }
             return Optional.of(content);
@@ -81,9 +88,11 @@ public final class RenewContentSeeder {
         try {
             parsed = X509RequestContentParser.parse(oldCertificate);
         } catch (RuntimeException e) {
+            // Malformed ASN.1 surfaces as unchecked BouncyCastle exceptions whose messages may carry internals,
+            // and this one reaches the client as a 422 body; log it and answer with fixed wording.
+            log.warn("The certificate's identity could not be decoded for a re-keyed request", e);
             throw new ValidationException(
-                    "The certificate's identity could not be decoded, so it cannot be carried into a re-keyed request. Error: "
-                            + e.getMessage());
+                    "The certificate's identity could not be decoded, so it cannot be carried into a re-keyed request");
         }
         if (!parsed.unsupportedSans().isEmpty()) {
             throw new ValidationException(
@@ -93,8 +102,8 @@ public final class RenewContentSeeder {
         try {
             return X509RequestContentRenderer.toExtensions(parsed.content());
         } catch (IOException e) {
-            throw new ValidationException(
-                    "Failed to build the subject alternative name of the re-keyed request. Error: " + e.getMessage());
+            log.warn("Failed to encode the subject alternative name of a re-keyed request", e);
+            throw new ValidationException("Failed to build the subject alternative name of the re-keyed request");
         }
     }
 
@@ -111,23 +120,6 @@ public final class RenewContentSeeder {
             }
         }
         return false;
-    }
-
-    /**
-     * Narrows parsed content to the identity the renew wire carries. A CSR source also yields key usage, extended key
-     * usage and requested extensions; those stay on the CSR travelling beside the content, so both sources put the same
-     * shape on the wire and nothing extension-shaped can reach it reduced.
-     */
-    private static X509RequestContent identityOnly(X509RequestContent content) {
-        content.setKeyUsage(null);
-        content.setExtendedKeyUsage(null);
-        content.setExtensions(null);
-        return content;
-    }
-
-    /** Content with neither dimension would fail the wire model's own "something must be provided" assertion. */
-    private static boolean isEmpty(List<?> values) {
-        return values == null || values.isEmpty();
     }
 
     private static boolean operatorSuppliedCsr(Certificate newCertificate, ClientCertificateRenewRequestDto request) {
--- a/src/main/java/com/otilm/core/certificate/request/RenewContentSeeder.java
+++ b/src/main/java/com/otilm/core/certificate/request/RenewContentSeeder.java
@@ -13,6 +13,7 @@ import com.otilm.core.util.X509RequestContentRenderer;
 import java.io.IOException;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
+import java.util.List;
 import java.util.Optional;
 import lombok.extern.slf4j.Slf4j;
 import org.bouncycastle.asn1.x509.Extensions;
@@ -44,7 +45,12 @@ public final class RenewContentSeeder {
                                 parsed.unsupportedSans());
                 return Optional.empty();
             }
-            return Optional.of(parsed.content());
+            X509RequestContent content = identityOnly(parsed.content());
+            if (isEmpty(content.getSubject()) && isEmpty(content.getSubjectAltNames())) {
+                log.warn("Not seeding structured renew content: no subject or subject alternative name to carry");
+                return Optional.empty();
+            }
+            return Optional.of(content);
         } catch (CertificateException | CertificateRequestException | RuntimeException e) {
             log.warn("Not seeding structured renew content: the identity to carry forward could not be read", e);
             return Optional.empty();
@@ -56,7 +62,14 @@ public final class RenewContentSeeder {
      * platform builds this CSR itself, so nothing downstream can recover a SAN dropped here.
      */
     public static Extensions rekeySanExtensions(X509Certificate oldCertificate) {
-        ParsedRequestContent parsed = X509RequestContentParser.parse(oldCertificate);
+        ParsedRequestContent parsed;
+        try {
+            parsed = X509RequestContentParser.parse(oldCertificate);
+        } catch (RuntimeException e) {
+            throw new ValidationException(
+                    "The certificate's identity could not be decoded, so it cannot be carried into a re-keyed request. Error: "
+                            + e.getMessage());
+        }
         if (!parsed.unsupportedSans().isEmpty()) {
             throw new ValidationException(
                     "The certificate carries a subject alternative name the platform cannot re-request: %s"
@@ -68,6 +81,23 @@ public final class RenewContentSeeder {
             throw new ValidationException(
                     "Failed to build the subject alternative name of the re-keyed request. Error: " + e.getMessage());
         }
+    }
+
+    /**
+     * Narrows parsed content to the identity the renew wire carries. A CSR source also yields key usage, extended key
+     * usage and requested extensions; those stay on the CSR travelling beside the content, so both sources put the same
+     * shape on the wire and nothing extension-shaped can reach it reduced.
+     */
+    private static X509RequestContent identityOnly(X509RequestContent content) {
+        content.setKeyUsage(null);
+        content.setExtendedKeyUsage(null);
+        content.setExtensions(null);
+        return content;
+    }
+
+    /** Content with neither dimension would fail the wire model's own "something must be provided" assertion. */
+    private static boolean isEmpty(List<?> values) {
+        return values == null || values.isEmpty();
     }
 
     private static boolean operatorSuppliedCsr(Certificate newCertificate, ClientCertificateRenewRequestDto request) {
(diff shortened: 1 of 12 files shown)


src/main/java/com/otilm/core/certificate/request/X509RequestContentParser.java
```diff
@@ -63,6 +65,19 @@ public static ParsedRequestContent parse(CertificateRequest request) {
...
+    /**
+     * Parses an issued certificate's identity into typed content: subject RDNs and SAN entries, no extensions. SAN
+     * kinds {@link GeneralNameType} cannot model are reported in {@link ParsedRequestContent#unsupportedSans()}.
+     */
+    public static ParsedRequestContent parse(X509Certificate certificate) {
+        X509RequestContent x509 = new X509RequestContent();
+        x509.setCertificateType(CertificateType.X509);
+        x509.setSubject(parseSubject(X500Name.getInstance(certificate.getSubjectX500Principal().getEncoded())));
+        List<String> unsupportedSans = new ArrayList<>();
+        x509.setSubjectAltNames(parseSans(certificateSans(certificate), unsupportedSans));
```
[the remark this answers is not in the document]
klaraf755: Fixed in 026641d. Both halves were wrong: the parser labelled every `ASN1String` as `UTF8_STRING`, and the renderer had no `PRINTABLE_STRING` branch either. UTF8/IA5/Printable now each keep their own encoding, and any other string type travels as Base64(DER), which `decodeDerOrFallback` restores as its original primitive. Tests cover the per-type encodings, BMPString falling back to DER, and a four-way render/re-parse round-trip.

The lossiness predates this PR — harmless while the parser only fed validation, real once the rekey leg re-encodes its output.

src/main/java/com/otilm/core/certificate/request/RenewContentSeeder.java
```diff
@@ -0,0 +1,86 @@
...
+        try {
+            ParsedRequestContent parsed = operatorSuppliedCsr(newCertificate, request)
+                    ? X509RequestContentParser.parse(storedCsr(newCertificate))
+                    : X509RequestContentParser.parse(storedCertificate(oldCertificate));
+            if (!parsed.unsupportedSans().isEmpty()) {
+                log
+                        .warn("Not seeding structured renew content: subject alternative name {} has no typed representation",
+                                parsed.unsupportedSans());
+                return Optional.empty();
+            }
```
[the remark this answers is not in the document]
klaraf755: Fixed in 026641d, and it pointed at more than the guard: `parse(CertificateRequest)` also fills `keyUsage`/`extendedKeyUsage`/`extensions`, so the supplied-CSR branch was putting extensions on the wire while the certificate branch was not.

Both sources are now narrowed to subject + SAN, which removes the reduction path rather than detecting it — nothing extension-shaped is carried at all. The CSR still travels beside the content and stays authoritative for the rest. Narrowing also needed a guard for content left with neither subject nor SAN, which would fail `isRequestContentProvided()`.

Your suppressed comment on line 59 was right too, same commit: `rekeySanExtensions` now converts a parser failure into a `ValidationException`, while renew seeds nothing and falls back to the CSR.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
```diff
@@ -131,6 +132,9 @@ public AdapterOperationResult renew(Certificate oldCert, Certificate newCert, Cl
             wire.setFormat(newCert.getCertificateRequest().getCertificateRequestFormat());
         }
         wire.setExistingCertificate(oldCert.getCertificateContent().getContent());
+        if (capabilityService.supports(authority, FeatureFlag.CERTIFICATE_REQUEST_STRUCTURED)) {
+            RenewContentSeeder.seed(oldCert, newCert, req).ifPresent(wire::setRequestContent);
```
lubomirw, line 136: **[High]** | Correctness | Confidence: High

An operator-supplied rekey CSR can carry a different or absent SAN, but `rekeyCertificateAction` calls `adapter.renew(..., null)`. `RenewContentSeeder` therefore treats the request as platform-generated and sends the predecessor SAN as authoritative `requestContent`, overriding the CSR for structured-capable authorities.

_Suggested:_ Propagate whether rekey used an uploaded CSR and seed from the successor CSR in that case; use the predecessor only for platform-generated CSRs. Add integration coverage for a differing or absent SAN.

<sub>flagged by Guided PR review and Superpowers; verified against the rekey action and subject-only CSR validation</sub>

src/main/java/com/otilm/core/certificate/request/RenewContentSeeder.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+
+    /**
+     * Narrows parsed content to the identity the renew wire carries. A CSR source also yields key usage, extended key
+     * usage and requested extensions; those stay on the CSR travelling beside the content, so both sources put the same
+     * shape on the wire and nothing extension-shaped can reach it reduced.
+     */
+    private static X509RequestContent identityOnly(X509RequestContent content) {
+        content.setKeyUsage(null);
```
lubomirw: **[Medium]** | Correctness | Confidence: High

The seeder clears key usage, extended key usage, and requested extensions even though `CertificateRenewRequestDtoV3` and the connector documentation define `requestContent` as authoritative for both subject identity and extensions. A conforming structured connector can therefore omit extension intent that remains only in the CSR.

_Suggested:_ Define and implement field-level fallback semantics: structured subject/SAN are authoritative, while omitted extension fields come from the CSR. Update the interfaces contract, connector behavior, and documentation together.

<sub>flagged by Copilot CLI and Guided PR review; verified against the DTO schema and structured-content documentation</sub>

src/main/java/com/otilm/core/certificate/request/X509RequestContentParser.java
```diff
@@ -63,6 +67,19 @@ public static ParsedRequestContent parse(CertificateRequest request) {
...
     }
 
+    /**
+     * Parses an issued certificate's identity into typed content: subject RDNs and SAN entries, no extensions. SAN
+     * kinds {@link GeneralNameType} cannot model are reported in {@link ParsedRequestContent#unsupportedSans()}.
+     */
+    public static ParsedRequestContent parse(X509Certificate certificate) {
+        X509RequestContent x509 = new X509RequestContent();
+        x509.setCertificateType(CertificateType.X509);
+        x509.setSubject(parseSubject(X500Name.getInstance(certificate.getSubjectX500Principal().getEncoded())));
```
lubomirw, line 77: **[Medium]** | Correctness | Confidence: High

A compound RDN such as `CN=host+O=Acme` is flattened into independent `RdnEntry` values. The renderer rebuilds each entry with a separate `addRDN`, so authoritative renewal content changes the X.500 subject semantics.

_Suggested:_ Detect compound RDNs and omit `requestContent` so the CSR preserves the exact DN, or extend the wire model to retain RDN grouping. Add a certificate test built with `addMultiValuedRDN`.

<sub>flagged by Superpowers; verified against the parser's RDN loop and the renderer's one-entry-per-RDN behavior</sub>

src/main/java/com/otilm/core/certificate/request/RenewContentSeeder.java
```diff
@@ -0,0 +1,116 @@
...
+     * platform builds this CSR itself, so nothing downstream can recover a SAN dropped here.
+     */
+    public static Extensions rekeySanExtensions(X509Certificate oldCertificate) {
+        ParsedRequestContent parsed;
+        try {
+            parsed = X509RequestContentParser.parse(oldCertificate);
+        } catch (RuntimeException e) {
+            throw new ValidationException(
+                    "The certificate's identity could not be decoded, so it cannot be carried into a re-keyed request. Error: "
+                            + e.getMessage());
```
lubomirw: **[Low]** | Security | Confidence: High

Raw Bouncy Castle/parser exception text is appended to `ValidationException`; `ExceptionHandlingAdvice` returns that text directly in the HTTP 422 body. The same issue exists in the rendering failure branch.

_Suggested:_ Log the original exceptions server-side and return fixed client-safe validation messages from both catch blocks.

<sub>flagged by Copilot CLI, Guided PR review, and Superpowers; verified against the HTTP validation-exception handler and nearby info-leak safeguards</sub>

src/test/java/com/otilm/core/certificate/request/RenewContentSeederTest.java
```diff
@@ -0,0 +1,324 @@
...
+
+    @Test
+    void rekeySanExtensions_returnsNull_whenThePredecessorHasNoSan() throws Exception {
+        // given
+        X509Certificate oldCertificate = CertificateTestUtil.createCertificateWithSubjectAndSans("CN=old.example.com");
+
+        // when
+        Extensions extensions = RenewContentSeeder.rekeySanExtensions(oldCertificate);
+
+        // then — the CSR is built exactly as before this change for a certificate with no SAN
```
lubomirw: **[Low]** | Maintainability | Confidence: High

The phrase "exactly as before this change" records implementation history instead of describing the current contract, so the comment can become misleading as the code evolves.

_Suggested:_ Reword it as current-state behavior, for example: "A predecessor without SAN produces no extension block."

<sub>flagged by Hygiene; verified on the added test line</sub>

lubomirw (changes requested):
Automated multi-reviewer triage - 5 findings (REQUEST_CHANGES).

src/main/java/com/otilm/core/certificate/request/X509RequestContentParser.java
lubomirw: This is a distinct correctness issue from the structured-content fallback findings. The current model stores subject entries as a flat ordered list, so a multi-valued RDN like `CN=host+O=Acme` cannot round-trip: parsing splits it into separate `RdnEntry` values, and rendering rebuilds them as separate RDNs.

I’d treat this as requiring a fail-closed guard for now: detect compound RDNs when seeding authoritative `requestContent` and omit/reject that path until the wire model can preserve RDN grouping. Add a regression test using a subject built with `addMultiValuedRDN`.

src/main/java/com/otilm/core/service/handler/authority/AuthorityProviderV3Adapter.java
klaraf755: Fixed in b1ae645. Confirmed: the uploaded-CSR branch validates the public key and subject DN against the predecessor but not the SAN, so the operator's SAN could differ and was being overridden.

`rekeyCertificateAction` now passes a renew DTO carrying the uploaded CSR instead of `null`, so the existing seed rule applies unchanged. Safe because the v2 adapter ignores `req` entirely and the v3 adapter reads it only for this decision. Integration test added; with the old `null` it fails on the predecessor's SAN, so it is not passing vacuously.

src/main/java/com/otilm/core/certificate/request/RenewContentSeeder.java
klaraf755: Fixed in b1ae645 — both catch blocks now log the exception and answer with fixed wording.

src/test/java/com/otilm/core/certificate/request/RenewContentSeederTest.java
klaraf755: Fixed in b1ae645 — reworded to "a predecessor without SAN produces no extension block".

src/main/java/com/otilm/core/certificate/request/X509RequestContentParser.java
klaraf755: Guard implemented in 8f2b4d8, on both seed sources: `hasMultiValuedRdn` over the source subject, and no `requestContent` when it matches. Regression test uses `addMultiValuedRDN`, and a negative control confirmed both new tests fail without the guard.

Confirmed empirically before implementing: the two forms differ in DER, in `X500Name.equals`, in `X500Principal.equals`, and in this platform's own normalized-string comparison (`PlatformX500NameStyle.appendRDN` joins a multi-valued RDN with `+`). Rekey's CSR was never affected, since `createRequestFromKeys` passes the certificate's own `X500Principal` through.

One clarification for anyone reading #1688 alongside this: "multi-valued RDN" there means repeated attribute types such as two OUs, which are separate RDNs and round-trip fine. The compound `+` form is a different thing. I renamed a test of mine in 28bae206 that had blurred the two.

src/main/java/com/otilm/core/certificate/request/RenewContentSeeder.java
klaraf755: Changed in b1ae645: content seeded from a CSR now keeps the key usage, extended key usage and extensions requested in it. Because the content is authoritative, a partial one is never sent — both `unsupportedSans` and `unrepresentableExtensionValues` gate it, so a CSR requesting a key usage bit the model cannot name sends nothing rather than a narrowed usage.

Not closed by that: content seeded from the *predecessor certificate* still carries subject + SAN only, so on a reuse-CSR renew a conforming connector reads "no extensions requested". Choosing a source cannot fix it — the CSR states what was requested, possibly overridden by the CA already, and the certificate states what was issued, which is CA-owned and why seeding it is deferred. The field-level fallback you suggested is the real answer, and it is already expressible: `@JsonInclude(NON_NULL)` means an absent field is "not stated" while `[]` is "explicitly none", and this branch leaves those fields null rather than empty. Making that normative touches the schema descriptions and otpki-connector#5, so it sits outside this PR.

lubomirw (approved):

PR #2176, opened 2026-08-31
Serve the content-signing formatting attribute aggregate as a discovery endpoint

### LINKED ISSUE #2124: Content-signing formatting attribute discovery endpoint
### Description

The content-signing save path validates stored formatting attributes against the per-operation connector schemas (`ContentSigningFormattingAttributes.aggregate`, OmniTrustILM/core#1974), but no public endpoint exposes that aggregate, so a client cannot learn which attributes exist before submitting them. The existing `GET /v1/signingProfiles/signatureFormattingConnectors/{connectorUuid}/formattingAttributes` serves the legacy timestamping contract only, and its OpenAPI description now says so. OmniTrustILM/fe-administrator#1997 is already specified against the merged descriptor set, re-fetched when the operator changes the level maximum.

**The contract rule: discovery is the save-time aggregation, exposed read-only.** Same code, same inputs, same failure modes, so a form built from discovery always validates on save.

- New route on the signing-profile API, a sibling of the legacy one: `GET /v1/signingProfiles/signatureFormattingConnectors/{connectorUuid}/contentSigningFormattingAttributes?family=...&maxLevel=...`, returning the merged flat `List<BaseAttribute>`. The contract addition lands in `interfaces` (`SigningProfileController`). A second endpoint, not a parameter on the legacy route - mirroring the connector-contract split keeps each endpoint on one contract with no default to guess.
- `family` and `maxLevel` are **required** because both are form state, not stored state, at the only moments discovery is called: on create no profile exists, and on update the operator is mid-edit changing them. The save path has no default for either, so discovery must not invent one. `signingProfileUuid` stays optional and authorization-only, matching the legacy route.
- The answer is the save-time aggregate: the operations reachable under `maxLevel`, merged by name into one flat set with the definition-consistency guard. Discovery reuses the save predicates rather than forking them - a connector declaring one name with two definitions, a connector not advertising the family, or a `maxLevel` above the connector's declared per-rung ceiling for that family (DP-16) all fail discovery with the same errors profile save would produce, so the operator learns at form-build time, not at save.
- If the save-time aggregation ever grows an input, discovery grows it in lockstep.

### Definition of Done

- [ ] Endpoint implemented against the `interfaces` contract, wired to the same aggregation code the save path uses - no second merge implementation
- [ ] `family` and `maxLevel` required; name-collision, family and DP-16 ceiling violations rejected with the same errors as profile save
- [ ] Authorization and audit wiring match the legacy discovery route; `signingProfileUuid` authorization-only
- [ ] The legacy `formattingAttributes` route is left with no content-signing caller: until this endpoint exists, a content-signing caller of the legacy route gets a timestamping-shaped descriptor set (connector declaring both interfaces) or an opaque connector error (content-signing-only connector), and core is knowingly left that way rather than made to refuse - this endpoint plus OmniTrustILM/fe-administrator#1997 is what ends it

ivosh:
**TL;DR — this implements the discovery endpoint declared by OmniTrustILM/interfaces#933, so a content-signing profile form can fetch the formatting attributes it must fill in before the profile is saved.**

Closes #2124.

## What it does

`GET /v1/signingProfiles/signatureFormattingConnectors/{connectorUuid}/contentSigningFormattingAttributes?family=…&maxLevel=…`

Three steps:

1. Validate that the connector serves the requested family and reaches the requested ceiling.
2. Ask the connector for the formatting attributes of every operation that ceiling reaches.
3. Return them merged by name into one flat set, carrying the connector's default values.

## No second merge implementation

The endpoint reuses `ContentSigningFormattingAttributes.aggregate` — **the same call profile save uses.** The validation is the same too: I extracted `ContentSigningWorkflowValidator.requireConnectorServes` out of `validate`, so save and discovery share one copy of the connector-facing rules.

That is the point of the endpoint. A form built from a discovery answer cannot fail save on the connector's account, and it cannot drift, because there is no second code path to drift from.

## Two bugs this would have shipped with

**The query enums would not have bound.** `SignatureFamily` and `SignatureLevel` are `IPlatformEnum`s whose JSON form is the lowercase code (`pades`, `signed`) — which is what the published contract documents. Spring's default converter calls `Enum.valueOf`, so **every documented value would have answered 400.** Fixed with `SignatureFamilyConverter` and `SignatureLevelConverter`, registered in the existing `@InitBinder` beside `SigningWorkflowTypeConverter`.

I verified the fix rather than assuming it: with the two registrations removed, `SigningProfileFormattingDiscoveryITest` fails `expected:<404> but was:<400>`; with them in place it passes.

**The transaction would have spanned an HTTP call.** The method is `@Transactional(propagation = Propagation.NOT_SUPPORTED)`, so an unresponsive connector cannot hold a database connection for its whole response time. Safe because `getConnectorForApiClient` returns an immutable snapshot and the connector entity is fetched through an entity graph, so nothing lazy crosses the boundary.

I applied the same annotation to the pre-existing timestamping sibling `listSignatureFormattingConnectorAttributes`, which had the same problem. It had no test at all, so it has one now.

## Failure modes

- **404** — the connector does not exist. The optional signing profile UUID is used only for authorization and is never dereferenced by this route.
- **422** — the connector cannot serve the requested family and level.

The 422 message is the validator's own, which is the same message profile save returns. The rule list is deliberately not restated in the OpenAPI description, so it cannot go stale as levels and feature flags are added.

## Tests

- `SigningProfileFormattingDiscoveryITest` — 2 MockMvc tests pinning that both wire codes bind over HTTP. This reuses the existing `@AutoConfigureMockMvc` + `BaseSpringBootTest` signature, so it costs **no new Spring context**; `ContextSignatureGuardTest.BASELINE` is unchanged at 61.
- `SigningProfileServiceImplITest` — 5 tests: the round trip, the two refusals, one proving `maxLevel` actually widens the operations fetched, and the new coverage for the timestamping sibling.

Affected suites on this branch: **125 tests, 0 failures.** `spotless:check` and `checkstyle:check` clean.

## Merge order

**OmniTrustILM/interfaces#933 first, this second.**


src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -275,12 +277,26 @@ public List<BaseAttribute> listSignatureAttributesForCertificate(SecuredUUID cer
...
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    public List<BaseAttribute> listContentSigningFormattingConnectorAttributes(UUID connectorUuid,
+            SignatureFamily family, SignatureLevel maxLevel, SecuredUUID signingProfileUuid)
+            throws NotFoundException, ConnectorException, AttributeException {
+        ContentSigningWorkflowValidator
+                .requireConnectorServes(
+                        connectorService.getConnectorEntityWithInterfaces(SecuredUUID.fromUUID(connectorUuid)), family,
+                        maxLevel);
+        return contentSigningFormattingAttributes
+                .aggregate(connectorService.getConnectorForApiClient(connectorUuid), maxLevel);
```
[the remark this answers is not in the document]
ivosh: Not changed.

Premise confirmed on the mechanism. `getConnectorEntityWithInterfaces` and `getConnectorForApiClient` are two reads — `findWithInterfacesByUuid`, then `findByUuid` in `ConnectorServiceImpl` — so a connector edit can commit between them.

The parity claim is refuted. Profile save does the same two reads: it validates on `getConnectorEntityWithInterfaces` (`SigningProfileServiceImpl:946-951`) and aggregates on `getConnectorForApiClient` (`:1214-1217`). Discovery and save share one window, so they agree with each other.

The combined snapshot is not taken here. `ImmutableConnectorInfo` is a package-private record in `service/v2/impl`, so building one means either widening its visibility or adding a `ConnectorService` method, and it only holds if the save path changes with it — this PR does not touch save. The endpoint is read-only and advisory: save re-validates, and the discovery-to-save window a form is open for is wider than this one by orders of magnitude.

src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -275,12 +277,26 @@ public List<BaseAttribute> listSignatureAttributesForCertificate(SecuredUUID cer
 
     @Override
     @ExternalAuthorization(resource = Resource.SIGNING_PROFILE, action = ResourceAction.ANY)
-    @Transactional(readOnly = true)
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
```
vyskocilm, line 280: ✅ note, not a finding — this is the PR's one behaviour change to existing code, so worth stating plainly for the next reviewer: `readOnly` was *wrong* here, not merely suboptimal. `listSignatureFormattingConnectorAttributes` reaches an outbound connector HTTP call through `fetchFormattingAttributeDefinitions` (`:1224-1230`), so the old boundary held a database connection for the connector's entire response time.

The one way `NOT_SUPPORTED` could have broken it — entity access after the session closes — does not apply: `findWithInterfacesByUuid` carries `@EntityGraph(attributePaths = {"interfaces"})`, and `ConnectorInterfaceEntity.features` is a `@Column` SQL array (`ConnectorInterfaceEntity:49-52`), not a lazy association, so `declares(…)` cannot throw `LazyInitializationException` outside a transaction. `timestampingDiscovery_stillServesItsOwnDescriptors` covers it.

vyskocilm (commented):
**Checked and fine** — recorded because each is something a reviewer would otherwise have to take on trust:

- **The validator extraction preserves save behaviour.** After extraction the check order is identical (`requireFamily` → `requireMaxLevel` → interface resolution → `requireFlag` → `requireRung` → `requireExecutableMaxLevel` → `validateTimestampSource`), the messages are byte-identical, and `requireConnectorServes` re-runs two null guards on values already proven non-null, which is idempotent. The 20 `CreateTests` and 8 `UpdateTests` that drive save are green.
- **Nothing raw reaches the wire on the 422 path** — measured with a throwaway MockMvc probe, not reasoned: `family=not-a-family` → 422 `["Unknown signature family not-a-family"]`; `maxLevel=not-a-level` → 422 `["Unknown signature level not-a-level"]`; unknown connector → 404 `"Object of type 'Connector' identified by … is not found."`. Every body is a `ValidationException` shaped in-house, and the echoed text is either the caller's own submitted code or a connector display name already visible to that caller. No SQL fragment, no stack frame, no internal identifier.
- **The new action needs no authorization plumbing, and none was written.** It reaches the auth service from `@ExternalAuthorization` alone — no Rego, no permission seed, no migration anywhere in the diff.
- No migration, entity, native query, filter field, `@Value`, yml change or env-var override — so nothing owed to the Helm side either.

**The open thread on the two connector reads:** I checked the rebuttal against the code instead of taking it on trust. It stands — reply posted on that thread.

**Not verified:** the full regression suite — I ran the architecture fences plus the two touched test classes, not everything; and Sonar new-code coverage parity, which needs a full JaCoCo run. Coverage risk is low by construction: `pom.xml:32` excludes `core/api/**` and `core/util/converter/**`, which removes the new controller method and both `PropertyEditor`s from the metric entirely, leaving roughly 6 lines in the service and 14 in the validator, all driven by the four new nested tests.

src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
```diff
@@ -275,12 +277,26 @@ public List<BaseAttribute> listSignatureAttributesForCertificate(SecuredUUID cer
...
     public List<BaseAttribute> listSignatureFormattingConnectorAttributes(UUID connectorUuid,
             SecuredUUID signingProfileUuid) throws NotFoundException, ConnectorException, AttributeException {
         return fetchFormattingAttributeDefinitions(connectorUuid);
     }
 
+    @Override
+    @ExternalAuthorization(resource = Resource.SIGNING_PROFILE, action = ResourceAction.ANY)
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    public List<BaseAttribute> listContentSigningFormattingConnectorAttributes(UUID connectorUuid,
```
vyskocilm, line 289: ❓ q: discovery takes no `signingScheme`, but save refuses a formatting connector outright when the scheme is not `MANAGED` (`SigningProfileServiceImpl:941-945`, `requireNoSignatureFormattingConnector`). So a form for a non-managed profile can call this, receive a descriptor list, and then fail save for a reason discovery never mentioned.

Narrower than — and not a contradiction of — the PR's "a form built from a discovery answer cannot fail save on the connector's account", which is literally true as written. Whether the UI can reach that state is a product question I can't settle from the diff.

src/main/java/com/otilm/core/api/web/SigningProfileControllerImpl.java
```diff
@@ -154,6 +160,16 @@ public List<BaseAttribute> listSignatureFormattingConnectorAttributes(UUID conne
                 .listSignatureFormattingConnectorAttributes(connectorUuid, SecuredUUID.fromUUID(signingProfileUuid));
     }
 
+    @Override
+    @AuditLogged(module = Module.SIGNING, resource = Resource.SIGNING_PROFILE, operation = Operation.LIST_ATTRIBUTES)
+    public List<BaseAttribute> listContentSigningFormattingConnectorAttributes(UUID connectorUuid,
+            SignatureFamily family, SignatureLevel maxLevel, UUID signingProfileUuid)
```
vyskocilm: ❓ q: `signingProfileUuid` carries no `@LogResource`, so the `SIGNING_PROFILE` audit row for this operation has no object identifier. The sibling directly above (`:158-161`) has exactly the same gap, so this extends an existing one rather than introducing it — not raising it as a finding. Cheap to close here while the method is new, if you want it.

src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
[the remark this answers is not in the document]
vyskocilm: Checked this rebuttal against the code rather than taking it on trust. It stands.

Both cited ranges do what they say. `createSigningProfile` (`:424-437`) is `NOT_SUPPORTED`, calls `validateContentSigningWorkflow` at `:433` — which reads the connector via `getConnectorEntityWithInterfaces` at `:947-950` — then `fetchFormattingAttributeDefinitions` at `:434`, which reads it a second time via `getConnectorForApiClient` at `:1214-1217`. `updateSigningProfile` (`:473-483`) is the same shape. So this method reproduces the save path's existing window rather than opening a new one, and closing it means changing both paths together — which is what makes the "widen `ImmutableConnectorInfo` or add a `ConnectorService` method" objection real rather than a deferral.

"Save re-validates" holds concretely too: both save entry points call `validateContentSigningWorkflow` unconditionally, so a discovery answer built from a connector state that has since changed cannot be persisted without re-passing `requireConnectorServes`. Worst case is a briefly-wrong form field list, rejected at save — strictly less severe than the identical window already accepted on the write path.

Not a blocker for this PR.

src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
ivosh: Not changed — recording agreement, since you filed it as a note rather than a finding.

Both halves check out. `fetchFormattingAttributeDefinitions` is the outbound connector call, so `readOnly = true` held a database connection for that connector's whole response time. And `ConnectorInterfaceEntity.features` is a `@Column` SQL array rather than a lazy association, so `declares(...)` cannot throw `LazyInitializationException` once the session is closed.

Left open — it is your record for the next reviewer, not mine to close.

src/main/java/com/otilm/core/service/impl/SigningProfileServiceImpl.java
ivosh: Not changed.

Premise confirmed: `validateContentSigningWorkflow:940-943` refuses a formatting connector outright when the scheme is not `MANAGED`, and discovery takes no `signingScheme`.

The route stays connector-scoped. Its inputs are a connector and what that connector must serve, while `requireNoSignatureFormattingConnector` is a rule about the shape of the profile being saved. Discovery's guarantee is the connector-facing one — the descriptors it returns will not be refused on the connector's account. It does not cover a scheme-level rejection of having a formatting connector at all, and adding `signingScheme` here would make a connector-scoped route adjudicate a profile shape it is not given.

To be clear about what I am not claiming: nothing stops such a call from being made. The route does not require the connector to be attached to any profile, so a non-managed form can reach it. Whether the UI does reach it is the product question you named, and I have not settled it either.

src/main/java/com/otilm/core/api/web/SigningProfileControllerImpl.java
ivosh: Fixed in 89db1fc1.

`signingProfileUuid` now carries `@LogResource(uuid = true)`, so the `SIGNING_PROFILE` audit row for this operation gets its object identifier.

Safe when the parameter is omitted: `AuditLogAspect.getResourceUuidsFromParameter:313-316` returns `null` for a null parameter value before it reads the annotation, so an authorization-only call with no profile logs exactly as it did before. `SigningProfileFormattingDiscoveryITest` is green (2 tests).

The sibling at `:158-161` is left alone, as you scoped it.

vyskocilm (approved):

PR #2177, opened 2026-09-01
Grant transitive MEMBERS permissions to the attribute-content-resolver role

### LINKED ISSUE #2153: Attributes - Resource references: dereferencing certificate content fails with 403 "Required 'Members' permission for 'RA Profile'"
### Description

Any v3 authority operation that re-resolves the authority's stored attributes fails with 403 when one of them is a RESOURCE/CERTIFICATES attribute with non-empty content (otpki's `tlsTrust`), regardless of the caller's permissions — including superadmin.

Chain: `OperationAttributeResolver` elevates to the system identity `attribute-content-resolver` for RESOURCE dereference (`runAsSystem`), so the caller's principal is irrelevant from there on. Resolving certificate content goes through `CertificateServiceImpl.getCertificateEntity`, whose cross-bean call `RaProfileInternalService.evaluateCertificateRaProfilePermissions` is guarded by `@ExternalAuthorization(CERTIFICATE, DETAIL, parentResource = RA_PROFILE, parentAction = MEMBERS)`. The seeded resolver role (migration `V202607031200__CreateAttributeContentResolverUserAndPermissions`) grants CERTIFICATE:DETAIL but has no `raProfiles` entry at all and `allowAllResources=false`, so OPA denies the parent check and `raProfiles/members` is recorded as the denied permission. The migration anticipated the analogous transitive gate for SECRET (VAULT_PROFILE:MEMBERS) but missed the certificate one.

Blast radius: every `authorityAttributesFor` call site in `AuthorityProviderV3Adapter` — listing RA-profile attribute descriptors, issue, renew, revoke, identify, poll status, cancel. On async poll threads there is no request context, so the same failure surfaces without the descriptive message. v2 authorities are unaffected (the v2 adapter does not resolve stored attributes).

Suggested fix: new migration adding RA_PROFILE:MEMBERS (and consider GROUP:MEMBERS for the group fallback) to the resolver role. Test gap: `OperationAttributeResolverTest` stubs `runAsSystem` and mocks the builder, so the permission gate under the resolver identity is never exercised.

### Steps to Reproduce

1. Create an otpki authority and select at least one CA certificate in its tlsTrust attribute
2. As superadmin, start creating an RA profile and select that authority (FE calls `GET /v1/authorities/{uuid}/attributes/raProfile`)
3. Observe the 403

### Expected Behavior

RA-profile attribute descriptors load; the system resolver identity carries all transitive grants required to dereference RESOURCE attribute content, as it already does for SECRET references.

### Actual Behavior

HTTP 403 with message "Access Denied. Required 'Members' permission for 'RA Profile'"; the toast reads "Failed to get RA Profile Attribute Descriptor list (403): …". All v3 operations on that authority fail the same way, including certificate issuance.

### Severity

Critical

### Module

Auth

lubomirw:
## TLDR

Fixes #2153 - any v3 authority operation on an authority with a non-empty RESOURCE/CERTIFICATES attribute (otpki `tlsTrust`) returned 403 "Required 'Members' permission for 'RA Profile'" for every caller, superadmin included. One Flyway migration adds the missing transitive grant plus two reasoned fail-safe grants to the resolver role; no code changes.

## Root cause

`OperationAttributeResolver` elevates to the system identity `attribute-content-resolver` for RESOURCE dereference, so the caller's permissions are irrelevant from there on. Certificate content resolution passes through `RaProfileInternalService.evaluateCertificateRaProfilePermissions`, whose authorization names `RA_PROFILE:MEMBERS` as the parent gate. The seed migration (V202607031200) anticipated the analogous transitive gate for SECRET dereference (`VAULT_PROFILE:MEMBERS`) but missed the certificate one - the role has no `raProfiles` entry at all and `allowAllResources=false`, so OPA denies regardless of who is calling.

## What is granted

- `RA_PROFILE:MEMBERS` - the live fix.
- `GROUP:MEMBERS` - fail-safe: covers the group-membership fallback; unreachable under the current policy shape, but this bug class is exactly a transitive gate surfacing after policy evolution. Failure mode if ever missing: explicit 403.
- `ATTRIBUTE:MEMBERS` - fail-safe: covers the custom-attribute content filter (`AttributeEngine#loadCustomAttributesSecurityResourceFilter`), whose failure mode without a role-level grant is silent content filtering rather than a 403 - the harder-to-diagnose gap should the dereference path ever load custom attributes.

Pairs are seeded before the role update for fresh installs (catalog sync runs only after Flyway); `updateRolePermissions` merges additively, existing grants preserved.

## Verification

- Checksum registered via the guard cycle: `JavaMigrationChecksumTest` failed with the computed value, registered, green (`DatabaseMigrationTest` green too).
- Reviewed by four independent reviewers (Copilot CLI, guided, superpowers, hygiene): core mechanics confirmed (merge semantics, additive seeding, gate location `RaProfileServiceImpl:681-684`); all surviving findings addressed in the javadoc or consciously accepted (no behavioral migration test - repo precedent, the harness has no live auth service).
- Follow-up (not this PR): `documentation` repo's `roles-permissions.md` row for this role needs the updated permission list; tracked for the docs release.

### DIFF
--- a/src/main/java/db/migration/V202608311000__GrantRaProfileMembersToAttributeContentResolver.java
+++ b/src/main/java/db/migration/V202608311000__GrantRaProfileMembersToAttributeContentResolver.java
@@ -1,0 +1,64 @@
+package db.migration;
+
+import com.otilm.api.model.core.auth.Resource;
+import com.otilm.core.model.auth.ResourceAction;
+import com.otilm.core.util.AuthHelper;
+import com.otilm.core.util.DatabaseAuthMigration;
+import com.otilm.core.util.DatabaseMigration;
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+
+import java.util.EnumMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Grants {@code RA_PROFILE:MEMBERS} and {@code GROUP:MEMBERS} to the {@code attribute-content-resolver} role.
+ * <p>
+ * A CERTIFICATE dereference runs as this identity and passes through
+ * {@code RaProfileInternalService.evaluateCertificateRaProfilePermissions}, whose authorization names
+ * {@code RA_PROFILE:MEMBERS} as the parent gate — the certificate counterpart of the SECRET path's
+ * {@code VAULT_PROFILE:MEMBERS}. {@code GROUP:MEMBERS} covers the group-membership fallback the authorization
+ * core evaluates when a direct decision denies, so a policy-shape change cannot reintroduce a transitive 403
+ * for group-assigned certificates. Note the fallback is not certificate-scoped: should it ever become reachable
+ * for this role, the grant authorizes group-based DETAIL/LIST on any group-eligible resource — deliberate
+ * fail-safe breadth, consistent with the seed migration's coverage philosophy.
+ * <p>
+ * {@code ATTRIBUTE:MEMBERS} guards the custom-attribute content filter
+ * ({@code AttributeEngine#loadCustomAttributesSecurityResourceFilter}): without a role-level grant that filter
+ * is an empty allowlist and content is silently dropped rather than denied with a 403 — the harder-to-diagnose
+ * gap should the dereference path ever load custom attributes.
+ */
+// Flyway mandates the V<version>__<Description> class-name format, which cannot match Sonar's S101 identifier pattern.
+@SuppressWarnings("java:S101")
+public class V202608311000__GrantRaProfileMembersToAttributeContentResolver extends BaseJavaMigration {
+
+    @Override
+    public Integer getChecksum() {
+        return DatabaseMigration.JavaMigrationChecksums.V202608311000__GrantRaProfileMembersToAttributeContentResolver
+                .getChecksum();
+    }
+
+    @Override
+    public void migrate(Context context) throws Exception {
+        Map<Resource, List<ResourceAction>> addedResourceActions = new EnumMap<>(Resource.class);
+        addedResourceActions.put(Resource.RA_PROFILE, List.of(ResourceAction.MEMBERS));
+        addedResourceActions.put(Resource.GROUP, List.of(ResourceAction.MEMBERS));
+        addedResourceActions.put(Resource.ATTRIBUTE, List.of(ResourceAction.MEMBERS));
+
+        // On a fresh install this migration runs before Core's catalog sync, and the auth service rejects
+        // permissions naming an unknown resource/action. Additive no-op where the pair is already known.
+        DatabaseAuthMigration.seedResources(addedResourceActions);
+
+        String roleUuid = DatabaseAuthMigration
+                .getSystemRolesMapping()
+                .get(AuthHelper.ATTRIBUTE_CONTENT_RESOLVER_USERNAME);
+        if (roleUuid == null) {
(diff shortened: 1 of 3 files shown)


src/main/java/db/migration/V202608311000__GrantRaProfileMembersToAttributeContentResolver.java
```diff
@@ -0,0 +1,64 @@
...
+import com.otilm.core.util.DatabaseMigration;
+import org.flywaydb.core.api.migration.BaseJavaMigration;
+import org.flywaydb.core.api.migration.Context;
+
+import java.util.EnumMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Grants {@code RA_PROFILE:MEMBERS} and {@code GROUP:MEMBERS} to the {@code attribute-content-resolver} role.
```
ivosh: **~ Minor** | Documentation

src/main/java/db/migration/V202608311000__GrantRaProfileMembersToAttributeContentResolver.java:16 says the class grants `RA_PROFILE:MEMBERS` and `GROUP:MEMBERS`, but `migrate` also grants `ATTRIBUTE:MEMBERS` at line 47. Because the opening sentence reads as the scope summary for a security-role migration, maintainers can easily overlook the third fail-safe grant.

_Suggested:_ Include `ATTRIBUTE:MEMBERS` in the opening sentence, or replace the enumeration with a summary that explicitly directs readers to all three rationale sections.

src/main/java/db/migration/V202608311000__GrantRaProfileMembersToAttributeContentResolver.java
```diff
@@ -0,0 +1,64 @@
...
+ * <p>
+ * A CERTIFICATE dereference runs as this identity and passes through
```
ivosh: **. Nitpick** | Maintainability

src/main/java/db/migration/V202608311000__GrantRaProfileMembersToAttributeContentResolver.java:18-30 packs the certificate path, parent gate, secret-path analogy, group fallback, future policy behavior, broad authorization consequence, and attribute-filter failure mode into two dense paragraphs. The safety-critical breadth of `GROUP:MEMBERS` is buried at lines 23-25 inside a long sentence instead of being surfaced as its own concern.

_Suggested:_ Restructure the Javadoc into a one-line summary followed by short `<p><b>RA profile:</b>`, `<p><b>Group fallback:</b>`, and `<p><b>Attribute filtering:</b>` sections, deferring neighboring implementation detail to those classes' own Javadoc.

ivosh (dismissed):

src/main/java/db/migration/V202608311000__GrantRaProfileMembersToAttributeContentResolver.java
lubomirw: Fixed in e0068d12b — the opening sentence now names all three grants, including ATTRIBUTE:MEMBERS.

src/main/java/db/migration/V202608311000__GrantRaProfileMembersToAttributeContentResolver.java
lubomirw: Fixed in e0068d12b — restructured into a one-line summary plus a labeled section per grant (RA profile / group fallback / attribute filtering), so the GROUP:MEMBERS breadth is now its own concern.

ivosh (approved):

Issue #2030, opened 2026-08-12
Dynamic ORDER BY support in SecurityFilterRepository


Jakub-Bartal-01:
### Description

`SecurityFilterRepositoryImpl` accepts a single hardcoded `BiFunction<Root, CriteriaBuilder, Order>` and applies it via `cr.orderBy(order.apply(root, cb))`. Paging is applied with `setFirstResult`/`setMaxResults`, so a `Sort` carried inside `PageRequest` is silently ignored. This ticket makes ordering request-driven.

Scope:
- Accept a sort specification (field source, field identifier, direction) instead of a fixed lambda.
- Resolve a PROPERTY sort target to an ORDER BY path using the `joinAttributes` and `fieldAttribute` metamodel paths already carried by the `FilterField` enum.
- Support more than one `Order` so a deterministic tie-break (uuid) can be appended - a single-column ORDER BY over a non-unique column makes paging unstable.
- Handle the `SELECT DISTINCT` interaction: PostgreSQL requires ORDER BY expressions to appear in the select list when `distinct(true)` is set.
- Preserve today's default ordering when no sort is supplied.

This is a shared code path - all 16 controllers consuming `SearchRequestDto` route through this repository - so the change must be behaviour-preserving for the 14 not migrated in this Epic.

### Definition of Done

- Ordering is driven by the request; callers passing no sort keep their current order.
- Tie-break ordering makes paging stable across pages.
- Unit tests cover property sorts, the DISTINCT interaction and the no-sort default.
- Sonar quality gate passes and new code has at least 80% coverage.

PR #2178, opened 2026-09-01
Order the configurable-column listings by the field the request names

### LINKED ISSUE #2031: Server-side sorting on the configurable-column list endpoints

Jakub-Bartal-01:
The ORDER BY engine landed with #2030 but nothing called it. This applies it to the seven listings that ship configurable columns: certificates, cryptographic keys, discoveries, connectors (`POST /v2/connectors/list`), secrets, CBOMs and signing records. A request that names no sort keeps each endpoint's existing default ordering.

## The two-phase listings were discarding the sort

Certificates and cryptographic keys fetch in two phases: a page of uuids in the requested order, then the objects by `uuid IN (...)`. Both second-phase queries carried their own ordering — `ORDER BY c.created DESC` on `findCertificateDtosByUuidsIn`, and the `OrderByCreatedAtDesc` baked into `findFullByUuidInOrderByCreatedAtDesc`. Ordering only the first phase is therefore replaced by the second, so a sort would be accepted and then silently thrown away.

Both orderings are removed and the rows are ranked back into the uuid list's order with a new `SortOrderBuilder.rankBy`, which is where the ordering now lives. Ranking is applied whether or not a sort was named: the ranked list is the one carrying the uuid tie-break, so it is also what makes a default-ordered page stable across a page boundary when the default term ties.

## A latent mismatch in the key listing

`getCountsOfAssociations` returned bare counts that the key listing lined up with its key items **by position**. That held only while both queries ordered by the same non-unique column and it had no ties; under a requested sort they do not line up at all, and a count would be rendered against the wrong key. It now returns the uuid alongside the count, through a Spring Data interface projection, and the caller joins on it.

This is a pre-existing defect rather than one this change introduces, but request-driven ordering is what turns it from unlikely into certain.

## Refusing a sort the catalogue withholds

A sort naming a field the catalogue reports as not sortable is refused against `SearchHelper.isOrderableField` — the same predicate that populates the published `sortable` flag — rather than answered with a silently unordered page. `PRIVATE_KEY` is the illustrative case: it resolves to a path and would order fine, but what the column displays is a comparison against an expected value, not the value that path holds.

## Tests

- `SortOrderBuilderTest` gains the ranking cases, which hold without a persistence context.
- The orderability refusal needs a resolved `FilterField`, so it sits in `SecurityFilterRepositoryITest` beside the unknown-field and wrong-resource cases already there.
- `KeyListingSortITest` exercises the key listing end to end, since that is where the ordering was being discarded rather than where it is built. The keys are seeded so the alphabetical order the sort asks for is the reverse of the listing's own default: an assertion that only checked the returned set, or a sort that never reached the query, would pass against the default order. **All four cases fail when the ranking is removed** — I checked rather than assumed.
- The default-order case reads its expectation back from the seeded rows instead of writing it out, because four rows saved in one test can share a `createdAt` and then the uuid tie-break decides rather than the seeding order.

Full suite green at 6377 tests before the rebase onto the merged #2162.

`CertificateListingSortITest` does the same for the certificate listing, the other two-phase fetch and the one whose second query carried the `ORDER BY c.created DESC`. All five of its cases fail when the ranking is removed. It also carries the AC5 case: a sorted listing still returns only the rows the filter admits, ordered within that set rather than across the table.

## Still owed on this ticket

Not claiming every acceptance criterion is met yet. Outstanding:

- End-to-end sort coverage for the five single-phase listings — ACs 1 and 3 ask for each of the seven, and discoveries, connectors, secrets, CBOMs and signing records are covered only at the repository level by `SecurityFilterRepositoryITest`, not through their own service. They all take the same one-argument path, so the risk is wiring rather than behaviour.
- AC5's second half: composition with row-level authorization, as distinct from the filter case now covered.
- Sonar and the coverage gate.

Closes #2031

### DIFF
--- a/src/main/java/com/otilm/core/dao/repository/CertificateRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/CertificateRepository.java
@@ -602,7 +602,11 @@ public interface CertificateRepository
             LEFT JOIN OwnerAssociation oa ON oa.objectUuid = c.uuid
             LEFT JOIN RaProfile ra ON ra.uuid = c.raProfileUuid
             WHERE c.uuid IN ?1
-            ORDER BY c.created DESC
             """)
+    /**
+     * The page of certificates named by the uuid list. Carries no ORDER BY: the ordering the listing asked for lives in
+     * the rank of that list, and an ORDER BY here would replace it. Callers rank the result with
+     * {@code SortOrderBuilder.rankBy}.
+     */
     List<CertificateDto> findCertificateDtosByUuidsIn(List<UUID> uuids);
 }
--- a/src/main/java/com/otilm/core/dao/repository/CertificateRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/CertificateRepository.java
@@ -575,11 +575,14 @@ public interface CertificateRepository
             @Param("oldCode") String oldCode);
 
     /**
-     * Populates almost all of the {@link CertificateDto} properties.
+     * Populates almost all of the {@link CertificateDto} properties for the page of certificates the uuid list names.
      *
      * <p>
      * Groups need to be retrieved separately and set to the DTO.
-     * </p>
+     *
+     * <p>
+     * Carries no ORDER BY: the ordering the listing asked for lives in the rank of that list, and an ORDER BY here
+     * would replace it. Callers rank the result with {@code SortOrderBuilder.rankBy}.
      */
     @Query("""
             SELECT new com.otilm.api.model.core.certificate.CertificateDto(
@@ -603,10 +606,5 @@ public interface CertificateRepository
             LEFT JOIN RaProfile ra ON ra.uuid = c.raProfileUuid
             WHERE c.uuid IN ?1
             """)
-    /**
-     * The page of certificates named by the uuid list. Carries no ORDER BY: the ordering the listing asked for lives in
-     * the rank of that list, and an ORDER BY here would replace it. Callers rank the result with
-     * {@code SortOrderBuilder.rankBy}.
-     */
     List<CertificateDto> findCertificateDtosByUuidsIn(List<UUID> uuids);
 }
--- a/src/main/java/com/otilm/core/dao/repository/CryptographicKeyItemRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/CryptographicKeyItemRepository.java
@@ -30,8 +30,13 @@ public interface CryptographicKeyItemRepository extends SecurityFilterRepository
     @EntityGraph(attributePaths = {"key", "key.tokenProfile"})
     List<CryptographicKeyItem> findByUuidIn(List<UUID> uuids);
 
+    /**
+     * The key items named by the uuid list. Deliberately carries no ordering: the ordering the listing asked for lives
+     * in the rank of that list, and the {@code OrderByCreatedAtDesc} this method used to name would replace it. Callers
+     * rank the result with {@code SortOrderBuilder.rankBy}.
+     */
     @EntityGraph(attributePaths = {"key", "key.tokenProfile", "key.groups", "key.owner"})
-    List<CryptographicKeyItem> findFullByUuidInOrderByCreatedAtDesc(List<UUID> uuids);
+    List<CryptographicKeyItem> findFullByUuidIn(List<UUID> uuids);
 
     @EntityGraph(attributePaths = {"key", "key.items"})
     List<CryptographicKeyItem> findWithKeyByUuidIn(List<UUID> uuids);
@@ -61,8 +66,16 @@ public interface CryptographicKeyItemRepository extends SecurityFilterRepository
             nativeQuery = true)
     Integer insertWithFingerprintConflictResolve(@Param("cki") CryptographicKeyItem keyItem);
 
+    /**
+     * How many certificates each of the named key items is associated with, keyed by the item's uuid.
+     *
+     * <p>
+     * Returns the uuid rather than counts alone because the caller has to line each count up with its key item. Two
+     * queries ordered by the same non-unique column line up only while that column has no ties, and once the listing
+     * orders by a field the request names they do not line up at all.
+     */
     @Query(value = """
-            SELECT COUNT(c.uuid)
+            SELECT cki.uuid AS uuid, COUNT(c.uuid) AS associations
                 FROM CryptographicKeyItem cki
                 JOIN cki.key ck
                 LEFT JOIN Certificate c
@@ -70,9 +83,16 @@ public interface CryptographicKeyItemRepository extends SecurityFilterRepository
                     OR c.altKeyUuid = ck.uuid
                 WHERE cki.uuid IN :uuids
                 GROUP BY cki.uuid
-                ORDER BY cki.createdAt DESC
             """)
-    List<Integer> getCountsOfAssociations(@Param("uuids") List<UUID> uuids);
+    List<KeyItemAssociationCount> getCountsOfAssociations(@Param("uuids") List<UUID> uuids);
+
+    /** One key item's certificate-association count. */
+    interface KeyItemAssociationCount {
+
+        UUID getUuid();
+
+        int getAssociations();
+    }
 
     @EntityGraph(attributePaths = {"key", "key.tokenInstanceReference", "key.tokenInstanceReference.connector"})
     Optional<CryptographicKeyItem> findWithConnectorByUuid(UUID uuid);
--- a/src/main/java/com/otilm/core/dao/repository/CryptographicKeyItemRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/CryptographicKeyItemRepository.java
@@ -32,8 +32,8 @@ public interface CryptographicKeyItemRepository extends SecurityFilterRepository
 
     /**
      * The key items named by the uuid list. Deliberately carries no ordering: the ordering the listing asked for lives
-     * in the rank of that list, and the {@code OrderByCreatedAtDesc} this method used to name would replace it. Callers
-     * rank the result with {@code SortOrderBuilder.rankBy}.
+     * in the rank of that list, and an ORDER BY here would replace it. Callers rank the result with
+     * {@code SortOrderBuilder.rankBy}.
      */
     @EntityGraph(attributePaths = {"key", "key.tokenProfile", "key.groups", "key.owner"})
     List<CryptographicKeyItem> findFullByUuidIn(List<UUID> uuids);
--- a/src/main/java/com/otilm/core/dao/repository/SecurityFilterRepositoryImpl.java
+++ b/src/main/java/com/otilm/core/dao/repository/SecurityFilterRepositoryImpl.java
@@ -146,10 +146,17 @@ public class SecurityFilterRepositoryImpl<T, ID> extends SimpleJpaRepository<T, 
     }
 
     /**
-     * The uuids of a page ordered by a field of a joined entity. The join gives a root as many rows as it has matches,
-     * so a window cut over those rows would underfill the page and let the same root reappear on the next one. The
-     * query therefore groups by the root's uuid and orders by the aggregate the ordering resolves - one row per root,
-     * which is what the window is allowed to cut.
+     * The uuids of a page in the order the request asked for, for the two sorts the entity query cannot carry itself.
+     *
+     * <p>
+     * A sort through a join gives a root as many rows as the join has matches, so a window cut over those rows would
+     * underfill the page and let the same root reappear on the next one. A sort by an attribute resolves to a
+     * correlated scalar subquery, and the entity query selects DISTINCT, which the database will not order by an
+     * expression that is absent from the select list.
+     *
+     * <p>
+     * This query answers both: it selects the sort key alongside the uuid, groups by the uuid and orders by the
+     * aggregate the ordering resolves - one row per root, carrying its key, which is what the window is allowed to cut.
      */
     private List<UUID> findUuidsOrderedBySortKey(final SecurityFilter filter,
             final TriFunction<Root<T>, CriteriaBuilder, CriteriaQuery<?>, Predicate> additionalWhereClause,
--- a/src/main/java/com/otilm/core/dao/repository/SecurityFilterRepositoryImpl.java
+++ b/src/main/java/com/otilm/core/dao/repository/SecurityFilterRepositoryImpl.java
@@ -21,6 +21,7 @@ import jakarta.persistence.EntityManager;
 import jakarta.persistence.NoResultException;
 import jakarta.persistence.Tuple;
 import jakarta.persistence.TypedQuery;
+import jakarta.persistence.criteria.CommonAbstractCriteria;
 import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaDelete;
 import jakarta.persistence.criteria.CriteriaQuery;
@@ -117,8 +118,8 @@ public class SecurityFilterRepositoryImpl<T, ID> extends SimpleJpaRepository<T, 
     public List<T> findUsingSecurityFilter(final SecurityFilter filter, List<String> fetchAssociations,
             final TriFunction<Root<T>, CriteriaBuilder, CriteriaQuery<?>, Predicate> additionalWhereClause,
             final Pageable p, final BiFunction<Root<T>, CriteriaBuilder, Order> order, final SortSpecification sort) {
-        if (SortOrderBuilder.traversesJoin(sort)) {
-            return loadInUuidOrder(findUuidsOrderedByJoinedField(filter, additionalWhereClause, p, sort),
+        if (SortOrderBuilder.needsRankedUuidQuery(sort)) {
+            return loadInUuidOrder(findUuidsOrderedBySortKey(filter, additionalWhereClause, p, sort),
                     fetchAssociations);
         }
         final CriteriaQuery<T> cr = createCriteriaBuilder(filter, fetchAssociations, additionalWhereClause, order, sort,
@@ -137,8 +138,8 @@ public class SecurityFilterRepositoryImpl<T, ID> extends SimpleJpaRepository<T, 
     public List<UUID> findUuidsUsingSecurityFilter(final SecurityFilter filter,
             final TriFunction<Root<T>, CriteriaBuilder, CriteriaQuery<?>, Predicate> additionalWhereClause,
             final Pageable p, final BiFunction<Root<T>, CriteriaBuilder, Order> order, final SortSpecification sort) {
-        if (SortOrderBuilder.traversesJoin(sort)) {
-            return findUuidsOrderedByJoinedField(filter, additionalWhereClause, p, sort);
+        if (SortOrderBuilder.needsRankedUuidQuery(sort)) {
+            return findUuidsOrderedBySortKey(filter, additionalWhereClause, p, sort);
         }
         final CriteriaQuery<UUID> cr = createCriteriaBuilderUuid(filter, additionalWhereClause, order, sort, p != null);
         return window(entityManager.createQuery(cr), p).getResultList();
@@ -150,7 +151,7 @@ public class SecurityFilterRepositoryImpl<T, ID> extends SimpleJpaRepository<T, 
      * query therefore groups by the root's uuid and orders by the aggregate the ordering resolves - one row per root,
      * which is what the window is allowed to cut.
      */
-    private List<UUID> findUuidsOrderedByJoinedField(final SecurityFilter filter,
+    private List<UUID> findUuidsOrderedBySortKey(final SecurityFilter filter,
             final TriFunction<Root<T>, CriteriaBuilder, CriteriaQuery<?>, Predicate> additionalWhereClause,
             final Pageable p, final SortSpecification sort) {
         final Class<T> entity = this.entityInformation.getJavaType();
@@ -159,7 +160,7 @@ public class SecurityFilterRepositoryImpl<T, ID> extends SimpleJpaRepository<T, 
         final Root<T> root = cr.from(entity);
         final Path<?> uuid = root.get(UniquelyIdentified_.UUID);
 
-        final SortOrderBuilder.GroupedOrdering ordering = SortOrderBuilder.resolveGrouped(root, cb, sort);
+        final SortOrderBuilder.GroupedOrdering ordering = SortOrderBuilder.resolveGrouped(root, cb, cr, sort);
         cr.multiselect(uuid, ordering.sortKey());
         cr.groupBy(uuid);
         cr.orderBy(ordering.orders());
@@ -199,10 +200,10 @@ public class SecurityFilterRepositoryImpl<T, ID> extends SimpleJpaRepository<T, 
     }
 
     private List<Order> resolveOrdering(final Root<T> root, final CriteriaBuilder cb,
-            final BiFunction<Root<T>, CriteriaBuilder, Order> order, final SortSpecification sort,
-            final boolean paged) {
+            final CommonAbstractCriteria query, final BiFunction<Root<T>, CriteriaBuilder, Order> order,
+            final SortSpecification sort, final boolean paged) {
         final Order defaultOrder = order == null ? null : order.apply(root, cb);
-        return SortOrderBuilder.resolve(root, cb, sort, defaultOrder, paged);
+        return SortOrderBuilder.resolve(root, cb, query, sort, defaultOrder, paged);
     }
 
     private void applyPredicates(final CriteriaQuery<?> cr, final SecurityFilter filter,
@@ -440,7 +441,7 @@ public class SecurityFilterRepositoryImpl<T, ID> extends SimpleJpaRepository<T, 
 
         fetchAssociations(root, fetchAssociations);
 
-        final List<Order> orders = resolveOrdering(root, cb, order, sort, paged);
+        final List<Order> orders = resolveOrdering(root, cb, cr, order, sort, paged);
         if (!orders.isEmpty()) {
             cr.orderBy(orders);
         }
@@ -460,7 +461,7 @@ public class SecurityFilterRepositoryImpl<T, ID> extends SimpleJpaRepository<T, 
 
         cr.select(root.get("uuid"));
 
-        final List<Order> orders = resolveOrdering(root, cb, order, sort, paged);
+        final List<Order> orders = resolveOrdering(root, cb, cr, order, sort, paged);
         if (!orders.isEmpty()) {
             cr.orderBy(orders);
         }
--- a/src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
@@ -27,6 +27,7 @@ import com.otilm.api.model.core.search.SearchFieldDataByGroupDto;
 import com.otilm.api.model.core.search.SearchFieldDataDto;
 import com.otilm.core.attribute.engine.AttributeColumnProjector;
 import com.otilm.core.attribute.engine.AttributeEngine;
+import com.otilm.core.attribute.engine.ListingSortResolver;
 import com.otilm.core.attribute.engine.records.ObjectAttributeContentInfo;
 import com.otilm.core.client.ConnectorApiFactory;
 import com.otilm.core.comparator.SearchFieldDataComparator;
@@ -38,7 +39,6 @@ import com.otilm.core.dao.repository.CertificateContentRepository;
 import com.otilm.core.dao.repository.ConnectorRepository;
 import com.otilm.core.dao.repository.DiscoveryCertificateRepository;
 import com.otilm.core.dao.repository.DiscoveryRepository;
-import com.otilm.core.dao.repository.SortSpecification;
 import com.otilm.core.enums.FilterField;
 import com.otilm.core.events.data.DiscoveryResult;
 import com.otilm.core.events.handlers.DiscoveryFinishedEventHandler;
@@ -94,6 +94,8 @@ public class DiscoveryServiceImpl implements DiscoveryExternalService, Discovery
 
     private AttributeEngine attributeEngine;
     private AttributeColumnProjector attributeColumnProjector;
+
+    private ListingSortResolver listingSortResolver;
 
     private TriggerInternalService triggerInternalService;
     private DiscoveryRepository discoveryRepository;
@@ -162,6 +164,11 @@ public class DiscoveryServiceImpl implements DiscoveryExternalService, Discovery
     }
 
     @Autowired
+    public void setListingSortResolver(ListingSortResolver listingSortResolver) {
+        this.listingSortResolver = listingSortResolver;
+    }
+
+    @Autowired
     public void setDiscoveryRepository(DiscoveryRepository discoveryRepository) {
         this.discoveryRepository = discoveryRepository;
     }
@@ -197,7 +204,8 @@ public class DiscoveryServiceImpl implements DiscoveryExternalService, Discovery
                 cb, cr) -> FilterPredicatesBuilder.getFiltersPredicate(cb, cr, root, request.getFilters());
         final List<DiscoveryListDto> listedDiscoveriesDTOs = discoveryRepository
                 .findUsingSecurityFilter(filter, List.of(), additionalWhereClause, p,
-                        (root, cb) -> cb.desc(root.get("created")), SortSpecification.from(request.getSort()))
+                        (root, cb) -> cb.desc(root.get("created")),
+                        listingSortResolver.resolve(Resource.DISCOVERY, request.getSort()))
                 .stream()
                 .map(Discovery::mapToListDto)
                 .toList();
--- a/src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
@@ -38,6 +38,7 @@ import com.otilm.core.dao.repository.CertificateContentRepository;
 import com.otilm.core.dao.repository.ConnectorRepository;
 import com.otilm.core.dao.repository.DiscoveryCertificateRepository;
 import com.otilm.core.dao.repository.DiscoveryRepository;
+import com.otilm.core.dao.repository.SortSpecification;
 import com.otilm.core.enums.FilterField;
 import com.otilm.core.events.data.DiscoveryResult;
 import com.otilm.core.events.handlers.DiscoveryFinishedEventHandler;
@@ -196,7 +197,7 @@ public class DiscoveryServiceImpl implements DiscoveryExternalService, Discovery
                 cb, cr) -> FilterPredicatesBuilder.getFiltersPredicate(cb, cr, root, request.getFilters());
         final List<DiscoveryListDto> listedDiscoveriesDTOs = discoveryRepository
                 .findUsingSecurityFilter(filter, List.of(), additionalWhereClause, p,
-                        (root, cb) -> cb.desc(root.get("created")))
+                        (root, cb) -> cb.desc(root.get("created")), SortSpecification.from(request.getSort()))
                 .stream()
                 .map(Discovery::mapToListDto)
                 .toList();
--- a/src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
@@ -443,7 +443,7 @@ public class DiscoveryServiceImpl implements DiscoveryExternalService, Discovery
                 .getResourceSearchableFields(Resource.DISCOVERY, false);
 
         List<SearchFieldDataDto> fields = List
-                .of(SearchHelper.prepareSearch(FilterField.CKI_NAME),
+                .of(SearchHelper.prepareSearch(FilterField.DISCOVERY_NAME),
                         SearchHelper
                                 .prepareSearch(FilterField.DISCOVERY_STATUS,
                                         Arrays.stream(DiscoveryStatus.values()).map(DiscoveryStatus::getCode).toList()),
--- a/src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
+++ b/src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
@@ -10,6 +10,7 @@ import com.otilm.api.model.common.enums.IPlatformEnum;
 import com.otilm.api.model.core.auth.Resource;
 import com.otilm.api.model.core.search.FilterConditionOperator;
 import com.otilm.api.model.core.search.FilterFieldSource;
+import com.otilm.core.attribute.engine.AttributeEngine.CustomAttributeContentFilter;
 import com.otilm.core.dao.entity.AttributeContent2Object;
 import com.otilm.core.dao.entity.AttributeContent2Object_;
 import com.otilm.core.dao.entity.AttributeContentItem_;
@@ -25,9 +26,11 @@ import com.otilm.core.dao.entity.UniquelyIdentified_;
 import com.otilm.core.dao.entity.cbom.CryptoAssetSource;
 import com.otilm.core.dao.entity.cbom.CryptoAssetSource_;
 import com.otilm.core.dao.entity.cbom.CryptoAsset_;
+import com.otilm.core.dao.repository.SortSpecification;
 import com.otilm.core.enums.FilterField;
 import com.otilm.core.enums.ResourceToClass;
 import com.otilm.core.enums.SearchFieldTypeEnum;
+import com.otilm.core.model.AttributeFieldIdentifier;
 import com.otilm.core.model.cbom.CryptoAssetIdentityGuard;
 import jakarta.persistence.EntityManager;
 import jakarta.persistence.Query;
@@ -135,25 +138,20 @@ public class FilterPredicatesBuilder {
 
         final AttributeType attributeType = filterDto.getFieldSource().getAttributeType();
         final String identifier = filterDto.getFieldIdentifier();
-        final String[] fieldIdentifier = identifier.split("\\|");
-        final AttributeContentType contentType = AttributeContentType.valueOf(fieldIdentifier[1]);
-        final String attributeName = fieldIdentifier[0];
+        final AttributeFieldIdentifier fieldIdentifier = AttributeFieldIdentifier.parse(identifier);
+        if (fieldIdentifier == null || fieldIdentifier.contentType() == null) {
+            throw new ValidationException(ValidationError
+                    .create("Filter field identifier %s does not name an attribute.".formatted(identifier)));
+        }
+        final AttributeContentType contentType = fieldIdentifier.contentType();
+        final String attributeName = fieldIdentifier.attributeName();
         final boolean isNotExistCondition = List
                 .of(FilterConditionOperator.NOT_EQUALS, FilterConditionOperator.NOT_CONTAINS,
                         FilterConditionOperator.EMPTY, FilterConditionOperator.NOT_MATCHES)
                 .contains(filterDto.getCondition());
 
-        // attributes content for cryptographic key items are stored under resource CRYPTOGRAPHIC_KEY, but for meta
-        // attributes, object uuid is uuid of cryptographic key item and for custom and data attribute it is uuid of
-        // cryptographic key
-        // place for improvement is to consolidate resource for attributes content
-        final Resource resource = root.getJavaType().equals(CryptographicKeyItem.class)
-                ? Resource.CRYPTOGRAPHIC_KEY
-                : ResourceToClass.getResourceByClass(root.getJavaType());
-        final String objectUuidPath = resource == Resource.CRYPTOGRAPHIC_KEY
-                && (attributeType == AttributeType.CUSTOM || attributeType == AttributeType.DATA)
-                        ? CryptographicKeyItem_.keyUuid.getName()
-                        : UniquelyIdentified_.uuid.getName();
+        final Resource resource = attributeResourceOf(root);
+        final String objectUuidPath = attributeObjectUuidPath(root, attributeType);
 
         List<Predicate> predicates = new ArrayList<>(attributeCorrelationPredicates(criteriaBuilder, root, subqueryRoot,
                 joinDefinition, attributeType, contentType, attributeName, resource, objectUuidPath));
@@ -1147,6 +1145,10 @@ public class FilterPredicatesBuilder {
     /**
      * The resource an object's attribute content is stored under. Key items carry the key's attributes, so a key item
      * root resolves to {@code CRYPTOGRAPHIC_KEY} rather than to a resource of its own.
+     *
+     * <p>
+     * A place for improvement is to consolidate the resource attribute content is stored under, so this mapping is not
+     * needed at all.
      */
     private static <T> Resource attributeResourceOf(final Root<T> root) {
         return root.getJavaType().equals(CryptographicKeyItem.class)
@@ -1157,9 +1159,13 @@ public class FilterPredicatesBuilder {
     /**
      * Which uuid of the root the content is keyed by. For a key item, meta attributes are attached to the item while
      * custom and data attributes are attached to the key it belongs to.
+     *
+     * <p>
+     * Decided from the root rather than from the resource, because the key uuid is a column of the key item and of
+     * nothing else: a root that is not one has no such path to read, whatever resource its content is filed under.
      */
-    private static String attributeObjectUuidPath(final Resource resource, final AttributeType attributeType) {
-        return resource == Resource.CRYPTOGRAPHIC_KEY
+    private static <T> String attributeObjectUuidPath(final Root<T> root, final AttributeType attributeType) {
+        return root.getJavaType().equals(CryptographicKeyItem.class)
                 && (attributeType == AttributeType.CUSTOM || attributeType == AttributeType.DATA)
                         ? CryptographicKeyItem_.keyUuid.getName()
                         : UniquelyIdentified_.uuid.getName();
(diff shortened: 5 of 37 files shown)


src/main/java/com/otilm/core/util/SortOrderBuilder.java
```diff
@@ -123,12 +155,21 @@ private static FilterField resolveField(SortSpecification sort) {
...
         } catch (IllegalArgumentException e) {
             throw new ValidationException(
                     ValidationError.create("Unknown sort field identifier %s.".formatted(sort.fieldIdentifier())));
         }
+        // Refused against the same predicate the catalogue publishes as `sortable`, so a request the frontend was told
+        // to withhold is answered with an error rather than with a silently unordered page. Checked before the path is
+        // resolved, because several of these fields do resolve to a path - it is just not the value the column shows.
+        if (!SearchHelper.isOrderableField(field)) {
+            throw new ValidationException(
+                    ValidationError.create("Field %s cannot be used to order this listing.".formatted(field.name())));
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed in 30aeec78f. `isSortable` now reports a field sortable when its path can be ordered by AND its resource is one whose listing applies the sort.

The resource half is why I had not simply wired it to `isOrderableField`: the flag is published per field while the ordering is wired per listing, and nine of the sixteen listings that take a search request are not wired. A bare `isOrderableField` would advertise ordering on crypto assets, audit logs and the other seven unwired listings, which would accept the sort and discard it — the same "client sorts, silently gets the default order" failure the flag was hard-coded false to avoid. `SORT_WIRED_RESOURCES` is explicit because nothing on a `FilterField` knows whether its listing was wired.

The catalogue assertion is updated in both directions: a wired listing advertises its orderable fields, and a field whose path cannot be ordered by stays non-sortable even there. `CryptographicAssetSearchableFieldsITest.noFieldReportsSortable` already held the unwired side and still passes, which is what shows the resource gate works.

Attribute fields get the same treatment on the stacked #2179, gated additionally on `displayable`.

src/main/java/com/otilm/core/service/impl/DiscoveryServiceImpl.java
```diff
@@ -196,7 +197,7 @@ public DiscoveryResponseDto listDiscoveries(final SecurityFilter filter, final S
                 cb, cr) -> FilterPredicatesBuilder.getFiltersPredicate(cb, cr, root, request.getFilters());
         final List<DiscoveryListDto> listedDiscoveriesDTOs = discoveryRepository
                 .findUsingSecurityFilter(filter, List.of(), additionalWhereClause, p,
-                        (root, cb) -> cb.desc(root.get("created")))
+                        (root, cb) -> cb.desc(root.get("created")), SortSpecification.from(request.getSort()))
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Correct, and this is the bug becoming reachable rather than a new one. The discovery catalogue published `CKI_NAME` — a cryptographic-key field pointing at `CryptographicKeyItem_.name` — as its Name field. Filtering worked only by accident: the filter path resolves a field by path and never checks which resource declares it, so `name` matched `Discovery.name`. Ordering does check the declaring resource, so the advertised Name column answered a sort with a validation error.

Fixed in 30aeec78f: it now publishes `DISCOVERY_NAME`, which has existed in `FilterField` all along and points at `Discovery_.name`.

On compatibility — a client already sending `CKI_NAME` as a discovery filter is unaffected. I checked the filter path for a root-resource check before changing anything: there is none, so `CKI_NAME` still resolves by path exactly as it did. Only what the catalogue advertises changes. Saved views resolve stored column identifiers against the live catalogue and degrade to a skipped column by design, and the saved-view table shipped days ago, so a stored `CKI_NAME` discovery column is unlikely to exist and would not break anything if it did.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -1128,4 +1124,111 @@ public static String buildPathToProperty(List<Attribute> joinAttributes, Attribu
...
+                .function(JSONB_EXTRACT_PATH_TEXT_FUNCTION_NAME, String.class,
+                        joinContentItem.get(AttributeContentItem_.json),
+                        criteriaBuilder.literal(contentType.isFilterByData() ? "data" : "reference"));
+        final Expression<?> value = castedAttributeContentData.contains(contentType)
+                ? ((JpaExpression<String>) extracted).cast(contentType.getContentDataClass())
+                : extracted;
+
+        subquery
+                .select(value)
+                .where(attributeCorrelationPredicates(criteriaBuilder, root, subqueryRoot, joinDefinition,
```
lubomirw: **[High]** | Security | Confidence: High

The sort subquery applies none of the authorization gates the attribute *projection* path applies. `attributeCorrelationPredicates` (:1137-1145) correlates only on type / contentType / name / objectType / objectUuid. Projection additionally enforces the caller's `allowedDefinitionUuids` / `forbiddenDefinitionUuids` (`AttributeColumnProjector:152-153`, `AttributeContent2ObjectRepository:132-133`), `encryptedContent`, `WITHHELD_CONTENT_TYPES` = {SECRET, CODEBLOCK} and `isVisible(definition)` (`AttributeColumnProjector:230-232`). Ordering by a value the same response blanks is a comparative oracle over data the caller is denied.

_Suggested:_ thread the same `CustomAttributeContentFilter` into the subquery predicates for `AttributeType.CUSTOM`, and add `ad.enabled = true`, matching `getProjectedAttributesContent`. The catalogue gate in the sibling comment on `SortOrderBuilder:110` covers the visible / SECRET / CODEBLOCK / encrypted half in one check; the definition-uuid lists still need explicit predicates here.

Calibration: the identical gap already exists in `getAttributeFilterPredicate`, which offers a strictly stronger oracle (`STARTS_WITH` extracts character by character). This is a parity gap against a doctrine the codebase states explicitly, not a new class of capability — worth a follow-up on the filter path too.

<sub>flagged by GPT 5.6, Copilot CLI, guided-pr-review and superpowers; verified against both paths</sub>

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -161,23 +162,59 @@ public static SearchFieldDataDto prepareSearchForJSON(final SearchFieldObject at
...
+     * The flag is published per field while the ordering is wired per listing, and seven of the sixteen listings that
+     * accept a search request are wired. A field of any other resource is reported not sortable however orderable its
+     * path is, because a catalogue reporting {@code true} there would advertise an ordering that listing discards: the
+     * client sorts, receives the default order, and is told nothing.
+     *
+     * <p>
+     * Explicit rather than derived: the wiring lives in each listing service and nothing on a {@link FilterField} knows
+     * whether its listing was wired. A listing that gains ordering adds itself here.
+     */
+    private static final Set<Resource> SORT_WIRED_RESOURCES = Set
```
lubomirw, line 184: **[High]** | Correctness | Confidence: High

`SORT_WIRED_RESOURCES` includes `SECRET` and `CBOM`, so `applyAttributeSortability` marks their displayable attribute columns `sortable=true` — but `ResourceToClass` has no `Secret.class` or `Cbom.class` entry (only `CRYPTO_ASSET` under the `// CBOMS` comment at :75), and `getResourceByClass` returns `null` (:133). `FilterPredicatesBuilder.attributeResourceOf` therefore yields null, the correlation becomes `objectType = null`, no row matches, every sort key is null, and the page falls back to the uuid tie-break. That is exactly the "catalogue advertises an ordering the listing discards" failure this PR set out to eliminate.

The same null also disables the cross-resource guard at `SortOrderBuilder:216` (`resource != null && ...`), so on those two roots a `FilterField` belonging to another resource is accepted — and the javadoc's stated reason for the guard (a field on a mapped superclass resolves against any entity that inherits it) is the case that slips through.

_Suggested:_ add `SECRET(Resource.SECRET, Secret.class)` and `CBOM(Resource.CBOM, Cbom.class)` to `ResourceToClass` (this also repairs the pre-existing attribute *filter* on secrets, broken the same way), and make an unmapped sort root fail closed rather than skip the check. Add an `AttributeColumnSortITest` case on the secret listing.

The "Still owed" section says secrets and CBOMs "all take the same one-argument path, so the risk is wiring rather than behaviour" — this is behaviour, and the missing end-to-end coverage is what hid it.

<sub>flagged by Copilot CLI and superpowers independently; verified at source</sub>

src/main/java/com/otilm/core/util/SortOrderBuilder.java
```diff
@@ -53,10 +74,12 @@ public static boolean traversesJoin(SortSpecification sort) {
...
      * @param paged whether the query will be windowed, which is what makes an unstable order observable
      */
-    public static List<Order> resolve(Root<?> root, CriteriaBuilder criteriaBuilder, SortSpecification sort,
-            Order defaultOrder, boolean paged) {
+    public static List<Order> resolve(Root<?> root, CriteriaBuilder criteriaBuilder, CommonAbstractCriteria query,
+            SortSpecification sort, Order defaultOrder, boolean paged) {
         List<Order> orders = new ArrayList<>();
         if (sort != null) {
+            // Property-only by construction: needsRankedUuidQuery sends every attribute sort to resolveGrouped, and
+            // resolveField refuses a non-property source rather than letting one build an unorderable query here.
```
lubomirw, line 82: **[High]** | Correctness | Confidence: High

Null placement differs between the two sort paths on the same endpoint. `resolveGrouped` pins `NullPrecedence.LAST` in both directions (:120-123) with a comment explaining that rows a column has nothing to show for must never lead the page — but `resolve()` on the next line adds its term with no null precedence, so PostgreSQL's default applies: NULLS LAST for ASC, NULLS FIRST for DESC.

So a DESC sort on a nullable direct column leads the page with blank cells, while the same DESC sort through a join puts them last. `DISCOVERY_END_TIME` is null while a discovery runs; `Certificate.commonName` is nullable. Only the grouped path is tested.

_Suggested:_ extract a shared `primary(criteriaBuilder, expression, direction)` helper applying `NullPrecedence.LAST`, used by both `resolve()` and `resolveGrouped()`, so the two cannot drift again. Add a direct-path counterpart to `AttributeColumnSortITest.anObjectWithoutAValueSortsLastInBothDirections`.

<sub>flagged by guided-pr-review and superpowers; verified at both call sites</sub>

src/main/java/com/otilm/core/util/SortOrderBuilder.java
```diff
@@ -76,15 +99,56 @@ public static List<Order> resolve(Root<?> root, CriteriaBuilder criteriaBuilder,
...
-        Expression<?> sortKey = aggregate(criteriaBuilder, resolveExpression(root, field), sort.direction());
+            CommonAbstractCriteria query, SortSpecification sort) {
+        String fieldName;
+        Expression<?> sortKey;
+        if (sort.fieldSource() == FilterFieldSource.PROPERTY) {
+            FilterField field = resolveField(root, sort);
+            fieldName = field.name();
+            sortKey = aggregate(criteriaBuilder, resolveExpression(root, field), sort.direction());
+        } else {
+            fieldName = sort.fieldIdentifier();
```
lubomirw, line 110: **[Medium]** | Correctness | Confidence: High

The attribute branch has no catalogue gate, unlike the property branch. `resolveField` refuses a non-orderable property with `"Field %s cannot be used to order this listing."` (:199-202), but this else-branch goes straight to `getAttributeSortKey` with no equivalent.

Two consequences: a request can order by exactly the attribute fields `applyAttributeSortability` publishes as `sortable=false` (SECRET, CODEBLOCK, encrypted, `visible=false`), and an identifier naming no registered definition yields all-null keys and a silently unordered page — the failure the PR eliminated on the property side.

_Suggested:_ resolve the attribute field against the resource's catalogue (the data `ListViewServiceImpl.catalogueOf` already assembles from `AttributeEngine.getResourceSearchableFields`) and throw when the field is absent or reports `sortable=false`, mirroring the property message.

To be precise about what already works: malformed identifiers and unknown content types **do** throw today (`FilterPredicatesBuilder:1187-1200`). It is the existence-and-sortability check that is missing, not validation in general.

<sub>flagged by GPT 5.6, guided-pr-review and superpowers; verified against the property path</sub>

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -1128,4 +1124,111 @@ public static String buildPathToProperty(List<Attribute> joinAttributes, Attribu
...
+                        attributeType, contentType, attributeName, resource, objectUuidPath)
+                        .toArray(new Predicate[]{}));
+        ((JpaSubQuery) subquery)
+                .orderBy(criteriaBuilder.asc(subqueryRoot.get(AttributeContent2Object_.order)))
```
lubomirw: **[Medium]** | Correctness | Confidence: High

The subquery orders by `aco.order` alone before `fetch(1)`, but `getProjectedAttributesContent` orders `aci.attributeDefinitionUuid, aco.order` (`AttributeContent2ObjectRepository:149`).

One attribute name can map to several definition rows — `SearchHelper.mergeFieldsWithSameIdentifier` exists for exactly that — and `item_order` restarts at 0 per definition per object, so `FETCH FIRST 1` picks arbitrarily among the `order = 0` rows and the key can change between two identical requests. Even when it happens to be stable it is not the value the cell shows, which contradicts this method's own javadoc claim that "a column sorts by what the cell displays".

_Suggested:_ order the subquery by `aci.attributeDefinitionUuid, aco.order` to mirror the projection exactly. One line, and it makes the javadoc true.

<sub>flagged by GPT 5.6 and superpowers; verified against both ORDER BY clauses</sub>

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -1128,4 +1124,111 @@ public static String buildPathToProperty(List<Attribute> joinAttributes, Attribu
...
+        final Subquery subquery = query.subquery(valueClass);
+        final Root<AttributeContent2Object> subqueryRoot = subquery.from(AttributeContent2Object.class);
+        final Join joinContentItem = subqueryRoot.join(AttributeContent2Object_.attributeContentItem, JoinType.INNER);
+        final Join joinDefinition = joinContentItem.join(AttributeContentItem_.attributeDefinition, JoinType.INNER);
+
+        final Expression<String> extracted = criteriaBuilder
+                .function(JSONB_EXTRACT_PATH_TEXT_FUNCTION_NAME, String.class,
+                        joinContentItem.get(AttributeContentItem_.json),
+                        criteriaBuilder.literal(contentType.isFilterByData() ? "data" : "reference"));
+        final Expression<?> value = castedAttributeContentData.contains(contentType)
```
lubomirw, line 1247: **[Medium]** | Correctness | Confidence: High

FILE and RESOURCE attributes sort by a value the cell never shows. Both have `filterByData=false`, so this extracts `jsonb_extract_path_text(json, 'reference')`, while `AttributeColumnProjector.toProjectedValue` renders a reduced identity in the column: `fileName` + `mimeType` for FILE (`FILE_IDENTITY_FIELDS`), and `resource` / `uuid` / `name` for RESOURCE (`RESOURCE_IDENTITY_FIELDS`).

Neither content type is in `WITHHELD_CONTENT_TYPES` (only SECRET and CODEBLOCK are), so both stay displayable and `applyAttributeSortability` now marks them sortable. CREDENTIAL and OBJECT have the same shape.

_Suggested:_ either exclude the non-scalar content types from the sortability grant in `applyAttributeSortability`, or extract the same identity sub-path the projector renders (e.g. `data`->`fileName` for FILE) so the ordering matches the column.

<sub>flagged by GPT 5.6, Copilot CLI and superpowers; verified against AttributeContentType and the projector</sub>

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -161,23 +162,59 @@ public static SearchFieldDataDto prepareSearchForJSON(final SearchFieldObject at
...
+        return fields;
+    }
+
+    /**
+     * What the catalogue advertises as sortable: a field whose path can be ordered by, on a listing that applies the
+     * ordering.
      */
     private static boolean isSortable(final FilterField filterField) {
-        return false;
+        return SORT_WIRED_RESOURCES.contains(filterField.getRootResource()) && isOrderableField(filterField);
```
lubomirw, line 226: **[Medium]** | Correctness | Confidence: High

Usage columns sort by the persistence bitmask rather than the displayed value. `Certificate.keyUsage` and `CryptographicKeyItem.usage` are `private int` bitmask columns; the DTOs expose lists via `CertificateKeyUsage.convertBitMaskToSet` / `getUsage`. Both `KEY_USAGE` (`FilterField:129`) and `CKI_USAGE` (`FilterField:210`) carry `nativeArrayField=false`, no `jsonPath` and no `expectedValue`, so `isOrderableField` admits them and `isSortable` publishes them — ordering the page by an integer that has no relationship to the rendered usage list.

_Suggested:_ exclude bitmask-backed derived/list fields from `isOrderableField`, or define an explicit documented sort key for them.

<sub>flagged by GPT 5.6; verified in the entities and FilterField</sub>

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -161,23 +162,59 @@ public static SearchFieldDataDto prepareSearchForJSON(final SearchFieldObject at
...
+     * path is, because a catalogue reporting {@code true} there would advertise an ordering that listing discards: the
+     * client sorts, receives the default order, and is told nothing.
+     *
+     * <p>
+     * Explicit rather than derived: the wiring lives in each listing service and nothing on a {@link FilterField} knows
+     * whether its listing was wired. A listing that gains ordering adds itself here.
+     */
+    private static final Set<Resource> SORT_WIRED_RESOURCES = Set
+            .of(Resource.CERTIFICATE, Resource.CRYPTOGRAPHIC_KEY, Resource.DISCOVERY, Resource.CONNECTOR,
+                    Resource.SECRET, Resource.CBOM, Resource.SIGNING_RECORD);
```
lubomirw, line 186: **[Medium]** | Error handling | Confidence: Medium

The listings *not* in this set still accept a sort and drop it on the floor. `CryptographicAssetServiceImpl.validatePaging:219-224` sets the platform precedent — it throws a shaped `ValidationException` for any sort, citing that the contract permits sorting only on fields marked sortable. `LocationServiceImpl`, `EntityInstanceServiceImpl`, `VaultInstanceServiceImpl`, `VaultProfileServiceImpl`, `SigningProfileServiceImpl`, `TspProfileServiceImpl` and `TimeQualityConfigurationServiceImpl` accept `SearchRequestDto.sort` and ignore it.

`SORT_WIRED_RESOURCES` stops the catalogue from lying, but a client that ignores the flag still gets precisely the "sorts, receives the default order, and is told nothing" outcome the javadoc just above argues against.

_Suggested:_ follow the `CryptographicAssetServiceImpl` precedent in the remaining unwired listings, or centralise it so any listing that has not opted into `SORT_WIRED_RESOURCES` rejects rather than ignores.

Unverified: I confirmed the precedent verbatim but did not check each of the seven named listings individually.

src/test/java/com/otilm/core/integration/search/KeyListingSortITest.java
```diff
@@ -0,0 +1,168 @@
...
+                .toList();
+    }
+
+    private static SearchSortRequestDto sortByName(SortDirection direction) {
+        return new SearchSortRequestDto(FilterFieldSource.PROPERTY, "CKI_NAME", direction);
+    }
+
+    @Test
+    void ordersTheListingByTheRequestedFieldAscending() {
+        Assertions.assertEquals(SEEDED_NAMES, listNames(sortByName(SortDirection.ASC), 1, 10));
```
lubomirw, line 170: **[Medium]** | Testing | Confidence: Medium

The association-count realignment has no test. `getCountsOfAssociations` moved from a bare `List<Integer>` lined up by position to a uuid-keyed `KeyItemAssociationCount` projection joined on uuid — a real fix for a count that could render against the wrong key — but nothing asserts `KeyItemDto.associations`. The new Spring Data interface projection (a `COUNT` returning `Long` into an `int` getter, alias-to-getter binding) is therefore entirely unexercised, and this class asserts only names.

_Suggested:_ seed two key items with different numbers of associated certificates, list under a requested sort whose order differs from `createdAt DESC`, and assert each returned `KeyItemDto` carries its own count.

Unverified: I did not run the `setAssociations` / `getAssociations` grep across `src/test` myself.

src/test/java/com/otilm/core/integration/search/ColumnCatalogueFlagsITest.java
```diff
@@ -116,13 +141,30 @@ void aDerivedPropertyFieldWouldNotBeOrderable(FilterField filterField) {
...
 
         SearchFieldDataDto attribute = field(discoveryService.getSearchableFieldInformationByGroup(),
                 "catalogue-flag-probe|" + AttributeContentType.TEXT.name()).orElseThrow();
         Assertions.assertEquals(true, attribute.getDisplayable());
+        Assertions.assertEquals(true, attribute.getSortable());
+    }
+
+    /**
+     * A column the catalogue withholds cannot be ordered on either - there is nothing to order by when the value is
+     * never rendered. Secret content is the clearest case.
```
lubomirw, line 174: **[Medium]** | Testing | Confidence: High

This asserts only that the catalogue *reports* `sortable:false` for a withheld attribute field — nothing asserts that a request naming it is refused. The property path has both halves: this flag test plus the enforcement test `SecurityFilterRepositoryITest.rejectsSortFieldTheCatalogueWithholdsFromSorting`. The asymmetry in the tests mirrors the missing second lock in the code (see the comment on `SortOrderBuilder:110`).

Also uncovered: unknown attribute identifiers, names containing the `|` separator, complex content types, duplicate-definition determinism, and attribute sorting on the secret and CBOM listings.

_Suggested:_ add the attribute-path counterpart — register a SECRET (or `visible=false`) custom attribute, sort by it, assert a `ValidationException`. Written first it fails today, which is the point. Then add the regressions alongside the other fixes.

<sub>flagged by GPT 5.6 and superpowers; verified by comparing the two test classes</sub>

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -1128,4 +1124,111 @@ public static String buildPathToProperty(List<Attribute> joinAttributes, Attribu
...
+     * <p>
+     * An attribute may hold several values for one object, which leaves the key ambiguous. The subquery therefore
+     * orders by {@code item_order} and takes the first row, so a multi-valued attribute sorts on its lowest-ordered
+     * value - the one the cell shows first. A row with no value for the field yields no row and so a null key, which is
+     * ordered last in both directions by the caller rather than being dropped.
+     */
+    public static <T> Expression<?> getAttributeSortKey(final CriteriaBuilder criteriaBuilder,
+            final CommonAbstractCriteria query, final Root<T> root, final FilterFieldSource fieldSource,
+            final String fieldIdentifier) {
+        final String[] identifierParts = fieldIdentifier.split("\\|");
```
lubomirw: **[Low]** | Correctness | Confidence: High

The identifier is parsed with `split("\\|")` here, while `AttributeColumnProjector:192` deliberately parses the same format with `lastIndexOf(FIELD_IDENTIFIER_SEPARATOR)`. Attribute names carry no separator restriction in the interface model.

Concretely, for an attribute named `team|region` with content type TEXT, the identifier `team|region|TEXT` splits into three parts, `AttributeContentType.valueOf("region")` throws, and the caller gets `"Unknown attribute content type region"` — so a column the projector renders happily cannot be sorted, and the error names the wrong thing. (Not a silent misparse; it is a rejection with a misleading message.)

`getAttributeFilterPredicate:138` has the same `split`, so this is consistency debt the PR inherits rather than introduces — but the correct approach was available to copy from in the same blast radius.

_Suggested:_ extract one shared last-separator parser (name = before the last `|`, content type = after) and use it in projection, filtering and sorting.

<sub>flagged by GPT 5.6, Copilot CLI and superpowers; verified at all three sites</sub>

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -1128,4 +1124,111 @@ public static String buildPathToProperty(List<Attribute> joinAttributes, Attribu
...
+     * object correlation as the filter, extracted from the stored json with the same {@code jsonb_extract_path_text}
+     * and the same per-content-type cast, so a column sorts by what the cell displays.
+     *
+     * <p>
+     * An attribute may hold several values for one object, which leaves the key ambiguous. The subquery therefore
+     * orders by {@code item_order} and takes the first row, so a multi-valued attribute sorts on its lowest-ordered
+     * value - the one the cell shows first. A row with no value for the field yields no row and so a null key, which is
+     * ordered last in both directions by the caller rather than being dropped.
+     */
+    public static <T> Expression<?> getAttributeSortKey(final CriteriaBuilder criteriaBuilder,
```
lubomirw, line 1197: **[Low]** | Error handling | Confidence: Medium

`getAttributeSortKey` is public and validates the identifier format, but never checks the field source. Called with `FilterFieldSource.PROPERTY`, `fieldSource.getAttributeType()` is null and the correlation degrades to `ad.type IS NULL` — a query that silently matches nothing rather than an error.

Latent today, since only `resolveGrouped` calls it and that branches on the source first. But the method is public and the guard is one line.

_Suggested:_ reject a null `fieldSource.getAttributeType()` with a `ValidationException` at the top, beside the existing identifier-format checks.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -153,14 +155,8 @@ private static <T> Predicate getAttributeFilterPredicate(final CriteriaBuilder c
...
 
-        List<Predicate> predicates = new ArrayList<>();
-        predicates.add(criteriaBuilder.equal(joinDefinition.get(AttributeDefinition_.type), attributeType));
-        predicates.add(criteriaBuilder.equal(joinDefinition.get(AttributeDefinition_.contentType), contentType));
-        predicates.add(criteriaBuilder.equal(joinDefinition.get(AttributeDefinition_.name), attributeName));
-        predicates.add(criteriaBuilder.equal(subqueryRoot.get(AttributeContent2Object_.objectType), resource));
-        predicates
-                .add(criteriaBuilder
-                        .equal(subqueryRoot.get(AttributeContent2Object_.objectUuid), root.get(objectUuidPath)));
+        List<Predicate> predicates = new ArrayList<>(attributeCorrelationPredicates(criteriaBuilder, root, subqueryRoot,
```
lubomirw, line 156: **[Low]** | Maintainability | Confidence: High

`attributeCorrelationPredicates` is documented as shared by the filter predicate and the sort key "so the two cannot disagree about which rows belong to a field" — but the two inputs most likely to drift are still computed inline here: the `CryptographicKeyItem` -> `CRYPTOGRAPHIC_KEY` resource mapping and the `keyUuid`-vs-`uuid` choice for CUSTOM/DATA (:150-155). `getAttributeSortKey` calls the new `attributeResourceOf` / `attributeObjectUuidPath` helpers instead, so there are now two copies of exactly the rule the shared helper was meant to protect.

_Suggested:_ replace the inline ternaries with `attributeResourceOf(root)` and `attributeObjectUuidPath(resource, attributeType)`, and move the existing "place for improvement is to consolidate resource for attributes content" comment onto the helpers.

src/main/java/com/otilm/core/dao/repository/CertificateRepository.java
```diff
@@ -602,7 +602,11 @@ void updateCertificateIssuerDN(@Param("oid") String oid, @Param("newCode") Strin
             LEFT JOIN OwnerAssociation oa ON oa.objectUuid = c.uuid
             LEFT JOIN RaProfile ra ON ra.uuid = c.raProfileUuid
             WHERE c.uuid IN ?1
-            ORDER BY c.created DESC
             """)
+    /**
```
lubomirw: **[Low]** | Documentation | Confidence: High

This doc comment sits after the closing paren of `@Query` and before the method signature, so it is an ordinary block comment — javadoc will not attach it to the method. The explanation is therefore invisible in the generated API docs and in IDE hover, which is exactly where the next person would be about to re-add an `ORDER BY`.

The equivalent comment on `CryptographicKeyItemRepository.findFullByUuidIn` is placed correctly, before its annotation.

_Suggested:_ move the `/** ... */` block above the `@Query` annotation.

<sub>flagged by guided-pr-review and superpowers; verified</sub>

src/main/java/com/otilm/core/dao/repository/CryptographicKeyItemRepository.java
```diff
@@ -30,8 +30,13 @@ public interface CryptographicKeyItemRepository extends SecurityFilterRepository
     @EntityGraph(attributePaths = {"key", "key.tokenProfile"})
     List<CryptographicKeyItem> findByUuidIn(List<UUID> uuids);
 
+    /**
+     * The key items named by the uuid list. Deliberately carries no ordering: the ordering the listing asked for lives
+     * in the rank of that list, and the {@code OrderByCreatedAtDesc} this method used to name would replace it. Callers
```
lubomirw: **[Low]** | Documentation | Confidence: High

This javadoc records the removed method's history rather than its current contract: "the `OrderByCreatedAtDesc` this method used to name". Describe only the current unordered-query / rank-restoration contract — the reader needs to know the method carries no ordering and why, not what it was called before.

Same pattern in two more places added by this PR:
- `CertificateListingSortITest:31` — "The second query used to carry `ORDER BY c.created DESC`, which replaced the ordering the first had established"
- `KeyListingSortITest:36` — "The second query used to name its own ordering, which replaced the first's"

_Suggested:_ keep the present-tense rationale, drop the history. E.g. here: "Deliberately carries no ordering: the ordering the listing asked for lives in the rank of that list, and an ORDER BY here would replace it. Callers rank the result with `SortOrderBuilder.rankBy`."

<sub>flagged by GPT 5.6 and the hygiene pass; verified verbatim at all three sites</sub>

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -161,23 +162,59 @@ public static SearchFieldDataDto prepareSearchForJSON(final SearchFieldObject at
...
+
+    /**
+     * Marks the attribute fields of one resource's catalogue sortable where that resource's listing applies a sort.
+     *
+     * <p>
+     * Ordering an attribute resolves to a correlated scalar subquery over the stored json, which works on any resource;
+     * what varies is whether the listing passes the sort on. A field the catalogue withholds as a column is withheld
+     * from ordering too - there is nothing to order a column by when the column cannot be shown.
+     */
+    public static List<SearchFieldDataDto> applyAttributeSortability(final List<SearchFieldDataDto> fields,
```
lubomirw: **[Low]** | Maintainability | Confidence: High

`applyAttributeSortability` reads at its four `AttributeEngine` call sites as a pure transform wrapping `prepareSearchForJSON`, but it mutates each element in place via `field.setSortable(...)` and returns the same list. Safe today only because `prepareSearchForJSON` always builds fresh DTOs and nothing caches the catalogue — a fluent signature over an in-place mutation invites a future caller to hand it a shared list.

It is also a bolt-on the caller must remember: `prepareSearchForJSON` hardcodes `setSortable(false)` and every one of the four call sites must wrap it, so a fifth added later silently publishes `sortable:false` with no failure.

_Suggested:_ thread the resource into `prepareSearchForJSON` and decide `sortable` where `displayable` is decided, so the two flags cannot disagree and no call site can forget. Failing that, make it `void` and name it `markAttributeSortability`.

<sub>flagged by guided-pr-review and superpowers</sub>

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -161,23 +162,59 @@ public static SearchFieldDataDto prepareSearchForJSON(final SearchFieldObject at
...
     /**
-     * What the catalogue advertises as sortable, which is nothing yet.
+     * The resources whose listing passes the sort a request carries to the repository.
      *
      * <p>
-     * A secured search can be ordered by a field, but no listing service passes the sort a request carries to the
-     * repository, so a catalogue reporting {@code true} would advertise an ordering that does not happen: a client
-     * sorting on such a field would get the default order back and no indication why. {@link #isOrderableField} is the
-     * predicate this becomes once the listings are wired.
+     * The flag is published per field while the ordering is wired per listing, and seven of the sixteen listings that
```
lubomirw: **[Nit]** | Documentation | Confidence: High

"seven of the sixteen listings" is accurate today — `SORT_WIRED_RESOURCES` has exactly seven entries — but it is a counted claim in prose next to the set it counts, so it goes stale the moment a listing is wired and nobody thinks to update the sentence.

_Suggested:_ drop the numbers and keep the claim, e.g. "the flag is published per field while the ordering is wired per listing, and only the listings named below are wired."

<sub>flagged by GPT 5.6 and the hygiene pass</sub>

src/main/java/com/otilm/core/dao/repository/SecurityFilterRepositoryImpl.java
```diff
@@ -150,7 +151,7 @@ public List<UUID> findUuidsUsingSecurityFilter(final SecurityFilter filter,
      * query therefore groups by the root's uuid and orders by the aggregate the ordering resolves - one row per root,
      * which is what the window is allowed to cut.
      */
-    private List<UUID> findUuidsOrderedByJoinedField(final SecurityFilter filter,
+    private List<UUID> findUuidsOrderedBySortKey(final SecurityFilter filter,
```
lubomirw, line 179: **[Nit]** | Documentation | Confidence: Medium

The method was renamed from `findUuidsOrderedByJoinedField` to `findUuidsOrderedBySortKey` precisely because it now also serves attribute sorts, and `needsRankedUuidQuery`'s javadoc was updated to explain both reasons — but this method's javadoc (:145-155) still opens "The uuids of a page ordered by a field of a joined entity" and explains only the row-multiplication rationale, never the DISTINCT / select-list reason the attribute path needs it for.

_Suggested:_ extend it with the attribute-subquery case, mirroring the two-reason structure now in `needsRankedUuidQuery`.

lubomirw (changes requested):
Automated multi-reviewer triage — 19 findings (Request changes).

Reviewed at `5d12edfb1` (including the merged stacked #2179) against merge base `342d3cbfa`, through four independent reviewers — Copilot CLI, guided-pr-review, superpowers and a comment-hygiene pass — with a fifth external review folded in and deduped. Every finding below was re-verified against the code before posting; where I could not confirm the decisive fact myself, the comment says so in a clause.

The mechanics of the change are sound: the two-phase ranking, the `getCountsOfAssociations` realignment, and row-level authorization on the new ranked path all hold up, and there is no injection surface. Three things read as blocking:

1. **`FilterPredicatesBuilder:1225`** — the attribute sort subquery applies none of the gates the projection path applies.
2. **`SearchHelper:184`** — `SECRET` and `CBOM` are wired sortable but unmapped in `ResourceToClass`, so attribute sorting on those two listings silently does nothing.
3. **`SortOrderBuilder:82`** — nulls lead the page on a DESC direct sort while the grouped path pins them last.

Findings 1, 2 and 4 are closely related and largely close each other: the catalogue gate suggested on `SortOrderBuilder:110` covers most of what finding 1 needs, and the `ResourceToClass` entries fix both the silent no-op and the disabled cross-resource guard in one edit. Findings 5 and 3 are one line each.

### Not anchored to a diff line

- **`SearchHelperColumnFlagsTest.java:60`** — Nit | Testing | Confidence: Medium. `noAttributeFieldIsSortableUntilAttributeSortingLands` still passes, because `prepareSearchForJSON` defers the flag — but attribute sorting has now landed, so the name describes a state of the world that no longer exists and reads as a contradiction next to `applyAttributeSortability`. _Suggested:_ rename it to say what it actually pins — that the per-definition builder leaves `sortable` false because the decision is made per resource. (Out of diff, so no inline anchor.)

### Raised and deliberately not included

For the record, four further findings came out of the review and were set aside rather than posted: the `ListViewServiceImpl.validateSort` gap that lets a saved view 400 its own listing, the `CKI_NAME` → `DISCOVERY_NAME` published-identifier change shipping without a migration, the correlated-subquery cost at production volume, and the missing `DISTINCT` on `createCriteriaBuilderUuid` under group-scoped authorization (AC5's second half). Happy to open them as separate issues if useful.

One correction worth recording, since it came up during review and is easy to get wrong: the attribute correlation *is* index-backed. `V202604091500__attribute_content_versioning.sql:9-10` creates `idx_aco_type_uuid_version ON attribute_content_2_object (object_type, object_uuid, object_version)`, whose leading prefix matches the correlation key exactly. The residual performance question is the per-row subquery evaluation before paging and the non-indexable `jsonb_extract_path_text` + cast — not a missing index.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
Jakub-Bartal-01: Confirmed and fixed. The sort subquery correlated on definition identity only; none of the projection's gates applied.

Routed through the same gates rather than duplicating them:

- New `ListingSortResolver` (`com.otilm.core.attribute.engine`) builds the `SortSpecification` for a listing. For an attribute-sourced sort it resolves the identifier against the resource's own published catalogue (`AttributeEngine.getResourceSearchableFields`) and refuses anything not reported `sortable`, then attaches `AttributeEngine.loadCustomAttributeContentFilter()` — the same record `AttributeColumnProjector` reads — to the specification. All seven wired listings build their sort through it now instead of `SortSpecification.from`.
- `getAttributeSortKey` adds `attributeReadabilityPredicates`: `aci.encrypted_data IS NULL` always, plus `ad.enabled = true` and the allowed/forbidden definition-uuid predicates for `AttributeType.CUSTOM` — the same three `getProjectedAttributesContent` applies.
- A specification carrying no content filter is refused rather than run unrestricted, so a listing wired without the resolver fails loudly instead of ordering by content the caller may not read.

The catalogue gate covers the `visible` / SECRET / CODEBLOCK / encrypted half, as you said it would.

Regression test: `AttributeColumnSortITest.anAttributeTheCallerMayNotReadCannotOrderThePage` restricts ATTRIBUTE/MEMBERS to an allow-list the definition is not on and asserts the ASC and DESC pages are identical — every key null, uuid tie-break deciding both. Verified to fail when the permission predicates are removed.

Agreed on the calibration: `getAttributeFilterPredicate` has the same gap and a stronger oracle. Threading the content filter into `getFiltersPredicate` reaches every listing service, so it is not in this PR and wants a change of its own.

One residual, stated so it is not mistaken for coverage: `visible` is enforced by the catalogue gate, and the catalogue merges definitions sharing a name and content type (`SearchHelper.mergeFieldsWithSameIdentifier`), so two definitions of one identifier that disagree about `visible` are answered by whichever survived the merge. The projector re-checks per row because it holds the definition document; a query-level equivalent would mean extracting `visible` out of `ad.definition` in SQL across two attribute-document versions, which I did not want to guess at here. It is the same merge behaviour the filter path has had.

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Confirmed. `ResourceToClass` had no `Secret` or `Cbom` entry, `getResourceByClass` returned null, and the correlation became `objectType = null`.

Two changes:

- `ResourceToClass` gains `SECRET`, `CBOM` and `OID`. This also repairs the pre-existing attribute *filter* on secrets, which was broken the same way.
- The sort no longer guesses the resource from the root. `SortSpecification` carries the resource the listing selects, supplied by `ListingSortResolver` at the call site, and `getAttributeSortKey` uses it; the root is only a fallback for a specification built without one (direct repository calls in tests). `SortOrderBuilder.resolveField` prefers `sort.resource()` over `resourceOf(root)` for the cross-resource guard, so the listing path can no longer skip it.

Pushing back on making an unmapped sort root fail closed: `CustomOidEntry` has no `ResourceToClass` entry either, and `SubclassFilterFieldSearchITest.ordersBySubclassDeclaredCodeAcrossEveryCategory` sorts that root through the repository directly — failing closed there breaks a legitimate, tested path. Carrying the resource from the caller is the stronger form of the same fix, and OID is mapped now as well.

End-to-end coverage: `AttributeColumnSortITest.ordersASecretListingByTheAttributeValue` orders the real secret listing by a custom attribute. There is no CBOM counterpart because `Resource.CBOM.hasCustomAttributes()` is false — `updateCustomAttributeDefinition` rejects the resource, so no custom content can exist to sort by. CBOM stays in `SORT_WIRED_RESOURCES` for its property fields.

src/main/java/com/otilm/core/util/SortOrderBuilder.java
Jakub-Bartal-01: Confirmed. `resolve()` added its term with no null precedence while `resolveGrouped` pinned `NullPrecedence.LAST`.

Extracted `primary(criteriaBuilder, expression, direction)` applying `NullPrecedence.LAST`, used by both paths.

Tests: `SecurityFilterRepositoryITest.aRowWithoutAValueSortsLastInBothDirectionsOnTheDirectPath` (nullable `commonName`, DESC) and `...OnTheGroupedPath`. Both fail when the null precedence is dropped from the shared helper.

src/main/java/com/otilm/core/util/SortOrderBuilder.java
Jakub-Bartal-01: Confirmed and fixed. The attribute branch resolves the field against the resource's published catalogue first — `ListingSortResolver.requireSortableAttributeField` reads `AttributeEngine.getResourceSearchableFields` and throws `Unknown sort field identifier %s.` when the field is absent and `Field %s cannot be used to order this listing.` when it reports `sortable:false`, matching the property path's messages.

Reading the flag off the catalogue rather than reimplementing its rules is deliberate: the value the client was given and the answer it gets back come from the same computation.

Tests in `AttributeColumnSortITest`: `aSecretAttributeCannotBeOrderedOn`, `aHiddenAttributeCannotBeOrderedOn`, `anAttributeWhoseCellIsNotItsSortKeyCannotBeOrderedOn`, `anIdentifierNamingNoRegisteredAttributeIsRejected`.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
Jakub-Bartal-01: Confirmed and fixed. The subquery orders by `aci.attributeDefinitionUuid, aco.order` now, mirroring `getProjectedAttributesContent`, and the javadoc says so.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
Jakub-Bartal-01: Half confirmed.

FILE and RESOURCE: agreed. Both are excluded from the sortability grant — `SearchHelper.COMPOSITE_CELL_CONTENT_TYPES`.

CREDENTIAL and OBJECT are not the same shape. `AttributeColumnProjector.toProjectedValue` maps both through `referenceOnly`, so the cell carries the reference and nothing else — which is exactly what the sort key extracts. Ordering already matches the column there, so they stay sortable.

Chose exclusion over extracting `data->>'fileName'` / `data->>'name'`: it is what `isOrderableField` already does for property fields whose display is not what the attribute holds, and it avoids a second per-content-type identity mapping that could drift from the projector's.

Tests: `SearchHelperColumnFlagsTest.anAttributeWhoseCellIsNotItsSortKeyIsNotSortable` and `anAttributeWhoseCellIsItsReferenceIsSortable`; `ColumnCatalogueFlagsITest.anAttributeWhoseCellIsNotItsSortKeyIsDisplayableButNotSortable`.

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Confirmed and fixed. `isOrderableField` excludes bitmask-backed fields via `BitMaskEnum.class.isAssignableFrom(filterField.getEnumClass())` — derived rather than listed, so `CERT_REQUEST_KEY_USAGE` and any bitmask field added later are covered without an edit.

Test: `ColumnCatalogueFlagsITest.aBitmaskBackedPropertyFieldWouldNotBeOrderable` over `KEY_USAGE`, `CKI_USAGE` and `CERT_REQUEST_KEY_USAGE`.

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Confirmed at each site, and centralised as you suggested. `RequestValidatorHelper.revalidateSearchRequestDto(dto, resource)` throws when the request names a sort and `SearchHelper.listingAppliesSort(resource)` is false, so a listing that has not opted into `SORT_WIRED_RESOURCES` rejects rather than ignores.

Applied to nine listings: your seven plus the audit-log and custom-OID listings, which have the same shape. `CryptographicAssetServiceImpl` keeps its own message. Five of the nine did not call `revalidateSearchRequestDto` at all and read `getPageNumber()` unguarded, so they pick up the null defaults with it.

Test: `ColumnCatalogueFlagsITest.anUnwiredListingRefusesASortRatherThanDiscardingIt`.

src/test/java/com/otilm/core/integration/search/KeyListingSortITest.java
Jakub-Bartal-01: Confirmed — nothing asserted `KeyItemDto.associations`.

Added `KeyListingSortITest.eachListedKeyCarriesItsOwnAssociationCount`: two certificates on `alpha`, one on `charlie`, none on the other two, then `[2, 0, 1, 0]` under a name-ASC sort and `[0, 1, 0, 2]` under DESC. A count lined up by position rather than by uuid fails one of the two directions. `aKeyWithoutCertificatesReportsNoAssociations` covers the zero case, which is what exercises the `getOrDefault` fallback rather than the projection.

src/test/java/com/otilm/core/integration/search/ColumnCatalogueFlagsITest.java
Jakub-Bartal-01: Confirmed. The enforcement half is in `AttributeColumnSortITest` now: `aSecretAttributeCannotBeOrderedOn` registers a SECRET custom attribute and asserts the listing refuses a sort naming it, with `aHiddenAttributeCannotBeOrderedOn` for `visible=false`.

Of the rest you listed: unknown identifiers are `anIdentifierNamingNoRegisteredAttributeIsRejected`, complex content types are `anAttributeWhoseCellIsNotItsSortKeyCannotBeOrderedOn`, and attribute sorting on the secret listing is `ordersASecretListingByTheAttributeValue`. Duplicate-definition determinism is now a property of the query rather than of the data (the subquery orders by definition uuid then `aco.order`). Names containing `|` are covered by the shared parser rather than by a case of their own — see the reply on `FilterPredicatesBuilder:1186`.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
Jakub-Bartal-01: Confirmed and fixed. New `com.otilm.core.model.AttributeFieldIdentifier` splits at the last separator and is used by the projector, the attribute filter and the attribute sort.

The filter path had the same bug and is fixed with it; it also answers a malformed identifier with a `ValidationException` now instead of the raw `IllegalArgumentException` `AttributeContentType.valueOf` threw.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
Jakub-Bartal-01: Confirmed and fixed. `getAttributeSortKey` rejects a source with no attribute type before anything else. It also rejects a specification carrying no custom-attribute content filter, which is the same guard in its stronger form: the method now refuses to build a query it cannot authorize, rather than one that silently matches nothing.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
Jakub-Bartal-01: Confirmed and fixed. The filter path calls `attributeResourceOf(root)` and `attributeObjectUuidPath(root, attributeType)`, so the two rules exist once.

`attributeObjectUuidPath` is keyed on the root rather than on the resource now: the key uuid is a column of the key item and of nothing else, so a root that is not one has no such path to read whatever resource its content is filed under. Behaviour is unchanged — `CRYPTOGRAPHIC_KEY` only ever arrives with a key-item root — and the "place for improvement is to consolidate resource for attributes content" note moved onto `attributeResourceOf`.

src/main/java/com/otilm/core/dao/repository/CertificateRepository.java
Jakub-Bartal-01: Confirmed and fixed. The method already had a javadoc block above `@Query`; the ordering note is merged into it rather than stacked as a second block.

src/main/java/com/otilm/core/dao/repository/CryptographicKeyItemRepository.java
Jakub-Bartal-01: Confirmed and fixed at all three sites — each now states the current contract in the present tense with no reference to what the method or query used to be called.

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Confirmed, and taken the first option rather than the fallback. `applyAttributeSortability` is gone; `prepareSearchForJSON` takes the resource and decides `sortable` where `displayable` is decided, so the two flags come from one place and a fifth call site cannot forget to apply it.

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Fixed — the counts are dropped and the claim stands on the set below it.

src/main/java/com/otilm/core/dao/repository/SecurityFilterRepositoryImpl.java
Jakub-Bartal-01: Confirmed and fixed. The javadoc carries both reasons now: the join case that multiplies rows a window would cut, and the attribute case where the entity query's DISTINCT will not order by an expression absent from the select list — and what this query does about both, which is to select the key alongside the uuid and group by it.

lubomirw (approved):

PR #2181, opened 2026-09-01
Surface request-attribute validation warnings under lenient mode

### LINKED ISSUE #1951: Surface request-attribute validation warnings under lenient mode
### Description

Request-attribute policy warnings are produced but thrown away. `ProtocolRequestAttributeValidator.validate(...)` returns `void` and its only consumer of warnings is a `log.warn` line; `RequestAttributeValidationResult.warnings` never reaches a response. fe-administrator#1780 recorded the consequence when it shipped the external-CSR compliance panel: *"lenient warnings are not returned by the current API, so there is nothing to display for lenient (a backend follow-up is needed to surface them)"*. That follow-up was never filed — this is it.

Two changes:

**1. Run the whitelist under lenient.** `ProtocolRequestAttributeValidator` builds `new RequestAttributePolicy(strict, strict)`, so lenient validation skips the RDN / SAN / extension whitelist entirely and produces no findings at all. Build `new RequestAttributePolicy(strict, true)` instead: the whitelist always runs, and `RequestAttributePolicy` already downgrades errors to warnings when not strict. An operator can then see which extensions their real ACME / SCEP / CMP / EST clients actually send *before* switching the profile to strict, instead of discovering it from `badCsr` responses in production.

**2. Return the warnings.** Populate `ClientCertificateDataResponseDto.requestAttributeWarnings` (added by the contract issue) from the validation result on the REST issuance paths, so the operator UI can display them. `validate(...)` needs to hand its result back rather than swallow it.

Scope notes:

- Protocol flows (ACME / SCEP / CMP / EST) have RFC-shaped responses with nowhere to carry warnings; they stay log-only. Keep the log line.
- Under strict nothing changes: a violation is still rejected with 422 and the existing error list.
- Keep the warning list out of `toLogData()` on the response DTO — audit records must not carry a variable-length list.

This is the surviving half of OmniTrustILM/core#1883, which is closed as not planned: its proposed built-in allowlist of "benign" standard extensions was rejected, because strict must mean "only what the resolved set maps". The structured Key Usage / Extended Key Usage mapping targets under OmniTrustILM/ilm#325 are what make that configurable in practice.

### Follow-up considered and deferred

A pre-submit validation endpoint (e.g. `POST /v2/operations/authorities/{authorityUuid}/raProfiles/{raProfileUuid}/csr/validate`) returning errors *and* warnings would let the operator see problems before issuing rather than after acceptance, and would work for strict too. Deliberately out of scope here — file separately if wanted.

### Acceptance Criteria

- [ ] Lenient validation runs the whitelist and produces warnings instead of skipping it; strict behaviour unchanged.
- [ ] Warnings returned in `ClientCertificateDataResponseDto.requestAttributeWarnings` on the REST issuance paths (`POST /certificates`, `/certificates/{uuid}/issue`, `/renew`, `/rekey`, `/certificates/register`).
- [ ] Warnings still logged for protocol flows, which carry no warning field.
- [ ] Warning list absent from `toLogData()`.
- [ ] Tests: a lenient profile with an unmapped extension returns a warning and still issues; the same profile under strict returns 422; a fully mapped CSR returns no warnings.
- [ ] `mvn -B verify` passes; new-lines coverage meets the Sonar gate.

klaraf755:
Closes #1951

Request-attribute policy warnings were produced and thrown away: `ProtocolRequestAttributeValidator.validate(...)` returned `void`, and lenient validation skipped the whitelist entirely, so it produced no findings at all. An operator had no way to see what their real ACME / SCEP / CMP / EST clients send before switching a profile to strict.

## Changes

**1. The whitelist always runs.** `ProtocolRequestAttributeValidator` built `new RequestAttributePolicy(strict, strict)`, which meant lenient validation never ran the RDN / SAN / extension whitelist. It now builds `new RequestAttributePolicy(strict, true)`; `RequestAttributePolicy` already routes findings to warnings when not strict, so strict behaviour is unchanged.

**2. The warnings are returned.** `validate(...)` now returns its warning list instead of only logging it. Inside `ClientOperationServiceImpl` the list travels to the response via two private carrier records, and `ClientCertificateDataResponseDto.requestAttributeWarnings` is populated on the REST issuance paths.

`submitCertificateRequest` is declared on two service interfaces, so its signature is untouched: the body moved to a private `submitCertificateRequestInternal` returning the carrier, and the annotated public method delegates. This is behaviour-identical — issue/renew/rekey already reached the body by self-invocation (which is why the explicit `checkCreatePermissions()` probe is load-bearing), and the proxied callers (`CertificateControllerImpl`, `ScepServiceImpl`) still enter through the annotation.

**3. `/certificates/{uuid}/issue` now validates its CSR.** That path accepts an operator CSR for a `REGISTERED` placeholder but ran no request-attribute validation at all — neither there nor in the async `issueCertificateAction`. Validation is placed after the challenge gate (so a caller who cannot authorize cannot probe the policy) and before the CSR attach (so a strict violation persists nothing). Under strict this is a behaviour change: a policy-violating CSR on this endpoint is now rejected with 422, consistent with the other uploaded-CSR paths. See **Behaviour changes beyond REST** below — that change is not confined to this endpoint, and not confined to strict.

Protocol flows keep the `log.warn` and carry no warning field — their RFC-shaped responses have nowhere to put one.

## Behaviour changes beyond REST

Two things item 3 implies that are worth stating explicitly, since a reader of the summary above would not infer them.

**It reaches SCEP and CMP.** `issueExistingCertificate` has two protocol callers, and `RequestAttributePolicyViolationException` extends `ValidationException`. So under a strict RA profile, a policy-violating CSR now also fails SCEP and CMP *registration completion*, not just the REST endpoint. Both are shaped the way each protocol's plain enrolment path already shapes the same exception — SCEP `BAD_REQUEST` carrying the platform-authored message, CMP `badCertTemplate` — rather than being folded into the generic anti-enumeration rejection, which would have been indistinguishable from "no registration matches your CSR".

**It is not strict-only.** `runKernel` converts an `X509RequestContentParser` failure into a policy violation regardless of the strict flag, and `/certificates/{uuid}/issue` plus the two completion paths reach that parser for the first time here. So a CSR that `CertificateRequestUtils.createCertificateRequest` accepts but the content parser rejects — a malformed `otherName`, an `iPAddress` that is not 4 or 16 octets — now fails a completion that previously succeeded, with strict off. Gating that on strict was considered and rejected: it would contradict `rejectsMalformedUploadedCsr_asValidationError_notServerFault`, which pins 422-under-lenient on the existing submit path. Aligning the two is a separate change.

**A third state on the wire.** Lenient validation fails open, so a resolution failure used to return an empty warning list — indistinguishable from "checked, nothing found". Since an operator reads that empty list as licence to switch the profile to strict, a resolution failure under lenient now returns one platform-authored warning saying validation could not be performed, and why.

## Scope notes

- **`POST /certificates/register` carries no warnings, by construction.** It takes no CSR — `subjectDn` / `subjectAltName` / `extensions` / `csrAttributes`, validated by the attribute engine and `CertificateRequestAttributeProjector`, never the whitelist. The acceptance criteria list it, but there is no external CSR to warn about, so the field stays at its default empty list.
- **`CertificateRequestContentValidator.validate(CertificateRequest, List, boolean)` was left alone.** It carries the same `(!lenient, !lenient)` pattern, but has no production caller — only its own test class exercises it. Not touched here rather than half-fixing dead code.
- `requestAttributeWarnings` is `List<String>`, matching the strict error path: `ExceptionHandlingAdvice.handleValidationException` already flattens `ValidationError` to `getErrorDescription()`, so the 422 body is a JSON array of strings too.

## Testing

`mvn -B verify`: **6297 tests, 0 failures, 0 errors, 2 skipped**; Spotless clean, 0 Checkstyle violations.

New coverage:
- `ProtocolRequestAttributeValidatorTest` — lenient returns the missing-required-field warning; lenient warns on an unmapped extension; strict rejects the same CSR with that finding as an error; a fully mapped CSR yields no warnings.
- `ClientOperationRequestAttributePropagationITest` — a lenient profile returns the warning on the issue response and still issues; a fully mapped CSR returns none; the response's `toLogData()` carries only the uuid.
- `ClientOperationServiceRegisterITest` — `/certificates/{uuid}/issue` returns the warning under lenient, and under strict rejects with nothing attached to the placeholder.

`ContextSignatureGuardTest.BASELINE` is unchanged at 61 — every test went into an existing class, and the register suite autowires the real `RaProfileCertificateRequestAttributeService` rather than adding a mock type.


src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1875,13 +1895,27 @@ public ClientCertificateDataResponseDto issueExistingCertificate(final SecuredPa
...
+            // strict violation rejects with nothing persisted. After the challenge gate above, so a caller who
+            // cannot authorize cannot probe the policy. A null format is not validated here: the parser selects on
+            // format, and the attach below already rejects it with a clear message.
+            if (request.getFormat() != null) {
+                try {
+                    requestAttributeWarnings = validateUploadedRequestAttributes(request.getRequest(),
+                            request.getFormat(), raProfile);
+                } catch (CertificateException e) {
+                    throw new ValidationException(
+                            ValidationError.create("Invalid certificate signing request: " + e.getMessage()));
```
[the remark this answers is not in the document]
klaraf755: Good catch — fixed in 1a4ce4a.

The catch conflated a local CSR parse failure (client fault, 422) with the `CertificateException` a strict profile raises when its attribute set cannot be resolved (authority outage). It now parses locally and shapes only that as a 422, letting the availability failure propagate to the 500 handler; CMP and SCEP shape it into their native issuance failure. No interfaces change — `ClientOperationController` already declared `throws CertificateException`.

Regression test: `issueExistingReportsUnresolvableAttributeSetAsServerFailure_notClientFault`.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1875,13 +1895,17 @@ public ClientCertificateDataResponseDto issueExistingCertificate(final SecuredPa
...
         if (registered) {
             // The CSR attach and completion attributes commit before the async ISSUE probe can deny, so gate here.
             // A verified challenge is itself the holder's authorization; demanding the operator permission on top
             // would make challenge-based self-service unusable for the identities it exists for.
             if (!challengeAuthorized) {
                 certificateService.checkCreatePermissions();
             }
+            // Runs after the challenge gate above, so a caller who cannot authorize cannot probe the policy, and
+            // before the attach below, so a strict violation rejects with nothing persisted.
+            requestAttributeWarnings = validateCompletionRequestAttributes(request, raProfile);
```
vyskocilm, line 1909: 🔴 **bug — the authority connector's raw response body now reaches a 500 REST body.**

This is a new unshaped exit. When the request-attribute set cannot be resolved, `ProtocolRequestAttributeValidator.resolveDefinitions` wraps the `ConnectorException` as the *cause* of a `CertificateException`, and that propagates out of `issueExistingCertificate` and out of the controller (which declares `throws CertificateException` and adds no shaping). `ExceptionHandlingAdvice.handleCertificateException` then does `messageBuilder.append(", cause=").append(ex.getCause().getMessage())` — and for a connector HTTP failure that message *is* the upstream response body verbatim: `BaseApiClient` builds `ConnectorClientException` / `ConnectorServerException` straight from `clientResponse.bodyToMono(String.class)`. Whatever the connector put in its error body — stack text, SQL, internal identifiers, vendor detail — is echoed to the API caller.

The sibling path already refuses exactly this: `submitAndShapeFailure`'s javadoc says a connector exception "relaying upstream text … contributes no message". The three protocol callers of this method all catch the exception (`ScepServiceImpl:1108`, `CrmfIrCrMessageHandler:170`), so the REST controller is the only exit that leaks.

Not hypothetical — it's the scenario your own new test `issueExistingReportsUnresolvableAttributeSetAsServerFailure_notClientFault` sets up (strict profile, MERGE mode, connector throws). The test calls the service directly, so the leak is invisible to it.

**Fix:** shape it before it leaves the service — catch it here, log the original with its cause, and rethrow a `CertificateException` carrying only platform-authored text and no cause, mirroring `validateUploadedRequestAttributes`. Please don't fix it in the advice: that handler is shared, and changing it is a far larger blast radius than this PR.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -3070,35 +3099,60 @@ private void validateSubjectDnForCertificate(String certificateContent, Certific
...
+            RaProfile raProfile) throws CertificateException {
+        if (request.getFormat() == null) {
+            return List.of();
+        }
+        CertificateRequest parsedRequest;
+        try {
+            parsedRequest = CertificateRequestUtils.createCertificateRequest(request.getRequest(), request.getFormat());
+        } catch (CertificateRequestException | IllegalArgumentException e) {
+            throw new ValidationException(
+                    ValidationError.create("Invalid certificate signing request: " + e.getMessage()));
```
vyskocilm: 🟡 **risk — this forwards a JDK/BouncyCastle runtime message into a 422 body, re-laundering the text the global advice deliberately suppresses.**

`handleIllegalArgumentException` carries an explicit rationale for *not* returning that message: it "also sees what the JDK and third-party libraries throw — an internal class name or a serializer's reference chain; a rejection whose message the caller should read belongs in `ValidationException` instead". Wrapping the message in a `ValidationException` routes it to `handleValidationException`, which returns the descriptions to the client — defeating that suppression.

Your own sibling method twenty lines below (line 3138-3140) makes the opposite choice on the identical exception pair: fixed text, cause logged at debug. Concretely reachable today: `normalizeAndDecodeCsr` does a bare `Base64.getDecoder().decode(...)`, so a mis-encoded CSR yields JDK text like `Illegal base64 character 2d`. Low impact on today's inputs, but the guard is on the exception *type*, not on the content, so it forwards whatever a future library version throws.

**Fix:**
```java
logger.debug("Failed to parse completion CSR for request-attribute validation", e);
throw new ValidationException(ValidationError.create("Certificate signing request could not be parsed"));
```

src/main/java/com/otilm/core/certificate/request/ProtocolRequestAttributeValidator.java
```diff
@@ -33,16 +33,16 @@ public ProtocolRequestAttributeValidator(RaProfileCertificateRequestAttributeSer
...
-    public void validate(CertificateRequest request, RaProfile raProfile) throws CertificateException {
+    public List<String> validate(CertificateRequest request, RaProfile raProfile) throws CertificateException {
         if (raProfile == null) {
-            return;
+            return List.of();
         }
         List<BaseAttribute> definitions = resolveDefinitions(raProfile);
         if (definitions.isEmpty()) {
-            return;
+            return List.of();
```
vyskocilm, line 53: 🟡 **risk — "not checked" and "fully compliant" are indistinguishable on the wire, which defeats the feature this PR exists to deliver.**

`resolveDefinitions` swallows a `ConnectorException`/`NotFoundException` and returns an empty list, so this early return hands back no warnings. To the operator reading the response, `"requestAttributeWarnings": []` means "your clients are clean".

The failure mode is precisely the one the feature is meant to prevent: an operator on a lenient profile whose authority connector is down — or whose attribute set was never configured on the connector — issues certificates, sees an empty warning list, concludes the fleet is compliant, flips the profile to strict, and every real client request starts returning 422.

**Fix:** distinguish the two states. Simplest version is to return a single warning saying validation could not be performed (and why, in platform-authored words), so an empty list keeps meaning "checked, nothing found".

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -1105,6 +1105,13 @@ private ScepResponse completeRegistration(ScepRequest scepRequest, Certificate m
             scepRegistrationTrackingWriter.discardPollMapping(scepRequest.getTransactionId(), scepProfile.getUuid());
             logger.info("SCEP registration completion rejected: {}", e.getMessage());
             throw new ScepException(REGISTRATION_REJECTION, FailInfo.BAD_MESSAGE_CHECK);
```
vyskocilm, line 1113: 🟡 **risk — the strict-mode behaviour change reaches SCEP and CMP registration completion too, and neither the description nor #1951 says so.**

The description frames item 3 as a change to one REST endpoint. But `issueExistingCertificate` has two protocol callers, and `RequestAttributePolicyViolationException` extends `ValidationException` — so a strict policy violation now lands in this `catch (ValidationException | NotFoundException)` and in `CrmfIrCrMessageHandler`'s, and both convert it into the generic anti-enumeration rejection: SCEP `BAD_MESSAGE_CHECK` + `REGISTRATION_REJECTION`, CMP `badMessageCheck` + the same text. That's a new rejection path for real SCEP and CMP registration completions against any strict RA profile, indistinguishable on the wire from "no registration matches your CSR".

Two separate things to settle:

1. **The record.** Add a line to the description and to #1951 — someone reading only the PR body would not know to look. No code change needed for this part.
2. **The failInfo.** Both *other* SCEP enrolment sites (around lines 682 and 741) catch this exception explicitly and return `BAD_REQUEST` with the platform-authored message, commented as safe. This completion path is the odd one out, so the same policy-violating CSR gets a different — and misleading — failInfo depending on whether it arrives via register→complete or via plain enrolment. Worth catching `RequestAttributePolicyViolationException` before the `ValidationException` arm here for consistency.

Diagnosability is fine on this path at least: line 1106 logs the denial message.

src/main/java/com/otilm/core/service/cmp/message/handler/CrmfIrCrMessageHandler.java
```diff
@@ -167,7 +167,7 @@ private ClientCertificateDataResponseDto completeRegistration(CertReqMessages cr
             // mean "no active registration to complete" — surface the single generic rejection, detail logged.
             throw new CmpProcessingException(tid, PKIFailureInfo.badMessageCheck,
```
vyskocilm, line 172: 🟡 **risk — on the CMP path the same newly-reachable exception has its detail destroyed entirely.**

A strict `RequestAttributePolicyViolationException` now lands in the `catch (ValidationException | NotFoundException e)` above. Here `e` is neither logged nor attached as a cause, and `CmpProcessingException(tid, failureInfo, errorDetails)` doesn't log either — so nothing survives.

Concretely: a CMP client completes a pre-registration under a strict profile whose set maps only CN, the CRMF carries `keyUsage`, the client receives `badMessageCheck` plus the generic registration-rejection text (which reads as a MAC failure), and the server log contains nothing at all about why. The ordinary ir/cr path ~35 lines above deliberately shapes the same exception into `badCertTemplate` with the platform-authored message.

**Fix:** add a `catch (RequestAttributePolicyViolationException e)` arm before the `ValidationException` one — or at minimum log `e` here.

Separately: nothing in the suite exercises the `CertificateException` arm of the widened catch on the next line.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
```diff
@@ -1105,6 +1105,13 @@ private ScepResponse completeRegistration(ScepRequest scepRequest, Certificate m
...
+        } catch (CertificateException e) {
+            // A strict RA profile whose request-attribute set cannot be resolved is an authority-side outage, not
+            // a client fault: no ISSUE reached the broker, so drop the staged mapping and report a server failure
+            // rather than the anti-enumeration rejection above.
+            scepRegistrationTrackingWriter.discardPollMapping(scepRequest.getTransactionId(), scepProfile.getUuid());
+            logger.error("SCEP registration completion could not be validated", e);
+            throw new ScepException("Unable to complete certificate registration");
```
vyskocilm: 🔵 **nit — the single-argument constructor silently sends `BAD_REQUEST`, while the comment three lines above says "report a server failure".**

`new ScepException(String)` defaults to `FailInfo.BAD_REQUEST` ("Transaction not permitted or supported"). A SCEP client reading `badRequest` treats its CSR as permanently invalid and does not retry — the wrong reaction to an authority-connector outage, which is the case this arm exists for. Every other `ScepException` in this file passes a `FailInfo` explicitly. The cause is also dropped (`e` is logged, not chained).

SCEP has no server-failure failInfo, so `BAD_REQUEST` may well be the only option — but pass it explicitly so the intent is visible, and consider softening the comment, which currently promises more than the protocol can express.

This arm (lines 1112-1114) is also untested: no test in the repo names this message, and `ScepRegistrationEnrolmentITest` never mentions `externalCsrValidationStrict` or `resolveIssueAttributeSet`.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -3070,35 +3099,60 @@ private void validateSubjectDnForCertificate(String certificateContent, Certific
...
+     * Validates the operator's completion CSR for a registered certificate against the resolved request-attribute set.
+     *
+     * <p>
+     * A null format is not validated: the parser selects on format, and the CSR attach rejects it with a clear message.
+     * Only the parse is a client fault — a strict profile whose attribute set cannot be resolved is a server-side
+     * inability, so that {@link CertificateException} propagates to the caller unshaped.
+     */
+    private List<String> validateCompletionRequestAttributes(ClientCertificateIssueRequestDto request,
+            RaProfile raProfile) throws CertificateException {
+        if (request.getFormat() == null) {
```
vyskocilm, line 3120: 🔵 **nit — the javadoc above asserts behaviour the code doesn't have: a null format is not rejected with a clear message, it's an unhandled NPE.**

Line 3106 justifies this guard by promising the downstream attach rejects a null format cleanly. It doesn't. `addCertificateRequestToExisting` calls `CertificateRequestUtils.createCertificateRequest(String, format)`, whose body is an arrow switch over the enum with no `case null` label — a null selector throws NPE, and NPE is not in that method's `catch (IllegalArgumentException | CertificateRequestException e)`. Result is a 500, not a clear 422.

The NPE predates this PR (format was never null-checked), but this PR is what documents the branch as handled, and a comment that's wrong about a hidden invariant is worse than no comment. It is reachable: the DTO field defaults to `PKCS10` but carries no `@NotNull`, so an explicit `"format": null` in the JSON body sets it to null.

**Fix:** either make the guard do what the comment claims — `throw new ValidationException(ValidationError.create("Certificate signing request format is required"))` — or drop the second sentence of the javadoc. The first is better: one line, and it closes a real 500.

src/main/java/com/otilm/core/certificate/request/ProtocolRequestAttributeValidator.java
```diff
@@ -81,9 +81,9 @@ private RequestAttributeValidationResult runKernel(List<BaseAttribute> definitio
             RaProfile raProfile, boolean strict) {
         try {
             ParsedRequestContent parsed = X509RequestContentParser.parse(request);
-            // Whitelist enforcement is tied to strict mode, so lenient mode does NOT run the whitelist check.
+            // The whitelist always runs; RequestAttributePolicy routes its findings to warnings when not strict.
             return CertificateRequestContentValidator
-                    .validate(definitions, parsed, new RequestAttributePolicy(strict, strict));
+                    .validate(definitions, parsed, new RequestAttributePolicy(strict, true));
```
vyskocilm, line 101: 🔵 **nit — after this line the codebase holds two contradictory definitions of "lenient", and the test suite certifies the obsolete one.**

I verified the scope note's claim: `CertificateRequestContentValidator.validate(CertificateRequest, List, boolean)` genuinely has no production caller. But it's still a public method on a live `@Component`, and it still builds `new RequestAttributePolicy(!lenient, !lenient)` — i.e. lenient still skips the whitelist. More awkwardly, `CertificateRequestContentValidatorTest` (around lines 569, 582, 596) asserts that obsolete semantic, so the green suite now reads as a decision that lenient skips the whitelist: the exact opposite of what this line decides. The next caller to reach for that seam inherits the pre-PR behaviour with test coverage vouching for it.

Same shape one file over: `RequestAttributePolicy.lenient()` still returns `(false, false)`, which now contradicts the only production construction and has no production caller — an obviously-named factory that will silently disable the whitelist for whoever reaches for it next.

**Fix:** either delete the unused overload (and the three tests, which duplicate coverage the policy-parameterised overload already has) plus the `lenient()` factory, or change both to `(…, true)` and update the three assertions. "Not touched rather than half-fixed" is a fair instinct, but leaving a fixture that certifies the opposite rule is worse than either option.

src/main/java/com/otilm/core/certificate/request/ProtocolRequestAttributeValidator.java
```diff
@@ -81,9 +81,9 @@ private RequestAttributeValidationResult runKernel(List<BaseAttribute> definitio
             RaProfile raProfile, boolean strict) {
         try {
             ParsedRequestContent parsed = X509RequestContentParser.parse(request);
```
vyskocilm, line 98: 🔵 **nit — lenient is not warning-only on this line, so the behaviour change isn't strict-only as the description says.**

`runKernel` converts a parse failure into `RequestAttributePolicyViolationException` regardless of `strict`. `/certificates/{uuid}/issue` (and the CMP/SCEP completion paths) reach this parser for the first time in this PR, so a CSR that `CertificateRequestUtils.createCertificateRequest` accepts but `X509RequestContentParser.parse` throws on — a malformed `otherName`, an `iPAddress` that isn't 4 or 16 octets, an unparseable RDN value — now fails a completion with 422 that previously succeeded, **with strict off**.

The description says only "under strict this is a behaviour change". Either gate the parse-failure rejection on `strict` (returning it as a warning otherwise, which matches the feature's intent), or widen that sentence in the description to cover it.

vyskocilm (commented):
A review of this branch — three passes: an architecture/convention pass, a correctness pass, and a security pass. Nine findings, one of which I'd treat as merge-blocking (the connector response body reaching a 500 REST body on the new `/certificates/{uuid}/issue` validation path). The rest are cheap: a fixed message, a "not checked" signal, two error-shaping alignments, and three doc/dead-code corrections.

**What I ran** at head `1cd1c897213739213302ff8c4b88973df20288d2`:

| Command | Result |
|---|---|
| architecture test package (`com.otilm.core.architecture.*Test`) | 102 tests, 0 failures — includes the authorization, transaction-boundary and context-signature guards |
| `spotless:check checkstyle:check` | 1956 files clean, 0 violations |
| `ProtocolRequestAttributeValidatorTest`, `CertificateRequestContentValidatorTest`, `RequestAttributeValidationResultTest` | 51/51 |
| `ClientOperationRequestAttributePropagationITest`, `AcmeFinalizeRequestAttributeValidationITest` | 4/4 (real PostgreSQL) |
| `ClientOperationServiceRegisterITest` | 106/106 |

**What I could not run**, so treat as unchecked rather than clean: the full suite (you report 6297 tests; I ran ~263 targeted), and coverage parity against the new-code gate formula. On that second one, a named risk found by inspection rather than measurement: `ScepServiceImpl:1112-1114` has no test — no test in the repo names "Unable to complete certificate registration", and `ScepRegistrationEnrolmentITest` never mentions `externalCsrValidationStrict` or `resolveIssueAttributeSet`. Same for the `CertificateException` arm of the widened catch at `CrmfIrCrMessageHandler:170` and the null-format guard at `ClientOperationServiceImpl:3112`. Roughly 5 uncovered new executable lines; probably over the gate, but unmeasured.

**Claims in the description I checked and can confirm.** The `(strict, strict)` → `(strict, true)` flip really is strict-neutral: findings route on the strict flag, and the newly-enabled whitelist pass is pure collection/string work over already-parsed content, so it cannot introduce a rejection. The `submitCertificateRequest` → `submitCertificateRequestInternal` split really is behaviour-identical: `submitAndShapeFailure` already self-invoked the public method, so the proxy was already bypassed and the annotation already didn't run on that path — which is exactly why the explicit `checkCreatePermissions()` probe is load-bearing, and it moved with the body. The class-level `@Transactional` is REQUIRED, so no transaction boundary moved. Validation on `/certificates/{uuid}/issue` really does sit behind every gate — authorization, profile-under-authority, secured entity load, ownership equality, state, challenge, then the create-permission probe — so it is not a policy oracle. The context-signature baseline of 61 is machine-verified unchanged. And the `CertificateRequestContentValidator` boolean overload really does have no production caller.

**Also clean:** no resource, action, permission, policy or auth migration is touched. The contract side (`requestAttributeWarnings`) is already on `interfaces` `main` and consumed, not invented locally — the ordering is right. `toLogData()` still returns only the uuid, so warnings never enter the audit record. No transaction or row lock spans the new connector call (`issueExistingCertificate` is `NOT_SUPPORTED`, and the challenge gate's lock commits before returning). The new SCEP compensation is correct — the exception can only arise before the enqueue, so nothing had been acknowledged. And the widened window between challenge and attach is safe: the attach re-reads under a pessimistic lock and re-asserts both the state and the null request-uuid.

**Retracted:** nothing from the mechanical pre-flight — it produced no candidates on this diff.

Two things worth a word in the release note rather than the code: running the whitelist under lenient means a `log.warn` per enrolment on any profile whose resolved set is narrower than real client CSRs (that's the point of the feature, but it's a log-volume change on hot protocol paths), and the new validation adds a second connector round-trip plus a second CSR parse per completion.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Fixed in 6e8c092. Rethrown causeless with the platform-authored message, cause logged at error — so the advice has nothing to append. Your read of the mechanism was exact: `ConnectorClientException`/`ConnectorServerException` are constructed straight from the body string, so the cause message *is* the upstream body. The regression test now asserts `getCause()` is null, which is the part it was missing.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: Fixed in 6e8c092 — fixed text plus `logger.debug`, exactly as the sibling method does.

src/main/java/com/otilm/core/certificate/request/ProtocolRequestAttributeValidator.java
klaraf755: Fixed in 6e8c092. `resolveDefinitions` now returns the reason it could not resolve, and a lenient failure returns one platform-authored warning saying validation could not be performed and why — so an empty list keeps meaning "checked, nothing found". Both reasons are covered by unit tests.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
klaraf755: Both parts done in 6e8c092. The exception gets its own arm ahead of the `ValidationException` one in both completion paths, shaped as the plain enrolment sites already shape it — SCEP `BAD_REQUEST` with the platform-authored message, CMP `badCertTemplate` — and SCEP discards the staged poll mapping, since the failure precedes the enqueue. Recorded in the PR body under "Behaviour changes beyond REST" and in a note on #1951.

src/main/java/com/otilm/core/service/cmp/message/handler/CrmfIrCrMessageHandler.java
klaraf755: Fixed in 6e8c092 — `catch (RequestAttributePolicyViolationException e)` before the generic arm, carrying `e.getMessage()` as `badCertTemplate`, matching the ir/cr path above. I left the generic arm untouched; its "detail logged" comment is pre-existing and, as you spotted, not true.

src/main/java/com/otilm/core/service/scep/impl/ScepServiceImpl.java
klaraf755: Fixed in 6e8c092 — `FailInfo.BAD_REQUEST` passed explicitly, cause chained, and the comment now says BAD_REQUEST is the closest the protocol allows rather than promising a server failure it cannot express.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: I don't think this one holds — `addCertificateRequestToExisting` null-checks the format before any parsing:

```java
if (issueRequest.getFormat() == null) {
    throw new CertificateRequestException("A certificate signing request format (PKCS10 or CRMF) is required");
}
```

That's ahead of the `createCertificateRequest` call, so a null format never reaches the arrow switch. It surfaces as `CertificateRequestException` → `ValidationException` → 422, and `issueRegisteredCertificateWithNullCsrFormatIsRejected` pins exactly that (it passes today; it also caught me when I first placed the validation before the guard). So the javadoc's second sentence is accurate and I've left both as they are — happy to reopen if you're looking at a different entry point.

src/main/java/com/otilm/core/certificate/request/ProtocolRequestAttributeValidator.java
klaraf755: Fixed in 6e8c092 by deleting rather than realigning: the overload, `RequestAttributePolicy.lenient()`, the `InstanceOverload` tests, and the factory-semantics test. The two tests that genuinely wanted the whitelist off — one is named for it — now construct `new RequestAttributePolicy(false, false)` directly, so their intent survives without a misleadingly-named factory. Agreed that leaving a fixture certifying the opposite rule was the worst option.

src/main/java/com/otilm/core/certificate/request/ProtocolRequestAttributeValidator.java
klaraf755: Confirmed, and recorded rather than changed. Gating the parse failure on `strict` would contradict `rejectsMalformedUploadedCsr_asValidationError_notServerFault`, which pins 422-under-lenient on the existing submit path — so gating it here would make the two paths disagree instead of agree. Widened the description instead ("It is not strict-only", with the `otherName`/`iPAddress` examples); aligning parse-failure handling across both paths is worth its own change.

vyskocilm (approved):

PR #2185, opened 2026-09-01
Populate the caller's resource actions in the user profile

### LINKED ISSUE #2183: Populate the caller's resource actions in the user profile
### Description

The Core half of OmniTrustILM/interfaces#920, which adds `allowedActions` to `UserProfilePermissionsDto`. That issue defines the contract; this one fills it in.

`AuthServiceImpl.getAuthProfile` currently reports `allowedListings` only, so a client can tell which resources it may list and nothing else. Narrow actions such as `UPDATE_BRANDING`, `GET_SECRET_CONTENT` and `UPDATE_SOURCE_VAULT_PROFILE` guard endpoints that `UPDATE` does not reach, leaving a client no way to know whether a control it is about to offer will be refused.

The authorization data needed is already resolved for the request — `AuthHelper.getUserProfile()` carries the per-resource permissions, and `ContextRefreshListener` carries the action catalogue — so nothing new is fetched.

Two details that decide correctness:

- Actions must be filtered out of the resource catalogue rather than mapped out of the grant. The authorization service can hold an action code this build no longer declares, and `ResourceAction.findByCode` throws on one, which would fail the whole profile request over a permission the caller does not even need.
- `NONE` and `ANY` are `NOT_GRANTABLE` sentinels the auth service rejects as unknown actions, so reporting them would name a grant that cannot exist.

Object-scoped grants stay out, per the contract on the field: an action held on one object says nothing about the next, so this is not the union `getAllowedResourceListings` computes for `LIST`.

First consumer is the Appearance tab in OmniTrustILM/fe-administrator#2041, which withholds itself from a viewer lacking `UPDATE_BRANDING` instead of offering a save Core refuses.

### Definition of Done

- [ ] `getAuthProfile` reports the caller's granted actions per resource alongside `allowedListings`.
- [ ] A role holding `SETTINGS` + `UPDATE` but not `UPDATE_BRANDING` does not have `UPDATE_BRANDING` reported.
- [ ] A role with `allowAllResources` has every catalogue action reported, `UPDATE_BRANDING` included.
- [ ] An action code the auth service holds but this build does not declare does not fail the profile request.
- [ ] `NOT_GRANTABLE` sentinels are never reported.
- [ ] Object-scoped grants are not reported as resource-level actions.
- [ ] Sonar quality gate passes and new code meets at least 80% coverage.

Jakub-Bartal-01:
Fills in the contract added by OmniTrustILM/interfaces#936, which adds `allowedActions` to `UserProfilePermissionsDto`. **Depends on that PR merging first** — `@AllArgsConstructor` means the constructor signature changes.

`getAuthProfile` reported `allowedListings` only, so a client could tell which resources it may list and nothing else. `UPDATE_BRANDING`, `GET_SECRET_CONTENT` and `UPDATE_SOURCE_VAULT_PROFILE` guard endpoints that `UPDATE` does not reach, leaving a client no way to know whether a control it was about to offer would be refused.

Nothing new is fetched. The per-resource permissions already come in on `AuthHelper.getUserProfile()`, and the action catalogue is already held by `ContextRefreshListener` for the auth-service sync.

## Two details that decide correctness

**Actions are filtered out of the catalogue, not mapped out of the grant.** The authorization service can hold an action code this build no longer declares — after a rename, or a downgrade — and `ResourceAction.findByCode` throws a `ValidationException` on one. Mapping the grant directly would fail the entire profile request, and therefore every page load, over a permission the caller does not even need. Intersecting with the catalogue means only codes this build declares can reach `findByCode`.

**`NONE` and `ANY` are never reported.** Both are `AccessType.NOT_GRANTABLE`; the auth service rejects them as unknown actions, so reporting them would name a grant that cannot exist. `registerResourceActions` guards `Resource.NONE` for the parent resource but not the action, so the sentinel can reach the catalogue.

## Semantics

Resource-level grants only, per the contract on the field. Object-scoped grants stay out, so this is not the union `getAllowedResourceListings` computes for `LIST` — an action held on one object says nothing about the next. Resources with no granted action are omitted rather than reported empty, and the output is ordered by resource code so the response is stable.

## Tests

Three added to `AuthServiceITest`, using the fixtures already there:

- The `localhost` role holds `settings` with `detail`, `list` and `update`. `UPDATE_BRANDING` is asserted absent, which is the case the field was added for.
- The changed profile allows `members` on one group object only; that grant is asserted absent from `allowedActions` while `GROUP` stays in `allowedListings`, pinning the deliberate difference between the two fields.
- The `allowAllResources` role is asserted to hold `UPDATE_BRANDING`, and no `NOT_GRANTABLE` action anywhere in the response.

Closes #2183

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/AuthServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/AuthServiceImpl.java
@@ -4,6 +4,7 @@ import com.otilm.api.exception.NotFoundException;
 import com.otilm.api.model.client.auth.UpdateUserRequestDto;
 import com.otilm.api.model.core.auth.AuthResourceDto;
 import com.otilm.api.model.core.auth.Resource;
+import com.otilm.api.model.core.auth.ResourceActionsDto;
 import com.otilm.api.model.core.auth.ResourcePermissionsDto;
 import com.otilm.api.model.core.auth.UserDetailDto;
 import com.otilm.api.model.core.auth.UserProfileDetailDto;
@@ -11,6 +12,7 @@ import com.otilm.api.model.core.auth.UserProfileDto;
 import com.otilm.api.model.core.auth.UserProfilePermissionsDto;
 import com.otilm.core.auth.ContextRefreshListener;
 import com.otilm.core.model.auth.ResourceAction;
+import com.otilm.core.model.auth.ResourceSyncRequestDto;
 import com.otilm.core.security.authn.client.ResourceApiClient;
 import com.otilm.core.security.authn.client.UserManagementApiClient;
 import com.otilm.core.security.authz.AnyPrincipalEndpoint;
@@ -68,8 +70,8 @@ public class AuthServiceImpl implements AuthExternalService {
         UserDetailDto userDetailDto = userManagementApiClient.getUserDetail(userProfileDto.getUser().getUuid());
 
         // load listing permissions
-        return new UserProfileDetailDto(userDetailDto,
-                new UserProfilePermissionsDto(getAllowedResourceListings(userProfileDto)));
+        return new UserProfileDetailDto(userDetailDto, new UserProfilePermissionsDto(
+                getAllowedResourceListings(userProfileDto), getAllowedResourceActions(userProfileDto)));
     }
 
     @Override
@@ -143,6 +145,64 @@ public class AuthServiceImpl implements AuthExternalService {
             }
         }
         return withDefaultListings(allowedListings);
+    }
+
+    /**
+     * The actions the caller holds on each resource as a whole, so a client can gate a control on the action that
+     * actually guards the endpoint instead of inferring it from listing access.
+     * <p>
+     * Object-scoped grants are deliberately left out - see {@link UserProfilePermissionsDto#getAllowedActions()} - so
+     * this is not the union {@link #getAllowedResourceListings} computes for {@code LIST}, and the two fields are not
+     * derivable from each other.
+     */
+    private List<ResourceActionsDto> getAllowedResourceActions(UserProfileDto userProfileDto) {
+        boolean allowAllResources = Boolean.TRUE.equals(userProfileDto.getPermissions().getAllowAllResources());
+        Map<Resource, ResourcePermissionsDto> mappedUserPermissions = allowAllResources
+                ? Map.of()
+                : userProfileDto
+                        .getPermissions()
+                        .getResources()
+                        .stream()
+                        .collect(Collectors
+                                .toMap(resource -> Resource.findByCode(resource.getName()), resource -> resource));
+
+        List<ResourceActionsDto> allowedActions = new ArrayList<>();
+        for (ResourceSyncRequestDto syncResource : contextRefreshListener.getResources()) {
+            Resource resource = Resource.findByCode(syncResource.getName().getCode());
+            ResourcePermissionsDto resourcePermissions = mappedUserPermissions.get(resource);
+            // Filtered out of the catalogue rather than mapped out of the grant: the authorization service may hold an
+            // action code this build no longer declares, and findByCode throws on one, which would fail the whole
+            // profile request over a permission the caller does not even need.
+            List<ResourceAction> actions = syncResource
+                    .getActions()
+                    .stream()
+                    .filter(actionCode -> isActionGranted(actionCode, allowAllResources, resourcePermissions))
+                    .map(ResourceAction::findByCode)
+                    // NONE and ANY are sentinels the auth service rejects as unknown actions, so reporting them would
+                    // name a grant that cannot exist.
+                    .filter(action -> action.getAccessType() != ResourceAction.AccessType.NOT_GRANTABLE)
+                    .sorted(Comparator.comparing(ResourceAction::getCode))
+                    .toList();
+
+            if (!actions.isEmpty()) {
+                allowedActions.add(new ResourceActionsDto(resource, actions));
+            }
+        }
+        return allowedActions.stream().sorted(Comparator.comparing(entry -> entry.getResource().getCode())).toList();
+    }
+
+    private static boolean isActionGranted(String actionCode, boolean allowAllResources,
+            ResourcePermissionsDto resourcePermissions) {
+        if (allowAllResources) {
+            return true;
+        }
+        if (resourcePermissions == null) {
+            return false;
+        }
+        if (Boolean.TRUE.equals(resourcePermissions.getAllowAllActions())) {
+            return true;
+        }
+        return resourcePermissions.getActions() != null && resourcePermissions.getActions().contains(actionCode);
     }
 
     private List<Resource> withDefaultListings(List<Resource> listings) {
--- a/src/main/java/com/otilm/core/service/impl/AuthServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/AuthServiceImpl.java
@@ -101,6 +101,15 @@ public class AuthServiceImpl implements AuthExternalService {
                         certificateFingerprint);
     }
 
+    /** Shared by both permission-derived fields, so a change to how a grant is looked up reaches each of them. */
+    private static Map<Resource, ResourcePermissionsDto> mapPermissionsByResource(UserProfileDto userProfileDto) {
+        return userProfileDto
+                .getPermissions()
+                .getResources()
+                .stream()
+                .collect(Collectors.toMap(resource -> Resource.findByCode(resource.getName()), resource -> resource));
+    }
+
     private List<Resource> getAllowedResourceListings(UserProfileDto userProfileDto) {
         List<Resource> allowedListings;
         List<Resource> allListings = contextRefreshListener
@@ -115,11 +124,7 @@ public class AuthServiceImpl implements AuthExternalService {
             return withDefaultListings(allListings);
         }
 
-        Map<Resource, ResourcePermissionsDto> mappedUserPermissions = userProfileDto
-                .getPermissions()
-                .getResources()
-                .stream()
-                .collect(Collectors.toMap(resource -> Resource.findByCode(resource.getName()), resource -> resource));
+        Map<Resource, ResourcePermissionsDto> mappedUserPermissions = mapPermissionsByResource(userProfileDto);
         ResourcePermissionsDto groupPermissions = mappedUserPermissions.get(Resource.GROUP);
         boolean hasGroupMembersPermissions = groupPermissions != null
                 && (groupPermissions.getAllowAllActions() || groupPermissions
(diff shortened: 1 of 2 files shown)


src/main/java/com/otilm/core/service/impl/AuthServiceImpl.java
```diff
@@ -145,6 +147,64 @@ private List<Resource> getAllowedResourceListings(UserProfileDto userProfileDto)
         return withDefaultListings(allowedListings);
     }
 
+    /**
```
klaraf755: Third copy of this rationale - the field Javadoc and the `@Schema` description in OmniTrustILM/interfaces#936 both state it. The `{@link UserProfilePermissionsDto#getAllowedActions()}` link is the half that earns its place; suggest keeping the link and dropping the restated prose.

src/main/java/com/otilm/core/service/impl/AuthServiceImpl.java
```diff
@@ -145,6 +147,64 @@ private List<Resource> getAllowedResourceListings(UserProfileDto userProfileDto)
...
+     * The actions the caller holds on each resource as a whole, so a client can gate a control on the action that
+     * actually guards the endpoint instead of inferring it from listing access.
+     * <p>
+     * Object-scoped grants are deliberately left out - see {@link UserProfilePermissionsDto#getAllowedActions()} - so
+     * this is not the union {@link #getAllowedResourceListings} computes for {@code LIST}, and the two fields are not
+     * derivable from each other.
+     */
+    private List<ResourceActionsDto> getAllowedResourceActions(UserProfileDto userProfileDto) {
+        boolean allowAllResources = Boolean.TRUE.equals(userProfileDto.getPermissions().getAllowAllResources());
+        Map<Resource, ResourcePermissionsDto> mappedUserPermissions = allowAllResources
```
klaraf755, line 158: The same map is built at line 118, so `getAuthProfile()` computes it twice and a fix to one silently misses the other - worth one private helper. Note the asymmetry while you're in there: this method guards unknown *action* codes carefully, then maps *resource* codes straight through the same throwing `findByCode`.

src/main/java/com/otilm/core/service/impl/AuthServiceImpl.java
```diff
@@ -145,6 +147,64 @@ private List<Resource> getAllowedResourceListings(UserProfileDto userProfileDto)
...
+            // action code this build no longer declares, and findByCode throws on one, which would fail the whole
+            // profile request over a permission the caller does not even need.
+            List<ResourceAction> actions = syncResource
+                    .getActions()
+                    .stream()
+                    .filter(actionCode -> isActionGranted(actionCode, allowAllResources, resourcePermissions))
+                    .map(ResourceAction::findByCode)
+                    // NONE and ANY are sentinels the auth service rejects as unknown actions, so reporting them would
+                    // name a grant that cannot exist.
+                    .filter(action -> action.getAccessType() != ResourceAction.AccessType.NOT_GRANTABLE)
```
klaraf755: `ResourceAction` already hosts this kind of classification via `isGrantableToReadOnlyRole()`, and `ReadOnlyRolePermissions` states the intent - classifying on the enum should be the only step. An `isGrantable()` sibling would keep the rule in one place instead of here and in the test.

The reason needs a correction too: only `ANY` is rejected, skipped at sync via `AnyActionName` in `ResourceService.SyncResourcesAsync`. `NONE` isn't skipped and would become a real, grantable action row. It can't reach the payload today - all 706 annotated endpoints pass an explicit action, only the `action()` default is `NONE` - so this guards a live footgun rather than a rejected value.

src/test/java/com/otilm/core/integration/service/AuthServiceITest.java
```diff
@@ -132,6 +133,84 @@ void testAuthProfileAllowAllResourcesDefaultListings() {
...
+    @Test
+    void testAuthProfileAllowedActionsExcludeObjectScopedGrants() {
+        injectLocalhostUserProfileChangedToContext();
+
+        UserProfileDetailDto userProfileDto = authService.getAuthProfile();
+
+        Assertions
+                .assertEquals(List.of(ResourceAction.CREATE, ResourceAction.DETAIL, ResourceAction.LIST),
+                        allowedActionsFor(userProfileDto, Resource.GROUP));
+        // The same grant does put GROUP in allowedListings, which is the union the two fields deliberately differ on.
```
klaraf755: This doesn't pin the difference the comment claims. The fixture grants GROUP a resource-level `list`, so it reaches `allowedListings` through that grant - delete the whole `objects` block and the assertion still passes. `CERTIFICATE` is the real contrast: always listed via `hasOwner()`, but `[CREATE, DETAIL]` with no LIST in `allowedActions`.

src/test/java/com/otilm/core/integration/service/AuthServiceITest.java
```diff
@@ -132,6 +133,84 @@ void testAuthProfileAllowAllResourcesDefaultListings() {
                         "DASHBOARD must appear exactly once");
     }
 
+    /**
```
klaraf755: Javadoc above test methods - the names carry it. Same at 162. The block at line 78 clears the bar because it explains the annotation scan; these two restate their assertions.

src/test/java/com/otilm/core/integration/service/AuthServiceITest.java
```diff
@@ -132,6 +133,84 @@ void testAuthProfileAllowAllResourcesDefaultListings() {
...
+        Assertions
+                .assertFalse(
+                        allowedActionsFor(userProfileDto, Resource.SETTINGS).contains(ResourceAction.UPDATE_BRANDING),
+                        "SETTINGS granted update must not imply UPDATE_BRANDING");
+        Assertions
+                .assertEquals(List.of(ResourceAction.CREATE, ResourceAction.DETAIL),
+                        allowedActionsFor(userProfileDto, Resource.CERTIFICATE));
+        Assertions
+                .assertTrue(allowedActionsFor(userProfileDto, Resource.CONNECTOR).isEmpty(),
+                        "a resource the role grants nothing on must be omitted, not reported empty");
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed. The assertion now runs over the reported entries — `noneMatch(entry -> entry.getResource() == Resource.CONNECTOR)` — rather than through the lookup helper, whose `orElseGet(List::of)` made absence and a present-but-empty action list indistinguishable.

Checked that it bites: replacing `if (!actions.isEmpty())` with `if (true)` in the service, so every catalogue resource is emitted, fails this assertion.

src/test/java/com/otilm/core/integration/service/AuthServiceITest.java
```diff
@@ -132,6 +133,84 @@ void testAuthProfileAllowAllResourcesDefaultListings() {
...
+                .assertEquals(List.of(ResourceAction.CREATE, ResourceAction.DETAIL, ResourceAction.LIST),
+                        allowedActionsFor(userProfileDto, Resource.GROUP));
+        // The same grant does put GROUP in allowedListings, which is the union the two fields deliberately differ on.
+        Assertions
+                .assertTrue(userProfileDto.getPermissions().getAllowedListings().contains(Resource.GROUP),
+                        "allowedListings keeps its own semantics");
+    }
+
+    @Test
+    void testAuthProfileAllowedActionsAllowAllResources() {
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Added `testAuthProfileAllowedActionsIgnoreUndeclaredActionCodes`: a profile granting `certificates` the actions `["create", "detail", "actionThisBuildNoLongerDeclares"]`, asserting the profile request still returns and reports `[CREATE, DETAIL]`.

Checked that it bites: switching the pipeline to read the grant's action list instead of the catalogue makes it error with `Unknown Resource Action Name actionThisBuildNoLongerDeclares`.

src/test/java/com/otilm/core/integration/service/AuthServiceITest.java
```diff
@@ -132,6 +133,84 @@ void testAuthProfileAllowAllResourcesDefaultListings() {
...
+        injectAllowAllResourcesUserProfileToContext();
+
+        UserProfileDetailDto userProfileDto = authService.getAuthProfile();
+
+        Assertions
+                .assertTrue(
+                        allowedActionsFor(userProfileDto, Resource.SETTINGS).contains(ResourceAction.UPDATE_BRANDING),
+                        "a role holding every resource must hold UPDATE_BRANDING");
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed. The test now compares the complete set of reported resource/action pairs against every pair in `contextRefreshListener.getResources()`, minus the two sentinel codes; the UPDATE_BRANDING probe stays as the named case.

The comparison is not vacuous: the catalogue carries 35 `ANY` entries (35 annotations pass `ResourceAction.ANY`), so removing the sentinel filter from the service fails this assertion on `certificates:ANY`, `raProfiles:ANY` and the rest.

src/main/java/com/otilm/core/service/impl/AuthServiceImpl.java
Jakub-Bartal-01: Trimmed to the link: `/** See {@link UserProfilePermissionsDto#getAllowedActions()} for the contract this computes. */`.

src/main/java/com/otilm/core/service/impl/AuthServiceImpl.java
Jakub-Bartal-01: Extracted `mapPermissionsByResource(UserProfileDto)`; both `getAllowedResourceListings` and `getAllowedResourceActions` call it, each keeping its own `allowAllResources` short-circuit, so behaviour is unchanged.

On the asymmetry: confirmed, and it is not merely uneven — a resource code the authorization service still holds that this build no longer declares makes `findByCode` throw and fails the whole profile request, for both fields. It is pre-existing on `main`: `getAllowedResourceListings` has always mapped resource codes that way, and the extraction has not moved it. Guarding it would change `allowedListings` too, which is behaviour this PR is not otherwise touching, so I have left it and would rather do it as its own change. Say the word and I will raise it.

src/main/java/com/otilm/core/service/impl/AuthServiceImpl.java
Jakub-Bartal-01: The correction is right and the comment now states it: only `ANY` is skipped at sync (`AnyActionName` in `ResourceService`), so no grant on it can exist; `NONE` is not skipped and would be stored as a real action, and stays out of the catalogue only because every annotated endpoint passes an explicit action rather than leaving the `action()` default in place. Verified against the tree — 35 annotations pass `ResourceAction.ANY`, none rely on the default — and against the running catalogue: with the sentinel filter removed, `ANY` pairs appear in the reported set and no `NONE` pair does.

Pushing back on the classification itself. `isGrantableToReadOnlyRole()` is `accessType == READ`, so reusing it here would drop every `WRITE` and `SENSITIVE_READ` action — `UPDATE_BRANDING` included, which is the action this field exists for. The two rules select different sets and cannot share a predicate.

An `isGrantable()` sibling is the right home for the `!= NOT_GRANTABLE` half, but `ResourceAction` ships from interfaces, so it is a change on #936 and this PR could not compile against it until that publishes. The duplication you flagged is gone in the meantime: the test no longer restates the `AccessType` comparison, it names the two sentinel codes it excludes from the expected catalogue set, so the classification is stated once, in the service. If you would rather have the enum method now, I will add it to #936 and switch this filter to it.

src/test/java/com/otilm/core/integration/service/AuthServiceITest.java
Jakub-Bartal-01: Correct — the fixture's resource-level `list` on groups puts GROUP in `allowedListings` on its own, so deleting the `objects` block leaves that assertion passing and it pinned nothing.

Replaced with the CERTIFICATE contrast you named: present in `allowedListings` through `hasOwner()`, reported as `[CREATE, DETAIL]` with no LIST in `allowedActions`. The GROUP equality assertion stays, since that is the one that pins the single-object MEMBERS grant being excluded.

src/test/java/com/otilm/core/integration/service/AuthServiceITest.java
Jakub-Bartal-01: Both removed. Where the fixture detail behind an assertion is not in the test name, it is a one-line comment on the assertion instead — the single-object MEMBERS grant, and why CERTIFICATE is the contrast between the two fields.

The new undeclared-action-code test does carry a Javadoc, on the same grounds as the block at line 78: its name says what it asserts, not why the case exists. Drop it if you read that one differently.

klaraf755 (approved):

PR #2173, opened 2026-08-31
Repair the primitive identity helpers for core#2165 (1 of 2, must land before core#2170)

vyskocilm:
4918
mmary

Part 1 of 2 in the core#2165 adjudication series, and the first link in a stack
that has to land in order: `main` <- **this PR** <- core#2170. core#2072 has
landed on `main`, so this PR carries the primitive-layer core#2165 repairs that
can merge independently of the generated identity-table work. core#2170's
264-vector fixture already encodes the behaviour this PR settles, so it must not
merge first.

This lands the core#2165 repairs that fit in the primitive identity layer now
present on `main`:

- make timestamp normalization strict, preserve non-timestamp fallbacks, and
  fold colonized RFC 3339 offsets
- sanitize occurrence locations in the safe order and cap by code point; keep
  duplicate triples and refuse non-integral numeric positions
- constrain cipher-suite identifier parsing and render each token to even-width
  hex up to `0xFFFF`
- store the contracted redaction envelope `{"redacted": true, "length": n}` with
  no digest
- prevent certificate digest sub-delimiter forgery and avoid keying on
  self-contradictory digest algorithms
- apply the reference whitespace set to lookup separators

## Measured cost

Against the 2026-08-31 corpus (180 documents, 4 815 assets), two of these
changes move a row and the rest are free:

- **duplicate occurrence triples** -- 12 occurrence-bearing assets across 4
  documents repeat an identical `location#line#offset`
  (`Chat2DB_cdxgen` collapses 79 -> 73 and 38 -> 10 under the old `TreeSet`)
- **the fourth cipher-suite encoding** -- `["0x1301"]`, `["0x1302"]`,
  `["0x1303"]`, one row each in CycloneDX's own
  `valid-cryptography-full-1.7` fixture, now merging with the 17 per-byte rows
  spelling the same suites; the 2 133 -> 2 134 identity split does not happen

Zero-row today, so free to settle now and not after first ingest: 0 of 490
distinct validity strings are calendar-invalid, 0 of 390 certificate hash
entries repeat an algorithm, 0 `alg`/`content` values carry a colon, 0
fractional numeric positions, and 0 occurrence locations carry userinfo, `#`,
CR or LF or exceed 194 characters.

## Review findings repaired in this branch

The second commit answers a review against the ai-toolkit review contract and
the architecture gate. Two findings were behavioural:

- the fraction pattern was anchored to the extended time-of-day spelling only,
[...]

PR #2188, opened 2026-09-01
Serve the cryptographic asset detail, statistics and sync visibility

### LINKED ISSUE #2145: Implement cryptographic asset detail, statistics and sync visibility endpoints

karolhrdina:
Closes #2145. Second half of the read surface on top of #2167's plumbing: fills the two `NotSupportedException` stubs and pins the CBOM sync-state fields to the wire.

## What

- **Detail** `GET /v1/cryptoAssets/{uuid}`: elected representative payload served verbatim beside every source's original (never merged), verdict provenance block, normalized fields, per-source capped evidence with true pre-cap `occurrenceCount`, OIDs with the `refuted` flag, sources fetch-joined to their CBOM headers in first-seen order.
- **Statistics** `GET /v1/statistics/cryptoAssets`: seven security-scoped aggregates on virtual threads (`DelegatingSecurityContextExecutorService`), reshaped by a pure package-private calculator — dense enum maps, null-bucket folds, top-10 families with `distinctAlgorithmFamilyCount`/`unassignedAssetCount` overflow badges, sync-completeness block. Gated by CRYPTO_ASSET LIST (deliberately shares the list permission). Aggregation failures fail the request rather than serve zeros that no longer reconcile.
- **Sync visibility**: `assetSyncState`/`assetSyncedAt` were already mapped on the CBOM DTOs; new tests pin them on list and detail.

## Acceptance criteria

- Detail complete; source links resolve (`cbomUuid` + `serialNumber`/`version`): `CryptographicAssetDetailITest`.
- Per-source originals beside the elected payload, disagreement covered: `servesElectedPayloadBesideDisagreeingPerSourceOriginals`.
- Refuted OID renders flagged and is never served as the display name: `flagsARefutedOidAndSuppressesItAsName`.
- Statistics reconcile exactly with list totals under identical permission filtering (asserted with absolute counts under an OPA object restriction) and the completeness block is correct under partial sync: `CryptographicAssetStatisticsITest`.
- No response carries the identity key: ArchUnit fence + wire-level regex and recomputed-literal assertions on list, detail, statistics and the 404 body.
- 403 without permission on every operation, server-side: `CryptographicAssetAuthorizationHttpITest`.
- Coverage on new code 0.9095 by the Sonar line+branch formula (gate 0.80).

## Map-typed payload hand-check (AC, recorded)

The lexical exposure fence cannot see a key travelling as runtime data, so the four map channels were traced by hand to their write paths:

1. `electedPayload` ← `merged_crypto_properties` ← `recomputeMergeFromSources` elects one source's stored payload verbatim → reduces to channel 2.
2. `sources[].payload` ← `original_crypto_properties` ← `CryptoAssetSourceWriter.upsertSource`, producer-derived and redaction-enveloped; no production caller exists yet (sync lands with core#2073), today fixture-only.
3. `verdict.evaluatedFields` ← `applyPqcVerdict` — written by core#2151, not here; fixture-only today. The write-side allowlist raised on core#2151 is the fix; this surface serves the column verbatim and is the backstop.
4. `CbomDetailDto.content` — pre-existing external-fetch channel, untouched by this diff.

## Coordination notes for reviewers

- PR #2173 (core#2165 items 3 and 12: credential-strip-before-cap, redaction envelope) is still open. This endpoint serves stored values, so rows ingested before those pipeline fixes land keep their pre-fix shapes until re-ingest — flagged on the issue as a pre-ship coordination point.
- Contract friction, follow-up filed: the detail DTO marks `verdict` REQUIRED, but a never-evaluated asset serves no fabricated provenance block (row-level `pqcVerdict` = `unknown`); OmniTrustILM/interfaces#938 flips it to NOT_REQUIRED.

## Verification

All four CI profile groups + full `mvn verify`: 6319 tests, 0 failures (2 pre-existing environmental `ProxyPropertiesTest` DNS errors, unrelated). Spotless/Checkstyle clean. Adversarial whole-branch review ran before this PR; its one finding (a reconciliation test that could pass vacuously) is fixed in the diff.

### DIFF
--- a/src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetSourceRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetSourceRepository.java
@@ -27,6 +27,19 @@ public interface CryptoAssetSourceRepository extends SecurityFilterRepository<Cr
     boolean existsByCbomUuid(UUID cbomUuid);
 
     /**
+     * The detail endpoint's source list: every CBOM's contribution with its document header, oldest contribution first,
+     * uuid as the deterministic tiebreak. The fetch join loads the read-only {@link CryptoAssetSource#getCbom()}
+     * association in the same query, so callers can read header fields without a session.
+     */
+    @Query("""
+            SELECT s FROM CryptoAssetSource s
+            JOIN FETCH s.cbom
+            WHERE s.assetUuid = :assetUuid
+            ORDER BY s.firstSeenAt ASC, s.uuid ASC
+            """)
+    List<CryptoAssetSource> findWithCbomByAssetUuid(@Param("assetUuid") UUID assetUuid);
+
+    /**
      * The assets a CBOM contributes to -- the work list for detaching that CBOM before its row is deleted, since the
      * foreign key is {@code RESTRICT}.
      */
--- a/src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
@@ -39,6 +39,7 @@ import com.otilm.core.security.authz.ExternalAuthorization;
 import com.otilm.core.security.authz.ObjectFilterAspect;
 import com.otilm.core.security.authz.SecuredUUID;
 import com.otilm.core.security.authz.SecurityFilter;
+import com.otilm.core.security.authz.SecurityResourceFilter;
 import com.otilm.core.service.CryptographicAssetExternalService;
 import com.otilm.core.service.ResourceExtensionService;
 import com.otilm.core.util.FilterPredicatesBuilder;
@@ -52,9 +53,11 @@ import jakarta.persistence.criteria.Root;
 import jakarta.persistence.criteria.Subquery;
 import java.time.OffsetDateTime;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
@@ -81,6 +84,11 @@ public class CryptographicAssetServiceImpl implements CryptographicAssetExternal
 
     /** Top-N cap on the served algorithm-family distribution; the badge pair carries the overflow. */
     private static final int TOP_ALGORITHM_FAMILIES = 10;
+
+    // Per-source evidence is already bounded at 50 occurrences / 64KB by OccurrenceEvidenceCapper; the asset-level
+    // fan-out -- one row per contributing CBOM -- is not. 100 x ~70KB bounds the response while sourceCbomCount
+    // keeps carrying the true total, the same served-vs-true pattern occurrenceCount already uses.
+    private static final int MAX_SERVED_SOURCES = 100;
 
     // The OpenAPI schema documents 1000 as the maximum items per page; this is where it is actually enforced.
     private static final int MAX_ITEMS_PER_PAGE = 1000;
@@ -149,7 +157,40 @@ public class CryptographicAssetServiceImpl implements CryptographicAssetExternal
                 .findByUuid(uuid)
                 .orElseThrow(() -> new NotFoundException(CryptoAsset.class, uuid));
         List<CryptoAssetSource> sources = cryptoAssetSourceRepository.findWithCbomByAssetUuid(asset.getUuid());
-        return toDetailDto(asset, sources);
+        return toDetailDto(asset, sources, visibleCbomUuids(asset.getUuid()));
+    }
+
+    /**
+     * The CBOM documents whose content the caller may read, scoped by the caller's own {@code cboms:list} object access
+     * -- serial numbers and payloads belong to the CBOM resource, the same rule
+     * {@link #cbomSerialNumbersScopedToCaller} applies to the source-CBOM value list. Narrowed to the documents that
+     * actually contribute to this asset, mirroring {@link #hasContributedAssets}, so the EXISTS subquery stays cheap.
+     */
+    private Set<UUID> visibleCbomUuids(UUID assetUuid) {
+        SecurityFilter cbomFilter = SecurityFilter.create();
+        objectFilterAspect.populateSecurityFilter(Resource.CBOM, ResourceAction.LIST, null, null, cbomFilter);
+        // Unpaged: Pageable.unpaged() throws from window()'s getOffset()/getPageSize(), unsupported on that
+        // sentinel. A null Pageable and null order is this codebase's own idiom for "every matching uuid, no
+        // paging, no particular order" -- see CryptographicKeyServiceImpl#filterKeyItemsBySecurityFilter and
+        // CertificateServiceImpl#bulkDeleteCertificateBatch.
+        return new HashSet<>(
+                cbomRepository.findUuidsUsingSecurityFilter(cbomFilter, contributesToAsset(assetUuid), null, null));
+    }
+
+    /**
+     * {@link #hasContributedAssets}, narrowed to one asset: the EXISTS subquery a per-detail visibility scope needs.
+     */
+    private static TriFunction<Root<Cbom>, CriteriaBuilder, CriteriaQuery<?>, Predicate> contributesToAsset(
+            UUID assetUuid) {
+        return (root, cb, query) -> {
+            Subquery<Integer> contributed = query.subquery(Integer.class);
+            Root<CryptoAssetSource> source = contributed.from(CryptoAssetSource.class);
+            contributed
+                    .select(cb.literal(1))
+                    .where(cb.equal(source.get(CryptoAssetSource_.cbomUuid), root.get(UniquelyIdentified_.uuid)),
+                            cb.equal(source.get(CryptoAssetSource_.assetUuid), assetUuid));
+            return cb.exists(contributed);
+        };
     }
 
     /**
@@ -232,6 +273,13 @@ public class CryptographicAssetServiceImpl implements CryptographicAssetExternal
             CryptographicAssetStatisticsDto dto = CryptographicAssetStatisticsCalculator
                     .assemble(totalAssets.get(), byType.get(), byVerdict.get(), byFamily.get(), TOP_ALGORITHM_FAMILIES,
                             sourceCbomCount.get(), bySyncState.get(), lastCompletedSyncAt.get());
+            if (deniesEverything(cbomFilter)) {
+                // A permission-shaped zero reads as "nothing has ever synced"; omitting the document-derived block
+                // instead states plainly that access, not the estate, is why nothing is served. Asset-side
+                // statistics above are untouched -- they are scoped by the independent CRYPTO_ASSET gate.
+                dto.setSourceCbomCount(null);
+                dto.setSyncCompleteness(null);
+            }
             log.debug("Cryptographic asset statistics calculated in {} ms", (System.nanoTime() - start) / 1_000_000L);
             return dto;
         } catch (InterruptedException e) {
@@ -242,6 +290,19 @@ public class CryptographicAssetServiceImpl implements CryptographicAssetExternal
             // request rather than silently serve zeros that no longer reconcile with the list.
             throw new IllegalStateException("Cryptographic asset statistics aggregation failed", e.getCause());
         }
+    }
+
+    /**
+     * Whether the populated filter denies every object of its resource outright: {@code areOnlySpecificObjectsAllowed}
+     * true with an empty allow-list -- the shape {@code ObjectFilterAspect#getResourceFilter} builds from an OPA "not
+     * allowed for the whole group of objects" vote that also names no individually allowed uuid, which is what
+     * {@code BaseSpringBootTest.denyObjectAccess} stubs. A partial restriction (some uuids allowed, or the whole group
+     * allowed with some forbidden) does not match this and keeps serving scoped counts as before.
+     */
+    private static boolean deniesEverything(SecurityFilter filter) {
+        SecurityResourceFilter resourceFilter = filter.getResourceFilter();
+        return resourceFilter != null && resourceFilter.areOnlySpecificObjectsAllowed()
+                && resourceFilter.getAllowedObjects().isEmpty();
     }
 
     /** A document contributed to the inventory when at least one asset-source row points at it. */
@@ -424,21 +485,65 @@ public class CryptographicAssetServiceImpl implements CryptographicAssetExternal
         return dto;
     }
 
-    private static CryptographicAssetDetailDto toDetailDto(CryptoAsset asset, List<CryptoAssetSource> sources) {
+    private static CryptographicAssetDetailDto toDetailDto(CryptoAsset asset, List<CryptoAssetSource> sources,
+            Set<UUID> visibleCbomUuids) {
         CryptographicAssetDetailDto dto = new CryptographicAssetDetailDto();
         dto.setUuid(asset.getUuid());
         dto.setName(servedName(asset.getName(), asset.getOid(), asset.getIdentityGuard()));
         dto.setType(asset.getAssetType());
         dto.setPqcVerdict(servedVerdict(asset.getPqcVerdict()));
+        // GLOBAL badges: computed over every loaded row, before the CBOM visibility filter below, so they keep
+        // reconciling with the list (scoped by CRYPTO_ASSET, not CBOM) exactly as the list endpoint serves them --
+        // the visibility gate below is on per-document CONTENT, not on whether a document contributed.
         dto.setSourceCbomCount(asset.getSourceCount());
         dto.setOccurrenceCount(sources.stream().mapToLong(CryptoAssetSource::getOccurrenceCount).sum());
         dto.setQuarantined(quarantined(asset.getIdentityGuard()));
         dto.setVerdict(toVerdictDto(asset));
         dto.setNormalizedFields(toNormalizedFieldsDto(asset));
-        dto.setElectedPayload(asset.getMergedCryptoProperties());
-        dto.setSources(sources.stream().map(CryptographicAssetServiceImpl::toSourceDto).toList());
+        dto.setElectedPayload(servedElectedPayload(asset, sources, visibleCbomUuids));
+        dto.setSources(servedSources(sources, visibleCbomUuids));
         dto.setOids(toOidDtos(asset));
         return dto;
+    }
+
+    /**
+     * Serial numbers, versions and payloads belong to the CBOM resource -- same rule as
+     * {@link #cbomSerialNumbersScopedToCaller} -- so a source is served only for a CBOM the caller may see. Visible
+     * rows are then capped at {@link #MAX_SERVED_SOURCES}, oldest-first: {@code sources} already arrives in that order
+     * (see {@code CryptoAssetSourceRepository#findWithCbomByAssetUuid}), so filtering before limiting keeps it.
+     */
+    private static List<CryptographicAssetSourceDto> servedSources(List<CryptoAssetSource> sources,
+            Set<UUID> visibleCbomUuids) {
+        return sources
+                .stream()
+                .filter(source -> visibleCbomUuids.contains(source.getCbomUuid()))
+                .limit(MAX_SERVED_SOURCES)
+                .map(CryptographicAssetServiceImpl::toSourceDto)
+                .toList();
+    }
+
+    /**
+     * The elected payload is one document's payload verbatim -- {@code asset.getMergedCryptoProperties()} is not
+     * synthesised, it is the electing source's own content -- so unlike the row-level badges (counts) it is gated by
+     * that ONE document's visibility, the same rule {@link #servedSources} applies per row. Election itself stays
+     * global and deterministic: an invisible electing source suppresses the payload rather than re-electing among
+     * whatever the caller can see, which would make the served payload depend on who is asking.
+     */
+    private static Map<String, Object> servedElectedPayload(CryptoAsset asset, List<CryptoAssetSource> sources,
+            Set<UUID> visibleCbomUuids) {
+        UUID electingSourceUuid = asset.getPropertiesSourceUuid();
+        if (electingSourceUuid == null) {
+            return null;
+        }
+        // A pointer naming a row absent from the loaded list (should not happen, but is not this method's invariant
+        // to enforce) is treated the same as an invisible one: no visible row, no payload.
+        boolean electingDocumentVisible = sources
+                .stream()
+                .filter(source -> electingSourceUuid.equals(source.getUuid()))
+                .findFirst()
+                .map(source -> visibleCbomUuids.contains(source.getCbomUuid()))
+                .orElse(false);
+        return electingDocumentVisible ? asset.getMergedCryptoProperties() : null;
     }
 
     /**
@@ -500,11 +605,37 @@ public class CryptographicAssetServiceImpl implements CryptographicAssetExternal
     /** Stored evidence is the capped producer occurrence array; the keys are CycloneDX occurrence field names. */
     private static CryptographicAssetEvidenceDto toEvidenceDto(Map<String, Object> occurrence) {
         CryptographicAssetEvidenceDto dto = new CryptographicAssetEvidenceDto();
-        dto.setLocation(occurrence.get("location") instanceof String location ? location : null);
-        dto.setLine(occurrence.get("line") instanceof Number line ? line.intValue() : null);
-        dto.setOffset(occurrence.get("offset") instanceof Number offset ? offset.intValue() : null);
-        dto.setSymbol(occurrence.get("symbol") instanceof String symbol ? symbol : null);
+        dto.setLocation(asServedText(occurrence.get("location")));
+        dto.setLine(asServedInteger(occurrence.get("line")));
+        dto.setOffset(asServedInteger(occurrence.get("offset")));
+        dto.setSymbol(asServedText(occurrence.get("symbol")));
         return dto;
+    }
+
+    /** A producer occasionally emits a numeric evidence field as a string; the value still names a real position. */
+    private static Integer asServedInteger(Object value) {
+        if (value instanceof Number number) {
+            return number.intValue();
+        }
+        if (value instanceof String text) {
+            try {
+                return Integer.valueOf(text.trim());
+            } catch (NumberFormatException e) {
+                return null;
+            }
+        }
+        return null;
+    }
+
+    /** A locator or symbol served verbatim when textual, or in its decimal form when a producer emitted a number. */
+    private static String asServedText(Object value) {
+        if (value instanceof String text) {
+            return text;
+        }
+        if (value instanceof Number number) {
+            return String.valueOf(number);
+        }
+        return null;
     }
 
     private static List<CryptographicAssetOidDto> toOidDtos(CryptoAsset asset) {
--- a/src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
@@ -1,7 +1,6 @@
 package com.otilm.core.service.impl;
 
 import com.otilm.api.exception.NotFoundException;
-import com.otilm.api.exception.NotSupportedException;
 import com.otilm.api.exception.ValidationException;
 import com.otilm.api.model.client.certificate.SearchFilterRequestDto;
 import com.otilm.api.model.client.certificate.SearchRequestDto;
@@ -11,17 +10,27 @@ import com.otilm.api.model.common.PaginationResponseDto;
 import com.otilm.api.model.core.auth.Resource;
 import com.otilm.api.model.core.cryptoasset.CryptographicAssetDetailDto;
 import com.otilm.api.model.core.cryptoasset.CryptographicAssetDto;
+import com.otilm.api.model.core.cryptoasset.CryptographicAssetEvidenceDto;
+import com.otilm.api.model.core.cryptoasset.CryptographicAssetNormalizedFieldsDto;
+import com.otilm.api.model.core.cryptoasset.CryptographicAssetOidDto;
+import com.otilm.api.model.core.cryptoasset.CryptographicAssetSourceDto;
+import com.otilm.api.model.core.cryptoasset.CryptographicAssetVerdictDto;
 import com.otilm.api.model.core.cryptoasset.PqcVerdict;
 import com.otilm.api.model.core.scheduler.PaginationRequestDto;
 import com.otilm.api.model.core.search.FilterFieldSource;
 import com.otilm.api.model.core.search.SearchFieldDataByGroupDto;
 import com.otilm.api.model.core.search.SearchFieldDataDto;
 import com.otilm.core.comparator.SearchFieldDataComparator;
+import com.otilm.core.dao.entity.Cbom;
 import com.otilm.core.dao.entity.Cbom_;
+import com.otilm.core.dao.entity.UniquelyIdentified_;
 import com.otilm.core.dao.entity.cbom.CryptoAsset;
+import com.otilm.core.dao.entity.cbom.CryptoAssetSource;
+import com.otilm.core.dao.entity.cbom.CryptoAssetSource_;
 import com.otilm.core.dao.entity.cbom.CryptoAsset_;
 import com.otilm.core.dao.repository.CbomRepository;
 import com.otilm.core.dao.repository.cbom.CryptoAssetRepository;
+import com.otilm.core.dao.repository.cbom.CryptoAssetSourceRepository;
 import com.otilm.core.enums.FilterField;
 import com.otilm.core.model.auth.ResourceAction;
 import com.otilm.core.model.cbom.CryptoAssetIdentityGuard;
@@ -40,39 +49,45 @@ import jakarta.persistence.criteria.CriteriaQuery;
 import jakarta.persistence.criteria.Expression;
 import jakarta.persistence.criteria.Predicate;
 import jakarta.persistence.criteria.Root;
+import jakarta.persistence.criteria.Subquery;
+import java.time.OffsetDateTime;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.function.TriFunction;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Pageable;
+import org.springframework.security.concurrent.DelegatingSecurityContextExecutorService;
 import org.springframework.stereotype.Service;
 
 /**
- * Serves the ratified cryptographic asset inventory contract. List and searchable-fields are real reads over the
- * deduplicated cross-CBOM asset projection; detail, statistics and sync visibility remain {@link NotSupportedException}
- * refusals until core#2145 builds them. The service also serves the per-object listing behind {@code @AuthEndpoint} --
- * the role editor's object picker.
- *
- * <p>
- * {@link NotSupportedException} maps to HTTP 501 in {@code ExceptionHandlingAdvice}, so a permitted caller learns a
- * still-refused operation is not implemented while an unpermitted caller is refused with 403 by the authorization
- * aspect ahead of this body.
+ * Serves the ratified cryptographic asset inventory contract: list, detail, searchable-fields and the dashboard
+ * statistics are all real reads over the deduplicated cross-CBOM asset projection. The service also serves the
+ * per-object listing behind {@code @AuthEndpoint} -- the role editor's object picker.
  */
 @Service(Resource.Codes.CRYPTO_ASSET)
+@Slf4j
 public class CryptographicAssetServiceImpl implements CryptographicAssetExternalService, ResourceExtensionService {
 
-    private static final String NOT_IMPLEMENTED = "Cryptographic asset inventory is not implemented yet";
+    /** Top-N cap on the served algorithm-family distribution; the badge pair carries the overflow. */
+    private static final int TOP_ALGORITHM_FAMILIES = 10;
 
     // The OpenAPI schema documents 1000 as the maximum items per page; this is where it is actually enforced.
     private static final int MAX_ITEMS_PER_PAGE = 1000;
 
     private CryptoAssetRepository cryptoAssetRepository;
+
+    private CryptoAssetSourceRepository cryptoAssetSourceRepository;
 
     private CbomRepository cbomRepository;
 
@@ -81,6 +96,11 @@ public class CryptographicAssetServiceImpl implements CryptographicAssetExternal
     @Autowired
     public void setCryptoAssetRepository(CryptoAssetRepository cryptoAssetRepository) {
         this.cryptoAssetRepository = cryptoAssetRepository;
+    }
+
+    @Autowired
+    public void setCryptoAssetSourceRepository(CryptoAssetSourceRepository cryptoAssetSourceRepository) {
+        this.cryptoAssetSourceRepository = cryptoAssetSourceRepository;
     }
 
     @Autowired
@@ -125,7 +145,11 @@ public class CryptographicAssetServiceImpl implements CryptographicAssetExternal
     @Override
     @ExternalAuthorization(resource = Resource.CRYPTO_ASSET, action = ResourceAction.DETAIL)
     public CryptographicAssetDetailDto getCryptographicAsset(SecuredUUID uuid) throws NotFoundException {
-        throw new NotSupportedException(NOT_IMPLEMENTED);
+        CryptoAsset asset = cryptoAssetRepository
+                .findByUuid(uuid)
+                .orElseThrow(() -> new NotFoundException(CryptoAsset.class, uuid));
+        List<CryptoAssetSource> sources = cryptoAssetSourceRepository.findWithCbomByAssetUuid(asset.getUuid());
+        return toDetailDto(asset, sources);
     }
 
     /**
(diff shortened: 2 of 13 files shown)


src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -125,7 +145,11 @@ public PaginationResponseDto<CryptographicAssetDto> listCryptographicAssets(Secu
     @Override
     @ExternalAuthorization(resource = Resource.CRYPTO_ASSET, action = ResourceAction.DETAIL)
     public CryptographicAssetDetailDto getCryptographicAsset(SecuredUUID uuid) throws NotFoundException {
-        throw new NotSupportedException(NOT_IMPLEMENTED);
+        CryptoAsset asset = cryptoAssetRepository
```
vyskocilm, line 156: 🔴 bug: **the detail endpoint hands CBOM data across the resource boundary with no scoping.**

This one is worth spelling out because it is an authorization finding rather than a defect in the reshaping. `findWithCbomByAssetUuid` is plain JPQL with no `SecurityFilter`, so every source row of the asset is returned regardless of whether the caller may see the document it came from. A caller holding `cryptoAssets:detail` but denied CBOM object access reads, for each contributing document: `cbomUuid`, `serialNumber`, `version`, `source` (the CycloneDX `metadata.component.name`) and the raw producer `payload`.

Scenario: operator restricted to CBOM `A`; asset X is reported by both `A` and `B`. `GET /v1/cryptoAssets/X` returns B's serial number, its metadata component name and its full crypto-properties payload.

This file already establishes the opposite rule 190 lines down — `cbomSerialNumbersScopedToCaller()` exists precisely because serial numbers belong to the CBOM resource — so the diff contradicts its own stated principle rather than leaving a question open.

Fix: populate a `Resource.CBOM/LIST` filter the way the statistics path already does and filter the source rows through it, or add a security-filtered finder.

Inert today: nothing in `src/main` calls `CryptoAssetSourceWriter`, so `crypto_asset_source` stays empty until core#2073. Worth closing before it fills.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -310,15 +419,101 @@ private static CryptographicAssetDto toDto(CryptoAssetListRow row) {
...
+
+    private static CryptographicAssetDetailDto toDetailDto(CryptoAsset asset, List<CryptoAssetSource> sources) {
+        CryptographicAssetDetailDto dto = new CryptographicAssetDetailDto();
+        dto.setUuid(asset.getUuid());
+        dto.setName(servedName(asset.getName(), asset.getOid(), asset.getIdentityGuard()));
+        dto.setType(asset.getAssetType());
+        dto.setPqcVerdict(servedVerdict(asset.getPqcVerdict()));
+        dto.setSourceCbomCount(asset.getSourceCount());
+        dto.setOccurrenceCount(sources.stream().mapToLong(CryptoAssetSource::getOccurrenceCount).sum());
+        dto.setQuarantined(quarantined(asset.getIdentityGuard()));
```
vyskocilm, line 499: 🔴 bug: `verdict` is `requiredMode = REQUIRED` in the ratified DTO and is omitted from every response.

`toVerdictDto` returns `null` when `pqcEvaluatedAt == null`; `CryptographicAssetDetailDto` carries `@JsonInclude(NON_NULL)` and the wire mapper is built with global `NON_NULL` (`ObjectMapperFactory.wire():32`). Confirmed from the jar: `verdict -> requiredMode=REQUIRED`.

`pqcEvaluatedAt` is written only by `CryptoAssetRepository.applyPqcVerdict` <- `CryptoAssetWriter.applyPqcVerdict`, which has no production caller. So this is not an edge case: **every** detail response is schema-invalid until core#2151 wires evaluation, and a generated strict client or a contract test rejects it.

The body calls this "contract friction". It is a live contract break. Fix: land interfaces#938 first, or serve the block with `evaluatedAt == null`.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -176,7 +200,69 @@ public List<SearchFieldDataByGroupDto> getSearchableFieldInformationByGroup() {
     @Override
     @ExternalAuthorization(resource = Resource.CRYPTO_ASSET, action = ResourceAction.LIST)
     public CryptographicAssetStatisticsDto getCryptographicAssetStatistics() {
-        throw new NotSupportedException(NOT_IMPLEMENTED);
+        // No SecurityFilter parameter reaches this method (the contract takes none), so both scopes are built
+        // by hand: assets under the caller's CRYPTO_ASSET LIST scope -- which is what makes these numbers
+        // reconcile with the list -- and the document-level completeness block under the CBOM LIST scope.
+        SecurityFilter assetFilter = SecurityFilter.create();
+        objectFilterAspect.populateSecurityFilter(Resource.CRYPTO_ASSET, ResourceAction.LIST, null, null, assetFilter);
```
vyskocilm: 🔴 bug: **a CBOM object restriction silently serves zeros that read as a never-synced estate.**

Second authorization finding, so again in full. `cbomFilter` is built for `CBOM/LIST`. A denied vote yields "only specific objects, none allowed", and the three CBOM aggregates then return 0/0/null rather than being withheld.

Scenario: operator with `cryptoAssets:list` and no CBOM access opens the dashboard and gets `totalAssets: 2, sourceCbomCount: 0, cbomStatBySyncState` all zero, `lastCompletedSyncAt: null`. A payload asserting that two assets exist and that zero documents ever contributed to them — byte-identical to what a genuinely unsynced platform returns. The operator cannot tell the difference, and neither can a support engineer reading a screenshot.

`CryptographicAssetStatisticsITest:206-225` asserts this as intended behaviour, so the test pins the wrong contract. And it is the same harm the method's own comment forbids twenty lines up: *"must fail the request rather than silently serve zeros that no longer reconcile"*. A permission-shaped zero is exactly a zero that no longer reconciles.

Unlike the two findings above this is reachable **today** — `Cbom` rows are populated by `CbomSyncTask`.

Fix: gate the completeness block on CBOM resource access and omit it entirely when absent, or carry an explicit "scoped" marker. A value list degrading to empty is defensible; a count degrading to zero is a false statement.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -294,10 +380,33 @@ private static Expression<String> displayLabel(Root<CryptoAsset> root, CriteriaB
...
+     * describes exactly one guard: REFUTED_CERTIFICATE_DIGEST (contradicting claims, reconciliation queue).
+     * BARE_CN_SUBJECT is a permanent DN-scope split with, by the ratified addendum, no reconciliation path at all, and
+     * REFUTED_OID is one component contradicting itself; flagging either would advertise an operator queue that can
+     * never drain. Widening this needs the flag re-worded in interfaces, not a wider test here.
+     */
+    private static boolean quarantined(CryptoAssetIdentityGuard guard) {
+        return guard == CryptoAssetIdentityGuard.REFUTED_CERTIFICATE_DIGEST;
     }
 
     private static CryptographicAssetDto toDto(CryptoAssetListRow row) {
```
vyskocilm, line 472: 🟡 risk: no cap on `sources` or on per-source `evidence`. `findWithCbomByAssetUuid` is unpaginated and `toEvidenceDtos` serves every retained occurrence.

`OccurrenceEvidenceCapper` bounds one *source row* at 50 occurrences / 64 KB; the asset-level fan-out is `sourceCount`, one row per CBOM. `rsa-2048` in a 5,000-CBOM estate -> 5,000 x (payload + up to 64 KB evidence), hundreds of MB in one response, on an endpoint any authenticated operator can hit.

The list endpoint caps at `MAX_ITEMS_PER_PAGE = 1000` and serves no payloads at all; detail serves all payloads uncapped. Fix: cap the source list and say so on the wire, the way the occurrence cap already does with `occurrenceCount`.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -176,7 +200,69 @@ public List<SearchFieldDataByGroupDto> getSearchableFieldInformationByGroup() {
...
+    private static Predicate hasContributedAssets(Root<Cbom> root, CriteriaBuilder cb, CriteriaQuery<?> query) {
+        Subquery<Integer> contributed = query.subquery(Integer.class);
+        Root<CryptoAssetSource> source = contributed.from(CryptoAssetSource.class);
+        contributed
+                .select(cb.literal(1))
+                .where(cb.equal(source.get(CryptoAssetSource_.cbomUuid), root.get(UniquelyIdentified_.uuid)));
+        return cb.exists(contributed);
+    }
+
+    private OffsetDateTime lastCompletedSyncAt(SecurityFilter cbomFilter) {
```
vyskocilm, line 317: 🔵 nit: `lastCompletedSyncAt` does two round trips and the second is unfiltered — `findUuidsUsingSecurityFilter` then `cbomRepository.findById(...)`.

Harmless for authorization (the uuid came from a filtered query), but `orElse(null)` reports "never synced" over a live estate if the row is deleted between the two calls. A projection (`SELECT MAX(assetsSyncedAt)`, or selecting the timestamp in the paged query) removes both the race and the round trip.

vyskocilm (commented):
Feature review, Opus 5, independent of the convention checklist. **3 bugs / 3 risks / 5 nits**, inline. Podman is broken on my box (`conmon.pid: no such file`), so no `*ITest` ran — the calculator, the ArchUnit fence and the DTO contract were driven directly instead, and every behavioural claim below carries its executed output. What I could not run is listed at the end.

The three bugs are, in order of how soon they bite: statistics serving permission-shaped zeros that read as a never-synced estate (**live today**), `verdict` omitted from every response against a REQUIRED contract (**live the moment any asset exists**), and the detail endpoint returning CBOM data with no CBOM scoping (**inert until core#2073**).

### Two claims in the PR body that do not hold

> the two assertions below (the vocabulary regex and the literal recomputed key) are checking something real rather than passing vacuously over a near-empty response

Half true. The seeded maps are `{"name":"ecdsa","curve":"P-256"}` and `{"checked":"field"}` — neither contains the key literal, so `doesNotContain(theSeededKeyLiteral())` cannot fail regardless of what the mapper does with the map channels. It proves the mapper injects no key. It does not prove a key sitting *inside* a stored payload would be withheld — and that test would fail, because there is no read-side redaction at all. Worth saying plainly, since the hand-check exists precisely to cover what the lexical fence cannot see.

> this surface serves the column verbatim and is the backstop

`setElectedPayload(asset.getMergedCryptoProperties())`, `setPayload(source.getOriginalCryptoProperties())` and `setEvaluatedFields(asset.getPqcEvaluatedFields())` are unfiltered pass-through to Jackson. A pass-through is the opposite of a backstop; the guarantee rests entirely on write paths that do not exist yet. Also, "verbatim" holds at the value level only — `merged_crypto_properties` is a `jsonb` copy, and Postgres reorders keys and drops duplicates.

Two smaller ones: "Aggregation failures fail the request rather than serve zeros" is true for exceptions (verified: every `.get()` propagates, no swallow) but false for a CBOM restriction, which serves exactly those zeros by design. And `deniesDetailWhenTheDetailActionIsDenied` / `deniesStatisticsWhenTheListActionIsDenied` would pass unchanged against `origin/main`, where both methods threw `NotSupportedException` *after* the method-security check — they pin the acceptance criterion, not this diff. (`detailNotFoundIsA404NotALeak` does pin new behaviour: main returned 501.)

### Verified correct — please do not spend a second reviewer on these

- Statistics and list use the **identical** scoping path: `populateSecurityFilter(CRYPTO_ASSET, LIST, null, null, f)` funnels into the same method the `@ExternalAuthorization` aspect uses, `parentName = NONE` both times, predicates from the same `getPredicates`, no join introduced either side — so `count(root)` and the list count the same row set. `totalAssets` reconciliation is sound.
- No lazy load and no transaction crosses a thread boundary. No `@Transactional` on the method; the injected `EntityManager` is Spring's shared proxy, so each virtual thread gets its own EM and connection (7 + request thread against a pool of 40 — the same shape `CertificateServiceImpl:2147` already ships). `JOIN FETCH s.cbom` initialises the only lazy association read.
- Executor shutdown is correct on every path, including the `InterruptedException` branch: `close()` runs before the catch clauses, `Thread.interrupt()` after, so it cannot livelock.
- Fail-the-request is real: any task failure surfaces at the first `.get()` as `ExecutionException` -> `IllegalStateException` -> 500, nothing caught per task, no zero substituted.
- No raw exception message reaches the wire (`ExceptionHandlingAdvice:682` -> `"Internal server error."`); the 404 body is class name + uuid only.
- Top-10 tie-break is deterministic — 11 all-tied families across 5 shuffled insertion orders gave identical output every time. `sorted()` is stable, comparator total on unique keys, no JDK or hash-order dependence.
- Dense enum maps are complete and match the `@Check` constraints exactly (5/4/4); a constant added later cannot vanish, because `densifiesEveryEnumCodeWithZero` asserts exact `Map.equals`.
- A refuted OID reaches no display-name path: `displayLabel` (SQL) and `servedName` (in-memory) both null it, and `toOidDtos` serves it only alongside `refuted: true`.
- `lastCompletedSyncAt` semantics are right — `markFailed` passes `syncedAt = null` into `COALESCE`, so a non-null timestamp really does mean at least one successful sync.
- The ArchUnit fence covers both new/changed main files and demonstrably fires: planted `dto.setName(asset.getIdentityKey())`, a planted `private String identityKey;` field and a planted `log.debug("identity_key {}", k)` all produce violations, the real scan is clean, neither file is on `SOURCE_ALLOWLIST`. `IdentityKeyExposureFence*` green, 26 tests.
- Detail is object-level authorized (`ExternalMethodAuthorizationManager:109-114` extracts `SecuredUUID` for the OPA vote), and statistics sharing `CRYPTO_ASSET LIST` is genuinely enforced, not just documented.
- Detail's `occurrenceCount` matches the list's `COALESCE(SUM(...), 0L)` over the same rows, and `JOIN FETCH` cannot drop a row (`cbom_uuid` NOT NULL, RESTRICT FK).
- The reconciliation test is **not** vacuous — it asserts the absolute expected count (5, then 3) before the relational equalities, so a silently-broken OPA stub fails the first assertion. The one thing the previous adversarial pass fixed stayed fixed.

### Not runnable here

`CryptographicAssetDetailITest`, `CryptographicAssetStatisticsITest`, `CryptographicAssetAuthorizationHttpITest`, `CryptographicAssetControllerITest`, `CbomServiceITest` — all 10 attempts died with `Could not find a valid Docker environment`. So unverified by execution: the enum-code group keys against a live Hibernate query, the `hasContributedAssets` EXISTS subquery compiling inside the count query, the jsonPath shapes, the 403/404 statuses, the wire-level regex over real responses, and both the 6319-test and 0.9095-coverage figures. Everything else above was executed.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -310,15 +419,101 @@ private static CryptographicAssetDto toDto(CryptoAssetListRow row) {
...
+    private static CryptographicAssetNormalizedFieldsDto toNormalizedFieldsDto(CryptoAsset asset) {
+        if (asset.getAlgorithmFamily() == null && asset.getPrimitive() == null && asset.getParameterSet() == null
+                && asset.getCurve() == null && asset.getMode() == null && asset.getPadding() == null
+                && asset.getVariant() == null) {
+            return null;
+        }
+        CryptographicAssetNormalizedFieldsDto dto = new CryptographicAssetNormalizedFieldsDto();
+        dto.setAlgorithmFamily(asset.getAlgorithmFamily());
+        dto.setPrimitive(asset.getPrimitive());
+        dto.setParameterSet(asset.getParameterSet());
```
vyskocilm, line 576: 🔵 nit: `occurrence.get("line") instanceof Number` drops a producer field of the wrong JSON type silently. A producer emitting `"line": "42"` yields `line` absent from the wire, and the stored value is reachable by no other path since `evidence` is not served raw. Same for `offset`, `location`, `symbol`. Coerce numeric strings, or log at debug.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -310,15 +419,101 @@ private static CryptographicAssetDto toDto(CryptoAssetListRow row) {
...
+    }
+
+    private static CryptographicAssetSourceDto toSourceDto(CryptoAssetSource source) {
+        Cbom cbom = source.getCbom();
+        CryptographicAssetSourceDto dto = new CryptographicAssetSourceDto();
+        dto.setCbomUuid(source.getCbomUuid());
+        dto.setSerialNumber(cbom.getSerialNumber());
+        dto.setVersion(cbom.getVersion());
+        dto.setOccurrenceCount(source.getOccurrenceCount());
+        dto.setSource(cbom.getSource());
```
vyskocilm, line 591: 🔵 nit: two more REQUIRED fields this surface can omit. `CryptographicAssetSourceDto.source` is REQUIRED but `CbomServiceImpl:260,638` writes `getMetadataComponentName(content).orElse(null)`. `CryptographicAssetDto.name` is REQUIRED and `servedName` returns null for a `REFUTED_OID` row with no producer name — asserted at `CryptographicAssetDetailITest:163`.

Both mirror a pre-existing platform violation (`CbomDto.source` is REQUIRED and equally nullable), hence nit rather than bug — but the `verdict` thread above means this diff now has three REQUIRED fields it can omit.

src/main/java/com/otilm/core/service/impl/CryptographicAssetStatisticsCalculator.java
```diff
@@ -0,0 +1,89 @@
...
+            Map<String, Long> cbomBySyncState, OffsetDateTime lastCompletedSyncAt) {
+        CryptographicAssetStatisticsDto dto = new CryptographicAssetStatisticsDto();
+        dto.setTotalAssets(totalAssets);
+        dto.setSourceCbomCount(sourceCbomCount);
+        dto.setStatByType(dense(byType, typeCodes(), null));
+        dto.setStatByPqcVerdict(dense(byVerdict, verdictCodes(), PqcVerdict.UNKNOWN.getCode()));
+        Map<String, Long> families = new HashMap<>(byFamily);
+        Long unassigned = families.remove(UNASSIGNED_KEY);
+        dto.setUnassignedAssetCount(unassigned == null ? 0L : unassigned);
+        dto.setDistinctAlgorithmFamilyCount((long) families.size());
```
vyskocilm, line 50: 🟡 risk: the two overflow badges do not let a client reconcile the truncated chart.

Executed — 11 families each at count 5, `totalAssets = 55`: `statByAlgorithmFamily` sums to **50**, `distinctAlgorithmFamilyCount = 11`, `unassignedAssetCount = 0`. The badges say that truncation happened, never how many assets fell outside the top 10, and the DTO has no "other" bucket. Any client rendering "chart + unassigned = total" is silently wrong.

Fix: an `otherAssetCount` in the contract, or serve the remainder as one folded slot.

src/main/java/com/otilm/core/service/impl/CryptographicAssetStatisticsCalculator.java
```diff
@@ -0,0 +1,89 @@
...
+        completeness.setLastCompletedSyncAt(lastCompletedSyncAt);
+        dto.setSyncCompleteness(completeness);
+        return dto;
+    }
+
+    /** Every code present with 0 when none, in enum declaration order; a NULL bucket folds into foldNullInto. */
+    private static Map<String, Long> dense(Map<String, Long> sparse, List<String> codes, String foldNullInto) {
+        Map<String, Long> result = new LinkedHashMap<>();
+        for (String code : codes) {
+            result.put(code, sparse.getOrDefault(code, 0L));
```
vyskocilm, line 68: 🟡 risk: `dense()` silently discards any group key not in the enum code list, including the NULL bucket when `foldNullInto == null`.

Executed: `assemble(5, {"algorithm":3,"Unassigned":2}, ...)` -> `statByType` sums to **3** against `totalAssets = 5`, no error, no finding.

Unreachable today — `CryptoAsset.assetType` and `Cbom.assetSyncState` are both `nullable = false` and the code lists match the DB check constraints exactly (5/4/4, verified). But the two `null` arguments at L46 and L51 are the only thing between a future nullable column and assets vanishing from a chart with no exception and no failing test. Make `dense()` throw or fold to an explicit bucket on an unmapped key.

src/main/java/com/otilm/core/service/impl/CryptographicAssetStatisticsCalculator.java
```diff
@@ -0,0 +1,89 @@
...
+ * Pure transformations behind the cryptographic asset statistics, kept free of Spring and persistence so they can be
+ * unit-tested directly. The SQL aggregation lives in the service; this class only reshapes its results.
+ */
+final class CryptographicAssetStatisticsCalculator {
+
+    /**
+     * The literal {@code SecurityFilterRepositoryImpl.countGroupedUsingSecurityFilter} substitutes for a NULL group
+     * key. For verdicts it means "never evaluated" and folds into {@code unknown}; for families it is the contract's
+     * unassignedAssetCount and never occupies a top-N slot.
+     */
```
vyskocilm, line 33: 🟡 risk: the `"Unassigned"` sentinel is a literal duplicated from `SecurityFilterRepositoryImpl.java:251`, and its safety rests on an undocumented invariant.

Executed: a genuine family spelled `unassigned` and the sentinel `Unassigned` coexist correctly (`families={unassigned=7}`, `unassignedAssetCount=3`) — but only because `CryptoAssetIdentityFields.fold()` lowercases. This is the first third-party-producer-supplied free-text column routed through `countGroupedUsingSecurityFilter`, whose `Collectors.toMap` has no merge function: a stored `Unassigned` beside any NULL family throws `IllegalStateException: Duplicate key`, wrapped into the 500 at `CryptographicAssetServiceImpl:240`.

Fix: state the dependency in the javadoc, or share one constant and give the collector a merge function.

src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetSourceRepository.java
```diff
@@ -26,6 +26,19 @@ public interface CryptoAssetSourceRepository extends SecurityFilterRepository<Cr
 
     boolean existsByCbomUuid(UUID cbomUuid);
 
+    /**
+     * The detail endpoint's source list: every CBOM's contribution with its document header, oldest contribution first,
+     * uuid as the deterministic tiebreak. The fetch join loads the read-only {@link CryptoAssetSource#getCbom()}
+     * association in the same query, so callers can read header fields without a session.
+     */
+    @Query("""
```
vyskocilm, line 34: 🔵 nit (pre-existing, surfaced by this endpoint): the "true pre-cap `occurrenceCount`" counts JSON-null array elements. `CryptoAssetSourceWriter:81-83` stores `occurrences.size()` while `OccurrenceEvidenceCapper.cap` does `filter(Objects::nonNull)` first, so a producer array `[{...}, null, null]` stores count 3 for one real occurrence.

Detail and list agree with each other (both read the same column), so it is an accuracy issue rather than a divergence — worth a look while the table is still empty.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Fixed in 7f2dd363. Source rows now pass through a CBOM/LIST visibility set built by the same populateSecurityFilter path the statistics block uses (EXISTS-scoped to this asset, resolved by the SecurityFilterRepository machinery — no duplicated predicate logic). Row badges (sourceCbomCount, occurrenceCount) deliberately stay global so detail keeps reconciling with the list; the gate is on per-document content. The scoped re-review of the fix also caught the twin channel: electedPayload is byte-for-byte the electing document's payload, so it is now served only when that document (via the provenance pointer) is visible — no per-caller re-election, the election stays contractually deterministic. Tests: deny-all (sources empty, badges global, electedPayload absent), partial restriction serving the electing document, and the sharp case where the electing document is hidden while another stays visible (visible row served, electedPayload absent, no substitution).

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Agreed it is a live contract break for strict clients, and took your first option: interfaces#940 is open flipping verdict to NOT_REQUIRED with the absence semantics in the description (closes interfaces#938). Serving behavior stays as shipped — a fabricated all-default block would present "never evaluated" as a decision. If #940 is rejected, the fallback is serving the block with evaluatedAt null; happy to pivot. Your nit on source/name is noted on interfaces#938 for the same reconciliation sweep.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Fixed in 7f2dd363, your first option: a fully denied CBOM scope now omits sourceCbomCount and syncCompleteness entirely (both contract-optional; NON_NULL drops them) — absence instead of a false statement. A partial object restriction still serves the scoped counts, which is the document-level-counts disclosure the contract note already accepts. The ITest that pinned the wrong contract is flipped to assert absence under deny-all, and a new test pins scoped counts under partial restriction.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Fixed in 7f2dd363: served sources capped at MAX_SERVED_SOURCES = 100, applied after the visibility filter, oldest-first (the repository's existing deterministic order); sourceCbomCount keeps carrying the true total — the same served-vs-true pattern occurrenceCount already uses. A 101-document test pins the cap, the true badges, and which rows survive (oldest in, newest out). Documenting the cap in the contract schema text can ride along with interfaces#941.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Kept as-is, deliberately: the uuid comes from the filtered query, so the second read is authorization-safe; a deletion between the two calls serves null once and self-heals on the next request. A filter-aware MAX needs new SecurityFilterRepository surface, which felt disproportionate for that race window — say the word if you want it and I will add the projection.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Fixed in 7f2dd363: line/offset now accept a Number or a numeric string, location/symbol accept a String or a Number (stringified); covered by test. One wrinkle the fix surfaced: OccurrenceEvidenceCapper blanks non-string locations to "" at write time, so the numeric-location case is only reachable via direct row writes — coerced anyway so the read side is total.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Raised on the contract side rather than papered over here: verdict is interfaces#940; source and name are flagged on interfaces#938 as the remaining REQUIRED-but-omittable fields for the same sweep (relax vs. guarantee a value is a contract ruling, not a serving-side one). Serving behavior unchanged — a refuted OID must not become the display name.

src/main/java/com/otilm/core/service/impl/CryptographicAssetStatisticsCalculator.java
karolhrdina: Filed as interfaces#941 proposing otherAssetCount (Long, not required): your folded-slot alternative was considered and rejected because the map keys are producer-derived family names, so a synthetic key is collision-prone and distorts the stated key semantics. Serving side is a two-line change once the field lands.

src/main/java/com/otilm/core/service/impl/CryptographicAssetStatisticsCalculator.java
karolhrdina: Fixed in 7f2dd363: dense() now throws on any unmapped key — including the NULL-bucket sentinel when there is no fold target — naming the key and the expected code list. Consistent with the endpoint's fail-loud rule; a silently discarded bucket is a chart that lies. Unit tests cover both throw paths and the still-folding verdict path.

src/main/java/com/otilm/core/service/impl/CryptographicAssetStatisticsCalculator.java
karolhrdina: Documented in 7f2dd363: the sentinel javadoc now states both invariants — the literal is duplicated from SecurityFilterRepositoryImpl's NULL-key substitution, and collision with stored data is prevented because CryptoAssetIdentityFields.fold() lowercases producer text, so a stored "Unassigned" cannot exist. Left the shared repository's Collectors.toMap alone — a merge function there changes behavior for every resource and deserves its own change if wanted.

src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetSourceRepository.java
karolhrdina: Fixed in 7f2dd363 (writer-side, while the table is still empty as you suggested): the stored count now counts only non-null array elements, using the same predicate cap() uses to decide what an occurrence is — the count and the retained evidence can no longer disagree about JSON nulls. Covered in CryptoAssetInventoryITest.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -125,7 +145,11 @@ public PaginationResponseDto<CryptographicAssetDto> listCryptographicAssets(Secu
     @Override
     @ExternalAuthorization(resource = Resource.CRYPTO_ASSET, action = ResourceAction.DETAIL)
     public CryptographicAssetDetailDto getCryptographicAsset(SecuredUUID uuid) throws NotFoundException {
-        throw new NotSupportedException(NOT_IMPLEMENTED);
+        CryptoAsset asset = cryptoAssetRepository
+                .findByUuid(uuid)
+                .orElseThrow(() -> new NotFoundException(CryptoAsset.class, uuid));
+        List<CryptoAssetSource> sources = cryptoAssetSourceRepository.findWithCbomByAssetUuid(asset.getUuid());
+        return toDetailDto(asset, sources);
```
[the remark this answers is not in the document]
karolhrdina: Accepted as a known property rather than changed: per-query snapshots are the shipped dashboard pattern (CertificateServiceImpl's addCertificateStatistics runs the same shape), a single-snapshot variant would serialize seven aggregates on the request path, and the skew window is bounded by query time on an endpoint whose numbers are already time-varying between calls. If exact cross-map consistency becomes a product requirement, a repeatable-read read-only transaction around sequential queries is the shape we would pick.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -310,15 +480,171 @@ private static CryptographicAssetDto toDto(CryptoAssetListRow row) {
...
+    }
+
+    /**
+     * Verdict provenance exists only once a rule set has evaluated the asset; until core#2151 populates verdicts, a
+     * fabricated all-default block would present "never evaluated" as a decision, so the block is omitted and the
+     * row-level verdict alone says UNKNOWN.
+     */
+    private static CryptographicAssetVerdictDto toVerdictDto(CryptoAsset asset) {
+        if (asset.getPqcEvaluatedAt() == null) {
+            return null;
```
vyskocilm, line 555: 🔴 bug: this returns `null` for a never-evaluated asset, but `CryptographicAssetDetailDto.verdict` is `requiredMode = REQUIRED` in the ratified contract (interfaces `CryptographicAssetDetailDto.java:20-21`). The shared object mapper is configured NON_NULL, so the field is not serialized as `null` — it is omitted from the JSON entirely.

Since no verdict exists anywhere until core#2151 populates them, *every* detail response served today violates the published schema, and this diff's own `servesDetailWireShapeAndOmitsVerdictUntilEvaluated` pins that omission as expected. Generated clients type `verdict` as non-optional and will either fail deserialization or hand callers an undefined nobody null-checks.

The omission itself is the right behaviour — fabricating an all-default provenance block would present "never evaluated" as a decision. The problem is the ordering: core shouldn't ship a shape the ratified contract forbids and file the contract fix as follow-up. Your own OmniTrustILM/interfaces#938 flips it to `NOT_REQUIRED`; that wants to land and be consumed here before this merges.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -310,15 +480,171 @@ private static CryptographicAssetDto toDto(CryptoAssetListRow row) {
...
+    }
+
+    private static CryptographicAssetSourceDto toSourceDto(CryptoAssetSource source) {
+        Cbom cbom = source.getCbom();
+        CryptographicAssetSourceDto dto = new CryptographicAssetSourceDto();
+        dto.setCbomUuid(source.getCbomUuid());
+        dto.setSerialNumber(cbom.getSerialNumber());
+        dto.setVersion(cbom.getVersion());
+        dto.setOccurrenceCount(source.getOccurrenceCount());
+        dto.setSource(cbom.getSource());
```
vyskocilm, line 591: 🔴 bug: `Cbom.source` is a nullable column (`Cbom.java:48`) but `CryptographicAssetSourceDto.source` is `requiredMode = REQUIRED`. `CbomServiceImpl:260` and `:638` both write `CbomUtil.getMetadataComponentName(content).orElse(null)`, so any uploaded CBOM whose metadata names no component yields `sources[].source` absent from the wire on a field the schema declares required. That is the ordinary upload path, not a corner case — and the `newCbom()` helpers in the new tests never set a source, so the null path is exercised and unasserted.

Worth deciding in `interfaces` alongside the verdict flip: either this field goes `NOT_REQUIRED`, or ingest writes a placeholder. Note `CbomDto.source` already has the identical mismatch from the same `orElse(null)`, so one decision should cover both rather than leaving two divergent copies.

vyskocilm (commented):
A review of this branch — one pass over the diff plus the architecture and fence tests, run on a local merge of `origin/main` into the head (see the note at the end).

**Four findings, three of them one shape:** the detail payload can omit fields the ratified schema marks required, so the contract fix has to land in `interfaces` before this merges. The fourth is an authorization-scoping pattern divergence on the statistics endpoint. Details are inline on the lines they concern.

**What ran**

- All architecture/fence tests: 102/102 green, including the identity-key exposure fence (`IdentityKeyExposureFenceArchTest` 3+1, `IdentityKeyExposureFenceSelfTest` 22), the transactional-boundary fence 5/5, the external-service authorization fence 6/6, the native-query schema fence, and the object-mapper centralization fence.
- The 7 test classes covering the diff: 127/127 pass (`CryptographicAssetDetailITest` 16, `CryptographicAssetStatisticsITest` 8, `CryptographicAssetStatisticsCalculatorTest` 9, `CbomServiceITest` 44, `CryptoAssetInventoryITest` 39, `CryptographicAssetControllerITest` 2, `CryptographicAssetAuthorizationHttpITest` 9).
- `spotless:check checkstyle:check`: 1998 files clean, 0 violations.
- Coverage on added production lines, recomputed independently with the pom's exclusions and the line+branch formula: **0.9298** (207+58)/(215+70). That is a partial run over the covering classes only, so it is a lower bound — your 0.9095 is comfortably clear of the gate. Uncovered: the two catch arms at 285-291, the `stored == null` guard at 600, and the `NumberFormatException` arm at 623-624.
- Test-context count: unchanged at 61, guard 2/2 green. Both new ITest classes extend `BaseSpringBootTest` with no extra annotations and the calculator test is plain JUnit, so no new Spring context is purchased. Claim confirmed.

**Claims I checked that hold up**

- No new resource or action, no policy bundle, no permission seed, no auth migration — statistics reuses the already-registered CRYPTO_ASSET + LIST pair, so nothing new reaches the startup sync.
- The identity key does not reach any response. A grep-level check flagged four lines naming the key in `CryptographicAssetAuthorizationHttpITest` — **retracting that**: they are assertions of its *absence* using the fence's own regex, in test sources the fence deliberately does not scan, and the machine fence passes. Your four-map-channel hand-trace also holds as far as it can be checked from here: the elected and per-source payloads are producer CycloneDX JSON, `verdict.evaluatedFields` has no production writer yet, and `neitherDetailNorStatisticsResponseCarriesTheIdentityKey` asserts the literal recomputed key value absent from both live responses with all four channels populated.
- No transaction is held anywhere in `CryptographicAssetServiceImpl`; both OPA calls run outside one. Read-only endpoints, so nothing to race and no external acknowledgment to diverge from.
- Nothing leaks to the wire: the aggregation `IllegalStateException` and `dense()`'s unmapped-key message (which does embed a DB-derived group key) both land on the advice's fixed `"Internal server error."`. The group key reaches the log only, and it is an asset type / verdict / family code.
- The virtual-thread fan-out matches an established pattern in this repo (`CertificateServiceImpl:2147` submits 9+ tasks through the same `DelegatingSecurityContextExecutorService`), so 7 tasks against a 40-thread prod pool is not a new risk. The null-Pageable/null-order idiom is real (`SecurityFilterRepositoryImpl.window` 217-222 no-ops on null). Detail's `occurrenceCount` and `sourceCbomCount` use the same expressions as the list's `findListRowsByUuids`, so the reconciliation criterion is structurally sound and not merely test-asserted. `verdict.decidedAt` cannot be null when `pqcEvaluatedAt` is set — `applyPqcVerdict` COALESCEs it in the same statement. The `cbom` association is `insertable = false, updatable = false`, so the fetch join is safe. Nulling `sourceCbomCount` / `syncCompleteness` is contract-legal; neither is marked required.
- No new `@Value` or env-var override, so no Helm-chart follow-up. `MAX_SERVED_SOURCES` / `TOP_ALGORITHM_FAMILIES` read as response-shape constants beside the existing `MAX_ITEMS_PER_PAGE`, not operator tunables. Both controller methods already carry audit logging. `FilterField` untouched. Every issue reference is repo-prefixed.

**Worth your call, not findings**

- `CryptoAssetSourceRepository.findWithCbomByAssetUuid` loads every source row with its Cbom fetch-joined, unbounded; `MAX_SERVED_SOURCES = 100` caps only what is served. The sum at line 499 needs the full set, so the read is load-bearing — but an asset contributed by very many documents materialises every row plus its header on the heap. Could `occurrenceCount` come from a SUM projection so the row load can be bounded too? No demonstrated failure, just a question.

**What could not be checked**

- The head does not compile against the current `interfaces` snapshot on its own merge-base: `SigningProfileControllerImpl` and `TokenInstanceControllerImpl` do not implement `listContentSigningFormattingConnectorAttributes` / `listTokenAttributes`, which main adapted to in `73cd5366` / `9cf58de4`. Not a defect in this diff — it needs a rebase. Everything above was run on a local merge of `origin/main` into the head (clean, no conflicts), which is what CI tests anyway.
- The full test run was not executed; roughly four CI shards beyond the architecture suite and the covering classes were skipped. Nothing in the diff points at them, but I cannot claim the whole suite is green.
- No spec or plan document in this repo mentions core#2145 (no `docs/` tree, no markdown hit), so if one lives elsewhere the reconcile-the-spec step is unverified from here.
- Whether an independent review pass iterated to convergence — the PR body's hand-trace evidences a pass, not the loop.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -176,7 +241,89 @@ public List<SearchFieldDataByGroupDto> getSearchableFieldInformationByGroup() {
     @Override
     @ExternalAuthorization(resource = Resource.CRYPTO_ASSET, action = ResourceAction.LIST)
     public CryptographicAssetStatisticsDto getCryptographicAssetStatistics() {
-        throw new NotSupportedException(NOT_IMPLEMENTED);
+        // No SecurityFilter parameter reaches this method (the contract takes none), so both scopes are built
+        // by hand: assets under the caller's CRYPTO_ASSET LIST scope -- which is what makes these numbers
+        // reconcile with the list -- and the document-level completeness block under the CBOM LIST scope.
+        SecurityFilter assetFilter = SecurityFilter.create();
+        objectFilterAspect.populateSecurityFilter(Resource.CRYPTO_ASSET, ResourceAction.LIST, null, null, assetFilter);
```
vyskocilm: 🟡 risk: this method already carries `@ExternalAuthorization(CRYPTO_ASSET, LIST)`, and the filter aspect populates a `SecurityFilter` argument from that annotation automatically — `ObjectFilterAspect:57-89`, with its Javadoc at 91-94 saying to prefer the annotation and reach for the public helper only when the filter must be built outside an authorized method.

The comment's premise doesn't hold: the *HTTP* contract takes no filter, but `CryptographicAssetExternalService` is core-owned and free to. `SigningRecordExternalService.getSigningRecordStatistics(period, SecurityFilter)` is the in-repo precedent for exactly this shape — no filter on the HTTP surface, one on the service signature, `SecurityFilter.create()` passed from the controller.

As written, deleting line 248 makes the endpoint serve unscoped counts with no fence anywhere to catch it — and unscoped counts silently stop reconciling with the list, which is the one property the statistics AC turns on. Suggest adding a `SecurityFilter` parameter to the service method, passing `SecurityFilter.create()` from `StatisticsControllerImpl`, and dropping the manual asset-filter block. Keep the manual CBOM call below it — a cross-resource scope genuinely has to be hand-built, same as `cbomSerialNumbersScopedToCaller`.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
```diff
@@ -310,15 +480,171 @@ private static CryptographicAssetDto toDto(CryptoAssetListRow row) {
...
+        if (stored == null) {
+            return null;
+        }
+        return stored.stream().map(CryptographicAssetServiceImpl::toEvidenceDto).toList();
+    }
+
+    /** Stored evidence is the capped producer occurrence array; the keys are CycloneDX occurrence field names. */
+    private static CryptographicAssetEvidenceDto toEvidenceDto(Map<String, Object> occurrence) {
+        CryptographicAssetEvidenceDto dto = new CryptographicAssetEvidenceDto();
+        dto.setLocation(asServedText(occurrence.get("location")));
```
vyskocilm: 🟡 risk: same required-field shape as the two above. `asServedText` returns `null` for anything that is neither String nor Number, and `CryptographicAssetEvidenceDto.location` is `requiredMode = REQUIRED`. `OccurrenceEvidenceCapper.withoutAdditionalContext` copies the producer's own keys through, and `sanitizeLocation` only fires when a `location` key is actually present — an occurrence lacking the key stores no location at all, and this then serves `null` on a required field.

Lower confidence than the other two only because nothing in production calls `upsertSource` yet (sync lands with core#2073), so I can't point at a live row. But the deliberate design of `asServedText` — read whatever shape is in the column, independent of what today's writer produces, as the javadoc on `evidenceFieldsCoerceNumericStringsAndNumbers` says — is exactly what keeps this reachable once a real producer writes.

Either default to `""` the way `sanitizeLocation` already does for a non-string value, or fold this field into the `interfaces` requiredness decision.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Ordering accepted: interfaces#940 (one-line flip + absence semantics, closes interfaces#938) is open and this PR should not merge before it lands and the snapshot is consumed here — could you review #940? It is exactly the change your first option names. Serving behavior stays as is per the agreed rationale. The branch is now rebased onto main (c6116c99), so it compiles against the live snapshot.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Agreed one ruling should cover both copies — that exact framing (relax vs. ingest placeholder, covering CryptographicAssetSourceDto.source AND CbomDto.source, plus CryptographicAssetDto.name for the refuted-OID case) is staged on interfaces#938 as a comment for the contract owners. Serving side deliberately unchanged until that ruling; the null path stays unasserted on purpose so we don't pin a shape mid-decision.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Done in c6116c99, exactly as you laid out: getCryptographicAssetStatistics(SecurityFilter) on the core-owned interface, StatisticsControllerImpl passes SecurityFilter.create() mirroring the signingRecords line, the manual asset-scope block is gone (the aspect populates the parameter), and the hand-built CBOM half stays with its comment corrected. The reconciliation ITest (absolute totals under an OPA restriction) still proves the aspect mechanism end to end.

src/main/java/com/otilm/core/service/impl/CryptographicAssetServiceImpl.java
karolhrdina: Done in c6116c99: location now serves "" whenever the coerced value is null — absent key and uncoercible type collapse to the same substitution — mirroring sanitizeLocation's own non-string fallback, so the read side never omits the required field. symbol stays null-omitted (optional). Test covers a producer occurrence with no location key at all.

vyskocilm (approved):

PR #2189, opened 2026-09-01
Connector-origin auth and server errors surface as 502

### LINKED ISSUE #2152: Connectors - Attribute callbacks: v2 connector problem+json 401/403 propagated verbatim to API clients
### Description

Core returns a v2 connector's problem+json HTTP status verbatim to its own API clients. `ExceptionHandlingAdvice.handleConnectorProblemException` responds with `ex.getProblemDetail().getStatus()` and no translation, so a connector-origin 401/403 (e.g. `CREDENTIAL_INVALID` when the otpki connector's upstream credentials are rejected) is indistinguishable from Core's own authentication/authorization failures. The frontend keys logout purely on status, so a connector misconfig terminates the operator's session; a connector 403 shows up as a caller permission problem. The machine-readable `errorCode` is also lost — the body is downgraded to the legacy `{"message": ...}` shape.

The same connector condition currently yields five different Core statuses depending on transport and content type: 401/403 verbatim (v2 problem+json), 400 (v1 legacy body), 400 without connector attribution (AMQP proxy), 502 (problem+json with an `errorCode` outside the `ErrorCode` enum), 503 (text/html body).

Expected direction: Core never propagates connector-origin auth statuses verbatim; 401/403 on Core's API are reserved for the caller's own session/permissions, and connector auth failures surface as an upstream fault (e.g. 502) with connector attribution and the original `errorCode` preserved. `DiscoveryConnectorErrors` already maps `UNAUTHORIZED`/`FORBIDDEN`/`CREDENTIAL_INVALID` to "the connector refused Core's credentials" for discovery ticks — the same classification, generalized. `handleConnectorProblemException` currently has no unit test pinning either behavior.

Related: #1990 (transport-level classification), OmniTrustILM/interfaces#783 (`errorCode` enum robustness — its "no Core change needed" note predates this finding), #1947 (connector error message rendering). Note: this also touches `interfaces` (`BaseApiClient`) and `fe-administrator` (status-keyed logout) for related client and frontend work.

### Steps to Reproduce

1. Configure an otpki authority whose connector-side credentials against OTPKI are wrong (or lack permission to list end entity / CA profiles)
2. As any logged-in user, open a form that triggers a connector attribute callback (e.g. RA profile creation, end entity profile select)
3. The connector answers 401 `application/problem+json` with `errorCode: CREDENTIAL_INVALID`
4. Observe the response of Core's callback endpoint and the FE behavior

### Expected Behavior

Core reports a connector/upstream fault (e.g. HTTP 502) with the connector named and the original `errorCode` preserved; the operator stays logged in and sees an actionable connector error.

### Actual Behavior

Core returns HTTP 401 (or 403) verbatim with body `{"message": "... Error is related to connector 'otpki' (<uuid>)."}`. The FE redirects to the login page on 401; on 403 it shows a permission toast implying the caller lacks permissions.

### Severity

Major

### Module

Connectors

lubomirw:
## TLDR

Fixes #2152 - a v2 connector's problem+json status was returned verbatim by Core, so a connector with rejected upstream credentials surfaced as Core's own 401: clients read it as an expired session and the FE logged the operator out; a connector 403 read as the caller lacking permissions. Connector-origin 401/403/5xx (and sub-400 oddities) now surface as **502** with the original status in the message; 404/422 keep their entity and validation semantics.

## Behavioral changes (release notes - 2.20.0)

| Connector-origin condition | Before | After |
|---|---|---|
| problem+json 401 / 403 | 401 / 403 (FE logout / permission toast) | **502**, message ends "Original response code NNN." |
| problem+json 5xx | verbatim (connector 500 looked like a Core bug) | **502** |
| problem+json sub-400 (e.g. 303) | verbatim status with error body | **502** |
| legacy-body 401 / 403 | 400 | **502** |
| problem+json 404 / 422 / other 4xx | verbatim | unchanged |
| status suffix format (client and server legacy handlers) | `Original response code 401 UNAUTHORIZED.` | numeric only on all three suffix sites: `Original response code 401.` |
| logging (problem+json path) | every connector problem at ERROR | all 4xx at WARN (translated or verbatim - this is the alert-noise fix), 5xx at ERROR |
| logging (legacy path) | every row at INFO | translated 401/403 at WARN; untranslated rows keep INFO |

Anyone scripting against the old statuses should read this table; the original code stays available in the message text.

## Design context

Stage 1 of the connector error model umbrella (verbatim status propagation conflated connector-upstream auth with the caller's session - the same condition previously yielded five different Core statuses depending on transport and content type). Stage 2 (problem+json on Core's own API, machine-readable errorCode preservation) is deliberately out of scope here and follows in 2.21+/3.0 - body shapes are unchanged in this PR. The MQ/proxy path benefits too: proxied connector 401s arrive as ConnectorClientException and translate identically.

## Verification

- 17 unit pins in ExceptionHandlingAdviceConnectorTest (translated rows, verbatim rows incl. 400/499 boundaries, null-status row, numeric-suffix pin on both paths, deliberately-unchanged neighbor handlers) + MockMvc callback integration pin (connector 401 problem on /v2/connectors/{uuid}/callback -> 502). The 599 case also pins the removal of the former HttpStatus.valueOf crash.
- Full suite: 6478 tests, 0 failures.
- Four-reviewer pass (Copilot CLI, guided, superpowers, hygiene): all surviving findings applied (sub-400 predicate, log-level split, suffix normalization, boundary pins); Copilot's errorCode-preservation finding is deliberate stage-2 scope.
- Sonar coverage: advice class is under the api/** coverage exclusion; tests exist regardless.


src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
```diff
@@ -335,21 +346,30 @@ public ErrorMessageDto handleConnectorCommunicationException(ConnectorCommunicat
...
+            messageBuilder.append(" Original response code ").append(originalStatus).append(".");
         }
 
-        ErrorMessageDto errorMessageDto = ErrorMessageDto.getInstance(errorMessage);
-        return ResponseEntity.status(ex.getProblemDetail().getStatus()).body(errorMessageDto);
+        if (originalStatus >= 500) {
+            LOG.error("HTTP {}: {} {}", responseStatus, messageBuilder, ex.getProblemDetail());
+        } else {
+            LOG.warn("HTTP {}: {} {}", responseStatus, messageBuilder, ex.getProblemDetail());
+        }
```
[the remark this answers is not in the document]
lubomirw: Intentional, though the PR body's logging row was genuinely misleading here - fixed it. The spec (stage 1 of the connector error model) moves ALL connector-origin problem 4xx from ERROR to WARN, translated or not: before this PR every one of these rows logged at ERROR, which is exactly the alert noise the change removes (a routine connector 422 firing error alerts). So the verbatim 404/422/499 rows at WARN are a deliberate step down from ERROR, not an accidental step up from INFO. The 'unchanged rows keep INFO' note applies to the legacy ConnectorClientException path only, whose untranslated rows really were INFO before and stay INFO. Body table now states the two paths separately.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
```diff
@@ -269,11 +268,23 @@ public ErrorMessageDto handleConnectorClientException(ConnectorClientException e
         }
 
         if (ex.getHttpStatus() != null) {
-            messageBuilder.append(" ").append("Original response code ").append(ex.getHttpStatus()).append(". ");
+            messageBuilder
+                    .append(" ")
+                    .append("Original response code ")
+                    .append(ex.getHttpStatus().value())
```
ivosh, line 274: **~ Minor** | Maintainability

This PR normalises the "Original response code" suffix to a bare number on the `ConnectorClientException` path (the added `.append(ex.getHttpStatus().value())` at line 274) and appends the same numeric form on the problem+json path, but leaves the third site that renders this suffix untouched: `src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java:314`, inside `handleConnectorServerException`, still appends `ex.getHttpStatus()` — and `HttpStatus.toString()` renders `500 INTERNAL_SERVER_ERROR`, not `500`.

Before this PR the two legacy handlers agreed on the format; after it they disagree, and which one an operator sees depends only on how the connector framed its error. `BaseApiClient` maps a legacy (non-problem+json) 5xx body to `ConnectorServerException` (interfaces `src/main/java/com/otilm/api/clients/BaseApiClient.java:763`) and a problem+json 5xx to `ConnectorProblemException`. So one and the same connector 500 now surfaces as a Core 502 whose message ends `Original response code 500 INTERNAL_SERVER_ERROR.` on the legacy route and `Original response code 500.` on the problem+json route.

Two places overstate the normalisation as a result. The release-note row "legacy status suffix format → numeric only" reads as if the whole legacy surface changed, when only the `ConnectorClientException` handler did; and the new pin `src/test/java/com/otilm/core/api/ExceptionHandlingAdviceConnectorTest.java:143` is named `bothPathsEmitNumericOnlyStatusSuffix` while a third path still emits the enum name (its sibling `connectorServerExceptionStaysBadGateway` pins only that handler's `@ResponseStatus`, not its message format).

_Suggested:_ Either apply the same `.value()` normalisation in `handleConnectorServerException`, or narrow the release-note row and the `bothPathsEmitNumericOnlyStatusSuffix` test name to say that only the ConnectorClientException path changed.

<sub>flagged by Reference integrity + Correctness & races</sub>

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
```diff
@@ -269,11 +268,23 @@ public ErrorMessageDto handleConnectorClientException(ConnectorClientException e
...
+                    .append(". ");
         }
 
-        LOG.info("HTTP 400: {}", messageBuilder);
-        return ErrorMessageDto.getInstance(messageBuilder.toString());
+        boolean authOrigin = ex.getHttpStatus() == HttpStatus.UNAUTHORIZED
```
ivosh, line 278: **. Nitpick** | Error handling

The new 401/403 → 502 translation in `handleConnectorClientException` (ExceptionHandlingAdvice.java:278) leaves the message text as the only place the origin is recorded, and that message's connector segment is built from `ex.getConnector()`. On the MQ/proxy path the PR description calls out as benefiting from this change, that field is never set for the auth categories: `ProxyClientImpl.java:364` builds `new ConnectorClientException(errorMessage, HttpStatus.UNAUTHORIZED)` and `ProxyClientImpl.java:366` the FORBIDDEN twin, both without a connector, while sibling branches of the same switch do pass it (`timeout`/`connection` at ProxyClientImpl.java:370, `default` at ProxyClientImpl.java:377). So a proxied connector 401 reaches the operator as a 502 reading `... Original response code 401.` with no `Error is related to connector name=…, uuid=….` segment — the one detail the 502 exists to make identifiable is missing on exactly that transport. Pre-existing, but this PR is what makes the message the sole carrier of origin.

_Suggested:_ Switch the authentication and authorization branches of `ProxyClientImpl.throwProxyError` to the `ConnectorClientException(message, httpStatus, connector)` constructor so the translated 502 names the connector.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
```diff
@@ -335,21 +346,30 @@ public ErrorMessageDto handleConnectorCommunicationException(ConnectorCommunicat
...
+        int responseStatus = translated ? HttpStatus.BAD_GATEWAY.value() : originalStatus;
+
+        StringBuilder messageBuilder = new StringBuilder(ex.getFullMessage(false));
+        if (translated) {
+            messageBuilder.append(" Original response code ").append(originalStatus).append(".");
         }
 
-        ErrorMessageDto errorMessageDto = ErrorMessageDto.getInstance(errorMessage);
-        return ResponseEntity.status(ex.getProblemDetail().getStatus()).body(errorMessageDto);
+        if (originalStatus >= 500) {
```
ivosh, line 373: **~ Minor** | Testing

This PR adds two new log-level branches — `ExceptionHandlingAdvice.java:367-371` (problem+json: 5xx at ERROR, everything else at WARN) and `ExceptionHandlingAdvice.java:282-286` (legacy: translated 401/403 at WARN, the rest at INFO). The PR description lists both as behavioural changes in the 2.20.0 release-note table and calls the WARN downgrade "the alert-noise fix", i.e. the operational point of the change. Neither new test class asserts a level: `ExceptionHandlingAdviceConnectorTest.java` asserts only statuses and message text, and `CallbackControllerErrorTest.java` only the HTTP status and body. A future edit that collapses the branch back to a single `LOG.error` would keep all 17 pins green while re-creating the alert noise. The pattern to reuse already exists for this exact class: `ExceptionHandlingAdviceTest.java` (same package, same class under test) has `captureLogsOfAdvice()`, a logback `ListAppender` helper, and two tests that filter `logged.list` by `event.getLevel()`; three other test classes in the tree use `ListAppender` the same way.

_Suggested:_ Add two ListAppender-backed pins modelled on ExceptionHandlingAdviceTest.captureLogsOfAdvice() — one that a connector 500 logs at ERROR while a connector 401 logs at WARN, one that a legacy 401 logs at WARN while a legacy 409 stays INFO.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
```diff
@@ -335,21 +346,30 @@ public ErrorMessageDto handleConnectorCommunicationException(ConnectorCommunicat
     }
 
     /**
-     * Handler for {@link ConnectorProblemException}.
-     *
-     * @return ResponseEntity with status code from the exception and body containing error message.
+     * Handler for {@link ConnectorProblemException}. Auth (401/403) and server (5xx) statuses from a connector are an
+     * upstream fault, never the caller's session or a Core bug — they surface as 502. Entity (404) and validation (422)
+     * semantics pass through verbatim.
```
ivosh: **~ Minor** | Documentation

The Javadoc added at `ExceptionHandlingAdvice.java:349-351` states the rule as "Auth (401/403) and server (5xx) statuses ... surface as 502. Entity (404) and validation (422) semantics pass through verbatim." The predicate it documents, at `ExceptionHandlingAdvice.java:358-359`, is `originalStatus &lt; 400 || originalStatus == 401 || originalStatus == 403 || originalStatus &gt;= 500`. Two facts are missing on both sides of the rule. First, every sub-400 status also translates to 502 — this PR's own `connectorProblem303BecomesBadGateway` pin and the inline comment at `ExceptionHandlingAdvice.java:356-357` say so, but the method Javadoc (the surface a caller sees in the IDE and in generated docs) does not, so the enumeration reads as exhaustive and is not. Second, the pass-through set is not "404 and 422" but every 4xx except 401/403 — this PR's own pins cover 400, 499 and 409 passing through verbatim. Naming only the two semantically interesting statuses invites a future reader to "fix" 400 or 409 into the translated set.

_Suggested:_ Restate the Javadoc as the predicate it documents — 401, 403, anything below 400, and anything 5xx become 502; every other 4xx passes through verbatim, notably 404 and 422 which keep their entity and validation semantics.

src/test/java/com/otilm/core/api/ExceptionHandlingAdviceConnectorTest.java
```diff
@@ -0,0 +1,170 @@
...
+    void legacyConnectorNullStatusStaysBadRequest() {
+        ResponseEntity<ErrorMessageDto> response = advice
+                .handleConnectorClientException(new ConnectorClientException("x", (HttpStatus) null));
+        assertEquals(HttpStatus.BAD_REQUEST, response.getStatusCode());
+        assertNotNull(response.getBody());
+        assertFalse(response.getBody().getMessage().contains("Original response code"));
+    }
+
+    @Test
+    void bothPathsEmitNumericOnlyStatusSuffix() {
```
ivosh: **. Nitpick** | Testing

`ExceptionHandlingAdviceConnectorTest.java:143-151` exercises two independent handlers in one method — the problem+json path at line 144 and the legacy path at lines 145-148 — with one assertion each, and the name ("bothPaths") announces it. Every other test in this class pins one behaviour, so a failure here does not say which path regressed. Two smaller points on the same method: it chains `.getBody().getMessage()` at lines 144 and 147-148 without the `assertNotNull(response.getBody())` guard that lines 48, 104, 113 and 138 all use, so a null body surfaces as an NPE rather than a named assertion failure; and it pins the numeric-only suffix only on translated rows, while the legacy path also appends the suffix on untranslated rows (`ExceptionHandlingAdvice.java:271-275` runs for any non-null status), which the release-note table lists as a changed format and `legacyConnector409StaysBadRequest` at line 126 does not assert.

_Suggested:_ Split it into problemPathEmitsNumericOnlyStatusSuffix and legacyPathEmitsNumericOnlyStatusSuffix, add the assertNotNull(getBody()) guard the sibling tests use, and extend the legacy one to a pass-through status such as 409.

src/test/java/com/otilm/core/api/ExceptionHandlingAdviceConnectorTest.java
```diff
@@ -0,0 +1,170 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Pins the connector-origin status translation: auth and server statuses from a connector are an upstream fault (502),
```
ivosh: **. Nitpick** | Maintainability

ExceptionHandlingAdviceConnectorTest.java:18-20. The first two sentences of the class Javadoc — "auth and server statuses from a connector are an upstream fault (502), while entity (404) and validation (422) semantics pass through — on both the problem+json and the legacy path" — say exactly what the 15 method names underneath already say (`connectorProblem401BecomesBadGateway`, `connectorProblem404StaysVerbatim`, `connectorProblem422StaysVerbatim`, `legacyConnector401BecomesBadGateway`, `legacyConnector409StaysBadRequest`, …). The names plus the assertion messages are the label, so a reader learns nothing new from the preamble. The one clause that isn't already in a name is the rationale in the third sentence: why the three deliberately-unchanged neighbour handlers (`connectorServerExceptionStaysBadGateway`, `connectorCommunicationExceptionStaysServiceUnavailable`, `connectorEntityNotFoundStaysNotFound`) are pinned at all. The direct sibling in this package, ExceptionHandlingAdviceTest.java:30, carries no class Javadoc.

_Suggested:_ Trim the Javadoc to the neighbour-handler rationale — the one thing the method names cannot carry — and drop the two sentences that restate them.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
```diff
@@ -335,21 +346,30 @@ public ErrorMessageDto handleConnectorCommunicationException(ConnectorCommunicat
...
-                    .error("HTTP %d: %s %s"
-                            .formatted(ex.getHttpStatus().value(), errorMessage, ex.getProblemDetail().toString()));
+        int originalStatus = ex.getProblemDetail().getStatus();
+        // Sub-400 statuses reach this handler too: the client throws for ANY non-2xx problem+json response, so a
+        // connector's 3xx problem document must not surface as a bodyless-redirect-shaped Core response.
+        boolean translated = originalStatus < 400 || originalStatus == 401 || originalStatus == 403
+                || originalStatus >= 500;
+        int responseStatus = translated ? HttpStatus.BAD_GATEWAY.value() : originalStatus;
+
+        StringBuilder messageBuilder = new StringBuilder(ex.getFullMessage(false));
```
ivosh, line 368: **. Nitpick** | Security

This PR reshapes the connector-origin auth response but keeps the message text ungated, which is exactly the case CLAUDE.md's "Don't leak runtime details to the wire" section names ("upstream error detail that should not be exposed").

**Legacy path.** `ExceptionHandlingAdvice.java:287` returns `messageBuilder`, whose first segment is `ex.getMessage()` (`ExceptionHandlingAdvice.java:256`). For a `ConnectorClientException` raised by the client library, that message *is* the raw response body: `OmniTrustILM/interfaces` `BaseApiClient.java:752` builds it from `clientResponse.bodyToMono(String.class)` inside the `is4xxClientError()` branch. So a 401 from a connector's upstream IdP — an HTML sign-in page, a reverse-proxy error page, a JSON blob echoing the rejected principal or an internal hostname — is copied byte-for-byte into Core's 502 body.

**problem+json path.** `ExceptionHandlingAdvice.java:362` seeds the message from `ex.getFullMessage(false)`, which is the connector-authored `ProblemDetailExtended.detail` (or `title`) — free text Core never shapes.

The exception *type* is a controlled domain type, but the CLAUDE.md rule is about the text being shaped by us, and this text is not. The 401/403 rows are the worst case for it: upstream credential-rejection bodies are the ones most likely to carry principal names, internal URLs, or token fragments. This is pre-existing forwarding rather than something this PR introduced, and the description states "body shapes are unchanged in this PR" — but the description's disclosure is about shape and `errorCode` preservation, not about the raw-text passthrough, and this PR is the one deciding what a connector auth failure looks like on the wire.

A benign reading: connector responses are considered trusted internal infrastructure, in which case the passthrough is fine and only needs saying so.

_Suggested:_ Either gate the forwarded text on the 401/403 rows — emit a fixed "The connector rejected the platform's upstream credentials." plus the numeric status, keeping the connector's own text in the WARN log only — or add a one-line comment on `handleConnectorProblemException` stating that connector-supplied `detail` is treated as trusted and deliberately passed to the caller.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
```diff
@@ -335,21 +346,30 @@ public ErrorMessageDto handleConnectorCommunicationException(ConnectorCommunicat
...
-        if (LOG.isErrorEnabled()) {
-            LOG
-                    .error("HTTP %d: %s %s"
-                            .formatted(ex.getHttpStatus().value(), errorMessage, ex.getProblemDetail().toString()));
+        int originalStatus = ex.getProblemDetail().getStatus();
+        // Sub-400 statuses reach this handler too: the client throws for ANY non-2xx problem+json response, so a
+        // connector's 3xx problem document must not surface as a bodyless-redirect-shaped Core response.
+        boolean translated = originalStatus < 400 || originalStatus == 401 || originalStatus == 403
+                || originalStatus >= 500;
+        int responseStatus = translated ? HttpStatus.BAD_GATEWAY.value() : originalStatus;
```
ivosh, line 364: **~ Minor** | Documentation

`ExceptionHandlingAdvice.java:360` (problem+json path) and `:280` (legacy `ConnectorClientException` path) now return **502** for connector-origin 401/403/5xx/sub-400. `ConnectorProblemException` is raised from the shared `BaseApiClient.handleProblemDetailResponse` (interfaces `src/main/java/com/otilm/api/clients/BaseApiClient.java:777`), so it can surface on *any* controller that calls a connector — not only the ones documenting connector errors.

Sweeping the interfaces repo, 20 server controllers that declare `throws ConnectorException` also declare `@ApiResponse(responseCode = "502", description = "Connector Error", content = ErrorMessageDto)` at class level (`CallbackController`, `ConnectorController`, `CertificateController`, …). Four do not, and inherit only the 400/401/403/500 set from `AuthProtectedController` (`interfaces/src/main/java/com/otilm/api/interfaces/AuthProtectedController.java:42-48`):

- `core/web/CryptographicKeyController.java` (13 `ConnectorException` references)
- `core/web/CryptographicOperationsController.java` (9)
- `core/web/TokenInstanceController.java` (10)
- `core/web/SigningProfileController.java` (5)

On those four, a connector 401 previously surfaced as 401 — a status their contract *does* document. After this PR it surfaces as 502, which their contract does not document at all, so a client generated from the spec sees an undeclared status on exactly the failure this PR is meant to make legible. (The 502 gap on those four was already reachable via `ConnectorServerException`; this PR widens it from an edge case to the primary connector-error status.)

_Suggested:_ Add the class-level `@ApiResponse(responseCode = "502", … ErrorMessageDto)` block that the other 20 connector-facing controllers already carry to CryptographicKeyController, CryptographicOperationsController, TokenInstanceController and SigningProfileController in the interfaces repo.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
```diff
@@ -335,21 +346,30 @@ public ErrorMessageDto handleConnectorCommunicationException(ConnectorCommunicat
...
     public ResponseEntity<ErrorMessageDto> handleConnectorProblemException(ConnectorProblemException ex) {
-        String errorMessage = ex.getFullMessage(false);
-        if (LOG.isErrorEnabled()) {
-            LOG
-                    .error("HTTP %d: %s %s"
-                            .formatted(ex.getHttpStatus().value(), errorMessage, ex.getProblemDetail().toString()));
+        int originalStatus = ex.getProblemDetail().getStatus();
+        // Sub-400 statuses reach this handler too: the client throws for ANY non-2xx problem+json response, so a
+        // connector's 3xx problem document must not surface as a bodyless-redirect-shaped Core response.
+        boolean translated = originalStatus < 400 || originalStatus == 401 || originalStatus == 403
```
ivosh: **~ Minor** | Correctness

The new `translated` predicate at `ExceptionHandlingAdvice.java:358` deliberately excludes 422, and the new Javadoc at `:350-351` states that "validation (422) semantics pass through verbatim". The status passes through, but the body shape does not match what the published contract promises for 422.

Every one of the 146 `responseCode = "422"` declarations under `interfaces/src/main/java/com/otilm/api/interfaces/core/web/` types 422 as an array of strings — `@Content(array = @ArraySchema(schema = @Schema(implementation = String.class)), examples = @ExampleObject("[\"Error Message 1\",\"Error Message 2\"]"))` (e.g. `CallbackController.java:29-31`). `handleValidationException` (`ExceptionHandlingAdvice.java:241`) honours that: it returns `List&lt;String&gt;`. The connector problem+json path returns `ErrorMessageDto` — `{"message": …}` — at line 372.

The two paths diverge for the same condition: `BaseApiClient` maps a *legacy* connector 422 to `ValidationException` (→ `String[]`, contract-correct), but a *problem+json* connector 422 to `ConnectorProblemException` (→ `ErrorMessageDto`, contract-violating). A client generated from the spec will fail to deserialize the second. This predates the PR, but the PR's own Javadoc now asserts the semantics are preserved, so the claim and the wire disagree.

_Suggested:_ Either map a connector-origin problem+json 422 to `ValidationException` so the body matches the documented `String[]` schema, or add an explicit second 422 content shape for `ErrorMessageDto` to the affected controller contracts and soften the new Javadoc's "pass through verbatim" wording to say only the status is preserved.

ivosh: **. Nitpick** | Maintainability

`ExceptionHandlingAdvice.java:358-359` expresses the same auth/server rule as bare numbers — `originalStatus &lt; 400 || originalStatus == 401 || originalStatus == 403 || originalStatus &gt;= 500` — whereas `handleConnectorClientException`, rewritten a few lines up in this same PR, expresses the identical rule with named constants: `ex.getHttpStatus() == HttpStatus.UNAUTHORIZED || ex.getHttpStatus() == HttpStatus.FORBIDDEN` (`ExceptionHandlingAdvice.java:278-279`). Two spellings of one policy in one class make it easy for a later edit to update one branch and miss the other, and the intent of `&lt; 400` and `&gt;= 500` is carried only by the comment above it.

Working from the raw `int` is the right call here — `HttpStatus.valueOf` on a non-standard code such as 599 is exactly the crash this PR removes — so the fix is only to name the numbers, not to change the type: compare against `HttpStatus.UNAUTHORIZED.value()` / `HttpStatus.FORBIDDEN.value()`, and express the ranges via `HttpStatus.Series.resolve(originalStatus)` or named private constants. *Clean Code* (Martin), "replace magic numbers with named constants".

_Suggested:_ Rewrite the predicate at ExceptionHandlingAdvice.java:358-359 in terms of `HttpStatus.UNAUTHORIZED.value()`, `HttpStatus.FORBIDDEN.value()` and `HttpStatus.Series` (or named constants) so both connector handlers spell the same policy the same way.

ivosh (changes requested):

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
lubomirw: Normalized the server handler too - all three suffix sites emit the numeric form now (7a54da657), and the release-note row says so.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
lubomirw: Done - both auth branches construct the exception, set the connector, then throw, so the translated 502 names the connector on MQ like the other transports.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
lubomirw: Added two ListAppender pins following ExceptionHandlingAdviceTest.captureLogsOfAdvice(): problem 500 at ERROR vs 401 at WARN, and legacy 401 at WARN vs 409 at INFO.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
lubomirw: Restated as the predicate: 401, 403, sub-400 and 5xx become 502; every other 4xx passes through, with 404/422 called out for their semantics.

src/test/java/com/otilm/core/api/ExceptionHandlingAdviceConnectorTest.java
lubomirw: Split per path with the getBody() guards, and the legacy test now pins the untranslated 409 row's suffix (present and numeric).

src/test/java/com/otilm/core/api/ExceptionHandlingAdviceConnectorTest.java
lubomirw: Trimmed to the neighbour-handler rationale - the method names carry the rest.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
lubomirw: Kept the passthrough and added the trust comment. The platform error contract makes connector-authored problem detail the operator-facing channel, with the connector responsible for its safety (bounded, no secrets, no echo on internal faults - SDK-renderer enforcement is part of the 2.21 catalog work). The legacy raw-body case you describe disappears with the shared classifier in the same release: unparseable bodies get a synthesized, controlled detail and the raw text stays in logs.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
lubomirw: Valid, and it belongs in the interfaces repo - opened OmniTrustILM/interfaces#949 adding the class-level 502/503 block to those four controllers.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
lubomirw: Real, and it predates this PR - the shared classifier (2.21) maps a problem-422 to ConnectorValidationException, which the advice renders as the documented String[] shape. Softened the javadoc here to claim only status preservation for 422 in the meantime.

lubomirw: Rewritten with HttpStatus.UNAUTHORIZED.value()/FORBIDDEN.value() and HttpStatus.Series - both handlers spell the policy the same way now.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
```diff
@@ -335,21 +350,34 @@ public ErrorMessageDto handleConnectorCommunicationException(ConnectorCommunicat
     }
 
     /**
-     * Handler for {@link ConnectorProblemException}.
-     *
-     * @return ResponseEntity with status code from the exception and body containing error message.
+     * Handler for {@link ConnectorProblemException}. The rule mirrors the predicate below: 401, 403, anything below
+     * 400, and anything 5xx surface as 502 — an upstream fault, never the caller's session or a Core bug. Every other
+     * 4xx passes through verbatim, notably 404 (entity) and 422 (validation) — though for 422 only the status is
+     * preserved; aligning the body with the documented String[] validation shape lands with the shared connector-error
```
ivosh: **. Nitpick** | Documentation

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java:356 says the body alignment "lands with" another classifier, recording planned sequencing rather than the current contract of handleConnectorProblemException. That planning detail can become stale; the preceding statement that only the 422 status is preserved already documents the relevant behavior.

_Suggested:_ Trim the Javadoc after "only the status is preserved" so it documents current behavior without the future-work reference.

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
ivosh: **Re-check: partial.**

ExceptionHandlingAdviceConnectorTest adds problemPathLogsServerErrorsAtErrorAndAuthAtWarn and legacyPathLogsTranslatedAtWarnAndUntranslatedAtInfo, but each test only asserts that both expected levels occur somewhere after two handler calls and does not associate each event with its originating status.

The tests catch collapse to a single log level, but would still pass if the two status-to-level mappings were reversed.

ivosh (dismissed):

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
lubomirw: Trimmed in 1b79740f5 - the javadoc now ends at the status-preservation statement, and the same pass dropped two more prose bits without informational value (a 'mirrors the predicate below' meta-phrase and an issue reference in the callback test's javadoc).

src/main/java/com/otilm/core/api/ExceptionHandlingAdvice.java
lubomirw: Right - the set was pinned but not the mapping. Each event is now asserted by index against its originating call, so swapping the two mappings fails both tests.

ivosh (approved):

Issue #934, opened 2025-01-27
Link public keys with existing certificates in inventory


lubomirw:
Currently when certificate is added to inventory (uploaded, discovered or issued through `CZERTAINLY` platform), it is linked with its key only when there exists such a key with same fingerprint in the inventory. It would be convenient to handle also other case, when public key of certificate is not present in the inventory by adding new key pair with only public key item.

It is necessary to implement following:
- add reference to key to `CertificateRequestEntity`
- make token instance reference optional in `CryptographicKey` entity
- Java migration to link certificate (and certificate request if present) public key with key from inventory or add new key with public key item and link it to certificate
- include key uuid to certificate request DTO

PR #2197, opened 2026-09-02
Order comment listings by a requested, tie-stable direction

### LINKED ISSUE #2182: Apply comment sort direction through Pageable with a stable uuid tie-break
### Description

`CommentServiceImpl` pages comments through two repository methods that bake the ordering into their names — `findByResourceAndObjectUuidAndParentUuidIsNullOrderByCreatedAtAsc` and `findByParentUuidOrderByCreatedAtAsc`. A derived-query `OrderBy` clause is emitted ahead of any `Sort` the `Pageable` carries, so the static term wins and a caller-supplied direction can never take effect. Drop the suffixes and let the service build the ordering:

```java
Sort.by(direction, "createdAt", "uuid")
```

The `uuid` term is a tie-break, not decoration. `createdAt` is stamped in `@PrePersist` from `OffsetDateTime.now()`, so comments written in the same instant tie, and under an unstable sort a tied row can appear on two pages or be skipped. The same tie-break is already the convention in the search stack — see `SortSpecification`. The defect exists in the current ascending-only code and is fixed here because the `ORDER BY` is being rewritten anyway.

The tie-break takes the requested direction rather than being pinned ascending. Pinned ascending, descending is not the reverse of ascending inside a tie group: both directions return the same first row, so a "latest comment" preview fetching one page descending shows the row the oldest-first listing shows. (This paragraph and the snippet above were corrected during review — the original text specified `.and(Sort.by(Sort.Direction.ASC, "uuid"))`, which has that defect.)

Worth being precise about what the tie-break buys: it totalises the order within a single snapshot. Offset paging remains unstable across requests under concurrent writes, and descending is the direction that exposes it, since new rows land at the front of the window. Keyset cursors would address that and are out of scope.

No new index on `created_at` — the existing `idx_comment_resource_object_uuid` does not serve the `ORDER BY`, but comment counts per object do not justify another index.

Depends on the `interfaces` change introducing `SortedPaginationRequestDto` — core needs an interfaces snapshot installed before it compiles.

Note: this also touches `interfaces` for the request DTO.

### Definition of Done

- Both repository methods take ordering from the `Pageable` instead of a method-name `OrderBy`
- `listComments` and `listReplies` honour `sortDirection`, with `uuid` appended in the same direction so descending is the exact reverse of ascending
- Tests cover asc and desc for thread roots and replies, that descending reverses ascending within a tie group, and that paging a tie group one row at a time matches the whole page
- A request omitting `sortDirection` still orders ascending, and so does a blank one
- Lowercase direction codes bind from the query string — query parameters convert through the conversion service, not Jackson, so `SortDirection` needs a registered converter

klaraf755:
> Depends on interfaces#947 (issue #934). Merge that first — core will not compile against a published snapshot that lacks `SortedPaginationRequestDto`.

## TLDR

Fixes #2182 — takes the sort direction from the request on both comment listings, and fixes two defects found on the way: the ordering was unstable when `created_at` ties, and lowercase direction codes were rejected outright.

## The ordering moves into the Pageable

Both derived queries carried their ordering in their method names — `findByResourceAndObjectUuidAndParentUuidIsNullOrderByCreatedAtAsc`, `findByParentUuidOrderByCreatedAtAsc`. A method-name `OrderBy` is emitted ahead of any `Sort` the `Pageable` carries (`AbstractQueryCreator.createQuery` does `complete(criteria, tree.getSort().and(dynamicSort))`), so the static term wins and a caller-supplied direction could never take effect. The suffixes come off and the service builds the ordering instead.

Named methods per direction were the alternative. The `Sort` becomes the same SQL `ORDER BY`, so it costs nothing at runtime, and it keeps two repository methods instead of four with the tie-break restated in every name. The methods now carry a javadoc line saying the caller owns the ordering, since the signature no longer shows it.

## The tie-break

`created_at` is stamped in `@PrePersist` from `OffsetDateTime.now()`, so comments written in the same instant tie. Under a tied sort a row can surface on two pages or on none, so the primary key follows `created_at` to make the order total. This defect was already present while the listings were ascending-only.

The tie-break takes the requested direction rather than being pinned ascending — otherwise descending is not the reverse of ascending inside a tie, and a "latest comment" preview fetching one page descending would show the same row the oldest-first listing shows.

Worth being precise about what this buys: it totalises the order within a single snapshot. Offset paging is still unstable across requests under concurrent writes, and descending is the direction that exposes it, since new rows land at the front of the window. Keyset cursors would be the fix for that, and are out of scope here.

## Lowercase codes were rejected

Query parameters bind through the conversion service, not Jackson, so `SortDirection`'s `@JsonCreator fromCode` never runs for them. Spring's default enum converter is `Enum.valueOf`, which wants `DESC` — while the published schema advertises `desc`. Verified before fixing: `?sortDirection=desc` returned 422. The converter is registered alongside the ones `KeyAlgorithm`, `OidCategory` and `SigningRecordStatisticsPeriod` already have. It is the first `SortDirection` binding from a string, so no existing caller relied on the uppercase form.

Blank is real too: `?sortDirection=` converts to null and overwrites the request object's own default, so the service defaults it explicitly.

## Out of scope

No index on `created_at` — the existing `idx_comment_resource_object_uuid` does not serve the `ORDER BY`, but comment counts per object do not justify another one.

Pagination bounds remain unenforced on these listings: `?pageNumber=0` yields `PageRequest.of(-1, …)` and a raw 400, `?itemsPerPage=1000000` materialises a million rows. `RequestValidatorHelper.revalidatePaginationRequestDto` only null-fills, and the gap is shared with the scheduler, approval and notification listings, so it belongs in its own change.

## Testing

`CommentSortOrderITest` covers both directions for roots and replies, that descending is the exact reverse of ascending within a tie group, and that paging a tie group one row at a time matches the whole page. `CommentSortDirectionBindingITest` drives the endpoint through MockMvc for the lowercase code, the omitted default, the blank value, and that an unknown code is the documented 422.

Rows are seeded with explicit `created_at` stamps rather than through `createComment`, so a `now()` collision cannot leave an assertion resolving on the tie-break instead of on the ordering under test.

152 tests green across the comment surface.


src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
```diff
@@ -279,6 +280,17 @@ private void changeResolution(UUID uuid, boolean resolved) throws NotFoundExcept
...
 
+    private static PageRequest pageByCreationTime(SortedPaginationRequestDto pagination) {
+        // The converter maps a blank sortDirection to null, which overwrites the request object's own default
+        SortDirection requested = Objects.requireNonNullElse(pagination.getSortDirection(), SortDirection.ASC);
+        Sort.Direction direction = Sort.Direction.fromString(requested.getCode());
+        // createdAt is stamped per request and can tie, so the primary key follows it to make the order total.
+        // It takes the requested direction too, or descending would not be the reverse of ascending within a tie.
+        return PageRequest
+                .of(pagination.getPageNumber() - 1, pagination.getItemsPerPage(),
+                        Sort.by(direction, "createdAt", "uuid"));
```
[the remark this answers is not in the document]
klaraf755: Reconciled on the requirement side: the issue was stale, not the code.

I wrote that DoD bullet before testing the tie-break, and it was wrong. Pinned ascending, `desc` is not the reverse of `asc` inside a tie group — both directions return the same first row, so a "latest comment" preview fetching one page descending shows exactly the row the oldest-first listing shows. Verified before changing it: with `.and(Sort.by(ASC, "uuid"))`, four roots sharing a `created_at` came back in byte-identical order for both directions.

`#2182` now specifies the shipped behaviour, and `ordersRootsSharingACreatedAtValueInReverseWhenTheDirectionIsDescending` pins it.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
```diff
@@ -279,6 +280,17 @@ private void changeResolution(UUID uuid, boolean resolved) throws NotFoundExcept
                 eventData, UUID.fromString(actor.getUuid()), null));
     }
 
+    private static PageRequest pageByCreationTime(SortedPaginationRequestDto pagination) {
+        // The converter maps a blank sortDirection to null, which overwrites the request object's own default
+        SortDirection requested = Objects.requireNonNullElse(pagination.getSortDirection(), SortDirection.ASC);
+        Sort.Direction direction = Sort.Direction.fromString(requested.getCode());
```
lubomirw: **[Nit]** | Maintainability | Confidence: High

`Sort.Direction.fromString(requested.getCode())` round-trips the enum to its code string and re-parses it; it works only because `getCode()` is constrained to asc/desc. The codebase already has the direct idiom in `SortOrderBuilder`.

_Suggested:_ `Sort.Direction direction = requested == SortDirection.DESC ? Sort.Direction.DESC : Sort.Direction.ASC;`

src/main/java/com/otilm/core/config/WebAppConfig.java
```diff
@@ -105,6 +106,9 @@ public SigningRecordStatisticsPeriod convert(String source) {
         registry
                 .addConverter(String.class, OidCategory.class,
                         source -> StringUtils.isBlank(source) ? null : OidCategory.fromCode(source));
+        registry
+                .addConverter(String.class, SortDirection.class,
+                        source -> StringUtils.isBlank(source) ? null : SortDirection.fromCode(source));
```
lubomirw: **[Low]** | Maintainability | Confidence: Medium

The blank→null mapping (mirroring the `OidCategory` converter) overwrites this DTO's own default `ASC` when a caller sends `?sortDirection=`. Safe today only because `pageByCreationTime` re-defaults null→ASC, but the declared `defaultValue="asc"` is not self-enforcing — a future endpoint reading `getSortDirection()` directly gets null on a blank param.

Unverified: whether any planned reuse reads the field without a null-guard; today's single caller is safe.

_Suggested:_ return `SortDirection.ASC` on blank instead of null, so the DTO default and blank-input behavior agree without each consumer repeating the guard.

lubomirw (commented):
Automated multi-reviewer triage — 2 findings (COMMENT). Both are maintainability nits; the code is correct and test-covered.

src/main/java/com/otilm/core/config/WebAppConfig.java
klaraf755: Fixed.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
klaraf755: Fixed.

lubomirw (approved):

PR #2200, opened 2026-09-02
Take comment notifications to the comment they are about

### LINKED ISSUE #2199: Internal comment notifications should reference the thread they are about
### Description

Clicking a comment notification should take the user to the comment it is about. Today it cannot: the notification stores its message text plus the host resource and host object UUID as its target, with `detail` deliberately null, so the recipient learns only that somebody commented on the object. Finding the comment means opening the object and expanding threads — the panel lists ten roots per page with replies collapsed and paged at twenty.

**What the notification carries.** A notification gains an optional *subject*: the object inside the target that it is about, for events whose subject is not the target itself. It is stored as one JSON column on the notification (the way audit records and compliance results are stored) and exposed flat on `NotificationDto` beside the existing target fields:

- `subjectObjectType` — `COMMENT` for comment events;
- `subjectObjectIdentification` — the comment the notification is about: the new root, the new reply, or the resolved thread;
- `subjectParentIdentification` — the object the subject is nested in: the thread root for a reply, absent for a root comment or a resolution.

Both values are already in the event payload, so this is a secondary reference rather than new data. The host target must stay as it is, since a comment has no page of its own and the deep link has to open a real page (see #2116). The comment body is still never persisted in the notification. Nothing is parsed out of `detail`.

**What the API needs for the redirect to work.** Carrying the references is not sufficient. The panel is a paged list — ten roots per page, replies paged at twenty under their root — so knowing which comment to show does not say *where* it is. A single-comment read would not help either: it would render the comment outside the list it belongs to, and the panel would still not know which page holds it.

Both listings therefore accept an anchor and return the page holding it:

- `GET /v1/comments/{resource}/{objectUuid}?anchorUuid={rootUuid}` returns the page of roots holding that thread; the anchor must be a root, since the caller tells a stale anchor by its absence from the page and a reply is never on a page of roots;
- `GET /v1/comments/{uuid}/replies?anchorUuid={replyUuid}` returns the page of replies holding that reply.

`CommentResponseDto` already extends the pagination response, so the returned `pageNumber` tells the panel where it landed and no new response shape is needed. The panel anchors the thread listing on the parent reference when present and on the subject otherwise, and the replies listing on the subject; both calls can be issued at once.

A stale anchor — the comment was deleted in the meantime, never belonged to this object, or is a reply passed to the thread listing — must not fail the object listing: that call can still serve the object's remaining comments, and failing it would take the whole panel down through the widget-lock path. It returns the requested page instead (the first, unless another was asked for), and the caller can tell what happened because the returned page does not contain the anchored UUID, so the panel reports that the comment no longer exists rather than landing the user silently on page one. The replies call is different: if the thread root itself was deleted, `GET /v1/comments/{uuid}/replies` addresses a comment that is gone and answers 404 as it does today.

Note: this also touches `interfaces` for the endpoint and notification contract and `fe-administrator` for the panel that consumes it (OmniTrustILM/fe-administrator#2014).

### User Story

As a user notified about a comment, I want clicking that notification to take me straight to the comment, so that I do not have to search the object's comment panel for it.

### Use Case

A reply lands on a thread I took part in, on an object that already carries a dozen threads. Today the notification tells me only that someone replied on that object; I open it and expand threads until I find the one that changed. With this, the notification opens the object, expands the thread and scrolls to the reply.

### Acceptance Criteria

- [ ] Comment notifications carry a subject alongside the host object target: the comment the notification is about, and for a reply the thread root it sits in.
- [ ] The subject is stored as a single JSON column and exposed on `NotificationDto` as `subjectObjectType`, `subjectObjectIdentification` and `subjectParentIdentification`; notifications for other events leave it absent.
- [ ] The host target is unchanged, so the deep link still opens a page that exists.
- [ ] The comment body is still never persisted in the notification.
- [ ] Both listings accept `anchorUuid` and return the page containing the named comment, reporting that page number; the thread listing accepts a thread root only, a reply anchor is ignored there.
- [ ] A stale or foreign anchor on the object listing returns the requested page rather than failing, so the panel still renders; the absent anchor in the returned page is what tells the caller the comment is gone. A deleted thread root still answers 404 on the replies call.
- [ ] Clicking a notification whose comment has since been deleted still opens the host object and its comment panel, and the panel shows an error saying the comment no longer exists — rather than silently landing the user on an unrelated page.
- [ ] End to end: clicking a comment notification opens the host object, expands the thread and brings the comment into view.

klaraf755:
Closes #2199

Clicking a comment notification should take the user to the comment. Today the persisted notification names only the host object, and the comment listings have no way to open the page holding a given comment (Epic: OmniTrustILM/ilm#246).

## What's here

- **Anchored listings** — both listings accept `anchorUuid` and compute the page index from a count of roots (or replies) sorted ahead of the anchor in the requested direction (#2197's `sortDirection`): created before it when ascending, after it when descending, with the listing's uuid tie-break. The count and the read are separate statements, so when the anchor is not on the page its position pointed at, they are repeated once before the requested page is served. The thread listing accepts a root only, because the caller tells a stale anchor by its absence from the page and a reply is never on a page of roots. A stale anchor, one from another object, or a reply on the thread listing leaves the caller on the requested page; the listing must keep serving the object's other comments. Tests pin the page arithmetic in both directions, tied timestamps across a page boundary, and the stale, foreign and reply-on-thread-listing cases.
- **Notification subject** — `Notification` gains one nullable jsonb `subject` column holding a `NotificationSubject` record (type, identification, parent identification), mapped the way `AuditLog.logRecord` is. `NotificationInternalService` gets subject-carrying overloads; the existing ones delegate with no subject, so nothing else changes. Comment events fill it with the comment and, for a reply, the thread root; the target stays the host object and the body still never reaches the notification. The listener unit test verifies the record for roots and replies; the handlers ITest asserts the persisted subject end to end.

## Notes for reviewers

- The subject is one JSON column rather than three, on the reasoning that it is a small typed value owned by the row and never queried, like the compliance results and audit records already stored that way.
- Companion contract: OmniTrustILM/interfaces#948 (merge first).


src/main/java/com/otilm/core/dao/repository/CommentRepository.java
```diff
@@ -24,6 +24,12 @@ Page<Comment> findByResourceAndObjectUuidAndParentUuidIsNullOrderByCreatedAtAsc(
 
     Page<Comment> findByParentUuidOrderByCreatedAtAsc(UUID parentUuid, Pageable pageable);
 
+    // Position of a comment within its listing, so a caller anchored at one can be given the page holding it.
+    long countByResourceAndObjectUuidAndParentUuidIsNullAndCreatedAtLessThan(Resource resource, UUID objectUuid,
+            OffsetDateTime createdAt);
+
+    long countByParentUuidAndCreatedAtLessThan(UUID parentUuid, OffsetDateTime createdAt);
```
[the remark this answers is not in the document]
klaraf755: Fair. Both listings now order by createdAt then uuid, and the position counts use the same tie-break, evaluated in the database, so a tied anchor lands on the page the database actually sorts it onto. Tied timestamps across a page boundary are covered for roots and replies.

src/main/java/com/otilm/core/dao/entity/notifications/Notification.java
```diff
@@ -57,6 +67,11 @@ public NotificationDto mapToDto() {
         dto.setDetail(this.detail);
         dto.setSentAt(this.sentAt);
         dto.setTargetObjectType(this.targetObjectType);
+        if (this.subject != null) {
+            dto.setSubjectObjectType(this.subject.type());
```
ivosh: **! Major** | Correctness

NotificationRecipientRepository.java:17 and :27 still construct NotificationDto through the legacy seven-argument projection, which does not include any of the three new subject fields. Although Notification.java:71-73 maps the persisted subject when a notification is created, listNotifications uses these repository projections, so the actual notification-list response omits the comment UUID and thread-root UUID needed for the deep link; this PR's primary user flow therefore cannot consume the persisted subject.

_Suggested:_ Update both notification-list projections, and the DTO projection constructor if needed, to populate all three subject fields from Notification.subject.

<sub>flagged by Persistence & schema + OpenAPI & schema fidelity</sub>

src/test/java/com/otilm/core/integration/service/CommentServiceValidationITest.java
```diff
@@ -52,12 +59,135 @@ private CommentDto post(UUID objectUuid, String body, UUID parentUuid) throws No
...
+            CommentResponseDto anchored = commentService.listReplies(root.getUuid(), anchor.getUuid(), pagination);
+            assertThat(anchored.getComments()).extracting(CommentDto::getUuid).contains(anchor.getUuid());
+        }
+    }
+
+    @Test
+    void aStaleAnchorLeavesTheListingOnTheRequestedPage() throws NotFoundException {
+        post(raProfileUuid, "still here", null);
+
+        CommentResponseDto anchored = listAnchoredAt(raProfileUuid, UUID.randomUUID(), 10);
```
ivosh: **~ Minor** | Testing

CommentServiceValidationITest.java:152-171 says stale and foreign anchors leave the listing on the requested page, but `listAnchoredAt` never sets `pageNumber`, so both cases use the default first page. An implementation that always reset invalid anchors to page 1—and, for the foreign case, even one that consulted the foreign anchor but happened to compute page 1—would keep these tests green, leaving this PR's fallback contract unverified.

_Suggested:_ Extend the helper with a requested page number, create enough roots for multiple pages, and assert the exact later page and contents for stale and foreign anchors.

src/main/java/com/otilm/core/dao/entity/notifications/Notification.java
```diff
@@ -57,6 +67,11 @@ public NotificationDto mapToDto() {
         dto.setDetail(this.detail);
         dto.setSentAt(this.sentAt);
         dto.setTargetObjectType(this.targetObjectType);
+        if (this.subject != null) {
```
ivosh: **. Nitpick** | Testing

Notification.java:70-73 is the only step that copies the persisted subject into `NotificationDto`, but no test reads any of the three new DTO properties. CommentEventHandlersITest.java:294-295 and 331-333 stop at the entity field, while NotificationListenerTest.java:224-227 mocks the service; therefore a dropped or swapped assignment in `mapToDto` would leave this PR's tests green while comment notifications still reached clients without a usable deep link.

_Suggested:_ Add a notification service test that creates and reloads a subject-bearing notification, lists it through `NotificationExternalService`, and asserts all three subject DTO fields for both a root and a reply.

src/test/java/com/otilm/core/integration/events/CommentEventHandlersITest.java
```diff
@@ -286,10 +287,12 @@ void globallyBoundProfileDeliversOnResolve() throws Exception {
                 .anySatisfy(message -> assertThat(message.getNotificationProfileUuids()).contains(profileUuid));
 
         notificationListener.processMessage(profileMessages.getFirst());
-        // The persisted notification deep-links to the host object, not the comment the trigger evaluated
+        // The persisted notification targets the host object and names the thread; a root has no parent
```
ivosh: **. Nitpick** | Maintainability

CommentEventHandlersITest.java:290 and CommentEventHandlersITest.java:328 repeat the target and subject relationships expressed by the immediately following AssertJ chains without explaining a non-obvious constraint, leaving two prose copies of the tested behavior to maintain.

_Suggested:_ Delete both assertion-narrating comments and let the descriptive assertions carry the behavior.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
```diff
@@ -158,6 +189,15 @@ public CommentResponseDto listReplies(UUID uuid, PaginationRequestDto pagination
         return CommentMapper.toResponseDto(replies, replyDtos);
     }
 
+    /** The anchored reply when it still exists and still belongs to this thread; null leaves the requested page. */
```
ivosh: **. Nitpick** | Maintainability

CommentServiceImpl.java:192 restates `anchoredReply`, its null guards, and the caller's page fallback without recording any additional rationale or invariant, so the Javadoc adds maintenance overhead to an otherwise self-explanatory helper.

_Suggested:_ Delete the one-line Javadoc above `anchoredReply`.

ivosh (changes requested):

src/main/java/com/otilm/core/dao/entity/notifications/Notification.java
klaraf755: Confirmed, the list projections bypassed the subject entirely. Both now project into a core row that carries it and map through the same path as a fresh notification; a listing test covers root, reply and subject-less notifications, read and unread.

src/test/java/com/otilm/core/integration/service/CommentServiceValidationITest.java
klaraf755: Both tests now request page 2 of five roots and assert that page and its contents.

src/main/java/com/otilm/core/dao/entity/notifications/Notification.java
klaraf755: Covered by the new listing test, which reads all three subject fields through the external service.

src/test/java/com/otilm/core/integration/events/CommentEventHandlersITest.java
klaraf755: Removed.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
klaraf755: Removed.

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
```diff
@@ -109,15 +109,26 @@ public void setAuditOperationDataOverride(AuditOperationDataOverride auditOperat
...
+                : Math
+                        .toIntExact(commentRepository
+                                .countRootsBefore(hostResource, objectUuid.getValue(), anchoredThread.getCreatedAt(),
+                                        anchoredThread.getUuid())
+                                / pagination.getItemsPerPage());
+        Page<Comment> roots = pageOfRoots(hostResource, objectUuid.getValue(), pageIndex, pagination);
+        if (anchoredThread != null && !contains(roots, anchoredThread)) {
+            // The position was counted in a separate read: a root added or removed ahead of the anchor in between
+            // moved it to another page, and the requested page is the documented answer when the anchor cannot be shown
+            roots = pageOfRoots(hostResource, objectUuid.getValue(), pagination.getPageNumber() - 1, pagination);
```
ivosh: **~ Minor** | Correctness

**Under concurrent commenting an anchored listing can land one page away from the anchor — the fallback is safe, just not the closest page.**

`listComments` runs with `Propagation.NOT_SUPPORTED`, so three reads happen with no shared snapshot: the anchor lookup (`:118`), the count of roots ahead of it (`:119-126`), and the page fetch (`:127`). A root inserted or deleted ahead of the anchor between the count and the fetch moves the anchor off the counted page. `listReplies` repeats the same sequence at `:189-196`.

The comment on this line already names that race and picks the requested page as the answer, so this is a deliberate choice rather than something missed. Flagging it only because the anchor is still live and still one hop away.

**When the current behaviour is fine:** the returned page is genuine and correctly ordered, its pagination metadata matches, and the anchor is not lost — it sits on a neighbouring page. The client already copes with an absent anchor, since that is the stale-anchor contract.

**When it costs something:** a race-displaced live anchor is indistinguishable from a stale one, so the deep link degrades to the requested page when the right page was adjacent.

_Suggested:_ on mismatch, recount and re-fetch once, falling back to the requested page only if the second attempt also misses the anchor. Two extra reads on a rare path, and it takes the failure from one race window to two consecutive ones. A repeatable-read transaction spanning all three reads would also work, but `NOT_SUPPORTED` looks deliberate here and this endpoint also runs authorization and a resource lookup, so widening the transaction costs more than the problem does.

ivosh (dismissed):

src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
klaraf755: Taken as suggested: on a miss the count and read are repeated once, then the requested page is served. The helper is shared by both listings, and with #2197 merged the position is now counted in the requested direction too.

ivosh (dismissed):

ivosh (approved):

PR #2201, opened 2026-09-02
Match every published TXT record when validating an ACME DNS-01 challenge

3keyroman:
## Problem

`_acme-challenge.<identifier>` can carry more than one TXT record — one left behind by an earlier attempt
alongside the current one. The DNS provider returns all TXT records of a name as separate values of a single
`TXT` attribute, but validation iterated the returned *attributes* and read only the first value of each. A
name holding more than one record was therefore matched against a single record, chosen by the order the
resolver answered in, so validation of a correctly published record failed whenever a stale record was
answered first.

A failed challenge set only its own status. Its authorization stayed `PENDING`, so a client waiting for the
authorization to reach a final status never learned that validation had failed and waited until its own
request deadline expired instead. The reason was not reported either: `Challenge.error` was always absent,
leaving the cause visible only in the server log.

## Change

- Every TXT record of the queried name is collected and matched. A record published as several
  character-strings is matched by its concatenation as well.
- A failed challenge invalidates its authorization and its order (RFC 8555 §7.1.6) and records the reason,
  which is returned to the client as the challenge's problem document.
- A challenge that has already settled, or whose authorization has, reports its recorded state instead of
  being validated again, so a repeated accept cannot move either out of a final status. A client that wants
  to try again places a new order, as it does against any ACME server that settles authorizations this way.
- A wildcard identifier is validated at the challenge label of its base domain (RFC 8555 §8.4).
- A client's request to deactivate an authorization (RFC 8555 §7.5.2) is honoured; the status was compared by
  reference before, so it never was. A deactivated authorization invalidates an open order, and deactivating an
  account settles its orders under the same order lock.
- An authorization can be read or deactivated, and an order finalized, only by the account that placed the
  order: the request must name that account with `kid`, and a key supplied inline is refused (RFC 8555 §6.2).
- An order becomes `READY` only once every one of its authorizations is `VALID`.
- Challenge validation runs outside any transaction; the outcome is written by a writer bean that locks the order
  row, re-reads the challenge and authorization under that lock and leaves them alone if they settled meanwhile, so
  concurrent accepts of the same challenge, of sibling challenges or of sibling authorizations cannot overwrite a
  final result. The account's failed-order count is incremented in the database.
- Statuses left stale by an instance that does not yet propagate failures are settled when a client asks — before
  the expiry of the authorization or order is enforced: an authorization still `PENDING` behind an `INVALID`
  challenge becomes `INVALID`, and an order still `PENDING` or `READY` behind such an authorization becomes
  `INVALID`. An order that has already requested or received a certificate keeps its status in every case.

The reason recorded on the challenge is authored text. Resolver responses, published record values and
exception messages stay in the server log, because the challenge name can carry verification tokens
belonging to third parties.

## Schema

`acme_challenge` gains the nullable columns `error_problem` and `error_detail`.

A second migration settles the rows the previous behaviour left behind: an authorization still `PENDING`
behind a challenge that is `INVALID` becomes `INVALID`; a `PENDING` or `READY` order with an `INVALID`
authorization becomes `INVALID` and is counted among the account's failed orders. Orders that already requested
or received a certificate are left as they are.


src/main/java/com/otilm/core/service/acme/AcmeChallengeStateMachine.java
```diff
@@ -0,0 +1,122 @@
...
+        AcmeAuthorization authorization = challenge.getAuthorization();
+        AcmeOrder order = authorization.getOrder();
+
+        if (result.valid()) {
+            challenge.setValidated(new Date());
+            challenge.setStatus(ChallengeStatus.VALID);
+            authorization.setStatus(AuthorizationStatus.VALID);
+            if (order.getStatus() == OrderStatus.PENDING) {
+                order.setStatus(OrderStatus.READY);
+            }
```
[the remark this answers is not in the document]
3keyroman: An order now becomes `READY` only once every one of its authorizations is `VALID` (`AcmeChallengeStateMachine.applyValidationResult`).

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -618,28 +613,75 @@ public ResponseEntity<Challenge> validateChallenge(String acmeProfileName, Strin
                 .putLogResourceInfo(com.otilm.api.model.core.auth.Resource.ACME_ORDER, true, order.getUuid().toString(),
                         order.getOrderId());
 
-        boolean isValid;
-        if (challenge.getType().equals(ChallengeType.HTTP01)) {
-            isValid = validateHttpChallenge(challenge);
-        } else {
-            isValid = validateDnsChallenge(challenge);
+        if (challenge.getStatus() != ChallengeStatus.PENDING
+                || authorization.getStatus() != AuthorizationStatus.PENDING) {
```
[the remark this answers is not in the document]
3keyroman: Validation now runs outside any transaction. `AcmeChallengeWriter.applyValidationResult` locks the order row, re-reads the challenge and authorization under that lock and leaves a result that settled in the meantime unchanged.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -618,28 +613,75 @@ public ResponseEntity<Challenge> validateChallenge(String acmeProfileName, Strin
...
+
+    /**
+     * Persists an order whose status changed, counting it against its account when it became invalid.
+     */
+    private void recordOrderTransition(AcmeOrder order, OrderStatus statusBefore) {
+        if (order.getStatus() == statusBefore) {
+            return;
+        }
+        if (order.getStatus() == OrderStatus.INVALID) {
+            incrementFailedOrdersCount(order);
```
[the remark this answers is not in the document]
3keyroman: The failed-order count is incremented in the database (`AcmeAccountRepository.incrementFailedOrders`), called only from the writer.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -573,6 +567,7 @@ public ResponseEntity<Authorization> getAuthorization(String acmeProfileName, St
                     .putLogResourceInfo(com.otilm.api.model.core.auth.Resource.ACME_ORDER, true,
                             authorization.getOrder().getUuid().toString(), authorization.getOrder().getOrderId());
         }
+        settleStaleStatuses(authorization);
```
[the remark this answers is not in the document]
3keyroman: Stale rows are settled before the expiry rejection in `getAuthorization`, `validateChallenge`, `getOrder` and `finalizeOrder`, through `AcmeChallengeWriter.settleOrder`.

src/main/resources/db/migration/V202609021100__settle_authorizations_behind_failed_challenges.sql
```diff
@@ -0,0 +1,29 @@
...
+SET status = 'INVALID', i_upd = CURRENT_TIMESTAMP
+WHERE a.status = 'PENDING'
+  AND EXISTS (SELECT 1 FROM acme_challenge c WHERE c.authorization_uuid = a.uuid AND c.status = 'INVALID');
+
+-- The failed-order count of an account is kept as orders fail, and orders already invalid are never counted
+-- again, so the orders settled here are counted now.
+WITH settled AS (
+    UPDATE acme_order o
+    SET status = 'INVALID', i_upd = CURRENT_TIMESTAMP
+    WHERE o.status IN ('PENDING', 'READY')
```
lubomirw, line 26: **[High]** | Correctness | Confidence: High

The repair updates every PENDING or READY order behind an invalid authorization, including orders with a populated `certificate_ref`. That contradicts both the migration comment and the runtime state machine's `isOpen()` guard, and it also increments `failed_orders` for an order whose certificate already exists.

_Suggested:_ Add `AND o.certificate_ref IS NULL` to this order update and cover a READY certificate-backed order in an execution-level migration test.

<sub>flagged by Guided PR review, Copilot CLI, and Superpowers</sub>

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -604,42 +606,59 @@ public ResponseEntity<Authorization> getAuthorization(String acmeProfileName, St
...
+            logger
+                    .debug("Challenge {} is {} on a {} authorization, returning its current state without validating",
+                            challengeId, challenge.getStatus(), authorization.getStatus());
+            return challengeResponse(acmeProfileName, challenge, isRaProfileBased);
         }
 
-        acmeOrderRepository.save(order);
-        acmeChallengeRepository.save(challenge);
-        acmeAuthorizationRepository.save(authorization);
+        ChallengeValidationResult result = challenge.getType().equals(ChallengeType.HTTP01)
```
lubomirw, line 646: **[Medium]** | Security | Confidence: High

This terminal validation path can be triggered with only a challenge ID. The controller does not accept a JWS body, so the request is neither authenticated nor checked against the owning account before a failed attempt invalidates the authorization and order. RFC 8555 sections 6.2 and 7.5.1 require this POST to carry a verified account-signed JWS.

_Suggested:_ Extend the challenge endpoint contract to accept and validate the JWS body, require ownership, and only then perform and persist validation.

<sub>flagged by Copilot CLI; verified against the controller signature and RFC 8555</sub>

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -1011,32 +1031,14 @@ private void validateAccount(AcmeAccount acmeAccount) throws AcmeProblemDocument
     private void deactivateOrders(AcmeAccount acmeAccount) {
         int failedOrdersCount = 0;
         for (AcmeOrder order : acmeAccount.getOrders()) {
-            // Order might have already been invalid and accounted for, only update count for changed status
-            if (order.getStatus() != OrderStatus.INVALID) {
+            // An order that was already invalid has been counted; only a change of status counts
+            if (acmeChallengeWriter.deactivateOrder(order.getUuid())) {
```
lubomirw, line 1050: **[Medium]** | Correctness | Confidence: High

Account deactivation loads the account before acquiring these order locks, then later derives and saves `failedOrders` from that stale entity. A concurrent challenge failure can commit the new database-side increment while this call waits for an order lock, after which the final full account save can overwrite that increment.

_Suggested:_ Refresh or lock the account after all order locks are acquired, or update status, contact, and the counter through guarded database updates that cannot write a stale counter value.

<sub>flagged by Guided PR review and Superpowers</sub>

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -1646,6 +1616,12 @@ private AcmeOrder validateOrder(String orderId) throws AcmeProblemDocumentExcept
                 acmeOrderRepository.save(order);
             }
         }
+        settleStaleStatuses(order);
```
lubomirw, line 1622: **[Medium]** | Error handling | Confidence: High

After this settlement changes the order to `INVALID`, `getOrder` converts that normal ACME terminal state into HTTP 400 `serverInternal` instead of returning the order representation. Clients therefore see a server fault rather than the validation outcome defined by RFC 8555.

_Suggested:_ Return the mapped order with status `INVALID`; persist and expose an order problem document as well if the model supports it, and update the tests that currently expect the exception.

<sub>flagged by Guided PR review; verified against RFC 8555 order states</sub>

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -1663,36 +1639,50 @@ private void incrementOrderCounts(OrderStatus newStatus, AcmeOrder order) {
     }
 
     private void incrementFailedOrdersCount(AcmeOrder order) {
-        order.getAcmeAccount().setFailedOrders(order.getAcmeAccount().getFailedOrders() + 1);
-        acmeAccountRepository.save(order.getAcmeAccount());
+        acmeChallengeWriter.countFailedOrder(order.getAcmeAccountUuid());
```
lubomirw, line 1644: **[Medium]** | Correctness | Confidence: High

The database-side increment is atomic, but the decision to increment is not. Two concurrent certificate-status polls can both read the same PROCESSING order, both decide it changed to INVALID, and both execute this increment because the order has no lock or optimistic version.

_Suggested:_ Move certificate-derived status transitions into a writer method that locks and re-reads the order before comparing its status and increments the account only for the single persisted transition.

src/test/java/com/otilm/core/service/acme/AcmeDnsChallengeValidatorTest.java
```diff
@@ -0,0 +1,177 @@
...
+        Assertions.assertEquals(List.of(STALE_KEY_AUTHORIZATION, EXPECTED_KEY_AUTHORIZATION), values);
+    }
+
+    @Test
+    void extractsNoValuesWhenTheNameHasNoTxtRecord() throws NamingException {
+        Assertions.assertEquals(List.of(), AcmeDnsChallengeValidator.extractTxtValues(new BasicAttributes(true)));
+    }
+
+    /**
+     * The reported failure: the expected record is published alongside stale ones and must still be accepted.
```
lubomirw: **[Low]** | Maintainability | Confidence: High

`The reported failure` records issue history, while the remaining sentence repeats the descriptive test name.

_Suggested:_ Remove this Javadoc; the method name already preserves the durable scenario.

<sub>flagged by Hygiene</sub>

src/main/resources/db/migration/V202609021100__settle_authorizations_behind_failed_challenges.sql
```diff
@@ -0,0 +1,29 @@
+-- Until a failed challenge invalidated its authorization and order, a failure left both pending behind the
```
lubomirw, line 5: **[Low]** | Maintainability | Confidence: High

Suggestion: since this repair and `V202609021000__acme_challenge_validation_error.sql` are both unreleased parts of the same change, consider creating one migration from the two before either reaches a persistent database.

_Suggested:_ Move the repair into the first migration after the schema changes, include the certificate-backed-order guard, and remove this second migration. Carry the final migration unchanged across the hotfix and forward release lines.

<sub>follow-up migration review</sub>

lubomirw (changes requested):
Automated multi-reviewer triage - 7 selected findings (REQUEST_CHANGES).

src/main/resources/db/migration/V202609021100__settle_authorizations_behind_failed_challenges.sql
3keyroman: Fixed in 79ef3f03b. The order update now carries `AND o.certificate_ref IS NULL`, so it matches the runtime rule and no longer counts such an order as failed.

Flyway is disabled in the test profile, so the suite cannot execute a migration. Verified instead against PostgreSQL 15 on a facsimile of the four tables, with one order per case:

```
order A (PENDING, no certificate)  -> INVALID
order B (READY, certificate_ref)   -> READY      (left alone)
order C (PROCESSING)               -> PROCESSING (left alone)
order D (READY, no certificate)    -> INVALID
failed_orders 5 -> 7               (only A and D counted)
```

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
3keyroman: Fixed in 79ef3f03b. `AcmeChallengeWriter.deactivateOrder` now counts each order it closes through the atomic `UPDATE ... failed_orders + 1`, and the sweep re-reads the account before the caller writes it back, so an increment recorded while the order locks were being taken is no longer overwritten. `AcmeAccount` has no `@DynamicUpdate`, so that write-back really did carry every column.

Covered by `testUpdateAccount_deactivationKeepsACountRecordedWhileItRan`, which records a count mid-deactivation and asserts it survives.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
3keyroman: Fixed in 79ef3f03b. `getOrder` returns the order with `status: invalid` instead of 400 `serverInternal`, per RFC 8555 section 7.1.3.

I left `Order.error` unset deliberately: in RFC 8555 that field carries the error from processing the order, while a validation failure is reported per identifier on the challenges of the order authorizations, which the client reaches through the order. Two of my tests that asserted the exception now assert the returned status, plus a new `testGetOrder_reportsAnInvalidOrderWithItsStatus`. It also makes `AcmeProtocolFlowITest`s `failFast` on an invalid order work as intended, which the exception was defeating.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
3keyroman: Fixed in 79ef3f03b. The transition moved into `AcmeChallengeWriter.reconcileCertificateStatus`, which locks the order row, re-reads it, decides and counts inside that lock; `validateOrder` now calls it. Valid orders are counted with a new atomic `incrementValidOrders`, so neither counter is a read-modify-write any more.

Worth noting this was a regression from making the counter atomic: the previous read-modify-write lost one of two concurrent increments, which masked the double count.

Covered by `testWriter_countsACertificateDerivedTransitionOnce`.

src/test/java/com/otilm/core/service/acme/AcmeDnsChallengeValidatorTest.java
3keyroman: Fixed in 79ef3f03b, Javadoc removed. Agreed on the reasoning, and it matches the repository guidance about not recording issue history in comments.

src/main/resources/db/migration/V202609021100__settle_authorizations_behind_failed_challenges.sql
3keyroman: Done in 79ef3f03b. The repair now lives in `V202609021000__acme_challenge_validation_error.sql` after the column additions, with the certificate-backed-order guard, and the second migration is gone. Neither had been released, so no deployment can have applied the old checksum.

Verified that the single file both adds the two columns and performs the repair when applied by Flyway. The 2.17 hotfix line carries no migration at all, so this file crosses only the forward release lines.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
3keyroman: Confirmed, and left open deliberately.

The gap is real: the interface is `validateChallenge(String acmeProfileName, String challengeId)`, so the endpoint takes no body and nothing is authenticated or checked against the owning account. That is unchanged from 2.17.0, but this PR raises its impact, because a failed attempt is now terminal for the authorization and the order rather than a retryable challenge failure.

Closing it properly means adding the JWS body to `AcmeController` in the `interfaces` repository first, then both controller implementations and the service signature here, and the same for the 2.17 hotfix line. That is cross-repository, so it is not something to fold into this PR unnoticed. Everything else you raised is fixed in 79ef3f03b.

Leaving this thread open for that decision.

src/main/resources/db/migration/V202609021100__settle_authorizations_behind_failed_challenges.sql
3keyroman: Reopening this: I merged them in 79ef3f03b and have reverted that in e8ce20f0d. The merge was wrong, and my resolving this was premature.

In one file the two run in one transaction, so the `ACCESS EXCLUSIVE` lock the `ALTER TABLE` takes on `acme_challenge` is still held while the repair waits for the row locks below it. During a rolling upgrade an instance of the previous version can hold one of those order rows and then read a challenge: it blocks on the DDL lock, the migration blocks on its order row, and PostgreSQL breaks the cycle by aborting one of them. If it aborts the migration, Flyway records a failed migration and the pod does not start.

As two migrations they run in separate transactions: the columns are added and committed, the lock released, and the repair then only reads `acme_challenge` with `ACCESS SHARE`, which is compatible with a concurrent writer. No cycle.

Both are still unreleased, so two files cost nothing in checksum terms, and the reason is now recorded at the top of the repair migration. The certificate-backed-order guard you asked for stays. Happy to reconsider if you see a way to keep one file without holding the DDL lock across the row waits.

3keyroman: Resolving this as a considered decline rather than a change.

The suggestion is sound on tidiness, and both migrations are indeed unreleased, so there is no checksum cost either way. The blocker is operational: in one file the column additions and the repair share a transaction, so the `ACCESS EXCLUSIVE` lock the `ALTER TABLE` holds on `acme_challenge` is still held while the repair waits for row locks on `acme_authorization` and `acme_order`. During a rolling upgrade an instance of the previous version that holds one of those rows and then reads a challenge closes the cycle, and if PostgreSQL resolves it by aborting the migration, Flyway records a failed migration and the pod does not start.

As two migrations they run in separate transactions, the DDL lock is released before the repair takes anything, and the repair only reads `acme_challenge` with `ACCESS SHARE`, which a concurrent writer does not conflict with.

The reason is recorded at the top of the repair migration so it is not lost, and the certificate-backed-order guard you asked for is in place. Reopen if you would rather take the merge with the deadlock risk, or if you see a formulation that keeps one file without holding the DDL lock across the row waits.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
3keyroman: Tracked separately as #2214 and resolving here.

The finding stands, and the RFC is unambiguous, so it is recorded rather than dismissed: #2214 carries the §6.2, §6.3, §6.4 and §7.5.1 citations, the four endpoints that take no JWS (`validateChallenge`, plus `getOrder`, `listOrders` and `downloadCertificate` on POST-as-GET), the reason this PR raises the stakes on the first of them, and the order of work — the signatures live in the `interfaces` repository, so it starts there.

Keeping it out of this PR is deliberate: the fix is cross-repository and touches three endpoints unrelated to DNS-01 validation, and folding it in silently would hide an API contract change inside a bug fix.

lubomirw (approved):

PR #2203, opened 2026-09-02
Make the Mockito and HQL-rendering test shapes hold on both framework lines

### LINKED ISSUE #2202: Make the Mockito and HQL-rendering test shapes hold on both framework lines
### Description

**Two test classes assert on framework details that Spring Boot 4 and Hibernate 7 spell
differently, and both can be made version-neutral on the 3.5 line — so they land as
preparation rather than as migration churn.**

Neither is a production defect. Both are carved out of #2017, which recorded the second as
RC#4 and attributed the first nowhere.

### RC#4 — criteria operands render quoted in `toHqlString()` (8 tests)

`FilterPredicatesBuilderITest` asserts the rendered HQL of a predicate's operand. What changed
is [HHH-17002](https://hibernate.atlassian.net/browse/HHH-17002) (query-plan caching for
`CriteriaQuery`), fixVersion **7.0.0.CR1**, never backported to 6.6. On 6.6.53 a value bind
appended the raw value and only a `String` literal was quoted; from 7.0 binds and literals
share one `SqmLiteral.appendHqlString`. For a non-null value with a non-null `JavaType`, that
quotes everything except `Number`, `Boolean` and `Enum`; `null` renders as `null`.

| Expression | 6.6.53 | 7.4.5 |
|---|---|---|
| `cb.equal(path, "test")` right-hand — `String` value bind | `test` | `'test'` |
| `cb.equal(expr, true)` right-hand — `Boolean` value bind | `true` | `true` |
| `cb.literal("test")` — `String` literal | `'test'` | `'test'` |
| `cb.literal(<LocalDateTime>)` — non-`String` literal | `2026-…` | `'2026-…'` |

The two forms already disagreed on 6.6, which is why one assertion in the file expected
`'test'` with quotes while the rest expected `test` without. Five of the eight sites are value
binds and three are literals; the bind at 316 carries a `Boolean`, which both versions render
unquoted, so the helper is inert there. All eight `toHqlString()` sites now go through a helper
that strips one surrounding pair of quotes, so no assertion depends on the rendered spelling.
Node shape is kept deliberately: the helper would otherwise stop telling a literal apart from a
value bind at the MATCHES site, so that site also asserts its argument is a `SqmLiteral`.

`toHqlString()` appears in this one test file and nowhere in `src/main`.

### Mockito annotations are no longer opened for a `@SpringBootTest` (3 tests)

`ConnectorServiceMockITest` is a `@SpringBootTest` that uses plain `@Mock` and `@InjectMocks`
with no `@ExtendWith(MockitoExtension.class)` and no `openMocks` call. Boot 3 opened those
annotations through `org.springframework.boot.test.mock.mockito.MockitoTestExecutionListener`.
Boot 4 removes that listener together with the `@MockBean` package, so every annotated field
stays null and `setUp` fails:

```
NullPointerException: Cannot invoke "ConnectorServiceImpl.setConnectorAdapters(java.util.Map)"
  because "this.connectorService" is null
```

The test now opens the annotations itself and closes them in `@AfterEach`.

**This is the only test in the repo relying on that implicit behaviour.** 80 test files use
plain `@Mock` or `@InjectMocks`; the other 79 already declare `MockitoExtension` or call
`openMocks`.

### Evidence

Same commit, both framework lines, both test classes:

| Line | Result |
|---|---|
| `main` — Boot 3.5, Hibernate 6.6.53 | 55 tests, 0 failures |
| `refactor/2016-spring-boot-4.1-baseline` — Boot 4.1.1, Hibernate 7.4.5 | 55 tests, 0 failures |

Before the change, the same two classes contribute 8 failures and 3 errors on the 4.1
baseline branch and pass on `main`.

### Definition of Done

- [ ] Both classes pass on Spring Boot 3.5 / Hibernate 6.6 — this must land as a no-op there.
- [ ] Both classes pass unchanged on the 4.1 baseline branch, verified by applying the commit
      to it rather than by inspection.
- [ ] No assertion depends on whether Hibernate quotes a rendered value.
- [ ] `mvn -B verify` passes; Spotless and Checkstyle gates pass.

### Follow-up

Removes 11 of the failures on `refactor/2016-spring-boot-4.1-baseline`. The remainder of
#2017 is RC#8, the follow-on locking defect, which cannot be prepared on 3.5 because it does
not reproduce there.

Preparation for the Spring Boot 4.1 migration, carved out of #2017. Lands on `main`, not on
the integration branch. Part of the preparation list on #1941.

ivosh:
Closes #2202.

**Two test classes assert on framework details that Spring Boot 4 and Hibernate 7 spell
differently. Both are made version-neutral here, on the 3.5 line, so they never become
migration churn.**

No production code changes. Both classes pass on `main` today and keep passing; the point of
the change is that they also pass on the 4.1 baseline, where they currently contribute 8
failures and 3 errors.

## `FilterPredicatesBuilderITest` — 8 tests

The assertions compare the rendered HQL of a predicate's operand. What changed is
[HHH-17002](https://hibernate.atlassian.net/browse/HHH-17002) (query-plan caching for
`CriteriaQuery`), fixVersion **7.0.0.CR1**, never backported to 6.6. On 6.6.53 a value bind
appended the raw value and only a `String` literal was quoted; from 7.0 binds and literals
share one `SqmLiteral.appendHqlString`. For a non-null value with a non-null `JavaType`, that
quotes everything except `Number`, `Boolean` and `Enum`; `null` renders as `null`.

| Expression | 6.6.53 | 7.4.5 |
|---|---|---|
| `cb.equal(path, "test")` right-hand — `String` value bind | `test` | `'test'` |
| `cb.equal(expr, true)` right-hand — `Boolean` value bind | `true` | `true` |
| `cb.literal("test")` — `String` literal | `'test'` | `'test'` |
| `cb.literal(<LocalDateTime>)` — non-`String` literal | `2026-…` | `'2026-…'` |

The two forms already disagreed on 6.6, which is why one assertion in this file expected
`'test'` **with** quotes while the other seven expected `test` **without**. The test was
pinning the rendered spelling of whichever internal form Hibernate happened to pick, which is
what breaks across the versions.

Five of the eight sites are value binds (278, 300, 316, 350, 2171) and three are literals
(323, 406, 419). The bind at 316 carries a `Boolean`, which both versions render unquoted, so
the helper is inert there. 406 and 419 are the sites that could have broken and do not — a
temporal rendered as `{ts '…'}` would survive the helper untouched, but 7.4.5 has no temporal
`appendHqlString` override, so `LocalDateTime` takes the generic quoted branch.

All eight `toHqlString()` sites now go through `unquoted(...)`, which strips one surrounding
pair of quotes, so no assertion depends on the rendered spelling. Node *shape* is a different
property and is deliberately kept: `unquoted()` would otherwise stop telling a literal apart
from a value bind at the MATCHES site, so that site now asserts its argument is a
`SqmLiteral`.

`toHqlString()` appears in this one test file and nowhere in `src/main`.

## `ConnectorServiceMockITest` — 3 tests

A `@SpringBootTest` using plain `@Mock` and `@InjectMocks`, with no
`@ExtendWith(MockitoExtension.class)` and no `openMocks` call. Boot 3 opened those annotations
through `org.springframework.boot.test.mock.mockito.MockitoTestExecutionListener`. Boot 4
removes that listener along with the `@MockBean` package, so the fields stay null:

```
NullPointerException: Cannot invoke "ConnectorServiceImpl.setConnectorAdapters(java.util.Map)"
  because "this.connectorService" is null
```

The test now opens the annotations itself and closes them in `@AfterEach`.

**This is the only test in the repo relying on that implicit behaviour.** 80 test files use
plain `@Mock` or `@InjectMocks`; the other 79 already declare `MockitoExtension` or call
`openMocks`. So this is a one-off, not the first of a sweep.

## Verification

The same commit was run against both framework lines — the 4.1 result is from applying this
commit to `refactor/2016-spring-boot-4.1-baseline`, not from inspection:

| Line | `FilterPredicatesBuilderITest` | `ConnectorServiceMockITest` |
|---|---|---|
| `main` — Boot 3.5, Hibernate 6.6.53 | 52 tests, 0 failures | 3 tests, 0 failures |
| 4.1 baseline — Boot 4.1.1, Hibernate 7.4.5 | 52 tests, 0 failures | 3 tests, 0 failures |

## Scope

Preparation for the Spring Boot 4.1 migration, carved out of #2017, which recorded the first
as RC#4 and attributed the second nowhere. Removes 11 failures from the integration branch
before it is cut — the same pattern as #2149.

What remains in #2017 is RC#8, the follow-on locking defect (HHH-20744). That one cannot be
prepared on 3.5, because it does not reproduce there.

### DIFF
--- a/src/test/java/com/otilm/core/integration/service/ConnectorServiceMockITest.java
+++ b/src/test/java/com/otilm/core/integration/service/ConnectorServiceMockITest.java
@@ -32,11 +32,13 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
 import org.springframework.boot.test.context.SpringBootTest;
 
 import static org.mockito.Mockito.any;
@@ -100,8 +102,12 @@ class ConnectorServiceMockITest {
 
     private Endpoint endpoint1, endpoint2, endpoint3;
 
+    private AutoCloseable mocks;
+
     @BeforeEach
     void setUp() {
+        mocks = MockitoAnnotations.openMocks(this);
+
         endpoint1 = new Endpoint();
         endpoint1.setUuid(UUID.fromString("abfbc322-29e1-11ed-a261-0242ac120002"));
         endpoint1.setName("endpoint1");
@@ -141,6 +147,11 @@ class ConnectorServiceMockITest {
 
         when(connectorApiFactory.getConnectorApiClient(any(ApiClientConnectorInfo.class)))
                 .thenReturn(connectorApiClient);
+    }
+
+    @AfterEach
+    void tearDown() throws Exception {
+        mocks.close();
     }
 
     @Test
--- a/src/test/java/com/otilm/core/integration/util/FilterPredicatesBuilderITest.java
+++ b/src/test/java/com/otilm/core/integration/util/FilterPredicatesBuilderITest.java
@@ -101,6 +101,7 @@ import java.util.UUID;
 import java.util.stream.Collectors;
 import org.hibernate.query.sqm.ComparisonOperator;
 import org.hibernate.query.sqm.function.SelfRenderingSqmFunction;
+import org.hibernate.query.sqm.tree.expression.SqmLiteral;
 import org.hibernate.query.sqm.tree.predicate.SqmComparisonPredicate;
 import org.hibernate.query.sqm.tree.predicate.SqmExistsPredicate;
 import org.hibernate.query.sqm.tree.predicate.SqmJunctionPredicate;
@@ -318,6 +319,7 @@ class FilterPredicatesBuilderITest extends BaseSpringBootTest {
         SelfRenderingSqmFunction<?> leftHandExpressionHandExpression = (SelfRenderingSqmFunction<?>) comparisonPredicateTest
                 .getLeftHandExpression();
         Assertions.assertEquals("textregexeq", leftHandExpressionHandExpression.getFunctionName());
+        Assertions.assertInstanceOf(SqmLiteral.class, leftHandExpressionHandExpression.getArguments().getLast());
         Assertions
                 .assertEquals(testValue,
                         unquoted(leftHandExpressionHandExpression.getArguments().getLast().toHqlString()));
--- a/src/test/java/com/otilm/core/integration/util/FilterPredicatesBuilderITest.java
+++ b/src/test/java/com/otilm/core/integration/util/FilterPredicatesBuilderITest.java
@@ -275,7 +275,7 @@ class FilterPredicatesBuilderITest extends BaseSpringBootTest {
         Assertions.assertEquals(ComparisonOperator.EQUAL, ((SqmComparisonPredicate) predicateTest).getSqmOperator());
         Assertions
                 .assertEquals(testValue,
-                        ((SqmComparisonPredicate) predicateTest).getRightHandExpression().toHqlString());
+                        unquoted(((SqmComparisonPredicate) predicateTest).getRightHandExpression().toHqlString()));
     }
 
     @Test
@@ -297,7 +297,7 @@ class FilterPredicatesBuilderITest extends BaseSpringBootTest {
                                 ((SqmComparisonPredicate) predicate).getSqmOperator());
                 Assertions
                         .assertEquals(testValue,
-                                ((SqmComparisonPredicate) predicate).getRightHandExpression().toHqlString());
+                                unquoted(((SqmComparisonPredicate) predicate).getRightHandExpression().toHqlString()));
             } else {
                 Assertions.assertFalse(predicate.isNull().isNegated());
             }
@@ -313,14 +313,14 @@ class FilterPredicatesBuilderITest extends BaseSpringBootTest {
         Assertions.assertInstanceOf(SqmComparisonPredicate.class, predicateTest);
         SqmComparisonPredicate comparisonPredicateTest = (SqmComparisonPredicate) predicateTest;
         Assertions.assertEquals(ComparisonOperator.EQUAL, comparisonPredicateTest.getSqmOperator());
-        Assertions.assertEquals("true", comparisonPredicateTest.getRightHandExpression().toHqlString());
+        Assertions.assertEquals("true", unquoted(comparisonPredicateTest.getRightHandExpression().toHqlString()));
         Assertions.assertInstanceOf(SelfRenderingSqmFunction.class, comparisonPredicateTest.getLeftHandExpression());
         SelfRenderingSqmFunction<?> leftHandExpressionHandExpression = (SelfRenderingSqmFunction<?>) comparisonPredicateTest
                 .getLeftHandExpression();
         Assertions.assertEquals("textregexeq", leftHandExpressionHandExpression.getFunctionName());
         Assertions
-                .assertEquals("'" + testValue + "'",
-                        leftHandExpressionHandExpression.getArguments().getLast().toHqlString());
+                .assertEquals(testValue,
+                        unquoted(leftHandExpressionHandExpression.getArguments().getLast().toHqlString()));
     }
 
     @Test
@@ -346,7 +346,8 @@ class FilterPredicatesBuilderITest extends BaseSpringBootTest {
             if (predicate instanceof SqmLikePredicate) {
                 Assertions.assertTrue(predicate.isNegated());
                 Assertions
-                        .assertEquals("%" + testValue + "%", ((SqmLikePredicate) predicate).getPattern().toHqlString());
+                        .assertEquals("%" + testValue + "%",
+                                unquoted(((SqmLikePredicate) predicate).getPattern().toHqlString()));
             } else {
                 Assertions.assertFalse(predicate.isNull().isNegated());
             }
@@ -402,7 +403,7 @@ class FilterPredicatesBuilderITest extends BaseSpringBootTest {
                         ((SqmComparisonPredicate) predicateTest).getSqmOperator());
         Assertions
                 .assertEquals(getFormattedDate(),
-                        ((SqmComparisonPredicate) predicateTest).getRightHandExpression().toHqlString());
+                        unquoted(((SqmComparisonPredicate) predicateTest).getRightHandExpression().toHqlString()));
     }
 
     @Test
@@ -415,7 +416,7 @@ class FilterPredicatesBuilderITest extends BaseSpringBootTest {
                 .assertEquals(ComparisonOperator.LESS_THAN, ((SqmComparisonPredicate) predicateTest).getSqmOperator());
         Assertions
                 .assertEquals(getFormattedDate(),
-                        ((SqmComparisonPredicate) predicateTest).getRightHandExpression().toHqlString());
+                        unquoted(((SqmComparisonPredicate) predicateTest).getRightHandExpression().toHqlString()));
     }
 
     @NotNull
@@ -2156,9 +2157,18 @@ class FilterPredicatesBuilderITest extends BaseSpringBootTest {
                 .collect(Collectors.toSet());
     }
 
+    /**
+     * Strips the quoting Hibernate applies inconsistently across versions.
+     */
+    private static String unquoted(final String hqlString) {
+        return hqlString.length() > 1 && hqlString.startsWith("'") && hqlString.endsWith("'")
+                ? hqlString.substring(1, hqlString.length() - 1)
+                : hqlString;
+    }
+
     private void testLikePredicate(final Predicate predicate, final String value) {
         Assertions.assertInstanceOf(SqmLikePredicate.class, predicate);
-        Assertions.assertEquals(value, ((SqmLikePredicate) predicate).getPattern().toHqlString());
+        Assertions.assertEquals(value, unquoted(((SqmLikePredicate) predicate).getPattern().toHqlString()));
     }
 
     private SearchFilterRequestDto prepareDummyFilterRequest(final FilterConditionOperator condition) {


karolhrdina (commented):
## Review — verified rather than inspected, then two findings and one follow-up

I checked this at the source level and by running the tests, on both branches. What held:

| Claim | Result |
|---|---|
| Both classes pass on `main` (pre-change) | **Verified** — 55 tests, 0 failures, run locally on `7889edd2` |
| Both classes pass on this PR head | **Verified** — 55 tests, 0 failures, run locally on `308972c2` |
| All eight `toHqlString()` sites wrapped, none missed | **Verified** — `main` had exactly 8 (278, 300, 316, 323, 349, 405, 418, 2161); all 8 now go through `unquoted(...)`, and `grep -c unquoted` on `main` is 0 |
| `toHqlString()` absent from `src/main` | **Verified** — 0 hits |
| `unquoted()` is sufficient on Hibernate 7.4.5 at all 8 sites | **Verified** against the official `-sources.jar` for 6.6.0, 6.6.53, 7.0.0, 7.1.0, 7.2.0 and 7.4.5 |
| One-off, not the first of a sweep | **Verified exactly** — 80 files use plain `@Mock`/`@InjectMocks`; 78 declare `MockitoExtension`, one calls `openMocks` (`StatisticsServiceImplTest:44`), and this is the 80th. Superclass chains walked 6 deep, no inherited mechanism anywhere. Also: of 33 `@SpringBootTest` classes this is the only one with plain `@Mock`/`@InjectMocks` |
| Nothing else left in `org.springframework.boot.test.mock.mockito` | **Verified** — 0 occurrences of `@MockBean`, `@SpyBean`, `MockitoTestExecutionListener`, `ResetMocksTestExecutionListener`, or the package import. The repo is already on `@MockitoBean`/`@MockitoSpyBean` (64 files), which lives in `spring-test` and keeps its own `MockitoResetTestExecutionListener`, so Boot 4 dropping Boot's listener is a non-event |
| No other version-sensitive rendering assertion anywhere in `src/test` | **Verified** — no test asserts on generated SQL/DDL text, and no golden file pins SQL |
| Spotless + Checkstyle | **Pass**. The gate's `checkstyle.xml` carries only `AvoidStarImport`, `NeedBraces`, `UnusedImports`, `RedundantImport` — no member-ordering rule for the new static helper to trip |

Two parts of the Mockito change are non-obvious enough that I want the verification on the record:

**The double-open on the 3.5 line is harmless.** Boot 3.5.16's `MockitoTestExecutionListener.prepareTestInstance` already calls `openMocks`, so the annotations are now opened twice per test method. I reproduced the exact field shapes against mockito-core 5.17.0: `FieldInitializer` short-circuits on the non-null `@InjectMocks` fields, so the second pass re-injects into the *same* instances instead of creating new ones; both targets end up holding the generation-2 mocks the `@Mock` fields point at; and both `AutoCloseable`s are closed — yours in `@AfterEach`, the listener's in `afterTestMethod`. No stale-mock split, no leaked closeable. (Corroborating the premise: all 17 classes in that package carry `@Deprecated(since = "3.4.0", forRemoval = true)` in 3.5.16.)

**`openMocks` rather than `@ExtendWith(MockitoExtension.class)` is the correct call, not merely the smaller one.** The extension pins `Strictness.STRICT_STUBS` and validates per method. `ConnectorV1Adapter.validateFunctionGroups` (`src/main/java/com/otilm/core/service/handler/ConnectorV1Adapter.java:276`) calls `findByCode` once per returned function group, and `testConnect_UnknownFunctionGroup` supplies only `AUTHORITY_PROVIDER` — so the `findByCode(CREDENTIAL_PROVIDER)` stub at lines 145-146 goes unused in that test. Reproduced: `UnnecessaryStubbingException`. Swapping in the extension for tidiness would regress the suite.

Findings are inline: one lost assertion worth fixing here, one correction to the recorded mechanism (no code change, but that table gets reused for the rest of the migration), and one follow-up.

Recorded so nobody re-derives them, none of them actionable: `unquoted()` is a provable no-op at 7 of the 8 sites on 6.6 today, so there it is pure forward-insurance; the `mocks.close()` NPE if `setUp` threw before the assignment cannot mask a cause, since `openMocks` is the first statement and JUnit reports the original as primary with the NPE suppressed; and `QuotingHelper` doubles an embedded `'`, so a future filter value containing a quote would render as `'o''brien'` and `unquoted()` would yield `o''brien` — unreachable while `testValue = "test"`.

src/test/java/com/otilm/core/integration/util/FilterPredicatesBuilderITest.java
```diff
@@ -313,14 +313,14 @@ void testMatchesPredicate() {
...
+        Assertions.assertEquals("true", unquoted(comparisonPredicateTest.getRightHandExpression().toHqlString()));
         Assertions.assertInstanceOf(SelfRenderingSqmFunction.class, comparisonPredicateTest.getLeftHandExpression());
         SelfRenderingSqmFunction<?> leftHandExpressionHandExpression = (SelfRenderingSqmFunction<?>) comparisonPredicateTest
                 .getLeftHandExpression();
         Assertions.assertEquals("textregexeq", leftHandExpressionHandExpression.getFunctionName());
         Assertions
-                .assertEquals("'" + testValue + "'",
-                        leftHandExpressionHandExpression.getArguments().getLast().toHqlString());
+                .assertEquals(testValue,
+                        unquoted(leftHandExpressionHandExpression.getArguments().getLast().toHqlString()));
```
karolhrdina, line 325: This is the one site whose *expected* value changed, and the change drops a real assertion.

The old form pinned the **quoted** shape:

```java
.assertEquals("'" + testValue + "'", ...getArguments().getLast().toHqlString());
```

On 6.6 only a `SqmLiteral` renders quoted — `ValueBindJpaCriteriaParameter` appends the raw value. So that assertion was implicit proof that `FilterPredicatesBuilder.java:242` builds the `textregexeq` argument with `criteriaBuilder.literal(filterValues.getFirst())` rather than handing the raw value to the criteria API. `unquoted()` collapses the two forms, so the assertion is now satisfied by either.

The concrete regression it stops catching: change `FilterPredicatesBuilder.java:242` from `criteriaBuilder.literal(filterValues.getFirst())` to `filterValues.getFirst()` and `testMatchesPredicate` stays green. Before this PR it failed.

Calibrating that honestly — `textregexeq` is Hibernate's PostgreSQL-dialect name for `~`, and a bind parameter would still execute correctly, so this is lost coverage of the SQL shape rather than a live defect. It is also cheap to keep both properties, by asserting the node type next to the value:

```java
Assertions
        .assertInstanceOf(SqmLiteral.class,
                leftHandExpressionHandExpression.getArguments().getLast());
Assertions
        .assertEquals(testValue,
                unquoted(leftHandExpressionHandExpression.getArguments().getLast().toHqlString()));
```

(with `import org.hibernate.query.sqm.tree.expression.SqmLiteral;`). The node type is what the two Hibernate versions agree on, so that stays version-neutral while restoring the property the old string comparison was asserting by accident.

src/test/java/com/otilm/core/integration/util/FilterPredicatesBuilderITest.java
```diff
@@ -2156,9 +2157,18 @@ private Set<UUID> getUuidsFromListCertificatesResponse(CertificateResponseDto ce
                 .collect(Collectors.toSet());
     }
 
+    /**
+     * Strips the quoting Hibernate applies inconsistently across versions.
+     */
+    private static String unquoted(final String hqlString) {
```
karolhrdina, line 2165: The helper is correct and sufficient — I verified all eight sites against the official `-sources.jar` for 6.6.0, 6.6.53, 7.0.0, 7.1.0, 7.2.0 and 7.4.5. But the mechanism recorded here and in #2202 is off in three ways, and that table is going to be read again for the rest of the migration.

**1. In 7.x, binds and literals are the same code path, and the quoting is type-dependent.** Both funnel through one static `SqmLiteral.appendHqlString`:

```java
if ( value instanceof Number || value instanceof Boolean ) sb.append( value );        // unquoted
else if ( value instanceof Enum<?> e ) sb.append( ... ).append( "." ).append( e.name() );  // unquoted
else appendSingleQuoteEscapedString( sb, javaType.toString( value ) );                // quoted
```

6.6.53 quoted only when `javaType.getJavaTypeClass() == String.class`, and appended raw otherwise. So the table's `cb.literal("test")`: `'test'` → `'test'` row is true **only for String** — for `LocalDateTime` the literal path went unquoted → quoted. The corrected model predicts breakage the recorded one says cannot happen, which is exactly the kind of thing that matters when the table is reused.

**2. Three of the eight sites are literals, not value binds.** Lines 323, 406 and 419 come from `criteriaBuilder.literal(...)` — `FilterPredicatesBuilder.java:242` for MATCHES, `:210` and `:213` for GREATER/LESSER. Only 278, 300, 350 and 2171 are `ValueBindJpaCriteriaParameter`.

Worth saying plainly that 406/419 are the sites that could have broken and don't: a temporal rendered as `{ts '…'}` or `datetime …` would pass through `unquoted()` untouched. It cannot happen — 7.4.5 has no temporal-specific `appendHqlString` override and no typed-literal or JDBC-escape rendering anywhere under `org/hibernate/query/sqm/tree/`, so `LocalDateTime` falls into the generic quoted branch, and `LocalDateTimeJavaType.toString` is byte-identical in both versions (`DateTimeFormatter.ISO_DATE_TIME.format(value)`, which for this value equals `LocalDateTime.toString()`). That is the one place the approach could have quietly failed, so it deserves to be in the record rather than inferred.

**3. Version and issue.** This is HHH-17002 ("Query plan caching for CriteriaQuery based on query structure"), commit `d31837969f0db851497eb5db1e88189a5ba4911a`, fixVersion **7.0.0.CR1** — not "Hibernate 7.2" as the description says. Never backported to 6.6; bisected across 6.6.0 / 6.6.53 / 7.0.0 / 7.1.0 / 7.2.0 / 7.4.5.

One consequence for the helper itself: line 316 is a `Boolean`, which `sb.append(value)` renders unquoted in *both* versions, so `unquoted()` is inert there. Harmless — just not one of the sites that needed it.

src/test/java/com/otilm/core/integration/service/ConnectorServiceMockITest.java
```diff
@@ -100,8 +102,12 @@ class ConnectorServiceMockITest {
 
     private Endpoint endpoint1, endpoint2, endpoint3;
 
+    private AutoCloseable mocks;
```
karolhrdina, line 105: Now that the test opens the annotations itself, `@SpringBootTest` on this class is vestigial — and it was the only thing keeping the class compliant where it sits.

Nothing in the 204 lines touches the context: no `@Autowired`, `@Value`, `@MockitoBean`, `@TestConfiguration`, `EntityManager` or `ApplicationContext`. The subject is built by Mockito. Before this PR `@SpringBootTest` was load-bearing precisely because Boot 3's `MockitoTestExecutionListener` was what opened the annotations; after it, all it does is boot a context.

Measured on a worktree at this PR's head with `@SpringBootTest` and its import removed:

- the 3 tests still pass, in **2.85 s** instead of **23.26 s**
- `CiTestSplitTest.integrationRootContainsOnlyContextTests` **fails**: `Expecting empty but was: ["service/ConnectorServiceMockITest.java"]` — *"Non-context (pure unit) test classes found in the integration root. Move them out to a unit location and rename *Test."*

So the repo's own taxonomy guard already classifies this as a unit test that passes only because of an annotation it no longer needs. The complete move is: relocate out of `com/otilm/core/integration/`, rename to `ConnectorServiceMockTest`, drop `@SpringBootTest`.

Being straight about the payoff, because the 20 s does not translate: `ContextSignatureGuardTest` stays green with `BASELINE` at 61, since the bare-`@SpringBootTest` signature is shared with `ApplicationITest`, `service/scep/CryptographicProviderITest`, `util/AcmeUtilITest` and `cryptography/PQCITest` — and `CryptographicProviderITest` sits in the same `test-integration-service-1` leg, so whichever runs first pays the boot either way and CI's marginal saving is small. The real value is that the class stops depending on Spring's test infrastructure at all, which makes this exact breakage structurally impossible instead of patched.

Follow-up sized rather than blocking — it needs a move and a rename, not a one-liner, and this PR is deliberately minimal.

src/test/java/com/otilm/core/integration/util/FilterPredicatesBuilderITest.java
ivosh: Fixed in 78f98904. `assertInstanceOf(SqmLiteral.class, ...)` now sits above the value assertion, with the import.

Two corrections to the repro. The stated mutation does not compile: `cb.function(name, type, args...)` takes `Expression<?>[]`, so a bare `filterValues.getFirst()` gives `method function in interface jakarta.persistence.criteria.CriteriaBuilder cannot be applied to given types; required: java.lang.String,java.lang.Class<T>,jakarta.persistence.criteria.Expression<?>[]`. And `testMatchesPredicate` builds its filter with `FilterFieldSource.PROPERTY`, so it executes the MATCHES case in `getPropertyFilterPredicate` at `FilterPredicatesBuilder.java:613`, not the one at `:242` in `getAttributeFilterConditionPredicate`. Mutating `:242` changes nothing this test reaches — that cost me a wrong conclusion before I checked which path runs.

Substituting `((HibernateCriteriaBuilder) criteriaBuilder).value(filterValues.getFirst())` at `:613`, the nearest compiling bind, reproduces the finding exactly:

| Test state | Result |
|---|---|
| `main`, pre-change | fails — `expected: <'test'> but was: <test>` |
| this PR, `unquoted()` alone | passes — the lost coverage |
| this PR, with the instance check | fails — `expected: <SqmLiteral> but was: <ValueBindJpaCriteriaParameter>` |

The node type is version-neutral: 55 tests, 0 failures on both 6.6.53 and 7.4.5.

src/test/java/com/otilm/core/integration/util/FilterPredicatesBuilderITest.java
ivosh: Fixed in 78f98904. All three corrections are in the PR description and in #2202, with one amendment to the second.

**1.** Confirmed against the sources jars. 6.6.53 `SqmLiteral.appendHqlString` quotes only when `javaType.getJavaTypeClass() == String.class` and appends raw otherwise; 7.4.5 `ValueBindJpaCriteriaParameter.appendHqlString` delegates to `SqmLiteral.appendHqlString`. Recorded with the guard your wording omits: the Number/Boolean/Enum split applies to a non-null value with a non-null `JavaType`; `null` renders as `null`.

**2.** Confirmed for the three literals — 323, 406 and 419 are `criteriaBuilder.literal(...)` — and 406/419 are recorded as the sites that could have broken and do not. One amendment: **316 is a fifth value bind, not a non-bind.** `cb.equal(expr, true)` routes through `SqmCriteriaNodeBuilder.equal(Expression, Object)` → `value(y, x)`, and with the default `BIND` handling mode that yields a `ValueBindJpaCriteriaParameter`. Probed on the running test: `class=org.hibernate.query.sqm.tree.expression.ValueBindJpaCriteriaParameter rendered=true`. Your conclusion is unaffected — it is inert either way, unquoted on 6.6 via `sb.append(value)` and unquoted on 7.4.5 via the `Boolean` branch — but the split is five binds and three literals.

**3.** Corrected to HHH-17002, fixVersion 7.0.0.CR1. Independent corroboration from a separate run: pinning Hibernate to 7.1.34 on `refactor/2016-spring-boot-4.1-baseline` leaves all 8 of these failures standing, so the change is present well before 7.2.

src/test/java/com/otilm/core/integration/service/ConnectorServiceMockITest.java
ivosh: Not changed. Premise confirmed, and your guard measurement reproduces here: removing `@SpringBootTest` and its import turns `CiTestSplitTest.integrationRootContainsOnlyContextTests` red with `Expecting empty but was: ["service/ConnectorServiceMockITest.java"]`.

Declining the full proposal — relocate, rename to `ConnectorServiceMockTest`, drop `@SpringBootTest` — on its own cost and benefit rather than on scope. The benefit is the one you measured and discounted: `ContextSignatureGuardTest` stays at `BASELINE` 61 because the bare-`@SpringBootTest` signature is shared, and `CryptographicProviderITest` pays the boot in the same CI leg regardless, so the marginal saving is small. The cost is a path change that moves the class between surefire include/exclude sets in the CI split profiles, which is a change to how the suite is partitioned rather than to a test.

That trade is worth making on its own, with the partitioning change visible in its own diff. Left open for you.

karolhrdina (approved):

Issue #951, opened 2025-02-09
Integrating notifications with events and workflow engine


lubomirw:
Existing notification types should be replaced with events that occurs in system and integrate them with workflows which gives user flexibility how to automatize processes as reaction on events in system. In that way, notification can be only one part of action when some event occurs.

- separate certificate discovered event from discovery finished event and remove `eventResource` of trigger
- implement endpoint to list all events
- replace `producedResource` event property with list of resources up the hierarchy (or producing event as underlying) that can override event
- update `ResourceService.listResourceEvents` method accordingly to list events based on resource or overriding resource.
- add column/property `event` to `TriggerAssociation` table to link it with resource
- refactor how event and notification queue is used

PR #2211, opened 2026-09-03
Serve group users from auth group membership

### LINKED ISSUE #2210: Serve group users from auth group membership
### Description

Implement the `GET /v1/groups/{uuid}/users` endpoint declared in the `interfaces` repo, returning the users assigned to a group.

Add `listGroupUsers(SecuredUUID)` to `GroupService` / `GroupServiceImpl`, annotated `@ExternalAuthorization(resource = Resource.GROUP, action = ResourceAction.MEMBERS)`. The action already exists and is already granted by migrations, so nothing needs to change in `auth` or `auth-opa-policies` - `AuthResourceSynchronizer.register()` syncs resource actions at startup.

Membership lives in the auth service, denormalized onto the user record: auth has no group table, only a `groups` text column on `user` holding `{uuid, name}` pairs that Core writes via `groupUuids` on `AddUserRequestDto` / `UpdateUserRequestDto` (`UserManagementServiceImpl.resolveGroups`).

Resolve the group first (which also enforces the access check and produces the 404), then fetch users through `userManagementApiClient.getUsers()` and filter by group **uuid** - the pattern already used by `NotificationServiceImpl.createNotificationForGroup`.

Deliberately not delegating to the auth service's `GET /auth/users?group=` filter: it matches on group **name** with a case-sensitive comparison (`UserService.GetAsync`), and it filters after paging, so results silently truncate past its 1000-row page. Filtering in Core by uuid avoids both, needs no new method on `UserManagementApiClient`, and costs the same round trip.

### User Story

As an administrator, I want the platform to return the users assigned to a group, so that the Group detail page can show me who is in it.

### Use Case

- An access or ownership review needs to see who belongs to a group. Membership is load-bearing beyond display: a member can action approval steps assigned to that `groupUuid` (`ApprovalServiceImpl` matches the caller's group uuids when approving or rejecting), and members receive internal notifications addressed to the group.

### Acceptance Criteria

- [ ] `GroupService.listGroupUsers(SecuredUUID)` returns `List<UserDto>` for the users whose auth group membership contains the group uuid.
- [ ] The method is guarded by `@ExternalAuthorization(resource = GROUP, action = MEMBERS)`; a caller without that permission is denied.
- [ ] An unknown group uuid produces `NotFoundException` (404).
- [ ] A group with no members returns an empty list, not an error.
- [ ] Matching is by group uuid, not group name.
- [ ] No new `ResourceAction` value and no change to `auth` or `auth-opa-policies`.
- [ ] Unit tests cover: members returned, empty group, unknown group, and a user belonging to several groups.
- [ ] Sonar quality gate passes with no new issues on the changed code.
- [ ] Code coverage on new and changed code is at least 80%.

klaraf755:
Closes #2210

Implements the group users endpoint declared in OmniTrustILM/interfaces#951, so the Group detail page can list the users assigned to a group (Epic: OmniTrustILM/ilm#360).

## What's here

- `GroupExternalService` / `GroupServiceImpl` - `getGroupUsers(SecuredUUID)` returning `List<NameAndUuidDto>` (uuid + username). Guarded by `@ExternalAuthorization(USER, LIST, parentResource = GROUP, parentAction = MEMBERS)` with the group uuid as `SecuredParentUUID`: users are the listed resource, the group is the scope - the nested-resource shape the other parent-guarded listings use. Two votes: `GROUP:MEMBERS` object-scoped on the group, then `USER:LIST` resource-level. Runs with `@Transactional(TxType.NOT_SUPPORTED)` so the auth round-trip does not hold a pooled connection. The group is resolved first (that is the 404), then `userManagementApiClient.getUsers()` is filtered by group uuid and projected.
- `GroupControllerImpl` - delegates, with `@AuditLogged(..., affiliatedResource = Resource.USER, operation = LIST)` matching `RoleManagementControllerImpl.getRoleUsers`.
- `groupMembersDummyMethod` removed - its only purpose was to register `GROUP:MEMBERS` for sync, which the real annotated method now does.
- `GroupServiceITest` - members returned as name+uuid; empty group; unknown group (resolved before the auth call, so no user stub); a user in several groups; denial without `GROUP:MEMBERS`; denial without `USER:LIST`; and a truncated directory (auth `totalCount` above the rows served) still returning the members it did serve.

## Notes for reviewers

- **Why `USER:LIST` on top of `GROUP:MEMBERS`, and why the method takes no `SecurityFilter`.** When either vote is denied, `ExternalAuthorizationCore.decide` falls back to the group/owner-association check, which grants on empty object uuids *if* a `SecurityFilter` parameter is present. This method has none, so `shouldSkipAuthorizationCheck` denies outright; both denial tests would flip red if one were added. `GROUP:MEMBERS` is the object-dereference gate in this codebase (`ExternalAuthorizationCore`), and it is granted object-unscoped to the `attribute-content-resolver` system role. The same name+uuid projection of users is reachable today only behind `USER:LIST` (`ResourceServiceImpl.getResourceObjects`); this endpoint now sits at that bar rather than below it.
- **Why name and uuid.** `UserDto.groups` is every group a member belongs to; returning it would let a caller authorised on one group map the membership of all the others through shared members, plus emails. The detail page needs an identity and a link.
- **Why unpaginated.** Auth's `GetUsersAsync` binds only `group` - never page or pageSize - so `/auth/users` always serves page 1 at the 1000-row `MaxPageSize`, and Core has no page 2 to ask for. An earlier revision paged the in-memory list and reported `totalItems`/`totalPages` computed from it; that presented the ceiling as a complete count. This revision returns what auth served and logs a warning when `totalCount` exceeds it. The ceiling is pre-existing and shared with `listUsers()`, group notifications and the access-control object pickers; lifting it is an auth change (bind the query DTO, filter by uuid in SQL before paging), which this PR does not attempt. The platform's own users page is the same shape: `GET /v1/users` returns a bare list that `CustomTable` pages client-side.
- **No sort.** With the whole (capped) list in one response, order cannot skip or duplicate anyone; auth returns uuid-ascending via its `SortBy` default and the FE sorts client-side.
- Both denial tests were checked to fail without their gate: the `USER:LIST` one by dropping the parent annotation, the 404 one by bypassing the group lookup.
- Depends on OmniTrustILM/interfaces#951; merge that first.


src/test/java/com/otilm/core/integration/service/GroupServiceITest.java
```diff
@@ -33,6 +39,9 @@ public class GroupServiceITest extends BaseSpringBootTest {
     @Autowired
     private GroupRepository groupRepository;
 
+    @MockitoBean
+    private UserManagementApiClient userManagementApiClient;
```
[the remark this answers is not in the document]
klaraf755: Fixed in ff1eaf4 - swapped the per-class `@MockitoBean` for `@Import(ManagementApiMocks.class)` and autowired the shared mock. `ContextSignatureGuardTest` passes.

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
```diff
@@ -131,6 +140,18 @@ public void deleteGroup(SecuredUUID uuid) throws NotFoundException {
...
 
+    @Override
+    @ExternalAuthorization(resource = Resource.GROUP, action = ResourceAction.MEMBERS)
+    public List<UserDto> getGroupUsers(SecuredUUID uuid) throws NotFoundException {
+        String groupUuid = getGroupEntity(uuid).getUuid().toString();
+        return userManagementApiClient
+                .getUsers()
+                .getData()
+                .stream()
+                .filter(user -> user.getGroups().stream().anyMatch(g -> g.getUuid().equals(groupUuid)))
```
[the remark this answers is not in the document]
klaraf755: Correct, and my PR description was wrong to claim this avoided the truncation. Paged in 20af425 over the member-scoped result. Note it does not lift the real ceiling: auth's `UsersController.GetUsersAsync` binds only `group`, never page/pageSize, so no caller can get past page 1 at the 1000-row `MaxPageSize`. Core's own `listUsers()` and `createNotificationForGroup` share that ceiling; lifting it needs an auth change, filed separately.

src/test/java/com/otilm/core/integration/service/GroupServiceITest.java
```diff
@@ -145,4 +154,62 @@ void testGetResourceObject() throws NotFoundException {
         Assertions.assertEquals(group.getUuid().toString(), nameAndUuidDto.getUuid());
         Assertions.assertEquals(group.getName(), nameAndUuidDto.getName());
     }
+
+    @Test
+    void getGroupUsersReturnsUsersAssignedToTheGroup() throws NotFoundException {
```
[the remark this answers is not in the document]
klaraf755: You were right that the harness exists - `denyResourceAccess` is in `BaseSpringBootTest`. Added in ff1eaf4; verified it catches the break by switching the annotation to `LIST`, which made it stop throwing.

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
```diff
@@ -131,6 +142,32 @@ public void deleteGroup(SecuredUUID uuid) throws NotFoundException {
         groupRepository.delete(group);
     }
 
+    @Override
+    @ExternalAuthorization(resource = Resource.GROUP, action = ResourceAction.MEMBERS)
```
ivosh: **? Question** | Security — Major if confirmed

`GroupServiceImpl.java:146-156` gates the new endpoint solely on `GROUP:MEMBERS`, yet the body it returns is the same auth payload that `UserManagementServiceImpl.java:158-160` (`listUsers()` -&gt; `userManagementApiClient.getUsers().getData()`) protects with `USER:LIST` — `UserDto` carries `username`, `firstName`, `lastName`, `email`, `description`, `enabled`, `systemUser` and the user's full `groups` list. So a principal with no USER permission at all now reads a slice of the user directory. Two reasons this deserves an explicit decision rather than the PR description's "nothing changes in auth": (1) in this codebase `GROUP:MEMBERS` is not a directory-read action, it is the object-dereference gate — `ExternalAuthorizationCore.java:224-240` uses `GROUP:MEMBERS` as the fallback that lets a group member LIST/DETAIL objects *associated* with the group, and `GroupServiceImpl.java:218-221` keeps a `groupMembersDummyMethod` whose comment says the action exists only to sync the code with auth; (2) `V202608311000__GrantRaProfileMembersToAttributeContentResolver.java:48` grants `GROUP:MEMBERS` at role level (object-unscoped) to the `attribute-content-resolver` system role purely as a dereference fail-safe, its Javadoc calling the breadth deliberate for "group-based DETAIL/LIST on any group-eligible resource" — a scope statement written before a user-listing hung off that action, so that identity silently acquires member enumeration for every group. The peer analogue `RoleManagementServiceImpl.java:179-181` (`getRoleUsers`, the endpoint this PR's audit annotation mirrors) gates on `ROLE:DETAIL`, not on a members/dereference action, and the `NotificationServiceImpl.java:102-113` pattern the PR reuses has no `@ExternalAuthorization` at all because it never reaches a wire surface.

_Suggested:_ Require a USER-resource read permission in addition to GROUP:MEMBERS (e.g. an AuthHelper USER:LIST/DETAIL check at the top of getGroupUsers), or, if MEMBERS alone is the intended gate, confirm it here and eventually record that decision in the method Javadoc together with the fact that role-level MEMBERS grants such as the attribute-content-resolver's now imply member enumeration.

ivosh: **? Question** | Security — Major if confirmed

GroupServiceImpl.java:146 gates the new `GET /v1/groups/{uuid}/users` on `GROUP:MEMBERS`. Before this PR that action had no external endpoint at all: its only carriers were `groupMembersDummyMethod` (GroupServiceImpl.java:219, a no-op that exists only to register the action) and the authorization core's group-membership fallback (`ExternalAuthorizationCore.evaluateGroupMembersPermissions`, ExternalAuthorizationCore.java:234, and `ObjectFilterAspect` line 145), where `GROUP:MEMBERS` means "the principal is a member of this group, so let it see objects assigned to it". After this PR the same grant also authorizes reading the group's member list — `UserDto` carries `email`, `firstName`, `lastName` and *each member's full list of other group memberships* — so a principal learns identities and cross-group membership it previously could not query. That matters because the grant is already handed out for the object-visibility fallback and not for user enumeration: V202608311000__GrantRaProfileMembersToAttributeContentResolver.java:48 grants `GROUP:MEMBERS` to the `attribute-content-resolver` role as a "fail-safe", its own Javadoc calling the breadth "deliberate" precisely because it is scoped to group-based DETAIL/LIST. The two peers that expose users are gated differently: `RoleManagementServiceImpl.getRoleUsers` (line 179) — the endpoint this PR names as its model — requires `ROLE:DETAIL`, and `UserManagementServiceImpl.listUsers` (line 158) requires `USER:LIST`; this endpoint requires neither `GROUP:DETAIL` nor `USER:LIST`. The PR body's "nothing changes in `auth` or `auth-opa-policies`" is true of the action catalogue but understates the capability change for existing holders of the action.

_Suggested:_ Confirm whether existing GROUP:MEMBERS holders are meant to gain member-list reads: if yes, disclose it in a release note as a permission-semantics change; if no, add USER:LIST (or GROUP:DETAIL) as a second gate on getGroupUsers.

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
```diff
@@ -131,6 +142,32 @@ public void deleteGroup(SecuredUUID uuid) throws NotFoundException {
...
+    public GroupUserResponseDto getGroupUsers(SecuredUUID uuid, PaginationRequestDto paginationRequestDto)
```
ivosh: **~ Minor** | Performance

`GroupServiceImpl` carries a class-level `@Transactional` (GroupServiceImpl.java:39-40), so the new `getGroupUsers` (GroupServiceImpl.java:147) runs inside an ambient transaction. It first touches the DB (`getGroupEntity` at GroupServiceImpl.java:150, which acquires the pooled connection) and only then makes the remote call `userManagementApiClient.getUsers()` (GroupServiceImpl.java:151-153) — up to the 1000-row auth page. The connection is therefore held from the pool for the whole auth round-trip, which is exactly what CLAUDE.md "Transactions and external calls" forbids ("Don't hold a database transaction or row lock across an HTTP call to an external system"). The established neighbour pattern for HTTP-bearing service methods in this repo is `@Transactional(propagation = Propagation.NOT_SUPPORTED)` — e.g. ComplianceServiceImpl.java:326, CertificateServiceImpl.java:1021, TspProfileBasicCredentialServiceImpl.java:89. Nothing here needs the transaction: the only DB access is a single `findByUuid` and no entity is written or lazily navigated after the call, so the method is safe to run with no ambient tx. (`NotificationServiceImpl.createNotificationForGroup`, the pattern this PR reuses for the filter, does keep the tx — but it performs local writes after the call and is not reachable from an operator-facing endpoint.)

_Suggested:_ Annotate `getGroupUsers` with `@Transactional(propagation = Propagation.NOT_SUPPORTED)` so the auth call does not run inside a transaction holding a pooled connection.

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
```diff
@@ -131,6 +142,32 @@ public void deleteGroup(SecuredUUID uuid) throws NotFoundException {
...
+                .stream()
+                .filter(user -> user.getGroups().stream().anyMatch(g -> g.getUuid().equals(groupUuid)))
+                .toList();
+
+        int itemsPerPage = paginationRequestDto.getItemsPerPage();
+        int pageNumber = paginationRequestDto.getPageNumber();
+
+        GroupUserResponseDto responseDto = new GroupUserResponseDto();
+        responseDto
+                .setUsers(members.stream().skip((long) (pageNumber - 1) * itemsPerPage).limit(itemsPerPage).toList());
```
ivosh: **~ Minor** | Correctness

GroupServiceImpl.java:163 slices `members` with `skip(...).limit(...)` in the order `userManagementApiClient.getUsers().getData()` happened to arrive, and this PR applies no sort anywhere between GroupServiceImpl.java:151 and 163. Each page request is a separate auth call, so if auth's user listing has no deterministic ORDER BY, a caller walking page 1 then page 2 can see the same member twice and miss another — the classic unstable-paging defect. Every other paged listing in Core pages over an ordered result: `SecurityFilterRepositoryImpl.createResourceObjectsQuery` always applies `query.orderBy(...)` (SecurityFilterRepositoryImpl.java:357-361) before `setFirstResult`/`setMaxResults` at 379-381, and the `PageRequest.of` peers (e.g. ApprovalProfileServiceImpl.java:77-81) page inside the DB query. This one relies on an upstream order the PR description does not claim anything about; the description discloses the 1000-row ceiling and the name/case problems with `GET /auth/users?group=`, but not ordering. If auth already returns users in a stable order, that fact is worth a one-line comment since it is a contract this endpoint now depends on.

_Suggested:_ Sort the filtered members on a stable key (e.g. `Comparator.comparing(UserDto::getUsername)` or `getUuid`) before `skip`/`limit`, or state in a comment that the auth listing order is guaranteed stable.

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
```diff
@@ -131,6 +142,32 @@ public void deleteGroup(SecuredUUID uuid) throws NotFoundException {
...
+        responseDto.setItemsPerPage(itemsPerPage);
+        responseDto.setPageNumber(pageNumber);
+        responseDto.setTotalItems((long) members.size());
+        responseDto.setTotalPages((int) Math.ceil((double) members.size() / itemsPerPage));
```
ivosh: **~ Minor** | Correctness

`GroupServiceImpl.java:167` divides by the client-supplied `itemsPerPage` with no lower bound. `PaginationRequestDto` documents `@Schema(defaultValue = "10", maximum = "1000")` but declares no `minimum`, carries no `@Min`, and `RequestValidatorHelper.revalidatePaginationRequestDto` (line 149) only substitutes a value when the field is `null` — so `GET /v1/groups/{uuid}/users?itemsPerPage=0` binds 0, survives revalidation, and `Math.ceil((double) members.size() / 0)` evaluates to `Infinity`, which `(int)` saturates to 2147483647. The response is a 200 whose `totalPages` field — documented in `PaginationResponseDto` as "Number of pages available" and marked `requiredMode = REQUIRED` — carries 2147483647 alongside an empty `users` list and a truthful `totalItems`, so the payload contradicts its own published schema rather than rejecting the request. The neighbour that does this arithmetic in one place guards it: `src/main/java/com/otilm/core/mapper/workflows/PaginationResponseMapper.java:29` writes `itemsPerPage &gt; 0 ? (int) Math.ceil(...) : 0`. (The sibling service-level computations — `ApprovalServiceImpl:450`, `NotificationProfileServiceImpl:148`, `AuditLogServiceImpl:105` — share the unguarded form, but they reach a JPA `setMaxResults`/`PageRequest` that rejects the input first, whereas this PR pages in memory with `limit(0)`, which is silently legal.) The same missing floor also means `pageNumber=0` reaches `Stream.skip(-itemsPerPage)`, answering with `ExceptionHandlingAdvice`'s opaque 400 "The request could not be processed." instead of a validation message.

_Suggested:_ Reject `itemsPerPage < 1` and `pageNumber < 1` with a `ValidationException` in `revalidatePaginationRequestDto`, or at minimum guard the division as `PaginationResponseMapper:29` does.

<sub>flagged by OpenAPI & schema fidelity + Java engineering & idioms + Correctness & races</sub>

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
ivosh: **~ Minor** | Maintainability

`GroupServiceImpl.java:147-169` (declared at `GroupExternalService.java:31`) does something a reader cannot infer from the name or the signature: it pulls **every** user from auth over HTTP (`userManagementApiClient.getUsers()`, which sends no query string), filters on the denormalized group-uuid copy carried on the auth user row, and then pages that list in memory — so `totalItems`/`totalPages` describe the set auth happened to return, not the group's real membership. Two facts make this deliberate rather than sloppy, and both exist only in this PR's description: (a) auth's `GET /auth/users?group=` is not delegated to because it matches on group *name*, case-sensitively, and filters after paging; (b) `getUsers()` is capped at auth's 1000-row `MaxPageSize`, so this endpoint pages an already-truncated set. The next reader of `GroupServiceImpl` — human or AI — will not have the PR body, and CLAUDE.md's "Writing for future AI sessions" asks precisely for "method-level Javadoc on non-obvious behaviour". The sibling methods on the same interface (`GroupExternalService.java:18-29`: `getGroup`, `deleteGroup`, `bulkDeleteGroup`) legitimately carry no Javadoc because their names say all there is to say; this one does not.

_Suggested:_ Add a short Javadoc on getGroupUsers recording that membership comes from the group uuids denormalized onto the auth user row, that filtering and paging happen in Core because auth's group filter matches by name and applies after paging, and that the totals are bounded by auth's 1000-row cap.

src/test/java/com/otilm/core/integration/service/GroupServiceITest.java
```diff
@@ -145,4 +161,120 @@ void testGetResourceObject() throws NotFoundException {
...
+    void getGroupUsersReturnsEmptyListWhenNoUserIsAssigned() throws NotFoundException {
+        stubUsers(user("22222222-2222-2222-2222-222222222222", "outsider",
+                new NameAndUuidDto("33333333-3333-3333-3333-333333333333", "otherGroup")));
+
+        Assertions.assertTrue(groupService.getGroupUsers(group.getSecuredUuid(), page(1, 10)).getUsers().isEmpty());
+    }
+
+    @Test
+    void getGroupUsersThrowsNotFoundForUnknownGroup() {
+        stubUsers(user("11111111-1111-1111-1111-111111111111", "member",
```
ivosh: **. Nitpick** | Testing

`GroupServiceITest.java:188-195` stubs a user who belongs to `group` before asserting `NotFoundException` for a different uuid, which reads as "the group is unknown even though a member exists". It cannot mean that: `GroupServiceImpl.java:149-150` validates the pagination request and then resolves the group, throwing before `userManagementApiClient.getUsers()` is ever called, so the stub at `:189-190` is never consumed (it does not fail the run only because `ManagementApiMocks` builds the mock with plain `Mockito.mock`, which is lenient about unused stubbing). Dropping it makes the test say what it actually verifies — resolution happens before the auth call — and removes a setup line a future reader may try to reason about.

_Suggested:_ Delete the stubUsers(...) setup from getGroupUsersThrowsNotFoundForUnknownGroup so the test shows the group is resolved before any auth call is made.

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
ivosh: **. Nitpick** | Maintainability

`GroupServiceImpl.java:218-221` keeps `groupMembersDummyMethod()`, whose comment states its only purpose: "Method is used just to sync MEMBERS resource action for GROUP resource with Auth service". This PR adds `@ExternalAuthorization(resource = Resource.GROUP, action = ResourceAction.MEMBERS)` on a real method of the same bean (`getGroupUsers`, `GroupServiceImpl.java:146`), and `ContextRefreshListener.handleContextRefresh` registers the pair from any `@ExternalAuthorization`-annotated declared method of a singleton bean — so the pair is now synced by `getGroupUsers` regardless. Local grep over `src/main` and `src/test` finds no caller of `groupMembersDummyMethod` (only its declaration), and org-wide `gh search code --owner OmniTrustILM --owner 3KeyCompany` for the name returns nothing, so it is unreachable placeholder code that this PR makes obsolete. The PR description does not mention it.

_Suggested:_ Delete `groupMembersDummyMethod()` now that `getGroupUsers` carries the same `GROUP:MEMBERS` annotation, or keep it and update its comment to say why the sync still needs a dedicated method.

<sub>flagged by Reference integrity + API contract & compat</sub>

src/main/java/com/otilm/core/api/web/GroupControllerImpl.java
```diff
@@ -83,4 +85,12 @@ public void deleteGroup(@LogResource(uuid = true) @PathVariable String uuid) thr
     public void bulkDeleteGroup(@LogResource(uuid = true) List<String> groupUuids) {
         groupService.bulkDeleteGroup(SecuredUUID.fromList(groupUuids));
     }
+
+    @Override
+    @AuditLogged(module = Module.CORE, resource = Resource.GROUP, affiliatedResource = Resource.USER,
+            operation = Operation.LIST)
+    public GroupUserResponseDto getGroupUsers(@LogResource(uuid = true) @PathVariable String uuid,
```
ivosh: **. Nitpick** | Documentation

`GroupControllerImpl.java:92` implements `GroupUserResponseDto getGroupUsers(String, PaginationRequestDto)`, which matches the code in the dependency (fetched via `gh pr diff 951 --repo OmniTrustILM/interfaces`: `GroupController.getGroupUsers` returning `GroupUserResponseDto extends core.scheduler.PaginationResponseDto`). The prose of that PR does not: interfaces#951's body says the endpoint returns `List&lt;UserDto&gt;` and "mirrors `RoleManagementController.getRoleUsers` exactly: same path shape, same response DTO, same error contract", while the merged contract is a paginated envelope, unlike `getRoleUsers`. Since this PR states "Depends on OmniTrustILM/interfaces#951; merge that first", a reviewer approving #951 on its description would be signing off on a shape neither repo ships.

_Suggested:_ Ask for interfaces#951's description to be corrected to the paginated `GroupUserResponseDto` before it is merged.

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
ivosh: **~ Minor** | Security

`GroupServiceImpl.java:151-163` puts the raw `UserDto` objects returned by auth into the response, and `UserDto` exposes `groups` — the name and uuid of *every* group that user belongs to — plus `email`, `firstName`, `lastName` and `description`. The authorization on this PR's endpoint is scoped to the single requested group (the `SecuredUUID` parameter is collected as an object uuid by `ExternalMethodAuthorizationManager.java:113`), so a caller holding `GROUP:MEMBERS` on group A alone learns the existence, names and uuids of groups B..N through their shared members, and their contact details, without any permission on those groups or on the USER resource. The Group detail page described in the PR needs the member identity, not the full cross-group membership graph.

_Suggested:_ Project each member onto the fields the Group detail page consumes before returning it — or at least reduce each user's `groups` to the requested group — so the response cannot disclose groups the caller is not authorized for.

ivosh (changes requested):

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
klaraf755: Agreed - `MEMBERS` is the dereference gate here, not a directory-read action. e77fdd07a adds `parentResource = USER, parentAction = LIST` to the annotation; the parent vote is resource-level (no `SecuredParentUUID`), so `USER:LIST` is checked before the group's `MEMBERS` vote. That is the bar the same projection already carries via `ResourceServiceImpl.getResourceObjects`. Covered by `getGroupUsersDeniesCallerWithoutUserListPermission`, verified to fail with the parent gate removed.

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
klaraf755: Projected to `NameAndUuidDto(uuid, username)` in e77fdd07a - the projection `listResourceObjects` already uses for users. No `groups`, email or names leave the endpoint.

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
klaraf755: Added `@Transactional(TxType.NOT_SUPPORTED)` at method level in e77fdd07a - jakarta's form, to stay with the class-level import rather than mix in Spring's.

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
klaraf755: Moot in e77fdd07a - pagination is gone and the whole (capped) list comes back in one response, so no boundary exists to skip or repeat across. Left unsorted deliberately: auth returns uuid-ascending via its `SortBy` default and the FE sorts client-side.

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
klaraf755: Gone with the pagination in e77fdd07a - no `itemsPerPage` is read.

src/test/java/com/otilm/core/integration/service/GroupServiceITest.java
klaraf755: Dropped in e77fdd07a - the test now shows the group is resolved before the auth call.

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
klaraf755: Removed in e77fdd07a.

src/main/java/com/otilm/core/api/web/GroupControllerImpl.java
klaraf755: interfaces#951's description and #950 are updated to the shape both repos now ship: `List<NameAndUuidDto>`, unpaginated.

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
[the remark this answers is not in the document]
klaraf755: Correcting my earlier reply: 20af425 paged the response, not the auth fetch, so it did not address this. "Fetch every page" is not available - `GetUsersAsync` binds only `group`, so `/auth/users` always serves page 1 at the 1000-row `MaxPageSize`. e77fdd07a drops the in-memory pagination, returns what auth served and logs a warning when `totalCount` exceeds it. The real fix is your second option - filter by uuid in auth before paging - which is an auth change outside this PR. The ceiling is shared with `listUsers()` and group notifications.

src/main/java/com/otilm/core/service/impl/GroupServiceImpl.java
klaraf755: Follow-up in 990efdf25: flipped the layout to `resource = USER, action = LIST, parentResource = GROUP, parentAction = MEMBERS` with the group uuid as `SecuredParentUUID` - users are the listed resource and the group is the scope, matching the other parent-guarded listings. Same two votes: `GROUP:MEMBERS` object-scoped on the group, `USER:LIST` resource-level. Both denial tests still pass; the fallback that could grant on empty object uuids only applies when a `SecurityFilter` parameter is present, which this method deliberately has none of.

ivosh (dismissed):

ivosh (approved):

PR #2216, opened 2026-09-04
Gate attribute filters by the caller's attribute-content permissions

### LINKED ISSUE #2195: Attribute filters apply none of the authorization gates the projection and sort paths apply
### Description

`FilterPredicatesBuilder.getAttributeFilterPredicate` correlates an attribute filter on attribute
definition identity alone. It does not apply the readability gates that
`getProjectedAttributesContent` applies when the same attribute is *projected* into a column:

- `aci.encrypted_data IS NULL`
- for `AttributeType.CUSTOM`, `ad.enabled = true` and the allowed / forbidden definition-uuid
  predicates derived from the caller's attribute content filter

The result is that a caller can filter a listing on attribute content they are not permitted to
read. Because the rows that come back are those matching the predicate, the filter behaves as an
oracle over a value the caller cannot see.

This is the same class of gap that OmniTrustILM/core#2178 closed on the **sort** path, and it was
found while fixing it. The filter is the stronger of the two: a sort reveals only the relative
order of a hidden value, whereas a filter answers direct questions about it, so a value can be
recovered far more efficiently.

The fix is not local. `getAttributeFilterPredicate` is reached through `getFiltersPredicate`, which
every listing service calls, so the caller's content filter has to be threaded down to it the way
core#2178 threaded it into `SortSpecification`. That breadth is why it was scoped out of #2178
rather than folded into it.

### Steps to Reproduce

1. Define a custom attribute and restrict its definition so a given role may not read its content
   (a disabled definition, an encrypted attribute, or one excluded by that role's attribute content
   filter).
2. Set some objects' values for that attribute and leave others unset.
3. As a user holding that role, list the resource and apply a filter on that attribute's content.
4. Compare the returned rows against the same listing with the filter removed.

### Expected Behavior

The filter is refused, or it matches nothing, on an attribute whose content the caller may not
read — the same outcome the projection path produces, where the column is simply not returned.

### Actual Behavior

The filter is applied against the stored content. The row set narrows according to a value the
caller has no permission to read, so the response distinguishes objects by that value.

### Severity

Major

### Module

Core

Jakub-Bartal-01:
Closes #2195

`FilterPredicatesBuilder.getAttributeFilterPredicate` correlated an attribute filter on attribute-definition identity alone. It applied none of the readability gates that `AttributeContent2ObjectRepository.getProjectedAttributesContent` and `AttributeColumnProjector` apply when the same attribute is projected into a listing column, so resource `LIST` access was enough to ask after a value the response would blank. The rows a filter returns are the rows that matched, so the filter answers direct questions about that content — one predicate at a time. #2178 closed the same gap on the sort path, where an ordering reveals only relative order; the filter is the stronger of the two and was scoped out of that change because the fix is not local.

## The gate

An attribute filter's `EXISTS` subquery now carries the projection's own row checks, reached through the helper the sort key already uses:

- encrypted content is skipped;
- a definition marked not visible is skipped, read from the stored definition with `jsonb_extract_path_text` at the path `AttributeDefinitionProperties.isVisible` reads, with an absent value counting as visible;
- for `CUSTOM` content, `enabled = true` plus the caller's allowed and forbidden definition uuids;
- a filter naming `SECRET` or `CODEBLOCK` content is refused outright, as `ListingSortResolver` refuses those content types for an ordering.

A positive condition on restricted content therefore matches nothing, and a negated one matches every row — which is what the projection's "no value" state implies. No row is dropped by a restriction the caller cannot see, and every object answers alike, so there is no differential to read.

The ciphertext gate is the one exception to "applies either way": it is skipped for `EMPTY` and `NOT_EMPTY`. `SearchHelper.prepareSearchForJSON` narrows an encrypted field to exactly those two operators, so excluding ciphertext rows there would leave the catalogue advertising two operators that can never hold. The permission, `enabled` and visibility gates do apply to presence, because whether an object carries a value for a withheld attribute is itself something the projection does not reveal.

## Threading the permissions

`getFiltersPredicate` is reached from every listing, so all 25 call sites now supply the caller's permissions and four services gain the engine. They arrive as a supplier, read only once a filter actually reaches attribute content: resolving them is a synchronous authorization call, and a listing filtered on properties alone — which is most of them — must not pay for one. `AttributeColumnProjector` loads them on the same terms, after establishing that a column asked for attribute content.

A listing builds its predicate twice, for the page and again for the count, so `AttributeEngine.customAttributeContentFilterOnce` returns a supplier that resolves on first use and reuses the answer, and each listing holds one across both builds. That keeps it at one resolution per request.

`AuditLogServiceImpl` takes the engine `@Lazy`: its authentication client logs through that service, so an eager injection closes a bean cycle and no context starts.

Every listing supplies the permissions rather than being trusted to know it has no attribute fields — audit logs and custom OIDs publish property-only catalogues today, but what a request names is the caller's choice, not the listing's. A call that reaches attribute content without them is refused rather than run unauthorized, the rule `getAttributeSortKey` already applies to an ordering.

## Behaviour worth documenting

A caller whose role scopes `ATTRIBUTE` `MEMBERS` now gets zero rows from a positive filter on a custom attribute they may not read, and every row from a negated one, with no error. That is the outcome this issue asks for and the one the projection already produces, but it is a visible change for such callers and belongs in the operator guide alongside the configurable-columns documentation.

## Tests

`AttributeFilterAuthorizationITest` covers the permission, `enabled`, ciphertext, visibility and withheld-content-type gates, both directions of the presence/value split, the negated-condition symmetry, and the number of authorization calls one listing makes. `FilterPredicatesBuilderITest` gains the refusal case and the two that pin when the permissions are read.

Each new case was checked against the removal of the gate it covers: dropping the readability predicates, extending the ciphertext gate to presence, resolving the permissions eagerly, and dropping the memoization each fail exactly the cases that claim them.

All four CI test groups pass locally, and `spotless:check` is clean.


src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -1421,6 +1421,30 @@ public CustomAttributeContentFilter loadCustomAttributeContentFilter() {
...
+    public Supplier<CustomAttributeContentFilter> customAttributeContentFilterOnce() {
+        return new Supplier<>() {
+
+            private CustomAttributeContentFilter resolved;
+
+            @Override
+            public CustomAttributeContentFilter get() {
+                if (resolved == null) {
+                    resolved = loadCustomAttributeContentFilter();
+                }
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Correct about the mechanism, and the transaction rule it cites is the repo's own. But this PR is not where the boundary moves, for two reasons.

The call already happens there. `DiscoveryServiceImpl.listDiscoveries` is under class-level `@Transactional` and already calls `listingSortResolver.resolve(...)` on line 208, which resolves the same permissions through the same `AttributeEngine.loadCustomAttributeContentFilter` for an attribute sort; `CertificateServiceImpl.listCertificates` does the same on line 594 and then calls `attributeColumnProjector.project(...)`, which loads them again for attribute columns. Two of the three paths that read attribute content already make this call inside the ambient transaction. This PR adds the third.

And the marginal cost is smaller than it looks: the supplier is read only once a filter actually reaches attribute content, so every property-only listing — most of them — makes no such call at all, where before the change it also made none. An attribute-filtered listing goes from zero to one, memoized across the page and count queries.

Fixing it only for the filter path would leave the transaction spanning an OPA call anyway, via the sort resolver two lines up. The real fix is a `NOT_SUPPORTED` boundary (or non-transactional orchestrators) around all three attribute-content reads in the listing methods, which touches every listing service and belongs in its own change rather than in a security fix for the filter gate.

Raising that as a follow-up. Resolving here on that basis — reopen if you would rather this PR carry the boundary change.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -1264,7 +1308,47 @@ public static <T> Expression<?> getAttributeSortKey(final CriteriaBuilder criter
...
+            .of(AttributeContentType.SECRET, AttributeContentType.CODEBLOCK);
+
+    /**
+     * Refuses a filter on content no path may read, rather than answering it. Refused for the reason
+     * {@code ListingSortResolver} refuses the same content types for an ordering: the catalogue withholds the field,
+     * and answering anyway would make the flag it withheld it with meaningless.
+     */
+    private static void requireFilterableContentType(final SearchFilterRequestDto filterDto) {
+        final AttributeFieldIdentifier identifier = AttributeFieldIdentifier.parse(filterDto.getFieldIdentifier());
+        if (identifier != null && UNFILTERABLE_CONTENT_TYPES.contains(identifier.contentType())) {
```
klaraf755: `identifier.contentType()` returns null for an unknown content-type name, and `Set.of(...).contains(null)` throws, so a filter on `environment|BOGUS` now 500s before the `does not name an attribute` validation at line 181 is reached. Moving this check after that parse fixes it and removes the second `AttributeFieldIdentifier.parse` of the same string.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -1264,7 +1308,47 @@ public static <T> Expression<?> getAttributeSortKey(final CriteriaBuilder criter
     }
 
     /**
-     * The predicates that keep a sort from reading content the same response would withhold, mirroring
+     * Content whose values a listing never reads, whatever a request asks for: a secret is rendered nowhere, and a code
+     * block is multi-line by construction. {@code AttributeColumnProjector} withholds both from a column and
+     * {@code SearchHelper} reports them undisplayable, so a filter that read them would be the one path left that does.
+     */
+    private static final Set<AttributeContentType> UNFILTERABLE_CONTENT_TYPES = Set
```
klaraf755: The catalogue does not withhold these fields for filtering. `SearchHelper.isDisplayable` only flips `displayable`; the conditions list is narrowed for ENCRYPTED alone, so a SECRET or CODEBLOCK field is still published with EMPTY/NOT_EMPTY and the FE will offer it, then get a 400 here. A presence filter on a secret reads no value, which is the same reasoning that keeps presence filters for ENCRYPTED. Either narrow the catalogue the way ENCRYPTED is, or allow EMPTY/NOT_EMPTY here.

Also, this is the third copy of the pair: `AttributeColumnProjector.WITHHELD_CONTENT_TYPES` and `isDisplayable` inline the same two. One shared constant would keep the filter gate from drifting from the projection.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -1274,12 +1358,21 @@ public static <T> Expression<?> getAttributeSortKey(final CriteriaBuilder criter
...
     private static List<Predicate> attributeReadabilityPredicates(final CriteriaBuilder criteriaBuilder,
             final Join joinContentItem, final Join joinDefinition, final AttributeType attributeType,
-            final CustomAttributeContentFilter contentFilter) {
+            final CustomAttributeContentFilter contentFilter, final boolean readsStoredValue) {
         final List<Predicate> predicates = new ArrayList<>();
-        predicates.add(criteriaBuilder.isNull(joinContentItem.get(AttributeContentItem_.encryptedData)));
+        if (readsStoredValue) {
+            predicates.add(criteriaBuilder.isNull(joinContentItem.get(AttributeContentItem_.encryptedData)));
+        }
+        predicates.add(definitionIsVisible(criteriaBuilder, joinDefinition));
```
klaraf755: This applies to DATA and META definitions too. Connectors routinely register technical attributes with `visible: false`, and the catalogue still lists them as filterable, so `EQUALS` now returns 0 rows and `EMPTY` every row with no error. Visibility on connector attributes is a display hint, not a permission. Either scope this gate to CUSTOM, or refuse the field the way SECRET is refused so the client can tell "withheld" from "no match".

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -121,16 +140,36 @@ public static <T> Predicate getFiltersPredicate(final CriteriaBuilder criteriaBu
                             .add(getPropertyFilterPredicate(criteriaBuilder, query, root, filterDto, joinedAssociations,
                                     refutedOidsOptedIn));
                 } else {
-                    predicates.add(getAttributeFilterPredicate(criteriaBuilder, query, root, filterDto));
+                    if (!contentFilterRead) {
+                        contentFilter = contentFilterSource == null ? null : contentFilterSource.get();
```
klaraf755: This resolves for every non-PROPERTY filter, but `attributeReadabilityPredicates` returns before reading the lists unless the type is CUSTOM. A META- or DATA-only filter pays the OPA round trip, and fails if OPA is down, for a result that is never used. Gating the `get()` on `getAttributeType() == CUSTOM` avoids both.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
```diff
@@ -587,7 +589,7 @@ public CertificateResponseDto listCertificates(SecurityFilter filter, Certificat
         RequestValidatorHelper.revalidateSearchRequestDto(request);
         Pageable p = PageRequest.of(request.getPageNumber() - 1, request.getItemsPerPage());
         TriFunction<Root<Certificate>, CriteriaBuilder, CriteriaQuery<?>, Predicate> additionalWhereClause = getAdditionalWhereClause(
-                request.getFilters(), request.isIncludeArchived());
+                request.getFilters(), request.isIncludeArchived(), attributeEngine.customAttributeContentFilterOnce());
```
klaraf755: "Once per request" holds only for the filter path. `listingSortResolver.resolve` on 596 and `attributeColumnProjector.project` further down each still call `loadCustomAttributeContentFilter()` eagerly, so a filtered, sorted, column-projected listing makes three OPA calls. Passing this same supplier into both (or making `SortSpecification` carry a `Supplier`) would make the claim true and keep filter and sort on one permission snapshot.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -1421,6 +1421,30 @@ public CustomAttributeContentFilter loadCustomAttributeContentFilter() {
...
+    /**
+     * The same permissions behind a supplier that resolves them on first use and then reuses the answer.
+     *
+     * <p>
+     * Resolving is a synchronous authorization call, and a listing builds its predicate more than once - once for the
+     * page and again for the count - so a caller that passed the load method itself would pay for each. Held for the
+     * span of one request by the method that starts it, and read on that request's own thread, which is what lets it
+     * memoize without synchronization.
+     */
+    public Supplier<CustomAttributeContentFilter> customAttributeContentFilterOnce() {
```
klaraf755, line 1426: `SingletonSupplier.of(this::loadCustomAttributeContentFilter)` from spring-core does this in one line and is thread-safe, which retires the caveat in the Javadoc. `getFiltersPredicate` then adds a second memo layer with `contentFilter`/`contentFilterRead`; with the supplier memoizing, those locals and the downstream null throw can collapse to one fail-fast check on a null supplier.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -1264,7 +1308,47 @@ public static <T> Expression<?> getAttributeSortKey(final CriteriaBuilder criter
...
+    /**
+     * Whether the definition behind a content row says its attribute may be shown to a user.
+     *
+     * <p>
+     * Read out of the stored definition document, because visibility is a property of the serialized attribute rather
+     * than a column of its own - which is why {@code AttributeColumnProjector} answers it in Java, after loading. A
+     * query has to answer it before, so it reads the same path with {@code jsonb_extract_path_text}. Absent counts as
+     * visible, matching {@code AttributeDefinitionProperties.isVisible}, whose properties may be null.
+     */
+    private static Predicate definitionIsVisible(final CriteriaBuilder criteriaBuilder, final Join joinDefinition) {
```
klaraf755, line 1316: Absent-means-visible makes this fail open: if the interfaces model renames or nests `visible`, the extraction returns NULL for every definition and hidden content becomes filterable again while the projector still hides it, with no test catching it. The entity already mirrors `required`, `readOnly` and `enabled` as columns; `visible` following that precedent would give projection, sort and filter one predicate. Fine as a follow-up.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -100,9 +106,22 @@ private FilterPredicatesBuilder() {
                     FilterConditionOperator.ENDS_WITH, FilterConditionOperator.MATCHES,
                     FilterConditionOperator.NOT_MATCHES);
 
+    /**
+     * The predicate a listing applies for the filters a request carries.
+     *
+     * @param contentFilterSource the caller's custom-attribute permissions, which an attribute-sourced filter is gated
```
klaraf755: The new Javadoc blocks here and on `getAttributeFilterPredicate`, `requireFilterableContentType`, `definitionIsVisible` and `customAttributeContentFilterOnce` narrate the design rationale and name `SearchHelper`, `AttributeColumnProjector` and `ListingSortResolver`. That goes stale without a compile error, and one claim already has (see the comment on `UNFILTERABLE_CONTENT_TYPES`). A behavioural sentence each is enough; the rationale is in the PR description.

src/test/java/com/otilm/core/integration/search/AttributeFilterAuthorizationITest.java
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    @Test
+    void anAttributeTheCallerMayNotReadCannotNarrowTheListing() {
+        restrictObjectAccess(Resource.ATTRIBUTE, ResourceAction.MEMBERS);
+
+        Assertions.assertEquals(List.of(), listNamesFilteredByEnvironment("alpha"));
+    }
+
+    /** The values do narrow the listing when the caller may read them, which is what the case above removes. */
```
klaraf755: Ten of the eleven tests here, and the three new ones in `FilterPredicatesBuilderITest`, carry Javadoc that restates the method name. The names already carry the behaviour; a `//` comment for non-obvious setup (the `clearInvocations` before the OPA count) is the only one worth keeping.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
Jakub-Bartal-01: Fixed. The content-type gate moved below the identifier parse and now takes the already-parsed `contentType`, so `environment|BOGUS` reaches the `does not name an attribute` validation as before, and the string is parsed once.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
Jakub-Bartal-01: Both halves, since either alone still leaves a reachable 400.

`SearchHelper.prepareSearchForJSON` now narrows the published conditions to `EMPTY`/`NOT_EMPTY` for these content types as it already did for `ENCRYPTED`, and the builder refuses only a value condition — a presence filter is answered, on the reasoning you give. The pair is one constant, `AttributeColumnProjector.WITHHELD_CONTENT_TYPES`, now public and read by `isDisplayable` and the filter gate as well.

Three cases pin it: a value filter on a secret is refused, the catalogue offers that field presence conditions alone, and a presence filter on it returns the object that carries a value.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
Jakub-Bartal-01: Scoped to CUSTOM. `definitionIsVisible` moved below the `attributeType != CUSTOM` early return, so a data or metadata definition's `visible` — a connector's display hint — no longer silences a filter the catalogue publishes. `aHiddenMetadataDefinitionStillFiltersByValue` covers it: metadata registered `visible: false` still matches by value.

The comment on the early return now names both reasons for it, the unset `enabled` column and visibility not being a permission on connector attributes.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
Jakub-Bartal-01: Fixed by pushing the supplier down rather than by gating the call site: `attributeReadabilityPredicates` takes the `Supplier` and reads it inside the CUSTOM branch, after the early return. A META- or DATA-only filter now makes no authorization call at all. `aMetadataFilterDoesNotResolveCustomAttributePermissions` asserts the counting supplier is never read.

src/main/java/com/otilm/core/service/impl/CertificateServiceImpl.java
Jakub-Bartal-01: The claim is now true. `ListingSortResolver.resolve` and `AttributeColumnProjector.project` both take the supplier, `SortSpecification` carries a `Supplier` instead of a resolved filter, and each of the seven listings that does all three passes one supplier to all three. Neither class calls `loadCustomAttributeContentFilter` any more, and the projector's `AttributeEngine` dependency is gone with it.

`oneListingResolvesTheCallersAttributePermissionsOnce` now issues one request that filters, orders and projects the same attribute, and still counts a single `ATTRIBUTE`/`MEMBERS` check.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
Jakub-Bartal-01: Both done. `customAttributeContentFilterOnce` is `SingletonSupplier.of(this::loadCustomAttributeContentFilter)` and the thread-safety caveat is gone with the anonymous class.

The second memo layer is gone too — `contentFilter`/`contentFilterRead` no longer exist and `getFiltersPredicate` passes the supplier straight down. The null check is now the single one in `getAttributeFilterPredicate`. I tried it as an eager `requireNonNull` at the top of `getFiltersPredicate` first; it breaks every test that mocks `AttributeEngine` (`CbomServiceITest`, for two) where a property-only listing has no reason to need permissions at all, so the check stays where the supplier is actually needed.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
Jakub-Bartal-01: Agreed on the failure mode — the extraction returning NULL for every definition reads as "visible" and the projector would still hide the content, with nothing failing. Tracked as #2220: mirror `visible` as an `attribute_definition` column following `required`, `readOnly` and `enabled`, with a backfill migration, and have projection, sort and filter read the column.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
Jakub-Bartal-01: Cut to a sentence each. `getFiltersPredicate`'s `@param` now says only when the supplier is read and that every listing supplies it; `getAttributeFilterPredicate`, `requireFilterableContentType`, `definitionIsVisible` and `attributeReadabilityPredicates` each keep one behavioural line. The `SearchHelper`, `AttributeColumnProjector` and `ListingSortResolver` cross-references are gone, and with them the `UNFILTERABLE_CONTENT_TYPES` claim that had already gone stale.

src/test/java/com/otilm/core/integration/search/AttributeFilterAuthorizationITest.java
Jakub-Bartal-01: Removed from all of them. The `clearInvocations` line kept its `//`, reworded to say what it is for, and the class Javadoc stays since it carries the security argument rather than a method name. The three cases in `FilterPredicatesBuilderITest` lost theirs too; the one non-obvious point among them — that only custom content is gated — is a `//` inside the metadata case.

klaraf755 (approved):

PR #2217, opened 2026-09-04
Publish the missing cryptographic-key columns in the filter-field catalogue

### LINKED ISSUE #2163: Add the missing cryptographic-key columns to the filter-field catalogue
### Description

The cryptographic-keys inventory renders three columns that the filter-field catalogue does not carry, so the column picker delivered in fe-administrator#2085 cannot offer them: `CKI_ENABLED`, `CKI_CREATED` and `CK_ASSOCIATIONS`. They appear in the platform default set but can never be picked, renamed or sorted.

Two are direct additions to `com.otilm.core.enums.FilterField`, both backed by real columns on `CryptographicKeyItem`:

- `CKI_ENABLED(Resource.CRYPTOGRAPHIC_KEY, null, null, CryptographicKeyItem_.enabled, Constants.ENABLED, SearchFieldTypeEnum.BOOLEAN)` — the same shape as `LOCATION_ENABLED` and `SECRET_ENABLED`.
- `CKI_CREATED(Resource.CRYPTOGRAPHIC_KEY, null, null, CryptographicKeyItem_.createdAt, Constants.CREATED_AT, SearchFieldTypeEnum.DATETIME)` — the same shape as `APPROVAL_CREATED_AT` and `COMMENT_CREATED_AT`.

Both also become filters on the keys inventory, which it does not have today.

`CK_ASSOCIATIONS` cannot be a `FilterField` in its current form. It is an aggregate: `CryptographicKeyItemRepository.getCountsOfAssociations` runs `COUNT(c.uuid) ... LEFT JOIN Certificate c ON c.keyUuid = ck.uuid OR c.altKeyUuid = ck.uuid`, and the count is set onto the DTO after the listing query has run. There is no mapped relation and no metamodel attribute for it, while a `FilterField` is a root resource, an optional join path and one metamodel attribute, which `FilterPredicatesBuilder` and `SortOrderBuilder` resolve to a JPA `Path`. An aggregate has no such path.

Two ways forward, and this issue should record which one is taken:

1. Add mapped `@OneToMany` relations on `CryptographicKey` for both certificate key columns and teach the predicate and sort builders to handle an aggregate field.
2. Leave the column display-only and outside the catalogue, and say so in the frontend default set so the gap is intentional rather than pending.

Also worth aligning: `Constants.ENABLED` is "Enabled" while the frontend default set labels that column "Status", so the picker will show "Enabled" once the field ships.

### Definition of Done

- `CKI_ENABLED` and `CKI_CREATED` exist in `FilterField`, and the cryptographic-key catalogue endpoint publishes them with the correct `displayable` and `sortable` flags.
- Filtering and sorting on both fields work end to end against the keys listing endpoint.
- The `CK_ASSOCIATIONS` decision is recorded on this issue; if it is to be catalogued, a follow-up issue exists for the mapped relations and the aggregate support in the builders.
- Tests cover the two new catalogue fields.

Jakub-Bartal-01:
Closes #2163

The cryptographic-keys inventory renders three columns the filter-field catalogue does not publish, so the column picker cannot offer them and no saved view can sort or rename them. Two of the three are ordinary columns and are added here; both also become filters on the keys listing, which had neither.

## The two fields

- **`CKI_ENABLED`** — `CryptographicKeyItem.enabled`, boolean, published as `Constants.ENABLED`.
- **`CKI_CREATED`** — the key's audited `created`, reached through a join on `CryptographicKeyItem_.key`, **not** the item's own `created_at` that the issue proposed. `CryptographicKeyItem.mapToDto` fills the listing DTO's `creationTime` from `key.getCreated()`, so a field on the item would filter and order the page by a timestamp no cell renders — the failure `SearchHelper.isOrderableField` exists to prevent for property fields generally. The two columns are written together in the flows that exist today, so the difference shows only once an item is added to a key that already exists. The test seeds the key's timestamp alone and leaves the item's at its default, so the item-based definition fails it.

Both use the labels the catalogue already publishes for the same concepts — `Constants.ENABLED` and `Constants.CREATED_AT` — rather than the keys page's own "Status" and "Creation Date". `CKI_STATE` publishes "State" on this resource, so a "Status" beside it would read as a second field rather than the same one. Aligning the page heading is a frontend change and is tracked in #2219.

## `CK_ASSOCIATIONS` stays out

It is an aggregate over a join predicate no mapped relation expresses — `LEFT JOIN Certificate c ON c.keyUuid = ck.uuid OR c.altKeyUuid = ck.uuid` — counted after the listing query has run, and a `FilterField` resolves to one scalar JPA path. Cataloguing it means two new mapped relations plus aggregate support in the predicate and sort builders that all sixteen `SearchRequestDto` listings run through: out of proportion to one column on one inventory, and better costed against a real request to sort by it.

Leaving it out means saving a list view from the platform default key column set still fails, since `validateColumns` rejects an identifier the catalogue does not offer. That is tracked in #2219, which records both ways forward and the `keyTableHelpers.tsx` comment that needs narrowing.

## Tests

`KeyCatalogueColumnsITest` covers the catalogue entries and their labels, and filtering and ordering by each of the two fields across the listing. It also pins the catalogue against `FilterField.getEnumsForResource(CRYPTOGRAPHIC_KEY)` by identifier set — the catalogue is assembled by naming fields one at a time, so the next field added to the enum and forgotten there is this same bug, and nothing would otherwise report it. The catalogue field readers are shared with `ColumnCatalogueFlagsITest` through `CatalogueFields`.

All four CI test groups pass locally and `spotless:check` is clean.

## Summary by CodeRabbit

* **New Features**
  * Added searchable and sortable fields for cryptographic key status and creation date.
  * Key catalogues now expose the key creation time as “Created At” and support filtering and ordering by this value.

* **Tests**
  * Added integration coverage verifying key catalogue fields, filtering, sorting, labels, and ordering.


src/main/java/com/otilm/core/enums/FilterField.java
```diff
@@ -211,6 +211,17 @@ public enum FilterField {
             SearchFieldTypeEnum.LIST, KeyUsage.class, null, false, null),
     CKI_LENGTH(Resource.CRYPTOGRAPHIC_KEY, null, null, CryptographicKeyItem_.length, "Key Size",
             SearchFieldTypeEnum.NUMBER),
+    // Labelled as the keys inventory already labels these two columns, not as the platform labels the
```
klaraf755: This rationale does not hold. `ListViewServiceImpl.catalogueOf` is enum-derived and `rejectUnknown` refused `CKI_ENABLED` before this PR, so no stored view can carry it; and the FE's `resolveColumns` keeps a view's own label override regardless. Every other boolean `enabled` field uses `Constants.ENABLED`, and `CKI_STATE` already publishes "State" on this resource, so the picker now shows "State" and "Status" side by side. If matching the page heading is the product choice, one line saying so is enough; otherwise `Constants.ENABLED` with the FE fallback heading adjusted in the follow-up.

src/main/java/com/otilm/core/enums/FilterField.java
```diff
@@ -211,6 +211,17 @@ public enum FilterField {
...
+    // same concepts elsewhere. The picker lists the catalogue's label beside headings the page renders,
+    // and a stored view refreshes its heading from the catalogue - so a different label here renames a
+    // saved view's column rather than merely reading oddly next to it.
+    CKI_ENABLED(Resource.CRYPTOGRAPHIC_KEY, null, null, CryptographicKeyItem_.enabled, Constants.STATUS,
+            SearchFieldTypeEnum.BOOLEAN),
+    // The key's creation time rather than the item's: the listing DTO reads `key.getCreated()` into `creationTime`,
```
klaraf755: Worth keeping, but as one line: the key's `created`, not the item's, because that is what the listing renders as `creationTime`. The enum carries comments only where an invariant would otherwise break, and this is the one here.

src/main/java/com/otilm/core/enums/FilterField.java
```diff
@@ -540,6 +551,9 @@ private static class Constants {
         public static final String ENABLED = "Enabled";
         public static final String STATUS = "Status";
         public static final String CREATED_AT = "Created At";
+
+        /** The keys inventory's own heading for a key's creation time; see CKI_CREATED. */
+        public static final String CREATION_DATE = "Creation Date";
```
klaraf755: Every other entry in `Constants` is used three or more times; that is what the class is for. A single-use label is inline elsewhere ("Key Size", "Token profile"), and the three other audited-created fields use `CREATED_AT`. Either inline the string on `CKI_CREATED` or use `CREATED_AT`.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -1673,6 +1673,8 @@ private List<SearchFieldDataByGroupDto> getSearchableFieldsMap() {
                                 .prepareSearch(FilterField.CKI_USAGE,
                                         Arrays.stream(KeyUsage.values()).map(KeyUsage::getCode).toList()),
                         SearchHelper.prepareSearch(FilterField.CKI_LENGTH),
+                        SearchHelper.prepareSearch(FilterField.CKI_ENABLED),
+                        SearchHelper.prepareSearch(FilterField.CKI_CREATED),
```
klaraf755, line 1677: Saving a view from the default key columns still fails after this: `KEY_COLUMNS` includes `CK_ASSOCIATIONS`, `resolveColumns` marks standard columns available, `toCreateRequest` sends them all, and `validateColumns` rejects the identifier. Leaving it out is reasonable, but that failure is the symptom the issue describes, so it wants a tracked follow-up rather than a comment on the closed issue. The FE comment in `keyTableHelpers.tsx` also goes stale.

src/test/java/com/otilm/core/integration/search/KeyCatalogueColumnsITest.java
```diff
@@ -0,0 +1,223 @@
...
+    @Test
+    void theCatalogueCarriesEveryKeyField() {
+        List<SearchFieldDataDto> published = cryptographicKeyService
+                .getSearchableFieldInformation()
+                .stream()
+                .filter(group -> group.getFilterFieldSource() == FilterFieldSource.PROPERTY)
+                .flatMap(group -> group.getSearchFieldData().stream())
+                .toList();
+
+        Assertions.assertEquals(FilterField.getEnumsForResource(Resource.CRYPTOGRAPHIC_KEY).size(), published.size());
```
klaraf755: A count pin passes when one key field is dropped and another published twice, or when a `SECRET_*` field is pasted in by mistake, since `prepareSearch` takes any `FilterField`. Set equality on identifiers, the way `CryptoAssetSearchITest` uses `containsExactlyInAnyOrder`, catches both.

src/test/java/com/otilm/core/integration/search/KeyCatalogueColumnsITest.java
```diff
@@ -0,0 +1,223 @@
...
+    void loadData() {
+        authService = new WireMockServer(WireMockPorts.AUTH_SERVICE);
+        authService.start();
+        WireMock.configureFor("localhost", authService.port());
+        authService
+                .stubFor(WireMock
+                        .get(WireMock.urlPathEqualTo("/auth/users"))
+                        .willReturn(WireMock.okJson("{\"data\": []}")));
+
+        Connector connector = new Connector();
```
klaraf755: Nothing here filters or sorts by token profile, and `seedKey` accepts null profile and instance (`ScepPkiOperationITest` already does that), so the connector/instance/profile block can go. `field(...)` and the PROPERTY flatten also exist in `ColumnCatalogueFlagsITest` in this package; a shared helper would serve both.

src/test/java/com/otilm/core/integration/search/KeyCatalogueColumnsITest.java
```diff
@@ -0,0 +1,223 @@
...
+    void theCatalogueOffersTheCreatedColumn() {
+        SearchFieldDataDto created = field(FilterField.CKI_CREATED.name()).orElseThrow();
+
+        Assertions.assertEquals(true, created.getSortable());
+        Assertions.assertEquals("Creation Date", created.getFieldLabel());
+    }
+
+    /**
+     * The catalogue is assembled by naming fields one at a time, so a field added to {@code FilterField} and forgotten
+     * there is exactly the gap this issue closes - and nothing would report it. Pinned by count, as the secret and
```
klaraf755: "This issue" names no issue and means nothing once it is closed. The test names already carry the behaviour, so the method Javadoc can go; on `seedItem` a one-line `//` about `@CreationTimestamp` not being settable is the part worth keeping. The labelling rationale is now stated here, in `FilterField` and in the PR body.

src/test/java/com/otilm/core/integration/search/KeyCatalogueColumnsITest.java
```diff
@@ -0,0 +1,223 @@
...
+    @Test
+    void theListingCanBeFilteredByEnabled() {
+        seedItem("enabled-one", true, OffsetDateTime.now().minusDays(3));
+        seedItem("enabled-two", true, OffsetDateTime.now().minusDays(2));
+        seedItem("disabled-one", false, OffsetDateTime.now().minusDays(1));
+
+        SearchFilterRequestDto filter = aPropertyFilter(FilterField.CKI_ENABLED, FilterConditionOperator.EQUALS, false);
+
+        Assertions.assertEquals(List.of("disabled-one"), listNames(filter, null));
+    }
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Added. `theListingCanBeFilteredAndOrderedByEnabled` now asserts the filter and both sort directions — `disabled-one` first ascending, last descending — so the catalogue's `sortable: true` is backed by an applied ordering.

src/test/java/com/otilm/core/integration/search/KeyCatalogueColumnsITest.java
```diff
@@ -0,0 +1,223 @@
...
+    void theListingCanBeOrderedByCreated() {
+        seedItem("oldest", true, OffsetDateTime.now().minusDays(3));
+        seedItem("newest", true, OffsetDateTime.now().minusDays(1));
+        seedItem("middle", true, OffsetDateTime.now().minusDays(2));
+
+        SearchSortRequestDto ascending = new SearchSortRequestDto(FilterFieldSource.PROPERTY,
+                FilterField.CKI_CREATED.name(), SortDirection.ASC);
+
+        Assertions.assertEquals(List.of("oldest", "middle", "newest"), listNames(null, ascending));
+    }
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Added. `theListingCanBeFilteredAndOrderedByCreated` now also filters: a `GREATER` bound returns only `newest`, and a wider one returns all three. The bound sits half a day from the nearest row so no timezone the JDBC session applies can move a row across it.

src/main/java/com/otilm/core/enums/FilterField.java
Jakub-Bartal-01: The rationale was wrong on both counts — `catalogueOf` is enum-derived and `rejectUnknown` refused `CKI_ENABLED`, so no stored view can carry it, and `resolveColumns` keeps the view's own `label` override either way. It also contradicted what I had already recorded on the issue, which argued for `Constants.ENABLED` on the grounds you give here.

`CKI_ENABLED` now publishes `Constants.ENABLED`, and the comment is gone. "State" and "Status" side by side in the picker is the deciding point: the catalogue stays consistent with itself, and aligning the page heading is a frontend change, tracked in #2219 along with the stale `keyTableHelpers.tsx` comment.

src/main/java/com/otilm/core/enums/FilterField.java
Jakub-Bartal-01: Cut to one line: `// The key's created, not the item's, because that is the value the listing renders as creationTime.`

src/main/java/com/otilm/core/enums/FilterField.java
Jakub-Bartal-01: `Constants.CREATION_DATE` is gone. `CKI_CREATED` uses `Constants.CREATED_AT`, which follows the three other audited-created fields and is the same consistency argument as the `Constants.ENABLED` thread above.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
Jakub-Bartal-01: Confirmed by reading it back: `KEY_COLUMNS` carries `CK_ASSOCIATIONS`, `resolveColumns` marks a platform default available even when the catalogue does not publish it, `toCreateRequest` sends every selected column, and `validateColumns` rejects the identifier. Saving the default set still fails.

Tracked as #2219, under the same Epic. It records both ways forward — catalogue the aggregate, or stop the save from failing — and the `keyTableHelpers.tsx` comment that needs narrowing to `CK_ASSOCIATIONS` alone. The PR body no longer claims the issue comment is the record.

src/test/java/com/otilm/core/integration/search/KeyCatalogueColumnsITest.java
Jakub-Bartal-01: Switched to set equality on identifiers, `containsExactlyInAnyOrderElementsOf` against `FilterField.getEnumsForResource(CRYPTOGRAPHIC_KEY)` mapped to names, so a dropped field, a duplicate and a pasted `SECRET_*` field each fail.

src/test/java/com/otilm/core/integration/search/KeyCatalogueColumnsITest.java
Jakub-Bartal-01: The connector, token instance and token profile are gone, along with their three repositories and the fields that held them — `seedKey(name, null, null, ...)`. `loadData` is now the auth stub alone, renamed `startAuthService`.

`field` and the PROPERTY flatten moved to a package-private `CatalogueFields` in this package, with `allFields` as well; `ColumnCatalogueFlagsITest` static-imports all three and its own copies are gone.

src/test/java/com/otilm/core/integration/search/KeyCatalogueColumnsITest.java
Jakub-Bartal-01: The method Javadoc is gone, and the `seedItem` block is one `//` about `@CreationTimestamp` not being settable and equal instants falling to the uuid tie-break. The class Javadoc stays — it says why an unpublished column cannot be picked at all, which no test name carries.

klaraf755 (approved):

PR #2218, opened 2026-09-04
Record the replies removed with a thread root in its audit data

### LINKED ISSUE #2213: Comments - Thread root deletion - Audit Logs: cascaded reply deletions leave no audit record
### Description
Deleting a comment thread root that has replies cascades to the replies, but only the root's body is written to the audit log. The replies are destroyed and no record of their content survives anywhere in the platform.

Comments are append-only by design — there is no edit operation — so the audit log is the only place the text of a deleted comment can be recovered from. Deleting a reply **directly** does produce an audit record containing that reply's body, which shows the per-comment logging path works correctly; the cascade path bypasses it.

This defeats the rationale the Epic gives for the cascade rule itself. Decision 6 on OmniTrustILM/ilm#246 states: *"Once replies exist, only the host object's owner or an `UPDATE` holder may delete it, and the delete cascades to the replies - a root author must not be able to erase other users' words. Audit-logged with the full text, so nothing is silently lost."* The rule exists to stop a root author from erasing other users' words, yet with only the root logged, those words are exactly what is lost.

For completeness: the acceptance criterion in OmniTrustILM/core#2025 reads *"the audit record carries the comment body"* in the singular, so the literal wording is satisfied. The gap is between that wording and the stated intent, and it would help to confirm which of the two is authoritative.

### Steps to Reproduce
1. Open the detail page of any **RA Profile** as a user who can post comments.
2. Post a comment in the comment panel — this creates a thread root.
3. Post two replies to that root, ideally as two different users.
4. Delete the thread root. The cascade removes both replies.
5. Open **Audit Logs** and filter on that RA Profile.
6. Count the delete entries for the operation and inspect their **Operation data**.

### Expected Behavior
- The audit log records the destroyed content of every comment removed by the operation — the root and both replies.
- Someone reading the audit log can reconstruct what text was destroyed, as Decision 6 states.

### Actual Behavior
- Exactly one audit entry is written for the whole operation.
- Its **Operation data** carries only the root's body. The two replies are not represented in it at all.
- No audit entry exists for either cascaded reply, so their text is unrecoverable.

Single entry produced by the operation (values redacted):

```json
{
  "body": "<root comment text>",
  "resource": "raProfiles",
  "createdAt": "<timestamp>",
  "authorUuid": "<uuid>",
  "objectName": "<RA profile name>",
  "objectUuid": "<uuid>",
  "commentUuid": "<root uuid>",
  "authorUsername": "<user>"
}
```

### Severity
Major

Comments cannot be edited, so the audit log is the only record of deleted text. The loss is silent, irreversible, and hits precisely the content the cascade rule was written to protect.

### Module
Core

### Attachments

_No response_

### Logs

_No response_

### Environment
ILM Core: 2.20.0-SNAPSHOT (3 Sept 2026, 16:12)
Commit: a6a6acd
Branch: main
Build time: 2026-09-03T14:12:27Z
ILM Frontend: 2.19.0 (3 Sept 2026, 14:11)
Client:
- macOS Tahoe 26.5.1
- Chrome 150.0.0.0 (arm64)

klaraf755:
Closes #2213

Deleting a thread root cascades to its replies, but only the root's body reached the audit record; the replies' text was destroyed without a trace (Epic: OmniTrustILM/ilm#246). Comments cannot be edited, so the audit record is the only place that text could have survived - which is the very rationale Decision 6 gives for the cascade rule.

## What's here

- `CommentWriter.deleteRoot` — reads the replies under the root's `FOR UPDATE` lock, so none can land between the read and the delete unrecorded, and returns them. The sole-author re-check runs over that list instead of a second query.
- `CommentServiceImpl.deleteComment` — the deletion's audit data carries the root and every cascaded reply, each with its verbatim body, author, timestamp and parent.
- `CommentDeletionData` (core, `com.otilm.core.model.comment`) — extends `CommentEventData` with `cascadedReplies`. Audit operation data is a plain `Serializable`, so the field lives on a core-side audit type and the event payload consumed by notifications never carries it. Reply deletions keep exactly the record they have today; root deletions gain one array.
- Tests — the audit ITest deletes a root with two replies and asserts both bodies in order with their parent and author; a second test pins that a reply deletion has no cascade field and a reply-less root records an empty one.

## Notes for reviewers

- On the issue's question of which wording is authoritative: the fix follows the Epic's intent (nothing silently lost) rather than the singular "comment body" in #2025's acceptance criterion.
- One audit record per operation, as for every other operation; the cascaded replies are nested in it rather than emitted as separate records.
- No interfaces change; independent of OmniTrustILM/core#2200.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved comment deletion handling when replies are removed concurrently, preventing incorrect cascade reporting.
  * Deletion audit records now accurately include replies removed with a root comment.

* **Tests**
  * Expanded coverage for root, reply, and reply-less comment deletion audit records.
  * Added concurrency coverage for simultaneous root and reply deletion.


src/main/java/com/otilm/core/service/impl/CommentServiceImpl.java
```diff
@@ -234,8 +235,14 @@ public void deleteComment(UUID uuid) throws NotFoundException {
...
         }
-        recordAuditData(baseEventData(comment, hostObject.getName()));
-        commentWriter.deleteRoot(uuid, mayCascade ? null : comment.getAuthorUuid());
+        // Comments cannot be edited, so the audit record is the only place the text of a deleted comment survives:
+        // the replies removed with the root go into it alongside the root itself
+        List<Comment> cascadedReplies = commentWriter.deleteRoot(uuid, mayCascade ? null : comment.getAuthorUuid());
+        CommentDeletionData deletionData = fillEventData(new CommentDeletionData(), comment, hostObject.getName());
+        deletionData
+                .setCascadedReplies(
+                        cascadedReplies.stream().map(reply -> baseEventData(reply, hostObject.getName())).toList());
```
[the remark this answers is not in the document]
klaraf755: Fair point on the mechanics, but the cliff predates this change: a single body is 65,536 characters, which for non-Latin text is already near 256 KB, so a directly deleted large comment has the same exposure today. The bound belongs to the audit producer rather than to comment deletion, so that records the broker rejects are never dropped silently; that is a follow-up on the producer rather than a cap on the cascade here.

src/main/java/com/otilm/core/service/writer/CommentWriter.java
```diff
@@ -64,14 +65,20 @@ public int delete(UUID uuid) {
      * caller when they may delete only because every comment in the thread is theirs; null when they hold the cascade
      * privilege and other users' replies go with the root.
      */
+    /**
+     * @return the replies removed along with the root, read under the same lock so none slips in unrecorded; the caller
+     * owes them to the audit record, which is the only place their text survives
+     */
```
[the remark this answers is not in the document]
klaraf755: Merged into the original block.

src/main/java/com/otilm/core/dao/repository/CommentRepository.java
```diff
@@ -30,6 +30,8 @@ public interface CommentRepository extends SecurityFilterRepository<Comment, UUI
      */
     Page<Comment> findByParentUuid(UUID parentUuid, Pageable pageable);
 
+    List<Comment> findByParentUuidOrderByCreatedAtAscUuidAsc(UUID parentUuid);
```
ivosh: **! Major** | Correctness

CommentRepository.java:33 performs an unlocked read before CommentWriter.java:79 deletes the root. The root's `FOR UPDATE` lock serializes concurrent reply inserts, but it does not prevent another request from deleting an existing reply after CommentWriter.java:74 reads it and before the root cascade executes; in that interleaving, both deletions succeed and the root's audit record falsely reports that it cascaded a reply actually deleted by the other operation.

_Suggested:_ Add a pessimistic write lock or equivalent `FOR UPDATE` query to the ordered reply finder and cover a concurrent reply-deletion race.

src/test/java/com/otilm/core/integration/api/web/CommentControllerAuditITest.java
```diff
@@ -142,17 +144,42 @@ void resolveAndUnresolveCarryTheThreadBody() throws NotFoundException {
...
+                .containsExactly("a reply that cascades away", "and another one");
+        assertThat(data.getCascadedReplies()).allSatisfy(reply -> {
+            assertThat(reply.getParentUuid()).isEqualTo(root.getUuid());
+            assertThat(reply.getAuthorUsername()).isNotBlank();
+        });
         assertAffiliatedHost(message);
     }
 
+    @Test
+    void deletingAReplyOrAReplylessRootRecordsNoCascade() throws NotFoundException {
```
ivosh: **. Nitpick** | Testing

CommentControllerAuditITest.java:166 combines reply deletion and replyless-root deletion in one method with separate controller calls and assertion phases. Unlike the single-mutation test at CommentControllerAuditITest.java:144, a failure in the first contract prevents the second from being exercised and reported independently.

_Suggested:_ Split this method into separate reply-deletion and replyless-root-deletion tests with independent fixtures and act/assert phases.

src/main/java/com/otilm/core/model/comment/CommentDeletionData.java
```diff
@@ -0,0 +1,20 @@
+package com.otilm.core.model.comment;
+
+import com.otilm.api.model.common.events.data.CommentEventData;
+import java.util.List;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+
+/**
+ * Audit data of a thread root's deletion. Comments cannot be edited, so the audit record is the only place the text of
```
ivosh: **. Nitpick** | Documentation

CommentDeletionData.java:10-12 packs the class purpose, editability rationale, completeness invariant, and audit-versus-event boundary into one 40-word semicolon-chained sentence. The same audit-only rationale is independently restated at CommentServiceImpl.java:309-310 and CommentWriter.java:68-69, creating three documentation surfaces that can drift while making the authoritative class contract harder to scan.

_Suggested:_ Rewrite the class Javadoc as a one-line summary followed by separate `<p><b>Completeness:</b>` and `<p><b>Audit scope:</b>` paragraphs, and remove the duplicated rationale from CommentServiceImpl and CommentWriter.

src/main/java/com/otilm/core/dao/repository/CommentRepository.java
ivosh: **~ Minor** | Performance

CommentRepository.java:33 returns every reply as a `List<Comment>`, and CommentWriter.java:74 materializes that list before deleting the root. Reply listings are otherwise paginated, no per-thread reply-count bound exists in the service or schema, and each body may be 65,536 characters; moreover, the existing `parent_uuid` index cannot satisfy the added `created_at, uuid` ordering, so the database must sort the complete thread. A sufficiently large valid thread can therefore consume unbounded database, heap, and audit-message capacity, potentially leaving the deletion committed before the single oversized audit message fails to publish.

_Suggested:_ Enforce an explicit per-thread reply cap at creation and bound this fetch to the same limit, including a safe handling path for any existing over-limit threads.

ivosh (changes requested):

src/main/java/com/otilm/core/dao/repository/CommentRepository.java
klaraf755: Taken. The replies are now read with FOR UPDATE in the same transaction, so a concurrent reply deletion waits and then finds nothing, which the service reports as 404; a concurrency test covers that interleaving.

src/test/java/com/otilm/core/integration/api/web/CommentControllerAuditITest.java
klaraf755: Split.

src/main/java/com/otilm/core/model/comment/CommentDeletionData.java
klaraf755: Trimmed: the reasoning lives on the class, the service and the writer's @return say only what they do.

src/main/java/com/otilm/core/dao/repository/CommentRepository.java
klaraf755: A per-thread cap is a product rule the Epic never set, and adding one inside an audit fix, with handling for existing threads, would outgrow this change. The sort runs over one thread narrowed by the parent index, so it is cheap at any realistic size; the message-size point is the audit-producer follow-up from the earlier thread.

ivosh: **Re-check: not addressed.**

The delta merely replaces `findByParentUuidOrderByCreatedAtAscUuidAsc` with `CommentRepository.findRepliesWithLockByParentUuid`, whose query still has no `LIMIT` and returns every reply as a `List<Comment>`; `CommentWriter.deleteRoot()` still materializes that full list, and no reply-cap creation logic or over-limit handling was added.

The added row locking addresses concurrency, not the requested bound on thread size, fetch size, or audit payload size.

ivosh (approved):

PR #2223, opened 2026-09-05
Attribute signing records to the authenticated caller

### LINKED ISSUE #2096: Timestamping: Signing records do not identify the authenticated user for Basic Auth requests
### Description

When a TSP profile uses Basic Authentication and the supplied credentials are correctly mapped to an ILM user, signing records do not capture or display that authenticated user.

The signing operation succeeds, but the user field in the resulting signing record contains only a dash (`-`).

The same missing attribution affects the signing-records dashboard, where the **Signings by user** graph groups these operations under an undefined user.

### Steps to Reproduce

1. Configure a TSP profile to use Basic Authentication.
2. Configure the Basic Auth credentials so they map to an existing ILM user.
3. Enable signing records.
4. Authenticate using the configured credentials and perform a signing operation.
5. Open the signing record created for the operation.
6. Inspect the user field.

### Expected Behavior

The signing record identifies the ILM user mapped to the Basic Auth credentials used for the signing request.

### Actual Behavior

The signing operation succeeds, but the signing record does not contain the authenticated user. The user field (requested by) displays a dash (`-`), preventing the operation from being attributed to the user who requested it.

### Severity

Major

### Module

Signing

### Screenshot

_No response_

### Logs

_No response_

### Environment

ILM 2.19.0

3keyroman:
Signing records left `requestedBy` unset, so the record detail showed a dash where the requester belongs and the per-user signing statistics grouped every operation under no user.

- `AuthHelper.getActingUserOrNull()` returns the acting user's uuid and username, or `null` for an anonymous or unauthenticated thread.
- `TimestampSigningRecordFactory` and `ContentSigningRecordFactory` populate `requestedBy` from it instead of always writing `null`.

Closes #2096

## Summary by CodeRabbit

* **New Features**
  * Signing and timestamp records now identify the authenticated requester by UUID and name when available.
  * Anonymous or unauthenticated requests remain unattributed.

* **Bug Fixes**
  * Corrected requester attribution for content-signing and timestamp records.

* **Tests**
  * Added coverage for authenticated, anonymous, and unauthenticated request scenarios.
  * Added verification that authenticated requester details are preserved across signing and timestamp record creation.

### DIFF
--- a/src/main/java/com/otilm/core/signing/tsa/TimestampSigningRecordFactory.java
+++ b/src/main/java/com/otilm/core/signing/tsa/TimestampSigningRecordFactory.java
@@ -28,9 +28,9 @@ import org.springframework.stereotype.Component;
  * {@code requestMetadataJson}, unconditionally.
  *
  * <p>
- * {@code requestedBy} is the identity on the thread that issued the token: the caller a TSP authenticator put into the
- * SecurityContext, or — for in-process issuance, which has no caller of its own — whoever requested the signature the
- * timestamp is going into. It stays unset when the issuing thread carries no identity.
+ * {@code requestedBy} is the identity on the thread that issued the token. On the TSP path a TSP authenticator put it
+ * into the SecurityContext; in-process issuance has no caller of its own, so the record names whoever requested the
+ * signature the timestamp goes into. It stays unset when the issuing thread carries no identity.
  */
 @Component
 public class TimestampSigningRecordFactory {
--- a/src/main/java/com/otilm/core/signing/tsa/TimestampSigningRecordFactory.java
+++ b/src/main/java/com/otilm/core/signing/tsa/TimestampSigningRecordFactory.java
@@ -9,6 +9,7 @@ import com.otilm.core.signing.record.SigningRecordInput;
 import com.otilm.core.signing.record.SigningRecordInputSource;
 import com.otilm.core.signing.record.TimestampTokenSerialNumbers;
 import com.otilm.core.signing.tsa.messages.TspRequest;
+import com.otilm.core.util.AuthHelper;
 import java.math.BigInteger;
 import java.time.Instant;
 import java.util.LinkedHashMap;
@@ -27,8 +28,9 @@ import org.springframework.stereotype.Component;
  * {@code requestMetadataJson}, unconditionally.
  *
  * <p>
- * {@code requestedBy} is left {@code null}: the TSP caller identity is resolved in the protocol layer and is not
- * currently threaded down to the TSA engine, and in-process issuance has no caller identity of its own.
+ * {@code requestedBy} is the identity on the thread that issued the token: the caller a TSP authenticator put into the
+ * SecurityContext, or — for in-process issuance, which has no caller of its own — whoever requested the signature the
+ * timestamp is going into. It stays unset when the issuing thread carries no identity.
  */
 @Component
 public class TimestampSigningRecordFactory {
@@ -63,7 +65,7 @@ public class TimestampSigningRecordFactory {
                 .signingProfile(signingProfile)
                 .protocol(protocol)
                 .signingTime(genTime)
-                .requestedBy(null)
+                .requestedBy(AuthHelper.getActingUserOrNull())
                 .displayName(signingProfile.name() + " #" + serialHex)
                 .requestMetadataJson(buildRequestMetadataJson(signingProfile, request, serialHex))
                 .signedDocument(encodedToken)
--- a/src/main/java/com/otilm/core/util/AuthHelper.java
+++ b/src/main/java/com/otilm/core/util/AuthHelper.java
@@ -251,6 +251,21 @@ public class AuthHelper {
         }
     }
 
+    /**
+     * The acting user's uuid and username, or {@code null} when the thread carries no identifiable user — the
+     * {@link NameAndUuidDto} counterpart of {@link #getActingUserUuidOrNull()}, for callers that persist the username
+     * next to the uuid. Absence has the same two shapes described there, and an anonymous caller is reported as no user
+     * rather than as the literal "anonymousUser" its principal carries.
+     */
+    public static NameAndUuidDto getActingUserOrNull() {
+        try {
+            NameAndUuidDto actingUser = getUserIdentification();
+            return NullUtil.parseUuidOrNull(actingUser.getUuid()) == null ? null : actingUser;
+        } catch (ValidationException | IllegalArgumentException e) {
+            return null;
+        }
+    }
+
     public static UserProfileDto getUserProfile() {
         UserProfileDto userProfileDto;
         try {
--- a/src/main/java/com/otilm/core/util/AuthHelper.java
+++ b/src/main/java/com/otilm/core/util/AuthHelper.java
@@ -244,11 +244,8 @@ public class AuthHelper {
      * users (acme, scep, cmp, localhost) are ordinary auth-service users and are carried like any other.
      */
     public static UUID getActingUserUuidOrNull() {
-        try {
-            return NullUtil.parseUuidOrNull(getUserIdentification().getUuid());
-        } catch (ValidationException | IllegalArgumentException e) {
-            return null;
-        }
+        NameAndUuidDto actingUser = getActingUserOrNull();
+        return actingUser == null ? null : UUID.fromString(actingUser.getUuid());
     }
 
     /**
--- a/src/test/java/com/otilm/core/integration/signing/tsa/TsaServiceImplITest.java
+++ b/src/test/java/com/otilm/core/integration/signing/tsa/TsaServiceImplITest.java
@@ -6,6 +6,7 @@ import com.otilm.api.interfaces.core.tsp.error.TspFailureInfo;
 import com.otilm.api.model.client.cryptography.key.KeyRequestType;
 import com.otilm.api.model.client.signing.profile.SigningProfileDto;
 import com.otilm.api.model.client.signing.profile.record.SigningRecordPersistenceMode;
+import com.otilm.api.model.common.NameAndUuidDto;
 import com.otilm.api.model.common.enums.cryptography.DigestAlgorithm;
 import com.otilm.api.model.common.enums.cryptography.KeyAlgorithm;
 import com.otilm.api.model.core.connector.v2.ConnectorDetailDto;
@@ -17,6 +18,7 @@ import com.otilm.core.dao.entity.Certificate;
 import com.otilm.core.dao.repository.signing.TspProfileRepository;
 import com.otilm.core.helpers.CertificateGeneratorHelper;
 import com.otilm.core.helpers.TestCertificateAuthority;
+import com.otilm.core.security.authn.PlatformUserDetails;
 import com.otilm.core.security.authz.SecuredParentUUID;
 import com.otilm.core.security.authz.SecuredUUID;
 import com.otilm.core.security.authz.SecurityFilter;
@@ -47,6 +49,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.core.context.SecurityContextHolder;
 
 import static com.otilm.core.signing.tsa.messages.TspRequestBuilder.aTspRequest;
 import static com.otilm.core.util.builders.ConnectorRequestDtoBuilder.aV1ConnectorRequest;
@@ -218,6 +221,14 @@ class TsaServiceImplITest extends BaseSpringBootTest {
         return signingProfile;
     }
 
+    private SigningRecordDto theOnlySigningRecord() throws Exception {
+        List<SigningRecordListDto> records = signingRecordService
+                .listSigningRecords(aSearchRequest().build(), SecurityFilter.create())
+                .getItems();
+        assertThat(records).hasSize(1);
+        return signingRecordService.getSigningRecord(SecuredUUID.fromString(records.getFirst().getUuid()));
+    }
+
     // ── processTspRequestForTspProfile ────────────────────────────────────────
 
     @Nested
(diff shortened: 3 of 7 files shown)


src/test/java/com/otilm/core/integration/signing/tsa/TsaServiceImplITest.java
```diff
@@ -336,6 +341,29 @@ void persistsSigningRecord_whenTimestampGranted() throws Exception {
             assertThat(signingRecord.getDtbs()).isNull();
         }
 
+        /**
+         * Without this the operator sees a dash where the requester belongs, and the per-user signing statistics group
+         * the operation under nobody.
+         */
+        @Test
+        void signingRecordNamesTheAuthenticatedCaller_whenTimestampGranted() throws Exception {
```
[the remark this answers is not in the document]
3keyroman: The premise here is wrong: this was not "the exact failure reported in #2096". The Basic Auth path was installing the mapped user correctly. The defect was in the record factory, whose pre-fix Javadoc said so outright — *"the TSP caller identity is resolved in the protocol layer and is not currently threaded down to the TSA engine"*. It affected Bearer and client-certificate callers identically; nothing about it was Basic-specific.

The secondary point — that this test would still pass if the Basic path stopped installing the mapped user — is true, but that regression is already covered. `TspAuthenticationFilterTest` drives the **real** `BasicPasswordAuthenticator` and `TspSecurityContextWriter` through the real filter and asserts `authenticateAsUser(mappedUser)` receives the mapped UUID, plus 401 and no call on a fingerprint mismatch.

One genuine gap did exist, so I closed it in 53a7b0f: `AuthHelper.authenticateAsUser` — the step that actually installs the resolved user — was mocked everywhere and never asserted. The chain is now covered end to end:

1. Basic credentials → `authenticateAsUser(mappedUser)` — `TspAuthenticationFilterTest`
2. `authenticateAsUser` → user in the `SecurityContext` — `AuthHelperTest` (new)
3. `SecurityContext` → `requestedBy` on the persisted record — the factory tests and `TsaServiceImplITest`

I'm not adding the HTTP-level test. It needs MockMvc plus the connector/token/certificate setup, so it forks a new Spring context signature and forces raising `ContextSignatureGuardTest.BASELINE` — which this repo's CLAUDE.md calls "a deliberate purchase — do not raise it casually". That is permanent CI cost to re-prove links that already have focused tests.

src/test/java/com/otilm/core/integration/signing/tsa/TsaServiceImplITest.java
```diff
@@ -336,6 +341,29 @@ void persistsSigningRecord_whenTimestampGranted() throws Exception {
...
+            // given
+            SigningProfileDto profile = createTimestampingSigningProfile("attributed-sp");
+            PlatformUserDetails caller = (PlatformUserDetails) SecurityContextHolder
+                    .getContext()
+                    .getAuthentication()
+                    .getPrincipal();
```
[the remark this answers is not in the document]
3keyroman: The premise here is wrong: this was not "the exact failure reported in #2096". The Basic Auth path was installing the mapped user correctly. The defect was in the record factory, whose pre-fix Javadoc said so outright — *"the TSP caller identity is resolved in the protocol layer and is not currently threaded down to the TSA engine"*. It affected Bearer and client-certificate callers identically; nothing about it was Basic-specific.

The secondary point — that this test would still pass if the Basic path stopped installing the mapped user — is true, but that regression is already covered. `TspAuthenticationFilterTest` drives the **real** `BasicPasswordAuthenticator` and `TspSecurityContextWriter` through the real filter and asserts `authenticateAsUser(mappedUser)` receives the mapped UUID, plus 401 and no call on a fingerprint mismatch.

One genuine gap did exist, so I closed it in 53a7b0f: `AuthHelper.authenticateAsUser` — the step that actually installs the resolved user — was mocked everywhere and never asserted. The chain is now covered end to end:

1. Basic credentials → `authenticateAsUser(mappedUser)` — `TspAuthenticationFilterTest`
2. `authenticateAsUser` → user in the `SecurityContext` — `AuthHelperTest` (new)
3. `SecurityContext` → `requestedBy` on the persisted record — the factory tests and `TsaServiceImplITest`

I'm not adding the HTTP-level test. It needs MockMvc plus the connector/token/certificate setup, so it forks a new Spring context signature and forces raising `ContextSignatureGuardTest.BASELINE` — which this repo's CLAUDE.md calls "a deliberate purchase — do not raise it casually". That is permanent CI cost to re-prove links that already have focused tests.

src/main/java/com/otilm/core/signing/tsa/TimestampSigningRecordFactory.java
```diff
@@ -27,8 +28,9 @@
  * {@code requestMetadataJson}, unconditionally.
  *
  * <p>
- * {@code requestedBy} is left {@code null}: the TSP caller identity is resolved in the protocol layer and is not
- * currently threaded down to the TSA engine, and in-process issuance has no caller identity of its own.
+ * {@code requestedBy} is the identity on the thread that issued the token: the caller a TSP authenticator put into the
```
ivosh: **~ Minor** | Security

This PR's new Javadoc at TimestampSigningRecordFactory.java:31-33 states the record names "the caller a TSP authenticator put into the SecurityContext", and TimestampSigningRecordFactory.java:68 now reads that thread via `AuthHelper.getActingUserOrNull()`. There is a TSP path where the filter admits the request but the thread carries no usable identity.

`TspSecurityContextWriter.authenticateAsUser` (src/main/java/com/otilm/core/security/authn/tsp/TspSecurityContextWriter.java:50-53) returns `true` whenever the user-proxy call does not throw. The codebase documents the case that misses, in `EventHandler.requireImpersonated` (src/main/java/com/otilm/core/events/EventHandler.java:372-375): "a deleted or disabled owner instead has the auth service answer 'not authenticated', which the client turns into an anonymous principal — so the call returns normally". Confirmed in `PlatformAuthenticationClient.createAuthenticationInfo` (src/main/java/com/otilm/core/security/authn/client/PlatformAuthenticationClient.java:193-201): a non-authenticated response yields `AuthenticationInfo.getAnonymousAuthenticationInfo()` rather than an exception, and `AuthHelper.authenticateAsUser` (AuthHelper.java:178-188) wraps it in a `PlatformAuthenticationToken` whose `isAuthenticated()` is hard-coded `true` (PlatformAuthenticationToken.java:16-17) with a null userUuid and ROLE_ANONYMOUS. `TspAuthenticationFilter` (TspAuthenticationFilter.java:92) then continues the chain instead of writing the 401 challenge. The sibling method in the very same class, `setFromAuthInfo` (TspSecurityContextWriter.java:35), guards exactly this with `authInfo.isAnonymous()`; the mapped-user path — reached from both the cache-hit and the cache-miss branch of `BasicPasswordAuthenticator` (BasicPasswordAuthenticator.java:82, :97) — does not.

What I could not settle from this repo: whether `authorizationEnforcer.enforce(TSP_PROFILE, TIMESTAMP, …)` in `TsaServiceImpl.processTspRequestForTspProfile` (src/main/java/com/otilm/core/signing/tsa/impl/TsaServiceImpl.java:62-63) denies a ROLE_ANONYMOUS principal — the OPA policy is not in this tree. If it denies, the cost is the misleading outcome (a generic TSP rejection rather than 401) plus a signing record that never gets written. If it allows, a Basic credential whose mapped user was disabled or deleted keeps issuing timestamps, and the "deliberately accepted, TTL-bounded trade-off" documented in `CredentialVerificationCache` (src/main/java/com/otilm/core/security/authn/client/CredentialVerificationCache.java:16-21) is not in fact TTL-bounded, because the post-TTL revalidation path lands on the same unguarded call.

_Suggested:_ Guard `TspSecurityContextWriter.authenticateAsUser` the way `EventHandler.requireImpersonated` does — assert the installed principal's uuid equals the mapped user uuid, and clear the context and return false otherwise.

src/test/java/com/otilm/core/util/AuthHelperTest.java
```diff
@@ -141,6 +144,61 @@ void getActingUserUuidOrNull_returnsTheAuthenticatedUsersUuid() {
...
+
+        assertNull(AuthHelper.getActingUserOrNull());
+    }
+
+    /**
+     * The user-proxy step every TSP authenticator ends in — a Basic credential resolves to a mapped user, and this is
+     * what puts that user where {@link AuthHelper#getActingUserOrNull()} reads it from.
+     */
+    @Test
+    void authenticateAsUser_putsTheResolvedUserInTheSecurityContext() {
```
ivosh, line 189: **~ Minor** | Documentation

src/test/java/com/otilm/core/util/AuthHelperTest.java:34-39 still declares the class as the "Contract of the scoped `runAsSystem` elevation" and closes with "The remote elevation is stubbed so these assertions test only the save/restore discipline, not the auth-service call." After this PR the class holds 11 tests and only 4 are `runAsSystem_*`; the other 7 are acting-user resolution. One of the four tests this PR adds contradicts the second sentence outright: `authenticateAsUser_putsTheResolvedUserInTheSecurityContext` (:186) mocks `PlatformAuthenticationClient` (:188-191) and asserts on the username that auth-service call returned (:199) — an auth-service call is exactly what the class doc promises is out of scope here. The scope claim was already loose before this PR (three `getActingUserUuidOrNull_*` tests sat outside it), but this PR adds four more tests and the one that makes the closing sentence untrue, so the doc now mis-describes the majority of the file — a maintainer looking for where acting-user resolution is covered will not find it from this header.

_Suggested:_ Widen the AuthHelperTest class Javadoc so it names both concerns it now covers — the scoped runAsSystem elevation and acting-user/user-proxy resolution — and drop or qualify the "not the auth-service call" sentence.

<sub>flagged by Maintainability & tests + Reference integrity</sub>

src/main/java/com/otilm/core/signing/tsa/TimestampSigningRecordFactory.java
ivosh: **. Nitpick** | Documentation

src/main/java/com/otilm/core/signing/tsa/TimestampSigningRecordFactory.java:31-33 replaces the old paragraph with a single ~40-word sentence: "{@code requestedBy} is the identity on the thread that issued the token: the caller a TSP authenticator put into the SecurityContext, or — for in-process issuance, which has no caller of its own — whoever requested the signature the timestamp is going into." Everything in it is accurate (I confirmed `TspSecurityContextWriter` does populate the context for the TSP path, and `InternalTimestampSource` runs inline on the content-signing thread) and none of it is redundant — it is just structured for a linear parser rather than a reader: three concerns (what the field holds, where the TSP path gets it, where in-process issuance gets it) fused into one block, behind a colon and a nested em-dash parenthetical, so the subject is lost before "whoever requested the signature". The sibling paragraph directly above it (:26-28) reads easily because it carries one concern. Restructuring drops nothing: a one-line summary, then one short bold-led sentence per source — e.g. "{@code requestedBy} is the identity on the thread that issued the token." / "<p><b>TSP:</b> the caller a TSP authenticator put into the SecurityContext." / "<p><b>In-process issuance:</b> it has no caller of its own, so the record names whoever requested the signature the timestamp goes into." / "<p>It stays unset when the issuing thread carries no identity."

_Suggested:_ Split the requestedBy paragraph into a one-line summary plus one short sentence per issuance path, keeping all four facts.

src/main/java/com/otilm/core/util/AuthHelper.java
```diff
@@ -251,6 +251,21 @@ public static UUID getActingUserUuidOrNull() {
...
+    /**
+     * The acting user's uuid and username, or {@code null} when the thread carries no identifiable user — the
+     * {@link NameAndUuidDto} counterpart of {@link #getActingUserUuidOrNull()}, for callers that persist the username
+     * next to the uuid. Absence has the same two shapes described there, and an anonymous caller is reported as no user
+     * rather than as the literal "anonymousUser" its principal carries.
+     */
+    public static NameAndUuidDto getActingUserOrNull() {
+        try {
+            NameAndUuidDto actingUser = getUserIdentification();
+            return NullUtil.parseUuidOrNull(actingUser.getUuid()) == null ? null : actingUser;
```
ivosh, line 260: **. Nitpick** | Maintainability

This PR adds `AuthHelper.getActingUserOrNull()` (AuthHelper.java:260-267) as the documented `NameAndUuidDto` counterpart of `getActingUserUuidOrNull()` (AuthHelper.java:246-252), but the rule that decides *what counts as no user* is now written out twice: both call `getUserIdentification()`, both run it through `NullUtil.parseUuidOrNull`, and both catch the same `ValidationException | IllegalArgumentException` pair. At AuthHelper.java:263 the parse result is discarded outright — `parseUuidOrNull` is invoked purely as a validity predicate — so the two methods only agree by construction, and a later change to the absence rule (say, also treating a blank username as no user) has to be made in both or they silently diverge, with `requestedBy` on signing records attributing an operation the uuid-only helper would have left unattributed. The Javadoc at AuthHelper.java:257 already promises they behave alike ("Absence has the same two shapes described there"), which is exactly the invariant worth expressing in code rather than in prose (*Clean Code*, DRY / single point of definition).

_Suggested:_ Have `getActingUserUuidOrNull()` delegate to `getActingUserOrNull()` (returning `UUID.fromString(actingUser.getUuid())` when non-null), so the "no identifiable user" rule is defined in exactly one place.

ivosh (dismissed):

src/main/java/com/otilm/core/util/AuthHelper.java
3keyroman: Fixed in fcad9a0c. \`getActingUserUuidOrNull()\` now delegates, so the absence rule is defined once:

```java
public static UUID getActingUserUuidOrNull() {
    NameAndUuidDto actingUser = getActingUserOrNull();
    return actingUser == null ? null : UUID.fromString(actingUser.getUuid());
}
```

`UUID.fromString` cannot throw here — `getActingUserOrNull()` only returns non-null after `parseUuidOrNull` has already parsed that string — and the discarded-predicate use of `parseUuidOrNull` you noted is gone with it. Behaviour is identical for all five inputs (no authentication, anonymous, blank uuid, malformed uuid, valid uuid); the three existing `getActingUserUuidOrNull_*` tests still pass, as do both callers' suites (932 tests across the discovery, approval and signing packages).

src/test/java/com/otilm/core/util/AuthHelperTest.java
3keyroman: Fixed in fcad9a0c. The closing sentence was outright false once this PR added a test that mocks `PlatformAuthenticationClient` and asserts on what it returned. The header now names both concerns and scopes the stubbing claim to the elevation tests only:

> Two `AuthHelper` concerns: the scoped `runAsSystem` elevation, and resolving the acting user from — and installing it into — the `SecurityContext`.
>
> The elevation tests stub the remote call, so they assert only the save/restore discipline: …

src/main/java/com/otilm/core/signing/tsa/TimestampSigningRecordFactory.java
3keyroman: Took the readability point, not the proposed structure. Four `<p>` blocks with bold labels is more markup and more lines for facts you confirmed were already accurate and non-redundant, and this file's sibling paragraphs are flowing prose. Split into three plain sentences instead, same three lines (fcad9a0c):

> `requestedBy` is the identity on the thread that issued the token. On the TSP path a TSP authenticator put it into the SecurityContext; in-process issuance has no caller of its own, so the record names whoever requested the signature the timestamp goes into. It stays unset when the issuing thread carries no identity.

All four facts survive, and the subject no longer has to carry a nested em-dash aside.

lubomirw (approved):

PR #2225, opened 2026-09-05
Keep the signing volume history when signing records are deleted

### LINKED ISSUE #2097: Timestamping: Deleting signing records alters historical data in the Signings over time dashboard graph
### Description

The signing-records dashboard contains a **Signings over time** graph showing the number of signatures created during each time period.

The graph displays correct values while the underlying signing records exist. When signing records are deleted, their corresponding counts also disappear from the graph. Deleting detailed records therefore rewrites the displayed signing history.

Historical signing counts should be immutable and must remain available even when individual signing records are removed.

### Steps to Reproduce

1. Enable signing records.
2. Perform several signing operations.
3. Open the signing-records dashboard.
4. Verify that the operations appear in the **Signings over time** graph.
5. Delete the corresponding signing records.
6. Return to or refresh the dashboard.
7. Inspect the same time period in the graph.

### Expected Behavior

The **Signings over time** graph continues to show the number of signing operations performed during the affected time period.

Deleting individual signing records must not alter previously recorded historical signing counts.

### Actual Behavior

After the signing records are deleted, their counts disappear from the **Signings over time** graph. The dashboard therefore indicates that those signing operations never occurred.

### Severity

Major

### Module

None

### Screenshot

_No response_

### Logs

_No response_

### Environment

ILM 2.19.0

3keyroman:
Closes #2097

- Add `signing_record_volume`, an append-only count of signings per signing profile per UTC hour, with no foreign key to `signing_profile` so the counts outlive the profile.
- Fold signing records into that table in the same statement that deletes them, claiming the rows first so two delete paths reaching one record cannot count it twice.
- Read the signing volume series and the 24h and 7d counts as the retained records plus that history; the retained total and the breakdown maps continue to describe what is currently retained.
- Read the statistics under one snapshot, and open both window halves on an hour boundary, so deleting a record cannot change either count.
- Add `sumGroupedUsingSecurityFilter` to `SecurityFilterRepository` for the grouped read of the new table.

Schema descriptions for the three changed figures are in OmniTrustILM/interfaces#957; merge that first so the published spec matches.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -92,7 +92,7 @@ Capture entry state before the external call (`final State entryState = entity.g
 
 ## Historical figures must not be derived from deletable rows
 
-A dashboard series, counter, or report that aggregates a table rows get deleted from is rewritten by every deletion — and retention sweeps and delete-after-retrieval mean history changes with nobody touching it. Derive such a figure from a separate append-only aggregate instead, and fold rows into it in the **same statement** that deletes them:
+A dashboard series, counter, or report that aggregates a table whose rows get deleted is rewritten by every one of those deletions — and retention sweeps and delete-after-retrieval mean the history changes with nobody touching it. Derive such a figure from a separate append-only aggregate instead, and fold rows into it in the **same statement** that deletes them:
 
 ```sql
 WITH victim AS MATERIALIZED (SELECT uuid, ... FROM detail WHERE ... LIMIT :limit)
@@ -105,7 +105,7 @@ DELETE FROM detail WHERE uuid IN (SELECT uuid FROM victim)
 
 PostgreSQL runs a data-modifying CTE exactly once and to completion, and every part reads the same snapshot of the materialized `victim` rows, so the roll-up and the delete cover the identical set and commit together: no double counting, no gap, and nothing added to the write path. The statement's own result is still the DELETE's row count, so batch loops keep working. Native, so `@Modifying(flushAutomatically = true)` — Hibernate cannot auto-flush for SQL it does not parse.
 
-**Claim the rows.** Without a locking clause a second deleter sees the same rows in its own snapshot, folds them in too, and then deletes nothing — one row counted twice. Which clause depends on the caller's contract: a keyed delete **waits** (`FOR UPDATE`), because its caller is told the row is gone and skipping would report success for a row the winner might yet roll back; a batch **skips** (`FOR UPDATE ... SKIP LOCKED`), because it has no per-record contract and waiting lets two sweeps picking overlapping victims in different index orders deadlock. Lock only the table you delete from (`FOR UPDATE OF sr`), never the ones you join for the predicate.
+**Claim the rows, and fold them in in key order.** Without a locking clause a second deleter sees the same rows in its own snapshot, folds them in too, and then deletes nothing — one row counted twice. Without an `ORDER BY` on the grouped upsert, two writers carrying rows for the same aggregate keys can take those locks in opposite orders and deadlock. Which clause depends on the caller's contract: a keyed delete **waits** (`FOR UPDATE`), because its caller is told the row is gone and skipping would report success for a row the winner might yet roll back; a batch **skips** (`FOR UPDATE ... SKIP LOCKED`), because it has no per-record contract and waiting lets two sweeps picking overlapping victims in different index orders deadlock. Lock only the table you delete from (`FOR UPDATE OF sr`), never the ones you join for the predicate.
 
 Assemble the figure under one snapshot — `@Transactional(readOnly = true, isolation = Isolation.REPEATABLE_READ)`. Read across the two tables in separate READ COMMITTED statements and a row deleted between them is counted as both live and history.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -95,9 +95,11 @@ Capture entry state before the external call (`final State entryState = entity.g
 A dashboard series, counter, or report that aggregates a table whose rows get deleted is rewritten by every one of those deletions — and retention sweeps and delete-after-retrieval mean the history changes with nobody touching it. Derive such a figure from a separate append-only aggregate instead, and fold rows into it in the **same statement** that deletes them:
 
 ```sql
-WITH victim AS MATERIALIZED (SELECT uuid, ... FROM detail WHERE ... LIMIT :limit)
+WITH victim AS MATERIALIZED (
+    SELECT uuid, ... FROM detail WHERE ... LIMIT :limit FOR UPDATE OF detail [SKIP LOCKED]
+)
 , rolled AS (
-    INSERT INTO aggregate AS a (...) SELECT ... FROM victim GROUP BY ...
+    INSERT INTO aggregate AS a (...) SELECT ... FROM victim GROUP BY ... ORDER BY <aggregate key>
     ON CONFLICT (...) DO UPDATE SET row_count = a.row_count + EXCLUDED.row_count
 )
 DELETE FROM detail WHERE uuid IN (SELECT uuid FROM victim)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -90,6 +90,29 @@ try {
 
 Capture entry state before the external call (`final State entryState = entity.getState();`) so the restoration path has something to restore *to*.
 
+## Historical figures must not be derived from deletable rows
+
+A dashboard series, counter, or report that aggregates a table rows get deleted from is rewritten by every deletion — and retention sweeps and delete-after-retrieval mean history changes with nobody touching it. Derive such a figure from a separate append-only aggregate instead, and fold rows into it in the **same statement** that deletes them:
+
+```sql
+WITH victim AS MATERIALIZED (SELECT uuid, ... FROM detail WHERE ... LIMIT :limit)
+, rolled AS (
+    INSERT INTO aggregate AS a (...) SELECT ... FROM victim GROUP BY ...
+    ON CONFLICT (...) DO UPDATE SET row_count = a.row_count + EXCLUDED.row_count
+)
+DELETE FROM detail WHERE uuid IN (SELECT uuid FROM victim)
+```
+
+PostgreSQL runs a data-modifying CTE exactly once and to completion, and every part reads the same snapshot of the materialized `victim` rows, so the roll-up and the delete cover the identical set and commit together: no double counting, no gap, and nothing added to the write path. The statement's own result is still the DELETE's row count, so batch loops keep working. Native, so `@Modifying(flushAutomatically = true)` — Hibernate cannot auto-flush for SQL it does not parse.
+
+**Claim the rows.** Without a locking clause a second deleter sees the same rows in its own snapshot, folds them in too, and then deletes nothing — one row counted twice. Which clause depends on the caller's contract: a keyed delete **waits** (`FOR UPDATE`), because its caller is told the row is gone and skipping would report success for a row the winner might yet roll back; a batch **skips** (`FOR UPDATE ... SKIP LOCKED`), because it has no per-record contract and waiting lets two sweeps picking overlapping victims in different index orders deadlock. Lock only the table you delete from (`FOR UPDATE OF sr`), never the ones you join for the predicate.
+
+Assemble the figure under one snapshot — `@Transactional(readOnly = true, isolation = Isolation.REPEATABLE_READ)`. Read across the two tables in separate READ COMMITTED statements and a row deleted between them is counted as both live and history.
+
+The aggregate keeps the column its detail rows were access-controlled by, and **no FK to it**: the parent is usually deletable once its detail rows are gone, and the aggregate has to outlive it. Read it back through the parent half of the security filter only — a rolled-up row has no detail identity left for a row-level filter to match on.
+
+Align the live half's window to the aggregate's bucket boundary, not to the exact cutoff. Against an exact cutoff a row inside the first bucket but before the cutoff counts for nothing while live and for one once deleted — deleting it *raises* the figure. Floored on both sides the two are exactly equivalent, and deletion becomes invisible. Say on the API field that the window opens on a bucket boundary.
+
 ## Transactional boundaries live on services, not repositories
 
 Repositories in `com.otilm.core.dao.repository.*` must not carry `@Transactional`. Every guarded write uses a bean-pair: an **orchestrator** (no class-level `@Transactional`; HTTP-bearing methods use `NOT_SUPPORTED` when needed to avoid holding locks across external calls) and a **writer** bean (suffix `Writer`) whose short `@Transactional`(REQUIRED) methods each execute one `@Modifying @Query`. Two beans is mandatory — Spring proxy self-invocation silently skips advice. Writers use REQUIRED so they compose: joining an ambient tx if present, starting one otherwise. `@Modifying @Query` bypasses JPA dirty-checking; set audit columns (`c.updated = CURRENT_TIMESTAMP`) in the SQL.
--- a/src/main/java/com/otilm/core/dao/entity/signing/SigningRecordVolume.java
+++ b/src/main/java/com/otilm/core/dao/entity/signing/SigningRecordVolume.java
@@ -1,0 +1,54 @@
+package com.otilm.core.dao.entity.signing;
+
+import com.otilm.core.dao.entity.UniquelyIdentified;
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.Table;
+import jakarta.persistence.UniqueConstraint;
+import java.time.Instant;
+import java.util.UUID;
+import lombok.Getter;
+import lombok.Setter;
+
+/**
+ * How many signings a signing profile performed in one UTC hour, kept after the {@link SigningRecord} rows themselves
+ * are gone. A record proves what a signing contained; this row preserves that the signing happened, which has to
+ * survive the record being deleted by an operator, by retention, or by delete-after-retrieval.
+ *
+ * <p>
+ * Rows are written only by the roll-up-then-delete statements on
+ * {@link com.otilm.core.dao.repository.signing.SigningRecordRepository}: a bucket is created or incremented in the same
+ * statement that removes the records it accounts for, so a count can neither double-count nor drift. Nothing amends a
+ * bucket afterwards and nothing deletes one, which is what makes the history immutable.
+ */
+@Getter
+@Setter
+@Entity
+@Table(name = "signing_record_volume", uniqueConstraints = @UniqueConstraint(name = "uq_signing_record_volume_bucket",
+        columnNames = {"signing_profile_uuid", "bucket_start"}))
+public class SigningRecordVolume extends UniquelyIdentified {
+
+    /**
+     * The signing profile the signings were performed under. Deliberately not a mapped association: a profile becomes
+     * deletable once its records are gone and these counts outlive it, so the reference would be unsatisfiable. The
+     * column stays the access-control anchor — signing history is scoped by signing-profile access.
+     */
+    @Column(name = "signing_profile_uuid", nullable = false)
+    private UUID signingProfileUuid;
+
+    @Column(name = "bucket_start", nullable = false)
+    private Instant bucketStart;
+
+    @Column(name = "signing_count", nullable = false)
+    private long signingCount;
+
+    @Override
+    public boolean equals(Object o) {
+        return super.equals(o);
+    }
+
+    @Override
+    public int hashCode() {
+        return super.hashCode();
+    }
+}
--- a/src/main/java/com/otilm/core/dao/repository/SecurityFilterRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/SecurityFilterRepository.java
@@ -70,6 +70,14 @@ public interface SecurityFilterRepository<T, ID> extends JpaRepository<T, ID> {
             SingularAttribute<?, ?> groupBy, BiFunction<Root<T>, CriteriaBuilder, Expression<?>> groupByExpression,
             TriFunction<Root<T>, CriteriaBuilder, CriteriaQuery<?>, Predicate> additionalWhereClause);
 
+    /**
+     * Grouped total of an already-aggregated column, for entities whose rows each carry a count rather than being one.
+     * Same grouping and security filtering as the variant above, only summing {@code sumOf} instead of counting rows.
+     */
+    Map<String, Long> sumGroupedUsingSecurityFilter(SecurityFilter filter, SingularAttribute<T, Long> sumOf,
+            BiFunction<Root<T>, CriteriaBuilder, Expression<?>> groupByExpression,
+            TriFunction<Root<T>, CriteriaBuilder, CriteriaQuery<?>, Predicate> additionalWhereClause);
+
     Long countUsingSecurityFilter(SecurityFilter filter);
 
     Long countUsingSecurityFilter(SecurityFilter filter,
--- a/src/main/java/com/otilm/core/dao/repository/SecurityFilterRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/SecurityFilterRepository.java
@@ -72,7 +72,9 @@ public interface SecurityFilterRepository<T, ID> extends JpaRepository<T, ID> {
 
     /**
      * Grouped total of an already-aggregated column, for entities whose rows each carry a count rather than being one.
-     * Same grouping and security filtering as the variant above, only summing {@code sumOf} instead of counting rows.
+     * {@code groupByExpression} is required here, where the grouped count above also accepts a join and attribute.
+     * Unlike that count, this sum carries no DISTINCT: correct only where no predicate can duplicate the root row,
+     * since a join that fans one row out multiplies the count that row carries.
      */
     Map<String, Long> sumGroupedUsingSecurityFilter(SecurityFilter filter, SingularAttribute<T, Long> sumOf,
             BiFunction<Root<T>, CriteriaBuilder, Expression<?>> groupByExpression,
--- a/src/main/java/com/otilm/core/dao/repository/SecurityFilterRepositoryImpl.java
+++ b/src/main/java/com/otilm/core/dao/repository/SecurityFilterRepositoryImpl.java
@@ -251,19 +251,41 @@ public class SecurityFilterRepositoryImpl<T, ID> extends SimpleJpaRepository<T, 
     public Map<String, Long> countGroupedUsingSecurityFilter(SecurityFilter filter, Attribute<?, ?> join,
             SingularAttribute<?, ?> groupBy, BiFunction<Root<T>, CriteriaBuilder, Expression<?>> groupByExpression,
             TriFunction<Root<T>, CriteriaBuilder, CriteriaQuery<?>, Predicate> additionalWhereClause) {
+        BiFunction<Root<T>, CriteriaBuilder, Expression<?>> grouping = groupByExpression != null
+                ? groupByExpression
+                : groupingByAttribute(join, groupBy);
+        return aggregateGroupedUsingSecurityFilter(filter, grouping, (root, cb) -> cb.countDistinct(root),
+                additionalWhereClause);
+    }
+
+    /** Groups by {@code groupBy}, read off the root or off a left join to {@code join} when one is given. */
+    private BiFunction<Root<T>, CriteriaBuilder, Expression<?>> groupingByAttribute(Attribute<?, ?> join,
+            SingularAttribute<?, ?> groupBy) {
+        return (root, cb) -> {
+            From<?, ?> from = join == null ? root : root.join(join.getName(), JoinType.LEFT);
+            return from.get(groupBy.getName());
+        };
+    }
+
+    @Override
+    public Map<String, Long> sumGroupedUsingSecurityFilter(SecurityFilter filter, SingularAttribute<T, Long> sumOf,
+            BiFunction<Root<T>, CriteriaBuilder, Expression<?>> groupByExpression,
+            TriFunction<Root<T>, CriteriaBuilder, CriteriaQuery<?>, Predicate> additionalWhereClause) {
+        return aggregateGroupedUsingSecurityFilter(filter, groupByExpression, (root, cb) -> cb.sum(root.get(sumOf)),
+                additionalWhereClause);
+    }
+
+    private Map<String, Long> aggregateGroupedUsingSecurityFilter(SecurityFilter filter,
+            BiFunction<Root<T>, CriteriaBuilder, Expression<?>> groupByExpression,
+            BiFunction<Root<T>, CriteriaBuilder, Expression<? extends Number>> aggregation,
+            TriFunction<Root<T>, CriteriaBuilder, CriteriaQuery<?>, Predicate> additionalWhereClause) {
         final Class<T> entity = this.entityInformation.getJavaType();
         final CriteriaBuilder cb = entityManager.getCriteriaBuilder();
         CriteriaQuery<AggregateResultDto> cr = cb.createQuery(AggregateResultDto.class);
 
         final Root<T> root = cr.from(entity);
-        Expression<?> groupBySelection;
-        if (groupByExpression != null) {
-            groupBySelection = groupByExpression.apply(root, cb);
-        } else {
-            From from = join == null ? root : root.join(join.getName(), JoinType.LEFT);
-            groupBySelection = from.get(groupBy);
-        }
-        cr.multiselect(groupBySelection, cb.countDistinct(root));
+        Expression<?> groupBySelection = groupByExpression.apply(root, cb);
+        cr.multiselect(groupBySelection, aggregation.apply(root, cb));
         cr.groupBy(groupBySelection);
 
         final List<Predicate> predicates = getPredicates(filter, additionalWhereClause, root, cb, cr);
--- a/src/main/java/com/otilm/core/dao/repository/SecurityFilterRepositoryImpl.java
+++ b/src/main/java/com/otilm/core/dao/repository/SecurityFilterRepositoryImpl.java
@@ -271,6 +271,7 @@ public class SecurityFilterRepositoryImpl<T, ID> extends SimpleJpaRepository<T, 
     public Map<String, Long> sumGroupedUsingSecurityFilter(SecurityFilter filter, SingularAttribute<T, Long> sumOf,
             BiFunction<Root<T>, CriteriaBuilder, Expression<?>> groupByExpression,
             TriFunction<Root<T>, CriteriaBuilder, CriteriaQuery<?>, Predicate> additionalWhereClause) {
+        Objects.requireNonNull(groupByExpression, "groupByExpression");
         return aggregateGroupedUsingSecurityFilter(filter, groupByExpression, (root, cb) -> cb.sum(root.get(sumOf)),
                 additionalWhereClause);
     }
--- a/src/main/java/com/otilm/core/dao/repository/signing/SigningRecordRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/signing/SigningRecordRepository.java
@@ -8,45 +8,53 @@ import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.query.Param;
 import org.springframework.stereotype.Repository;
 
+import static com.otilm.core.dao.repository.signing.SigningRecordRollupSql.CLAIM_VICTIMS;
+import static com.otilm.core.dao.repository.signing.SigningRecordRollupSql.ROLL_UP_THEN_DELETE;
+import static com.otilm.core.dao.repository.signing.SigningRecordRollupSql.SKIP_CONTENDED_ROWS;
+import static com.otilm.core.dao.repository.signing.SigningRecordRollupSql.WAIT_FOR_CONTENDED_ROWS;
+
+/**
+ * Every statement here that removes signing records rolls them into {@code signing_record_volume} first; see
+ * {@link SigningRecordRollupSql}. They are native because a data-modifying CTE has no JPQL equivalent, and native
+ * statements are opaque to Hibernate's auto-flush, hence {@code flushAutomatically}.
+ */
 @Repository
 public interface SigningRecordRepository extends SecurityFilterRepository<SigningRecord, UUID> {
     boolean existsBySigningProfileUuidAndSigningProfileVersion(UUID signingProfileUuid, int version);
 
     boolean existsBySigningProfileUuid(UUID signingProfileUuid);
 
-    @Modifying
-    @Query("DELETE FROM SigningRecord sr WHERE sr.uuid = :uuid")
+    @Modifying(flushAutomatically = true)
+    @Query(value = CLAIM_VICTIMS + """
+            SELECT sr.uuid, sr.signing_profile_uuid, sr.signing_time
+            FROM {h-schema}signing_record sr
+            WHERE sr.uuid = :uuid
+            """ + WAIT_FOR_CONTENDED_ROWS + ROLL_UP_THEN_DELETE, nativeQuery = true)
     int deleteByUuid(@Param("uuid") UUID uuid);
 
-    @Modifying
-    @Query(value = """
-            DELETE FROM {h-schema}signing_record
-            WHERE ctid IN (
-                SELECT sr.ctid
-                FROM {h-schema}signing_record sr
-                JOIN {h-schema}signing_profile_version spv
-                  ON sr.signing_profile_uuid = spv.signing_profile_uuid
-                 AND sr.signing_profile_version = spv.version
-                WHERE spv.retention_days IS NOT NULL
-                  AND sr.signing_time < NOW() - make_interval(days => spv.retention_days)
-                LIMIT :limit
-            )
-            """, nativeQuery = true)
+    @Modifying(flushAutomatically = true)
+    @Query(value = CLAIM_VICTIMS + """
+            SELECT sr.uuid, sr.signing_profile_uuid, sr.signing_time
+            FROM {h-schema}signing_record sr
+            JOIN {h-schema}signing_profile_version spv
+              ON sr.signing_profile_uuid = spv.signing_profile_uuid
+             AND sr.signing_profile_version = spv.version
+            WHERE spv.retention_days IS NOT NULL
+              AND sr.signing_time < NOW() - make_interval(days => spv.retention_days)
+            LIMIT :limit
+            """ + SKIP_CONTENDED_ROWS + ROLL_UP_THEN_DELETE, nativeQuery = true)
     int deleteExpiredByRetention(@Param("limit") int limit);
 
-    @Modifying
-    @Query(value = """
-            DELETE FROM {h-schema}signing_record
-            WHERE ctid IN (
-                SELECT sr.ctid
-                FROM {h-schema}signing_record sr
-                JOIN {h-schema}signing_profile_version spv
-                  ON sr.signing_profile_uuid = spv.signing_profile_uuid
-                 AND sr.signing_profile_version = spv.version
-                WHERE spv.delete_after_retrieval = true
-                  AND sr.signed_document_retrieved_at IS NOT NULL
-                LIMIT :limit
-            )
-            """, nativeQuery = true)
+    @Modifying(flushAutomatically = true)
+    @Query(value = CLAIM_VICTIMS + """
+            SELECT sr.uuid, sr.signing_profile_uuid, sr.signing_time
+            FROM {h-schema}signing_record sr
+            JOIN {h-schema}signing_profile_version spv
+              ON sr.signing_profile_uuid = spv.signing_profile_uuid
+             AND sr.signing_profile_version = spv.version
+            WHERE spv.delete_after_retrieval = true
+              AND sr.signed_document_retrieved_at IS NOT NULL
+            LIMIT :limit
+            """ + SKIP_CONTENDED_ROWS + ROLL_UP_THEN_DELETE, nativeQuery = true)
     int deleteRetrievedAndFlagged(@Param("limit") int limit);
 }
--- a/src/main/java/com/otilm/core/dao/repository/signing/SigningRecordRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/signing/SigningRecordRepository.java
@@ -8,7 +8,7 @@ import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.query.Param;
 import org.springframework.stereotype.Repository;
 
-import static com.otilm.core.dao.repository.signing.SigningRecordRollupSql.CLAIM_VICTIMS;
+import static com.otilm.core.dao.repository.signing.SigningRecordRollupSql.OPEN_VICTIM_CTE;
 import static com.otilm.core.dao.repository.signing.SigningRecordRollupSql.ROLL_UP_THEN_DELETE;
 import static com.otilm.core.dao.repository.signing.SigningRecordRollupSql.SKIP_CONTENDED_ROWS;
 import static com.otilm.core.dao.repository.signing.SigningRecordRollupSql.WAIT_FOR_CONTENDED_ROWS;
@@ -25,7 +25,7 @@ public interface SigningRecordRepository extends SecurityFilterRepository<Signin
     boolean existsBySigningProfileUuid(UUID signingProfileUuid);
 
     @Modifying(flushAutomatically = true)
-    @Query(value = CLAIM_VICTIMS + """
+    @Query(value = OPEN_VICTIM_CTE + """
             SELECT sr.uuid, sr.signing_profile_uuid, sr.signing_time
             FROM {h-schema}signing_record sr
             WHERE sr.uuid = :uuid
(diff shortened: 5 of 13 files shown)


src/main/java/com/otilm/core/dao/repository/signing/SigningRecordRollupSql.java
```diff
@@ -0,0 +1,70 @@
...
+                SELECT gen_random_uuid(),
+                       victim.signing_profile_uuid,
+                       date_trunc('hour', victim.signing_time AT TIME ZONE 'UTC') AT TIME ZONE 'UTC',
+                       count(*)
+                FROM victim
+                GROUP BY victim.signing_profile_uuid,
+                         date_trunc('hour', victim.signing_time AT TIME ZONE 'UTC') AT TIME ZONE 'UTC'
+                ON CONFLICT (signing_profile_uuid, bucket_start)
+                DO UPDATE SET signing_count = srv.signing_count + EXCLUDED.signing_count
+            )
```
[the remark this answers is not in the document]
3keyroman: Fixed in 7a981539d. The grouped upsert now feeds from a named `bucket` CTE and inserts `ORDER BY signing_profile_uuid, bucket_start`, so two sweeps carrying rows for the same profile and hour take the bucket locks in the same order.

Added `deleteExpiredByRetention_rollsUpEveryBucketOneBatchSpans` — one batch spanning two profiles across two hours. Nothing previously covered the grouped upsert with more than one bucket in a single statement.

CLAUDE.md
```diff
@@ -90,6 +90,29 @@ try {
 
 Capture entry state before the external call (`final State entryState = entity.getState();`) so the restoration path has something to restore *to*.
 
+## Historical figures must not be derived from deletable rows
+
+A dashboard series, counter, or report that aggregates a table rows get deleted from is rewritten by every deletion — and retention sweeps and delete-after-retrieval mean history changes with nobody touching it. Derive such a figure from a separate append-only aggregate instead, and fold rows into it in the **same statement** that deletes them:
```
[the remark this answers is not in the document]
3keyroman: Fixed in 7a981539d — rewritten as "aggregates a table whose rows get deleted".

src/main/java/com/otilm/core/dao/repository/signing/SigningRecordRepository.java
```diff
@@ -8,45 +8,53 @@
...
+    @Query(value = CLAIM_VICTIMS + """
+            SELECT sr.uuid, sr.signing_profile_uuid, sr.signing_time
+            FROM {h-schema}signing_record sr
+            JOIN {h-schema}signing_profile_version spv
+              ON sr.signing_profile_uuid = spv.signing_profile_uuid
+             AND sr.signing_profile_version = spv.version
+            WHERE spv.retention_days IS NOT NULL
+              AND sr.signing_time < NOW() - make_interval(days => spv.retention_days)
+            LIMIT :limit
+            """ + SKIP_CONTENDED_ROWS + ROLL_UP_THEN_DELETE, nativeQuery = true)
```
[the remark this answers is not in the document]
3keyroman: Fixed in 7a981539d, arrived at independently from the same finding by Copilot. The grouped rows now come from a named `bucket` CTE and the upsert inserts `ORDER BY bucket.signing_profile_uuid, bucket.bucket_start`, so concurrent sweeps take the `signing_record_volume` locks in the same order.

Also added `deleteExpiredByRetention_rollsUpEveryBucketOneBatchSpans` — one batch spanning two profiles across two hours. Nothing previously covered the grouped upsert with more than one bucket in a single statement.

The nitpick on the lock probe in `SigningRecordRepositoryITest` is fixed in a473237d9: the predicate now names the current database, excludes the probe's own backend, and requires the stalled statement to touch `signing_record`. You were right that a cluster-wide match would have released the competing transaction before the delete began waiting, leaving the assertions to pass over an interleaving the test never produced.

src/test/java/com/otilm/core/integration/service/SigningRecordServiceITest.java
```diff
@@ -510,6 +532,194 @@ void reportsFailureForMissingRecordButDeletesExisting() {
...
+            // given
+            SigningRecord signingRecord = insertRecordSignedAt(hoursAgo(2));
+
+            // when
+            signingRecordService.deleteSigningRecord(SecuredUUID.fromUUID(signingRecord.getUuid()));
+
+            // then
+            SigningRecordStatisticsDto statistics = statistics();
+            assertEquals(1L, statistics.getCountLast24h());
+            assertEquals(1L, statistics.getCountLast7d());
```
ivosh, line 595: **~ Minor** | Testing

`signingsSince` (SigningRecordServiceImpl.java:223-228) sums `deletedVolume(...).values()` straight into the returned figure, with no dense-bucket trimming — unlike `volumeOverTime`, where `denseBuckets` drops any key outside `[from, now]`. So for the two window counters the *only* thing keeping pre-window history out is the `bucketStart &gt;= from` predicate at SigningRecordServiceImpl.java:255.

Nothing in this PR exercises that predicate. All eight tests in `StatisticsTests` place their signings inside the window they assert on: `hoursAgo(2)` for the 24h tests (SigningRecordServiceITest.java:546, 560, 574, 587, 652), the floored 24h edge at :619-620, and `hoursIntoTheDayBeforeYesterday` for the 7d test at :601-602. Delete the `greaterThanOrEqualTo` at line 255 and every test in this PR still passes, as does `SigningRecordStatisticsServiceITest` (which deletes nothing). A one-line regression there would make `countLast24h` report the whole retained-plus-history total for the profile, silently and forever.

The cheapest cover sits right beside `windowCountsKeepTheSigningAfterItsRecordIsDeleted` (SigningRecordServiceITest.java:584-596): insert and delete a signing three days old, then assert `getCountLast24h()` is 0 while `getCountLast7d()` is 1 — one test, both directions of the predicate.

_Suggested:_ Add a test that deletes a signing older than 24h and asserts countLast24h excludes it while countLast7d includes it.

src/test/java/com/otilm/core/integration/dao/repository/signing/SigningRecordRepositoryITest.java
```diff
@@ -204,6 +232,268 @@ void deleteRetrievedAndFlagged_keepsNotYetRetrievedRecordsOfFlaggedProfiles() {
...
+                delete.executeUpdate();
+            }
+            CountDownLatch started = new CountDownLatch(1);
+            Future<Integer> blocked = waiting.submit(() -> {
+                started.countDown();
+                return doInTransaction(() -> repository.deleteByUuid(signingRecord.getUuid()));
+            });
+            assertTrue(started.await(BLOCKED_DELETE_TIMEOUT.toSeconds(), TimeUnit.SECONDS),
+                    "the delete under test never started");
+            assertThrows(TimeoutException.class,
```
ivosh, line 414: **~ Minor** | Testing

`deleteBehindAnInFlightDelete` (SigningRecordRepositoryITest.java:398-421) is the only test covering `WAIT_FOR_CONTENDED_ROWS`, and its Javadoc at :393-396 names the exact hazard it has to rule out: "Letting the competing transaction go early would leave the delete nothing to find, and the assertions would then pass over an interleaving this test never produced."

The guard against that hazard is the `assertThrows(TimeoutException.class, () -&gt; blocked.get(BLOCK_CONFIRMATION, ...))` at :413-415, and it does not distinguish the two cases it needs to. `started.countDown()` at :408 fires *before* `doInTransaction(...)` is entered, so `started.await()` proves only that the worker thread was scheduled. A worker that has not yet opened its transaction, resolved the Spring Data proxy and issued the statement within `BLOCK_CONFIRMATION` (1s, :48) produces the same `TimeoutException` as a worker genuinely parked on the row lock. On that interleaving `inFlight.commit()` at :416 lands first, the keyed delete then finds no row, and :382-384 all pass green over a race that never happened — the precise scenario the Javadoc says it is preventing.

Postgres can answer the question the timeout is guessing at: poll `pg_blocking_pids()` / `pg_stat_activity.wait_event_type = 'Lock'` until a session is actually reported waiting, and fail on timeout instead of on non-timeout. That keeps the assertion the commit "Confirm the contended delete is blocked rather than polling for it" was after — a confirmation rather than an assumption — without tying it to how fast the box happens to be.

_Suggested:_ Confirm the block by polling pg_blocking_pids()/pg_stat_activity for a waiting session instead of inferring it from a 1s Future timeout.

src/main/java/com/otilm/core/dao/repository/SecurityFilterRepository.java
```diff
@@ -70,6 +70,14 @@ Map<String, Long> countGroupedUsingSecurityFilter(SecurityFilter filter, Attribu
             SingularAttribute<?, ?> groupBy, BiFunction<Root<T>, CriteriaBuilder, Expression<?>> groupByExpression,
             TriFunction<Root<T>, CriteriaBuilder, CriteriaQuery<?>, Predicate> additionalWhereClause);
 
+    /**
+     * Grouped total of an already-aggregated column, for entities whose rows each carry a count rather than being one.
```
ivosh, line 74: **~ Minor** | Documentation

The new Javadoc at SecurityFilterRepository.java:73-76 tells the next caller the two methods are interchangeable: "Same grouping and security filtering as the variant above, only summing {@code sumOf} instead of counting rows." Two things it says are not so.

The count variant aggregates with `cb.countDistinct(root)` (SecurityFilterRepositoryImpl.java:257); the sum variant with a bare `cb.sum(root.get(sumOf))` (:274). That DISTINCT is not decoration — it is what makes the count correct when the security filter or `additionalWhereClause` joins a to-many and fans the root row out. The sum has no equivalent, so the same filter that leaves the count right multiplies the sum. That is exactly the caveat the sibling `countRowsUsingSecurityFilter` spells out four declarations below, at SecurityFilterRepository.java:86-91 ("Correct only where no predicate can duplicate the root row: every joining shape must go through EXISTS subqueries and the resource's security filter must add no join of its own") — the house pattern for a non-DISTINCT variant is to state the precondition, and this one skips it.

"Same grouping" is also loose: the count variant accepts `join`/`groupBy` attributes and falls back to `groupingByAttribute` when `groupByExpression` is null, while here `groupByExpression` is dereferenced unconditionally at SecurityFilterRepositoryImpl.java:287, so it is mandatory. Today's only caller (`deletedVolume`, SigningRecordServiceImpl.java:251-256) passes one and reads a table with no joins, so nothing is wrong yet — the doc is what will mislead the second caller.

_Suggested:_ Restate the Javadoc to say groupByExpression is required and that the sum is not de-duplicated, mirroring the precondition wording on countRowsUsingSecurityFilter.

src/main/java/com/otilm/core/dao/entity/signing/SigningRecordVolume.java
```diff
@@ -0,0 +1,54 @@
...
+import jakarta.persistence.Entity;
+import jakarta.persistence.Table;
+import jakarta.persistence.UniqueConstraint;
+import java.time.Instant;
+import java.util.UUID;
+import lombok.Getter;
+import lombok.Setter;
+
+/**
+ * How many signings a signing profile performed in one UTC hour, kept after the {@link SigningRecord} rows themselves
```
ivosh, line 14: **. Nitpick** | Documentation

Two facts about the new table are each committed to the tree more than once, in near-identical words:

What the table is for — SigningRecordVolume.java:14-16 ("A record proves what a signing contained; this row preserves that the signing happened, which has to survive the record being deleted by an operator, by retention, or by delete-after-retrieval") and V202609051000__signing_record_volume.sql:1-4, which repeats that sentence almost verbatim and then repeats SigningRecordVolume.java:19-22's "Rows are written only by the roll-up-then-delete statements on SigningRecordRepository" as well.

Why there is no FK — SigningRecordVolume.java:31-35, V202609051000__signing_record_volume.sql:8-10, and CLAUDE.md:112, all making the same argument (the profile outlives its records, the column stays the access-control anchor) in three different phrasings.

Both clusters are aimed at the same reader — whoever is next changing this table — so none of the copies is the "different audience" case. Because the wordings differ, a future change has three independent places to go stale and grep from any one of them finds neither of the others. The migration is the authoritative home for the schema decision (it is where the FK is not created, and it is immutable once shipped); the entity Javadoc can keep the one thing the SQL cannot say, that `signingProfileUuid` is deliberately not a mapped association, and defer the rest. `SigningRecordVolumeRepository.java:8-10` already shows the right shape: a one-line cross-reference, not a re-derivation.

_Suggested:_ Keep the full rationale in the migration and reduce the entity and CLAUDE.md copies to the JPA-specific note plus a pointer.

src/main/java/com/otilm/core/service/impl/SigningRecordServiceImpl.java
```diff
@@ -128,23 +128,24 @@ public PaginationResponseDto<SigningRecordListDto> listSigningRecordsForProfile(
...
 
         SigningRecordStatisticsDto dto = new SigningRecordStatisticsDto();
         dto.setTotalRetained(signingRecordRepository.countUsingSecurityFilter(filter));
-        dto
-                .setCountLast24h(signingRecordRepository
-                        .countUsingSecurityFilter(filter, signedSince(now.minus(Duration.ofDays(1)))));
-        dto
-                .setCountLast7d(signingRecordRepository
-                        .countUsingSecurityFilter(filter, signedSince(now.minus(Duration.ofDays(7)))));
+        dto.setCountLast24h(signingsSince(filter, now.minus(Duration.ofDays(1))));
```
ivosh, line 147: **~ Minor** | Correctness

`SigningRecordServiceImpl.java:147-148` makes `countLast24h`/`countLast7d` "signings performed" (retained records **plus** rolled-up history) opened on an hour boundary. The consumer in `OmniTrustILM/fe-administrator:src/components/_pages/dashboard/SigningRecordsDashboard/index.tsx:142-146` renders that number as the tile "Signings – last 24h" and wires `onSetFilter={() =&gt; buildSigningTimeWindowFilter(availableFilters, SIGNING_WINDOW_HOURS.last24h)}`, i.e. clicking the tile navigates to the signing-records list filtered to an exact rolling 24h of `signingTime`. After this PR the tile can show N while the list it drills into shows fewer rows — every deleted signing counts in the tile but has no row to list — and the two windows also differ by up to an hour. Lines 152-156 do the same for `countLast7d`. Verified via `gh search code --owner OmniTrustILM` plus fetching the file; the fe-administrator repo has no open PR for this (checked `gh pr list --repo OmniTrustILM/fe-administrator --state open`), and this PR's description does not mention the drill-through.

_Suggested:_ File an fe-administrator follow-up issue (or note it in this PR body) so the tile's drill-through is relabelled or de-linked now that its count includes signings with no listable record.

src/main/java/com/otilm/core/service/impl/SigningRecordServiceImpl.java
```diff
@@ -128,23 +128,24 @@ public PaginationResponseDto<SigningRecordListDto> listSigningRecordsForProfile(
...
+        dto.setCountLast7d(signingsSince(filter, now.minus(Duration.ofDays(7))));
```
ivosh, line 148: **. Nitpick** | Documentation

`SigningRecordServiceImpl.java:147-148` changes the published meaning of `countLast24h`/`countLast7d`, and `volumeOverTime` at line 171. I verified the contract side is handled: `OmniTrustILM/interfaces` PR #957 (branch `fix/2097-immutable-signing-volume`, targeting `main`, still OPEN) rewrites exactly those three `@Schema` descriptions and the class Javadoc of `SigningRecordStatisticsDto`, including the hour-boundary wording this PR's own new CLAUDE.md rule demands. The link is one-way, though — #957 says "Documentation for OmniTrustILM/core#2097" while this PR body mentions no interfaces change, so a reviewer or merger here has no signal that a companion PR must land. Core depends on `com.otilm:interfaces:2.20.0-SNAPSHOT` (`pom.xml:70`), so nothing breaks the build either way; the only exposure is this PR merging first and the generated OpenAPI spec describing the old semantics until #957 lands.

_Suggested:_ Add a line to this PR's description linking OmniTrustILM/interfaces#957 and stating the intended merge order.

src/main/java/com/otilm/core/service/impl/SigningRecordServiceImpl.java
```diff
@@ -205,22 +206,79 @@ private Expression<String> signingSchemePairExpression(Root<SigningRecord> root,
...
+     * Start of the hour {@code cutoff} falls in. History is bucketed by the hour, so the retained half of a figure has
+     * to start on the same boundary: only then does a signing contribute the same count before and after its record is
+     * deleted. Against the exact cutoff, a signing in that first hour but before the cutoff would count for nothing
+     * while retained and for one once deleted — deletion would raise the number.
+     */
+    private static Instant windowStart(Instant cutoff) {
+        return cutoff.truncatedTo(ChronoUnit.HOURS);
+    }
+
+    /** Bucketed counts of the records still held in {@code signing_record}. */
```
ivosh: **. Nitpick** | Maintainability

SigningRecordServiceImpl.java:240 — `/** Bucketed counts of the records still held in {@code signing_record}. */` says only what `retainedVolume` and the `signingRecordRepository` call on the next line already say. Its neighbours in this PR all earn their doc: `deletedVolume` (line 247) adds the hourly-storage constraint and the hour-boundary precondition for `from`, `windowStart` (line 230) explains the deletion-raises-the-number trap, `signingProfileScope` (line 271) explains why only the parent filter carries over. This one is the sole pure restatement, and CLAUDE.md's "Writing for future AI sessions" lists "Comments restating method names or class purposes" under Avoid. Either drop it or give it the one fact the name cannot carry — that this half is the one that shrinks as records are deleted, which is what makes it the complement of `deletedVolume`.

_Suggested:_ Delete the one-line Javadoc on `retainedVolume`, or replace it with the fact that this half shrinks as records are deleted while `deletedVolume` grows.

src/main/java/com/otilm/core/service/impl/SigningRecordServiceImpl.java
```diff
@@ -128,23 +128,24 @@ public PaginationResponseDto<SigningRecordListDto> listSigningRecordsForProfile(
         return listRecords(request, filter, signingProfileUuid);
     }
 
+    /**
+     * Read under one snapshot: the signing figures are assembled from two tables, and a record deleted between the two
+     * queries would otherwise be counted both as retained and as history. {@code REPEATABLE_READ} costs nothing here —
+     * the transaction only reads, so it cannot fail to serialize.
```
ivosh: **. Nitpick** | Documentation

`SigningRecordServiceImpl.java:134` claims `REPEATABLE_READ` costs nothing because the transaction only reads and so cannot fail to serialize. The serialization half is right, but a PostgreSQL REPEATABLE READ transaction holds its snapshot — and therefore pins the cluster xmin horizon — from its first statement to commit, whereas a READ COMMITTED one releases it between statements. `getSigningRecordStatistics` (`SigningRecordServiceImpl.java:139`) now issues roughly a dozen statements, several of them full grouped scans over the blob-carrying `signing_record` table, so a slow dashboard read blocks reclamation of dead `signing_record` tuples for its whole duration.

This repo already treats that as a real cost: `SigningRecordRetentionSweeper.java:33-36` justifies its per-sweep batch cap precisely because "a long hold would keep a connection idle-in-transaction and pin the vacuum horizon". The isolation choice is still the right one — the snapshot is what makes the two-table read consistent — but the Javadoc understates what it buys it with, and this method is where a maintainer will copy the pattern from.

_Suggested:_ Reword the Javadoc to say the snapshot cannot fail to serialize but does pin the vacuum horizon for the read's duration, rather than that it costs nothing.

src/main/java/com/otilm/core/service/impl/SigningRecordServiceImpl.java
```diff
@@ -205,22 +206,79 @@ private Expression<String> signingSchemePairExpression(Root<SigningRecord> root,
...
+                        cb.literal(bucket == Bucket.HOUR ? BUCKET_HOUR_FORMAT : BUCKET_DAY_FORMAT));
+    }
+
+    /**
+     * The signing-profile half of {@code filter}. Signing-record visibility follows signing-profile access, and a
+     * rolled-up bucket has no record identity left for a record-level filter to match on, so only the parent filter
+     * carries over to the history rows. Nothing is lost with it: {@link Resource#SIGNING_RECORD} declares no
+     * object-level access, so the record half of the filter never narrows anything on its own.
+     */
+    private static SecurityFilter signingProfileScope(SecurityFilter filter) {
```
ivosh, line 276: **? Question** | Security — Minor if confirmed

SigningRecordServiceImpl.java:271-281 drops `filter.getResourceFilter()` for the history read and justifies it with "Resource#SIGNING_RECORD declares no object-level access, so the record half of the filter never narrows anything on its own." I confirmed the premise — in the `interfaces` module `SIGNING_RECORD(Codes.SIGNING_RECORD, "Signing Record")` uses the two-arg constructor, so `hasObjectAccess`/`hasGroups`/`hasOwner` are all false — but the code path that builds the filter never consults it: `ObjectFilterAspect.getResourceFilter` (ObjectFilterAspect.java:246) populates the record-level `SecurityResourceFilter` straight from the OPA `checkObjectAccess` vote for every resource, gated on nothing. So the guarantee lives in the OPA policy, not in this repo.

If that policy ever returns `actionAllowedForGroupOfObjects=false` for `signingRecords`, the two halves of the same figure diverge: the retained half gets `record ∧ parent` (`SecurityFilterRepositoryImpl.initObjectAccessPredicates`, line 588-589) while the history half gets `parent` only, so the history rows would be broader than what the caller may see live — silently, with no test or assertion catching it. If the intent is "this can only ever be a no-op", it is cheap to make that structural instead of asserted.

_Suggested:_ Either derive the drop from `Resource.SIGNING_RECORD.hasObjectAccess()` (and fail loudly if it ever becomes true), or add one line to the Javadoc naming the OPA policy as the thing that holds this invariant.

src/main/java/com/otilm/core/service/impl/SigningRecordServiceImpl.java
```diff
@@ -205,22 +206,79 @@ private Expression<String> signingSchemePairExpression(Root<SigningRecord> root,
...
+        Instant from = windowStart(now.minus(period.getWindow()));
+        Bucket bucket = period.getBucket();
+        Map<String, Long> sparse = SigningRecordStatisticsCalculator
+                .sumByBucket(retainedVolume(filter, bucket, from), deletedVolume(filter, bucket, from));
+        return SigningRecordStatisticsCalculator.denseBuckets(from, now, bucket, sparse);
+    }
+
+    /** Signings performed since {@code cutoff}, whether or not their records still exist. */
+    private Long signingsSince(SecurityFilter filter, Instant cutoff) {
+        Instant from = windowStart(cutoff);
```
ivosh, line 224: **? Question** | Correctness — Minor if confirmed

`SigningRecordServiceImpl.java:224` floors the count window to the hour, so `countLast24h` covers up to 25 hours. The dashboard renders that number as a clickable badge whose drill-through builds an exact window: `fe-administrator-pkcs11-connector/src/components/_pages/dashboard/SigningRecordsDashboard/index.tsx:142-148` pairs `data={statistics.countLast24h}` with `onSetFilter={() =&gt; buildSigningTimeWindowFilter(availableFilters, SIGNING_WINDOW_HOURS.last24h)}` — `signingTime &gt;= now - 24h`, computed client-side at click time. Records in `[floor_hour(now-24h), now-24h)` count for the badge and are filtered out of the list, so the badge reads higher than the list it opens on a fresh install with nothing deleted; once history exists, the deleted half widens the gap further. The same holds for the `countLast7d` badge and for the `Signings over Time` chart at `index.tsx:171-180`, whose bar click filters the record list by the bar's exact range. The PR description discloses the hour-boundary decision but not this consumer divergence.

_Suggested:_ Decide whether the drill-through should floor its window to the hour to match the badge, or the badge caption should stop implying the list will contain the same rows — and say which in the release note.

CLAUDE.md
```diff
@@ -90,6 +90,29 @@ try {
 
 Capture entry state before the external call (`final State entryState = entity.getState();`) so the restoration path has something to restore *to*.
 
+## Historical figures must not be derived from deletable rows
+
+A dashboard series, counter, or report that aggregates a table whose rows get deleted is rewritten by every one of those deletions — and retention sweeps and delete-after-retrieval mean the history changes with nobody touching it. Derive such a figure from a separate append-only aggregate instead, and fold rows into it in the **same statement** that deletes them:
+
+```sql
+WITH victim AS MATERIALIZED (SELECT uuid, ... FROM detail WHERE ... LIMIT :limit)
```
ivosh: **. Nitpick** | Documentation

The SQL block this PR adds at `CLAUDE.md:97-104` is the shape a future session will copy, but it is missing exactly the two things the paragraph at `CLAUDE.md:108` declares non-optional: the `victim` CTE at `CLAUDE.md:98` carries no locking clause ("Without a locking clause a second deleter sees the same rows in its own snapshot ... one row counted twice") and the grouped upsert at `CLAUDE.md:100` carries no `ORDER BY` ("Without an `ORDER BY` on the grouped upsert, two writers ... can take those locks in opposite orders and deadlock"). The shipped implementation does carry both — `SigningRecordRollupSql.java` appends `WAIT_FOR_CONTENDED_ROWS` / `SKIP_CONTENDED_ROWS` and orders the upsert by `signing_profile_uuid, bucket_start` — so the snippet is the only place in the doc set that shows the deadlock-prone, double-counting version. `CLAUDE.md:155` ("Writing for future AI sessions") is the reason this matters: the snippet, not the prose beneath it, is what gets lifted.

_Suggested:_ Add `FOR UPDATE OF detail [SKIP LOCKED]` to the `victim` CTE and an `ORDER BY` on the grouped upsert in the CLAUDE.md snippet so the copied shape is the one the prose mandates.

<sub>flagged by Doc-set & publication coherence</sub>

src/main/java/com/otilm/core/dao/repository/SecurityFilterRepositoryImpl.java
```diff
@@ -251,19 +251,41 @@ private <R> TypedQuery<R> window(final TypedQuery<R> query, final Pageable p) {
...
         final Root<T> root = cr.from(entity);
-        Expression<?> groupBySelection;
-        if (groupByExpression != null) {
-            groupBySelection = groupByExpression.apply(root, cb);
-        } else {
-            From from = join == null ? root : root.join(join.getName(), JoinType.LEFT);
-            groupBySelection = from.get(groupBy);
-        }
-        cr.multiselect(groupBySelection, cb.countDistinct(root));
+        Expression<?> groupBySelection = groupByExpression.apply(root, cb);
```
ivosh, line 288: **. Nitpick** | Error handling

The extracted helper at `SecurityFilterRepositoryImpl.java:287` now does `groupByExpression.apply(root, cb)` unconditionally. That is safe for `countGroupedUsingSecurityFilter`, which guards it at line 254 and falls back to `groupingByAttribute(join, groupBy)`, but `sumGroupedUsingSecurityFilter` (line 274) passes the caller's reference straight through with no fallback and no `join`/`groupBy` overload to fall back to. Its Javadoc in `SecurityFilterRepository.java:75` reads "Same grouping and security filtering as the variant above", which points a reader at the sibling that *does* tolerate `null` — the one caller (`SigningRecordServiceImpl.deletedVolume`) always passes an expression, so the next one is the one that finds out, via a bare NPE from inside a repository lambda with no parameter name in the message. *Effective Java* Item 49 (check parameters for validity) — either state the requirement or enforce it at the boundary.

_Suggested:_ Either add `Objects.requireNonNull(groupByExpression, "groupByExpression")` at the top of `sumGroupedUsingSecurityFilter`, or add an `@param groupByExpression` line to the interface Javadoc saying it is required here unlike in the counting variant.

src/main/java/com/otilm/core/dao/repository/signing/SigningRecordRollupSql.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+final class SigningRecordRollupSql {
+
+    /**
+     * Opens the {@code victim} CTE; the caller appends the SELECT naming the rows to delete, which must alias
+     * {@code signing_record} as {@code sr} and project {@code uuid}, {@code signing_profile_uuid} and
+     * {@code signing_time}. Materialized because both halves of the statement read it and must see the same rows — a
+     * re-evaluated {@code LIMIT} could pick a different set.
+     */
+    static final String CLAIM_VICTIMS = "WITH victim AS MATERIALIZED (";
```
ivosh: **. Nitpick** | Maintainability

`SigningRecordRollupSql.java:29` is `static final String CLAIM_VICTIMS = "WITH victim AS MATERIALIZED ("` — it opens a CTE and claims nothing. Its own Javadoc two lines up says so ("Opens the {@code victim} CTE"), and the class Javadoc at line 8 splits the roles explicitly: a statement "opens with `CLAIM_VICTIMS`, selects the rows it is about to delete, **claims them with** `WAIT_FOR_CONTENDED_ROWS` or `SKIP_CONTENDED_ROWS`". So the file names the claiming step twice and both times it is a different constant. The three peer constants here are named for what their own text does (`ROLL_UP_THEN_DELETE` really is the roll-up and the delete); this one is named for what the assembled statement does, which is what makes it read wrong at the `@Query` call sites in `SigningRecordRepository.java` where the fragments appear in sequence.

_Suggested:_ Rename `CLAIM_VICTIMS` to something describing the fragment itself, e.g. `OPEN_VICTIM_CTE`, and update the four static-import sites in `SigningRecordRepository`.

src/test/java/com/otilm/core/integration/service/SigningRecordServiceITest.java
```diff
@@ -510,6 +532,194 @@ void reportsFailureForMissingRecordButDeletesExisting() {
...
+        /**
+         * {@code hour} o'clock UTC on the day before yesterday — a whole day inside the 7-day window, so a daily bucket
+         * built from it is neither the partial leading day nor today.
+         */
+        private Instant hoursIntoTheDayBeforeYesterday(int hour) {
+            return Instant.now().minus(Duration.ofDays(2)).truncatedTo(ChronoUnit.DAYS).plus(Duration.ofHours(hour));
+        }
+
+        private long volumeIn(Instant bucket) {
+            return statistics().getVolumeOverTime().get(HOUR_BUCKET_KEY.format(bucket));
```
ivosh, line 721: **. Nitpick** | Testing

`SigningRecordServiceITest.java:706-707` declares `private long volumeIn(Instant bucket)` and returns `statistics().getVolumeOverTime().get(HOUR_BUCKET_KEY.format(bucket))` — a `Long` auto-unboxed to `long`. A missing key is exactly the regression these tests exist to catch (a bucket-key format drift between `BUCKET_HOUR_FORMAT` and the Java formatter, or a window start that no longer covers the bucket), and it surfaces as a bare `NullPointerException` inside the helper rather than as a failure naming the key and the map. *Effective Java* Item 61 — auto-unboxing a possibly-absent boxed value is the classic NPE source. The sibling `totalVolume()` at line 710 is unaffected because it streams the values. Note the neighbouring `dailySeriesSumsTheHourlyHistoryOfADeletedDay` at line 612 already gets this right by handing the raw `Long` to `assertEquals`, which reports a readable `expected: &lt;2L&gt; but was: &lt;null&gt;`.

_Suggested:_ Have `volumeIn` return `Long` (or assert with `assertThat(statistics().getVolumeOverTime()).containsEntry(key, 1L)`) so an absent bucket reports the key rather than throwing.

src/main/java/com/otilm/core/dao/repository/signing/SigningRecordRollupSql.java
```diff
@@ -0,0 +1,77 @@
+package com.otilm.core.dao.repository.signing;
+
+/**
+ * The two halves every statement that removes signing records is wrapped in, so that removing a record never removes
```
ivosh: **~ Minor** | Documentation

The class summary — the one line an IDE prints on hover and the generated class index shows — is `SigningRecordRollupSql.java:4-5`: *"The two halves every statement that removes signing records is wrapped in, so that removing a record never removes the fact that the signing happened."*

**The count is wrong against the file's own contents.** The paragraph directly beneath it (`:8-9`) names **three** fragments this class contributes around the caller's own `SELECT` — `CLAIM_VICTIMS`, one of `WAIT_FOR_CONTENDED_ROWS`/`SKIP_CONTENDED_ROWS`, and `ROLL_UP_THEN_DELETE` — and `SigningRecordRepository.java:28-32`, `:36-45` and `:49-58` assemble exactly that: opener + SELECT + lock clause + closer. The file ships four constants. A reader who counts gets a different answer from the summary that told them "two halves".

**And the sentence garden-paths.** Twenty-five words, with the relative pronoun dropped ("the two halves *that* every statement … is wrapped in") and the preposition stranded at the end: "The two halves every statement" reads as verb-plus-object until "is wrapped in" forces a re-parse. It also conjugates *remove* three times in one breath ("removes signing records", "removing a record", "never removes the fact"), and fuses what the class **is** with why it exists — so neither half lands on first read.

The two paragraphs under it repeat rather than add. `:8-13` re-derives what each constant's own Javadoc already states (`CLAIM_VICTIMS` — "Opens the `victim` CTE"; `ROLL_UP_THEN_DELETE` — "Rolls the claimed rows into their UTC hourly buckets and deletes them"), and `:16-19` restates the double-count rationale this same PR writes into `CLAUDE.md:108`. `CLAUDE.md`'s own **Avoid** list names "Comments restating method names or class purposes".

Fix shape — nothing factual dropped, only restructured: a short summary that does not commit to a count ("SQL fragments that wrap every signing-record delete, so deleting a record never deletes the fact that the signing happened."), then one concern per `<p>` with a bold lead-in (`<b>How a statement is assembled:</b>`, `<b>Why the rows are claimed first:</b>`), sentences around 20 words, and the per-fragment detail left to each constant's own Javadoc.

_Suggested:_ Replace the opening sentence with a short summary that commits to no fragment count, and split the rest into one bold-led `<p>` per concern, dropping what the per-constant Javadoc and `CLAUDE.md:108` already say.

ivosh (changes requested):

src/main/java/com/otilm/core/dao/repository/signing/SigningRecordRollupSql.java
```diff
@@ -0,0 +1,77 @@
...
+     */
+    static final String WAIT_FOR_CONTENDED_ROWS = "FOR UPDATE OF sr\n";
+
+    /**
+     * Leaves a row another statement holds where it is. For the batch sweeps, which have no per-record contract —
+     * whatever a sweep passes over is taken by the statement holding it or by the next sweep. Waiting instead would let
+     * two sweeps that pick overlapping victims in different index orders deadlock, or let one open transaction stall a
+     * sweep while it holds the cluster-wide sweep lock.
+     */
+    static final String SKIP_CONTENDED_ROWS = "FOR UPDATE OF sr SKIP LOCKED\n";
```
ivosh: **. Nitpick** | Documentation

**The class documents what each fragment is, but never states the whitespace rule that assembling them depends on.**

The fragments end three different ways. `CLAIM_VICTIMS` (`:29`) ends with `(` and no newline. `WAIT_FOR_CONTENDED_ROWS` (`:36`) and `SKIP_CONTENDED_ROWS` (`:44`) end with an escaped `\n` inside a one-line literal. `ROLL_UP_THEN_DELETE` (`:54`) gets its trailing newline implicitly, from the text block's closing delimiter sitting on its own line. Three mechanisms for one job, so a reader cannot infer the rule from the pattern.

**Not every joint is fragile — one is.** `CLAIM_VICTIMS` ending in `(` is safe, because `(` is a delimiter and cannot fuse with the `SELECT` that follows. The joint that matters is the next one: both lock constants begin with a bare `FOR`, carrying no leading whitespace of their own, so **the SELECT the caller supplies must end with a newline**. Today it does, because each text block in `SigningRecordRepository.java` (`:28-32`, `:36-45`, `:49-58`) puts its closing `"""` on its own line. That is a property of how the caller happened to be typed. This class neither states it nor can enforce it.

**When it breaks, the message points somewhere else.** A caller who closes the text block on the last content line produces `LIMIT :limitFOR UPDATE OF sr SKIP LOCKED`. That reads as a named parameter `limitFOR`, so the failure is a parameter-not-found error naming `limitFOR`; a fourth statement ending in a literal instead would give a plain SQL syntax error. In neither case does the word "newline" appear. And nothing catches it earlier — `@Query` values are compile-time constants, and the native SQL is never checked against the database at build time.

**The odds of a fourth caller are going up, not down.** `CLAUDE.md:97-104` in this same PR presents the shape as a snippet to copy for the next table that needs it.

The fix is one sentence rather than a code change, though the code change is cheap too: give `WAIT_FOR_CONTENDED_ROWS` and `SKIP_CONTENDED_ROWS` a leading `\n` and the joint stops depending on the caller at all. A blank line in the generated SQL costs nothing.

_Suggested:_ State on the class that every fragment — including the SELECT the caller supplies — must end with a newline, and give the two lock constants a leading `\n` so the joint no longer depends on where the caller put its closing `"""`.

src/test/java/com/otilm/core/integration/service/SigningRecordServiceITest.java
3keyroman: Fixed in 857b95a8a. Verified the claim first: replacing that `greaterThanOrEqualTo` with `null` left all 31 tests green, so nothing covered it.

Added `windowCountsExcludeADeletedSigningOlderThanTheWindow` — a deleted signing three days old, asserting `countLast24h` is 0 and `countLast7d` is 1. One test, both directions, and it now fails with the predicate removed.

src/main/java/com/otilm/core/dao/repository/SecurityFilterRepository.java
3keyroman: Fixed in 857b95a8a. Both points were right: `groupByExpression` is dereferenced unconditionally in `aggregateGroupedUsingSecurityFilter`, so it is mandatory, and `cb.sum` has no DISTINCT to match the count's. The Javadoc now states the precondition in the same shape as `countRowsUsingSecurityFilter`.

src/main/java/com/otilm/core/service/impl/SigningRecordServiceImpl.java
3keyroman: Fixed in 857b95a8a — dropped. It was the one line in the set that restated its own name.

src/test/java/com/otilm/core/integration/dao/repository/signing/SigningRecordRepositoryITest.java
3keyroman: Partly addressed in 857b95a8a, and I want to flag the trade-off rather than quietly pick one side.

You are right that `started.countDown()` fired before the transaction was entered, so it proved only that the worker was scheduled. The latch now counts down *inside* `doInTransaction`, so when it fires the transaction is open and the connection is in hand and issuing the statement is all that remains — which is what the following timeout then rules out.

I did not go back to polling `pg_stat_activity`/`pg_blocking_pids()` because that is precisely what the previous revision did, and SonarCloud failed it on two counts: `java:S2925` for the `Thread.sleep` pacing the poll and `java:S9142` for the `prepareStatement` inside the loop. Awaitility is not on the test classpath, so a poll needs one or the other. If you would rather have the definitive confirmation and accept those two smells (or add Awaitility), say so and I will switch — it is your call which side of that trade to take.

src/main/java/com/otilm/core/dao/entity/signing/SigningRecordVolume.java
3keyroman: Skipping this one as a nitpick, and saying so rather than leaving it unanswered.

The overlap is real, but the three copies are not equivalent: CLAUDE.md states the general pattern for any such aggregate rather than describing this table, and the migration comment and entity Javadoc reach different readers in practice — someone reading the schema, and someone editing the entity who will not open a shipped migration. Trimming to a pointer trades a little redundancy for a hop, on a decision unlikely to change.

src/main/java/com/otilm/core/dao/repository/signing/SigningRecordRollupSql.java
3keyroman: Fixed in b9d0053c9 — the best catch of the set. Both lock constants now open with a leading `\n`, so the joint no longer depends on where the caller put its closing `"""`, and the class summary states that every fragment following the SELECT brings its own newline.

src/main/java/com/otilm/core/dao/repository/signing/SigningRecordRollupSql.java
3keyroman: Fixed in b9d0053c9. "Two halves" was written before the WAIT/SKIP split added a third fragment, so it was simply stale. The summary now commits to no count, and the paragraph that re-derived each constant's own Javadoc is gone.

src/main/java/com/otilm/core/dao/repository/signing/SigningRecordRollupSql.java
3keyroman: Fixed in b9d0053c9 — renamed to `OPEN_VICTIM_CTE`, with the four sites updated. Two constants both claiming to "claim" was genuinely confusing at the call sites.

src/main/java/com/otilm/core/dao/repository/SecurityFilterRepositoryImpl.java
3keyroman: Fixed in b9d0053c9 — added `Objects.requireNonNull(groupByExpression, "groupByExpression")`, alongside the Javadoc change on the sibling thread. A named failure beats an NPE surfacing from inside a repository lambda.

CLAUDE.md
3keyroman: Fixed in b9d0053c9. The snippet was the one place in the doc set showing the version the prose beneath it forbids, and the snippet is what gets copied — it now carries `FOR UPDATE OF detail [SKIP LOCKED]` and the `ORDER BY`.

src/test/java/com/otilm/core/integration/service/SigningRecordServiceITest.java
3keyroman: Fixed in b9d0053c9 — `volumeIn` returns `Long`, so a missing bucket reports `expected: <1> but was: <null>` instead of throwing from inside the helper. A key that has drifted is exactly what these tests exist to catch.

src/main/java/com/otilm/core/service/impl/SigningRecordServiceImpl.java
3keyroman: Fixed in b9d0053c9. "Costs nothing" was wrong — it conflated the serialization half with the cost. The Javadoc now says the snapshot cannot fail to serialize but does hold the vacuum horizon for the length of the read, which is the same cost `SigningRecordRetentionSweeper` already documents.

src/main/java/com/otilm/core/service/impl/SigningRecordServiceImpl.java
3keyroman: Fixed in b9d0053c9. You are right that the guarantee does not live where my comment said it did: `ObjectFilterAspect.getResourceFilter` populates the record-level filter from the OPA vote unconditionally, without consulting `hasObjectAccess()`. The Javadoc no longer claims the enum holds the invariant — it now says the drop is a no-op while the authorization policy grants no object-level access on signing records, which is what the enum declaring none asks of it. I left it as documentation rather than a runtime assertion since the alternative half of the filter cannot be carried across anyway: it matches on record identity, which a rolled-up bucket no longer has.

src/main/java/com/otilm/core/service/impl/SigningRecordServiceImpl.java
3keyroman: Done — added a line to the PR description linking OmniTrustILM/interfaces#957 and stating that it should merge first so the published spec matches.

src/main/java/com/otilm/core/service/impl/SigningRecordServiceImpl.java
3keyroman: Confirmed, and it is a known consequence rather than an oversight — I raised the same divergence before this PR was opened. The tile counts signings; the list it opens can only show records that still exist, so once anything has been deleted the two must differ. That is the point of the fix, not a side effect of it.

Leaving this open: the remediation is a caption or de-link in fe-administrator, which is outside this PR, and I am checking whether to raise it there.

src/main/java/com/otilm/core/service/impl/SigningRecordServiceImpl.java
3keyroman: Right, and worth separating from the deleted-records half: the hour flooring alone makes the badge cover up to 25 hours while the drill-through filters an exact rolling 24, so the two can differ on a fresh install with nothing deleted.

I kept the flooring deliberately — against the exact cutoff, a signing inside the first hour but before the cutoff counts for nothing while retained and for one once deleted, so deleting a record would *raise* the badge. That is the bug this PR exists to remove, so the badge and the history have to share a boundary. It is the drill-through that should follow, not the badge.

Leaving this open with the sibling thread while I check whether to raise the fe-administrator side.

src/test/java/com/otilm/core/integration/dao/repository/signing/SigningRecordRepositoryITest.java
3keyroman: Addressed properly in the latest commit — I found a way to take your suggestion without the Sonar cost, so no decision needed from you after all.

The confirmation is now `pg_blocking_pids()`, asked **once** after the timeout has established that the delete has not returned, rather than polled in a loop. One prepared statement and no `Thread.sleep`, so neither `java:S9142` nor `java:S2925` applies, and the assertion is a positive report from the server instead of an inference from elapsed time. The predicate also names the current database and excludes the probe's own backend.

The latch move stands as well: it now counts down inside the transaction, so the timeout is ruling out a statement that is already issued rather than a thread that may not have got going.

src/main/java/com/otilm/core/service/impl/SigningRecordServiceImpl.java
3keyroman: Raised and fixed in OmniTrustILM/fe-administrator#2112.

The tiles now say `includes deleted records` in place of `by signing time`, so the count no longer implies the list will contain the same rows. That half cannot be fixed by filtering — no filter produces rows that no longer exist — so the label is the honest remedy.

src/main/java/com/otilm/core/service/impl/SigningRecordServiceImpl.java
3keyroman: Fixed in OmniTrustILM/fe-administrator#2112 — and you were right that this half is separable from the deleted-records one.

`buildSigningTimeWindowFilter` now floors the window start to the hour, so the drill-through opens on the same boundary the tile counts from and the two agree exactly when nothing has been deleted. Added a test with a mid-hour `now`; the existing ones all used an exactly-on-the-hour clock, so they could not have caught it.

The badge keeps the flooring, as you say it should: against an exact cutoff, deleting a record would raise it.

src/test/java/com/otilm/core/integration/dao/repository/signing/SigningRecordRepositoryITest.java
```diff
@@ -204,6 +233,286 @@ void deleteRetrievedAndFlagged_keepsNotYetRetrievedRecordsOfFlaggedProfiles() {
...
+    }
+
+    /**
+     * Whether PostgreSQL reports a session in this database blocked by another. Asked once, after the timeout above has
+     * established that the delete has not returned, so being stuck is confirmed by the server rather than inferred from
+     * how fast the box happens to be.
+     */
+    private boolean aSessionIsBlockedByAnother() throws SQLException {
+        try (Connection probe = testDataSource.getConnection(); PreparedStatement blocked = probe.prepareStatement("""
+                SELECT count(*) FROM pg_stat_activity
```
ivosh: **~ Minor** | Testing

SigningRecordRepositoryITest.java:433 scans every backend in the database and succeeds when any session has a blocker; it does not identify the worker executing `deleteByUuid` or the connection holding the contested row. If the worker is merely delayed after opening its transaction while unrelated application activity is blocked, this test commits the competing delete and all final assertions can pass without exercising the intended wait-and-recheck interleaving.

_Suggested:_ Capture both connections' PostgreSQL backend PIDs and assert that the holder PID appears in `pg_blocking_pids(workerPid)` before committing the competing delete.

src/test/java/com/otilm/core/integration/service/SigningRecordServiceITest.java
```diff
@@ -510,6 +532,208 @@ void reportsFailureForMissingRecordButDeletesExisting() {
...
+            signingRecordService.deleteSigningRecord(SecuredUUID.fromUUID(evening.getUuid()));
+
+            // then
+            Map<String, Long> daily = statistics(SigningRecordStatisticsPeriod.LAST_7D).getVolumeOverTime();
+            assertEquals(2L, daily.get(DAY_BUCKET_KEY.format(earlier)));
+        }
+
+        @Test
+        void windowCountsExcludeADeletedSigningOlderThanTheWindow() throws NotFoundException {
+            // given
```
ivosh: **. Nitpick** | Maintainability

src/test/java/com/otilm/core/integration/service/SigningRecordServiceITest.java:617, :620, and :623 add bare `// given`, `// when`, and `// then` labels around statements whose roles are already clear from the test name and code. This PR therefore adds three comments without preserving any non-obvious context.

_Suggested:_ Remove the three bare phase comments and let the test method and statements provide the structure.

ivosh (approved):

PR #2227, opened 2026-09-07
PQC readiness rule set and evaluator (1 of 2)

vyskocilm:
The rule set that decides a PQC readiness verdict, and the evaluator that runs it. First of two stacked
PRs for #2151; the follow-up stores verdicts and sweeps stale ones. **1 469 lines, one commit.**

No persistence, no scheduling, no Spring context needed to test any of it.

## What it decides

`PqcFamilies` dispositions **all 130** ratified algorithm families onto one of four verdicts, keyed on the
ratified spelling. `PqcFamiliesTest` fails the build if the identity tables gain a family without a
disposition — which is exactly what a generated `pqcFamilies` table (#2196 ruling **C12**) would give for
free, and is the standing cost of that ruling being carried rather than implemented. The map is shaped so
C12 can absorb it: swapping to a generated set is a lookup change, not a re-authoring.

Two orderings in the rule table are load-bearing. Asset-type rules run first, because an asset that is not
an algorithm has no family to read. The hybrid rule runs before every family rule, because a hybrid's
stored family is whichever construction the identity grammar elected — `X25519-ML-KEM-768` stores the
classical half, so a family-first order would report a migrated asset as un-migrated.

A hybrid inherits its post-quantum component's **own** disposition rather than a blanket ready:
`X25519-ML-KEM-768` → `ready`, `X25519-Kyber768` → `notReady`, because bare Kyber is a superseded draft
that is not wire-compatible with the scheme that replaced it. A rule testing only that a post-quantum
component exists would report the second as migrated — the "silently erasing a weak-crypto finding"
outcome an inventory must never produce.

## Two constraints made structural rather than conventional

`nistQuantumSecurityLevel` is **not a member of `PqcRuleInput`**. It is producer-supplied and observed to
disagree across producers for one asset (SHA-384 carries 0, 2 and 3; SHA-256 adds 5; one producer wrote a
non-numeric string). Keeping it out of the type a predicate can see is what makes "corroborates, never
decides" a property of the design instead of a convention someone has to keep.

The same record is the identity-key boundary. `pqc_evaluated_fields` reaches clients verbatim as
`CryptographicAssetVerdictDto.evaluatedFields`, and `IdentityKeyExposureFence` cannot see into it — the
fence is lexical, both ends are innocuously named, and the map's keys are runtime data. The guarantee is
that the input carries no key to begin with, so the evidence projection cannot record one.
`PqcEvidenceAllowlistTest` asserts that over the **rendered JSON**, which is the assertion the fence
[...]

PR #2196, opened 2026-09-02
Close the core#2165 identity adjudications that are repairs, and price the rest (3 of 3)

vyskocilm:
## Summary

Part 2 of 2 of the #2165 adjudication series. **The stack is gone: #2173 and #2170 have both merged,
this PR targets `main`, and it is the last of the three.**

**#2170 merged at the diff its reviewer read, so it carries none of the fixes answering that review —
all sixteen are here.** `karolhrdina` reviewed #2170 at `f7045c73` with 30 inline findings; the branch was
frozen for that review, so every fix routed onto this one. Sixteen are fixed here and fourteen are recorded
as open key-moving work on #2165. That makes this PR two things at once: part 2 of the adjudication series,
and the only place #2170's review is answered. See *Answering #2170's review* below.

#2173 landed the primitive-layer repairs. This closes the rest of the docket that is a repair,
implements one arm of the one adjudication that blocks nothing, and delivers measurement plus a stated
ask for the items that need the ratifier rather than a patch.

## Measured cost: 4 rows in two segments, one group merges

Every costing here comes from `CorpusKeySnapshotTest` (committed in this PR). **It now keys the population
ingest keys**, driving `CbomAssetExtractor.extract` rather than keying every component
`DocumentScope.walk` returns. An earlier revision of the instrument counted ordinary libraries, files and
frameworks as inventory on `backstop:unknown-type`, and two `library` components sharing a name in different
documents formed a size-2 key group ingest cannot produce. **The "8 048 components keyed / 4 836 distinct
identity keys" figures an earlier version of this section quoted were that walk population, not ingest's.**
Of the 8 048 components in the corpus, 3 054 are not cryptographic assets at all.

Corrected population, over the `corpora/` tree of the OmniTrustILM/ilm#299 working directory
`ilm299-t1-work-2026-08-18`: **200 `.json` files, 196 parsed, 0 unparseable, 4 non-object roots,
8 048 components, 4 994 keyed, 0 skipped**.

The cost is measured in two segments because a single end-to-end run is not available: the PR baseline
`9b9ee07f` no longer compiles against the `interfaces` 2.20.0-SNAPSHOT now in the local repository, so the
corrected instrument cannot be run on it. The segments are disjoint in the rows they move.

**Segment 1 — `9b9ee07f` to `149a7ac1`, 2 rows** (measured with the earlier walk-population instrument, so
its totals are the 8 048 / 4 836 pair; the two rows below are cryptographic assets and so belong to the
ingest population as well):

[...]

Issue #2151, opened 2026-08-26
Implement PQC verdict rule set, persistence and re-evaluation sweep


vyskocilm:
### Description

PQC readiness verdicts over the cryptographic asset inventory (ilm#299, design doc
section 6). Split out of [core#2072](https://github.com/OmniTrustILM/core/issues/2072),
which held both halves at 6 days together, estimate-neutral: that story kept extraction
and normalization and **merged on 2026-08-31 as PR core#2161**, so the merged view this
story evaluates over is live on `main`.

**What this story adds is the evaluator, its vocabulary rulings, and the sweep — not the
storage.** The whole persistence and exposure half already shipped ahead of it, and
re-deriving any of it is rework:

| Already on `main` | Where |
| --- | --- |
| `crypto_asset.pqc_verdict / pqc_rule_id / pqc_reason / pqc_ruleset_version / pqc_evaluated_fields / pqc_decided_at / pqc_evaluated_at`, with the `ck_crypto_asset_pqc_verdict` CHECK list and btree indexes on verdict and ruleset version | `V202608271000__crypto_asset_inventory.sql` |
| The single write path, `@Transactional`, with the two-timestamp semantics already in SQL (`pqc_evaluated_at` every call; `pqc_decided_at` on a verdict-value change, and filled if it was null) | `CryptoAssetWriter.applyPqcVerdict` → `CryptoAssetRepository.applyPqcVerdict` |
| Wire contract, frozen: four-value `PqcVerdict`, `CryptographicAssetVerdictDto` | interfaces#909 / #913, merged |
| Filtering and aggregation over the column | `FilterField.CBOM_ASSET_PQC_VERDICT`, `CBOM_ASSET_PQC_RULESET_VERSION`; detail + statistics endpoints live via PR core#2188 (2026-09-02) |

So this story ships four things:

1. The hardcoded Java rule set — registry tokens plus the PQC pseudo-families, NIST IR 8547
   framing, first-match-wins, its own integer rule-set version constant. No YAML, no Settings
   entry, nothing configurable.
2. A callable evaluator surface core#2073 can invoke per unit of work: it takes the normalized
   row fields plus the out-of-key rule inputs (`hybridComponents`, `nistQuantumSecurityLevel`)
   and persists **only** through the crypto-asset writer beans (see the persistence criterion
   below, which names which bean and why it cannot be `CryptoAssetWriter` itself).
3. The write-side allowlist that decides what `pqc_evaluated_fields` may contain.
[...]

PR #2228, opened 2026-09-07
Persist PQC verdicts and sweep the stale ones (2 of 2)

vyskocilm:
Stores what the rule set decides and keeps it current after a rule-set bump. Second of two PRs for #2151;
**#2227 is merged** (`ce0890d9`), so this one is no longer stacked and its base is `main`. **+1 653 / −219 over
31 files, three commits** — the feature, the review fixes, and a spotless reflow. It is over the 1K floor now
that the review round added a native work list, a per-row retry path and three test classes, and the natural
seam between cryptographic judgement and transaction mechanics still falls where #2227 ends: splitting
further would cut a class from its own tests.

## The batch write is its own bean, and had to be

`CryptoAssetPqcVerdictWriter` is a new `@Component`, **not** a method on `CryptoAssetWriter`. Two
architecture rules forbid that shape and both fail the build:

- `IdentityRulesetStampArchTest.theAssetWriterIsStillUnreachableFromProduction` fails if **any** production
  class depends on `CryptoAssetWriter`. It is the tripwire that forces `IdentityRuleset.VERSION` to be
  bumped when ingest is wired, and a sweeper reaching for that bean would spring it early for a reason it
  was not built to catch. Its documented remedy — bump the constant, delete the test — is exactly what this
  story must not do.
- `TransactionalBoundaryArchTest` Rule D requires every public method of a `@Service` in
  `..service.writer..` to be effectively `REQUIRED`, which a batch-scoped `REQUIRES_NEW` method is not.

`SigningRecordWriter.deleteExpiredBatch` is still the shape copied, and it escapes Rule D only because that
bean is a `@Component` — so this one is too. **The ticket's AC #8 asked for a `REQUIRES_NEW` entry point on
`CryptoAssetWriter`, which is unimplementable on this branch; the issue body has been corrected.** AC #8's
other clause — "verdicts are written with the existing `CryptoAssetRepository.applyPqcVerdict` SQL" — is
also inaccurate, and deliberately so: see the two statements below.

## Three correctness properties that only exist in a database

**The row version.** The sweep reads rows in its outer transaction and writes them in a later one, and a
concurrent writer can land in that window. The guard has to refuse a verdict computed from the earlier read,
and it guards on `xmin` — the transaction that wrote the row's current version — because `i_upd` cannot. A
first version of this PR compared `i_upd`, and review measured what that costs: `CURRENT_TIMESTAMP` is
`transaction_timestamp()`, microsecond resolution and non-monotonic, and **2.0 % of barrier-synchronised
transaction pairs share it exactly**. An equality check on a shared value is not a version check, and a
payload landing from such a transaction passed it, was overwritten by a verdict computed without it, and was
stamped current — the shape of failure nothing would find. `findStaleVerdictRows` is therefore native, so it
can project `xmin::text::bigint` in the **same statement** as the columns the verdict is computed from: the
sweep's transaction is READ COMMITTED, and a second statement would see a later snapshot, which is the same
hole in a different place. A tuple frozen between read and write reads `xmin` as 2 and can only cause a
spurious refusal the next sweep retries.

**What counts as stale.** Either half of the contract: a verdict from an older generation, or a verdict
older than the row it describes (`pqc_evaluated_at < i_upd`). Version alone leaves every asset whose merged
payload changes after its first sweep — a first source arriving after an identity-only upsert, a richer
source winning the election, a `detachCbom` re-election — carrying a verdict computed from the old payload
until the next generation bump. The guarded write sets both timestamps to one `CURRENT_TIMESTAMP`, so a
verdict does not re-offer itself, while any later writer does. The guard's staleness clause is widened
identically, because a row offered for a moved payload is already at the current generation and a
version-only clause would refuse every write the wider list asks for. The cost, accepted: identity refreshes
re-offer rows on every re-sync, bounded by the cap.

**The keyset cursor.** A written row leaves the work list, so an offset would skip exactly as many unread
rows as the previous batch wrote. Re-querying from the start instead terminates only if *every* claimed row
leaves — and a row the guard refuses does not: it would return at the head of every batch and strand the
rest of the table behind it, sweep after sweep. Ordering by uuid and carrying the last one forward makes the
sweep advance whatever happens to an individual row.

## The two statements stay two

`applyPqcVerdict` is ingest's write, `applyPqcVerdictIfStale` the sweep's, and their `SET` lists are
byte-identical: the `pqc_decided_at` `CASE … IS DISTINCT FROM`, `pqc_evaluated_at`, the `jsonb` cast, the
`i_upd` bump. One statement with the guards made optional would turn them into nullable parameters, which
reads worse than a duplicate the build refuses to let diverge — so `CryptoAssetPqcVerdictSqlTest` reads both
`@Query` values by reflection and asserts the text between `SET` and `WHERE` is identical.

## Failure, per row rather than per batch

A batch shares one `REQUIRES_NEW` transaction, so one row the database refuses rolls back every good write
beside it — and since the cursor restarts at the nil uuid every sweep, a refusal that persists is reached
first on every run and starves the whole tail. On a `RuntimeException` from the batch the same rows are
retried one per transaction; a row that still fails is counted, logged with its uuid alone, and the loop
continues.

A row whose *evaluation* throws is stamped `UNKNOWN / EVALUATION-FAILED` at the current generation, so it
leaves the work list and shows up in the inventory rather than only in logs.

## The trigger

A system `ScheduledJobTask` registered in `SystemScheduledJobs`, not a `@Scheduled` class. Every
cluster-locked sweep shipped so far uses `@Scheduled`, which is simpler and reaches nobody: under SaaS an
operator cannot touch yml, so such a sweep can never be paused. Registered as a system job it is reachable
through the Scheduler API — `changeScheduledJobState` carries no `isSystem` guard, while `updateScheduledJob`
refuses to edit its cron, which is the shape wanted (pause yes, re-tune no). The route needs the external
scheduler, which is not a new dependency: `SystemScheduledJobs.registerJobs` already calls it for
`CbomSyncTask`, and without ingest there is no inventory to sweep.

`performJob` is deliberately **not** `@Transactional`: `SchedulerListener` is already class-level
transactional, so the sweeper's own `REQUIRES_NEW` is the second in the chain and a third would hold a
connection idle for the whole run. No-op runs throw `ScheduledJobSkippedException`, or an hourly job would
write a SUCCESS history row forever for doing nothing.

It does catch, though. A failure of the sweep's own backend — a failover, a pooler restart, an
`idle_in_transaction_session_timeout` killing a session that is idle between batches — used to propagate,
roll the listener's transaction back with the STARTED history row inside it, and leave **no row at all**
where an operator would look for a failure. The catch runs on the listener's still-live connection and its
text is fixed, because an exception message here reaches the Scheduler API.

## What the history row says

The only place an operator learns anything, so it separates the four outcomes rather than implying them:

```
Read %d stale cryptographic asset(s) in %d batch(es); %d verdict(s) written, of which %d recorded as
UNKNOWN because the rule set could not be evaluated; %d refused and left for the next sweep; %d could
not be written
```

A refused row is ordinary — another writer got there first — and does not fail the run. A failed write or a
sweep that stopped early does. `applyStaleBatch` returns *which* rows it wrote, so "recorded as UNKNOWN" is
true of exactly the stamps that committed: an earlier version counted the stamps it had built, which a
refusal or a rollback made false.

## `PqcRuleset.VERSION` is 5

#2227's Yarrow and CMEA right guards move verdicts for nine glued spellings, and this constant's contract is
to move whenever a rule changes any row's verdict, rule id or evidence. Bumping is free — no row has ever
been stamped, since `CryptoAssetWriter` still has no production caller — and cheaper than spending the
documented `IdentityRuleset` exemption to save an integer.

Both `why` texts now name the direction of the error rather than only its size. As sized symmetric key
material `YarrowAES` reads `READY / MATERIAL-SYMMETRIC-READY` where it read `NOT_READY / CLASSICAL-LEGACY`,
because the material arms decide once no legacy family is named — a false `ready` on a glued legacy token
replacing a false legacy finding on a name that was merely a word (`Yarrowed`). `CMEAS` carries the same
sentence. Synthetic only: no in-repo fixture and no corpus row holds any Yarrow or CMEA spelling. The
regenerated artifact digest is `474e3dc95c5e…`, and all three pins move in the same commit.

## Verification

- `CryptoAssetPqcSweepITest` **10/10 against real PostgreSQL**, `PqcVerdictSweeperTest` 9/9,
  `CryptoAssetPqcSweepTaskTest` 7/7, `CryptoAssetPqcVerdictSqlTest` 1/1, and **1 070 tests / 0 failures**
  across `CryptoAsset*`, `Pqc*`, `Identity*`, `Normalization*`, `SchedulerServiceITest` and every
  `*ArchTest` — after merging `origin/main`.
- **Every new pin red-greened by mutation.** Seven mutations, each reddening exactly the test named for it:
  the `xmin` clause neutralised (two payload-window tests), the staleness clause deleted, the work list
  narrowed to version-only, the cursor line removed, the per-row retry removed, the stamp count taken from
  what was built rather than what landed, and one `SET` line drifted between the two statements.
- `IdentityRulesetStampArchTest` and `TransactionalBoundaryArchTest` green **without** relaxing either;
  `NativeQuerySchemaArchTest` and `ObjectMapperCentralizationArchTest` green over the new native query and
  the hoisted column mapper.
- `ContextSignatureGuardTest.BASELINE` unchanged — the ITest reuses an existing context, and the two new
  test classes boot no Spring context at all.
- No migration, no new column, no repository `@Transactional`, no contract change, `CbomSyncTask` untouched.

## Not in this PR

- **The verdict-distribution measurement (AC #10).** Undischarged, and with no remaining blocker: the
  material-scope question it used to wait on shipped in #2227's `8caec99a`. It needs a run over the corpus
  copy #2196's dossier ratifies (`documents=200 parsed=196 components=8048 keyed=4994`), with `unknown`
  scored against a denominator that excludes the correctly-`notApplicable` names. It stays open on #2151.
  The caveat behind the original scope finding stays on the record: "993 of 1 478 material components name a
  family in their own name" counted family tokens in component **names**, which is not what the pipeline
  stores, and material's algorithm properly arrives via `algorithmRef`, resolved against the document at
  ingest and never from a stored row. #2227's fallback derives from the one column that *is* stored.
- **No index serves the work-list query.** Correctness is unaffected and the table is empty today. Left
  recorded rather than guessed into a migration — and the predicate it has to serve is only now settled,
  which was the reason for waiting.
- **Helm charts.** `CRYPTO_ASSET_PQC_SWEEP_BATCH_SIZE` and `CRYPTO_ASSET_PQC_SWEEP_MAX_BATCHES_PER_SWEEP`
  need wiring in the charts repo; this repo cannot close that. The second name changed during the review
  round — relaxed binding already accepted the long spelling and it outranked the short one, so two live
  spellings existed and the documented one lost silently.
- **Three pre-existing findings surfaced while reviewing**, recorded in
  https://github.com/OmniTrustILM/core/pull/2228#issuecomment-5582374667: `IPsec` deciding
  `CLASSICAL-SHOR-COMPONENT` from the `ec` inside its own name, `arcfour` electing no family, and
  `PqcRuleInput.parameterSet` being read by no rule.

Refs #2151


src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
```diff
@@ -182,6 +186,53 @@ void applyPqcVerdict(@Param("uuid") UUID uuid, @Param("verdict") String verdict,
...
+     * longer the one the sweep read has changed under it, and the verdict computed from the earlier read is not
+     * written. The version clause restates what the work list calls stale. A refused row keeps its old generation and
+     * is offered again on the next sweep; a touch that changed nothing the rules read costs it that one wait.
+     *
+     * @return 1 if the row was written, 0 if it was touched since it was read or is no longer stale
+     */
+    @Modifying
+    @Query(value = """
+            UPDATE {h-schema}crypto_asset
+            SET pqc_verdict = :verdict,
```
karolhrdina, line 266: 🟡 **This `SET` list is a verbatim copy of `applyPqcVerdict`'s eleven lines above, and nothing pins the two equal.**

Measured: the two statements' `SET` clauses are byte-identical (11 lines each); they differ only in the `WHERE`:

```
applyPqcVerdict         WHERE uuid = :uuid
applyPqcVerdictIfStale  WHERE uuid = :uuid
                          AND (pqc_ruleset_version IS NULL OR pqc_ruleset_version < :rulesetVersion)
                          AND i_upd = :updatedAsRead
```

AC #8 in #2151 says *"Verdicts are written with the existing `CryptoAssetRepository.applyPqcVerdict` SQL."* They are written with a second statement that duplicates it. The `SET` list *is* the two-timestamp contract — the `pqc_decided_at` `CASE … IS DISTINCT FROM`, `pqc_evaluated_at`, the `jsonb` cast, the `i_upd` bump — and `CryptoAssetInventoryITest` pins it for the first statement only. No test compares the two. The day someone edits one `CASE` and not the other, ingest and the sweep write different row shapes for the same verdict, `decidedAt` dates itself differently depending on which path wrote the row, and nothing goes red.

**Fix** — either one statement with the two guards made optional, so both writers share it:

```sql
WHERE uuid = :uuid
  AND (CAST(:updatedAsRead AS timestamptz) IS NULL OR i_upd = :updatedAsRead)
  AND (CAST(:guardVersion AS int) IS NULL OR pqc_ruleset_version IS NULL OR pqc_ruleset_version < :guardVersion)
```

or, cheaper and honest about the duplication: a unit test that reads both `@Query` values by reflection and asserts the text between `SET` and `WHERE` is identical, so a divergence at least fails the build.

src/main/java/com/otilm/core/cbom/pqc/PqcEvaluator.java
```diff
@@ -314,8 +299,8 @@ public PqcRuleInput fromNormalized(NormalizedAsset asset, JsonNode cryptoPropert
      * The material tier derives no family -- {@code AssetNormalizer} leaves it null for every
      * {@code related-crypto-material} component, with or without an {@code algorithmRef} -- so a private key whose own
      * name says {@code RSA-2048} reached the rules with nothing to classify. The name is a column, so reading the
-     * family out of it is available identically to both callers. Confined to material: on an algorithm row a null
-     * family is the normalizer's decision, a cipher suite above all, and stands.
+     * family out of it is available to every caller. Confined to material: on an algorithm row a null family is the
+     * normalizer's decision, a cipher suite above all, and stands.
```
karolhrdina, line 303: 🟡 **The body's "scope finding that needs a decision" was decided before this PR — this Javadoc describes the decision — and with it the stated blocker for the distribution AC is gone.**

Body: *"The material tier derives no algorithm family at all … Options: narrow v1 to algorithms, defer asymmetric material under its own rule id as certificates are, or add a derivation."* And, under **Not in either PR**: *"The verdict-distribution measurement (AC #10) … the distribution it would report is a function of the material-scope decision above — measuring first would publish a number that changes."*

But five lines below this comment, on `main` and unchanged here:

```java
if (family == null && fields.assetType() == CryptographicAssetType.RELATED_CRYPTO_MATERIAL) {
    family = ratifiedFamily(normalizer.familyFromName(fields.name()));
}
```

The derivation was added in #2227's `8caec99a` ("Adjudicated your way"). The option the body lists as open is the one already shipped, the section is stale, and the reason given for not measuring AC #10 no longer holds. AC #10 — *"the four-value verdict distribution is measured, recorded in the PR body and pinned as a fixture … names the corpus artifact it ran over … `unknown` scored against a denominator that excludes the correctly-`notApplicable` names"* — is therefore undischarged in both PRs with no remaining stated blocker. You have been candid that it is not in either PR; the point is only that the reason is out of date and the criterion still stands open on the ticket, which is the one place a reader will look for it.

karolhrdina (commented):
## Review — the sweep is sound where it is hardest; the findings are at its edges

Verified at `2f7f99af`, seven parallel passes in isolated worktrees against real PostgreSQL 15
(Testcontainers), every pin the body names red-greened by mutation, every claim below reproduced.
Baseline on this head: `Pqc*Test,NormalizationRulesTest,IdentityTablesTest,IdentityVectorTest` **872 / 0**,
`CryptoAssetPqcSweepITest` **8 / 0**, `PqcStoredRowRoundTripITest` **8 / 0**. All seven worktrees left clean.

What is right, first, because it is the substance. The trigger/sweeper split with the `@Component` writer
escaping Rule D exactly as `SigningRecordWriter` does — and Rule D's *purpose*, an ambient transaction for
every `@Modifying` call, is met, not dodged (mutating the writer to `@Service` reddens Rule D; adding a
`CryptoAssetWriter` dependency reddens the tripwire). The keyset cursor over Postgres-side uuid ordering: seeded
across both sign bits, 13/13 written exactly once, and the Java side never compares a `UUID` (whose
`compareTo` disagrees with Postgres). The advisory lock: two real concurrent sweeps, one ran, one skipped in
1.9 ms, no double write. The topology is exactly as described — `pg_stat_activity` shows three backends, two
`idle in transaction`, one active on the `UPDATE`. The two-timestamp contract holds through the *new* SQL for
every transition (`READY→READY` keeps `decided_at`; `READY→NOT_READY`, `READY→EVALUATION-FAILED` move it;
`UNKNOWN→EVALUATION-FAILED` keeps it). The `i_upd` guard goes beyond the AC and closes a hole the AC's version
predicate alone cannot see, and the eighth ITest proves it. Termination and the cap are exact. The failure path
threw nothing over 280 vectors, 53 corpus assets and 129 600 synthetic rows. The parity-test deletion lost
nothing — its successors redden where it could not. `PqcStoredRowRoundTripITest` keeps the property it was
renamed for: making `ingest` evaluate the pre-upsert derivation reddens both multi-producer tests. The table
regeneration is byte-reproducible with all three pins equal to the real digest, and nothing in any in-repo
fixture moves.

**One critical.** The guard is a **timestamp**, and two transactions that begin in the same microsecond share
`CURRENT_TIMESTAMP`. Reproduced with the production writer beans and nothing forged: a co-start loop hit an
equal `now()` on attempt 69 (0.3 s); T1 wrote a verdict, the sweep read the row, T2's `upsertSource` landed
with a byte-equal `i_upd`, and the guarded write **passed** — `written = 1`, the row left the work list carrying
a verdict computed without its payload. 3 000 barrier-synchronised pairs collide 2.0 % of the time. The window
per row is a whole batch. On this branch the sweep is the only writer so exposure is narrow; the moment #2073
wires concurrent ingest, two writers straddling a sweep read become routine, and this is the "failure nothing
would find" the body itself describes. `xmin` is a write-unique token; a timestamp is not.

**The risks, each reproduced against Postgres or by mutation, in the order I would take them:**

1. **A failure on the sweep's own connection erases the run from history.** Terminate the advisory-lock
   holder mid-run: the writer's batch commits, the next read fails, the sweeper returns `aborted=true` — and
   then the `REQUIRES_NEW` proxy throws on rollback, `runScheduledJob`'s `finally` finalizes nothing, and the
   listener's rollback deletes the `STARTED` row. **Zero history rows**, 5 of 12 assets written. The
   `aborted → FAILED` contract covers only failures inside the writer's transaction. `performJob` is also
   entirely uncovered (lines 72–86, JaCoCo). Four-line fix.
2. **A payload landing after the stamp never re-enters the work list.** `UNKNOWN / FAMILY-UNRESOLVED` at v4,
   then the `size: 256` source arrives, `i_upd` moves — and the row is not stale, the rules say
   `READY / MATERIAL-SYMMETRIC-READY`, the stored verdict says `UNKNOWN` until the next bump. By the AC's
   version-only design and #2073's remit, but live today, since the sweep is the only writer.
3. **The cursor advance is unpinned.** Delete `cursor = rows.get(…)` and all 14 sweep tests stay green — a
   stuck cursor is a third shape the strengthened cursor test cannot see, because written rows leave the list.
   What the line buys: with one guard-refused page of five ahead of eight rows, shipped code stamps 8/8 in
   3 batches; without it, 20 batches, **0 written**, and `SUCCESS "Swept 100 … 0 verdict(s) written"`.
4. **The version clause is unpinned, and the body's account of it is wrong.** Delete
   `AND (pqc_ruleset_version IS NULL OR …)` and both suites stay green, because every writer also moves
   `i_upd` so the `i_upd` clause refuses alone. `theGuardedWriteRefusesABatchBuiltFromAStaleRead` was
   written, the body says, to catch exactly that deletion; it cannot. The `i_upd` clause *is* pinned — by the
   eighth, body-unmentioned test.
5. **The outcome record cannot tell a stamped failure from a refused one**, so "were recorded as UNKNOWN" is
   unknowable from the data the message is built from (copilot's line-78 comment, widened from the string
   to the record), a run whose every write was refused reports `SUCCESS`, and `failed > 0` is unreachable
   from any stored data — nine hostile rows all evaluate cleanly.
6. **One failing `UPDATE` rolls back its batch and starves the tail** while the cause persists — the
   starvation the body reports fixed for evaluation failures, reintroduced on the write side. Not reachable
   today (the CHECK matches the enum exactly; the reachable deadlock is transient and self-heals); needs a
   future drift such as a fifth verdict constant shipped without its migration.
7. **The sweep's `UPDATE` duplicates the ingest `UPDATE`'s eleven-line `SET` list byte-for-byte** and nothing
   pins the two equal. AC #8 said "written with the existing SQL". One `CASE` edited on one side and not the
   other is state divergence no test would see.
8. **The body's "scope finding that needs a decision" was decided before this PR** — the `familyFromName`
   fallback it lists as an option is on `main` at `PqcEvaluator.fromStoredRow:322-324` — and with it the stated
   blocker for the distribution AC (#10) is gone. The criterion stands open on the ticket with no remaining
   reason given.
9. **The two right guards move nine glued spellings** (`Yarrowed`, `YarrowAES`, `CMEAlgorithm`, `CMEAS`, …)
   without a `PqcRuleset.VERSION` bump — the first table move in that constant's history without one — and
   their material-tier mirror is `YarrowAES` as a sized `secret-key` flipping `notReady → ready`, which the
   recorded cost ("elects nothing") does not name. Synthetic only.

### Conformance against #2151

AC #2 (sweep configuration), #8 (the writer bean) and #9 (the sweep) are implemented as specified, with two
deviations in the PR's favour (the `i_upd` guard; a keyset cursor where the AC's template query had none) and
one against (the duplicated SQL, item 7). `changeScheduledJobState` has no `isSystem` guard and
`updateScheduledJob` refuses system jobs — "pause yes, re-tune no" reproduces; note `enabled` is never
consulted at execution, so pause relies on the external scheduler withholding the message (pre-existing,
scheduler-wide). `@Validated` on the record properties genuinely enforces: `batch-size: 0`, `-1` and a
non-numeric value all refuse startup, through env vars too. AC #10 (the distribution) is undischarged in
both PRs, as the body says, and its stated blocker no longer exists (item 8). AC #11: Sonar gate passed,
84.2 % on new code, `BASELINE` 61 reproduces (census stays 61 with both new ITests deleted).

### Numbers in the body

| claim | measured |
|---|---|
| "`CryptoAssetPqcSweepITest`: **7/7** against real PostgreSQL" | **8/8** — the eighth is the only pin on the `i_upd` clause |
| "`PqcVerdictSweeperTest` 6/6", "`BASELINE` still 61" | reproduce |
| "**295** tests on this branch" | reproduces as `Pqc*Test` + `NormalizationRulesTest` minus the separately quoted sweeper test (145+16+45+76+10+3); excludes `PqcStoredRowRoundTripITest` (8) and `IdentityTablesTest` (22), both in the diff |
| "**637 lines, one commit** — under the 1K floor" | one commit, yes; **+1104 / −218 over 29 files**. No pushed head ever measured 637 (GitHub stats: +748…750/−3 while stacked, then +1094, +1104). Stale — it predates the rebase, the parity deletion, the stored-row tests and the regeneration |
| "stacked on #2227, review that first" | #2227 is merged (`ce0890d9`); the base is `main` |
| "No migration, no new column, no repository `@Transactional`, no contract change, `CbomSyncTask` untouched" | all reproduce |
| "`IdentityRulesetStampArchTest` and `TransactionalBoundaryArchTest` green **without** relaxing either" | reproduces — both files unchanged, both red under the described mutations |
| "the task records FAILED" on an aborted sweep | reproduces for writer-side failures; **differs** for a failure of the sweep's own transaction (item 1) |
| "a row whose evaluation throws … leaves the work list" | reproduces when the write lands; **differs** when the guard refuses it or the batch aborts (item 5) — and the body also says "skipped" two paragraphs apart from "stamped"; the code stamps |
| copilot #2 (env var name) | confirmed, and sharper: relaxed binding *already* accepts `CRYPTO_ASSET_PQC_SWEEP_MAX_BATCHES_PER_SWEEP` and it **outranks** the placeholder's short name when both are set — two live spellings, the documented one losing silently. The one-line rename is the fix |

### Wiring, wire-safety and the suite — and a correction to my own #2227 advice

On #2227 I asked for `@Lazy` on the evaluator's beans so a bad `cbom/identity-tables.json` would not refuse
the whole application at boot, and you applied it. **It was inert, and this PR does not undo it because the
property never held.** A pre-existing production `@Component`, `ContextRefreshListener.handleContextRefresh:81`
(`git blame`: years old), calls `getBean` on every singleton bean definition at `ContextRefreshedEvent`,
lazy ones included — so `IdentityTables.load()` has always run at context start in this application.
Measured both ways: as shipped, a corrupted artifact refuses `ApplicationITest`'s context through
`pqcVerdictSweeper → pqcEvaluator → assetNormalizer → IdentityTables.load()`; with `@Lazy` on the sweeper's
injection point too, boot is *still* refused, via the listener. `PqcConfig`'s Javadoc claims a property the
app never had; that is documentation, not a finding, and the correction is mine to make. One trap to record
so nobody "fixes" this the obvious way: if the load *were* deferred to the first sweep, a
`BeanCreationException` is a `RuntimeException` caught per row — measured: `evaluated=0, written=3, failed=3`,
every stale row permanently stamped `EVALUATION-FAILED` at the current generation. A deferred load must be
resolved once per sweep, outside the per-row try. As shipped, the per-row try holds no I/O and no bean
resolution, which is the right shape.

Wire and logs are clean: no `getMessage()` in new code; `EVALUATION-FAILED`'s reason and id are literals,
its evidence renders as `{}` and lands in `jsonb`; the history message carries counts only; no
`PqcStaleVerdictRow` component is a key, no reachable exception message carries a producer value, no new
file is in `SOURCE_ALLOWLIST`, and the whole `com.otilm.core.architecture` package is green
(`IdentityKeyExposureFence*` 52/52). Full suite **7 908 tests, 0 failures, 3 errors** — the two deterministic
`ProxyPropertiesTest` hostname errors from the last round and one `JmsNetworkChaosITest` toxiproxy
container start (8/8 on isolated re-run); nothing attributable to the PR. Sonar: gate passed, 84.17 %,
**11 of 105 new lines uncovered, all in `performJob` (72–75, 77–80, 83, 84, 86)** plus one partial condition
at `PqcVerdictSweeper:104`; local JaCoCo reproduces it line for line. Generator regeneration byte-identical,
`68f01858…` before and after.


### Two things not this PR's fault

`main` currently compiles only against an `interfaces` snapshot newer than `-56` (it needs #2211's
group-users method) and older than `-59` (which added `listSupportedTokenProfileKeyUsages`, unimplemented by
`main`'s `TokenProfileControllerImpl`). This branch's last CI run at 13:49Z predates `-59` at 14:04Z by
fifteen minutes; a re-run today goes red on both `main` and this branch until core#2212 lands. I rebuilt
`interfaces@06b63eee` from source to review. And `ObjectMapperFactory.jsonColumn()` is constructed per row in
`evaluate` — 736 µs against 4 µs cached — about 7 s per 10 000-row sweep spent inside the lock-holding
transaction; a `static final` field removes it.

### Method

Seven passes: the sweeper loop and cursor; the guarded write, SQL and transactions against real Postgres
(barrier-synchronised co-starting transactions, injected triggers, terminated backends, provoked deadlocks in
both directions); the scheduled task and configuration end to end through the real `SchedulerListener`; the
table regeneration and both version constants over 1 302 inputs under four classpaths; the tests, by mutating
every production line the body names a pin for; wiring, wire-safety and the full suite; and conformance
against #2151. Conducted with Claude (Opus 5); every inline comment reproduced at `2f7f99af`.

src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
```diff
@@ -182,6 +186,53 @@ void applyPqcVerdict(@Param("uuid") UUID uuid, @Param("verdict") String verdict,
...
+                pqc_evaluated_fields = CAST(:evaluatedFields AS jsonb),
+                pqc_evaluated_at = CURRENT_TIMESTAMP,
+                pqc_decided_at = CASE
+                    WHEN crypto_asset.pqc_verdict IS DISTINCT FROM CAST(:verdict AS TEXT) THEN CURRENT_TIMESTAMP
+                    ELSE COALESCE(crypto_asset.pqc_decided_at, CURRENT_TIMESTAMP)
+                END,
+                i_upd = CURRENT_TIMESTAMP
+            WHERE uuid = :uuid
+              AND (pqc_ruleset_version IS NULL OR pqc_ruleset_version < :rulesetVersion)
+              AND i_upd = :updatedAsRead
```
karolhrdina: 🔴 **The guard is microsecond-timestamp equality, and two transactions that begin in the same microsecond share `CURRENT_TIMESTAMP` — a payload landing under the sweep from such a transaction passes the guard and is silently overwritten.**

`CURRENT_TIMESTAMP` is `transaction_timestamp()`: microsecond resolution, non-monotonic, and identical for backends whose transactions start together. This clause compares that value for equality, so it is a *value* check, not a version check.

Reproduced against real PostgreSQL 15 with the production writer beans and nothing forged. A loop starts T1 and T2 at a barrier and retries until their first `SELECT now()` agree — **first equal pair at attempt 69, 0.3 s**. Then:

```
T1  assetWriter.applyPqcVerdict(A, READY, "T1-OLD-GEN", 3)          commits   i_upd = 17:45:19.3218
    sweep reads A                                                   updated = 2026-09-08T15:45:19.321800Z
T2  sourceWriter.upsertSource(A, cbom, {secret-key, size 64})       commits   i_upd = 17:45:19.3218   (equal)
    verdictWriter.applyStaleBatch([NOT_READY @ updated], 4)         -> written = 1
row: source_count=1, pqc_rule_id=PROBE2-FROM-PRE-PAYLOAD-READ, pqc_ruleset_version=4, no longer on the work list
```

Expected — the contract `aPayloadLandingBetweenReadAndWriteKeepsTheRowOnTheWorkList` pins — `written = 0`. How often nature produces the collision: 3 000 barrier-synchronised transaction pairs → **61 (2.0 %) identical `now()`**, mean skew 20 µs. The per-row read→write window is a whole batch (260 ms per 500 rows locally), and after a rule-set bump every row of the inventory passes through such a window.

This is the failure the body itself describes as the one nothing would find: a verdict computed without the payload, stamped current, and never revisited until the next generation. On this branch the sweep is the only writer, so the window is narrow; the moment #2073 wires concurrent ingest, two writers straddling a sweep read become routine. I have rated it critical on that basis rather than on today's exposure.

The other ordering is **not** a defeat and I checked it: a writer that started before the sweep's read and commits after it moves `i_upd` *backwards* (`.627874 < .928146`), and equality refuses it.

**Fix** — compare a write-unique token, not a timestamp. `xmin` changes on every `UPDATE` and no two transactions share an xid: make the work list a native query projecting `xmin::text::bigint AS row_version`, carry `long rowVersion` in `PqcStaleVerdictRow`/`PqcVerdictWrite` in place of `updated`, and guard with `AND xmin::text::bigint = :rowVersion`. A tuple frozen between read and write reads `xmin` as 2 and can only cause a spurious refusal (retried next sweep), never a false pass. One-line alternative with a caveat: `i_upd = clock_timestamp()` in the four writer statements makes a later writer's stamp strictly later (writers serialise on the row lock), but changes `i_upd` semantics for every writer and departs from the `CURRENT_TIMESTAMP` convention.

src/main/java/com/otilm/core/tasks/CryptoAssetPqcSweepTask.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * A run that swept nothing is a skip: until ingest gains a caller this fires hourly and finds nothing, and a
+     * SUCCESS row an hour would bury the runs that did something. A row the rule set threw on was swept -- it now
+     * carries an UNKNOWN verdict at the current generation and no later sweep will revisit it -- so a run with any such
+     * row is a failure the operator has to see, as is an aborted sweep, whatever the earlier batches managed to write.
+     */
+    @Override
+    public ScheduledTaskResult performJob(final ScheduledJobInfo scheduledJobInfo, final Object taskData) {
+        PqcVerdictSweeper.SweepOutcome outcome = sweeper.sweep();
```
karolhrdina: 🟡 **A failure on the sweep's own connection mid-run leaves no history row at all — not FAILED, nothing.**

The `aborted → FAILED` contract added after the adversarial round covers exceptions thrown by `applyStaleBatch`. Any exception that leaves `sweeper.sweep()` *itself* takes a different path. Reproduced against real PostgreSQL through the real `SchedulerListener`: 12 stale rows, the sweep mid-run, and the sweep transaction's backend — the one holding the advisory lock, `idle in transaction` between batches — terminated (`pg_terminate_backend`, standing in for a failover, a pooler restart, an `idle_in_transaction_session_timeout`, or an ops runbook that kills idle-in-transaction sessions):

```
writer's batch commits (5 rows written)
next findStaleVerdictRows fails; the sweeper's catch sets aborted=true and returns normally
the REQUIRES_NEW proxy cannot end its transaction on the dead connection and throws:
   JpaSystemException: Unable to rollback against JDBC Connection ... Connection is closed
performJob propagates; runScheduledJob's finally sees result == null && !skipped -> finalizes nothing
the listener's transaction rolls back and takes the STARTED history row with it
=> scheduled_job_history: 0 rows.  assets written=5, stale left=7
```

In production the JMS catch-all logs "Unexpected error in endpoint 'schedulerListener'" and consumes the message. An operator paging job history sees a gap, not a failure, over a half-swept table. The next hourly run resumes correctly (the cursor restarts at the nil uuid), so this is visibility rather than data — but it is exactly the run the Javadoc above says "the operator has to see".

Contrast, verified: a failure *inside* the writer's transaction does reach `FAILED` (`SweepOutcome[aborted=true, evaluated=5, written=0, batches=0]` → `FAILED "The sweep aborted before completing…"`).

**Fix** — catch around the call, on the listener's still-live connection, with a fixed message (no exception text on the wire; `CbomSyncTask` puts `e.getMessage()` into history — do not copy that):

```java
PqcVerdictSweeper.SweepOutcome outcome;
try {
    outcome = sweeper.sweep();
} catch (RuntimeException e) {
    log.error("PQC verdict sweep failed before it could report its outcome", e);
    return new ScheduledTaskResult(SchedulerJobExecutionStatus.FAILED,
            "The sweep failed before it could report its outcome; see the application log", Resource.CRYPTO_ASSET, null);
}
```

Nothing on the skip path changes (`ScheduledJobSkippedException` is thrown after the call). The class has no logger today.

Also, measured for the record: this whole method is **uncovered** — JaCoCo over the PR's own tests leaves lines 72–86 unexecuted, which is the entirety of `performJob`; the sweeper, writer, properties and registration are at 100 %.

src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
```diff
@@ -182,6 +186,53 @@ void applyPqcVerdict(@Param("uuid") UUID uuid, @Param("verdict") String verdict,
...
+     * The sweep's work list. Keyset-cursored, and that is correctness rather than performance: a written row leaves
+     * this result set, so an offset would skip exactly as many unread rows as the last batch wrote, and re-querying
+     * from the start only terminates if every claimed row leaves -- one that throws would return at the head forever.
+     */
+    @Query("""
+            SELECT new com.otilm.core.model.cbom.PqcStaleVerdictRow(a.uuid, a.assetType, a.name, a.oid,
+                a.algorithmFamily, a.primitive, a.parameterSet, a.curve, a.mode, a.padding, a.variant,
+                a.mergedCryptoProperties, a.updated)
+            FROM CryptoAsset a
+            WHERE (a.pqcRulesetVersion IS NULL OR a.pqcRulesetVersion < :version) AND a.uuid > :after
```
karolhrdina: 🟡 **A payload that lands after a row is stamped current never re-enters this work list; `i_upd` guards the write window, not staleness.**

Real PostgreSQL: a `related-crypto-material` row `vault-key-2024` upserted with no source → `sweep#1 evaluated=1 written=1` → `UNKNOWN / FAMILY-UNRESOLVED` at version 4. Then `upsertSource(… {type: secret-key, size: 256} …)` — `i_upd` moves (`.027103 → .036280`), `merged_crypto_properties` is populated, and **the row is not on the work list**: `sweep#2 evaluated=0 written=0 batches=0`. Evaluating the row as it now stands: `READY / MATERIAL-SYMMETRIC-READY {materialType=secret-key, materialSize=256}`. The stored verdict stays `UNKNOWN` until `PqcRuleset.VERSION` bumps.

On this branch ingest does not evaluate (#2073 has not landed), so the sweep is the *only* verdict writer — and every asset whose merged payload changes after its first sweep keeps a verdict computed from the old payload: the first source arriving after an identity-only upsert, a richer source winning the election, a `detachCbom` re-election. The ITest Javadoc for `aPayloadLandingBetweenReadAndWriteKeepsTheRowOnTheWorkList` says a verdict/payload contradiction "would stand until the next generation bump" about the millisecond read→write window; the same is true of the unbounded window after the stamp.

Scoping, stated fairly: #2151 specifies the version-only work list and assigns per-unit-of-work evaluation to #2073, so this is a design gap the PR inherits and makes live, not a coding slip. Two honest dispositions:

- widen the list — `OR a.pqcEvaluatedAt < a.updated`. `applyPqcVerdictIfStale` sets both to the same `CURRENT_TIMESTAMP`, so a verdict does not re-offer itself, while any later writer (`upsertIdentity` on conflict, `recomputeMergeFromSources`) does. Mirror to weigh: identity refreshes then re-offer rows on every re-sync, bounded by the cap, and the deferred work-list index becomes more pressing;
- keep the version-only list and record — here and in #2073 — that *every* source change (upsert **and** detach) must re-evaluate in the same unit of work, and say in the body that until #2073 lands verdicts lag payloads.

src/main/java/com/otilm/core/cbom/pqc/PqcVerdictSweeper.java
```diff
@@ -0,0 +1,156 @@
...
+        int batches = 0;
+        boolean aborted = false;
+        try {
+            List<PqcStaleVerdictRow> rows;
+            do {
+                rows = assetRepository.findStaleVerdictRows(PqcRuleset.VERSION, cursor, PageRequest.of(0, batchSize));
+                if (rows.isEmpty()) {
+                    break;
+                }
+                cursor = rows.get(rows.size() - 1).uuid();
```
karolhrdina, line 92: 🟡 **This line is pinned by no test — delete it and all 14 sweep tests stay green — and it is what stops a refused page from starving the table.**

Mutation: line 89 removed (cursor stays at the nil uuid). `PqcVerdictSweeperTest` **6/6 green**, `CryptoAssetPqcSweepITest` **8/8 green**, `theSweepAdvancesAcrossBatchesWithoutSkippingARow` included. The body says that test was strengthened so "offset paging or `while(false)`" would fail it — and both of those *do* redden it (verified). A stuck cursor is a third shape it cannot see: written rows leave the work list, so re-querying from nil returns the next page anyway.

The cursor is load-bearing exactly when a row *stays* on the list after its batch — a guard refusal — which no test seeds. Measured on real Postgres with five rows the guard can never land (a `BEFORE UPDATE … RETURN NULL` trigger) given the five lowest uuids, ahead of eight ordinary rows:

```
shipped   sweep#1 evaluated=13 written=8  batches=3    normal rows 8/8 stamped, refused 0/5
          sweep#2 evaluated=5  written=0  batches=1    refused page re-evaluated once per sweep, bounded
line 89
removed   sweep#1 evaluated=100 written=0 batches=20   (the cap)  normal rows 0/8 stamped
          sweep#2 identical
```

Eight rows behind the refused page get no verdict on any run, while the job reads `SUCCESS "Swept 100 cryptographic asset(s) in 20 batch(es); 0 verdict(s) written"`.

**Fix** — in `PqcVerdictSweeperTest`: two full pages, writer returning 0, an `ArgumentCaptor` on the `after` parameter, assert `[nil, last(page1), last(page2)]` and `times(3)`. Optionally the ITest variant with a refused row. Three other unit mutations also stay green, for the record — short-batch exit, empty-page break, and `written` accumulation — none of them starvation-shaped.

src/test/java/com/otilm/core/integration/cbom/CryptoAssetPqcSweepITest.java
```diff
@@ -0,0 +1,293 @@
...
+     * <p>
+     * {@link #afresherVerdictIsNotOverwritten} writes the fresh verdict before the sweep runs, so the work list never
+     * offers the row and the guard is never reached -- removing the guard's clauses would leave that test green. This
+     * one calls the writer with a batch the sweep would have built from a stale read, after a fresher verdict has
+     * landed, which is one of the two orderings the guard exists for.
+     */
+    @Test
+    void theGuardedWriteRefusesABatchBuiltFromAStaleRead() {
+        UUID uuid = upsert("RSA", "rsa", "2048");
+        PqcVerdictWrite staleRead = new PqcVerdictWrite(uuid, staleRow(uuid).updated(), new PqcDecision(
```
karolhrdina: 🟡 **This test cannot see the clause the body says it was written for: delete the `pqc_ruleset_version` clause from the guarded `UPDATE` and it stays green.**

Body: *"deleting the `AND (pqc_ruleset_version …)` clause left it green … `theGuardedWriteRefusesABatchBuiltFromAStaleRead` drives the writer directly with a batch built from a stale read."* Measured: delete `CryptoAssetRepository.java:229` → `CryptoAssetPqcSweepITest` **8/8 green**, `PqcVerdictSweeperTest` **6/6 green**.

Cause: every writer of the row also sets `i_upd = CURRENT_TIMESTAMP` (`upsertIdentity`, `recomputeMergeFromSources`, `applyPqcVerdict`, `applyPqcVerdictIfStale`), so the fresh `INGEST-WON` verdict this test writes moves `i_upd`, and the `AND i_upd = :updatedAsRead` clause on line 230 refuses the write on its own. The test pins the guard as a whole; the version clause is pinned by nothing. Contrast: deleting line 230 instead reddens exactly one test — `aPayloadLandingBetweenReadAndWriteKeepsTheRowOnTheWorkList` (`expected: 0 but was: 1`, reproduced twice) — so the `i_upd` clause is pinned by the newest, body-unmentioned test, and the version clause by nothing.

Production behaviour is correct today; the clause is defence in depth behind `i_upd`. What is wrong is the verification claim, and the Javadoc on the repository already frames an `i_upd` refusal over "a touch that changed nothing the rules read" as a cost — if that clause is ever loosened, the version clause becomes the only protection for the ingest-wins ordering, and nothing would notice it going missing.

**Fix** — build the stale write with the row's `i_upd` as it is *after* `applyPqcVerdict` landed (re-read the entity's `updated` after the `INGEST-WON` write; keep `NOT_READY / SWEEP-STALE`). Then only the version clause can refuse it: with line 229 present `applyStaleBatch` returns 0; with it deleted the write lands and `getPqcRuleId()` flips to `SWEEP-STALE` — red. Keep the current shape as the both-clauses case if you like.

src/main/java/com/otilm/core/cbom/pqc/PqcVerdictSweeper.java
```diff
@@ -0,0 +1,156 @@
...
+                : ObjectMapperFactory.jsonColumn().valueToTree(row.mergedCryptoProperties());
+        return evaluator
+                .evaluate(evaluator.fromStoredRow(row.fields(), merged), PqcEvaluator.nistQuantumSecurityLevel(merged));
+    }
+
+    /**
+     * @param ran false when disabled or another node held the lock, which is a skip rather than an empty success
+     * @param failed rows the rule set threw on, each stamped {@code EVALUATION-FAILED} at the current generation
+     */
+    public record SweepOutcome(boolean ran, boolean aborted, int evaluated, int written, int failed, int batches) {
```
karolhrdina: 🟡 **The record cannot say whether a stamped failure landed, so the history message asserts something it cannot know — and `failed > 0` is unreachable from any stored data.**

This is copilot's line-78 comment, widened from the string to the record. `evaluated` and `failed` are counted before `applyStaleBatch`; `written` is a bare total. Measured with a mock writer:

| outcome | history message | database |
|---|---|---|
| poison row, guard refuses it: `evaluated=0 written=0 failed=1` | `FAILED "…1 could not be evaluated and were recorded as UNKNOWN"` | row carries no verdict, still on the work list |
| writer throws on the only batch: `evaluated=3 failed=1 written=0 batches=0` | `FAILED "…Swept 4 … in 0 batch(es) … 1 recorded as UNKNOWN"` | nothing written |
| guard refuses everything: `evaluated=5 written=0 failed=0 batches=1` | **`SUCCESS`** `"Swept 5 … 0 verdict(s) written"` | nothing changed; five rows come back next hour |
| `evaluated=3 failed=2 written=3` | `"2 … recorded as UNKNOWN"` | consistent with {3 good} **or** {2 failed + 1 good} — unknowable |

So the Javadoc on `failed` ("each stamped `EVALUATION-FAILED` at the current generation") and the task's "no later sweep will revisit it" are false in exactly the refused and aborted cases, and a run that wrote nothing reports `SUCCESS`.

Separately, and worth knowing before anyone spends more on this path: **`failed > 0` cannot be produced from stored data today.** Nine deliberately hostile rows — `name` NULL, non-numeric and 23-digit `parameter_set`, punctuation-only identity columns, `UNROUTABLE`/`PROTOCOL`/`CERTIFICATE` types, a payload with `type=12345`, `size="abc"`, `nistQuantumSecurityLevel="x"`, nested arrays, and a `relatedCryptoMaterialProperties` that is a string — all evaluate cleanly (`9 verdict(s) written, 0 could not be evaluated`). Every reader in `fromStoredRow` guards its input. A wider sweep of 280 vector components, 53 corpus assets and 129 600 synthetic rows through the exact stored-row route threw nothing. The `EVALUATION-FAILED` branch is reachable only by a mock, which is the right way to have it, but the message and the `FAILED`-on-`failed` branch are then describing a case that does not occur.

**Fix** — return per-row results from `applyStaleBatch` (or a count of stamped failures that actually landed) and carry `failedWritten` in the outcome; text-only minimum: drop "and were recorded as UNKNOWN", and report `swept − written` as *refused, retried next sweep*: `"Read %d stale cryptographic asset(s); %d verdict(s) written in %d committed batch(es); %d refused and left for the next sweep; %d could not be evaluated"`.

src/main/java/com/otilm/core/service/writer/cbom/CryptoAssetPqcVerdictWriter.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /**
+     * @return rows actually written; below {@code batch.size()} when ingest touched a row after the read, which is
+     * normal
+     */
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public int applyStaleBatch(List<PqcVerdictWrite> batch, int rulesetVersion) {
+        int written = 0;
+        for (PqcVerdictWrite write : batch) {
```
karolhrdina, line 44: 🟡 **One failing `UPDATE` rolls back its whole batch and, while the cause persists, starves every row behind it on every later sweep.**

Up to `batchSize` (500 in production) `UPDATE`s run in this one `REQUIRES_NEW` transaction with no per-row isolation. Reproduced against real PostgreSQL: 12 stale rows, test batch 5, the `UPDATE` of the row at uuid-order index 6 made to fail with SQLSTATE 23514 (a `BEFORE UPDATE` trigger raising a CHECK-violation code — the shape a fifth `PqcVerdict` constant shipped without its migration would produce):

```
sweep#1  SweepOutcome[aborted=true, evaluated=10, written=5, failed=0, batches=1]
         stamped: uuid-order 0..4 only; batch 2's four good rows rolled back; batch 3 never attempted
         performJob -> FAILED "The sweep aborted before completing. Swept 5 … in 0 batch(es) …"
sweep#2, #3 (cause still present)  aborted=true, written=0, batches=0   rows 10 and 11 still stale
cause removed -> sweep#4 written=7, stale=0
```

Because the cursor restarts at the nil uuid, the failing batch is reached first on every run, so every stale row after it in uuid order is never reached — the starvation the body reports fixed for *evaluation* failures, reintroduced on the *write* side (the `EVALUATION-FAILED` stamp only covers exceptions the evaluator throws). In production the cost is 499 good writes per run plus the whole tail of the table, under an hourly `FAILED` row.

Reachability at this head, stated honestly: no persistent per-row write failure exists today. `PqcVerdict` in the resolved `interfaces` jar has exactly the four constants `ck_crypto_asset_pqc_verdict` lists, and `evaluatedFields` values originate from TEXT/JSONB columns, so the `jsonb` cast cannot meet the one escape it rejects. The reachable failures (`deadlock detected`, provoked with a reversed-order two-row ingest) are transient and self-heal on the next run. The permanent variant needs a future drift, and when it comes the symptom is an hourly FAILED history row over a stalled inventory.

**Fix** — on `RuntimeException` from a batch, retry the same batch one row per `REQUIRES_NEW` transaction (a second writer method), counting a row that still fails as `failed` and moving on; the cursor has already advanced, so the loop simply continues. Minimal alternative: `continue` rather than abort, so only the poisoned batch is starved, while still reporting `aborted`.

scripts/cbom/build_identity_tables.py
```diff
@@ -173,13 +173,20 @@
     {"pattern": r"(?<![A-Za-z0-9])scrypt", "family": "scrypt", "why": "registry token"},
 
     # --- DRBG named after its primitives, same principle ---------------------------
-    {"pattern": r"(?<![A-Za-z0-9])Yarrow", "family": "Yarrow",
+    {"pattern": r"(?<![A-Za-z0-9])Yarrow(?![A-Za-z])", "family": "Yarrow",
```
karolhrdina, line 176: 🟡 **The two right guards move verdicts for glued spellings without a `PqcRuleset.VERSION` bump, and their material-tier mirror is a `notReady → ready` flip the recorded cost does not name.**

Regeneration is byte-reproducible, all three digest pins equal the file's real `68f01858e4cf…`, `IdentityVectorTest` is 541/541, and the in-repo corpus snapshot is unchanged — on every fixture the repo carries, **0** elections, keys, secondary-token sets or verdicts move, which reproduces your external-corpus zero in kind. `fromStoredRow` + `evaluate` is 0 lines different from `main` over 1 302 inputs, so the evaluator diff is Javadoc-only as claimed.

What does move, enumerated rather than trusted: nine synthetic spellings. `Yarrowed`, `YarrowAES`, `CMEAlgorithm`, `CMEAS` go `NOT_READY / CLASSICAL-LEGACY → UNKNOWN / FAMILY-UNRESOLVED` and change key tier (`alg:family → alg:name`); `YarrowDES`, `YarrowSHA1`, `CMEADES` keep `CLASSICAL-LEGACY-COMPONENT` but their evidence and keys move. Two consequences:

1. **The mirror.** The `why` text records the cost as "`YarrowAES` elects nothing". As sized symmetric-key material the same spelling now reads **`READY / MATERIAL-SYMMETRIC-READY`** where it read `NOT_READY / CLASSICAL-LEGACY` before — the material arms decide once no legacy family is named. Before the PR those rows carried a false legacy finding when the name was merely a word (`Yarrowed`); after it they carry a false `ready` when the name is a glued legacy token. Synthetic only — no in-repo fixture holds any Yarrow/CMEA/Skipjack spelling, the registry variant is hyphenated, and you report none externally. If the spelling-evidence decision stands, the acceptance should name this direction in the `why` and the commit; do not change rules here.
2. **The version constant.** `PqcRuleset`'s contract, as edited in this very PR, is "bump it whenever a rule changes any row's verdict, rule id or evidence", and this is the **first** table move on the history of that constant without a bump (1 → 2 → 3 → 4 each paired with a verdict-changing commit; verified from `git log -p`). Not reachable today — `CryptoAssetWriter` has no production caller, so no row was ever stamped 4 under the old tables — which is precisely the "one standing exemption, and it expires" that `IdentityRuleset`'s own Javadoc records with a tripwire. Nothing in-tree records the same ground for `PqcRuleset`. Bump to 5 (free), or one sentence in `PqcRuleset`'s Javadoc citing the `IdentityRuleset` exemption and naming the moved spellings.

src/main/java/com/otilm/core/tasks/CryptoAssetPqcSweepTask.java
```diff
@@ -0,0 +1,88 @@
...
+        int swept = outcome.evaluated() + outcome.failed();
+        if (!outcome.ran() || (swept == 0 && !outcome.aborted())) {
+            throw new ScheduledJobSkippedException();
+        }
+        String message = "Swept %d cryptographic asset(s) in %d batch(es); %d verdict(s) written, %d could not be evaluated and were recorded as UNKNOWN"
+                .formatted(swept, outcome.batches(), outcome.written(), outcome.failed());
```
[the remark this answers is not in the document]
vyskocilm: Fixed in `10598434`, and the record could not have supported the claim either: `applyStaleBatch` returned a bare count, so the sweep never knew which of its writes landed.

It now returns the uuids it wrote. `SweepOutcome` carries `unevaluated` as the stamps that actually committed, and the history message reports rows refused (ordinary — another writer got there first, retried next sweep) and rows whose own write transaction failed as separate figures:

```
Read %d stale cryptographic asset(s) in %d batch(es); %d verdict(s) written, of which %d recorded
as UNKNOWN because the rule set could not be evaluated; %d refused and left for the next sweep;
%d could not be written
```

`CryptoAssetPqcSweepTaskTest` pins each branch, and `PqcVerdictSweeperTest.aStampTheGuardRefusedIsNotReportedAsRecorded` goes red if the count reverts to the stamps built rather than the stamps landed.

src/main/resources/application.yml
```diff
@@ -430,3 +430,8 @@ signing-record:
     fallback-cron: ${SIGNING_RECORD_DELETE_AFTER_RETRIEVAL_CRON:0 0 3 * * *}
     batch-size: ${SIGNING_RECORD_DELETE_AFTER_RETRIEVAL_BATCH_SIZE:1000}
     max-batches-per-sweep: ${SIGNING_RECORD_DELETE_AFTER_RETRIEVAL_MAX_BATCHES_PER_SWEEP:10}
+
+crypto-asset:
+  pqc-sweep:
+    batch-size: ${CRYPTO_ASSET_PQC_SWEEP_BATCH_SIZE:500}
+    max-batches-per-sweep: ${CRYPTO_ASSET_PQC_SWEEP_MAX_BATCHES:20}
```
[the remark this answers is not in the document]
vyskocilm: Renamed to `CRYPTO_ASSET_PQC_SWEEP_MAX_BATCHES_PER_SWEEP` in `10598434`, matching the property.

Worth recording why it was more than a naming slip, per @karolhrdina's sharpening of it in the same round: relaxed binding *already* accepted the long spelling and it outranked the placeholder's short name when both were set, so two live spellings existed and the documented one lost silently.

src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
vyskocilm: Taken as written, with the cheaper of your two options.

The statements stay separate — one statement with nullable guard parameters reads worse than a duplicate the build refuses to let diverge — and `CryptoAssetPqcVerdictSqlTest` reads both `@Query` values by reflection and asserts the text between `SET` and `WHERE` is identical. It also asserts the extraction found the real list (`pqc_decided_at`, `i_upd = CURRENT_TIMESTAMP`), so two empty strings cannot match.

Mutation: one `pqc_evaluated_at` line changed to a `COALESCE` in the guarded statement only → red.

Separately, AC #8's other half is inaccurate in the same place and in the other direction — the sweep writes with a second statement by design, not with "the existing `applyPqcVerdict` SQL". Worth a one-line correction the next time #2151's body is touched.

src/main/java/com/otilm/core/cbom/pqc/PqcEvaluator.java
vyskocilm: Correct on both counts, and the body is rewritten.

The scope-finding section is gone: #2227's `8caec99a` shipped the derivation, so the option list this PR carried was describing a question that was already closed when the PR opened. The stated blocker for AC #10 goes with it.

AC #10 is therefore undischarged with no remaining blocker, and I have not smuggled that away. It needs a run over the corpus copy #2196's dossier ratifies (`documents=200 parsed=196 components=8048 keyed=4994`), with `unknown` scored against a denominator that excludes the correctly-`notApplicable` names. It stays open on #2151, which is the one place a reader will look for it.

The measurement caveat behind the original finding stays on the record in the body — "993 of 1 478 material components name a family in their own name" counted family tokens in component *names*, which is not what the pipeline stores.

src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
vyskocilm: Accepted, and fixed the way you prescribe.

`i_upd` was my choice, and the reasoning behind it was wrong in a way worth naming: I asked which columns every writer touches, which is the right question for coverage and the wrong one for uniqueness.

`findStaleVerdictRows` is now native and projects `xmin::text::bigint AS row_version` **in the same statement** as the columns the verdict is computed from — same statement rather than a cheap second one, because the sweep's transaction is READ COMMITTED and a later snapshot would make a write landing between the two invisible to the guard. `PqcStaleVerdictRow` and `PqcVerdictWrite` carry `long rowVersion` in place of `updated`, and the guard is `AND xmin::text::bigint = :rowVersion`. `merged_crypto_properties` comes back as text while I am there, since the rules want a `JsonNode` and the entity converter was building a `Map` only for the sweep to serialize it again.

Pinned by `aPayloadLandingWithoutMovingTheTimestampKeepsTheRowOnTheWorkList`. The collision is staged rather than raced — the payload is written with `i_upd` left at exactly the value the sweep read, which is the state your barrier-synchronised pairs produce, and is deterministic — and the test asserts both halves of the premise before asserting the refusal:

```
the payload landed and left i_upd exactly as the sweep read it   assertEquals(updatedAsRead)
the row version, unlike the timestamp, cannot be shared          assertNotEquals(read.rowVersion())
applyStaleBatch(...)                                             isEmpty()
```

Mutation: the clause neutralised with the parameter still bound → this test **and** `aPayloadLandingBetweenReadAndWriteKeepsTheRowOnTheWorkList` red, so the guard now has two tests that fail without it.

Your `clock_timestamp()` alternative rejected for the reason you gave: it changes `i_upd` semantics for every writer on the platform to fix one guard.

src/main/java/com/otilm/core/tasks/CryptoAssetPqcSweepTask.java
vyskocilm: Fixed as prescribed, message and all. `performJob` catches `RuntimeException` around `sweeper.sweep()`, on the listener's still-live connection, logs and returns FAILED with fixed text — no exception message on the wire, and `CbomSyncTask`'s `getMessage()` deliberately not copied. The skip path is unchanged, since `ScheduledJobSkippedException` is thrown after the call.

On the coverage measurement, which was the more useful half of this comment: `performJob` is now covered by `CryptoAssetPqcSweepTaskTest` — seven tests over the skip (nothing read, and a contended sweep), the three FAILED branches, SUCCESS, and this catch, including that the returned message carries no table name.

src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
vyskocilm: Taken the first way: the work list is widened.

```
WHERE (pqc_ruleset_version IS NULL
        OR pqc_ruleset_version < :version
        OR pqc_evaluated_at < i_upd)
```

The guard's staleness clause is widened identically, and has to be — a row offered because its payload moved is already at the current generation, so a version-only clause would refuse every write the wider list asks for. Widening one without the other would have introduced a defect of its own, which is the kind of thing that only shows up when both halves are written in the same sitting.

The mirror you asked me to weigh is accepted, not waved past: identity refreshes now re-offer rows on every re-sync, bounded by the per-sweep cap, and the missing work-list index becomes more pressing rather than less. That index stays recorded as open work rather than guessed into a migration — and the reason for waiting was that the predicate was unsettled, which it no longer is.

Pinned by `aPayloadLandingAfterTheStampPutsTheRowBackOnTheWorkList`: a material row swept to a verdict, then given a payload, is back on the list and re-evaluated to `MATERIAL-SYMMETRIC-WEAK` on the next sweep. Mutation: the clause removed from the work list → red.

src/main/java/com/otilm/core/cbom/pqc/PqcVerdictSweeper.java
vyskocilm: Correct, and the third shape you name is exactly why the strengthened test could not see it.

Fixed with the test you specify. `PqcVerdictSweeperTest.theCursorAdvancesEvenWhenTheGuardRefusesEveryWrite` serves two full pages keyed off the `after` argument, has the writer return no written uuids, captures `after` and asserts `[nil, last(page1), last(page2)]` over three reads. Mutation: the cursor line deleted → that test red, alone, out of nine.

The other three mutations you list stay unpinned. On your own reading none is starvation-shaped, and I would rather say so here than add tests whose only job is to fail a mutation — recorded so it is not mistaken for coverage.

src/test/java/com/otilm/core/integration/cbom/CryptoAssetPqcSweepITest.java
vyskocilm: Rebuilt as you prescribe, and renamed to say what it now pins: `theGuardedWriteRefusesARowThatIsNoLongerStale`.

It takes the row version *after* `applyPqcVerdict` landed, so the version half of the guard matches and only the staleness clause can refuse. Mutation: the staleness clause deleted → that test red, and it alone.

The other half is pinned by the two payload-window tests (see the `xmin` thread), so both clauses of the guard now have a test that fails without them — which was the state your comment says the PR claimed and did not have.

src/main/java/com/otilm/core/cbom/pqc/PqcVerdictSweeper.java
vyskocilm: Fixed at the record, not at the string.

`applyStaleBatch` returns the uuids it wrote. `SweepOutcome` is now `(ran, aborted, read, written, unevaluated, writeFailures, batches)` with `refused()` derived, and `unevaluated` counts only stamps that committed. The four rows of your table now read:

| outcome | history |
|---|---|
| poison row, guard refuses it | FAILED-free: `read 1, written 0, refused 1`, nothing claimed as recorded |
| writer throws on the only batch | the batch is retried per row (see the writer thread); a row that still fails is `writeFailures`, FAILED |
| guard refuses everything | SUCCESS, `written 0, refused 5` — stated in figures rather than implied |
| `read 3, unevaluated 2, written 3` | unambiguous: the 2 are stamps that landed |

`failed > 0` being unreachable from stored data — worth knowing, and it is now reachable in one more way than when you measured it: the merged payload arrives as text and is parsed in the sweep, so a payload not readable as JSON throws where your nine hostile rows evaluated cleanly. Still nothing the pipeline can store, so the branch stays mock-reached by design. What changed is that the message no longer describes a case it cannot distinguish.

src/main/java/com/otilm/core/service/writer/cbom/CryptoAssetPqcVerdictWriter.java
vyskocilm: Fixed with your primary option rather than the minimal one.

On a `RuntimeException` from the batch, the same rows are retried one per `REQUIRES_NEW` transaction (`CryptoAssetPqcVerdictWriter.applyStaleRow`); a row that still fails is counted as a write failure, logged with its uuid alone, and the loop continues — the cursor has already advanced. A write-side failure no longer reports `aborted` at all; `aborted` is now the *read* failing, which is the case the other thread on this file is about.

Pinned by `aBatchThatRollsBackIsRetriedOneRowPerTransaction`: the batch throws, one of three rows keeps throwing per-row, and the outcome is `written 2 / writeFailures 1 / refused 0 / aborted false / batches 1`, with `applyStaleRow` verified three times. Mutation: the retry removed → red.

Your reachability note stands and is the reason this is a small change rather than a redesign: no persistent per-row write failure exists at this head, and the deadlock variant self-heals on the next run. The cost is one extra transaction per row, and only on a batch that has already rolled back.

scripts/cbom/build_identity_tables.py
vyskocilm: Both taken.

`PqcRuleset.VERSION` is 5 — the free option, and the one the constant's own contract asks for rather than an exemption borrowed from `IdentityRuleset`'s Javadoc. Your point that this would have been the first table move without a bump is the whole argument; there is no reason to spend a documented exemption to save an integer.

Both `why` texts now name the direction of the error rather than only its size. As sized symmetric key material `YarrowAES` reads `READY / MATERIAL-SYMMETRIC-READY` where it read `NOT_READY / CLASSICAL-LEGACY`, because the material arms decide once no legacy family is named — so a false `ready` on a glued legacy token replaces a false legacy finding on a name that was merely a word (`Yarrowed`). `CMEAS` carries the same sentence. Rules unchanged, as you asked.

Regenerated: the artifact digest is now `474e3dc95c5e155f9dea80ee0f1620081f8170e20a9dc9f29e1e2ac09337e1d1`, and all three pins move in the same commit — `PqcFamiliesTest`, `IdentityTablesTest`, and `IdentityTables`'s Javadoc. `IdentityVectorTest` 541/541 and `NormalizationRulesTest` 145/145 on the new tables.

karolhrdina (approved):

PR #2233, opened 2026-09-07
Mirror attribute visibility as a column so every read path shares one predicate

### LINKED ISSUE #2220: Mirror attribute visibility as a column so every read path shares one predicate
### Description

`AttributeDefinition` stores the whole serialized attribute in its `definition` jsonb column and mirrors only some of its properties as real columns: `required`, `readOnly` and `enabled`. `visible` is not among them.

Three read paths have to answer "may this attribute's values be shown to a user", and each answers it differently:

- `AttributeColumnProjector` reads it in Java, after loading the definition.
- `SearchHelper.isDisplayable` reads it off `SearchFieldObject` when publishing the catalogue.
- `FilterPredicatesBuilder.definitionIsVisible`, added in core#2195, has to answer it inside a query, so it extracts `properties -> visible` with `jsonb_extract_path_text` and treats an absent value as visible.

That last one fails open. If the interfaces attribute model renames the property or nests it one level deeper, the extraction returns NULL for every definition, every predicate passes, and hidden content becomes filterable again — while the projector, reading the deserialized object, still hides it. Nothing fails to compile and no test catches the divergence.

Mirroring `visible` as a column, the way the other three properties already are, gives projection, sorting and filtering one predicate over one source of truth, and turns a rename into a mapping error rather than a silent authorization gap.

### Definition of Done

- `attribute_definition` carries a `visible` column, written on definition create and update alongside `required`, `readOnly` and `enabled`, with a Flyway migration that backfills it from the stored `definition` document.
- `FilterPredicatesBuilder` reads the column instead of extracting it from jsonb, and `definitionIsVisible` no longer treats an unreadable value as visible.
- The catalogue and the projection read the same column.
- A test asserts a hidden custom attribute is withheld from a column, an ordering and a filter, and that the three answers agree.

Jakub-Bartal-01:
`AttributeDefinition` mirrors `required`, `readOnly` and `enabled` as real columns but kept `visible` only inside the serialized definition document. Three read paths had to answer "may this attribute's values be shown to a user", and each reached it differently:

- `AttributeColumnProjector` read it off the deserialized definition, per row, after loading.
- `SearchHelper` derived it when building the catalogue.
- `FilterPredicatesBuilder.definitionIsVisible` extracted `properties -> visible` with `jsonb_extract_path_text` inside the query, and treated an absent value as visible.

The last one fails open. If the interfaces attribute model renames the property or nests it one level deeper, the extraction returns NULL for every definition, every predicate passes, and hidden content becomes filterable again — while the projector, reading the deserialized object, still hides it. Nothing fails to compile and no test catches the divergence.

## The change

`attribute_definition` carries `visible` alongside the other three, backfilled from the stored document by `V202609071200`. All three paths read that column.

`AttributeDefinition.setDefinition` re-derives the column from the definition it stores, which is what keeps the two in step: the entity is the only writer, and the value comes from `properties.isVisible()` — a typed accessor. A rename in the interfaces model is now a compile error rather than a silent authorization gap, which is the point of the issue.

The projection turns its per-row Java check into a query predicate, so `ProjectedAttributeContent` no longer carries the definition document it held solely for this flag, and the listing query stops loading a jsonb blob per content row.

## Scope held deliberately

The filter path applies visibility to `CUSTOM` definitions only, while the projection and the catalogue apply it to every type. That asymmetry predates this change and is preserved: on a data or metadata definition `visible` is a connector's display hint, so it governs what is rendered but not what is filterable, and enforcing it in filters would stop matching metadata callers filter on today. The comment above the early return now says so, so the new projection predicate is not read as evidence the split was meant to close.

## Verification

- `mvn -P test-integration-core` — 1133 tests, 0 failures.
- `HiddenAttributeAgreementITest` — one hidden custom attribute through the catalogue, a column, an ordering and a filter, with a visible control beside it.
- The migration was applied to a Postgres 16 instance and backfills correctly for every shape the document takes: explicit `false`, explicit `true`, `properties` without the key, and no `properties` at all. Core's tests build the schema from entities and never run Flyway, so this is the only coverage the migration has.

Closes #2220

## Summary by CodeRabbit

* **Bug Fixes**
  * Hidden attributes are now consistently excluded from search results, projections, filtering, and sorting.
  * Visible attributes continue to support display, filtering, sorting, and ordering.
  * Attribute visibility is persisted and applied consistently across search and discovery operations.
* **Tests**
  * Added coverage verifying behavior for hidden and visible custom attributes across search paths.

### DIFF
--- a/src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
+++ b/src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
@@ -215,11 +215,8 @@ public class AttributeColumnProjector {
 
         for (ProjectedAttributeContent content : stored) {
             // Encrypted content is ciphertext that only its own decryption path can read, and a listing does not take
-            // that path. An attribute whose definition says it is not visible is not to be shown to a user, which is
-            // what a column does with it. The catalogue withholds both kinds of field; a row that reaches here anyway
-            // is dropped, because the flag it withheld them with is a hint on a response the caller is free to ignore.
-            if (content.encryptedContent() != null || WITHHELD_CONTENT_TYPES.contains(content.contentType())
-                    || !AttributeDefinitionProperties.isVisible(content.definition())) {
+            // that path. The query already excludes a definition marked not visible.
+            if (content.encryptedContent() != null || WITHHELD_CONTENT_TYPES.contains(content.contentType())) {
                 continue;
             }
             RequestedColumn column = matchRequested(requested, content);
--- a/src/main/java/com/otilm/core/dao/entity/AttributeDefinition.java
+++ b/src/main/java/com/otilm/core/dao/entity/AttributeDefinition.java
@@ -35,6 +35,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.UUID;
+import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import lombok.Setter;
@@ -93,6 +94,7 @@ public class AttributeDefinition extends UniquelyIdentified implements ObjectAcc
      * visibility from one column rather than three readings of the document. Written only by {@link #setDefinition},
      * which is what keeps the two in step.
      */
+    @Setter(AccessLevel.NONE)
     @Column(name = "visible", nullable = false)
     private boolean visible = true;
 
--- a/src/main/java/com/otilm/core/dao/entity/AttributeDefinition.java
+++ b/src/main/java/com/otilm/core/dao/entity/AttributeDefinition.java
@@ -17,6 +17,7 @@ import com.otilm.api.model.common.attribute.common.properties.CustomAttributePro
 import com.otilm.api.model.common.attribute.common.properties.MetadataAttributeProperties;
 import com.otilm.api.model.common.attribute.v2.MetadataAttributeV2;
 import com.otilm.api.model.common.attribute.v3.CustomAttributeV3;
+import com.otilm.core.attribute.engine.AttributeDefinitionProperties;
 import com.otilm.core.attribute.engine.AttributeVersionHelper;
 import com.otilm.core.util.ObjectAccessControlMapper;
 import jakarta.persistence.Column;
@@ -87,6 +88,14 @@ public class AttributeDefinition extends UniquelyIdentified implements ObjectAcc
     @Column(name = "read_only")
     private Boolean readOnly;
 
+    /**
+     * Mirrors {@code properties.visible} of the stored definition, so projection, ordering and filtering answer
+     * visibility from one column rather than three readings of the document. Written only by {@link #setDefinition},
+     * which is what keeps the two in step.
+     */
+    @Column(name = "visible", nullable = false)
+    private boolean visible = true;
+
     @Column(name = "version", nullable = false)
     private int version;
 
@@ -132,6 +141,15 @@ public class AttributeDefinition extends UniquelyIdentified implements ObjectAcc
     @Column(name = "encrypted_data", length = Integer.MAX_VALUE)
     @JdbcTypeCode(SqlTypes.ARRAY)
     private List<String> encryptedData;
+
+    /**
+     * Replaces the definition and re-derives {@link #visible} from it. Overridden rather than left to Lombok because a
+     * definition written without refreshing the column would let a filter read a visibility the projection does not.
+     */
+    public void setDefinition(BaseAttribute definition) {
+        this.definition = definition;
+        this.visible = AttributeDefinitionProperties.isVisible(definition);
+    }
 
     public void setConnector(Connector connector) {
         this.connector = connector;
--- a/src/main/java/com/otilm/core/model/SearchFieldObject.java
+++ b/src/main/java/com/otilm/core/model/SearchFieldObject.java
@@ -8,7 +8,6 @@ import com.otilm.api.model.common.attribute.common.DataAttribute;
 import com.otilm.api.model.common.attribute.common.MetadataAttribute;
 import com.otilm.api.model.common.attribute.common.content.AttributeContentType;
 import com.otilm.api.model.common.attribute.common.content.data.ProtectionLevel;
-import com.otilm.core.attribute.engine.AttributeDefinitionProperties;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
@@ -33,10 +32,7 @@ public class SearchFieldObject {
 
     private ProtectionLevel protectionLevel;
 
-    /**
-     * Whether the definition says its values may be shown to a user. Read here rather than at each use, because the
-     * flag lives on the properties of whichever attribute shape the definition holds.
-     */
+    /** Whether the definition says its values may be shown to a user, as mirrored on the definition row. */
     private boolean visible = true;
 
     private List<String> contentItems;
@@ -53,12 +49,12 @@ public class SearchFieldObject {
     }
 
     public SearchFieldObject(String attributeName, AttributeContentType attributeContentType,
-            AttributeType attributeType, String label, BaseAttribute attributeDefinition) {
+            AttributeType attributeType, String label, boolean visible, BaseAttribute attributeDefinition) {
         this.attributeName = attributeName;
         this.attributeContentType = attributeContentType;
         this.attributeType = attributeType;
         this.label = label;
-        this.visible = AttributeDefinitionProperties.isVisible(attributeDefinition);
+        this.visible = visible;
 
         if (attributeType == AttributeType.CUSTOM || attributeType == AttributeType.DATA) {
             if (attributeDefinition instanceof CustomAttribute customAttribute) {
--- a/src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
+++ b/src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
@@ -89,11 +89,6 @@ public class FilterPredicatesBuilder {
             .of(AttributeContentType.INTEGER, AttributeContentType.FLOAT, AttributeContentType.DATE,
                     AttributeContentType.TIME, AttributeContentType.DATETIME);
     private static final String JSONB_EXTRACT_PATH_TEXT_FUNCTION_NAME = "jsonb_extract_path_text";
-
-    /** The path within a stored attribute definition at which its visibility lives. */
-    private static final String DEFINITION_PROPERTIES_KEY = "properties";
-
-    private static final String VISIBLE_KEY = "visible";
     private static final String TEXTREGEXEQ_FUNCTION_NAME = "textregexeq";
     private static final String ARRAY_CONTAINS_FUNCTION_NAME = PostgresFunctionContributor.ARRAY_CONTAINS;
 
@@ -1309,17 +1304,11 @@ public class FilterPredicatesBuilder {
     }
 
     /**
-     * Whether the definition behind a content row says its attribute may be shown to a user. Read out of the stored
-     * definition document with {@code jsonb_extract_path_text}, since visibility is a property of the serialized
-     * attribute rather than a column; absent counts as visible, as it does in {@code AttributeDefinitionProperties}.
+     * Whether the definition behind a content row says its attribute may be shown to a user. Reads the mirrored column,
+     * so this cannot fall open on a document whose shape has moved.
      */
     private static Predicate definitionIsVisible(final CriteriaBuilder criteriaBuilder, final Join joinDefinition) {
-        final Expression<String> visible = criteriaBuilder
-                .function(JSONB_EXTRACT_PATH_TEXT_FUNCTION_NAME, String.class,
-                        joinDefinition.get(AttributeDefinition_.definition),
-                        criteriaBuilder.literal(DEFINITION_PROPERTIES_KEY), criteriaBuilder.literal(VISIBLE_KEY));
-        return criteriaBuilder
-                .or(criteriaBuilder.isNull(visible), criteriaBuilder.notEqual(visible, Boolean.FALSE.toString()));
+        return criteriaBuilder.isTrue(joinDefinition.get(AttributeDefinition_.visible));
     }
 
     /**
@@ -1340,6 +1329,10 @@ public class FilterPredicatesBuilder {
         // Only custom definitions carry a permission model and a visibility the platform enforces. On a data or
         // metadata definition `visible` is a connector's display hint, and the nullable `enabled` column is unset,
         // so applying either would drop rows a listing is meant to return.
+        //
+        // The projection and the catalogue do withhold a hidden data or metadata field, because those decide what is
+        // rendered and a display hint governs that. Filtering is not rendering, so such a value stays answerable here
+        // by design: the split is deliberate, not an oversight the projection predicate was meant to close.
         if (attributeType != AttributeType.CUSTOM) {
             return predicates;
         }
--- a/src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
+++ b/src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
@@ -1218,9 +1218,10 @@ public class FilterPredicatesBuilder {
      *
      * <p>
      * Ordering reads a value, so it is gated like the projection that renders one: encrypted content is skipped, a
-     * disabled custom definition is skipped, and the caller's custom-attribute permissions narrow which definitions are
-     * readable at all. Whether the field may be ordered on - visible, not secret, not a code block - is settled before
-     * this by {@code ListingSortResolver} against the resource's published catalogue.
+     * definition marked not visible is skipped whatever its attribute type, a disabled custom definition is skipped,
+     * and the caller's custom-attribute permissions narrow which definitions are readable at all. That the field may be
+     * ordered on at all - not secret, not a code block, and visible in at least one of the definitions it collapses -
+     * is settled before this by {@code ListingSortResolver} against the resource's published catalogue.
      */
     public static <T> Expression<?> getAttributeSortKey(final CriteriaBuilder criteriaBuilder,
             final CommonAbstractCriteria query, final Root<T> root, final SortSpecification sort) {
@@ -1281,6 +1282,11 @@ public class FilterPredicatesBuilder {
         predicates
                 .addAll(attributeReadabilityPredicates(criteriaBuilder, joinContentItem, joinDefinition, attributeType,
                         contentFilterSource, true));
+        // A hidden definition must supply no sort key either: the order of a page is part of what it shows, and the
+        // projection filling its cells keeps only visible definitions. Custom is narrowed for every reader already.
+        if (attributeType != AttributeType.CUSTOM) {
+            predicates.add(definitionIsVisible(criteriaBuilder, joinDefinition));
+        }
 
         subquery.select(value).where(predicates.toArray(new Predicate[]{}));
         ((JpaSubQuery) subquery)
@@ -1330,9 +1336,8 @@ public class FilterPredicatesBuilder {
         // metadata definition `visible` is a connector's display hint, and the nullable `enabled` column is unset,
         // so applying either would drop rows a listing is meant to return.
         //
-        // The projection and the catalogue do withhold a hidden data or metadata field, because those decide what is
-        // rendered and a display hint governs that. Filtering is not rendering, so such a value stays answerable here
-        // by design: the split is deliberate, not an oversight the projection predicate was meant to close.
+        // A display hint governs what is rendered, so the projection and the ordering that arranges it withhold a
+        // hidden data or metadata value while a filter still matches it.
         if (attributeType != AttributeType.CUSTOM) {
             return predicates;
         }
--- a/src/main/java/com/otilm/core/util/SearchHelper.java
+++ b/src/main/java/com/otilm/core/util/SearchHelper.java
@@ -323,6 +323,7 @@ public class SearchHelper {
             .thenComparing(SearchFieldObject::getLabel, Comparator.nullsLast(Comparator.naturalOrder()))
             .thenComparing(SearchFieldObject::isList)
             .thenComparing(SearchFieldObject::isMultiSelect)
+            .thenComparing(SearchFieldObject::isVisible)
             .thenComparing(SearchFieldObject::getProtectionLevel, Comparator.nullsLast(Comparator.naturalOrder()))
             .thenComparing(field -> field.getContentItems() == null ? null : String.join("\0", field.getContentItems()),
                     Comparator.nullsLast(Comparator.naturalOrder()));
@@ -352,6 +353,11 @@ public class SearchHelper {
         // matchable.
         if (other.getProtectionLevel() != ProtectionLevel.ENCRYPTED) {
             merged.setProtectionLevel(other.getProtectionLevel());
+        }
+        // Projection and the filter predicates keep the content of every visible definition, so the collapsed field is
+        // shown whenever any definition behind it is visible.
+        if (other.isVisible()) {
+            merged.setVisible(true);
         }
         // A fixed-choice list input is only correct if every definition is a list; otherwise free-form input must
         // survive the merge, since a list rendering would make the free-form definitions' values un-enterable.
--- a/src/main/java/com/otilm/core/util/SearchHelper.java
+++ b/src/main/java/com/otilm/core/util/SearchHelper.java
@@ -1,5 +1,6 @@
 package com.otilm.core.util;
 
+import com.otilm.api.model.common.attribute.common.AttributeType;
 import com.otilm.api.model.common.attribute.common.content.AttributeContentType;
 import com.otilm.api.model.common.attribute.common.content.data.ProtectionLevel;
 import com.otilm.api.model.common.enums.BitMaskEnum;
@@ -161,7 +162,8 @@ public class SearchHelper {
         // filter the listing then refuses.
         if (attributeSearchInfo.getProtectionLevel() == ProtectionLevel.ENCRYPTED
                 || AttributeColumnProjector.WITHHELD_CONTENT_TYPES
-                        .contains(attributeSearchInfo.getAttributeContentType())) {
+                        .contains(attributeSearchInfo.getAttributeContentType())
+                || hasNoMatchableContent(attributeSearchInfo)) {
             conditionOperators = List.of(FilterConditionOperator.EMPTY, FilterConditionOperator.NOT_EMPTY);
         }
         fieldDataDto.setConditions(conditionOperators);
@@ -277,6 +279,23 @@ public class SearchHelper {
         return !AttributeColumnProjector.WITHHELD_CONTENT_TYPES.contains(attributeSearchInfo.getAttributeContentType())
                 && attributeSearchInfo.getProtectionLevel() != ProtectionLevel.ENCRYPTED
                 && attributeSearchInfo.isVisible();
+    }
+
+    /**
+     * Whether no value of this field is matchable, so the conditions published for it are the presence pair alone.
+     *
+     * <p>
+     * A custom definition marked not visible is excluded from every query that reads its content, filters included, so
+     * a value condition on it could only ever return nothing - and a published condition is one a saved list view
+     * accepts, not merely a hint. A presence condition survives because it needs no value to answer: with the content
+     * unreadable the field is empty everywhere, which is what the column beside it renders.
+     *
+     * <p>
+     * A data or metadata definition carries the same flag as a connector's display hint that filtering does not apply,
+     * so its values stay matchable and its conditions stay whole.
+     */
+    private static boolean hasNoMatchableContent(final SearchFieldObject attributeSearchInfo) {
+        return attributeSearchInfo.getAttributeType() == AttributeType.CUSTOM && !attributeSearchInfo.isVisible();
     }
 
     private static SearchFieldTypeEnum retrieveSearchFieldTypeEnumByContentType(
--- a/src/main/resources/db/migration/V202609071200__attribute_definition_visible.sql
+++ b/src/main/resources/db/migration/V202609071200__attribute_definition_visible.sql
@@ -1,0 +1,7 @@
+-- Mirrors properties.visible out of the stored definition document so projection, ordering and
+-- filtering share one predicate. Absent in the document means visible, which is the column default.
+ALTER TABLE attribute_definition ADD COLUMN visible BOOLEAN NOT NULL DEFAULT TRUE;
+
+UPDATE attribute_definition
+SET visible = FALSE
+WHERE definition -> 'properties' ->> 'visible' = 'false';
--- a/src/test/java/com/otilm/core/attribute/engine/ProjectedContentWithholdingTest.java
+++ b/src/test/java/com/otilm/core/attribute/engine/ProjectedContentWithholdingTest.java
@@ -1,0 +1,109 @@
+package com.otilm.core.attribute.engine;
+
+import com.otilm.api.model.client.certificate.SearchColumnRequestDto;
+import com.otilm.api.model.common.attribute.common.AttributeType;
+import com.otilm.api.model.common.attribute.common.content.AttributeContentType;
+import com.otilm.api.model.common.attribute.v3.content.BaseAttributeContentV3;
+import com.otilm.api.model.common.attribute.v3.content.StringAttributeContentV3;
+import com.otilm.api.model.core.auth.Resource;
+import com.otilm.api.model.core.search.AttributeProjectable;
+import com.otilm.api.model.core.search.FilterFieldSource;
+import com.otilm.core.attribute.engine.AttributeEngine.CustomAttributeContentFilter;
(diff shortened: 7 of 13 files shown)


src/main/java/com/otilm/core/dao/entity/AttributeDefinition.java
```diff
@@ -87,6 +88,14 @@ public class AttributeDefinition extends UniquelyIdentified implements ObjectAcc
     @Column(name = "read_only")
     private Boolean readOnly;
 
+    /**
+     * Mirrors {@code properties.visible} of the stored definition, so projection, ordering and filtering answer
+     * visibility from one column rather than three readings of the document. Written only by {@link #setDefinition},
+     * which is what keeps the two in step.
+     */
+    @Column(name = "visible", nullable = false)
+    private boolean visible = true;
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed. The field now carries `@Setter(AccessLevel.NONE)`, and `setDefinition` is written explicitly so it re-derives the mirror, making it the only application write path as the javadoc claims.

I checked all 64 `setVisible(` call sites in `src/main` and `src/test` first: none is on this entity — every one is on an attribute *properties* object, a DTO or a request object — so nothing had to be migrated. There is also no `BeanUtils.copyProperties` in `src/main`, so no reflective writer depended on it. Hibernate uses field access here, and the integration suites that register and read back definitions of both visibilities still pass.

src/main/java/com/otilm/core/model/SearchFieldObject.java
```diff
@@ -53,12 +49,12 @@ public SearchFieldObject(String attributeName, AttributeContentType attributeCon
...
 
     public SearchFieldObject(String attributeName, AttributeContentType attributeContentType,
-            AttributeType attributeType, String label, BaseAttribute attributeDefinition) {
+            AttributeType attributeType, String label, boolean visible, BaseAttribute attributeDefinition) {
         this.attributeName = attributeName;
         this.attributeContentType = attributeContentType;
         this.attributeType = attributeType;
         this.label = label;
-        this.visible = AttributeDefinitionProperties.isVisible(attributeDefinition);
+        this.visible = visible;
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed, and the analysis was right on both counts. `mergeDuplicateField` now sets `visible` to true when any definition behind the collapsed field is visible, which is the union-of-capabilities semantics the method's own javadoc states and agrees with what projection and the ordering/filter predicates already do. `DUPLICATE_MERGE_ORDER` also gained `.thenComparing(SearchFieldObject::isVisible)`, so the comparator remains a total order over every merge-relevant field now that visibility is one of them.

Covered by two tests in `SearchHelperITest`: the collapsed field is visible when one of two same-name, same-content-type definitions is — asserted in both input orders, since the merge mutates the surviving object in place — and stays hidden when every definition is. The second is a control, so the OR cannot silently degrade to a constant.

src/main/java/com/otilm/core/dao/entity/AttributeDefinition.java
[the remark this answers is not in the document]
Jakub-Bartal-01: Fixed — the generated setter is suppressed with `@Setter(AccessLevel.NONE)` and `setDefinition` is written explicitly to re-derive the mirror. No existing caller set it on this entity (all 64 `setVisible(` sites are on properties objects, DTOs or request objects), and there is no `BeanUtils.copyProperties` in `src/main`, so nothing needed migrating.

src/main/java/com/otilm/core/dao/entity/AttributeDefinition.java
```diff
@@ -87,6 +89,15 @@ public class AttributeDefinition extends UniquelyIdentified implements ObjectAcc
...
     private Boolean readOnly;
 
+    /**
+     * Mirrors {@code properties.visible} of the stored definition, so projection, ordering and filtering answer
+     * visibility from one column rather than three readings of the document. Written only by {@link #setDefinition},
+     * which is what keeps the two in step.
+     */
+    @Setter(AccessLevel.NONE)
+    @Column(name = "visible", nullable = false)
+    private boolean visible = true;
```
ivosh: **. Nitpick** | Correctness

AttributeDefinition.java:92-99 documents `visible` as "Written only by `setDefinition`", and this PR rests its safety argument on that ("the entity is the only writer"). It holds for Java: all five `setDefinition` call sites are in AttributeEngine (lines 645, 985, 1052, 1057, 1156) and nothing assigns the field directly. It does not hold for SQL. Two shipped migrations rewrite the same document in place — `src/main/resources/db/migration/V202511241530__attributes_v3.sql:4-9` and `:12-25` (`jsonb_set` over `{schemaVersion}`, `{version}`, `{content}`) and `V202602111633__definition_content.sql:1-2` — and `DatabaseAttributeMigration.moveTableDataAttributes` inserts `attribute_definition` rows with an explicit 13-column list that names `definition` but not `visible` (src/main/java/com/otilm/core/util/DatabaseAttributeMigration.java:68). None of them diverges today, only because every one is ordered before V202609071200 and its backfill sweeps the result. From now on the ordering runs the other way: a migration that touches `properties` — an interfaces rename, which is the scenario this PR is defending against, is exactly such a migration — would update the document and leave `visible` at whatever the row already held, and that stale column is now the sole authorization predicate on the filter path (FilterPredicatesBuilder.java:1311). That is the same silent-divergence failure the PR removes from the query, relocated to the write path, with no constraint or test that would catch it.

_Suggested:_ Derive the column in the database (a generated column or a trigger on `definition`) so no writer can skip it, or state on the field and in the migration that any later migration rewriting `definition` must re-run the backfill.

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -353,6 +354,11 @@ private static SearchFieldObject mergeDuplicateField(final SearchFieldObject mer
         if (other.getProtectionLevel() != ProtectionLevel.ENCRYPTED) {
             merged.setProtectionLevel(other.getProtectionLevel());
         }
+        // Projection and the filter predicates keep the content of every visible definition, so the collapsed field is
+        // shown whenever any definition behind it is visible.
+        if (other.isVisible()) {
```
ivosh, line 378: **? Question** | Security — Nitpick if confirmed

`mergeDuplicateField` (SearchHelper.java:359) now reports a collapsed field visible whenever any definition behind it is, and `isDisplayable`/`isSortable` (SearchHelper.java:276, 222) derive both `displayable` and `sortable` from that one flag. The justification at SearchHelper.java:357 cites the projection and the filter predicates, and both do hold: `getProjectedAttributesContent` (AttributeContent2ObjectRepository.java:120) applies `ad.visible = true` to every attribute type, and `definitionIsVisible` (FilterPredicatesBuilder.java:1311) covers CUSTOM.

Ordering is the third reader of that flag and it is gated by nothing else. `ListingSortResolver.requireSortableAttributeField` (ListingSortResolver.java:76) refuses only what the catalogue does not publish as sortable, and `getAttributeSortKey` reaches `attributeReadabilityPredicates` (FilterPredicatesBuilder.java:1282), which returns at FilterPredicatesBuilder.java:1336 for DATA and META without ever adding `definitionIsVisible`. So where two META (or DATA) definitions share a name and content type and only one is visible — the multi-connector case `mergeFieldsWithSameIdentifier` exists for, and one the catalogue queries do produce as two rows since `ad.definition` sits in the DISTINCT tuple — the catalogue now deterministically publishes `sortable: true`, while the sort-key subquery correlates on type, content type and name only (FilterPredicatesBuilder.java:1164-1166), orders by `attributeDefinitionUuid` and takes the first row. That row can be the hidden definition's, so the listing is ordered by a value the column beside it does not render, against `getAttributeSortKey`'s own contract that "a column sorts by what the cell displays" (FilterPredicatesBuilder.java:1218). Before this PR the merged flag was decided by a comparator tie and so was sometimes false; the union makes the exposure consistent.

Two readings. If ordering follows the same rule the added comment states for the projection — "those decide what is rendered and a display hint governs that" (FilterPredicatesBuilder.java:1333) — then the sort path needs the visibility predicate for every type, or the union needs restricting to CUSTOM. If the union is meant to make the merged field orderable regardless, say so at SearchHelper.java:357, because the same added comment at FilterPredicatesBuilder.java:1333 asserts the catalogue withholds a hidden data or metadata field, which after this change is only true when every duplicate behind it is hidden.

_Suggested:_ Decide whether ordering follows the projection's visibility rule: either add `definitionIsVisible` to the sort-key predicates for every attribute type (or restrict the union at SearchHelper.java:359 to CUSTOM), or record the intent at SearchHelper.java:357 and correct the catalogue claim at FilterPredicatesBuilder.java:1333.

<sub>flagged by Correctness & races + OpenAPI & schema fidelity + Security & info-leak</sub>

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
```diff
@@ -215,11 +215,8 @@ private Map<UUID, Map<FilterFieldSource, Map<String, List<BaseAttributeContentV3
 
         for (ProjectedAttributeContent content : stored) {
             // Encrypted content is ciphertext that only its own decryption path can read, and a listing does not take
-            // that path. An attribute whose definition says it is not visible is not to be shown to a user, which is
-            // what a column does with it. The catalogue withholds both kinds of field; a row that reaches here anyway
-            // is dropped, because the flag it withheld them with is a hint on a response the caller is free to ignore.
-            if (content.encryptedContent() != null || WITHHELD_CONTENT_TYPES.contains(content.contentType())
-                    || !AttributeDefinitionProperties.isVisible(content.definition())) {
+            // that path. The query already excludes a definition marked not visible.
```
ivosh, line 218: **~ Minor** | Documentation

`AttributeColumnProjectionITest.java:275-277` explains its own test with "The catalogue withholds such a field, but that flag is a hint on a response the caller is free to ignore, so **the projection drops the values again on the way out**." That was an accurate description of the `!AttributeDefinitionProperties.isVisible(content.definition())` branch this PR deletes from `AttributeColumnProjector.collectValues` (`AttributeColumnProjector.java:217-219`). After this PR the rows never arrive: the exclusion is `ad.visible = true` in the query (`AttributeContent2ObjectRepository.java:120`), whose new Javadoc asserts the opposite of the test comment — "never loaded only to be dropped afterwards" (`AttributeContent2ObjectRepository.java:111-112`). The test still passes, so the rot is silent, and it is the one comment in the tree that still describes the removed Java-side drop (`grep -rn "on the way out" src/test/java` returns this line alone).

_Suggested:_ Rewrite the comment above `AttributeColumnProjectionITest.aHiddenAttributeColumnProjectsNoValues` to say the projection query excludes a definition marked not visible, instead of that the projector drops the values on the way out. Or drop the comment altogether.

src/test/java/com/otilm/core/integration/search/HiddenAttributeAgreementITest.java
```diff
@@ -0,0 +1,201 @@
...
+    void theCatalogueStillOffersAVisibleFieldBesideIt() {
+        SearchFieldDataDto visible = field(discoveryService.getSearchableFieldInformationByGroup(), identifier(VISIBLE))
+                .orElseThrow();
+
+        Assertions.assertEquals(true, visible.getDisplayable());
+        Assertions.assertEquals(true, visible.getSortable());
+    }
+
+    @Test
+    void aColumnOnTheHiddenFieldProjectsNothing() {
```
ivosh: **. Nitpick** | Testing

The class Javadoc (`HiddenAttributeAgreementITest.java:46-48`) claims all four read paths are pinned, and the PR description says "with a visible control beside it" — but only two paths actually have one. The catalogue has `theCatalogueStillOffersAVisibleFieldBesideIt` (line 94) and the ordering has `anOrderingOnTheVisibleFieldIsApplied` (line 125, which asserts `[alpha, beta]` against the listing's default `created DESC` order at `DiscoveryServiceImpl.java:211`, so it genuinely distinguishes). The two paths whose only assertion is "nothing came back" have none:

- `aColumnOnTheHiddenFieldProjectsNothing` (line 103) asserts `getAttributeValues()` is null. A column whose identifier the projector cannot match is silently ignored rather than rejected — see `AttributeColumnProjectionITest.anUnknownAttributeColumnLeavesTheListingIntact:329` and `anIdentifierWithoutAContentTypeIsIgnored:338` — so a wrong source, name or content type in the request produces the same null.
- `aFilterOnTheHiddenFieldMatchesNothing` (line 133) asserts `List.of()`. `RequestValidatorHelper.revalidateSearchRequestDto` only defaults paging, and `FilterPredicatesBuilder.requireFilterableContentType` throws only for withheld content types, so a filter naming an unregistered attribute or the wrong content type also matches nothing.

Neither hidden case would fail if the request were mis-built or the visibility predicate were removed for an unrelated reason — CLAUDE.md's "Tests that look green but cover nothing". The sibling that gets this right is `AttributeFilterAuthorizationITest.java:117-119`, where `anAttributeTheCallerMayReadNarrowsTheListing` sits beside the `List.of()` case using the same helper.

_Suggested:_ Add a column case on `VISIBLE` asserting `production` is projected for alpha, and a filter case on `VISIBLE` asserting it returns alpha, so neither hidden-path absence assertion can pass on a mis-built request.

src/test/java/com/otilm/core/integration/search/HiddenAttributeAgreementITest.java
```diff
@@ -0,0 +1,201 @@
...
+        SearchRequestDto request = new SearchRequestDto();
+        request.setColumns(List.of(new SearchColumnRequestDto(FilterFieldSource.CUSTOM, identifier(HIDDEN))));
+
+        List<DiscoveryListDto> discoveries = list(request);
+
+        // Both fixtures on the page, so the per-entry assertion below is not vacuous. Unsorted, hence order-free.
+        Assertions.assertEquals(List.of(alpha.getName(), beta.getName()), names(request).stream().sorted().toList());
```
ivosh: **. Nitpick** | Testing

`HiddenAttributeAgreementITest.java:109-113`: the comment says "Both fixtures on the page, so the per-entry assertion below is not vacuous", but the guard on line 110 calls `names(request)`, which runs `list(request)` a *second* time (line 142-143), while the loop on lines 111-113 asserts over `discoveries` from the first call on line 107. So the guard establishes that the second invocation returned both fixtures, not that `discoveries` was non-empty — if `discoveries` were empty the loop would still be vacuous and the test would still be green. It also doubles the listing work for the assertion.

_Suggested:_ Derive the guard from `discoveries` itself — assert its mapped, sorted names equal `[alpha.getName(), beta.getName()]` — rather than re-running the listing through `names(request)`.

src/main/java/com/otilm/core/model/SearchFieldObject.java
```diff
@@ -53,12 +49,12 @@ public SearchFieldObject(String attributeName, AttributeContentType attributeCon
     }
 
     public SearchFieldObject(String attributeName, AttributeContentType attributeContentType,
-            AttributeType attributeType, String label, BaseAttribute attributeDefinition) {
+            AttributeType attributeType, String label, boolean visible, BaseAttribute attributeDefinition) {
```
ivosh: **. Nitpick** | Documentation

`SearchFieldObject.java:51-52` is a public constructor this PR modified, and the two facts a future reader needs are both off-page. First, it now receives `boolean visible` *and* `attributeDefinition`, which still carries `properties.visible` — the old body derived the flag from that very argument (`AttributeDefinitionProperties.isVisible(attributeDefinition)`), so the redundant parameter reads like an oversight unless you know the column is authoritative and the document must not be re-read. That is the whole point of the PR, and it currently lives only in the PR description; the field Javadoc on line 35 gets halfway ("as mirrored on the definition row") but says nothing about not re-deriving. Second, the constructor's only callers are the four JPQL constructor expressions at `AttributeDefinitionRepository.java:68, 74, 95, 101` — positional argument lists inside strings that the Java compiler cannot check against this signature, so the two have to be kept in step by hand. Per CLAUDE.md "Writing for future AI sessions", method-level Javadoc on non-obvious behaviour is the intended home for both.

_Suggested:_ Add a two-line Javadoc on the constructor saying `visible` comes from the mirrored `attribute_definition.visible` column and must not be re-derived from `attributeDefinition`, and that the argument order is duplicated in the `AttributeDefinitionRepository` JPQL constructor expressions.

src/main/resources/db/migration/V202609071200__attribute_definition_visible.sql
```diff
@@ -0,0 +1,7 @@
+-- Mirrors properties.visible out of the stored definition document so projection, ordering and
+-- filtering share one predicate. Absent in the document means visible, which is the column default.
+ALTER TABLE attribute_definition ADD COLUMN visible BOOLEAN NOT NULL DEFAULT TRUE;
+
+UPDATE attribute_definition
```
ivosh: **~ Minor** | Testing

V202609071200__attribute_definition_visible.sql:5 is the only thing that decides `visible` for every `attribute_definition` row that already exists, and this PR's Verification section says "Core's tests build the schema from entities and never run Flyway, so this is the only coverage the migration has". The first half checks out — `src/test/resources/application.yml:136` sets `spring.flyway.enabled: false` and `ddl-auto: create-drop` builds the schema from the entities. The conclusion does not: `src/test/java/com/otilm/core/integration/migration/` carries 13 migration ITests plus `BaseMigrationTest.java`, which exist *because* Flyway is off. `BaseMigrationTest` routes each class to its own Testcontainers schema, and the SQL-only ones read the script off the classpath and execute it against seeded pre-migration rows: `TokenInstanceConnectorInterfaceMigrationITest.java:26` drops the entity-created column, runs `V202608201200` and asserts `information_schema`, and `AlignCertificateSchemaWithEntitiesITest.java:23` says in its own javadoc "The regular test bootstrap applies the migration to an empty schema, so without this test the backfill UPDATEs conversion would never execute against data" — the same gap this migration has. The four document shapes the PR enumerates (explicit `false`, explicit `true`, `properties` without the key, no `properties`) are precisely what that harness pins; verified by hand on a local Postgres 16 they are pinned by nothing in CI, and a backfill that mis-reads the path fails open in the same direction #2220 was filed about.

_Suggested:_ Add a `BaseMigrationTest` subclass that drops the entity-created `visible` column, seeds the four definition-document shapes, executes V202609071200 off the classpath and asserts the backfilled column.

<sub>flagged by Reference integrity + Persistence & schema</sub>

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -1340,6 +1329,10 @@ private static List<Predicate> attributeReadabilityPredicates(final CriteriaBuil
         // Only custom definitions carry a permission model and a visibility the platform enforces. On a data or
         // metadata definition `visible` is a connector's display hint, and the nullable `enabled` column is unset,
         // so applying either would drop rows a listing is meant to return.
+        //
+        // The projection and the catalogue do withhold a hidden data or metadata field, because those decide what is
```
ivosh: **. Nitpick** | Documentation

FilterPredicatesBuilder.java:1332-1335 adds a second paragraph above the `attributeType != AttributeType.CUSTOM` early return. The pre-existing three lines directly above it (1329-1331) already give the reason the split exists — `visible` is a connector display hint on data/metadata definitions, and applying it would drop rows a listing is meant to return — and the new paragraph's first two sentences restate that in other words ("Filtering is not rendering"). Its closing clause then narrates how to read this PR rather than the code: "the split is deliberate, not an oversight the projection predicate was meant to close" only makes sense to someone who has the before/after of this change in front of them; a reader six months from now has no "projection predicate" event to anchor it against, and the substance is already in the PR description under "Scope held deliberately". The rendering-vs-filtering distinction is genuine rationale and is worth keeping — it is the review-facing tail that is residue.

_Suggested:_ Keep one sentence of the new paragraph stating that rendering paths withhold a hidden display hint while filtering does not, and drop the closing "the split is deliberate, not an oversight the projection predicate was meant to close" clause.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
```diff
@@ -1309,17 +1304,11 @@ private static void requireFilterableContentType(final String fieldIdentifier,
...
+     * so this cannot fall open on a document whose shape has moved.
      */
     private static Predicate definitionIsVisible(final CriteriaBuilder criteriaBuilder, final Join joinDefinition) {
-        final Expression<String> visible = criteriaBuilder
-                .function(JSONB_EXTRACT_PATH_TEXT_FUNCTION_NAME, String.class,
-                        joinDefinition.get(AttributeDefinition_.definition),
-                        criteriaBuilder.literal(DEFINITION_PROPERTIES_KEY), criteriaBuilder.literal(VISIBLE_KEY));
-        return criteriaBuilder
-                .or(criteriaBuilder.isNull(visible), criteriaBuilder.notEqual(visible, Boolean.FALSE.toString()));
+        return criteriaBuilder.isTrue(joinDefinition.get(AttributeDefinition_.visible));
```
ivosh, line 1317: **. Nitpick** | Documentation

FilterPredicatesBuilder.java:1311 turns `definitionIsVisible` into a hard `visible = true` predicate on every custom filter and sort, and this PR's `HiddenAttributeAgreementITest.aFilterOnTheHiddenFieldMatchesNothing` pins that a filter on the hidden field returns nothing. The published field catalogue for that same field is not adjusted to match: SearchHelper.java:276 derives only `displayable`/`sortable` from `isVisible()`, while SearchHelper.java:167 still hands out the full condition set for the content type and SearchHelper.java:169 still publishes the definition's predefined option list as `value`. So the `SearchFieldDataDto` for `internal-routing|TEXT` advertises `EQUALS`, `CONTAINS`, ... on a field whose filter can only ever return zero rows — the client filters, gets an empty page, and is told nothing. Three lines above, SearchHelper.java:159-166 is the established handling for exactly this shape: encrypted and withheld content types have their conditions narrowed to `EMPTY`/`NOT_EMPTY` because "offering a value condition would publish a filter the listing then refuses". Consumers do read `conditions` without gating on `displayable` — `OmniTrustILM/mcp-server` SearchServiceImpl.formatField prints `field.getConditions()` for every catalogue field it lists. If a not-visible custom field is meant to stay in the catalogue for completeness, that reading needs to be recorded next to `isDisplayable`; if not, the field or its value conditions belong out of the published catalogue.

_Suggested:_ Either omit a not-visible custom field from the published catalogue or narrow its conditions the way SearchHelper already narrows encrypted and withheld content types, and record the chosen reading next to isDisplayable.

src/test/java/com/otilm/core/attribute/engine/ProjectedContentWithholdingTest.java
```diff
@@ -0,0 +1,109 @@
...
+ * Which loaded content rows reach a column: a row of ciphertext, and a row of a content type no list cell renders, are
+ * dropped even when the query returns them.
+ */
+class ProjectedContentWithholdingTest {
+
+    private static final UUID OBJECT_UUID = UUID.randomUUID();
+
+    private static final String ATTRIBUTE_NAME = "environment";
+
+    private static final String FIELD_IDENTIFIER = ATTRIBUTE_NAME + "|STRING";
```
ivosh: **. Nitpick** | Maintainability

`ProjectedContentWithholdingTest.java:39` builds the identifier as `ATTRIBUTE_NAME + "|STRING"`, folding both the separator and the enum constant's name into one literal. Every other identifier-building site in the suite derives it from the typed pieces — including the other test file this PR adds (`HiddenAttributeAgreementITest.java:152`: `attributeName + "|" + AttributeContentType.TEXT.name()`), plus `AttributeColumnProjectionITest.java:151`, `KeyAttributeColumnProjectionITest.java:174` and `AttributeColumnSortITest.java:149`. `AttributeFieldIdentifier.SEPARATOR` already exists as a public constant for the `|`, and `AttributeFieldIdentifier.contentType()` resolves the second part with `AttributeContentType.valueOf`, so a rename of the enum constant leaves this literal compiling and the test silently exercising an identifier that parses to no content type (`parse` succeeds, `contentType()` returns null, and `parseRequestedColumns` skips the column — the row is dropped for the wrong reason). *Clean Code*'s no-duplicated-literal rule, and the same spirit as *Effective Java Item 34*: let the enum be the authority for its own name. Two of the three tests in the file assert only `assertNull`, so this failure mode would not surface.

_Suggested:_ Build FIELD_IDENTIFIER from AttributeContentType.STRING.name() and AttributeFieldIdentifier.SEPARATOR, matching HiddenAttributeAgreementITest.identifier().

ivosh (dismissed):

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Confirmed end to end, and fixed with option (a), applied to the sort path.

Reproduced first: two `META` definitions of `connector-region|STRING` under one connector, the hidden one holding `aaa` on `discovery-alpha` and the visible one `zzz` on `discovery-beta`. The catalogue publishes `sortable: true` because the merged flag is the union, `ListingSortResolver` therefore admits the sort, and `attributeReadabilityPredicates` returns at the early return for `META` — so the subquery's first row was the hidden definition's and the page came back `[discovery-alpha, discovery-beta]` while the column beside it rendered only beta's `zzz`. That is exactly the sequence the new test reports when the predicate is removed.

`getAttributeSortKey` now adds `definitionIsVisible` for every non-custom type (custom already carries it from the readability predicates). The predicate sits on the sort path rather than in the shared readability predicates, so the filter's documented split is untouched — `aHiddenMetadataFieldKeepsItsValueConditionsAndStaysFilterable` pins that, an `EQUALS` filter on a fully hidden metadata field still matching its row.

Not option (b): restricting the union at `SearchHelper.java:359` to `CUSTOM` would publish `displayable: false` for a merged metadata field whose visible definition the projection does render, so the catalogue would withhold a column that works.

The javadoc at `FilterPredicatesBuilder.java:1218` now lists visibility among what ordering is gated by, and the paragraph at `:1333` says the projection and the ordering withhold a hidden display hint while a filter still matches it.

Covered by `HiddenAttributeAgreementITest.anOrderingOnADuplicatedMetadataFieldReadsOnlyTheVisibleDefinition`, which asserts both halves of the contract: the order is `[discovery-beta, discovery-alpha]` and the cell on the first row reads `zzz` while alpha's is absent. The fixture puts the lower value behind the hidden definition and in the fixture's own order, so neither a sort that never ran nor one reading the hidden row can produce the expected sequence. Removing the predicate fails it with `expected: <[discovery-beta, discovery-alpha]> but was: <[discovery-alpha, discovery-beta]>`.

Ran on JDK 21: `mvn -P test-integration-core -Dtest='HiddenAttributeAgreementITest,AttributeColumnSortITest,AttributeColumnProjectionITest,SearchHelperITest'` — 52 tests, 0 failures — and the full `test-integration-core` profile.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
Jakub-Bartal-01: Fixed by narrowing the conditions, following the encrypted/withheld handling at `SearchHelper.java:159-166`.

`prepareSearchForJSON` now also narrows to `EMPTY`/`NOT_EMPTY` when `hasNoMatchableContent` holds — a custom definition marked not visible. The presence pair survives because it needs no value to answer: with `definitionIsVisible` on the filter subquery the field is empty for every row, which is what the column beside it renders. A value condition could only ever return nothing, and a published condition is not merely a hint — `ListViewServiceImpl.catalogueOf`/`accepts` validates a saved list view against it, so advertising `EQUALS` also let a view be saved on a filter that can never match.

The reading is recorded on `hasNoMatchableContent`, beside `isDisplayable`, including why the same flag changes nothing for the other sources: on a data or metadata definition it is a connector display hint that filtering does not apply, so those conditions stay whole. The check is gated on `AttributeType.CUSTOM` for that reason — narrowing a hidden metadata field would have removed filters that do work.

I did not drop the field from the catalogue. Its identifier is what a saved view and a filter request name, and `ListViewServiceImpl` refuses any field the catalogue does not offer, so omitting it would turn a saved view that currently returns an empty page into a validation error.

I left `value` as it is. The definition's predefined option list is published for encrypted and withheld fields too, so dropping it for this case alone would introduce a new asymmetry rather than remove one.

Covered by `HiddenAttributeAgreementITest.theCatalogueOffersTheHiddenFieldNoValueCondition`, which fails without the change reporting the full ten-operator set. Two controls beside it: `theCatalogueStillOffersAVisibleFieldBesideIt` now asserts the visible custom field still offers `EQUALS`, and `aHiddenMetadataFieldKeepsItsValueConditionsAndStaysFilterable` asserts the metadata branch keeps its value conditions and still matches its row.

Ran on JDK 21: `mvn -P test-integration-core -Dtest='HiddenAttributeAgreementITest,AttributeColumnSortITest,AttributeColumnProjectionITest,SearchHelperITest'` — 52 tests, 0 failures — and the full `test-integration-core` profile.

src/main/java/com/otilm/core/attribute/engine/AttributeColumnProjector.java
Jakub-Bartal-01: Fixed. The comment above `AttributeColumnProjectionITest.aHiddenAttributeColumnProjectsNoValues` now ends "so the projection query excludes the definition and its values are never loaded", which is what `ad.visible = true` in `getProjectedAttributesContent` does, and it no longer contradicts that query's javadoc.

`grep -rn "on the way out" src/test/java` now returns nothing.

Verified with `mvn -P test-integration-core -Dtest=AttributeColumnProjectionITest` — 14 tests, 0 failures.

src/main/java/com/otilm/core/util/FilterPredicatesBuilder.java
Jakub-Bartal-01: Fixed. The paragraph is one sentence: "A display hint governs what is rendered, so the projection and the ordering that arranges it withhold a hidden data or metadata value while a filter still matches it." The "the split is deliberate, not an oversight the projection predicate was meant to close" clause is gone.

It now names the ordering as a rendering path, which is true of the code after the sort-key predicate added on the `SearchHelper.java:359` thread.

src/test/java/com/otilm/core/attribute/engine/ProjectedContentWithholdingTest.java
Jakub-Bartal-01: Fixed. `FIELD_IDENTIFIER` is now `ATTRIBUTE_NAME + AttributeFieldIdentifier.SEPARATOR + AttributeContentType.STRING.name()`, so a rename of the enum constant breaks the compile rather than leaving the test exercising an identifier that parses to no content type.

Verified with `mvn -P test-non-integration -Dtest=ProjectedContentWithholdingTest` — 4 tests, 0 failures.

ivosh (approved):

src/main/resources/db/migration/V202609071200__attribute_definition_visible.sql
```diff
@@ -0,0 +1,7 @@
+-- Mirrors properties.visible out of the stored definition document so projection, ordering and
+-- filtering share one predicate. Absent in the document means visible, which is the column default.
+ALTER TABLE attribute_definition ADD COLUMN visible BOOLEAN NOT NULL DEFAULT TRUE;
```
ivosh: **? Question** | Security — Major if confirmed

src/main/resources/db/migration/V202609071200__attribute_definition_visible.sql:3-7 backfills `visible` once, while src/main/java/com/otilm/core/dao/entity/AttributeDefinition.java:151-153 synchronizes only writers running this PR's binary. The supported Helm chart advertises a Kubernetes Deployment and configurable replicas, so an old pod can insert or update a hidden definition after Flyway commits without writing the new column; its `TRUE` default then makes new projection, filter, and sort code treat that definition as visible indefinitely. Org-wide `gh search code` hit the API rate limit while checking rollout templates, so this is safe if upgrades stop every old pod before migration; otherwise the mixed-version window recreates the fail-open leak this PR is intended to remove.

_Suggested:_ Keep the column synchronized in the database during mixed-version deployments, such as with an INSERT/UPDATE trigger retained for one release, or enforce and document a stop-all-old-pods upgrade.

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -353,6 +373,11 @@ private static SearchFieldObject mergeDuplicateField(final SearchFieldObject mer
...
```
ivosh: **~ Minor** | Correctness

src/main/java/com/otilm/core/util/SearchHelper.java:371-379 independently merges “some definition is plaintext” and “some definition is visible.” For a duplicated custom field whose hidden definition is plaintext and whose visible definition is encrypted, the merged object becomes visible and plaintext even though no definition is both; src/main/java/com/otilm/core/util/SearchHelper.java:278-298 consequently advertises display, sorting, and value operators, while this PR's projection, filter, and sort predicates can return no readable value. Requests accepted from that catalogue therefore produce empty cells, no value matches, or an unapplied ordering.

_Suggested:_ Represent the merged capability as “there exists one visible, non-encrypted definition,” separately from metadata's hidden-but-filterable rule, and cover the crossed hidden/plain plus visible/encrypted case.

Issue #2192, opened 2026-09-02
Document the coupled interfaces and core PR flow


lubomirw:
### Description

The coupled-PR flow spans two repositories, has six ordered steps, and contains one step that will be done wrong from memory every time. It needs a written runbook before anyone but its author uses it.

The trap: the tail requires **"Re-run all jobs"**, never "Re-run failed jobs". On a coupled PR the `resolve-interfaces` job *succeeded* — it emitted `override=pr` — so a failed-jobs re-run does not re-execute it. The gate reads the stale output and stays red. Even if it went green, the compile and test evidence would still come from the PR-snapshot build rather than mainline.

The second trap: it is **two** workflows to re-run, not one. `test_docker_image.yaml` builds from the pom and has no override, so it fails at compile for the whole review and needs its own re-run afterwards.

The runbook must cover:

1. Open the `interfaces` PR first; its job summary prints the `Depends-On:` line.
2. Paste it into the `core` PR body; both PRs are reviewed in parallel.
3. Expect **three** non-green checks for the whole review — `Interfaces pin` red, `Build` red, and both arch contexts permanently "Expected". No container build and no Trivy scan happen until the tail, and Sonar produces no analysis either, so coupled PRs are reviewed without its input.
4. Merge `interfaces`; wait about three minutes for the publish.
5. Re-run **all jobs** on **both** workflows. Do not push — a push dismisses the approval and you cannot re-approve your own.
6. Merge `core`.

Plus:

- **Building locally.** CI overrides the version but the pom does not, so IntelliJ and local `mvn` will not find the new types. `core/.gitignore` already ignores `/.mvn`, so create `.mvn/maven.config` with a single `-Dinterfaces.version=…` line; both honour it and it cannot be committed by accident. Delete it when the work lands.
- Editing a PR body does not re-trigger CI, but the resolver reads the body live, so re-running all jobs does pick up an edited marker.
- The escape hatch: `Interfaces-Version: 2.20.0-M<n>-<sha8>` when someone else's merge reddens your unrelated PR. Pick the newest one below the merge that broke you. It unblocks work and review, not merge — remove it and re-run before merging. A pin stays resolvable for at least 90 days; the weekly cleanup prunes an `-M` build only once it is both outside the newest 50 and older than that.
- **Fork PRs have no coupled flow.** `publish-pr` skips forks, which get a read-only token and cannot publish, so a fork-authored breaking change falls back to today's behaviour.
[...]

PR #2236, opened 2026-09-08
Resolve the interfaces version from a PR marker during review

### LINKED ISSUE #2191: Resolve the interfaces version from a PR marker in CI
### Description

Let a `core` PR build against an unmerged `interfaces` PR without ever editing `core/pom.xml`.

The obvious design — commit the pin, revert it before merge — is rejected. `core`'s ruleset sets both `dismiss_stale_reviews_on_push` and `require_last_push_approval`, so the revert push dismisses the approval and the author cannot re-approve their own push. Every coupled merge would end waiting on a second human, which is the cost this work exists to remove.

Instead the pom keeps the mainline version permanently and CI overrides it from a marker in the PR body:

```
Depends-On: OmniTrustILM/interfaces#940     # build against that PR's snapshot
Interfaces-Version: 2.20.0-M904-bea47fb9     # build against that exact version
```

A `resolve-interfaces` job emits a `-Dinterfaces.version=…` argument consumed by `compile`, `test` and `package`, plus an `override` mode consumed by a `pin-gate` job that fails on anything but `none`. The gate wires into the aggregate `build` job's `needs:`, so the already-required `Build` context covers it and no ruleset change is needed.

The `Depends-On` form clears itself: once the referenced PR is merged the resolver drops the override and the build validates against mainline. `Interfaces-Version` does not — it means "I am knowingly building against something other than mainline" and must be removed by hand. One rule covers both: **no override may be active at merge time.** The escape hatch unblocks work and review, not merge.

Three implementation details are load-bearing.

**The resolver must fetch the PR body live**, not read `github.event.pull_request.body`. A workflow re-run replays the original event payload, so an edited marker would be invisible — and the escape hatch could then never be cleared without a push, reintroducing the exact failure this design avoids.

**It must look the coordinate up, not derive it** from `core`'s own property. Across a release cut `interfaces` publishes `2.21.0-alpha-PR<n>-SNAPSHOT` while `core` still reads `2.20.0`, and deriving would silently resolve a stale pre-bump artifact. Release-cut week is exactly when coupled PRs are in flight. Querying GitHub Packages for whatever ends in `-alpha-PR<n>-SNAPSHOT` also yields a legible error when the `interfaces` build has not published yet.

**Both extracted values must be regex-constrained** (`[0-9]{1,7}` and `[0-9][0-9A-Za-z.\-]{0,60}`). Both repos are public and PR bodies are attacker-controllable.

`formatting` is deliberately left off the resolver — `spotless:check` and `checkstyle:check` do not compile against `interfaces`.

Design and plan: `docs/superpowers/2026-09-01-interfaces-core-pr-coupling/` (design §4.1–§4.3, plan Phase 3).

### Definition of Done

- [ ] `<interfaces.version>` extracted into `<properties>` in `core/pom.xml`; the dependency uses `${interfaces.version}`; the value stays at the mainline SNAPSHOT.
- [ ] `resolve-interfaces` job fetches the PR body via the API, not the event payload, and needs no checkout.
- [ ] It resolves the PR coordinate by querying GitHub Packages, and fails with an explicit "No published snapshot for interfaces#N" when the `interfaces` build has not published yet.
- [ ] `pin-gate` fails for both override modes with a message naming the required action, and is reachable from the required `Build` context via `needs:`.
- [ ] `compile`, `test` and `package` consume the emitted argument; an empty value leaves the command unchanged.
- [ ] On `merge_group` the resolver returns empty and the build uses the pom value.
- [ ] Verified: a PR with no marker behaves exactly as before.
- [ ] Verified: a `Depends-On` PR builds green against the snapshot with `Interfaces pin` red.
- [ ] Verified: after the referenced PR merges, **re-running all jobs without pushing** turns the gate green and the existing approval is not dismissed.
- [ ] Verified: the escape-hatch happy path resolves and builds, and **removing** the marker then re-running is picked up — if it still reports `explicit`, the resolver is reading the payload rather than fetching live.
- [ ] Verified: a marker containing shell metacharacters is rejected by the regex rather than executed.
- [ ] Sonar quality gate and ≥80% coverage: not applicable — workflow YAML and one pom property, no compiled code.

---

**Updated 2026-09-04 for changes on `interfaces#953`.** The immutable coordinate now
carries the commit's short SHA: `<base>-M<count>-<sha8>`, e.g. `2.20.0-M904-bea47fb9`.
The resolver's `Interfaces-Version` regex already accepts that form, so no change is
needed there — verified against the longer coordinate.

Two publish preconditions also changed, which matters only for the diagnostic when no
coordinate exists: draft PRs now **do** publish (that filter was removed), while
**Dependabot** PRs never can, because GitHub forces a read-only token on their runs
whatever the workflow requests. Fork PRs still cannot publish.

lubomirw:
## TLDR

A coupled `interfaces` change cannot be validated here until it merges. This adds a marker in the PR body — `Depends-On: OmniTrustILM/interfaces#<n>` — that points this build at that PR's published snapshot for the duration of review, and a gate that keeps the override from ever reaching `main`.

The pom is never edited. `<interfaces.version>` becomes a property whose mainline value stays put; CI overrides it with `-Dinterfaces.version=`.

Closes #2191

## The two markers

Both go in the PR body, one per line.

- **`Depends-On: OmniTrustILM/interfaces#940`** — waits on a PR. Self-clearing: when that PR merges, the same unchanged line stops meaning "override" and starts meaning "nothing to do", so the gate goes green on a re-run with no body edit and no push.
- **`Interfaces-Version: 2.20.0-M906-480022e4`** — pins an exact published build, for when someone else's merge reddens an unrelated PR. Does not self-clear; delete it before merging.

`Interfaces pin` is red the whole time either is active. That is the point — it is what makes merging early impossible. It is reached transitively through the required `Build` job's `needs`, so it must not be dropped from that list.

## Why the marker parsing is as strict as it is

A marker that is *almost* right is the dangerous case: the author believes they pinned something, the build quietly does not, and a behaviour-only interfaces change merges having validated nothing.

So the extractors are strict — the line starts at column 0, nothing follows the value — while the *presence* probes are deliberately loose. A bulleted, indented, bold, blockquoted or URL-form marker fails the build with a message naming the accepted form, instead of vanishing. Verified against all of them.

Two things are deliberately **not** matched: markers inside a fenced code block or an HTML comment (so a PR body quoting this syntax is safe — this PR's own body would otherwise trip its own gate), and `Depends-On:` lines naming any other repository in the org.

## Notes for review

- **`generated` needs the override too.** `mvn test` runs the compile and test-compile phases, so that job builds the whole tree. Without this it died on javac for the whole review of exactly the PRs this mechanism exists for, and its diagnostics all blame the identity tables.
- **API reads are captured and checked.** A command substitution inside an `if` condition does not trip `set -e`, so a 404 or a rate limit used to read as "not merged".
- **The resolved version reaches the build jobs through `env:`**, not a `run:` interpolation, and its whole name is anchored rather than just the `-alpha-PR<n>-SNAPSHOT` suffix — the prefix comes from the version in the interfaces PR's own pom.
- **Publishing is opt-in.** An interfaces PR publishes a snapshot only while it carries the `publish-snapshot` label (OmniTrustILM/interfaces#955, merged). That is why the missing-snapshot error names the label first.

## Known and deliberately out of scope

- **No container build or Trivy scan during coupled review.** `test_docker_image.yaml` builds from the pom and carries no override, so the arch contexts stay non-green until the pin clears. That is the design's recorded decision; overriding it needs an `extra-maven-args` input on the org's `containers-test.yml`, which 31 repositories consume. Sonar took a cheaper route and now does report during coupled review — see below.
- **Contributor documentation stays in the planning workspace for now.** Phase 6 (#2192) publishes the runbook into this repository; a second copy written here would drift from it.
- **Mainline SNAPSHOT ordering in `interfaces`.** Two merges to `main` whose publishes finish out of order leave `2.20.0-SNAPSHOT` on the older tree. Zero occurrences across 30 publishes over two weeks; it belongs to `publish-package.yml`, not here.

## What a coupled pull request sees

`Interfaces pin` is red, `Build` is red behind it, and the two arch contexts are non-green — all four clear together once the interfaces PR merges, publishes, and all jobs are re-run.

Sonar is the exception: it now analyses whenever `Package` is green, so the quality gate and coverage report throughout the review rather than arriving after approval. `Interfaces pin` is the sole merge block.

## Testing

`InterfacesPinGuardTest` parses this workflow and asserts the two list memberships the merge block rests on: `pin-gate` inside `build`'s `needs`, and `$MVNARG` on every job running a Maven phase that compiles the tree. A third assertion pins the set of jobs the detection matches, so a detection that stops matching cannot pass over an empty set. Verified to fail on both invariants when either is broken.

The marker handling was exercised end to end on this pull request rather than only reasoned about — all three resolver modes, both gate messages, `Build` going red rather than skipped, Maven resolving a PR-only artifact, and four markers planted in fenced blocks and an HTML comment all correctly ignored.

Sonar quality gate and >=80% coverage: the guard test is the only compiled code added and is pure assertion over a workflow file.


.github/workflows/build_pr.yml
```diff
@@ -15,6 +15,171 @@ permissions:
...
+          # environment.
+          if ! BODY=$(gh api "repos/${GITHUB_REPOSITORY}/pulls/$PR" --jq '.body // ""'); then
+            fail "Could not read the body of ${GITHUB_REPOSITORY}#${PR}."
+          fi
+
+          # Fenced blocks and HTML comments go first: the regexes below are
+          # line-oriented and know nothing about Markdown, so a body that quotes
+          # the marker syntax - this feature's own PR body, say - would otherwise
+          # pin the build to whatever the example names.
+          BODY=$(printf '%s\n' "$BODY" | python3 -c 'import re,sys; t=sys.stdin.read(); t=re.sub(r"<!--.*?-->","",t,flags=re.S); sys.stdout.write(re.sub(r"^```.*?^```","",t,flags=re.S|re.M))')
```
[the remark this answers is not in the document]
lubomirw: Applied in a573c5be3. The stripper now matches `^[ \t]*(```|~~~).*?^[ \t]*\1`, so an indented or tilde fence is stripped too.

Verified against a body carrying a fence nested in a list item, a three-space-indented fence, a tilde fence and an HTML comment - each holding a marker. Before the change all four reached the probe; after it none does.

Also exercised in CI rather than only locally: run 34212818760 carried all four in the PR body and resolved `MODE: none` with a green `Interfaces pin`.

.github/workflows/build_pr.yml
```diff
@@ -15,6 +15,171 @@ permissions:
...
+            | head -1 || true)
+          # The extractors above are strict - line start, nothing after the
+          # value - so a marker with trailing junk is rejected rather than
+          # truncated: `Interfaces-Version: 2.20.0; rm -rf /` does not become a
+          # pin on 2.20.0. The probes below are deliberately looser, because a
+          # marker the author plainly meant has to fail loudly instead of
+          # vanishing; a bullet, an indent, bold or a blockquote all miss the
+          # extractor. The Depends-On probe names the repository, so a Depends-On
+          # line addressed to any other repo in the org is left alone.
+          if printf '%s' "$BODY" | grep -qiP '^[[:space:]>*_-]*Depends-On[[:space:]]*:[[:space:]*_]*(https?://github\.com/)?OmniTrustILM/interfaces' && [ -z "$NUM" ]; then
```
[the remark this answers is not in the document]
lubomirw: Applied in a573c5be3, with your lookahead.

Confirmed the reported behaviour first: `OmniTrustILM/interfaces-schema#12` matched the probe, missed the extractor, and hard-failed the job with a message naming the wrong repository. With `(?![0-9A-Za-z._-])` it is ignored, while `interfaces#12`, a bare `interfaces` and the pull-request URL form all still match - the last two being unusable markers *for* interfaces, which should fail loudly.

No such repository exists in the org today, so this was latent rather than live. Worth closing anyway: it is the same class as the `documentation#340` case fixed a round earlier, where I anchored the org and repo name but not its right edge.

.github/workflows/build_pr.yml
```diff
@@ -15,6 +15,205 @@ permissions:
...
+  # snapshot for the duration of review. The pom is never touched: core's
+  # ruleset dismisses approvals on push and forbids self-approval of the push,
+  # so a revert commit would cost the author a second review round.
+  resolve-interfaces:
+    name: Resolve interfaces version
+    runs-on: ubuntu-latest
+    outputs:
+      mvnarg:   ${{ steps.resolve.outputs.mvnarg }}
+      override: ${{ steps.resolve.outputs.override }}
+    steps:
```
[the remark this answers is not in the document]
lubomirw: Declared in 8b166939b, though not for the reason given — the 403 does not happen.

`core` is `visibility=public`, so the pulls API is readable without `pull-requests: read`. And the resolver has run eight times on this PR, all green, with the read wired to hard-fail:

```
if ! BODY=$(gh api "repos/${GITHUB_REPOSITORY}/pulls/$PR" --jq '.body // ""'); then
  fail "Could not read the body of ${GITHUB_REPOSITORY}#${PR}."
fi
```

A 403 would have failed the job. Instead one of those runs printed `Explicit override: 2.20.0-M907-1299756b`, a value that existed only in the PR body — so the read demonstrably works, and "will fail for every PR" is contradicted by every run on the branch.

Worth declaring anyway, for a different reason: it works *because* the repository is public, which is a silent dependency on a repo setting that nothing in the workflow mentions. Make `core` private and the failure you describe is exactly what happens. The sibling publish workflow in `interfaces` already declares the same permission on the job making this call, so this was also the odd one out.

Scoped to the job rather than the workflow: the workflow-level block covers all eight jobs, and widening every token to close a gap in one is the wrong shape. A job-level block replaces the default rather than extending it, so `contents: read` is restated and `packages: read` is dropped — this job resolves no artifacts.

No behaviour change today.

.github/workflows/build_pr.yml
```diff
@@ -15,6 +15,208 @@ permissions:
...
+  # marker in the PR body redirects this build at the interfaces PR's own
+  # snapshot for the duration of review. The pom is never touched: core's
+  # ruleset dismisses approvals on push and forbids self-approval of the push,
+  # so a revert commit would cost the author a second review round.
+  resolve-interfaces:
+    name: Resolve interfaces version
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
```
ivosh: **! Major** | Correctness

`.github/workflows/build_pr.yml:26-28` gives the new `resolve-interfaces` job a job-level `permissions:` block of `contents: read` + `pull-requests: read`. A job-level block replaces the workflow-level one rather than merging with it — every permission not listed becomes `none` — so this job alone loses the `packages: read` granted at `build_pr.yml:13-15`. It is also the only job in this workflow with a job-level block (the base file has exactly one `permissions:` key, at line 13), so nothing else in the file signals that the top-level grant is being narrowed.

The job then calls `/orgs/OmniTrustILM/packages/maven/com.otilm.interfaces/versions` at `build_pr.yml:120-123`, which is a GitHub Packages read (`read:packages`). Grounding the established pattern: the only other caller of that exact endpoint in the org, `OmniTrustILM/interfaces/.github/workflows/cleanup-pr-packages.yml`, declares `packages: write` at job level for the same call — it does not rely on inheritance.

If the read is refused, the guarded call at :120 takes the `fail "Could not list the published interfaces package versions."` branch, and it does so on *both* `Depends-On` paths: the merged-proof check at :132 and the snapshot lookup at :161. So `Depends-On:` would never resolve, and the self-clearing behaviour the PR advertises would never trigger either, while `Interfaces-Version:` keeps working because it returns at :105 before the packages call — a confusing split. If you have actually observed the listing succeed with `packages: none`, the fix is a one-line comment on :26 saying the endpoint needs no packages scope; otherwise the permission is missing.

_Suggested:_ Add `packages: read` to the `resolve-interfaces` job's `permissions:` block.

<sub>flagged by Correctness & races + Reference integrity</sub>

.github/workflows/build_pr.yml
```diff
@@ -15,6 +15,208 @@ permissions:
...
 jobs:
+  # A coupled interfaces change cannot be validated here until it merges, so a
+  # marker in the PR body redirects this build at the interfaces PR's own
+  # snapshot for the duration of review. The pom is never touched: core's
+  # ruleset dismisses approvals on push and forbids self-approval of the push,
+  # so a revert commit would cost the author a second review round.
+  resolve-interfaces:
+    name: Resolve interfaces version
+    runs-on: ubuntu-latest
+    permissions:
```
ivosh, line 24: **~ Minor** | Correctness

**The resolver's package lookup has never run under the token this job now carries.**

`build_pr.yml:26-28` gives `resolve-interfaces` a job-level `permissions:` block. A job-level block **replaces** the workflow-level one at `:13-15` rather than extending it. `packages: read` lives only in that workflow-level block, so this job no longer holds it. The runner confirms it: [run 34334966010](https://github.com/OmniTrustILM/core/actions/runs/34334966010) at `509f2f2c` prints `Contents: read`, `Metadata: read`, `PullRequests: read` — and no `Packages`.

This job does read the Packages API. `:120-122` calls `GET /orgs/OmniTrustILM/packages/maven/com.otilm.interfaces/versions`, and `:123` hard-fails the job if that call errors. The block's rationale — the job resolves no artifacts — holds for Maven resolution but does not cover this REST call.

**Why it probably still works.** `com.otilm.interfaces` is a **public** package. Unauthenticated, the endpoint returns `401`. With a classic PAT holding no `read:packages` scope it returns `200`, and the response's `X-Accepted-Oauth-Scopes` header comes back **empty** — the API declares no required scope for the request. So the listing looks gated on authentication rather than on the packages scope, and the documented `read:packages` requirement appears to apply to non-public packages.

**Why that is not proof.** That evidence comes from a classic PAT. `GITHUB_TOKEN` is an installation token, and its permission model is not the OAuth scope model. GitHub's REST reference for this endpoint carries no fine-grained / GitHub App permission section at all, so the docs cannot settle it either way. `OmniTrustILM/interfaces/.github/workflows/cleanup-pr-packages.yml:51-53` records the same open question about the sibling call and declares `packages: write` rather than find out.

**And nothing has exercised it.** The lookup has succeeded six times on this branch — every run from `7022c4da` (09-08 07:05) to `b8aa2e1d` (09-09 08:31) printed `is merged - using the mainline pom value`, which is only reachable after the listing returns. All six predate `8b166939b` (09-09 09:04:45), the commit that added the job-level block, and all six carried `Packages: read`. The only two runs since then printed `MODE: none` and exited at the `[ -z "$NUM" ]` check, because every marker in the current body is inline-quoted. **The headline path of this PR has therefore never executed under the token it will ship with.** If the call does need the permission, the first coupled PR after merge dies at `:123` with "Could not list the published interfaces package versions" — a message that points at publishing rather than at the token.

_Suggested:_ Put one unfenced `Depends-On: OmniTrustILM/interfaces#<open PR>` line in a PR body and re-run — a single run settles this for good. Otherwise add `packages: read` back to the job block; the workflow already grants it to all seven other jobs, so it widens nothing.

<sub>flagged by CI / build + Config & interface-doc conventions</sub>

.github/workflows/build_pr.yml
```diff
@@ -187,6 +394,9 @@ jobs:
 
   generated:
     name: Generated artifacts
+    # `mvn test` runs the compile and test-compile phases, so this job builds the
+    # whole tree and needs the override as much as compile/test/package do.
+    needs: resolve-interfaces
```
ivosh: **~ Minor** | Correctness

This PR routes the resolved version into every Maven invocation in `build_pr.yml` (`:260`, `:302`, `:368`, `:472`) and the comment at `build_pr.yml:397-398` reasons explicitly about which jobs build the tree — but `Test Docker image` is a third PR-time Maven build outside this file. `.github/workflows/test_docker_image.yaml:4-5` triggers on the same branch set as `Build PR` (`main*`, `feat/*`, `hotfix/*`), and `:18-23` calls `OmniTrustILM/.github/.github/workflows/containers-test.yml@main` with `pre-build: maven`; that reusable workflow runs `mvn -f pom.xml -U -B -ntp -T 1C clean package -DskipTests` (containers-test.yml:109) and exposes no input for extra Maven arguments. On a coupled PR it therefore compiles the branch against mainline `2.20.0-SNAPSHOT` and dies on javac — the exact failure the `generated` note in the PR description describes, for the whole review of exactly the PRs this mechanism exists for. `build_preview_docker.yml:76` has the same shape (`pre-build: maven` via `containers-build-and-push.yml`), though it is opt-in behind the `preview` label. Neither appears in the PR description's "Known and deliberately out of scope" list. `Build` is the only required Actions context per `build_pr.yml:491-503`, so this reddens a non-blocking check rather than letting anything through — but it leaves a permanently red check with no way to clear it during review.

_Suggested:_ Either add a Maven-args input to the org's `containers-test.yml` and thread the resolved version through `test_docker_image.yaml`, or list the two container workflows in the PR description's out-of-scope section.

<sub>flagged by Reference integrity + Dependency & CVE + Correctness & races</sub>

.github/workflows/build_pr.yml
```diff
@@ -15,6 +15,208 @@ permissions:
...
+          fi
+
+          # Fenced blocks and HTML comments go first: the regexes below are
+          # line-oriented and know nothing about Markdown, so a body that quotes
+          # the marker syntax - this feature's own PR body, say - would otherwise
+          # pin the build to whatever the example names.
+          BODY=$(printf '%s\n' "$BODY" | python3 -c 'import re,sys; t=sys.stdin.read(); t=re.sub(r"<!--.*?-->","",t,flags=re.S); t=re.sub(r"^[ \t]*(```|~~~).*?^[ \t]*\1","",t,flags=re.S|re.M); sys.stdout.write(t)')
+
+          # head -1 below would otherwise take the earlier line silently.
+          DUPD=$(printf '%s' "$BODY" | grep -ciP '^Depends-On[[:space:]]*:' || true)
```
ivosh: **~ Minor** | Correctness

`build_pr.yml:68` counts `^Depends-On[[:space:]]*:` across every repository, while the extractor at line 74 and the presence probe at line 88 both narrow to `OmniTrustILM/interfaces` — and the comment at lines 85-87 promises that "a Depends-On line addressed to any other repo in the org is left alone". It is not: a body carrying both `Depends-On: OmniTrustILM/interfaces#940` and `Depends-On: OmniTrustILM/some-other-repo#12` yields `DUPD=2` and hard-fails with "The PR body has 2 Depends-On lines. Keep exactly one.", a message that does not explain the real constraint. The stated purpose of the counter — "head -1 below would otherwise take the earlier line silently" — is about the interfaces extractor alone, so scoping it loses nothing. The asymmetry runs the other way too: the counter anchors `^Depends-On` strictly while the probe accepts leading `[[:space:]>*_-]*`, so a bulleted duplicate escapes the count entirely and the strict line silently wins.

_Suggested:_ Narrow the DUPD regex at build_pr.yml:68 to `OmniTrustILM/interfaces` so it counts only the lines the extractor competes over.

<sub>flagged by CI / build + Correctness & races</sub>

.github/workflows/build_pr.yml
```diff
@@ -15,6 +15,208 @@ permissions:
...
+          DUPV=$(printf '%s' "$BODY" | grep -ciP '^Interfaces-Version[[:space:]]*:' || true)
+          [ "${DUPD:-0}" -le 1 ] || fail "The PR body has ${DUPD} Depends-On lines. Keep exactly one."
+          [ "${DUPV:-0}" -le 1 ] || fail "The PR body has ${DUPV} Interfaces-Version lines. Keep exactly one."
+
+          NUM=$(printf '%s' "$BODY" \
```
ivosh, line 68: **. Nitpick** | Testing

`build_pr.yml:73-93` is the security-relevant core of this PR — two strict extractors plus two deliberately loose presence probes — and the PR body states the stakes itself ("A marker that is *almost* right is the dangerous case") and claims "Verified against all of them" for the bulleted / indented / bold / blockquoted / URL forms, plus the fenced-block and HTML-comment exclusions at `build_pr.yml:65`. None of that verification is committed. The Testing section says coverage is "not applicable — workflow YAML and one pom property, no compiled code", which is true of JaCoCo/Sonar but leaves ~15 regexes with no regression check at all: the next person who edits one of these patterns (or the `python3` fence stripper) has nothing that re-checks the accept/reject table, and a pattern that silently stops matching degrades to "marker vanished, build validated nothing" — the exact failure this mechanism exists to prevent. Extracting the script (see the sibling finding on `build_pr.yml:39`) makes this cheap: the same input/expected-output table you verified by hand becomes a few dozen lines of fixture.

_Suggested:_ Commit the verified marker forms as an accept/reject fixture exercised by a small script test, so a future edit to the extractors or the fence stripper re-checks them.

.github/workflows/build_pr.yml
```diff
@@ -15,6 +15,208 @@ permissions:
   packages: read
 
 jobs:
+  # A coupled interfaces change cannot be validated here until it merges, so a
```
ivosh: **~ Minor** | Documentation

This PR introduces a contributor-facing protocol spanning two repositories, and `grep -rniE 'Depends-On|Interfaces-Version|publish-snapshot' --include='*.md'` over the whole tree returns nothing; `docs/` holds only `superpowers`, and there is no PR template or CONTRIBUTING file. The job-level comment at `build_pr.yml:18-22` explains *why* the mechanism exists but never states the accepted syntax — the only in-repo statements of the form are inside two `fail` strings (`build_pr.yml:89` and `:92`). Everything a user needs beyond that lives in surfaces that are not in the tree: marker precedence and the "delete `Interfaces-Version:` before merging" rule are only in this PR's own body; the paste-able `Depends-On:` line and the `publish-snapshot` label requirement are only in a GitHub Actions step summary in `OmniTrustILM/interfaces` (`.github/workflows/publish-pr-snapshot.yml:210-217`, `publish-package.yml:232`). The PR body also defers the merge-window hazard to "the runbook", which does not exist here. Per CLAUDE.md "Writing for future AI sessions", the next reader has no conversation context and nothing to grep.

_Suggested:_ Add a short README or `documentation` section (or a PR-template line) giving both marker forms, their precedence, the `publish-snapshot` prerequisite, and the wait-for-publish rule, and point the job comment at it.

.github/workflows/build_pr.yml
```diff
@@ -15,6 +15,208 @@ permissions:
...
+          [ "${DUPD:-0}" -le 1 ] || fail "The PR body has ${DUPD} Depends-On lines. Keep exactly one."
+          [ "${DUPV:-0}" -le 1 ] || fail "The PR body has ${DUPV} Interfaces-Version lines. Keep exactly one."
+
+          NUM=$(printf '%s' "$BODY" \
+            | grep -oiP '^Depends-On:[[:space:]]*OmniTrustILM/interfaces#\K[0-9]{1,7}(?=[[:space:]]*$)' \
+            | head -1 || true)
+          VER=$(printf '%s' "$BODY" \
+            | grep -oiP '^Interfaces-Version:[[:space:]]*\K[0-9][0-9A-Za-z.\-]{0,60}(?=[[:space:]]*$)' \
+            | head -1 || true)
+          # The extractors above are strict - line start, nothing after the
```
ivosh: **. Nitpick** | Documentation

`build_pr.yml:79-87` is one unbroken paragraph carrying three distinct concerns — extractor strictness (with an inline `rm -rf /` example), why the presence probes are deliberately looser, and how the `Depends-On` probe closes off the repository name against a same-prefix sibling — in sentences of ~35-40 words each chaining em-dashes, colons and semicolons, with no topic sentence and no grouping. The content is accurate and worth keeping; it is just packed too tightly to scan, and the same shape recurs at `build_pr.yml:18-22` and `:100-102`. Restructured without dropping a fact, it reads as a one-line summary plus one short paragraph per concern with a lead-in, e.g. `# Strict extractors, loose probes.` / `# Strict: line start, nothing after the value - trailing junk is rejected, not truncated.` / `# Loose: a bulleted, indented, bold or blockquoted marker must fail loudly, not vanish.` / `# The Depends-On probe closes off the repo name, so a sibling repo's marker is left alone.`

_Suggested:_ Split `build_pr.yml:79-87` into one short paragraph per concern with a leading summary line, capping sentences at roughly 20 words. Alternatively, trim down heavily the prose.

.github/workflows/build_pr.yml
```diff
@@ -52,7 +255,9 @@ jobs:
       - name: Compile
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
-        run: mvn -B -U -ntp test-compile -Dmaven.compiler.proc=full
+          MVNARG: ${{ needs.resolve-interfaces.outputs.mvnarg }}
+        # $MVNARG unquoted on purpose: empty contributes no argument.
```
ivosh: **. Nitpick** | Maintainability

`build_pr.yml:259` records the load-bearing invariant ("$MVNARG unquoted on purpose: empty contributes no argument"), and `sonar.yaml:155` restates it in different words ("Unquoted: empty contributes no argument"). The three remaining expansions carry nothing: `build_pr.yml:302` (test), `:368` (package), `:472` (corpus key). Those are the sites a future reader — or a linter-driven cleanup — is most likely to "fix" by quoting, which would pass an empty `""` argument to `mvn` on every unpinned PR, i.e. break the common path rather than the rare one. Two differently-worded copies and three silent sites is the worst of both: the fact is duplicated where it is present and absent where it matters.

_Suggested:_ State the unquoted-on-purpose invariant once at the `resolve-interfaces` `mvnarg` output declaration (`build_pr.yml:30`) and reference it, or repeat the identical one-liner at all five expansions.

.github/workflows/build_pr.yml
```diff
@@ -15,6 +15,208 @@ permissions:
...
+          # Self-clearing: once the interfaces PR merges, the same unchanged
+          # marker stops meaning "override" and starts meaning "nothing to do".
+          if [ "$MERGED" = true ]; then
+            # Mainline SNAPSHOT carries the change only once the interfaces
+            # publish finishes, 220-367s after the merge. The merge commit's
+            # immutable -M build is the proof that it did.
+            if ! printf '%s\n' "$VERSIONS" | cut -f2 \
+                 | grep -qE "^[0-9][0-9A-Za-z.-]*-M[0-9]+-${MSHA:0:8}\$"; then
+              AGE=$(( $(date -u +%s) - $(date -u -d "${MAT:-@0}" +%s) ))
+              if [ "$AGE" -lt 900 ]; then
```
ivosh: **. Nitpick** | Documentation

`build_pr.yml:135` decides between a retryable failure and a warn-and-proceed at `AGE -lt 900`, but the comment above it (`build_pr.yml:129-131`) documents only the observed publish latency ("220-367s after the merge"), not where 900 comes from — a reader cannot tell whether it is 2.5x headroom over that measurement, an unrelated timeout, or a guess, and it is the constant that decides whether a post-merge re-run blocks or silently builds against mainline. The measurement itself is also stated more precisely than it can stay true (the PR body renders the same figure as "four to six minutes", i.e. 240-360s, so the two surfaces already disagree at the edges), and a change in the interfaces publish pipeline rots it silently.

_Suggested:_ Say in the comment why the threshold is 900s relative to the observed latency, and round the measured window ("roughly four to six minutes") so it does not rot at the second.

.github/workflows/sonar.yaml
```diff
@@ -91,6 +131,17 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-sonar-
       # SonarCloud Scan analysis and push
+      # continue-on-error: a run predating this artifact still analyses, just
+      # against the pom value.
+      - name: Download the resolved interfaces argument
```
ivosh: **. Nitpick** | Maintainability

In sonar.yaml the comment `# SonarCloud Scan analysis and push` (sonar.yaml:133) was a section label sitting immediately above `- name: SonarCloud Scan`. This PR inserts the new `Download the resolved interfaces argument` step, plus its own `# continue-on-error:` comment, between the two (sonar.yaml:134-143), so the pre-existing label now reads as a description of the artifact download and the actual scan step at sonar.yaml:145 is left unlabelled. Every other comment in this file sits directly above the step it describes.

_Suggested:_ Move the `# SonarCloud Scan analysis and push` line down to sit directly above `- name: SonarCloud Scan`, leaving the new `# continue-on-error:` comment attached to the download step.

.github/workflows/sonar.yaml
```diff
@@ -5,11 +5,51 @@ on:
...
+  # a coupled PR, which would skip Sonar for exactly the PRs that need it most.
+  # Every artifact this workflow downloads comes from Package.
+  gate:
+    name: Sonar precondition
+    if: github.event.workflow_run.conclusion != 'cancelled'
+    runs-on: ubuntu-latest
+    outputs:
+      analyse: ${{ steps.c.outputs.analyse }}
+    steps:
+      - id: c
```
ivosh: **. Nitpick** | Maintainability

The new gate step is `- id: c` (sonar.yaml:18) and is referenced as `steps.c.outputs.analyse` (sonar.yaml:16). Every other step id in this file is descriptive - `pr_number`, `get_pr_data`, `cache-month` - and the sibling resolver job this same PR adds uses `- id: resolve` (build_pr.yml:34). A single letter also gives the step no readable label in the Actions UI, since it carries no `name:` either.

_Suggested:_ Rename the step id from `c` to something descriptive such as `check` or `precondition` and update the `analyse` output reference.

.github/workflows/sonar.yaml
```diff
@@ -5,11 +5,51 @@ on:
     types: [completed]
 
 jobs:
+  # Gated on Package rather than the run conclusion: pin-gate fails by design on
+  # a coupled PR, which would skip Sonar for exactly the PRs that need it most.
+  # Every artifact this workflow downloads comes from Package.
+  gate:
```
ivosh, line 11: **. Nitpick** | Security

The `gate` job added at sonar.yaml:11-45 declares no `permissions:` block, and sonar.yaml has none at workflow level either (verified against `main`), so it runs with whatever the repository default grants — on a `workflow_run` trigger that is typically a read/write token. All it needs is the jobs-list read at sonar.yaml:31-32. The `sonar` job has always had this property, so this is not a regression, but the new job is the first place in this file where a scope can be declared, and the sibling workflow sets the example: build_pr.yml:13-15 pins a repo-wide read-only token with a comment explaining why.

_Suggested:_ Add `permissions: { contents: read, actions: read }` to the gate job so the new workflow_run job does not hold a write-capable token.

.github/workflows/build_pr.yml
```diff
@@ -15,6 +15,208 @@ permissions:
...
+      - name: Publish the resolved argument
+        env:
+          ARG: ${{ steps.resolve.outputs.mvnarg }}
+        run: printf '%s\n' "$ARG" > mvnarg.txt
+
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: interfaces-mvnarg
+          path: mvnarg.txt
+          retention-days: 1
```
ivosh: **~ Minor** | Correctness

build_pr.yml:184 gives `interfaces-mvnarg` `retention-days: 1`. Every other artifact `sonar.yaml` consumes from the same run outlives it by a wide margin: `pr-number`, `compiled-classes` and `coverage-reports` are uploaded with no `retention-days` at all (repo default, typically 90 days) and `compiled` uses 7. So re-running the Sonar workflow more than a day after `Build PR` — the normal reaction to a SonarCloud flake or a token rotation — finds all three of its real inputs present and only the mvnarg gone. `continue-on-error: true` on the download (sonar.yaml:137) plus `MVNARG=$(cat mvnarg.txt 2>/dev/null || true)` (sonar.yaml:154) then lets the scan proceed with no override, silently, against exactly the classpath mismatch the step was added to prevent (build_pr.yml:172-174 states that purpose). Related, same fallback: the artifact lands at `path: .` (sonar.yaml:143) after the PR head is checked out at sonar.yaml:40-45, so on a failed download `cat mvnarg.txt` reads an `mvnarg.txt` committed in the head branch instead. That is contained today only by the exposure the file already documents at sonar.yaml:35-39 (the `mvn` step executes the fork's own pom with `SONAR_TOKEN` in scope), so it is not a new escalation — but it does make the silent-no-override path head-selectable rather than merely accidental.

_Suggested:_ Give `interfaces-mvnarg` the same retention as the other Sonar inputs and key the `cat` on the download step's `outcome` so a missing artifact reddens the scan instead of silently dropping the override.

<sub>flagged by Dependency & CVE + CI / build + Test-selection & suppression-baseline</sub>

.github/workflows/sonar.yaml
```diff
@@ -91,6 +131,17 @@ jobs:
...
+      # continue-on-error: a run predating this artifact still analyses, just
+      # against the pom value.
+      - name: Download the resolved interfaces argument
+        continue-on-error: true
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: interfaces-mvnarg
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          run-id: ${{ github.event.workflow_run.id }}
+          path: .
```
ivosh: **. Nitpick** | Security

`sonar.yaml:143` downloads `interfaces-mvnarg` with `path: .` — the root of the fork/head tree checked out at `sonar.yaml:80-85` — and `sonar.yaml:154` reads it with `cat mvnarg.txt 2>/dev/null || true`, word-splitting it unquoted into the `mvn` invocation at line 156. Because the download is `continue-on-error: true`, the no-artifact fallback is not "use the pom value" as the comment at lines 134-135 states: it is "use whatever `mvnarg.txt` the head commit happens to contain", which lets arbitrary extra `mvn` flags reach the step that holds `SONAR_TOKEN`. This is not a new capability for fork PRs — the comment at lines 75-79 already accepts that this step executes the fork's own pom with that token in scope — but it is a second, non-obvious channel, and it also means a stray committed `mvnarg.txt` in a same-repo branch silently redirects dependency resolution for the scan.

_Suggested:_ Download the artifact to a directory outside the checkout (e.g. `path: ../mvnarg`) and read it from there, so the value can only come from the producing job.

<sub>flagged by CI / build</sub>

.github/workflows/sonar.yaml
ivosh, line 11: **~ Minor** | Documentation

The description's "Known and deliberately out of scope" bullet says "`SonarCloud Code Analysis` runs on a green `Build PR`, so a coupled PR gets no Sonar report until the pin clears. It is a required context, so this blocks the merge rather than letting anything through." The new gate in `.github/workflows/sonar.yaml:8-45` does the opposite on purpose: its comment at lines 8-10 states it is "gated on Package rather than the run conclusion: pin-gate fails by design on a coupled PR, which would skip Sonar for exactly the PRs that need it most", and line 13 only excludes `cancelled`, so with `Package` green (line 40) Sonar analyses and posts its context even though `pin-gate` reddens the run. So a coupled PR *does* get a Sonar report, and the thing blocking the merge is `Interfaces pin`, not a withheld Sonar context. That bullet reads like the pre-change behaviour left in the description, and it is the one place a reviewer would look to learn what quality reporting a pinned PR still gets.

_Suggested:_ Rewrite that bullet to state that Sonar now analyses a coupled PR whenever `Package` is green, with `Interfaces pin` as the sole merge block.

<sub>flagged by CI / build + Config & interface-doc conventions</sub>

.github/workflows/sonar.yaml
```diff
@@ -5,11 +5,51 @@ on:
     types: [completed]
 
 jobs:
+  # Gated on Package rather than the run conclusion: pin-gate fails by design on
+  # a coupled PR, which would skip Sonar for exactly the PRs that need it most.
+  # Every artifact this workflow downloads comes from Package.
```
ivosh: **. Nitpick** | Documentation

`sonar.yaml:10` asserts that every artifact this workflow downloads comes from `Package`, and thirty lines later the same PR adds a download of `interfaces-mvnarg`, which `build_pr.yml:180-184` uploads from `resolve-interfaces`. The gate is still sound — `package` lists `resolve-interfaces` in its own `needs` at `build_pr.yml:313`, so a successful Package implies the resolver succeeded and published — but the sentence as written no longer carries that reasoning.

_Suggested:_ Reword the comment to say the resolver is a transitive dependency of Package, so keying on Package still guarantees every downloaded artifact exists.

<sub>flagged by CI / build</sub>

.github/workflows/sonar.yaml
```diff
@@ -5,11 +5,51 @@ on:
...
+            ANALYSE=true
+          else
+            ANALYSE=false
+          fi
+          echo "analyse=$ANALYSE" >> "$GITHUB_OUTPUT"
+
   sonar:
     name: Sonar
+    needs: gate
+    if: needs.gate.outputs.analyse == 'true'
```
ivosh: **~ Minor** | Documentation

This PR's description lists under "Known and deliberately out of scope": "`SonarCloud Code Analysis` runs on a green `Build PR`, so a coupled PR gets no Sonar report until the pin clears. It is a required context, so this blocks the merge rather than letting anything through." The new `gate` job contradicts that: sonar.yaml:31-44 keys `analyse` on the `Package` job's conclusion, and sonar.yaml:50 runs `sonar` whenever that is `success`. On a coupled PR `Package` is green (only `pin-gate` and hence `Build` are red), so Sonar *does* analyse and *does* post its report — which is exactly what sonar.yaml:8-10 says it was written to achieve. A reviewer reading the description will believe a second, independent lock exists on coupled PRs; only `Build` blocks them.

The same condition also un-gates two cases the comment does not mention: a run reddened by `formatting` (build_pr.yml:239) or by `generated` (build_pr.yml:425-449) still leaves `Package` green, so those PRs now get a full analysis and a green `SonarCloud Code Analysis` context where previously the context stayed unreported. Nothing merges through it — `Build` is still red and required — but it is a widening of when the quality gate reports success, decided here in passing.

_Suggested:_ Replace the description's Sonar bullet with what the gate now does — Sonar analyses whenever `Package` succeeds, including runs reddened by `pin-gate`, `formatting` or `generated` — and note in the sonar.yaml:8-10 comment that the last two are intentionally included.

<sub>flagged by Test-selection & suppression-baseline</sub>

.github/workflows/build_pr.yml
```diff
@@ -276,7 +487,7 @@ jobs:
 
   build:
     name: Build
-    needs: [package, formatting, generated]
+    needs: [package, formatting, generated, pin-gate]
```
ivosh, line 479: **~ Minor** | Testing

The whole merge-blocking property of this PR hangs on two list memberships that nothing checks:

1. `pin-gate` staying inside `needs` at build_pr.yml:490 — build_pr.yml:199-200 says so in prose ("so do not drop it from that list"). Drop it and an active `Depends-On`/`Interfaces-Version` override leaves every check green, which is the failure this PR exists to make impossible.
2. Every tree-building job passing `$MVNARG` — build_pr.yml:397-398 states the rule for `generated`; a job added later that runs `mvn test`/`package` without it just reddens for the wrong reason (the very symptom the note says cost a full review round).

Both are silent when they drift, and this repo already has the established shape for exactly that: `src/test/java/com/otilm/core/architecture/CiTestSplitTest.java` parses pom.xml with XPath to hold the CI split that build_pr.yml's matrix consumes (including `transferredClassesMustExist`, a pure staleness check), and `WireMockPortsGuardTest` parses a YAML config with snakeyaml for the same reason — "a drift on either side surfaces … nowhere near the edit that caused it". A sibling guard parsing `.github/workflows/build_pr.yml` could assert both invariants in a few lines. The PR's Testing section ("not applicable — workflow YAML and one pom property") is what leaves the gap.

_Suggested:_ Add a guard beside `CiTestSplitTest` that parses build_pr.yml and asserts `pin-gate` is in `build.needs` and that every job invoking a compile/test/package phase passes `$MVNARG`.

<sub>flagged by Test-selection & suppression-baseline</sub>

.github/workflows/build_pr.yml
```diff
@@ -15,6 +15,208 @@ permissions:
...
+
+      # sonar.yaml is a separate workflow and cannot read job outputs. Without
+      # this it analyses the compiled classes against a classpath they were not
+      # built with.
+      - name: Publish the resolved argument
+        env:
+          ARG: ${{ steps.resolve.outputs.mvnarg }}
+        run: printf '%s\n' "$ARG" > mvnarg.txt
+
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
```
ivosh: **. Nitpick** | Maintainability

`.github/workflows/build_pr.yml:180` adds `- uses: actions/upload-artifact@043fb46d…` with no `name:`, so it shows in the run log as "Run actions/upload-artifact@043fb46d…". Every other artifact upload in the same file carries one — "Upload compiled output" (265), "Upload coverage exec" (304), "Upload PR number artifact" (375), "Upload Compiled Classes" (382), "Upload Coverage Reports" (389) — and it is the only step in this workflow without a name. The neighbouring step this PR adds at line 175 ("Publish the resolved argument") does follow the convention.

_Suggested:_ Give the step a `name:` such as "Upload the resolved interfaces argument" to match the sibling upload steps.

<sub>flagged by Config & interface-doc conventions</sub>

.github/workflows/build_pr.yml
```diff
@@ -15,6 +15,208 @@ permissions:
...
+          if [ -n "$VER" ] && [ -n "$NUM" ]; then
+            echo "::warning::Both markers present - Interfaces-Version wins; ignoring Depends-On: OmniTrustILM/interfaces#${NUM}."
+          fi
+
+          # Reaches the build jobs through the environment, never interpolated
+          # into their run: blocks, so the value is not shell source text there.
+          # The regexes still constrain it; this removes the dependence on them.
+          if [ -n "$VER" ]; then
+            echo "Explicit override: $VER"
+            emit "-Dinterfaces.version=$VER" "explicit"; exit 0
```
ivosh: **. Nitpick** | Error handling

build_pr.yml:105 emits `-Dinterfaces.version=$VER` and exits before the resolver ever fetches the published-version list, which happens later at build_pr.yml:120-124 on the `Depends-On` path only. So an `Interfaces-Version:` pin is validated for *shape* but never for *existence*. A stale or mistyped build id — the documented form is `2.20.0-M906-480022e4`, eight hex characters that are easy to fat-finger — clears both probes, sets `override=explicit`, and then reappears as four independently-red jobs (compile at :260, test at :302, package at :368, generated at :471) whose only diagnosis is a raw `Could not resolve dependencies ... com.otilm:interfaces:jar:<v>`. That is the failure mode the PR's own rationale singles out — the author believes they pinned something and the build quietly did not — arriving without the named message the strict extractors exist to produce. The list is already in hand for the other path, so the check is a reordering.

Provenance is unpinned on this path too: pom.xml:460-472 declares `central-portal` (`https://central.sonatype.com/repository/maven-snapshots/`) as a second repository after `github`, and `com.otilm:interfaces` really is published to both — I fetched `https://central.sonatype.com/repository/maven-snapshots/com/otilm/interfaces/maven-metadata.xml` live and it lists `2.18.1-SNAPSHOT` and `2.20.0-SNAPSHOT`. An explicit pin can therefore resolve from a feed the GitHub-Packages-validated `Depends-On` path would never choose. Both feeds are org-owned (`repo1.maven.org/maven2/com/otilm/` lists `dependencies` and `interfaces`, so the namespace is verified), so this is a diagnostics-and-provenance gap rather than a dependency-confusion risk. Checked with OSV (`api.osv.dev/v1/query` for `com.otilm:interfaces` at `2.20.0-SNAPSHOT`, and `org.apache.tomcat.embed:tomcat-embed-core` at `10.1.59`): both returned `{}`, and no advisory bears on this PR.

_Suggested:_ Move the VERSIONS fetch above the explicit-`VER` branch and fail with a named message when the pinned version is absent from the org's published interfaces versions.

<sub>flagged by Dependency & CVE</sub>

ivosh (changes requested):

.github/workflows/build_pr.yml
lubomirw: Right, and it was my regression — two hours old. Restored in 005148892.

The rationale I wrote for that block ("this job resolves no artifacts") held for Maven resolution and not for the Packages REST call at `:120`, which is the whole point of the job. Confirmed on the runner exactly as you describe: the resolver step at `509f2f2c` prints `Contents: read / Metadata: read / PullRequests: read` and nothing else.

Your point that the lookup has never executed under this token is the part that mattered most — six green lookups, all predating `8b166939b`, and the two runs since both exited at `[ -z "$NUM" ]`. The headline path really had gone unexercised.

I did not run the one-marker experiment you suggest. The line is free, the workflow grants it to the other seven jobs, and the answer does not change what I do here. It is still worth knowing for its own sake, though — `cleanup-pr-packages.yml` in `interfaces` records the same open question and declares `packages: write` rather than settle it, so one deliberate run would retire it in both places.

.github/workflows/build_pr.yml
lubomirw: Fixed in 005148892. The counter now carries the probe's own pattern:

```
'^[[:space:]>*_-]*Depends-On[[:space:]]*:[[:space:]*_]*(https?://github\.com/)?OmniTrustILM/interfaces(?![0-9A-Za-z._-])'
```

That closes both halves. A `Depends-On` for another repository no longer counts, so the comment's promise holds; and because the counter now accepts the same leading decoration the probe does, a bulleted duplicate is counted rather than silently losing to the strict line.

.github/workflows/build_pr.yml
lubomirw: Taking the sibling guard (`build_pr.yml:490`) and leaving this one for now.

You are right that the accept/reject table is verified and uncommitted. The reason for the split is cost: the guard on the two list memberships is a few dozen lines against the parsed YAML, while a fixture over the extractors wants the script out of the workflow first — and extracting it changes how the resolver runs, which I would rather not fold into this PR alongside everything else.

Recorded rather than dropped. The table exists in the PR discussion; the work is the extraction, not the fixtures.

.github/workflows/build_pr.yml
lubomirw: Agreed on the gap; differing on where the fix belongs.

Phase 6 of this epic (#2192) exists to publish the runbook into this repository, and the plan is explicit that it snapshots to one repo only because "three snapshots means three copies that drift". Writing a second description here now creates exactly that.

What I did instead: the two rules that were documented nowhere at all — the marker must start at column 0, and nothing may follow the value — are now in the runbook source, along with the closed-unmerged behaviour and the post-merge wait. Those were genuinely missing, not merely unpublished.

The ordering is still backwards, as you say: the mechanism lands before its documentation. If #2192 slips, a short pointer here is the right correction.

.github/workflows/build_pr.yml
lubomirw: Restructured in 005148892, close to your shape:

```
# Strict extractors, loose probes.
#
# Strict: line start, nothing after the value. Trailing junk is
# rejected, not truncated - `Interfaces-Version: 2.20.0; rm -rf /`
# does not become a pin on 2.20.0.
#
# Loose: a bulleted, indented, bold or blockquoted marker misses the
# extractor, so the probe has to catch it and fail loudly.
#
# The Depends-On probe closes off the repository name, so a marker for
# another repo - or a sibling starting with the same word - is ignored.
```

No fact dropped. The other two sites you name are shorter and I left them.

.github/workflows/sonar.yaml
lubomirw: Fixed in 005148892 — the label sits above `- name: SonarCloud Scan` again, and the `continue-on-error` note stays with the download step it explains.

.github/workflows/sonar.yaml
lubomirw: Renamed to `check` in 005148892, with the output reference updated. It also gained a `name:`, so it has a readable label in the Actions UI rather than none.

.github/workflows/sonar.yaml
lubomirw: Added in 005148892: `contents: read` and `actions: read`, which is what the jobs-list read needs and nothing more.

Worth noting it is not only a `workflow_run` default — this repository's default workflow permission is `read`, so the token was not write-capable in practice. The reason to declare it is the one you give: this is the first place in the file where a scope can be stated, and leaving it implicit makes the job depend on a repository setting nothing in the file mentions.

.github/workflows/build_pr.yml
lubomirw: Both fixed in 005148892.

`retention-days: 1` is gone, so the artifact now matches the other Sonar inputs. And the silent path is closed: the scan keys on the download step's `outcome` and fails with a message naming the cause, rather than proceeding with no override.

The second half of your finding is the one I had not seen. Reading `mvnarg.txt` from the tree root meant a failed download fell through to whatever the head commit contains, so the "use the pom value" fallback the comment claimed was really "use whatever the branch supplies". The download now lands in `$RUNNER_TEMP/mvnarg`, outside the checkout, so the value can only come from the producing job.

.github/workflows/sonar.yaml
lubomirw: Fixed together with the retention finding — the artifact now lands in `$RUNNER_TEMP/mvnarg` rather than the tree root, and the scan fails when the download did not succeed instead of reading whatever is there.

You are right that this was not a new capability given what the file already accepts about fork PRs and `SONAR_TOKEN`. It was the difference between an accidental fallback and a selectable one, which is worth removing on its own.

.github/workflows/sonar.yaml
lubomirw: Rewritten. That bullet described the behaviour before the gate was added and I did not update it.

Both bullets in that section were stale, in fact — the merge-window one too, since the resolver now holds the pin until the merge commit's immutable `-M` build is published. The section now lists what is actually still out of scope, and a new "What a coupled pull request sees" section states the current picture: `Interfaces pin` red, `Build` red behind it, the arch contexts non-green, and Sonar reporting throughout with the pin as the sole merge block.

.github/workflows/build_pr.yml
lubomirw: Added in 005148892 as `InterfacesPinGuardTest`, beside the precedents you name.

It asserts both invariants — `pin-gate` inside `build.needs`, and `$MVNARG` on every job running a phase that compiles the tree — by parsing the workflow with snakeyaml, the same way `WireMockPortsGuardTest` reads the test yml.

A third assertion pins the set of jobs the detection matches (`compile`, `test`, `package`, `generated`). Without it a detection that quietly stopped matching would leave the override check iterating an empty set and passing, which is the `transferredClassesMustExist` idea from `CiTestSplitTest`.

Verified by breaking both invariants and re-running: 3 tests, 2 failures, one per invariant, each with its intended message. `formatting` is correctly excluded — `spotless:check checkstyle:check` names plugin goals and no lifecycle phase.

.github/workflows/build_pr.yml
lubomirw: Named in 005148892 — "Upload the resolved interfaces argument", matching the five sibling uploads.

.github/workflows/build_pr.yml
lubomirw: Reconsidered — a section now exists in the README (46d40f869), so the mechanism no longer ships ahead of any in-repo statement of it.

It gives both forms, the column-0 and end-of-line rules, the `publish-snapshot` prerequisite, the fenced-block exclusion, the fork limitation, and the "re-run all jobs, not the failed one" step, with the self-clearing and delete-before-merge distinction between the two markers.

Deliberately no pointer to the fuller runbook: that lands in this repository later under its own phase, and a link to something not yet here would read as a dead reference. It also stays out of the PR-template trap you flagged — the examples sit inside fenced blocks, and the resolver strips those before matching, so the README cannot pin anything if it is pasted into a body.

.github/workflows/build_pr.yml
lubomirw: Correcting what I told you here: the test asserted two of the three lines the merge block needs, not both invariants as I claimed. Deleting `build`'s `if: ${{ !cancelled() }}` left all three tests green — the aggregate is then skipped when the gate fails, and a skipped required check counts as passing. That is the bug #2196 fixed, reachable again by an innocent simplification.

Fixed in 6019dc4cf. Two assertions added for the other legs, and a Maven call is now recognised in any command position, so `cd x && mvn` and `integration-test` no longer slip past — the old detection required the line to start with `mvn `.

Re-verified by mutation. Five ways to break the block, five reds:

```
remove build's if: !cancelled()            Failures: 1
gut the needs scan (toJSON removed)        Failures: 1
new job: cd . && mvn verify, no MVNARG     Failures: 2
new job: mvn integration-test, no MVNARG   Failures: 2
drop pin-gate from build.needs             Failures: 1
```

Three of those five passed before. The javadoc carried the same overclaim — it opened with "the two list memberships" — and now says what the test actually pins.

.github/workflows/build_pr.yml
lubomirw: Not filing this after all, and the reason is close to the opposite of the one I gave you.

I justified it on the defect rate: three marker-parsing bugs shipped in two days, all of them the silent-ignore kind. But those came from the parser being under construction. Once this merges it becomes plumbing nobody touches until the marker protocol itself changes.

Against that, the cost is ~250 lines, the first `ProcessBuilder` test in `src/test`, and `@EnabledOnOs(LINUX)` — the script needs `python3` and GNU `grep -P`, so it would skip on every developer machine here and run only in CI. It would guard a window that is closing.

Two things I checked while deciding, in case either is useful elsewhere. Translating the patterns into Java is not an option: `java.util.regex` reads `[[:space:]]` as a class of the letters `:space`, so `^Interfaces-Version[[:space:]]*:` returns false for `Interfaces-Version :` and true for `Interfaces-Versionsss:` — inverted, silently. And extracting the script to `.github/scripts/` to use the org's `*.test.sh` idiom would add a checkout to a job deliberately built without one, plus a new job that must sit in `build.needs` unfiltered, since the aggregate scan counts a skipped upstream as failure.

Your point stands on its merits. I am judging the ratio, not the finding.

ivosh (approved):

PR #2239, opened 2026-09-08
Require External Account Binding on ACME profiles that configure a key

### LINKED ISSUE #2230: Configure External Account Binding credentials on ACME profiles
### Description

An ACME profile can carry External Account Binding credentials, and a profile that carries at least one requires every new account to present one. This controls who may register an account; it does not change how identifiers are authorized, so http-01 and dns-01 continue to run until the pre-authorized identifier work lands.

- `acme_profile` gains `eab_secret_uuids`, a UUID array of Secrets holding HMAC keys, following the `groupUuids` precedent on `ProtocolCertificateAssociations`.
- The EAB key identifier is the Secret's UUID. The operator creates the Secret themselves in the vault profile of their choice; the platform only links it.
- A stateless endpoint generates a suitable key value, base64url of 32 random bytes, and returns it without persisting it, so operators do not have to invent one. The client passes that value verbatim as its EAB HMAC key and the server base64url-decodes the Secret content to obtain the MAC key.
- The directory response reports `externalAccountRequired` as true whenever the profile has at least one linked Secret. Nothing is stored: the requirement is derived from the array, so there is no separate flag to keep consistent with it.
- `newAccount` then requires a binding whose kid is one of the linked Secrets, verifying the HS256 MAC per RFC 8555 section 7.3.4, including that the protected header names this newAccount URL, carries the account key as its payload, and does not carry a nonce.
- `acme_account` records the Secret it was created with, so a leaked credential can be traced to its accounts.
- Deleting a Secret linked to an ACME profile is refused, naming the profiles, because the last removal would silently reopen the profile to unauthenticated registration. Disabling a Secret is allowed and stops it working at once.
- Both the profile-based and the RA-profile-based ACME endpoints.

Two consequences worth stating up front. The ACME endpoint is unauthenticated, so reading the Secret needs either an internal accessor or a permission grant for the protocol user, which is work inside the Secrets module. And there is no lockout on repeated wrong bindings, since a Secret has no attempt counter; a 256-bit key makes that acceptable, and it removes the denial of service the per-registration challenge accepted.

### Acceptance Criteria

- [ ] A profile with no linked Secret behaves exactly as today.
- [ ] A profile with a linked Secret advertises `externalAccountRequired`, and `newAccount` without a binding answers `externalAccountRequired`.
- [ ] A valid binding creates an account that records the Secret it used.
- [ ] Unknown kid, a kid not linked to this profile, a wrong MAC, a nonce in the protected header, a binding for another URL or another account key, and a disabled Secret all answer one generic `unauthorized` problem.
- [ ] A vault or Secret read failure answers `serverInternal`, not `unauthorized`.
- [ ] The generator returns a distinct value per call and persists nothing.
- [ ] Deleting a linked Secret is refused and names the profiles.
- [ ] The vault is read outside any transaction.
- [ ] Sonar quality gate passed, >= 80% coverage on new code.

klaraf755:
Depends-On: OmniTrustILM/interfaces#960

## Problem

Any client that can reach an ACME profile's directory can register an account against it. There is no way for an operator to require that a client already hold a credential the platform issued, so enrolment through ACME cannot be restricted to a known population.

## Change

An ACME profile carries `eab_secret_uuids`, a list of the secrets whose content is an HMAC key a newAccount binding may verify under (RFC 8555 §7.3.4). A non-empty list is what makes the directory meta advertise `externalAccountRequired` — the requirement is derived, so a profile can never advertise it with nothing to verify against.

- `AcmeEabVerifier` resolves the binding's `kid` to one of the profile's secrets, checks the URL and the account key, and verifies the MAC. It runs `Propagation.NOT_SUPPORTED`, because reading a key goes to the vault over HTTP and must not be held inside the newAccount transaction.
- Every failed binding gets the same `unauthorized` — which check failed would tell a caller which key identifiers exist. An unreadable or unusable key is `serverInternal`; a missing binding is `externalAccountRequired`. A binding carrying a nonce is refused: §7.3.4 does not list one, and the enclosing newAccount JWS already carries the replay protection.
- The key is read as the `acme` system user every ACME request already runs as, granted `SECRET:GET_SECRET_CONTENT` by `V202609101100`. Registering a secret on a profile requires, on that secret, both permissions the read is gated on — `SECRET:GET_SECRET_CONTENT` and `VAULT_PROFILE:MEMBERS` on the vault profile it is sourced from — so an operator can only delegate a read they could perform themselves. Checking only the first would let someone who cannot read a secret still put it in charge of a profile.
- `POST /v1/acmeProfiles/eabKeys` returns a fresh 256-bit key, base64url. Nothing is stored or associated; the operator creates the secret from it.
- Deleting a secret an ACME profile uses as a binding key is refused — the profile would keep advertising `externalAccountRequired` against a key that no longer exists. Disabling it stays allowed.
- The account records which secret it was admitted with (`acme_account.eab_secret_uuid`).

Four unrelated ACME defects found along the way: the terms-of-service check rejected agreement instead of its absence, and now also requires the profile to carry a terms URL; a newAccount without `contact` hit an NPE when the profile required contacts; an omitted `contact` was serialized as null rather than an empty list; and `editAcmeProfile` guarded the new-order switch against a null request field and then overwrote it unconditionally seventy lines later, so any edit omitting `termsOfServiceChangeDisable` wrote `disable_new_orders = NULL` and every later ACME request on that profile failed unboxing it. The overwrite is gone and the read is null-safe, so rows already carrying NULL recover.

One change reaches beyond ACME and wants a reviewer's eye: `ConnectorRequestAttributesBuilder.resolveContent` now calls `loadFullCredentialData` only when an attribute actually names a credential. That loader authorizes `CREDENTIAL:DETAIL` at method entry, so calling it unconditionally charged every caller a permission for work it would then skip — which would have meant granting the `acme` role a credential permission it has no use for. The authority-instance, entity-instance and callback paths share this skeleton; behaviour there is unchanged except that a caller referencing no credential is no longer denied. `AttributeReferenceExpanderFenceArchTest` exempts this class by name, so the fence is untouched.

## Tests

`AcmeEabVerifierTest` covers the verification outcomes, the indistinguishable rejection, and the internal-error classifications. `ExternalAccountBindingJwsTest` covers parsing, MAC verification and key decoding. `AcmeEabAccountITest` covers the newAccount wiring end to end: what the directory advertises, and that a refused binding leaves no account behind. `AcmeProfileServiceITest` gains the secret-resolution, derived-requirement, generator and terms-URL cases, and `ConnectorRequestAttributesBuilderTest` the credential-skip.

The role grants in `V202609101100` are not covered by any test — `BaseSpringBootTest` mocks `OpaClient` permissively, so no integration test can observe a missing permission. They were verified by hand against a running auth service, registering an account with acme.sh and issuing a certificate through http-01.

Closes #2230


src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -1010,11 +1030,41 @@ private AcmeAccount addNewAccount(String profileName, String publicKey, NewAccou
         account.setDefaultRaProfile(!isRaProfileBased);
         account.setAccountId(accountId);
         account.setContact(SerializationUtil.serialize(accountRequest.getContact()));
```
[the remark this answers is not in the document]
klaraf755: Confirmed: `SerializationUtil.serialize(null)` produces the literal `null`, which `deserializeArrayString` reads back as a null list. An omitted contact is now serialized as an empty list.

The fix survived the branch being rewritten for profile-level External Account Binding, and `AcmeEabAccountITest` exercises the newAccount path that reaches it.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -846,6 +869,7 @@ private ClientCertificateDataResponseDto registerPlatformLevel(RaProfile raProfi
         ClientCertificateDataResponseDto response = new ClientCertificateDataResponseDto();
         response.setUuid(certificate.getUuid().toString());
         response.setCertificateData("");
+        attachEabCredential(response, generatedEabKey);
```
ivosh: **~ Minor** | Correctness

This PR wires `generatedEabKey` into both register paths, but only guards one of them. In `registerCertificate` the PR wrapped the whole post-acceptance tail in a new `try` (ClientOperationServiceImpl.java:750-787) whose catch calls `discardUndeliveredEabCredential(certificate, generatedEabKey)` at line 785 — including the `stateMachine.transition(certificate, CertificateState.REGISTERED)` at line 768.

The peer `registerPlatformLevel` runs the same steps unguarded: `generatedEabKey` is assigned at line 848 (inside the try), but the tail at lines 863-867 — `stateMachine.transition(certificate, CertificateState.REGISTERED)` then `fireRegistrationEventIfChallengeProtected` — sits *outside* any try, and `attachEabCredential` at line 872 is only reached if that tail succeeds. `fireRegistrationEventIfChallengeProtected` swallows its own `RuntimeException` (line 1221), but a failure in the transition at line 863 propagates out with the challenge already stored and committed (`registerCertificate` is `Propagation.NOT_SUPPORTED`, line 618) and the credential never returned — exactly the state `discardUndeliveredEabCredential` was written to prevent, described in its own Javadoc at lines 1107-1112.

No test covers it either: `undeliverableGeneratedCredentialIsRemovedWhenTheConnectorAcceptedButALocalStepFailed` (ClientOperationServiceRegisterITest.java) stubs `registeringAdapter().register(...)` and so only exercises the connector-backed path.

_Suggested:_ Extend `registerPlatformLevel`'s try to cover the REGISTERED transition at line 863 (or wrap lines 863-867 in the same catch) so `discardUndeliveredEabCredential` runs there too, and add the matching platform-level test.

<sub>flagged by Maintainability & tests + Correctness & races</sub>

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -1009,12 +1029,44 @@ private AcmeAccount addNewAccount(String profileName, String publicKey, NewAccou
...
+    /**
+     * Flushes the bound account so a second binding of the same registration racing past the binder's existence check
+     * fails here, on the unique constraint, with the same generic rejection; the rollback rule on {@code newAccount}
+     * turns the poisoned transaction into a plain rollback.
+     */
+    private void saveBoundAccount(AcmeAccount account) throws AcmeProblemDocumentException {
+        try {
+            acmeAccountRepository.saveAndFlush(account);
+        } catch (DataIntegrityViolationException e) {
+            if (!isRegistrationBindingCollision(e)) {
```
ivosh: **~ Minor** | Testing

`saveBoundAccount` (AcmeServiceImpl.java:1050-1061) is the only defence against two newAccount requests racing past the binder's `existsByRegistrationCertificateUuid` pre-check, and its whole behaviour hinges on `isRegistrationBindingCollision` at AcmeServiceImpl.java:1054 matching `CryptoAssetConstraintTranslator.constraintNameOf(e)` against `uq_acme_account_registration_certificate`. No test reaches that branch: `AcmeRegistrationBindingITest.secondAccountOnTheSameRegistrationIsRejectedGenerically` (line 279) is stopped by the pre-check in `AcmeRegistrationBinder`, and `oneAccountPerRegistrationIsEnforcedByTheSchema` (line 297) asserts only that a duplicate `saveAndFlush` raises `DataIntegrityViolationException` — it never asserts the exception carries a constraint name the filter recognises. If the name Hibernate surfaces ever differs from the constant, the `if (!isRegistrationBindingCollision(e)) throw e;` arm rethrows and the losing racer gets a 500 SERVER_INTERNAL instead of the generic rejection, with nothing failing.

_Suggested:_ Extend oneAccountPerRegistrationIsEnforcedByTheSchema to assert CryptoAssetConstraintTranslator.constraintNameOf(thrown) equals AcmeAccount.UNIQUE_REGISTRATION_CERTIFICATE_CONSTRAINT, pinning the name the race branch matches on.

<sub>flagged by Correctness & races + Transaction & concurrency</sub>

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -1000,6 +1015,11 @@ private AcmeAccount addNewAccount(String profileName, String publicKey, NewAccou
         } else {
             return oldAccount;
```
ivosh: **? Question** | Security — Nitpick if confirmed

AcmeServiceImpl.java:1016 returns the existing account for a known public key before the registration-mode branch at :1020 ever runs. `AcmeProfileServiceImpl.editAcmeProfile` lets an operator move a live profile to CERTIFICATE_REGISTRATION (the new `editKeepsTheStoredChallengeSourceWhenOmittedAndReplacesItWhenGiven` test pins exactly that), so every account created while the profile was PROTOCOL_DEFAULT keeps being admitted with no External Account Binding and keeps `registration_certificate_uuid` NULL, even though the directory now advertises `externalAccountRequired`. Returning an existing account for a known key is RFC 8555 section 7.3 behaviour, so this may be deliberate grandfathering — but nothing in the code or the tests states it.

_Suggested:_ Decide whether a profile switching into CERTIFICATE_REGISTRATION should refuse pre-existing unbound accounts (reject the edit while such accounts exist, or deny them at newAccount) and, if grandfathering is intended, record that on AcmeProfile.isRegistrationMode with a one-line comment.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -1009,12 +1029,44 @@ private AcmeAccount addNewAccount(String profileName, String publicKey, NewAccou
...
+     * Flushes the bound account so a second binding of the same registration racing past the binder's existence check
+     * fails here, on the unique constraint, with the same generic rejection; the rollback rule on {@code newAccount}
+     * turns the poisoned transaction into a plain rollback.
+     */
+    private void saveBoundAccount(AcmeAccount account) throws AcmeProblemDocumentException {
+        try {
+            acmeAccountRepository.saveAndFlush(account);
+        } catch (DataIntegrityViolationException e) {
+            if (!isRegistrationBindingCollision(e)) {
+                throw e;
```
ivosh: **. Nitpick** | Error handling

AcmeServiceImpl.java:1053-1055 translates only `uq_acme_account_registration_certificate`; every other integrity violation is rethrown raw. V202609081100 also adds `fk_acme_account_registration_certificate`, and AcmeRegistrationBinder.java:85-86 reads the certificate well before the flush at AcmeServiceImpl.java:1052 commits, in the same READ COMMITTED transaction. An operator deleting the pre-registered certificate in that window makes the insert fail on the foreign key, `isRegistrationBindingCollision` returns false, and the DataIntegrityViolationException propagates — the ACME client gets a 500 instead of the single generic problem document that every other registration-mode rejection produces.

_Suggested:_ Treat fk_acme_account_registration_certificate the same as the unique constraint in isRegistrationBindingCollision so a registration that vanished mid-request also answers with REGISTRATION_REJECTION.

src/test/java/com/otilm/core/integration/service/acme/AcmeRegistrationBindingITest.java
```diff
@@ -0,0 +1,387 @@
...
+        assertEquals(0, failedAttempts(registration), "a replay against a bound registration reaches no gate");
+    }
+
+    @Test
+    void oneAccountPerRegistrationIsEnforcedByTheSchema() throws Exception {
+        Certificate registration = seedRegistration();
+        acmeAccountRepository.saveAndFlush(boundAccountRow(registration));
+        AcmeAccount duplicate = boundAccountRow(registration);
+
+        assertThrows(DataIntegrityViolationException.class, () -> acmeAccountRepository.saveAndFlush(duplicate));
```
ivosh: **. Nitpick** | Testing

`oneAccountPerRegistrationIsEnforcedByTheSchema` (AcmeRegistrationBindingITest.java:297-303) is the only test in this PR that actually provokes the unique violation, and its single assertion at line 302 checks nothing but `DataIntegrityViolationException.class`. That type is produced by any constraint on `acme_account`, so the test is green whichever constraint fired.

The branch that depends on the *name* is `AcmeServiceImpl.isRegistrationBindingCollision` (AcmeServiceImpl.java:1063-1067), which requires `CryptoAssetConstraintTranslator.constraintNameOf(e)` to return exactly `uq_acme_account_registration_certificate`. Nothing exercises it: `secondAccountOnTheSameRegistrationIsRejectedGenerically` (line 279) is short-circuited by the binder's `existsByRegistrationCertificateUuid` pre-check and never reaches `saveBoundAccount`, and this test writes rows straight through the repository. If the name comes back schema-qualified, differently cased, or absent, `saveBoundAccount` (AcmeServiceImpl.java:1050-1060) rethrows the raw `DataIntegrityViolationException` onto the ACME wire instead of the generic rejection, and the `@Transactional(rollbackOn = ...)` rule added at AcmeServiceImpl.java:289 protects a fallback that never fires.

The sibling suite already pins this exactly: `CryptoAssetInventoryITest.java:419` and `:728` assert `constraintNameOf(failure)).contains("...")`, the latter with `describedAs("the refusal must come from the name bound, not from another constraint or from the JPA layer")`.

_Suggested:_ Capture the thrown exception in `oneAccountPerRegistrationIsEnforcedByTheSchema` and assert `CryptoAssetConstraintTranslator.constraintNameOf(ex)` contains `AcmeAccount.UNIQUE_REGISTRATION_CERTIFICATE_CONSTRAINT`, the way CryptoAssetInventoryITest:728 does.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1038,12 +1072,24 @@ private int registrationWindowDays() {
...
+     */
+    private String maybeCreateRegistrationAuthorization(Certificate certificate,
             ClientCertificateRegistrationDto request) {
-        String secret = request.getAuthorizationSecret();
-        if (secret == null || secret.isBlank()) {
-            return;
+        if (!hasChallenge(request)) {
+            return null;
         }
+        String generated = request.isGenerateEabCredential() ? AcmeEabCredential.generateChallenge() : null;
```
ivosh: **. Nitpick** | Maintainability

ClientOperationServiceImpl.java:1091 calls `AcmeEabCredential.generateChallenge()` from `maybeCreateRegistrationAuthorization`, the generic register path, and stores the result as the registration challenge on `CertificateRegistrationAuthorization` — the same row a CMP `senderKID` enrolment resolves against. The value itself carries nothing ACME-specific: `AcmeEabCredential.generateChallenge()` is `RandomUtil.generateRandomNonceBase64Url(32)`.

This PR made the opposite call for the neighbouring rule: the UTF-8 MAC-key derivation was deliberately hoisted out of `CmpConfigurationContext` into `RegistrationResolver.macKey` (RegistrationResolver.java:38) precisely so one shared home owns "the rule every protocol shares". The challenge *generator* is the one piece of that pair left behind in `com.otilm.core.service.acme.registration`, which is why a v2 client-operations service now imports an ACME class (ClientOperationServiceImpl.java:105). Only `toEabHmacKey` is genuinely ACME-shaped.

_Suggested:_ Move `generateChallenge()` next to `RegistrationResolver.macKey` in `com.otilm.core.service.registration` and leave only the ACME-specific `toEabHmacKey` in `AcmeEabCredential`.

src/test/java/com/otilm/core/integration/service/AcmeProfileServiceITest.java
```diff
@@ -224,6 +225,64 @@ void testEditAcmeProfile() throws ConnectorException, AttributeException, NotFou
...
+        Assertions.assertEquals(ProtocolChallengeSource.CERTIFICATE_REGISTRATION, dto.getChallengeSource());
+        Assertions
+                .assertTrue(acmeProfileRepository
+                        .findByUuid(UUID.fromString(dto.getUuid()))
+                        .orElseThrow()
+                        .isRegistrationMode());
+    }
+
+    @Test
+    void requiringTermsAgreementNeedsATermsUrlOnCreateAndEdit() throws Exception {
```
ivosh: **. Nitpick** | Testing

`requiringTermsAgreementNeedsATermsUrlOnCreateAndEdit` (AcmeProfileServiceITest.java:255-269) runs three act/assert cycles: create-without-URL rejected, create-with-URL accepted, edit-without-URL rejected. The name announces the bundling (`...OnCreateAndEdit`), and a failure in the first cycle hides the other two.

Two more do the same: `editKeepsTheStoredChallengeSourceWhenOmittedAndReplacesItWhenGiven` (AcmeProfileServiceITest.java:272-284, two edit calls asserting two different rules) and `termsOfServiceRequirementAdmitsOnlyTheClientThatAgreed` (AcmeRegistrationBindingITest.java:306-324, a rejected newAccount then an accepted one).

The surrounding class is consistently one behaviour per method with a `_case` suffix — `testAddAcmeProfile` / `testAddAcmeProfile_validationFail` / `testAddAcmeProfile_alreadyExist` (lines 161, 189, 195), `testEditAcmeProfile` / `testEditAcmeProfile_validationFail` (203, 300) — so these three are the outliers, and they are exactly the shape the repo's `decompose-eager-tests` skill targets (multiple act/assert cycles, a `verbAndOtherVerb` name).

_Suggested:_ Split each of the three into one test per behaviour, keeping the shared fixture in a private helper.

src/main/java/com/otilm/core/service/registration/RegistrationResolver.java
```diff
@@ -0,0 +1,92 @@
...
+    public static byte[] macKey(String challenge) {
+        return challenge.getBytes(StandardCharsets.UTF_8);
+    }
+
+    /**
+     * @param registrationUuid the referenced certificate, or null when the reference did not parse
+     * @param eligible the protocol's rule for which certificate of the RA profile admits the operation (state,
+     * archival)
+     */
+    public Certificate resolveEligible(RaProfile raProfile, UUID registrationUuid, Predicate<Certificate> eligible)
```
ivosh: **. Nitpick** | Documentation

`RegistrationResolver.resolveEligible` (RegistrationResolver.java:47) is one of the two public methods of the seam this PR introduces to be shared by CMP and ACME, yet its Javadoc (lines 42-46) is two `@param` tags and nothing else: no summary sentence saying what it returns, and no `@throws` for the checked `RegistrationRejectedException` it declares — even though the two rejection reasons it produces (lines 50 and 55-56) are what each protocol turns into its own generic wire message.

Its own sibling one line below, `verifyMac` (lines 61-66), opens with a summary; so does the collaborator it delegates to, `RegistrationChallengeGate.verify`, which carries a summary plus `@return`. `resolveEligible` is the outlier of the three. (`ExternalAccountBindingJws.parse`, ExternalAccountBindingJws.java:52, has the same shape — a Javadoc that starts at `@return`.)

_Suggested:_ Add a one-line summary and an `@throws RegistrationRejectedException` clause to `resolveEligible`, matching the shape of `verifyMac` just below it.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -274,6 +284,9 @@ public ResponseEntity<?> getNonce(String acmeProfileName, Boolean isHead, URI re
 
     @Override
     @ProtocolEndpoint
+    // A rejected registration binding is a checked problem exception thrown after a failed flush; the rollback rule
+    // turns the commit of the poisoned transaction into a plain rollback.
+    @Transactional(rollbackOn = AcmeProblemDocumentException.class)
```
ivosh: **. Nitpick** | Maintainability

AcmeServiceImpl.java:289 adds `@Transactional(rollbackOn = AcmeProblemDocumentException.class)` using the class's jakarta annotation (import at line 89). The sibling protocol endpoint two hundred lines down, `validateChallenge` at AcmeServiceImpl.java:628-631, carries an explicit comment for the opposite choice: "Spring's annotation rather than the class's jakarta one: the transaction-boundary rules read Spring's, and the rollback attribute is what those rules require of a method declaring a checked exception." That reason applies verbatim here — this is also a protocol endpoint declaring a checked exception and setting a rollback attribute. It is concrete, not cosmetic: `TransactionalBoundaryArchTest.effectivePropagation` (src/test/java/com/otilm/core/architecture/TransactionalBoundaryArchTest.java:172-180) resolves only `org.springframework.transaction.annotation.Transactional`, so it reads `newAccount` as having no transactional boundary at all and Rule A's guard does not cover the new `acmeAccountRepository.saveAndFlush` call site at AcmeServiceImpl.java:1052. Runtime behaviour is equivalent (Spring's JTA parser maps `rollbackOn` onto a rollback rule over the unchecked default), so this is about the boundary staying visible to the rules that police it.

_Suggested:_ Declare newAccount with org.springframework.transaction.annotation.Transactional(rollbackFor = AcmeProblemDocumentException.class), as validateChallenge at AcmeServiceImpl.java:630 does.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1055,6 +1101,41 @@ private void maybeCreateRegistrationAuthorization(Certificate certificate,
...
+            logger
+                    .warn("Failed to record the undelivered EAB credential for certificate {}: {}",
+                            certificate.getUuid(), e.getMessage());
+        }
+    }
+
+    private static void attachEabCredential(ClientCertificateDataResponseDto response, String generatedChallenge) {
+        if (generatedChallenge != null) {
+            response.setEabKid(response.getUuid());
+            response.setEabHmacKey(AcmeEabCredential.toEabHmacKey(generatedChallenge));
```
ivosh: **~ Minor** | Security

`attachEabCredential` at ClientOperationServiceImpl.java:1137 hands the caller `base64url(UTF-8(challenge))` as `eabHmacKey`. But a `generateEabCredential` registration creates an ordinary authorization row, so `fireRegistrationEventIfChallengeProtected` (ClientOperationServiceImpl.java:770 for the connector path, :867 for the platform path — guarded only by `existsByCertificateUuid`) also fires CERTIFICATE_REGISTERED, and CertificateRegisteredEventHandler.java:46 puts `registrationChallengeStore.resolvePlaintext(authorization)` — the raw challenge text — into the event data that the handler's own comment says "rides the event-data object to the external provider".

Two consequences this PR does not disclose. (1) The contract text this PR relies on (interfaces#960, `ClientCertificateDataResponseDto.eabHmacKey`: "returned exactly once; the platform … cannot show it again") is not true — the same secret is exported a second time, to whatever external notification provider is configured. (2) What the notification carries is the SCEP/CMP form of the credential (the challenge text, what `ScepServiceImpl:990` and `RegistrationResolver.macKey` consume), not the `--eab-hmac-key` form built at :1137; an operator who forwards the notified value to a device gets a MAC mismatch that spends a gate attempt against the lockout counter.

If shipping the credential on that event is intended for the ACME case too, it needs the `toEabHmacKey` encoding and a note that "once" means once per delivery channel; if it is not intended, the event should skip the credential for a generated-EAB registration.

_Suggested:_ Encode the credential with `AcmeEabCredential.toEabHmacKey` before `CertificateRegisteredEventHandler` sets it for a generated-EAB registration, or suppress it on that event and correct the "shown only once" wording.

<sub>flagged by Secrets & sensitive data + Security & info-leak</sub>

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -1000,6 +1015,11 @@ private AcmeAccount addNewAccount(String profileName, String publicKey, NewAccou
         } else {
             return oldAccount;
         }
+        // In registration mode the binding is the account's admission: resolved and MAC-verified before any row
+        // exists, so a rejected binding leaves nothing behind.
+        Certificate boundRegistration = acmeProfile.isRegistrationMode()
```
ivosh: **? Question** | Security — Minor if confirmed

AcmeServiceImpl.java:1020 resolves and MAC-verifies the EAB only while the account row is being created, and `registrationCertificateUuid` is read afterwards by nothing except `AcmeAccountRepository.existsByRegistrationCertificateUuid` (src/main/java/com/otilm/core/dao/repository/acme/AcmeAccountRepository.java:23) — `createCert` (AcmeServiceImpl.java:1429) still calls `issueCertificate`, an ordinary fresh issuance. So the pre-registration's guard rails bound admission and nothing after it: once bound, the account keeps ordering certificates after the authorization's `expiresAt` passes, after the gate moves it to LOCKED, and after the pre-registered certificate is deleted — `ON DELETE SET NULL` (src/main/resources/db/migration/V202609081100__acme_account_registration_binding.sql:10-11) silently unbinds the row and leaves the account VALID and enabled. Flipping an existing profile to CERTIFICATE_REGISTRATION is likewise not retroactive: accounts already created on it keep working, and a newAccount replayed with an existing key returns at AcmeServiceImpl.java:1016, before the binder runs. If this is the intended scope (issue #2230 lists only the binding), the exposure still needs writing down, because an operator who sees the directory advertise `externalAccountRequired` will read registration mode as gating issuance; if orders were meant to be scoped to the bound registration, that scoping is missing.

_Suggested:_ Add a one-line note on `AcmeProfile.challengeSource` and in the registration-mode operator notes stating that the EAB gates account creation only, so the issuance window, the lockout and deleting the pre-registered certificate do not revoke an already-bound account.

src/main/java/com/otilm/core/service/acme/registration/ExternalAccountBindingJws.java
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public static ExternalAccountBindingJws parse(ExternalAccountBinding binding) {
+        if (binding == null || binding.getProtectedHeader() == null || binding.getPayload() == null
+                || binding.getSignature() == null) {
+            return null;
+        }
+        try {
+            JWSObject jws = new JWSObject(new Base64URL(binding.getProtectedHeader()),
+                    new Base64URL(binding.getPayload()), new Base64URL(binding.getSignature()));
+            if (!JWSAlgorithm.HS256.equals(jws.getHeader().getAlgorithm())) {
```
ivosh, line 42: **~ Minor** | Correctness

RFC 8555 section 7.3.4 lists four criteria for the External Account Binding protected header: `alg` MUST be MAC-based, `kid` MUST be present, `url` MUST equal the outer JWS URL, and **the `nonce` field MUST NOT be present**. `ExternalAccountBindingJws.parse` (ExternalAccountBindingJws.java:35-50) enforces the first via the HS256 check on line 43, and `registrationUuid()`/`isForUrl()` cover the second and third — but nothing rejects a header that carries `nonce`, so a binding that violates the MUST NOT is accepted and MAC-verified.

The established neighbour pattern is in the same protocol implementation this PR extends: `AcmeServiceImpl.validateKey` (AcmeServiceImpl.java:1126-1129) implements the analogous RFC 8555 section 7.3.5 rule for the key-change inner JWS and rejects it with `Problem.MALFORMED` when `jwsInnerObject.getHeader().toJSONObject().containsKey("nonce")`. The new section 7.3.4 parser should hold the same line for the same reason.

_Suggested:_ Return null from `ExternalAccountBindingJws.parse` when `jws.getHeader().toJSONObject().containsKey("nonce")`, mirroring the section 7.3.5 check in `AcmeServiceImpl.validateKey`, and pin it with a case in `ExternalAccountBindingJwsTest.structurallyInvalidBindingsDoNotParse`.

<sub>flagged by Protocol / PKI semantics</sub>

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -971,16 +984,18 @@ private AcmeAccount addNewAccount(String profileName, String publicKey, NewAccou
         String accountId = AcmeRandomGeneratorAndValidator.generateRandomId();
         AcmeAccount oldAccount = acmeAccountRepository.findByPublicKey(publicKey);
         if (acmeProfile.isRequireContact() != null && acmeProfile.isRequireContact()
-                && accountRequest.getContact().isEmpty()) {
+                && (accountRequest.getContact() == null || accountRequest.getContact().isEmpty())) {
```
ivosh, line 1013: **~ Minor** | Security

AcmeServiceImpl.java:988 (`logger.error("Contact not found for Account: {}", accountRequest)`) and :999 (`logger.error("Terms of Service not agreed for the new Account: {}", accountRequest)`) print the whole `NewAccountRequest`; :303 does the same at DEBUG. With the contract this PR depends on (interfaces#960), `NewAccountRequest.toString()` appends `externalAccountBinding` and `ExternalAccountBinding` is a Lombok `@Data`, so all three lines now emit the client's EAB protected header, payload and HMAC signature — and this PR is what makes clients send one.

The signature is an HMAC-SHA256 tag over known plaintext under the registration challenge, i.e. an offline-guessing oracle for that challenge. For a platform-generated challenge (32 SecureRandom bytes, `AcmeEabCredential.RANDOM_BYTES`) that is harmless. But the same binder path verifies against an operator-supplied `authorizationSecret` too (AcmeRegistrationBinder.java:90 → the shared gate), and such a secret only has to clear the HS256 256-bit minimum — a ≥32-character passphrase qualifies, and it is the same secret that gates issue/renew/rekey over the platform API, SCEP and CMP. Both ERROR lines are reachable by an unauthenticated client on a requireContact / requireTermsOfService profile, before the binding is checked at all.

Note the asymmetry with the same contract PR, which deliberately kept `eabHmacKey` out of `ClientCertificateDataResponseDto.toString()` while adding `externalAccountBinding` to `NewAccountRequest.toString()`. If the binding in the log is wanted for diagnostics, say so in a comment; otherwise the rejection lines only need the contact list and the account key thumbprint.

_Suggested:_ Log the rejection with the contact list and account key thumbprint instead of the whole `accountRequest`, or exclude `externalAccountBinding` from `NewAccountRequest.toString()` the way `eabHmacKey` is excluded from the response DTO's.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
ivosh: **i Info** | Security

AcmeServiceImpl.java:1055 rethrows the raw `DataIntegrityViolationException` from a protocol endpoint, which reads like a CLAUDE.md "Don't leak runtime details to the wire" violation — those messages routinely carry SQL fragments and table/column identifiers. Checked the exit path: `AcmeExceptionHandlingAdvice.handleOtherException` (AcmeExceptionHandlingAdvice.java:39) catches every non-`AcmeProblemDocumentException` from `com.otilm.core.api.acme`, logs it server-side and answers a bare `Problem.SERVER_INTERNAL` document with no detail. The rethrow is safe as written; recording it here only so the next reader does not have to re-derive it.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -743,31 +747,49 @@ public ClientCertificateDataResponseDto registerCertificate(SecuredParentUUID au
...
+                fireRegistrationEventIfChallengeProtected(certificate.getUuid());
             }
-        } else {
-            stateMachine.transition(certificate, CertificateState.REGISTERED);
-            logger.info("Certificate {} registered by authority", certificate.getUuid());
-            fireRegistrationEventIfChallengeProtected(certificate.getUuid());
+        } catch (RuntimeException | ConnectorAcceptedButLocalFailureException e) {
+            // Post-acceptance local failure: state stays as it is (state-divergence rule). The binding writer records
+            // its own divergence; the other steps are recorded here, the same line the register catch writes.
+            if (!(e instanceof ConnectorAcceptedButLocalFailureException)) {
```
ivosh: **. Nitpick** | Maintainability

The post-acceptance guard this PR wraps around `persistRegistrationMeta`/`persistRegistrationBinding` catches a two-type union at `ClientOperationServiceImpl.java:772` and then immediately splits it again by type at `:775` (`if (!(e instanceof ConnectorAcceptedButLocalFailureException))`). Multi-catch exists for branches that handle every caught type *uniformly*; branching on the runtime type inside it is what separate `catch` clauses express, and the compiler then checks the split instead of the reader. Two clauses — `catch (ConnectorAcceptedButLocalFailureException e) { discardUndeliveredEabCredential(...); throw e; }` followed by `catch (RuntimeException e) { addEventHistory(...); discardUndeliveredEabCredential(...); throw e; }` — cost one extra line and remove the negated instanceof entirely (Clean Code, small focused blocks; *Effective Java* Item 73 context on handling exceptions at the right abstraction). Note this is not the established neighbour pattern: the two peer sites at `:2219` and `:2521` use `instanceof` because their condition is compound (`connectorAccepted || e instanceof ...`) over a `catch (Exception e)`, which genuinely cannot be split into clauses. This one can.

_Suggested:_ Replace the multi-catch at ClientOperationServiceImpl.java:772 with a `catch (ConnectorAcceptedButLocalFailureException e)` clause followed by a `catch (RuntimeException e)` clause and drop the `instanceof` test.

src/main/java/com/otilm/core/service/acme/registration/ExternalAccountBindingJws.java
```diff
@@ -0,0 +1,99 @@
...
+     * @throws UnusableRegistrationChallengeException when {@code key} is shorter than the HS256 minimum, so no binding
+     * could ever verify under it
+     */
+    public boolean verify(byte[] key) {
+        try {
+            return jws.verify(new MACVerifier(key));
+        } catch (KeyLengthException e) {
+            throw new UnusableRegistrationChallengeException(
+                    "registration challenge is shorter than the HS256 minimum key length", e);
+        } catch (JOSEException e) {
```
ivosh, line 102: **. Nitpick** | Error handling

`ExternalAccountBindingJws.verify` (`ExternalAccountBindingJws.java:95-96`) catches `JOSEException` and returns `false` with no log and no comment. That return value is not inert: the predicate is handed to `RegistrationResolver.verifyMac`, which runs it inside `registrationChallengeGate.verify`, so `false` means "wrong MAC" and spends one counted attempt against the pre-registration — the very thing `AcmeRegistrationBindingITest.wrongKeyIsRejectedGenericallyAndCountedAgainstTheRegistration` pins. A `JOSEException` that is *not* a `KeyLengthException` (a missing/failed JCA HMAC provider, a verifier construction fault) would therefore burn attempts and eventually lock out a perfectly good registration, with no trace anywhere: the class has no logger, and `AcmeRegistrationBinder.rejection` only logs the generic reason. This PR already draws the distinction for one such fault by rethrowing `UnusableRegistrationChallengeException` for `KeyLengthException`; the residual branch is the one with nothing behind it (*Effective Java* Item 77, never ignore exceptions). The sibling swallow in `bindsAccountKey` at `:78-79` is harmless by comparison because it runs before the gate.

_Suggested:_ Add a logger to ExternalAccountBindingJws and log the swallowed `JOSEException` at warn (with the exception) before returning false at line 96.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -1009,12 +1029,44 @@ private AcmeAccount addNewAccount(String profileName, String publicKey, NewAccou
...
+                throw e;
+            }
+            logger.info("ACME registration-mode newAccount rejected: registration bound concurrently");
+            throw new AcmeProblemDocumentException(HttpStatus.UNAUTHORIZED, Problem.UNAUTHORIZED,
+                    AcmeRegistrationBinder.REGISTRATION_REJECTION);
+        }
+    }
+
+    private static boolean isRegistrationBindingCollision(DataIntegrityViolationException e) {
+        return CryptoAssetConstraintTranslator
```
ivosh: **. Nitpick** | Maintainability

`AcmeServiceImpl.isRegistrationBindingCollision` at `AcmeServiceImpl.java:1063-1067` calls `CryptoAssetConstraintTranslator.constraintNameOf(e)`. The method itself is domain-neutral (it walks the cause chain for a Hibernate `ConstraintViolationException` and folds the bare name with `Locale.ROOT`), but its class is named for, and its Javadoc explicitly scoped to, the cryptographic-asset inventory: "Turns a cryptographic-asset constraint violation into a sentence that is safe to hand to a caller" (`src/main/java/com/otilm/core/dao/CryptoAssetConstraintTranslator.java:9-18`), and its `BY_CONSTRAINT` map holds only `crypto_asset`/`cbom` names. A reader of the ACME path now has to open a CBOM class to learn that the import is not a layering mistake. The only precedent for using it outside CBOM is generic logging in `ExceptionHandlingAdvice.java:701` — this PR is the first business decision taken on it. A name should say what the class does (Clean Code); the neutral half of this one wants its own home.

_Suggested:_ Move `constraintNameOf` into a domain-neutral helper (e.g. `com.otilm.core.dao.ConstraintViolations`) that `CryptoAssetConstraintTranslator` and `AcmeServiceImpl` both delegate to.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
```diff
@@ -1055,6 +1101,41 @@ private void maybeCreateRegistrationAuthorization(Certificate certificate,
...
+        } catch (RuntimeException e) {
+            logger
+                    .warn("Failed to record the undelivered EAB credential for certificate {}: {}",
+                            certificate.getUuid(), e.getMessage());
+        }
+    }
+
+    private static void attachEabCredential(ClientCertificateDataResponseDto response, String generatedChallenge) {
+        if (generatedChallenge != null) {
+            response.setEabKid(response.getUuid());
```
ivosh: **. Nitpick** | Maintainability

`ClientOperationServiceImpl.java:1136` sets `response.setEabKid(response.getUuid())` — the helper recovers the registration UUID out of the response object rather than being told it. Both call sites (`:792` and `:872`) happen to set `uuid` on the line before, so it works today, but the contract is an undeclared call-order dependency: moving `attachEabCredential` one line up, or adding a third caller that builds the DTO in a different order, yields `eabKid = null` silently and the client gets an EAB credential it cannot use. The method already takes the other half of the credential as a parameter; the kid should arrive the same way (a method's inputs should be its parameters — Clean Code; *Effective Java* Item 51 on designing method signatures).

_Suggested:_ Change `attachEabCredential` to take the certificate UUID as a parameter and pass `certificate.getUuid().toString()` from both call sites.

ivosh (changes requested):

src/main/java/com/otilm/core/service/acme/registration/ExternalAccountBindingJws.java
klaraf755: Fixed as suggested. `ExternalAccountBindingJws` exposes `hasNonce()` and the verifier rejects a binding whose protected header carries one, with the same generic `unauthorized` as every other failed binding. `ExternalAccountBindingJwsTest.aNonceInTheProtectedHeaderIsVisible` pins the parse side and `AcmeEabVerifierTest.aBindingCarryingANonceIsRejected` the rejection.

The class moved package in the rework, from `service.acme.registration` to `service.acme.eab`.

src/main/java/com/otilm/core/service/acme/registration/ExternalAccountBindingJws.java
klaraf755: Fixed. `ExternalAccountBindingJws` now has a logger and warns with the exception before returning false.

Your stated consequence no longer applies, though: there is no attempt gate and no lockout in this design, so a `JOSEException` cannot burn attempts or lock a registration out — the binding key is profile-level and its strength is the whole defence. What remains is the diagnosability half of the argument, which is reason enough on its own: the client sees the same generic 401 either way, so without the log a broken JCA HMAC provider leaves no trace anywhere.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
klaraf755: Addressed, on the contract side. `ExternalAccountBinding.signature` now carries `@ToString.Exclude` (interfaces#960), so the three log lines you name no longer emit the MAC — the offline-guessing oracle is closed at the source rather than at each call site. The protected header and payload still render; both are public inputs, the header naming a secret UUID and the payload being the account's own public JWK.

The second half of your argument is gone with the design: there is no operator-supplied `authorizationSecret` on this path any more. The binding key is 256 bits of platform-generated randomness held in a secret, never a passphrase, and it gates account registration rather than issue/renew/rekey across the platform API, SCEP and CMP.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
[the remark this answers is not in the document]
klaraf755: Fixed, and the fix survived the rework: an omitted `contact` is serialized as an empty list rather than the literal `null`. `AcmeEabAccountITest` exercises the newAccount path that reaches it.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: No longer applies. This branch was rewritten: External Account Binding is now configured on the ACME Profile as a list of secrets holding accepted HMAC keys, and is no longer bound to a certificate pre-registration. `generateEabCredential`, `attachEabCredential`, `discardUndeliveredEabCredential`, `AcmeRegistrationBinder`, `acme_account.registration_certificate_uuid` and the `RegistrationResolver` extraction are all gone, and `ClientOperationServiceImpl` and the CMP path are byte-identical to main again. The code this comment points at no longer exists.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
klaraf755: No longer applies. This branch was rewritten: External Account Binding is now configured on the ACME Profile as a list of secrets holding accepted HMAC keys, and is no longer bound to a certificate pre-registration. `generateEabCredential`, `attachEabCredential`, `discardUndeliveredEabCredential`, `AcmeRegistrationBinder`, `acme_account.registration_certificate_uuid` and the `RegistrationResolver` extraction are all gone, and `ClientOperationServiceImpl` and the CMP path are byte-identical to main again. The code this comment points at no longer exists.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: No longer applies. This branch was rewritten: External Account Binding is now configured on the ACME Profile as a list of secrets holding accepted HMAC keys, and is no longer bound to a certificate pre-registration. `generateEabCredential`, `attachEabCredential`, `discardUndeliveredEabCredential`, `AcmeRegistrationBinder`, `acme_account.registration_certificate_uuid` and the `RegistrationResolver` extraction are all gone, and `ClientOperationServiceImpl` and the CMP path are byte-identical to main again. The code this comment points at no longer exists.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: No longer applies. This branch was rewritten: External Account Binding is now configured on the ACME Profile as a list of secrets holding accepted HMAC keys, and is no longer bound to a certificate pre-registration. `generateEabCredential`, `attachEabCredential`, `discardUndeliveredEabCredential`, `AcmeRegistrationBinder`, `acme_account.registration_certificate_uuid` and the `RegistrationResolver` extraction are all gone, and `ClientOperationServiceImpl` and the CMP path are byte-identical to main again. The code this comment points at no longer exists.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
klaraf755: No longer applies. This branch was rewritten: External Account Binding is now configured on the ACME Profile as a list of secrets holding accepted HMAC keys, and is no longer bound to a certificate pre-registration. `generateEabCredential`, `attachEabCredential`, `discardUndeliveredEabCredential`, `AcmeRegistrationBinder`, `acme_account.registration_certificate_uuid` and the `RegistrationResolver` extraction are all gone, and `ClientOperationServiceImpl` and the CMP path are byte-identical to main again. The code this comment points at no longer exists.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
klaraf755: No longer applies. The rollback attribute is gone: `newAccount` carries no method-level `@Transactional` at all in the rewritten branch, because there is no `saveAndFlush` and no unique-constraint race to convert into a rejection. The account is written only after the binding has verified.

The rule you cite is still the right one, and the one new transactional boundary this branch does add follows it — `AcmeEabVerifier.verify` uses Spring's annotation with `Propagation.NOT_SUPPORTED`, so `TransactionalBoundaryArchTest.effectivePropagation` can resolve it. Reading a binding key is a vault HTTP call and must not be held inside the newAccount transaction.

src/main/java/com/otilm/core/service/v2/impl/ClientOperationServiceImpl.java
klaraf755: No longer applies — `attachEabCredential` does not exist in the rewritten branch, and `ClientOperationServiceImpl` is byte-identical to main again. External Account Binding is now configured on the ACME Profile as a list of secrets holding accepted HMAC keys, so no credential is built into a registration response and there is no kid to pass anywhere.

To answer the question behind it: the kid is the secret's own UUID, which the verifier reads from the binding's protected header and matches against the profile's configured list. Nothing recovers it from a response object.

src/main/java/com/otilm/core/service/acme/eab/AcmeEabVerifier.java
```diff
@@ -0,0 +1,140 @@
...
+        if (!verified) {
+            throw reject(acmeProfile, "the binding MAC does not verify under the named key");
+        }
+        return keyUuid;
+    }
+
+    private byte[] macKeyOf(AcmeProfile acmeProfile, UUID keyUuid) throws AcmeProblemDocumentException {
+        SecretContent content;
+        try {
+            content = secretService.getSecretContent(keyUuid);
```
ivosh, line 101: **! Major** | Performance

**`newAccount` is a public endpoint, and every EAB-enabled call now holds two database connections while the vault answers.**

`AcmeServiceImpl` is transactional at the class level (`AcmeServiceImpl.java:141`), and `newAccount` does not override it. That transaction has already used its connection before verification runs — `findByPublicKey` at lines 302-303, `getProfiles` at 971. Hibernate holds the connection until commit, and nothing in `application.yml` changes the default handling mode.

`AcmeEabVerifier.verify` is `NOT_SUPPORTED` (`AcmeEabVerifier.java:62`), which **suspends** that transaction rather than completing it. `JpaTransactionManager.doSuspend` unbinds the holders from the thread but never closes the `EntityManager`. The connection stays checked out.

`getSecretContent` then opens a second one. `SecretServiceImpl` is class-level transactional (`SecretServiceImpl.java:131`) and line 900 has no override, so `REQUIRED` starts a fresh transaction. The connector call runs inside it at lines 925-927. The fingerprint writes at 941 and 944 come after it, so the transaction cannot end before the vault answers.

That is two connections per in-flight request. The pool is 40 (`application.yml:254`). **Twenty concurrent EAB `newAccount` requests take all of it**, and every other operation on the platform then fails after the 30s `connectionTimeout` at line 252. A wedged vault holds each request for roughly 45 seconds — 3s connect plus 35s response, or 10s pending-acquire (lines 402-407).

**What is new here is not the pattern, it is who can reach it.** Calling the vault inside a transaction is pre-existing behaviour of `getSecretContent`. Until this PR that method had exactly one caller: `SecretManagementControllerImpl:72`, an authenticated management endpoint. This PR adds the second, and it is reached from an unauthenticated, directory-reachable endpoint. The doubling and the public reachability together are what make this worth fixing; neither alone would be.

_Suggested:_ Move the boundary to the entry point, the way `validateChallenge` already does at `AcmeServiceImpl.java:626` — declare `newAccount` itself `NOT_SUPPORTED`, and wrap the local reads and the `save` at 1033 in short transactional methods. That removes the suspended connection and leaves only the single one every existing `getSecretContent` caller already pays. It also gives the profile reload-and-lock a natural home, which is where the TOCTOU at line 1016 has to be fixed anyway.

<sub>flagged by Correctness &amp; races</sub>

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -1000,7 +1011,14 @@ private AcmeAccount addNewAccount(String profileName, String publicKey, NewAccou
         } else {
             return oldAccount;
         }
+        // The binding is the account's admission: verified before any row exists, so a rejected one leaves
+        // nothing behind. Its key is read from the vault, which is why the verifier runs outside this transaction.
+        UUID eabSecretUuid = acmeProfile.isExternalAccountRequired()
```
ivosh, line 1046: **! Major** | Correctness

AcmeServiceImpl.java:1016 reads the managed profile's EAB list and then, after the vault round-trip, persists the account at line 1033 without locking or reloading the profile. A concurrent edit can enable EAB after the empty-list check and still allow an unbound account, or remove the selected key while verification is in flight and still admit the account under that key because the suspended transaction retains the stale entity.

_Suggested:_ Reload and lock the ACME profile in a short post-verification transaction, revalidate the requirement and selected key, and insert the account in that same transaction.

src/main/java/com/otilm/core/service/impl/SecretServiceImpl.java
```diff
@@ -622,6 +630,7 @@ public void deleteSecret(UUID uuid, boolean deleteInVaults) throws NotFoundExcep
         authorizationEnforcer
                 .enforce(Resource.VAULT_PROFILE, ResourceAction.MEMBERS,
                         SecuredUUID.fromUUID(secret.getSourceVaultProfile().getUuid()));
+        refuseWhenBoundToAcmeProfile(uuid, secret.getName());
```
ivosh, line 633: **! Major** | Correctness

SecretServiceImpl.java:633 checks profile references only when the delete request is queued, while the actual deletion runs asynchronously at line 661 and may wait for approval. AcmeProfileServiceImpl.java:351 can assign the still-existing secret during that gap, and neither path locks the secret or has a foreign-key constraint, so both operations can pass and deletion can leave a profile advertising `externalAccountRequired` for a missing key.

_Suggested:_ Serialize profile assignment and deletion on the secret row, reject secrets already pending deletion, and repeat the reference check in `deleteSecretAction` immediately before the local delete.

src/main/java/com/otilm/core/service/acme/eab/ExternalAccountBindingJws.java
```diff
@@ -0,0 +1,113 @@
...
+    public boolean verify(byte[] key) {
+        try {
+            return jws.verify(new MACVerifier(key));
+        } catch (KeyLengthException e) {
+            throw new UnusableEabKeyException("binding key is shorter than the HS256 minimum key length", e);
+        } catch (JOSEException e) {
+            // Anything other than a short key is a fault in the JCA HMAC provider or the verifier, not a wrong MAC.
+            // It reaches the client as the same generic rejection, so without this line it leaves no trace at all.
+            logger.warn("External Account Binding MAC could not be evaluated", e);
+            return false;
```
ivosh: **~ Minor** | Error handling

ExternalAccountBindingJws.java:110 returns `false` for a `JOSEException` even though the preceding comment identifies it as a JCA provider or verifier fault rather than a wrong MAC. `AcmeEabVerifier` consequently follows its bad-MAC branch and returns `unauthorized`, so valid clients receive a credential rejection for an internal service failure instead of the documented `serverInternal` response.

_Suggested:_ Propagate `JOSEException` as an internal EAB verification failure and map it to `serverInternal` rather than returning `false`.

ivosh (changes requested):

src/main/java/com/otilm/core/service/acme/eab/ExternalAccountBindingJws.java
klaraf755: Fixed. `verify` now throws `UnusableEabKeyException` for a `JOSEException` instead of returning false, so the verifier maps it to `serverInternal` alongside the short-key case rather than following the bad-MAC branch. The Javadoc says both are faults that must not reach a client as a rejection, and `aMacThatCannotBeEvaluatedIsNotReportedAsAWrongMac` pins it.

The class logger is gone with it — the warn line existed only because the fault was being swallowed, and the verifier already logs at the point where it classifies the failure.

src/main/java/com/otilm/core/service/impl/SecretServiceImpl.java
klaraf755: Partly fixed, and the remainder is a design question I would rather not settle silently.

`refuseWhenBoundToAcmeProfile` now also runs in `deleteSecretAction`, immediately before the local delete, so the check happens on the side of the queue and the approval where the row actually goes. That closes the window you describe for the case that matters: a profile claiming the secret after the request was queued.

What I have not done is serialize the two operations on the secret row or reject secrets already pending deletion. Both are worth doing, but they change the locking behaviour of a shared secret path used well beyond ACME, and the residual race is now narrow — assignment would have to land between the recheck and the delete in the same action. I would rather carry that with the transaction restructure the two Major findings on this PR call for, or as its own change, than fold it in here.

src/main/java/com/otilm/core/service/acme/eab/AcmeEabVerifier.java
klaraf755: Addressed by restructuring the endpoint rather than tuning around it, though not completely.

`newAccount` now carries `@Transactional(propagation = NOT_SUPPORTED)` — Spring's annotation, so `TransactionalBoundaryArchTest.effectivePropagation` can resolve it. There is no ambient transaction to suspend, so this method no longer holds a connection across the vault call: each read runs in its own short transaction and returns its connection, and the account is written in an explicit transaction opened after verification has finished.

What that does not remove is the transaction the secret read opens for itself. `SecretServiceImpl.getSecretContent` is `REQUIRED` and writes a secret version after the connector answers, so the HTTP call still runs inside one. This takes the endpoint from two connections per in-flight request to one, not to zero, and the Javadoc on `newAccount` and `AcmeEabVerifier.verify` has been corrected to say so. Removing the last one means splitting `getSecretContent` into read, connector call, and a short transaction for the version write — pre-existing code with several callers beyond ACME, so I have left it rather than fold it in here.

One correction to the framing. `getSecretContent` is not reached only from an authenticated operator endpoint: `SecretServiceImpl.getResourceObjectContent` calls it as the `AttributeResourceService` hook, `ResourceServiceImpl.loadResourceObjectContentData` dispatches to that, and it is reached from `ConnectorRequestAttributesBuilder` and from the authority-instance, entity-instance, token-instance, notification-instance, callback and discovery paths. A RESOURCE-typed attribute naming a secret already pulls vault content inside a transaction on the ordinary connector-request path.

What is new is the reachability rather than the pattern: every one of those callers sits behind an operator action or a connector operation the platform itself initiates, whereas newAccount can be driven at whatever rate a client chooses. The pool arithmetic stands.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
klaraf755: Fixed, though my first attempt at it was a no-op and the finding is worth reading twice because of why.

The account is written by `insertAccountAgainstTheProfileAsItStands`, which opens an explicit transaction, re-reads the profile through a new `findAndLockByUuid` (`@Lock(PESSIMISTIC_WRITE)`), and decides again on the locked row before inserting. That covers both directions you name: a profile that had External Account Binding switched on during the vault round-trip, and one whose selected key was withdrawn.

The locked re-read alone was not enough. `spring.jpa.open-in-view` is unset, so it defaults to true and an `EntityManager` stays bound for the whole request; the profile read before the vault call therefore remained managed, and the locked read handed back that same instance with its pre-vault state, since Hibernate returns the cached entity and only upgrades the lock mode rather than re-hydrating, and `AcmeProfile` carries no version column to catch it. The check compared the stale snapshot against itself. The pre-vault copy is now detached before the locked read, so the row is what gets read in both environments. Worth noting that the integration test passed throughout, for the reason that hides the problem: `BaseSpringBootTest` is not transactional and has no open-in-view, so every repository call there gets a fresh persistence context.

Two fixes came with it. `commit()` is no longer inside the `try` — a failure there has already rolled back and completed the transaction, so rolling back again replaced the real cause with an `IllegalTransactionStateException`, and an INSERT failure surfaces exactly there because the assigned-id entity defers its insert to flush. And the requirement-switched-on-mid-request case now answers `externalAccountRequired` rather than `unauthorized`, since the client sent no binding and needs telling to bring one; a withdrawn key still answers `unauthorized` with `AcmeEabVerifier.REJECTION`, so it reads no differently from a binding that never verified.

`aKeyWithdrawnWhileTheBindingWasVerifiedDoesNotAdmitTheAccount` covers the revalidation.

src/main/java/com/otilm/core/service/acme/eab/AcmeEabVerifier.java
klaraf755: Correction to my previous reply, which claimed more than the change achieves.

The connection this method held is gone, but the vault call still runs inside a transaction — `SecretServiceImpl.getSecretContent` is `REQUIRED` and writes a secret version after the connector answers, so it opens one of its own around the HTTP call. The restructure took this from two connections per in-flight request to one; it did not take it to zero. I have corrected the Javadoc on both `newAccount` and `AcmeEabVerifier.verify`, which said otherwise.

Removing the remaining one means splitting `getSecretContent` into read, call, and a short transaction for the version write. That is pre-existing code on a path used well beyond ACME, so I have left it, but it is load-bearing for this PR and should not be described as solved.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
klaraf755: Reopened and fixed properly — the revalidation I described was a no-op in a real request.

`spring.jpa.open-in-view` is unset, so it defaults to true and an `EntityManager` is bound for the whole request. The profile read before the vault call therefore stayed managed, and the locked read handed back that same instance with its pre-vault state: Hibernate returns the cached entity and only upgrades the lock mode rather than re-hydrating, and `AcmeProfile` carries no version column to catch it. The check compared the stale snapshot against itself.

The integration test passed for the reason that hides it: `BaseSpringBootTest` is not transactional and has no open-in-view, so every repository call there gets a fresh persistence context.

`insertAccountAgainstTheProfileAsItStands` now detaches the pre-vault copy before the locked read, so the row is what gets read in both environments. Two further fixes came with it: `commit()` is no longer inside the `try`, since a failure in commit has already rolled back and completed the transaction and rolling back again replaced the real cause with an `IllegalTransactionStateException` — an INSERT failure surfaces there, because the assigned-id entity defers its insert to flush. And the requirement-switched-on-mid-request case now answers `externalAccountRequired` rather than `unauthorized`, since the client sent no binding and needs to be told to bring one.

src/main/java/com/otilm/core/service/acme/eab/AcmeEabVerifier.java
```diff
@@ -0,0 +1,142 @@
...
+    public void setSecretService(SecretExternalService secretService) {
+        this.secretService = secretService;
+    }
+
+    /**
+     * Resolves the binding to one of the profile's keys and verifies its MAC.
+     *
+     * <p>
+     * Runs outside the caller's transaction: reading a key goes to the vault over HTTP, and no caller's transaction
+     * should be held open for that. The secret read still opens one of its own, which is not this class to fix. The
```
ivosh: **. Nitpick** | Documentation

src/main/java/com/otilm/core/service/acme/eab/AcmeEabVerifier.java:55 appends `which is not this class to fix` to an otherwise useful transaction warning; that phrase records a scope decision instead of explaining runtime behavior and gives future readers no actionable constraint.

_Suggested:_ Rewrite the sentence as `The secret read still opens a transaction of its own.`

src/main/java/com/otilm/core/service/impl/SecretServiceImpl.java
ivosh: **Re-check: partial.**

In `SecretServiceImpl.deleteSecretAction()`, the delta adds `refuseWhenBoundToAcmeProfile(secretUuid, secret.getName())`, but places it before the vault-deletion loop; the unchanged local `secretRepository.delete(secret)` occurs afterward, and the delta adds neither secret-row locking nor a pending-deletion rejection.

The asynchronous action now rechecks profile references, but the requested serialization and pending-deletion guard remain absent, and assignment can still race between the recheck and local deletion.

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -272,8 +293,21 @@ public ResponseEntity<?> getNonce(String acmeProfileName, Boolean isHead, URI re
...
+     *
+     * <p>
+     * This removes the connection this method would otherwise hold across the vault call. It does not remove the one
+     * the secret read opens for itself: {@code SecretServiceImpl.getSecretContent} is transactional and writes a secret
+     * version after the connector answers, so the call still runs inside a transaction of its own.
+     */
     @Override
     @ProtocolEndpoint
+    @org.springframework.transaction.annotation.Transactional(
+            propagation = org.springframework.transaction.annotation.Propagation.NOT_SUPPORTED)
```
ivosh: **. Nitpick** | Maintainability

`Propagation` is fully qualified here, but it is already imported at line 139 and nothing shadows it.

The FQN on line 309 is necessary: `jakarta.transaction.Transactional` (line 88) occupies that simple name, for the class-level annotation and for `finalizeOrder`. The prefix on `Propagation` is redundant on top of that.

`validateChallenge` at line 657 writes the same construct the short way — `propagation = Propagation.NOT_SUPPORTED` — and so does the new `AcmeEabVerifier.verify()` at line 64.

_Suggested:_ `@org.springframework.transaction.annotation.Transactional(propagation = Propagation.NOT_SUPPORTED)`

ivosh (dismissed):

src/main/java/com/otilm/core/service/acme/eab/AcmeEabVerifier.java
klaraf755: Fixed in 68252f51e. The sentence that "records a scope decision instead of explaining runtime behavior" is gone; the paragraph now reads "The secret read still opens a transaction of its own."

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
klaraf755: Fixed in 68252f51e. The annotation now reads `@org.springframework.transaction.annotation.Transactional(propagation = Propagation.NOT_SUPPORTED)` — the qualified name stays only on `Transactional`, where `jakarta.transaction.Transactional` occupies the simple name, and the redundant prefix on `Propagation` is dropped.

ivosh (dismissed):

ivosh (approved):

PR #2240, opened 2026-09-08
Offer as columns only the fields a listing carries

### LINKED ISSUE #2180: Decide per-field displayability instead of advertising every property field as a column
### Description

`SearchHelper.prepareSearch` sets `displayable` unconditionally: every property field in every catalogue is advertised as a column a user may pick. For a number of fields nothing on the listing side can supply the value, so the column is offered and renders blank.

Verified on certificates. `CertificateDetailDtoMapper.toListDto` sets neither `keyUsage` nor `subjectAlternativeNames`, and the protocol association is not mapped either, so `KEY_USAGE`, `SUBJECT_ALTERNATIVE_NAMES` and `CERTIFICATE_PROTOCOL` are pickable today and would each render empty. Those three are the worked examples; the same question has to be answered for the other six inventories.

**There is no mechanical rule to derive this from, which is why it was held out of `core#2162` rather than guessed at.** Both candidate rules were checked and both fail:

1. *"A field with no attribute behind it has nothing to render."* All 181 `FilterField` entries have a non-null `fieldAttribute`, so this excludes none of them.
2. *"A field reached through a join is not in the list DTO."* Structure predicts nothing: `CERTIFICATE_PROTOCOL` is reached through a join and is absent from the list DTO, while `raProfile` and `groups` are reached through joins and are present.

A reflective contract test is not available either: it would need a `FilterField` → DTO-property mapping that does not exist. `CERTIFICATE_PROTOCOL` publishes `protocolAssociation` as its attribute and nothing links that to a DTO getter, so any such test needs a hand-written exception list — the judgement call again, with machinery around it.

So the work is an explicit per-field displayability decision across the seven listing mappers, taken with the inventory owners and whoever owns the frontend render registry. A wrong guess silently removes a column a user can pick today, which is worse than leaving one blank, so it wants deciding rather than inferring.

**Scope boundary.** Fields the DTO *does* carry but which nothing renders on the frontend — `FINGERPRINT`, `SUBJECTDN`, `KEY_SIZE` and similar — are `fe-administrator#2067`'s scope, with an acceptance criterion being added to `fe-administrator#2020` so the pilot cannot pass while shipping blank certificate columns. This issue is the other case: the value is not carried at all, so no frontend change can render it.

**A lead worth costing before adding DTO fields.** `AttributeColumnProjector.parseRequestedColumns` drops a requested column whose `fieldSource` has no attribute type, which is what excludes `PROPERTY`. The frontend's `getProjectedContent` keys on `column.fieldSource` without caring which source it is, and `FilterFieldSource` already has `Property`, so `attributeValues.property.KEY_USAGE` would render as-is. Relaxing that one check may be cheaper than adding N DTO fields and N renderers. Worth costing first, and it may change the shape of the answer for every inventory.

This is `core#2163` in reverse: that issue adds catalogue fields that are missing, this one stops the catalogue promising columns that cannot be shown.

Note: `sortable` had the same over-promise and was handled this way — reported hard `false` in `core#2162`, then turned on with a per-resource gate in `core#2031`. With that now done, `displayable` is the last capability flag still set unconditionally.

Raised at the request of the reviewer on `OmniTrustILM/core#2162`, where it was agreed the decision does not belong inside that PR.

### Definition of Done

- [ ] A per-field displayability decision exists for all seven inventories, agreed with the inventory and frontend-registry owners, and is recorded somewhere a reader can check it against the mappers.
- [ ] The projection route above is costed against adding DTO fields and renderers, and the choice is recorded with its reason.
- [ ] `SearchHelper` stops setting `displayable` unconditionally and reflects that decision.
- [ ] `KEY_USAGE`, `SUBJECT_ALTERNATIVE_NAMES` and `CERTIFICATE_PROTOCOL` on certificates either carry a value or are no longer offered as columns.
- [ ] A test pins the decision so a new `FilterField` cannot silently become a blank column.
- [ ] Sonar quality gate passes and new code meets at least 80% coverage.

Jakub-Bartal-01:
`SearchHelper.prepareSearch` set `displayable` unconditionally, so every property field of every catalogue was advertised as a column a user may pick. For the fields the listing DTO carries no value for, picking one produced a heading whose every cell was empty.

The flag is now decided per field: a field is offered when its listing is wired to the configurable-column pipeline at all, and that listing carries the value the column would show.

Closes #2180.

## The decision

There is no rule to derive this from, and the issue's two candidates were both re-checked and both fail. Every `FilterField` of these seven listings has a non-null `fieldAttribute`, so "no attribute behind it" excludes none of them. Structure predicts nothing either: `CERTIFICATE_PROTOCOL` is reached through a join and is absent from the listing, while `RA_PROFILE_NAME` and `GROUP_NAME` are reached through joins and are present.

So it was decided by reading what each of the seven listings actually selects. Six go through a mapper; the certificate listing does not, building each `CertificateDto` from a JPQL constructor projection with `CertificateDetailDtoMapper.toListDto` serving only the protocol-specific listings. The result is an explicit `ABSENT_FROM_LISTING` set in `SearchHelper`, naming the query each group was read against.

Seventeen fields stop being offered:

| Inventory | Fields | Why |
|---|---|---|
| Certificate | `CERT_LOCATION_NAME`, `KEY_USAGE`, `SUBJECT_TYPE`, `SUBJECT_ALTERNATIVE_NAMES`, `OCSP_VALIDATION`, `CRL_VALIDATION`, `SIGNATURE_VALIDATION`, `CERTIFICATE_PROTOCOL`, `ACME_PROFILE`, `SCEP_PROFILE`, `CMP_PROFILE`, `ACME_ACCOUNT`, `SUCCEEDING_CERTIFICATES`, `PRECEDING_CERTIFICATES` | absent from the listing's constructor projection (`findCertificateDtosByUuidsIn`) and from `CertificateDto` itself |
| Connector | `CONNECTOR_AUTH_TYPE` | the v2 `ConnectorDto` the listing returns carries no authentication type; only the v1 detail DTO does |
| Secret | `SECRET_SYNC_VAULT_PROFILE` | `Secret.setCommonFields` sets the source vault profile and not the sync ones |
| Signing record | `SIGNING_RECORD_SIGNED_DOCUMENT_RETRIEVED_AT` | `SigningRecordMapper.toListDto` sets the retrieval timestamp on the detail DTO only |

Cryptographic keys, discoveries and CBOMs lose nothing — every field of those three is carried.

`SIGNING_RECORD_SIGNING_PROFILE_VERSION` stays offered although it is nested: `toListDto` pins it onto the `signingProfile` member from the record's own captured version. Whether a renderer reaches it is `fe-administrator#2067`'s question, not the catalogue's.

## The projection route, costed

The issue flagged a cheaper-looking route: relax the `fieldSource.getAttributeType() == null` check in `AttributeColumnProjector.parseRequestedColumns`, so a property column survives parsing and arrives as `attributeValues.property.KEY_USAGE`, which `getProjectedContent` on the frontend already reads without caring about the source.

It does not work, and the reason is one level below that check. The projector's only value source is `attributeContent2ObjectRepository.getProjectedAttributesContent(...)`, which matches stored rows by attribute type and attribute name against the attribute-content tables. A property field has no row there at all — `KEY_USAGE` is a column on `certificate`. Letting the column past the parse gate therefore returns nothing and renders the same empty cell, having spent a query to do it.

Making it work needs a second value source that reads the entity, which is the listing mappers again. That is strictly more than adding a DTO field and a renderer for the handful of fields anyone actually wants, so the route is not taken and the fields are withdrawn instead.

## Sortable follows displayable, on the listings that publish columns

Ordering there is triggered by clicking a column header, so a field that cannot be shown has no header to click, however orderable its path is. The attribute path already coupled the two flags this way; the property path now does too. `CERTIFICATE_PROTOCOL` is the case that disagreed with itself: it resolves to one scalar the repository can order by, and shows nothing.

The refusal in `SortOrderBuilder.resolveField` moves onto that same predicate, so the comment claiming it matches the published flag is true rather than nearly true. It is widened by one clause for the fields of listings that publish no columns at all, where an ordering is not a column's: a caller naming a `SortSpecification` against the repository directly can order by `OID_ENTRY_CODE`, which no picker ever offered. `SubclassFilterFieldSearchITest` covers that side, and the new test covers both.

Nothing a client can request reaches the widened branch — `RequestValidatorHelper` already refuses a requested sort on any listing outside the pipeline — so for client requests the gate and the catalogue agree exactly.

## Fields of unwired listings

A listing outside the pipeline returns a DTO the projector cannot fill, so its fields now report `false` for both flags rather than advertising a column that listing cannot project. `SORT_WIRED_RESOURCES` is renamed `CONFIGURABLE_COLUMN_RESOURCES` to say that it now gates both.

The attribute path takes the same gate. `prepareSearchForJSON` derived `displayable` from content visibility alone, so an ordinary custom, data or metadata attribute of an unwired resource stayed advertised as a column although that listing never invokes `AttributeColumnProjector` - `LOCATION` is the plainest case. Its `sortable` was already gated per resource; `displayable` now is too, and `sortable` reads it rather than repeating the check.

## Saved views are held to the same flags

`ListViewServiceImpl` validated a view's columns and ordering by catalogue *membership*, which is the filter capability. So the flags published here would not have stopped a view from storing `CERTIFICATE_PROTOCOL`: as a column it comes back blank in every row, and as an ordering it is refused by `SortOrderBuilder` when the view is applied, failing the whole listing request rather than merely ordering it differently.

The catalogue now carries the three capabilities separately - the conditions a field accepts, whether it may be a column, whether it may be an ordering - read off the published catalogue rather than from a second copy of its rules. Columns and orderings are validated against the latter two. Filters stay on membership, because a listing filters on values it does not select: a certificate is still found by its protocol.

Views stored before this lands need no migration, and keep their columns.

A read returns every column whose field the catalogue still offers, including one the listing can no longer show, and drops only a field that has left the catalogue outright - the omission `ListViewDto.columns` already documents. Two things depend on that. `fe-administrator` resolves availability against the same catalogue and renders a withdrawn column as unavailable, named in a notice and removable from the picker, which withholding it silently prevents. And a view whose columns were all withdrawn would otherwise read back as an empty column list, which `ListViewUpdateRequestDto` refuses with a 422, so the next rename or pin of it would fail over a column the caller never touched. The shape a read returns has to be a shape a write accepts.

The column gate therefore applies to what a request introduces rather than to what the view holds: a creation is held to the current catalogue in full, while an edit may keep or remove a column the stored view already carries. A withdrawn column cannot be introduced, and cannot strand the view holding it.

An ordering keeps both the strict gate and the drop on read, because the two costs are not alike: an undisplayable column costs one blank heading, while an ordering `SortOrderBuilder` refuses costs every application of the view. Such a view opens in the listing's own order instead of erroring, and `sort: null` is always a legal write shape.

## The pin

`PropertyColumnDisplayabilityITest` asserts the published set of columns per listing equals a written-out literal, field by field over the whole enum. A field added to a wired listing is displayable by default, so it lands outside that map and turns the test red until someone has read the mapper and classified it. Neither a new blank column nor a quietly withdrawn one reaches the picker without that.

`ColumnCatalogueFlagsITest.aNativeArrayPropertyFieldIsDisplayable` moves from the time-quality-configuration catalogue to the connector one, since the first is not a column pipeline listing and now offers no columns at all.

## Still open

The decision is one engineer's reading of the mappers. It wants confirming with the inventory owners before it ships, particularly the certificate list: fourteen columns is the largest withdrawal, and a column someone uses today is worse to lose than a blank one is to keep.

### DIFF
--- a/src/main/java/com/otilm/core/service/impl/ListViewServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/ListViewServiceImpl.java
@@ -208,16 +208,22 @@ public class ListViewServiceImpl implements ListViewExternalService, ListViewInt
                 .setColumns(view
                         .getColumns()
                         .stream()
-                        .filter(column -> catalogue.offers(CatalogueField.of(column)))
+                        .filter(column -> catalogue.canDisplay(CatalogueField.of(column)))
                         .toList());
         dto.setFilters(view.getFilters());
-        dto.setSort(view.getSort());
+        // Returning an ordering the listing would now refuse hands the client a view whose every application answers
+        // an error; dropping it opens the view in the listing's own order instead.
+        dto
+                .setSort(view.getSort() != null && catalogue.canSortBy(CatalogueField.of(view.getSort()))
+                        ? view.getSort()
+                        : null);
         return dto;
     }
 
     /**
      * Rejects anything the resource's catalogue cannot apply, so a view is stored only in a shape the listing can
-     * actually use. A field that disappears afterwards is a different case, and is skipped on read instead.
+     * actually use. A field that afterwards disappears, or stops being one the listing can show or order by, is a
+     * different case and is dropped on read instead.
      */
     private void validateRequest(Resource resource, ListViewUpdateRequestDto request) {
         Catalogue catalogue = catalogueOf(resource);
@@ -250,6 +256,17 @@ public class ListViewServiceImpl implements ListViewExternalService, ListViewInt
                         .filter(column -> !catalogue.offers(CatalogueField.of(column)))
                         .map(ListViewColumnDto::getFieldIdentifier)
                         .toList());
+
+        List<String> unshowable = columns
+                .stream()
+                .filter(column -> !catalogue.canDisplay(CatalogueField.of(column)))
+                .map(ListViewColumnDto::getFieldIdentifier)
+                .toList();
+        if (!unshowable.isEmpty()) {
+            throw new ValidationException(ValidationError
+                    .create("Resource %s does not offer these fields as columns: %s."
+                            .formatted(resource.getCode(), String.join(", ", unshowable))));
+        }
     }
 
     /**
@@ -282,8 +299,16 @@ public class ListViewServiceImpl implements ListViewExternalService, ListViewInt
     }
 
     private static void validateSort(Resource resource, SearchSortRequestDto sort, Catalogue catalogue) {
-        if (sort != null && !catalogue.offers(CatalogueField.of(sort))) {
+        if (sort == null) {
+            return;
+        }
+        if (!catalogue.offers(CatalogueField.of(sort))) {
             rejectUnknown(resource, List.of(sort.getFieldIdentifier()));
+        }
+        if (!catalogue.canSortBy(CatalogueField.of(sort))) {
+            throw new ValidationException(ValidationError
+                    .create("Resource %s cannot be ordered by %s."
+                            .formatted(resource.getCode(), sort.getFieldIdentifier())));
         }
     }
 
@@ -295,27 +320,42 @@ public class ListViewServiceImpl implements ListViewExternalService, ListViewInt
     }
 
     /**
-     * Every field the resource's listing can address, with the conditions each of them accepts. Properties come from
-     * the filter-field enum, everything else from the attribute definitions currently registered for the resource.
+     * Every field the resource's listing can address, with what each of them may be used for. Properties come from the
+     * filter-field enum, everything else from the attribute definitions currently registered for the resource.
+     *
+     * <p>
+     * Read from the published catalogue rather than from a copy of its rules, so the flags the client picked a view out
+     * of and the answer it gets back when it saves one cannot disagree.
      */
     private Catalogue catalogueOf(Resource resource) {
-        Map<CatalogueField, List<FilterConditionOperator>> fields = new HashMap<>();
+        Map<CatalogueField, Capabilities> fields = new HashMap<>();
         FilterField
                 .getEnumsForResource(resource)
                 .forEach(field -> fields
                         .put(new CatalogueField(FilterFieldSource.PROPERTY, field.name()),
-                                SearchHelper.availableConditions(field)));
+                                new Capabilities(SearchHelper.availableConditions(field),
+                                        SearchHelper.isDisplayable(field), SearchHelper.isSortableField(field))));
         attributeEngine
                 .getResourceSearchableFields(resource, false)
                 .forEach(group -> group
                         .getSearchFieldData()
                         .forEach(field -> fields
                                 .put(new CatalogueField(group.getFilterFieldSource(), field.getFieldIdentifier()),
-                                        field.getConditions())));
+                                        new Capabilities(field.getConditions(),
+                                                Boolean.TRUE.equals(field.getDisplayable()),
+                                                Boolean.TRUE.equals(field.getSortable())))));
         return new Catalogue(fields);
     }
 
-    private record Catalogue(Map<CatalogueField, List<FilterConditionOperator>> fields) {
+    /**
+     * What one field of a catalogue may be used for. The three are separate capabilities rather than one membership: a
+     * listing filters on values it does not select - a certificate is found by its protocol without the listing
+     * carrying one - and orders only by the values it shows.
+     */
+    private record Capabilities(List<FilterConditionOperator> conditions, boolean displayable, boolean sortable) {
+    }
+
+    private record Catalogue(Map<CatalogueField, Capabilities> fields) {
 
         boolean isEmpty() {
             return fields.isEmpty();
@@ -325,9 +365,20 @@ public class ListViewServiceImpl implements ListViewExternalService, ListViewInt
             return fields.containsKey(field);
         }
 
+        boolean canDisplay(CatalogueField field) {
+            Capabilities capabilities = fields.get(field);
+            return capabilities != null && capabilities.displayable();
+        }
+
+        boolean canSortBy(CatalogueField field) {
+            Capabilities capabilities = fields.get(field);
+            return capabilities != null && capabilities.sortable();
+        }
+
         boolean accepts(CatalogueField field, FilterConditionOperator condition) {
-            List<FilterConditionOperator> conditions = fields.get(field);
-            return conditions != null && conditions.contains(condition);
+            Capabilities capabilities = fields.get(field);
+            return capabilities != null && capabilities.conditions() != null
+                    && capabilities.conditions().contains(condition);
         }
     }
 
--- a/src/main/java/com/otilm/core/service/impl/ListViewServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/ListViewServiceImpl.java
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.stream.Collectors;
 import org.hibernate.exception.ConstraintViolationException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.dao.DataIntegrityViolationException;
@@ -89,7 +90,7 @@ public class ListViewServiceImpl implements ListViewExternalService, ListViewInt
     public ListViewDto createView(ListViewRequestDto request) throws AlreadyExistException {
         UUID userUuid = loggedUserUuid();
         Resource resource = request.getResource();
-        validateRequest(resource, request);
+        validateRequest(resource, request, Set.of());
 
         serializeWritesFor(userUuid, resource);
         if (listViewRepository.existsByUserUuidAndResourceAndName(userUuid, resource, request.getName())) {
@@ -111,7 +112,7 @@ public class ListViewServiceImpl implements ListViewExternalService, ListViewInt
             throws NotFoundException, AlreadyExistException {
         UUID userUuid = loggedUserUuid();
         ListView view = ownView(uuid, userUuid);
-        validateRequest(view.getResource(), request);
+        validateRequest(view.getResource(), request, columnsOf(view));
 
         serializeWritesFor(userUuid, view.getResource());
         if (listViewRepository
@@ -204,11 +205,15 @@ public class ListViewServiceImpl implements ListViewExternalService, ListViewInt
         dto.setName(view.getName());
         dto.setResource(view.getResource());
         dto.setDefaultView(view.isDefaultView());
+        // A column the listing can no longer show is still returned: the client has the same catalogue this is read
+        // from, so it can mark the column unavailable and offer to remove it, and a view whose every column was
+        // withdrawn still reads back in a shape it can be saved in. Only a field that has left the catalogue outright
+        // is dropped, having nothing left to label it with.
         dto
                 .setColumns(view
                         .getColumns()
                         .stream()
-                        .filter(column -> catalogue.canDisplay(CatalogueField.of(column)))
+                        .filter(column -> catalogue.offers(CatalogueField.of(column)))
                         .toList());
         dto.setFilters(view.getFilters());
         // Returning an ordering the listing would now refuse hands the client a view whose every application answers
@@ -222,10 +227,22 @@ public class ListViewServiceImpl implements ListViewExternalService, ListViewInt
 
     /**
      * Rejects anything the resource's catalogue cannot apply, so a view is stored only in a shape the listing can
-     * actually use. A field that afterwards disappears, or stops being one the listing can show or order by, is a
-     * different case and is dropped on read instead.
+     * actually use.
+     *
+     * <p>
+     * {@code carriedAlready} are the columns the stored view holds, which are exempt from the column gate. A field can
+     * stop being one the listing shows after a view stored it, and rejecting it would leave that view unsaveable: the
+     * client reads it back, renames it, and the rename is refused over a column it did not touch. So a withdrawn column
+     * can be kept or removed but not introduced, and a creation - which carries nothing already - is held to the
+     * current catalogue in full.
+     *
+     * <p>
+     * An ordering has no such exemption. It is applied by re-issuing the listing request, which refuses a field that is
+     * not sortable, so keeping one would answer an error on every application rather than show a blank column. Such an
+     * ordering is dropped on read instead.
      */
-    private void validateRequest(Resource resource, ListViewUpdateRequestDto request) {
+    private void validateRequest(Resource resource, ListViewUpdateRequestDto request,
+            Set<CatalogueField> carriedAlready) {
         Catalogue catalogue = catalogueOf(resource);
         if (catalogue.isEmpty()) {
             throw new ValidationException(ValidationError
@@ -233,12 +250,17 @@ public class ListViewServiceImpl implements ListViewExternalService, ListViewInt
                             .formatted(resource.getCode())));
         }
 
-        validateColumns(resource, request.getColumns(), catalogue);
+        validateColumns(resource, request.getColumns(), catalogue, carriedAlready);
         validateFilters(resource, request.getFilters(), catalogue);
         validateSort(resource, request.getSort(), catalogue);
     }
 
-    private static void validateColumns(Resource resource, List<ListViewColumnDto> columns, Catalogue catalogue) {
+    private static Set<CatalogueField> columnsOf(ListView view) {
+        return view.getColumns().stream().map(CatalogueField::of).collect(Collectors.toUnmodifiableSet());
+    }
+
+    private static void validateColumns(Resource resource, List<ListViewColumnDto> columns, Catalogue catalogue,
+            Set<CatalogueField> carriedAlready) {
         Set<CatalogueField> seen = new LinkedHashSet<>();
         List<String> duplicated = columns
                 .stream()
@@ -259,6 +281,7 @@ public class ListViewServiceImpl implements ListViewExternalService, ListViewInt
 
         List<String> unshowable = columns
                 .stream()
+                .filter(column -> !carriedAlready.contains(CatalogueField.of(column)))
                 .filter(column -> !catalogue.canDisplay(CatalogueField.of(column)))
                 .map(ListViewColumnDto::getFieldIdentifier)
                 .toList();
--- a/src/main/java/com/otilm/core/util/SearchHelper.java
+++ b/src/main/java/com/otilm/core/util/SearchHelper.java
@@ -19,6 +19,7 @@ import jakarta.persistence.metamodel.Attribute;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
+import java.util.EnumSet;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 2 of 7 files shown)


src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -85,8 +128,8 @@ public static SearchFieldDataDto prepareSearch(final FilterField filterField, Ob
             values = withoutNull;
         }
         fieldDataDto.setValue(values);
-        fieldDataDto.setDisplayable(true);
-        fieldDataDto.setSortable(isSortable(filterField));
+        fieldDataDto.setDisplayable(isDisplayable(filterField));
+        fieldDataDto.setSortable(isSortableField(filterField));
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Confirmed and fixed in 1aed49b3f. `catalogueOf` did put every `FilterField` in with only its condition list, and both `validateColumns` and `validateSort` used membership, so a view could store `CERTIFICATE_PROTOCOL` as a column and — worse than the blank heading — store an ordering on it that `SortOrderBuilder.resolveField` now refuses, failing the whole listing request when the view is applied.

The catalogue entry is now a `Capabilities` record carrying the conditions, `displayable` and `sortable` separately, read off the published catalogue (`SearchHelper.isDisplayable`/`isSortableField` for properties, the `SearchFieldDataDto` flags for attributes) rather than from a second copy of the rules. Columns validate against `canDisplay`, orderings against `canSortBy`. Filters stay on membership deliberately: a listing filters on values it does not select, so `CERTIFICATE_PROTOCOL` remains a legitimate filter.

Existing rows need no migration. `toDto` drops a column that is no longer displayable, the way it already dropped one whose field had left the catalogue, and drops an ordering that is no longer sortable so such a view opens in the listing default order instead of erroring.

Covered by four new tests in `ListViewServiceITest`: `aColumnTheListingCannotShowIsRejectedOnWrite`, `anOrderingTheListingWouldRefuseIsRejectedOnWrite`, `aColumnTheListingNoLongerShowsIsSkippedOnRead`, `anOrderingTheListingNoLongerAppliesIsDroppedOnRead`, plus `aFilterOnAFieldTheListingCannotShowIsAccepted` holding the filter side open. All four fail on the parent commit (verified by reverting `src/main/java` and re-running: 4 failures of 34). The PR description now records the boundary under "Saved views are held to the same flags".

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -174,20 +217,21 @@ public static SearchFieldDataDto prepareSearchForJSON(final SearchFieldObject at
...
      *
      * <p>
-     * The flag is published per field while the ordering is wired per listing, and only the listings named here are
-     * wired. A field of any other resource is reported not sortable however orderable its path is, because a catalogue
-     * reporting {@code true} there would advertise an ordering that listing discards: the client sorts, receives the
-     * default order, and is told nothing.
+     * Both flags are published per field while the pipeline is wired per listing, and only the listings named here are
+     * wired. A field of any other resource is reported neither displayable nor sortable however orderable its path is,
+     * because a catalogue reporting {@code true} there would advertise a column that listing cannot project and an
+     * ordering it discards: the client asks, receives the default columns in the default order, and is told nothing.
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Correct, and fixed in 1aed49b3f. `prepareSearchForJSON` already had the `resource` in hand for `sortable`, and `displayable` was ignoring it, so an ordinary custom, data or metadata attribute of an unwired resource such as `LOCATION` stayed advertised as a column although that listing never invokes `AttributeColumnProjector`.

`isDisplayable(SearchFieldObject, Resource)` now applies `CONFIGURABLE_COLUMN_RESOURCES` first, and `isSortable` reads it rather than repeating the resource check — the two flags now come from one gate on both the property and the attribute path.

Asserted by `SearchHelperColumnFlagsTest.noAttributeOfAnUnwiredListingIsOfferedAsAColumn` (TEXT/`ProtectionLevel.NONE`/visible on `LOCATION`, expecting `displayable:false`), which fails on the parent commit with `expected: <false> but was: <true>`. It sits next to the existing `noAttributeOfAnUnwiredListingIsSortable`, so the pair now pins both flags on an unwired resource.

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -32,21 +33,63 @@ public class SearchHelper {
...
+         * {@code RA_PROFILE_NAME} and {@code GROUP_NAME} are joined and present. What decides it is whether the query
+         * the listing runs selects the value, which only reading that query answers.
+         *
+         * <p>
+         * A field is listed here when the listing DTO has no property for it, or has one the list mapper leaves unset.
+         * A field the DTO does carry stays displayable even where no frontend renderer draws it yet: that is a column
+         * waiting for its cell, not a column that can never have one.
+         */
+        private static final Set<FilterField> ABSENT_FROM_LISTING = EnumSet
+                .of(
```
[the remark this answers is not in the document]
Jakub-Bartal-01: Accurate, and unchanged. The first item of the issue's Definition of Done asks for a decision "agreed with the inventory and frontend-registry owners", and no such agreement exists: the seventeen withdrawn fields are one engineer's reading of the seven listing mappers, recorded field by field in `ABSENT_FROM_LISTING` and in the PR description with the query each group was read against so a reviewer can check it, but read by one person only.

This is not something the code can settle, so nothing here changes for it. The thread stays open until those owners have confirmed the set — the certificate list in particular, where fourteen columns are withdrawn and a column somebody uses today is worse to lose than a blank one is to keep. If they identify a field that is carried after all, or one that is relied on, the set is a single `EnumSet` and `PropertyColumnDisplayabilityITest` names the expected columns per listing, so the correction is one entry in each.

src/main/java/com/otilm/core/service/impl/ListViewServiceImpl.java
```diff
@@ -208,16 +208,22 @@ private ListViewDto toDto(ListView view, Map<Resource, Catalogue> catalogues) {
                 .setColumns(view
                         .getColumns()
                         .stream()
-                        .filter(column -> catalogue.offers(CatalogueField.of(column)))
+                        .filter(column -> catalogue.canDisplay(CatalogueField.of(column)))
```
lubomirw: **[Medium]** | Correctness | Confidence: High

When every stored column becomes undisplayable, this filter returns `columns: []`. The frontend visually falls back to its standard columns, but rename and pin operations still submit the empty stored array; `ListViewUpdateRequestDto.columns` is `@NotEmpty`, so those updates return 422.

_Suggested:_ Coordinate the frontend to submit the resolved standard columns whenever `fellBackToStandard` is true, or define an API-side empty-view migration, and add a regression test for an all-withdrawn saved view.

<sub>flagged by Superpowers; verified against Core's read path, fe-administrator's `patchActive`/`toUpdateRequest`, and the interfaces validation contract</sub>

lubomirw: **[Medium]** | Correctness | Confidence: High

This drop-on-read behavior changes the saved-view preservation contract. `fe-administrator` intentionally keeps undisplayable columns as unavailable so users can inspect or remove them; fresh reads now lose that UX, while a stale page or a visibility change between read and save can submit a column that Core newly rejects with 422. This is an explicit product/contract judgment rather than an accidental implementation detail.

_Suggested:_ Choose one cross-repo contract: either preserve unavailable entries with applicability metadata, or update the frontend to accept their removal and strip them before every write.

<sub>flagged by Copilot CLI and Superpowers; verified against `fe-administrator/main` column-resolution and saved-view write helpers</sub>

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -220,16 +264,45 @@ public static boolean listingAppliesSort(final Resource resource) {
...
+    }
+
+    /**
+     * Whether a listing may be ordered by this field, which is what a requested sort is refused against.
+     *
+     * <p>
+     * Wider than the catalogue's {@code sortable} flag, and deliberately so. On a listing that publishes columns the
+     * two agree, because ordering there is triggered by clicking a column header and a field that listing cannot show
+     * has no header to click. Every other listing orders from its own code rather than from a header - the OID entries
+     * listing orders by {@code OID_ENTRY_CODE}, which no column picker ever offered - so only the path matters there.
```
lubomirw: **[Nit]** | Maintainability | Confidence: High

The Javadoc says the OID entries listing orders by `OID_ENTRY_CODE`, but `CustomOidEntryServiceImpl.listCustomOidEntries` applies a fixed descending OID sort. The supporting `SubclassFilterFieldSearchITest` invokes the repository directly rather than exercising the listing.

_Suggested:_ Describe this as a repository-level ordering path, or replace it with an example from a listing that actually applies its own requested field ordering.

<sub>flagged by Superpowers and Hygiene; verified against the OID service and integration test</sub>

src/main/java/com/otilm/core/util/SearchHelper.java
```diff
@@ -32,21 +33,63 @@ public class SearchHelper {
...
+        /**
+         * Fields that share one attribute with another field of the same resource, and so display something drawn out
+         * of it rather than the attribute itself. Derived rather than listed, so a field added later is classified
+         * without an edit here: the certificate validation-check fields are the case that exists today, each reading
+         * one serialized validation result.
+         */
+        private static final Set<FilterField> SHARING_AN_ATTRIBUTE = fieldsSharingAnAttribute();
+
+        /**
+         * Fields of a configurable-column listing whose value that listing does not carry. Offering one as a column
```
lubomirw, line 55: **[Nit]** | Maintainability | Confidence: High

Several rationale comments are substantially longer than their contracts: `ABSENT_FROM_LISTING`, `CONFIGURABLE_COLUMN_RESOURCES`, attribute displayability, and the property-column pin test. This was classified as likely noise because the issue explicitly requires recording non-derivable mapper decisions, but the inline narration is still worth a maintainability judgment.

_Suggested:_ If the team prefers terse inline documentation, retain the mapper/source invariants and trim the surrounding decision narrative at `SearchHelper.java:54`, `SearchHelper.java:220`, `SearchHelper.java:340`, and `PropertyColumnDisplayabilityITest.java:18`.

<sub>flagged by Hygiene; verified against the issue's requirement to record the manual classification</sub>

lubomirw (commented):
Automated multi-reviewer triage — 4 findings (COMMENT).

src/main/java/com/otilm/core/service/impl/ListViewServiceImpl.java
Jakub-Bartal-01: Confirmed and fixed in f9a4f1e54.

Reproduced first: `aViewWhoseEveryColumnWasWithdrawnStillReadsBackASaveableShape` stores a certificate view whose columns are `CERTIFICATE_PROTOCOL` and `KEY_USAGE`, both withdrawn by this PR, and on the parent commit the read answers `expected: <[CERTIFICATE_PROTOCOL, KEY_USAGE]> but was: <[]>`. `ListViewUpdateRequestDto.columns` does carry `@NotEmpty`, and `ExceptionHandlingAdvice` maps `MethodArgumentNotValidException` to 422, so the rename that follows is refused. The frontend side holds too: `toUpdateRequest` seeds the body from the stored row (`columns: view.columns`) and `patchActive` routes rename and pin through it, and the only non-empty guard in that repo is inside `ColumnPicker.handleSave`, which the rename and pin paths do not pass through. The visual fallback to the standard set is render-only.

Rather than coordinate the frontend to substitute the standard columns, the read no longer produces the empty list. A read returns every column whose field the catalogue still offers and drops only a field that has left it outright - the omission `ListViewDto.columns` already documents. So the filter can no longer empty a view that was stored non-empty, and the shape a read returns is a shape a write accepts, which is the invariant that was broken.

The write gate moved with it, since preserving the column means the client resubmits it: `validateColumns` now exempts the columns the stored view already carries, so a withdrawn column can be kept or removed but not introduced. A creation carries nothing already and is still held to the current catalogue in full.

No API-side migration, and no frontend change is required.

Three tests cover it, all failing on the parent commit (2 failures and 1 error of 37): the one above, `aWithdrawnColumnTheViewAlreadyCarriesSurvivesARename`, and `aColumnTheListingNoLongerShowsIsKeptOnRead`. `aWithdrawnColumnCannotBeAddedToAnExistingView` pins that the exemption is not wider than the columns the view holds.

One case is untouched and predates this PR: a view whose every column has left the catalogue entirely still reads back empty. The frontend strips those from every write itself (`toStorableColumns` filters against the raw catalogue), so it could not save such a view either way, and closing it would mean returning identifiers with nothing left to label them or relaxing `@NotEmpty` in interfaces.

Verified with JDK 21: `spotless:apply`, then `spotless:check checkstyle:check` clean, and the four CI profiles green - `test-non-integration` 4530, `test-integration-core` 1327, `test-integration-service-1` 1113, `test-integration-service-2` 1155.

Jakub-Bartal-01: Correct on both halves, and fixed in f9a4f1e54. I read the frontend before deciding, and it settles the choice rather than leaving it open.

`fe-administrator` is already built for the preserving contract, so there is nothing to define. `resolveColumns` (`src/utils/columnPicker.ts`) resolves a stored column against the displayable-filtered catalogue and returns `{ ...column, available: false }` when it does not match, keeping it at its stored index; `resolveView` counts those into `fellBackToStandard`; `UnresolvedColumnsNotice` names them ("Expiry and Cost centre cannot be shown, so this view is showing 3 of its 5 columns") and offers Review columns; `SelectedColumnRow` renders the row with an Unavailable badge, no drag handle, no rename, and a remove button. `toStoredColumnsKeepingUnavailable` then carries them through a save that is not a column edit. Since a field that has left the catalogue outright was already omitted on read, the `displayable: false` column was the only remaining case that machinery had left to act on - so drop-on-read did not degrade the UX, it removed its last input.

So the contract is the first of your two options, minus the metadata: the read preserves the entry, and the client derives applicability from the catalogue it already fetches for the picker. Adding a flag to `ListViewColumnDto` would be an interfaces change publishing something the client can compute.

The second half - the stale page and the visibility change between read and save - is fixed by the same commit and was the sharper of the two. Preserving the column means the client resubmits it (`toStorableColumns` filters against the raw catalogue, so a catalogued-but-undisplayable column survives the write), so the write gate had to move or every such view would be unrenamable. `validateColumns` now exempts the columns the stored view carries: a withdrawn column can be kept or removed, never introduced, and a creation is held to the current catalogue in full. That closes the race in the direction that matters - the request is accepted rather than 422 - without letting a client author a blank column.

The ordering keeps both its strict write gate and its drop on read, and the asymmetry is deliberate: an undisplayable column costs one blank heading, while an ordering the listing refuses costs every application of the view. `sort: null` is always a legal write shape, so that path round-trips too.

Covered by `aColumnTheListingNoLongerShowsIsKeptOnRead` and `aWithdrawnColumnTheViewAlreadyCarriesSurvivesARename` (both failing on the parent commit), with `aWithdrawnColumnCannotBeAddedToAnExistingView` holding the exemption to the columns the view holds. No `fe-administrator` change is needed; the PR description now records the contract.

Verified with JDK 21: `spotless:check checkstyle:check` clean and all four CI profiles green (4530 / 1327 / 1113 / 1155).

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Confirmed and fixed in f9a4f1e54. The javadoc was wrong twice over: `CustomOidEntryServiceImpl.listCustomOidEntries` passes `(root, cb) -> cb.desc(root.get(CustomOidEntry_.oid))` as the default order, and its `RequestValidatorHelper.revalidateSearchRequestDto(request, Resource.OID)` refuses a requested sort outright, since `OID` is not a `CONFIGURABLE_COLUMN_RESOURCES` member. So no OID listing request reaches `resolveField` at all, and the listing is not what the widened predicate serves.

What does reach it is a caller naming a `SortSpecification` against the repository directly, which is exactly what `SubclassFilterFieldSearchITest:141` does. Checked the other listings that skip the resource-aware overload for the same mistake: `CryptographicAssetServiceImpl` also passes a fixed ordering lambda, and the rest are pipeline resources, so the repository path is the only one.

The javadoc on `isOrderableOnListing` now describes it as the repository-level path and says so in those terms. `PropertyColumnDisplayabilityITest` repeated the same claim, so its javadoc is corrected too and the test is renamed `aFieldNoPickerOffersCanStillOrderARepositoryQuery` - the assertions are unchanged, since the predicate itself was right.

src/main/java/com/otilm/core/util/SearchHelper.java
Jakub-Bartal-01: Trimmed in f9a4f1e54, keeping the mapper and source invariants and cutting the narration around them. Roughly 30 lines of comment removed across the four sites.

- `ABSENT_FROM_LISTING`: three paragraphs to two. The rule for membership stays, as does the reason the set is listed rather than derived, since that is what stops someone replacing it with a predicate - `CERTIFICATE_PROTOCOL` is joined and absent while `RA_PROFILE_NAME` and `GROUP_NAME` are joined and present. The per-group comments naming each mapper and projection are untouched: they are the non-derivable record the issue asks for, and a reader cannot re-derive them without opening seven mappers.
- `CONFIGURABLE_COLUMN_RESOURCES`: dropped the walk-through of what a wrong `true` looks like to the client, keeping the clause that says it advertises a column the listing cannot project. The "adds itself here" invariant stays - it is the footgun.
- Attribute displayability: kept the four withheld content kinds, one clause each, and cut the explanatory asides.
- `PropertyColumnDisplayabilityITest`: cut the paragraph restating why displayability is judged per field, which `SearchHelper` already says. The red-on-a-new-field invariant stays, and so does the one line explaining why it needs a persistence context, since that is the reason it extends `BaseSpringBootTest`.

Also removed two duplicated sentences on `isSortable` and `isDisplayable` ("Decided here rather than by a pass over the assembled catalogue...") - same claim in both places, and pure decision narrative.

No behaviour touched; `spotless:check checkstyle:check` clean and all four CI profiles green.

lubomirw (approved):

Issue #2209, opened 2026-09-03
Recount per-CBOM cryptographic statistics from the walked components during asset ingest


karolhrdina:
### Description

The counts on the `cbom` table come from the repository's `cryptoStats` feed entry. OmniTrustILM/cbom-repository#138 decided (section 11 of the [decision document](https://github.com/OmniTrustILM/cbom-repository/blob/feat/138-search-hardening-change-feed/docs/design/2026-09-02-change-feed-decision.md)) that the repository will not rewrite existing objects: objects uploaded before the counting fix keep a shallow count (nested components not counted), new uploads carry `crypto-stats-version: 2` in S3 metadata, and that marker is not on the JSON wire. Core therefore cannot tell a legacy count from a current one and must treat every repository count as advisory.

#2073 walks every component of every CBOM (iterative, depth-first, document order, depth at most 1000, `components` tree only, as in `CbomAssetExtractor` and `DocumentScope.walk`). The header counts are a by-product of that walk. This issue recounts them there and refreshes the `cbom` table.

### Acceptance Criteria

- [ ] During the per-CBOM unit of work, the counts are computed from the walked components with the same rules as the repository's version-2 count: `type == cryptographic-asset` with a non-null `cryptoProperties`, `components` tree only (not `metadata.component`, `metadata.tools`, `formulation`, `services`), depth at most 1000.
- [ ] The computed counts replace the repository-provided counts on the `cbom` row for that CBOM; the repository values are used only until the first ingest of that CBOM has run.
- [ ] `cryptoStats: null` from the feed (see #2208) leaves the counts unset until ingest, with no error.
- [ ] The statistics and sync-visibility endpoints shipped by #2145 read the recounted values.
- [ ] Test: a CBOM whose nested components hold three cryptographic assets under a library, fed with a legacy shallow count of one, ends with three after ingest.
- [ ] The ticket states whether this lands inside #2073 or as a follow-up immediately after it.

### Out of scope

Any rewrite of repository objects or metadata (deferred operator tool in cbom-repository); changes to the repository's counting rules; PQC verdicts (#2151).

Related: #2073 (part of, or adjacent to), #2072, OmniTrustILM/cbom-repository#138 (decision section 11), #2208.

karolhrdina:
[...]

Issue #2073, opened 2026-08-14
Extend CBOM sync with per-CBOM asset onboarding and deletion lifecycle


karolhrdina:
### Description

Extend CbomSyncTask with asset onboarding as a single per-CBOM unit of
work — no separate header and asset phases (ilm#299, design doc section 7):

- Discovery via the paged cbom-repository cursor feed Core already walks
  (core#2208, PR core#2245); "original" pseudo-version filtered;
  tombstoned (serialNumber, version) pairs and already-known rows skipped.
- Per CBOM, bounded per run: fetch once (outside any transaction) → parse,
  normalize, redact, evaluate rules in memory → writer transactions:
  header upsert (state SYNCING) → batched asset/link upserts
  (READ COMMITTED, migration-defined arbiters) → per-URN supersede diff
  with orphan GC and source_count maintenance → state SYNCED. Crash leaves
  SYNCING/FAILED; next run redoes the unit idempotently (epic AC13).
- Cluster safety: new ClusterOperationSynchronizer.Operation
  CBOM_ASSET_SYNC held by a thin outer transaction while writer
  transactions commit batches (SigningRecordRetentionSweeper pattern).
- Deletion lifecycle: deleteCbom / bulkDeleteCbom extend to batched link
  GC → orphan-asset GC → header delete → tombstone insert; weekly
  tombstone-aware reconciliation re-list over the same paged feed.
- Unified clean-skip when the repository is unconfigured, unreachable or
  absent (fixes the SaaS 404 stack traces, core#1330); buffer-drift guard
  comparing fetch-buffer expectations at run start; failure reasons stored
  operator-visibly; JMS listener occupancy documented for operators.
- First commit strengthens the thin sync test baseline (current ITest pins
  only the task wrapper) so compatibility deltas are enumerated and pinned.

### Acceptance Criteria

- [ ] Idempotence: re-run over an unchanged repository creates zero new
      rows (epic AC9).
- [ ] Resumability: kill mid-unit leaves consistent state; next run
      completes the remainder (epic AC13).
- [ ] Supersede: a new version of a URN removes vanished source links,
      GCs orphaned assets, keeps counts consistent.
- [ ] User-deleted CBOMs stay deleted across hourly sync AND weekly
      reconciliation (tombstone test).
- [ ] Header behavior compatible except the enumerated deltas (404 clean
      skip; "original" handled deliberately instead of erroring; additive
      sync-state fields); strengthened baseline tests pass.
[...]

PR #2245, opened 2026-09-09
Consume the paged cbom-repository feed with bounded retry of skipped entries

### LINKED ISSUE #2208: Consume paged cbom-repository search and the corrected feed contract (limit, created_at, null cryptoStats)
### Description

cbom-repository 0.2.0 (OmniTrustILM/cbom-repository#138, deliverable 1, branch `feat/138-search-hardening-change-feed`, PR pending) changes the search contract that `CbomSyncTask` and `CbomServiceImpl.sync()` consume through `CbomRepositoryClient`:

- `GET /api/v1/bom?after=` accepts an optional `limit` (1 to 1000). With `limit`, entries are sorted by the store's `LastModified` then object key, and the client protocol is: advance `after` to the last `created_at` within a run, stop on a page shorter than `limit`, and between runs use `watermark = run_start - overlap`. Core already does the last step (last successful run start minus 60 s).
- The entry field is `created_at`, not `timestamp`. `BomEntryDto.timestamp` is never populated today.
- Objects with missing or invalid statistics become visible with `cryptoStats: null` and `warnings: ["crypto-stats-missing"]` or `["crypto-stats-invalid"]` instead of being hidden. If Core dereferences `cryptoStats` without a null check, the entry throws, is counted as skipped, and is never offered again because the watermark moves past it at the next successful run. Baseline: today the repository drops such objects, so Core never saw them; a null-dereference skip leaves the identical end state (Core catches per-entry failures and the run still counts as successful), so there is no regression, and for invalid metadata the change is a strict improvement because one bad object no longer fails the whole feed with 500. Handling the null so that the entry is recovered is this issue's job.
- `version` can be the string `original`; Core already filters `original` out.
- The repository's statistics are advisory: legacy objects hold a shallow count and the JSON does not say which (the `crypto-stats-version` marker is S3 metadata only). The recount during ingest is a separate issue (see Related).

The [decision document](https://github.com/OmniTrustILM/cbom-repository/blob/feat/138-search-hardening-change-feed/docs/design/2026-09-02-change-feed-decision.md) (section 13, question 3) recommends keeping the watermark rule `last successful run start - overlap`; this issue assumes that choice unless the Tech Lead picks the alternative.

### Acceptance Criteria

- [ ] `CbomRepositoryClient` sends `limit` (proposed value 1000) and iterates pages per the documented protocol; a short page terminates the run; an empty first page is a valid empty run.
- [ ] `BomEntryDto` maps `created_at`; the dead `timestamp` field is removed or annotated as the JSON name `created_at`.
- [ ] `cryptoStats: null` and `warnings` are tolerated: no exception; the warning codes are logged with the entry identity; the entry is still stored, with counts left for the ingest recount or computed from the fetched document.
- [ ] A skipped entry is not lost silently: skipped identities `(serialNumber, version)` are recorded with the failure reason and retried on the next 3 runs (default, configurable); after that they are recorded as permanently skipped and the watermark advances as today. A run with skipped entries still counts as successful, so one permanently failing entry (a corrupt document, a Core validation bug) can never freeze the watermark.
- [ ] The watermark rule stays `last successful run start - overlap`; the overlap is configurable with a default of 60 s and is documented.
- [ ] Deduplication on `(serialNumber, version)` is unchanged.
- [ ] Tests against a stubbed repository cover: multiple pages, short-page termination, `cryptoStats: null` with each warning code, `version: "original"`, and the bounded retry of skipped identities (retried on 3 runs, then permanently skipped).
- [ ] When cbom-repository ships the cursor (OmniTrustILM/cbom-repository#144), adopting `cursor` and `Link` is a one-line follow-up noted on the ticket, not a blocker.

### Out of scope

Asset ingest and tombstones (#2073); the recount itself (separate issue); any webhook receiver (OmniTrustILM/cbom-repository#26).

Related: OmniTrustILM/cbom-repository#138, #2073, ilm#299.

karolhrdina:
Core consumes the cbom-repository 0.3.0 search feed the way the contract now describes it, and stops losing entries it could not store. Closes #2208. **25 files, one commit**, rebased onto `main` after #2236 and #2228. Follow-ups: #2249, #2250, #2251.

## What changes on the wire

`CbomServiceImpl.sync()` opens a run with `GET /api/v1/bom?after=<watermark>&limit=1000` and follows
`Link: <bom?cursor=…&limit=…>; rel="next"` until a page carries no `Link` — the absence of the header,
not the page size, ends the run. The target is a relative-path reference, so `CbomRepositoryClient`
appends only its query to Core's own `/api/v1/bom` URL (the README's recipe for a client that builds URLs
from a configured base): host, mount prefix and any ingress rewrite stay Core's, and the cursor token is
forwarded byte for byte (`search_followsALinkCursorThatContainsBase64PaddingVerbatim` pins padding and
percent-escapes). An absolute, path-absolute, network-path, foreign-path or query-less target, a target
that smuggles `after` next to `cursor` (compared decoded, so `%61fter` counts), and an immediately repeated
cursor are refused with a Core-shaped 502 whose detail names none of the target — an absolute one carries
the repository's own host, and the detail reaches a REST caller of the sync endpoint as well as the job
result; the target goes to the Core log. A failed page request likewise keeps its status but not the
repository's own sentence. The `Link` header itself is parsed by a small hand-walked RFC 8288 scanner
(`LinkHeader`), quoted parameters included, with no regex recursion to overflow on a long header.

`BomEntryDto` maps `created_at` (the dead `timestamp` is gone), `cryptoStats` is nullable and `warnings`
is read. The watermark rule is unchanged — start of the last successful run minus the overlap — but the
overlap is now `cbom.sync.overlap` (default 60 s, `@DurationUnit(SECONDS)` so `CBOM_SYNC_OVERLAP=60`
means seconds). Any non-2xx to a page request fails the run, so the watermark holds and the whole run is
repeated, exactly as the contract asks for a replica that does not know `cursor` yet.

## Nothing is skipped silently any more

An entry the sync cannot store — its document unreadable, the repository refusing it, a Core validation
refusing the stored document, the database refusing the row for anything but the `(serialNumber,
version)` uniqueness — is recorded in the new `cbom_sync_skip` table with an operator-safe reason and
the counts the feed reported, and retried at the end of each of the next `cbom.sync.skipped-retry-runs`
runs (default 3). After that it is kept as `PERMANENTLY_SKIPPED`, the watermark advances as today, and
the run still counts as successful — one corrupt document cannot hold the inventory back. A run loads
only the `RETRYING` rows (the live retry set; written-off rows are looked up by identity when an entry
fails or is stored, so the per-run load does not grow with every document ever written off). A skip
record is deleted the moment the document is stored, whether the retry phase stored it or the overlap
re-offered it, and whichever state it is in. `cryptoStats: null` entries are stored with zero counts, left for the ingest recount (#2209), with the
warning codes logged against the entry identity; `crypto-stats-shallow`/`-truncated` keep the reported
counts. `version: "original"` is ignored without a read and without a skip record.

Server-side read failures need one more rule, found by the adversarial review: Core maps its own refused
connections and the 35 s response timeout to HTTP 503, so "503 fails the run" would let a single hanging
document freeze the watermark. A 503 is therefore deferred: it is charged to its document like any other
failure unless the run shows a repository-wide outage — no read succeeded at all and at least two of the
**feed pass's** documents failed that way — in which case the run is skipped (503 →
`ScheduledJobSkippedException`) and nothing is charged. A retry's failure never counts towards that
verdict: the page requests of the same run just succeeded, and a retry exempt from being charged would be
exempt in every quiet run to come — skipped hourly, watermark never recorded, budget never spent — which
is the unbounded retry the table exists to prevent (review finding; `aPermanentlyHangingDocument…` pins
it). The status that drives the decision is the HTTP response's, never the one a problem-detail body
claims. Every other status, 5xx included, is that document's verdict and is charged at once.

## Transaction boundaries

`sync()` and `syncAuthorized()` run `@Transactional(NOT_SUPPORTED)`, and so does the scheduler's
`CbomSyncTask.performJob` (as `DiscoveryCertificateTask` already did): the run pages an external service
and reads one document per entry, which must not sit inside the platform's 120 s default transaction.
Entry stores keep their `REQUIRES_NEW` transaction; skip bookkeeping goes through the REQUIRED
`CbomSyncSkipWriter`, whose native upsert (`ON CONFLICT (serial_number, version)`) carries the attempt
arithmetic and uses the unique constraint as the concurrency arbiter. The `Cbom` entity now names
`cbom_serial_version_unique` (the migration already did), so the entity-generated test schema carries the
constraint the sync classifies duplicates by. Loop guards: a run-level cursor set, "an empty page never
carries a `Link`", and a 10 000-page ceiling — each fails the run with a Core-shaped 502.

`CbomSyncTask` no longer forwards `e.getMessage()` to the scheduler result for arbitrary exceptions;
only a `PlatformException`'s own sentence gets through, anything else becomes "unexpected error, see the
Core log".

## Configuration

`application.yml` gains `cbom.sync.page-size` (`CBOM_SYNC_PAGE_SIZE`, 1000), `cbom.sync.overlap`
(`CBOM_SYNC_OVERLAP`, 60 s) and `cbom.sync.skipped-retry-runs` (`CBOM_SYNC_SKIPPED_RETRY_RUNS`, 3),
bound and validated by `CbomSyncProperties`. Both reviews made the same point — a SaaS operator can reach
neither yml nor an environment variable, and `skipped-retry-runs` is operator policy — so **#2249** moves all
three into the Settings UI, which also retires the env-var placeholders and the Helm wiring they would need.
This PR keeps the yml placement until then.

## Prerequisites and follow-ups (handled outside this PR)

- Every cbom-repository replica must run **0.3.0 or newer**: Core always sends `limit` and ends the run on a
  missing `Link`, so a 0.2.0 replica would truncate a run silently. Core WARNs when the opening page is
  full and `Link`-less, which is that signature.
- `SchedulerListener`'s class-level `jakarta.transaction.Transactional` still keeps the job-history commit
  inside the 120 s transaction (pre-existing): a successful run longer than that leaves no SUCCESS row and
  the watermark waits for a run that fits. Filed as **#2251**.
- `cbom_sync_skip` has no retention and no operator view or reset yet: **#2250** adds a sweep of
  `PERMANENTLY_SKIPPED` rows after a retention period, a UI that names the document and its reason with a
  retry, everything configurable from the Settings UI. #2073 builds its per-CBOM unit of work on this loop;
  #2209 recounts the zero counts.

## Tests

Against a WireMock repository speaking the 0.3.0 contract, with raw JSON bodies: `LinkHeaderTest` 22,
`CbomRepositoryClientTest` 20, `CbomHeaderCountsTest` 3, `CbomSyncPropertiesTest` 7,
`CbomSyncWatermarkTest` 4, new `CbomSyncITest` 28 (three-page run, empty first page, 400 on a cursor page,
cursor loop, empty page with `Link`, null stats per warning code, `original`, the five-run retry budget,
recovery with the reported counts, duplicate re-offers, the outage rule and the hanging document that
must not become one, a written-off row resolved by a late store, writer-level upsert transitions),
`CbomServiceITest` 45, `CbomSyncTaskITest` 9, plus the ArchUnit and context-budget guards
(`ContextSignatureGuardTest` stays at 61). Spotless and Checkstyle clean.

### DIFF
--- a/src/main/java/com/otilm/core/cbom/client/CbomRepositoryClient.java
+++ b/src/main/java/com/otilm/core/cbom/client/CbomRepositoryClient.java
@@ -4,14 +4,20 @@ import com.otilm.api.exception.CbomRepositoryException;
 import com.otilm.api.model.core.cbom.CbomUploadRequestDto;
 import com.otilm.api.model.core.settings.PlatformSettingsDto;
 import com.otilm.api.model.core.settings.SettingsSection;
+import com.otilm.core.config.CbomSyncProperties;
 import com.otilm.core.model.cbom.BomCreateResponseDto;
 import com.otilm.core.model.cbom.BomEntryDto;
 import com.otilm.core.model.cbom.BomResponseDto;
 import com.otilm.core.model.cbom.BomSearchRequestDto;
 import com.otilm.core.model.cbom.BomVersionDto;
 import com.otilm.core.settings.SettingsCache;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Optional;
 import java.util.function.Function;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
@@ -23,6 +29,7 @@ import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ProblemDetail;
+import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Component;
 import org.springframework.web.reactive.function.client.ClientResponse;
 import org.springframework.web.reactive.function.client.ExchangeFilterFunction;
@@ -66,21 +73,150 @@ public class CbomRepositoryClient {
                 .getBody(), request);
     }
 
-    public List<BomEntryDto> search(final BomSearchRequestDto query) throws CbomRepositoryException {
-        String baseUrl = getCbomRepositoryBaseUrl();
+    /** The last path segment of {@link #CBOM_SEARCH}; a Link target may name it or leave the path empty. */
+    private static final String SEARCH_SEGMENT = "bom";
+
+    /** How much of an unusable Link target is quoted in the Core log. */
+    private static final int LINK_TARGET_ECHO_LIMIT = 200;
+
+    /**
+     * Opens a search run: documents created strictly after {@code query.after} (null reads as 0), the first page when
+     * {@code query.limit} is set, the whole unpaged legacy listing when it is not. Follow pages with
+     * {@link #nextPage(BomSearchPage)} until {@link BomSearchPage#hasNext()} is false; the absence of the repository's
+     * {@code Link rel="next"} header -- not the page size -- says the run is complete.
+     */
+    public BomSearchPage search(final BomSearchRequestDto query) throws CbomRepositoryException {
+        final Integer limit = query.getLimit();
+        if (limit != null && (limit < CbomSyncProperties.MIN_PAGE_SIZE || limit > CbomSyncProperties.MAX_PAGE_SIZE)) {
+            throw new IllegalArgumentException("CBOM Repository search limit must be within %d..%d, was %d"
+                    .formatted(CbomSyncProperties.MIN_PAGE_SIZE, CbomSyncProperties.MAX_PAGE_SIZE, limit));
+        }
+        final String baseUrl = getCbomRepositoryBaseUrl();
+        final UriComponentsBuilder builder = UriComponentsBuilder
+                .fromUriString(baseUrl)
+                .path(CBOM_SEARCH)
+                .queryParam("after", query.getAfter() == null ? 0L : query.getAfter());
+        if (limit != null) {
+            builder.queryParam("limit", limit);
+        }
+        return fetchPage(builder.build().toUri());
+    }
+
+    /** Fetches the page a previous page's {@code Link rel="next"} header pointed at. */
+    public BomSearchPage nextPage(final BomSearchPage page) throws CbomRepositoryException {
+        if (!page.hasNext()) {
+            throw new IllegalArgumentException("The page carries no next link; the run is complete");
+        }
+        return fetchPage(page.nextPage());
+    }
+
+    private BomSearchPage fetchPage(final URI pageUri) throws CbomRepositoryException {
         final WebClient.RequestBodyUriSpec request = client.method(HttpMethod.GET);
-        return processRequest(r -> r
-                .uri(uriBuilder -> UriComponentsBuilder
-                        .fromUriString(baseUrl)
-                        .path(CBOM_SEARCH)
-                        .queryParam("after", query.getAfter())
-                        .build()
-                        .toUri())
-                .retrieve()
-                .toEntity(new ParameterizedTypeReference<List<BomEntryDto>>() {
-                })
-                .block()
-                .getBody(), request);
+        final ResponseEntity<List<BomEntryDto>> response;
+        try {
+            response = processRequest(
+                    r -> r.uri(pageUri).retrieve().toEntity(new ParameterizedTypeReference<List<BomEntryDto>>() {
+                    }).block(), request);
+        } catch (CbomRepositoryException e) {
+            throw pageRequestFailed(e);
+        }
+        final Optional<String> target = LinkHeader.nextTarget(response.getHeaders().get(HttpHeaders.LINK));
+        final URI next = target.isPresent() ? resolveNextPage(pageUri, target.get()) : null;
+        return new BomSearchPage(response.getBody(), next);
+    }
+
+    /**
+     * A failed page request keeps its status -- the sync run classifies a 503 by it, and any other status fails the run
+     * -- but not the repository's own {@code detail}: a failed page fails the whole run, whose message an operator
+     * reads in the scheduler's job result, and that sentence is chosen by the other side. It is logged for the Core log
+     * and replaced here with Core's own.
+     */
+    private static CbomRepositoryException pageRequestFailed(final CbomRepositoryException failure) {
+        final ProblemDetail reported = failure.getProblemDetail();
+        if (reported == null) {
+            return failure;
+        }
+        logger
+                .warn("CBOM Repository failed a page request with HTTP {}: {}", reported.getStatus(),
+                        reported.getDetail());
+        final ProblemDetail shaped = ProblemDetail.forStatus(reported.getStatus());
+        shaped.setDetail("CBOM Repository failed a page request (HTTP " + reported.getStatus() + ")");
+        return new CbomRepositoryException(shaped);
+    }
+
+    /**
+     * The repository sends the next page as a relative-path reference, {@code bom?cursor=..&limit=..}. Only its query
+     * is used: it is appended to the URL that was just requested, so the host, the mount prefix and any path an ingress
+     * rewrites stay Core's own -- the README names this as the way for a client that builds URLs from a configured
+     * base. An absolute or path-absolute target, a path other than the search segment, a target without a query, or a
+     * query that names {@code after} (the contract forbids combining a cursor with it) means the contract changed under
+     * Core; the run fails rather than following it.
+     */
+    static URI resolveNextPage(final URI requested, final String target) throws CbomRepositoryException {
+        final URI reference;
+        try {
+            reference = new URI(target);
+        } catch (URISyntaxException e) {
+            throw unusableLink(target);
+        }
+        final boolean relativeToSearch = reference.getScheme() == null && reference.getRawAuthority() == null
+                && !reference.getRawSchemeSpecificPart().startsWith("//")
+                && (reference.getRawPath().isEmpty() || SEARCH_SEGMENT.equals(reference.getRawPath()));
+        if (!relativeToSearch || StringUtils.isBlank(reference.getRawQuery())
+                || namesAfterParameter(reference.getRawQuery())) {
+            throw unusableLink(target);
+        }
+        // The cursor is opaque and may itself carry characters such as the '=' of base64 padding. Decomposing the
+        // query into name/value pairs -- as UriComponentsBuilder#replaceQuery followed by build(true) does, to
+        // validate each value against Type.QUERY_PARAM -- rejects such a legal value with an unchecked
+        // IllegalArgumentException instead of the 502 below. Appending the raw query verbatim to the request URL's
+        // own base (its query stripped) keeps the token exactly as the repository sent it.
+        final String base = UriComponentsBuilder.fromUri(requested).replaceQuery(null).build(true).toUriString();
+        final URI next;
+        try {
+            next = URI.create(base + "?" + reference.getRawQuery());
+        } catch (IllegalArgumentException e) {
+            throw unusableLink(target);
+        }
+        if (next.equals(requested)) {
+            throw new CbomRepositoryException(ProblemDetail
+                    .forStatusAndDetail(HttpStatus.BAD_GATEWAY, "CBOM Repository repeated the page cursor"));
+        }
+        return next;
+    }
+
+    /**
+     * The contract forbids combining a cursor with {@code after}; a Link that does anyway is not safe to follow. The
+     * name is compared decoded, because the server decodes it before it reads it: {@code %61fter} is {@code after}, and
+     * comparing the raw text would let a target carry the parameter past this guard. The escapes are already known to
+     * be well formed -- the target parsed as a URI, which rejects the rest.
+     */
+    private static boolean namesAfterParameter(final String rawQuery) {
+        for (final String pair : rawQuery.split("&")) {
+            final int equals = pair.indexOf('=');
+            final String rawName = equals >= 0 ? pair.substring(0, equals) : pair;
+            if ("after".equals(URLDecoder.decode(rawName, StandardCharsets.UTF_8))) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * The target goes to the Core log, not into the detail: it is server-supplied text of unbounded length, an absolute
+     * one names the repository's own host and mount path, and the detail reaches more than the scheduler's job result
+     * -- the sync endpoint's REST caller gets it through the error advice too. Only its first
+     * {@value #LINK_TARGET_ECHO_LIMIT} characters are logged.
+     */
+    private static CbomRepositoryException unusableLink(final String target) {
+        if (logger.isWarnEnabled()) {
+            logger
+                    .warn("CBOM Repository returned an unusable Link header for the next page: {}",
+                            StringUtils.abbreviate(target, LINK_TARGET_ECHO_LIMIT));
+        }
+        return new CbomRepositoryException(ProblemDetail
+                .forStatusAndDetail(HttpStatus.BAD_GATEWAY,
+                        "CBOM Repository returned an unusable Link header for the next page"));
     }
 
     public BomResponseDto read(final String urn, final Integer version) throws CbomRepositoryException {
@@ -155,21 +291,20 @@ public class CbomRepositoryClient {
 
     public static Mono<ClientResponse> handleHttpExceptions(final ClientResponse clientResponse) {
         if (clientResponse.statusCode().isError()) {
-            return clientResponse
-                    .bodyToMono(ProblemDetail.class)
-                    .flatMap(problemDetail -> Mono.<ClientResponse>error(new CbomRepositoryException(problemDetail)))
-                    .switchIfEmpty(Mono.defer(() -> {
-                        ProblemDetail pd = ProblemDetail.forStatus(clientResponse.statusCode());
-                        return Mono.error(new CbomRepositoryException(pd));
-                    }))
-                    .onErrorResume(e -> {
-                        if (e instanceof CbomRepositoryException) {
-                            return Mono.error(e);
-                        }
-                        return Mono
-                                .error(new CbomRepositoryException(
-                                        ProblemDetail.forStatus(clientResponse.statusCode())));
-                    });
+            return clientResponse.bodyToMono(ProblemDetail.class).flatMap(problemDetail -> {
+                // The response's status is the one the callers classify by, not whatever the body says: a body
+                // that is JSON but no problem detail decodes with status 0, and a body may name another status.
+                problemDetail.setStatus(clientResponse.statusCode().value());
+                return Mono.<ClientResponse>error(new CbomRepositoryException(problemDetail));
+            }).switchIfEmpty(Mono.defer(() -> {
+                ProblemDetail pd = ProblemDetail.forStatus(clientResponse.statusCode());
+                return Mono.error(new CbomRepositoryException(pd));
+            })).onErrorResume(e -> {
+                if (e instanceof CbomRepositoryException) {
+                    return Mono.error(e);
+                }
+                return Mono.error(new CbomRepositoryException(ProblemDetail.forStatus(clientResponse.statusCode())));
+            });
         }
         return Mono.just(clientResponse);
     }
--- a/src/main/java/com/otilm/core/config/CbomSyncProperties.java
+++ b/src/main/java/com/otilm/core/config/CbomSyncProperties.java
@@ -1,0 +1,50 @@
+package com.otilm.core.config;
+
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.bind.DefaultValue;
+import org.springframework.boot.convert.DurationUnit;
+
+/**
+ * Tuning for the CBOM header sync against cbom-repository, bound from {@code cbom.sync.*} and validated at startup.
+ *
+ * <p>
+ * These are deployment-time protocol parameters, not operator settings: the page size is bounded by the repository's
+ * contract, and the overlap has to cover clock skew between Core and the object store plus the longest upload the
+ * deployment expects (S3 stamps a multipart upload with its initiation time). The repository documents 60 s.
+ *
+ * @param pageSize the {@code limit} sent with {@code GET /api/v1/bom}; 1..1000 per the contract, 1000 by default
+ * @param overlap how far behind the last successful run's start the next run lists from; 60 s by default. A bare number
+ * binds as seconds: without {@code @DurationUnit} Spring reads it as milliseconds, so {@code 60} would silently mean 60
+ * ms and the overlap would vanish
+ * @param skippedRetryRuns how many later runs retry an entry that could not be stored before it is recorded as
+ * permanently skipped; 3 by default, 0 writes an entry off at its first failure
+ */
+@ConfigurationProperties(prefix = "cbom.sync")
+public record CbomSyncProperties(@DefaultValue("1000") int pageSize,
+        @DefaultValue("60s") @DurationUnit(ChronoUnit.SECONDS) Duration overlap,
+        @DefaultValue("3") int skippedRetryRuns) {
+
+    public static final int MIN_PAGE_SIZE = 1;
+    public static final int MAX_PAGE_SIZE = 1000;
+
+    public CbomSyncProperties {
+        if (pageSize < MIN_PAGE_SIZE || pageSize > MAX_PAGE_SIZE) {
+            throw new IllegalArgumentException("cbom.sync.page-size must be within %d..%d, was %d"
+                    .formatted(MIN_PAGE_SIZE, MAX_PAGE_SIZE, pageSize));
+        }
+        if (overlap == null || overlap.isNegative()) {
+            throw new IllegalArgumentException("cbom.sync.overlap must be a non-negative duration, was " + overlap);
+        }
+        if (skippedRetryRuns < 0) {
+            throw new IllegalArgumentException(
+                    "cbom.sync.skipped-retry-runs must be zero or positive, was " + skippedRetryRuns);
+        }
+    }
+
+    /** The first failed attempt plus the configured retries: the attempt count at which an entry is written off. */
+    public int maxAttempts() {
+        return 1 + skippedRetryRuns;
+    }
+}
--- a/src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
+++ b/src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
@@ -28,14 +28,18 @@ import com.otilm.core.attribute.engine.AttributeColumnProjector;
 import com.otilm.core.attribute.engine.AttributeEngine;
 import com.otilm.core.attribute.engine.AttributeEngine.CustomAttributeContentFilter;
 import com.otilm.core.attribute.engine.ListingSortResolver;
+import com.otilm.core.cbom.client.BomSearchPage;
 import com.otilm.core.cbom.client.CbomRepositoryClient;
 import com.otilm.core.comparator.SearchFieldDataComparator;
+import com.otilm.core.config.CbomSyncProperties;
 import com.otilm.core.dao.CryptoAssetConstraintTranslator;
 import com.otilm.core.dao.entity.Cbom;
 import com.otilm.core.dao.entity.Cbom_;
 import com.otilm.core.dao.entity.ScheduledJobHistory;
+import com.otilm.core.dao.entity.cbom.CbomSyncSkip;
 import com.otilm.core.dao.repository.CbomRepository;
 import com.otilm.core.dao.repository.ScheduledJobHistoryRepository;
+import com.otilm.core.dao.repository.cbom.CbomSyncSkipRepository;
 import com.otilm.core.enums.FilterField;
 import com.otilm.core.events.transaction.TransactionHandler;
 import com.otilm.core.logging.LoggerWrapper;
@@ -45,12 +49,15 @@ import com.otilm.core.model.cbom.BomEntryDto;
 import com.otilm.core.model.cbom.BomResponseDto;
 import com.otilm.core.model.cbom.BomSearchRequestDto;
 import com.otilm.core.model.cbom.BomVersionDto;
+import com.otilm.core.model.cbom.CbomHeaderCounts;
+import com.otilm.core.model.cbom.CbomSyncSkipState;
 import com.otilm.core.model.cbom.CryptoStatsDto;
 import com.otilm.core.security.authz.ExternalAuthorization;
 import com.otilm.core.security.authz.SecuredUUID;
 import com.otilm.core.security.authz.SecurityFilter;
 import com.otilm.core.service.CbomExternalService;
 import com.otilm.core.service.CbomInternalService;
+import com.otilm.core.service.writer.cbom.CbomSyncSkipWriter;
 import com.otilm.core.tasks.CbomSyncTask;
 import com.otilm.core.util.CbomUtil;
 import com.otilm.core.util.FilterPredicatesBuilder;
@@ -60,8 +67,13 @@ import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaQuery;
 import jakarta.persistence.criteria.Predicate;
 import jakarta.persistence.criteria.Root;
+import java.net.URI;
+import java.time.Duration;
+import java.time.OffsetDateTime;
 import java.util.ArrayList;
 import java.util.Date;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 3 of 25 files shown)


src/main/java/com/otilm/core/cbom/client/CbomRepositoryClient.java
```diff
@@ -66,21 +71,115 @@ public BomCreateResponseDto create(final CbomUploadRequestDto data) throws CbomR
...
+
+    /** The contract forbids combining a cursor with {@code after}; a Link that does anyway is not safe to follow. */
+    private static boolean namesAfterParameter(final String rawQuery) {
+        for (final String pair : rawQuery.split("&")) {
+            final int equals = pair.indexOf('=');
+            final String name = equals >= 0 ? pair.substring(0, equals) : pair;
+            if ("after".equals(name)) {
+                return true;
+            }
+        }
```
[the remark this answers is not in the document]
karolhrdina: Addressed in 807a5c4f. `namesAfterParameter` now compares the **decoded** name (`URLDecoder.decode(rawName, UTF_8)`), so `bom?cursor=c1&%61fter=0&limit=1` is refused with the same Core-shaped 502 as the plain form. No extra guard around the decoding: the target has already parsed as a `URI`, which rejects malformed escapes itself.

`CbomRepositoryClientTest.search_refusesAnAbsoluteOrPathAbsoluteOrForeignLinkTarget` now covers `%61fter` and the fully encoded `%61%66%74%65%72`; both fail against the previous raw comparison.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -426,196 +481,389 @@ private Cbom getEntity(SecuredUUID uuid) throws NotFoundException {
...
+     * A run shows an outage only when nothing proves the repository serves documents at all: no read succeeded and
+     * either the feed offered nothing new or several documents failed server-side. One new document failing on its own
+     * is not that evidence, so it is charged like any other failure -- one hanging document must never hold the
+     * watermark.
+     */
+    private static boolean looksLikeOutage(SyncRun run) {
+        if (run.deferred.isEmpty() || run.successfulReads > 0) {
+            return false;
+        }
+        return run.read == 0 || run.deferred.size() >= 2;
```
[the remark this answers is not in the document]
karolhrdina: Addressed in 807a5c4f. `SyncRun` carries a new `feedReads` — the document reads the feed pass actually attempted, i.e. what is left after the originals, the invalid identities and the entries already stored — and `looksLikeOutage` branches on that instead of `run.read`.

So the run you describe (feed holds only overlap duplicates, the sole read is a skipped entry's retry answering 503) is an outage again: the run is skipped, nothing is charged to the retry budget, and the watermark holds. That is what the documented rule "the feed offered nothing new" always meant; the entry count could not stand in for it.

New test `CbomSyncITest.aRetryPhaseFailureIsNotChargedWhenTheFeedOnlyRepeatedStoredEntries`, which fails if `run.read` is put back. The Javadoc on `sync()` and `looksLikeOutage` now says "the feed pass read no document" and spells out why the entry count is not the same thing.

src/main/java/com/otilm/core/tasks/CbomSyncTask.java
```diff
@@ -66,9 +67,13 @@ public ScheduledTaskResult performJob(final ScheduledJobInfo scheduledJobInfo, f
...
             }
 
+            // Only a shaped domain exception's own message is operator-safe; anything else (JPA, the WebClient
+            // stack, a bare RuntimeException) could quote driver or framework internals, so it is replaced with the
+            // fallback below. The stack trace itself is still logged, for the Core log.
+            final String safeReason = PlatformException.safeMessage(e, "unexpected error, see the Core log");
             final String errorMessage = String
                     .format("Unable to sync CBOMs for job %s. Error: %s",
-                            scheduledJobInfo == null ? "" : scheduledJobInfo.jobName(), e.getMessage());
+                            scheduledJobInfo == null ? "" : scheduledJobInfo.jobName(), safeReason);
```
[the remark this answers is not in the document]
karolhrdina: Addressed in 807a5c4f, at the client rather than here: `CbomRepositoryClient.fetchPage` reshapes a failed page request into `CBOM Repository failed a page request (HTTP n)`. The status is kept — the sync classifies a 503 by it, so a repository-wide outage still becomes a skipped job — and the repository's own `detail` is logged for the Core log instead of travelling into the job result.

Page requests only: `create`, `read` and `versions` serve the CBOM API, whose callers are meant to see the repository's problem detail, and the per-document read path already shapes its own text (`repository failed the document read (HTTP n)`).

Covered by `CbomRepositoryClientTest.search_keepsTheStatusOfAFailedPageButNotTheRepositoryDetail` and `nextPage_propagatesA400FromAReplicaThatDoesNotKnowTheCursor`, plus `CbomSyncITest.aFailingCursorPageFailsTheRunAndKeepsWhatEarlierPagesStored`, which asserts the run's message no longer contains the repository's sentence.

src/main/resources/application.yml
```diff
@@ -410,6 +410,16 @@ connector:
...
+  sync:
+    # Page size sent as `limit` to the repository search (1..1000); pages are followed through `Link rel="next"`.
+    page-size: ${CBOM_SYNC_PAGE_SIZE:1000}
+    # Each run lists documents created after (start of the last successful run - overlap). The overlap must cover
+    # clock skew between Core and the object store plus the longest upload duration (S3 stamps a multipart upload with
+    # its initiation time); the repository documents 60 s. Re-read entries are deduplicated on (serialNumber, version).
+    # A bare number is read as seconds; a unit suffix (60s, 2m) is also accepted.
+    overlap: ${CBOM_SYNC_OVERLAP:60s}
+    # An entry that could not be stored is retried on this many later runs, then recorded as permanently skipped.
+    skipped-retry-runs: ${CBOM_SYNC_SKIPPED_RETRY_RUNS:3}
```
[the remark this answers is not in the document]
karolhrdina: Not changed in this PR — noting it as a decision rather than a defect, so this thread stays open.

`page-size` and `overlap` are bounded by the repository's contract, and the retry budget was put beside them for that reason. The counter-argument that the budget is operator policy is fair; acting on it means a Settings entry plus a UI surface, and `overlap` would follow the same rule. Tracking it separately rather than widening this change.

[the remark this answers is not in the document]
karolhrdina: @lubomirw This is a design decision, could you please tell me what you'd like me to do here?

vyskocilm (commented):
A review of this PR: two passes, one for correctness and one for the platform's own architecture conventions. Seven findings inline — one leak of an internal URL to a REST caller, two ways the sync can wedge or write a document off during an outage, unbounded growth of the new skip table, plus wiring and reporting items. Nothing in the authorization area is wrong: no resource/action change, no policy or permission work, nothing that needs a migration there.

**What ran** (against head `807a5c4f`, in a clean worktree):

- architecture test suite — 129 tests, 0 failures (all 21 classes)
- the eight test classes this PR touches (`CbomSyncITest`, `CbomServiceITest`, `CbomSyncTaskITest`, `CbomRepositoryClientTest`, `LinkHeaderTest`, `CbomSyncPropertiesTest`, `CbomHeaderCountsTest`, `CbomSyncWatermarkTest`) — 135 tests, 0 failures, with the container runtime healthy so the Testcontainers suites really executed
- `spotless:check checkstyle:check` — 2082 files clean, 0 violations
- context-signature guard — baseline unchanged, so `CbomSyncITest` bought no new Spring context; no ArchUnit freeze-store entry was added either

**What could not be checked:** the full test suite (so regression elsewhere from the `BomEntryDto.timestamp` -> `createdAt` rename is unverified — CI covers it); new-lines coverage parity with the quality gate, which needs a JaCoCo run over the whole suite (`CbomServiceImpl`'s ~500 new lines and `LinkHeader` count, `dao/entity/**` is excluded); and the process items — whether the ticket, the spec and the plan doc were reconciled, and whether an independent review pass was run, which this PR's shape (new control flow, new transaction boundary, new external interaction) is exactly the case for.

**Verified fine, worth saying:** the run holds no transaction across the HTTP calls; entry stores keep `REQUIRES_NEW`; skip bookkeeping is a separate writer bean with the unique constraint as arbiter, and the resolve-vs-record race self-heals on the next run; `LinkHeader` is a hand-walked scanner with no backtracking risk; every cursor-loop guard (run-level URI set, empty-page-with-Link, the 10 000-page ceiling, `next.equals(requested)`) is reachable and correct; the entity's new `cbom_serial_version_unique` mirror really matches `V202601271114`; the native query carries `{h-schema}`; the migration version sorts after main's highest. `resolveNextPage` cannot NPE — the `getScheme() == null` short-circuit covers every case where `getRawPath()` can be null (checked against `java.net.URI` directly). `Closes #2208` is correct as-is in the commit trailer.

**Worth your call, not findings:**

1. `CbomSyncTask.performJob:54` still declares `Propagation.REQUIRES_NEW` (pre-existing, outside this diff). This PR correctly moves `sync()`/`syncAuthorized()` to `NOT_SUPPORTED` — on main they inherited the class-level `REQUIRED` and joined the task's transaction for the whole run of HTTP calls — but the caller still opens a transaction under the platform's 120 s default timeout and then suspends it for a run this PR makes explicitly unbounded (up to 10 000 pages). Whether that suspended transaction pins a pool connection depends on Hibernate's delayed acquisition and I did not measure it; `performJob` issues no statement before the suspend, so it may hold nothing. The in-repo precedent for this shape, `DiscoveryCertificateTask.performJob`, is `NOT_SUPPORTED`. The javadoc added at `CbomServiceImpl:514` reads as though the caller had been changed too.
2. `findAllByOrderByFirstSkippedAtAsc()` has no unique tie-break, so retry order among rows sharing `first_skipped_at` is arbitrary. Harmless here (not a paged listing), but the fix for the skip-table finding touches the same finder.
3. Three small dead ends: `BomEntryDto.WARNING_CRYPTO_STATS_{MISSING,INVALID,SHALLOW,TRUNCATED}` are never read anywhere in main or test (the sync just logs `entry.getWarnings()`) and a static-analysis bot will flag them; `BomEntryDto.createdAt` carries `@NotNull` but feed entries get no `@Valid`, so the annotation has no effect and the field is never read; and `resolveNextPage` also accepts a network-path reference with an empty authority (`//?cursor=x`), harmless since only `getRawQuery()` is used, but wider than the javadoc's "relative-path reference".

src/main/java/com/otilm/core/cbom/client/CbomRepositoryClient.java
```diff
@@ -66,21 +73,144 @@ public BomCreateResponseDto create(final CbomUploadRequestDto data) throws CbomR
...
+    /**
+     * The target is echoed so an operator can see what the repository sent, but it is server-supplied text of unbounded
+     * length that ends up in the scheduler's job result -- so only its first {@value #LINK_TARGET_ECHO_LIMIT}
+     * characters are quoted.
+     */
+    private static CbomRepositoryException unusableLink(final String target) {
+        return new CbomRepositoryException(ProblemDetail
+                .forStatusAndDetail(HttpStatus.BAD_GATEWAY,
+                        "CBOM Repository returned an unusable Link header for the next page: "
+                                + StringUtils.abbreviate(target, LINK_TARGET_ECHO_LIMIT)));
```
vyskocilm: 🔴 bug: this echo reaches a REST client, not only the job result. The javadoc weighs it as operator-visible text in the scheduler's result, but the authorized sync endpoint is on the same path with nothing catching it: `CbomControllerImpl.sync()` -> `CbomServiceImpl.syncAuthorized()` -> `runSync` -> `readFeed` -> `nextPage` -> `fetchPage` -> `resolveNextPage` -> here (`fetchPage`'s catch wraps `processRequest`, not `resolveNextPage`), and `ExceptionHandlingAdvice.handleCbomRepositoryException` copies `problemDetail.getDetail()` straight into `ErrorMessageDto`. The trigger case is precisely an absolute target, so a replica behind an ingress that emits an RFC-8288-legal absolute `Link` makes Core answer 502 with the repository's internal host, port and mount path in the body.

`pageRequestFailed`, twelve lines up, already applies the opposite discipline (log the repository's detail, return Core's own), and `CbomSyncTask:73` swaps `e.getMessage()` for `PlatformException.safeMessage`. Same treatment here: log the full target at WARN, and let the detail be the constant prefix with nothing appended. `CbomRepositoryClientTest.search_refusesAnAbsoluteOrPathAbsoluteOrForeignLinkTarget:273` currently asserts the echoed target, so it pins the leak and needs updating with the fix.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -426,196 +481,413 @@ private Cbom getEntity(SecuredUUID uuid) throws NotFoundException {
...
+     * either the feed pass read no document -- so the retry phase's unanswered reads are the only evidence the run has
+     * -- or several documents failed server-side. One new document failing on its own is not that evidence, so it is
+     * charged like any other failure -- one hanging document must never hold the watermark.
+     *
+     * <p>
+     * The feed's entry count cannot stand in for {@code feedReads}: a page of entries the overlap re-offers, of
+     * original uploads or of invalid identities is read as documents the repository served, although not one of them
+     * was read.
+     */
+    private static boolean looksLikeOutage(SyncRun run) {
```
vyskocilm, line 834: 🔴 bug: `feedReads == 0` reads a quiet hour as a repository outage, and one hanging document then freezes the watermark for good.

One document that permanently answers 503 — or just exceeds the client's response timeout, which is mapped to 503. Run 1: `feedReads == 1`, `deferred == 1`, so no outage, attempts=1, RETRYING. Every later run in which the feed offers nothing new: `successfulReads == 0`, `feedReads == 0`, `deferred == 1` -> outage -> `settleUnavailable` throws 503 -> `CbomSyncTask` turns it into `ScheduledJobSkippedException` -> `SchedulerServiceImpl.finalizeFinishedScheduledJob` deletes the history row. The job never records SUCCESS again, `getLastSyncTimestamp()` keeps returning the same stale watermark (so the listing window grows every hour and re-lists an ever larger prefix), and that entry's retry budget never spends — the bounded retry is unbounded for exactly the hanging-document case the class comment says it bounds.

`aRetryPhaseWithOnlyServerSideFailuresDoesNotChargeTheBudget` pins the not-charging rule, so the rule is deliberate; it is the perpetual-skip consequence that is untested. A retry-phase 503 needs its own evidence of an outage — the `deferred.size() >= 2` arm, or charging after N consecutive skipped runs — rather than the absence of new documents.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -426,196 +481,413 @@ private Cbom getEntity(SecuredUUID uuid) throws NotFoundException {
...
+        } catch (CbomRepositoryException e) {
+            final ProblemDetail problemDetail = e.getProblemDetail();
+            if (problemDetail == null) {
+                return StoreOutcome.unavailable("repository failed the document read (status unknown)");
             }
+            final String reason = "repository failed the document read (HTTP " + problemDetail.getStatus() + ")";
+            // A 503 is not a verdict on this document: the repository answers it when it is unavailable as a whole, and
+            // the client synthesizes it when it never got an answer -- a refused connection, or the response timeout.
+            // Any other status the repository chose, 5xx included, is about this document and is charged to it.
+            return problemDetail.getStatus() == HttpStatus.SERVICE_UNAVAILABLE.value()
```
vyskocilm, line 742: 🔴 bug: the 503-or-charge decision reads the status out of the error **body**, not the HTTP response. `problemDetail` here is `CbomRepositoryClient.handleHttpExceptions`' `clientResponse.bodyToMono(ProblemDetail.class)`, and the default mapper ignores unknown properties. So a repository or ingress that answers HTTP 503 with `{"error":"service unavailable"}` decodes to a `ProblemDetail` whose `status` is `0`: the read is charged to the document, and after `skipped-retry-runs` runs of the same outage the entry is `PERMANENTLY_SKIPPED` and never synced again — entry loss caused by the outage rather than by the document. The operator-visible reason also reads `repository failed the document read (HTTP 0)`. The inverse leaks the other way: a real 500 whose body claims `"status":503` is deferred, and feeds the watermark freeze flagged above.

`clientResponse.statusCode()` is the value that should drive this (the same body-derived status also reaches `CbomRepositoryClient:142` `pageRequestFailed`, hence the task's skip-or-fail decision for page requests). Every test builds the body with the `problem(status, detail)` helper, which always writes `status` into it, so this path is untested.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -426,196 +481,413 @@ private Cbom getEntity(SecuredUUID uuid) throws NotFoundException {
...
+        return syncResultMessage;
+    }
+
+    private Map<SyncIdentity, CbomSyncSkip> loadSkipRecords() {
+        final Map<SyncIdentity, CbomSyncSkip> skips = new LinkedHashMap<>();
+        for (CbomSyncSkip skip : syncSkipRepository.findAllByOrderByFirstSkippedAtAsc()) {
+            skips.put(new SyncIdentity(skip.getSerialNumber(), skip.getVersion()), skip);
+        }
+        return skips;
+    }
```
vyskocilm, line 556: 🟡 risk: this loads the whole `cbom_sync_skip` table on every run, and a `PERMANENTLY_SKIPPED` row can never leave it — so both the table and the per-run load grow without bound on an hourly job.

The finder has no state filter and no limit. The only delete path is `CbomSyncSkipWriter.resolve` via `resolveSkipIfRecorded`, which needs the feed to re-offer the entry — but the watermark has long moved past it, and `retrySkippedEntry` returns at its first line for any state other than `RETRYING`, before it even reaches `existsBySerialNumberAndVersion`. A written-off row is therefore immortal: read into the map every run, forever, only to be skipped.

Compounding it, the row's stated purpose is unreachable — there is no controller, DTO, service method or filter-field constant for `CbomSyncSkip` anywhere in main, so the `reason` column that the entity javadoc and the migration header both call operator-visible is visible to nobody; `run.summary()`'s counters are the only signal and they do not name the documents. Pick one: narrow the finder to `RETRYING` and look `previous` up per identity (`findBySerialNumberAndVersion` already exists) so the load is bounded by the live retry set; or sweep written-off rows on retention; or expose them so the record is actually a record. Whichever way, the "kept as the record of a permanently skipped document" claim in the migration header and the javadoc should match.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -426,196 +481,413 @@ private Cbom getEntity(SecuredUUID uuid) throws NotFoundException {
...
 
-    public boolean isCbomRepositoryClientConfigured() {
-        return cbomRepositoryClient.isConfigured();
+    /**
+     * Records one entry's failure and charges it to the counter it belongs to: an entry this deployment had already
+     * written off is counted apart, so the run's report does not read as if it had just given up on it.
+     */
+    private void recordFailure(SyncIdentity identity, String reason, CbomHeaderCounts counts, SyncRun run,
+            CbomSyncSkip previous) {
+        if (recordSkip(identity, reason, counts, run, previous)) {
```
vyskocilm: 🔵 nit: an entry written off on this run is reported as both retried and abandoned. `recordSkip` increments `permanentlySkipped` and returns `false` when the row crosses into `PERMANENTLY_SKIPPED` for the first time, so this branch then also does `failed++`. With the defaults, an entry at attempts=3 failing on run 4 yields a job result reading `1 entries could not be stored and were recorded for retry; ... 1 entries are now permanently skipped` — queued for retry and given up on, in one sentence. `anEntryWhoseDocumentCannotBeReadIsRecordedRetriedOnThreeRunsThenPermanentlySkipped` only asserts on the retried/resolved/permanently-skipped clause, so it does not catch this.

src/main/resources/application.yml
vyskocilm, line 422: 🟡 risk: these three env-var overrides have no Helm-chart wiring, so in a deployed environment all three sit at their compiled-in defaults. `CBOM_SYNC_PAGE_SIZE`, `CBOM_SYNC_OVERLAP` and `CBOM_SYNC_SKIPPED_RETRY_RUNS` exist only as placeholders here; until the charts repo passes them through, a deployment cannot move the overlap even though the `CbomSyncProperties` javadoc grounds the right value in that deployment's clock skew and its longest multipart upload — the knob an actual incident needs turned. The PR body says this is handled separately; worth confirming the charts change is really queued, since the two land in different repos and a placeholder that nothing passes through looks configurable and is not.

src/main/java/com/otilm/core/config/CbomSyncProperties.java
```diff
@@ -0,0 +1,50 @@
...
+ * @param overlap how far behind the last successful run's start the next run lists from; 60 s by default. A bare number
+ * binds as seconds: without {@code @DurationUnit} Spring reads it as milliseconds, so {@code 60} would silently mean 60
+ * ms and the overlap would vanish
+ * @param skippedRetryRuns how many later runs retry an entry that could not be stored before it is recorded as
+ * permanently skipped; 3 by default, 0 writes an entry off at its first failure
+ */
+@ConfigurationProperties(prefix = "cbom.sync")
+public record CbomSyncProperties(@DefaultValue("1000") int pageSize,
+        @DefaultValue("60s") @DurationUnit(ChronoUnit.SECONDS) Duration overlap,
+        @DefaultValue("3") int skippedRetryRuns) {
```
vyskocilm, line 27: 🟡 risk: `skippedRetryRuns` looks like operator policy rather than deploy-time plumbing, and yml is unreachable for a SaaS customer — the settings subsystem is their only configuration surface. Its two siblings are fine: `pageSize` is bounded by the repository's own contract (1..1000, enforced both here and in `CbomRepositoryClient.search`) and no operator has a reason to move it, and `overlap` is grounded in this deployment's clock skew and upload duration, which is per-environment plumbing. `skippedRetryRuns` has neither justification — it decides how many hourly runs Core keeps trying before writing a CBOM document off, with an operator-visible consequence, and it is needed long after the settings store exists: this very client already reads its base URL from `SettingsCache.getSettings(SettingsSection.PLATFORM)` (`CbomRepositoryClient:248`), so settings are demonstrably available on this path. Either move it into settings, or extend the record javadoc with the reason it must be deploy-time, the way the other two parameters already carry theirs.

src/main/java/com/otilm/core/cbom/client/CbomRepositoryClient.java
karolhrdina: Confirmed and fixed in fc3044f4. The chain is as described — `CbomControllerImpl.sync()` → `syncAuthorized()` → … → `resolveNextPage`, and `handleCbomRepositoryException` copies `detail` into `ErrorMessageDto` — so an absolute target put the repository's host into a 502 body.

`unusableLink` now logs the (capped) target at WARN and returns the constant sentence with nothing appended; `search_refusesAnAbsoluteOrPathAbsoluteOrForeignLinkTarget` asserts the constant instead of the echo. While there: a network-path reference with an empty authority (`//?cursor=x`, your item 3) is refused as well, and is on that test's target list.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
karolhrdina: Confirmed — the trace is right, and it is worse than a quiet hour: the exemption was self-perpetuating. Fixed in fc3044f4 by changing what counts as evidence.

The verdict now reads `successfulReads == 0 && feedDeferred >= 2`, where `feedDeferred` counts only the feed pass's unanswered reads. A retry's failure never weighs in, however many retries fail: the page requests of the same run just succeeded, so the repository is serving, and a retry exempt from being charged would stay exempt in every quiet run to come. I did not take the `deferred.size() >= 2` arm alone — two hanging documents in the retry set would recreate the freeze — and a persisted consecutive-skip counter buys little for its machinery. The price is one attempt charged to a retry when an outage begins between a run's page requests and its retries.

Tests: `aPermanentlyHangingDocumentIsWrittenOffAfterTheBudgetEvenWhenTheFeedStaysQuiet` (four runs, every one completes, ends `PERMANENTLY_SKIPPED`; one run re-offers an already-stored duplicate to cover the Copilot variant) and `aRetryPhaseFailureDoesNotCountTowardsAnOutage` (feed document 503 + retry 503: charged, run completes). Both error out with the old arm restored. The two tests that pinned the exemption are gone; the two-feed-documents outage test stays. Javadoc on `sync()`, `looksLikeOutage` and `settleUnavailable` rewritten; PR body updated.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
karolhrdina: Confirmed and fixed in fc3044f4. `handleHttpExceptions` now sets `problemDetail.setStatus(clientResponse.statusCode().value())` after decoding the body, so the status every caller classifies by is the response's — a body that is JSON but no problem detail (decodes with status 0) and a body that names another status are both overridden.

Tests: `read_classifiesAFailedDocumentReadByTheResponseStatusNotByTheBody` (HTTP 503 with `{"error":"service unavailable"}` → 503) and `search_classifiesAFailedPageByTheResponseStatusNotByTheBody` (HTTP 500 with a body claiming `"status":503` → 500, Core-shaped detail).

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
karolhrdina: Confirmed on both counts; the load is bounded in fc3044f4, retention and visibility are not in this PR.

- `loadSkipRecords` now reads `findAllByStateOrderByFirstSkippedAtAscUuidAsc(RETRYING)` — the live retry set, with the `uuid` tie-break from your item 2 — so the per-run load grows with open retries, not with every document ever written off. `previous` for a failing entry and the row to resolve for a stored one are looked up by identity when the map does not hold them (`findBySerialNumberAndVersion`, one indexed read on the failure and store paths, both dominated by the document GET). A written-off row is therefore still deleted when its document finally arrives: `aWrittenOffEntryIsStillResolvedWhenItsDocumentFinallyStores`, and the existing "failed again" test now exercises the lookup.
- Written-off rows are not swept and not exposed here. The entity javadoc and the migration header now say exactly that — kept `PERMANENTLY_SKIPPED`, not loaded again, found in the database and in the Core log's write-off warning, retention and an operator's view as follow-up work — instead of calling the row a record nobody can read. @karolhrdina has the retention/reset follow-up on his list; the operator-facing view belongs with it.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
karolhrdina: Confirmed and fixed in fc3044f4. `recordSkip` returns `RETRYING | WRITTEN_OFF | ALREADY_WRITTEN_OFF` and `recordFailure` increments exactly one counter per outcome (`recordedForRetry`, `permanentlySkipped`, `alreadyPermanent`), so the fourth run reads `0 entries could not be stored and were recorded for retry; … 1 entries are now permanently skipped`. `anEntryWhoseDocumentCannotBeReadIsRecordedRetriedOnThreeRunsThenPermanentlySkipped` now asserts the retry clause on that run too.

src/main/resources/application.yml
karolhrdina: Not in this PR by design — the env-var pass-through lives in the charts repo, and @karolhrdina is carrying that change together with the ≥ 0.3.0 prerequisite note. Agreed that a placeholder nothing passes through is not configuration; Karol, please link the charts PR here so the two land together.

src/main/java/com/otilm/core/config/CbomSyncProperties.java
karolhrdina: Not changed here — this is a decision, and it is Karol's. Your split of the three is fair: `pageSize` is bounded by the contract and `overlap` is per-environment plumbing, while `skippedRetryRuns` is policy with a visible consequence and no deploy-time reason of its own; I will not write a justification into the javadoc that I do not believe. Moving it means a Settings field, its DTO/UI surface and a migration, so it is either a follow-up or a widening of this PR. Copilot raised the same point (thread above); both stay open until Karol rules.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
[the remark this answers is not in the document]
karolhrdina: Follow-up after the human review: the rule changed once more in fc3044f4. Counting feed reads still let one permanently hanging document freeze the watermark in every quiet run, so the verdict now counts only the feed pass's unanswered reads (`feedDeferred >= 2`) and a retry's failure never weighs in -- see https://github.com/OmniTrustILM/core/pull/2245#discussion_r3980337954.

src/main/resources/application.yml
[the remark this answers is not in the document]
karolhrdina: Follow-up: #2249 moves all three `cbom.sync.*` tunables into the Settings UI (parent OmniTrustILM/ilm#299). Resolving against it.

src/main/java/com/otilm/core/config/CbomSyncProperties.java
karolhrdina: Follow-up: #2249 moves all three `cbom.sync.*` tunables into the Settings UI (parent OmniTrustILM/ilm#299). Resolving against it.

src/main/resources/application.yml
karolhrdina: Will be in a follow-up https://github.com/OmniTrustILM/core/issues/2249

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
karolhrdina: Retention and the operator view are filed as **#2250** *Retention and operator view for the CBOM documents the sync could not store* (parent OmniTrustILM/ilm#299, 2 MD): a batched sweep of `PERMANENTLY_SKIPPED` rows after a retention period, a UI that names the document, its state and its reason and lets an operator request another attempt, and every value of it configurable from the Settings UI next to what #2249 moves there. Whether the operator sees it as a view of `cbom_sync_skip` or as a `cbom` row carrying a sync state the way `asset_sync_state` already does is the story's first decision. The bounded load and the identity lookup stay as landed here (`loadSkipRecords` reads `RETRYING` only). Resolving against #2250.

vyskocilm (approved):

PR #2170, opened 2026-08-28
Carry the ratified decision tables and the identity chain that reads them (2 of 3, on top of core#2173)

vyskocilm:
## Where all the lines come from

Three input files plus one script. Script reads inputs, writes one JSON. That JSON is the table.

```
src/main/cbom/identity/
  cryptography-defs.json         CycloneDX registry snapshot — families, curves
  cryptography-defs.schema.json  CycloneDX schema — the legal family enum
  oid-strand.json                OID -> family, ratified here, no upstream source
                 |
                 v
scripts/cbom/build_identity_tables.py   + grammar tables written inside the script
                 |
                 v
src/main/resources/cbom/identity-tables.json   4 256 lines, committed, on the classpath
```

**Four things go in, four things come out.**

1. *Computed from the registry.* Curves. 246 spellings collapse to 197 canonical curves; 29 of those have more than one spelling. 510 aliases fall out of the same pass. Nobody typed these — `secp256r1`, `prime256v1` and `NIST P-256` are one curve because the registry says so.
2. *Merged from schema + registry + this repo.* Families. Schema enum, plus families the data carries but the enum forgot, plus 33 pseudo-families this repo adds (RSA, EC, Kyber, Falcon, Fernet …) because the registry has no token for a bare RSA key and the inventory filter must still catch one.
3. *Copied from `oid-strand.json`.* 196 OID → family rows and 11 blocked prefixes. Copied, not derived. See the provenance table below.
4. *Pasted from the script.* 139 name-grammar rules, the size stoplist, mode and padding tokens, primitive defaults. Regexes a human wrote and a human reviews.

**Script checks itself before it writes.** Every family the grammar or the OID table can produce must be legal, must be spelled the way the family lookup spells it, and must have a primitive default. Every closed-vocabulary token must be printable ASCII. A miss is `SystemExit`, not a warning — a bad table never reaches disk.

**Why not derive everything.** Only ~32% of the output is genuinely computed from the registry. The rest is decisions. Decisions have no upstream to derive from. So: write them once, in a place a reviewer can read, and make a machine prove the committed bytes still match.

**Two things stop a silent edit.**

- CI regenerates and diffs → catches a hand-edit no input justifies.
- SHA-256 pin in `IdentityTablesTest` → catches an input change, which re-keys the whole inventory and must be ratified with the vectors in the same commit.

Change an input, re-run the script, ratify the new hash. That is the whole loop.

Closes #2168.

[...]

Issue #2071, opened 2026-08-14
Implement cryptographic asset inventory data model and persistence


karolhrdina:
### Description

Flyway migrations, entities and repositories for the cryptographic asset
inventory (ilm#299, design doc section 3):

- crypto_asset: typed identity and filter columns (asset type, name, oid,
  algorithm family, primitive, parameter set, curve, mode, padding, variant),
  identity_key with a plain UNIQUE arbiter, ruleset_version recording the rules
  that keyed the row, properties_leaf_count and properties_hash (the merge
  bookkeeping — also on crypto_asset_source), unindexed JSONB merged
  crypto_properties, PQC verdict
  columns (verdict, rule id, reason, rule-set version, evaluated-fields
  snapshot), properties-source pointer, source_count.

- crypto_asset_alias: (absorbed_key, canonical_key, reason, decided_by,
  decided_at) — the repair path for an accidental duplicate. Nothing in the
  keying path reads it, so it cannot affect conformance; it survives re-ingest,
  it is exactly reversible because per-source payloads are retained, and an
  alias is REFUSED where a safety rule caused the split (a refuted certificate
  digest, or a bare-CN row facing a full-DN row).

- crypto_asset_source: many-to-many link with per-source original
  crypto_properties, capped evidence JSONB plus occurrence_count,
  first/last_seen_at; UNIQUE (asset_uuid, cbom_uuid); FK ON DELETE CASCADE
  to the asset, RESTRICT to the cbom row (deletion goes through the
  service path). capped evidence JSONB plus occurrence_count — capping DROPS the
  occurrence's additionalContext field rather than truncating it, because the
  1.7 schema documents that field as "e.g. a code snippet" and at a
  secret-scanner finding the snippet is the secret line ...

- cbom_tombstone table; asset_sync_state / asset_sync_error /
  assets_synced_at columns on the existing cbom table.

- Btree indexes for every filter column plus lower(name) / lower(oid) for
  free-text. No GIN, no CREATE EXTENSION, no partitioning in v1.

- Entities via the house JSONB mapping, repositories extending
  SecurityFilterRepository, writer beans for all modifying paths,
  FilterField enum entries for the whole filter set.

- Entities via the house JSONB mapping, repositories extending
  SecurityFilterRepository, writer beans for all modifying paths, FilterField
  enum entries for the whole filter set EXCEPT identity_key, which must never
  appear in FilterField, in any DTO, or in any response: given the key, a
[...]

PR #2247, opened 2026-09-10
Filter cryptographic assets by curve membership (curve TEXT → text[])

### LINKED ISSUE #2194: Per-curve membership filtering for composite curves (curve TEXT → text[])
### Product decision (2026-09-02)

An asset whose curve is a `+`-joined composite (`other/curve25519+other/curve448` — hybrid schemes such as X25519/X448) is stored as one scalar value. The curve filter is `LIST`/`EQUALS`, so the composite is offered as its own filter value and matches exactly the rows holding it — but selecting `other/curve25519` alone does **not** return those rows.

**Decision: the inventory should filter by per-curve membership.** The product's core query is "show every asset that touches curve X" (deprecation sweeps, compliance review, PQC migration planning); hybrid schemes are precisely the transitional pattern this inventory exists to track, and a per-curve sweep that silently excludes hybrids is a completeness trap. Exact-composite selection stays available alongside membership.

Decided by Claude on Karol's dispatch, 2026-09-02, closing the question carried in #2166 and in PR #2170 discussion [r3880812408](https://github.com/OmniTrustILM/core/pull/2170#discussion_r3880812408). If product disagrees, close this ticket with the reasoning — the shipped equality behavior is coherent and test-pinned, so declining costs nothing.

### Shape — an existing pattern, not a filter-semantics invention

- Migrate `crypto_asset.curve` `TEXT` → `text[]`; the write path stores the composite split on `+` (single curve = one-element array, absent = `NULL`).
- Swap `FilterField.CBOM_ASSET_CURVE` (`FilterField.java:406`) from `SearchFieldTypeEnum.LIST` to `SearchFieldTypeEnum.NATIVE_ARRAY` — gains `CONTAINS`/`NOT_CONTAINS`, keeps `EQUALS`/`NOT_EQUALS`/`EMPTY`/`NOT_EMPTY`. The `= ANY(column)`/array-contains predicate paths already exist in `FilterPredicatesBuilder`, and the pattern is already served in production fields (Features, NTP Servers).
- Wire compatibility: `NATIVE_ARRAY` reports `FilterFieldType.LIST` (see `SearchHelper.java:75`), so the interfaces contract is unchanged — only the advertised condition list grows.
- Rework `CryptoAssetRepository.findDistinctCurve()` — the recursive-CTE loose index scan assumes a scalar column; distinct over unnested members instead, so the value list offers each curve once rather than each combination once.

### Constraints

- **Identity preimage unchanged.** The ratified vectors keep the `+`-joined form in the preimage (`ALG|ECDH|256|other/Curve25519+other/Curve448|||` in `identity-key-vectors.json`); only the stored projection and filter semantics change. No re-key; `RULESET_VERSION` stays 2.
- **Cheapest strictly before core#2073's first ingest.** Today this is a column-type migration over an empty table. After data exists it becomes a data migration plus the saved list-view value sweep that #2166 already prices for the canonicalization fold.
- **core#2145 statistics:** if any statistic groups by curve, that ticket must define multi-membership counting (an asset with two curves appearing in two buckets vs. distinct-asset counting).
- Tests to update with the swap: `CryptographicAssetSearchableFieldsITest.theCurveFieldOffersTheStoredNormalizedSpellingOnlyOnce`, `CryptoAssetListQueryITest.distinctValueQueriesReturnCanonicalSortedValuesWithoutNullsOrDuplicates`.

Refs: #2166 (carried the question) · #2074 (shipped the equality shape, closed 2026-08-31) · PR #2170 discussions [r3880812408](https://github.com/OmniTrustILM/core/pull/2170#discussion_r3880812408) → [r3904650809](https://github.com/OmniTrustILM/core/pull/2170#discussion_r3904650809) (correct-as-shipped verification).

vyskocilm:
Closes #2194.

A hybrid scheme (X25519/X448, and every transitional pairing after it) reports one curve value holding both members joined on `+`. Held as scalar `TEXT`, the query this inventory exists to serve — *show me every asset that touches curve25519* — silently skipped exactly those rows, which are the transitional pattern a PQC migration most needs to find.

`crypto_asset.curve` becomes `TEXT[]`, storing the members the normalizer had already sorted and deduplicated, and `CBOM_ASSET_CURVE` moves from `LIST` to `NATIVE_ARRAY` so `EQUALS` asks membership. `findDistinctCurve()` offers each curve once rather than each combination once.

**The identity preimage is untouched.** It still hashes the `+`-joined spelling, so no row is re-keyed and `RULESET_VERSION` stays 2. `CompositeCurve` rejoins the members for the API DTO, leaving the wire contract byte-identical — `NATIVE_ARRAY` still reports `FilterFieldType.LIST`, so `interfaces` needs no change and only the advertised condition list grows.

## Two things a reviewer should not have to discover

**1. This changes a shared filter predicate, not just the curve field.**

`PostgresFunctionContributor.ARRAY_CONTAINS` moves from `CAST(?1 AS TEXT) = ANY(?2)` to `CAST(?2 AS TEXT[]) @> ARRAY[CAST(?1 AS TEXT)]`, which applies to every `NATIVE_ARRAY` field — `CONNECTOR_FEATURES`, `OID_ENTRY_ALT_CODES` and `TIME_QUALITY_CONFIGURATION_NTP_SERVERS` as well as the curve.

It is forced rather than opportunistic: PostgreSQL has **no index path for `scalar = ANY(column)` at all**, so the GIN index this migration adds would have been decorative. Measured on the pinned `postgres:15-alpine`, 2 000 rows, `enable_seqscan = off` — `= ANY` plans a disabled-cost `Seq Scan`, containment plans a `Bitmap Index Scan`. `CryptoAssetCurveMembershipMigrationITest.theCurveIndexAnswersThePredicateTheFilterEmits` renders the predicate from the shipped constant rather than a copy, so the pairing cannot drift; reverting the constant fails it.

The cast on the column is what keeps the other three fields working in the **entity-generated** schema the integration suite builds: there a `List<String>` under `SqlTypes.ARRAY` generates `varchar[]`, and `@>` is not defined across `varchar[]` and `text[]`.

**Corrected during review** (thread from @karolhrdina): an earlier version of this paragraph and of the constant's Javadoc concluded that indexability is a property of `columnDefinition`, and that a GIN index on the other three would buy nothing until their type moved. That is false for the shipped schema — every array column the migrations ship is already `TEXT[]` (`custom_oid_entry.alt_codes`, `connector_interface.features`, `time_quality_configuration.ntp_servers`), `columnDefinition` shapes only the generated schema, and in production the cast is a no-op on all four. A GIN index on any of them would be matched today, with no type change needed. The Javadoc now says that.

**Six operators returned a 500 on this field, and the gate for that is central.**

Raised by @karolhrdina, who ran all six against both branches on real Postgres. `SearchFieldTypeEnum` gates which conditions are *advertised*, but nothing held a **submitted** condition to that set outside saved list views. While `curve` was `TEXT`, `STARTS_WITH`, `ENDS_WITH`, `MATCHES`, `NOT_MATCHES`, `GREATER` and `LESSER` produced valid if unadvertised SQL; against `text[]` they fail inside the query as a `DataAccessException`, which `ExceptionHandlingAdvice` does not handle — so **500, not 422**.

`FilterPredicatesBuilder.getPropertyFilterPredicate` now refuses any condition outside `SearchHelper.availableConditions(filterField)` before a predicate is built, on **every listing** — the same set `ListViewServiceImpl` already validates stored views against, so the two paths agree.

This is wider than the curve. The gate failed four existing tests on three unrelated resources, all using operators their fields never advertised and all of which *worked*: `ENTITY_CONNECTOR_NAME`/`ENTITY_KIND` `CONTAINS`, `LOCATION_ENTITY_INSTANCE` `ENDS_WITH`, `AUDIT_LOG_RESOURCE_NAME` `NOT_CONTAINS`. So the defect is that the advertise list and the predicate builder disagree; the 500 is only where that disagreement meets a column type that cannot absorb it. **Ruled: gate centrally**, and hold those four tests to what their fields advertise — a field's advertised set otherwise means nothing. `AUDIT_LOG_RESOURCE_NAME`'s json-array not-present branch stays covered through `EMPTY`, which reaches the same code.

**The `+` separator now has one definition, and the split happens in Java.**

@karolhrdina found it spelled in five places with no test crossing more than one — mutating `CompositeCurve.SEPARATOR` left 55 tests green. `CompositeCurve` now owns both directions, `CryptoAssetWriter` binds the members, the upsert drops `string_to_array`, and `AssetNormalizer` joins through the same constant. Two SQL sites remain, each pinned by a test that fails when it moves: `array_to_string` in `findStaleVerdictRows` and `string_to_array` in the migration. This is also what #2166 needs — a writer-side fold cannot see members while the split happens inside the SQL statement.

**`EMPTY`/`NOT_EMPTY` lose their index path.** Measured by @karolhrdina: `array_ops` GIN indexes elements, so it answers neither `curve IS NULL` nor `cardinality(curve) = 0`. `EQUALS` gains an index path it never had; those two lose the one the btree gave them. Taken deliberately — `EQUALS` is the query this inventory exists to answer — and written into the migration, which names the partial index that would restore it.

**Merge order: this must land before #2073.** `ALTER COLUMN ... TYPE ... USING` takes `ACCESS EXCLUSIVE` and rewrites the table; the `CREATE INDEX` blocks writes again. Free only because `crypto_asset` is empty in every environment today. Behind ingest they are outage-class, and `CONCURRENTLY` is unavailable because Flyway wraps the script in a transaction. Stated in the migration, and carried to [#2073](https://github.com/OmniTrustILM/core/issues/2073#issuecomment-5436951889).

**One measured semantic edge.** For an array holding a NULL element that does not contain the searched value, `= ANY` returned NULL where containment returns `false`. Under `EQUALS` both filter the row out. Under `NOT_EQUALS` the old form dropped such a row and the new form includes it — the correct answer, but a change to three other resources' listings. No production writer produces such arrays; `alt_codes` is the one whose elements come from a client payload.

**2. The GIN index is the escalation v1 deferred.**

v1 shipped "no GIN, no extensions, no partitioning — documented escalations", aimed at not indexing the JSONB long tail. A btree cannot answer membership on an array at all, so the type change forces this one. It is documented in the migration where it is taken.

## Verification

- **8181 tests, 0 failures, 0 errors** (full suite, after the review fixes in `6d642f93`)
- 806 tests green: all `com.otilm.core.architecture.*` plus every crypto-asset, PQC, CBOM, native-array, custom-OID and connector suite
- `spotless:check checkstyle:check` clean; `ContextSignatureGuardTest` BASELINE unchanged — no new Spring context bought
- `CryptoAssetCurveMembershipMigrationITest` runs the migration in a scratch schema and asserts the `USING` conversion directly: a composite splits, a single curve becomes one element, an absent curve stays NULL. The suite's schema is entity-generated, so that clause runs nowhere else.
- The architecture gate was run twice and both its findings fixed; the second pass re-verified them independently, including a mutation check that the new index assertion can fail.

## A defect this found late

`findStaleVerdictRows` selected `curve` raw while `PqcStaleVerdictRow` reads it with `row.get("curve", String.class)`. A `text[]` arriving there fails the type check, so the first PQC sweep over any inventory holding one curve-bearing row would have died in the work-list mapping, before per-row error handling could catch it — `ClassCastException: Cannot cast [Ljava.lang.String; to java.lang.String`. Fixed with `array_to_string(curve, '+') AS curve`.

Every existing sweep fixture seeded a NULL curve, and NULL passes the type check, so the whole suite was green over it. `aRowWithAHybridCurveIsReadBackJoinedAndSwept` now seeds one; it was written before the fix and observed to fail.

## Open, and not for this PR

- **The ticket body says "Exact-composite selection stays available alongside membership." It does not.** Raised again by Copilot in review; the full ruling and the requested correction are recorded on [#2194](https://github.com/OmniTrustILM/core/issues/2194#issuecomment-5660109637), for the ticket owner to apply. `NATIVE_ARRAY` `EQUALS` is membership, never array equality, and the value list now offers members, so the composite spelling is not selectable. Ruled 2026-09-10 to keep membership — it is what the product decision in #2194 asks for, and "touches both" is two ANDed filters. The ticket sentence needs correcting to match.
- **The empty-inventory premise is settled, and it is what makes the DDL free.** There is no deployed database, and `CryptoAssetWriter` has no production caller — asserted, not remembered, by `IdentityRulesetStampArchTest.theAssetWriterIsStillUnreachableFromProduction`, which fails the build on the first production caller. An earlier version of this section listed this as an unrun query; it is neither unrun nor a query. The migration converts existing data and is tested to, so nothing here depends on it either way.
- #2166 (curve class representatives) is unaffected: the value list still offers distinct stored spellings, now per member. A fold placed at the writer will need to fold per member.

### DIFF
--- a/src/main/java/com/otilm/core/cbom/asset/CompositeCurve.java
+++ b/src/main/java/com/otilm/core/cbom/asset/CompositeCurve.java
@@ -1,0 +1,29 @@
+package com.otilm.core.cbom.asset;
+
+import java.util.List;
+
+/**
+ * Renders the stored curve members back into the composite spelling.
+ *
+ * <p>
+ * A hybrid scheme names more than one curve, and the normalizer renders those as a single {@code +}-joined token in
+ * sorted, deduplicated order. That token is what the ratified identity vectors hash, so its shape can never change; the
+ * column holds the members split out of it, which is what lets a filter ask whether an asset touches a curve rather
+ * than whether its whole composite equals one. The split preserves the normalizer's order, so joining the members back
+ * reproduces the preimage spelling byte for byte -- which is what the API contract keeps carrying.
+ */
+public final class CompositeCurve {
+
+    private static final String SEPARATOR = "+";
+
+    private CompositeCurve() {
+    }
+
+    /** The stored members in the spelling the identity preimage and the API contract both use, or {@code null}. */
+    public static String join(List<String> members) {
+        if (members == null || members.isEmpty()) {
+            return null;
+        }
+        return String.join(SEPARATOR, members);
+    }
+}
--- a/src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
+++ b/src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
@@ -64,6 +64,11 @@ public interface CryptoAssetRepository extends SecurityFilterRepository<CryptoAs
      * reconciling a genuine disagreement is an explicit decision rather than a side effect of sync order.
      *
      * <p>
+     * <b>Curve:</b> bound as the normalized composite and stored split on {@code +}, so a hybrid scheme's members can
+     * each be matched on their own. The caller keeps passing the joined spelling because that is what the identity
+     * preimage hashes; the split is a storage projection, and {@code CompositeCurve} is its inverse on the read side.
+     *
+     * <p>
      * <b>Identity guard:</b> an existing guard survives, because it is a safety refusal rather than a field. A guard
      * says this row was deliberately kept separate — a refuted certificate digest, a bare common name facing a full
      * subject DN — and {@code CryptoAssetAliasWriter} refuses an alias by reading the guard that is on the row now.
@@ -81,7 +86,7 @@ public interface CryptoAssetRepository extends SecurityFilterRepository<CryptoAs
                     algorithm_family, primitive, parameter_set, curve, mode, padding, variant, identity_guard,
                     properties_leaf_count, source_count, i_cre, i_upd)
             VALUES (:uuid, :key, :rulesetVersion, :assetType, :name, :oid, :algorithmFamily, :primitive,
-                    :parameterSet, :curve, :mode, :padding, :variant, :identityGuard,
+                    :parameterSet, string_to_array(CAST(:curve AS TEXT), '+'), :mode, :padding, :variant, :identityGuard,
                     0, 0, CURRENT_TIMESTAMP, CURRENT_TIMESTAMP)
             ON CONFLICT (identity_key) DO UPDATE SET
                 ruleset_version = EXCLUDED.ruleset_version,
@@ -97,7 +102,8 @@ public interface CryptoAssetRepository extends SecurityFilterRepository<CryptoAs
                 variant = COALESCE(crypto_asset.variant, EXCLUDED.variant),
                 identity_guard = COALESCE(crypto_asset.identity_guard, EXCLUDED.identity_guard),
                 i_upd = CURRENT_TIMESTAMP
-            """, nativeQuery = true)
+            """,
+            nativeQuery = true)
     void upsertIdentity(@Param("uuid") UUID uuid, @Param("key") String key, @Param("rulesetVersion") int rulesetVersion,
             @Param("assetType") String assetType, @Param("name") String name, @Param("oid") String oid,
             @Param("algorithmFamily") String algorithmFamily, @Param("primitive") String primitive,
@@ -194,7 +200,8 @@ public interface CryptoAssetRepository extends SecurityFilterRepository<CryptoAs
      * from -- the sweep's transaction is {@code READ COMMITTED}, so a second statement would see a later snapshot and a
      * write landing between the two would be invisible to the guard. {@code merged_crypto_properties} comes back as
      * text because the evaluator wants a {@code JsonNode}: the entity converter would build a {@code Map} only for the
-     * sweep to serialize it again.
+     * sweep to serialize it again. {@code curve} comes back joined on {@code +} because the record and the rules read
+     * the identity spelling, which the column stores split.
      *
      * <p>
      * <b>Stale</b> is either half of the contract: a verdict from an older generation of the rules, or a verdict older
@@ -215,7 +222,7 @@ public interface CryptoAssetRepository extends SecurityFilterRepository<CryptoAs
                    algorithm_family,
                    primitive,
                    parameter_set,
-                   curve,
+                   array_to_string(curve, '+') AS curve,
                    mode,
                    padding,
                    variant,
@@ -335,14 +342,17 @@ public interface CryptoAssetRepository extends SecurityFilterRepository<CryptoAs
             """, nativeQuery = true)
     List<String> findDistinctParameterSet();
 
+    /**
+     * Every curve any asset touches, once each -- not every combination once.
+     *
+     * <p>
+     * The sibling finders skip along a btree with a recursive loose index scan. That cannot work here: the distinct
+     * values are the array's elements, and no index orders them. The unnest is a sequential scan, which is what the
+     * value list for a membership filter costs.
+     */
     @Query(value = """
-            WITH RECURSIVE vals AS (
-                SELECT min(curve) AS v FROM {h-schema}crypto_asset
-                UNION ALL
-                SELECT (SELECT min(curve) FROM {h-schema}crypto_asset WHERE curve > vals.v)
-                FROM vals WHERE vals.v IS NOT NULL
-            )
-            SELECT v FROM vals WHERE v IS NOT NULL ORDER BY v
+            SELECT DISTINCT member FROM {h-schema}crypto_asset a, unnest(a.curve) AS member
+            WHERE member IS NOT NULL ORDER BY member
             """, nativeQuery = true)
     List<String> findDistinctCurve();
 
--- a/src/main/java/com/otilm/core/enums/FilterField.java
+++ b/src/main/java/com/otilm/core/enums/FilterField.java
@@ -408,7 +408,10 @@ public enum FilterField {
             SearchFieldTypeEnum.LIST),
     CBOM_ASSET_PARAMETER_SET(Resource.CRYPTO_ASSET, null, null, CryptoAsset_.parameterSet, "Parameter Set",
             SearchFieldTypeEnum.LIST),
-    CBOM_ASSET_CURVE(Resource.CRYPTO_ASSET, null, null, CryptoAsset_.curve, "Elliptic Curve", SearchFieldTypeEnum.LIST),
+    // A hybrid scheme's curve is stored as its members, so EQUALS asks whether the asset touches the curve rather
+    // than whether its whole composite is that curve. NATIVE_ARRAY still reports FilterFieldType.LIST on the wire.
+    CBOM_ASSET_CURVE(Resource.CRYPTO_ASSET, null, null, CryptoAsset_.curve, "Elliptic Curve",
+            SearchFieldTypeEnum.NATIVE_ARRAY),
     CBOM_ASSET_MODE(Resource.CRYPTO_ASSET, null, null, CryptoAsset_.mode, "Mode", SearchFieldTypeEnum.LIST),
     CBOM_ASSET_PADDING(Resource.CRYPTO_ASSET, null, null, CryptoAsset_.padding, "Padding", SearchFieldTypeEnum.LIST),
     CBOM_ASSET_VARIANT(Resource.CRYPTO_ASSET, null, null, CryptoAsset_.variant, "Variant", SearchFieldTypeEnum.LIST),
--- a/src/main/java/com/otilm/core/util/PostgresFunctionContributor.java
+++ b/src/main/java/com/otilm/core/util/PostgresFunctionContributor.java
@@ -12,6 +12,32 @@ public class PostgresFunctionContributor implements FunctionContributor {
     public static final String ARRAY_CONTAINS = "text_array_contains";
     public static final String ARRAY_ITEM_CONTAINS = "text_array_item_contains";
 
+    /**
+     * Scalar membership in a native array column, written as containment rather than as the equivalent
+     * {@code CAST(?1 AS TEXT) = ANY(?2)}.
+     *
+     * <p>
+     * The two answer the same question for the single non-null value this is ever called with -- every caller in
+     * {@code FilterPredicatesBuilder} passes {@code value.toString()} as a literal -- but only containment can be
+     * answered by a GIN index. PostgreSQL has no index path for {@code scalar = ANY(column)} at all, so the
+     * {@code = ANY} form made every membership filter a sequential scan whatever index the column carried.
+     *
+     * <p>
+     * The column is cast because the array columns are not all one type: a {@code List<String>} mapped with
+     * {@code SqlTypes.ARRAY} generates {@code varchar[]} unless the entity says otherwise, and {@code @>} is not
+     * defined across {@code varchar[]} and {@code text[]}. On a {@code text[]} column the cast is to the type the
+     * column already has, which PostgreSQL drops, so the index still matches -- pinned by
+     * {@code CryptoAssetCurveMembershipMigrationITest}.
+     *
+     * <p>
+     * That last part is what indexability rests on, and it is a property of the column rather than of the index: on a
+     * {@code varchar[]} column the cast is a real one, and a GIN index there is never matched through it. An entity
+     * earns the indexable type by declaring {@code columnDefinition = "TEXT[]"}, as {@code CryptoAsset.curve} does and
+     * the other array-mapped fields do not -- so adding an array index to one of those would silently buy nothing until
+     * its column type moves too.
+     */
+    public static final String ARRAY_CONTAINS_PATTERN = "CAST(?2 AS TEXT[]) @> ARRAY[CAST(?1 AS TEXT)]";
+
     @Override
     public void contributeFunctions(FunctionContributions functionContributions) {
         BasicType<Integer> resultType = functionContributions
@@ -26,10 +52,9 @@ public class PostgresFunctionContributor implements FunctionContributor {
         functionContributions
                 .getFunctionRegistry()
                 .registerPattern(JSONB_CONTAINS, "jsonb_contains(?1, ?2::jsonb)", booleanType);
-        // CAST(?1 AS TEXT) = ANY(?2) — check scalar membership in a native text[] column
         functionContributions
                 .getFunctionRegistry()
-                .registerPattern(ARRAY_CONTAINS, "CAST(?1 AS TEXT) = ANY(?2)", booleanType);
+                .registerPattern(ARRAY_CONTAINS, ARRAY_CONTAINS_PATTERN, booleanType);
         // EXISTS over unnest(...) + LIKE — check substring match in any native text[] item
         functionContributions
                 .getFunctionRegistry()
--- a/src/main/resources/db/migration/V202609101000__crypto_asset_curve_membership.sql
+++ b/src/main/resources/db/migration/V202609101000__crypto_asset_curve_membership.sql
@@ -1,0 +1,17 @@
+-- A hybrid scheme (X25519/X448, and every transitional pairing after it) reports one curve value holding both
+-- members joined on '+'. Held as scalar TEXT, "show me every asset that touches curve25519" -- the sweep this
+-- inventory exists to serve -- silently skipped exactly those rows. The column becomes an array so a curve can be
+-- matched by membership; a single-curve asset is a one-element array and an absent curve stays NULL.
+--
+-- The identity preimage is unaffected: it keeps the '+'-joined spelling, so no row is re-keyed and the identity
+-- rule-set version does not move. This is a projection change, not an identity change.
+ALTER TABLE "crypto_asset"
+    ALTER COLUMN "curve" TYPE TEXT[] USING CASE WHEN "curve" IS NULL THEN NULL ELSE string_to_array("curve", '+') END;
+
+-- The btree answered equality on the scalar and cannot answer membership in an array. GIN can, but only for a
+-- containment predicate: PostgreSQL has no index path for 'scalar = ANY(column)' at all, which is why the membership
+-- filter is emitted as containment (PostgresFunctionContributor.ARRAY_CONTAINS_PATTERN). The pairing is pinned by
+-- CryptoAssetCurveMembershipMigrationITest, which plans the shipped predicate against this index.
+DROP INDEX "idx_crypto_asset_curve";
+
+CREATE INDEX "idx_crypto_asset_curve" ON "crypto_asset" USING GIN ("curve");
--- a/src/test/java/com/otilm/core/integration/search/CryptoAssetSearchITest.java
+++ b/src/test/java/com/otilm/core/integration/search/CryptoAssetSearchITest.java
@@ -188,6 +188,32 @@ class CryptoAssetSearchITest extends BaseSpringBootTest {
                 .isEmpty();
     }
 
+    /**
+     * The reason the column is an array. A hybrid scheme names two curves in one asset, and the query this inventory
+     * exists to answer -- every asset that touches curve X -- has to reach it from either member. The composite itself
+     * is deliberately not selectable: it is a spelling of the pair, not a curve, and it is not offered as a value.
+     */
+    @Test
+    void aHybridAssetIsFoundByEitherOfItsCurvesAndNotByTheirCompositeSpelling() {
+        UUID hybrid = upsert(new CryptoAssetIdentityFields(CryptographicAssetType.ALGORITHM, "X25519/X448",
+                "1.3.101.110", "ecdh", "key-agree", null, "other/curve25519+other/curve448", null, null, null), null);
+
+        assertThat(search(aPropertyEqualsFilter(FilterField.CBOM_ASSET_CURVE, "other/curve25519")))
+                .containsExactly(hybrid);
+        assertThat(search(aPropertyEqualsFilter(FilterField.CBOM_ASSET_CURVE, "other/curve448")))
+                .containsExactly(hybrid);
+        assertThat(search(aPropertyEqualsFilter(FilterField.CBOM_ASSET_CURVE, "other/curve25519+other/curve448")))
+                .describedAs("the composite matches no member, so it selects nothing")
+                .isEmpty();
+        assertThat(search(aPropertyNotEmptyFilter(FilterField.CBOM_ASSET_CURVE)))
+                .describedAs("a hybrid has curves like any other asset")
+                .containsExactlyInAnyOrder(populated, hybrid);
+        assertThat(search(
+                aPropertyFilter(FilterField.CBOM_ASSET_CURVE, FilterConditionOperator.NOT_EQUALS, "other/curve25519")))
+                .describedAs("excluding a member excludes every asset that touches it")
+                .doesNotContain(hybrid);
+    }
+
     // ---- free text, refuted-OID and source-CBOM filters ----
 
     @Test
--- a/.github/workflows/build_pr.yml
+++ b/.github/workflows/build_pr.yml
@@ -15,6 +15,197 @@ permissions:
   packages: read
 
 jobs:
+  # A marker in the PR body redirects this build at an unmerged interfaces PR's
+  # snapshot. The pom is never touched: core dismisses approvals on push and
+  # forbids self-approval, so a revert commit would cost a second review round.
+  resolve-interfaces:
+    name: Resolve interfaces version
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: read
+      pull-requests: read
+    outputs:
+      mvnarg:   ${{ steps.resolve.outputs.mvnarg }}
+      override: ${{ steps.resolve.outputs.override }}
+    steps:
+      # No checkout: nothing here reads the working tree.
+      - id: resolve
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          EVENT: ${{ github.event_name }}
+          PR: ${{ github.event.pull_request.number }}
+        run: |
+          set -euo pipefail
+
+          emit() { echo "mvnarg=$1" >>"$GITHUB_OUTPUT"; echo "override=$2" >>"$GITHUB_OUTPUT"; }
+          fail() { echo "::error::$1"; exit 1; }
+
+          # merge_group carries no pull_request object. Unreachable under the
+          # current triggers; kept so adding one cannot silently resolve wrong.
+          if [ "$EVENT" != pull_request ]; then
+            echo "Not a pull_request event - using the pom value."
+            emit "" "none"; exit 0
+          fi
+
+          # Live, not github.event.pull_request.body: a re-run replays the original
+          # payload, so an edited marker would be invisible and the override could
+          # never clear without a push, which dismisses the approval.
+          if ! BODY=$(gh api "repos/${GITHUB_REPOSITORY}/pulls/$PR" --jq '.body // ""'); then
+            fail "Could not read the body of ${GITHUB_REPOSITORY}#${PR}."
+          fi
+
+          # Fenced blocks and HTML comments go first: the regexes are line-oriented,
+          # so a body quoting the marker syntax would otherwise pin the build.
+          BODY=$(printf '%s\n' "$BODY" | python3 -c 'import re,sys; t=sys.stdin.read(); t=re.sub(r"<!--.*?-->","",t,flags=re.S); t=re.sub(r"^[ \t]*(```|~~~).*?^[ \t]*\1","",t,flags=re.S|re.M); sys.stdout.write(t)')
+
+          # head -1 below would otherwise take the earlier line silently.
+          DUPD=$(printf '%s' "$BODY" | grep -ciP '^[[:space:]>*_+#.)0-9-]*Depends-On[[:space:]*_]*:[[:space:]*_]*(https?://github\.com/)?OmniTrustILM/interfaces(?![0-9A-Za-z._-])' || true)
+          DUPV=$(printf '%s' "$BODY" | grep -ciP '^[[:space:]>*_+#.)0-9-]*Interfaces-Version[[:space:]*_]*:' || true)
+          [ "${DUPD:-0}" -le 1 ] || fail "The PR body has ${DUPD} Depends-On lines. Keep exactly one."
+          [ "${DUPV:-0}" -le 1 ] || fail "The PR body has ${DUPV} Interfaces-Version lines. Keep exactly one."
+
+          NUM=$(printf '%s' "$BODY" \
+            | grep -oiP '^Depends-On:[[:space:]]*OmniTrustILM/interfaces#\K[0-9]{1,7}(?=[[:space:]]*$)' \
+            | head -1 || true)
+          VER=$(printf '%s' "$BODY" \
+            | grep -oiP '^Interfaces-Version:[[:space:]]*\K[0-9][0-9A-Za-z.\-]{0,60}(?=[[:space:]]*$)' \
+            | head -1 || true)
+          # Strict extractors, loose probes.
+          #
+          # Strict: line start, nothing after the value - `Interfaces-Version:
+          # 2.20.0; rm -rf /` is rejected, not truncated to a pin on 2.20.0.
+          #
+          # Loose: a decorated marker misses the extractor, so the probe catches it
+          # and fails loudly. The Depends-On probe closes off the repository name,
+          # so another repo's marker - or a same-prefix sibling - is ignored.
+          if printf '%s' "$BODY" | grep -qiP '^[[:space:]>*_+#.)0-9-]*Depends-On[[:space:]*_]*:[[:space:]*_]*(https?://github\.com/)?OmniTrustILM/interfaces(?![0-9A-Za-z._-])' && [ -z "$NUM" ]; then
+            fail "A Depends-On: line for interfaces is present but not in the accepted form. It must begin the line and read: Depends-On: OmniTrustILM/interfaces#<number> - a pull-request URL is not accepted."
+          fi
+          if printf '%s' "$BODY" | grep -qiP '^[[:space:]>*_+#.)0-9-]*Interfaces-Version[[:space:]*_]*:' && [ -z "$VER" ]; then
+            fail "An Interfaces-Version: line is present but not in the accepted form. It must begin the line and name a published version, e.g. Interfaces-Version: 2.20.0-M906-480022e4"
+          fi
+
+          # Both markers set: name the loser rather than letting it disappear.
+          if [ -n "$VER" ] && [ -n "$NUM" ]; then
+            echo "::warning::Both markers present - Interfaces-Version wins; ignoring Depends-On: OmniTrustILM/interfaces#${NUM}."
+          fi
+
+          # Reaches the build jobs through the environment, never interpolated into
+          # their run: blocks, so the value is never shell source text.
+          if [ -n "$VER" ]; then
+            echo "Explicit override: $VER"
+            emit "-Dinterfaces.version=$VER" "explicit"; exit 0
+          fi
+
+          if [ -z "$NUM" ]; then
+            emit "" "none"; exit 0
+          fi
+
+          # Captured, not tested inline: a substitution inside an `if` does not trip
+          # `set -e`, so a 404 or rate limit would read as "not merged".
+          if ! IPR=$(gh api "repos/OmniTrustILM/interfaces/pulls/$NUM" --jq '[.merged, .state, (.merge_commit_sha // ""), (.merged_at // "")] | @tsv'); then
+            fail "Cannot read OmniTrustILM/interfaces#${NUM} - check that the number names an existing pull request."
+          fi
+          read -r MERGED ISTATE MSHA MAT <<<"$IPR"
+
+          if ! VERSIONS=$(gh api --paginate \
+                  "/orgs/OmniTrustILM/packages/maven/com.otilm.interfaces/versions?per_page=100" \
+                  --jq '.[] | [.created_at, .name] | @tsv'); then
+            fail "Could not list the published interfaces package versions."
+          fi
+
+          # Self-clearing: once the interfaces PR merges, the same unchanged
+          # marker stops meaning "override" and starts meaning "nothing to do".
+          if [ "$MERGED" = true ]; then
+            # Mainline SNAPSHOT carries the change only once the interfaces
+            # publish finishes. The merge commit's -M build is evidence that run
+            # reached its pin job, not that the SNAPSHOT deploy in it succeeded -
+            # that job mints the pin even when it did not.
+            if ! printf '%s\n' "$VERSIONS" | cut -f2 \
+                 | grep -qE "^[0-9][0-9A-Za-z.-]*-M[0-9]+-${MSHA:0:8}\$"; then
+              AGE=$(( $(date -u +%s) - $(date -u -d "${MAT:-@0}" +%s) ))
+              if [ "$AGE" -lt 900 ]; then
+                fail "interfaces#${NUM} merged ${AGE}s ago but its build has not published yet. Wait for it, then re-run all jobs."
+              fi
+              echo "::warning::interfaces#${NUM} merged ${AGE}s ago with no -M build; a release-cut merge produces none. Proceeding against mainline."
+            fi
+            echo "interfaces#$NUM is merged - using the mainline pom value."
+            emit "" "none"; exit 0
+          fi
+
+          # A PR closed without merging would otherwise read as "still open" and keep
+          # this build pinned to an abandoned branch; its snapshot outlives it.
+          if [ "$ISTATE" = closed ]; then
+            fail "OmniTrustILM/interfaces#${NUM} was closed without merging. Remove the Depends-On line, or repoint it at whatever superseded it."
+          fi
+
+          # Looked up, not derived from core's own property: across a release cut
+          # interfaces publishes 2.21.0-alpha-PR<n>-SNAPSHOT while this pom still
+          # reads 2.20.0.
+          #
+          # The whole name is anchored, not just the suffix - the prefix comes from
+          # the interfaces PR's own pom, so it would otherwise carry PR-controlled
+          # text into the mvn argument. Newest by created_at, not API order, which
+          # is not contractual and matters when a PR spans a release cut.
+          PRV=$(printf '%s\n' "$VERSIONS" \
+            | awk -F'\t' -v n="$NUM" '$2 ~ "^[0-9][0-9A-Za-z.-]*-alpha-PR" n "-SNAPSHOT$"' \
+            | sort -r | head -1 | cut -f2)
+
+          if [ -z "$PRV" ]; then
+            fail "No published snapshot for interfaces#${NUM}. It needs the publish-snapshot label and a finished build. Fork PRs never publish - their token is read-only."
+          fi
+
+          echo "interfaces#$NUM is open - building against $PRV"
+          emit "-Dinterfaces.version=$PRV" "pr"
+
+      # sonar.yaml is a separate workflow and cannot read job outputs; without this
+      # it analyses the compiled classes against a classpath they were not built with.
+      - name: Publish the resolved argument
+        env:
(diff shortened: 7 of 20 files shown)


src/main/java/com/otilm/core/util/PostgresFunctionContributor.java
```diff
@@ -12,6 +12,32 @@ public class PostgresFunctionContributor implements FunctionContributor {
...
+     * defined across {@code varchar[]} and {@code text[]}. On a {@code text[]} column the cast is to the type the
+     * column already has, which PostgreSQL drops, so the index still matches -- pinned by
+     * {@code CryptoAssetCurveMembershipMigrationITest}.
+     *
+     * <p>
+     * That last part is what indexability rests on, and it is a property of the column rather than of the index: on a
+     * {@code varchar[]} column the cast is a real one, and a GIN index there is never matched through it. An entity
+     * earns the indexable type by declaring {@code columnDefinition = "TEXT[]"}, as {@code CryptoAsset.curve} does and
+     * the other array-mapped fields do not -- so adding an array index to one of those would silently buy nothing until
+     * its column type moves too.
```
karolhrdina: **The production claim here is inverted.** The Javadoc says only `CryptoAsset.curve` earns `text[]` and that "adding an array index to one of [the other array-mapped fields] would silently buy nothing until its column type moves too". That is false for the shipped schema: all three are already declared `TEXT[]` by their own migrations —

- `alt_codes TEXT[]` — `V202507221212__custom_oid_entry.sql:7`
- `features TEXT[] NULL` — `V202602131307__connectors_v2.sql:21`
- `"ntp_servers" TEXT[] NOT NULL` — `V202605011200__time_quality_configuration.sql:5`

and no later migration alters any of them. `columnDefinition` on the entity only shapes the *entity-generated* schema the ITests run against; it has no bearing on production DDL. So `CAST(col AS TEXT[])` is a no-op on all four columns in production, and a GIN index on `alt_codes`/`features`/`ntp_servers` would be matched today, with no type change needed.

Concrete cost: this constant is called out in the PR description as the thing a reviewer should not have to discover, and #2170 builds on it. Whoever acts on this paragraph either writes a pointless `ALTER COLUMN ... TYPE TEXT[]` migration for a column that is already `text[]`, or declines an index that would in fact have worked.

The cast itself is still worth keeping (the test schema really does generate `varchar(255)[]`), so only the conclusion needs correcting — say that the cast exists for the entity-generated schema, and that production columns are already `text[]`.

src/main/java/com/otilm/core/util/PostgresFunctionContributor.java
```diff
@@ -26,10 +52,9 @@ public void contributeFunctions(FunctionContributions functionContributions) {
         functionContributions
                 .getFunctionRegistry()
                 .registerPattern(JSONB_CONTAINS, "jsonb_contains(?1, ?2::jsonb)", booleanType);
-        // CAST(?1 AS TEXT) = ANY(?2) — check scalar membership in a native text[] column
         functionContributions
                 .getFunctionRegistry()
-                .registerPattern(ARRAY_CONTAINS, "CAST(?1 AS TEXT) = ANY(?2)", booleanType);
+                .registerPattern(ARRAY_CONTAINS, ARRAY_CONTAINS_PATTERN, booleanType);
```
karolhrdina, line 66: Stale comment left behind by this change, on the enum that declares the type: `SearchFieldTypeEnum.NATIVE_ARRAY` (`src/main/java/com/otilm/core/enums/SearchFieldTypeEnum.java:48-49`) still reads

```java
// Like LIST but the underlying column is a native PostgreSQL array (text[]).
// EQUALS/NOT_EQUALS use = ANY(column) instead of column = value.
```

That is now the shape the migration comment and `ARRAY_CONTAINS_PATTERN`'s Javadoc both say must never come back — `= ANY` has no index path at all, so restoring it un-indexes every `NATIVE_ARRAY` membership filter. The enum is where someone adding the next array-backed field will look first, and it is the only place in the codebase still documenting the old predicate. `CryptoAssetCurveMembershipMigrationITest` would catch the revert, but only after the reader has been told the wrong thing.

src/test/java/com/otilm/core/integration/search/CryptoAssetSearchITest.java
```diff
@@ -188,6 +188,32 @@ void aFilterCombinesWithAnother() {
...
+        assertThat(search(aPropertyEqualsFilter(FilterField.CBOM_ASSET_CURVE, "other/curve25519+other/curve448")))
+                .describedAs("the composite matches no member, so it selects nothing")
+                .isEmpty();
+        assertThat(search(aPropertyNotEmptyFilter(FilterField.CBOM_ASSET_CURVE)))
+                .describedAs("a hybrid has curves like any other asset")
+                .containsExactlyInAnyOrder(populated, hybrid);
+        assertThat(search(
+                aPropertyFilter(FilterField.CBOM_ASSET_CURVE, FilterConditionOperator.NOT_EQUALS, "other/curve25519")))
+                .describedAs("excluding a member excludes every asset that touches it")
+                .doesNotContain(hybrid);
```
karolhrdina: `doesNotContain(hybrid)` is the one assertion in this test that cannot fail for the thing the predicate change actually risks.

`getNativeArrayNotEqualsExpression` renders `curve IS NULL OR NOT (curve @> ARRAY[v])`. The `IS NULL` arm is exactly what the `= ANY` → `@>` move interacts with: under `= ANY` a NULL array yielded SQL NULL rather than false, so the arm was load-bearing for keeping absent-curve rows in a `NOT_EQUALS` result; under containment the arm is still needed, because `NULL @> ARRAY[v]` is still NULL. `bare` (curve NULL) and `populated` (curve `{secp256r1}`) both have to come back here, and neither is asserted — the line passes if the predicate returns nothing at all.

Suggest pinning the set, which is what makes the NULL arm non-removable:

```suggestion
        assertThat(search(
                aPropertyFilter(FilterField.CBOM_ASSET_CURVE, FilterConditionOperator.NOT_EQUALS, "other/curve25519")))
                .describedAs("excluding a member excludes every asset that touches it, and keeps the rest -- including an absent curve")
                .containsExactlyInAnyOrder(populated, bare);
```

src/main/java/com/otilm/core/enums/FilterField.java
```diff
@@ -408,7 +408,10 @@ public enum FilterField {
             SearchFieldTypeEnum.LIST),
     CBOM_ASSET_PARAMETER_SET(Resource.CRYPTO_ASSET, null, null, CryptoAsset_.parameterSet, "Parameter Set",
             SearchFieldTypeEnum.LIST),
-    CBOM_ASSET_CURVE(Resource.CRYPTO_ASSET, null, null, CryptoAsset_.curve, "Elliptic Curve", SearchFieldTypeEnum.LIST),
+    // A hybrid scheme's curve is stored as its members, so EQUALS asks whether the asset touches the curve rather
+    // than whether its whole composite is that curve. NATIVE_ARRAY still reports FilterFieldType.LIST on the wire.
+    CBOM_ASSET_CURVE(Resource.CRYPTO_ASSET, null, null, CryptoAsset_.curve, "Elliptic Curve",
```
karolhrdina, line 413: **Six operators that worked on `main` now throw a 500 on this field.**

`SearchFieldTypeEnum` gates which conditions are *advertised*, but nothing validates the condition a client actually submits on this path. `ListViewServiceImpl.validateFilters` is the only operator check in the codebase and it covers saved list views; the crypto-asset listing goes through `CryptographicAssetServiceImpl.listCryptographicAssets` → `RequestValidatorHelper.revalidateSearchRequestDto`, which validates paging and sort only. So an unadvertised operator reaches `FilterPredicatesBuilder`'s generic branch and is applied to the column.

While `curve` was `TEXT` that produced valid — if unadvertised — SQL. Against `text[]` it does not. I ran the same six operators through `FilterPredicatesBuilder` on both branches, real Postgres:

| operator | `main` | this branch |
|---|---|---|
| `STARTS_WITH` | works | `SemanticException: Operand of 'like' is of type 'java.util.List<java.lang.String>' which is not a string` |
| `ENDS_WITH` | works | same |
| `MATCHES` | works | `PSQLException: function textregexeq(text[], unknown) does not exist` |
| `NOT_MATCHES` | works | same |
| `GREATER` | works | `SemanticException: Cannot compare left expression of type 'java.util.List<java.lang.String>' with 'java.lang.String'` |
| `LESSER` | works | same |

All six surface as `InvalidDataAccessResourceUsageException`, and `ExceptionHandlingAdvice` has no handler for `DataAccessException`, so they fall through to **500**, not 422.

None of these was ever advertised for `CBOM_ASSET_CURVE`, so this is not a contract break in the strict sense. But any stored or scripted filter using one flips from returning rows to erroring, and it errors as a server fault rather than a rejected request. Worth either rejecting unadvertised operators up front on this path, or noting the change explicitly.

src/main/resources/db/migration/V202609141000__crypto_asset_curve_membership.sql
```diff
@@ -0,0 +1,17 @@
...
+ALTER TABLE "crypto_asset"
+    ALTER COLUMN "curve" TYPE TEXT[] USING CASE WHEN "curve" IS NULL THEN NULL ELSE string_to_array("curve", '+') END;
+
+-- The btree answered equality on the scalar and cannot answer membership in an array. GIN can, but only for a
+-- containment predicate: PostgreSQL has no index path for 'scalar = ANY(column)' at all, which is why the membership
+-- filter is emitted as containment (PostgresFunctionContributor.ARRAY_CONTAINS_PATTERN). The pairing is pinned by
+-- CryptoAssetCurveMembershipMigrationITest, which plans the shipped predicate against this index.
+DROP INDEX "idx_crypto_asset_curve";
+
+CREATE INDEX "idx_crypto_asset_curve" ON "crypto_asset" USING GIN ("curve");
```
karolhrdina, line 45: **`EMPTY` / `NOT_EMPTY` lose their index path here — measured, and it is not mentioned anywhere in the PR.**

The dropped btree indexed NULLs and could answer `curve IS NULL`. `array_ops` GIN cannot, and it cannot answer `cardinality(curve) = 0` either. `EXPLAIN` on `postgres:15-alpine`, 20 000 rows, `enable_seqscan = off`:

```
OLD btree, EMPTY:
  Index Scan using zz_old_ix on zz_old
    Index Cond: (curve IS NULL)

NEW GIN, EMPTY  (curve IS NULL OR cardinality(curve) = 0):
  Seq Scan on zz_new  (cost=10000000000.00..10000000507.00)

NEW GIN, NOT_EMPTY:
  Seq Scan on zz_new  (cost=10000000000.00..10000000507.00)

NEW GIN, EQUALS (the shipped containment predicate):
  Bitmap Heap Scan -> Bitmap Index Scan on zz_new_ix   ✓
```

So of the six conditions `NATIVE_ARRAY` now advertises on this field, exactly one is index-answerable. `EQUALS` gains an index path it genuinely never had; `EMPTY`/`NOT_EMPTY` lose one they had. `NOT_EQUALS`/`NOT_CONTAINS` are negations OR'd with `IS NULL` and were never GIN-answerable; `CONTAINS`/`NOT_CONTAINS` go through `ARRAY_ITEM_CONTAINS` (`unnest` + `LIKE`), which no index serves.

That is very likely the right trade — `EQUALS` is the query the inventory exists to serve, and the table is empty today — but the migration comment and the `ARRAY_CONTAINS_PATTERN` Javadoc both present this as strictly fixing indexability, and a reader would not learn that two conditions regressed. A partial index (`WHERE curve IS NULL`) would restore it if `EMPTY` matters; otherwise a sentence in the migration comment.

src/main/java/com/otilm/core/cbom/asset/CompositeCurve.java
```diff
@@ -0,0 +1,29 @@
...
+ * <p>
+ * A hybrid scheme names more than one curve, and the normalizer renders those as a single {@code +}-joined token in
+ * sorted, deduplicated order. That token is what the ratified identity vectors hash, so its shape can never change; the
+ * column holds the members split out of it, which is what lets a filter ask whether an asset touches a curve rather
+ * than whether its whole composite equals one. The split preserves the normalizer's order, so joining the members back
+ * reproduces the preimage spelling byte for byte -- which is what the API contract keeps carrying.
+ */
+public final class CompositeCurve {
+
+    private static final String SEPARATOR = "+";
```
karolhrdina, line 23: **The round-trip invariant is spelled out in five independent places, and the DTO half is unpinned.**

`join(split(x)) == x` is what lets the identity preimage stay untouched while storage changes shape. The `+` that makes it true is written separately at:

1. `AssetNormalizer.joinCurves` — `String.join("+", distinct)` (`AssetNormalizer.java:1044`)
2. `CompositeCurve.SEPARATOR` — this line
3. `CryptoAssetRepository:89` — `string_to_array(CAST(:curve AS TEXT), '+')`
4. `CryptoAssetRepository:225` — `array_to_string(curve, '+')`
5. `V202609101000:9` — `string_to_array("curve", '+')`

No shared constant, and no test crosses more than one of them. I mutated each in isolation:

| mutated | caught by |
|---|---|
| `array_to_string` sep (4) | `CryptoAssetPqcSweepITest` ✓ |
| `CompositeCurve.SEPARATOR` (2) | **nothing** — `CryptographicAssetControllerITest`, `CryptoAssetInventoryITest`, `PqcStoredRowRoundTripITest`, `CryptographicAssetSearchableFieldsITest`: **55 tests, all green with the separator changed to `~`** |

Only `CompositeCurveTest` catches (2), and it hand-builds the `List`, so it never crosses Hibernate or the DTO. The PR's headline claim — *"leaving the wire contract byte-identical"* — therefore has no end-to-end pin for a hybrid. The only DTO-level curve assertion in the repo is `CryptographicAssetDetailITest:439`, and it is a single curve.

Two cheap fixes, and the second matters for what comes after this PR:

- Add one `+`-curve vector to `PqcStoredRowRoundTripITest.singleProducerVectors` — that single line covers the entity array-order round-trip *and* the joined-equals-normalizer-spelling claim.
- Give `CompositeCurve` the `split` to match its `join`, bind `List<String>` from the writer, and drop `string_to_array` from the upsert. Then (1)–(3) collapse to one definition and the projection lives in Java.

That second one is also the forward-compat point for **#2166**. Its note says *"a fold placed at the writer will need to fold per member."* With the split happening inside the SQL statement, a writer-side fold cannot see members — it would have to fold the joined string before binding, or the split has to move to Java anyway. Since this PR is meant to be the baseline that work builds on, moving it now is cheaper than moving it later.

src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
```diff
@@ -335,14 +342,17 @@ SELECT min(parameter_set) AS v FROM {h-schema}crypto_asset
...
     @Query(value = """
-            WITH RECURSIVE vals AS (
-                SELECT min(curve) AS v FROM {h-schema}crypto_asset
-                UNION ALL
-                SELECT (SELECT min(curve) FROM {h-schema}crypto_asset WHERE curve > vals.v)
-                FROM vals WHERE vals.v IS NOT NULL
-            )
-            SELECT v FROM vals WHERE v IS NOT NULL ORDER BY v
+            SELECT DISTINCT member FROM {h-schema}crypto_asset a, unnest(a.curve) AS member
+            WHERE member IS NOT NULL ORDER BY member
```
karolhrdina, line 367: **Consequence worth naming beyond the scan itself: this neutralises the sibling finders' optimisation.**

`getSearchableFieldInformationByGroup` calls all eight value-list queries in one request, sequentially (`CryptographicAssetServiceImpl.java:218-242`). The javadoc two methods up (`:306-310`) explains that `SELECT DISTINCT` was rejected precisely because a full scan costs *"hundreds of milliseconds per column at millions of rows, for eight columns on every filter-panel open"*, and that the recursive loose index scan makes each one *"reliably milliseconds"*.

Once one of those eight is a full unnest scan, the panel's latency is that scan. The seven skip scans still avoid seven scans, but they no longer buy the panel the property they were written for — the request cannot be fast while curve is in it. So the cost is not "one column got slower", it is "the documented budget for this endpoint no longer holds".

Nothing to change in this PR necessarily — the table is empty until core#2073 — but the sibling javadoc at `:306-310` now overstates what it delivers, and whatever fixes this (expression index feeding a skip scan, a side table of members, or a cached/materialised value list) is the same work that would let curve rejoin the budget. Worth a line there pointing at this method so the next reader does not trust the old guarantee.

src/main/resources/db/migration/V202609101000__crypto_asset_curve_membership.sql
```diff
@@ -0,0 +1,17 @@
+-- A hybrid scheme (X25519/X448, and every transitional pairing after it) reports one curve value holding both
+-- members joined on '+'. Held as scalar TEXT, "show me every asset that touches curve25519" -- the sweep this
+-- inventory exists to serve -- silently skipped exactly those rows. The column becomes an array so a curve can be
+-- matched by membership; a single-curve asset is a one-element array and an absent curve stays NULL.
+--
+-- The identity preimage is unaffected: it keeps the '+'-joined spelling, so no row is re-keyed and the identity
+-- rule-set version does not move. This is a projection change, not an identity change.
+ALTER TABLE "crypto_asset"
+    ALTER COLUMN "curve" TYPE TEXT[] USING CASE WHEN "curve" IS NULL THEN NULL ELSE string_to_array("curve", '+') END;
```
karolhrdina: **This rewrite is free only because it lands before ingest — worth stating as a merge-order constraint rather than leaving it implicit.**

`ALTER COLUMN ... TYPE ... USING` takes `ACCESS EXCLUSIVE` and fully rewrites the table; the `CREATE INDEX` on line 17 then blocks writes again. Both are free today: `IdentityRuleset.java:12` records that `CryptoAssetWriter` has no production caller, and grep confirms no caller outside tests, so `crypto_asset` is empty in every environment.

If this slips behind core#2073, the same two statements become an outage-class migration on a populated inventory — and `CONCURRENTLY` is not an available remedy here, because Flyway wraps the script in a transaction and `CREATE INDEX CONCURRENTLY` cannot run inside one. It would need `executeInTransaction=false`, which no migration in this repo uses.

The ITest javadoc mentions the table is expected to be empty. Since that is what makes the DDL safe rather than just makes the test convenient, it belongs in this comment too.

Separately, two conversion shapes the `USING` clause handles differently from what the surrounding code assumes. On real PG 15, `string_to_array('', '+')` returns `{}` — an empty array, not `NULL` and not `{""}`. A pre-existing row with an empty-string curve therefore converts to `{}`, and then:

- `CompositeCurve.join` maps an empty list to `null`, but `array_to_string('{}', '+')` returns `''` — so the same row reports `null` on the detail endpoint and `''` to the PQC sweep.
- `PqcEvaluator:240` tests `input.curve() != null`, which `''` passes.
- `CryptographicAssetServiceImpl:582` tests `asset.getCurve() == null`, which `{}` does not satisfy, so a `normalizedFields` block is emitted with every field absent.

`CryptoAssetIdentityFields.fold` maps blank to `null`, so this is unreachable on data this codebase wrote — but the migration's stated job is converting data it did not write. `USING CASE WHEN NULLIF("curve", '') IS NULL THEN NULL ELSE string_to_array(...) END` closes it in one token, and makes the existing `CASE WHEN ... IS NULL` (redundant, since `string_to_array(NULL, '+')` is already NULL) actually earn its place.

karolhrdina (commented):
## Review summary

Reviewed at `4546bb9e`, with the branch checked out and the suite run locally against the pinned `postgres:15-alpine`. The design is right and the PR body is unusually honest about its own edges. Nothing here blocks merge on correctness; the substantive items are two measured behaviour changes the PR does not mention, and one test gap that leaves its headline claim unpinned.

### What I verified rather than read

| check | result |
|---|---|
| `CompositeCurveTest`, `CryptoAssetCurveMembershipMigrationITest`, `CryptoAssetSearchITest` | 30 tests green |
| Other `NATIVE_ARRAY` fields + PQC paths after the predicate change (connector features 16, OID alt codes 26, NTP servers 18, PQC sweep 11, round-trip 8, searchable fields 6, `SearchHelper` 17) | 102 tests green — the shared-predicate change breaks nothing |
| **Mutation:** revert `ARRAY_CONTAINS_PATTERN` to `= ANY` | `theCurveIndexAnswersThePredicateTheFilterEmits` **fails** — the pinning claim holds |
| **Mutation:** `array_to_string` separator → `~` | `CryptoAssetPqcSweepITest` **fails** — sweep side pinned |
| **Mutation:** `CompositeCurve.SEPARATOR` → `~` | **55 integration tests still green** — DTO side unpinned |
| Six unadvertised operators on `CBOM_ASSET_CURVE`, `main` vs this branch | all six work on `main`, all six 500 here |
| `EXPLAIN` of `EMPTY`/`NOT_EMPTY`/`EQUALS` against btree vs GIN | `EQUALS` gains an index path, `EMPTY`/`NOT_EMPTY` lose one |

Also confirmed sound, so nobody needs to re-check: the `?1`/`?2` reversal in the new pattern (Hibernate's `PatternRenderer` is index-based; `@>` outranks `=`/`AND`/`OR`/`NOT` in PostgreSQL precedence, so no parenthesisation hazard); the `+`-split assumption (no canonical or alias curve token in `identity-tables.json` contains `+`, and `CURVE_ALTERNATIVES` already splits on it, so `join`/`split` is a true inverse); the identity preimage genuinely cannot see the column (`CryptoAssetIdentity` reads `NormalizedAsset`, and the key is computed before `CryptoAssetWriter.upsertIdentity` is called); NULL-column handling across all six conditions (the `isNull(...) OR` arms at `FilterPredicatesBuilder:826-827` and `:856-857` are what keep NULL-curve rows in the negations); sorting is unreachable (`CRYPTO_ASSET` is not in `CONFIGURABLE_COLUMN_RESOURCES`, and the listing rejects any sort at `CryptographicAssetServiceImpl:379-383`); `COALESCE` on `text[]`, the `ALTER` having no dependent constraint or view, migration ordering, and the scratch-schema / `search_path` / `enable_seqscan` hygiene in the new ITest.

### Findings

Posted inline, in rough order of what I'd want addressed:

1. **Six operators that worked on `main` now 500** (`FilterField.java`) — nothing validates unadvertised operators on this path, and `ExceptionHandlingAdvice` has no `DataAccessException` handler.
2. **`EMPTY`/`NOT_EMPTY` lose their index path** (migration `:17`) — measured; only 1 of the 6 advertised conditions is index-answerable, while the migration comment and the pattern Javadoc present the swap as strictly fixing indexability.
3. **The `+` invariant lives in five places and the DTO half is unpinned** (`CompositeCurve.java`) — mutation-proven; includes the forward-compat argument for moving `split` into Java before #2166 builds on this.
4. **The sibling value-list optimisation is neutralised** (`CryptoAssetRepository:355`) — extends Copilot's scan finding: all eight finders run in one request, so the endpoint's documented budget no longer holds regardless of the other seven.
5. **Migration sequencing is load-bearing and implicit** (migration `:9`) — the rewrite is free only while `crypto_asset` is empty; plus the `''` → `{}` conversion shape, where `CompositeCurve.join` and `array_to_string` disagree.
6. **`ARRAY_CONTAINS_PATTERN` Javadoc has test and production inverted** — already posted separately. `alt_codes`, `features` and `ntp_servers` are all `TEXT[]` in Flyway (`V202507221212:7`, `V202602131307:21`, `V202605011200:5`) and never altered; the `varchar[]` shape exists only in the entity-generated test schema. So *"adding an array index to one of those would silently buy nothing until its column type moves too"* is false — a GIN index on any of the three is matchable today. Given the PR names this constant as the thing a reviewer should not have to rediscover, the wrong conclusion is the one most likely to get acted on.

### Test gaps

- **`NOT_EQUALS` against an array holding a NULL element** — the one semantic change the PR body explicitly flags, affecting three resources it does not name, has zero coverage in either direction. No test in the repo seeds a NULL *element* into any array column. One case in `NativeArrayFilterSearchITest` seeding `["OFI", null]` pins it.
- **No hybrid curve is ever read back through Hibernate and joined.** `PqcStoredRowRoundTripITest.singleProducerVectors` produces no multi-member curve, so the array-order round-trip is never exercised.
- **No DTO assertion that a hybrid serialises to the joined spelling.**
- **The `COALESCE(crypto_asset.curve, EXCLUDED.curve)` ON CONFLICT path is untested** — the fill-once tests cover `name`, `mode`, `oid`, never `curve`.
- `columnType()` asserts `data_type = 'ARRAY'`, which does not distinguish `text[]` from `varchar[]` — and `text[]`-ness is the premise of the whole indexability argument. `udt_name = '_text'` would pin it. Likewise the plan assertion could require `Index Cond:` alongside the index name, so it stays non-vacuous if an `ORDER BY` ever introduces a full-index-scan path.

### Doc drift

- `SearchFieldTypeEnum:48-49` still documents `EQUALS/NOT_EQUALS use = ANY(column)` — the exact shape this PR removes (posted inline).
- `V202608271000__crypto_asset_inventory.sql:3` still asserts "No GIN … in v1".
- `CryptographicAssetServiceImpl:210-212` still describes curve values as "the distinct stored spellings of the normalized column"; they are now members, never composites.

### On the two open questions in the description

The ticket-vs-implementation conflict is called correctly. `NATIVE_ARRAY` `EQUALS` is membership, never array equality; "touches both" is two ANDed filters; and the composite is a spelling of a pair, not a curve. #2194's acceptance criterion is what needs correcting, not this code. Worth noting the resulting asymmetry out loud, though, since Copilot raised it from the other side: the detail endpoint renders `other/curve25519+other/curve448` as the value of `curve`, and that is now the one spelling of that field which selects nothing. A client that round-trips a displayed value into a filter gets an empty result, not an error.

`CONTAINS`/`NOT_CONTAINS` are newly advertised here as a side effect of the `LIST → NATIVE_ARRAY` move. On a `FilterFieldType.LIST` field the UI renders a value picker, so `CONTAINS` applies `LIKE '%value%'` to a value the user *selected from a list* — and there is exactly one substring pair among the 197 canonical tokens, `other/Ed448` ⊂ `other/Ed448-Goldilocks`. Low stakes, but it is a widening nobody asked for.

### One process note

The description says this is a baseline for #2170, but **#2170 merged on 2026-09-02** (`a7404454`), ten days before this branch's tip. The live downstream is #2166 (open) and #2165 (open) — which is what I reviewed forward-compatibility against, and what finding 3 is aimed at. If a different PR is actually meant to stack on this, worth naming it, because "is this the right shape to build on" was the main lens here.

src/test/java/com/otilm/core/integration/search/CryptoAssetSearchITest.java
```diff
@@ -188,6 +188,32 @@ void aFilterCombinesWithAnother() {
...
+        UUID hybrid = upsert(new CryptoAssetIdentityFields(CryptographicAssetType.ALGORITHM, "X25519/X448",
+                "1.3.101.110", "ecdh", "key-agree", null, "other/curve25519+other/curve448", null, null, null), null);
+
+        assertThat(search(aPropertyEqualsFilter(FilterField.CBOM_ASSET_CURVE, "other/curve25519")))
+                .containsExactly(hybrid);
+        assertThat(search(aPropertyEqualsFilter(FilterField.CBOM_ASSET_CURVE, "other/curve448")))
+                .containsExactly(hybrid);
+        assertThat(search(aPropertyEqualsFilter(FilterField.CBOM_ASSET_CURVE, "other/curve25519+other/curve448")))
+                .describedAs("the composite matches no member, so it selects nothing")
+                .isEmpty();
```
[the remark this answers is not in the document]
vyskocilm: Confirmed, and the ticket is what changes, not the code. Ruled 2026-09-10: `EQUALS` on this field means membership, never array equality. The composite is a spelling of a pair, not a curve, the value list offers members, and "touches both" is two ANDed member filters.

So #2194's "Exact-composite selection stays available alongside membership" is false. It is @karolhrdina's ticket, so I have not edited the body; the ruling and the requested correction are recorded at https://github.com/OmniTrustILM/core/issues/2194#issuecomment-5660109637.

No code change. This test asserts the contract as ruled.

src/main/java/com/otilm/core/enums/FilterField.java
```diff
@@ -408,7 +408,10 @@ public enum FilterField {
...
+            SearchFieldTypeEnum.NATIVE_ARRAY),
```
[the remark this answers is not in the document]
vyskocilm: Same finding as the thread on `CryptoAssetSearchITest`, same disposition: membership is the ruling (2026-09-10), and #2194's body sentence is what needs correcting. Recorded with the rationale at https://github.com/OmniTrustILM/core/issues/2194#issuecomment-5660109637 and raised with the ticket owner there.

No code change.

src/main/resources/db/migration/V202609101000__crypto_asset_curve_membership.sql
```diff
@@ -0,0 +1,17 @@
...
+
+-- The btree answered equality on the scalar and cannot answer membership in an array. GIN can, but only for a
+-- containment predicate: PostgreSQL has no index path for 'scalar = ANY(column)' at all, which is why the membership
+-- filter is emitted as containment (PostgresFunctionContributor.ARRAY_CONTAINS_PATTERN). The pairing is pinned by
+-- CryptoAssetCurveMembershipMigrationITest, which plans the shipped predicate against this index.
+DROP INDEX "idx_crypto_asset_curve";
```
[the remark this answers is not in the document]
vyskocilm: Fixed in 6d642f93. `DROP INDEX` now precedes `ALTER COLUMN ... TYPE`, with the reason in the migration.

Also in that commit, from @karolhrdina's thread on the same statement: the `USING` clause converts an empty-string curve to NULL rather than to `{}` (`NULLIF("curve", '')`), and the merge-order constraint and the `EMPTY`/`NOT_EMPTY` index regression are written into the comment.

src/main/java/com/otilm/core/util/PostgresFunctionContributor.java
```diff
@@ -12,6 +12,32 @@ public class PostgresFunctionContributor implements FunctionContributor {
     public static final String ARRAY_CONTAINS = "text_array_contains";
     public static final String ARRAY_ITEM_CONTAINS = "text_array_item_contains";
 
+    /**
+     * Scalar membership in a native array column, written as containment rather than as the equivalent
+     * {@code CAST(?1 AS TEXT) = ANY(?2)}.
+     *
+     * <p>
+     * The two answer the same question for the single non-null value this is ever called with -- every caller in
```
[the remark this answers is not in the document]
vyskocilm: Both confirmed and fixed in 6d642f93.

The pattern's javadoc no longer claims equivalence. It now states the one measured divergence: for an array holding a NULL element and not holding the searched value, `= ANY` yields SQL NULL where containment yields false — same outcome under `EQUALS`, and under `NOT_EQUALS` the old form dropped such a row where this one keeps it, which is the correct answer and a change to every `NATIVE_ARRAY` field, not only the curve.

`SearchFieldTypeEnum:48-49` rewritten: it describes containment, and says explicitly that `= ANY` must not come back because it has no index path at all.

src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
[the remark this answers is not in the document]
vyskocilm: Confirmed, and @karolhrdina's thread on the same method names the consequence I had not: the cost is not "one column got slower", it is that the documented budget for the whole endpoint no longer holds, because the panel asks for all eight lists in one request.

No catalogue in this PR — the table is empty until #2073, so there is nothing to maintain a catalogue of and no measurement that could size one. What 6d642f93 does change is the documentation that would otherwise mislead: the javadoc at `:306-310` now says the budget does not hold while curve is in the request, and points at `findDistinctCurve()`; `findDistinctCurve()` points back.

Recorded as open work on #2166, which touches the same value list — an expression index feeding a skip scan, a side table of members, or a cached list are the same work that would let curve rejoin the budget.

vyskocilm: Confirmed — this is the part I had not stated, and the framing is right: the seven skip scans no longer buy the endpoint the property they were written for.

Fixed the documentation in 6d642f93, both directions as you suggested: `findDistinctAlgorithmFamily()`'s javadoc now says the budget no longer holds for the panel while curve is one of the eight, and names `findDistinctCurve()`; `findDistinctCurve()` points back at it for the budget it spends. Recorded as open work on #2166, naming the three candidate fixes and that none is measurable until #2073 puts rows in the table.

src/main/java/com/otilm/core/util/PostgresFunctionContributor.java
vyskocilm: Confirmed — I verified all three against the migrations before changing anything, and you are right on every point:

- `alt_codes TEXT[]` — `V202507221212__custom_oid_entry.sql:7`
- `features TEXT[] NULL` — `V202602131307__connectors_v2.sql:21`
- `"ntp_servers" TEXT[] NOT NULL` — `V202605011200__time_quality_configuration.sql:5`

and no later migration alters any of them. The paragraph was reasoning from `columnDefinition`, which shapes only the entity-generated schema, and then asserting it of production DDL.

Rewritten in 6d642f93 to your conclusion: the cast exists for the entity-generated schema the ITests build (where a `List<String>` under `SqlTypes.ARRAY` really does generate `varchar[]`), and in production all four columns are already `text[]`, so the cast is a no-op and a GIN index on any of the four would be matched today with no type change needed.

Thanks for chasing the DDL — the wrong version of this paragraph would have cost someone either a pointless `ALTER COLUMN ... TYPE TEXT[]` or a declined index that works.

src/main/java/com/otilm/core/util/PostgresFunctionContributor.java
vyskocilm: Confirmed and fixed in 6d642f93. `SearchFieldTypeEnum.NATIVE_ARRAY` now describes membership and the containment predicate, names `ARRAY_CONTAINS_PATTERN`, and says plainly that `= ANY` answers the same question but cannot be answered by any index, so it must not come back.

Your point about ordering stands and is why it needed fixing rather than relying on the test: the enum is where someone adding the next array-backed field looks first, and the migration ITest would only have caught the revert after the reader had already been told the wrong thing.

src/test/java/com/otilm/core/integration/search/CryptoAssetSearchITest.java
vyskocilm: Confirmed — the line passes if the predicate returns nothing, which is exactly the failure the `= ANY` → `@>` move risks. Applied your suggestion in 6d642f93, with the description naming the `IS NULL` arm as what keeps `bare` in the result, so the arm is now non-removable.

src/main/java/com/otilm/core/enums/FilterField.java
vyskocilm: Confirmed, reproduced, and fixed centrally in 6d642f93 — thank you for running all six against both branches, the table is what made the scope decidable.

`FilterPredicatesBuilder.getPropertyFilterPredicate` now refuses any condition outside `SearchHelper.availableConditions(filterField)` before a predicate is built, for every listing rather than only for this field. That is the same set `ListViewServiceImpl` already validates a stored view against, so the two paths now agree.

The decision this forced is worth stating, because it reaches past this PR. Your six are not the only unadvertised operators in use — the gate failed four existing tests on three unrelated resources, all of which **worked**:

| field | operator | test |
|---|---|---|
| `ENTITY_CONNECTOR_NAME` | `CONTAINS` | `EntityInstanceSearchITest` |
| `ENTITY_KIND` | `CONTAINS` | `EntityInstanceSearchITest` |
| `LOCATION_ENTITY_INSTANCE` | `ENDS_WITH` | `LocationsSearchITest` |
| `AUDIT_LOG_RESOURCE_NAME` | `NOT_CONTAINS` | `FilterPredicatesBuilderITest` |

So the real defect is that the advertise list and the predicate builder disagree, and the 500 is only where that disagreement meets a column type that cannot absorb it. Three resolutions were on the table: narrow the gate to `NATIVE_ARRAY` (fixes the 500, breaks nothing, leaves the disagreement); widen the advertise lists to match the builder (loses nothing, but changes the wire contract of three unrelated resources in a PR about curves); or gate centrally and hold those tests to what their fields advertise. **Ruled: gate centrally.** A listing should answer every unadvertised filter the same way, and the alternative leaves a field's advertised set meaning nothing.

The four tests now use advertised operators. `AUDIT_LOG_RESOURCE_NAME`'s json-array not-present branch is still covered — `EMPTY` reaches the same `getNotPresentPredicate(isJsonArray=true)`, so that test was retargeted rather than deleted, and a new test asserts the refusal itself.

Two new tests pin the gate: the refusal on `AUDIT_LOG_RESOURCE_NAME`, and a parameterized one over your exact six operators on `CBOM_ASSET_CURVE`.

Full suite after the change: 8181 tests, 0 failures, 0 errors.

src/main/resources/db/migration/V202609141000__crypto_asset_curve_membership.sql
vyskocilm: Confirmed — you are right that the migration comment and the pattern javadoc both presented this as strictly fixing indexability, and a reader would not have learned that two conditions regressed.

Taken the sentence, not the partial index, and said so explicitly in 6d642f93. The migration comment now states the trade with your measurement: of the six conditions the field advertises, `EQUALS` gains an index path it never had and `EMPTY`/`NOT_EMPTY` lose the one the btree gave them, `NOT_EQUALS`/`NOT_CONTAINS` are negations OR'd with `IS NULL` and were never GIN-answerable, and `CONTAINS` goes through `unnest` + `LIKE` which no index serves. It records that `EQUALS` is the query this inventory exists to answer and `EMPTY` is not, so the trade is deliberate, and names the partial index on `WHERE curve IS NULL` as the remedy if `EMPTY` ever earns one.

No partial index now: the table is empty until #2073, so an index for a condition with no known caller is a write cost bought on speculation.

src/main/resources/db/migration/V202609101000__crypto_asset_curve_membership.sql
vyskocilm: Both confirmed and both fixed in 6d642f93.

**Merge order.** Agreed it belongs in the migration rather than in the ITest javadoc — it is what makes the DDL safe, not what makes the test convenient. The comment now carries it: both statements are free only while `crypto_asset` is empty, which holds because `CryptoAssetWriter` has no production caller; behind #2073 they are outage-class on a populated inventory; and `CONCURRENTLY` is not a remedy because Flyway wraps the script in a transaction and no migration here sets `executeInTransaction=false`. It states that this must land before ingest.

I have also carried the constraint to where the other half of it lives: appended as item 5 to the #2071→#2073 obligations comment (https://github.com/OmniTrustILM/core/issues/2073#issuecomment-5436951889), and recorded on #2194.

**Empty-string conversion.** Verified `string_to_array('', '+')` returns `{}` on the pinned image, and your three consequences follow. Closed with exactly the token you suggested:

```sql
USING CASE WHEN NULLIF("curve", '') IS NULL THEN NULL ELSE string_to_array("curve", '+') END
```

and the comment now says why the `IS NULL` arm earns its place rather than being redundant. `CryptoAssetCurveMembershipMigrationITest` seeds a fourth row with an empty-string curve and asserts it converts to SQL NULL — the point being, as you put it, that converting data this codebase did not write is the statement's whole job.

src/main/java/com/otilm/core/cbom/asset/CompositeCurve.java
vyskocilm: Confirmed, and I took both fixes in 6d642f93 — including the second, on your forward-compat argument for #2166.

**The split moved into Java.** `CompositeCurve` now owns both directions, `CryptoAssetWriter` binds `String[]` members, and the upsert drops `string_to_array`. `AssetNormalizer.joinCurves` joins through `CompositeCurve.join`. Your (1)–(3) are now one definition.

**Remaining sites: two, each pinned by a test that fails when it moves.** `array_to_string(curve, '+')` in `findStaleVerdictRows` (caught by `CryptoAssetPqcSweepITest`, as you measured) and `string_to_array` in the migration (caught by `CryptoAssetCurveMembershipMigrationITest`).

**One correction to what the round-trip vector buys.** I added the `+`-curve vector to `singleProducerVectors` as you suggested, then re-ran your mutation against it — and with the definitions collapsed it does **not** catch a changed separator, because `join(split(x)) == x` now holds for any separator. Re-measured: `CompositeCurveTest` catches it (2 failures), and `NormalizationRulesTest` catches it (4 failures), which is the one that matters — a changed preimage spelling re-keys, and the ratified vectors are what refuse it. So the vector is worth keeping, but for the array's element order, its Hibernate mapping and the writer's split surviving a real round trip, not for the separator. The comment on it says that rather than the stronger claim.

That correction is yours by implication — your mutation table is what prompted me to re-run it instead of assuming the new test closed the gap.

karolhrdina (approved):

PR #2248, opened 2026-09-10
Pre-authorize identifiers listed on an ACME profile

### LINKED ISSUE #2231: Pre-authorize identifiers on ACME profiles
### Description

An ACME profile can list the identifiers its accounts may obtain without proving control. An order whose identifiers all match the list gets authorizations created valid and becomes ready immediately, so http-01 and dns-01 are never issued for them.

- `acme_profile` gains the identifier policy, a list of entries each carrying a value, a match type and a wildcard flag, plus the mode governing identifiers the policy does not cover. Stored as JSONB, since the policy is always read whole with the profile and never queried entry by entry.
- Match types are explicit rather than glob or regex, so the policy stays deterministic and auditable:
  - `EXACT` matches that identifier only.
  - `SUBDOMAIN` matches descendants of that name at any depth, not the name itself; covering both needs an `EXACT` entry alongside.
  - `SUBDOMAIN` applies to dns identifiers; ip identifiers (RFC 8738) are `EXACT`.
  - Matching is case-insensitive for dns and by octets for ip.
- `allowWildcard` sits on the entry, not the profile, so pre-authorizing the wildcard identifier `*.apps.example.com` is a decision per pattern rather than a switch covering every listed name. A wildcard identifier is covered only when the entry allows it and everything the wildcard stands for is already inside that entry's scope.
- Mode governs identifiers the policy does not cover:
  - `PREAUTHORIZED_OR_CHALLENGE`, the default: they proceed through the normal http-01 and dns-01 flow, so a profile with an empty policy behaves exactly as today and the migration needs no data.
  - `PREAUTHORIZED_ONLY`: the order is refused with `rejectedIdentifier`.
- `PREAUTHORIZED_ONLY` with an empty policy is rejected on create and on edit, checked against the merged profile so a partial edit cannot reach it. The message says the mode needs at least one entry, and points at disabling new orders as the way to stop a profile.
- An order mixing covered and uncovered identifiers pre-authorizes the covered ones and treats the rest according to the mode.
- Both the profile-based and the RA-profile-based ACME endpoints.

Worked example, for entries `EXACT server01.example.com`, `SUBDOMAIN apps.example.com` and the same subdomain entry with `allowWildcard`:

| Ordered identifier | EXACT server01 | SUBDOMAIN apps | SUBDOMAIN apps + allowWildcard |
|---|---|---|---|
| `server01.example.com` | covered | no | no |
| `SERVER01.example.com` | covered | no | no |
| `apps.example.com` | no | no | no |
| `web.apps.example.com` | no | covered | covered |
| `db.eu.apps.example.com` | no | covered | covered |
| `other.example.com` | no | no | no |
| `*.apps.example.com` | no | no | covered |

Access control needs nothing new: editing a profile is already gated per profile and audit logged as `PROTOCOLS / ACME_PROFILE / UPDATE`.

### Acceptance Criteria

- [ ] An order for a covered identifier is ready at creation, its authorization valid, and it carries no challenges.
- [ ] `EXACT` matches only that name; `SUBDOMAIN` matches descendants at any depth and not the name itself.
- [ ] A wildcard identifier is pre-authorized only when its entry allows it.
- [ ] ip identifiers match by octets regardless of rendering.
- [ ] Under `PREAUTHORIZED_OR_CHALLENGE` an uncovered identifier gets the normal challenges; under `PREAUTHORIZED_ONLY` the order is refused.
- [ ] An order mixing covered and uncovered identifiers applies the mode to the uncovered ones only.
- [ ] `PREAUTHORIZED_ONLY` with an empty policy is refused on create and on partial edit.
- [ ] A profile with an empty policy issues http-01 and dns-01 exactly as today.
- [ ] Sonar quality gate passed, >= 80% coverage on new code.

klaraf755:
Depends-On: OmniTrustILM/interfaces#967

## Problem

Every ACME order goes through http-01 or dns-01. A fleet whose devices have neither DNS write access nor an HTTP listener cannot prove control of anything, so it cannot use the platform's ACME endpoint at all.

## Change

An ACME profile lists identifiers its accounts may obtain without proving control, and a mode saying what happens to everything else. An order whose identifiers are all covered has its authorizations created valid, carries no challenges, and is `ready` at creation.

- `acme_profile` gains `preauthorized_identifiers` (JSONB — the policy is always read whole with the profile, never queried entry by entry) and `identifier_authorization_mode`.
- `AcmeIdentifierPolicy` is the matcher, a pure function over the policy and one identifier: `EXACT` covers that name only; `SUBDOMAIN` covers descendants at any depth and not the name itself, so covering both takes two entries. DNS comparison is case-insensitive and tolerates a trailing root dot; IP comparison is by parsed octets, and only literals are accepted — a name would otherwise be resolved, letting DNS decide what a policy covers.
- A wildcard identifier is covered only when its entry allows it *and* everything the wildcard stands for is already inside that entry's scope, which an `EXACT` entry can never satisfy.
- `PREAUTHORIZED_OR_CHALLENGE`, the default, sends an uncovered identifier through the usual flow, so a profile with an empty policy behaves exactly as today and the migration needs no data. `PREAUTHORIZED_ONLY` refuses the order with `rejectedIdentifier`.
- The whole order is decided before the first row is written, so a refused order leaves nothing behind. An order mixing covered and uncovered identifiers pre-authorizes the covered ones and applies the mode to the rest.
- An entry that could never match is refused on create and on edit: a value that is not a well-formed name or address literal, a wildcard written where a name belongs, an address asked to match by subdomain. Such an entry would otherwise sit in the policy looking like cover the operator does not have. `PREAUTHORIZED_ONLY` with an empty policy is refused likewise, so the two rules together mean the mode always has something usable behind it. Both are checked against the merged profile so a partial edit cannot reach either. Omitting either property on edit keeps what the profile has; an empty array clears it.
- The RFC 8555 §7.1.6 readiness rule moves out of `applyValidationResult` into `AcmeChallengeStateMachine.readyWhenEveryAuthorizationIsValid`, so challenge validation and order creation ask the same question rather than each deciding it. It gained an emptiness guard, since `allMatch` over no authorizations is vacuously true.

Two behaviour changes that are not about pre-authorization and apply to every order:

- **An identifier the platform does not issue for is now refused.** Today any type is accepted and given an http-01 and a dns-01 challenge, so an `email` or `TNAuthList` identifier gets challenges it can never satisfy and the order sits pending until it expires. Such an order is now `unsupportedIdentifier` at creation. `dns` and `ip` are the supported set, enumerated once in `AcmeIdentifierType` so what a policy entry may declare and what an order may ask for cannot drift apart.
- An order naming no identifiers threw a NullPointerException out of `generateOrder` and returned 500. It is now `malformed`.

Known gap: `validateCsr` compares identifier values against stringified SAN values and is type-aware only for `dns`. Pre-authorization lets an `ip` order reach finalize without first failing a challenge, so it makes an existing weakness easier to hit. Not addressed here.

Access control needs nothing new — editing a profile is already gated per profile and audit logged as `PROTOCOLS / ACME_PROFILE / UPDATE`.

## Tests

`AcmeIdentifierPolicyTest` asserts the worked example from the issue as a table, so the documented behaviour and the matcher cannot drift, plus the label-boundary, trailing-dot, wildcard-scope, IP-octet and never-resolve-a-name cases. `AcmePreauthorizedOrderITest` covers the wiring end to end: valid-with-no-challenge, unchanged challenges for an uncovered identifier, the mixed order, a profile with no policy, and what `PREAUTHORIZED_ONLY` refuses without writing rows.

Closes #2231

### DIFF
--- a/src/main/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicy.java
+++ b/src/main/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicy.java
@@ -39,47 +39,51 @@ public final class AcmeIdentifierPolicy {
     private static final int MAX_NAME_LENGTH = 253;
     private static final int IPV6_BYTES = 16;
 
+    /** The longest an address literal can be, so an oversized value is refused before it is parsed. */
+    private static final int MAX_ADDRESS_LENGTH = 45;
+
     private AcmeIdentifierPolicy() {
     }
 
-    /** Whether any entry of the policy covers this identifier. */
+    /**
+     * Whether any entry of the policy covers this identifier. The ordered value is parsed once rather than once per
+     * entry: it is caller-supplied and a policy may hold many entries.
+     */
     public static boolean covers(List<AcmePreauthorizedIdentifierDto> policy, Identifier identifier) {
         if (policy == null || policy.isEmpty() || !isSupportedType(identifier) || identifier.getValue() == null) {
             return false;
         }
-        return policy.stream().anyMatch(entry -> coveredBy(entry, identifier));
+        if (isIp(identifier)) {
+            Optional<byte[]> ordered = addressBytes(identifier.getValue());
+            return ordered.isPresent() && policy.stream().anyMatch(entry -> coversAddress(entry, ordered.get()));
+        }
+        String ordered = normalizeName(identifier.getValue());
+        return ordered != null && policy.stream().anyMatch(entry -> coversName(entry, ordered));
     }
 
-    private static boolean coveredBy(AcmePreauthorizedIdentifierDto entry, Identifier identifier) {
-        if (entry == null || entry.getValue() == null || entry.getMatchType() == null) {
-            return false;
-        }
-        return isIp(identifier)
-                ? coversAddress(entry, identifier.getValue())
-                : coversName(entry, identifier.getValue());
+    private static boolean usable(AcmePreauthorizedIdentifierDto entry) {
+        return entry != null && entry.getValue() != null && entry.getMatchType() != null;
     }
 
     /**
      * RFC 8738 addresses have no hierarchy to descend, so only an exact entry can cover one, and both sides must parse
      * as literals.
      */
-    private static boolean coversAddress(AcmePreauthorizedIdentifierDto entry, String orderedValue) {
-        if (entry.getMatchType() != AcmeIdentifierMatchType.EXACT) {
+    private static boolean coversAddress(AcmePreauthorizedIdentifierDto entry, byte[] ordered) {
+        if (!usable(entry) || entry.getMatchType() != AcmeIdentifierMatchType.EXACT) {
             return false;
         }
-        Optional<byte[]> ordered = addressBytes(orderedValue);
         Optional<byte[]> pattern = addressBytes(entry.getValue());
-        return ordered.isPresent() && pattern.isPresent() && Arrays.equals(ordered.get(), pattern.get());
+        return pattern.isPresent() && Arrays.equals(ordered, pattern.get());
     }
 
-    private static boolean coversName(AcmePreauthorizedIdentifierDto entry, String orderedValue) {
+    private static boolean coversName(AcmePreauthorizedIdentifierDto entry, String ordered) {
+        if (!usable(entry)) {
+            return false;
+        }
         String pattern = normalizeName(entry.getValue());
         if (pattern == null || pattern.startsWith(WILDCARD_PREFIX)) {
             // An entry is a name, never a pattern: the match type is what widens it.
-            return false;
-        }
-        String ordered = normalizeName(orderedValue);
-        if (ordered == null) {
             return false;
         }
         if (ordered.startsWith(WILDCARD_PREFIX)) {
@@ -150,6 +154,9 @@ public final class AcmeIdentifierPolicy {
 
     /** The bytes of an IP literal, or empty when the value is not one. */
     private static Optional<byte[]> addressBytes(String value) {
+        if (value.length() > MAX_ADDRESS_LENGTH) {
+            return Optional.empty();
+        }
         return value.indexOf(':') >= 0 ? ipv6Bytes(value) : ipv4Bytes(value);
     }
 
@@ -184,8 +191,14 @@ public final class AcmeIdentifierPolicy {
             return Optional.empty();
         }
         int compression = value.indexOf("::");
-        Optional<byte[]> head = hextets(compression < 0 ? value : value.substring(0, compression));
-        Optional<byte[]> tail = hextets(compression < 0 ? "" : value.substring(compression + 2));
+        // RFC 4291 section 2.2 form 3 puts the dotted quad at the end of the whole address, so it belongs to
+        // whichever segment ends it: the tail when the address is compressed, otherwise the only segment there is.
+        Optional<byte[]> head = compression < 0
+                ? hextets(value, true)
+                : hextets(value.substring(0, compression), false);
+        Optional<byte[]> tail = compression < 0
+                ? Optional.of(new byte[0])
+                : hextets(value.substring(compression + 2), true);
         if (head.isEmpty() || tail.isEmpty()) {
             return Optional.empty();
         }
@@ -204,8 +217,8 @@ public final class AcmeIdentifierPolicy {
         return Optional.of(address);
     }
 
-    /** One colon-separated run of hextets, optionally ending in a dotted quad. Empty text is an empty run. */
-    private static Optional<byte[]> hextets(String text) {
+    /** One colon-separated run of hextets, ending in a dotted quad only where the address allows one. */
+    private static Optional<byte[]> hextets(String text, boolean quadAllowed) {
         if (text.isEmpty()) {
             return Optional.of(new byte[0]);
         }
@@ -214,7 +227,7 @@ public final class AcmeIdentifierPolicy {
         for (int i = 0; i < parts.length; i++) {
             String part = parts[i];
             if (part.indexOf('.') >= 0) {
-                if (i != parts.length - 1) {
+                if (!quadAllowed || i != parts.length - 1) {
                     return Optional.empty();
                 }
                 Optional<byte[]> quad = ipv4Bytes(part);
--- a/src/main/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicy.java
+++ b/src/main/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicy.java
@@ -8,10 +8,16 @@ import java.net.UnknownHostException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
+import java.util.regex.Pattern;
 
 /**
  * Whether an ACME profile's pre-authorization policy covers an ordered identifier, so the authorization for it can be
- * created valid and carry no challenge. Pure: the policy and the identifier in, a decision out.
+ * created valid and carry no challenge. Pure: the policy and the identifier in, a decision out, no I/O of any kind.
+ *
+ * <p>
+ * Everything here fails closed. A value this class cannot parse with certainty — a malformed name, an address that is
+ * not a literal, an identifier type it does not know — is not covered, because the alternative is issuing a certificate
+ * for a name nobody proved control of.
  */
 public final class AcmeIdentifierPolicy {
 
@@ -19,12 +25,23 @@ public final class AcmeIdentifierPolicy {
     private static final String IP = "ip";
     private static final String WILDCARD_PREFIX = "*.";
 
+    /** A DNS label: letters, digits and inner hyphens, per the preferred name syntax of RFC 1035 section 2.3.1. */
+    private static final Pattern LABEL = Pattern.compile("[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?");
+
+    private static final Pattern IPV4 = Pattern.compile("(\\d{1,3})\\.(\\d{1,3})\\.(\\d{1,3})\\.(\\d{1,3})");
+
+    /** The character set of an IPv6 literal, including a zone identifier. No hostname can match it. */
+    private static final Pattern IPV6_SHAPED = Pattern
+            .compile("[0-9A-Fa-f:.%\\p{Alnum}_-]*:[0-9A-Fa-f:.%\\p{Alnum}_-]*");
+
+    private static final int MAX_NAME_LENGTH = 253;
+
     private AcmeIdentifierPolicy() {
     }
 
     /** Whether any entry of the policy covers this identifier. */
     public static boolean covers(List<AcmePreauthorizedIdentifierDto> policy, Identifier identifier) {
-        if (policy == null || policy.isEmpty() || identifier == null || identifier.getValue() == null) {
+        if (policy == null || policy.isEmpty() || !isSupportedType(identifier) || identifier.getValue() == null) {
             return false;
         }
         return policy.stream().anyMatch(entry -> coveredBy(entry, identifier));
@@ -34,14 +51,34 @@ public final class AcmeIdentifierPolicy {
         if (entry == null || entry.getValue() == null || entry.getMatchType() == null) {
             return false;
         }
-        if (isIp(identifier)) {
-            // RFC 8738 addresses have no hierarchy to descend, so only an exact entry can cover one, and the
-            // comparison is on the address rather than on how it was written.
-            return entry.getMatchType() == AcmeIdentifierMatchType.EXACT
-                    && sameAddress(entry.getValue(), identifier.getValue());
+        return isIp(identifier)
+                ? coversAddress(entry, identifier.getValue())
+                : coversName(entry, identifier.getValue());
+    }
+
+    /**
+     * RFC 8738 addresses have no hierarchy to descend, so only an exact entry can cover one. Both sides must parse as
+     * literals: a name is never resolved, since that would let whoever controls DNS decide what a policy covers.
+     */
+    private static boolean coversAddress(AcmePreauthorizedIdentifierDto entry, String orderedValue) {
+        if (entry.getMatchType() != AcmeIdentifierMatchType.EXACT) {
+            return false;
         }
-        String ordered = normalize(identifier.getValue());
-        String pattern = normalize(entry.getValue());
+        byte[] ordered = addressBytes(orderedValue);
+        byte[] pattern = addressBytes(entry.getValue());
+        return ordered != null && pattern != null && Arrays.equals(ordered, pattern);
+    }
+
+    private static boolean coversName(AcmePreauthorizedIdentifierDto entry, String orderedValue) {
+        String pattern = normalizeName(entry.getValue());
+        if (pattern == null || pattern.startsWith(WILDCARD_PREFIX)) {
+            // An entry is a name, never a pattern: the match type is what widens it.
+            return false;
+        }
+        String ordered = normalizeName(orderedValue);
+        if (ordered == null) {
+            return false;
+        }
         if (ordered.startsWith(WILDCARD_PREFIX)) {
             return coversWildcard(entry, ordered.substring(WILDCARD_PREFIX.length()), pattern);
         }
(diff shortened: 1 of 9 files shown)


src/main/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicy.java
```diff
@@ -0,0 +1,120 @@
...
+        return left != null && Arrays.equals(left, addressBytes(ordered));
+    }
+
+    /**
+     * Parsed rather than compared as text, so the same address written two ways matches. Only literals are accepted: a
+     * name would otherwise be resolved, letting DNS decide what a policy covers.
+     */
+    private static byte[] addressBytes(String value) {
+        String literal = value.trim();
+        if (!InetAddresses.isLiteral(literal)) {
```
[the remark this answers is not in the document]
klaraf755: Fixed, and this was the serious one. `Character.digit('e', 16)` is 14, so `example.com` passed the first-character test and `InetAddress.getByName` then performed a real lookup, exactly as described.

The matcher now parses literals itself and never calls the resolver on anything that could be a hostname: IPv4 is parsed group by group with a leading-zero rejection (octal to some parsers, decimal to others), and IPv6 is recognised by a character class no hostname can match before parsing. Both sides of an address comparison must parse as literals, so neither a name entry against an address identifier nor an address entry against a name can match.

The test that was supposed to prevent this was itself the reason it survived: it used `localhost`, whose leading `l` is one of the few characters that short-circuited the old check. It is now a parameterized case over five name/address pairs that all begin with characters a first-character test waves through.

src/main/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicy.java
```diff
@@ -0,0 +1,120 @@
...
+    private static final String DNS = "dns";
+    private static final String IP = "ip";
+    private static final String WILDCARD_PREFIX = "*.";
+
+    private AcmeIdentifierPolicy() {
+    }
+
+    /** Whether any entry of the policy covers this identifier. */
+    public static boolean covers(List<AcmePreauthorizedIdentifierDto> policy, Identifier identifier) {
+        if (policy == null || policy.isEmpty() || identifier == null || identifier.getValue() == null) {
```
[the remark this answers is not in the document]
klaraf755: Fixed. `covers` now returns false for any identifier whose type is neither `dns` nor `ip`, so `isSupportedType` is on the production path rather than only in tests.

The order path rejects such an identifier outright as well, with `unsupportedIdentifier`: an authorization for a type the platform cannot validate would otherwise sit pending until the order expired, and a null-typed identifier could reach `ready` and then hit the pre-existing NPE in `validateCSR`. `AcmePreauthorizedOrderITest.anIdentifierTypeTheServerDoesNotIssueForIsRefused` covers it.

src/main/java/com/otilm/core/service/impl/AcmeProfileServiceImpl.java
```diff
@@ -296,6 +312,20 @@ public AcmeProfileDto editAcmeProfile(SecuredUUID uuid, AcmeProfileEditRequestDt
...
+     * the profile could receive. Checked on the merged profile, since an edit can supply the mode alone or empty the
+     * policy alone and neither request sees the other half.
+     */
+    private static void requirePolicyWhenPreauthorizedOnly(AcmeProfile acmeProfile) {
+        if (acmeProfile.effectiveIdentifierAuthorizationMode() == AcmeIdentifierAuthorizationMode.PREAUTHORIZED_ONLY
+                && acmeProfile.preauthorizedIdentifierList().isEmpty()) {
+            throw new ValidationException(ValidationError
+                    .create("preauthorizedOnly needs at least one pre-authorized identifier; to stop the profile "
+                            + "accepting orders, disable new orders instead"));
+        }
```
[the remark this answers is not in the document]
klaraf755: Fixed. `AcmeProfileServiceITest` gains five cases exercising the check through the service API rather than the entity: `preauthorizedOnlyWithAnEmptyPolicyIsRefusedOnCreate`, `preauthorizedOnlyWithEntriesIsAcceptedOnCreate`, `aModeOnlyEditCannotStrandAProfileWithoutEntries`, `emptyingThePolicyOfAPreauthorizedOnlyProfileIsRefused`, and `anEditOmittingThePolicyKeepsItAndAnEmptyListClearsIt`.

The mode-only and empty-list edits are the two you name, and both are the reason the check runs against the merged profile rather than the request — neither request carries the other half of the state that makes the combination invalid.

lubomirw (commented):
Automated multi-reviewer triage ? 7 selected findings (COMMENT).

src/main/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicy.java
```diff
@@ -0,0 +1,260 @@
...
+
+    /**
+     * Whether any entry of the policy covers this identifier. The ordered value is parsed once rather than once per
+     * entry: it is caller-supplied and a policy may hold many entries.
+     */
+    public static boolean covers(List<AcmePreauthorizedIdentifierDto> policy, Identifier identifier) {
+        if (policy == null || policy.isEmpty() || !isSupportedType(identifier) || identifier.getValue() == null) {
+            return false;
+        }
+        if (isIp(identifier)) {
```
lubomirw, line 56: **[Medium]** | Correctness | Confidence: Medium

This path can make an `ip` order ready at creation, but CSR validation compares order identifier values to stringified SAN values without type-aware IP handling. That can cause a correct IPv4 or IPv6 CSR to fail with `badCSR` during finalization.

_Suggested:_ Compare CSR SANs by identifier type, using canonical IP octets for `iPAddress`, and add end-to-end finalize tests for IPv4 and IPv6 pre-authorized orders.

_Unverified:_ an end-to-end CSR against this exact Bouncy Castle version would settle the final SAN rendering detail.

<sub>flagged by Superpowers; verified against the new IP pre-authorization path and existing CSR validation structure.</sub>

src/main/java/com/otilm/core/service/impl/AcmeProfileServiceImpl.java
```diff
@@ -296,6 +312,20 @@ public AcmeProfileDto editAcmeProfile(SecuredUUID uuid, AcmeProfileEditRequestDt
...
     }
 
+    /**
+     * PREAUTHORIZED_ONLY refuses every identifier the policy does not cover, so with no entries it refuses every order
+     * the profile could receive. Checked on the merged profile, since an edit can supply the mode alone or empty the
+     * policy alone and neither request sees the other half.
+     */
+    private static void requirePolicyWhenPreauthorizedOnly(AcmeProfile acmeProfile) {
+        if (acmeProfile.effectiveIdentifierAuthorizationMode() == AcmeIdentifierAuthorizationMode.PREAUTHORIZED_ONLY
+                && acmeProfile.preauthorizedIdentifierList().isEmpty()) {
```
lubomirw, line 322: **[Medium]** | Correctness | Confidence: High

`PREAUTHORIZED_ONLY` is guarded only by a non-empty policy list. Entries the matcher can never use, such as malformed literals, wildcard-prefixed values, or incompatible match types, can satisfy the invariant and then reject every order.

_Suggested:_ Validate policy-entry syntax and match-type compatibility on create/edit, and require at least one semantically usable entry for `PREAUTHORIZED_ONLY`.

<sub>flagged by Copilot CLI and Superpowers; verified against the merged-profile guard and fail-closed matcher behavior.</sub>

src/test/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicyTest.java
```diff
@@ -0,0 +1,323 @@
...
+
+    private static final AcmePreauthorizedIdentifierDto EXACT_SERVER = entry("server01.example.com",
+            AcmeIdentifierMatchType.EXACT, false);
+    private static final AcmePreauthorizedIdentifierDto SUBDOMAIN_APPS = entry("apps.example.com",
+            AcmeIdentifierMatchType.SUBDOMAIN, false);
+    private static final AcmePreauthorizedIdentifierDto SUBDOMAIN_APPS_WILDCARD = entry("apps.example.com",
+            AcmeIdentifierMatchType.SUBDOMAIN, true);
+
+    /**
+     * The table from the issue, asserted as written so the documented behaviour and the code cannot drift.
```
lubomirw, line 25: **[Low]** | Maintainability | Confidence: High

The test Javadoc refers to `the issue` without identifying a durable source readers can follow from the code.

_Suggested:_ Remove the issue reference or name the durable behavior being pinned instead.

<sub>flagged by Hygiene; verified against the added test Javadoc.</sub>

src/test/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicyTest.java
```diff
@@ -0,0 +1,323 @@
...
+                AcmeIdentifierPolicy
+                        .covers(List.of(entry("192.0.2.1", AcmeIdentifierMatchType.SUBDOMAIN, false)), ip("192.0.2.1")),
+                "an address has no hierarchy to descend");
+    }
+
+    /**
+     * Resolving either side would let whoever controls DNS decide what a policy covers: an entry naming a host would
+     * pre-authorize whatever address it resolves to, and an address entry would pre-authorize every name pointing at
+     * it. The values here start with characters a first-character test would wave through, which is how this went
+     * unnoticed once.
```
lubomirw, line 105: **[Low]** | Maintainability | Confidence: High

The Javadoc records how the defect previously escaped detection instead of only explaining the current invariant: `which is how this went unnoticed once`.

_Suggested:_ Reword to: `The values start with characters that a first-character check would admit.`

<sub>flagged by Hygiene; verified against the added regression-test Javadoc.</sub>

src/test/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicyTest.java
```diff
@@ -0,0 +1,323 @@
...
+            "0:1.2.3.4::",
+            "1:2:3:4:5:1.2.3.4::",
+            "1.2.3.4:5::6"})
+    void aMalformedIpv6ValueIsNotAnAddress(String value) {
+        assertFalse(
+                AcmeIdentifierPolicy.covers(List.of(entry(value, AcmeIdentifierMatchType.EXACT, false)), ip(value)));
+    }
+
+    /**
+     * The dotted quad is the address-final 32 bits. Accepting it earlier gave every entry a second spelling: the head
```
lubomirw, line 183: **[Low]** | Maintainability | Confidence: High

The Javadoc describes prior parser behavior rather than only stating the lasting aliasing constraint.

_Suggested:_ Reword to: `Allowing a dotted quad before the elision gives an entry an alias because the rewritten head parses to the same bytes.`

<sub>flagged by Hygiene; verified against the added IPv6 alias regression test.</sub>

src/main/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicy.java
```diff
@@ -0,0 +1,260 @@
...
+    private static boolean isAscii(String value) {
+        for (int i = 0; i < value.length(); i++) {
+            if (value.charAt(i) > 0x7F) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    /** The bytes of an IP literal, or empty when the value is not one. */
```
lubomirw, line 155: **[Nit]** | Maintainability | Confidence: High

The private-method comment mostly restates the method name and `Optional` return behavior.

_Suggested:_ Remove `/** The bytes of an IP literal, or empty when the value is not one. */` if you want stricter hygiene on private helpers.

<sub>flagged by Hygiene; verified against the private helper.</sub>

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
```diff
@@ -140,6 +144,12 @@ public class AcmeServiceImpl implements AcmeExternalService {
 
     private static final Logger logger = LoggerFactory.getLogger(AcmeServiceImpl.class);
 
+    /** How many refused identifiers are named before the rest are counted instead. */
```
lubomirw, line 147: **[Nit]** | Maintainability | Confidence: High

The constant comment duplicates the nearby `summarize` helper's explanation of how refused identifiers are listed and counted.

_Suggested:_ Remove the constant-level comment if you want that explanation to live only with `summarize`.

<sub>flagged by Hygiene; verified against the constant and helper documentation.</sub>

src/main/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicy.java
klaraf755: Fixed in fe9e41969, and the "Unverified" note is now settled: it is a real defect, and a total one.

Against the pinned Bouncy Castle 1.84, `IETFUtils.valueToString` on an `iPAddress` subjectAltName returns the hex of its DER octet string, not text — `192.0.2.1` becomes `#0404c0000201`, `2001:db8::1` becomes `#041020010db8000000000000000000000001`. An ordered value can never equal that, so `sans.containsAll(identifiers)` failed for every `ip` order. This is independent of pre-authorization; it is true on main today. Pre-authorization only makes it easy to reach, since such an order arrives `ready` with no challenge to fail first.

The extraction was where the defect lived, not the comparison, so both moved into `AcmeCsrIdentifiers`: a CSR is read once and each type held as the thing it is — DNS names as text, addresses as octets. `validateCSR` is now seven lines. Addresses compare by octets through the same `AcmeIdentifierPolicy.addressBytes` parser pre-authorization uses, so neither side's spelling decides the match, and an address must be a subjectAltName (RFC 8738 section 3) rather than a common name.

It also closed an adjacent gap in the same place: `AcmeIdentifierPolicy` folds DNS case and a trailing root dot but `validateCSR` did not, so a pre-authorized `SERVER01.example.com` order would have reached finalize and failed `badCSR`.

Tests are `AcmeCsrIdentifiersTest` — six cases on real signed CSRs, no mocks and no Spring context, covering IPv4, IPv6 in three spellings, the address-versus-name-that-reads-the-same case, DNS case and root-dot folding, and unknown types failing closed.

src/test/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicyTest.java
klaraf755: Fixed in aa512ea6d. The Javadoc no longer refers to "the issue"; it now names what the table pins — that the behaviour a profile's editor sees and the matcher cannot drift apart.

src/test/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicyTest.java
klaraf755: Fixed in aa512ea6d, taking the suggested wording: "The values start with characters that a first-character check would admit."

src/test/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicyTest.java
klaraf755: Fixed in aa512ea6d, taking the suggested wording: "Allowing one before the elision gives an entry an alias, because the rewritten head parses to the same bytes."

src/main/java/com/otilm/core/service/acme/impl/AcmeServiceImpl.java
klaraf755: Fixed in bca7de01f. The constant-level comment is removed; the explanation of how refused identifiers are listed and counted lives only with `summarize`.

src/main/java/com/otilm/core/service/acme/identifier/AcmeIdentifierPolicy.java
klaraf755: Partly. The comment you quoted — `/** The bytes of an IP literal, or empty when the value is not one. */` — is gone in fe9e41969.

The method is not left bare, though: it became public in the same commit so that CSR validation at finalize reads an address through the same parser, and it now carries a Javadoc saying why it is shared and that it never reaches a resolver. That is a constraint rather than a restatement of the name, so I kept it.

src/main/java/com/otilm/core/service/impl/AcmeProfileServiceImpl.java
klaraf755: Fixed in a2ce6f650, taking the stronger of the two options you offered: rather than requiring at least one semantically usable entry, an entry that could never match is now refused outright on create and on edit. A non-empty policy is therefore all-usable by construction, and "at least one usable entry for `preauthorizedOnly`" follows from the existing non-empty guard.

`AcmeIdentifierPolicy.isUsable` is the single decision, so the rule that accepts an entry and the matcher that acts on it are the same code. It refuses a value that is not a well-formed name or address literal, a wildcard written where a name belongs (an entry is a name; the match type is what widens it), an address asked to match by subdomain, and an address given a wildcard allowance. The message names the offending entries back as the operator wrote them.

Both rules are checked against the merged profile, so neither a mode-only nor an entries-only edit can slip past.

On placement: the entry syntax is checked in core rather than on the interfaces DTO, because deciding it needs the DNS-label and IP-literal parsers, and putting a second copy of those in interfaces is the drift this PR removed elsewhere by making `AcmeIdentifierType` the single enumeration of types. What does live on the DTO is the part that needs no parser — an `ip` entry may declare neither `subdomain` nor a wildcard allowance.

Tests: `anEntryThatCouldNeverMatchIsNotUsable` covers the nine shapes as a table, plus `anEntryThatCanMatchIsUsable` and `anIncompleteEntryIsNotUsable`; `AcmeProfileServiceITest` pins the refusal through the service on both create and edit, that a refused edit stores nothing, and that a policy of only unusable entries can no longer satisfy `preauthorizedOnly`.

lubomirw (dismissed):

lubomirw (approved):

PR #2252, opened 2026-09-11
Offer only commentable resources as a comment condition's host

### LINKED ISSUE #2229: Workflows - Conditions - Comment: Host Resource offers every platform resource instead of the 22 commentable ones
### Description
When building a condition on the **Comment** resource in **Workflows → Conditions**, the **Host Resource** field offers every value of the platform's `Resource` enum — 67 options — instead of only the resources that can host a comment. Comments are supported on 22 resources, so 45 of the offered values can never match anything. An operator can build a rule on one of them, save it without any warning, and never find out why the rule does not fire.

The Epic names the source of truth for this set explicitly: *"This set is the single source of truth: core validates requests against it and registers the `COMMENT` action for exactly these resources at startup."* Request validation and permission registration both use `Resource.commentableResources`; this condition field does not — its value list is populated from the whole `Resource` enum.

Some of the offered values were excluded from commenting by an explicit decision on the Epic:

- `certificateRequests` — Decision 1: *"`CERTIFICATE_REQUEST` is excluded from round one"*
- `endEntityProfiles` — listed under **Out of Scope**
- `comments` — a comment as the host of a comment; the Epic states that `Resource.COMMENT` exists *"for event/audit identity only"*
- `NONE` and `ANY` — pseudo-values with no object behind them

The remainder are platform internals and non-commentable resources, among them `dashboard`, `settings`, `auditLogs`, `credentials`, `users`, `roles`, `groups`, `attributes`, `jobs`, `proxies`, `oids`, `rules`, `actions`, `triggers`, `conditions`, `executions`, `listViews`, `searchFilters`, `resourceEvents`, `notificationInstances`, `keyItems`, `cboms`, `cryptoAssets`, `acmeAccounts`, `acmeOrders`, `acmeChallenges`, `cmpTransactions`, `complianceRules`, `customAttributes`, `globalMetadata`, `trustedCertificates`, `signingRecords` and `tspProfileBasicCredentials`.

### Steps to Reproduce
1. Go to **Workflows → Conditions** and create a new condition.
2. Select **Comment** as the resource.
3. Open the **Host Resource** field and inspect the values offered.
4. Compare the list with the 22 commentable resources defined for the feature.

### Expected Behavior
- Only the 22 resources in `Resource.commentableResources` are offered.
- The resources excluded by the Epic (`certificateRequests`, `endEntityProfiles`), `comments` itself, and the pseudo-values `NONE` and `ANY` are not offered.

### Actual Behavior
- 67 values are offered — the full `Resource` enum.
- 45 of them can never match a comment, so a rule built on any of them saves successfully and then never fires.

### Severity
Minor

A rule built on an impossible value saves without warning and never fires, which costs operator time to diagnose. No impact on data or access.

### Module
Workflows

### Attachments

_No response_

### Logs

_No response_

### Environment
Commit: 7a414fc
Branch: main
Client:
- macOS Tahoe 26.5.1

klaraf755:
Closes #2229

The Host Resource field of a comment condition offered every constant of the `Resource` enum, 67 values, though only 22 resources can host a comment (Epic: OmniTrustILM/ilm#246). A rule built on one of the other 45 saved without a warning and never fired.

## What's here

- `FilterField` — a list field backed by an enum can declare the subset of its constants it offers, through one more constructor and `getEnumValues()`, which falls back to every constant. The field listing in `ResourceServiceImpl` asks the field for its values instead of reading the enum class; predicate building and trigger evaluation still go through the enum class and are untouched.
- `COMMENT_HOST_RESOURCE` declares `Resource.getCommentableResources()`, the same set request validation and the startup action registration use.
- While at it, the same bug class on the other `Resource`-backed fields: `AUDIT_LOG_AFFILIATED_RESOURCE` and `APPROVAL_RESOURCE` no longer offer `NONE` or `ANY`, and `AUDIT_LOG_RESOURCE` no longer offers `ANY`. `NONE` stays on the audit resource field because module-level operations are audited with it. The set is the complement of the two wildcards, computed rather than listed, so it cannot drift.
- Tests — the resource service ITest pins the host resource values to exactly the commentable set, and the three resource fields to the presence and absence of the wildcards.

## Notes for reviewers

- Not done here, on purpose: the per-check certificate validation fields offer all eight `CertificateValidationStatus` values while the validator can only produce four for OCSP and CRL and three for the signature check, and `APPROVAL_ACTION` offers every action while approvals arise for a handful. Both subsets live in code that is not `FilterField`, so declaring them here would duplicate knowledge; they want a single source of truth first and are a separate change.

### DIFF
--- a/src/main/java/com/otilm/core/enums/FilterField.java
+++ b/src/main/java/com/otilm/core/enums/FilterField.java
@@ -84,7 +84,9 @@ import com.otilm.core.model.cbom.CryptoAssetIdentityGuard;
 import jakarta.persistence.metamodel.Attribute;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.EnumSet;
 import java.util.List;
+import java.util.Set;
 import lombok.Getter;
 
 @Getter
@@ -283,10 +285,11 @@ public enum FilterField {
     AUDIT_LOG_ACTOR_NAME(Resource.AUDIT_LOG, null, null, AuditLog_.actorName, "Actor name", SearchFieldTypeEnum.STRING),
     AUDIT_LOG_ACTOR_AUTH_METHOD(Resource.AUDIT_LOG, null, null, AuditLog_.actorAuthMethod, "Actor Auth method",
             SearchFieldTypeEnum.LIST, AuthMethod.class),
+    // NONE is a recorded resource for module-level operations; an absent affiliated resource is stored as no value
     AUDIT_LOG_RESOURCE(Resource.AUDIT_LOG, null, null, AuditLog_.resource, "Resource", SearchFieldTypeEnum.LIST,
-            Resource.class),
+            Resource.class, resourcesExcept(Resource.ANY)),
     AUDIT_LOG_AFFILIATED_RESOURCE(Resource.AUDIT_LOG, null, null, AuditLog_.affiliatedResource, "Affiliated resource",
-            SearchFieldTypeEnum.LIST, Resource.class),
+            SearchFieldTypeEnum.LIST, Resource.class, resourcesExcept(Resource.NONE, Resource.ANY)),
     AUDIT_LOG_OPERATION(Resource.AUDIT_LOG, null, null, AuditLog_.operation, "Operation", SearchFieldTypeEnum.LIST,
             Operation.class),
     AUDIT_LOG_OPERATION_RESULT(Resource.AUDIT_LOG, null, null, AuditLog_.operationResult, "Operation result",
@@ -317,7 +320,7 @@ public enum FilterField {
 
     // Approval
     APPROVAL_RESOURCE(Resource.APPROVAL, null, null, Approval_.resource, "Resource", SearchFieldTypeEnum.LIST,
-            Resource.class),
+            Resource.class, resourcesExcept(Resource.NONE, Resource.ANY)),
     APPROVAL_ACTION(Resource.APPROVAL, null, null, Approval_.action, "Action", SearchFieldTypeEnum.LIST,
             ResourceAction.class),
     APPROVAL_STATUS(Resource.APPROVAL, null, null, Approval_.status, Constants.STATUS, SearchFieldTypeEnum.LIST,
@@ -556,6 +559,11 @@ public enum FilterField {
     /** The values a list field backed by an enum offers: the subset it declares, or every constant otherwise. */
     public IPlatformEnum[] getEnumValues() {
         return enumValues != null ? enumValues : enumClass.getEnumConstants();
+    }
+
+    /** Every resource but the given wildcards, which scope a grant and are never the resource of a stored record. */
+    private static Set<Resource> resourcesExcept(Resource first, Resource... rest) {
+        return EnumSet.complementOf(EnumSet.of(first, rest));
     }
 
     public boolean isNativeArrayField() {
--- a/src/main/java/com/otilm/core/enums/FilterField.java
+++ b/src/main/java/com/otilm/core/enums/FilterField.java
@@ -83,6 +83,7 @@ import com.otilm.core.model.auth.ResourceAction;
 import com.otilm.core.model.cbom.CryptoAssetIdentityGuard;
 import jakarta.persistence.metamodel.Attribute;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 import lombok.Getter;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/otilm/core/enums/FilterField.java
```diff
@@ -282,10 +285,11 @@ public enum FilterField {
     AUDIT_LOG_ACTOR_NAME(Resource.AUDIT_LOG, null, null, AuditLog_.actorName, "Actor name", SearchFieldTypeEnum.STRING),
     AUDIT_LOG_ACTOR_AUTH_METHOD(Resource.AUDIT_LOG, null, null, AuditLog_.actorAuthMethod, "Actor Auth method",
             SearchFieldTypeEnum.LIST, AuthMethod.class),
+    // NONE is a recorded resource for module-level operations; an absent affiliated resource is stored as no value
     AUDIT_LOG_RESOURCE(Resource.AUDIT_LOG, null, null, AuditLog_.resource, "Resource", SearchFieldTypeEnum.LIST,
-            Resource.class),
+            Resource.class, resourcesExcept(Resource.ANY)),
     AUDIT_LOG_AFFILIATED_RESOURCE(Resource.AUDIT_LOG, null, null, AuditLog_.affiliatedResource, "Affiliated resource",
-            SearchFieldTypeEnum.LIST, Resource.class),
+            SearchFieldTypeEnum.LIST, Resource.class, resourcesExcept(Resource.NONE, Resource.ANY)),
```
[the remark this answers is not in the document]
klaraf755: Right, the fallback in SearchHelper still expanded every constant. It now asks the field for its values, so every catalogue that calls it without values gets the subset; the audit-log catalogue is covered by a test on both resource fields.

src/test/java/com/otilm/core/integration/service/ResourceServiceITest.java
```diff
@@ -363,6 +365,39 @@ void testListResourceRuleFilterFields() throws NotFoundException {
      * this mismatch, a missing {@code ResourceToClass} constant, or the next one -- into a build failure, which is why
      * this asserts nothing about the cause.
      */
+    @Test
```
ivosh, line 355: **~ Minor** | Documentation

ResourceServiceITest.java:368 makes the existing “Every resource that declares filter fields” Javadoc describe `commentHostResourceOffersOnlyTheCommentableResources()`, although that test exercises only `Resource.COMMENT`; the actual `everyResourceWithFilterFieldsCanBeListed()` method at line 402 is now detached from its explanation. Since Javadoc documents the immediately following declaration, this leaves a false description contrary to Google Java Style §7 and CLAUDE.md's “Writing for future AI sessions” guidance.

_Suggested:_ Move the existing Javadoc so it immediately precedes `everyResourceWithFilterFieldsCanBeListed()`. Or drop it altogether.

<sub>flagged by Java engineering & idioms + Maintainability & tests</sub>

src/main/java/com/otilm/core/enums/FilterField.java
```diff
@@ -530,10 +554,21 @@ public enum FilterField {
...
         this.settable = settable;
         this.expectedValue = expectedValue;
     }
 
+    /** The values a list field backed by an enum offers: the subset it declares, or every constant otherwise. */
+    public IPlatformEnum[] getEnumValues() {
+        return enumValues != null ? enumValues : enumClass.getEnumConstants();
+    }
+
+    /** Every resource but the given wildcards, which scope a grant and are never the resource of a stored record. */
```
ivosh: **~ Minor** | Documentation

FilterField.java:567 says every supplied wildcard is never the resource of a stored record, while FilterField.java:288 explicitly says `NONE` is stored for module-level audit rows; the missing qualifier is that a value cannot occur in the calling field. This PR repeats the corrected `NONE`/`ANY` rationale again in AuditLogServiceITest.java:104 and ResourceServiceITest.java:378, spreading the field-specific contract across four prose sites and leaving the generic helper text misleading and likely to drift.

_Suggested:_ Document the `NONE` and `ANY` exclusions once beside each affected `FilterField` constant, make `resourcesExcept`'s Javadoc purely mechanical, and remove the duplicate test comments.

ivosh (changes requested):

src/test/java/com/otilm/core/integration/service/ResourceServiceITest.java
klaraf755: Moved back onto the test it describes; the two new tests sit above it without one.

src/main/java/com/otilm/core/enums/FilterField.java
klaraf755: Done: the helper's Javadoc is mechanical, each field says beside its declaration why it drops what it drops, and the two test comments are gone.

src/main/java/com/otilm/core/enums/FilterField.java
```diff
@@ -530,10 +556,21 @@ public enum FilterField {
...
         this.settable = settable;
         this.expectedValue = expectedValue;
     }
 
+    /** The values a list field backed by an enum offers: the subset it declares, or every constant otherwise. */
+    public IPlatformEnum[] getEnumValues() {
+        return enumValues != null ? enumValues : enumClass.getEnumConstants();
+    }
+
+    /** Every resource but the given ones. */
```
ivosh, line 569: **. Nitpick** | Maintainability

FilterField.java:569 says only that `resourcesExcept` excludes the supplied resources, which the method name and its single `EnumSet.complementOf` expression already make explicit.

_Suggested:_ Remove the redundant one-line Javadoc above `resourcesExcept`.

ivosh (approved):

src/main/java/com/otilm/core/enums/FilterField.java
klaraf755: Keeping it, not worth another commit before merging.

Issue #2249, opened 2026-09-10
Make the CBOM sync tunables configurable from the Settings UI


karolhrdina:
### Description

OmniTrustILM/core#2245 (core#2208) introduced three tunables of the CBOM header sync as `application.yml` properties with environment-variable overrides:

| property | env var | default | meaning |
|---|---|---|---|
| `cbom.sync.page-size` | `CBOM_SYNC_PAGE_SIZE` | 1000 | entries requested per feed page (repository contract: 1..1000) |
| `cbom.sync.overlap` | `CBOM_SYNC_OVERLAP` | 60 s | how far before the last successful run's start each run re-lists |
| `cbom.sync.skipped-retry-runs` | `CBOM_SYNC_SKIPPED_RETRY_RUNS` | 3 | how many further runs Core retries an entry it could not store before writing it off as `PERMANENTLY_SKIPPED` |

Both reviews of that PR (Copilot, Michal) made the same point: `application.yml` and environment variables are unreachable for a SaaS operator, whose only configuration surface is the Settings UI, and `skipped-retry-runs` in particular is operator policy — it decides when a CBOM document is given up on — not deploy-time plumbing. The PR was merged with the yml placement and this story carries the move.

**Everything the CBOM sync exposes as a tunable must be configurable from the Settings UI.** All three values move into the platform settings (`PlatformSettingsDto`, the settings section the CBOM repository URL already lives in), with the defaults and validation bounds they have today, and the sync reads them at the start of each run through `SettingsCache`. `CbomSyncProperties` and the three env-var placeholders go away with them, which also closes the Helm-chart wiring question raised on the PR.

### Acceptance criteria

- The Settings UI shows the three CBOM sync values under the CBOM repository configuration, with the defaults above pre-filled and validation matching today's (`page-size` 1..1000, `overlap` ≥ 0 s, `skipped-retry-runs` ≥ 0).
- A change in the Settings UI takes effect on the next sync run without a restart.
- `application.yml` no longer carries `cbom.sync.*`, and `CBOM_SYNC_PAGE_SIZE`, `CBOM_SYNC_OVERLAP`, `CBOM_SYNC_SKIPPED_RETRY_RUNS` are gone from the codebase and from the Helm charts' scope.
- `CbomSyncPropertiesTest` is replaced by tests of the settings binding and of the sync reading the values at run start; `CbomSyncITest` keeps passing.

### Touch points

[...]

Issue #2072, opened 2026-08-14
Implement CBOM cryptographic asset extraction and normalization


karolhrdina:
### Description

Extraction and normalization of cryptographic assets out of CBOM documents
(ilm#299, design doc section 4) — the first half of the former
extraction-and-verdicts story, split because the two halves were 6 days
together. The PQC verdict half is [core#2151](https://github.com/OmniTrustILM/core/issues/2151); it is evaluated
over the merged view this story produces, so this story lands first.

Depends on [core#2071](https://github.com/OmniTrustILM/core/issues/2071) (tables, entities, repositories, writer
beans) and [interfaces#874](https://github.com/OmniTrustILM/interfaces/issues/874) (CryptographicAssetType, the DTO
shapes and the platform enums).

Streaming CycloneDX walker (nested component trees, 1.6 and 1.7);
cryptographic-asset components without cryptoProperties are skipped and
logged, never fatal (epic AC1); the ratified T1 normalization pipeline
(derivation tables, whitelist and stoplist, curve canonicalization,
primitive defaults); identity tuple and backstop computation; deterministic
property merge; redaction of material values; evidence capture (capped list
plus true count); document-scoped bom-ref arrays kept as opaque data
flagged for UI labeling. Pure, Spring-free normalizer classes.

- A cryptographic-asset component without `cryptoProperties`, or with an
  assetType the specification does not know, is NOT skipped: it keys on the
  ratified backstop tier WITH its component name and is reported as unroutable.
  Four exist in the validation corpus. (Epic AC1 needs the same correction.)
- The ingest ObjectMapper is asserted to keep INCLUDE_SOURCE_IN_LOCATION
  disabled: with it enabled, one `log.warn(msg, e)` on a malformed document
  prints hundreds of characters of raw, pre-redaction content into a log, which
  the redaction proof does not cover.

### Acceptance Criteria

- [ ] Corpus round-trips reproduce the ratified T1 harness results exactly
      (zero wrong merges, 1.6/1.7 parity, third-party corpus included).
- [ ] Redaction test proves no material value survives to persistence.
- [ ] Wild-input robustness pinned by tests: integer specVersion,
      non-numeric nistQuantumSecurityLevel, absent bom-refs, evidence-less
      assets, nonstandard fields all parse without failing the run.
- [ ] Bounded memory on a maximum-size document; per-component skip on
      malformed input; determinism across permuted ingest order.
[...]

PR #2256, opened 2026-09-14
Ingest a CBOM's cryptographic assets as one unit of work

vyskocilm:
Part of #2073 — **PR 1 of a 5-PR plan**, and deliberately not `Closes`: this lands the unit of work and the resume pass only. Supersede and the orphan rule, the deletion lifecycle, the duplicate-`bom-ref` refusal with the findings channel, and the weekly reconciliation follow as separate PRs.

## What this is

`CbomAssetExtractor` and every `crypto_asset*` writer have existed since #2071 and #2072 **with no production caller**. This composes them: per CBOM, extract in memory, write the assets in batch transactions under a new `CBOM_ASSET_SYNC` cluster lock, and leave the row saying `SYNCED` or `FAILED`.

**Two discovery sources, because the feed alone cannot resume.** `processFeedEntry` skips every entry whose header row already exists — which is exactly what a crashed ingest leaves behind, and exactly what a CBOM uploaded through Core's own API looks like (`assetSyncState` defaults to `PENDING`). Those assets would never be ingested at all. So a bounded pass over the rows that still owe an ingest runs beside the inline one, re-reading each document outside any transaction. Without it the resumability acceptance criterion cannot pass.

That pass has to tell a document's own failure from the repository's, and it has to answer that itself: the header pass's guard returns at once when the feed deferred nothing, which is the ordinary state of a run working a backlog down. A document the repository refuses is recorded as failed — an untouched row would be re-offered at the same position every run, spending a budget slot and saying nothing — while reads that got no answer at all are deferred and settled on the header pass's rule: an outage only when no read succeeded and at least two failed that way.

## Boundaries, which are the point

- State writes commit in transactions of their own, or the `FAILED` row would roll back with the write that failed.
- `ALIAS_DECISION_LOCK` is taken **before the first `crypto_asset` row lock**, because it outranks every one of them. Asserted in a unit test, since a database would show the inversion only as an intermittent deadlock.
- The document is a **parameter, never a fetch**, so no HTTP call happens under the cluster lock.
- The advisory lock is transaction-scoped and released at each batch commit — which is why a batch is bounded by `cbom.sync.asset-batch-size` rather than by the document.

## Two things worth a reviewer's attention

**`IdentityRuleset.VERSION` moves 2 → 3 and `IdentityRulesetStampArchTest` is deleted**, exactly as that tripwire instructs the first production caller of `CryptoAssetWriter` to do. Generation 2 was sound only while no environment could hold a keyed row; this commit is what ends that. The free re-key window closes here.

**A document whose cross-component scope could not be built is refused rather than ingested.** With no scope a fabricated placeholder digest is trusted and every certificate's public-key slot empties — both over-merges, and an over-merge costs a re-key to undo where a refusal is simply retried.

## New schema and configuration

- `V202609141200__cbom_asset_sync_attempted_at.sql`. `assets_synced_at` records only success, so it cannot say whether a claim is live; the retry window is measured from the start of an attempt and never renewed during one.
- `cbom.sync.asset-ingest-enabled` (true) — the kill switch, covering **both** discovery paths and checked inside the unit of work as well as by its two callers. `max-ingest-documents: 0` disables only the backlog pass, so it never was one.
- `cbom.sync.asset-batch-size` (100), `cbom.sync.max-ingest-documents` (50), `cbom.sync.ingest-retry-after` (30m).

> [!IMPORTANT]
> The four new values carry deploy-time env overrides (`CBOM_SYNC_ASSET_INGEST_ENABLED`, `CBOM_SYNC_ASSET_BATCH_SIZE`, `CBOM_SYNC_MAX_INGEST_DOCUMENTS`, `CBOM_SYNC_INGEST_RETRY_AFTER`) that are **not yet wired into the Helm charts**, so they are unreachable in a deployed environment until that lands in the charts repo. Whether these belong in the Settings UI instead should be decided with #2249, which is already moving the `cbom.sync` tunables — they extend an existing `@ConfigurationProperties` record whose three older values live in yml today, and a new settings section is an `interfaces` release rather than a `core` commit.

## Concurrency

The backlog pass claims each row **before** it reads the document. Selecting the work list is a plain read, so every node's hourly job selects the same rows, and without a claim each spends the same HTTP document read and the same extraction before the cluster lock serialises the first write. The writes converge — every one is an idempotent upsert — so what the claim saves is load on the CBOM repository, at its worst exactly when a backlog is being worked down.

It is a **compare-and-swap on what the work list saw**, not a state allowlist: an allowlist admitting `IN_PROGRESS` would let the loser claim the row the winner just took. A claim the run cannot use is given back, because an outage answers no document read at all and rows left `IN_PROGRESS` by it would claim work no node is doing.

## Deletion, and why a switch rather than the lifecycle

`crypto_asset_source_to_cbom_key` is `ON DELETE RESTRICT` and this is the first production writer of `crypto_asset_source`, so from this PR on a CBOM that has acquired sources cannot be deleted through the API until the withdrawal (`CryptoAssetSourceWriter.detachCbom`) and the `cbom_tombstone` land. That wiring stays its own PR; folding it in here doubles this one.

`cbom.sync.asset-ingest-enabled` is what bounds the interval: turning ingest off stops any further CBOM acquiring sources. `CryptoAssetSourceWriter.detachCbom`'s Javadoc now says the exposure is live and names the switch, rather than describing the wiring as unreachable.

## Verification

1703 tests, 0 failures under `mvn verify`, against real PostgreSQL via Testcontainers: `CbomServiceITest` 53, `CbomSyncITest` 29, `CryptoAssetInventoryITest` 39, `CbomSyncTaskITest` 9, `CryptoAssetInventoryMigrationITest`, `CryptoAssetSearchITest`, `CryptographicAssetServiceITest`, `CryptographicAssetStatisticsITest`, `CbomSyncPropertiesTest` 11, `CbomAssetIngestServiceTest` 12, `CbomAssetExtractorTest` 44, plus the whole `com.otilm.core.architecture.*` and `com.otilm.core.cbom.**` suites — `IdentityKeyExposureFenceArchTest`, `TransactionalBoundaryArchTest`, `NativeQuerySchemaArchTest`, `ContextSignatureGuardTest` (BASELINE 61 unchanged — no new Spring context was bought). `spotless:apply checkstyle:check` clean.

The convention gate (`ilm-arch-check`) and an independent `copilot` correctness pass both ran over the stack; their findings against this PR's code are fixed here rather than deferred. The review of 2026-09-14 raised thirteen further findings and two blocking items; all are fixed in `ea5146f2`, with one recorded rather than changed — `applyPqcVerdict` bumping `i_upd` to the same instant as `pqc_evaluated_at`, which is the sweep's own query and predates this PR.

**Not done:** Sonar new-lines coverage parity has not been computed locally, so the gate's `new_coverage` figure is unverified ahead of CI.


karolhrdina (commented):
Reviewed the full diff plus the collaborators it composes. Several findings below were confirmed empirically against real PostgreSQL rather than inferred; those say so.

The composition is well argued and the boundaries it documents are the right ones to document. Most of what follows is about what happens at the edges of those boundaries, plus two things that block.

## Blocking

### 1. CI is red on this branch

`CbomSyncITest.anEntryAlreadyInTheDatabaseIsADuplicateWithoutADocumentRead` fails:

```
Expected exactly 0 requests matching the following pattern but received 1:
{ "urlPath" : "/api/v1/bom/urn:uuid:dup", "method" : "GET" }
```

`ingestPending` re-reads the document for exactly the entry the feed pass skipped as a duplicate, which is the contract that test pins. That may well be the intended new behaviour — but it is a deliberate change to a stated contract, and it should be made deliberately in that test rather than discovered in CI. The verification section lists `CbomServiceITest`, `CryptoAssetInventoryITest`, `CbomSyncTaskITest`, `CbomSyncPropertiesTest` and `CbomAssetIngestServiceTest`; `CbomSyncITest` is not among them.

Worth noting the production consequence separately from the test: from this PR on, every duplicate feed entry whose assets are still `PENDING` costs a document read that the feed pass was written to avoid.

### 2. Every successfully ingested CBOM becomes undeletable through the API, and ingest cannot be turned off

`crypto_asset_source_to_cbom_key` is `ON DELETE RESTRICT`, and the migration that declared it assigned the remedy to this ticket by name (`V202608271000__crypto_asset_inventory.sql:112-120`):

> The delete path must therefore detach this CBOM's sources first -- `CryptoAssetSourceWriter.detachCbom` over `CryptoAssetSourceRepository.findAssetUuidsByCbomUuid` -- and then record a `cbom_tombstone` so the next sync does not simply re-ingest the document. [...] **The ingest ticket owns that wiring, and it is not optional** -- without it the first CBOM to acquire a source becomes undeletable through the API, refused by this constraint with no endpoint able to detach.

This PR is the first production writer of `crypto_asset_source`. `detachCbom` still has zero production callers — `grep` finds it only in `CryptoAssetInventoryITest` — and `deleteCbom`/`bulkDeleteCbom` only shape the refusal message. No `cbom_tombstone` is recorded either. The new `CbomServiceITest.setUp` had to add `cryptoAssetSourceRepository.deleteAll()` ahead of `cbomRepository.deleteAll()` for exactly this reason, which is the same failure an operator will hit.

It compounds because there is no kill switch: `cbom.sync.max-ingest-documents: 0` disables only `ingestPending`; the inline ingest in `store()` is ungated. So an operator who hits this cannot stop further CBOMs from acquiring sources without a rollback — and per the PR body the three new tunables are not wired into the charts, so even the backlog pass cannot be disabled in a deployed environment.

Either land the detach path (plus the tombstone) here, or gate the whole ingest — both passes — behind something reachable in a deployed environment.

## Also worth fixing before merge

Left inline, in rough order of impact:

- `CbomServiceImpl:825` — the inline ingest sits outside `store()`'s try/catch and breaks its documented "never throws" contract; one DB hiccup aborts the whole run and skips `settleUnavailable`.
- `CbomAssetIngestService:202` — under open-in-view (the REST `sync` path) the verdict is computed from a stale entity, and the mis-stamp is permanent because the sweeper's staleness test can never see it. Reproduced.
- `CbomAssetIngestService:207` — one unevaluable asset rolls back its whole batch and fails the document for ever.
- `CbomAssetIngestService:149` / `:170` — the global cluster lock and the row claim work against each other, and `LOCKED_ELSEWHERE` strands the claim it cannot use.
- `CbomAssetIngestService:181` — lock-order gap against `CryptoAssetPqcVerdictWriter.applyStaleBatch`.
- `CbomAssetIngestService:262`, `CbomRepository:125`, the new index — smaller, each described inline.

I have also confirmed Copilot's `:184` comment is real and larger than an edge case; details inline there.

## Checked and clean

Recording these so they do not get re-litigated. All verified, several by execution:

- `updateAssetSyncState`'s `:expectedStates IS NULL OR ... IN :expectedStates` with a null `Collection` **works** — Spring Data JPA 3.5.13 rewrites the second occurrence to its own parameter, so Hibernate 6.6.53 keeps multi-valued binding. Confirmed by emitted SQL and by running the ITests. (It is Spring-Data-specific: the same HQL through a plain `EntityManager` throws `ClassCastException: RegularEnumSet cannot be cast to Enum`. Worth a one-line note so nobody lifts it out.)
- `NULLS FIRST` in JPQL and `Limit` as a query-method parameter are both supported at these versions.
- The claim's timestamp compare-and-swap round-trips exactly through `TIMESTAMPTZ`; no timezone or precision hazard, given the column is only ever written by SQL.
- `storedUuid.get()` cannot be null at the ingest call — every path that skips the `set` returns first, and `@PrePersist` assigns the uuid before `save` returns.
- No stale work-list reads and no `LazyInitializationException`: `Cbom` has no associations, and `claimForIngest` is `clearAutomatically = true` with `claimedFrom` captured beforehand.
- The documented lock ranking is obeyed, and `pg_advisory_xact_lock` really is re-entrant within a transaction (verified).
- `seenAt` does not participate in the cross-source merge election, so a backlog re-read stamped `now` cannot beat a newer CBOM.
- Format specifiers in `SyncRun.summary()` match their arguments on both halves.
- No duplicate `CbomAssetExtractor` bean, no circular dependency, `subList` views are safe over the immutable `Extraction.assets()`, and a null `cryptoProperties` would be handled (though `MaterialRedaction` means it never occurs).
- `ingestPending` cannot re-pick rows the same run ingested inline.

## Questions

1. **`asset_sync_error` has no read path.** `CbomDto` in the pinned `interfaces` artifact exposes only `assetSyncState` and `assetSyncedAt`; `Cbom.mapToDto` does not copy the error, and nothing else in `src/main` reads the column. So every operator-safe sentence this PR is careful to shape is, today, invisible. Is that waiting on the findings channel in a later PR?
2. **Backlog drain rate.** 50 documents an hour on an hourly cron is 1 200 a day. Is that sized against the CBOM population this will meet on first rollout, given every existing row is `PENDING`?
3. Stale prose that the tripwire's removal makes wrong, and that a future reader will act on: `IdentityTablesTest:47-51` still says generation 2 has never keyed a row and that the bump belongs to whoever first gives the writer a production caller; `CryptoAssetPqcVerdictWriter:18-19` cites `IdentityRulesetStampArchTest` as live; `CryptoAssetSourceWriter:47` says "nothing composes these writers today"; `CryptoAssetRepository:320` says the curve-list budget is "unmeasurable until ingest puts rows in the table". Also `PqcConfig`'s javadoc claim that a bad table load "fails the sweep that asks for it rather than refusing the whole application at boot" is no longer true — the eager `CbomAssetIngestService` defeats `@Lazy` on both `AssetNormalizer` and `PqcEvaluator`, which `CbomIngestConfig` states for the extractor but not for these.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -791,7 +816,14 @@ private StoreOutcome store(SyncIdentity identity, CbomHeaderCounts counts, SyncR
...
             return StoreOutcome.failed("storing the CBOM row failed unexpectedly (see the Core log)");
         }
-        return isDuplicate.get() ? StoreOutcome.DUPLICATE : StoreOutcome.STORED;
+        if (isDuplicate.get()) {
+            return StoreOutcome.DUPLICATE;
+        }
+        // The header row is committed, so the assets are ingested with the document already in hand rather than left
+        // to ingestPending, which would re-read it. A failure here is the CBOM's ingest state to record, never the
+        // header's: the row is stored either way, and the run's own counters carry what the ingest did.
+        countIngest(assetIngestService.ingest(storedUuid.get(), document, run.startedAt), run);
```
karolhrdina: **`store()`'s "never throws" contract is broken here, and one transient database failure aborts the whole run.**

The javadoc twelve lines above this method says *"Never throws for one entry's failure -- the outcome is the caller's signal."* This call sits outside the try/catch that closes at :818, and `ingest` is not exception-safe: `markInProgress` (`CbomAssetIngestService:126`), `markSynced` (`:157`) and the `markFailed` in `refuse`/`fail` (`:217`, `:222`) are all unwrapped, as is `valueToTree` (`:119`). Only `extract` and the batch loop are guarded.

So a `CannotAcquireLockException` / `QueryTimeoutException` / `JDBCConnectionException` on the `cbom` row update for one entry propagates out of `store()` → `processFeedEntry` → `readFeed` → `runSync` → `sync()`. The cost is not that entry:

- the remaining pages and entries of the run are abandoned;
- **`settleUnavailable(run)` at :560 never runs**, so every entry already on `run.deferred` gets no skip record and no retry-budget charge at all — silently dropped from this run's bookkeeping;
- `ingestPending(run)` never runs and `run.summary()` is lost;
- `CbomSyncTask` records `FAILED`, and since `getLastSyncTimestamp()` reads only the last SUCCESS row the watermark does not advance and the whole feed window is re-read next hour.

Same shape at `:1222` (`ingestOnePending`'s ingest call is outside its try/catch) and at the unguarded writer calls in `ingestPending` — `claimed(cbom)` at :1102, `markFailed` at :1197/:1210/:1219, and `settleUnreadable`'s writes at :1160/:1164.

Wrapping this call and the body of the `ingestPending` loop in a per-entry catch would restore the isolation the rest of `store()` already has.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -0,0 +1,264 @@
...
+     *
+     * <p>
+     * Read back rather than evaluated from the document: an asset several CBOMs report is evaluated on the payload the
+     * merge elected, not on whichever document happened to arrive last. Stamping here rather than leaving the rows to
+     * {@code PqcVerdictSweeper} is a latency decision -- a null {@code pqc_ruleset_version} is already stale to the
+     * sweep -- and it matters most on the first ingest, when the sweep has the largest backlog it will ever have.
+     */
+    private void stampVerdicts(Iterable<UUID> assetUuids) {
+        for (UUID assetUuid : assetUuids) {
+            final CryptoAsset row = assetRepository.findById(assetUuid).orElse(null);
```
karolhrdina: **The verdict is computed from a stale entity whenever `sync` is invoked over HTTP, and the wrong verdict is permanent.** Reproduced against PostgreSQL via Testcontainers.

`spring.jpa.open-in-view` is not set anywhere, so Spring Boot's default (`true`) applies and `OpenEntityManagerInViewInterceptor` is registered — and other code in this repo already assumes it is on (`AcmeServiceImpl:1078`). `CbomControllerImpl.sync()` → `cbomService.syncAuthorized()` → `runSync()` reaches this code inside that bound `EntityManager`, so `JpaTransactionManager.doBegin` reuses the OSIV-bound `EntityManagerHolder` instead of creating a new one: **all of a document's `REQUIRES_NEW` batch transactions share one persistence context**, and a commit does not evict entities.

Within a single batch the read is fresh — nothing before `stampVerdicts` loads a `CryptoAsset` (`findUuidByIdentityKey` and `lockForSourceChange` are scalar projections, the rest are native `@Modifying`). The exposure is **cross-batch within one document**: any asset uuid touched in batch *i* and again in batch *j>i* is served from the first-level cache, without what `recomputeMergeFromSources` wrote in between. For any CBOM with more than `asset-batch-size` (100) assets, a common algorithm straddling batches is close to certain — and the comment at :177-178 already acknowledges that two components can key as the same asset.

Measured, same document, `asset-batch-size=1`, two components keying to one RSA-2048 asset (sparse first, rich second):

```
NO OSIV    evaluated={"algorithmFamily":"RSA","nistQuantumSecurityLevel":0}
WITH OSIV  evaluated={"algorithmFamily":"RSA"}
```

with `merged_crypto_properties` holding the rich payload in the database in both cases.

It does not self-heal. `recomputeMergeFromSources` sets `i_upd = CURRENT_TIMESTAMP` and `applyPqcVerdict` sets `pqc_evaluated_at` *and* `i_upd` to the same `transaction_timestamp()` — measured equal to the microsecond — so `findStaleVerdictRows`' `pqc_evaluated_at < i_upd` is false and `pqc_ruleset_version` is current. `PqcVerdictSweeper` never re-offers the row until the ruleset version moves.

Cross-*document* staleness is not affected: `markInProgress` goes through `updateAssetSyncState`, which is `@Modifying(clearAutomatically = true)`, and that does clear the shared context.

Options: read the payload through a native projection the way `findStaleVerdictRows` already does (`merged_crypto_properties::text`), or `em.refresh`, or add `clearAutomatically = true` to `recomputeMergeFromSources`. Separately worth reconsidering that `applyPqcVerdict` bumps `i_upd` to the same instant as `pqc_evaluated_at` — that defeats the sweeper's own staleness detector for *any* mis-stamp, not just this one.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -0,0 +1,264 @@
...
+            if (row == null) {
+                continue;
+            }
+            final JsonNode merged = JSON_COLUMN.valueToTree(row.getMergedCryptoProperties());
+            final PqcDecision decision = evaluator
```
karolhrdina: **No per-asset failure isolation: one unevaluable asset destroys the whole ingest, permanently.**

This `evaluate` call has no try/catch. A `RuntimeException` from it propagates out of `writeBatchUnderClusterLock`, rolls back the entire batch transaction — **including every `upsertIdentity` and `upsertSource` already done in that batch** — and is caught at :152, which marks the whole CBOM `FAILED`. `findAssetIngestRetries` then re-offers it every `ingest-retry-after` for ever, with the same outcome and the same cost.

The established path guards exactly this per row: `PqcVerdictSweeper.decideOrRecordFailure` (`:175-186`) catches `RuntimeException`, counts `crypto_asset.pqc_sweep.evaluation_failed` and stamps an `EVALUATION_FAILED` sentinel, so one bad row costs one row.

The javadoc directly above calls stamping here "a latency decision" — by that argument it should not be able to fail the unit of work. A try/catch around the loop body that logs the uuid and moves on would leave `pqc_ruleset_version` null, which is precisely how the sweeper finds the row later.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -0,0 +1,264 @@
...
+            return refuse(cbomUuid,
+                    "the document's cross-component scope could not be built, so its assets cannot be keyed safely");
+        }
+
+        try {
+            for (List<CbomAssetExtractor.ExtractedAsset> batch : batches(extraction.assets())) {
+                final boolean locked = transactionHandler
+                        .runInNewTransaction(() -> writeBatchUnderClusterLock(cbomUuid, batch, seenAt));
+                if (!locked) {
+                    return IngestOutcome.LOCKED_ELSEWHERE;
```
karolhrdina: **`LOCKED_ELSEWHERE` strands the claim it cannot use — the one case the rest of the PR handles and this one does not.**

`markInProgress` at :126 is unconditional and runs *before* the ingest knows it can do any work. Returning here writes no further state, so the row is left `asset_sync_state = IN_PROGRESS`, `asset_sync_error = NULL`, `asset_sync_attempted_at = now`, with batches `0..k-1` written and `k..N` not.

The javadoc on the enum says *"the CBOM keeps owing an ingest and the next run offers it again."* It does not. The row has left `findPendingAssetIngests` (PENDING only) and `findAssetIngestRetries` requires `assetSyncAttemptedAt < startedAt - ingest-retry-after`, so it is invisible for at least 30 minutes — it skips the next hourly run entirely, with a half-ingested asset set and no error text an operator could read.

`settleUnreadable` already does the right thing for the sibling case, and says why: *"a row left IN_PROGRESS by an outage claims work no node is doing, and it would wait out `cbom.sync.ingest-retry-after` instead of being offered again at once."* That is exactly this situation. Either release the claim here, or take the cluster lock before `markInProgress` so the claim is never written for work that cannot start.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -0,0 +1,264 @@
...
+        return IngestOutcome.INGESTED;
+    }
+
+    /**
+     * One batch, inside the transaction that holds the cluster lock. Returns false when another node holds it, which is
+     * a skip rather than a failure: the CBOM keeps owing an ingest and the next run offers it again.
+     */
+    private boolean writeBatchUnderClusterLock(UUID cbomUuid, List<CbomAssetExtractor.ExtractedAsset> batch,
+            OffsetDateTime seenAt) {
+        if (!clusterSynchronizer.tryLock(ClusterOperationSynchronizer.Operation.CBOM_ASSET_SYNC)) {
```
karolhrdina: **`CBOM_ASSET_SYNC` is one global key, which cancels out the row claim this PR ships alongside it.**

`Operation.CBOM_ASSET_SYNC` carries a single lock key for the whole operation, not one per CBOM. So at most one node in the cluster can be writing assets at any instant, for any document.

That defeats the stated purpose of the claim. Per `claimAssetIngest`'s javadoc the claim exists so that "the load on the CBOM repository" does not multiply by the number of nodes. But the claim is taken *before* the document read and the lock is tried *after* it: a node that loses this `tryLock` has already spent the HTTP read and the extraction, and then throws both away. Combined with the stranding above, the loser also parks each of its rows for `ingest-retry-after`.

Concretely, hourly cron `0 0 * ? * *` fires on every node in the same minute. With three nodes, two of them claim roughly two-thirds of the 50-document budget, spend 50 document reads and 50 extractions between them, write nothing, and move those rows from `PENDING` into the retry list — which `ingestPending` only consults once the pending list is exhausted. Cluster throughput is capped at one node's rate however many nodes there are, and the extra nodes make the backlog *slower* to drain, not faster.

Per-entity advisory keys are already the pattern here — `"signing-profile:" + uuid` (`SigningProfileServiceImpl:495`), `"list-view:" + ...` (`ListViewServiceImpl:151`), `CommentWriter.hostLockKey`. Keying this lock on the CBOM uuid would make the claim and the lock agree; it needs a `tryLock(String)` overload on `ClusterOperationSynchronizer`, which has only the blocking `lock(String)` today. Alternatively drop the cluster lock entirely — the row claim already serialises per CBOM and every write is an idempotent upsert.

Two smaller things in the same area:

- `clusterSynchronizer.lock(ALIAS_DECISION_LOCK)` at :175 blocks with no timeout while holding `CBOM_ASSET_SYNC`, and no `lock_timeout`/`statement_timeout` is configured anywhere. One stuck operator alias-repair transaction therefore stalls the entire cluster's ingest.
- The implied third rank — `CBOM_ASSET_SYNC` above `ALIAS_DECISION_LOCK` — is safe only because `CBOM_ASSET_SYNC` is never acquired with a blocking call. Worth writing down next to the existing ranking so a future blocking acquisition does not quietly close a cycle.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -0,0 +1,264 @@
...
+        }
+        // Before the first crypto_asset row lock any writer below will take. Re-entrant within the transaction, so
+        // taking it here costs the writers' own acquisitions nothing.
+        clusterSynchronizer.lock(CryptoAssetAliasWriter.ALIAS_DECISION_LOCK);
+
+        // A set, not a list: two components of one document can key as the same asset, and the verdict pass reads
+        // each row back once, after every source of the batch has been merged into it.
+        final Set<UUID> written = new LinkedHashSet<>();
+        for (CbomAssetExtractor.ExtractedAsset asset : batch) {
+            final UUID assetUuid = assetWriter.upsertIdentity(asset.identityKey(), fieldsOf(asset), asset.guard());
```
karolhrdina, line 253: **Lock-order gap against the PQC sweep, on `crypto_asset` row locks.**

This loop takes `crypto_asset` row locks in **document order** (`upsertIdentity`'s `ON CONFLICT`, then `lockForSourceChange`, then `applyPqcVerdict`). `CryptoAssetPqcVerdictWriter.applyStaleBatch` updates up to `batchSize` rows in **one** transaction in **uuid order**, because `findStaleVerdictRows` ends `ORDER BY uuid`.

The two run concurrently — the sweep holds `CRYPTO_ASSET_PQC_SWEEP`, not `CBOM_ASSET_SYNC` — and every asset this ingest has just created is immediately on the sweep's work list, since `upsertIdentity` leaves `pqc_ruleset_version` null. Two transactions updating an overlapping row set in opposite orders is a textbook PostgreSQL deadlock. On this side the loser surfaces as `DeadlockLoserDataAccessException` → the catch at :152 → the CBOM reads `FAILED` and waits out `ingest-retry-after`.

Sorting the batch's asset uuids before writing would close it, and costs nothing — the comment at :177-178 already shows the set is being built for its own reasons.

(Inferred from the two orderings rather than reproduced live, so treat the timing as plausible rather than measured — but the orderings themselves are not in question.)

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -0,0 +1,264 @@
...
+        final JsonNode retained = asset.retainedProperties();
+        return retained == null || retained.isNull() ? null : JSON_COLUMN.convertValue(retained, Map.class);
+    }
+
+    /**
+     * The operator-visible half of a failure. A driver message is never it: a constraint violation's DETAIL line
+     * carries the failing row, and for {@code crypto_asset} that row carries the identity key.
+     */
+    private static String safeReason(RuntimeException e) {
+        return CryptoAssetConstraintTranslator.describe(e);
```
karolhrdina, line 376: **Every non-constraint failure is recorded as "it would violate a database constraint", which is false.**

`CryptoAssetConstraintTranslator.describe` walks the cause chain for a Hibernate `ConstraintViolationException` and returns `GENERIC_REJECTION` — *"The cryptographic asset inventory rejected the change: it would violate a database constraint."* — for anything else. But `fail()` routes **every** `RuntimeException` from the write phase through here, and several of them are not constraint violations and carry text the operator could actually act on:

- `CryptoAssetWriter.upsertIdentity` → `requireNoAlias` throws `ValidationException` with *"The {guard} safety rule would keep this cryptographic asset separate, but an alias already merges it into another asset. Withdraw that alias first..."* — actionable, and already operator-shaped. It becomes "would violate a database constraint".
- `requireIdentityKeyShape` → `ValidationException("Cryptographic asset identity key has invalid shape")`, which is a pipeline bug, reported as a database constraint.
- The `IllegalStateException` for "the cryptographic asset row disappeared between its upsert and its lookup".
- Lock-acquisition failures, query timeouts, connection failures.

`CbomAssetIngestServiceTest.aRefusedWriteLeavesTheCbomFailedWithTextNoDriverWrote` does not catch this: it throws a `DataIntegrityViolationException` whose message is a driver string but which has **no `ConstraintViolationException` cause**, so `describe` takes the generic fallback and the assertions (`doesNotContain("identity_key")`, `doesNotContain("DETAIL")`) pass trivially. The test proves the fallback is safe, not that the mapping works.

Passing a `ValidationException`'s own message through, and keeping `describe` for the `DataIntegrityViolationException` branch, would preserve the safety property (neither path touches a driver message) while saying something true.

src/main/java/com/otilm/core/dao/repository/CbomRepository.java
```diff
@@ -29,6 +31,103 @@ public interface CbomRepository extends SecurityFilterRepository<Cbom, UUID> {
...
+     * made -- the read was tried and the repository did not answer -- and that column records attempts, not successes.
+     *
+     * <p>
+     * Guarded on {@code IN_PROGRESS}, which is the claim this caller holds: a row another node has since taken further
+     * is not this run's to rewind.
+     */
+    @Modifying(clearAutomatically = true, flushAutomatically = true)
+    @Query("""
+            UPDATE Cbom c
+            SET c.assetSyncState = :previousState
```
karolhrdina, line 142: **Releasing a claim permanently erases that row's `asset_sync_error`.**

`claimAssetIngest` sets `c.assetSyncError = NULL` (:95); this statement restores only `assetSyncState`. Reachable path: a `FAILED` row is offered by `findAssetIngestRetries`, claimed (error cleared), the document read then answers 503, `unavailableAsAWhole` defers it, and `settleUnreadable` releases the claim back to `FAILED`. The row ends the run `FAILED` with a null reason, and `ingest-retry-after` means it sits that way for at least 30 minutes.

The javadoc on `updateAssetSyncState` calls this column operator-visible, so losing it is a real loss even though nothing reads it yet. Either carry the previous error back in the release, or do not clear it at claim time — the ingest's own `markInProgress` clears it again anyway, at the point where an attempt genuinely starts.

src/main/resources/db/migration/V202609141200__cbom_asset_sync_attempted_at.sql
```diff
@@ -0,0 +1,14 @@
...
+-- failed? The `cbom` table is not audited (it carries no `i_upd`), so there is no existing column that moves on every
+-- attempt.
+--
+-- Null means never attempted, which is what every row holds today: nothing has ever ingested assets.
+ALTER TABLE "cbom"
+    ADD COLUMN "asset_sync_attempted_at" TIMESTAMPTZ;
+
+-- The retry list reads (state, attempted_at) together and orders by the timestamp; the existing state index alone
+-- leaves the ordering to a sort over every pending row.
+CREATE INDEX "idx_cbom_asset_sync_attempt" ON "cbom" ("asset_sync_state", "asset_sync_attempted_at");
```
karolhrdina, line 23: **This index does not supply the ordering the comment credits it with.**

Measured on PostgreSQL 16 with 500 000 rows, the index present, and the exact predicate and `ORDER BY` from `findAssetIngestRetries`:

```
Limit -> Sort (Sort Key: asset_sync_attempted_at NULLS FIRST, uuid)
          -> Bitmap Heap Scan on cbom
               Filter: ((asset_sync_attempted_at IS NULL) OR (asset_sync_attempted_at < ...))
               -> Bitmap Index Scan on idx_cbom_asset_sync_attempt
                    Index Cond: (asset_sync_state = ANY ('{IN_PROGRESS,FAILED}'))
```

Three independent reasons the sort survives: the `assetSyncAttemptedAt IS NULL OR ... < :retryBefore` disjunction is not an index range condition, so the timestamp lands in the heap `Filter` rather than the `Index Cond`; `NULLS FIRST` is the opposite of a default btree's ASC/NULLS LAST; and `c.uuid` is the second sort key and is not in the index. Re-running without `NULLS FIRST`, without the `uuid` tiebreaker, and without the `IS NULL` arm left a `Sort` node in every variant.

So over the pre-existing `idx_cbom_asset_sync_state` this index buys nothing for this query, while making that single-column index a redundant strict prefix the migration does not drop. Either drop the ordering claim from the comment (and consider dropping `idx_cbom_asset_sync_state`), or index `("asset_sync_state", "asset_sync_attempted_at" NULLS FIRST)` and restructure the disjunction so the timestamp can be a range condition.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -0,0 +1,264 @@
...
+        clusterSynchronizer.lock(CryptoAssetAliasWriter.ALIAS_DECISION_LOCK);
+
+        // A set, not a list: two components of one document can key as the same asset, and the verdict pass reads
+        // each row back once, after every source of the batch has been merged into it.
+        final Set<UUID> written = new LinkedHashSet<>();
+        for (CbomAssetExtractor.ExtractedAsset asset : batch) {
+            final UUID assetUuid = assetWriter.upsertIdentity(asset.identityKey(), fieldsOf(asset), asset.guard());
+            sourceWriter
+                    .upsertSource(assetUuid, cbomUuid, propertiesOf(asset), asset.evidence(),
+                            asset.reportedOccurrences(), seenAt);
```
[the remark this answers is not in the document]
karolhrdina: Confirming Copilot's reading of this, and that it is larger than an edge case.

`ExtractedAsset` is produced **per component** — `CbomAssetExtractor:178-193` appends one per cryptographic-asset component with no dedup by key, and `reportedOccurrences` is that component's own `occurrences.size()`. The arbiter is `ON CONFLICT (asset_uuid, cbom_uuid)`, and every content column in the `DO UPDATE` is assigned rather than accumulated, gated on `EXCLUDED.last_seen_at >= crypto_asset_source.last_seen_at`. Since `seenAt` is one constant for the whole document, that gate is a tie, `>=` is true, and the later component **overwrites** `original_crypto_properties`, `properties_leaf_count`, `properties_hash`, `evidence` and `occurrence_count`.

So a CBOM that reports one algorithm from several components keeps only the last one. Component A with a rich `cryptoProperties` and 40 occurrences followed by component B with `{}` and 2 leaves the source row holding B's empty payload and `occurrence_count = 2`; `recomputeMergeFromSources` then re-elects on the impoverished row, and the list page's `COALESCE(SUM(s.occurrenceCount), 0)` under-reports. It crosses batch boundaries too, since each batch is its own transaction.

Worth noting the new `reportedOccurrences` overload is not the cause — the old one lost the same data — but it does change *which* number survives to one component's pre-cap count rather than anything aggregate, which is the opposite of what the overload's javadoc is arguing for.

Grouping the batch by `identityKey` (or by resolved `assetUuid`) before writing — concatenating evidence, summing `reportedOccurrences`, and picking the payload with the highest `CryptoPropertiesDigest.leafCount` — would fix it in one place.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Fixed. The claim goes back.

`ingest` reads the row's state before `markInProgress` (a scalar projection, `CbomRepository.findAssetSyncState`, so nothing lands in a bound persistence context) and a `LOCKED_ELSEWHERE` return calls `releaseClaim` with it. `SYNCED` is excluded, and `releaseClaim` is itself guarded on `IN_PROGRESS`, so a row another node has taken further is not rewound.

`aContendedClusterLockGivesTheClaimBackAndLeavesTheCbomOwingAnIngest` pins it.

Much less reachable now in any case: with the lock keyed per CBOM (thread below) contention here means another node is ingesting *this* document.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Fixed, on the pattern you name.

The evaluate-and-stamp loop body is wrapped per row: a `RuntimeException` counts `crypto_asset.ingest.verdict_failed`, logs the uuid (never the key) and moves on, leaving `pqc_ruleset_version` null — which is exactly what `findStaleVerdictRows` selects on.

`CbomAssetIngestServiceTest.anUnevaluableAssetDoesNotFailTheDocument` pins it: the document still reads `SYNCED`.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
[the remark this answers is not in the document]
vyskocilm: Fixed, and confirmed as larger than an edge case — both readings in this thread are the one implemented.

Folded before batching, in `CbomAssetExtractor.ExtractedAsset.coalesceByIdentity`: occurrence counts sum, evidence concatenates, and the payload with the highest `CryptoPropertiesDigest.leafCount` survives. A tie keeps the earlier component, so the result does not depend on document order.

Grouped by `identityKey` rather than by resolved `assetUuid`, so the fold happens once per document instead of once per batch and cannot be split across a batch boundary.

It lives on the record rather than in the ingest because the fold reads the identity key: `IdentityKeyExposureFenceArchTest` refuses a method that reads a carrier accessor and hands a carrier back, and registering the ingest's helper broke two sibling rules of the same fence. Reading its own field inside the record is what the fence is content with, and it is the right home anyway.

Agreed that the `reportedOccurrences` overload was not the cause. With the fold it now carries the document's total, which is what its Javadoc was arguing for.

Covered by `CbomAssetExtractorTest.componentsThatKeyAsOneAssetFoldIntoOne` and `anEquallyRichPayloadDoesNotDisplaceTheEarlierComponent`, and end to end by `CbomAssetIngestServiceTest.componentsThatKeyAsOneAssetAreWrittenOnce`.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Fixed, and thank you for measuring it — the permanence is what makes it worth the extra query.

`stampVerdicts` no longer reads the entity. New `CryptoAssetRepository.findVerdictRowsByUuids` / `verdictRowsByUuids` is the same native projection `findStaleVerdictRows` uses (`merged_crypto_properties::text`, `array_to_string(curve, '+')`, `xmin`), so no first-level cache can serve it whatever persistence context is bound. `CompositeCurve.join` is no longer needed on this path and the `findById`-based `fieldsOf(CryptoAsset)` overload is gone.

Ordered by `uuid`, which also settles the lock-order thread below.

`applyPqcVerdict` setting `i_upd` to the same instant as `pqc_evaluated_at` is not changed here. It is the sweep's own query and the behaviour predates this PR; changing it would re-offer every ingested row to the next sweep. Recorded as open work on #2073 rather than carried in this PR.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Fixed. The batch writes in uuid order.

`writeBatchUnderClusterLock` resolves each asset's existing row first (`findUuidByIdentityKey`, a plain read taking no lock) and sorts on it; an asset with no row yet sorts last and keeps document order, because its uuid is minted by the insert and it holds no lock the sweep can be waiting on. `stampVerdicts` reads its rows through a query that is itself `ORDER BY uuid`.

It narrows rather than closes: a row another node inserts between the resolution and the write is ordered as new. Said so in the comment, and the sweep's retry-one-transaction-at-a-time path stays the backstop.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Fixed — keyed per CBOM.

New `ClusterOperationSynchronizer.tryLock(String)`, the non-blocking twin of `lock(String)` over the same `hashtext` keyspace. The key is `CbomAssetIngestService.assetSyncLockKey(uuid)` = `"cbom-asset-sync:" + uuid`, so two nodes working different documents do not contend and the row claim and the lock now agree. `Operation.CBOM_ASSET_SYNC` is removed rather than left unused.

The two smaller ones:

- `lock(ALIAS_DECISION_LOCK)` still blocks with no timeout, but it is now held under a per-document lock rather than a cluster-wide one, so a stuck alias repair stalls the documents that share its assets, not every ingest on every node. No `lock_timeout` is configured anywhere in this repo; adding one is a platform-wide change and not this PR's.
- The third rank is written down, on the class Javadoc: asset-sync above `ALIAS_DECISION_LOCK` above every `crypto_asset` row lock, safe only because asset-sync is never acquired with a blocking call.

src/main/java/com/otilm/core/dao/repository/CbomRepository.java
vyskocilm: Fixed, the second way you suggest.

`claimAssetIngest` no longer sets `assetSyncError = NULL`. The ingest's own `markInProgress` clears it where an attempt genuinely starts, so nothing is lost and the release path has nothing to carry back. Written down on the query: a claim is not an attempt.

src/main/resources/db/migration/V202609141200__cbom_asset_sync_attempted_at.sql
vyskocilm: Fixed — the claim goes, and so does the redundant index.

The comment now says the index carries the timestamp to keep the retry list's filter off the heap, and states plainly that it does not supply the ordering and that no index can while the query keeps that disjunction, the `NULLS FIRST` and the `uuid` tiebreaker.

`idx_cbom_asset_sync_state` is dropped in the same migration, since the composite has it as a strict prefix.

`CryptoAssetInventoryMigrationITest` now applies this migration after `V202608271000` and asserts the resulting index set, so the drop is covered rather than asserted only in prose.

src/main/java/com/otilm/core/dao/repository/CbomRepository.java
```diff
@@ -29,6 +31,103 @@ public interface CbomRepository extends SecurityFilterRepository<Cbom, UUID> {
...
+    @Modifying(clearAutomatically = true, flushAutomatically = true)
+    @Query("""
+            UPDATE Cbom c
+            SET c.assetSyncState = :inProgress,
+                c.assetSyncError = NULL,
+                c.assetSyncAttemptedAt = CURRENT_TIMESTAMP
+            WHERE c.uuid = :uuid
+              AND c.assetSyncState = :expectedState
+              AND COALESCE(c.assetSyncAttemptedAt, :neverAttempted) = :expectedAttemptedAt
+            """)
```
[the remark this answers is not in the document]
vyskocilm: Partly fixed, and the residual is deliberate.

The downgrade is real and is closed: `markInProgress` is now guarded on `NOT_YET_SYNCED`, the same guard `markFailed` carries, so a stale attempt cannot walk a `SYNCED` row back to `IN_PROGRESS`.

No claim token, because there is nothing left for one to protect. The other two transitions a stale attempt could make are already safe: `markFailed` is guarded the same way, and `markSynced` is unconditional on purpose — the run that actually ingested the assets is entitled to say so, and both runs' writes converge because every `crypto_asset` write is an idempotent upsert and the cluster lock (now keyed per CBOM) admits one of them at a time. A token would add a column and a parameter to every transition to change the loser's outcome from "writes the same row twice" to "writes nothing".

Written down on `markInProgress`: the attempt timestamp is a retry window, never an ownership token.

src/main/resources/application.yml
```diff
@@ -420,6 +420,13 @@ cbom:
     overlap: ${CBOM_SYNC_OVERLAP:60s}
     # An entry that could not be stored is retried on this many later runs, then recorded as permanently skipped.
     skipped-retry-runs: ${CBOM_SYNC_SKIPPED_RETRY_RUNS:3}
+    # Assets written per ingest transaction. Bounds how long one ingest holds the alias advisory lock and its row locks.
+    asset-batch-size: ${CBOM_SYNC_ASSET_BATCH_SIZE:100}
+    # CBOMs one run ingests from the pending work list (crashed ingests, earlier failures, documents uploaded through
+    # Core's own API), on top of the ones the same run's feed pass stored. 0 disables the pass.
+    max-ingest-documents: ${CBOM_SYNC_MAX_INGEST_DOCUMENTS:50}
```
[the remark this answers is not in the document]
vyskocilm: A kill switch landed here; the Settings move did not.

New `cbom.sync.asset-ingest-enabled` (default true, `CBOM_SYNC_ASSET_INGEST_ENABLED`) gates **both** paths — the inline ingest and the backlog pass — and is checked inside `CbomAssetIngestService.ingest` as well as by its two callers, so "nothing is written" holds for a future caller too. `max-ingest-documents: 0` never was a kill switch: it disables only the backlog pass.

Settings rather than yml is still the right end state and is not this PR's to decide — #2249 is already moving the `cbom.sync` tunables, and a new settings section is an `interfaces` release (`SettingsSection` and the DTOs live there), so it cannot land as a `core` commit. Recorded on #2249.

The Helm wiring for all four values is still owed; the IMPORTANT block in the body says so.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Fixed, exactly as proposed.

`safeReason` passes a `ValidationException`'s own message through — text the inventory's writers shaped for an operator — and keeps `CryptoAssetConstraintTranslator.describe` for everything else, so neither path touches a driver message.

You were right about the test proving nothing: `aRefusedWriteLeavesTheCbomFailedWithTextNoDriverWrote` now says in its Javadoc that it pins the fallback, and a new `aWriterRefusalKeepsTheSentenceTheWriterShaped` pins the mapping — the alias sentence survives, and "would violate a database constraint" does not appear.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
vyskocilm: Fixed. `store()`'s contract holds again.

The inline call moved into `ingestInline`, which catches `RuntimeException`, logs it and charges the entry — so `settleUnavailable`, `ingestPending` and `run.summary()` all still run, and the watermark advances.

Same treatment in the backlog pass: the `ingestPending` loop body is wrapped per entry (covering `claimed`, `ingestOnePending` and the `markFailed` calls inside it), and `settleUnreadable` is wrapped separately.

karolhrdina (approved):

PR #2257, opened 2026-09-15
Supersede earlier CBOM versions and collect what they orphan

vyskocilm:
Part of #2073 — **PR 2 of a 5-PR plan**, on top of #2256. Deliberately not `Closes`: the deletion lifecycle, the duplicate-`bom-ref` refusal with the findings channel, and the weekly reconciliation follow as separate PRs.

## What this is

#2256 made every CBOM version write its own `crypto_asset_source` links and never take any away. Every `(serialNumber, version)` keeps its own live `cbom` row, so **the revision that last named an asset goes on sourcing it for ever**: version 2 adds its links beside version 1's, and an asset version 2 no longer names stays in the inventory, sourced by a document that does not mention it.

**The newest ingested version owns the URN.** Ingesting version N+1 writes its own sources first and *then* withdraws every earlier version's links. That order is the point: a run dying in between overstates a source count — recomputed from the links, so the next run repairs it — where the other order would leave the inventory saying nothing about a document that still exists.

A revision an **ingested** later one supersedes is not extracted at all. It reports the new `SUPERSEDED` outcome, and the check runs before the document is read and before the row is claimed, so such a revision costs no HTTP read and takes no claim.

*Ingested*, not merely stored, is load-bearing. A `SYNCED` write-off is irreversible — nothing moves a row back onto a work list — and after the #2256 upgrade every pre-existing row defaults to `PENDING`, so a serial's revisions are commonly all unsynced at once. Writing the older ones off against a newer row whose document turns out to be unreadable would leave the URN contributing nothing at all, where before it contributed a stale but present inventory.

## The orphan rule, which is the decision in this PR

`CbomAssetDetachService` is a bean of its own, because the deletion lifecycle needs the same unit of work. It settles what a withdrawal leaves behind:

> An asset whose **last** source is withdrawn is **deleted**, *unless an alias points at it* — that is, unless its key is some alias's **canonical** key — in which case the row is kept with `source_count` at zero and its payload cleared.

The inventory is meant to say what the documents currently say, and a row nothing sources any more says nothing. But `crypto_asset_alias_to_canonical_key` is `ON DELETE CASCADE`, so collecting the canonical orphan blindly would take an operator's merge decision with it, silently and on a schedule nobody triggered.

The **absorbed** side is collected like any other orphan. `absorbed_key` carries no foreign key at all — the migration that declared the table says an absorbed row that no longer exists is its normal state — so nothing there was ever at risk, and retaining those rows would keep assets the documents no longer mention for no gain.

`CryptoAssetSourceWriter`'s Javadoc ruled the opposite — that every orphan is retained — and now says this.

## Where the orphan decision's SQL lives, and why

In `CryptoAssetRepository.findOrphanRows`, one native query answering both questions for a whole batch, rather than in the service against a key it was handed. The identity key is a hash over a low-entropy pre-image and `IdentityKeyExposureFenceArchTest` is what keeps the set of sources naming it from growing; the fence caught the service naming it, and it was fixed structurally rather than by widening the allowlist.

Two other things the query's shape is deliberate about:

- **The absence of a source row, not `source_count`.** The counter is bookkeeping that only `recomputeMergeFromSources` maintains; `crypto_asset_source` is the fact it is derived from. What the answer authorizes is an irreversible `DELETE` whose `crypto_asset_source_to_crypto_asset_key` cascade takes every *other* CBOM's provenance with it, so gating it on a derived counter would turn any divergence into data loss instead of a stale number.
- **One round trip per batch.** The caller runs inside the transaction holding the cluster-wide `ALIAS_DECISION_LOCK` and already pays three statements per asset to detach it; asking both questions per asset would make it five, all serialized behind a lock every ingest batch on every node also blocks on.

## Concurrency

The supersede test is re-read **under the cluster lock at every batch**, and once more before the row is marked synced. That lock is transaction-scoped, so it is released at each batch commit — and in that gap a newer revision can be stored, ingested, and withdraw everything written so far. Resuming would re-attach a superseded revision's links to assets the newer one now owns, permanently: the newer row is `SYNCED` and never looks again, and withdrawal only ever runs from the ingesting version.

When that happens the unit is abandoned **and this revision's own contribution is withdrawn** before the row is settled. Marking it synced without giving those links back would strand whatever it had already written, for the same reason.

A batch that cannot take the withdrawn CBOM's asset-sync lock abandons the rest of the unit. That is **not** all-or-nothing and the code no longer claims it is: each batch commits in its own transaction, so the batches that already committed stay committed, orphans included, and a collected orphan is not reversible. `Withdrawal.complete()` says whether the rest was reached, and the counts describe what really happened either way. Safe, because the CBOM is left owing the work and the next run redoes the whole withdrawal idempotently.

The lock is keyed on the **document**, not the operation — as #2256 left it — so a withdrawal and an unrelated document's ingest do not contend, while a withdrawal and an ingest of the same CBOM still exclude each other. `ALIAS_DECISION_LOCK` is taken before the first `crypto_asset` row lock here too: deleting an orphan deletes its aliases by cascade, so this transaction decides about aliases as well as rows.

## Known deviations

**A written-off revision cannot be told from one that contributed.** Both read `SYNCED`. `markSuperseded` deliberately leaves `assets_synced_at` alone, so the dashboard's `lastCompletedSyncAt`, the `CbomDto.assetSyncedAt` field and the `CBOM_ASSETS_SYNCED_AT` filter are never advanced by a run that ingested nothing — but `sourceCbomCount` (CBOMs with at least one source row) and `cbomStatBySyncState` (CBOMs by state) no longer reconcile in one `syncCompleteness` response, because a written-off revision is `SYNCED` and sources nothing. Distinguishing them needs a new `CbomAssetSyncState` constant, which lives in the `interfaces` contract this ticket does not own.

**A kept orphan is visible as a live asset.** `source_count = 0` with no payload is a row state nothing produced before this PR, and no read path has a `source_count > 0` predicate, so it is listed, paginated and counted in `totalAssets` like a sourced asset. Kept deliberately for now — the row exists precisely so an operator can still see and unmerge the decision holding it, and hiding it would make that alias unreachable through the API that created it — but the inventory's totals therefore count assets no document currently mentions.

## Open work recorded on #2073

- **Deleting the superseding revision strands the older one.** A CBOM naming no cryptographic assets acquires no source rows, so it is deletable today; delete it after it superseded an earlier revision and that earlier revision stays `SYNCED`, sourcing nothing, off both work lists for ever. The repair belongs to whatever removes the row — the deletion-lifecycle PR — and its candidate is to re-offer the newest surviving version of the serial.
- **Orphan collection races an unrelated concurrent ingest.** An asset sourced only by the withdrawn revision can be collected while another CBOM's ingest, mid-document, is about to report it; that ingest then re-creates it under a fresh UUID, resetting `pqc_decided_at` and breaking external references to the old one. The per-document lock keying is what makes the two paths concurrent by design. Aliases are unaffected. Deferring collection past a quiet point is the weekly reconciliation's shape.

## Verification

1709 tests → **8347 under `mvn verify`**, 0 failures, against real PostgreSQL via Testcontainers: `CryptoAssetInventoryITest`, `CbomServiceITest`, `CbomSyncITest`, `CbomSyncTaskITest`, `CryptoAssetInventoryMigrationITest`, `CryptoAssetSearchITest`, `CryptographicAssetServiceITest`, `CryptographicAssetStatisticsITest`, plus the whole `com.otilm.core.architecture.*` and `com.otilm.core.cbom.**` suites — `IdentityKeyExposureFenceArchTest`, `TransactionalBoundaryArchTest`, `NativeQuerySchemaArchTest`, `ContextSignatureGuardTest` (BASELINE unchanged — no new Spring context was bought; the shared test fixtures are values, not beans). `spotless:apply checkstyle:check` clean.

No migration, no new configuration, no `interfaces` change.

**Not done:** Sonar new-lines coverage parity has not been computed locally, so the gate's `new_coverage` figure is unverified ahead of CI.


src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -196,6 +212,9 @@ public IngestOutcome ingest(UUID cbomUuid, JsonNode document, OffsetDateTime see
                     return lockedElsewhere(cbomUuid, entryState);
                 }
             }
+            if (!withdrawSupersededVersions(cbomUuid)) {
```
karolhrdina, line 215: **The supersede check is a TOCTOU, and the per-batch lock release is what opens it: a version's links can survive a newer version's withdrawal, permanently.**

`hasLaterVersion` is read exactly once, at entry (L177). The asset-sync lock is transaction-scoped and is released at *every batch commit* — the class Javadoc states that deliberately ("a second node can take it in the gap"). For one serial number:

1. node A enters `ingest(v2)`; `hasLaterVersion(v2)` is false (only v1, v2 exist);
2. node A writes batch 1 of v2's sources, commits, releases `cbom-asset-sync:<v2>`;
3. v3 is stored; node B ingests v3, writes v3's sources, then `withdrawSupersededVersions(v3)` → `detachService.withdraw(v2)` takes `cbom-asset-sync:<v2>` **in that gap** and removes everything batch 1 wrote (deleting the orphans it left);
4. node A resumes at batch 2 and writes the remaining v2 sources, then `withdrawSupersededVersions(v2)` (which only knows about v1), then marks v2 SYNCED.

End state: v2 — a superseded revision — still sources every asset from batch 2..n, for ever. v3 is SYNCED and will never look again; `findSupersededVersionUuids` is only ever consulted by the *ingesting* version. This is precisely the "an asset stays in the inventory, sourced by a document that does not mention it" failure the PR exists to remove, and nothing here repairs it — the weekly reconciliation is a later PR.

Cheapest closure: re-read `cbomRepository.hasLaterVersion(cbomUuid)` inside `writeBatchUnderClusterLock`, under the lock that was just taken, and abandon the unit when it has become true (it is one indexed lookup per batch, and the CBOM is left owing the work exactly as `lockedElsewhere` already leaves it).

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -165,7 +174,14 @@ public IngestOutcome ingest(UUID cbomUuid, JsonNode document, OffsetDateTime see
         if (!enabled) {
             return IngestOutcome.DISABLED;
         }
-        // Captured before the claim overwrites it, so a batch that finds the lock taken can put the row back in the
+        if (cbomRepository.hasLaterVersion(cbomUuid)) {
```
karolhrdina, line 177: **"Ahead of the claim as well, so a revision nothing will parse never takes one" is not true of the backlog pass, and the check is too deep to save what it is meant to save.**

It holds for `markInProgress` on L187. It does not hold for `CbomServiceImpl.ingestPending`, which is the pass that actually works the backlog: it calls `claimed(cbom)` → `claimForIngest` (the row *is* claimed), and then `ingestOnePending` pays a full HTTP `read(serialNumber, version)` against the CBOM Repository, and only then reaches this method.

So a superseded revision still costs, every run: the claim, one document read, one of the run's `cbom.sync.max-ingest-documents` slots, and a `run.ingestReads++` (which is the input to `settleUnreadable`'s outage verdict). A repository carrying many historical revisions in `PENDING`/`FAILED` — exactly the population a backlog pass exists for — spends its whole ingest budget on documents discarded on arrival, and never gets to the entries that still owe real work.

The root-cause fix is one level up, in the work-list selection: add `AND NOT EXISTS (SELECT 1 FROM Cbom newer WHERE newer.serialNumber = c.serialNumber AND newer.version > c.version)` to `findPendingAssetIngests` / `findAssetIngestRetries` so those rows are never offered (they can be settled to SYNCED in bulk), or at minimum test `hasLaterVersion` in `ingestOnePending` *before* `read(...)`. Keeping the guard here as well is fine as the authoritative backstop.

src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
```diff
@@ -37,6 +37,37 @@ public interface CryptoAssetRepository extends SecurityFilterRepository<CryptoAs
...
+     * count.
+     *
+     * <p>
+     * Native and scalar so it reads the row the detachment just wrote rather than an entity the persistence context
+     * cached before it: {@code recomputeMergeFromSources} is a {@code @Modifying} native statement, which leaves any
+     * managed copy of the row stale.
+     */
+    @Query(value = "SELECT EXISTS (SELECT 1 FROM {h-schema}crypto_asset a WHERE a.uuid = :uuid AND a.source_count = 0)",
+            nativeQuery = true)
+    boolean isOrphaned(@Param("uuid") UUID uuid);
```
karolhrdina, line 51: **An irreversible `DELETE` is authorized from a derived counter rather than from the fact it is derived from.**

`source_count` is bookkeeping that only `recomputeMergeFromSources` maintains; `crypto_asset_source` is the truth. Gating `CbomAssetDetachService.settleOrphan` → `assetWriter.delete` on the counter means any divergence becomes data loss rather than a stale number: a count that reads low for an asset that still has a source row deletes the asset, and `crypto_asset_source_to_crypto_asset_key` is `ON DELETE CASCADE`, so the *other* CBOM's provenance goes with it — silently, on a schedule nobody triggered, which is the exact class of harm the alias carve-out below was written to avoid.

The authoritative predicate costs the same and is served by `uq_crypto_asset_source`, which leads with `asset_uuid`:

```sql
SELECT NOT EXISTS (SELECT 1 FROM {h-schema}crypto_asset_source s WHERE s.asset_uuid = :uuid)
```

(One behavioural difference to keep in mind if you take it: for a uuid with no `crypto_asset` row this answers `true` where the current form answers `false`. `settleOrphan` then runs `isNamedByAnAlias` (false) and `delete` (0 rows), so it stays harmless — but it is worth a line in the Javadoc.)

The stated reason for native+scalar (reading past a stale managed copy after a `@Modifying` recompute) is satisfied either way.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
```diff
@@ -0,0 +1,145 @@
...
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    public Optional<Withdrawal> withdraw(UUID cbomUuid) {
+        final List<UUID> assets = sourceRepository.findAssetUuidsByCbomUuid(cbomUuid);
+        Withdrawal total = Withdrawal.NOTHING;
+        for (int from = 0; from < assets.size(); from += batchSize) {
+            final List<UUID> batch = assets.subList(from, Math.min(from + batchSize, assets.size()));
+            final Withdrawal done = transactionHandler
+                    .runInNewTransaction(() -> withdrawBatchUnderClusterLock(cbomUuid, batch));
+            if (done == null) {
+                return Optional.empty();
```
karolhrdina, line 94: **The documented contract of the empty return is false as soon as there is more than one batch.**

The `@return` on L82-83 promises "the CBOM still sources every asset it did before", and the class Javadoc (L39-41) promises the whole unit is abandoned rather than half-withdrawn. But each batch commits in its own transaction (L91-92), and the advisory lock is transaction-scoped — so batch 1 can commit (source rows gone, orphans **deleted**, which is not reversible) and batch 2 then lose `tryLock` and land here. The caller is told nothing happened.

The convergence argument still works (`lockedElsewhere` gives the claim back, the next run redoes the unit and the second attempt finishes the withdrawal), so this is not corruption — but two things follow that the code does not acknowledge:

- the Javadoc on L82-83 and L39-41 should say what actually holds: *no more than the caller's unit is lost, and the withdrawal may already be partly committed*; a reader taking the current wording at face value will build the deletion lifecycle on a guarantee that isn't there;
- the `Withdrawal` counts of the batches that did commit are discarded with the `Optional`, so `withdrawSupersededVersions`' log line and the run report understate what the node actually did.

`CbomAssetDetachServiceTest.aContendedClusterLockWithdrawsNothingAndSaysSo` only exercises the single-batch case, so the assertion `verify(sourceWriter, never()).detachCbom(...)` reads as a general property when it is not one. A test with `service(1)`, two assets and `tryLock` returning `true, false` would pin the real behaviour.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
```diff
@@ -0,0 +1,145 @@
...
+        DELETED,
+        KEPT,
+        NOT_ORPHANED
+    }
+
+    private OrphanOutcome settleOrphan(UUID assetUuid) {
+        if (!assetRepository.isOrphaned(assetUuid)) {
+            return OrphanOutcome.NOT_ORPHANED;
+        }
+        if (assetRepository.isNamedByAnAlias(assetUuid)) {
```
karolhrdina, line 136: **The absorbed side of the orphan carve-out rests on a cascade that does not exist.**

The rule's stated reason — repeated on L28-31, L106-107, in `CryptoAssetSourceWriter`'s new Javadoc and in the PR description — is that "`crypto_asset_alias` cascades from the asset, so collecting the orphan blindly would take an operator's merge decision with it". That is true for the **canonical** side only. In `V202608271000__crypto_asset_inventory.sql` the alias table has exactly one FK:

```sql
-- The key of the row that was absorbed. No FK: by the time the alias exists that row is gone...
"absorbed_key"  TEXT NOT NULL,
...
CONSTRAINT "crypto_asset_alias_to_canonical_key" FOREIGN KEY ("canonical_key")
    REFERENCES "crypto_asset" ("identity_key") ON DELETE CASCADE
```

Deleting an orphan whose key is only some alias's `absorbed_key` cannot destroy that alias — and per the migration's own comment, an absent absorbed row is the alias table's *normal* state. So the `al.absorbed_key = a.identity_key` arm of `isNamedByAnAlias` retains rows against the ticket's acceptance criterion ("the inventory says what the documents currently say") for a risk that isn't present on that side.

Either drop the absorbed arm and keep only the canonical one — which is what the cascade argument actually supports — or keep both and change the justification everywhere it is stated, because the current text will mislead the deletion-lifecycle PR that reuses this unit of work. `CryptoAssetInventoryITest.anOrphanAnAliasNamesIsKeptWithNothingLeftToAttribute` currently pins the behaviour *and* the wrong reason (`describedAs("both sides of the merge are named by the decision, so neither is collected")`).

src/test/java/com/otilm/core/integration/cbom/CryptoAssetInventoryITest.java
```diff
@@ -877,6 +890,107 @@ void aDeletedCbomIsTombstonedOnceAndByItsOwnUuid() {
...
+                .describedAs("the superseded version sources nothing")
+                .isEmpty();
+        assertThat(assetRepository.findAll())
+                .describedAs("the asset the new version stopped naming is collected, the one it still names is not")
+                .singleElement()
+                .satisfies(surviving -> {
+                    assertThat(surviving.getName()).isEqualTo("aes-256");
+                    assertThat(surviving.getSourceCount()).isEqualTo(1);
+                });
+        assertThat(cbom(first.getUuid()).getAssetSyncState()).isEqualTo(CbomAssetSyncState.SYNCED);
```
karolhrdina, line 972: **This pins a second, undocumented instance of the "known deviation" — and this one is not the SUPERSEDED branch.**

The PR body owns one deviation: a `SUPERSEDED` outcome stamps `assets_synced_at` for a document that was never read. But this assertion pins a different row in the same wrong state: `first` was genuinely ingested, then had **every one of its links withdrawn**, and it is left reading `SYNCED` with `assets_synced_at` still set. `asset_sync_state` is on the wire (`CbomDto.assetSyncState`), and for this row it now means "its assets are in the inventory" about a document that sources nothing at all.

The practical consequence is not cosmetic: `findPendingAssetIngests` and `findAssetIngestRetries` select on that column, so a `SYNCED` superseded row is never offered again. That is what makes the hole in the next comment (deleting the newest revision) unrecoverable.

Worth at least a `describedAs` saying what SYNCED means for a withdrawn revision, and a line in the PR's "Known deviation" section covering it — the `interfaces` argument for not adding a state applies to this case too, but the case itself is currently undisclosed.

src/main/java/com/otilm/core/dao/repository/CbomRepository.java
```diff
@@ -29,6 +29,35 @@ public interface CbomRepository extends SecurityFilterRepository<Cbom, UUID> {
...
+     * again.
+     */
+    @Query("""
+            SELECT COUNT(newer) > 0
+            FROM Cbom self
+            JOIN Cbom newer
+                ON newer.serialNumber = self.serialNumber
+            WHERE self.uuid = :uuid AND newer.version > self.version
+            """)
+    boolean hasLaterVersion(@Param("uuid") UUID uuid);
```
karolhrdina, line 59: **`hasLaterVersion` is evaluated against rows that can disappear, and the SYNCED-superseded row is never re-offered — so the inventory can lose a serial number permanently.**

A CBOM with no cryptographic components acquires no `crypto_asset_source` rows, so `crypto_asset_source_to_cbom_key` (RESTRICT) does not refuse its deletion — it is deletable through the API today. Sequence:

1. v1 names assets, is ingested, SYNCED;
2. v2 names none; its ingest withdraws v1's links and collects the orphans (correct — the newest revision says nothing);
3. an operator deletes v2. It has no sources, so the delete succeeds and a `cbom_tombstone` stops the sync re-storing it.

Now the newest surviving revision of the serial is v1, which *does* name assets — but v1 reads `SYNCED`, so neither `findPendingAssetIngests` nor `findAssetIngestRetries` will ever offer it again, and nothing else recomputes this. The serial contributes nothing to the inventory, for ever, with no operator-visible signal.

The supersede decision is derived state with no record that it was taken; anything that removes the superseding row has to undo it. Either the withdrawal should put the superseded rows back to `PENDING` on deletion of their superseder, or the deletion path must re-offer the newest surviving version of the serial. Worth pinning as an explicit follow-up on the deletion-lifecycle PR rather than leaving it implied.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    private enum OrphanOutcome {
+        DELETED,
+        KEPT,
+        NOT_ORPHANED
+    }
+
+    private OrphanOutcome settleOrphan(UUID assetUuid) {
+        if (!assetRepository.isOrphaned(assetUuid)) {
```
karolhrdina, line 133: **Two extra scalar round trips per asset, inside the transaction that holds the cluster-wide `ALIAS_DECISION_LOCK`.**

Per asset this loop already costs three statements in `detachCbom` (`lockForSourceChange`, the delete, `recomputeMergeFromSources`); `settleOrphan` adds `isOrphaned` and, for every orphan, `isNamedByAnAlias` — five round trips per asset, all serialized behind the alias advisory lock, which is a *global* lock every ingest batch on every node also blocks on. On a document with a few thousand assets that is the whole cluster's ingest throughput.

Both probes are set-shaped and the batch is already in hand. One query per batch answers them together:

```sql
SELECT a.uuid,
       EXISTS (SELECT 1 FROM {h-schema}crypto_asset_alias al
               WHERE al.absorbed_key = a.identity_key OR al.canonical_key = a.identity_key) AS aliased
FROM {h-schema}crypto_asset a
WHERE a.uuid IN (:uuids) AND a.source_count = 0
```

run once after the batch's detaches, which reduces the pair to one round trip per batch and leaves `assetWriter.delete` the only per-asset call on the orphan path. It keeps the identity key inside the repository, which is the fence constraint the current shape was chosen for.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -208,6 +227,34 @@ public IngestOutcome ingest(UUID cbomUuid, JsonNode document, OffsetDateTime see
...
+     * <p>
+     * After the new version's own sources are written, not before. A run that dies in between leaves an asset sourced
+     * by two revisions of one document -- a count too high, which the next run corrects -- where the other order would
+     * leave the inventory saying nothing about a document that still exists.
+     *
+     * @return false when another node holds the cluster lock, which leaves the CBOM owing the whole unit: it is not
+     * marked synced, and the next run redoes it idempotently
+     */
+    private boolean withdrawSupersededVersions(UUID cbomUuid) {
+        for (UUID superseded : cbomRepository.findSupersededVersionUuids(cbomUuid)) {
```
karolhrdina, line 243: **One `findAssetUuidsByCbomUuid` per historical revision, on every ingest, for ever.**

`findSupersededVersionUuids` returns *all* earlier versions of the serial, and `withdraw` opens each one with its own `SELECT DISTINCT s.asset_uuid ...`. For a serial on its 40th revision that is 40 queries per ingest, 39 of which are guaranteed to return nothing because those revisions were withdrawn the first time they were superseded — and it repeats on every re-ingest and every retry.

Narrow the selection to revisions that still have something to withdraw, which is one indexed check (`idx_crypto_asset_source_cbom` leads with `cbom_uuid`):

```java
WHERE self.uuid = :uuid AND older.version < self.version
  AND EXISTS (SELECT 1 FROM CryptoAssetSource s WHERE s.cbomUuid = older.uuid)
```

Same result, and the steady state (nothing left to withdraw) costs one query instead of N.

src/test/java/com/otilm/core/integration/cbom/CryptoAssetInventoryITest.java
```diff
@@ -927,4 +1041,25 @@ private Cbom cbom(String serialNumber, int version) {
...
+    private static JsonNode twoAlgorithms() {
+        return read("{\"components\":[" + component("AES-256") + "," + component("RSA-2048") + "]}");
+    }
+
+    private static String component(String name) {
+        return "{\"type\":\"cryptographic-asset\",\"name\":\"" + name + "\",\"cryptoProperties\":"
+                + "{\"assetType\":\"algorithm\",\"algorithmProperties\":{}}}";
+    }
+
+    private static JsonNode read(String json) {
```
karolhrdina, line 1058: **These fixtures already exist, three lines from where the new tests call them.**

`oneAlgorithm()`, `twoAlgorithms()`, the `component(...)` builder and this `read(...)` are a near-copy of `CbomAssetIngestServiceTest`'s L397-410 — same names, same shape, a different component spelling, and a fresh `new ObjectMapper()` on every call where the sibling holds a `private static final ObjectMapper MAPPER`. `CbomAssetDetachServiceTest.service(...)` likewise re-derives the `CbomSyncProperties` seven-arg fixture that `CbomAssetIngestServiceTest.properties(int)` already owns.

Three copies of the same CycloneDX stub drift independently, and each one is a place where a future `CbomAssetExtractor` change has to be found. A small package-private helper in `com.otilm.core.cbom.ingest` (or `..cbom.asset`) holding `algorithmDocument(String... names)` and the properties fixture would serve all three, and would not buy a new Spring context.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -139,7 +142,13 @@ public enum IngestOutcome {
          * {@code cbom.sync.asset-ingest-enabled} is off. Nothing was read and nothing was written, and the CBOM is left
          * exactly as it was found -- so turning the switch back on resumes rather than repairs.
          */
-        DISABLED
+        DISABLED,
+        /**
+         * A later version of the same serial number is already stored, so this document no longer speaks for its URN.
+         * Nothing was written and the CBOM reads {@code SYNCED}: it owes no ingest, because the version that supersedes
+         * it owns the inventory.
```
karolhrdina, line 149: **"Nothing was written and the CBOM reads `SYNCED`: it owes no ingest, because the version that supersedes it owns the inventory."** — that second clause is not checked anywhere.

`hasLaterVersion` (`CbomRepository:52-59`) asks only whether a row with a higher `version` exists in `cbom`. It says nothing about `newer.assetSyncState`. So the older version is written off as `SYNCED` on the strength of a newer *row* that may never have contributed anything — and that write-off is irreversible: `markSynced` also stamps `assets_synced_at`, making the row indistinguishable from one that really ingested, and nothing in `src/main/java` ever resets `SYNCED` back to `PENDING`.

Note the asymmetry with the other half of this PR: the *withdrawal* is correctly driven by the newer version's **successful** ingest (`withdrawSupersededVersions` runs after the batch loop), but the *suppression* is driven by mere row existence.

**A — the URN goes blank.** After the core#2073 upgrade every pre-existing header row defaults to `PENDING` (`Cbom:76`), so v1 and v2 are commonly both `PENDING`.
1. The backlog pass offers v1 → `hasLaterVersion` true → `SUPERSEDED`, `markSynced`, contributes nothing.
2. It offers v2, whose document is permanently unusable — `extraction.documentScopeUnavailable()` (`:197`), a 404 (`CbomServiceImpl:1232`), or an extractor throw (`:192`) → `REFUSED`/`FAILED`.
3. v2 retries forever and keeps failing. The serial number contributes **zero** assets, although v1's document was perfectly ingestible, and v1 is `SYNCED` so neither `findPendingAssetIngests` (`PENDING` only) nor `findAssetIngestRetries` (`IN_PROGRESS`/`FAILED` only) will ever offer it again.

Before this PR v1 would have been ingested and the URN would be present — stale, but present. That is a regression in inventory completeness, not just a change of which revision wins. The same hole opens when the newer version's document cannot be re-read (the report's `ingestUnavailable`), or while it sits beyond the `maxIngestDocuments` budget and then fails.

**B — a partial inventory is stranded permanently.**
1. Run 1: v1's ingest writes batch 1, batch 2 throws → `fail(v1)` (`:218`) → `FAILED` with an arbitrary **subset** of its `crypto_asset_source` rows attached.
2. Run 2: v2 is stored, its inline ingest is refused → `FAILED`.
3. Run 3 (after `ingest-retry-after`): the retry list offers v1 (`INGEST_RETRY_STATES` = `{IN_PROGRESS, FAILED}`) → `hasLaterVersion` true → `SUPERSEDED` → `markSynced(v1)`.

That subset is now permanent. v1 is `SYNCED` so nothing will finish its ingest, and withdrawal only ever runs from a *successful* ingest of a later version (`:215`), which v2 will never be. The run report counts it as "superseded by a later version".

Both cases close with `AND newer.assetSyncState = SYNCED` in `hasLaterVersion`. B additionally wants the `SUPERSEDED` arm to withdraw this row's own contribution before marking it synced — as written it marks the row synced and leaves whatever it already wrote in place.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -165,7 +174,14 @@ public IngestOutcome ingest(UUID cbomUuid, JsonNode document, OffsetDateTime see
         if (!enabled) {
             return IngestOutcome.DISABLED;
         }
-        // Captured before the claim overwrites it, so a batch that finds the lock taken can put the row back in the
+        if (cbomRepository.hasLaterVersion(cbomUuid)) {
+            // Checked before the document is even parsed: a superseded revision has nothing to contribute, and
```
karolhrdina, line 178: **"Checked before the document is even parsed" is not true on either production call path, and the test that pins it uses an overload nothing calls.**

Both callers hit the `Map` overload, which materialises the whole tree first:

```java
public IngestOutcome ingest(UUID cbomUuid, Map<String, Object> document, OffsetDateTime seenAt) {
    final JsonNode tree = JSON_COLUMN.valueToTree(document);   // runs before the guard below
    return ingest(cbomUuid, tree, seenAt);
}
```

- `CbomServiceImpl:843` (inline pass) passes a `Map<String,Object>`.
- `CbomServiceImpl:1259` (backlog pass) passes a `BomResponseDto`, and `BomResponseDto extends LinkedHashMap<String, Object>` (`model/cbom/BomResponseDto.java:11`) — so it binds to the same overload.

What the guard actually avoids is the **extraction**, not the parse. `CbomAssetIngestServiceTest.aSupersededVersionIsNotEvenParsed` (`:341`) passes because it calls the `JsonNode` overload directly — a path no caller uses — so it asserts a property production does not have. Renaming it to say "not extracted", or moving the guard above `valueToTree`, would make the comment and the test agree with the code.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -165,7 +174,14 @@ public IngestOutcome ingest(UUID cbomUuid, JsonNode document, OffsetDateTime see
...
+            // ingesting it would attach the inventory to a document another row already speaks for. Ahead of the
+            // claim as well, so a revision nothing will parse never takes one.
+            runInOwnTransaction(() -> stateWriter.markSynced(cbomUuid, seenAt));
```
karolhrdina, line 181: **The "Known deviation" section understates this: `assets_synced_at` has three live consumers today, one of them the public API contract.**

1. **Dashboard statistic.** `CryptographicAssetServiceImpl.lastCompletedSyncAt` (`:329-338`) takes the max `assetsSyncedAt` over *every* CBOM row, with no state or supersede predicate, and serves it as `CryptographicAssetSyncCompletenessDto.lastCompletedSyncAt` (wired at `:283`). A run that processed only superseded rows now advances "last completed sync at" although it ingested nothing.
2. **Public API.** `Cbom:112` → `dto.setAssetSyncedAt(this.assetsSyncedAt)`, and `CbomServiceImpl:268` on the detail DTO. A superseded CBOM is served as `assetSyncState = SYNCED` with a fresh `assetSyncedAt`, while it sources zero assets.
3. **User-facing filter.** `FilterField.CBOM_ASSETS_SYNCED_AT` ("Assets Synced At", `DATETIME`), offered by `CbomServiceImpl:492`.

Related divergence introduced in the same block: `sourceCbomCount` counts CBOMs with at least one `crypto_asset_source` row (`hasContributedAssets`, `:320-327`) while `cbomStatBySyncState` counts by state (`:280-282`). After this PR those two numbers in one `syncCompleteness` response stop reconciling, because superseded rows are `SYNCED` and source nothing — and this file explicitly cares about that ("serve zeros that no longer reconcile with the list", `:301`).

The deviation itself may still be the right trade, but "nothing reads that column for a superseded row today" is not accurate, and the PR body is where a reviewer would go looking. Worth either correcting that paragraph or excluding superseded rows from `lastCompletedSyncAt`.

*(Refuted while checking, for the record: the work lists do **not** read `assets_synced_at` — `findPendingAssetIngests` selects on `assetSyncState`, `findAssetIngestRetries` on `assetSyncState` + `assetSyncAttemptedAt` — and there is no stale-resync sweep over the column anywhere, only the `idx_cbom_assets_synced_at` index.)*

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
```diff
@@ -0,0 +1,145 @@
...
+ * Withdraws one CBOM's contribution to the cryptographic asset inventory, and settles what that withdrawal leaves
+ * behind.
+ *
+ * <p>
+ * The orchestrator half of a bean pair, like {@link CbomAssetIngestService}: it owns the transaction boundaries and
+ * composes the writers, and writes nothing itself.
+ *
+ * <p>
+ * <b>The orphan rule.</b> An asset whose last source is withdrawn is deleted, unless an alias names it -- on either
+ * side of a merge -- in which case the row is kept with {@code source_count} at zero and its payload cleared. The
```
karolhrdina, line 29: **`source_count = 0` is a row state that could not exist before this PR, and nothing on the read side accounts for it.**

`detachCbom` had no production caller until now, and on the ingest path `upsertIdentity` is always followed by `upsertSource` inside the same batch transaction — so a zero-source `crypto_asset` row was never externally visible. The orphan rule now creates them deliberately and durably.

There is no `source_count > 0` predicate anywhere on the read side:
- `CryptographicAssetServiceImpl.listCryptographicAssets` (`:136-162`) applies only the security filter and the caller's own filters, so kept-for-alias rows are listed and paginated like any other asset — with `sourceCbomCount = 0`, no `merged_crypto_properties`, and a PQC verdict re-derived from a null payload (`recomputeMergeFromSources` bumps `i_upd`, which makes the row stale to `findStaleVerdictRows`).
- `getCryptographicAssetStatistics` counts them in `totalAssets`, `byType`, `byVerdict` and `byFamily`.
- `FilterField.CBOM_ASSET_SOURCE_COUNT` exposes the column to users but nothing defaults it.

That reads directly against this class's own stated rationale — "the inventory is meant to say what the documents currently say, and a row nothing sources any more says nothing". Keeping the row to protect the merge decision is the right call; leaving it in the list and in the statistics as a live asset is the part that has no answer yet. Either exclude `source_count = 0` from the list and the aggregates, or say in this Javadoc why these rows stay visible.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
```diff
@@ -0,0 +1,145 @@
...
+        if (!assetRepository.isOrphaned(assetUuid)) {
+            return OrphanOutcome.NOT_ORPHANED;
+        }
+        if (assetRepository.isNamedByAnAlias(assetUuid)) {
+            log
+                    .debug("CBOM asset withdrawal: asset {} has no source left but an alias names it; keeping the row",
+                            assetUuid);
+            return OrphanOutcome.KEPT;
+        }
+        assetWriter.delete(assetUuid);
```
karolhrdina, line 142: **Orphan collection races an unrelated concurrent ingest, and the asset comes back under a new UUID.**

`isOrphaned` is a correct *committed-state* test — `detachCbom` joins this batch transaction and `recomputeMergeFromSources` sets `source_count` from a live `count(*)` before the read. But "committed" excludes another CBOM's ingest that is mid-document, because ingest also commits per batch.

Asset A is sourced only by v1. Node C starts ingesting an unrelated CBOM `Z` that also reports A, and A falls in Z's batch 5. While Z is on batch 1, node B ingests v2, and this line deletes A: `source_count` is 0 and no alias names it. Z's batch 5 then re-inserts A through `upsertIdentity`, which mints a fresh `UUID.randomUUID()` (`CryptoAssetWriter:90`).

The asset UUID is API-visible — it is the identity of `CryptoAssetListRow` and of the detail endpoint — so external references to it break. `i_cre`, `pqc_verdict`, `pqc_rule_id` and `pqc_decided_at` are reset with it, and `pqc_decided_at` is documented as "dat[ing] the finding rather than the last time anybody looked" (`CryptoAssetRepository:200`), so the age of a finding is silently lost.

Aliases are correctly protected by `isNamedByAnAlias`, so that part does not occur. Either way this is new exposure: before this PR `detachCbom` had no production caller and no asset was ever deleted.

src/test/java/com/otilm/core/cbom/ingest/CbomAssetDetachServiceTest.java
```diff
@@ -0,0 +1,104 @@
...
+    @Test
+    void linksAreWithdrawnInBatchesOfTheConfiguredSize() {
+        when(sourceRepository.findAssetUuidsByCbomUuid(CBOM))
+                .thenReturn(List.of(UUID.randomUUID(), UUID.randomUUID(), UUID.randomUUID()));
+        when(synchronizer.tryLock(anyString())).thenReturn(true);
+
+        service(2).withdraw(CBOM);
+
+        // One lock acquisition per batch transaction: three links at a batch size of two is two transactions.
+        verify(synchronizer, times(2)).tryLock(CbomAssetIngestService.assetSyncLockKey(CBOM));
```
karolhrdina, line 75: **Three gaps here are mutation-escaping — I ran the mutants against the suite to check.** All four unit tests pass and were stable under `-Dsurefire.runOrder=reversealphabetical`; these are the assertions that do not bite.

**1. This test cannot catch a batching bug.** Its only assertion is the `tryLock` count. Replacing `assets.subList(from, Math.min(from + batchSize, assets.size()))` with `assets.subList(0, Math.min(batchSize, assets.size()))` — batch 1 processed twice, assets 3..N **never withdrawn** — leaves it green. It never verifies which assets each batch contained, never verifies `detachCbom` was called three times, and discards the return value. That is exactly the data-loss shape a batching test exists to catch.

Secondary: `service(int)` builds the bean with `new TransactionHandler()` (`:101`), an unproxied POJO, so `REQUIRES_NEW` is inert. "Split into multiple **transactions**" is not tested at all — only that the loop iterates twice.

**2. `Withdrawal.plus` has no coverage anywhere.** Changing it to `new Withdrawal(detached + other.detached, other.deleted, kept + other.kept)` — dropping accumulation of `deleted` — leaves all 19 unit tests green, and it escapes the ITests too: `asset-batch-size` defaults to 100 and every ITest withdraws ≤ 2 assets, so `plus` is only ever called as `NOTHING.plus(done)`, which is identity for any mutant returning `other`'s field. No test anywhere executes a multi-batch accumulation, yet the record's own Javadoc says it exists "for the run report and for the tests that pin the orphan rule".

**3. Nothing proves `isNamedByAnAlias` is asset-specific.** Dropping `WHERE a.uuid = :uuid` from the query — "any alias row anywhere keeps every orphan" — passes every ITest: `withdrawingTheLastSourceCollectsTheAssetItOrphaned` runs with an empty alias table, and `anOrphanAnAliasNamesIsKeptWithNothingLeftToAttribute` expects `kept` for both assets anyway. The missing case is an alias naming asset X plus an unrelated orphan Y that must still be **deleted** — which is the failure mode that would quietly stop the collector collecting anything at all on a tenant that uses merges.

Worth knowing alongside these: `isOrphaned` is mocked in every unit test and Mockito's default for `boolean` is `false`, so all four take the `NOT_ORPHANED` path and neither `assetWriter.delete` nor `isNamedByAnAlias` is ever invoked in a unit test. The orphan rule — this PR's headline behaviour — rests entirely on the Testcontainers ITest. (Consequence: `verify(assetWriter, never()).delete(any())` at `:63` is unfalsifiable, since `delete` is unreachable there whatever the production code does.)

What does bite, verified: the sources-then-withdraw ordering (`CbomAssetIngestServiceTest:268` `InOrder`) fails when the withdrawal is moved above the write loop; `anAssetThatKeepsASourceIsLeftAlone` fails when `!isOrphaned` is inverted; and `kept = 2` in the alias ITest can only hold if both arms of the `OR` match, so both sides of a merge are genuinely pinned.

karolhrdina (commented):
Reviewed with five independent passes (concurrency/transactions, persistence/schema, sync-state lifecycle, test quality with mutation runs, and a general correctness sweep). Everything below was verified against the code; hypotheses that did not survive are listed at the end so they don't get re-raised.

The design reasoning in this PR is unusually careful and most of it holds up — the ordering (sources first, then withdraw), the lock ranking, the decision to put the orphan SQL behind the identity-key fence, and the choice to keep an aliased orphan are all correct and correctly argued. Six things below are not yet right.

## Blocking

**1. `SUPERSEDED` is decided on a row's existence, not on that row having ingested** (`CbomAssetIngestService:147-151, 177`). Two deterministic single-node outcomes: a URN that goes blank (a regression against pre-PR behaviour), and a partial inventory stranded permanently as `SYNCED`. This is the one I'd fix first — it needs no concurrency and no cluster. Details inline. Note the asymmetry it creates: withdrawal is gated on the newer version's *successful* ingest, suppression is gated on mere existence.

**2. The withdrawal is not fenced against an in-flight ingest of the version being withdrawn.** Already raised by Copilot and by the earlier review pass — confirming it independently, and adding two points neither covered:
- It does not need the older version to be multi-batch. The window between `hasLaterVersion` (`:177`) and the first batch write also spans `markInProgress` and the whole `extractor.extract(document)` call, which is seconds on a large document.
- When `findAssetUuidsByCbomUuid` comes back **empty**, the loop body never runs, so `withdraw` takes **no lock at all** — there is not even batch-level exclusion between a starting ingest of v1 and a withdrawal of v1.

A fix that keeps the current ranking intact (both acquisitions stay non-blocking `tryLock`, so the no-cycle argument still holds): re-read `hasLaterVersion(cbomUuid)` *inside* `writeBatchUnderClusterLock` after the `tryLock` succeeds and abandon the unit if it has become true, and re-read the asset list per batch under that batch's own lock rather than snapshotting once outside every transaction.

**3. Orphan collection races an unrelated concurrent ingest**, and the asset returns under a new API-visible UUID. Details inline.

## Worth settling before merge

**4.** `source_count = 0` is a row state that could not exist before this PR, and no read path accounts for it.
**5.** The "Known deviation" paragraph is not accurate — `assets_synced_at` has three live consumers, one of them the public `CbomDto`.
**6.** Three test gaps that escape mutation, including one that would hide a batching bug that silently drops assets.

## Also: `aSupersededVersionIsNotEvenParsed` asserts something production does not do

`BomResponseDto extends LinkedHashMap<String, Object>`, so both production callers bind to the `Map` overload and `valueToTree` runs before the guard. Inline.

## Checked and clean — please don't let these get re-raised

- **Security filtering on the `JOIN Cbom older` alias.** `SecurityFilterRepository` is an explicit Criteria-API surface, not a Hibernate filter; there is no `@FilterDef`/`@Filter` in `src/main/java`, no soft delete, and no tenant column on `cbom`. Both aliases see the same rows.
- **HQL validity.** The unmapped entity join has precedent in this very file (`findVersionsByUuid:20-28`), and `SELECT COUNT(x) > 0` is admitted by Hibernate 6.6's `selectExpression()` grammar (alternative 4, `expressionOrPredicate()`).
- **NULL / tied `serial_number`+`version`.** `NOT NULL` on both, plus `cbom_serial_version_unique UNIQUE (serial_number, version)`, so `<`/`>` is a strict total order within a serial and `self` excludes itself.
- **Stale `source_count` at `isOrphaned`.** No `CryptoAsset`/`CryptoAssetSource` entity is ever loaded on the detach path, so there is nothing to go stale and no dirty flush to revert the recompute. The missing `flushAutomatically`/`clearAutomatically` on `deleteForAssetAndCbom` and `deleteAsset` is harmless here.
- **FK RESTRICT on `assetWriter.delete`.** `crypto_asset_source_to_cbom_key` RESTRICT points at `cbom`, not at `crypto_asset`; the asset side is `ON DELETE CASCADE`, and `properties_source_uuid` is `ON DELETE SET NULL`. There are exactly two FKs referencing `crypto_asset` in the whole migration set and no PQC side tables.
- **Lock-order deadlock.** Asset-sync keys are only ever taken with the non-blocking `tryLock` (the two acquisitions are `CbomAssetIngestService:265` and `CbomAssetDetachService:103`), and nothing holding `ALIAS_DECISION_LOCK` ever reaches for one, so no advisory cycle can close. Row-lock order is consistent too: `findAssetUuidsByCbomUuid` is `ORDER BY s.assetUuid`, matching the ingest and the PQC sweep.
- **Arch tests.** `NativeQuerySchemaArchTest`, `IdentityKeyExposureFenceArchTest`, `TransactionalBoundaryArchTest` all pass, and the fence genuinely permits `isNamedByAnAlias` — `CryptoAssetRepository.java` is allowlisted for the `identity_key|absorbed_key|canonical_key` vocabulary, and the key never leaves the repository. Moving that join into the repository rather than widening the allowlist was the right call.
- **Run report.** `%d` count and argument order both line up in the new summary string.
- **Ordering.** Every single-node processing order I traced ends in a correct state; three-version chains are handled because `findSupersededVersionUuids` returns all older versions. The wrong-state cases are (1) and (2), not plain ordering.
- **New batching loop** cannot spin: `CbomSyncProperties` rejects `assetBatchSize < 1` at binding.
- **Deletion path.** The Javadoc is still accurate, and the supersede path makes API deletion strictly better, never worse — a withdrawn earlier version ends with zero source rows and becomes deletable. A kept-for-alias asset blocks nothing.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -165,7 +174,14 @@ public IngestOutcome ingest(UUID cbomUuid, JsonNode document, OffsetDateTime see
...
+            return IngestOutcome.SUPERSEDED;
```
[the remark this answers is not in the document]
vyskocilm: Right, and fixed. `CbomAssetIngestService.settleWithoutReading(uuid)` is now public and `CbomServiceImpl.ingestOnePending` calls it **before** `read(...)`. A superseded revision whose old document 404s is settled as SUPERSEDED, not FAILED, and an unavailable read never comes into it. The guard inside `ingest` stays as the backstop for inline callers. Same ground as karolhrdina's thread on :177.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
```diff
@@ -0,0 +1,145 @@
...
+            return new Withdrawal(detached + other.detached, deleted + other.deleted, kept + other.kept);
+        }
+    }
+
+    /**
+     * Withdraws every link the given CBOM contributed, applying the orphan rule to each asset it leaves without a
+     * source.
+     *
+     * @return empty when another node holds the cluster lock and the work was left for it -- the CBOM still sources
+     * every asset it did before, and the caller must not report the withdrawal as done
```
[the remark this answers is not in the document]
vyskocilm: Correct, and the Javadoc said something false. `Withdrawal` now carries `complete`, `withdraw` returns it instead of `Optional`, and the counts of the batches that did commit survive the abandonment. Class comment and `@return` both say what actually holds: no more than the caller's own unit is lost, the withdrawal may already be partly committed, and a collected orphan is not reversible. Pinned by `anAbandonedUnitStillReportsWhatTheCommittedBatchesDid`.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
```diff
@@ -0,0 +1,145 @@
...
+     */
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    public Optional<Withdrawal> withdraw(UUID cbomUuid) {
+        final List<UUID> assets = sourceRepository.findAssetUuidsByCbomUuid(cbomUuid);
+        Withdrawal total = Withdrawal.NOTHING;
+        for (int from = 0; from < assets.size(); from += batchSize) {
+            final List<UUID> batch = assets.subList(from, Math.min(from + batchSize, assets.size()));
+            final Withdrawal done = transactionHandler
+                    .runInNewTransaction(() -> withdrawBatchUnderClusterLock(cbomUuid, batch));
```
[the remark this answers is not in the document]
vyskocilm: Real, and closed from the other side. `hasIngestedLaterVersion` is now re-read inside `writeBatchUnderClusterLock` (under the lock just taken) and once more after the batch loop before the row is marked synced. A late source written by the older ingest is therefore withdrawn by that ingest itself: on seeing the newer revision it abandons the rest and calls `supersede`, which withdraws its own contribution and settles the row. So the snapshot no longer has to be complete for the end state to be.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Confirmed and taken as written. `hasIngestedLaterVersion(cbomUuid)` is re-read inside `writeBatchUnderClusterLock`, under the lock, and a true answer abandons the unit through a new `BatchOutcome.SUPERSEDED`.

One addition on top of your closure: abandoning is not enough on its own, because batch 1..k-1 are already committed and the newer revision's withdrawal has been and gone. So the abandon path goes through `supersede(uuid, entryState)`, which **withdraws this revision's own links** and only then settles the row. Same handler the entry check uses, for the same reason. Also re-checked after the loop, for the version that was on its last batch when the newer one landed — nothing else would have looked again.

Pinned by `aVersionOvertakenBetweenBatchesAbandonsTheRestAndGivesBackWhatItWrote`.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Right on both counts, and the comment was wrong. Took the `ingestOnePending` option rather than the work-list predicate: `settleWithoutReading(uuid)` is now public and called **before** `read(...)`, so a superseded revision costs no HTTP read, no `ingestReads++`, and it leaves the list for good on the first encounter instead of being re-offered.

Why not the `NOT EXISTS` in `findPendingAssetIngests` / `findAssetIngestRetries`: rows filtered out of both selectors are never settled by anything, so they sit at `PENDING` for ever and `cbomStatBySyncState` reports a backlog that no run will ever work down. The bulk settle you mention would fix that, but it is a second write path over the same rows with no claim on them. One cheap pass that settles each row once buys the same steady state.

The guard in `ingest` stays as the authoritative backstop, as you suggest. Pinned by `aSupersededRevisionIsSettledByTheBacklogPassWithoutADocumentRead`, which asserts zero document reads.

src/main/java/com/otilm/core/dao/repository/cbom/CryptoAssetRepository.java
vyskocilm: Agreed — authorizing an irreversible cascade from a derived counter is the wrong predicate. Replaced with the source-row test, and folded into the batch query below, so it now reads:

```sql
WHERE a.uuid IN (:uuids)
  AND NOT EXISTS (SELECT 1 FROM {h-schema}crypto_asset_source s WHERE s.asset_uuid = a.uuid)
```

The missing-row difference you flag is now moot in the new shape: a uuid with no `crypto_asset` row is simply not returned, which is the same answer as "nothing to collect". Said so in the Javadoc, along with the reason the counter is not the predicate.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
vyskocilm: Both consequences taken.

- `Withdrawal` gained `complete`, `withdraw` returns it rather than `Optional`, and `NOTHING.plus(...)` carries the committed batches' counts through an abandonment. `withdrawSupersededVersions` logs what was done before it returns false.
- The `@return` and the class comment now say what holds: the batches that committed stay committed, a collected orphan is not reversible, and what the caller is promised is only that no more than its own unit is lost.

`aContendedClusterLockWithdrawsNothingAndSaysSo` keeps the single-batch case, and `anAbandonedUnitStillReportsWhatTheCommittedBatchesDid` adds the two-batch one you describe — `service(1)`, two assets, `tryLock` returning `true, false`, asserting `Withdrawal(1, 1, 0, false)`.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
vyskocilm: Checked against the migration, and you are right — `crypto_asset_alias` has exactly one FK and it is on `canonical_key`. The cascade argument supports the canonical arm only.

Dropped the absorbed arm rather than rewriting the justification: retaining a row the documents no longer mention, to protect a decision nothing could have destroyed, loses on the acceptance criterion for nothing. The predicate is now `al.canonical_key = a.identity_key` alone, and the reason is restated in the repository Javadoc, the class comment, `CryptoAssetSourceWriter.detachCbom` and the PR description.

The test is renamed `anOrphanAnAliasPointsAtIsKeptAndTheAbsorbedSideIsCollected` and now asserts `Withdrawal(2, 1, 1, true)` with the absorbed row gone and the alias still standing — it pins the new rule and its real reason.

src/test/java/com/otilm/core/integration/cbom/CryptoAssetInventoryITest.java
vyskocilm: Fair — that row was a second, undisclosed instance and it was not the SUPERSEDED branch.

Half of it is now actually fixed rather than disclosed: `markSuperseded` writes `SYNCED` with a **null** `syncedAt`, and `updateAssetSyncState` coalesces, so no superseded or withdrawn-then-settled revision gets an `assets_synced_at` it did not earn. That takes `lastCompletedSyncAt`, the DTO field and the filter out of it (see the thread on :181).

What remains is the state itself: `SYNCED` cannot distinguish "contributed" from "written off", and both now source nothing. Added the `describedAs` you asked for, said so in `markSuperseded`'s Javadoc, and the PR body's deviation section now covers this case too. Telling them apart needs a new `CbomAssetSyncState` constant, which lives in `interfaces` — recorded as open work on core#2073.

src/main/java/com/otilm/core/dao/repository/CbomRepository.java
vyskocilm: Confirmed, and it survives the other fix in this round: `hasIngestedLaterVersion` now requires the newer row to be `SYNCED`, which stops the write-off happening against a revision that never contributed — but it does nothing once that revision has contributed and is then deleted. Your sequence still ends with v1 `SYNCED`, sourcing nothing, and off both work lists for ever.

Not fixed here: the repair belongs to whatever removes the superseding row, which is the deletion lifecycle. Recorded as open work on core#2073 and carried explicitly in the deletion-lifecycle PR's description — the candidate there is to re-offer the newest surviving version of the serial when a CBOM is deleted, rather than to have the withdrawal try to anticipate it.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
vyskocilm: Taken. Both probes are now one set-shaped query per batch, run after the batch's detaches:

```sql
SELECT a.uuid,
       EXISTS (SELECT 1 FROM {h-schema}crypto_asset_alias al
               WHERE al.canonical_key = a.identity_key) AS aliased
FROM {h-schema}crypto_asset a
WHERE a.uuid IN (:uuids)
  AND NOT EXISTS (SELECT 1 FROM {h-schema}crypto_asset_source s WHERE s.asset_uuid = a.uuid)
```

`assetWriter.delete` is the only per-asset call left on the orphan path. The identity key stays inside the repository, as you note. `isOrphaned` and `isNamedByAnAlias` are gone; the alias arm is canonical-only per your thread on :136, and the source test replaces `source_count` per :51.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Taken as written:

```java
WHERE self.uuid = :uuid AND older.version < self.version
  AND EXISTS (SELECT 1 FROM CryptoAssetSource s WHERE s.cbomUuid = older.uuid)
```

Javadoc says why, including that the steady state is one query rather than N and that the probe is served by `idx_crypto_asset_source_cbom`.

src/test/java/com/otilm/core/integration/cbom/CryptoAssetInventoryITest.java
vyskocilm: Taken. New `CbomIngestTestFixtures` in `com.otilm.core.cbom.ingest` (test sources) holds `algorithmDocument(String... names)`, `algorithmSeenAt`, the shared `ObjectMapper` and the `CbomSyncProperties` fixture in both its enabled and disabled forms. All three classes now go through it, and `CryptoAssetInventoryITest`'s own `component`/`read` and `CbomAssetIngestServiceTest`'s `properties`/`algorithm`/`read` are deleted. Values only — no Spring, so no new context.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Both cases confirmed, and both closed.

`hasLaterVersion` is now `hasIngestedLaterVersion`, with `AND newer.assetSyncState = :ingested` bound to `SYNCED`. Renamed at every call site so the new meaning is visible where it is used, and the Javadoc carries your reasoning — the write-off is irreversible, so it has to be earned by the newer revision contributing rather than by its header row existing, and after the upgrade every pre-existing row is `PENDING`. Pinned by `aLaterVersionThatWasNeverIngestedDoesNotSupersedeTheOlderOne` (ITest) and `aLaterVersionThatHasNotItselfBeenIngestedSupersedesNothing` (unit).

For B, the second half you ask for: the `SUPERSEDED` arm now runs `supersede(uuid, entryState)`, which calls `detachService.withdraw(uuid)` on **this** row first and marks it synced only when that withdrawal is complete — an incomplete one leaves the row owing the unit. So a partial contribution is given back rather than stranded. Pinned by `aSupersededVersionWithdrawsWhatItHadAlreadyContributed` and `aSupersededVersionWhoseOwnWithdrawalIsContendedIsNotWrittenOff`.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Right — both callers bind to the `Map` overload, `BomResponseDto extends LinkedHashMap` included, so the tree was always built first.

Moved the guard rather than renaming around it. The kill switch and the supersede check are now `settleWithoutReading(uuid)`, called at the top of **both** overloads, ahead of `valueToTree`. The test is renamed `aSupersededVersionIsNotExtracted` and still asserts `extractor` is never called; what it now also asserts is true of the path production actually takes.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Thank you for chasing the consumers — and for the refutation, which I have not re-litigated.

Fixed rather than re-documented: `markSuperseded` writes `SYNCED` with a null `syncedAt`, which `updateAssetSyncState` coalesces, so the column keeps whatever the row honestly earned (nothing, usually). That takes all three consumers out of it — `lastCompletedSyncAt` cannot be advanced by a run that ingested nothing, `CbomDto.assetSyncedAt` is not served a fresh time for a document sourcing zero assets, and the filter matches only rows that really synced. The writer's Javadoc names all three, so the next person to reach for `markSynced` here sees why there are two methods.

The `sourceCbomCount` / `cbomStatBySyncState` divergence you note is real and stays: a written-off revision is `SYNCED` and sources nothing, so the two figures in one `syncCompleteness` response no longer reconcile. That needs the state constant (`interfaces`) — recorded as open work on core#2073, and the PR body now says so instead of claiming nothing reads the column.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
vyskocilm: Accurate, and it is a row state nothing produced before this class.

Taken the documentation option rather than the predicate, deliberately: the row is kept precisely so an operator can still see and unmerge the decision holding it, and a `source_count > 0` filter on the list would make that alias unreachable through the API that created it. Hiding it from `listCryptographicAssets` while counting it in `totalAssets` — or the reverse — is worse than either.

So the class Javadoc now states it outright: a kept orphan is listed, paginated and counted like a sourced asset, the inventory's totals therefore count assets no document currently mentions, and that is unresolved. Recorded as open work on core#2073. If you would rather it were excluded from the aggregates only, say so and I will do it here — it is a small change and I would rather take your call than guess.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
vyskocilm: Confirmed, and not closed by anything in this round. The per-CBOM key is what makes it possible: the withdrawal and the unrelated ingest hold different keys by design, so nothing serializes them, and `isOrphaned` reading committed state is exactly right and exactly insufficient. The uuid churn, the reset `pqc_decided_at` and the lost finding age all follow.

Swapping the counter for the source-row test (your :51 thread) does not change it — both read committed state. Closing it properly means the orphan decision taking something the concurrent ingest also takes, which is a lock the per-document keying was chosen to avoid, or deferring collection past a quiet point, which is the weekly reconciliation's shape.

Recorded as open work on core#2073 with your sequence written out, so the reconciliation PR meets it as a stated requirement rather than rediscovering it. Aliases being safe is noted there too.

src/test/java/com/otilm/core/cbom/ingest/CbomAssetDetachServiceTest.java
vyskocilm: Thank you for running the mutants — all three were live and all three are closed.

1. `linksAreWithdrawnInBatchesOfTheConfiguredSize` is replaced by `eachBatchWithdrawsItsOwnSliceOfTheLinks`, which verifies `detachCbom` per asset and `times(3)` overall, and captures the argument to the new batch query to assert the slices are `[first, second]` then `[third]`. Your `subList(0, ...)` mutant fails it on both counts. The `new TransactionHandler()` point stands — this class tests the loop, not the propagation; the transaction boundary is the ITest's.
2. `theCountsOfEveryBatchAreAddedUp` runs `service(1)` over two assets with one deleted and one kept, asserting `Withdrawal(2, 1, 1, true)`. Your `plus` mutant fails it. `anAbandonedUnitStillReportsWhatTheCommittedBatchesDid` covers the accumulation-then-abandon case.
3. The asset-specific question moved into the batch query, so the analogous mutant is dropping `a.uuid IN (:uuids)`. `anUnrelatedAliasDoesNotSaveAnOrphanFromCollection` is the missing case you describe — an alias naming X, an unrelated orphan Y, asserting Y is deleted and X kept.

Also addressed the `isOrphaned` default-`false` observation: the orphan branches are no longer unreachable in unit tests. `anOrphanNoAliasPointsAtIsCollected` and `anOrphanAnAliasPointsAtIsKept` stub the batch query and verify `assetWriter.delete`, so `verify(..., never()).delete(...)` elsewhere is falsifiable now.

karolhrdina (approved):

PR #2259, opened 2026-09-15
Treat an absent value as not met instead of failing the condition

### LINKED ISSUE #2234: Workflows - Conditions - Comment: date comparison on the empty Resolved At field fails with an internal error
### Description
A condition on the **Comment** resource that compares the **Resolved At** field with a date fails with an internal error whenever the field has no value. On the **Comment created** event `Resolved At` is always empty — the thread has not been resolved yet — so every date comparison on this field hits the failure.

The evaluation reason shown to the operator is `Condition is not set properly: Cannot invoke "java.util.Date.toInstant()" because "o" is null`. Two problems in one:

1. A null value in a nullable date field should simply not match the condition. Instead the evaluator dereferences it and fails.
2. The message states that the condition is set incorrectly, which is wrong — the condition is valid and was accepted when saved. It also exposes an internal Java error string to the operator, who has no way to act on it.

Both `greater than` and `in next` were tested and behave the same way, so this is not specific to one operator. `Resolved At` is null for every comment until its thread is resolved, and any other nullable date field evaluated through the same path would be affected as well.

### Steps to Reproduce
1. Go to **Workflows → Conditions** and create a condition on the **Comment** resource: field **Resolved At**, operator **greater than**, any date. Save it — it saves without complaint.
2. Create a rule containing that condition, an action, and a trigger bound to the **Comment created** event.
3. Post a comment on any commentable object.
4. Open the object's **Event History**, find the **Comment created** entry and open **Evaluation details**.

### Expected Behavior
- `Resolved At` has no value on a newly created comment, so the condition does not match: result **Not met**, with no error.
- If a condition genuinely cannot be evaluated, the reason names the cause in terms the operator can act on, without internal error text.

### Actual Behavior
- Result is **Not met** with the reason `Condition is not set properly: Cannot invoke "java.util.Date.toInstant()" because "o" is null`.
- The reason misdirects the operator to a condition that is in fact correct.
- The same happens with the `in next` operator (tested with `P7D`).

### Severity
Minor

A workaround exists — `is empty` and `is not empty` work on this field — and there is no impact on data or access. The main cost is the misleading message, which sends the operator to fix a condition that is not broken.

### Module
Workflows

### Attachments

Image

### Logs

_No response_

### Environment
Commit: 7a414fc
Branch: main
Client:
- macOS Tahoe 26.5.1

klaraf755:
Closes #2234

A condition comparing a date field that has no value failed inside the operator and was reported as `Condition is not set properly: Cannot invoke "java.util.Date.toInstant()" because "o" is null`. Comments surfaced it, since `Resolved At` is empty on every comment when its created event fires, but the evaluator behaved this way for any field without a value and any operator but `is empty` / `is not empty`.

## What's here

- **Absent values have an answer.** `TriggerEvaluator` decides a null value before any operator lambda sees it: `is empty` and `not equals` are met, everything else is not. Applies to plain fields and to items inside collections.
- **An absent association is an absent value.** A field read through an association the object does not hold, the RA profile name of a certificate without one, used to fail while resolving the path. The path is now walked one link at a time, so the same rule applies: `equals` is not met, `not equals` and `is empty` are, as the query-side filter already does for a null. One handler test flipped accordingly: an ignore trigger on an empty RA profile name now ignores a certificate without an RA profile, which is what it asks; a genuinely unevaluable condition, a regex that does not compile, still does not cost the certificate its import.
- **Operator applicability is checked first.** An operator the field type does not support, `greater than` on a string for instance, is still refused as a misconfiguration, regardless of the value, so the null answer never hides a wrong operator.
- **Reasons the operator can act on.** Both refusals name the field, the operator and, where there is one, the value: `Condition on field 'Resolved At' is not set properly: value 'not-a-date' cannot be used with operator 'Greater than'`. The Java detail goes to the debug log.
- **Tests** — `greater than`, `in next`, `equals`, `not equals` and `is not empty` against an unresolved comment; the comparison met once resolved; the same on a certificate with no serial number; an unparseable date producing a reason that names the field and value and carries no Java text. One assertion in the existing exception test changed: `contains` with a numeric value on a null common name expected an exception, and the exception it pinned was this null dereference rather than a value problem; it now expects "not met".

## Notes for reviewers

- Not caused by the comment subsystem; the pinned assertion above shows the behaviour predates it.
- `not equals` on an absent value is met, matching how `is empty` reads and how SQL-side filters treat a null as different from any value.


src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
```diff
@@ -275,10 +275,14 @@ private boolean evaluatePropertyConditionItem(T object, Resource resource, Strin
         // whether the condition is satisfied
         try {
             if (!(objectValue instanceof Collection<?> objectValues)) {
-                if (objectValue != null && filterField.getEnumClass() != null) {
+                BiPredicate<Object, Object> comparison = comparisonFor(fieldType, operator, filterField.getLabel());
+                if (objectValue == null) {
```
ivosh, line 279: **~ Minor** | Correctness

TriggerEvaluator.java:279 only sees null after `getPropertyValue` has resolved the entire property path. For fields such as `RA_PROFILE_NAME` (`raProfile.name`), an unset optional `Certificate.raProfile` throws during resolution at TriggerEvaluator.java:262 and is converted to `RuleException`, so this PR still makes `RA Profile not equals X` fail instead of returning true; positive comparisons are likewise reported as misconfigured rather than simply not met.

_Suggested:_ Resolve joined paths null-safely so an absent intermediate association reaches `evaluateAbsentValue`, and add an unresolved-RA-profile test covering `EQUALS` and `NOT_EQUALS`.

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
```diff
@@ -275,10 +275,14 @@ private boolean evaluatePropertyConditionItem(T object, Resource resource, Strin
         // whether the condition is satisfied
         try {
             if (!(objectValue instanceof Collection<?> objectValues)) {
-                if (objectValue != null && filterField.getEnumClass() != null) {
+                BiPredicate<Object, Object> comparison = comparisonFor(fieldType, operator, filterField.getLabel());
```
ivosh, line 278: **~ Minor** | Testing

TriggerEvaluator.java:278-280 deliberately resolves the comparison before handling a null value, but TriggerEvaluatorITest.java:684-686 only asserts that null `COMMON_NAME` plus `GREATER` throws some `RuleException`. That assertion already passed before this PR because invoking `test` on the missing operator entry was caught and wrapped as the same exception type, so it cannot distinguish the new controlled applicability rejection from the old dereference path.

_Suggested:_ Capture that `RuleException` and assert the controlled reason names `Common Name` and `Greater than` as an unsupported combination.

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
```diff
@@ -317,6 +332,26 @@ private boolean evaluateItemsInCollection(FilterConditionOperator operator, Obje
...
+        BiPredicate<Object, Object> comparison = fieldTypeToOperatorActionMap.get(fieldType).get(operator);
+        if (comparison == null) {
+            throw new RuleException("Condition on field '%s' is not set properly: operator '%s' cannot be applied to it"
+                    .formatted(fieldLabel, operator.getLabel()));
+        }
+        return comparison;
+    }
+
+    /**
+     * What a condition can say about a value the object does not have: it is absent and equals nothing, so only the
```
ivosh: **. Nitpick** | Documentation

TriggerEvaluator.java:346-349 combines the absent-value model, operator-result matrix, comparison fallback, and error-classification rationale in one paragraph. Its first sentence is over 40 words with both a colon and semicolon, while the safety-critical fact that null must not reach operator lambdas is buried at the end rather than surfaced for maintainers.

_Suggested:_ Drop the narration and what-it-is-not. State only plain facts. Rewrite it as a one-line summary followed by short `<p>` sections with bold lead-ins such as `<b>Null semantics:</b>` and `<b>Error boundary:</b>`.

src/test/java/com/otilm/core/integration/evaluator/TriggerEvaluatorITest.java
```diff
@@ -407,6 +407,82 @@ void testCommentRuleEvaluatorOnProperties() throws RuleException {
...
+        Assertions.assertFalse(commentTriggerEvaluator.evaluateConditionItem(condition, comment, Resource.COMMENT));
+        condition.setOperator(FilterConditionOperator.EQUALS);
+        condition.setValue("2019-12-01T22:10:00.274+00:00");
+        Assertions.assertFalse(commentTriggerEvaluator.evaluateConditionItem(condition, comment, Resource.COMMENT));
+        condition.setOperator(FilterConditionOperator.NOT_EQUALS);
+        Assertions.assertTrue(commentTriggerEvaluator.evaluateConditionItem(condition, comment, Resource.COMMENT));
+        condition.setOperator(FilterConditionOperator.NOT_EMPTY);
+        Assertions.assertFalse(commentTriggerEvaluator.evaluateConditionItem(condition, comment, Resource.COMMENT));
+
+        // Once resolved, the same comparison is answered on the value
```
ivosh: **. Nitpick** | Maintainability

`TriggerEvaluatorITest.java:436` says the comment is now resolved immediately before `setResolvedAt`, selecting `GREATER`, and asserting true. The comment adds no behavior or rationale beyond those self-explanatory statements.

_Suggested:_ Remove this step-narration comment and let the setup and assertion document the transition.

src/test/java/com/otilm/core/integration/evaluator/TriggerEvaluatorITest.java
```diff
@@ -609,10 +685,11 @@ void testExceptions() throws RuleException, ParseException {
                 .assertThrows(RuleException.class, () -> certificateTriggerEvaluator
                         .evaluateConditionItem(condition, certificate, Resource.CERTIFICATE));
 
+        // An absent common name is simply not met; the earlier exception here came from dereferencing the null
```
ivosh: **. Nitpick** | Documentation

`TriggerEvaluatorITest.java:688` records why the earlier assertion threw rather than explaining a current invariant. The field/operator setup and `assertFalse` already capture the intended absent-value behavior, while the historical null-dereference detail belongs in this PR rather than the test.

_Suggested:_ Delete the change-history comment and let the setup plus `assertFalse` document the current behavior.

ivosh (changes requested):

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
klaraf755: Taken: the path is walked one link at a time and an absent link yields an absent value, so a certificate without an RA profile answers 'not met' to equals and 'met' to not equals; both covered.

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
klaraf755: The assertion now checks the reason names 'Common Name' and 'greater than'.

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
klaraf755: Rewritten as a one-line summary with null-semantics and error-boundary sections.

src/test/java/com/otilm/core/integration/evaluator/TriggerEvaluatorITest.java
klaraf755: Removed.

src/test/java/com/otilm/core/integration/evaluator/TriggerEvaluatorITest.java
klaraf755: Removed.

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
klaraf755: Follow-up: making the association resolvable flipped one handler test. An ignore trigger on an empty RA profile name now ignores a certificate without an RA profile, which is what it asks and what the query-side filter does for a null; the test pins that, and a separate one keeps 'unevaluable does not block the import' with a regex that does not compile.

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
```diff
@@ -581,7 +621,18 @@ protected void performSendNotificationAction(Resource resource, ResourceEvent ev
     private Object getPropertyValue(Object object, List<Attribute> joinAttributes, Attribute fieldAttribute)
             throws InvocationTargetException, IllegalAccessException, NoSuchMethodException {
         String pathToProperty = FilterPredicatesBuilder.buildPathToProperty(joinAttributes, fieldAttribute);
-        return PropertyUtils.getProperty(object, pathToProperty);
+        if (pathToProperty.isEmpty()) {
+            return object;
```
ivosh: **~ Minor** | Correctness

This PR makes TriggerEvaluator.java:625 return the root object for an empty path. `CBOM_ASSET_FREE_TEXT` at FilterField.java:432 has neither join attributes nor a field attribute and is exposed as a rule field by ResourceServiceImpl.java:209-217, so its advertised `CONTAINS` condition now evaluates `CryptoAsset.toString()` rather than the free-text columns, silently producing false or accidental matches instead of a configuration error.

_Suggested:_ Handle `CBOM_ASSET_FREE_TEXT` with its query-side free-text semantics or reject an empty property path with `RuleException` instead of returning the root entity.

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
```diff
@@ -581,7 +621,18 @@ protected void performSendNotificationAction(Resource resource, ResourceEvent ev
...
+        }
+        // Walked one link at a time: an association the object does not hold leaves the property absent
```
ivosh: **. Nitpick** | Maintainability

TriggerEvaluator.java:627 restates the immediately following loop and null branch; the implementation already shows that it walks each link and returns null for an absent association, so the comment adds no non-obvious rationale.

_Suggested:_ Remove the comment above the `current` variable.

ivosh (dismissed):

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
klaraf755: Right, returning the root object was wrong: an empty path now refuses the condition, naming the field. In-memory free-text would have to repeat the column set the query-side branch owns, so it stays a query-only field. No evaluator exists for CRYPTO_ASSET today, so this is not reachable yet, but the silent comparison against the entity was mine to fix.

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
klaraf755: Removed.

src/main/java/com/otilm/core/evaluator/TriggerEvaluator.java
```diff
@@ -578,10 +618,28 @@ protected void performSendNotificationAction(Resource resource, ResourceEvent ev
...
+     *
+     * @throws RuleException when the field names no property of its own: a free-text field searches several columns at
+     * once and only a query can answer it, so there is nothing here to read
+     */
+    private Object getPropertyValue(Object object, List<Attribute> joinAttributes, Attribute fieldAttribute,
+            FilterField filterField)
+            throws InvocationTargetException, IllegalAccessException, NoSuchMethodException, RuleException {
         String pathToProperty = FilterPredicatesBuilder.buildPathToProperty(joinAttributes, fieldAttribute);
-        return PropertyUtils.getProperty(object, pathToProperty);
+        if (pathToProperty.isEmpty()) {
```
ivosh, line 631: **~ Minor** | Error handling

TriggerEvaluator.java:631 throws the no-single-property `RuleException` before `comparisonFor` is reached. Because RuleServiceImpl.java:157-167 persists any non-null field/operator pair, a `CBOM_ASSET_FREE_TEXT` condition using `GREATER` reaches this branch and is reported as a field-resolution problem rather than the unsupported operator, contradicting this PR's operator-applicability-first behavior and giving the operator the wrong remediation.

_Suggested:_ Validate and cache `comparisonFor(...)` immediately after resolving `filterField`, before calling `getPropertyValue`.

ivosh (approved):

PR #2260, opened 2026-09-15
Name the host object of a history entry evaluated on a comment

### LINKED ISSUE #2235: Workflows - Event firing details: the object link for a comment event points to a non-existent comment detail page (404)
### Description
In **Event firing details** for a comment event, the object UUID is rendered as a link to `/administrator/#/comments/detail/<comment uuid>`. No detail page exists for a comment — comments live only as a panel on their host object — so the route renders a bare `404` with no way back other than the browser's back button.

The same failure mode was recognised and handled elsewhere in this Epic: OmniTrustILM/core#2116 makes `NotificationSubjectResolver` resolve a comment event's subject to the **host object**, precisely because *"a `COMMENT` target would deep-link nowhere"*. Notification deep links therefore point at the host object. The event firing details view was not covered by that change.

Filed against `core` because the event's recorded subject is the comment itself and core#2116 established that comment events resolve their subject to the host object; if the link is built purely on the frontend from resource plus UUID, the fix belongs there instead.

### Steps to Reproduce
1. Create a trigger on the **Comment created** event with any rule, and post a comment so the event fires.
2. Open the host object's **Event History** and find the **Comment created** entry.
3. Open **Event firing details** and click the object UUID link.

### Expected Behavior
- The link opens the **host object's** detail page, where the comment panel lives — consistent with the deep-link decision already made for notifications in core#2116.
- Or, if there is no navigable target, the UUID is shown as plain text instead of a link.

### Actual Behavior
- The link navigates to `/administrator/#/comments/detail/<comment uuid>`.
- The page renders a bare `404` with no navigation, so the operator is stranded.

### Severity
Minor

A dead link in a diagnostic view, with no impact on data or access. It does strand the operator on an empty page with no way back.

### Module
Workflows

### Attachments

_No response_

### Logs

_No response_

### Environment
Commit: 7a414fc
Branch: main
Client:
- macOS Tahoe 26.5.1

klaraf755:
Closes #2235
Depends-On: OmniTrustILM/interfaces#973

The event firing details linked every evaluated object to `/{event resource}/detail/{uuid}`. For a comment event the evaluated object is the comment, which has no page, so the link opened a 404 (Epic: OmniTrustILM/ilm#246).

## What's here

- `EventServiceImpl.getEventHistory` — for the comment histories on the page, one repository call loads the comments and names each one's host resource, uuid and name; `EventHistoryMapper` puts it on the object summary as `hostObject`. Other histories carry nothing. A host deleted since is left out rather than named half-way; comments cascade with their host, so that is only ever a race.
- The host name is resolved with the internal lookup: a reader of the event history already sees the comment uuid and the trigger data, and the name is the one addition, which is accepted.
- Test — a comment history under `COMMENT_CREATED` names its RA profile host with its name; the certificate history on the same page names nothing.

## Notes for reviewers

- Follows the `origin` precedent on `ObjectEventHistoryDto` rather than a new lookup endpoint: notifications already carry the host as their target and audit data already carries the host, so the trigger history was the only surface holding a bare comment uuid, and a per-row fetch in the dialog would have cost more than the link is worth. A comment lookup endpoint remains the right primitive if deep-linking into the comment panel is wanted later.
- Frontend: the firing details dialog should prefer `hostObject` for the link when present — OmniTrustILM/fe-administrator#2144.
- Companion contract: OmniTrustILM/interfaces#973 (merge first).


src/main/java/com/otilm/core/mapper/workflows/EventHistoryMapper.java
```diff
@@ -87,7 +88,7 @@ public static EventHistoryDto toEventHistoryDto(EventHistory eventHistory, int o
                 .stream()
                 .map(objectUuid -> EventHistoryMapper
                         .toTriggerHistoryObjectSummaryDto(triggerHistoriesPerObject.getOrDefault(objectUuid, List.of()),
-                                objectUuid))
+                                objectUuid, hostObjects.get(objectUuid)))
```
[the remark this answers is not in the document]
klaraf755: Right, an immutable empty map refuses a null key. The lookup is skipped for a missing uuid, and a test lists a page holding an ignored object without a uuid.

src/main/java/com/otilm/core/service/impl/EventServiceImpl.java
```diff
@@ -75,6 +84,42 @@ public PaginationResponseDto<ObjectEventHistoryDto> getEventHistory(Resource res
...
+                .stream()
+                .flatMap(perObject -> perObject.values().stream())
+                .flatMap(Collection::stream)
+                .filter(th -> th.getObjectResource() == Resource.COMMENT && th.getObjectUuid() != null)
+                .map(TriggerHistory::getObjectUuid)
+                .distinct()
+                .toList();
+        if (commentUuids.isEmpty()) {
+            return Map.of();
+        }
```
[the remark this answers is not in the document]
klaraf755: Fixed at the mapper: a missing uuid is never looked up, so the map's null tolerance no longer matters. Hosts shared by several comments on the page are now also looked up once.

klaraf755:
Superseded by #2262, the same branch from the org repository so Depends-On can resolve the interfaces snapshot. Copilot's findings from this PR are already addressed in the branch.

klaraf755:
Reopened: this stays the core PR. The interfaces side moved to OmniTrustILM/interfaces#973 (org branch, snapshot published), and the Depends-On line now points there.

src/main/java/com/otilm/core/service/impl/EventServiceImpl.java
```diff
@@ -75,6 +84,51 @@ public PaginationResponseDto<ObjectEventHistoryDto> getEventHistory(Resource res
...
+        List<Comment> comments = commentRepository.findAllById(commentUuids);
```
ivosh, line 105: **~ Minor** | Correctness

EventServiceImpl.java:105 derives host data exclusively by reloading live `Comment` rows. CommentServiceImpl.java:291 and CommentServiceImpl.java:309 support ordinary reply and root deletion, while `trigger_history.object_uuid` has no foreign key to `comment` (V202405201225__workflows_trigger_actions.sql:123-135), so the historical row survives but `findAllById` returns no comment; `hostObject` is then absent and the frontend falls back to the same dead comment link this PR is intended to fix.

_Suggested:_ Persist the comment's host resource and UUID on the history when the event is handled, then map `hostObject` from that association instead of requiring the live `Comment` row.

src/main/java/com/otilm/core/mapper/workflows/EventHistoryMapper.java
```diff
@@ -74,7 +75,7 @@ public static TriggerHistoryObjectSummaryDto toTriggerHistoryObjectSummaryDto(Li
 
     public static EventHistoryDto toEventHistoryDto(EventHistory eventHistory, int objectsEvaluated, int objectsMatched,
             int objectsIgnored, List<UUID> paginatedObjectUuids, int objectsPageNumber, int objectsItemsPerPage,
-            Map<UUID, List<TriggerHistory>> triggerHistoriesPerObject) {
+            Map<UUID, List<TriggerHistory>> triggerHistoriesPerObject, Map<UUID, ResourceObjectDto> hostObjects) {
```
ivosh: **~ Minor** | Documentation

src/main/java/com/otilm/core/mapper/workflows/EventHistoryMapper.java:76-91 adds a public `Map<UUID, ResourceObjectDto> hostObjects` parameter, but neither its name nor documentation states that keys are evaluated comment UUIDs rather than host UUIDs, or that absent entries deliberately omit the host. Only src/main/java/com/otilm/core/service/impl/EventServiceImpl.java:113-117 encodes that contract, so a future caller can compile with a host-keyed map and silently produce no links.

_Suggested:_ Add method-level Javadoc and rename the parameter to `hostObjectsByEvaluatedObjectUuid`, documenting that missing entries yield no host object.

src/main/java/com/otilm/core/service/impl/EventServiceImpl.java
```diff
@@ -75,6 +84,51 @@ public PaginationResponseDto<ObjectEventHistoryDto> getEventHistory(Resource res
...
+            }
+        }
+        return hostObjects;
+    }
+
+    private ResourceObjectDto hostObjectOf(Resource resource, UUID hostUuid) {
+        try {
+            return new ResourceObjectDto(resource, hostUuid,
+                    resourceService.getResourceObjectInternal(resource, hostUuid).getName());
+        } catch (NotFoundException e) {
```
ivosh, line 127: **~ Minor** | Testing

src/main/java/com/otilm/core/service/impl/EventServiceImpl.java:123-129 introduces a special `NotFoundException` path for deletion after comments have been loaded, but src/test/java/com/otilm/core/integration/service/EventServiceITest.java:309-371 covers only a successful lookup, a non-comment history, and a null object UUID. src/test/java/com/otilm/core/integration/service/CommentHostDeletionITest.java:77-88 proves steady-state cascade deletion but cannot enter this interleaving, so a regression that propagates the lookup failure would leave this PR's tests green.

_Suggested:_ Add a focused EventServiceImpl test that supplies a comment from CommentRepository, makes ResourceInternalService throw NotFoundException for its host, and asserts getEventHistory succeeds with hostObject null.

ivosh (changes requested):

src/main/java/com/otilm/core/service/impl/EventServiceImpl.java
klaraf755: Confirmed and leaving it: object_uuid is polymorphic with no foreign key, nothing prunes the history, so a deleted comment's row keeps a uuid that resolves to nothing. The panel shows a plain uuid for that row rather than a link (fe-administrator#2144); a deleted comment has no thread to open anyway. Persisting the host would mean columns on trigger_history and a hook in the generic evaluator path for one resource, which is worth judging on its own rather than inside a link fix.

src/main/java/com/otilm/core/service/impl/EventServiceImpl.java
klaraf755: Added: a comment whose host object no longer exists lists with no host and does not fail the page. No mocking needed - a comment's host reference is polymorphic, so it can simply name a profile that is not there.

ivosh (approved):

PR #2261, opened 2026-09-15
Update the cryptograhy key management to support Cryptography Provider V2

### LINKED ISSUE #2068: Support the synchronous V2 cryptographic key lifecycle
### Description

Allow administrators to manage the synchronous lifecycle of cryptographic keys associated with token profiles that use a Cryptography Provider API V2 connector.

For V2 tokens, the connector declares which key types it supports. The key-creation flow must offer and accept only those types instead of presenting unsupported options and failing after submission.

After creation, administrators must be able to view a key, update its Core-managed metadata, and delete it safely. Metadata changes do not alter the key material held by the cryptography provider.

### Technical scope

- Support synchronous V2 key creation, viewing, metadata modification, and deletion through the Core API and administrator frontend.
- Select the V1 or V2 key flow from the cryptography provider associated with the chosen token profile and token.
- Load the key types supported by the selected V2 token and use them as the allowed creation options.
- Enforce supported key types in both the frontend flow and the Core API.
- Send the selected token, token-profile, and required connector context when creating or destroying a V2 key.
- Create the Core inventory record only after the connector confirms successful key creation.
- Store the connector key reference and metadata required for later lifecycle operations.
- Allow updates to Core-managed key metadata, including ownership, groups, and custom attributes, without changing the key material.
- Destroy the key through the V2 connector before removing its Core inventory record.
- Preserve the Core inventory record when connector destruction fails so the key remains traceable and deletion can be retried safely.
- Preserve the existing V1 key lifecycle behavior.

### Acceptance criteria

- [ ] Core selects the correct key lifecycle flow according to whether the token profile's token uses a V1 or V2 cryptography provider.
- [ ] Selecting a V2-backed token profile loads the key types supported by its token.
- [ ] The frontend offers only the key types declared by the selected V2 token.
- [ ] Core rejects attempts to create a key type that the selected V2 token does not support, including requests submitted outside the frontend.
- [ ] An administrator can synchronously create each key type declared as supported by the selected V2 token.
- [ ] A successfully created V2 key is represented by exactly one Core inventory record with the connector reference needed for later operations.
- [ ] A failed connector creation does not leave a completed or usable key record in Core inventory.
- [ ] An administrator can view a V2 key and its permitted metadata.
- [ ] An administrator can update Core-managed metadata such as owner, groups, and custom attributes without replacing or modifying the key material.
- [ ] Fields that identify the provider-held key material, key type, token, or token profile cannot be changed through metadata modification.
- [ ] Deletion requests destroy the provider-held key before removing the corresponding Core inventory record.
- [ ] If connector destruction fails, the Core inventory record remains available and the failure is reported so deletion can be retried.
- [ ] A confirmed connector destruction is followed by removal of the Core inventory record without intentionally leaving an orphaned inventory entry.
- [ ] Key lifecycle actions use the existing authorization model and produce the required audit records.
- [ ] Existing synchronous V1 key creation, viewing, metadata modification, and deletion continue to work without behavioral regressions.
- [ ] The Sonar quality gate passes.
- [ ] New code has at least 80% test coverage.
- [ ] V1 regression compatibility is verified where the affected path is shared.

### Testing requirements

- Test selection of the V1 or V2 flow from the token profile and associated token.
- Test dynamic loading of V2-supported key types and their presentation in the frontend.
- Test backend rejection of unsupported key types, including direct API requests.
- Test successful synchronous creation for every supported key type.
- Test connector creation failures and verify that no usable Core inventory record remains.
- Test inventory creation, connector reference storage, and duplicate prevention after successful key creation.
- Test viewing and updating owner, groups, custom attributes, and other editable metadata.
- Test that metadata updates cannot alter key material, key type, token, or token profile association.
- Test successful deletion ordering: connector destruction followed by Core inventory removal.
- Test connector destruction failure and verify that the Core inventory record remains traceable and retryable.
- Test authorization and audit logging for create, metadata update, and delete operations.
- Run regression tests for the corresponding synchronous V1 key lifecycle.

### Dependencies / blocked-by relationships

Blocked by the task that adds V2 token-profile management. Record this as a native GitHub blocked-by relationship when the issues are created.

### Out of scope

- Asynchronous key creation, status tracking, or cancellation.
- Modification, replacement, or rotation of existing key material.
- Changing an existing key's type, token, or token-profile association.
- Key import, export, or scan/adopt operations.
- Signing, verification, encryption, decryption, or random generation.
- Migration of existing V1 keys to V2.
- Migration or implementation of a cryptography provider.
- Removal or redesign of the V1 key lifecycle.

LukasNajman:
### Changes to the behavior

- Make key enable/disable operations idempotent: unchanged items produce no errors, history entries, database updates, or cache eviction.

- Standardize key mutation authorization around the parent key’s token, removing separate profile DETAIL checks and enforcing token access even for empty parent keys.

### Known limitation
Creation callbacks still use KEY_PAIR definitions for SECRET requests, which can fail with V2 profiles. Resolving the correct creation schema remains deferred.


src/main/java/com/otilm/core/dao/repository/CryptographicKeyItemRepository.java
```diff
@@ -17,6 +21,35 @@ public interface CryptographicKeyItemRepository extends SecurityFilterRepository
...
+            """)
+    int updateEnabledIfChanged(@Param("uuid") UUID uuid, @Param("enabled") boolean enabled);
+
+    @Modifying(flushAutomatically = true, clearAutomatically = true)
+    @Query("""
+            UPDATE CryptographicKeyItem item
+            SET item.keyData = NULL, item.state = :finalState, item.updatedAt = CURRENT_TIMESTAMP
+            WHERE item.uuid = :uuid
+            """)
+    int finalizeKeyItemDestruction(@Param("uuid") UUID uuid, @Param("finalState") KeyState finalState);
```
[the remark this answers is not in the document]
LukasNajman: Fixed by moving the logic directly to SQL

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -454,258 +405,147 @@ public KeyDetailDto createKey(UUID tokenInstanceUuid, SecuredParentUUID tokenPro
...
-                .setCustomAttributes(attributeEngine
-                        .updateObjectCustomAttributesContent(Resource.CRYPTOGRAPHIC_KEY, key.getUuid(),
-                                request.getCustomAttributes()));
-        keyDetailDto
-                .setAttributes(attributeEngine
-                        .updateObjectDataAttributesContent(ObjectAttributeContentInfo
-                                .builder(Resource.CRYPTOGRAPHIC_KEY, key.getUuid())
-                                .connector(tokenInstanceReference.getConnectorUuid())
-                                .build(), request.getAttributes()));
+        CryptographicKeyFullModel key = persistCreatedKey(tokenProfile, request, remotelyCreatedKeyItems);
```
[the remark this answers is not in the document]
LukasNajman: This is too big change, we won't do it in this PR. We are mirroring the state as-is.

src/main/java/com/otilm/core/service/handler/key/KeyProviderV2Adapter.java
```diff
@@ -0,0 +1,213 @@
...
+    private ProviderKeyItem toCreatedKeyItem(String name, KeyDataV2Dto data, List<MetadataAttribute> keyMeta,
+            KeyMaterial material) {
+        KeyType type = switch (data.getType()) {
+            case PUBLIC -> KeyType.PUBLIC_KEY;
+            case PRIVATE -> KeyType.PRIVATE_KEY;
+            case SECRET -> KeyType.SECRET_KEY;
+        };
+        RemoteKeyReference reference = new RemoteKeyReference.MetadataReference(keyMeta);
+        return new ProviderKeyItem(name, type, data.getAlgorithm(), data.getLength(), reference, material,
+                data.getMetadata());
```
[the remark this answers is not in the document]
LukasNajman: Both REST and MQ clients already reject missing/empty handles before returning to the adapter.

src/main/java/com/otilm/core/service/handler/token/TokenProviderV2Adapter.java
```diff
@@ -66,28 +69,34 @@ public TokenInstanceStatusDetailDto getStatus(TokenInstanceBasicModel tokenInsta
...
         }
         return response;
     }
 
-    private TokenScopedRequestV2Dto scopedRequest(TokenInstanceBasicModel tokenInstance) throws ConnectorException {
+    @Override
+    public List<KeyRequestType> listSupportedKeyRequestTypes(TokenProfileBasicModel tokenProfile)
+            throws ConnectorException {
+        return tokenApiClient.listSupportedKeyRequestTypes(connectorInfo, tokenProfileScopedRequest(tokenProfile));
+    }
```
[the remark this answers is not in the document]
LukasNajman: Updated the API clients in the interfaces repo to not allow nulls

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -2896,7 +2902,8 @@ private void deleteObjectAttributeDefinitionContent(UUID definitionUuid, Resourc
         }
     }
 
-    private SecurityResourceFilter loadCustomAttributesSecurityResourceFilter() {
+    @Transactional(propagation = org.springframework.transaction.annotation.Propagation.NOT_SUPPORTED)
+    public SecurityResourceFilter loadCustomAttributesSecurityResourceFilter() {
```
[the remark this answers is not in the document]
LukasNajman: This is a broader problem our services has - @Transactional annotation on services. We are slowly removing this. We will not update the AttributeEngine in this PR.

src/main/java/com/otilm/core/api/web/CryptographicOperationControllerImpl.java
```diff
@@ -48,6 +48,12 @@ public EncryptDataResponseDto encryptData(@LogResource(uuid = true, affiliated =
                         UUID.fromString(uuid), UUID.fromString(keyItemUuid), request);
     }
 
+    @Override
+    public List<BaseAttribute> listDecryptAttributes(String tokenInstanceUuid, String tokenProfileUuid, String uuid,
```
ivosh, line 52: **! Major** | Correctness

`src/main/java/com/otilm/core/api/web/CryptographicOperationControllerImpl.java:52` (also lines 80, 116, and 134) implements the four operation-specific discovery routes by unconditionally returning `List.of()`. The dependent interface promises provider-derived schemas and 404/422/502/503 failures, while the neighboring `listCipherAttributes`, `listSignatureAttributes`, and `listRandomAttributes` methods delegate through the service and connector path; these new methods instead report `200 []` even for nonexistent resources or V2 providers that require attributes.

_Suggested:_ Clarify if this is intended. If no, then implement all four methods through a provider-version adapter, or return an explicit 501 and disclose the deferral instead of returning successful empty schemas.

<sub>flagged by Reference integrity + OpenAPI & schema fidelity + API contract & compat + Doc-set & publication coherence</sub>

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -454,258 +405,147 @@ public KeyDetailDto createKey(UUID tokenInstanceUuid, SecuredParentUUID tokenPro
...
 
-        CryptographicKey key;
-        if (type.equals(KeyRequestType.KEY_PAIR)) {
-            key = createKeyTypeOfKeyPair(connector, tokenProfile, request, createKeyRequestDto);
-        } else {
-            key = createKeyTypeOfSecret(connector, tokenProfile, request, createKeyRequestDto);
-        }
+        List<ProviderKeyItem> remotelyCreatedKeyItems = keyProviderAdapterFactory
+                .forToken(tokenProfile.tokenInstance())
+                .createKey(tokenProfile, type, request.getAttributes(), request.getName());
```
ivosh, line 443: **! Major** | Correctness

CryptographicKeyServiceImpl.java:423 completes provider creation before duplicate-fingerprint validation at line 426 and before the transactional createKeyWithItems call at line 1212. If either step fails, the provider retains a live key whose V2 metadata handle exists only in memory, so Core cannot list, authorize, destroy, or reconcile it. The neighboring certificate-operation flow avoids this failure mode by persisting a pending record before connector acceptance.

_Suggested:_ Persist a PENDING Core key and stable provider request identifier before calling the connector, then attach the returned handles in a short writer transaction while retaining the pending record for reconciliation on any post-acceptance failure.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -1040,31 +756,246 @@ public void updateKeyUsages(BulkKeyUsageRequestDto request) {
...
+        Map<UUID, CryptographicKeyFullModel> keys = new HashMap<>();
+        for (CryptographicKeyItemBasicModel keyItem : keyItems) {
+            UUID parentKeyUuid = keyItem.parentKeyUuid();
+            CryptographicKeyFullModel key = keys.get(parentKeyUuid);
+            if (key == null) {
+                key = getCryptographicKeyFullModel(parentKeyUuid);
+                keys.put(parentKeyUuid, key);
+            }
+            if (key.tokenInstance() != null) {
+                keyProviderAdapterFactory.forToken(key.tokenInstance()).destroyKeyItem(key, keyItem.reference());
```
ivosh, line 965: **! Major** | Correctness

CryptographicKeyServiceImpl.java:946 destroys every selected item remotely, but the only local cleanup is the single writer call at line 950 after the entire loop. If a later connector call throws, or if that writer transaction fails, the outer catch retains every local row, including items already destroyed remotely; CryptographicKeyServiceImplBulkDeleteTest explicitly expects no local deletion after a connector failure. This PR therefore leaves Core advertising unusable keys and makes recovery depend on provider-specific not-found behavior.

_Suggested:_ Process each item as a durable pending deletion followed by remote destruction and short local cleanup, retaining the pending marker for reconciliation whenever post-acceptance cleanup fails.

src/main/java/com/otilm/core/service/writer/CryptographicKeyWriter.java
```diff
@@ -0,0 +1,508 @@
...
+     * Deletes selected local key items and their attributes and history. Parents left without items are deleted with
+     * their certificate references, attributes, owner/group associations, and comments. All changes commit or roll back
+     * together; parents with remaining items are preserved.
+     *
+     * @param keyItemUuids non-null list of item UUIDs to delete; missing items are ignored
+     * @return number of existing items deleted, counting duplicate UUIDs only once
+     */
+    @Transactional
+    public int deleteKeyItemsWithAssociations(List<UUID> keyItemUuids) {
+        List<CryptographicKeyItem> keyItems = cryptographicKeyItemRepository.findByUuidIn(keyItemUuids);
```
ivosh, line 232: **! Major** | Correctness

CryptographicKeyWriter.java:222 loads selected items without locking their parent rows, then lines 230-231 decides whether each parent is empty from a transaction-local collection snapshot. Two requests deleting different final siblings can each observe the other sibling, both skip parent cleanup, and commit an empty parent with its associations still present. The adjacent deleteKeyIfEmpty method already uses findForUpdateByUuid, confirming that parent serialization is required for this check.

_Suggested:_ Collect the affected parent UUIDs, lock them in deterministic order, then reload and delete the items and re-check parent emptiness while those locks are held.

src/main/java/com/otilm/core/service/impl/CryptographicOperationServiceImpl.java
```diff
@@ -192,12 +193,12 @@ public EncryptDataResponseDto encryptData(SecuredParentUUID tokenInstanceUuid, S
         requestDto.setCipherAttributes(request.getCipherAttributes());
```
ivosh, line 193: **! Major** | Performance

CryptographicOperationServiceImpl.java:171 opens a transaction around encryptData and its connector HTTP call at lines 200-201; decryptData, signData, and verifyData repeat the same pattern. A slow or unavailable provider therefore holds a database transaction and connection for the entire network round trip, while the same class already demonstrates the correct NOT_SUPPORTED boundary for generateCsr at line 476. This PR updates all four operation paths without applying that established boundary.

_Suggested:_ Mark encryptData, decryptData, signData, and verifyData as NOT_SUPPORTED and move their event-history writes into short transactional writer methods after each connector call.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -454,258 +405,147 @@ public KeyDetailDto createKey(UUID tokenInstanceUuid, SecuredParentUUID tokenPro
...
-        }
-        attributeEngine
-                .updateObjectCustomAttributesContent(Resource.CRYPTOGRAPHIC_KEY, key.getUuid(),
-                        request.getCustomAttributes());
+        CryptographicKeyFullModel updatedKey = cryptographicKeyWriter
+                .update(key.uuid(), request, owner, customAttributeResourceFilter);
+        updatedKey.items().forEach(item -> evictKeyItemCache(item.uuid()));
 
-        logger.debug("Key details updated. Key: {}", key);
+        logger.debug("Key details updated. Key: {}", updatedKey);
```
ivosh: **! Major** | Security

CryptographicKeyServiceImpl.java:471 now logs `updatedKey`, whose concrete type is the new `ImmutableCryptographicKeyFullModel`; its generated record string traverses `items`, and CryptographicKeyItemBasicModel.java:16 includes the persisted `keyData`. KeyProviderV1Adapter.java:116-123 populates that value from serialized provider key material for private and secret keys, so every key edit can emit the material at DEBUG. Before this PR, the same log statement received a `CryptographicKey` entity whose `items` relation was explicitly excluded from `toString()`.

_Suggested:_ Replace the whole-model log with an identifier-only message that cannot stringify key items.

src/main/java/com/otilm/core/mapper/crypto/CryptographicKeyDtoMapper.java
```diff
@@ -0,0 +1,223 @@
...
+        KeyItemDetailDto dto = new KeyItemDetailDto();
+        dto.setUuid(item.uuid().toString());
+        dto.setName(item.name());
+        if (item.reference() instanceof RemoteKeyReference.UuidReference reference && reference.uuid() != null) {
+            dto.setKeyReferenceUuid(reference.uuid().toString());
+        }
+        dto.setKeyAlgorithm(item.algorithm());
+        dto.setType(item.type());
+        dto.setLength(item.length());
+        dto.setFormat(item.format());
```
ivosh: **! Major** | Correctness

CryptographicKeyDtoMapper.java:162 maps `item.format()` into a field that both `KeyItemDto` and `KeyItemDetailDto` declare REQUIRED. KeyProviderV2Adapter.java:127 and :164 deliberately supply no material for V2 secret and private keys, while CryptographicKeyWriter.java:123-127 persists a format only when material exists; because ObjectMapperFactory.java:32 suppresses nulls, every such item is serialized without the required `format` property.

_Suggested:_ Make the Core DTO format optional for V2 keys without material, or persist a meaningful non-null format for every created key item.

src/main/java/com/otilm/core/service/handler/key/KeyProviderV2Adapter.java
```diff
@@ -0,0 +1,222 @@
...
+                        .builder(Resource.TOKEN_PROFILE, tokenProfile.uuid())
+                        .connector(connectorUuid)
+                        .build());
+        List<RequestAttribute> resolvedTokenProfileAttributes = operationAttributeResolver
+                .resolveForConnectorRequestAsSystem(connectorUuid, storedProfileAttributes);
+
+        TokenProfileScopedRequestV2Dto request = new TokenProfileScopedRequestV2Dto();
+        request.setTokenAttributes(resolvedTokenAttributes);
+        request.setTokenProfileAttributes(resolvedTokenProfileAttributes);
+        request.setKeyUsages(Set.copyOf(tokenProfile.usages()));
```
ivosh, line 225: **! Major** | Correctness

KeyProviderV2Adapter.java:218 and TokenProviderV2Adapter.java:135 copy a profile's usages into `TokenProfileScopedRequestV2Dto.keyUsages`, whose dependent contract requires a non-empty set. Core still permits profiles without usages—TokenProfileWriter.java:64-67 leaves the persisted bitmask empty when the optional request field is omitted—so the new V2 create, destroy, creation-schema, and request-type operations can send `keyUsages: []` and be rejected by a compliant connector.

_Suggested:_ Require at least one usage when creating or updating a V2 token profile before building scoped requests, or relax the shared V2 request schema if an empty set is valid.

src/test/java/com/otilm/core/integration/service/CryptographicKeyServiceITest.java
```diff
@@ -813,6 +1904,119 @@ void testEnableDisableKey() throws NotFoundException {
...
+        // when
+        boolean changed = cryptographicKeyWriter.setKeyItemEnabled(missingItemUuid, enabled);
+
+        // then
+        Assertions.assertFalse(changed);
+    }
+
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    void setKeyItemsEnabled_doesNotRecordHistory_whenAlreadyInRequestedState(boolean enabled) throws NotFoundException {
```
ivosh, line 2096: **~ Minor** | Testing

CryptographicKeyServiceITest.java:2002 verifies unchanged enable/disable operations only by comparing event-history counts, while the separate test at line 1971 checks the database timestamp. The conditional return at CryptographicKeyServiceImpl.java:1320 is the sole guard preventing cache eviction, but no changed test makes the writer return false and verifies that CacheEvictor remains untouched, so the advertised no-eviction behavior could regress while these tests stay green.

_Suggested:_ Add an unchanged-state enable/disable test that returns false from setKeyItemEnabled and verifies no cache eviction occurs.

src/main/java/com/otilm/core/model/crypto/RemoteKeyReference.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.UUID;
+
+/** Provider-owned key identity, interpreted in the context of its connector and token. */
+public sealed interface RemoteKeyReference {
+
+    record UuidReference(UUID uuid) implements RemoteKeyReference {
+    }
+
+    /** Opaque durable key handle returned by a stateless provider. */
+    record MetadataReference(List<MetadataAttribute> keyMeta) implements RemoteKeyReference {
```
ivosh, line 39: **~ Minor** | Correctness

RemoteKeyReference.java:14 stores the caller-supplied metadata list directly, and both CryptographicKeyItemBasicModel.java:26 and CryptographicKeyServiceImpl.java:1042 pass the entity's list into it. Clearing or otherwise structurally modifying either exposed list therefore changes an already-created API or cached operation snapshot, despite these models being used as immutable snapshots; CryptographicKeySnapshotsTest.java:76 checks the other snapshot collections but omits this one.

_Suggested:_ Copy the metadata list in MetadataReference's compact constructor and add a test proving that mutations to the source list cannot alter the snapshot.

src/main/java/com/otilm/core/attribute/engine/AttributeEngine.java
```diff
@@ -1635,14 +1635,20 @@ public List<ResponseAttribute> replaceObjectDataAttributesContent(ObjectAttribut
     public List<ResponseAttribute> updateObjectCustomAttributesContent(Resource objectType, UUID objectUuid,
             List<RequestAttribute> requestAttributes)
             throws ValidationException, NotFoundException, AttributeException {
+        SecurityResourceFilter securityResourceFilter = loadCustomAttributesSecurityResourceFilter();
+        return updateObjectCustomAttributesContent(objectType, objectUuid, requestAttributes, securityResourceFilter);
+    }
+
+    public List<ResponseAttribute> updateObjectCustomAttributesContent(Resource objectType, UUID objectUuid,
```
ivosh: **~ Minor** | Documentation

AttributeEngine.java:1642 adds a public overload that accepts a preloaded SecurityResourceFilter, and AttributeEngine.java:2905 exposes the NOT_SUPPORTED loader used to obtain it. Neither method has method-level Javadoc explaining the non-obvious contract: the filter is captured before entering the writer transaction, null permits all custom-attribute content, and the overload must reuse that authorization decision without another lookup; only AttributeEngineITest.java:876 currently makes the transaction-suspension requirement discoverable.

_Suggested:_ Add method-level Javadoc to both methods documenting filter loading, null semantics, and the requirement to obtain the filter outside the writer transaction.

src/main/java/com/otilm/core/service/CryptographicKeyExternalService.java
```diff
@@ -147,21 +152,31 @@ KeyDetailDto createKey(UUID tokenInstanceUuid, SecuredParentUUID tokenProfileUui
     void deleteKeyItems(SecurityFilter filter, List<String> keyItemUuids) throws ConnectorException;
 
     /**
-     * Destroy a key
+     * Destroys selected items of a key after validating the complete selection and its authorization. Duplicate item
```
ivosh: **~ Minor** | Documentation

CryptographicKeyExternalService.java:155, CryptographicKeyExternalService.java:170, and CryptographicKeyExternalService.java:278 each compress validation order, deduplication, continue-on-error behavior, completion reporting, and non-rollback semantics into one paragraph. The safety-critical fact that earlier successful destruction is not undone is buried at the end and repeated three times with slightly different wording, making this load-bearing API contract difficult to scan and maintain consistently.

_Suggested:_ Rewrite each destructive-operation Javadoc as a one-line summary followed by separate `<p><b>Validation:</b>`, `<p><b>Failure handling:</b>`, and `<p><b>Atomicity:</b>` paragraphs while retaining the existing facts.

src/test/java/com/otilm/core/mapper/crypto/CryptographicKeyDtoMapperTest.java
```diff
@@ -0,0 +1,377 @@
...
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import static com.otilm.core.util.builders.CryptographicKeyFullModelBuilder.aKeySnapshot;
+import static com.otilm.core.util.builders.CryptographicKeyItemBasicModelBuilder.aKeyItemSnapshot;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
ivosh: **. Nitpick** | Testing

CryptographicKeyDtoMapperTest.java:33 imports JUnit assertEquals and alternates it with AssertJ inside the same assertion blocks, including lines 58-66. The adjacent TokenProfileDtoMapperTest.java:12 uses AssertJ consistently, so this new mapper test introduces a one-off convention and less consistent assertion diagnostics.

_Suggested:_ Replace the JUnit equality assertions in CryptographicKeyDtoMapperTest with AssertJ assertThat assertions.

src/main/java/com/otilm/core/service/writer/CryptographicKeyWriter.java
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    /**
+     * Persists the supplied key entity and returns its basic model.
+     *
+     * @param key key entity containing the changes to persist
+     * @return immutable basic model of the saved key
+     */
+    @Transactional
+    public CryptographicKeyBasicModel update(CryptographicKey key) {
```
ivosh: **~ Minor** | Maintainability

This PR adds the transactional `CryptographicKeyWriter.update(CryptographicKey)` method at `CryptographicKeyWriter.java:504`, but it finds no caller. The actual edit path uses the separate `update(UUID, EditKeyRequestDto, ...)` overload, so this undisclosed entry point is dead service API.

_Suggested:_ Remove `update(CryptographicKey)`, or wire the intended production caller before merging.

.aiignore
```diff
@@ -0,0 +1,12 @@
+# An .aiignore file follows the same syntax as a .gitignore file.
```
ivosh: **. Nitpick** | Maintainability

.aiignore:1-9 in this PR includes stock template comments such as “you can ignore files” and “or folders” that merely narrate the patterns below them.

_Suggested:_ Remove the four template comments and retain only the ignore patterns.

src/main/java/com/otilm/core/service/handler/key/KeyProviderV2Adapter.java
```diff
@@ -0,0 +1,222 @@
...
+            AttributeEngine attributeEngine, OperationAttributeResolver operationAttributeResolver,
+            OutboundSecretContainment outboundSecretContainment) {
+        this.connectorInfo = connectorInfo;
+        this.attributeEngine = attributeEngine;
+        this.operationAttributeResolver = operationAttributeResolver;
+        this.outboundSecretContainment = outboundSecretContainment;
+        this.keyManagementSyncApiClient = connectorApiFactory.getKeyManagementApiClientV2(connectorInfo);
+    }
+
+    /** Placeholder until stateless provider key listing is implemented. */
```
ivosh: **. Nitpick** | Documentation

KeyProviderV2Adapter.java:67 describes when listing might be implemented rather than documenting the method’s current contract, effectively leaving a TODO in prose that will become stale.

_Suggested:_ Replace the placeholder with a current-contract statement such as `V2 providers do not expose key listing`, or remove the method until support exists.

src/main/java/com/otilm/core/service/handler/token/TokenProviderAdapter.java
```diff
@@ -18,16 +20,50 @@
...
+    /**
+     * Lists the key request types supported for the specified token profile.
+     *
+     * @param tokenProfile the token profile for which to determine supported key request types
+     * @return the supported key request types
+     * @throws ConnectorException if the connector operation fails
+     */
+    List<KeyRequestType> listSupportedKeyRequestTypes(TokenProfileBasicModel tokenProfile) throws ConnectorException;
+
+    ;
```
ivosh: **. Nitpick** | Maintainability

TokenProviderAdapter.java:68 contains a standalone semicolon after the final method declaration; it has no effect and is leftover syntax noise.

_Suggested:_ Remove the standalone semicolon.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -1040,31 +756,246 @@ public void updateKeyUsages(BulkKeyUsageRequestDto request) {
...
+
+    /**
+     * Finds one ordered page of accessible key items matching a validated search request, with its total count.
+     * Configures the supplied security filter's parent property for token-level authorization.
+     *
+     * @param contentFilter shared lazy attribute permissions, also used when projecting the response columns
+     */
+    private Page<CryptographicKeyItem> findAccessibleKeyItemsMatchingSearch(SecurityFilter filter,
+            SearchRequestDto request, Supplier<CustomAttributeContentFilter> contentFilter) {
+        // Convert the API's one-based page number to Spring's zero-based index and limit the page size.
```
ivosh, line 884: **. Nitpick** | Maintainability

CryptographicKeyServiceImpl.java:864-892 adds line-by-line comments for constructing `page`, setting the parent property, selecting a default sort, fetching entities, counting rows, and returning `PageImpl`; those statements and names already communicate the same information. The comments at lines 876 and 886 do carry non-obvious pagination and ordering invariants.

_Suggested:_ Remove the narration at lines 864, 873, 879, 881, 884, 890, and 892 while retaining the invariant comments at lines 876 and 886.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -378,10 +324,13 @@ public List<KeyDto> listKeyPairs(Optional<String> tokenProfileUuid, SecurityFilt
         }
 
         List<KeyDto> response = cryptographicKeyRepository
-                .findUsingSecurityFilter(filter, List.of("groups", "owner"), additionalWhereClause, null,
-                        (root, cb) -> cb.desc(root.get("created")))
+                .findUsingSecurityFilter(filter, List
+                        .of("groups", "owner", "items", "certificates", "altCertificates", "tokenProfile",
```
ivosh: **~ Minor** | Performance

CryptographicKeyServiceImpl.java:327-339 changes the unpaginated key-pair lookup to fetch-join five plural associations—groups, items, primary certificates, alternative certificates, and every sibling token profile—and only afterward filters for two active items in Java. The database must materialize their Cartesian product before DISTINCT deduplicates root keys, whereas the neighboring listCryptographicKeys path pages UUIDs before loading associations; keys with many certificates or profiles can therefore make this endpoint consume excessive time and memory.

_Suggested:_ Filter eligible key-pair UUIDs in SQL, bound the result, and batch-load the required associations and counts instead of fetch-joining every plural path in one query.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -1040,31 +756,246 @@ public void updateKeyUsages(BulkKeyUsageRequestDto request) {
...
+        List<CryptographicKeyItemBasicModel> keyItems = cryptographicKeyItemRepository
+                .findBasicModelsByUuidIn(permittedUuids);
+
+        logger.debug("Going to delete key items with UUIDs {}", permittedUuids);
+        Map<UUID, CryptographicKeyFullModel> keys = new HashMap<>();
+        for (CryptographicKeyItemBasicModel keyItem : keyItems) {
+            UUID parentKeyUuid = keyItem.parentKeyUuid();
+            CryptographicKeyFullModel key = keys.get(parentKeyUuid);
+            if (key == null) {
+                key = getCryptographicKeyFullModel(parentKeyUuid);
```
ivosh: **~ Minor** | Performance

CryptographicKeyServiceImpl.java:938-942 calls getCryptographicKeyFullModel once per distinct parent inside the key-item loop. This PR removes the prior findWithKeyByUuidIn batch fetch, so a default 500-item deletion can now issue up to 500 full-graph parent queries—including items, certificates, groups, and token profiles—before performing connector work.

_Suggested:_ Add an IN-based batched finder for the distinct parent UUIDs and build the parent-model map before iterating over key items.

src/main/java/com/otilm/core/dao/repository/TokenProfileRepository.java
```diff
@@ -55,14 +57,16 @@ default Optional<TokenProfileFullModel> findFullModelByUuidAndTokenInstanceRefer
     }
 
     default List<TokenProfileFullModel> findFullModelsUsingSecurityFilter(SecurityFilter filter) {
-        return findUsingSecurityFilter(filter, List.of("tokenInstanceReference"), null)
+        return findUsingSecurityFilter(filter,
+                List.of("tokenInstanceReference.connectorInterface", "tokenInstanceReference.tokenProfiles"), null)
```
ivosh: **~ Minor** | Performance

TokenProfileRepository.java:61 adds tokenInstanceReference.tokenProfiles to an unpaginated query whose root is already TokenProfile. For N profiles belonging to one token, that fetch join produces N×N SQL rows before DISTINCT, even though TokenProfileDtoMapper only reads tokenInstance().status() from the expanded token snapshot.

_Suggested:_ Use a lightweight listing projection containing the token status and omit tokenInstanceReference.tokenProfiles from both unpaginated listing fetches.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -1188,261 +1083,231 @@ public List<NameAndUuidDto> listResourceObjects(SecurityFilter filter, List<Sear
...
+            throw new ValidationException(ValidationError.create(message));
+        }
+        return requestedUuids.stream().map(belongingItems::get).toList();
+    }
+
+    private List<UUID> parseKeyItemUuids(List<String> keyItemUuids) {
+        return keyItemUuids == null ? List.of() : keyItemUuids.stream().map(UUID::fromString).toList();
+    }
+
+    private List<CryptographicKeyItemBasicModel> getRequestedKeyItems(List<UUID> keyItemUuids) {
```
ivosh, line 1281: **~ Minor** | Security

CryptographicKeyServiceImpl.java:1250-1267 loads every requested item and emits a validation response enumerating absent UUIDs before `verifyPermissionsForKeyItemParents` runs in the bulk mutation methods at lines 803, 812, 842, and 852, or the equivalent check at line 823 for destruction. A caller admitted by the generic method-action check but denied DETAIL on the target token can therefore distinguish an absent item from an existing unauthorized item by observing the validation response versus the later authorization denial.

_Suggested:_ Resolve bulk selections through a parent-token-scoped security filter and return the same generic not-found-or-not-authorized failure before exposing any item UUID.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -1040,31 +756,246 @@ public void updateKeyUsages(BulkKeyUsageRequestDto request) {
...
+    @ExternalAuthorization(resource = Resource.CRYPTOGRAPHIC_KEY, action = ResourceAction.LIST)
+    public List<SearchFieldDataByGroupDto> getSearchableFieldInformation() {
+        return getSearchableFieldsMap();
+    }
+
+    @Override
+    @ExternalAuthorization(resource = Resource.CRYPTOGRAPHIC_KEY, action = ResourceAction.ENABLE,
+            parentResource = Resource.TOKEN, parentAction = ResourceAction.DETAIL)
+    public void enableKeyItems(List<String> uuids) throws NotFoundException {
+        List<CryptographicKeyItemBasicModel> keyItems = getRequestedKeyItems(parseKeyItemUuids(uuids));
```
ivosh: **~ Minor** | Documentation

`src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java:802` (also lines 811, 820, 842, and 851) now preloads the complete key-item selection, and `CryptographicKeyServiceImpl.java:1265` throws `ValidationException` when any UUID is absent. Consequently, enable, disable, destroy, compromise, and usage-update requests that previously skipped missing items and returned 204 now reject the whole batch with HTTP 422 before mutation, but this client-visible contract change is absent from this PR's behavior notes.

_Suggested:_ Add a release note naming the five affected bulk key-item endpoints and their new all-or-nothing HTTP 422 behavior.

src/main/java/com/otilm/core/api/web/CryptographicKeyControllerImpl.java
```diff
@@ -122,7 +122,7 @@ public KeyItemDetailDto getKeyItem(@LogResource(uuid = true, affiliated = true)
     @AuditLogged(module = Module.CRYPTOGRAPHIC_KEYS, resource = Resource.CRYPTOGRAPHIC_KEY,
             affiliatedResource = Resource.TOKEN, operation = Operation.CREATE)
     public KeyDetailDto createKey(@LogResource(uuid = true, affiliated = true) String tokenInstanceUuid,
-            String tokenProfileUuid, KeyRequestType type, KeyRequestDto request) throws AlreadyExistException,
+            String tokenProfileUuid, KeyRequestType type, @Valid KeyRequestDto request) throws AlreadyExistException,
```
ivosh: **~ Minor** | Documentation

`src/main/java/com/otilm/core/api/web/CryptographicKeyControllerImpl.java:125` adds `@Valid`, activating the dependent `KeyRequestDto` constraints that require nonblank names and descriptions and non-null attributes. Requests that the previous endpoint passed to the service can therefore now fail at the HTTP boundary with 422, but this PR does not disclose the stricter request contract.

_Suggested:_ Add a release note documenting the newly required nonblank name and description and non-null attributes fields for key creation.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -1040,31 +756,246 @@ public void updateKeyUsages(BulkKeyUsageRequestDto request) {
...
+        setKeyItemsUsages(itemUuids(items), request.getUsage());
         logger.info("Key usages updated: {}", uuid);
     }
 
+    @Override
+    @ExternalAuthorization(resource = Resource.CRYPTOGRAPHIC_KEY, action = ResourceAction.DETAIL)
+    public List<KeyEventHistoryDto> getEventHistory(UUID uuid, UUID keyItemUuid) throws NotFoundException {
+        logger.debug("Request to get the list of events for the key item");
+        CryptographicKeyBasicModel key = getCryptographicKeyBasicModel(uuid);
+        verifyPermissionsForAssociatedToken(key, "get key item history", ResourceAction.MEMBERS);
```
ivosh: **~ Minor** | Documentation

`src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java:772` adds a parent-token `MEMBERS` authorization check to `getEventHistory`, and line 773 now requires the item to belong to the key named in the route. Previously the endpoint required only cryptographic-key `DETAIL` access and ignored the parent/item relationship; this PR describes token authorization changes only for mutations, so clients with direct key access but no token membership can be newly denied without notice.

_Suggested:_ Document the added token-membership and parent-item matching requirements for the key-item history endpoint.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
```diff
@@ -128,63 +126,40 @@
...
-import org.springframework.transaction.TransactionStatus;
 import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
-import org.springframework.transaction.support.DefaultTransactionDefinition;
 
 import static java.util.function.Predicate.not;
 
 @Service(Resource.Codes.CRYPTOGRAPHIC_KEY)
-@Transactional(noRollbackFor = ValidationException.class)
+@Transactional(propagation = Propagation.NOT_SUPPORTED)
```
ivosh, line 144: **! Major** | Correctness

CryptographicKeyServiceImpl.java:142 makes every method `NOT_SUPPORTED`, so `listCryptographicKeys` no longer keeps a persistence context open around its repository reads and DTO mapping. At CryptographicKeyServiceImpl.java:885 it fetches items through `findFullByUuidIn`, whose graph at CryptographicKeyItemRepository.java:101 omits `key.tokenInstanceReference`, but CryptographicKeyItem.java:197-199 dereferences that lazy association while mapping. Direct non-web callers, including the existing integration tests, therefore hit `LazyInitializationException` for token-backed keys; HTTP only works while OSIV happens to mask the defect.

_Suggested:_ Add `key.tokenInstanceReference` to `findFullByUuidIn`'s entity graph so detached listing rows contain every association `mapToSummaryDto` reads.

.aiignore
ivosh: This file currently excludes nothing that isn't already excluded or already absent by `.gitignore`. I recommend to remove it from the PR.

ivosh (changes requested):
Thank you for the PR.
I suspect many of the findings are flagging pre-existing issues. Therefore I suggest to file follow-up issues for these problems.

src/main/java/com/otilm/core/api/web/CryptographicOperationControllerImpl.java
LukasNajman: These are unused endpoints for now. Actually, it would be better to not include them, in this PR. However, next PR will fill in the implementation, so I think it is not worth the effort removing them from the PR.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
LukasNajman: The local items are now cleaned up in the for loop

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
LukasNajman: If we persist the key before remote creation, we will not have the remote handle, se we won't be able to identify the key anyway. Also, the fingerprint can not be computed without the data returned by the provider.

Persisting PENDING core key creation is a big task,  we have dedicated issue for this. https://github.com/OmniTrustILM/ilm/issues/388

LukasNajman: I have at least improved the logging, so it is clearly stated that that the key could have been orphaned in the provider.

src/main/java/com/otilm/core/service/impl/CryptographicOperationServiceImpl.java
LukasNajman: Exiting problem, will be addressed when updating the key operations to work with cryptography provider v2

src/main/java/com/otilm/core/service/writer/CryptographicKeyWriter.java
LukasNajman: The parents are not locked before item deletion

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
LukasNajman: I have created a specialized interface that provides method for unique identification without exposing any sensitive information

src/main/java/com/otilm/core/mapper/crypto/CryptographicKeyDtoMapper.java
LukasNajman: The mapper now provides CUSTOM format with an explanatory message that the key content is managed by provider and not available in the core.

src/main/java/com/otilm/core/service/handler/key/KeyProviderV2Adapter.java
LukasNajman: This problem was already present. For example, fronted does not allow submitting form with empty key usages, so the main problem is a missing validation of the request. But that is a breaking change, the v1 connector was fine with empty key usages (though not very useful to be empty and the FE denied this anyway). For now, I would leave it as is, the V2 connectors will refuse such requests and fail. We can tighten this when the V1 is removed. The whole controller is missing a request validation anyway, so we should implement it in the future. This PR is already too big for this.

src/test/java/com/otilm/core/integration/service/CryptographicKeyServiceITest.java
LukasNajman: Updated the test

src/main/java/com/otilm/core/model/crypto/RemoteKeyReference.java
LukasNajman: I am not sure how that would help. Returning a copy that would somebody modify in faith it will have some impact will probably fail the call to connector the same way as modifying the original metadata. I am not addressing this.

src/test/java/com/otilm/core/mapper/crypto/CryptographicKeyDtoMapperTest.java
LukasNajman: unified

src/main/java/com/otilm/core/service/writer/CryptographicKeyWriter.java
LukasNajman: removed

src/main/java/com/otilm/core/service/handler/key/KeyProviderV2Adapter.java
LukasNajman: will throw instead

src/main/java/com/otilm/core/service/handler/token/TokenProviderAdapter.java
LukasNajman: removed

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
LukasNajman: I find those useful, leaving them here

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
LukasNajman: Yeah, but that is why we cache them. I don't think it is probably somebody would be deleting 500 key items, more so 500 items each belonging to the different key. I will not optimize this prematurely.

src/main/java/com/otilm/core/dao/repository/TokenProfileRepository.java
LukasNajman: implemented

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
LukasNajman: The unified message is returned so the caller can't distinguish between those two states.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
LukasNajman: A new list model was used to reduce the graph size

.aiignore
LukasNajman: removed

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
LukasNajman: added

src/main/java/com/otilm/core/mapper/crypto/CryptographicKeyDtoMapper.java
```diff
@@ -0,0 +1,254 @@
...
+        }
+        dto.setKeyAlgorithm(item.algorithm());
+        dto.setType(item.type());
+        dto.setLength(item.length());
+        dto.setFormat(responseFormat(item));
+        dto.setState(item.state());
+        dto.setEnabled(item.enabled());
+        dto.setUsage(item.usages());
+        dto.setReason(item.reason());
+        dto.setKeyData(item.keyData() == null ? PROVIDER_MANAGED_KEY_DATA : item.keyData());
```
ivosh, line 193: **~ Minor** | Correctness

CryptographicKeyDtoMapper.java:193 treats every null `keyData` as provider-managed, but destruction deliberately clears this field for both V1 and V2 items. Consequently a destroyed item nested in a key detail is returned with explanatory text and `CUSTOM` format, while the direct item-detail mapper in `CryptographicKeyItem.java:182` returns null and the stored format for the same item.

_Suggested:_ Apply the provider-managed placeholder and `CUSTOM` format only to eligible metadata-backed V2 items, preserving null data and the stored format for destroyed items.

src/main/java/com/otilm/core/api/web/CryptographicOperationControllerImpl.java
ivosh: **Re-check: not addressed.**

The fix delta contains no change to `CryptographicOperationControllerImpl.java`; the current `listDecryptAttributes()`, `listVerifyAttributes()`, `listSignAttributes()`, and `listEncryptAttributes()` methods still unconditionally return `List.of()`.

The reply acknowledges deferral to an unspecified next PR, but the endpoints still return successful empty schemas rather than using a provider adapter or returning an explicit 501.

src/main/java/com/otilm/core/model/crypto/RemoteKeyReference.java
ivosh: **Re-check: not addressed.**

`RemoteKeyReference.java:39` still declares `MetadataReference(List<MetadataAttribute> keyMeta)` without a compact constructor or `List.copyOf(keyMeta)`; the delta also adds no source-list mutation test to `CryptographicKeySnapshotsTest`.

The metadata list remains caller-owned and mutable after snapshot construction, exactly as flagged.

src/main/java/com/otilm/core/service/handler/key/KeyProviderV2Adapter.java
ivosh: **Re-check: not addressed.**

No usage validation or schema relaxation appears in the fix delta; `KeyProviderV2Adapter.tokenProfileScopedRequest()` still calls `request.setKeyUsages(Set.copyOf(tokenProfile.usages()))` at line 225 without rejecting an empty set, and `TokenProviderV2Adapter` retains the same behavior at line 135.

The author explicitly chose to leave the behavior unchanged, so empty usages can still reach V2 connector requests.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
ivosh: **Re-check: partial.**

`deleteKeyItemsBatch()` now performs `deleteKeyItemsWithAssociations(...)` inside the item loop at lines 966-968, after each remote `destroyKeyItem(...)` call at line 964, but it creates no pending-deletion marker before the remote call.

Earlier successful items are cleaned up before a later connector failure, but a writer failure still leaves the currently destroyed remote item advertised locally.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
ivosh: **Re-check: partial.**

`listKeyPairs()` now omits certificate collections and sibling token profiles from the fetch graph and obtains certificate counts separately in `getKeyCertificateCounts()` at lines 362-374, but lines 331-334 still run an unbounded fetch of plural `groups` and `items`, and lines 344-357 still filter eligible pairs in Java.

The graph is smaller, but SQL eligibility filtering, result bounding, and batched association loading remain unimplemented.

src/main/java/com/otilm/core/service/impl/CryptographicKeyServiceImpl.java
ivosh: **Re-check: not addressed.**

`deleteKeyItemsBatch()` at lines 958-963 still calls `cryptographicKeyRepository.findFullModelByUuid` through `computeIfAbsent` once for every distinct parent encountered; no IN-based parent-model finder is used.

The map avoids duplicate lookups for siblings but does not address the flagged one-full-query-per-distinct-parent behavior.

src/main/java/com/otilm/core/service/impl/CryptographicOperationServiceImpl.java
ivosh: **Re-check: not addressed.**

`CryptographicOperationServiceImpl.java` still annotates `encryptData` (line 171), `decryptData` (line 227), `signData` (line 299), and `verifyData` (line 374) with plain `@Transactional`; their connector calls and inline `eventHistoryService.addEventHistory(...)` writes remain, while the fix delta changes only logging in these methods.

The reply promises a future provider-v2 update but cites no tracked issue or follow-up PR, and the requested transaction-boundary refactor is absent.

ivosh (commented):
Unaddressed or only partially addressed findings.

PR #2262, opened 2026-09-15
Name the host object of a history entry evaluated on a comment

### LINKED ISSUE #2235: Workflows - Event firing details: the object link for a comment event points to a non-existent comment detail page (404)

klaraf755:
Closes #2235
Depends-On: OmniTrustILM/interfaces#973

The event firing details linked every evaluated object to `/{event resource}/detail/{uuid}`. For a comment event the evaluated object is the comment, which has no page, so the link opened a 404 (Epic: OmniTrustILM/ilm#246).

## What's here

- `EventServiceImpl.getEventHistory` — for the comment histories on the page, one repository call loads the comments and names each one's host resource, uuid and name; `EventHistoryMapper` puts it on the object summary as `hostObject`. Other histories carry nothing. A host deleted since is left out rather than named half-way; comments cascade with their host, so that is only ever a race.
- The host name is resolved with the internal lookup: a reader of the event history already sees the comment uuid and the trigger data, and the name is the one addition, which is accepted.
- Test — a comment history under `COMMENT_CREATED` names its RA profile host with its name; the certificate history on the same page names nothing.

## Notes for reviewers

- Follows the `origin` precedent on `ObjectEventHistoryDto` rather than a new lookup endpoint: notifications already carry the host as their target and audit data already carries the host, so the trigger history was the only surface holding a bare comment uuid, and a per-row fetch in the dialog would have cost more than the link is worth. A comment lookup endpoint remains the right primitive if deep-linking into the comment panel is wanted later.
- Frontend: the firing details dialog should prefer `hostObject` for the link when present — OmniTrustILM/fe-administrator#2144.
- Companion contract: OmniTrustILM/interfaces#973 (merge first).


klaraf755:
Closed in favour of #2260, which stays as the core PR; only the interfaces side needed to move into the org repository for its snapshot to publish.

PR #2263, opened 2026-09-15
Delete a CBOM through the cryptographic asset inventory, and remember that it went

vyskocilm:
Part of #2073 — **PR 3 of a 5-PR plan**, on top of #2257. Deliberately not `Closes`: the duplicate-`bom-ref` refusal with the findings channel, and the weekly reconciliation, follow as separate PRs.

## What this is

`crypto_asset_source_to_cbom_key` is `ON DELETE RESTRICT`. From the moment #2256's ingest wrote a source row, **a CBOM could not be deleted through the API at all** — the migration that declared the constraint assigned the remedy to this ticket by name. Both delete paths now withdraw the inventory first, then remove the header and record a `cbom_tombstone` in one transaction.

The tombstone is read back where it has to be. The repository goes on offering a document an operator deleted here — deletion in Core is not deletion there — so the feed pass and the skip retry both test the table before storing. Nothing read `cbom_tombstone` before this PR; it was written for it.

## The two withdrawal contentions, and why they differ

A keyed delete **waits** for the CBOM's asset-sync lock rather than skipping. An operator is told the document is gone, and the deletion cannot proceed while the inventory still references it, so a skipped withdrawal would turn the delete into a refusal for as long as another node happened to be syncing. The housekeeping withdrawal still **skips**: its caller learns it from `Withdrawal.complete()` and leaves the CBOM owing the work.

`deleteCbom` became `NOT_SUPPORTED` like the bulk path, because the withdrawal runs a transaction per batch.

**The blocking acquisition carries a ranking obligation, and `CbomAssetIngestService`'s lock ranking is amended to say so.** It asserted that the asset-sync lock is only ever acquired with a non-blocking `tryLock`; `withdrawWaiting` makes that false. What keeps it safe is the ranking rather than the call — the wait is the **first** lock of its transaction, above `ALIAS_DECISION_LOCK` and above every `crypto_asset` row lock, so a waiter holds nothing a holder could go on to want. A path that took either of those and *then* waited on the asset-sync lock would close a cycle; that is what the ranking forbids. The header delete takes the same key under the same rule.

## The work list is read under the lock, one page per batch

The lock is transaction-scoped, so it is released at every batch commit. A withdrawal that read its work list once at entry would therefore promise more than it delivered: every `crypto_asset_source` row an ingest of the same document attached after that read — including in the gap between two batch commits — went unwithdrawn, and `complete()` was still returned. A CBOM sourcing nothing took no lock at all, so a delete could race an ingest writing that document's *first* sources with no exclusion whatsoever.

The withdrawal now reads one page per batch inside the transaction that holds the lock, and ends on a page the lock itself says is empty. `deleteAndTombstone` takes the same lock, which closes the last window — between the final batch commit and the header delete — in both directions: the `RESTRICT` refusal on this side, and on the other an ingest writing sources against a header that has just been deleted. On the ingest side, `writeBatchUnderClusterLock` re-reads whether the header still exists, next to the supersede re-read that is there for the same reason, and reports `DELETED`.

## What a half-done deletion leaves behind

The withdrawal commits before the header delete, so a delete that fails in between leaves a row reading `SYNCED` that sources nothing — and `markFailed` deliberately refuses to overwrite `SYNCED`, which would keep the backlog pass from ever rebuilding it. Both paths use `markWithdrawnButNotDeleted`, which writes over `SYNCED` — and which is **not** an ingest attempt, so it leaves `asset_sync_attempted_at` alone. Stamping it would exclude the row from the retry list for a full `cbom.sync.ingest-retry-after` window and then sort it behind every older candidate: the row that most urgently owes a rebuild made to wait the longest.

Three things that write is **not** used for:

- **A CBOM that owes its first ingest.** `PENDING` is excluded. Nothing was withdrawn from a row whose assets were never ingested, and flipping it would move it off the fast pending list onto the slow retry one under a sentence untrue of it.
- **The inventory's own refusal**, identified by constraint name rather than by exception type. The `RESTRICT` foreign key refusing means the row is genuinely sourced again, which makes `SYNCED` true of it; recording that a deletion withdrew the inventory would force a re-ingest of a document nothing is wrong with. `crypto_asset_source_to_cbom_key` is today the only foreign key in the schema referencing `cbom(uuid)` and the tombstone insert cannot conflict, so the two readings coincide — but only the constraint name says what is actually meant, and the next foreign key added to `cbom` would otherwise join the carve-out silently.
- **A withdrawal that withdrew nothing.** Not the same as a withdrawal that *threw*: every batch commits on its own, so a failure on batch *k* leaves *1..k-1* withdrawn and their orphan assets irreversibly deleted. `WithdrawalFailedException` carries what committed, and the failure is recorded for exactly the withdrawals that changed the inventory. `deleteCbom` guards its withdrawal the way the bulk path does — record the partial withdrawal, shape the message, log the `FAILURE` audit entry — rather than letting it propagate.

**And `markSynced` can no longer undo that record.** Success is otherwise unconditional and neither writer holds the asset-sync lock, so an ingest sitting between its last asset batch and its `markSynced` would put the row straight back to `SYNCED` with the error cleared — the exact stranded state the failure write exists to prevent. `markSynced` now refuses to overwrite a row carrying the deletion sentence. It is not a trap: the next backlog pass claims the row through `markInProgress`, which clears the error, and the re-ingest records its success normally.

And the record itself cannot displace what it describes. It is a transaction against a database that has just failed, so the failures reaching it — a lost connection, a lock timeout, a deadlock victim — are the ones most likely to make it fail too. Unguarded it threw out of `bulkDeleteCbom`: every uuid after the failing one went unattempted, the deletions already committed were reported nowhere, and the caller got a bare 500 instead of the per-entry list the method's contract is. `recordWithdrawnButNotDeleted` swallows and logs it, saying in the log line that the row will not be rebuilt until its state is corrected.

## The tombstone is tested twice, on purpose

The pre-check runs before the document read, and that read is an HTTP call. An operator deleting the document while it is in flight would otherwise have it stored straight back seconds later under a uuid of its own — which the delete cannot remove, and which nothing else notices, since `createCbomEntry`'s own pre-check looks at the live table only. So it is tested again inside the transaction that inserts, and that outcome is `StoreOutcome.TOMBSTONED`.

The feed pass also **resolves the skip** of an entry it settles by tombstone. That could not be left to the retry pass: the identity is in the run's attempted set by the time the tombstone is tested, and the retry pass returns for anything already attempted, so its own tombstone branch was never reached. Since the repository goes on offering a deleted document every run, such a skip row survived for ever.

## Known deviations

**A deleted superseding revision strands the one it superseded.** A CBOM naming no cryptographic assets acquires no source rows, so it is deletable today; delete it after it superseded an earlier revision and that earlier revision stays `SYNCED`, sourcing nothing, off both work lists for ever. This PR removes the row — it is the right place to notice, and it does not fix it: the repair is to re-offer the newest surviving version of the serial, which needs a work list this stack does not have yet. Recorded as open work on #2073, promised to this PR on #2257.

**A tombstone is permanent and has no operator-facing way out.** From the first deletion of a `(serial_number, version)`, a sync run can never store that identity again — no expiry, no endpoint that removes a row. A mistaken deletion is undone by uploading the document again through `createCbom`, which leaves the tombstone behind, or not at all. Related, and now said on `CbomTombstoneWriter.record`: because the insert swallows a conflict, a document deleted, re-uploaded and deleted again keeps the **first** tombstone, so `uuid`, `deleted_at` and `deleted_by` describe the first deletion rather than the latest. The row is a record that the document is not wanted, not an audit trail of who removed which copy of it. Recorded as open work on #2073; the weekly reconciliation PR is the natural place to give it a way out.

**The in-transaction tombstone test is a check-then-act.** It shrinks the window from the length of an HTTP read to the gap between one statement and the next in the same transaction; it does not close it. Closing it needs both the delete and the store to take a cluster lock keyed on the identity, which is a lock on the sync hot path and a decision worth making on its own.

**The wait for the asset-sync lock is not bounded per delete.** Each batch transaction carries `spring.transaction.default-timeout`, but a document is `ceil(sources / batch)` of them and `bulkDeleteCbom` multiplies that by the list length, so the total an operator's request can spend waiting has no ceiling, holding a pooled connection throughout. A `lock_timeout` surfacing as "another node is syncing this CBOM, try again" would bound it — at the cost of reintroducing the refusal `withdrawWaiting` exists to avoid, and of a new tunable that under SaaS belongs in the Settings subsystem rather than in yml. Recorded as open work on #2073.

## Verification

`mvn verify` against real PostgreSQL via Testcontainers.

Eleven tests were added or strengthened for the review findings: the locked re-read and the empty-list case, a source attached between two batches, both halves of the partial-withdrawal distinction, the lock-ranking `InOrder` on the blocking acquisition, the ingest-side probe for a header deleted mid-ingest, the attempt clock, the `PENDING` lower bound, the `markSynced` race and its recovery, a bulk delete against a real `RESTRICT` refusal, and the merged payload of an asset whose elected source was the deleted document.

**Not done:** Sonar new-lines coverage parity has not been computed locally, on this PR or on the earlier ones in the stack, so the gate's `new_coverage` figure is unverified ahead of CI.

No migration, no new configuration, no `interfaces` change.


src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -424,18 +520,25 @@ public List<BulkActionMessageDto> bulkDeleteCbom(List<UUID> uuids) {
                 continue;
             }
             try {
-                transactionHandler.runInNewTransaction(() -> cbomRepository.deleteById(uuid));
+                // Outside the catch below, which says the inventory was withdrawn. A withdrawal that threw withdrew
+                // nothing, and recording DELETION_WITHDREW_THE_INVENTORY for it would flip a genuinely SYNCED row to
+                // FAILED carrying a sentence that is not true of it -- one an operator can now search for.
+                assetDetachService.withdrawWaiting(uuid);
```
karolhrdina, line 619: **A withdrawal that threw did not necessarily withdraw nothing.**

The comment here — and the matching carve-out in `deleteWithdrawnRow` — rests on "A withdrawal that threw withdrew nothing, so the row still says what it said". `withdrawInBatches` commits **one transaction per batch**, and `CbomAssetDetachService`'s own class javadoc says so explicitly: *"the batches that already committed stay committed, their orphans included, and deleting an orphan is not reversible."*

So a `RuntimeException` on batch *k* (deadlock victim, lost connection, lock timeout, a failing `assetWriter.delete`) propagates out of `withdrawWaiting` with batches `1..k-1` already committed. Concretely: a `SYNCED` CBOM with 250 sources at `assetBatchSize = 100`; batch 3 throws; 200 sources are gone and their orphan assets are permanently deleted; this `catch` deliberately does **not** call `recordWithdrawnButNotDeleted`, so the row stays `SYNCED`, stays off `findPendingAssetIngests` and off `findAssetIngestRetries`, and no backlog pass ever rebuilds it. That is exactly the state `markFailedEvenIfSynced` was added for.

`testBulkDeleteCbom_withdrawalFailure_leavesTheRowsStateAlone` stubs `withdrawWaiting` to throw before it does anything, so it only pins the zero-batch case and cannot catch this.

The distinction the carve-out needs is not "threw vs. returned" but "withdrew nothing vs. withdrew something". `withdrawWaiting` could carry that — e.g. throw a checked/typed failure that wraps the partial `Withdrawal`, or have the detach service report what it committed before rethrowing — and the caller marks the row failed only when `detached + deleted + kept > 0`.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -380,11 +418,21 @@ public CbomDto createCbom(CbomUploadRequestDto request)
...
+     * one long-lived transaction holding every row it touched. It waits for the cluster lock rather than skipping --
+     * see {@link CbomAssetDetachService#withdrawWaiting}.
+     */
     @Override
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
     @ExternalAuthorization(resource = Resource.CBOM, action = ResourceAction.DELETE)
     public void deleteCbom(UUID uuid) throws NotFoundException {
         Cbom cbom = getEntity(SecuredUUID.fromUUID(uuid));
-        deleteRow(cbom);
+        assetDetachService.withdrawWaiting(uuid);
```
karolhrdina: **`deleteCbom` does not guard the withdrawal at all** — the same defect as the bulk path, one step worse.

If `withdrawWaiting` throws partway through (see the comment on the bulk path: batches commit individually, so a throw on batch *k* leaves `1..k-1` permanently withdrawn and their orphans deleted), the exception propagates straight out of `deleteCbom`:

- no `recordWithdrawnButNotDeleted`, so a `SYNCED` row keeps claiming an inventory contribution that is partly gone and no backlog pass will rebuild it;
- no shaped message — the operator gets whatever the detach service threw, which for a `DataIntegrityViolationException` is a driver message whose `DETAIL` line quotes the failing row (the leak `CryptoAssetConstraintTranslator` exists to prevent);
- no `OperationResult.FAILURE` audit entry, while `bulkDeleteCbom` logs one for the same failure via `recordBulkDeleteFailure`.

Wrapping this call the way the bulk path does — record the partial withdrawal, shape the message, log the failure — would close all three.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
```diff
@@ -111,12 +117,48 @@ Withdrawal plus(Withdrawal other) {
...
+    }
+
+    /** What a batch does when another node holds the withdrawn CBOM's asset-sync lock. */
+    private enum Contention {
+        SKIP,
+        WAIT
+    }
+
+    private Withdrawal withdrawInBatches(UUID cbomUuid, Contention contention) {
         final List<UUID> assets = sourceRepository.findAssetUuidsByCbomUuid(cbomUuid);
```
karolhrdina: **The asset list is read before any lock is taken, so `withdrawWaiting`'s promise is weaker than its javadoc.**

`findAssetUuidsByCbomUuid` runs here, outside any transaction and outside the asset-sync lock; only the per-batch transactions below take the lock. So "waiting is what makes `Withdrawal.complete()` true here by construction" is true only of *this snapshot*: any `crypto_asset_source` row an ingest of the same CBOM attaches after this read — including one it attaches in the gap between two batch commits, since the lock is transaction-scoped and released at each — is never withdrawn, and `complete() == true` is still returned.

The caller then hits `crypto_asset_source_to_cbom_key` on the header delete and, because of the `DataIntegrityViolationException` carve-out, the row is left `SYNCED` while sourcing only the subset that survived. Blocking for the lock buys nothing if the list it acts on was read before the lock.

Two shapes that would actually deliver the promise: re-read (or re-check-empty) inside the locked batch transaction, or loop `withdrawWaiting` until `findAssetUuidsByCbomUuid` comes back empty under the lock.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -393,20 +441,68 @@ public void deleteCbom(UUID uuid) throws NotFoundException {
...
+     * longer true, so the operator is told to retry instead. Every other failure leaves a row that says SYNCED and
+     * sources nothing, which is what that state exists for.
      */
-    private void deleteRow(Cbom cbom) {
+    private void deleteWithdrawnRow(UUID uuid) {
         try {
-            cbomRepository.delete(cbom);
-            cbomRepository.flush();
+            transactionHandler.runInNewTransaction(() -> deleteAndTombstone(uuid));
         } catch (DataIntegrityViolationException e) {
```
karolhrdina: **The `DataIntegrityViolationException` carve-out is broader than its justification.**

The reasoning is "something re-attached sources, so the row really is synced and only the delete failed". Two gaps:

1. **The premise is not guaranteed.** The withdrawal releases the asset-sync lock at every batch commit, and the asset list was read before any lock at all (see the comment on `withdrawInBatches`). The surviving sources can be a *partial* re-attachment, or the single source an ingest added after the list read — in which case the row is left `SYNCED` while sourcing a subset of its assets, and nothing repairs it: the re-ingesting run has already marked the row `SYNCED` and never looks again.
2. **It is not narrowed to the constraint it reasons about.** Any `DataIntegrityViolationException` takes this branch. `crypto_asset_source_to_cbom_key` is the only FK to `cbom` today, but the carve-out is stated as a general rule and will silently absorb the next one. `CryptoAssetConstraintTranslator.constraintNameOf(e)` is already used for exactly this kind of narrowing in `store()` (`DEDUP_CONSTRAINT::equalsIgnoreCase`) — the same guard here would make the carve-out say what it means.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -69,9 +69,15 @@
...
  * <p>
  * <b>Lock ranking, third rank included.</b> The asset-sync lock is taken above {@code ALIAS_DECISION_LOCK}, which is
- * above every {@code crypto_asset} row lock. The asset-sync lock is only ever acquired with a non-blocking
- * {@code tryLock}, so a node that cannot get it abandons the batch rather than joining a wait chain; acquiring it with
- * a blocking call would close a cycle against any path that takes the two in the other order.
+ * above every {@code crypto_asset} row lock. On this path it is acquired with a non-blocking {@code tryLock}, so a node
+ * that cannot get it abandons the batch rather than joining a wait chain.
+ *
+ * <p>
+ * It may also be acquired <b>blocking</b> -- {@link CbomAssetDetachService#withdrawWaiting(UUID)} does, for a caller
```
karolhrdina, line 76: **The ranking is now safe, but the complementary check on the ingest side is still missing.**

The deadlock argument holds — I checked every `lock(...)` site, and nothing takes `ALIAS_DECISION_LOCK` or a `crypto_asset` row lock and then waits on the asset-sync lock. But this PR makes the opposite race reachable for the first time, and the fix is one line away from code that already does the same job.

`writeBatchUnderClusterLock` re-reads `hasIngestedLaterVersion(cbomUuid)` under the lock, for precisely this reason: the lock is released at every batch commit and the world can change in the gap. It does **not** re-read whether the `cbom` header still exists. After `deleteAndTombstone` commits, an ingest batch for the same document passes the lock and the supersede test and inserts `crypto_asset_source` rows against a `cbom_uuid` that is gone — a foreign-key violation, a noisy failed document, and a `markFailed` that updates zero rows.

Note that the window is not narrow: a CBOM whose `findAssetUuidsByCbomUuid` returns empty takes **no lock at all** (`withdrawInBatches` never enters the loop), so a delete of a CBOM mid-first-ingest races the ingest with no exclusion whatsoever.

An `existsById`/`existsByUuid` probe next to the existing `hasIngestedLaterVersion` call closes it at the same cost as the check already there.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
```diff
@@ -126,8 +168,11 @@ public Withdrawal withdraw(UUID cbomUuid) {
     }
 
     /** One batch, inside the transaction that holds the cluster lock. Null when another node holds it. */
-    private Withdrawal withdrawBatchUnderClusterLock(UUID cbomUuid, List<UUID> assets) {
-        if (!clusterSynchronizer.tryLock(CbomAssetIngestService.assetSyncLockKey(cbomUuid))) {
+    private Withdrawal withdrawBatchUnderClusterLock(UUID cbomUuid, List<UUID> assets, Contention contention) {
+        final String lockKey = CbomAssetIngestService.assetSyncLockKey(cbomUuid);
+        if (contention == Contention.WAIT) {
+            clusterSynchronizer.lock(lockKey);
```
karolhrdina: **An unbounded blocking wait now sits on a request thread.**

`ClusterOperationSynchronizer.lock` is `pg_advisory_xact_lock` with no timeout, and nothing in the deployment sets `lock_timeout` or `statement_timeout`. `deleteCbom`/`bulkDeleteCbom` are synchronous API calls, so a delete blocks an HTTP worker and a pooled connection for as long as the holder runs — and the holder is an ingest batch that itself blocks on the **cluster-global** `ALIAS_DECISION_LOCK` (line 180 below), which every ingest batch on every node also queues on. `bulkDeleteCbom` pays that wait once per uuid, serially, and per batch within each.

`Contention.WAIT` is the right call for the semantics; a wait with no ceiling is what worries me. A bounded form keeps both properties: `SET LOCAL lock_timeout` around the acquisition, or a bounded retry over `tryLock`, and a delete that genuinely cannot get the lock in N seconds refuses with a shaped message rather than holding a worker indefinitely.

src/test/java/com/otilm/core/cbom/ingest/CbomAssetDetachServiceTest.java
```diff
@@ -178,6 +178,18 @@ void anOrphanAnAliasPointsAtIsKept() {
...
+    /** The keyed caller waits instead of skipping, or another node's sync would read as a refused deletion. */
+    @Test
+    void anOperatorsDeleteWaitsForTheClusterLockRatherThanSkippingTheWithdrawal() {
+        UUID asset = sourcedAsset();
+
+        service(100).withdrawWaiting(CBOM);
+
+        verify(synchronizer).lock(CbomAssetIngestService.assetSyncLockKey(CBOM));
+        verify(synchronizer, never()).tryLock(anyString());
+        verify(sourceWriter).detachCbom(asset, CBOM);
```
karolhrdina, line 193: **The invariant this test exists to protect is the one it does not assert.**

The safety of the new blocking acquisition rests entirely on the claim in the javadoc above and on `CbomAssetIngestService`: *"the wait must be the **first** lock of its transaction, above `ALIAS_DECISION_LOCK`"*. This test asserts that `lock(assetSyncLockKey)` happens and that `tryLock` does not — neither of which would fail if someone moved the `lock(ALIAS_DECISION_LOCK)` call above it and closed the cycle.

This class's own header says it pins "the one ordering a database would reveal only as an intermittent deadlock", and `theAliasLockIsTakenBeforeTheFirstAssetRowLock` uses `InOrder` for exactly that. One more line here does the same for the new rank:

```suggestion
    /** The keyed caller waits instead of skipping, or another node's sync would read as a refused deletion. */
    @Test
    void anOperatorsDeleteWaitsForTheClusterLockRatherThanSkippingTheWithdrawal() {
        UUID asset = sourcedAsset();

        service(100).withdrawWaiting(CBOM);

        verify(synchronizer).lock(CbomAssetIngestService.assetSyncLockKey(CBOM));
        verify(synchronizer, never()).tryLock(anyString());
        verify(sourceWriter).detachCbom(asset, CBOM);
        // The blocking acquisition must be the FIRST lock of the transaction: a waiter that already held
        // ALIAS_DECISION_LOCK would close a cycle against any ingest holding the asset-sync lock.
        InOrder ranked = inOrder(synchronizer);
        ranked.verify(synchronizer).lock(CbomAssetIngestService.assetSyncLockKey(CBOM));
        ranked.verify(synchronizer).lock(CryptoAssetAliasWriter.ALIAS_DECISION_LOCK);
    }
```

src/main/java/com/otilm/core/service/writer/cbom/CbomAssetSyncStateWriter.java
```diff
@@ -125,6 +125,23 @@ public int markSuperseded(UUID cbomUuid) {
...
+     * For the one caller entitled to overrule a success: a deletion withdraws the CBOM's contribution to the inventory
+     * before removing the header, so a deletion that fails in between leaves a row that says SYNCED and sources
+     * nothing. The guard {@link #markFailed} applies is there to stop a stale run from overwriting another node's
+     * success; here the row genuinely is no longer ingested, and leaving it SYNCED would keep the backlog pass from
+     * ever rebuilding it.
+     *
+     * @param operatorSafeError see {@link #markFailed}
+     */
+    @Transactional
+    public int markFailedEvenIfSynced(UUID cbomUuid, String operatorSafeError) {
```
karolhrdina: **This write stamps `asset_sync_attempted_at`, which delays the rebuild it exists to trigger.**

`updateAssetSyncState` sets `assetSyncAttemptedAt = CURRENT_TIMESTAMP` unconditionally. `findAssetIngestRetries` only offers a `FAILED` row once `assetSyncAttemptedAt < retryBefore`, so a row marked here sits out a full `cbom.sync.ingest-retry-after` (30 minutes by default) before the backlog pass touches it — while sourcing nothing at all, which is the one state where the retry window buys nothing. The row's assets are known to be gone; there is no "the run that claimed it might still be slow" to wait out.

The javadoc says "leaving it SYNCED would keep the backlog pass from ever rebuilding it", which is right, but the row still is not rebuilt for half an hour. Worth either saying so here, or writing `NEVER_ATTEMPTED` for this one caller so the row lands at the head of the retry list instead of the tail.

src/main/java/com/otilm/core/service/writer/cbom/CbomTombstoneWriter.java
```diff
@@ -15,10 +15,10 @@
...
  *
  * <p>
- * <b>No production caller yet.</b> The delete path must record a tombstone as it deletes, or the next sync finds the
- * document upstream, sees nothing in the live {@code cbom} table, and re-ingests exactly what an operator removed —
- * which is the scenario this table exists to prevent. That wiring arrives with the ingest ticket, alongside the source
- * detachment the RESTRICT foreign key requires.
+ * The delete path records one as it deletes, or the next sync finds the document upstream, sees nothing in the live
+ * {@code cbom} table, and re-ingests exactly what an operator removed — which is the scenario this table exists to
+ * prevent. {@code CbomServiceImpl} is the caller, and reads it back through
+ * {@code CbomTombstoneRepository.existsBySerialNumberAndVersion} before storing a feed entry.
```
karolhrdina: **Two other javadocs still say this PR has not happened.** This one was correctly updated; these were missed, and both now state the opposite of the code:

`src/main/java/com/otilm/core/service/writer/cbom/CryptoAssetSourceWriter.java:119` (on `detachCbom`):
> **The API's delete path still does not call this.** `CbomServiceImpl` must withdraw every asset a CBOM contributes to before deleting the row, because `crypto_asset_source_to_cbom_key` is RESTRICT; until it does, a CBOM that has acquired a source cannot be deleted through the API at all. […] The supersede path here is the first caller of this unit of work, not the deletion lifecycle.

`CbomServiceImpl` is now a caller, through `CbomAssetDetachService.withdrawWaiting`, and the deletion lifecycle is exactly what this PR wires.

`src/main/java/com/otilm/core/config/CbomSyncProperties.java:27` (on `assetIngestEnabled`):
> […] a CBOM that has them cannot be deleted through the API until the deletion lifecycle lands. **Turning this off is what stops that spreading**, and it is why the value carries a deploy-time override.

The stated justification for the kill switch no longer exists. In a codebase where these comments are load-bearing, a stale one is worse than none — the next reader will turn ingest off for a reason that has been fixed.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -680,7 +800,14 @@ private void processFeedEntry(BomEntryDto entry, SyncRun run, Map<SyncIdentity,
                     .debug("CBOM Sync: CBOM serialNumber {} and version {}: already exists. Skipping the sync",
                             identity.serialNumber(), identity.version());
             run.duplicates++;
-            resolveSkipIfRecorded(identity, skips, run);
+            resolveSkipIfRecorded(identity, skips, run, "is stored");
+            return;
+        }
+        if (isTombstoned(identity, run)) {
```
karolhrdina: **Two existence probes per feed entry, where one round trip would do.**

The feed pass already pays `cbomRepository.existsBySerialNumberAndVersion` per entry; this adds a second per-entry query against `cbom_tombstone`, for every entry of every page of every run — 2000 round trips on a full 1000-entry page instead of 1000. And a tombstoned identity is, by this PR's own premise, re-offered and re-probed on every run for as long as it stays inside the listing window, so the cost does not decay.

Both probes are set-shaped and the page is already in hand. One `IN`-list query per page against each table (or one against both), resolved into a `Set<SyncIdentity>` before the loop, gives the same answers in two round trips per page. `CbomAssetDetachService` already makes this trade explicitly — *"Once for the batch […] asking them per asset would put two more round trips per asset inside the transaction"*.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -938,10 +1089,32 @@ private SkipOutcome recordSkip(SyncIdentity identity, String reason, CbomHeaderC
...
 
     /**
-     * Deletes the entry's skip row, if it has one, now that the document is stored. The live retry set answers for
-     * {@code RETRYING} rows; a written-off row is looked up, so it too is resolved when its document finally arrives.
+     * Whether an operator deleted this identity. The repository goes on offering the document -- deletion here is not
+     * deletion there -- so without this test every run would re-store what the operator removed, and the next ingest
+     * would rebuild the inventory behind it.
      */
-    private void resolveSkipIfRecorded(SyncIdentity identity, Map<SyncIdentity, CbomSyncSkip> skips, SyncRun run) {
+    private boolean isTombstoned(SyncIdentity identity, SyncRun run) {
```
karolhrdina: **A predicate with side effects, and the third copy of the same outcome.**

`isTombstoned` reads as a question but increments `run.tombstoned` and writes a log line, so the two call sites cannot test the tombstone without also recording a run outcome — and `store()` has its own separate `run.tombstoned++` plus a near-identical debug line for the in-transaction case. Three places now encode "an operator deleted this; count it and say so".

Folding the counting into the outcome the callers already switch on — have the tombstone test return a `StoreOutcome`/boolean and let one place own `run.tombstoned++` and the log — removes the duplication and makes the predicate safe to call from a future third site without silently double-counting.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
```diff
@@ -111,12 +117,48 @@ Withdrawal plus(Withdrawal other) {
...
+     * Blocking on that key carries a ranking obligation, and this is the only path that takes it blocking: the wait
+     * must be the <b>first</b> lock of its transaction, above {@code ALIAS_DECISION_LOCK} and above every
+     * {@code crypto_asset} row lock, so that a waiter holds nothing a holder could go on to want. See the lock ranking
+     * on {@link CbomAssetIngestService}.
+     */
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    public Withdrawal withdrawWaiting(UUID cbomUuid) {
+        final Withdrawal withdrawn = withdrawInBatches(cbomUuid, Contention.WAIT);
+        if (!withdrawn.complete()) {
+            throw new IllegalStateException("A waiting withdrawal cannot be abandoned");
```
karolhrdina: **This branch is unreachable.** With `Contention.WAIT`, `withdrawBatchUnderClusterLock` never returns `null` — the `return null` is behind the `tryLock` arm only — so `withdrawInBatches` never reaches `Withdrawal.ABANDONED` and `complete()` is always `true`.

The javadoc calls it "the invariant, not a case the caller is expected to handle", which is fair, but as written it is a guard no input can trip, so nothing pins the invariant either. If it is meant as documentation, an `assert`/comment says that more honestly; if it is meant as a tripwire for a future `Contention` value, it belongs where the contention is decided, next to the `if (done == null)`.

src/test/java/com/otilm/core/integration/cbom/CryptoAssetInventoryITest.java
```diff
@@ -445,41 +445,58 @@ void aCbomStillReferencedByTheInventoryCannotBeDeletedAndTheRefusalLeaksNothing(
...
+        sourceWriter
+                .upsertSource(assetUuid, leanCbom.getUuid(), Map.of("primitive", "signature"), List.of(),
+                        OffsetDateTime.now());
+        sourceWriter
+                .upsertSource(assetUuid, richCbom.getUuid(), Map.of("primitive", "signature"), List.of(),
+                        OffsetDateTime.now());
+
+        cbomService.deleteCbom(leanCbom.getUuid());
+
+        assertThat(asset(assetUuid).getSourceCount()).isEqualTo(1);
```
karolhrdina: **The test asserts less than its javadoc claims.** The comment promises *"with the deleted document's contribution gone from its merged payload"*, but the only assertion is `getSourceCount() == 1`.

The merged payload is the interesting half here: `detachCbom` calls `recomputeMergeFromSources`, and the `crypto_asset_to_properties_source_key` FK is `ON DELETE SET NULL`, so an asset whose `properties_source_uuid` pointed at the *deleted* CBOM's source row is exactly the case where re-election must have happened rather than the FK having quietly nulled the pointer. Neither `mergedCryptoProperties` nor `propertiesSourceUuid` is checked, so the test passes either way.

Giving the two CBOMs distinguishable payloads and asserting the survivor's merged payload is the surviving one would make the javadoc true.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -393,20 +441,68 @@ public void deleteCbom(UUID uuid) throws NotFoundException {
...
+                    .error("CBOM {}: its cryptographic assets were withdrawn, the deletion then failed, and recording that failed too. The row still reads SYNCED and sources nothing; no backlog pass will rebuild it until its state is corrected.",
+                            uuid, e);
         }
     }
 
+    /**
+     * The tombstone commits with the deletion or with neither. Without it the next sync finds the document still
+     * offered by the repository, sees nothing in the live table, and re-ingests exactly what the operator removed.
+     */
+    private void deleteAndTombstone(UUID uuid) {
```
karolhrdina, line 589: **A tombstone is permanent and has no operator-facing way out.**

This makes `cbom_tombstone` load-bearing for the first time: from here on, a `(serial_number, version)` that has been deleted once can never be stored by a sync run again — the pre-check, the in-transaction check, and the retry pass all refuse it, permanently, with no expiry and no endpoint that removes a row. A mistaken delete is unrecoverable through the API; the only remedies are a manual `DELETE FROM cbom_tombstone` or re-uploading the document by hand through `createCbom`, which leaves the stale tombstone behind.

Related: because `recordTombstone` is `ON CONFLICT DO NOTHING` over both constraints, a CBOM deleted, re-uploaded under a new uuid with the same serial and version, and deleted again keeps the **first** tombstone — so `uuid`, `deleted_at` and `deleted_by` describe a different deletion than the one that just happened, and the second deletion has no record of its own. The repository javadoc argues that swallowing the conflict is right (it is), but the audit consequence is not stated anywhere and `deleted_by` cannot be read as "who deleted this".

Worth at least recording alongside the other known deviations; the weekly reconciliation PR is the natural place to give it a way out.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -393,20 +441,68 @@ public void deleteCbom(UUID uuid) throws NotFoundException {
...
+     * The write is a transaction of its own against a database that has just failed, so the failures that reach it are
+     * the ones most likely to make it fail too -- a lost connection, a lock timeout, a deadlock victim, an exhausted
+     * pool. Left unguarded it would replace the caller's outcome: the operator's shaped refusal becomes a bare 500, and
+     * one entry's failure in {@link #bulkDeleteCbom} throws out of the loop, abandoning every uuid after it and
+     * reporting none of the deletions already committed. The row then reads SYNCED while sourcing nothing and no
+     * backlog pass will rebuild it, which is why the log line says so.
+     */
+    private void recordWithdrawnButNotDeleted(UUID uuid) {
+        try {
+            assetSyncStateWriter.markFailedEvenIfSynced(uuid, DELETION_WITHDREW_THE_INVENTORY);
```
karolhrdina: **`markSynced` can undo this write immediately — the race is only closed in one direction.**

`markSynced` is itself unconditional: `updateAssetSyncState(cbomUuid, SYNCED, null, syncedAt, null)` (`CbomAssetSyncStateWriter.java:101-103`), called in its own transaction from `CbomAssetIngestService.java:265`, and it also clears `asset_sync_error`. Neither writer holds the asset-sync lock when it runs — `pg_advisory_xact_lock` is transaction-scoped (`ClusterOperationSynchronizer.java:84-89`), the withdrawal drops it at every batch commit, and both delete paths are `NOT_SUPPORTED`.

Interleaving:

1. Node B is ingesting the same CBOM. It has written its last `crypto_asset_source` batch and released the lock, and is between that and `markSynced` — e.g. inside `hasIngestedLaterVersion` / `withdrawSupersededVersions` (`CbomAssetIngestService.java:254-257`).
2. Node A's withdrawal takes the lock, detaches B's rows, commits.
3. A's header delete fails; `recordWithdrawnButNotDeleted` writes `FAILED` + the sentence.
4. B reaches `markSynced`. Unconditional, so the row goes back to `SYNCED`, `assets_synced_at` is stamped, and `asset_sync_error` is nulled.

The row now reads `SYNCED`, sources nothing, and carries no error. `findPendingAssetIngests` matches `PENDING` only and `findAssetIngestRetries` matches `{IN_PROGRESS, FAILED}` (`CbomServiceImpl.java:112-113`), so it is on neither work list and nothing ever rebuilds it — exactly the outcome the javadoc at `:476-477` says this method exists to prevent.

The milder version of the same gap: if B fails instead of succeeding it calls `markFailed`, whose guard `NOT_YET_SYNCED` *includes* `FAILED`, so it silently replaces `DELETION_WITHDREW_THE_INVENTORY` with an unrelated ingest reason. Nothing is stranded there, but the operator-facing sentence is lost.

Closing this needs the two unconditional writers to stop racing — e.g. have `markSynced` refuse to overwrite a `FAILED` row carrying the deletion sentence, or take the asset-sync lock around the failure write.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -424,18 +520,25 @@ public List<BulkActionMessageDto> bulkDeleteCbom(List<UUID> uuids) {
                 continue;
             }
             try {
-                transactionHandler.runInNewTransaction(() -> cbomRepository.deleteById(uuid));
+                // Outside the catch below, which says the inventory was withdrawn. A withdrawal that threw withdrew
+                // nothing, and recording DELETION_WITHDREW_THE_INVENTORY for it would flip a genuinely SYNCED row to
+                // FAILED carrying a sentence that is not true of it -- one an operator can now search for.
```
karolhrdina: **`CBOM_ASSET_SYNC_ERROR` does not exist.**

`grep -rn "CBOM_ASSET_SYNC_ERROR" src/` returns exactly two hits, both added by this PR: this comment, and the javadoc at `CbomServiceITest.java:1795`.

`FilterField` has `CBOM_ASSET_SYNC_STATE` (`:392`) and `CBOM_ASSETS_SYNCED_AT` (`:394`), and nothing for the error column. The searchable-field list at `CbomServiceImpl.java:599-612` offers no such field, and `Cbom.mapToDto()` (`Cbom.java:96-114`) copies `assetSyncState` and `assetsSyncedAt` but not `assetSyncError`. The sentence is therefore reachable only through direct database access, or from the ERROR log line at `:485-486` on the total-failure path.

This is load-bearing, not just a stale comment: searchability is the stated reason for the carve-out — *"would flip a genuinely SYNCED row to FAILED carrying a sentence that is not true of it — one an operator can now search for."* Either add the filter field and the DTO field, or drop the claim from both places and rest the carve-out on the state flip alone.

src/main/java/com/otilm/core/service/writer/cbom/CbomAssetSyncStateWriter.java
karolhrdina: Two effects of this write that the javadoc does not account for.

**It re-stamps the attempt clock, delaying the rebuild it is asking for.** `updateAssetSyncState` sets `c.assetSyncAttemptedAt = CURRENT_TIMESTAMP` on every call (`CbomRepository.java:235`), and the retry work list is gated on `assetSyncAttemptedAt IS NULL OR assetSyncAttemptedAt < :retryBefore`, with `retryBefore = runStart − cbom.sync.ingest-retry-after` (default 30m, `CbomServiceImpl.java:1291`). So the row this method just put on the backlog is excluded from it for a full retry window, and `ORDER BY assetSyncAttemptedAt NULLS FIRST` then sorts it behind every older candidate. With the hourly cron (`CbomSyncTask.java:24`, `0 0 * ? * *`), a delete that fails at 10:45 is skipped by the 11:00 run and first picked up at 12:00. This is not an ingest attempt, so the stamp buys nothing and costs the row its place in the queue — the row that most urgently owes a rebuild is the one made to wait longest.

**It has no lower bound either.** With `expectedStates = null` the write also lands on `PENDING`. A CBOM whose assets were never ingested has nothing to withdraw — `withdrawWaiting` returns `NOTHING` — and a failing header delete then flips `PENDING` → `FAILED`, moving the row off the fast `findPendingAssetIngests` list onto the 30-minute retry list and attaching a sentence that is untrue of it: nothing was withdrawn. The PR guards the analogous case for a withdrawal that *threw* (`CbomServiceImpl.java:522-530`) but not for one that legitimately found nothing. Guarding on `{IN_PROGRESS, SYNCED, FAILED}` — or on the `Withdrawal` having actually detached something — would cover it.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
karolhrdina, line 533: This treats **any** `DataIntegrityViolationException` as "an ingest re-attached sources, so the row really is SYNCED".

That is correct today, and I checked it rather than assuming: `crypto_asset_source_to_cbom_key` is the only foreign key in the entire schema that references `cbom(uuid)` (`V202608271000__crypto_asset_inventory.sql:119-120`; no other `REFERENCES "cbom"` in any migration), and the tombstone insert is `ON CONFLICT DO NOTHING` with no arbiter, so nothing else inside `deleteAndTombstone` can raise one.

But nothing enforces or tests that invariant, and the failure mode when it lapses is the silent one. The next FK added to `cbom` — or an `ON CONFLICT` dropped from `recordTombstone` — turns this branch into "leave the row SYNCED, sourcing nothing, off both work lists", with an inventory-worded message on top: `describe` falls back to `GENERIC_REJECTION`, which also begins *"The cryptographic asset inventory rejected the change"*.

Cheap fix, and the same shape `store()` already uses for `DEDUP_CONSTRAINT` at `:933-945`:

```java
} catch (DataIntegrityViolationException e) {
    if (CryptoAssetConstraintTranslator.constraintNameOf(e)
            .filter("crypto_asset_source_to_cbom_key"::equalsIgnoreCase)
            .isEmpty()) {
        recordWithdrawnButNotDeleted(uuid);
    }
    throw new ValidationException(ValidationError.create(CryptoAssetConstraintTranslator.describe(e)));
}
```

The mirror at `:538` wants the same narrowing.

src/test/java/com/otilm/core/integration/cbom/CryptoAssetInventoryITest.java
```diff
@@ -445,41 +445,58 @@ void aCbomStillReferencedByTheInventoryCannotBeDeletedAndTheRefusalLeaksNothing(
...
         List<BulkActionMessageDto> messages = cbomService.bulkDeleteCbom(List.of(leanCbom.getUuid()));
 
-        assertThat(messages).singleElement().satisfies(message -> {
-            assertThat(message.getMessage()).contains("still referenced by the cryptographic asset inventory");
-            assertLeaksNothing(message.getMessage(), identityKey);
-        });
-        assertThat(cbomRepository.findById(leanCbom.getUuid()))
-                .describedAs("a refused deletion leaves the row where it was")
-                .isPresent();
+        assertThat(messages).isEmpty();
```
karolhrdina, line 477: The test replaced here was the only cover for the bulk path's `DataIntegrityViolationException` branch, and nothing picked it up.

After this PR every `bulkDeleteCbom` test throws a plain `RuntimeException` (`CbomServiceITest.java:1808`, `:1853`, `:1908`), and this one asserts `messages` is empty. So:

* `recordBulkDeleteFailure`'s `CryptoAssetConstraintTranslator.describe(ex)` arm (`CbomServiceImpl.java:560-562`) is never exercised — only the `"Error deleting CBOM entry"` arm is;
* the `if (!(ex instanceof DataIntegrityViolationException))` skip at `:538-540` — the decision that leaves the row `SYNCED` — is only ever taken in the true direction.

The single-delete counterpart, `aDeleteTheInventoryRefusesLeavesTheRowSynced`, does not fill the gap: it stubs `cbomRepositorySpy.delete(any(Cbom.class))` with a synthetic violation, so no delete is attempted and nothing rolls back, and it asserts only `contains(...)` on the message. No surviving test shows that a *real* refusal leaves the CBOM row in place.

A bulk twin of the test that was removed — real source row, real `RESTRICT` refusal, assert the translated message and that the row is still present — would cover both, and is close to what was deleted.

(For the record the "no driver DETAIL leak" property itself is not lost. `CryptoAssetConstraintTranslatorTest` pins it per constraint against a planted DETAIL line, and `aCbomStillReferencedByTheInventoryCannotBeDeletedAndTheRefusalLeaksNothing` at `:420` still pins it end-to-end against a live driver exception.)

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
karolhrdina, line 198: This wait is unbounded, and it holds a pooled connection for its whole duration.

`lock` is `pg_advisory_xact_lock` (`ClusterOperationSynchronizer.java:84-89`). No `lock_timeout` or `statement_timeout` is configured anywhere in main config — both appear only in one test and in comments — and there is no HTTP request-processing timeout either (`spring.mvc.async.request-timeout` unset, `server.tomcat.connection-timeout` unset). The only bound is `spring.transaction.default-timeout: 120` (`application.yml:343`), and it applies **per batch**: `withdrawInBatches` runs `ceil(sources / 100)` of them, each with a fresh budget, so the total wait for one delete has no ceiling, and `bulkDeleteCbom` multiplies it by the list length. Virtual threads mean no platform thread is pinned, but the Hikari connection is held throughout (pool 40, `application.yml:255`).

And because the lock is released at every batch commit, a same-CBOM ingest on another node can retake it in each gap, so the two ping-pong rather than the delete progressing monotonically.

A bounded wait would be friendlier than an unbounded one: `SET LOCAL lock_timeout` on this acquisition, surfacing as a shaped *"another node is syncing this CBOM, try again"*. That gives up less than it appears to, because under contention this path can still end in the `RESTRICT` refusal anyway — see the thread on the source-list snapshot at `:159`.

karolhrdina (commented):
Reviewed with the stack in hand (#2256, #2257). The shape is right — withdraw, then delete-and-tombstone in one transaction, with the tombstone read back on both sync passes — and several of the load-bearing claims in the description check out when tested rather than taken on trust:

* the lock ranking holds. `withdrawWaiting`'s blocking acquisition really is the first lock of its transaction, and no path in the codebase takes `ALIAS_DECISION_LOCK` or a `crypto_asset` row lock and then waits on the asset-sync key;
* the tombstone insert is `ON CONFLICT DO NOTHING` with no arbiter, so the delete → re-upload → delete cycle cannot throw;
* `AuditorAware<String>` is unambiguous and returns `"system"` rather than empty off-request; `deleted_by` is nullable either way;
* the `NOT_SUPPORTED` change on `deleteCbom` costs nothing — the class-level `@Transactional` was the only transaction in play, the sole caller is not transactional, `Cbom` has no lazy associations, and `logEvent` is SLF4J-only and never was transactional;
* no dependency cycle in the new setter injections.

Six findings below, plus two replies on Copilot's existing threads — it found the partial-withdrawal problem independently and it does reproduce; I've added the evidence and what I think the fix should be.

The two I'd want resolved before merge:

1. **`markSynced` is unconditional, so `markFailedEvenIfSynced` closes only one direction of the race.** A concurrent ingest finishing just after the delete's failure write puts the row straight back to `SYNCED` with `asset_sync_error` cleared — the exact stranded state this PR adds the method to prevent.
2. **The "one an operator can now search for" justification is not backed by anything.** `CBOM_ASSET_SYNC_ERROR` does not exist; `assetSyncError` is on the entity only, absent from `CbomDto` and from `FilterField`. That sentence is the stated reason for the carve-out, so either the field lands or the reasoning changes.

The rest: a latent mis-attribution in the `DataIntegrityViolationException` catch, a rebuild delayed by up to a retry window, a `PENDING` row demoted and mislabelled, and one real coverage regression on the bulk path.

Checked and **not** findings, so nobody re-derives them: the `Error`-bypasses-recovery path in `deleteWithdrawnRow` (an `Error` means the JVM is already gone, and the bulk path makes the same call); reporting a concurrently-deleted uuid as a bulk success (`deleteById` behaved identically on `main`); the tombstone insert as a source of `DataIntegrityViolationException`; and the "no driver DETAIL leak" property, which survives in `CryptoAssetConstraintTranslatorTest` and in `CryptoAssetInventoryITest:420`.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
```diff
@@ -111,12 +117,48 @@ Withdrawal plus(Withdrawal other) {
...
+        }
```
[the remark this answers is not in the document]
karolhrdina: Confirming this one — it reproduces on the code as written, and the consequence is worse than "hidden from the backlog": it is unrecoverable without manual SQL.

Evidence:

* Each batch is its own `REQUIRES_NEW` transaction (`TransactionHandler.runInNewTransaction`), and `cbom.sync.asset-batch-size` defaults to **100** (`application.yml:428`). So any CBOM with more than 100 sources is exposed to a throw in batch *N>1*, and the class comment already states the consequence — *"the batches that already committed stay committed, their orphans included, and deleting an orphan is not reversible"* (`CbomAssetDetachService.java:59-64`).
* The row is then left `SYNCED`, and nothing ever reselects it. The backlog pass draws from exactly two queries: `findPendingAssetIngests(PENDING, …)` and `findAssetIngestRetries(INGEST_RETRY_STATES, …)` with `INGEST_RETRY_STATES = {IN_PROGRESS, FAILED}` (`CbomServiceImpl.java:112-113`, `:1288-1295`). `SYNCED` is in neither. That is the precise state `markFailedEvenIfSynced` was added to prevent, and this path is the one place it is deliberately not called.
* The single-delete path is the weaker of the two: `deleteCbom` has no `try` around `withdrawWaiting` at all (`CbomServiceImpl.java:434`), so a partial withdrawal there is not recorded, not marked, and not even reported per-entry the way `bulkDeleteCbom` manages.
* The new test cannot catch it. `testBulkDeleteCbom_withdrawalFailure_leavesTheRowsStateAlone` stubs the whole method — `doThrow(...).when(detachServiceSpy).withdrawWaiting(savedUuid)` — so it only pins the zero-batches-committed case, which is the case the comment in `bulkDeleteCbom` describes correctly. Nothing exercises a throw from *inside* `withdrawInBatches`.

On the fix: of the two options you list, "conservatively mark the row failed on withdrawal errors" looks clearly preferable here. A needless re-ingest is idempotent and costs one document read; a `SYNCED` row sourcing nothing is invisible to every work list and to the operator. If the needless re-ingest is the thing worth avoiding, the narrower version is to have `withdrawInBatches` wrap a mid-loop failure in an exception carrying the partial `Withdrawal`, and let the caller mark failed only when something actually committed.

Either way the comment at `CbomServiceImpl.java:523-525` ("A withdrawal that threw withdrew nothing") needs to go — it is true only for a single-batch CBOM.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
```diff
@@ -111,12 +117,48 @@ Withdrawal plus(Withdrawal other) {
...
+    private enum Contention {
+        SKIP,
+        WAIT
+    }
+
+    private Withdrawal withdrawInBatches(UUID cbomUuid, Contention contention) {
         final List<UUID> assets = sourceRepository.findAssetUuidsByCbomUuid(cbomUuid);
         Withdrawal total = Withdrawal.NOTHING;
         for (int from = 0; from < assets.size(); from += batchSize) {
             final List<UUID> batch = assets.subList(from, Math.min(from + batchSize, assets.size()));
```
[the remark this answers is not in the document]
karolhrdina: Confirming, and the suggested fix does not go far enough.

`clusterSynchronizer.lock` is `pg_advisory_xact_lock` (`ClusterOperationSynchronizer.java:84-89`) — **transaction**-scoped — and every batch is its own `REQUIRES_NEW` transaction. So acquiring the lock before the `findAssetUuidsByCbomUuid` read and re-reading under it would close two of the four windows, not all of them:

1. between the snapshot read (`:156`) and the first batch's acquisition (`:174`) — closed by your suggestion;
2. each inter-batch gap, because the lock is released at every batch commit — closed only if the list is re-read per batch;
3. **after the last batch commits and before the header delete runs** — not closed by either. `deleteWithdrawnRow` opens its own `runInNewTransaction` (`CbomServiceImpl.java:459`) and holds no asset-sync lock while `deleteAndTombstone` deletes and flushes (`:494-504`);
4. a CBOM with zero sources, where the loop body never executes and no lock is taken at all — your case.

Window 3 is also where the PR's own third "known deviation" lives (an ingest writing `crypto_asset_source` rows against a header that was just deleted). Both directions of that race close with the same change: hold the asset-sync lock across the header delete as well, i.e. take it once in the transaction that runs `deleteAndTombstone`, or re-check `findAssetUuidsByCbomUuid` under it there.

Worth noting the irony against the stated motivation for `withdrawWaiting` — *"a skipped withdrawal would turn the delete into a refusal for as long as another node happened to be syncing"* (`CbomAssetDetachService.java:128-131`). Waiting removes the `complete == false` refusal but not the refusal: under a concurrent same-CBOM ingest the delete can still fail on the `RESTRICT` FK via window 3 or 4, now after paying an unbounded wait and after irreversibly deleting orphan assets in the batches that did commit.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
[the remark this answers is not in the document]
vyskocilm: Fixed. The distinction the code now makes is "withdrew nothing" rather than "threw".

`withdrawInBatches` wraps a mid-loop failure in `CbomAssetDetachService.WithdrawalFailedException`, which carries the `Withdrawal` the committed batches produced and answers `withdrewSomething()`. Both delete paths record `markWithdrawnButNotDeleted` for exactly the withdrawals that changed the inventory, and for no others; `deleteCbom` no longer calls the withdrawal bare at all (see the thread on `CbomServiceImpl:434`).

The comment you flagged at `CbomServiceImpl:523-525` is gone, along with the claim it rested on.

Two tests, both in `CbomAssetDetachServiceTest`, since you are right that stubbing `withdrawWaiting` whole cannot reach this:

* `aFailurePartwayThroughCarriesWhatTheCommittedBatchesWithdrew` — batch 1 commits a detach and an orphan deletion, batch 2 throws; the exception carries `Withdrawal(1, 1, 0, true)`;
* `aFailureBeforeTheFirstCommitReportsAnUntouchedInventory` — the zero-batch case, which is the one the carve-out is for.

And `CbomServiceITest.aWithdrawalThatEmptiedPartOfTheInventoryBeforeFailingLeavesTheRowOwingARebuild` pins the consequence end to end: the row reads `FAILED` with the sentence, so the retry pass sees it.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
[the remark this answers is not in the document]
vyskocilm: Fixed, including windows 3 and 4 — thank you for enumerating them, the count was right.

`withdrawInBatches` no longer reads a snapshot. Each batch takes the lock, reads one page of `findAssetUuidsByCbomUuid(cbomUuid, Limit.of(batchSize))` inside that transaction, withdraws it, and commits; the loop ends on a page the lock itself says is empty. That closes 1 and 2, and 4 with them — a CBOM sourcing nothing now takes the lock to establish that it sources nothing, rather than assuming it.

Window 3 is closed by `deleteAndTombstone` taking the same key as the first lock of its own transaction. That also closes the PR's third known deviation from the other side, and the deviation is struck from the description; `writeBatchUnderClusterLock` re-reads `existsById` under the lock as the ingest-side half (see the thread on `CbomAssetIngestService:76`).

What survives: the loop converges because an ingest of one document is finite, not because anything bounds it. That is now stated on the class rather than left implicit.

Tests: `aCbomThatSourcesNothingIsStillWithdrawnUnderTheLock` and `aSourceAttachedBetweenTwoBatchesIsWithdrawnTooRatherThanMissed`.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
vyskocilm: Fixed — the carve-out now turns on "withdrew nothing", not on "threw", exactly as you put it.

`withdrawInBatches` wraps a mid-loop failure in `WithdrawalFailedException`, carrying the partial `Withdrawal`; `recordPartialWithdrawal` records the failure only when `withdrewSomething()`. The comment this thread is on has been rewritten around that, and the 250-sources-at-batch-3 case is covered by `aWithdrawalThatEmptiedPartOfTheInventoryBeforeFailingLeavesTheRowOwingARebuild` in `CbomServiceITest` and by `aFailurePartwayThroughCarriesWhatTheCommittedBatchesWithdrew` in `CbomAssetDetachServiceTest`.

`testBulkDeleteCbom_withdrawalFailure_leavesTheRowsStateAlone` stays, and is now honest about what it pins: the zero-batch case, named as such in its javadoc, with a pointer to its twin.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
vyskocilm: Fixed, with one correction to the middle point.

`deleteCbom` now calls `withdrawForDeletion`, which does what the bulk path does: `recordPartialWithdrawal` when the withdrawal committed something, a shaped message, and an `OperationResult.FAILURE` audit entry with that message. A constraint violation is raised as a translated `ValidationException`; anything else is rethrown.

The correction: the raw message was not reaching the wire. `ExceptionHandlingAdvice` handles `DataIntegrityViolationException` by logging the class, SQL state and constraint name and returning `"Internal server error."`, and the `Exception` handler returns the same, so an operator saw neither the `DETAIL` line nor the driver text. Both other points hold — nothing recorded the partial withdrawal, and nothing logged a failure audit entry — and the shaped message is worth having for the refusal an operator can act on, which is why it is there now. But this was not an information leak, and I did not want to record it as one.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
vyskocilm: Fixed, by the second of your two shapes.

`withdrawInBatches` reads one page per batch inside the locked transaction (`findAssetUuidsByCbomUuid(cbomUuid, Limit.of(batchSize))`) and stops on a page read under the lock that comes back empty — so "complete" is a claim the lock witnessed, not one the snapshot remembered. Sources attached between two batch commits are picked up by the next page rather than missed.

The `withdrawWaiting` javadoc is amended to say what it does and does not promise: the emptiness holds for the transaction that read it, and keeping it true until the header is gone is the caller's job — which `deleteAndTombstone` now does by taking the same lock.

Test: `aSourceAttachedBetweenTwoBatchesIsWithdrawnTooRatherThanMissed`.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Fixed, as the one line you described.

`writeBatchUnderClusterLock` now re-reads `cbomRepository.existsById(cbomUuid)` under the lock, directly after the `hasIngestedLaterVersion` re-read and for the same stated reason, and returns a new `BatchOutcome.DELETED`. That surfaces as `IngestOutcome.DELETED`, counted in the run report as "deleted mid-ingest" rather than as a failure — no state write, because there is no row to write to.

Your note about the zero-source CBOM taking no lock at all is fixed on the other side too: the withdrawal now reads its work list under the lock, so the empty case acquires it (see the thread on `CbomAssetDetachService:159`), and `deleteAndTombstone` holds the same key across the header delete.

The PR's third known deviation is therefore struck rather than carried.

Test: `aCbomDeletedBetweenTwoBatchesStopsTheIngestRatherThanSourcingAMissingHeader`.

src/test/java/com/otilm/core/cbom/ingest/CbomAssetDetachServiceTest.java
vyskocilm: Applied, near enough verbatim — the `InOrder` on the two acquisitions is in `anOperatorsDeleteWaitsForTheClusterLockRatherThanSkippingTheWithdrawal`, with the comment saying what moving the alias lock above it would close.

You are right that the test asserted everything except the invariant it exists for.

src/test/java/com/otilm/core/integration/cbom/CryptoAssetInventoryITest.java
vyskocilm: Fixed. The two CBOMs now carry distinguishable payloads, the deleted one is the **elected** source — which is the case that needs re-election to have happened — and the test asserts the survivor's merged payload and its `propertiesSourceUuid`, plus a pre-assertion that the richer payload really was elected first, so the setup cannot silently stop testing what it means to.

That makes the `ON DELETE SET NULL` failure mode visible: without re-election the pointer is null and the stale payload stays, which the old `getSourceCount() == 1` could not see.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
vyskocilm: Recorded, in both places.

`CbomTombstoneWriter.record` now says it: the first deletion of a `(serial_number, version)` is the one the row describes, so after a delete → re-upload → delete cycle `uuid`, `deleted_at` and `deleted_by` name the first one and the second has no record of its own — the row is a record that the document is not wanted, not an audit trail of who removed which copy of it. The audit log is where that lives. And that the identity has no way back through the API: no expiry, no endpoint that removes a row, so a mistaken deletion is undone by re-uploading through `createCbom`, which leaves the tombstone behind.

Both are now a "Known deviations" entry on the PR as well, pointing at the weekly reconciliation PR as the place to give it a way out, and recorded as open work on #2073. Not fixed here — an unsuppress endpoint is a new operator-facing action, and it belongs with the pass that reconciles against the repository rather than bolted to the delete path.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
vyskocilm: Fixed — this was the sharper of the two, and the interleaving is real.

`markSynced` now writes through `updateAssetSyncStateUnlessError`, which refuses to overwrite a row carrying `DELETION_WITHDREW_THE_INVENTORY`. A state guard could not express it, since the thing that must not be overruled is a fact about the inventory rather than a state; the constant moved to `CbomAssetSyncStateWriter` so both writers read the same one.

It does not deadlock the row into `FAILED`: the next backlog pass claims it through `markInProgress`, which clears the error, and the re-ingest records its success normally. `anIngestFinishingAfterADeletionWithdrewTheInventoryCannotClaimTheRowIsSynced` pins both halves — the refused write, and the recovery after a claim.

On the milder version: left as it is, deliberately. `markFailed` replacing the sentence with an ingest reason loses operator-facing text, but the row stays `FAILED` and stays on the retry list, so nothing is stranded — and guarding it would mean a genuine ingest failure could not record itself, which is worse. Recorded as open work on #2073 rather than fixed.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
vyskocilm: You are right, and I took the second of your two options: the claim is dropped rather than the field added.

The sentence is gone from this comment and from the `CbomServiceITest` javadoc at :1795, so `CBOM_ASSET_SYNC_ERROR` appears nowhere. The carve-out rests on the state flip alone, which is the part that was always doing the work — and its reasoning has changed shape anyway, since it is now keyed on the constraint name and on whether the withdrawal committed anything rather than on the exception type.

Adding the filter field would have been in scope (`FilterField` is in `core`), but the DTO half is not — `CbomDto` lives in `interfaces` — and a searchable field nothing renders is a worse answer than no claim.

src/main/java/com/otilm/core/service/writer/cbom/CbomAssetSyncStateWriter.java
vyskocilm: Both fixed.

The write no longer goes through `updateAssetSyncState`. `markWithdrawnButNotDeleted` uses a new `updateAssetSyncStateKeepingAttempt`, which sets the state and the error and nothing else — this is not an ingest attempt, and your 10:45 → 12:00 example is exactly what stamping it cost.

And it now has a lower bound: `{IN_PROGRESS, SYNCED, FAILED}`, excluding `PENDING`, for the reason you give — a CBOM whose assets were never ingested has nothing to withdraw. I took the state guard rather than the `Withdrawal`-inspecting version because the two callers reach this write from different places, and the state is the thing that is true regardless of which.

Tests: `recordingAWithdrawnButUndeletedRowDoesNotRestartTheRetryClock` and `aDeleteFailureLeavesARowThatOwesItsFirstIngestAlone`. The second one also forced `testBulkDeleteCbom_deleteFailure_returnsErrorMessage` to say which state it means — it was asserting `FAILED` on a row it left `PENDING`, which under the new bound is the case that must not flip.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
vyskocilm: Fixed, in the shape you wrote, at both sites.

`theInventoryRefused(...)` reads the constraint name through `CryptoAssetConstraintTranslator.constraintNameOf` and compares it to `INVENTORY_SOURCE_CONSTRAINT`; `deleteWithdrawnRow` and the bulk path's carve-out both go through it. The translated message still reaches the caller either way — only the "leave the row SYNCED" decision is narrowed.

Your verification of the invariant is now recorded where the decision is, on the constant and on `deleteWithdrawnRow`, so the next foreign key added to `cbom` meets an argument rather than a coincidence.

src/test/java/com/otilm/core/integration/cbom/CryptoAssetInventoryITest.java
vyskocilm: Fixed. `CbomServiceITest.aBulkDeleteTheInventoryRefusesIsTranslatedAndLeavesTheRowSynced` is the bulk twin: a real `crypto_asset` row, a real `crypto_asset_source` row, and the real `RESTRICT` foreign key refusing the header delete, with the withdrawal suppressed through `detachServiceSpy` so the source row survives into it without needing a second node.

It asserts the translated sentence, that the uuid the driver's `DETAIL` line quotes is not in the message, and that the row is still present and still `SYNCED` — which exercises `recordBulkDeleteFailure`'s `describe` arm and the false direction of the carve-out at `:538`, both of which had no cover.

It went in `CbomServiceITest` rather than here because it needs `detachServiceSpy`, and adding a `@MockitoSpyBean` to this class would fork a new Spring context signature and require raising `ContextSignatureGuardTest.BASELINE` — not a purchase worth making for one test.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetDetachService.java
vyskocilm: Recorded as open work on #2073 rather than fixed, and stated as a known deviation on the PR.

The measurement stands: the bound is per batch, and a document is `ceil(sources / batch)` batches with `bulkDeleteCbom` multiplying by the list length, so the total an operator's request can wait has no ceiling and holds a pooled connection throughout.

Two reasons not to take `SET LOCAL lock_timeout` in this PR:

* it reintroduces the refusal `withdrawWaiting` exists to remove. Your note that the path can end in the `RESTRICT` refusal anyway is no longer true after the change on the `:159` thread — the withdrawal now reads its work list under the lock and `deleteAndTombstone` holds the same key across the header delete, so once this path has the lock the delete goes through. Trading a certain success for a timed-out retry is a different decision than it was when you wrote this;
* the bound is a tunable, and under SaaS an operator-facing tunable belongs in the Settings subsystem rather than in yml, which is a larger change than a review round should carry — plus `ClusterOperationSynchronizer` is platform-wide, and a timeout-shaped acquisition there wants its own PR.

The ping-pong you describe is real and is now said on `CbomAssetDetachService`: the loop converges because an ingest of one document is finite, not because anything bounds it.

karolhrdina (approved):

PR #2264, opened 2026-09-15
Keep a scheduled job's history out of the job's transaction

### LINKED ISSUE #2251: Scheduled jobs longer than the 120 s default transaction leave no job-history row
### Description

`SchedulerListener` carries a class-level `jakarta.transaction.Transactional`, so every scheduled job runs inside one platform transaction with the default `spring.transaction.default-timeout: 120` seconds. `SchedulerServiceImpl.runScheduledJob` writes the `STARTED` job-history row in that transaction, calls the task's `performJob`, and writes the final status (`finalizeFinishedScheduledJob`) in the same transaction, which commits when `processMessage` returns.

Tasks that page an external system already declare `performJob` as `@Transactional(propagation = NOT_SUPPORTED)` (`DiscoveryCertificateTask`, `CbomSyncTask`). That suspends the listener's transaction for the duration of the job, but the transaction's deadline keeps running. A job that takes longer than 120 s therefore resumes into a transaction that is already past its deadline:

- the history writes and the commit fail (`transaction timeout expired`), the transaction rolls back, and both the `STARTED` row and the final `SUCCESS` / `FAILED` status vanish. The run leaves **no job-history row at all**, although everything the job committed on its own (entries written in `REQUIRES_NEW`, skip records, discovery results) stands;
- the failure surfaces only as `Unexpected error in endpoint 'scheduler…'` in the Core log — `AbstractJmsEndpointConfig` catches it, so the message is acknowledged and the job does not run twice, but nothing points an operator from the missing history row to that log line;
- the one-time-job deregistration (an HTTP call) and the finished-job event are sent from inside `finalizeFinishedScheduledJob`, before the commit that then fails, so the history can disagree with what was sent;
- while any job runs, its `STARTED` row is uncommitted, so the job history shows nothing running.

The CBOM header sync (core#2245) derives its watermark from the last `SUCCESS` row of `CbomSyncTask` (`getLastSyncTimestamp`). A run longer than 120 s — the first sync of a large repository, or any run with many documents to read — never records `SUCCESS`, so the next run re-lists from the previous watermark. The sync recovers on its own only because already-stored entries are skipped without a read, which makes the following run shorter; a repository whose feed alone takes longer than 120 s to page never gets a `SUCCESS` row and re-lists everything on every run. The same holds for every other long-running job.

Found during the review of core#2245 (Michal, item "worth your call" 1); the PR moved `CbomSyncTask.performJob` to `NOT_SUPPORTED` and left the listener as it was, because this is scheduler-wide.

### Proposed fix

- `SchedulerListener` becomes `@Transactional(propagation = NOT_SUPPORTED)`, like the platform's other listeners (`EventListener`, `NotificationListener`, `TimeQuality*Listener`), so a scheduled job never runs inside an ambient transaction.
- Job-history bookkeeping moves to a `*Writer` bean with short `REQUIRED` transactions, per the repository's transactional-boundary rule: one that commits the `STARTED` row before `performJob` (so the row is visible while the job runs), one that writes the final status or removes the row of a skipped run afterwards. The `ScheduledJob` lookup and `authenticateAsUser` stay where they are.
- The one-time-job deregistration and the finished-job event are sent after the final status is committed (commit first, then call externally).
- Verify that the `@TransactionalEventListener(AFTER_COMMIT)` paths keep the transaction they need: `handleScheduledJobFinishedEvent` is fed by `DiscoveryFinishedEventHandler`, not by the scheduler listener, and `JmsConfig` deliberately configures no JMS transaction manager so those handlers run under the message's identity — that comment must still hold after the change.

### Acceptance criteria

- A scheduled job that runs longer than `spring.transaction.default-timeout` ends with a `SUCCESS` / `FAILED` history row carrying its result message, and the CBOM sync's watermark advances after such a run.
- The `STARTED` row of a running job is visible from another transaction while `performJob` is executing (this is also the test that proves no ambient transaction spans the job — `SchedulerServiceITest`, no new Spring context).
- A skipped run (`ScheduledJobSkippedException`) still leaves no history row; a failed task class still leaves a `FAILED` row.
- `TransactionalBoundaryArchTest` passes with the new writer; `SchedulerServiceITest`, `SchedulerListenerITest`, `CbomSyncITest` and `CbomSyncTaskITest` keep passing.

### Touch points

`SchedulerListener`, `SchedulerServiceImpl.runScheduledJob` / `finalizeFinishedScheduledJob` / `registerJobHistory`, a new `ScheduledJobHistoryWriter` under `service/writer`, `SchedulerServiceITest`.

Related: core#2245 (PR body, "Prerequisites and follow-ups"), core#2208.

### Addendum 2026-09-15 — what the proposed fix above leaves out

Read against `main` (`8a3c1dfb`) before implementation. Nothing in the listener or `SchedulerServiceImpl` changed since this was filed; two things make it more urgent: core#2256 put asset ingest inside the same `CbomSyncTask` run (up to `cbom.sync.max-ingest-documents` extra document reads per run), so runs over 120 s are now the expected case for a repository with a backlog, and `CryptoAssetPqcSweepTask` (core#2228) was added with no propagation on `performJob` — it inherits the ambient transaction exactly as described here (its own `sweep()` is `REQUIRES_NEW`, so its writes survive; only the history row is lost).

**A task that throws.** `runScheduledJob` finalizes only when `skipped || result != null`. Any `RuntimeException` other than `ScheduledJobSkippedException` escapes with `result == null`, so nothing finalizes and the exception ends in `AbstractJmsEndpointConfig`'s catch-all. Today the listener transaction rolls back and takes the `STARTED` row with it. Once `STARTED` is committed before `performJob`, that same path leaves a committed `STARTED` row with `jobEndTime = null` forever: the history shows a phantom running job and `deleteScheduledJob` refuses (`existsByScheduledJobUuidAndSchedulerExecutionStatusAndJobEndTimeIsNull`). The fix must catch it, write `FAILED` with the exception message, and rethrow (the endpoint still logs and acknowledges — no redelivery, no double run). **`null` is not failure**: `DiscoveryCertificateTask` returns `null` on purpose when the discovery runs asynchronously and `handleScheduledJobFinishedEvent` finalizes the row later. Rule: *threw* → `FAILED`; *returned `null`* → the `STARTED` row stays for the later finalization.

**Tasks and what changes for them.** `CbomSyncTask` and `DiscoveryCertificateTask` declare `performJob` `NOT_SUPPORTED` — unchanged behaviour, now without a ticking deadline underneath. `CryptoAssetPqcSweepTask` declares nothing — after the change it runs with no ambient transaction, which is what its `REQUIRES_NEW` sweeper wanted anyway. `UpdateCertificateStatusTask` and `UpdateIntuneRevocationRequestsTask` carry their own class-level `REQUIRED`, so they will open their own 120 s transaction instead of joining the listener's: same timeout as today, but a timeout now surfaces as a task failure with the `STARTED` row surviving and a `FAILED` row written, rather than as a vanished history. State this in the PR; do not change those tasks here.

**Events published inside `performJob`.** With the listener at `NOT_SUPPORTED`, an event published from inside a task without its own transaction has no transaction to be `AFTER_COMMIT` of; a `@TransactionalEventListener` without `fallbackExecution = true` would silently not run. Sweep every `publishEvent` reachable from a `ScheduledJobTask` and every `@TransactionalEventListener(AFTER_COMMIT)` (`SchedulerServiceImpl.handleScheduledJobFinishedEvent`, `CertificateHandler`, `CertificateEventHistoryServiceImpl`, `NotificationProducer`, `TspProfileSecretEvictionListener`, `TimeQualityRegisterImpl`, `TimeQualityMonitorInitializer`) and record the result in the PR. `handleScheduledJobFinishedEvent` itself is fed by `DiscoveryFinishedEventHandler` (`@Transactional`, reached from `EventListener`), not by the scheduler listener, so it keeps its transaction.

**Architecture rules.** `TransactionalBoundaryArchTest.effectivePropagation` reads only Spring's `@Transactional`, so the listener's `jakarta` annotation is invisible to it today; after switching to Spring's `@Transactional(propagation = NOT_SUPPORTED)` Rule A applies to `processMessage` (it calls no repository — fine). The new `ScheduledJobHistoryWriter` goes under `service/writer`, is a `@Service`, and every public method is effectively `@Transactional` (`REQUIRED`) — Rule D; `ScheduledJobHistoryRepository` has no `@Modifying` methods, so Rule C is not engaged. The one-time-job deregistration (HTTP) and `ScheduledJobFinishedEvent` publication move out of the writer, after its commit returns.

**Stale text to correct in the same PR.** `CryptoAssetPqcSweepTask` Javadoc ("Not `@Transactional`: `SchedulerListener` already opens one, and the sweeper's own is the second" and the `performJob` note "It runs on the listener's transaction, which is still live") and the `JmsConfig` comment on the missing JMS transaction manager, which currently explains itself with "the JPA transaction then commits inside `processMessage`" — the identity argument still holds, the sentence about where the commit happens does not.

**Tests.** `SchedulerServiceITest` calls `runScheduledJob` directly, i.e. with no ambient transaction, so it cannot reproduce this bug and will pass before and after — the new ITest must go through the listener bean (as `SchedulerListenerITest` does) with a task bean that blocks mid-`performJob` while a second `REQUIRES_NEW` transaction reads the `STARTED` row; no new Spring context (`BaseSpringBootTest`). `SchedulerServiceMockedTest` (`runScheduledJob` / history save / event production) is rewritten around the writer. Add: throwing task → one `FAILED` row with the message; `null`-returning task → `STARTED` row still present with `jobEndTime = null`.

### Acceptance criteria (additions)

- A task that throws a `RuntimeException` other than `ScheduledJobSkippedException` leaves exactly one history row, `FAILED`, carrying the exception message; the exception still reaches the endpoint's catch-all and the message is acknowledged once.
- A task that returns `null` leaves its `STARTED` row (`jobEndTime = null`) for a later finalization — `DiscoveryCertificateTask`'s asynchronous path keeps working end to end (`DiscoveryCertificateTask` + `handleScheduledJobFinishedEvent`).
- Every `@TransactionalEventListener(AFTER_COMMIT)` reachable from a scheduled job is listed in the PR with its transaction source after the change; none depends on the removed listener transaction.
- `CryptoAssetPqcSweepTask`'s and `JmsConfig`'s comments describe the new shape.
- No change to `UpdateCertificateStatusTask`, `UpdateIntuneRevocationRequestsTask`, `DiscoveryCertificateTask`, `CbomSyncTask` code; their behaviour differences are stated in the PR body.

### Definition of done

- Branch off `main`, one PR, no interfaces change, no `Depends-On:` marker, no ticket dependency (independent of core#2249 / core#2250; land before core#2250 adds its sweep task).
- `mvn -B -ntp spotless:check checkstyle:check` clean; `TransactionalBoundaryArchTest`, `SchedulerServiceITest`, `SchedulerListenerITest`, `SchedulerServiceMockedTest`, `CbomSyncITest`, `CbomSyncTaskITest`, `CryptoAssetPqcSweepTask`'s tests green locally against PostgreSQL; Sonar gate ≥ 80 % on new code.
- Adversarial review (independent max-effort pass over the diff, correctness / races / state divergence, per the repository's CLAUDE.md) run and its findings closed before the PR is opened; PR body names the review and states merge order (none) and the behaviour changes per task.
- Scope: pure scheduler. Out: making the 120 s default configurable, changing any task's own transaction, the merge queue (core#2193).

karolhrdina:
Closes #2251.

## Problem

`SchedulerListener` ran every scheduled job inside one platform transaction with `spring.transaction.default-timeout: 120`. A task's `NOT_SUPPORTED` `performJob` suspended that transaction but not its deadline, so a run longer than 120 s resumed into an expired transaction: the history writes rolled back, the run left **no history row at all**, and the CBOM sync's watermark (the last `SUCCESS` row of `CbomSyncTask`) never advanced — every following run re-listed from the previous watermark. Since #2245 and #2256 the sync also ingests assets, so runs over 120 s are the expected case for a repository with a backlog.

## Change

- `SchedulerListener` is `@Transactional(propagation = NOT_SUPPORTED)`, like `EventListener` and `NotificationListener`, and `SchedulerServiceImpl.runScheduledJob` carries the same annotation itself, so no caller can put a job back under an ambient transaction.
- New `ScheduledJobHistoryWriter` (`service/writer/scheduler`, `@Service`, plain `REQUIRED` methods under Rule D). Both callers run `NOT_SUPPORTED` — `runScheduledJob`, and `handleScheduledJobFinishedEvent`, which suspends the committed-but-still-registered publishing transaction of the `AFTER_COMMIT` phase — so each write starts and commits a transaction of its own. Each close is one `@Modifying` statement (`ScheduledJobHistoryRepository.closeRun` / `removeRun`); a row count of 0 — the row vanished, e.g. the job was deleted mid-run and cascaded — is an `IllegalStateException`, the same policy for all three finalizers.
- `runScheduledJob`, three rules:
  1. **STARTED is committed before `performJob`** and visible to other transactions while the job runs (the history shows the run while it runs; it also closes the old race where `DiscoveryCertificateTask` handed an uncommitted `jobHistoryUuid` to an async discovery).
  2. **Threw → `FAILED`.** A `RuntimeException` from `authenticateAsUser`/`performJob` closes the row with a shaped message (`PlatformException.safeMessage`, never the raw text) and is rethrown; if the bookkeeping write itself fails, it rides along as a suppressed exception and the task's own failure still surfaces. An `Error` is not caught (Sonar's rule and the platform's habit) but a `finally` closes its row too; whether the message is redelivered is then the JMS retry policy's call — a transient `Error` gets its retry, a deterministic one exhausts the policy. Without this the now-durable `STARTED` row would stay open forever and `deleteScheduledJob` would refuse the job as "executing".
  3. **Returned `null` → finishes later** (`DiscoveryCertificateTask` while the discovery is `PROCESSING`); the row stays `STARTED` for `handleScheduledJobFinishedEvent`, which is now `NOT_SUPPORTED` instead of `REQUIRES_NEW` (it suspends the committed publishing transaction so the writer commits on its own, and keeps the scheduler HTTP call and the JMS event outside any database transaction).
- `finalizeFinishedScheduledJob(job, historyUuid, result)`: status committed first, then the one-time deregistration (its client fails unchecked; caught as such), then the finished-job event; a failure to send the event is logged with the job name and swallowed — the status is committed and a redelivery would run the job again. **If the close write itself fails** (a failover between the task's return and `recordFinished`), a second close marks the row `FAILED` naming the real outcome and rethrows: the outside is told nothing, and a missing `SUCCESS` row only makes the next sync re-list. The listener swallows `RuntimeException` for the same reason and logs "did not run to completion".
- Skipped runs (`ScheduledJobSkippedException`) still leave no row; an unknown task class still leaves a `FAILED` row.

## Behaviour per task (one task changed, one statement)

| task | `performJob` | before → after |
|---|---|---|
| `CbomSyncTask`, `DiscoveryCertificateTask` | `NOT_SUPPORTED` | unchanged, minus the ticking deadline underneath. `DiscoveryCertificateTask` additionally shapes its `FAILED` message with `PlatformException.safeMessage` — it was the one path that still put a raw driver/constraint text into `result_message` — and logs the exception itself |
| `CryptoAssetPqcSweepTask` | none (`sweep()` is `REQUIRES_NEW`) | unchanged behaviour; its Javadoc no longer claims the listener opens a transaction |
| `UpdateCertificateStatusTask` (class-level jakarta `@Transactional`), `UpdateIntuneRevocationRequestsTask` (class-level Spring `@Transactional`) | `REQUIRED` | now open their **own** 120 s transaction at `performJob` instead of joining the listener's; a timeout surfaces as a `FAILED` row with the `STARTED` row surviving, not as a vanished history |

## AFTER_COMMIT sweep

Every `@TransactionalEventListener(AFTER_COMMIT)` in `src/main` and where its transaction comes from after this change: `SchedulerServiceImpl.handleScheduledJobFinishedEvent` ← `DiscoveryFinishedEventHandler` (class `@Transactional`, reached from `EventListener`); `CertificateHandler` ← `CertificateUploadedEventHandler`, `CertificateServiceImpl`, `LocationServiceImpl` (none reachable from a task); `CertificateEventHistoryServiceImpl` ← three `@Transactional` event handlers; `TimeQualityMonitorInitializer`, `TimeQualityRegisterImpl` ← `TimeQualityConfigurationServiceImpl` (API); `NotificationProducer`, `TspProfileSecretEvictionListener` carry `fallbackExecution = true`. `eventProducer.produceMessage` is JMS, not a Spring event. `UpdateCertificateStatusTask`'s `CertificateValidationEvent` now fires at the task's own commit. No listener depends on the removed transaction.

## Known limitation, deliberately left out

A process that dies mid-job (OOM kill, pod eviction) now leaves a committed `STARTED` row with `jobEndTime = null`; before, the rollback erased it. Such a row blocks `deleteScheduledJob` ("executing") until it is closed by hand. Closing it automatically needs either an age bound on that guard or a startup reaper that knows which node owned the run — neither is safe to guess here in a clustered deployment. Recorded as open work in a comment on #2251 so it survives the squash; follow-up PR when a remedy is chosen.

## Verification

- TDD: `SchedulerListenerITest.testProcessMessage_StartedRowIsVisibleToAnotherTransactionWhileTheJobRuns` and `…TaskThrowsRuntimeException_LeavesExactlyOneFailedRow` fail on `main` (uncommitted insert; zero rows after rollback) and pass here; `…Success_LeavesTheSuccessRowTheWatermarkLookupReads` guards the watermark join. `SchedulerServiceMockedTest` rewritten around the writer (+7: throws → FAILED and rethrow; `Error` → FAILED; bookkeeping failure suppressed; null → row stays open; event failure ≠ failed run; an unchecked deregistration failure still sends the event; a failed close write → FAILED and rethrow). `ScheduledJobHistoryWriterITest` (new, real PostgreSQL): commit-on-return, every result column, shaped FAILED, terminal unknown-task row, delete, vanished row reported by all three finalizers. `ScheduledJobFinishedEventITest` (new): the asynchronous finish end to end — a `ScheduledJobFinishedEvent` published inside a `TransactionTemplate` closes the row after that transaction commits. `SchedulerTransactionContractTest` pins `NOT_SUPPORTED` on the listener, `runScheduledJob` and the event handler. `DiscoveryCertificateTaskTest` (new): a `DataIntegrityViolationException` never reaches the result message.
- Local, against Testcontainers PostgreSQL: `SchedulerListenerITest`, `SchedulerServiceMockedTest`, `SchedulerServiceITest`, `ScheduledJobHistoryWriterITest`, `ScheduledJobFinishedEventITest`, `SchedulerTransactionContractTest`, `DiscoveryCertificateTaskTest`, `TransactionalBoundaryArchTest`, `CbomSyncITest`, `CbomSyncTaskITest`, `CryptoAssetPqcSweepTaskTest`, `DiscoveryServiceITest` — all green; `spotless:check checkstyle:check` clean.
- Adversarial review before opening: two independent maximum-effort passes (Opus, Fable) and a `/code-review high` pass over the diff, then a focused re-review of the fixes. Taken from them: the `Error` catch, `addSuppressed`, `NOT_SUPPORTED` on `runScheduledJob` itself, the non-null result guard, the explicit event-failure catch, the one-time deregistration catch widened to the client's unchecked failures, the one-statement writer and its vanished-row policy, the writer ITest, three Javadoc corrections. From vyskocilm's first round (second commit): the close-write fallback, the writer back under Rule D as `@Service`/`REQUIRED`, the event-path integration test in place of the reflection contract test, the terminal unknown-task row, the discovery message shaping, and the `Error`/redelivery trade stated.

**Merge order: none.** No `Depends-On`, no interfaces change, independent of core#2249; land before core#2250 adds its retention sweep task so that task does not inherit the bug.


src/main/java/com/otilm/core/service/impl/SchedulerServiceImpl.java
```diff
@@ -293,82 +307,114 @@ public void runScheduledJob(final String jobName) throws NotFoundException {
...
             }
+            throw e;
+        }
+
+        if (result == null) {
+            // Finishes asynchronously (DiscoveryCertificateTask while the discovery is PROCESSING);
+            // handleScheduledJobFinishedEvent closes the row later.
+            return;
         }
+        finalizeFinishedScheduledJob(scheduledJob, history.getUuid(), result);
```
vyskocilm, line 376: L360: 🔴 bug: `finalizeFinishedScheduledJob(...)` sits outside the try/catch that records FAILED, so the close write has no fallback.

A long `CbomSyncTask` run returns SUCCESS; the pooler restarts or the DB fails over between that return and `historyWriter.recordFinished`. The `DataAccessException` propagates out of `runScheduledJob` and the new `catch (RuntimeException)` in `SchedulerListener` swallows it. The committed STARTED row then stays open with `jobEndTime = null` permanently: `deleteScheduledJob` (L164-171) refuses the job as "executing" with no age bound and no reaper, and no SUCCESS row is written, so the sync watermark never advances. That is the end state #2251 describes, moved from the timeout path to the close path. Before this PR the same failure rolled the row back cleanly.

It is a likelier variant of the process-dies-mid-job limitation the description discloses, and it is not covered by it.

Fix: close best-effort as FAILED in a catch here, or bound the "executing" guard by age.

src/main/java/com/otilm/core/service/impl/SchedulerServiceImpl.java
```diff
@@ -293,82 +307,114 @@ public void runScheduledJob(final String jobName) throws NotFoundException {
...
-        final ScheduledJobHistory scheduledJobHistory = registerJobHistory(scheduledJob,
-                SchedulerJobExecutionStatus.STARTED, null);
-
-        if (scheduledJob.getUserUuid() != null) {
-            authHelper.authenticateAsUser(scheduledJob.getUserUuid());
-        }
+        // Committed before the task runs: the history shows the run while it runs, and nothing the task does --
+        // however long it takes -- can roll this row back. Runs with no transaction of its own (see
+        // SchedulerListener); every write below is the writer's own short transaction.
+        final ScheduledJobHistory history = historyWriter.recordStarted(scheduledJob);
```
vyskocilm, line 318: L318: 🔵 committing STARTED before `performJob` is the right fix, but the state it creates outlives this PR's description. A process that dies mid-job (OOM kill, pod eviction) leaves a committed row with `jobEndTime = null`; `deleteScheduledJob` (L164-171) then refuses that job indefinitely, and `SchedulerExternalService` exposes no operation to close it, so the only remedy is a manual database edit. The description names this and declines to file it — a squash merge takes that record with it. Worth recording as open work on #2251 or the epic, so the constraint survives the merge.

src/main/java/com/otilm/core/service/writer/scheduler/ScheduledJobHistoryWriter.java
```diff
@@ -0,0 +1,97 @@
...
+ * {@code SchedulerListener}, which runs with no transaction, and from an {@code AFTER_COMMIT} event handler, where a
+ * {@code REQUIRED} write would join a transaction that has already committed and be lost. {@code REQUIRES_NEW} is right
+ * from both call sites, and it makes a run's final status durable before any external call reports it.
+ *
+ * <p>
+ * Each close is one {@code @Modifying} statement, as the writer convention asks, and its row count is what tells a
+ * vanished row apart: the three finalizers all name the row {@link #recordStarted} committed moments earlier on the
+ * same thread, so a missing row means the same thing to each of them.
+ */
+@Component
```
vyskocilm: L19-30: 🟡 The stated reason for `@Component` + `REQUIRES_NEW` does not hold for the shape this PR ships. That matters, because it takes the writer out of the only machine-checked rule that exists for writers — `TransactionalBoundaryArchTest` selects `@Service` classes under `..service.writer..` and requires plain `REQUIRED` — and `ScheduledJobHistoryWriterContractTest` then freezes that exemption.

The javadoc says a `REQUIRED` write would "join a transaction that has already committed and be lost" in the AFTER_COMMIT phase. But the only AFTER_COMMIT caller, `handleScheduledJobFinishedEvent` (`SchedulerServiceImpl` L370), is annotated `NOT_SUPPORTED` in this same PR: it suspends and unbinds the EntityManagerHolder before the writer is reached, so a `REQUIRED` writer finds no ambient transaction and starts and commits its own. The other call site, `runScheduledJob`, is `NOT_SUPPORTED` too. `REQUIRES_NEW` is therefore not load-bearing at either call site.

Measured rather than argued. With the writer changed to `@Service` + plain `@Transactional`, `ScheduledJobHistoryWriterITest`, `SchedulerListenerITest`, `SchedulerServiceITest`, `TransactionalBoundaryArchTest` and a probe that publishes a real `ScheduledJobFinishedEvent` inside a `TransactionTemplate` all pass: 19 tests, 0 failures. Negative control, to show the probe can fail and the mechanism is real: with `REQUIRED` *and* `NOT_SUPPORTED` removed from the handler, the probe fails with `expected: <SUCCESS> but was: <STARTED>`. So the lost write happens only when the handler does not suspend — which this PR ensures it does.

Fix: `@Service` + plain `@Transactional` on all five methods. If `REQUIRES_NEW` is genuinely wanted as belt-and-braces against a future caller that does hold a transaction, say that instead of the AFTER_COMMIT argument — and then it is a change to the shared rule (widen it to every bean in that package), not an exemption from it.

src/main/java/com/otilm/core/service/writer/scheduler/ScheduledJobHistoryWriter.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /** The run has started; the row is visible to every other transaction the moment this returns. */
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public ScheduledJobHistory recordStarted(ScheduledJob job) {
+        return repository.save(newRow(job, SchedulerJobExecutionStatus.STARTED, null));
+    }
+
+    /** A run that could not start because its task class is unknown: one FAILED row, no STARTED before it. */
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public ScheduledJobHistory recordUnknownTask(ScheduledJob job, String message) {
```
vyskocilm, line 50: L47: 🔵 nit: `recordUnknownTask` writes a *terminal* FAILED row through `newRow(...)`, which never sets `jobEndTime`. It lands with `job_end_time = NULL` and renders through `ScheduledJobHistoryDto.endTime` as a run that never ended, indistinguishable in the UI from one still in flight. (It does not trip the delete guard, which also requires STARTED.) Inherited from the old `registerJobHistory`, but this is the PR that made the three finalizers consistent about end times, and it left the fourth writer out. Set `jobEndTime` here.

vyskocilm (commented):
A review of this PR. One blocking finding, two worth settling before merge, and some smaller notes. The transaction reshaping itself is right, and the three new integration tests do fail on `main` as the description claims.

**What ran**, against Testcontainers PostgreSQL: the full architecture test package (133 tests), plus `SchedulerListenerITest`, `ScheduledJobHistoryWriterITest`, `ScheduledJobHistoryWriterContractTest`, `SchedulerServiceMockedTest`, `SchedulerServiceITest`, `CbomSyncTaskITest`, `DiscoveryServiceITest` — all green. `spotless:check checkstyle:check` clean, 0 violations. The test-context signature guard passes unchanged at 61, and correctly so: the new ITest extends the shared base with no local mocks, and `SchedulerListenerITest`'s `@MockitoBean CbomServiceImpl` reproduces a signature `CbomSyncTaskITest` already buys. No growth in the frozen-violation store.

**Checked and clean**, so nobody re-derives them: `catch (SchedulerException | RuntimeException)` compiles because `SchedulerException extends Exception`; `ScheduledJob` has no lazy association, so the detached read under `NOT_SUPPORTED` is safe; `performJob` and `authenticateAsUser` declare no checked exceptions, so `catch (RuntimeException | Error)` really is total and no path leaves a committed STARTED row on a throw; the shared test base truncates the schema per test, so the writer's real commits escaping test rollback do not pollute `findAll()` assertions; `ScheduledJobHistory` and its superclass declare no audit columns, so the UPDATE owes none; `result_message` is `text`, not bounded; all five `ScheduledJobTask` implementations either suspend, open their own transaction, or delegate to a `REQUIRES_NEW` bean, so nothing in the job path relied on the listener's removed transaction. No authorization, resource/action, migration or search-field change anywhere in the diff.

**Not run**: the full suite, and a coverage-parity check against the Sonar exclusions.

src/main/java/com/otilm/core/service/writer/scheduler/ScheduledJobHistoryWriter.java
```diff
@@ -0,0 +1,97 @@
...
+        return repository.save(newRow(job, SchedulerJobExecutionStatus.FAILED, message));
+    }
+
+    /** The task returned a result: closes the row with it. */
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public void recordFinished(UUID historyUuid, ScheduledTaskResult result) {
```
vyskocilm, line 58: L53: 🟡 risk: the message-shaping guarantee this PR adds covers `recordFailed` only. `recordFinished` writes `ScheduledTaskResult.getResultMessage()` verbatim into the same column that `SchedulerServiceImpl` L343 describes as one that reaches the scheduler API.

`DiscoveryCertificateTask.performJob` (L96-102) builds that message as `String.format("Unable to create discovery %s for job %s. Error: %s", ..., e.getMessage())` from a `catch (Exception e)` around `discoveryService.createDiscovery(...)`, so a `DataIntegrityViolationException` puts raw constraint or SQL text into `result_message`, which `getScheduledJobHistory` then serves.

Pre-existing and outside the diff, but it is now the only unshaped path into a column this PR otherwise gates. Either shape it in `close(...)`, or say explicitly that it is out of scope.

src/test/java/com/otilm/core/service/writer/scheduler/ScheduledJobHistoryWriterContractTest.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Pins the one thing about the writer that a tidy-up would break: it is called from inside an {@code AFTER_COMMIT}
+ * phase, where a {@code REQUIRED} write joins a transaction that has already committed and is lost. Making it a
+ * {@code @Service} would have {@code TransactionalBoundaryArchTest}'s Rule D demand exactly that.
+ */
+class ScheduledJobHistoryWriterContractTest {
+
+    @Test
+    void everyPublicMethodOpensItsOwnTransaction() {
+        assertFalse(ScheduledJobHistoryWriter.class.isAnnotationPresent(Service.class),
```
vyskocilm: L15-23: 🟡 risk: this test claims to pin the AFTER_COMMIT behaviour but cannot observe it — it reads annotations by reflection and never loads a context, opens a transaction, or writes a row. Its assertion message also restates the premise corrected in the writer comment, so it would block a later correct normalisation to `@Service` while citing a condition that no longer holds.

The path itself is uncovered. `SchedulerServiceMockedTest` L563/L580 are plain Mockito, so no transaction semantics at all, and neither new integration test publishes a `ScheduledJobFinishedEvent`. The asynchronous-finish path — `DiscoveryCertificateTask` returning null, the row closed later by the event — has no end-to-end test in the PR whose central design claim is about exactly that path. A future change to the handler's propagation would silently stop closing history rows with the suite still green.

Fix: replace the reflection assertion with an integration test that publishes the event from inside a `TransactionTemplate` and asserts the row is closed after commit. Extending the shared Spring base with no local mocks adds no context signature, so no baseline change is needed.

src/main/java/com/otilm/core/messaging/jms/listeners/SchedulerListener.java
```diff
@@ -30,6 +38,13 @@ public void processMessage(SchedulerJobExecutionMessage schedulerMessage) {
             schedulerService.runScheduledJob(schedulerMessage.getJobName());
         } catch (SchedulerException | NotFoundException e) {
             logger.error("Unable to process the job {}. Error: {}", schedulerMessage.getJobName(), e.getMessage(), e);
+        } catch (RuntimeException e) {
+            // Deliberately swallowed, so the endpoint acknowledges the message: a redelivery would run the job again.
+            // Whatever the run recorded stands -- a FAILED row when the task threw, SUCCESS or FAILED when a side
+            // effect after the status failed, nothing at all when the failure came before the STARTED row -- and this
+            // line names the job for whoever has to look. An Error is not caught here: its FAILED row is written all
+            // the same, and whether the message is redelivered is the JMS retry policy's call.
+            logger.error("Scheduled job '{}' did not run to completion", schedulerMessage.getJobName(), e);
```
vyskocilm, line 49: L41-47: ❓ q: `RuntimeException` is swallowed so the endpoint acks, the stated reason being that a redelivery would run the job again. `Error` is rethrown and left to the JMS retry policy — so it *does* redeliver and *does* run the job again, against a row already closed FAILED. Same argument, opposite handling. Deliberate?

Separately, the catch does not distinguish a deterministic job failure (where redelivery would loop) from a transient infrastructure failure before the STARTED row is written (where redelivery used to recover, and where the run now simply does not happen, leaving nothing but a log line). Your call rather than a finding, but worth stating in the description if it is the intended trade.

src/test/java/com/otilm/core/integration/messaging/jms/listeners/SchedulerListenerITest.java
```diff
@@ -75,4 +88,64 @@ void testProcessMessage_CbomSyncTaskThrowsException_DoesNotThrowUnexpectedRollba
...
+     * repository read would join whatever transaction the listener holds and prove nothing.
+     */
+    private TransactionTemplate otherTransaction() {
+        TransactionTemplate template = new TransactionTemplate(transactionManager);
+        template.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
+        return template;
+    }
+
+    @Test
+    void testProcessMessage_StartedRowIsVisibleToAnotherTransactionWhileTheJobRuns() throws Exception {
```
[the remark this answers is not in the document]
karolhrdina: Taken in two ways in e7dc5501: the writer is now `REQUIRED` (see vyskocilm's thread on it), so this test discriminates — an ambient transaction at the listener or at `runScheduledJob` would make `recordStarted` join it and the `REQUIRES_NEW` reader would see nothing; and `SchedulerTransactionContractTest` pins `NOT_SUPPORTED` on `SchedulerListener`, `runScheduledJob` and `handleScheduledJobFinishedEvent` directly.

src/main/java/com/otilm/core/service/impl/SchedulerServiceImpl.java
karolhrdina: Confirmed and fixed in e7dc5501: `finalizeFinishedScheduledJob` wraps `recordFinished`; when it fails, a second close marks the row FAILED with "The run finished with SUCCESS but its history could not be closed; see the Core log" (suppressed onto the original if that fails too) and rethrows — the outside is told nothing, and a missing SUCCESS row only makes the next sync re-list, which is safe. Pinned by `testRunScheduledJob_WhenTheCloseWriteFails_ClosesTheRowAsFailedAndRethrows`. The age bound on the guard I left alone: it is the process-death remedy, now recorded on #2251 per your other thread.

src/main/java/com/otilm/core/service/impl/SchedulerServiceImpl.java
karolhrdina: Recorded on #2251 as open work (comment dated today) with the two remedies — an age bound on the "executing" guard, or a startup reaper that knows which node owned the run — and why neither is safe to guess in a clustered deployment. No new ticket, per the epic's rule this week; the PR body now points at that comment instead of declining.

src/main/java/com/otilm/core/service/writer/scheduler/ScheduledJobHistoryWriter.java
karolhrdina: Measured the same way, and agree: with both callers `NOT_SUPPORTED`, `REQUIRES_NEW` was not load-bearing. e7dc5501: `@Service` + plain `@Transactional` on all five methods, so Rule D applies; the Javadoc now states the real dependency — the callers guarantee no ambient transaction, the handler by suspending the committed publishing one. The contract test that froze the exemption is gone (see the thread on it).

src/main/java/com/otilm/core/service/writer/scheduler/ScheduledJobHistoryWriter.java
karolhrdina: Fixed in e7dc5501: `recordUnknownTask` sets `jobEndTime = jobExecution`; `ScheduledJobHistoryWriterITest.recordUnknownTask_isTerminalWithAnEndTime` pins it.

src/main/java/com/otilm/core/service/writer/scheduler/ScheduledJobHistoryWriter.java
karolhrdina: Shaped at the source in e7dc5501: `DiscoveryCertificateTask` formats the failure with `PlatformException.safeMessage(e, …)` and logs the exception itself, so the raw cause still reaches the Core log; `DiscoveryCertificateTaskTest` feeds it a `DataIntegrityViolationException` carrying constraint text and asserts the result message carries none of it. `close(...)` stays verbatim on purpose: the writer cannot tell a shaped task message from a raw one, and every other producer (`CbomSyncTask`, the sweep, `UpdateCertificateStatusTask`) already shapes its own.

src/test/java/com/otilm/core/service/writer/scheduler/ScheduledJobHistoryWriterContractTest.java
karolhrdina: Replaced in e7dc5501. `ScheduledJobFinishedEventITest` publishes a real `ScheduledJobFinishedEvent` from inside a `TransactionTemplate` and asserts the row is SUCCESS with an end time after the commit — your probe, in the suite, on the shared base with no local mocks. `SchedulerTransactionContractTest` pins `NOT_SUPPORTED` on the three entry points instead of the writer's propagation, which Rule D now covers.

src/main/java/com/otilm/core/messaging/jms/listeners/SchedulerListener.java
karolhrdina: Deliberate, and now stated in the listener comment and the PR body. `RuntimeException`: swallowed, because it is the job's own failure and a redelivery would repeat it against a closed row. `Error`: not caught anywhere (Sonar's rule and the platform's habit agree; in e7dc5501 the row is closed from a `finally` instead), so the JMS retry policy keeps the redelivery decision — a transient one (memory back, a class present after a redeploy) gets its retry, a deterministic one exhausts the policy. On the second point: before this PR a failure before the STARTED row — `findByJobName`, the history insert — was already swallowed by `AbstractJmsEndpointConfig`'s `catch (Exception)`; only `JmsException | JMSException | IOException` were ever retried, and the one place those arise here is `produceMessage`, after the status commit, where a redelivery would re-run the job. So the trade is unchanged before the row and made explicit after it. Leaving this thread to you.

vyskocilm (approved):

PR #2265, opened 2026-09-16
Read the CBOM sync policy tunables from the platform settings

### LINKED ISSUE #2249: Make the CBOM sync tunables configurable from the Settings UI

karolhrdina:
Lane B of OmniTrustILM/ilm#299, story #2249. core#2245 and core#2256 made the CBOM header sync a scheduled job whose policy sat in `application.yml`, where a SaaS operator cannot reach it. Ruling of 2026-09-15 (in the issue): the **policy trio** moves to the platform settings; the protocol and lock plumbing stays deploy-time.

## What changes

| Value | Before | After |
|---|---|---|
| overlap | `cbom.sync.overlap` (60 s) | platform setting `cbomSyncOverlapSeconds`, default 60, 0..86400 |
| retry budget | `cbom.sync.skipped-retry-runs` (3) | platform setting `cbomSyncSkippedRetryRuns`, default 3, 0..1000 |
| ingest budget | `cbom.sync.max-ingest-documents` (50) | platform setting `cbomSyncMaxIngestDocuments`, default 50, 0..10000 |
| page size, asset-ingest kill switch, asset batch size, ingest retry window | `cbom.sync.*` | unchanged, still `CbomSyncProperties` |

- **`CbomSyncPolicy`** (new record in `com.otilm.core.cbom.sync`): the three values, their defaults, the contract's caps (`UtilsSettingsDto.MAX_CBOM_SYNC_*`, so core refuses exactly what the API refuses), `maxAttempts()`, and `fromSettings(UtilsSettingsDto)` with per-field fallback to the default.
- **`CbomServiceImpl.runSync`** snapshots the policy once from `SettingsCache` at the start of a run and keeps it on `SyncRun`: a setting changed mid-run applies to the next run on this node, other nodes follow within the 30 s cache refresh. One INFO line per run states the policy it runs with. `getLastSyncTimestamp` takes the overlap as a parameter; `ingestPending` reads the ingest budget from the run.
- **`test_docker_image.yaml`** reads the same dependency marker as `build_pr.yml` (same loose probes, same publish wait after the merge): while the interfaces PR it names is unmerged, or merged but not yet published, the image test waits (skipped with a notice and a step summary) instead of failing to compile against the published snapshot. The shared container workflow takes no Maven argument, so building the image against the PR snapshot would need an input added in `OmniTrustILM/.github`; merging stays blocked meanwhile by `Build`, so nothing is skipped past.
- **`CbomSyncProperties`** shrinks to `pageSize`, `assetIngestEnabled`, `assetBatchSize`, `ingestRetryAfter`; the three keys and their `${CBOM_SYNC_*}` placeholders leave `application.yml`.
- **`SettingServiceImpl`**: the utils section reads back with the defaults filled in, so the UI shows what the sync will use. A stored value that is not an integer or lies outside the contract's bounds (only a manual DB edit can put one there) reads as the default and is reported when it appears, whenever its text changes, and again after it was corrected, not on every 30 s refresh (the last reported text is kept per setting name on the bean; a valid or unset read clears it). On write the utils section is stored as sent: an unset value deletes its row (as the branding writes do) instead of persisting a `NULL` row; the find-or-create of a platform setting row is one helper for the certificates and utils categories; and, as for branding, a transaction-scoped advisory lock (`lockUtilsWrites`) is held before the rows are read, so two concurrent first updates cannot insert one name twice (the two URLs had that race before; five rows would have). The certificates rows go through the same find-or-create, so they get the same lock (`lockCertificateWrites`), taken after the utils one in a fixed order.

## Operator notes

- `CBOM_SYNC_OVERLAP`, `CBOM_SYNC_SKIPPED_RETRY_RUNS` and `CBOM_SYNC_MAX_INGEST_DOCUMENTS` are **ignored** from this PR on (unknown `cbom.sync.*` keys bind to nothing). They never shipped in a tagged release and were never wired in Helm, so there is nothing to remove there; a `main`-tracking deployment that set them now runs at the defaults until the values are set in Settings → Platform → Utils.
- A client that `PUT`s `utils` without the three fields resets them to the defaults (the section is stored as sent, as it always was for the URLs). The fe-administrator PR ships the fields; until it is deployed, saving the utils form from an older UI build resets the policy to 60 / 3 / 50.

## Tests

- `CbomSyncPolicyTest` (7): defaults, missing section, per-field fallback, seconds conversion, negatives and cap+1 refused, caps accepted.
- `CbomSyncPropertiesTest` (6): the four remaining values.
- `SettingServiceMockedTest` (2, plain Mockito, no Spring context): a platform update takes the utils lock, then the certificates lock, before reading the rows; a section left out takes no lock.
- `SettingServiceITest` (+5): defaults reported and editable and landing in `SettingsCache`; an unset value returns to the default, the URL is cleared, no row left behind; stored `99999999` / `-5` / `fifty` read as the defaults; `aSectionOrCertificatesGroupLeftOutOfThePlatformUpdateIsUntouched` pins the PUT semantics the interfaces operation documents (a section left out is untouched; inside `certificates` a group left out is untouched and a present one is stored as sent); `aCorruptCbomSyncTunableIsReportedWhenItAppearsAndAgainAfterACorrection` pins the warning cadence with a log appender: once while the corrupt text stays, again after a correction and re-break. `SettingControllerBrandingITest` (3) still green.
- `CbomSyncITest` (+1): with a retry budget of 0 in the platform settings an unreadable entry is `PERMANENTLY_SKIPPED` after one run; the bound-defaults test asserts only the deploy-time record.
- Wide run before push: `CbomSyncPolicyTest, CbomSyncPropertiesTest, CbomSyncITest, CbomSyncTaskITest, CbomAssetIngestServiceTest, CbomAssetDetachServiceTest, CbomServiceITest, SettingServiceITest, TransactionalBoundaryArchTest` green; `spotless:check checkstyle:check` clean.

## Review passes

Two independent max-effort adversarial agents (Opus, Fable) over `git diff origin/main...HEAD` and `/code-review high` (7 verifier batches). Both agents and the review converged on one MEDIUM: an out-of-range stored value reached `CbomSyncPolicy` and bricked every run with an opaque history message → the read path now treats it like non-integer text (default + one warning). Also folded in: the WARN-per-refresh noise, `NULL` rows on unset, the duplicated find-or-create, the `maxAttempts` copy on `SyncRun`, stale Javadoc naming the deleted property, the contradiction in `CbomSyncProperties`' Javadoc, an ITest that re-proved a unit test, a no-op re-cache in a test. A focused re-review of the fixes → SHIP-AFTER-FIXES (the dependency marker below) plus the utils write lock, an exact cap comparison and two Javadoc precisions, all in. Refuted by the reviewers and left as is: settings cache generations straddling a run (URL and policy are independent knobs), boot-order null cache (both entry points gate on the repository client being configured), env-var removal (never shipped). Michal's review (2026-09-16, incl. the ILM architecture gate: PASS WITH GAPS, no violation): both MEDs folded in — the warn-once set became a per-name last-reported value that clears on a valid read, and the certificates rows got the same advisory lock as the utils rows, in a fixed order; the two LOWs (cache repopulated after the lock releases; the compact-constructor guard is defence in depth, the read-side clamp is the live protection) are recorded as he had them.

## Merge order

1. ~~interfaces PR 974 merges first and publishes its `-M<build>` snapshot~~ — done 2026-09-16 (merge `bed419f1`). While it was open, a `Depends-On` marker in the first line of this body made the `Build PR` workflow build and test against the PR snapshot, kept `Interfaces pin` and `Build` red on purpose, and made the `Test Docker image` workflow wait.
2. ~~Then: remove that first line, re-run all jobs on both workflows (`Build PR` and `Test Docker image`)~~ — done; this PR now builds against mainline `2.20.0-SNAPSHOT`. **Blocked on core#2261**: mainline interfaces also carries interfaces#975 (cryptography provider v2, merged 2026-09-16 15:10), whose two new controller methods (`listVerifyAttributes`, `listSupportedKeyRequestTypes`) core `main` does not implement yet; #2261 does. Until it merges, every core PR that builds against mainline fails `Compile` on those two classes, this one included; nothing in this diff is involved. Once #2261 is in: rebase this branch, CI green, merge. It lands **before core#2250**, which adds its retention setting to the block this PR creates.
3. fe-administrator PR https://github.com/OmniTrustILM/fe-administrator/pull/2149 after this is deployed to `main`.

The `Depends-On` line stays out of this body from here on: with the dependency merged, it would only re-pin the build to a snapshot that no longer moves.

Closes #2249 (the operator-facing acceptance criterion is met once the fe-administrator PR lands; the platform side is complete here).


vyskocilm:
Deep review of this PR and its pair interfaces#974, against the ratified ruling in core#2249 ("Scope ruling 2026-09-15"). Core-side findings here; the contract-side ones are on interfaces#974.

I also ran the ILM architecture gate over head `6eef3e99`, built against interfaces#974 head in a throwaway worktree: **`GATE: PASS WITH GAPS`, no checklist violation.** 133 architecture tests and 32 unit tests green. Worth recording what it confirmed: no OPA bundle, permission seed or auth migration is authored and none is needed; `lockUtilsWrites()` mirrors `lockBrandingWrites()` and takes one advisory key per transaction, so no lock-ordering cycle; `ContextSignatureGuardTest.BASELINE` is correctly untouched — no new Spring context bought; the three `${CBOM_SYNC_*}` placeholders leave and an org-wide search finds no chart reference, so there is no Helm handoff. One ruling line — "`CbomAssetIngestService` reads `maxIngestDocuments` per call" — is moot rather than missed: that class never held the value.

Implementation matches the ruling item for item. Findings below are MED at worst.

**MED — `REPORTED_CORRUPT_UTILS_VALUES` suppresses a legitimate second warning, and never shrinks.** `SettingServiceImpl`, static key set on `name + "=" + value`. An operator who breaks a value, sees the warning, corrects it and later re-breaks it with the same text is never warned again for the life of the JVM — while the Javadoc promises "until it is corrected", and nothing un-reports on correction. It is also static mutable state no Spring test context resets: no flake today because no test asserts the warning, one waiting for the first test that does. A `ConcurrentHashMap<String, String>` holding the last reported value *per name*, warning when it changes, fixes both: three entries at most, and a re-break warns again.

**MED — the lock asymmetry is now shared-helper-shaped.** `lockUtilsWrites()` is right: taken before `findBySection` reads the rows, held by the class-level `@Transactional`. But `updateCertificateSettings` uses the same newly-extracted `platformSetting()` find-or-create and takes no lock, so the double-insert race this PR closes for utils stays open on `PLATFORM_CERTIFICATES`. Pre-existing and out of scope — the extraction is what makes it read as settled. Either add `lockCertificateWrites()`, or say on `platformSetting()` that the caller holds its category's lock. (Reached independently by the gate.)

**LOW — the cache is repopulated after the lock is released.** `cacheAfterCommit(...)` runs post-commit, so `pg_advisory_xact_lock` is already dropped. Two concurrent utils updates serialise their writes, but their post-commit re-reads can interleave, leaving `SettingsCache` on the earlier writer's state until the 30 s refresh. Identical to the branding path, bounded and self-healing, and this PR does not make it worse. Recording it, not asking for a fix here.

**LOW — `CbomSyncPolicy`'s compact constructor is unreachable from production.** `fromSettings` is fed by `utilsInteger`, which has already clamped to `0..cap`, so none of the four throws can fire from the only production caller — as its own Javadoc says. Defence in depth is fine. Worth naming only so the test record is read correctly: `CbomSyncPolicyTest`'s "negatives and cap+1 refused" proves the guard, and the clamp is what actually keeps a run alive on a corrupt row. `SettingServiceITest.aStoredCbomSyncTunableOutsideTheContractBoundsReadsAsItsDefault` is where that is proved, and it does.

## CI

The three red checks are the deliberate dependency gate, not defects: `Interfaces pin` fails with `Building against an unmerged interfaces PR. Merge it, wait for the publish, then re-run all jobs.` (`MODE: pr`), `Build` fails only because that gate did, and `test / Build app artifacts (once)` fails because the Docker workflow has no marker handling. Compile, all four Test jobs, SonarCloud and CodeQL green at this SHA.

Not verified on my side: the integration suites locally (CI ran them green at this SHA) and coverage parity (Sonar green).

karolhrdina:
Both MEDs folded in (dfd719cb), plus the CI half:

- **Warn-once set → per-name last-reported value.** `lastReportedCorruptUtilsValue` is a `ConcurrentHashMap<String, String>` on the bean instance, no JVM-wide state: a corrupt value is reported when it appears, whenever its text changes, and again after a valid or unset read has cleared the entry; three entries at most. `SettingServiceITest.aCorruptCbomSyncTunableIsReportedWhenItAppearsAndAgainAfterACorrection` pins the cadence with a log appender: once while the text stays, again after a correction and re-break.
- **Lock asymmetry.** `lockCertificateWrites()` beside `lockUtilsWrites()`, both taken in `updatePlatformSettings` before the rows are read — utils first, then certificates, always in that order, so the two keys cannot wait on each other; `platformSetting()` says its caller holds the category's lock, and the repository comment says what a duplicate row would do (the read maps by name last-wins, so one copy would silently win and the other would never be updated or deleted again). `SettingServiceMockedTest` (plain Mockito, no Spring context bought) pins the order and that an absent section takes no lock.
- The two LOWs are recorded as you had them.
- **CI.** `Interfaces pin` and `Build` stay red until interfaces#974 is merged and published: the gate. The `Test Docker image` workflow now reads the same marker (`interfaces-dependency` job: the loose probes of `build_pr.yml`, and its publish wait after the merge): while the dependency is unmerged or unpublished the image test waits with a notice instead of failing on the contract, its required contexts show as expected, and `Build` blocks the merge meanwhile. The shared `containers-test.yml` takes no Maven argument, so building the image against the PR snapshot itself would need an input added in `OmniTrustILM/.github`. Once #974 publishes: remove the marker line, re-run all jobs on both workflows.

karolhrdina:
Status after interfaces#974 merged (2026-09-16): the `Depends-On` marker is out of the body, `Resolve interfaces version` now reads "merged - using the mainline pom value", `Interfaces pin` is green, and the `Test Docker image` gate lets the image build run. Both workflows were re-run on `dfd719cb`.

`Compile` and `test / Build app artifacts (once)` fail, but not on anything in this PR:

```
CryptographicOperationControllerImpl.java:[31,8] ... does not override abstract method listVerifyAttributes(...)
TokenProfileControllerImpl.java:[37,8] ... does not override abstract method listSupportedKeyRequestTypes(...)
```

Mainline `2.20.0-SNAPSHOT` carries interfaces#975 (cryptography provider v2, merged 15:10 today, before #974), and core `main` does not implement those two controller methods yet; core#2261 does, and it is green against mainline. Until #2261 merges, every core PR that builds against mainline fails the same way. Once it is in, this branch gets rebased and re-run; no change to this diff is needed.

vyskocilm (approved):

PR #2266, opened 2026-09-16
Report what a CBOM's asset ingest found, and reconcile against the whole listing weekly

### LINKED ISSUE #2073: Extend CBOM sync with per-CBOM asset onboarding and deletion lifecycle

vyskocilm:
Closes #2073 

Closes the last two parts of core#2073's plan: the ingest findings channel with the duplicate-`bom-ref` refusal, and the weekly reconciliation. They were planned as PRs 4 and 5 of a five-PR stack; combined they sit inside the 1K–3K band, so they are one review rather than two. PRs 1–3 merged as #2256, #2257 and #2263.

## Refuse a document that repeats a `bom-ref`

CycloneDX requires `bom-ref` to be unique within a document. A repeat is invalid input rather than a shape to resolve — both readings of it move keys — so the ingest refuses the whole document.

The rule lives in `CbomAssetIngestService`, **not** in `DocumentScope`. `DocumentScope`'s "a duplicated ref resolves to nothing" rule stays exactly as it was, because it is also how a component is keyed outside ingest — the identity vector runner and the corpus measurements — where there is no document to refuse. `IdentityVectorTest`, `NormalizationRulesTest` and `CorpusKeySnapshotTest` are green: no accepted document's key moves.

**The refusal reads wider than the index does.** `bom-ref` uniqueness is scoped to the document, and three sections define one: `metadata.component`, `components[]` and `services[]`. The component index is built from `components[]` alone, because that is the only section a reference the extractor follows can resolve into. So `DocumentScope.duplicatedRefs` is a second bounded traversal over all three, and `ambiguousRefs()` is a superset of the refs `resolve` answers nothing for. Feeding the wide set back into the index would move the key of a component whose own ref is unique, on the strength of a repeat in a section nothing resolves into. `dependencies[]` cites definitions rather than making them, so it is not read.

`Extraction` gained `ambiguousRefs`. A ref with no UTF-8 encoding is replaced by a stand-in rather than dropped, so the count does not understate what the document carries.

## The findings channel

`MaterialRedaction.findings()` and `Extraction.skips` were computed and then discarded at the extraction boundary. `cbom_ingest_finding` is where they land.

- **Rolled up per distinct message**, not per component: one document reports the same finding for thousands of components and the count is the useful part. Each kind (`FINDING`, `SKIP`) keeps its 100 most frequent and the rest are counted in the log — a finding names the producer member it is about, so distinct messages are as unbounded as the document is.
- The roll-up kernel, `IngestFindingRollup`, is a pure function with its own unit tests and no Spring.
- The report is written **in its own transaction, before any refusal**: a document refused for what the report names owes the operator that report most of all, and enrolling it in the asset writes would roll it back with them. It **replaces** the previous attempt's rows.
- The FK to `cbom` is `ON DELETE CASCADE`, unlike `crypto_asset_source`'s `RESTRICT`: a finding is a statement about a document and means nothing once the document is gone.
- `detail` is cut to 512 **code points** on a code-point boundary, with `CHECK` backstops in the migration and the entity. It is a column of the report's unique index, so a btree tuple over ~2704 bytes is refused outright, and the text is producer-derived and otherwise unbounded. The cut happens before grouping, which is what keeps a `uq_cbom_ingest_finding` violation unreachable.
- A superseded revision's report goes with the contribution it withdraws — on the mainline supersession path as well as through `supersede()` — and so does an extraction that threw, the one refusal path that writes no report of its own.
- `CbomIngestFindingMigrationITest` runs the migration as Flyway will. The suite generates its schema from the entities, so nothing else executes that file, and the entity deliberately restates the same constraints under the same names; the test is what proves the two copies agree.

`FilterField.CBOM_ASSET_SYNC_ERROR` makes a refusal's reason searchable. It is registered in the CBOM searchable-field list **and** in `SearchHelper.ABSENT_FROM_LISTING`, which is what every other member of that set does: `ABSENT_FROM_LISTING` stops a registered field being offered as a column, it is not an alternative to registering it. `Cbom.mapToDto` never sets the reason and `CbomDto` has no property for it, so a column would be a heading over empty cells. Filtering works — the predicate reads the entity's column, which the listing simply does not return.

## A refusal the document earned is not retried for ever

Every asset-ingest failure before this PR was transient — a lock, a read, a database error — so the backlog needed no attempt counter and no terminal state. A repeated `bom-ref` is not: the extraction is a pure function of the document, so a retry spends an HTTP read, an extraction and one of `cbom.sync.max-ingest-documents` slots to reach the same verdict, every run, for ever.

`cbom.asset_sync_content_refusals` counts the refusals the document itself earned — `ambiguousRefs` and `documentScopeUnavailable`, and nothing else — and `findAssetIngestRetries` stops offering the row at `MAX_CONTENT_REFUSALS`. A counter rather than a flag because the two are not equally deterministic: `documentScopeUnavailable` is raised by an extraction that *threw*, which could have been the moment rather than the document. Every transient failure still goes through `markFailed`, leaves the counter alone and is retried as before.

The row stays `FAILED`, carrying the reason an operator can read and filter on. A terminal *state* is the better answer and needs a `CbomAssetSyncState` constant, so an `interfaces` release — the same reason `markSuperseded`'s javadoc already records.

## Weekly reconciliation

`CbomReconcileTask`, Sundays 02:30, registered in `SystemScheduledJobs`. It calls `CbomInternalService.reconcile()`, which is `runSync(THE_WHOLE_LISTING)` — **the same run** from `after = 0`, so every rule about duplicates, tombstones, skips and ingest holds for both passes rather than being implemented twice. Being one run, the two openings spend the same `cbom.sync.max-attempts` budget, which `SyncScope`'s javadoc now says.

What it reaches that the hourly pass cannot: `cbom.sync.overlap` and the skip retry budget cover an entry the feed *offered* and Core then failed on. Neither covers one the feed never offered inside any window Core asked for — a run that recorded SUCCESS after the listing moved past an entry, a back-dated document, an entry written off while its window closed. The overlap is not the answer to those; this is.

It does not touch the hourly watermark in either direction: that is read from `CbomSyncTask`'s own job history, and this run is recorded under `CbomReconcileTask`. `CbomSyncTask` and `CbomReconcileTask` now share `AbstractCbomFeedTask`.

### The outage verdict had to become scope-aware, and bounded

`looksLikeOutage` was `successfulReads == 0 && feedDeferred >= 2`. A whole-listing pass over an estate Core is already in step with reads **no documents at all** — almost every entry it is offered is already stored and is settled by an indexed existence check — so the first half of that test is satisfied by a *healthy* run rather than by a sick repository. Two broken documents would then have thrown `SERVICE_UNAVAILABLE`, abandoning the pass **before `ingestPending`** and charging nothing, for ever, in precisely the "a few entries are stuck" case the weekly pass exists to repair.

A whole-listing pass now diagnoses an outage only when nothing in the listing **was already stored**. That is `alreadyStored`, a counter of its own, not the aggregate `duplicates` the summary line reports: `duplicates` also counts an identity the feed offered twice within one run — a page overlapping a retried cursor does that — and one such repeat would flip the verdict for an entire run, in the direction that charges the whole estate. The hourly pass keeps the original rule, which its own overlap would otherwise defeat.

**And no run charges more than `MAX_CHARGED_READ_FAILURES_PER_RUN` entries to the retry budget.** Nothing else bounded it: `run.deferred` grows with the listing, and the weekly pass lists the estate. A repository serving its listing but not its documents would otherwise charge every entry Core does not hold, once a Sunday, writing the estate off in as few as `skipped-retry-runs + 1` of them — one, at the `0` that configuration allows. The cap raises the outage verdict rather than truncating the charges, because charging is all-or-nothing by construction: an uncharged entry has no skip row, so the retry pass never offers it again and the hourly watermark has moved past it.

The verdict is now a package-private pure function of its four counters, so every arm — including the cap, which a run of a thousand stubbed entries is the only other way to reach — is decided by a test that states the counters. The scope arms are pinned by integration tests as well.

## Registering a system job is race-free

`scheduled_job.job_name` is what every lookup resolves a job by, and nothing declared it unique. Registration is a read-then-write with no transaction of its own and no cluster lock, so two nodes booting together both read "absent" and both insert — after which `Optional<ScheduledJob> findByJobName` throws on every trigger, enable, disable and delete of that job, silently, for as long as the estate lives. The window has been open for every system job; for the existing four the row was written on a first-ever boot, before there was a cluster to race. A job added to estates that are already running and will be rolling-restarted lands squarely in it.

`V202609161300` adds `uq_scheduled_job_job_name`. The half in front of it is the careful part: `scheduled_job_history`'s FK is `ON DELETE CASCADE`, so the migration repoints any existing duplicates' history at the surviving row before deleting them. `registerScheduler` catches the violation and re-reads — the row the other node wrote *is* the registration — while any other refusal is still an error, with the driver's text logged rather than returned.

## Orphan collection carries its own condition

Carried from #2257's review. `crypto_asset`'s delete was unconditional, so `findOrphanRows` and the delete were two statements with a window between them. Nothing can attach a source in that window today — the withdrawal and the ingest both write under `ALIAS_DECISION_LOCK` — but that is a property of the callers, and it is the statement that destroys the row. It now deletes only while the row still has no source and no alias names it, and a declined delete is counted as *kept*, which is what the row then is.

The alias arm is the one that matters: `crypto_asset_alias` cascades from `canonical_key`, so a delete that raced an operator pointing an alias at the row would take that merge decision with it, and no re-ingest would bring it back.

**What this does not close**, and deliberately: an ingest of a *different* document that will source the same asset in a later batch can still have the row collected in between, so the asset returns under a fresh UUID with its PQC verdict re-stamped by the sweep. That is recoverable and invisible to an operator; the irreversible half — the merge decision — is what the alias arm protects. Closing the rest means not deleting on the withdrawal path at all, which is a lifecycle change and a ruling, not a fix.

## Verification

- `mvn verify`: **8464 tests, 0 failures, 0 errors**.
- New-lines coverage **95.90%**, computed with Sonar's branch-inclusive formula `(covered_lines + covered_branches) / (lines_to_cover + branches_to_cover)` against `git diff origin/main...HEAD` with `pom.xml`'s `sonar.coverage.exclusions` applied.
- Architecture gate (`/ilm-check`) run and its findings closed. Every new test is mutation-checked, with the mutation asserted to have applied.
- Review round from @karolhrdina and Copilot: all eleven findings fixed, each pinned by a test that fails without the fix.

## Open work — ticketed, please skip these in review

- **[core#2267](https://github.com/OmniTrustILM/core/issues/2267)** — serving `assetSyncError` and the `cbom_ingest_finding` report to clients. Needs a `CbomDto`/`CbomDetailDto` field, so it is an `interfaces` release. This is why `cbom_ingest_finding` is **written but not read by any production path** in this PR. There is no missing endpoint to look for.
- **[core#2268](https://github.com/OmniTrustILM/core/issues/2268)** — the `cbom.sync.*` tunables move out of `application.yml` into Settings. Ruled 2026-09-16: nothing lives in yml that need not, because an ILM user cannot reach it there. This PR introduces no new one, and nothing is to be wired into the Helm charts.


src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -408,6 +475,7 @@ private IngestOutcome supersede(UUID cbomUuid, CbomAssetSyncState entryState) {
                     .debug("CBOM asset ingest: CBOM {} is superseded; withdrew the {} links it had already contributed",
                             cbomUuid, withdrawn.detached());
         }
+        clearReport(cbomUuid);
```
karolhrdina, line 504: **The mainline supersession path does not clear the report.** The PR states *"A superseded revision's report goes with the contribution it withdraws"*, and the javadoc at `:463-466` says the same — but this loop, which is where a revision's contribution is actually withdrawn on supersession, never calls `clearReport(superseded)`.

`clearReport` has exactly two call sites (`:241` and `:478`), and both pass the CBOM currently being ingested. `:478` lives in `supersede(cbomUuid, …)`, whose three callers (`:222`, `:283`, `:294`) likewise all pass the CBOM being ingested. Nothing anywhere passes a uuid from `findSupersededVersionUuids`.

So for the ordinary case — v1 ingested and `SYNCED`, v2 arrives:

1. v2's batches write its sources;
2. `withdrawSupersededVersions(v2)` detaches **v1's** links here (`:359`);
3. v1 now sources nothing, its row still reads `SYNCED` from its own earlier ingest, **and its ingest report is still attached**.

Nothing later repairs it. `supersede(v1)` is the only thing that would clear the report, and it runs only if v1 is fed through `ingest`/`settleWithoutReading` again — but a `SYNCED` row is selected by neither work list (`findPendingAssetIngests` takes `PENDING`, `findAssetIngestRetries` takes `{IN_PROGRESS, FAILED}`). The report survives until the CBOM itself is deleted and the `ON DELETE CASCADE` takes it.

The result is a revision reading `SYNCED`, sourcing nothing, carrying a report about assets it no longer contributes — which is exactly the state the invariant was written to exclude. Once #2267 serves these to clients, that report is what an operator sees against a revision that has been silently emptied.

One line next to `:359` fixes it:

```java
for (UUID superseded : cbomRepository.findSupersededVersionUuids(cbomUuid)) {
    final CbomAssetDetachService.Withdrawal withdrawn = detachService.withdraw(superseded);
    ...
    if (!withdrawn.complete()) {
        return false;
    }
    clearReport(superseded);
}
```

after the completeness check, so an abandoned withdrawal leaves the report in place with the links it still describes.

`aSupersededVersionsReportGoesWithIt` (`CbomAssetIngestServiceTest.java:496`) covers only the `hasIngestedLaterVersion` → `supersede` path, which is why this is green.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -1099,16 +1138,16 @@ private void ingestInline(UUID cbomUuid, Map<String, Object> document, SyncRun r
...
+     * are offered again. Otherwise each deferred document, a retried one included, is charged for its own failure,
+     * exactly as any other failed entry is.
      */
-    private void settleUnavailable(SyncRun run) throws CbomRepositoryException {
+    private void settleUnavailable(SyncRun run, SyncScope scope) throws CbomRepositoryException {
         if (run.deferred.isEmpty()) {
             return;
         }
-        if (looksLikeOutage(run)) {
+        if (looksLikeOutage(run, scope)) {
```
karolhrdina, line 1206: **Nothing caps how many entries one reconcile run charges, and one configuration makes a single bad Sunday terminal.**

When this returns false the run falls through to `for (DeferredSkip deferred : run.deferred) recordFailure(...)`. `run.deferred` is a plain unbounded `ArrayList`, and the only ceiling in the run at all is `MAX_PAGES_PER_RUN = 10_000` at a page size of up to 1000 — so up to ten million listed entries, every one of which becomes a charged `cbom_sync_skip` attempt if Core does not already hold it.

The hourly pass never reached that scale: its window is one hour wide. The whole-listing pass is the entire estate, every Sunday, and the case that triggers it is not exotic — listing endpoint up, document endpoint down, Core holding at least one listed entry (see the thread on `run.duplicates`).

With defaults it takes four such Sundays to write the estate off: `maxAttempts() = 1 + skippedRetryRuns`, default `3` ⇒ 4, and one run charges exactly one attempt per entry (`run.attempted` dedups, and `retrySkippedEntry` returns early for anything the feed pass already touched).

**But `cbom.sync.skipped-retry-runs: 0` is accepted** — the validation only rejects negatives (`CbomSyncProperties.java:59-62`). That makes `maxAttempts == 1`, and `CbomSyncSkipRepository`'s `CASE WHEN 1 >= :maxAttempts THEN 'PERMANENTLY_SKIPPED'` arm fires on the first failure. One Sunday, whole estate written off.

The damage is recoverable — the weekly pass re-offers written-off entries through `processFeedEntry`, which never consults the skip table as a gate, so a healthy repository resolves them — but the intervening state is thousands of skip rows, N wasted HTTP reads, a per-entry WARN storm, and a `permanentlySkipped` figure in the job history that will read as a catastrophe.

Two cheap guards, either one sufficient: a per-run cap on charges, or "charged more than some fraction of the listing ⇒ treat as an outage after all", which is the same evidence the verdict already reasons about.

**Related, and not mentioned in the description:** `runSync(THE_WHOLE_LISTING)` calls `retrySkipped(run, skips)` exactly as the hourly pass does, so the weekly pass spends from the *same* `maxAttempts` budget. An entry now reaches `PERMANENTLY_SKIPPED` up to a run earlier than before this PR, with no overlap required. Worth a sentence in the reconcile javadoc at minimum.

One more, while here: `CbomReconcileTask.java:23-24` says the task "occupies the same listener as every other scheduled job -- while it holds that thread the jobs behind it wait." `messaging.concurrency.scheduler: 10` (`application.yml:353`) means they do not, across ten threads per node and N nodes. That matters because it is the reason a reconcile and an hourly sync can overlap at all, which the comment is trying to reason about.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
```diff
@@ -231,9 +235,33 @@ private IngestOutcome ingestExtracted(UUID cbomUuid, JsonNode document, OffsetDa
...
+            // the backlog pass, and strand this CBOM at IN_PROGRESS until the retry window expires.
+            log.warn("CBOM asset ingest: recording the ingest report failed for CBOM {}", cbomUuid, e);
+            return fail(cbomUuid, "the cryptographic asset ingest report could not be stored (see the Core log)");
+        }
+
+        if (!extraction.ambiguousRefs().isEmpty()) {
+            // CycloneDX requires bom-ref to be unique, so a document that repeats one is invalid input rather than a
+            // shape to resolve -- and both readings of a repeat are wrong in a way that moves keys. The refused
+            // document keeps its row and its reason; the producer's fix is a new version.
+            return refuse(cbomUuid,
```
karolhrdina: **This refusal is deterministic and permanent, but the retry machinery treats it as transient.**

`refuse(...)` is `markFailed` (`:540-543`), so the row goes back to `FAILED`. The asset-ingest backlog then re-offers it on the next run whose `retryBefore` has passed:

```
WHERE c.assetSyncState IN (IN_PROGRESS, FAILED)
  AND (c.assetSyncAttemptedAt IS NULL OR c.assetSyncAttemptedAt < :retryBefore)
```
(`CbomRepository.findAssetIngestRetries`; `ingest-retry-after` default 30m, `application.yml:433`)

There is no attempt counter and no terminal state on the asset-ingest side — `cbom` carries `asset_sync_attempted_at` and nothing else, and `markFailed`'s guard (`NOT_YET_SYNCED = {PENDING, IN_PROGRESS, FAILED}`) admits `FAILED`, so each pass just re-stamps the clock. So a document with a repeated `bom-ref`:

* is re-read from the CBOM repository over HTTP on every run, for ever — `ingestOnePending` reads the document before it can discover the refusal;
* is re-extracted and has its findings report deleted and rewritten each time (`recordReport` runs before the refusal);
* occupies one of the 50 slots of `cbom.sync.max-ingest-documents` on every run.

The comment says *"the producer's fix is a new version"* — which is right, and is exactly why this should stop being retried. Nothing about the document will change between attempts; the refusal is a pure function of its content.

The header pass already has the concept this needs: `cbom_sync_skip` counts attempts against `cbom.sync.max-attempts` and settles at `PERMANENTLY_SKIPPED`. The asset ingest has no equivalent, and before this PR it did not need one — every previous `REFUSED`/`FAILED` reason was transient (a lock, a read, a database error). `ambiguousRefs` and `documentScopeUnavailable` (`:255-271`) are the first refusals that are properties of the document itself.

With N such documents in an estate, that is N document reads per hour against the repository, for ever, and `min(N, 50)` backlog slots per run spent on work that cannot succeed. `ORDER BY assetSyncAttemptedAt NULLS FIRST` keeps them from starving the retryable ones outright, but they still take their share of every run.

Worth a terminal state (or at least an attempt bound) for the content-derived refusals specifically — the transient ones should keep retrying as they do now.

src/main/resources/db/migration/V202609141400__cbom_ingest_finding.sql
```diff
@@ -0,0 +1,32 @@
+-- What one CBOM's asset ingest had to tell the producer about the document, and what it could not turn into an asset.
+-- The extraction already computed both -- a dropped value, a withheld digest, an inlined secret, a component that
+-- could not be keyed -- and then discarded them at the boundary, so a document could be refused with no record of
+-- why and ingested with no record of what it cost. Rows are rolled up per distinct message rather than per
+-- component: one document reports the same finding for thousands of components, and the count is the useful part.
+CREATE TABLE "cbom_ingest_finding"
```
karolhrdina, line 6: **Nothing in the suite ever executes this file.**

`src/test/resources/application.yml:135-136` sets `spring.flyway.enabled: false` and `:140` sets `ddl-auto: create-drop`, so the test schema is generated from the entities. `grep -rl cbom_ingest_finding src/` returns four files — this migration, the entity, the repository, and one javadoc mention in `CbomAssetExtractor` — and none of them is a test. So a typo here, a constraint that never compiled, a wrong `ON DELETE` action, or a missing index would all ship green.

The repo already has the answer to this, and two of its instances are about this very subsystem: `CryptoAssetInventoryMigrationITest` and `CryptoAssetCurveMembershipMigrationITest`, both under `src/test/java/com/otilm/core/integration/migration/`. The first one's javadoc states the risk in as many words:

> *"The regular test bootstrap generates its schema from the entities, so nothing else in the suite ever executes this file: an index that was never created, a foreign key with the wrong delete action, or a check constraint that never compiled would all pass unnoticed."*

It pins `EXPECTED_INDEXES`, `EXPECTED_FOREIGN_KEY_ACTIONS`, `EXPECTED_CHECK_CONSTRAINTS` and the unique-constraint names for `crypto_asset*`. `cbom_ingest_finding` has no equivalent.

This matters more than usual here because the invariants are deliberately written twice — the entity's own comment says so:

> `CbomIngestFinding.java:48-49`: *"Stated here as well as in the migration, so the entity-generated schema the tests run against carries the same invariants and the same constraint names as production."*

The duplication is the design, and nothing proves the two copies agree. They **do** agree today — I checked every CHECK, the unique constraint name, the FK name and its `CASCADE`, and column nullability, line by line against the entity — but that is a hand check that expires with this review.

A `CbomIngestFindingMigrationITest` in the same shape as the two siblings would close it.

src/main/resources/db/migration/V202609141400__cbom_ingest_finding.sql
```diff
@@ -0,0 +1,32 @@
...
+    "component_name" TEXT,
+    -- Shaped by Core and operator-visible. A finding names the member it is about, never the member's value.
+    "detail"         TEXT        NOT NULL,
+    -- How many components raised this exact message.
+    "occurrences"    INT         NOT NULL,
+    "recorded_at"    TIMESTAMPTZ NOT NULL,
+    PRIMARY KEY ("uuid"),
+    -- CASCADE, unlike crypto_asset_source's RESTRICT: a finding is a statement about one document and means nothing
+    -- once that document is gone, where a source carries inventory provenance that must not vanish silently.
+    CONSTRAINT "cbom_ingest_finding_to_cbom_key" FOREIGN KEY ("cbom_uuid") REFERENCES "cbom" ("uuid") ON DELETE CASCADE,
```
karolhrdina, line 23: This FK falsifies a load-bearing comment that #2263 left in `CbomServiceImpl`, at `:523-528`:

> *"Narrowed to the one constraint that expresses that: `crypto_asset_source_to_cbom_key` is today the only foreign key in the schema referencing `cbom(uuid)`, and the tombstone insert cannot conflict, so every violation reaching here is the inventory's. ... the next foreign key added to `cbom` would otherwise silently join it."*

`cbom_ingest_finding_to_cbom_key` is that next foreign key, and this PR adds it.

The **logic is fine** — `deleteWithdrawnRow` was narrowed to `theInventoryRefused(e)` on the constraint name rather than on the exception type, which is exactly why adding an FK is safe here; and `ON DELETE CASCADE` can never refuse a delete, so this constraint cannot reach that catch at all. Nothing to fix in the code.

But the sentence is now false, and it is the kind of sentence a later reader will take as a schema invariant rather than as a statement about one moment. Worth updating it to say what is actually true: the inventory FK is the only `RESTRICT` reference to `cbom(uuid)`, which is the property the carve-out depends on.

src/test/java/com/otilm/core/cbom/ingest/IngestFindingRollupTest.java
```diff
@@ -0,0 +1,160 @@
...
+                        List.of()));
+
+        assertThat(report.rows())
+                .filteredOn(row -> row.kind() == CbomIngestFindingKind.FINDING)
+                .hasSize(IngestFindingRollup.MAX_ROWS_PER_KIND);
+        // 105 distinct dropped-member messages plus the repeated one, against a bound of 100.
+        assertThat(report.dropped()).isEqualTo(6);
+        assertThat(report.rows().get(0).detail())
+                .describedAs("the most frequent message is the one that survives first")
+                .isEqualTo("the same thing happened twice");
```
karolhrdina, line 92: The comparator here is the convergence property the PR rests on, and no assertion pins it.

```java
.sorted(Comparator.comparingInt(Row::occurrences).reversed().thenComparing(Row::detail))
```

As written this is correct and is a *total* order — `detail` is the map key of a per-kind `LinkedHashMap`, so no two rows compare equal, and the surviving 100 are a deterministic function of the input. That is what makes "replaces the previous attempt's rows" converge: a re-ingest writes the identical rows in the identical order.

But `aFloodOfDistinctFindingsIsBoundedWithoutCostingTheSkips` puts 105 messages at `occurrences == 1` against 99 remaining slots, so `thenComparing(Row::detail)` is what decides which 99 survive — and the test asserts only `hasSize(100)`, `dropped == 6`, and `rows().get(0)`. It never asserts *which* of `"member 0 was dropped" … "member 104 was dropped"` made it.

So both of the classic regressions pass this file unchanged:

* dropping `.thenComparing(Row::detail)` — the surviving set becomes dependent on stream order rather than being pinned;
* the mis-chain `comparingInt(Row::occurrences).thenComparing(Row::detail).reversed()`, which reverses *both* keys and silently changes the tie-break direction.

An assertion on the exact surviving set at a tie-heavy boundary would close it. A same-input-twice equality assertion on the whole `Rollup` would be the direct statement of the convergence claim, which the class javadoc at `:15-18` makes and nothing currently tests.

One more gap in the same file, cheap to close: `add` truncates *before* it groups (`:108` then `:110`), so two distinct over-long messages sharing a 507-code-point prefix merge into one row with a summed count. That ordering is what keeps a `uq_cbom_ingest_finding` violation unreachable, it is documented at `:126-128`, and no test pins it — moving the truncation into the writer would turn it into a unique violation with nothing failing.

src/main/java/com/otilm/core/tasks/SystemScheduledJobs.java
```diff
@@ -25,6 +25,7 @@ public Void registerJobs() throws SchedulerException {
         schedulerService.registerScheduledJob(UpdateCertificateStatusTask.class);
         schedulerService.registerScheduledJob(UpdateIntuneRevocationRequestsTask.class);
         schedulerService.registerScheduledJob(CbomSyncTask.class);
+        schedulerService.registerScheduledJob(CbomReconcileTask.class);
```
karolhrdina, line 28: This is the first job added to `SystemScheduledJobs` since deployments started carrying `scheduled_job` rows, which opens a window that has never been exercised.

`registerScheduler` (`SchedulerServiceImpl.java:453-488`) is a read-then-write with no transaction of its own, no upsert, and no cluster lock:

```java
Optional<ScheduledJob> scheduledJob = scheduledJobsRepository.findByJobName(jobName);
if (scheduledJob.isPresent()) { ... return ...; }
ScheduledJob scheduledJobEntity = new ScheduledJob();
...
scheduledJobsRepository.save(scheduledJobEntity);
```

and `scheduled_job.job_name` carries **no unique constraint** — `V202305151330__scheduler.sql` declares it `VARCHAR NOT NULL`, and `V202306121100__scheduler_indexes.sql` adds only the uuid primary key.

Two nodes of a cluster booting together both see "absent" and both insert. The result is two rows with `job_name = 'CbomReconcileTask'`, after which `Optional<ScheduledJob> findByJobName(...)` throws `IncorrectResultSizeDataAccessException` on every trigger — so the job never runs, and enable/disable/delete for it fail too.

The race is pre-existing and applies to all five system jobs, but for the existing four the row was written on a first-ever boot, before there was a cluster to race. This job lands on estates that are already running and will be rolling-restarted, so this is the boot where the window is actually open.

A unique index on `scheduled_job(job_name)` (plus `ON CONFLICT DO NOTHING`, or just catching the violation and re-reading) closes it, and is worth doing here rather than after someone hits it. If that feels out of scope for this PR, it is at least worth a note on #2073 — the failure is silent until an operator wonders why the weekly reconcile has never run.

karolhrdina (commented):
Reviewed with the whole stack in hand (#2256, #2257, #2263 merged). Main and test sources compile clean offline, migration versioning is in order (`V202609141400` is the highest, no collision, no out-of-order need), and most of the description's load-bearing claims survive checking rather than merely reading well:

* `reconcile()` really is `runSync(THE_WHOLE_LISTING)` and really does put `after=0` on the wire — the ternary short-circuits `getLastSyncTimestamp()` entirely, and `reconcile_listsTheWholeListingRatherThanTheWatermarkWindow` stubs only `after=0`, so a watermark run would get no stub at all. Watermark isolation is real: the reader filters on `CbomSyncTask.NAME` by equality, and the two job names differ.
* `AbstractCbomFeedTask` is a faithful extraction — I compared the moved `performJob` line for line against the deleted one. The `getJobClassName()` pull-up is safe for all four unrelated tasks; every prior body was byte-identical to the new default, and CGLIB dispatches on the target, so the stored class name does not move.
* `IngestFindingRollup` is genuinely pure and genuinely deterministic — no `HashMap`, no clock, no logger, and the comparator is a *total* order because `detail` is the map key, so the convergence the "replaces the previous attempt's rows" claim needs actually holds.
* Truncation happens before grouping, so the `uq_cbom_ingest_finding` violation one might expect is not reachable; the btree tuple has ~620 bytes of headroom at the CHECK's worst case, and both text columns are bounded before they reach SQL.
* The conditional orphan delete's SQL is right — correct columns (`al.canonical_key = a.identity_key`, not `absorbed_key`, not the uuid), both arms index-served, both mutation-caught by the new tests. `DocumentScope`'s rule is unchanged; the +23 there is javadoc only. The `UNENCODABLE_REF` stand-in cannot manufacture a false duplicate, because detection runs on the raw refs before substitution.
* `cbom_ingest_finding`'s `ON DELETE CASCADE` composes correctly with `crypto_asset_source`'s `RESTRICT` in both firing orders.

Seven findings below, plus four replies on Copilot's threads — it found the `run.duplicates` conflation, the `walk` blind spot, the unregistered filter field and the truncation unit mismatch, and all four hold up. I have added the evidence and, where it changes the fix, corrected the consequence.

The one I would want resolved before merge:

**`withdrawSupersededVersions` never clears the superseded revision's report.** The description says *"A superseded revision's report goes with the contribution it withdraws"*, and the javadoc at `CbomAssetIngestService:463-466` says the same, but `clearReport` has two call sites and both pass the CBOM being ingested. On the ordinary path — v1 `SYNCED`, v2 arrives, v2's loop detaches v1's links — v1 keeps its report for ever, because a `SYNCED` row is on neither work list and `supersede(v1)` never runs. Once #2267 serves these, that is a report shown against a revision that has been silently emptied. One line fixes it.

Then two about blast radius that are the same family as Copilot's outage finding, and are the reason I would not treat that one as cosmetic: nothing caps how many entries a single reconcile run charges to the retry budget, and `cbom.sync.skipped-retry-runs: 0` — which the validation accepts — makes one bad Sunday write off every entry Core does not hold.

The rest: a deterministic ingest refusal with no terminal state, a migration no test ever executes, a comment this PR falsifies, a first-boot registration race for the new job, and one test-coverage gap on the comparator that is the convergence claim.

Checked and **not** findings, so nobody re-derives them: `int` overflow on `occurrences`; a unique violation from truncation; the report write racing the cluster lock (it is taken later, and the report touches neither `crypto_asset` nor `ALIAS_DECISION_LOCK`); the `SyncRun.summary()` format arity (12 and 7, counted programmatically, both match); sorting or exporting by the new filter field; and the hourly pass's outage arm, which is bit-for-bit the old rule.

src/main/java/com/otilm/core/cbom/asset/identity/CbomAssetExtractor.java
```diff
@@ -237,10 +243,55 @@ public Extraction extract(JsonNode document, Set<String> batchRefutedDigests) {
...
-                skips.add(new Skip(nameOf(component), e.getClass().getSimpleName()));
+                // The name goes through the same encodability rule as everything else headed for storage. It
+                // cannot be refused the way an asset's is: this is the refusal path, and a component whose name has
+                // no UTF-8 encoding is exactly the one whose skip has to be reportable.
+                skips.add(new Skip(encodable(nameOf(component)), e.getClass().getSimpleName()));
             }
         }
-        return new Extraction(List.copyOf(assets), List.copyOf(skips), walk.depthLimitReached(), scopeUnavailable);
+        return new Extraction(List.copyOf(assets), List.copyOf(skips), walk.depthLimitReached(), scopeUnavailable,
+                encodable(scope.ambiguousRefs()));
```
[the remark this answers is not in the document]
karolhrdina: Confirmed — `DocumentScope.walk` starts at the root `components` array and recurses only into each component's own `components`:

```java
pushChildren(document == null ? null : document.get("components"), 1, pending, depths);
while (!pending.isEmpty()) {
    ...
    pushChildren(component.get("components"), depth + 1, pending, depths);
}
```

so `metadata.component` and `services[]` are never visited, and a `bom-ref` repeated between one of those and a `components[]` entry leaves `duplicated` empty.

Worth scoping the consequence precisely, because it is narrower than "the invalid document is ingested unchecked": `componentsByRef` is built from the *same* walk, so a ref defined only in `metadata.component` was never resolvable in the first place, and `DocumentScope`'s resolution behaviour is unchanged by the collision. What is missed is the **refusal**, not the key. The document is keyed exactly as it was before this PR, which is defensible — the component that wins is the `components[]` one, deterministically — while CycloneDX says the document is invalid and the operator is told nothing.

So: a gap in a new validation rather than a key-moving hazard. Still worth closing, and your fix is the right shape — a single ref-collecting traversal over every `bom-ref`-bearing section, with the same depth bound, feeding the same `duplicated` set. Two details for whoever takes it:

* `services[]` nests (`service.services[]`), so it needs the same bounded recursion, and `MAX_DEPTH` should be shared rather than duplicated.
* `dependencies[]` carries `ref`/`dependsOn`, but those are *references* to definitions rather than definitions, so they must not feed the duplicate set — only `metadata.component`, `components[]` and `services[]` define a `bom-ref`.

No test covers the blind spot today, and none covers the plural arm of the refusal message either (`"%d bom-ref values"`), so a fix here has room to add both.

src/main/java/com/otilm/core/cbom/ingest/IngestFindingRollup.java
```diff
@@ -0,0 +1,149 @@
...
+     * whose tail an operator will not see anyway.
+     */
+    private static String bounded(String text, int maxLength) {
+        if (text == null || text.length() <= maxLength) {
+            return text;
+        }
+        final int keep = maxLength - TRUNCATION_MARKER.length();
+        final int end = text.offsetByCodePoints(0, text.codePointCount(0, keep));
+        return text.substring(0, end) + TRUNCATION_MARKER;
+    }
```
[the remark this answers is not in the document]
karolhrdina: Confirming the unit mismatch, with two corrections to the consequence.

`bounded` guards on `text.length()` — UTF-16 units — while the javadoc at `:47-49`, the migration's `CHECK (length("detail") <= 512)` and PostgreSQL's `length()` all speak in code points. The *cut position* is code-point-correct (`codePointCount` + `offsetByCodePoints`), so no surrogate pair is ever split; only the bound is in the wrong unit.

**It cannot fail the insert.** Code points ≤ UTF-16 units always, so anything that passes the Java guard passes the CHECK, and a truncated string is at most 507 + 5 = 512 code points in the all-BMP case. The error is one-directional: over-truncation.

Measured against the actual method:

| input | in (utf16 / cp) | out (utf16 / cp / utf8 bytes) |
|---|---|---|
| `"z".repeat(513)` | 513 / 513 | 512 / 512 / 512 |
| 512 astral code points | 1024 / 512 | **513** / 259 / 1021 |
| 257 astral code points | 514 / 257 | **513** / 259 / 1021 |

Two things fall out:

1. A 512-code-point emoji string is cut to **259** code points, though the column would have taken all 512 — your point, and it is real.
2. The output can be **513** UTF-16 units, one over `MAX_DETAIL_LENGTH`. When char 506 is a high surrogate, `codePointCount(0, 507)` counts the dangling half as a code point and `offsetByCodePoints` then returns 508. Harmless against the database (259 code points), but it means the constant is not a post-condition of the method that owns it.

`theCutFallsOnACodePointBoundary` (`IngestFindingRollupTest.java:113-123`) asserts only well-pairedness and no size at all, which is precisely the shape that lets (2) through — the sibling ASCII test does assert the size. Adding `assertThat(detail).hasSizeLessThanOrEqualTo(IngestFindingRollup.MAX_DETAIL_LENGTH)` there fails today.

Also worth adjusting the javadoc at `:48-49`: "512 code points is at most 2048 bytes of UTF-8" is a true bound on the CHECK, but what the code enforces is tighter — ≤ 1536 bytes, since 512 UTF-16 units is at worst 512 three-byte BMP characters.

src/main/java/com/otilm/core/enums/FilterField.java
```diff
@@ -393,6 +393,10 @@ public enum FilterField {
             CbomAssetSyncState.class),
     CBOM_ASSETS_SYNCED_AT(Resource.CBOM, null, null, Cbom_.assetsSyncedAt, "Assets Synced At",
             SearchFieldTypeEnum.DATETIME),
+    // The reason asset ingest last refused or failed this CBOM. Filterable here is as far as this repository can
+    // carry it on its own: CbomDto has no field for the reason, so serving it in a response is an interfaces release.
+    CBOM_ASSET_SYNC_ERROR(Resource.CBOM, null, null, Cbom_.assetSyncError, "Asset Sync Error",
+            SearchFieldTypeEnum.STRING),
```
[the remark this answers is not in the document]
karolhrdina: Confirmed, and the inconsistency is sharper than "not discoverable": every other member of `SearchHelper.ABSENT_FROM_LISTING` **is** registered in its own resource's hand-written list.

* `CONNECTOR_AUTH_TYPE` → `ConnectorServiceImpl.java:530`
* `SIGNING_RECORD_SIGNED_DOCUMENT_RETRIEVED_AT` → `SigningRecordServiceImpl.java:108`
* `SECRET_SYNC_VAULT_PROFILE` → `SecretServiceImpl.java:307`

`ABSENT_FROM_LISTING` means "filterable but not a column" — it is not an alternative to registering the field, it is the flag that stops the registered field from being offered as a column. Registering in one and not the other leaves the new member in a state no existing member is in.

The rest of the reasoning in the PR does check out, for the record:

* the predicate builder resolves `Cbom_.assetSyncError` and the STRING branch applies, so filtering genuinely works at that layer;
* `isDisplayable` is false (`Resource.CBOM` is in `CONFIGURABLE_COLUMN_RESOURCES`, and the field is in `ABSENT_FROM_LISTING`), so no empty column heading and no saved view can introduce one;
* sorting is unreachable — `isOrderableOnListing` is `isOrderableField && (isDisplayable || !CONFIGURABLE_COLUMN_RESOURCES.contains(root))` = `true && (false || false)` = false, and a client that tries gets a clean `ValidationException` from `SortOrderBuilder`, not an NPE or a silently unordered page;
* CSV export is DTO-driven, so the field cannot reach it.

And the field *is* already reachable for rules and triggers, since `ResourceServiceImpl.listResourceRuleFilterFields` derives from `FilterField.getEnumsForResource` rather than from the hand-written list. So the gap is specifically the CBOM listing's filter picker.

One `SearchHelper.prepareSearch(FilterField.CBOM_ASSET_SYNC_ERROR)` at `CbomServiceImpl.java:707` closes it. Worth a test too — `grep -rn "CBOM_ASSET_SYNC_ERROR\|assetSyncError" src/test` returns nothing, so neither the filter working nor the sort being refused is pinned.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -1131,9 +1170,22 @@ private void settleUnavailable(SyncRun run) throws CbomRepositoryException {
...
+     * listing the repository serves, whatever its document store is doing. The hourly pass keeps the original rule,
+     * which its own overlap would otherwise defeat -- it re-offers stored entries every run by design.
      */
-    private static boolean looksLikeOutage(SyncRun run) {
-        return run.successfulReads == 0 && run.feedDeferred >= 2;
+    private static boolean looksLikeOutage(SyncRun run, SyncScope scope) {
+        if (run.successfulReads > 0 || run.feedDeferred < 2) {
+            return false;
+        }
+        return scope == SyncScope.SINCE_THE_LAST_RUN || run.duplicates == 0;
```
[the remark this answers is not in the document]
karolhrdina: Confirming this, with the exact contamination source.

`run.duplicates` has three increment sites, and only one of them is the evidence this predicate wants:

* `:928` — the feed offered the same identity twice **in this run** (`!run.attempted.add(identity)`). Says nothing about what Core holds.
* `:936` — `existsBySerialNumberAndVersion` matched. This is the signal the predicate means.
* `:965` — `StoreOutcome.DUPLICATE`, the dedup-constraint race.

`:965` is harmless here: it is only reached after `store()` completed a document read, so `run.successfulReads > 0` and `looksLikeOutage` has already short-circuited at `:1185`. So the fix is narrow — `:928` is the only site that needs separating.

The trigger is sharper than "an edge case". Take a genuine total outage of the document endpoint against a Core holding nothing of the listing: every entry defers, `duplicates == 0`, the run is correctly diagnosed as an outage and charges nothing. Now let the listing repeat **one** identity — which `:921-923`'s own comment says happens ("a page overlapping a retried cursor") and which is more likely on a whole-listing pass from `after = 0` than on an hourly window. `duplicates == 1`, `:1188` returns false, and every deferred entry in the estate is charged to the retry budget instead.

So a single repeated listing entry flips the verdict for the entire run, in exactly the direction that costs the most.

Separating the counters is the fix — the predicate wants "how many entries of this listing does Core already hold", which is `:936` alone. Worth keeping the aggregate `duplicates` for the summary line, since that is what it reports.

src/main/java/com/otilm/core/cbom/asset/identity/CbomAssetExtractor.java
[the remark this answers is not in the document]
vyskocilm: Fixed, with the scoping correction from the reply below taken as the specification: what was missing is the **refusal**, not the key.

`DocumentScope.duplicatedRefs` is a second traversal covering `metadata.component`, `components[]` and `services[]`, each bounded by the same `MAX_DEPTH` and each recursing through its own nesting. `dependencies[]` is deliberately not read — `ref`/`dependsOn` cite a definition rather than making one.

**The component index is unchanged.** Feeding the wider set back into `componentsByRef` would move the key of a component whose own ref is unique, on the strength of a repeat in a section nothing the extractor follows resolves into — and resolution is what `CorpusKeySnapshotTest`, `IdentityVectorTest` and `NormalizationRulesTest` pin. So the index keeps the narrow `components[]`-only rule and the refusal takes the wide one. `ambiguousRefs()` is now a superset of the refs `resolve` answers nothing for, and says so.

Seven tests in `DocumentScopeTest` and two in `CbomAssetIngestServiceTest`, including the plural arm of the refusal message you noted was uncovered, and one that asserts the metadata repeat **still resolves to its component** — that one fails if a later change points the index at the wide set. Mutation-checked in both directions: narrowing the refusal back to `walk` fails three of them, widening the index fails a fourth.

src/main/java/com/otilm/core/cbom/ingest/IngestFindingRollup.java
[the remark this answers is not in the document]
vyskocilm: Fixed. `bounded` now measures in code points — `text.codePointCount(0, text.length()) <= maxLength` — so the guard and the cut are in the same unit, which is the unit `length()` and both CHECKs count.

Both consequences the reply below measured are closed by it: a 512-code-point emoji message now survives whole rather than being cut to 259, and the 513-UTF-16-unit output is gone because the dangling-surrogate arithmetic it came from is gone.

`theCutFallsOnACodePointBoundary` gained the size assertion that was missing; `anAstralMessageIsKeptToItsFullAllowanceOfCodePoints` and `aMessageOfExactlyTheBoundIsLeftAlone` are new. Mutation-checked: restoring `text.length()` fails all three.

The `MAX_DETAIL_LENGTH` javadoc keeps "512 code points is at most 2048 bytes of UTF-8" — with the guard in code points that is now the bound the code actually enforces, rather than the looser one the reply correctly said it was — and a paragraph says why the unit matters.

src/main/java/com/otilm/core/enums/FilterField.java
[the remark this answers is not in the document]
vyskocilm: Fixed — `SearchHelper.prepareSearch(FilterField.CBOM_ASSET_SYNC_ERROR)` is in `CbomServiceImpl.getSearchableFieldInformationByGroup`, and the reply below is right that the inconsistency was the point: `ABSENT_FROM_LISTING` is what stops a *registered* field being offered as a column, not a substitute for registering it. `CONNECTOR_AUTH_TYPE`, `SIGNING_RECORD_SIGNED_DOCUMENT_RETRIEVED_AT` and `SECRET_SYNC_VAULT_PROFILE` are all in both, and now so is this.

`testGetSearchableFieldInformationByGroup` asserts the field is advertised and the count is 12, so dropping the registration turns it red.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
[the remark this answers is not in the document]
vyskocilm: Fixed, with the contamination source from the reply below as the specification: `:928` is the only site that needed separating.

`SyncRun` gained `alreadyStored`, incremented only at the `existsBySerialNumberAndVersion` branch, and `looksLikeOutage` reads that instead of the aggregate. `duplicates` is unchanged and still what the summary line reports, since that is the figure it means.

Two tests: `aListingThatRepeatsAnEntryDoesNotMakeAnOutageLookHealthy` in `CbomSyncITest` builds exactly your case — empty Core, one repeated listing entry, two failed reads — and asserts the run still fails as an outage with nothing charged; and `CbomSyncOutageVerdictTest` decides every arm of the verdict from stated counters.

The verdict is now a package-private pure function (`looksLikeOutage(deferred, feedDeferred, successfulReads, alreadyStored, wholeListing)`), which is also how the new charge cap is tested — see the thread on blast radius.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Fixed, and you traced it exactly right: `supersede(v1)` is unreachable for a `SYNCED` row, so nothing cleared it.

`clearReport(superseded)` now runs inside `withdrawSupersededVersions`, after the completeness check — an abandoned withdrawal leaves the report beside the links it still describes, as you suggested.

`theSupersededVersionsReportGoesWithTheLinksItDescribed` pins it on the mainline path, and `aContendedWithdrawalLeavesTheSupersededVersionsReportAlone` pins the ordering. Both mutation-checked: removing the call fails the first, moving it above `if (!withdrawn.complete())` fails the second.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
vyskocilm: Fixed, taking the second of your two guards: **charged more than a run's worth ⇒ treat as an outage after all**.

`MAX_CHARGED_READ_FAILURES_PER_RUN = 1000`, and `looksLikeOutage` returns true when `run.deferred.size()` reaches it, before any other arm. So a run that would charge more than that charges nothing, is skipped, and holds the watermark.

**Why the verdict rather than a cap on the loop**, which was my first instinct too: charging is all-or-nothing by construction. An entry charged nothing has no `cbom_sync_skip` row, so `retrySkipped` never offers it again, and the hourly watermark has moved past it — truncating the charges would lose exactly the entries it declined to charge. Raising the verdict keeps the existing invariant that every deferred entry of a run is charged or none is.

A thousand rather than a fraction of the listing because the budget is spent in entries, not proportion: an estate of a hundred documents losing all hundred is a small write-off and an estate of a million losing ten thousand is not.

**On `cbom.sync.skipped-retry-runs: 0`** — I have left it legal. It is documented on `CbomSyncProperties` as meaning "writes an entry off at its first failure", which is a deliberate value rather than an oversight, and the blast radius that made it terminal is what the cap closes: at `0` one bad Sunday now writes off nothing at all, because the run never reaches `recordFailure`. Rejecting it would be a second fix for a hazard the first one removes.

The two documentation points are in: `SyncScope`'s javadoc now says both openings spend the same `cbom.sync.max-attempts` budget and that an entry can therefore reach `PERMANENTLY_SKIPPED` a run earlier than before, and `CbomReconcileTask`'s comment about occupying the scheduler's single listener is replaced — `messaging.concurrency.scheduler` is 10, and the corrected sentence says that is *why* a reconcile and an hourly sync can overlap, which is what the comment was reaching for.

Tested in `CbomSyncOutageVerdictTest`: the cap arm fires with reads answered and entries recognised — every other arm saying "charge them" — and one below the cap still charges.

src/main/java/com/otilm/core/cbom/ingest/CbomAssetIngestService.java
vyskocilm: Fixed — an attempt bound, scoped to the content-derived refusals exactly as you framed it.

`cbom` gains `asset_sync_content_refusals` (`V202609161200`), incremented **only** by the two refusals that are verdicts on the document: `ambiguousRefs` and `documentScopeUnavailable`. `findAssetIngestRetries` gained `AND c.assetSyncContentRefusals < :maxContentRefusals`, so the backlog stops offering the row at the bound. Every other failure still goes through `markFailed`, leaves the counter alone, and is retried exactly as it always was — which is right, because every one of them can go the other way next time.

**A counter rather than a flag**, because the two are not equally deterministic. A repeated `bom-ref` is a pure function of the bytes and one attempt would do; `documentScopeUnavailable` is raised by an extraction that *threw*, which could have been the moment rather than the document. `MAX_CONTENT_REFUSALS = 3` is the smallest number that gives the second one a second and third chance rather than the smallest number that is correct for the first.

**Not a state, and not a `cbom.sync.*` tunable.** A terminal state is the better answer and I did not take it here: `CbomAssetSyncState` is in `interfaces`, so it is a release — the same reason `markSuperseded`'s javadoc already records. The row stays `FAILED` carrying the reason an operator can read and filter on; what changes is only that the backlog stops paying for it. The producer's fix is a new version, which arrives as its own row and its own ingest.

Three tests: `aRefusalTheDocumentEarnedIsCountedAgainstTheIngestsBound`, `aFailureThatIsNotTheDocumentsFaultIsNotCountedAgainstTheBound`, and `theRetryListStopsOfferingADocumentThatKeepsEarningItsRefusal`, which puts three rows at the bound, one below it and one at zero and asserts which two are offered.

src/main/resources/db/migration/V202609141400__cbom_ingest_finding.sql
vyskocilm: Fixed. `CbomIngestFindingMigrationITest` is in the same shape as the two siblings — scratch schema, `cbom` stub, apply, assert, drop — and asserts the FK's `CASCADE`, the four CHECK names, the unique constraint name, and every column's nullability and type.

It also executes the constraints rather than only naming them: a bad `kind`, a negative `occurrences`, a 513-code-point `detail` and a second row for one `(cbom_uuid, kind, detail)` are each asserted to be refused *by name*, a 512-astral-code-point `detail` is asserted to be accepted (the CHECK counts code points, not bytes, which is the property the writer's bound leans on), and a `DELETE FROM cbom` is asserted to take the findings with it.

That is what turns your hand check into a standing one, since the duplication between the entity and the migration is the design and neither copy was proving the other.

While there: `CryptoAssetInventoryMigrationITest` now also applies `V202609161200__cbom_asset_sync_content_refusals.sql` and pins its column and check — it alters `cbom` in the same chain, so it belongs to that file rather than to one of its own.

src/main/resources/db/migration/V202609141400__cbom_ingest_finding.sql
vyskocilm: Fixed — you are right that the code is fine and the sentence is not.

`deleteWithdrawnRow`'s javadoc now says what the carve-out actually rests on: `crypto_asset_source_to_cbom_key` is the only **RESTRICT** reference to `cbom(uuid)`. It names `cbom_ingest_finding_to_cbom_key` explicitly and says why it does not join — `ON DELETE CASCADE` cannot refuse a delete, so it cannot reach that catch at all — and the closing warning is narrowed to "the next foreign key added to `cbom` **that can refuse a delete**".

src/test/java/com/otilm/core/cbom/ingest/IngestFindingRollupTest.java
vyskocilm: Fixed, and you were right that the assertion I had would not have caught either regression — the sort is stable and the map is a `LinkedHashMap`, so with the tie-break dropped the surviving set was still `member 000 … member 099`. Asserting the exact set against an ascending fixture would have been another instrument that cannot fail.

So the fixture now offers the findings in **descending** order, which puts insertion order and message order in disagreement:

- `whichOfTheTiedMessagesSurviveIsDecidedByTheMessage` asserts the exact surviving list. Dropping `.thenComparing(Row::detail)` makes it the last 100 offered; the mis-chain `comparingInt(...).thenComparing(...).reversed()` reverses it.
- `theOrderTheDocumentOffersItsFindingsInDoesNotMoveTheReport` rolls the same findings up in both orders and asserts the two `Rollup`s are equal. That is the convergence claim stated directly — a same-input-twice equality would have passed under a `HashMap` too, since one JVM iterates the same keys the same way.

Both mutation-checked; dropping the tie-break fails both.

The truncate-before-group ordering is pinned too: `twoOverlongMessagesThatAgreeToTheCutBecomeOneRow` feeds two messages sharing a 512-code-point prefix and asserts one row with `occurrences == 2`. Moving the cut into the writer turns that into the `uq_cbom_ingest_finding` violation you describe, and this now fails first.

src/main/java/com/otilm/core/tasks/SystemScheduledJobs.java
vyskocilm: Fixed — constraint and race handling both, since you are right that this is the boot where the window is actually open.

`V202609161300__scheduled_job_unique_name.sql` adds `uq_scheduled_job_job_name`. The half in front of it is the part that needed care: an estate that already double-registered cannot take the constraint until those rows are reconciled, and `scheduled_job_history`'s FK is `ON DELETE CASCADE`, so deleting a loser outright takes the runs recorded against it. The migration repoints that history at the surviving row (lowest uuid per name) first, then deletes, then adds the constraint.

`registerScheduler` now catches `DataIntegrityViolationException` and re-reads: the row the other node wrote **is** the registration, which is all the caller asked for, so the loser returns the winner's detail rather than failing the boot. A refusal that turns out not to be the race — `findByJobName` still empty — is still an error, with the driver's text logged rather than put in the message that reaches an API response.

`ScheduledJob`'s `@Table` declares the same constraint under the same name, so the entity-generated test schema refuses it too.

Two tests: `ScheduledJobUniqueNameMigrationITest` sets up a job registered three times with history against each, runs the migration, and asserts only the lowest uuid survives, that all three runs are still recorded against it, and that a second row is now refused by name; and `SchedulerJobRegistrationRaceTest` covers the three paths through the catch without a Spring context.

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
vyskocilm: Correcting one thing in my own fix above, found by a second review pass over it.

The cap was tested **before** every other arm, which meant it overrode the one piece of evidence that settles the question. A run in which reads succeeded has watched the repository serve documents — its failures are those documents' own, however many there are, and charging them is exactly what the budget is for. Put first, the cap would call a healthy hourly run carrying a thousand genuinely unreadable documents an outage, skip it, and hold the watermark for ever: the mirror image of the self-skipping the scope-aware arm was added to prevent, and a behaviour change to the hourly pass that nothing here asked for.

It now bounds the whole-listing arm alone, which is the only place it was ever needed:

```java
if (successfulReads > 0 || feedDeferred < 2) {
    return false;
}
if (!wholeListing || alreadyStored == 0) {
    return true;
}
return deferred >= MAX_CHARGED_READ_FAILURES_PER_RUN;
```

Reaching the last line means the listing has proved the repository is serving while no document read succeeded, so the remaining question is only how many entries failed — a few are the stuck documents that pass exists to charge, an estate's worth of them are the document store. That is the shape the guard was always for.

Two tests now state the ordering rather than leaving it to be read off the implementation: `aRunWhoseReadsSucceededIsChargedHoweverManyFailedWithIt` and `anHourlyRunIsDecidedByTheOriginalRuleRatherThanByTheCap`. Restoring the cap-first ordering fails the first.

Fixed in `09f4a47a`.

karolhrdina (dismissed):

karolhrdina (approved):

PR #2271, opened 2026-09-17
Repair the timestamp test script and let it verify ML-DSA tokens

### LINKED ISSUE #2270: The manual TSA test script rejects every default request and reports success anyway
**TL;DR: `httpRequests/issue-timestamp.sh` has been rejecting every default invocation since the signing profiles gained an allow-list, and it exited `0` while doing so. It also cannot verify ML-DSA tokens, which is the algorithm the timestamping stack now provisions alongside RSA.**

## Observed failure

A default run against a locally provisioned stack:

```
$ ./httpRequests/issue-timestamp.sh
...
Status: Rejected
Failure info: unacceptedPolicy
Done.
$ echo $?
0
```

The script requested policy `1.2.3.4.1`, the TSA refused it, and the script reported `Done.` and exited successfully. Anyone using this to check a rebuilt connector would read that as a pass.

## Mechanism

Two independent defects compound.

**The default policy OID went stale.** `POLICY_OID` was hardcoded to `1.2.3.4.1`. Since development-environment#40 the provisioned signing profiles carry a populated `allowedPolicyIds` containing only `1.2.3.4.5.6`, so the TSA answers `unacceptedPolicy` for every default request. The script's own header warns that its defaults track `scripts/timestamping-setup.sh` in development-environment; this is exactly that drift.

**A rejection is not an HTTP error.** RFC 3161 carries the outcome in the response body as a `PKIStatus`, so a refusal arrives as HTTP 200 with a granted-looking envelope. The script checked only the transport, never the status, so the failure was invisible to the exit code — which is why the first defect went unnoticed.

## Scope

Five defects in one 500-line script:

1. `POLICY_OID` default rejected by every provisioned profile.
2. Exits `0` when the TSA rejects the request.
3. `-v` signature diagnostics dead for RSA — "Could not extract signature bytes". The extractor scanned the `TimeStampResp` envelope for "the last OCTET STRING" instead of locating the signature in the token's `SignerInfo`.
4. A misleading diagnostic: the recovered `DigestInfo` prefix was compared against `-d`, the *imprint* digest, when `DigestInfo` names the *signing* digest (SHA-384, from the profile). With the default `-d sha256` this reported MISMATCH on a perfectly valid token.
5. Mode `644` — the only one of 53 scripts in `httpRequests/` that was not executable, so `./issue-timestamp.sh` gave "Permission denied".

Plus a coverage gap: the script cannot verify ML-DSA tokens at all.

## ML-DSA

`openssl ts -verify` cannot verify a post-quantum token. `PKCS7_signatureVerify` goes through `EVP_DigestVerify`, which ML-DSA refuses:

```
provider signature not supported: ML-DSA-65 verify_init
```

So the verification has to be assembled by hand: the chain with `-purpose timestampsign`, the signature over the `signedAttrs` re-tagged from `[0]` to `SET OF` per RFC 5652 5.4 via `openssl pkeyutl -rawin`, the signed `messageDigest` bound to the returned `TSTInfo`, and the `TSTInfo` bound to the request's message imprint, nonce and requested policy — the last of which is what `-queryfile` gives the RSA path for free, and without which a replayed token from the same TSA verifies just as well.

## Suggested fix

Default `POLICY_OID` to empty rather than to `1.2.3.4.5.6`, so each profile answers with its own `defaultPolicyId` and the script works against non-qualified, qualified and both ML-DSA profiles without per-profile knowledge — and does not rot again the next time an allow-list changes. `-p` stays for exercising a specific OID deliberately.

Exit non-zero on rejection, fix the four remaining defects, and add the ML-DSA verification path, keeping the DER work in a separate helper beside the existing `httpRequests/asn1-dump.py`.

ivosh:
Closes #2270.

`httpRequests/issue-timestamp.sh` had been rejecting every default invocation since the signing profiles gained a populated `allowedPolicyIds`, and it exited `0` while doing so — so the breakage hid itself. This repairs it and teaches it the ML-DSA profiles the timestamping stack now provisions alongside RSA.

## The fixes

- **Default `POLICY_OID` is now empty** rather than the stale `1.2.3.4.1`. Requesting no policy makes each profile answer with its own `defaultPolicyId`, so one invocation works against non-qualified, qualified and both ML-DSA profiles without per-profile knowledge, and will not rot the next time an allow-list changes. `-p` still exercises a specific OID deliberately.
- **Exits 1 when the TSA rejects.** A rejection is an HTTP 200 carrying a `PKIStatus`, so the transport check alone could never see it.
- **`-v` diagnostics work again for RSA.** The old extractor scanned the `TimeStampResp` envelope for "the last OCTET STRING"; it now locates the signature structurally in the `SignerInfo`.
- **The `DigestInfo` diagnostic no longer lies.** It compared the recovered prefix against `-d`, the *imprint* digest, when `DigestInfo` names the *signing* digest — reporting MISMATCH on valid tokens under the default `-d sha256`.
- **Mode `644` → `755`**, the only script of 53 in `httpRequests/` that was not executable.

## ML-DSA

`openssl ts -verify` cannot verify a post-quantum token — `PKCS7_signatureVerify` goes through `EVP_DigestVerify`, which ML-DSA refuses with `provider signature not supported: ML-DSA-65 verify_init`. The script now reads the signer's key algorithm from the token and picks the path.

The ML-DSA path verifies four things: the chain with `-purpose timestampsign`, with `-c` as the sole trust anchor — `-no-CApath -no-CAstore` keep the default CA directory and store out, which `openssl ts -verify` on the RSA route also does, while the untrusted chain comes from the token's own `CertificateSet` and from `-t` when given; the signature over the `signedAttrs` re-tagged from `[0]` to `SET OF` per RFC 5652 5.4, via `pkeyutl -rawin`; the signed `messageDigest` against the returned `TSTInfo`; and the `TSTInfo` against the request's message imprint, nonce and requested policy. That last check is what `-queryfile` gives the RSA path for free — without it a replayed token from the same TSA verifies just as well.

Ahead of those, some structural preconditions: exactly one `SignerInfo`, and `eContentType` and the signed `content-type` attribute both `id-ct-TSTInfo`. The digest comes from `SignerInfo.digestAlgorithm` rather than being inferred from the attribute's length, which sha256 and sha3-256 share. These are selected checks, not conformance: the ESS `signingCertificateV2` binding is not validated, so a token whose signing-certificate attribute names a different certificate is not caught. Verifying an ML-DSA token needs OpenSSL 3.5 or newer — `-h` says so, and the script checks before it verifies.

The DER work lives in a new `httpRequests/cms-extract.py`, alongside the existing `asn1-dump.py`. `messageDigest` and `query` each print a verdict line, so a `MISMATCH` is reported rather than confused with a parse error; the structural and algorithm checks have no success line and fail as `MALFORMED <reason>` with exit 1.

## Verification

Against a local stack with all four profiles provisioned:

- 8 positive runs — RSA and ML-DSA × qualified and non-qualified × both routes, `-d sha256|sha384|sha512`, `-n`, `-p`, `-v`.
- Negative: rejected policy, wrong CA and unknown profile each exit 1.
- Replay, the case this PR closes: a token issued for payload A checked against payload B's query gives `query MISMATCH imprint`; the same payload requested twice gives `query MISMATCH nonce` — imprint identical, only the nonce differing. Both pass the `messageDigest` check, which is precisely why it is not sufficient on its own.
- Tamper: flipping a byte in the signature, in `signedAttrs`, or in the `TSTInfo` is detected.
- Malformed input — truncated, non-DER, and a `messageDigest` attribute with no `attrValues`, an empty `attrValues`, or a non-`OCTET STRING` value — reports `MALFORMED <reason>` instead of a traceback. So do a multi-signer token, an `eContentType` or signed `content-type` other than `id-ct-TSTInfo`, an unknown `digestAlgorithm` OID, and a `messageDigest` whose length disagrees with it.

RSA cross-check that the extracted parts are exact: `pkeyutl -verifyrecover` on the emitted `signature.bin` yields a `DigestInfo` whose digest equals SHA-384 of the emitted `signedattrs.der`.


httpRequests/issue-timestamp.sh
```diff
@@ -347,28 +385,59 @@ rm -f "$CURL_STDERR"
...
+
+    if ! openssl ts -reply -in "$RESPONSE_FILE" -token_out -out "$OUTPUT_DIR/token_content.der" 2>/dev/null; then
+        err "Could not extract the token from the TSP response"
+        exit 1
+    fi
+    SIGNER_PEM="$OUTPUT_DIR/signer_cert.pem"
+    if ! openssl pkcs7 -inform DER -in "$OUTPUT_DIR/token_content.der" \
+            -print_certs -out "$SIGNER_PEM" 2>/dev/null || [[ ! -s "$SIGNER_PEM" ]]; then
+        err "Could not extract the signer certificate; was the query sent without -cert?"
+        exit 1
```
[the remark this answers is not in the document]
ivosh: Fixed in cbf7e10e.

Confirmed: these tokens embed two certificates (leaf + `CN=MLDSA` root), and `openssl x509 -in` reads only the first. Reversing the certificate set made verification fail — wrong public key, and the chain check validated the root against itself.

`cms-extract.py` now resolves the certificate from `SignerInfo.sid`, handling both the `issuerAndSerialNumber` and `[0] subjectKeyIdentifier` choices, and writes it as `signer.der`; the shell uses that everywhere it previously took the first `pkcs7 -print_certs` entry, including in `diagnose_signature`. When certificates are embedded and none matches the sid, it reports `MALFORMED no embedded certificate matches the SignerInfo sid` rather than falling back to the first.

Verified by reordering the CertificateSet inside a real token: `pkcs7 -print_certs` then yields `CN=MLDSA` first, while the selected `signer.der` is still the leaf and its key verifies the signature.

httpRequests/issue-timestamp.sh
```diff
@@ -108,6 +114,61 @@ err() {
...
+}
+
+signer_cert_pem() {
+    local outdir="$1" out_pem="$2"
+    [[ -s "$outdir/signer.der" ]] || return 1
+    openssl x509 -inform DER -in "$outdir/signer.der" -out "$out_pem" 2>/dev/null
+}
+
+# openssl cannot verify the timestamp token for ML-DSA so we do it here.
+verify_mldsa_token() {
```
lubomirw, line 146: **[Medium]** | Correctness | Confidence: High

`verify_mldsa_token` checks the chain, the raw signature, `messageDigest` and the request binding, but skips three checks `openssl ts -verify` performs on the RSA route: the ESS `signingCertificateV2` attribute must hash to the signer certificate (`ts_check_signing_certs`), `signerInfos` must hold exactly one entry, and `eContentType` plus the signed `content-type` attribute must be `id-ct-TSTInfo`. The BC `TimeStampTokenGenerator` in the connector always emits ESSCertIDv2, so a defect there is what this script would now miss. A two-signer token built with `openssl cms -sign -signer a -signer b` gets `messageDigest OK` / `query OK` from `cms-extract.py` while `ts -verify` rejects it with `there must be one signer`.

_Suggested:_ in `cms-extract.py`, require `len(children(buf, signer_infos)) == 1`, compare `encapContentInfo.eContentType` to `1.2.840.113549.1.9.16.1.4`, locate the `1.2.840.113549.1.9.16.2.47` attribute (fallback `.12`), hash the resolved signer certificate with its `hashAlgorithm` (SHA-256 when absent) and compare with `certHash`; print `signingCertificate OK|MISMATCH` and route it through the `case` statement in this function, like `messageDigest` and `query`.

<sub>flagged by guided-pr-review, superpowers code review, Copilot CLI; verified against OpenSSL crypto/ts/ts_rsp_verify.c and timestamp-formatting-connector TimestampTokenAssembler</sub>

httpRequests/issue-timestamp.sh
```diff
@@ -108,6 +114,61 @@ err() {
...
+    local outdir="$1" cert_pem="$2" verdicts="$3"
+    local chain_args=(verify -CAfile "$CA_CERT")
```
lubomirw: **[Low]** | Security | Confidence: High

`openssl verify -CAfile` still loads the default CApath and CAstore (`apps/lib/apps.c` `setup_verify`), whereas `openssl ts -verify -CAfile` on the RSA route builds its store from the given file only (`apps/ts.c` `create_cert_store`). A signer that chains to any system-trusted or `SSL_CERT_DIR`-installed root therefore passes the ML-DSA path while the RSA path would reject it. Reproduced on OpenSSL 3.5.7 with the real CA only in `SSL_CERT_DIR` and an unrelated `-CAfile`: `verify` printed OK, `ts -verify` printed FAILED.

_Suggested:_ `local chain_args=(verify -CAfile "$CA_CERT" -no-CApath -no-CAstore)` so `-c` is the sole trust anchor in both branches.

<sub>flagged by superpowers code review; verified against OpenSSL apps/verify.c, apps/lib/apps.c and apps/ts.c</sub>

httpRequests/cms-extract.py
```diff
@@ -0,0 +1,336 @@
...
+            raise Malformed("messageDigest attribute has no attrValues SET")
+        values = children(buf, fields[1])
+        if len(values) != 1 or values[0]["tag"] != OCTET_STRING:
+            raise Malformed("messageDigest attrValues is not a single OCTET STRING")
+        signed = value_of(buf, values[0])
+        break
+    if signed is None:
+        raise Malformed("no messageDigest attribute")
+
+    name = DIGEST_BY_LENGTH.get(len(signed))
```
lubomirw: **[Low]** | Correctness | Confidence: High

`digest_verdict` maps 32/48/64 bytes to sha256/384/512 and never reads `SignerInfo.digestAlgorithm`. That matches the stack today (interfaces `SignatureAlgorithm.ML_DSA_65` commits to `id-sha512`, and every entry pairs a SHA-2 digest), but the same enum already carries SHAKE256-512 for Ed448, and `openssl cms -sign -md sha3-256` yields `messageDigest MISMATCH sha256` while `openssl cms -verify` succeeds. A 28-byte digest raises `MALFORMED`, which `issue-timestamp.sh` turns into exit 1 before `ts -verify` even runs, on the RSA route too.

_Suggested:_ read `children(buf, signer_info)[2]` (the digestAlgorithm SEQUENCE), take its OID content bytes and map them to a hashlib name (`2.16.840.1.101.3.4.2.1/2/3` to sha256/384/512, `.2.8/.2.9/.2.10` to sha3_256/384/512), raise `Malformed` for an unknown OID or when the attribute length disagrees with the size of that digest, and drop `DIGEST_BY_LENGTH`.

<sub>flagged by guided-pr-review, superpowers code review, Copilot CLI; verified against interfaces SignatureAlgorithm.java</sub>

httpRequests/issue-timestamp.sh
```diff
@@ -108,6 +114,61 @@ err() {
...
+
+    [[ -n "$TSA_CERT" ]] && chain_args+=(-untrusted "$TSA_CERT")
```
lubomirw, line 155: **[Low]** | Correctness | Confidence: High

The chain check sees only `-CAfile` and the optional `-t` file, whereas `openssl ts -verify` on the RSA route uses every certificate embedded in the CMS `CertificateSet` as untrusted. The connector embeds `certificateChain.chain()` (`TimestampTokenAssembler.java:180`), so a three-tier ML-DSA PKI (root, intermediate, TSA) fails here with `signer certificate does not chain` on a valid token unless the caller happens to pass the intermediate via `-t`. It works today only because the dev ML-DSA set is two-tier (leaf plus the `CN=MLDSA` root).

_Suggested:_ have `cms-extract.py` also write the non-signer certificates from the `CertificateSet` (one `untrusted.pem`, or per-cert DER files) and add `-untrusted "$outdir/untrusted.pem"` to `chain_args` when the file is non-empty, keeping `-CAfile` as the only trust-anchor source.

<sub>flagged by guided-pr-review, Copilot CLI; verified against timestamp-formatting-connector TimestampTokenAssembler</sub>

httpRequests/cms-extract.py
```diff
@@ -0,0 +1,336 @@
...
+def extract(buf, outdir, query=None):
+    signed_data = signed_data_of(buf)
+    econtent = econtent_of(buf, signed_data)
+    signer_info = signer_info_of(buf, signed_data)
+    signed_attrs = child_with_tag(buf, signer_info, CONTEXT_0, "signedAttrs")
+
+    write_file(f"{outdir}/signedattrs.der", signed_attrs_der(buf, signed_attrs))
+    write_file(f"{outdir}/signature.bin", signature_of(buf, signer_info, signed_attrs))
+
+    signer = signer_certificate(buf, signed_data, signer_info)
```
lubomirw: **[Low]** | Correctness | Confidence: High

`signer.der` is written only when the token embeds a matching certificate and a previous one is never removed, while `signedattrs.der` and `signature.bin` are always overwritten. `OUTPUT_DIR` defaults to the fixed `./tmp/tsa-test` and only `response.tsr` is cleared per run, so a token without certificates makes `signer_cert_pem` pick up the certificate from the previous run: the script then routes RSA/ML-DSA by the stale key and verifies against it instead of reporting the intended `was the query sent without -cert?` error. Reproduced: a certificate-less token run into a directory holding a previous RSA `signer.der` exits 0.

_Suggested:_ unlink an existing `<outdir>/signer.der` here when `signer is None` (ignoring `FileNotFoundError`), or remove `signer.der`, `signedattrs.der` and `signature.bin` in the shell before `extract_cms_parts`, mirroring the `rm -f "$RESPONSE_FILE"` at the request step.

<sub>flagged by guided-pr-review, superpowers code review, Copilot CLI</sub>

httpRequests/issue-timestamp.sh
```diff
@@ -108,6 +114,61 @@ err() {
...
+    chain_args+=(-purpose timestampsign "$cert_pem")
+    if ! openssl "${chain_args[@]}"; then
+        err "signer certificate does not chain to $CA_CERT for timestamping"
+        return 1
+    fi
+
+    openssl x509 -in "$cert_pem" -noout -pubkey > "$outdir/signer_pubkey.pem" 2>/dev/null
+    if ! openssl pkeyutl -verify -pubin -inkey "$outdir/signer_pubkey.pem" \
```
lubomirw: **[Low]** | Error handling | Confidence: Medium

`pkeyutl -verify` stderr is discarded (`>/dev/null 2>&1`) and the preceding `x509 -pubkey` failure is suppressed, so every failure, including `provider signature not supported` on OpenSSL < 3.5, collapses into `ML-DSA signature over signedAttrs is not valid`. On OpenSSL < 3.5 the situation is worse: `x509 -text` prints the ML-DSA key algorithm as a bare OID, `is_pqc_signer` returns false, and the token falls into `openssl ts -verify`, failing with `Verification failed` / `RSA decrypt failed (key may not be RSA...)`, which sends the developer debugging the stack instead of the toolchain. The usage text says both profile kinds are supported without naming the OpenSSL version that makes the ML-DSA path work.

Unverified: the exact `x509 -text` wording for an ML-DSA key on OpenSSL < 3.5.

_Suggested:_ capture `pkeyutl` / `x509` stderr into the `err` message instead of discarding it, state the OpenSSL 3.5+ requirement for ML-DSA next to "RSA and ML-DSA signing profiles are both supported", and optionally guard the ML-DSA branch with an `openssl version` check that names the requirement.

lubomirw (commented):
Automated multi-reviewer triage — 6 findings (COMMENT). Sources: Copilot CLI, guided-pr-review, superpowers code review, pr-hygiene; each finding was re-verified against the diff, OpenSSL source, and the sibling repos it depends on. One Medium (the ML-DSA path skips the ESS, one-signer and content-type checks that `openssl ts -verify` enforces), five Low.

httpRequests/issue-timestamp.sh
ivosh: Fixed in b361a503, minus the ESS check.

`signer_info_of` now requires exactly one entry in `signerInfos`; `econtent_of` compares `encapContentInfo.eContentType` against `1.2.840.113549.1.9.16.1.4`; `require_tstinfo_content_type` requires the signed `content-type` attribute to carry that same OID. The two-signer CMS you describe now gives `MALFORMED there must be one signer, found 2` and exit 1 — before the change it printed `messageDigest OK sha256` and exit 0. An `eContentType` of `id-data` gives `MALFORMED eContentType is not id-ct-TSTInfo`, and a token whose signed `content-type` disagrees gives `MALFORMED signed content-type is not id-ct-TSTInfo`.

They sit in `cms-extract.py`, so they gate the RSA route as well, where `ts -verify` was already enforcing them. The RSA tokens captured from the dev stack still give `messageDigest OK sha384` / `query OK`, and `openssl ts -verify` on them still gives `Verification: OK`.

`signingCertificateV2` is deliberately left out. Premise confirmed: the live ML-DSA-87 token does carry `id-smime-aa-signingCertificateV2`, `hashAlgorithm` sha512, a 64-byte `certHash` — so the check would have something to bind against. But ESSCertIDv2 with the `.12` fallback and the defaulted `hashAlgorithm` is the expensive part of the recipe, and this is a developer tooling script rather than a conformance verifier. Without that binding the script is not an RFC 3161 conformance check, and the module docstring now says so rather than claiming otherwise. Leaving the thread open for you.

httpRequests/issue-timestamp.sh
ivosh: Fixed in b361a503. `chain_args` is now `local chain_args=(verify -CAfile "$CA_CERT" -no-CApath -no-CAstore)`, verbatim as suggested.

Reproduced your result first, on OpenSSL 3.6.3: with the real root only in `SSL_CERT_DIR` and an unrelated `-CAfile`, `openssl verify` printed `leaf.pem: OK`. With the two flags added it prints `error 20 at 1 depth lookup: unable to get local issuer certificate` and fails. `-c` is the sole anchor on both routes now.

httpRequests/cms-extract.py
ivosh: Fixed in b361a503. `DIGEST_BY_LENGTH` is gone. `digest_name_of` reads `children(buf, signer_info)[2]`, takes its OID and maps it through `DIGEST_BY_OID`: `2.16.840.1.101.3.4.2.1/2/3/4` to sha256/384/512/224, `.2.8/.2.9/.2.10` to sha3_256/384/512, plus `1.3.14.3.2.26` for sha1. An unmapped OID raises `MALFORMED unsupported digestAlgorithm`, and an attribute whose length disagrees with `hashlib.new(name).digest_size` raises `MALFORMED messageDigest is 32 bytes, not a sha512 digest`.

A `-md sha3-256` token now prints `messageDigest OK sha3_256`; before the change it printed `messageDigest MISMATCH sha256`. The tokens captured from the dev stack are unaffected: the RSA ones still print `messageDigest OK sha384`, the ML-DSA-65, ML-DSA-87 and ML-DSA-44 ones `messageDigest OK sha512`.

httpRequests/issue-timestamp.sh
ivosh: Fixed in b361a503. `certificates_of` splits the `CertificateSet` and `cms-extract.py` writes the non-signer certificates to `untrusted.pem` (PEM, 64-char lines per RFC 7468); `verify_mldsa_token` appends `-untrusted "$outdir/untrusted.pem"` when that file is non-empty, leaving `-CAfile` as the only trust-anchor source.

Built the three-tier case to confirm it: root → intermediate → leaf, with the intermediate embedded in the token only. `openssl verify -CAfile root -no-CApath -no-CAstore` fails on that leaf; adding `-untrusted untrusted.pem` gives `OK`. Checked on 3.6.3 that multiple `-untrusted` options accumulate rather than overwrite, so a `-t` file still applies alongside it. On the real two-tier ML-DSA token `untrusted.pem` holds the `CN=MLDSA` CA and the chain still verifies.

httpRequests/cms-extract.py
ivosh: Fixed in b361a503. Took the first option: `write_file` now treats `None` as "discard what an earlier run left at this path", guarded by `FileNotFoundError`, and `extract` calls it that way for both `signer.der` and the new `untrusted.pem`.

Reproduced it first: a cert-bearing token followed by a `-nocerts` token into the same directory left `signer.der` holding `CN=TestLeaf` and exited 0. After the change the second run removes it, so `signer_cert_pem` fails and the caller gets the intended `was the query sent without -cert?` error.

httpRequests/issue-timestamp.sh
ivosh: Fixed in b361a503, including the optional guard.

`openssl x509 -noout -pubkey` and `openssl pkeyutl -verify` now run as `2>&1 >file` and `2>&1 >/dev/null`, so their stderr lands in `$failure` and is appended to the `err` line instead of being discarded. The usage text reads "verifying an ML-DSA token needs OpenSSL 3.5 or newer". `openssl_has_mldsa` gates the ML-DSA branch and reports `verifying an ML-DSA token needs OpenSSL 3.5 or newer; this is $(openssl version)`.

The guard runs first in `verify_mldsa_token`, ahead of the chain command — my first cut had it after, which would have let `openssl verify` fail on the ML-DSA certificate itself and report `signer certificate does not chain`, the same misdirection you flagged. Checked with a stubbed 3.4.1: the stub's `verify` arm never runs. Version comparison is numeric on major and minor, so 3.10 is not read as older than 3.5 — checked against 3.4.1, 3.5.0, 3.10.2, 4.0.0 and 2.8.3. macOS's `/usr/bin/openssl` is LibreSSL 3.3.6 and is blocked too, with the message naming it verbatim. Fallthrough happens only when the extracted major or minor is non-numeric, in which case the operation reports its own error rather than this one.

For that guard to fire the token has to reach the branch, so `is_pqc_signer` now also matches `Public Key Algorithm: 2.16.840.1.101.3.4.3.1[789]`.

Leaving this open: that last part is written against your description of how OpenSSL < 3.5 renders an ML-DSA key, which I could not verify — this box is 3.6.3 and prints `Public Key Algorithm: ML-DSA-87`. If an older build prints something other than the dotted OID, the pattern needs another alternative.

ivosh: Verified now — your reading was right, and the caveat above is closed.

OpenSSL 3.0.13 in `ubuntu:24.04`, against the live ML-DSA-87 signer certificate:

```
Public Key Algorithm: 2.16.840.1.101.3.4.3.19
```

The bare dotted OID, exactly as you described. The alternative added to `is_pqc_signer` matches it, so the token stays on the ML-DSA route and `verify_mldsa_token` reports:

```
[ERROR] verifying an ML-DSA token needs OpenSSL 3.5 or newer; this is OpenSSL 3.0.13 30 Jan 2024
```

For contrast, on that same build `openssl verify -CAfile ... -purpose timestampsign` gives `error 20 at 0 depth lookup: unable to get local issuer certificate` — which the script would have surfaced as `signer certificate does not chain`, the misdirection you flagged. That is why the guard runs ahead of the chain command rather than after it.

Not hypothetical for this team: Ubuntu 24.04 LTS and Linux Mint 22.x are both on 3.0.13 and stay there until 2029, since LTS point releases do not move the OpenSSL minor — "keeping up to date" does not clear the floor there. Where it is cleared: Ubuntu 25.10 (3.5.3), Ubuntu 26.04 LTS (3.5.5), Git for Windows 2.53+ (3.5.5, through 3.5.8 in 2.56.0-rc1), Homebrew `openssl@3` (3.6.3). macOS's own `/usr/bin/openssl` is LibreSSL 3.3.6 and is blocked too, correctly — it cannot do ML-DSA either.

lubomirw (approved):

PR #2273, opened 2026-09-17
Stop refusing a branding logo for its aspect ratio

### LINKED ISSUE #2272: Branding: stop refusing a logo for its aspect ratio
### Description

`BrandingLogoValidator` refuses an uploaded logo whose aspect ratio falls outside 1:1 to 3:1 (`MIN_ASPECT_RATIO` / `MAX_ASPECT_RATIO`), for a PNG from its IHDR dimensions and for an SVG from `width`/`height` or the `viewBox`. An SVG from which no ratio can be derived at all is refused as well.

The bound is miscalibrated, and it enforces a design preference rather than a property anything depends on.

**The platform's own logo fails it.** `ot-logo-color.svg` in `fe-administrator` is `viewBox="0 0 405.97 78.9"` — 5.15:1. An operator uploading OmniTrust's own wordmark is told *"has an aspect ratio of 5.15:1; it must be between 1:1 and 3:1"*. Wide wordmarks are the most common logo shape there is, so the rule refuses the ordinary case.

**Nothing downstream needs the bound.** `BrandLogo` renders the logo as `h-9 w-auto max-w-full object-contain`: the height is fixed, the width follows the image, and the width cap scales an over-wide mark down rather than overflowing or distorting it. No surface crops the image, so an unusual ratio cannot cut a logo in half — it can only make it render smaller than the operator hoped, which is theirs to judge and is visible to them in the Appearance tab's own preview.

**It also contradicts how the same feature treats the comparable question.** Brand colours that fail WCAG AA — a real accessibility cost — produce a warning the operator may override, on the stated principle that the brand belongs to the operator and the platform's job is to say what a choice costs rather than to overrule it. A 4:1 logo, which costs nothing, is refused outright.

The rules worth keeping hard are the ones that protect the system: the one mebibyte ceiling, which every anonymous login page fetches, and the format rules, on which the sanitizer and the renderer both depend. Aspect ratio protects neither.

Note: this also touches `fe-administrator`, which mirrors the rule on the client and will replace it with advisory guidance; that half cannot land until this one does, because Core's rejection is what the operator actually hits.

### User Story

As an operator branding my instance, I want to upload my organisation's real logo whatever shape it is, so that the platform carries my brand rather than the nearest shape the platform will accept.

### Use Case

An operator whose logo is a wide wordmark — the common case, and the shape of OmniTrust's own mark — uploads it to **Settings** -> **Platform** -> **Appearance**. Today they are refused and have to pad the image into a squarer canvas before it is accepted. After this change the logo is stored as supplied, and the operator judges from the preview whether it renders as they want.

### Acceptance Criteria

1. A logo is no longer refused for its aspect ratio: `validateAspectRatio` and `validateRatio` are gone, along with the SVG ratio derivation they exist for (`ratioFromWidthAndHeight`, `ratioFromViewBox`, `ratioOf` and the `SVG_LENGTH` pattern).
2. An SVG carrying neither usable `width`/`height` nor a `viewBox` is accepted, since there is no longer a ratio to determine. The rejection naming that case is gone.
3. The PNG chunk walk stays: a logo must still be a well-formed PNG, and a zero width or height is still refused, because both are questions about the file rather than about its shape.
4. The size ceiling and both format rules are unchanged.
5. `BrandingLogoValidatorTest` drops the cases asserting a ratio rejection and gains coverage that a wide wordmark — 5.15:1, the platform's own proportions — and an SVG with no determinable ratio are both stored.
6. The OpenAPI description of the logo fields no longer states an aspect-ratio constraint.

Jakub-Bartal-01:
`BrandingLogoValidator` refused an uploaded branding logo whose aspect ratio fell outside 1:1 to 3:1, for a PNG from its IHDR dimensions and for an SVG from its `width`/`height` or `viewBox`. An SVG from which no ratio could be derived at all was refused as well.

The bound enforced a design preference rather than a property anything depends on, and it refused the ordinary case.

The platform's own logo fails it. `ot-logo-color.svg` is `viewBox="0 0 405.97 78.9"` — 5.15:1 — so an operator uploading OmniTrust's own wordmark was told *"has an aspect ratio of 5.15:1; it must be between 1:1 and 3:1"*. Wide wordmarks are the most common logo shape there is.

Nothing downstream needs the bound either. The administration UI renders the logo as `h-9 w-auto max-w-full object-contain`: the height is fixed, the width follows the image, and the width cap scales an over-wide mark down rather than overflowing or distorting it. No surface crops the image, so an unusual ratio can only make a logo render smaller than the operator hoped — which is theirs to judge, and which the Appearance tab's preview already shows them.

It also sat oddly beside how the same feature treats the comparable question: brand colours that fail WCAG AA produce a warning the operator may override, on the principle that the brand belongs to the operator and the platform's job is to say what a choice costs rather than to overrule it.

## What stays

The rules that protect the system are unchanged: the encoded-length and decoded-size ceilings, both format rules, content sniffing over the declared media type, the PNG chunk walk with per-chunk CRC verification, DTD and external-entity refusal, and `SvgSanitizer`. The zero-width-or-height refusal stays too, moved into `validatePng` — a PNG with a zero dimension is a malformed file rather than an unusual shape.

## Tests

The ratio-rejection cases became acceptance cases rather than being dropped, with the platform's own 5.15:1 proportions as a regression fixture on both the PNG and the SVG path. `aPngWithNoWidthOrHeightIsRejected` is now parameterised over IHDR's width and height offsets, so both halves of the zero-dimension guard are exercised.

The frontend mirrors this rule and is removed in OmniTrustILM/fe-administrator#2158, which merges after this. Merging in this order is the safe one: until it lands the UI keeps refusing a wide logo it would have refused anyway, where the reverse order would let an upload pass the client and fail at the server.

Closes #2272


src/test/java/com/otilm/core/settings/branding/BrandingLogoValidatorTest.java
```diff
@@ -195,36 +191,28 @@ private static void repairChunkCrc(byte[] content, int offset) {
         System.arraycopy(fourBytes((int) crc.getValue()), 0, content, offset + 8 + dataLength, 4);
     }
 
+    /**
+     * Nothing derives proportions from an SVG, so these pin only that it is taken whatever it declares — a size in any
+     * unit, a percentage, a viewBox alone, or no size at all.
+     */
     @ParameterizedTest
     @ValueSource(strings = {
```
ivosh, line 201: **~ Minor** | Testing

BrandingLogoValidatorTest.java:199 covers positive, absent, percentage, mixed-unit, and non-finite SVG dimensions, but never `width="0"` or `height="0"`. Because this PR moved the zero-dimension rejection into `validatePng`, those SVGs are now accepted; without assertions for them, a later shared dimension guard could silently restore the old SVG rejection while this suite remains green.

_Suggested:_ Add separate zero-width and zero-height SVG fixtures to the accepted-size parameterization.

src/test/java/com/otilm/core/settings/branding/BrandingLogoValidatorTest.java
```diff
@@ -195,36 +191,28 @@ private static void repairChunkCrc(byte[] content, int offset) {
...
-
-        Assertions
-                .assertTrue(rejectionMessage(svgLogo(bare)).contains("neither usable width and height nor a viewBox"));
-    }
-
-    /**
-     * Every comparison against {@code NaN} is false, so a viewBox carrying one would pass a range check by failing both
-     * halves of it. {@code Infinity} parses just as readily and divides into a ratio that is not a number either.
-     */
+    /** The viewBox is never read, so a value that is not a finite size reaches the sanitizer like any other. */
```
ivosh: **. Nitpick** | Documentation

BrandingLogoValidatorTest.java:215 says the `viewBox` is never read, but `SvgSanitizer.java:170-173` reads every attribute value and applies CSS-escape and external-URL checks to it. The intended guarantee is narrower: the validator no longer interprets `viewBox` numerically to derive proportions.

_Suggested:_ Drop the comment because it rots easily, or rewrite it to say that `viewBox` is no longer interpreted as dimensions but is still inspected by `SvgSanitizer`.

src/test/java/com/otilm/core/settings/branding/BrandingLogoValidatorTest.java
```diff
@@ -70,31 +70,27 @@ void aClearedSlotIsAccepted() {
         Assertions.assertDoesNotThrow(() -> BrandingLogoValidator.validateAndSanitize(FIELD, null));
     }
 
+    /** 406x79 is the platform's own wordmark, so the list spans the proportions a real brand arrives in. */
     @ParameterizedTest
-    @CsvSource({"200,200", "300,100", "240,120", "1,1", "3,1"})
-    void aPngWithinTheAllowedAspectRatioIsAccepted(int width, int height) {
+    @CsvSource({"200,200", "300,100", "240,120", "1,1", "3,1", "406,79", "1000,50", "100,300"})
+    void aPngOfAnyShapeIsAccepted(int width, int height) {
         Assertions.assertDoesNotThrow(() -> BrandingLogoValidator.validateAndSanitize(FIELD, pngLogo(width, height)));
```
ivosh: **~ Minor** | Documentation

`src/test/java/com/otilm/core/settings/branding/BrandingLogoValidatorTest.java:77` makes the server accept wide and tall PNGs, so the merge-order link in this PR is operationally important. The description points to `OmniTrustILM/fe-administrator#2158`, but that number is the frontend issue and no PR exists at that URL; the open companion implementation is `OmniTrustILM/fe-administrator#2160`, which removes `logoRatioError` and converts the ratio limit to guidance. As written, reviewers cannot follow the advertised dependency or verify the stated merge order.

_Suggested:_ Replace `OmniTrustILM/fe-administrator#2158` with `OmniTrustILM/fe-administrator#2160` in this PR's description.

ivosh (dismissed):

src/test/java/com/otilm/core/settings/branding/BrandingLogoValidatorTest.java
Jakub-Bartal-01: Fixed in 2bcde8d10. `width='0' height='100'` and `width='200' height='0'` are now separate fixtures in the accepted parameterisation, and the doc comment above it says why zero belongs to PNG and not to SVG: for a PNG a zero side means a file no decoder can read, while an SVG that declares one is a well-formed document that simply renders nothing. A later shared dimension guard would now turn the suite red rather than pass silently.

src/test/java/com/otilm/core/settings/branding/BrandingLogoValidatorTest.java
Jakub-Bartal-01: Fixed in 2bcde8d10 — you are right that the attribute is still read. The comment now says no dimension is derived from the viewBox any more, and that `SvgSanitizer` still inspects it like any other attribute. I kept the parameterisation rather than dropping it: nothing interprets those values today, so it is there to fail if viewBox parsing is ever reintroduced.

ivosh (approved):

PR #2275, opened 2026-09-17
Sign and timestamp with the seven new SignatureAlgorithm parameter sets

### LINKED ISSUE #1020: Sign and timestamp with the seven new SignatureAlgorithm parameter sets
Parent: OmniTrustILM/ilm#268. Blocked by the interfaces issue that adds the constants.

**TL;DR: once `SignatureAlgorithm` gains the seven ML-DSA and SLH-DSA SHA2 constants, make Core's signing and timestamping tests prove it signs with every one of them, and fix the two test tables that cannot hold two parameter sets of one family.**

*Rewritten 2026-09-16. This issue was filed 2025-04-09 as "Add PQC algorithms to available signature algorithms" against the pre-rename `CzertainlyProvider` and `CzertainlySignatureService`, with no parent and no acceptance criteria. It is repurposed here as the Core half of that same goal, with the contract half split out to OmniTrustILM/interfaces and both linked under ilm#268.*

## Problem

Core's signing gate needs no code change: `StaticManagedKeySignerCreator` resolves the BouncyCastle parameter-set name and looks it up in the enum, and the signer classes derive everything from the enum's getters. What is missing is proof, and two tests stand in the way of adding it.

`TspControllerITest` and `TspSigningProfileControllerITest` key their per-algorithm tables by `KeyAlgorithm` - an `EnumMap<KeyAlgorithm, ...>` with one `AlgorithmSpec` per family. ML-DSA-44 and ML-DSA-87 both map to `KeyAlgorithm.MLDSA`, so the tables cannot hold them.

## What to change

- **Rekey the two TSP integration tables by `SignatureAlgorithm`** and add rows for ML-DSA-44, ML-DSA-87 and at least the two 128-bit SLH-DSA sets. The larger SLH-DSA sets are slow to sign in BouncyCastle; cover them in the unit tier rather than the integration tier.
- **`CryptographyUtilTest`**: `resolveSignatureAlgorithmName` rows for each new parameter set.
- **`KeyDataValidationTest`**: rows for the new sets.
- **`CryptographyProviderConnectorMock` and `RealSignerTransformer`**: sign with the new sets so the software path is exercised end to end.
- **A resolution test**: a managed key of each new parameter set resolves to its constant in `StaticManagedKeySignerCreator` instead of `MISCONFIGURED`.

## Reach

Core signs through the V1 cryptography client. It reaches `pkcs11-cryptography-provider`, the connector that generates these keys on hardware, only after the V2 operations adapter in ilm#268. This issue proves the software path and the gate; the hardware path follows the adapter.

## Definition of Done

- [ ] Core depends on the interfaces release that carries the seven constants
- [ ] The two TSP integration tables are keyed by `SignatureAlgorithm` and cover ML-DSA-44, ML-DSA-87, SLH-DSA-SHA2-128S and SLH-DSA-SHA2-128F
- [ ] `CryptographyUtilTest`, `KeyDataValidationTest`, the connector mock and the signer transformer cover every new parameter set
- [ ] A managed key of each new parameter set resolves in `StaticManagedKeySignerCreator`; `MISCONFIGURED` is reserved for codes the enum lacks
- [ ] Sonar clean and coverage >= 80 %

ivosh:
Depends-On: OmniTrustILM/interfaces#979

Parent OmniTrustILM/ilm#268, Core half of #1020. Contract half: OmniTrustILM/interfaces#979, which adds the seven constants this branch references — merge and publish that first.

**Core's signing gate needs no code change: `StaticManagedKeySignerCreator` resolves the BouncyCastle parameter-set name and looks it up in the enum, and the signer classes derive everything from the enum's getters. What was missing is proof, and two test tables stood in the way of adding it. This PR is tests only.**

## What changes

**The two TSP integration tables are rekeyed by `SignatureAlgorithm`.** They kept `EnumMap<KeyAlgorithm, ...>` with one `AlgorithmSpec` per family, so ML-DSA-44 and ML-DSA-87 — both `KeyAlgorithm.MLDSA` — could not coexist: the second row silently overwrote the first's certificate and private-key reference. `AlgorithmSpec` now carries the `SignatureAlgorithm` and derives both the display label and the JCA signing name from its code, so the row cannot disagree with itself. Rows added for ML-DSA-44 and ML-DSA-87 (SLH-DSA-SHA2-128F was already there), giving seven algorithms end to end with and without token signature validation.

The slower SLH-DSA sets stay in the unit tier, as the issue directs — signing with them at the integration tier would dominate the suite's runtime without reaching a different path.

**Per-algorithm key material is now provisioned lazily, from a shared fixture.** `setUp` built every row's key pair, connector key and certificate before each of the 19/17 tests, though each test uses exactly one; `TsaSigningMaterial` issues one when a test asks for it. That is what keeps the seven-row matrix from costing more than the old five-row one. The row record, the table, its lookup and that helper now live in `SigningAlgorithmSpecs` and `TsaSigningMaterial`, so both TSP classes share a single definition and each post-quantum parameter spec is derived from the algorithm's own code rather than restated per file.

**`CryptographyProviderConnectorMock.registerSigningKey`** takes a `SignatureAlgorithm` instead of a JCA name string, so the mock signs under exactly the algorithm the platform is expected to resolve.

**`ComputeDtbsEchoTransformer`** echoes the digest the request's signature algorithm commits to, instead of hardcoding SHA-256 as a real connector never would. Without this an ML-DSA run — which commits to SHA-512 — could only have passed by accident.

**Unit tier.** `CryptographyUtilTest` gains a table over all ten post-quantum parameter sets the enum names, asserting the resolved name *is the code of its platform constant* and that `findByCode` returns that constant; the three single-set tests it subsumes are removed. `StaticManagedKeySignerCreatorTest` proves the creator takes the parameter set from the public key item and resolves it through `Signer.getSignatureAlgorithm()`, and that a set the enum lacks — `ML-DSA-99` — is refused as `MISCONFIGURED`. A per-set table here would be tautological: the resolver returns the stored spec name verbatim, and interfaces' own `SignatureAlgorithmTest` already asserts the code round-trip for every constant.

**`ContentSigningEngineITest`** gains ML-DSA-65 and SLH-DSA-SHA2-128F rows — one per implicit digest — proving the new constants survive the content-signing path *and* `SignatureDigestCoherence`, which pins the authorized digest to the algorithm. `SignatureDigestCoherenceTest` needed no edit: its `values()` loop picks the new rows up on its own.

## Reach

Core signs through the V1 cryptography client. It reaches `pkcs11-cryptography-provider`, the connector that generates these keys on hardware, only after the V2 operations adapter in ilm#268. This PR proves the software path and the gate; the hardware path follows the adapter.

## Tests

Spotless and checkstyle clean; the full suite runs in CI. The affected classes: TSP 19 + 17, `ContentSigningEngineITest` 14, `CryptographyUtilTest` 76, `StaticManagedKeySignerCreatorTest` 8, `SignatureDigestCoherenceTest` 8.

Two local failures are environmental and unrelated to this diff: `JwtDecoderITest` (13) hard-binds `new WireMockServer(8082)` and another process held that port, and `ClusterOperationSynchronizerITest` timed out once under full-suite load and passes in isolation.

## Review passes

`/pr-hygiene` and `/review-pr` (7 lenses, branch mode), twice. The findings folded in: the eager `@BeforeEach` provisioning above; the doc-comment rationale repeated across five test files; and one that matters — **the content-signing test derived its expected digest from `SignatureAlgorithm.getDigestAlgorithmIdentifier()`, the same source the mock echo and the production coherence gate read**, so a wrong mapping would have moved all three together and left the test green. `PostQuantumKey` now carries a literal `expectedDigest`. Mutation-checked: flipping ML-DSA-65's digest to SHA-256 in interfaces now fails with `The document digest is SHA-512, but this Signing Profile signs a SHA-256 digest`.

## Merge order

1. interfaces#979 merges and publishes its `-M<build>` snapshot — while it is open, the `Depends-On` line above makes `Build PR` build against the PR snapshot and keeps `Interfaces pin` and `Build` red on purpose;
2. remove that first line, re-run all jobs on both workflows, then merge.

Closes #1020


src/test/java/com/otilm/core/util/CryptographyUtilTest.java
```diff
@@ -369,6 +331,53 @@ void resolveSlhDsaThrowsWhenSpkiAlgorithmOidDoesNotMatch() throws Exception {
...
+
+        // then
+        assertEquals(expected.getCode(), result);
+        assertEquals(expected, SignatureAlgorithm.findByCode(result));
+    }
+
+    static Stream<Arguments> postQuantumParameterSets() {
+        return Stream
+                .of(Arguments
+                        .of(SignatureAlgorithm.ML_DSA_44, KeyAlgorithm.MLDSA, "ML-DSA", MLDSAParameterSpec.ml_dsa_44),
```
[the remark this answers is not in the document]
ivosh: Fixed in 95bf2bef. Correct: `git diff main..HEAD` removed `resolveFalcon1024AlgorithmName` alongside the two ML-DSA single-set tests, but only the ML-DSA ones were subsumed by the table. The surviving `resolvePqcSpecNameFalcon1024` exercises `resolvePqcParameterSpecName`, not the resolver in question.

`postQuantumParameterSets` now carries a provider column and a `FALCON_1024 / KeyAlgorithm.FALCON / "Falcon" / FalconParameterSpec.falcon_1024 / BouncyCastlePQCProvider` row; the other nine keep `BouncyCastleProvider`. The table covers all ten parameter sets the enum names, so its section header is true as written.

Mutation-checked: expecting `ML_DSA_44` on the FALCON row fails with `expected: <ML-DSA-44> but was: <FALCON-1024>`, so the row resolves a real FALCON-1024 SPKI rather than passing vacuously.

FALCON-512 keeps its dedicated test: there is no `SignatureAlgorithm.FALCON_512` constant, so it cannot key a `SignatureAlgorithm`-typed row.

`CryptographyUtilTest`: 76 tests, 0 failures. spotless and checkstyle clean.

src/test/java/com/otilm/core/signing/engine/signer/StaticManagedKeySignerCreatorTest.java
```diff
@@ -141,5 +146,38 @@ void throwsMisconfigured_whenKeyHasNoPublicKeyItem() {
...
+        @ParameterizedTest(name = "{0}")
+        @MethodSource("com.otilm.core.signing.engine.signer.StaticManagedKeySignerCreatorTest#postQuantumParameterSets")
+        void resolvesThePostQuantumParameterSetToItsConstant(SignatureAlgorithm expected, KeyAlgorithm keyAlgorithm)
+                throws SigningEngineException {
+            // given
+            ResolvedStaticKeyManagedSigning scheme = new ResolvedStaticKeyManagedSigning(
+                    SigningCertificateBuilder.valid(),
+                    List
+                            .of(CryptographicKeyItemModelFixtures.activeSigningPrivateKey(keyAlgorithm),
+                                    CryptographicKeyItemModelFixtures.publicKey(keyAlgorithm, expected.getCode())),
```
klaraf755: The fixture is seeded with `expected.getCode()` and `resolveSignatureAlgorithmName` returns `pqcParameterSpecName` verbatim for the PQC key algorithms, so each row reduces to `findByCode(x.getCode()) == x`, which interfaces' `SignatureAlgorithmTest` already covers. One row is enough to prove `create()` reads the public item's spec name. What's missing is the negative case the PR body describes: an unknown name such as `ML-DSA-99` asserting `MISCONFIGURED` and the "which the platform does not support" message.

src/test/java/com/otilm/core/integration/api/tsp/TspControllerITest.java
```diff
@@ -108,39 +108,52 @@ public class TspControllerITest extends BaseSpringBootTest {
...
+                            FalconParameterSpec.falcon_1024, List.of()),
+                    new AlgorithmSpec(SignatureAlgorithm.ML_DSA_44, KeyAlgorithm.MLDSA, MLDSAParameterSpec.ml_dsa_44,
+                            List.of()),
+                    new AlgorithmSpec(SignatureAlgorithm.ML_DSA_65, KeyAlgorithm.MLDSA, MLDSAParameterSpec.ml_dsa_65,
+                            List.of()),
+                    new AlgorithmSpec(SignatureAlgorithm.ML_DSA_87, KeyAlgorithm.MLDSA, MLDSAParameterSpec.ml_dsa_87,
                             List.of()),
-                    new AlgorithmSpec("SLH-DSA-SHA2-128F", KeyAlgorithm.SLHDSA, SLHDSAParameterSpec.slh_dsa_sha2_128f,
-                            "SLH-DSA-SHA2-128F", List.of()));
+                    new AlgorithmSpec(SignatureAlgorithm.SLH_DSA_SHA2_128S, KeyAlgorithm.SLHDSA,
```
klaraf755: The Javadoc above says the slower SLH-DSA sets stay in the unit tier, but 128S is the slow variant: against BC 1.84 it's ~194 ms keygen + ~666 ms sign versus ~2 + ~30 ms for 128F, and 192F/256F are both under 100 ms. With `stubRealSigning()` unconditional and keys regenerated per test, that's ~3.4 s across the two classes for the same `KeyAlgorithm.SLHDSA` path 128F covers. Either drop the row or reword the comment to the rule you actually applied. Same in `TspSigningProfileControllerITest`.

src/test/java/com/otilm/core/integration/api/tsp/TspControllerITest.java
```diff
@@ -166,38 +179,45 @@ public void setUp() throws Exception {
...
-
-            // Uploading the leaf associates it (by public-key fingerprint) with the token-backed key
-            tsaCertificates
-                    .put(spec.keyAlgorithm(),
-                            trustedCa.issueTimestampingCertificate(keyPair, "CN=Test TSA " + spec.label()));
+        trustedCa = testCertificateAuthority.createTrustedCa("CN=TSP Test Root CA");
+    }
+
+    private Certificate tsaCertificateFor(SignatureAlgorithm signatureAlgorithm) throws Exception {
+        Certificate existing = tsaCertificates.get(signatureAlgorithm);
```
klaraf755: This branch can't be taken: per-method lifecycle, DB wiped in `BaseSpringBootTest`, and every test calls the helper once. Return the certificate directly and drop the map (same in `TspSigningProfileControllerITest` and `ContentSigningEngineITest.postQuantumSigningCertificateFor`). `privateKeyReferenceUuids` can stay, the mismatch tests read it.

src/test/java/com/otilm/core/util/mocks/ComputeDtbsEchoTransformer.java
```diff
@@ -55,16 +59,20 @@ public ResponseDefinition transform(ServeEvent serveEvent) {
         }
     }
 
-    private byte[] committedContent(ServeEvent serveEvent) throws Exception {
+    private static DigestAlgorithm committedDigest(JsonNode body) {
+        SignatureAlgorithm signatureAlgorithm = SignatureAlgorithm.findByCode(body.at("/signatureAlgorithm").asText());
```
klaraf755: Nit: `asText()` on a missing node is `""`, so `findByCode` throws and WireMock answers a bare 500. A regression that stops sending `signatureAlgorithm` would now fail inside `sign()` as `CONNECTOR_FAULT`, before the `containsExactly` assertion written for it. `signatureAlgorithmsReceivedBy` already anticipates the empty case; a textual check falling back to SHA-256 keeps that assertion first to fail.

src/test/java/com/otilm/core/integration/signing/contentsigning/ContentSigningEngineITest.java
```diff
@@ -402,6 +458,50 @@ private SigningProfileDto createContentSigningProfile(String name, SignatureFami
         return profile;
     }
 
+    private SigningProfileDto createPostQuantumContentSigningProfile(PostQuantumKey postQuantumKey) throws Exception {
```
klaraf755: Nit: same chain as `createContentSigningProfile` apart from the certificate and `List.of()`. A `(certificateUuid, signingAttributes)` parameter pair with the current 4-arg form delegating would do it. Attributes have to stay explicit, the 1-arg builder's RSA defaults get validated against the key algorithm at creation.

src/test/java/com/otilm/core/integration/api/tsp/TspControllerITest.java
```diff
@@ -108,39 +108,52 @@ public class TspControllerITest extends BaseSpringBootTest {
...
-            String jcaSignatureAlgorithm, List<RequestAttribute> signingAttributes) {
+    private record AlgorithmSpec(SignatureAlgorithm signatureAlgorithm, KeyAlgorithm keyAlgorithm,
+            AlgorithmParameterSpec keyParameterSpec, List<RequestAttribute> signingAttributes) {
     }
 
+    /**
+     * The slower SLH-DSA parameter sets are covered in the unit tier; signing with them here would dominate the suite's
+     * runtime without exercising a different path.
+     */
     private static final List<AlgorithmSpec> ALGORITHM_SPECS = List
```
klaraf755: Nit, pre-existing but this grows it: the record, table, `specFor` and `tsaCertificateFor` are now identical in both TSP tests bar two string prefixes, and the same algorithm-to-spec rows appear in three more files. `MLDSAParameterSpec.fromName` and friends accept the enum codes, so a single fixture could derive the spec from `getCode()`. Fine as a follow-up.

src/test/java/com/otilm/core/util/CryptographyUtilTest.java
```diff
@@ -369,6 +331,57 @@ void resolveSlhDsaThrowsWhenSpkiAlgorithmOidDoesNotMatch() throws Exception {
...
+        // when
+        String result = CryptographyUtil.resolveSignatureAlgorithmName(keyAlgorithm, publicKey, List.of());
+
+        // then
+        assertEquals(expected.getCode(), result);
+        assertEquals(expected, SignatureAlgorithm.findByCode(result));
+    }
+
+    static Stream<Arguments> postQuantumParameterSets() {
+        return Stream
```
klaraf755, line 352: Nit: `generatorName` and `provider` follow from `KeyAlgorithm`: `KeyPairGenerator.getInstance(keyAlgorithm.getCode())` resolves all three under the main BC provider (the TSP tests and `PQCITest` already generate Falcon that way). A `generatePublicKeyBase64(KeyAlgorithm, spec)` overload puts each row on one line.

klaraf755 (changes requested):

src/test/java/com/otilm/core/signing/engine/signer/StaticManagedKeySignerCreatorTest.java
ivosh: Fixed in df5fcbb8. Premise confirmed: `CryptographyUtil.resolveSignatureAlgorithmName` returns `pqcParameterSpecName` verbatim for `FALCON, MLDSA, SLHDSA`, and the fixture seeds `expected.getCode()`, so every row reduced to `findByCode(code) == constant` — which interfaces' `SignatureAlgorithmTest` already asserts for each constant.

The table is gone. `readsThePostQuantumParameterSetFromThePublicKeyItem` keeps a single ML-DSA-65 case, with the private item naming only `KeyAlgorithm.MLDSA` so the set can come from nowhere but the public item. `throwsMisconfigured_whenPostQuantumParameterSetIsNotAPlatformConstant` adds the missing negative case: `ML-DSA-99` asserting `MISCONFIGURED`, the operator message carrying the name and "which the platform does not support", and the client message. `StaticManagedKeySignerCreatorTest` 8 tests.

src/test/java/com/otilm/core/integration/api/tsp/TspControllerITest.java
ivosh: Fixed in df5fcbb8. Premise confirmed, and measured here on BC 1.84: `slh_dsa_sha2_128s` 169.9 ms keygen / 542.3 ms sign, against `128f` 2.0 / 29.4 — while `192f` (4.0 / 46.0) and `256f` (7.7 / 85.4) were excluded. The table carried the slowest set and omitted three faster ones, so the comment described a rule the rows did not follow.

Dropped the row. SLH-DSA is represented by `128f` alone, and the comment now states the rule actually applied: every SLH-DSA set takes the same platform path, so one stands in for the family here and the rest stay in the unit tier. Both TSP classes read the table from the shared `SigningAlgorithmSpecs`, so they cannot drift apart again. TSP 19 + 17.

src/test/java/com/otilm/core/integration/api/tsp/TspControllerITest.java
ivosh: Fixed in df5fcbb8. Premise confirmed: no `@TestInstance(PER_CLASS)` anywhere in the hierarchy, the maps are instance fields, and `tsaCertificateFor` had exactly one caller — `createEnabledProfiles`, which each test invokes once. Same shape for `ContentSigningEngineITest.postQuantumSigningCertificateFor` at its single call site.

Both certificate maps are gone; the helper issues and returns directly. `privateKeyReferenceUuids` stays, behind `TsaSigningMaterial.privateKeyReferenceUuid`, because the mismatch tests read it. Applied in `TspControllerITest`, `TspSigningProfileControllerITest` and `ContentSigningEngineITest`.

src/test/java/com/otilm/core/util/mocks/ComputeDtbsEchoTransformer.java
ivosh: Fixed in df5fcbb8. Premise confirmed, and demonstrated by mutation: with `dtbsRequest.setSignatureAlgorithm` removed from `ManagedContentSigningEngine`, the old transformer failed the run as `SigningEngineException: Internal error during signature formatting` inside `sign()` at line 280 — an error, before any assertion ran. With the textual check the same mutation fails at the `containsExactly` on line 288, which is the assertion written for it.

`committedDigest` now tests `isTextual()` and falls back to SHA-256, so the empty entry `signatureAlgorithmsReceivedBy` documents is what the test actually sees.

src/test/java/com/otilm/core/integration/signing/contentsigning/ContentSigningEngineITest.java
ivosh: Fixed in df5fcbb8, but by different means than you proposed — flagging it rather than resolving, in case you prefer yours.

The parameter pair works, but it makes the attributes nullable and the body forks on `signingAttributes != null` to pick between the one- and two-argument scheme. Instead `aContentSigningProfile(name, family, maxLevel, timestampSource)` returns the `SigningProfileRequestDtoBuilder` with the content-signing half set, and each caller finishes the chain with the scheme its key needs: `.withStaticKeyManagedSigning(signingCertificate.getUuid())` for the classical profile, `.withStaticKeyManagedSigning(certificate.getUuid(), List.of())` for the post-quantum one. `createEnabledSigningProfile` then applies the record policy, creates and enables. One definition of the chain, no nullable and no branch.

Your constraint is kept exactly: attributes stay explicit for the post-quantum profile, because the one-argument scheme defaults them to `PKCS1_v1_5` + SHA-256 RSA attributes that are validated against the key algorithm at creation.

src/test/java/com/otilm/core/integration/api/tsp/TspControllerITest.java
ivosh: Fixed in df5fcbb8 rather than deferred. The record, the table, `specFor` and `tsaCertificateFor` now live in `SigningAlgorithmSpecs` and `TsaSigningMaterial`; both TSP classes share them and the two string prefixes are constructor arguments. `SigningAlgorithmSpecs.parameterSpecFor` derives the spec from the code, as you suggested — `FalconParameterSpec.fromName("FALCON-1024")`, `MLDSAParameterSpec.fromName` and `SLHDSAParameterSpec.fromName` all accept the platform codes verbatim, checked against BC 1.84 for all ten sets. `ContentSigningEngineITest.PostQuantumKey` derives its spec the same way.

src/test/java/com/otilm/core/util/CryptographyUtilTest.java
ivosh: Fixed in df5fcbb8. Premise confirmed: on BC 1.84 here, `KeyPairGenerator.getInstance(code)` resolves `FALCON`, `ML-DSA` and `SLH-DSA` under the main `BC` provider, and a bare-lookup falcon_1024 key carries SPKI OID 1.3.9999.3.14 and still resolves to FALCON-1024. Added the overload as proposed; the `generatorName` and `provider` columns are gone and each row is one line. `CryptographyUtilTest` 76 tests.

klaraf755 (approved):

Issue #1565, opened 2026-06-08
Expose cache settings via Platform Settings API


ivosh:
## Description
Back the new **Caching** tab in **Settings / Platform Settings** (frontend: `OmniTrustILM/fe-administrator#1689`) with platform settings persistence, API support, and runtime application of cache configuration.

The system runs **multiple independently-configured caches**, not a single cache. See `com.czertainly.core.config.cache.CacheConfig` and the `*CacheProperties` records. Today each cache group is configured via immutable Spring `@ConfigurationProperties` records (e.g. `AuthCacheProperties`, prefix `caching.authentication`), bound from config at startup, and the Caffeine caches are built once in the `cacheManager` bean. Each group exposes exactly two parameters — `ttlMinutes` and `maxSize` (both `@Min(1)`).

This issue makes those parameters configurable per cache group at runtime via the Platform Settings API, layered over the existing config-bound defaults.

Cache groups (8 groups covering ~11 named caches):
- **Authentication** (`caching.authentication`) — backs `systemUserAuth`, `userUuidAuth`, `certificateAuth`, `tokenAuth`
- **Certificate chain** — `certificateChain`
- **Connector API client** — `connectorApiClient`
- **Cryptographic key item** — `cryptographicKeyItem`
- **Signing certificate** — `signingCertificate`
- **Signing profile** — `signingProfile`
- **Time quality configuration** — `timeQualityConfiguration`
- **TSP profile** — `tspProfile`

## User Story
As a platform administrator, I want each cache's **time-to-live** and **size limit** to be persisted, served, and applied by the backend so that the per-cache values configured in the UI are stored, validated, and actually take effect.

## Use Case
The frontend loads current per-cache settings via the Platform Settings API to populate the **Caching** tab, and submits updated **time-to-live** and **cache size limit** values for one or more cache groups. The backend validates, persists, and applies them so each affected cache honors the new configuration without redeployment.

## Acceptance Criteria
- [ ] Per-cache settings (**ttlMinutes**, **maxSize**) for each cache group are added to the platform settings model and persisted, layered over the existing `@ConfigurationProperties` startup defaults.
- [ ] Platform Settings REST API exposes endpoints to read and update cache settings keyed by cache group (DTOs, mapping, OpenAPI).
- [ ] Input validation enforces `>= 1` for both fields (consistent with the current `@Min(1)`), with clear error responses.
[...]

PR #2277, opened 2026-09-17
Cache OPA authorization decisions by the caller's effective permissions

### LINKED ISSUE #1658: Cache authorization (OPA) decisions, mirroring the authentication cache
## Description
Authorization decisions in Core are evaluated by calling the external OPA service over HTTP for every secured operation. Each `@ExternalAuthorization`-annotated `*Service` method triggers an `OpaClient` request (`checkResourceAccess` / `checkObjectAccess`), and object-level filtering issues further per-object checks. Unlike authentication — which is already backed by a Caffeine-based caching layer (`AuthenticationCache` / `PlatformAuthenticationCache`, configured in `CacheConfig`) with TTL, indices, and targeted eviction — authorization results are not cached at all, so identical authorization queries repeatedly hit OPA across the network.

Introduce a caching layer for authorization (OPA) decisions modelled on the existing authentication cache, so that repeated authorization checks for the same principal/resource/action are served from an in-memory cache and invalidated correctly when the underlying permissions change.

## User Story
As an operator of the platform, I want authorization (OPA) decisions to be cached the way authentication results already are, so that high-frequency API and protocol traffic does not repeatedly round-trip to OPA, reducing latency and load while still reflecting permission changes promptly.

## Use Case
- A client issues many requests in quick succession (e.g. ACME/SCEP/TSP protocol traffic, bulk listing, or UI polling). Today each secured method re-queries OPA; with caching, the first decision for a given `(principal, resource, action, objectUUIDs, policy)` key is reused for the cache TTL.
- An administrator changes a role's permissions, assigns/removes a role, or disables a user. The authorization cache must be evicted (per-user where possible, cache-wide for role-level changes) so the change takes effect immediately — mirroring how `evictByUserUuid` / `evictAll` work for the authentication cache today.

## Acceptance Criteria
- [ ] A new authorization-decision cache (Caffeine, registered in `CacheConfig` with its own `Auth*CacheProperties`-style TTL and max-size properties) caches OPA resource- and object-access decisions.
- [ ] Cache key captures everything that affects a decision: principal identity, policy name, resource + action, parent resource/action when present, and the requested object UUIDs.
- [ ] Cache integrates at the OPA call boundary (`OpaClient` / the `AbstractExternalAuthorizationManager` path) without changing the `@ExternalAuthorization` API surface for callers.
- [ ] Cache entries are evicted on permission-affecting mutations: role permission changes and role deletion evict broadly; user-scoped changes (role assignment, disable, delete) evict that user's entries — consistent with the authentication cache's eviction strategy and the existing `RoleManagementService` / `UserManagementService` eviction hooks.
- [ ] TTL, max size, and enable/disable are configurable via properties with sensible defaults; cache stats are recorded (`recordStats()`), as for the other caches.
- [ ] Anonymous / non-decidable outcomes are not cached (consistent with `unless = "#result.anonymous"` in the auth cache).
- [ ] Unit and integration tests cover cache hits/misses, key correctness, and eviction on user/role mutations (mirroring `AuthenticationCacheIntegrationTest` and `*CacheEvictionTest`).

klaraf755:
Closes #1658.

## Problem

Every `@ExternalAuthorization`-secured method round-trips to OPA over HTTP. A single `listCertificates` call makes five: two `method`-policy checks (parent, then resource) from `ExternalAuthorizationCore`, and three `objects`-policy filters (parent, resource, group members) from `ObjectFilterAspect`. Authentication has had a Caffeine cache for some time; authorization had none, so ACME/SCEP traffic, bulk listing and UI polling re-asked OPA the same questions continuously.

## Change

`OpaClient` consults a new `AuthorizationCache` before its HTTP call; on a hit the request never leaves the process. Callers of `OpaClient` and the `@ExternalAuthorization` surface are unchanged.

**Key.** SHA-256 over length-prefixed `(policy, principalDigest, canonicalResourceJson, details)`. The principal digest reduces the caller's profile to `{anonymous, permissions}` — the only two things either rego policy reads about a caller: every allow rule resolves through `input.principal.permissions`, and the single rule touching `input.principal.user` compares the username to `anonymousUser`. Two callers with equal effective permissions therefore share one entry, and because the auth service's permission merge is canonical (sorted resources, deduplicated sorted actions), two different role sets that grant the same access share it too. A permission change yields a different key, not a stale entry.

**Behaviour.**
- Two caches under `caching.authorization.*`: `resourceAuthz` (verdicts; keys carry the requested object UUIDs) and `objectAuthz` (object filters; few keys, large values). Kill switch `enabled`; startup warning when the authorization TTL exceeds the authentication TTL, since for anonymous callers it is the only staleness bound.
- Native Caffeine `get(key, fn)`: one loader run per concurrent key; nothing stored when the loader throws or returns `null` (OPA answers `{}` for an unloaded policy path). Returned DTOs are copies — the cached instance never escapes.
- The resource JSON used for the key has sorted properties and sorted `uuids`/`parentUUIDs`, which both policies consume as sets; the OPA request body itself is unchanged. The principal read rejects trailing tokens, and hashing goes through `IdentityDigests.sha256Hex`, so a malformed profile bypasses the cache instead of colliding.
- `RoleManagementServiceImpl` clears both caches at its seven existing `authenticationCache.evictAll()` sites — memory reclamation and operator expectation, not a correctness requirement.

**Deviations from the issue's acceptance criteria**, all deliberate: no per-user eviction (a permissions-derived key already makes user-scoped changes unreachable); anonymous decisions are cached (pure, correctly keyed; the `unless=#result.anonymous` analogue is a failed authentication, which has no authorization counterpart); the key is a digest of the tuple, not the literal tuple.

The five `AUTHZ_*` environment overrides are not wired into the Helm charts, matching the eight existing cache variables, none of which are.

## Tests

- `AuthorizationCacheKeysTest` (13, no Spring context) — equal permissions digest alike across different users and role sets, different permissions do not, anonymous keyed apart even with identical permissions, malformed shapes (no `user`, `permissions` null vs absent, trailing content), canonical JSON is order-insensitive, decision key changes with every component and resists cross-boundary collisions.
- `PlatformAuthorizationCacheTest` (17) — sharing and non-sharing, permuted UUIDs and differently ordered property maps share, absent-vs-empty UUIDs do not, `enabled=false` bypasses, `null` and throwing loaders cache nothing, `null`/malformed/surrogate principals bypass without throwing, a mutated returned copy leaves the next read intact, two concurrent misses run the loader once, `evictAll` clears both caches.
- `OpaClientTest` (9, `MockWebServer` per test) — a repeated check makes one HTTP request; equal-permission callers make one; different permissions make two; the two policies do not share entries.
- `AuthorizationCacheITest` (2) — a repeated decision is served from cache; deleting a role evicts it. Same context signature as `AuthenticationCacheITest`; `ContextSignatureGuardTest.BASELINE` stays 61.
- Existing integration tests mock `OpaClient` and are unaffected. Full `mvn verify`: 8868 tests, 0 failures.
- New-code coverage by SonarCloud's formula, computed locally with the project's exclusions: 93.2% (103/108 lines, 20/24 branches). Uncovered: the startup TTL warning in `CacheConfig` (fires only on a misconfigured TTL), the `url` branch of the canonical resource JSON (no production caller sets `url`), and the null guards on `properties` and on the `allow` list in the copy constructor.


src/main/java/com/otilm/core/config/cache/CacheConfig.java
```diff
@@ -148,4 +175,17 @@ public CacheManager cacheManager(AuthCacheProperties authCacheProperties,
 
         return mgr;
     }
+
+    // For authenticated callers authorization staleness is bounded by authentication staleness, but for
+    // anonymous callers the authorization TTL is the only bound, so it must not exceed the authentication TTL.
+    private static void warnIfAuthorizationTtlExceedsAuthenticationTtl(AuthCacheProperties authCacheProperties,
+            AuthorizationCacheProperties authorizationCacheProperties) {
+        if (authorizationCacheProperties.ttlMinutes() > authCacheProperties.ttlMinutes()) {
```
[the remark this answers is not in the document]
klaraf755: Fixed in 80625f52c. The comparison now runs only when `caching.authorization.enabled` is true — a disabled cache serves nothing, so its TTL bounds nothing and there is nothing to warn about. The predicate is extracted as `CacheConfig.authorizationTtlExceedsAuthenticationTtl` and covered by `CacheConfigTest` for the disabled, longer and equal cases.

src/main/resources/application.yml
```diff
@@ -7,6 +7,12 @@ caching:
   authentication:
     ttl-minutes: ${AUTH_CACHE_TTL_MINUTES:5}   # How long a cached identity is considered valid
     max-size: ${AUTH_CACHE_MAX_SIZE:500}       # Maximum number of cached identities
+  # Authorization cache caches OPA decisions, keyed by the caller's effective permissions rather than their identity
+  authorization:
+    enabled: ${AUTHZ_CACHE_ENABLED:true}                             # Kill switch; when false every check goes to OPA
+    ttl-minutes: ${AUTHZ_CACHE_TTL_MINUTES:5}                        # Must not exceed the authentication cache TTL
+    resource-max-size: ${AUTHZ_RESOURCE_CACHE_MAX_SIZE:10000}        # Verdicts; keys include requested object UUIDs
+    object-max-size: ${AUTHZ_OBJECT_CACHE_MAX_SIZE:1000}             # Object filters; few keys, large values
```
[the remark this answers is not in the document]
klaraf755: Moving these to the Settings API is tracked in #1565, which now lists the authorization cache as the ninth cache group with its `enabled` flag, its two size limits, and the rule that its TTL must not exceed the authentication cache's. Until that lands, this PR keeps it on the same application properties and environment overrides as the other eight caches — none of which are wired into the Helm chart either.

src/main/java/com/otilm/core/service/impl/RoleManagementServiceImpl.java
```diff
@@ -120,6 +128,7 @@ public void deleteRole(String roleUuid) {
         roleManagementApiClient.deleteRole(roleUuid);
         attributeEngine.deleteObjectAttributeContent(Resource.ROLE, UUID.fromString(roleUuid));
         authenticationCache.evictAll();
+        authorizationCache.evictAll();
```
ivosh: **! Major** | Security

RoleManagementServiceImpl.deleteRole at src/main/java/com/otilm/core/service/impl/RoleManagementServiceImpl.java:128-131 commits the role deletion to the auth service, then performs fallible local attribute cleanup before reaching either cache clear. If AttributeEngine.deleteObjectAttributeContent throws, both cached authentication profiles and this PR's authorization decisions retain the deleted role's permissions, so affected callers can remain authorized until the cache TTL expires.

_Suggested:_ Clear both caches in a finally block entered after the remote deletion succeeds so cleanup failures still propagate without preserving stale authorization state.

src/main/java/com/otilm/core/service/impl/RoleManagementServiceImpl.java
```diff
@@ -40,6 +41,7 @@ public class RoleManagementServiceImpl implements RoleManagementExternalService,
     private RoleManagementApiClient roleManagementApiClient;
     private AttributeEngine attributeEngine;
     private AuthenticationCache authenticationCache;
+    private AuthorizationCache authorizationCache;
```
ivosh, line 45: **! Major** | Performance

The class-wide transaction at src/main/java/com/otilm/core/service/impl/RoleManagementServiceImpl.java:38 wraps updateRole's local attribute reads, blocking auth-service WebClient call, and local writes at lines 109-119. A slow or unavailable auth service therefore holds the database transaction and any acquired connection or locks for the full HTTP latency; the neighboring NgCallbackDispatcher establishes the repository's NOT_SUPPORTED orchestration plus short explicit-transaction pattern for this exact boundary.

_Suggested:_ Remove the class-level transaction, run HTTP-bearing orchestration with Propagation.NOT_SUPPORTED, and move the local attribute writes into a separate Writer bean with short REQUIRED transactions.

src/test/java/com/otilm/core/security/authz/opa/PlatformAuthorizationCacheTest.java
```diff
@@ -0,0 +1,295 @@
...
+        empty.setObjectUUIDs(List.of());
+
+        resourceAccess(PrincipalFixtures.operator("1111", "alice"), certificateDetail());
+        resourceAccess(PrincipalFixtures.operator("1111", "alice"), empty);
+
+        assertThat(loads).hasValue(2);
+    }
+
+    @Test
+    void theTwoPolicyCachesAreSeparate() {
```
ivosh, line 177: **~ Minor** | Testing

PlatformAuthorizationCacheTest.java:177 exercises `getOrCheckObjectAccess` only once and checks only the loader count, while the cache-hit mutation test at line 230 covers only `OpaResourceAccessResult`. This PR's tests would therefore remain green if `PlatformAuthorizationCache` returned the cached `OpaObjectAccessResult` itself instead of the copy promised by `AuthorizationCache`, allowing one caller to alter the object filter seen by later callers.

_Suggested:_ Add an object-filter cache-hit test that mutates every field of the first returned DTO and verifies a second lookup retains the original allowed, forbidden, and group-action values.

src/main/java/com/otilm/core/security/authz/opa/AuthorizationCache.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * <p>
+ * Entries are keyed by the caller's effective permissions rather than their identity, so two callers holding different
+ * roles that grant the same access share one entry. The key omits {@code principal.user} and {@code principal.roles}
+ * because the method and object policies read only the effective permissions and the anonymous username; a policy that
+ * starts reading further user fields requires revisiting the key.
+ */
+public interface AuthorizationCache {
+
+    /**
```
ivosh, line 32: **~ Minor** | Documentation

AuthorizationCache.java:20 and AuthorizationCache.java:38 repeat near-identical unlabelled paragraphs that each combine lookup semantics, single-flight behavior, exception/null handling, bypass conditions, and copy ownership. On this public contract, the only per-method distinction—the cache and result kind—is buried, and future semantic changes must be synchronized across two prose copies.

_Suggested:_ Move the shared contract into short interface-level paragraphs with bold lead-ins for loading, bypass, and ownership, then leave each method with one sentence describing its cache and result type.

src/test/java/com/otilm/core/security/authz/opa/OpaClientTest.java
```diff
@@ -150,6 +146,56 @@ void throwsExceptionWhenRequestToOpaFails() {
...
     }
 
+    @Test
+    void repeatedCheckHitsOpaOnce() {
+        setUpSuccessfulResourceAccessResponse();
+
+        opaClient.checkResourceAccess(OpaPolicy.METHOD.policyName, getResource(), getPrincipal(), null);
+        opaClient.checkResourceAccess(OpaPolicy.METHOD.policyName, getResource(), getPrincipal(), null);
+
+        assertEquals(1, opaMock.getRequestCount());
```
ivosh: **. Nitpick** | Testing

OpaClientTest.java:156, along with lines 170, 185, and 196, adds four JUnit `assertEquals` assertions, whereas all four other cache test classes introduced by this PR use AssertJ. The older assertions explain the file's legacy style, but extending it makes the new cache suite internally inconsistent with the neighboring tests.

_Suggested:_ Replace the four new request-count assertions with AssertJ `assertThat(...).isEqualTo(...)` assertions.

src/main/resources/application.yml
```diff
@@ -7,6 +7,12 @@ caching:
   authentication:
     ttl-minutes: ${AUTH_CACHE_TTL_MINUTES:5}   # How long a cached identity is considered valid
     max-size: ${AUTH_CACHE_MAX_SIZE:500}       # Maximum number of cached identities
+  # Authorization cache caches OPA decisions, keyed by the caller's effective permissions rather than their identity
+  authorization:
+    enabled: ${AUTHZ_CACHE_ENABLED:true}                             # Kill switch; when false every check goes to OPA
+    ttl-minutes: ${AUTHZ_CACHE_TTL_MINUTES:5}                        # Must not exceed the authentication cache TTL
```
ivosh, line 13: **~ Minor** | Security

src/main/resources/application.yml:13 says only that the authorization TTL must not exceed the authentication TTL, but AuthorizationCacheKeys.java:89 keys on the policy name rather than a policy version, while CacheConfig.java:123-140 stores each decision in node-local caches. After an OPA policy or anonymous-access rule is tightened, every replica can therefore continue serving a cached allow decision until its own authorization TTL expires; neither the knob nor an operator/security document discloses that revocation window or that the environment-based kill switch requires restarting the replicas.

_Suggested:_ Document in the operator/security guide (`documentation` repository) that the TTL bounds stale OPA and anonymous decisions independently on every replica, and state that disabling the cache requires restarting every replica.

ivosh (changes requested):
## Review summary

2 major · 3 minor · 1 nitpick (6 verified).

**Verified clean:**
- All 7 existing `authenticationCache.evictAll()` sites in `RoleManagementServiceImpl` now have an adjacent `authorizationCache.evictAll()` call. _(Maintainability & tests)_
- The only 2 production call sites into `AuthorizationCache` are the resource and object paths in `OpaClient`; the remaining 4 production references are interface and implementation declarations. _(Maintainability & tests)_
- Org-wide `gh search code` across OmniTrustILM and 3KeyCompany for `getAllowedObjects()` and `getAllow()` found no production in-place mutations among the returned matches. _(Maintainability & tests)_
- All 3 AuthorizationCache methods have production callers: OpaClient calls both decision methods, and RoleManagementServiceImpl calls evictAll at 7 mutation sites. _(Reference integrity)_
- All 7 existing authenticationCache.evictAll() sites in RoleManagementServiceImpl now immediately pair authorizationCache.evictAll(). _(Reference integrity)_
- The complete current method_policy.rego and objects_policy.rego in OmniTrustILM/auth-opa-policies read principal.permissions; only method_policy.rego additionally reads principal.user.username for anonymousUser, matching this PR's principal digest. _(Reference integrity)_
- Scanned added lines across all 17 changed files: no TODO/FIXME/XXX/DEBUG/REMOVE ME markers, raw debug-print calls, or commented-out code blocks. _(Noise & hygiene)_
- Reviewed all 118 added comment, Javadoc, and YAML-comment lines: none cites an internal plan/review/chat path or narrates change history. _(Noise & hygiene)_
- Parsed the entire patch for added blank lines: no run of three or more consecutive added blank lines. _(Noise & hygiene)_
- All five permission and role-membership mutation paths at RoleManagementServiceImpl.java:142-203 clear both authentication and authorization caches immediately after their auth-service writes succeed. _(Transaction & concurrency)_
- Both PlatformAuthorizationCache lookup paths use Caffeine Cache.get(key, loader), and concurrentMissesOnOneKeyRunTheLoaderOnce exercises two concurrent callers against the same key. _(Transaction & concurrency)_
- All 3 production OpaClient call sites under src/main/java use fixed OpaPolicy.METHOD or OpaPolicy.OBJECTS values; no request-derived policy path was found. _(Security & info-leak)_
- Both cache accessors derive keys from the policy, principal permissions and anonymous status, canonical resource JSON, and request details; key-derivation failures bypass caching at PlatformAuthorizationCache.java:68-89. _(Security & info-leak)_
- The changed OpaClient path wraps HTTP and serialization failures in the fixed AccessDeniedException message at OpaClient.java:85 rather than forwarding raw exception text. _(Security & info-leak)_
- The only inline fully qualified type references in the changed production files, RoleManagementServiceImpl.java:92 and :110, disambiguate the core RoleRequestDto from the imported client RoleRequestDto; this PR adds no inline FQN. _(Java engineering & idioms)_
- Both static-only helpers added by this PR, AuthorizationCacheKeys.java:47 and PrincipalFixtures.java:24, prevent instantiation with private constructors. _(Java engineering & idioms)_
- PlatformAuthorizationCache.java:31 uses constructor injection like PlatformAuthenticationCache.java:25, while RoleManagementServiceImpl.java:68 follows that class's existing setter-injection pattern. _(Java engineering & idioms)_

src/main/java/com/otilm/core/service/impl/RoleManagementServiceImpl.java
klaraf755: Fixed in 07f7d7700. `deleteRole` and `updateRole` now evict both caches in a `finally` entered after the auth-service call has returned, so a failing local attribute step still propagates without keeping cached profiles or decisions that grant the role's old permissions. The seven eviction pairs now share one `evictPermissionCaches()` helper.

src/test/java/com/otilm/core/security/authz/opa/PlatformAuthorizationCacheTest.java
klaraf755: Added in 07f7d7700: `theCallerCannotMutateTheCachedObjectFilter` mutates all three fields of the first returned `OpaObjectAccessResult` and checks that a second lookup still carries the original allowed, forbidden and group-action values with a single load.

src/main/java/com/otilm/core/security/authz/opa/AuthorizationCache.java
klaraf755: Restructured in 07f7d7700: the shared contract now sits on the interface as Keying, Loading, Bypass and Ownership paragraphs, and each method keeps one sentence naming its cache and result type plus its tags.

src/test/java/com/otilm/core/security/authz/opa/OpaClientTest.java
klaraf755: Done in 07f7d7700 — the four request-count assertions now read `assertThat(opaMock.getRequestCount()).isEqualTo(n)`.

src/main/resources/application.yml
klaraf755: Both points are now in the documentation issue for this cache, OmniTrustILM/documentation#405: that the time-to-live bounds stale OPA and anonymous decisions independently on every replica, and that switching the cache off is a deploy-time setting that takes effect on restart until the configuration moves to Platform Settings (OmniTrustILM/core#1565).

src/main/java/com/otilm/core/service/impl/RoleManagementServiceImpl.java
klaraf755: Done in 1f37597e2. `RoleManagementServiceImpl` now runs with `@Transactional(propagation = NOT_SUPPORTED)` at class level, so no database transaction spans the `roleManagementApiClient` calls. A separate `Writer` bean turned out to be unnecessary: the only local writes are the custom-attribute calls, and `AttributeEngine` is itself class-level `@Transactional`, so each of them already runs in its own short `REQUIRED` transaction once the ambient one is gone. `TransactionalBoundaryArchTest` (Rule A) confirms no repository write is reachable from the `NOT_SUPPORTED` methods.

ivosh (approved):

PR #2279, opened 2026-09-17
Document the coupled interfaces and core pull request flow

### LINKED ISSUE #2192: Document the coupled interfaces and core PR flow

lubomirw:
## TLDR

The coupled-PR mechanism has been live since #2236 merged, three people are using it, and the only thing to read is a paragraph in the README. This adds the runbook and trims that paragraph to the syntax plus a pointer.

Closes #2192

## Why now

Two people learned the mechanism's edges by failing rather than by reading — one lost an afternoon to a fork pull request that silently published nothing, another to the label step. The next thing each of them hits is the one this page exists for: **"Re-run all jobs", never "Re-run failed jobs"**, on **both** workflows. `resolve-interfaces` *succeeds* when an override is active, so a failed-jobs re-run replays its cached answer and the pin stays red however many times it is pressed.

## What is in it

The ordered steps; what a coupled pull request actually shows while it waits; the `Interfaces-Version` escape hatch for an unrelated pull request reddened by somebody else's merge; how to read and trace an `-M` coordinate; the fork limitation; and what reviewing one means.

Two local-build traps are written down because both have already been hit:

- A locally installed `interfaces` branch is silently replaced in the shared `~/.m2` by the next upstream snapshot. The symptom does not point at the cause — one missing constant inside an annotation is an unrecoverable javac error, so annotation processing is skipped entirely and the build reports dozens of missing Lombok getters and JPA metamodel classes in files nobody touched.
- The PR coordinate is mutable, so a local build against it needs `-U` while that branch is moving. The page says explicitly not to carry the `-nsu` advice over from the other route.

## Accuracy

Every behavioural claim was checked against the workflows rather than written from the design document, and the check table is transcribed from a live coupled pull request (#2275) rather than described.

That mattered: the table was rewritten after #2265 added the `Interfaces dependency` gate to `test_docker_image.yaml` the same morning — the image test now waits rather than failing. The page also notes what that costs, which is easy to miss: while it waits, the required `test / Build amd64` and `test / Build arm64` contexts are not reported at all, so the merge box shows them waiting for status. That is a second, independent reason the pull request cannot merge, and it clears itself after step 5.

Three claims inherited from the design document were measured false and are not repeated: that Sonar produces no analysis on a coupled pull request (it does, since #2236); that a release-prep commit's `-M` hole comes from the publish job aborting (the pin job recognises a non-SNAPSHOT pom and deliberately mints nothing); and that `main~k` resolves an ordinal (`interfaces` main carries 68 merge commits against 341 first-parent, so it works near the tip and fails silently further back — which is why the page says not to).

## One honest gap

`Depends-On: interfaces#940` — the repository name shortened — is ignored **in silence**, while every other malformed marker fails loudly. That is the cost of scoping the probe to `OmniTrustILM/interfaces` so a marker for another repository is left alone; one probe cannot do both. The README now states the full repository name as a third rule and says plainly which failures are loud and which is not.

## Testing

Documentation only; Sonar quality gate and >=80% coverage not applicable.

Reviewed by four independent passes, and the findings that mattered all came from reviewers that went and read the workflows. Worth a read from someone who has **not** used the flow — @karolhrdina or @LukasNajman are both using the marker now and neither built it, so either would be a real test of whether it is followable without the design document.


docs/coupled-interfaces-changes.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+Copy the exact coordinate from the publish job summary rather than typing it —
+the prefix comes from the `interfaces` pom, not from `core`'s, and the two
+diverge across a release cut.
+
+Maven and IntelliJ both honour it. The coordinate is mutable: every push to the
+interfaces pull request replaces it, and Maven's default snapshot policy is
+daily, so build with `-U` while that branch is moving. Delete the file when the
+work lands.
```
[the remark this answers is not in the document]
lubomirw: All three applied in 06babcafc, including both suppressed ones — which were the better findings here, so thank you for putting them in the review body rather than dropping them.

**The pin across a release-cut hole.** Correct, and a self-inflicted one: an earlier round flagged "from that merge's publish summary" as ambiguous, and making it specific introduced this. The page explains two paragraphs further down that a release-prep commit mints no pin, so naming the immediately preceding merge contradicts its own text. It now says to take the newest pin that *exists* and walk back past any hole, keeping "usually the merge immediately before it" as the common case.

**The cleared gate.** Also correct and sharper than what I wrote. I had it as "evidence the run reached its pin job, not proof that mainline carries the change" — but both resolvers stop waiting after fifteen minutes and proceed with no pin at all, so past that point it is not evidence of that either. Reworded, and it now says to check the `Publish package` run rather than infer from the gate.

**The mutable coordinate.** Right — `stage` runs `mvn -B verify` before anything is staged, so only a push that builds replaces the coordinate and a failing one leaves the previous snapshot in place. Now says "every push whose build passes".

One note on the visible comment versus the suppressed ones: the two suppressed findings were Medium and both landed; the visible one was Low. Worth knowing that the suppression heuristic inverted the value ordering on this review.

docs/coupled-interfaces-changes.md
```diff
@@ -0,0 +1,207 @@
...
+they are green regardless, so the gate does the enforcing rather than the
+reviewer.
+
+What the approver should know is the tail: after `interfaces` merges, somebody
+has to re-run all jobs on both workflows, and a **push** at that point dismisses
+the approval and cannot be self-approved. Re-running does not.
+
+One thing that is easy to miss while the review is open: a push to the
+`interfaces` pull request republishes the same coordinate but re-runs nothing
+here. Re-run all jobs after each such push, or the green checks are evidence
```
ivosh: **~ Minor** | Correctness

docs/coupled-interfaces-changes.md:205 says to re-run the core workflows after each interfaces push, but `publish-pr-snapshot.yml` does not replace the mutable coordinate until its Verify and publish jobs finish. Until then `build_pr.yml` sees the already-existing coordinate and can produce green checks against the previous interfaces head, which is exactly the stale evidence this paragraph intends to prevent.

Severity: Medium | Category: Correctness | Fix: addition | Confidence: high | Verdict: fix now

_Suggested:_ Require the new interfaces head's `Publish PR snapshot` run to complete successfully before re-running all jobs on both core workflows.

docs/coupled-interfaces-changes.md
```diff
@@ -0,0 +1,207 @@
...
+   coordinate being refreshed; it does not delete what was already published.
+2. **Take the line from its job summary** and paste it into the `core` pull
+   request body, at column 0:
+
+   ```
+   Depends-On: OmniTrustILM/interfaces#940
+   ```
+
+   It must be a bare line in the body — **not inside a code fence or an HTML
+   comment**. Those are stripped before matching, and this is the one mistake CI
```
ivosh: **~ Minor** | Documentation

This PR's `docs/coupled-interfaces-changes.md:34` calls a fenced or commented marker “the one mistake CI does not report,” but `README.md:181-182` documents another silent case: shortening the repository name. The workflow confirms both behaviours: `.github/workflows/build_pr.yml:59` strips fences and comments, while the malformed-marker probe at `.github/workflows/build_pr.yml:80-83` only recognizes the full `OmniTrustILM/interfaces` name. A reader following only the runbook can therefore overlook the explicitly known shorthand failure.

_Suggested:_ Change “the one mistake” to “one silent mistake” and identify the shortened repository name as the other silent case.

<sub>flagged by Doc-set & publication coherence + Maintainability & tests</sub>

docs/coupled-interfaces-changes.md
```diff
@@ -0,0 +1,207 @@
...
+   types, looking like an ordinary compile error.
+
+   Both are now reviewed in parallel, against code that has actually run.
+3. **Expect the checks below to stay non-green for the whole review.**
+4. **Merge `interfaces`.** Watch the `Publish package` run on its `main` for
+   that merge commit — specifically the `Publish immutable build version` job,
+   which mints the `-M<count>-<sha8>` coordinate both `core` gates wait for.
+   Watch the run rather than the clock.
+
+   That job mints the pin even when the SNAPSHOT deploy in the same run failed,
```
ivosh: **~ Minor** | Documentation

At `docs/coupled-interfaces-changes.md:45-49`, this PR compresses the pin job's failed-deploy behavior, what a cleared gate proves, the 15-minute no-pin fallback, and the required pre-merge check into one 60-word sentence. The safety-critical fact that the gate can clear without publication is buried mid-sentence, making this load-bearing step difficult to scan.

Severity: Medium | Category: Documentation | Fix: correction | Confidence: high | Verdict: fix now

_Suggested:_ Rewrite this paragraph as a one-line warning followed by short bold-led paragraphs for gate evidence, the timeout, and the required pre-merge check.

ivosh (dismissed):

docs/coupled-interfaces-changes.md
lubomirw: Applied in 09f665c73, and you are right about more than the omission: the advice as written *caused* the problem the paragraph exists to prevent. Re-run immediately after the push and core resolves the previous coordinate, so the reader gets stale evidence with a fresh green tick on it — worse than no advice.

It now says to wait for that pull request's `Publish PR snapshot` run to finish first, and why: the coordinate is replaced only when the publish job completes, and `stage` runs a full `verify` before `preflight` and `publish` get anywhere near it.

This was the third timing-between-repos correction on this page, so I audited the other cross-repo handoffs rather than wait for a fourth:

| Handoff | State |
| --- | --- |
| Label to first publish (step 1) | covered — the snapshot appears only after that PR's own verify passes |
| Merge to mainline publish (step 4) | covered — names `Publish package`, the pin job, and the fifteen-minute escape |
| Tail re-run of both workflows (step 5) | no window — both read live state and are independent of each other |
| Push to republish mid-review | this one |

So one gap rather than a class of them. The pattern in all three corrections was the same: prose collapsing a multi-minute handoff into an instant.

ivosh (approved):

PR #2284, opened 2026-09-18
List, retry and expire the documents the CBOM sync gave up on

### LINKED ISSUE #2250: Retention and operator view for the CBOM documents the sync could not store
### Description

OmniTrustILM/core#2245 (core#2208) records every feed entry the CBOM header sync could not turn into a `cbom` row in the new `cbom_sync_skip` table — the document read failed (404, a repository 4xx/5xx, a 503 or timeout the run cannot attribute to a repository outage, an unreadable or oversized body) or storing the header failed (Core validation, database refusal). The row carries the identity (`serial_number`, `version`), the attempt count, the last operator-safe reason and the header counts the feed reported. The next runs retry it; after `cbom.sync.skipped-retry-runs` further failures it is written off as `PERMANENTLY_SKIPPED`. Asset processing is not involved — that lifecycle has its own `asset_sync_state` / `asset_sync_error` on the `cbom` row (core#2073).

Michal's review of the PR ([thread](https://github.com/OmniTrustILM/core/pull/2245#discussion_r3980337971)) left two things for a follow-up; the PR shipped with the entity javadoc and the migration header saying so:

- **Retention.** A `PERMANENTLY_SKIPPED` row is never removed. A run no longer loads it (only `RETRYING` rows are read), and it is deleted when its document finally stores, but a document that never stores leaves a row forever, so the table grows with the repository's history of broken documents.
- **Visibility.** The row exists to tell an operator which document was given up on and why, but nothing reads it: no endpoint, no DTO, no UI. The only signals are the counters in the job result and the write-off warning in the Core log, neither of which names the document to the operator.

### What this story delivers

- An operator sees, in the UI, the documents the sync could not store — identity, state (`RETRYING` / `PERMANENTLY_SKIPPED`), attempts, first and last attempt, reason — and can request another attempt for a written-off document (reset to `RETRYING`; the next run picks it up).
- `PERMANENTLY_SKIPPED` rows expire: a retention period after which a scheduled sweep removes them in batches, so the table is bounded by the operator's window, not by history.
- **Everything is configurable from the Settings UI** — the retention period, next to the sync tunables core#2249 moves there. Nothing new in `application.yml` or as an environment variable.

### Decide first: where the operator sees it

Two shapes; pick during refinement, the delivery above holds for both.

1. A dedicated view of `cbom_sync_skip`: a list endpoint with the platform's paging and filtering, a table under the CBOM inventory, a retry action. The smallest change, but a second list the operator has to know about.
2. Fold failed documents into the CBOM inventory itself: a `cbom` row with a sync state and reason, the way asset onboarding already shows `asset_sync_state` / `asset_sync_error`. One list, retry from the document's detail. Needs the header row to exist before the document was ever read — `spec_version` is `NOT NULL` and derived from the body today — so a migration and the FE state column come with it.

### Acceptance criteria

- The operator can list retrying and written-off entries with their reason from the UI, and a requested retry is attempted on the next sync run.
- `PERMANENTLY_SKIPPED` rows older than the configured retention are removed by a scheduled sweep; the sweep is batched and holds no long transaction (`*Writer` bean with a `@Modifying` query, as the repository's transactional rules require).
- The retention period is a platform setting with a default (proposed: 90 days), editable in the Settings UI, effective on the next sweep.
- Reads of the skip records go through the CBOM inventory's permissions; the reason column stays operator-safe (no raw exception text).
- The `CbomSyncSkip` entity javadoc and the `cbom_sync_skip` migration header describe the shipped behaviour instead of pointing at follow-up work.

### Touch points

- `core`: list and retry endpoints (or the `cbom` row state, per the decision), `CbomSyncSkipRepository` (retention delete, reset), `CbomSyncSkipWriter`, a sweep task next to `CbomSyncTask`, the setting read through `SettingsCache`.
- `interfaces`: DTOs for the list and, for shape 2, the sync state on the CBOM DTO; the retention setting in `PlatformSettingsDto`.
- `fe-administrator`: the table or the state column, the retry action, the setting input.

Related: core#2249 (sync tunables in the Settings UI), core#2073 (asset onboarding lifecycle on the `cbom` row).

### Shape ruling 2026-09-15 (Karol)

**Shape 1 — a dedicated view.** Shape 2 would add a placeholder `cbom` row (a migration: `spec_version` is NOT NULL and body-derived, and `CbomServiceImpl.createCbomEntry` rejects an empty one) and a new sync-state value into the `interfaces` enum while core#2073's remaining PRs (#2257 supersede/orphans, the duplicate-`bom-ref` refusal with its findings channel) are re-cutting exactly that row and that enum. Shape 1 touches none of it. Michal is informed on core#2073.

Delivered as:

- **PR 0, independent, first:** expose #2256's stored `asset_sync_error` — `CbomDto.assetSyncError` (`NOT_REQUIRED`, absent when null), `Cbom.mapToDto`, a row on the fe CBOM detail page. Today the column is written but nothing serves or renders it.
- **interfaces:** `CbomSyncSkipDto` (identity, state `RETRYING | PERMANENTLY_SKIPPED`, attempts, first/last attempt, operator-safe reason, header counts), `POST /v1/cboms/syncSkips` (list, platform paging/filtering), `GET /v1/cboms/syncSkips/search`, `POST /v1/cboms/syncSkips/{uuid}/retry`, and `UtilsSettingsDto.cbomSyncSkipRetentionDays` (`@Min(1)`, default 90) next to the sync tunables core#2249 adds.
- **core:** index on `cbom_sync_skip (state, first_skipped_at)`; `CbomSyncSkipRepository` promoted to `SecurityFilterRepository` with `@Modifying` batch delete and reset; `CbomSyncSkipWriter` stays `@Service`/`REQUIRED` (Rule D) with `deleteExpiredBatch` and `requestRetry`; `CbomSyncSkipRetentionSweeper` on the `SigningRecordRetentionSweeper` pattern (cluster lock, per-batch commits, `maxBatchesPerSweep` cap); `CbomSyncSkipRetentionTask` (`performJob` `NOT_SUPPORTED`, daily, system job) registered in `SystemScheduledJobs`; `FilterField` entries; service methods gated `@ExternalAuthorization(CBOM, LIST / UPDATE)`; controller with `@AuditLogged`; retention read from `SettingsCache` per sweep.
- **fe-administrator:** a `PagedList` page under the CBOM inventory with the retry action, plus the retention-days input in the utils settings form.

Order: after core#2249 (the retention setting lands in the settings block it creates) and after core#2251 (so the new sweep task does not inherit the 120 s ambient-transaction bug). Estimate revised to 3.25 MD (agent basis; 3 for shape 1 plus 0.25 for PR 0).

karolhrdina:
Lane B of OmniTrustILM/ilm#299, story #2250 (shape 1 as ruled on 2026-09-15, deviations below). The header sync has written the documents it cannot store to `cbom_sync_skip` since #2245, but nothing read the table, nothing pruned it, and the asset sync reason #2256 stores never reached an operator.

## What changes

| Endpoint | Permission | Behaviour |
|---|---|---|
| `POST /v1/cboms/syncSkips` | CBOM `LIST` | Pages `cbom_sync_skip` with the platform `SearchRequestDto`; newest failure first (`lastAttemptAt` descending, `uuid` ascending); filters on state (equals, not equals) and serial number (equals, not equals, contains, not contains, starts with); ordering by last attempt, first failure, attempts or serial number. An identifier or condition outside that surface is refused as 422 with a message naming both it and what would work. `columns` is accepted and ignored. |
| `GET /v1/cboms/syncSkips/search` | CBOM `LIST` | The two filter fields, served by `CbomSyncSkipSearch`; state is a `LIST` field over the new `CbomSyncSkipState` platform enum. |
| `POST /v1/cboms/syncSkips/{uuid}/retry` | CBOM `UPDATE` | A permanently skipped document goes back to `retrying` with `attempts` 0, keeping the reason and both timestamps of its last attempt until the next run. A row already retrying is answered unchanged. Unknown uuid is 404. Audited as `Operation.RETRY`. |

**Retention.** `CbomSyncSkipRetentionTask` (system job, daily at 03:15) removes permanently skipped rows whose `last_attempt_at` is older than the new **`cbomSyncSkipRetentionDays`** platform utils setting (default 90, bounds 1 to 3650; a stored value outside them reads as the default with the same once-per-name warning as the other three tunables). `CbomSyncSkipRetentionSweeper` takes the cluster lock (`CBOM_SYNC_SKIP_RETENTION`), deletes in batches of 500 through a `REQUIRES_NEW` writer, and stops after 20 batches per run, saying so in the history entry. Rows still retrying are kept whatever their age. New index `idx_cbom_sync_skip_state_last_attempt (state, last_attempt_at)` in `V202609171500`.

**`assetSyncError`** is mapped onto `CbomDto` and the detail, and leaves `SearchHelper.ABSENT_FROM_LISTING`, so the reason the asset ingest refused a document can be shown as a column and ordered on.

The core `model.cbom.CbomSyncSkipState` enum is replaced by the interfaces one, which is what the listing serialises.

## Deviations from the 2026-09-15 ruling (decided at pickup, after recon)

- **Listing mechanics:** the endpoint takes `SearchRequestDto`, but core serves the searchable fields by hand and builds its own `Specification` (`cbom.sync.CbomSyncSkipSearch`) instead of adding `FilterField` entries. `FilterField` roots at `Resource.CBOM` would collide with the CBOM listing's own sort and column pipeline, and a new `Resource` needs authorization plumbing for two fields. `CbomSyncSkipRepository` stays a plain `JpaRepository` plus `JpaSpecificationExecutor`.
- **Retention age is `last_attempt_at`**, not `first_skipped_at`: a written-off document the repository keeps offering keeps its row, one that disappeared from the repository expires. The index follows.
- **Batch size and per-run cap are constants** (500 and 20), not yml: the ticket asks for nothing new in yml, and the daily sweep catches up.
- The applied migration `V202609081800` is not edited (Flyway checksum); the new migration's header and the entity Javadoc describe the shipped behaviour.
- `Operation.RETRY` is added to the audit enum (interfaces); the permission action stays `UPDATE`, so there is no new `ResourceAction` and no OPA change.
- The retention floor is one day, not zero: a retention of nothing would remove a write-off as it lands.
- PR 0 of the ruling (`assetSyncError` alone) is folded into this train to keep the PR count down.

## Tests

Unit and architecture: `CbomSyncPolicyTest`, `CbomSyncSkipRetentionSweeperTest`, `CbomSyncSkipRetentionTaskTest`, `CbomSyncPropertiesTest`, `CbomAssetIngestServiceTest`, `CbomAssetDetachServiceTest`, `TransactionalBoundaryArchTest`, `NativeQuerySchemaArchTest`, `ExternalServiceAuthorizationArchTest`, `ContextSignatureGuardTest`.

Integration (Testcontainers PostgreSQL): `CbomSyncITest` (42 to 46 classes of scenario: default order, every advertised filter condition end to end, every sort key, paging with the uuid tie-break, the refusals and their messages, permission gating, the retry reset and the next run storing it, the retry no-op, not found, the catalogue), `CbomServiceITest` (the listing and the detail carry the reason, and omit it when there is none), `CbomSyncSkipRetentionSweeperITest`, `CbomSyncSkipRetentionMigrationITest`, `SettingServiceITest`, `SchedulerServiceITest`, `CbomSyncTaskITest`.

All green locally against PostgreSQL; `spotless:check checkstyle:check` clean.

## Review passes

Two independent max-effort review agents over `git diff origin/main...HEAD`, then a focused pass on each fix. What they found and what changed:

| Finding | Fix |
|---|---|
| Every new 422 message rendered its own template: `ValidationError` interpolates `{}`, not `%s`, so a rejected sort answered ``["The sync skip list cannot be ordered by %s; use one of %s"]`` | The four messages pre-format with `String.formatted`, as `ListingSortResolver` does, and the integration test asserts each one names the offending identifier and the allowed set |
| The migration integration test left the pooled connection's `search_path` on a schema it then dropped, so a later test borrowing that connection could fail on an unqualified statement | `RESET search_path` in the `finally`, as the three sibling scratch-schema tests do; the sweeper test's insert is schema-qualified |
| The catalogue published one of the four ordering keys, while the contract promised it named everything the list may be ordered by | The contract sentence is narrowed (interfaces#981, follow-up commit): the catalogue is the filter fields, the list operation names the ordering keys |
| `displayable` was set to `false` where the contract says it is never set | Left unset |
| The serial number filter lower-cased both sides and left LIKE's wildcards active, unlike every other platform string filter | Case-sensitive, wildcards escaped through `FilterPredicatesBuilder`; two values sent for the single-valued field are refused rather than silently truncated |
| A retry racing the sweep could answer 200 with a row that no longer exists | The re-read is unconditional for a written-off row, so a vanished row answers 404 |
| The settings read sat above the task's `try`, the one path its Javadoc promises is covered | Moved inside |
| Four Javadoc passages described behaviour the retention and the retry had just changed, and `CbomSyncPolicy.skipRetention()` had no caller | Corrected, and the accessor removed |
| The index used an `ix_` prefix; every other index in the repository uses `idx_` | Renamed, migration and its test together |
| The task unit test depended on whatever another test had left in the static settings cache | It seeds and restores the cache itself |
| Tests named for conditions asserted none of them | Thirteen integration tests added: every advertised condition end to end, including that the catalogue advertises only what the list serves; every ordering key; the tie-break; `columns` accepted and ignored; the retry keeping the last attempt's record |

`/code-review high` was dispatched but stopped on a session spend limit before reporting; the two agent passes and the focused re-review stand in its place.

**Not covered by a test:** the retry-versus-sweep race is defended by the unconditional re-read, but provoking it needs a delete between two statements of one method.

**Review round on this PR (Michal and Copilot, both on the same statement).** The retention batch delete tested `state` and `last_attempt_at` only inside its subselect, so the outer statement was `uuid IN <already-computed set>`: a row an operator's retry moved to `retrying`, or a run's fresh failure re-dated, while the delete waited for its lock was removed anyway, because the qual re-checked under EvalPlanQual still held. That swallowed an accepted retry, or dropped a live record and handed its document a fresh budget. Fixed in `8aa74478`: the subselect claims its victims with `FOR UPDATE SKIP LOCKED` and the delete repeats the predicate, which is what the repository's own batch rule prescribes and what `SigningRecordRollupSql` already does. A new integration test holds a row in an uncommitted retry and runs the sweep on its own thread with a deadline, because the regression blocks rather than fails: against the old statement it times out, against the fix it passes.

`722391e0` adds the asset sync reason to `PropertyColumnDisplayabilityITest`'s decided column list, which is the one test the interfaces pin had been keeping CI from running.

## Merge order

OmniTrustILM/interfaces#981 is **merged** (`fbca1eaf`) and its snapshot published, so the `Depends-On` marker is gone from this description and CI has resolved against mainline: every job is green, the image build included.

Mergeable once the branch is up to date. The fe-administrator PR for this story, OmniTrustILM/fe-administrator#2165, follows after this is merged and deployed to `main`.

core#2264 (story 1) is merged, so the new sweep task runs outside the scheduler listener's transaction as designed.

Closes #2250


src/main/java/com/otilm/core/dao/repository/cbom/CbomSyncSkipRepository.java
```diff
@@ -69,4 +76,34 @@ void upsertAttempt(@Param("uuid") UUID uuid, @Param("serialNumber") String seria
...
+    @Modifying(clearAutomatically = true, flushAutomatically = true)
+    @Query(value = """
+            DELETE FROM {h-schema}cbom_sync_skip
+            WHERE uuid IN (
+                SELECT uuid FROM {h-schema}cbom_sync_skip
+                WHERE state = 'PERMANENTLY_SKIPPED' AND last_attempt_at < :cutoff
+                ORDER BY last_attempt_at, uuid
+                LIMIT :limit)
+            """, nativeQuery = true)
+    int deleteWrittenOffBefore(@Param("cutoff") OffsetDateTime cutoff, @Param("limit") int limit);
```
vyskocilm, line 121: L100-108: bug: the batch DELETE takes no row lock, so it removes rows a concurrent writer has just moved out of scope.

The outer `WHERE` is `uuid IN (...)` and nothing else. `state` and `last_attempt_at` are tested only inside the subselect, on a different scan node from the one that takes the DELETE's row lock. The `LIMIT` blocks sublink pull-up, so the subselect is a SubPlan evaluated against the statement snapshot; under EvalPlanQual the qual rechecked after the lock is `uuid IN <already-computed uuid set>`, which stays true whatever the concurrent writer did.

Two ways that loses data:

- **Operator retry swallowed.** The sweep's subselect picks up row X. Before the DELETE reaches X's lock, `requestRetry` commits `state = 'RETRYING', attempts = 0`. X is deleted anyway. `retrySyncSkip`'s unconditional re-read then answers 404 for a retry the database had already accepted, and the document is never re-offered — the sync watermark moved past it long ago.
- **Live record deleted.** `upsertAttempt` commits `last_attempt_at = now` for a document the repository still offers, between the subselect and the lock. The row is deleted at zero days old. The next run re-creates it with `attempts` 1, handing a document that had already spent its budget a fresh full one.

CLAUDE.md prescribes the clause for exactly this case ("a batch **skips** (`FOR UPDATE ... SKIP LOCKED`)"), and `SigningRecordRollupSql` is the in-repo precedent — `SKIP_CONTENDED_ROWS` is `FOR UPDATE OF sr SKIP LOCKED`, and its victim CTE is `MATERIALIZED` with the Javadoc reason "a re-evaluated `LIMIT` could pick a different set".

Fix:

```sql
DELETE FROM {h-schema}cbom_sync_skip
WHERE uuid IN (
    SELECT uuid FROM {h-schema}cbom_sync_skip
    WHERE state = 'PERMANENTLY_SKIPPED' AND last_attempt_at < :cutoff
    ORDER BY last_attempt_at, uuid
    LIMIT :limit
    FOR UPDATE SKIP LOCKED)
  AND state = 'PERMANENTLY_SKIPPED' AND last_attempt_at < :cutoff
```

`FOR UPDATE` makes the subselect do its own per-row recheck and drop a row a retry has moved; `SKIP LOCKED` keeps the sweep from blocking behind an uncommitted retry while it holds the cluster lock. Repeating the predicate on the outer DELETE closes the EPQ hole.

One consequence worth a comment in `CbomSyncSkipRetentionSweeper`: `SKIP LOCKED` makes a short batch possible while eligible rows remain, and L61 ends the sweep on `lastBatch < BATCH_SIZE`. A concurrent `CbomSyncSkipWriter.resolve` can already do that today. One day's delay on a daily job is fine — but the loop currently reads as if a short batch proves the table is clean.

vyskocilm (commented):
One critical finding. Posted inline.

Checked and clean, so not raised: `Resource.CBOM` is declared with the two-arg constructor, so `hasObjectAccess` is false — the listing taking no `SecurityFilter` and the retry taking a bare `UUID` are correct, not a gap. `assetSyncError` reaching the wire for the first time is safe: every writer of that column passes a fixed sentence or `CbomAssetIngestService.safeReason`, which gates on `ValidationException` and otherwise routes through `CryptoAssetConstraintTranslator.describe`, so no driver text can land there. The retry is a compare-and-swap, not a read-then-write, so two concurrent retries and retry-versus-write-off both resolve correctly. The advisory lock is released on every path out of `sweep()`. `Operation.RETRY` and `ResourceAction.UPDATE` are both already in the synced catalogue, so no OPA change is owed. Migration, settings bounds, writer/orchestrator pairs and test-context count all check out; the architecture suite, the touched ITests and `spotless:check checkstyle:check` were executed, all green.

Also corrected on interfaces#981: I had claimed an operator-retried row whose document left the repository would never expire. `runSync` calls `retrySkipped` over every `RETRYING` row independent of the feed, so it does expire. Withdrawn there.

src/main/java/com/otilm/core/dao/repository/cbom/CbomSyncSkipRepository.java
karolhrdina: Fixed in `8aa74478`, with your SQL.

The analysis holds, and the repository's own rule prescribes the clause: CLAUDE.md says a batch **skips** (`FOR UPDATE ... SKIP LOCKED`) because it has no per-record contract and waiting lets two sweeps picking overlapping victims in different index orders deadlock, and `SigningRecordRollupSql.SKIP_CONTENDED_ROWS` is the in-repo precedent. The delete now claims its victims and repeats the predicate on the outer statement; neither half is redundant, and the Javadoc says which hole each one closes.

The loop comment is in too: a short batch no longer reads as proof that the table is clean, since `SKIP LOCKED` and a concurrent `resolve` can both shorten one. A day's delay on a daily job is the right trade.

Proved rather than asserted: `CbomSyncSkipRetentionSweeperITest.aRowAConcurrentWriterHoldsIsSkippedRatherThanDeletedUnderIt` holds a row in an uncommitted `state = 'RETRYING'` update, runs the sweep, and checks the free row went while the held row survived with the retry intact. The sweep runs on its own thread with a 30-second deadline, because the regression does not fail the statement — it blocks. Against the old SQL the test errors with `TimeoutException` after 51s; against the fix it passes in 18s.

src/main/java/com/otilm/core/dao/repository/cbom/CbomSyncSkipRepository.java
```diff
@@ -69,4 +76,34 @@ void upsertAttempt(@Param("uuid") UUID uuid, @Param("serialNumber") String seria
...
+     */
+    @Modifying(clearAutomatically = true, flushAutomatically = true)
+    @Query(value = """
+            DELETE FROM {h-schema}cbom_sync_skip
+            WHERE uuid IN (
+                SELECT uuid FROM {h-schema}cbom_sync_skip
+                WHERE state = 'PERMANENTLY_SKIPPED' AND last_attempt_at < :cutoff
+                ORDER BY last_attempt_at, uuid
+                LIMIT :limit)
+            """, nativeQuery = true)
```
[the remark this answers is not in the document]
karolhrdina: Fixed in `8aa74478`. The subselect now takes `FOR UPDATE SKIP LOCKED` and the delete repeats `state` and `last_attempt_at` on its own qual, so a row moved out of scope between the pick and the lock is dropped rather than deleted. `SKIP LOCKED` rather than a wait: the sweep holds the cluster lock, and this is a batch with no per-record contract.

An integration test holds a row in an uncommitted retry and asserts the sweep skips it and leaves the retry standing.

vyskocilm (approved):

PR #2289, opened 2026-09-20
Handle unavailable OAuth2 JWK Set URLs

### LINKED ISSUE #2286: Settings - Authentication - Provider Detail: 422 on unreachable JWK Set URL makes the provider impossible to open, edit or delete
### Description
Opening the detail page of an OAuth2 authentication provider whose **JWK Set URL** no longer resolves returns `422 Unprocessable Content` and the page never renders. **Edit** and **Delete** exist only on that detail page, so the provider becomes unmanageable from the UI: it stays in the **Settings → Authentication** list, cannot be inspected, cannot be corrected, and cannot be removed.

The read path performs an outbound network call and treats its failure as a validation error. `SettingServiceImpl.getOAuth2ProviderSettings()` calls `checkJwkSetValidity()` on every `GET`, which opens a connection to the stored `jwkSetUrl`; the resulting `IOException` is rethrown as `ValidationException`, which the API maps to `422`. The same check runs inside `validateOAuth2ProviderSettings()` on `PUT`, so the stale URL cannot be corrected through the form either.

`DELETE /v1/settings/authentication/oauth2Providers/{providerName}` does not run this check and still succeeds, but the provider list exposes only a **Create** action — no row-level delete — so that path is reachable only through the API.

Any change that makes a previously valid provider host unreachable triggers this: a decommissioned identity-provider domain, a DNS change, or an environment migration. Providers configured before such a change are the affected ones.

### Steps to Reproduce
1. Go to **Settings → Authentication**.
2. Create an OAuth2 provider with a valid **JWK Set URL**, e.g. `https://oidc.example.com/realms/demo/protocol/openid-connect/certs`.
3. Make that host unreachable — remove its DNS record, or set the stored URL to a host that does not resolve.
4. Return to **Settings → Authentication**; the provider is still listed.
5. Click the provider to open its detail page.
6. Inspect `GET /v1/settings/authentication/oauth2Providers/{providerName}` in the browser network tab.

### Expected Behavior
- The detail page opens and renders the stored configuration.
- An unreachable **JWK Set URL** is surfaced as a non-blocking warning on the page, not as a failure of the whole request.
- **Edit** and **Delete** stay available so the provider can be corrected or removed.
- A delete action is reachable even when the provider's detail page cannot be rendered.

### Actual Behavior
- `GET .../oauth2Providers/{providerName}` returns `422 Unprocessable Content` with `Unable to open connection for JWK Set URL: ...`.
- The detail widget stays locked; **Edit** and **Delete** are never rendered.
- `PUT` fails with the same validation, so the stored URL cannot be fixed from the UI.
- The provider stays in the list permanently; the only removal path is a direct `DELETE` API call.

### Severity
Critical

### Module
Auth

### Attachments

_No response_

### Logs

_No response_

### Environment
ILM Core: 2.20.0-SNAPSHOT (18 Sept 2026, 11:47)
Commit: 669b45c
Branch: main
Build time: 2026-09-18T09:47:49Z
ILM Frontend: 2.19.0 (18 Sept 2026, 12:37)
Client:
- macOS Tahoe 26.5.1

lubomirw:
## TLDR

Core keeps OAuth2 provider settings readable and editable when a configured JWK Set cannot be loaded. Retrieval returns the stored provider with empty keys plus a typed failure reason, while save-time validation remains network-free.

Depends-On: OmniTrustILM/interfaces#989

## What changed

- Stop fetching remote JWK Sets while saving provider settings, so an unavailable but well-formed URL can still be persisted or corrected.
- Return `unavailable`, `invalid`, or `tooLarge` through `jwkSetLoadFailure` when retrieval cannot produce keys, while retaining the empty key-list fallback.
- Handle legacy provider rows with neither explicit JWK source without failing retrieval.
- Restrict JWK Set URLs to HTTP(S) with a host and valid port, and preserve safe URI syntax detail in validation errors.
- Limit remote JWK Set responses to 1 MiB and avoid exposing raw network or parser exception details.
- Normalize blank URL and embedded JWK Set values, while continuing to validate embedded sets and public-key conversion on save.
- Cover reachable, unavailable, malformed, oversized, hostless, out-of-range-port, blank-value, legacy-row, embedded-key, and bulk-update paths.

## Testing

- `mvn -o "-Dtest=SettingServiceITest,JwtDecoderITest,WireMockPortsGuardTest,TransactionalBoundaryArchTest,InterfacesPinGuardTest" test jacoco:report` — 72 tests, 0 failures.
- `mvn -o spotless:check checkstyle:check` — passes.
- Sonar-equivalent new-code coverage — 93.88% (92/98 lines and branches).
- Independent correctness review — SHIP after the out-of-range-port follow-up fix.

Closes #2286


src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
```diff
@@ -974,55 +991,84 @@ private Map<String, Map<String, Setting>> mapSettingsByCategory(List<Setting> se
...
+            convertJwkToListOfKeyDtos(checkJwkSetValidity(settingsDto));
+            return;
+        }
+        parseJwkSetUrl(jwkSetUrl);
+    }
+
+    private URL parseJwkSetUrl(String jwkSetUrl) {
+        URI uri;
+        try {
+            uri = new URI(jwkSetUrl);
```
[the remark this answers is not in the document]
lubomirw: Fixed in 588d118d3. JWK Set URLs are now parsed as server authorities before host and port checks, with regression coverage rejecting both `:-1` and `:+1` signed-port forms.

klaraf755 (commented):
The backend now returns `jwkSetLoadFailure`, but nothing consumes it yet — an FE Task is needed to render the unreachable JWK Set URL as a non-blocking warning on the provider detail page. I could not find one in `fe-administrator`; worth filing alongside this PR.

src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
```diff
@@ -974,55 +991,84 @@ private Map<String, Map<String, Setting>> mapSettingsByCategory(List<Setting> se
...
+        try {
+            String scheme = uri.getScheme();
+            if (!"http".equalsIgnoreCase(scheme) && !"https".equalsIgnoreCase(scheme)) {
+                throw new ValidationException("JWK Set URL must use http or https.");
+            }
+            URL url = uri.toURL();
+            if (url.getHost() == null || url.getHost().isBlank()) {
+                throw new ValidationException("JWK Set URL must include a host.");
+            }
+            if (uri.getPort() > 65535) {
```
klaraf755: The port range is checked after `uri.toURL()`, so this only works because the current JDK happens to accept port 99999 in a `URL`. On a JDK that tightens `URL` construction, `toURL()` throws first and the catch below collapses it into the generic `"JWK Set URL is invalid."`, taking both port tests with it.

Related: that same `catch (MalformedURLException | IllegalArgumentException)` also spans the three `ValidationException` throws above, and only leaves their messages intact because `ValidationException extends RuntimeException` rather than `IllegalArgumentException` — a fragile invariant to rely on.

Suggest validating scheme, host and port on the `URI` (`uri.getHost()`, `uri.getPort()`), then calling `toURL()` inside a narrow try of its own.

src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
```diff
@@ -974,55 +991,84 @@ private Map<String, Map<String, Setting>> mapSettingsByCategory(List<Setting> se
...
         if (settingsDto.getJwkSetUrl() != null) {
             try {
-                URL url = new URI(settingsDto.getJwkSetUrl()).toURL();
+                URL url = parseJwkSetUrl(settingsDto.getJwkSetUrl());
                 URLConnection urlConnection = url.openConnection();
                 urlConnection.setConnectTimeout(5000);
                 urlConnection.setReadTimeout(5000);
-                try (InputStream stream = url.openStream()) {
-                    jwkSet = new String(stream.readAllBytes());
+                try (InputStream stream = urlConnection.getInputStream()) {
```
klaraf755, line 1048: The http/https guard covers the configured URL only — `HttpURLConnection` follows same-protocol redirects by default, so a JWKS host can still send the fetch to an internal address. Low impact (nothing is echoed back unless it parses as a JWK Set), but `setInstanceFollowRedirects(false)` on the connection would make the guard complete. Optional.

src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
```diff
@@ -856,6 +857,7 @@ public void updateAuthenticationSettings(AuthenticationSettingsUpdateDto authent
 
     @Override
     @ExternalAuthorization(resource = Resource.SETTINGS, action = ResourceAction.DETAIL)
+    @org.springframework.transaction.annotation.Transactional(propagation = Propagation.NOT_SUPPORTED)
```
klaraf755: The fully-qualified name is only needed because the simple name is taken by the `jakarta.transaction.Transactional` import on line 59 — which is the minority convention here (132 files import Spring's, 29 jakarta's). Flipping the class-level import lets both read plainly:

```diff
-import jakarta.transaction.Transactional;
+import org.springframework.transaction.annotation.Transactional;
```
```diff
-    @org.springframework.transaction.annotation.Transactional(propagation = Propagation.NOT_SUPPORTED)
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
```

No behavioral change — Spring's `AnnotationTransactionAttributeSource` resolves both to the same attribute, and both default to `REQUIRED` with rollback on `RuntimeException`/`Error`. One side effect in the right direction: `TransactionalBoundaryArchTest.effectivePropagation` only recognizes Spring's annotation, so today the class-level jakarta one is invisible to it and every other method in the class reports no effective `@Transactional`. Rules A and D should not bite (`service.impl`, not `service.writer`; the `NOT_SUPPORTED` method is read-only), but worth running that test to confirm rather than assuming.

src/test/java/com/otilm/core/service/SettingServiceMockedTest.java
```diff
@@ -23,10 +28,7 @@
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
-/**
- * The advisory locks a platform settings update takes, and their order. The integration tests are single-threaded, so
- * they would stay green with the locks gone or swapped; this pins what they cannot.
- */
+/** Unit coverage for settings service behavior that does not require a Spring context. */
```
klaraf755: This restates what the class name and `@ExtendWith(MockitoExtension.class)` already say. The doc it replaced carried something the code could not — why the lock-order tests exist at all ("the integration tests are single-threaded, so they would stay green with the locks gone or swapped; this pins what they cannot"). Suggest dropping this line and moving that rationale onto the two lock tests.

src/test/java/com/otilm/core/service/SettingServiceMockedTest.java
```diff
@@ -70,6 +75,21 @@ void aSectionLeftOutOfTheUpdateTakesNoLock() {
         verify(settingRepository, never()).lockUtilsWrites();
     }
 
+    @ParameterizedTest
+    @ValueSource(strings = {"http://example.com:-1/jwks", "http://example.com:+1/jwks"})
+    void anOAuth2ProviderUpdateRejectsAJwkSetUrlWithASignedPort(String jwkSetUrl) {
```
klaraf755, line 83: This is pure URL parsing sitting in a class otherwise about lock ordering, while its five siblings (non-http, hostless, out-of-range port, malformed URL, malformed embedded set) run as Spring integration tests in `SettingServiceITest`. Worth picking one home for the whole group — this class is the cheaper one, since none of them need a context.

src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
lubomirw: Fixed in 1dcaad7d6. Scheme, host, and port are now validated on the URI before conversion, and the catch now encloses only uri.toURL().

src/test/java/com/otilm/core/service/SettingServiceMockedTest.java
lubomirw: I kept the signed-port regression in the mocked service test because it needs no Spring or persistence behavior, while the existing integration cases continue to cover validation through the real service setup. Consolidating them would be a test-layout refactor rather than a correctness fix.

src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
lubomirw: Fixed in 1dcaad7d6. SettingServiceImpl now uses Spring's @Transactional at both class and method level, so TransactionalBoundaryArchTest can see the class-level REQUIRED boundary.

src/main/java/com/otilm/core/service/impl/SettingServiceImpl.java
lubomirw: Same-protocol redirects do not bypass the HTTP/HTTPS restriction, and direct internal HTTP/HTTPS URLs are currently permitted as well. Disabling redirects would break legitimate IdP or CDN redirects without establishing an outbound-host security boundary; if we introduce such a policy, destinations and every redirect hop should be validated together.

src/test/java/com/otilm/core/service/SettingServiceMockedTest.java
lubomirw: Fixed in 1dcaad7d6. The generic class documentation is removed, and the single-threaded integration-test rationale now sits directly on each lock test.

lubomirw:
Two cross-repo follow-ups remain:

- Core now exposes jwkSetLoadFailure, but e-administrator does not consume it. The non-blocking provider-detail warning is tracked in OmniTrustILM/fe-administrator#2174.
- The Interfaces pin remains red while OmniTrustILM/interfaces#989 is open.

Neither requires another Core code change.

klaraf755 (approved):

PR #2294, opened 2026-09-21
Surface a disabled secret as 422 instead of 500

### LINKED ISSUE #2292: Disabled secret in an operation-path attribute returns HTTP 500 instead of a readable 422
### Description

When an operation-path attribute references a SECRET resource that is disabled, the caller gets `HTTP 500` with the body `Internal server error.` instead of the readable `422` the platform already produces for this condition.

`SecretServiceImpl.getSecretContent` throws a `ValidationException` naming the secret, and `ExceptionHandlingAdvice` maps `ValidationException` to `422` with the message in the body. That message never reaches the caller, because `OperationAttributeResolver.resolveForConnectorRequestAsSystem` catches it and wraps it:

```java
} catch (AttributeException | NotFoundException | ValidationException e) {
    throw new ConnectorException(
            "Unable to resolve stored attribute references for connector request (connector "
                    + connectorUuid + ")", e);
}
```

`ExceptionHandlingAdvice` has handlers for the subclasses (`ConnectorClientException`, `ConnectorServerException`, `ConnectorCommunicationException`, `ConnectorProblemException`, `ConnectorEntityNotFoundException`) but none for the base `ConnectorException`, so the wrap falls through to `@ExceptionHandler(Exception.class)` and becomes a `500` with a generic body.

Three things are lost: the status (422 -> 500), the message the operator needs, and the distinction between a configuration mistake the operator can fix and a platform fault.

Suggested fix: drop `ValidationException` from that catch. It is a `RuntimeException`, so no signature change is needed, and `AuthHelper.runAsSystem` restores the security context in a `finally`, so letting it propagate is safe. It then reaches the existing `422` handler with its own message.

`NotFoundException` in the same catch has the same problem (a missing secret should be `404`), but it is checked and the method declares `throws ConnectorException`, so unwrapping it is a larger change worth deciding separately.

**Second, smaller defect in the same path.** The message reads `Secrethms-pin is not enabled` — two string concatenations in `SecretServiceImpl` are missing a space:

```java
"Source vault profile" + secret.getSourceVaultProfile().getName() + " is not enabled"   // line 911
"Secret" + secret.getName() + " is not enabled"                                          // line 914
```

The line immediately above them uses `.formatted()` correctly, so these two are the odd ones out.

### Steps to Reproduce

1. Register a PKCS#11 Cryptography Provider connector and create a Generic secret holding the token user PIN.
2. Create a token that references that secret through the `pin` attribute.
3. Disable the secret.
4. Open the token profile creation form for that token (`GET /v2/tokenProfiles/.../attributes`, or the key-usages call on the same path).

### Expected Behavior

`422 Unprocessable Entity` with the existing validation message, e.g. `Secret hms-pin is not enabled` — telling the operator which secret is disabled so they can re-enable it.

### Actual Behavior

`500 Internal Server Error` with the body `Internal server error.` The cause is only visible in the server log:

```
com.otilm.api.exception.ConnectorException: Unable to resolve stored attribute references for connector request (connector 0a816416-b627-4950-aaee-fa793eb37da9)
	at com.otilm.core.service.handler.OperationAttributeResolver.lambda$resolveForConnectorRequestAsSystem$0(OperationAttributeResolver.java:63)
	at com.otilm.core.service.handler.token.TokenProviderV2Adapter.tokenScopedRequest(TokenProviderV2Adapter.java:120)
	at com.otilm.core.service.impl.TokenInstanceServiceImpl.listSupportedKeyUsages(TokenInstanceServiceImpl.java:354)
	...
Caused by: Secrethms-pin is not enabled
	at com.otilm.core.service.impl.SecretServiceImpl.getSecretContent(SecretServiceImpl.java:914)
	at com.otilm.core.service.impl.SecretServiceImpl.getResourceObjectContent(SecretServiceImpl.java:1188)
	at com.otilm.core.attribute.engine.ConnectorRequestAttributesBuilder.resolveContent(ConnectorRequestAttributesBuilder.java:81)
	at com.otilm.core.attribute.engine.ConnectorRequestAttributesBuilder.dereferenceForConnectorRequest(ConnectorRequestAttributesBuilder.java:67)
```

The same wrap is reached from both `listTokenProfileAttributes` and `listSupportedKeyUsages`, so the whole token profile creation flow fails this way.

### Environment

Reproduced against a local Core with the PKCS#11 Cryptography Provider connector and a SoftHSM sidecar.

Verified present on `upstream/main`: the catch at `OperationAttributeResolver.java:59`, the two concatenations at `SecretServiceImpl.java:911` and `:914`, and no `@ExceptionHandler(ConnectorException.class)` in `ExceptionHandlingAdvice`.

klaraf755:
Fixes #2292

A disabled secret referenced by an operation-path attribute answered `500 Internal server error.` instead of the `422` the platform already produces for it.

`SecretServiceImpl.getSecretContent` throws a `ValidationException` naming the secret, which the advice maps to `422` with the message in the body. `OperationAttributeResolver` caught it and wrapped it in `ConnectorException`, which has no handler of its own — only its subclasses do — so it fell through to `@ExceptionHandler(Exception.class)` and lost both the status and the message.

`ValidationException` is no longer caught there. It is unchecked, so the signature is unchanged, and `runAsSystem` restores the security context in a `finally`, so letting it propagate is safe. This also brings the operation path into agreement with `DiscoveryV2Client.dereferenceAsSystem`, which never caught it.

Second, the message itself read `Secrettestsecret is not enabled` — two concatenations in `SecretServiceImpl` were missing a space. Both now use `.formatted()`, matching the line above them.

## Not fixed here

**`NotFoundException` stays wrapped.** A *missing* secret still renders as `500` rather than `404`. It is checked and the method declares `throws ConnectorException`, so unwrapping it needs a decision rather than a one-line change.

**`ConnectorException` is the wrong type for this in the first place.** Of the 22 base-class throw sites in Core, 20 read "Connector returned…" / "Connector did not…". The two exceptions are this one and `DiscoveryV2Client:309`, both reporting a failure to resolve Core's *own* stored references, and neither sets the `connector` field the class carries. The collapse is forced by `AuthHelper.runAsSystem` being `<T, E extends Exception>` — one checked type through the lambda — not by any claim about the connector; the discovery site says as much in its Javadoc.

Worth a follow-up: a dedicated checked type for reference resolution, mapped to `422`, thrown by both sites. That satisfies the single-checked-type constraint, stops naming the connector as the faulting party, and gives both a real status. Adding an `@ExceptionHandler(ConnectorException.class)` instead would be cheaper but would restatus all 22 sites at once.

## Testing

- `OperationAttributeResolverTest.propagatesValidationFromDerefUnwrapped` replaces the test that asserted the wrap — it encoded the bug.
- `SecretServiceITest.testGetSecretContentDisabled` now asserts both messages, not just the exception type.
- Both were confirmed failing before the fix. 132 tests pass across the resolver, secret, and Token/Key/Authority adapter suites; `spotless:check` and `checkstyle:check` clean.


src/main/java/com/otilm/core/service/handler/OperationAttributeResolver.java
```diff
@@ -56,10 +55,11 @@ public List<RequestAttribute> resolveForConnectorRequestAsSystem(UUID connectorU
         return authHelper.runAsSystem(AuthHelper.ATTRIBUTE_CONTENT_RESOLVER_USERNAME, () -> {
             try {
                 return connectorRequestAttributesBuilder.dereferenceForConnectorRequest(connectorUuid, stored);
-            } catch (AttributeException | NotFoundException | ValidationException e) {
-                // A stored reference may be unresolvable, or point at a disabled/invalid-state secret or vault profile
-                // (unchecked ValidationException). Surface all through the declared ConnectorException contract so the
-                // operation fails cleanly instead of escaping as a raw RuntimeException.
+            } catch (AttributeException | NotFoundException e) {
```
ivosh, line 65: **! Major** | Error handling

The narrowed catch in `OperationAttributeResolver.java:58` is right for the HTTP path, but one caller of `resolveForConnectorRequestAsSystem` is not on an HTTP path and recovers *only* from `ConnectorException`.

`AuthorityProviderV3Adapter.pollStatus` (`AuthorityProviderV3Adapter.java:396`) calls `authorityAttributesFor(authority)` at line 403 and `resolvedRaProfileAttributes(...)` at 404, both of which go through this resolver. Its caller is `CertificateStatusPollListener.processMessage`, which wraps the call in `try { status = async.pollStatus(cert, msg.op()); } catch (ConnectorException e) { ... }` at `CertificateStatusPollListener.java:123-134`. That catch is the *only* place the poll flow can terminate a failing operation: it calls `applyFailure(cert, msg.op(), timeoutReason(msg.op()))` on `isLastAttempt(msg)` (line 129-130).

Before this PR a disabled secret/vault profile behind an authority or RA-profile attribute arrived there as a wrapped `ConnectorException`, so the poll eventually failed the certificate out. After this PR it arrives as `ValidationException`, which:
- escapes `processMessage`,
- is swallowed by the generic `catch (Exception e) { logger.error("Unexpected error in endpoint ...") }` in `AbstractJmsEndpointConfig.java:109`, so the message is neither retried nor dead-lettered,
- leaves the `certificate_status_poll` row undeleted, so `CertificateStatusPollSweeper` re-enqueues it on every sweep ("the listener owns the async poll flow's state transitions and deletes the row" — `CertificateStatusPollSweeper.java:26`),
- and is not rescued by `PendingIssueReaper`, whose selection is `certificateRepository.findStalePendingIssueWithoutPollRow(...)` (`PendingIssueReaper.java:74`, query at `CertificateRepository.java:223-228`) — it deliberately skips certificates that still have a poll row.

Net effect: a certificate in PENDING_ISSUE/PENDING_REVOKE behind a disabled secret polls forever and never reaches FAILED, where it previously timed out. `CertificateStatusPollListenerTest` has no case for an unchecked exception out of `pollStatus`, so nothing catches this. (Grounded by local `git grep` over `origin/main` plus `gh search code --owner OmniTrustILM --owner 3KeyCompany`, which finds no other repo calling this resolver.)

_Suggested:_ Either catch `ValidationException` alongside `ConnectorException` in `CertificateStatusPollListener.processMessage` so the poll still fails the certificate on the last attempt, or keep the wrap on the poll path only and unwrap solely for the request-scoped callers.

<sub>flagged by Correctness & races + Reference integrity</sub>

src/main/java/com/otilm/core/service/handler/OperationAttributeResolver.java
```diff
@@ -56,10 +55,11 @@ public List<RequestAttribute> resolveForConnectorRequestAsSystem(UUID connectorU
...
+                // A stored reference may be unresolvable. Surface that through the declared ConnectorException
```
ivosh: **~ Minor** | Documentation

This PR changes the effective contract of the public `resolveForConnectorRequestAsSystem` (OperationAttributeResolver.java:44): its signature still says `throws ConnectorException`, but an unchecked `ValidationException` from the deref now escapes to the caller by design. That decision is recorded only in the 4-line inline comment inside the lambda at OperationAttributeResolver.java:59-62, where nobody reading the method signature or the class doc will see it.

The neighbour pattern points the other way. The class Javadoc at OperationAttributeResolver.java:15-30 is already the documented home for this class's cross-cutting behaviour, structured as `<b>Authorization:</b>` / `<b>Scope:</b>` / `<b>Placement:</b>` paragraphs, and `AuthorityProviderV3Adapter.java:549` explicitly sends readers there ("see its Javadoc for the system-identity trust model") — yet it says nothing about which failures surface as `ConnectorException` and which pass through unwrapped. CLAUDE.md's "Writing for future AI sessions" asks for "Method-level Javadoc on non-obvious behaviour over inline blocks", and the PR description itself cites the peer site `DiscoveryV2Client.dereferenceAsSystem` as one that "says as much in its Javadoc" (that class is not in this working tree — `find src -name 'DiscoveryV2Client*'` returns nothing — so I could only compare against the local neighbour). A caller author reading `throws ConnectorException` at :44 has no way to learn that a disabled secret arrives as a `RuntimeException` instead.

A fourth `<b>Failures:</b>` paragraph in the existing class Javadoc costs two sentences and keeps the block comment at :59-62 short: e.g. "<b>Failures:</b> unresolvable stored references surface as {@link ConnectorException}. A {@link ValidationException} (disabled or invalid-state secret or vault profile) propagates unwrapped — it names the object the operator must fix and maps to 422; wrapping it yields a 500 with no message."

_Suggested:_ Move the propagate-unwrapped rationale into a `<b>Failures:</b>` paragraph on the class Javadoc (or method Javadoc on `resolveForConnectorRequestAsSystem`), leaving a one-line pointer at the catch block.

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
```diff
@@ -803,12 +803,19 @@ void testGetSecretContentDisabled() {
         secret.setEnabled(false);
         secretRepository.save(secret);
         UUID secretUuid = secret.getUuid();
-        Assertions.assertThrows(ValidationException.class, () -> secretService.getSecretContent(secretUuid));
+        ValidationException disabledSecret = Assertions
```
ivosh: **. Nitpick** | Testing

`SecretServiceITest.testGetSecretContentDisabled` (SecretServiceITest.java:802-819) drives two independent behaviours in one method: disabled secret -> message (:806-809), then a state rewind (`secret.setEnabled(true)` + save at :810-811) followed by disabled vault profile -> message (:812-818). This PR did not introduce the bundling, but it doubled what each half asserts, so the method now carries four assertions across two act/assert cycles with a mutate-and-re-save step in the middle, and there is no arrange/act/assert separation between them. A failure in the first `assertEquals` at :809 leaves the secret disabled and the second half unrun, so one regression report covers two unrelated production branches (SecretServiceImpl.java:909-915 vault profile, :913-915 secret).

The siblings in this same class are one behaviour each: `testGetSecretContent` and `getSecretContent_whenFingerprintCalcThrows_shouldThrowValidationException` (SecretServiceITest.java:788-799) both do a single act and a single assert, and the latter also uses the `method_whenCondition_shouldOutcome` naming and an explicit `// act + assert` phase break that `testGetSecretContentDisabled` lacks.

_Suggested:_ Split it into `getSecretContent_whenSecretDisabled_shouldNameTheSecret` and `getSecretContent_whenVaultProfileDisabled_shouldNameTheProfile`, each arranging its own disabled state so neither depends on the other's rewind.

src/test/java/com/otilm/core/service/handler/OperationAttributeResolverTest.java
```diff
@@ -102,16 +102,20 @@ void propagatesConnectorExceptionFromDerefUnwrapped() throws Exception {
     }
 
     @Test
-    void wrapsUnresolvableReferenceValidationAsConnectorException() throws Exception {
+    void propagatesValidationFromDerefUnwrapped() throws Exception {
+        // A disabled/invalid-state secret or vault profile throws an unchecked ValidationException from the deref. It
```
ivosh: **. Nitpick** | Maintainability

This PR states the same non-obvious why — a `ValidationException` maps to 422 and naming the object matters, so wrapping it would yield a 500 with no message — three times, each in different words and none next to the other:

- OperationAttributeResolver.java:59-62 (the declaration site: "...names the object the operator has to fix and renders as 422, which wrapping would turn into a 500 carrying no message at all")
- OperationAttributeResolverTest.java:106-107 ("It names the offending object and maps to 422, so wrapping it would hand the caller a 500 with no message")
- OperationAttributeResolverTest.java:117-118, the `assertSame` failure message ("must reach the 422 handler, not be wrapped as a 500") — the third copy sits eleven lines below the second, inside the same test method

(SecretServiceITest.java:808 paraphrases it a fourth time in one line; that one reads as a genuinely different point — why assert the text rather than the type — so it is fine to keep.)

The authoritative home is the production catch block at OperationAttributeResolver.java:59-62, where the behaviour actually happens. The assertion message at :117-118 earns its place because it is what a maintainer sees when the test fails; the block comment at :106-107 adds nothing the test name `propagatesValidationFromDerefUnwrapped` plus that assertion message do not already carry, and gives a future edit a third place to go stale that grep will not find from either of the other two.

_Suggested:_ Drop the block comment at OperationAttributeResolverTest.java:106-107 and let the test name plus the `assertSame` message at :117-118 carry the why.

src/main/java/com/otilm/core/service/handler/OperationAttributeResolver.java
ivosh: **~ Minor** | Documentation

`OperationAttributeResolver.java:58` narrows the catch so an unchecked `ValidationException` raised inside the deref now escapes `resolveForConnectorRequestAsSystem` — a public method whose signature (`OperationAttributeResolver.java:44`) still advertises only `throws ConnectorException`. The new contract is recorded solely in an implementation comment inside the lambda (`OperationAttributeResolver.java:59-62`), which no caller and no IDE completion sees: `AuthorityProviderV3Adapter.authorityAttributesFor` (:554) and `resolvedRaProfileAttributes` (:564), and the ~20 call sites above them, read the signature only. *Effective Java* Item 74 asks that every exception a method can throw — unchecked included — be documented with an `@throws` tag, precisely because the compiler will not carry it. The sibling this PR aligns the catch with, `DiscoveryV2Client.dereferenceAsSystem`, does carry a Javadoc block explaining its exception collapse even though it is `private`; this method, which is public and is the class the class-level Javadoc calls "the single home" for operation-path resolution, has no method Javadoc at all.

_Suggested:_ Add a Javadoc block on `resolveForConnectorRequestAsSystem` with an `@throws ValidationException` line saying a disabled or invalid-state secret or vault profile propagates unwrapped and renders as 422.

ivosh (changes requested):

src/main/java/com/otilm/core/service/handler/OperationAttributeResolver.java
klaraf755: Confirmed — reproduced it as a failing test before fixing: the ValidationException escaped `pollStatus` and `assertThrows(ConnectorException.class)` got `ValidationException`. Took the second option: `pollStatus` wraps it back into `ConnectorException`, so the listener still fails the certificate out on the last attempt. The wrap is at `pollStatus` rather than in `authorityAttributesFor`/`resolvedRaProfileAttributes` — those feed 14 call sites, mostly request-scoped ones that want the 422. `pollStatus` is the only adapter method the listener calls.

src/main/java/com/otilm/core/service/handler/OperationAttributeResolver.java
klaraf755: Added a `<b>Failures:</b>` paragraph to the class Javadoc covering both halves, and it names `pollStatus` as the caller that wraps. The catch block is down to a one-line pointer.

Minor correction: `DiscoveryV2Client` is in the tree, at `src/main/java/com/otilm/core/service/handler/discovery/DiscoveryV2Client.java` — it catches only `AttributeException | NotFoundException`, which is what this change aligned the operation path with.

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
klaraf755: Split into `getSecretContent_whenSecretDisabled_shouldNameTheSecret` and `getSecretContent_whenVaultProfileDisabled_shouldNameTheProfile`, each arranging its own disabled state — no rewind between them.

src/test/java/com/otilm/core/service/handler/OperationAttributeResolverTest.java
klaraf755: Dropped the block comment; the test name and the `assertSame` message carry it.

src/test/java/com/otilm/core/integration/service/SecretServiceITest.java
```diff
@@ -799,16 +799,30 @@ void getSecretContent_whenFingerprintCalcThrows_shouldThrowValidationException()
...
-    void testGetSecretContentDisabled() {
+    void getSecretContent_whenSecretDisabled_shouldNameTheSecret() {
         secret.setEnabled(false);
         secretRepository.save(secret);
         UUID secretUuid = secret.getUuid();
-        Assertions.assertThrows(ValidationException.class, () -> secretService.getSecretContent(secretUuid));
-        secret.setEnabled(true);
-        secretRepository.save(secret);
+
+        // act + assert
```
ivosh: **. Nitpick** | Maintainability

SecretServiceITest.java:807 and SecretServiceITest.java:820 add `// act + assert` immediately before assertions in tests whose names already state the behavior. The labels add no scenario or invariant; at line 807, the label also separates the useful operator-facing rationale from the assertion it explains.

_Suggested:_ Remove both `// act + assert` lines while retaining the operator-facing rationale comment.

ivosh (approved):

PR #2298, opened 2026-09-22
Answer an empty CBOM identifier instead of failing the bulk delete

### LINKED ISSUE #2293: Bulk delete of CBOMs returns HTTP 500 when the uuid list contains an empty string
### Description

`DELETE /v1/cboms` (the bulk delete) answers **HTTP 500** `Internal server error.` when the list
contains an empty string. Jackson deserialises `""` to a **null** `UUID`, and
`CbomServiceImpl.bulkDeleteCbom` dereferences it on the not-found branch:

```java
Set<UUID> existingUuids = cbomRepository.findExistingUuids(uuids);
for (UUID uuid : uuids) {
    if (!existingUuids.contains(uuid)) {
        messages.add(BulkActionMessageDto.failureWithMessage(uuid.toString(), "", "CBOM entry not found"));
```

`CbomServiceImpl.java:640` — `uuid.toString()` on a null element.

Every other shape of bad input on this endpoint is already handled:

| body | status | response |
|---|---|---|
| `["00000000-0000-0000-0000-000000000000"]` (well-formed, unknown) | 200 | `[{"uuid":"00000000-…","name":"","message":"CBOM entry not found"}]` |
| `["not-a-uuid"]` (unparseable) | 400 | `{"message":"Unable to read HTTP message"}` |
| `[""]` | **500** | `{"message":"Internal server error."}` |

So `""` is the only input that reaches the null: it is valid JSON, it deserialises without
error, and it then fails inside the service.

Suggested fix: treat a null element the way the method already treats an unknown one — report
it as a failure item rather than dereferencing it (`uuid == null` first in the loop), or reject
the request with 422 when any element is null. Either keeps the endpoint's contract that a bulk
call answers 200 with a per-item verdict.

### Steps to Reproduce

1. `POST /v1/cboms/upload` or sync at least one document, so the endpoint has work to do
   (not required — the empty string alone is enough).
2. Call the bulk delete with an empty string in the list:

```
curl -X DELETE http://localhost:8280/api/v1/cboms \
  -H 'ssl-client-cert: <admin>' -H 'Content-Type: application/json' \
  -d '[""]'
```

### Expected Behavior

`400`/`422`, or `200` with a per-item failure — the same treatment a well-formed but unknown
UUID gets.

### Actual Behavior

```
{"message":"Internal server error."}
```

with, in the server log:

```
ERROR [tomcat-handler-63] [com.otilm.core.api.ExceptionHandlingAdvice - 730]: General error occurred:
Cannot invoke "java.util.UUID.toString()" because "uuid" is null
java.lang.NullPointerException: Cannot invoke "java.util.UUID.toString()" because "uuid" is null
	at com.otilm.core.service.impl.CbomServiceImpl.bulkDeleteCbom(CbomServiceImpl.java:640)
```

### Environment

Found by the core#2283 dev end-to-end pass (scenario S15, "delete and keep deleted"), against
`hub.omnitrustregistry.com/ilm/core:develop-latest`
`sha256:4bb33690fcdfe216a91496f48b8931804a1034df229b61434019a244696356b0` — the CI build of
`main` at `64fe940a` — on the `np/core-env` compose deployment.

The delete path is the one built in #2263 (part of #2073); the line is present on `main`.

vyskocilm:
Closes #2293

`DELETE /v1/cboms` answered HTTP 500 when the uuid list carried an empty string. Jackson deserialises `""` to a **null** `UUID`, and two places dereferenced it. Both had to go, because they fail in different deployments:

| audit output | where it failed |
|---|---|
| `NONE` (the default, and what the #2293 report ran on) | `CbomServiceImpl.bulkDeleteCbom` — `uuid.toString()` on the not-found branch |
| `DATABASE` / `CONSOLE` / `ALL` | `AuditLogAspect.getResourceUuidsFromParameter` — `UUID.fromString(v.toString())`, reached from `log()` **before** `joinPoint.proceed()` |

### The service (`dfed683a`)

A null element stays out of the existence query and gets its own per-item failure, so the endpoint keeps its contract of answering with a verdict for every element rather than rejecting the whole call:

```json
[{"uuid":"","name":"","message":"Missing CBOM identifier"}]
```

That matches how the platform's other bulk deletes treat an element they cannot act on — `SigningRecordServiceImpl` and `TspProfileServiceImpl` report per item and never reject the batch.

### The audit aspect (`4dd74351`)

The aspect reads a method's `@LogResource(uuid = true)` parameters to name the audited objects, and mapped every list element through `UUID.fromString(v.toString())`. A null element therefore failed inside the advice, so the caller got a 500 before the service could answer at all. A null element carries no identity to audit, so it is now skipped and the entry names the identifiers actually acted on.

This read serves **24 controller methods** taking a uuid list (TSP profile, signing profile, time quality configuration, compliance v2, …), so the same 500 was reachable on every one of them whenever audit logging was enabled.

### Verification

- Three new tests, all appended to existing classes — no new Spring context, so `ContextSignatureGuardTest.BASELINE` is untouched.
  - `CbomServiceITest#testBulkDeleteCbom_NullElement` — also asserts the non-null neighbour in the same list was really deleted.
  - `CbomServiceITest#testBulkDeleteCbom_OnlyNullElements`
  - `AuditLogAspectITest#nullUuidElement_isSkippedByTheAudit_andTheCallStillAnswers` — audit output `ALL`, drives `CbomController`, asserts both the returned verdict and the written audit entry.
- Each test was run against the unfixed code first and reproduces the reported failure exactly (`Cannot invoke "java.util.UUID.toString()" because "uuid" is null` at `CbomServiceImpl:640`; `Cannot invoke "Object.toString()" because "v" is null` at `AuditLogAspect:328`).
- `CbomServiceITest` 71/71, `AuditLogAspectITest` 6/6, architecture suite 133/133, `spotless:check` + `checkstyle:check` green.

### One open question for the reviewer

The `interfaces` contract declares 200 and 422 for this operation, with no element-level constraint on the body. This PR answers 200 with a per-item verdict; the alternative is `List<@NotNull UUID>` + `@Valid` in `interfaces`, which would answer the declared 422 instead. The per-item verdict was chosen because it matches the sibling bulk deletes and needs no contract change — say so if you would rather have the 422.


src/test/java/com/otilm/core/integration/aop/AuditLogAspectITest.java
```diff
@@ -276,6 +282,35 @@ void deniedWithRecordedPermission_namesThePermissionInTheAuditMessage() {
                         auditLogs.getFirst().getMessage());
     }
 
+    /**
+     * core#2293: an empty string in the request body reaches the controller as a null list element, and the aspect
```
ivosh: **. Nitpick** | Documentation

This PR explains the same non-obvious fact — Jackson turns an empty JSON string into a **null** `UUID` list element — four times, in four different wordings, and none of the four is the declaration of that behaviour (it happens in the API/deserialisation layer, not in any of these files): AuditLogAspect.java:327-329 ("Jackson produces one from an empty JSON string (core#2293)"), CbomServiceImpl.java:638-640 ("Jackson deserialises an empty JSON string to a null UUID"), AuditLogAspectITest.java:286-288 ("an empty string in the request body reaches the controller as a null list element"), and CbomServiceITest.java:1158-1159 ("A null element is what the API layer produces from an empty string in the request body"). This is the cross-file shape of the WHAT-vs-WHY duplication: each copy reads perfectly reasonable on its own, but a future change to how the API layer handles `""` has four independent places to go stale, and grep from any one of them finds none of the others because no two share a phrasing. The two production comments are the load-bearing copies — a maintainer reading either fix needs the why right there. The two test Javadocs are the reducible ones: `testBulkDeleteCbom_NullElement` and `nullUuidElement_isSkippedByTheAudit_andTheCallStillAnswers` already carry the WHAT in their names, and the remainder of each Javadoc restates the production comment rather than adding a test-specific constraint.

_Suggested:_ Keep the rationale on the two production sites and cut the two test Javadocs down to what is specific to the test (or drop them, since the method names already say what is asserted).

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -634,8 +635,18 @@ public List<BulkActionMessageDto> bulkDeleteCbom(List<UUID> uuids) {
...
+        // Jackson deserialises an empty JSON string to a null UUID, so the list can carry an element that identifies
+        // nothing. It stays out of the existence query, and the loop reports a per-item failure without rejecting the
+        // whole call.
+        final List<UUID> identifiers = uuids.stream().filter(Objects::nonNull).toList();
+        final Set<UUID> existingUuids = identifiers.isEmpty()
+                ? Set.of()
+                : cbomRepository.findExistingUuids(identifiers);
         for (UUID uuid : uuids) {
+            if (uuid == null) {
+                messages.add(BulkActionMessageDto.failureWithMessage("", "", "Missing CBOM identifier"));
```
ivosh: **. Nitpick** | Documentation

CbomServiceImpl.java:646-649 adds a new element of the `bulkDeleteCbom` return contract: a null element now yields a `BulkActionMessageDto` with an **empty** uuid and name and the literal message `"Missing CBOM identifier"`. That is the fact a caller needs and cannot infer from the signature — yet the interface Javadoc on `CbomExternalService#bulkDeleteCbom` (src/main/java/com/otilm/core/service/CbomExternalService.java:66-73: "The method will try to delete all CBOMs and return the result if some of them failed to be deleted... @return List of BulkActionMessageDto with the failed message of deletion for CBOM") still describes only the not-found/failed-delete case, and the new behaviour lives solely in an implementation comment a consumer of the interface never reads. This matters more than usual here because the PR's own open question is about whether this per-item verdict or a 422 is the right contract: whichever the reviewer picks, the answer belongs where the contract is declared. This is an observation about the untouched interface file, anchored on the changed line that introduces the behaviour.

_Suggested:_ Add one line to the `CbomExternalService#bulkDeleteCbom` Javadoc stating that a null element is reported as a per-item failure carrying an empty uuid rather than aborting the batch.

ivosh: **. Nitpick** | Correctness

CbomServiceImpl.java:647 answers a null element with `{"uuid":"","name":"","message":"Missing CBOM identifier"}`. Now that the aspect no longer fails first, the same element also reaches the sibling bulk deletes with audit enabled, and they name it differently: `SecuredUUID.fromUuidList` wraps null in a SecuredUUID whose `toString()` is the literal `SecuredUUID[null]` (SecuredUUID.java:54-55), and the per-item failure carries that as the identifier — TspProfileServiceImpl.java:260, TimeQualityConfigurationServiceImpl.java:224 and ComplianceProfileServiceImpl.java:235 (v2) all build `BulkActionMessageDto.failure(uuid.toString(), …)`. So the platform now reports the same malformed element two ways, and the sibling wording leaks an internal class name into a public response body. The per-item shape this PR chose is the better one; only the identifier text diverges.

_Suggested:_ Record the divergence in #2293 (or open a follow-up) so the SecuredUUID-based bulk deletes report a null element with the same empty identifier rather than "SecuredUUID[null]".

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
```diff
@@ -634,8 +635,18 @@ public List<BulkActionMessageDto> bulkDeleteCbom(List<UUID> uuids) {
             return messages;
         }
 
-        Set<UUID> existingUuids = cbomRepository.findExistingUuids(uuids);
+        // Jackson deserialises an empty JSON string to a null UUID, so the list can carry an element that identifies
+        // nothing. It stays out of the existence query, and the loop reports a per-item failure without rejecting the
```
ivosh: **. Nitpick** | Maintainability

CbomServiceImpl.java:638-640 — the first sentence earns its place (Jackson turning an empty JSON string into a null UUID is exactly the non-obvious fact a reader cannot get from the code). The second, "It stays out of the existence query, and the loop reports a per-item failure without rejecting the whole call", walks through the three statements immediately below it: the `filter(Objects::nonNull)` on line 641, the `identifiers.isEmpty()` guard on line 642, and the `if (uuid == null)` branch on lines 646-649 all read plainly. The one part of it that is not restatement is the contract decision (per-item verdict rather than rejecting the batch), which is worth a short clause.

_Suggested:_ Trim the comment to the Jackson fact plus the per-item-verdict rationale and drop the step-by-step narration of the filter, the guard and the loop branch.

src/main/java/com/otilm/core/aop/AuditLogAspect.java
```diff
@@ -324,8 +324,11 @@ private List<UUID> getResourceUuidsFromParameter(LogResource logResource, String
                     parameterValue = optional.get();
                 }
 
+                // A null element carries no identity to audit -- and this read happens before the advised method
+                // runs, so dereferencing it would turn the caller's malformed element into a 500 the endpoint
+                // never gets to answer for. Jackson produces one from an empty JSON string (core#2293).
                 return parameterValue instanceof List<?> listValues
-                        ? listValues.stream().map(v -> UUID.fromString(v.toString())).toList()
+                        ? listValues.stream().filter(Objects::nonNull).map(v -> UUID.fromString(v.toString())).toList()
```
ivosh: **. Nitpick** | Maintainability

AuditLogAspect.java:331 — this PR rewrites the `List` branch of `getResourceUuidsFromParameter` to `listValues.stream().filter(Objects::nonNull).map(v -> UUID.fromString(v.toString())).toList()`. `Stream.toList()` is unmodifiable, but the three other return branches of the same method (the `Optional` and scalar branches immediately below, and the `parameterName.equalsIgnoreCase("uuid")` branch further down) all deliberately wrap in `new ArrayList<>(List.of(...))`, i.e. the method's established contract for its own callers is a *mutable* `List<UUID>`. The mismatch is pre-existing — the PR only inserted the `filter` — but it is now on a changed line, and the inconsistency is the classic latent footgun described by *Effective Java Item 50* context: a future caller that follows the mutable-contract branch hits `UnsupportedOperationException` only for list-valued parameters. I confirmed every current consumer is read-only, so nothing is broken today: `constructResourceRecord` defensively re-wraps with `new ArrayList<>(...)` before touching the elements, `getObjectIdentityFromMDC` only calls `getFirst()`, and `AuditLogEnhancer.enrichObjectUuids` (AuditLogEnhancer.java:51-67) only streams. Worth settling while the line is already being touched.

_Suggested:_ Make the method return one mutability contract — either wrap this branch as `new ArrayList<>(listValues.stream().filter(Objects::nonNull).map(v -> UUID.fromString(v.toString())).toList())` like its siblings, or drop the `new ArrayList<>` wrappers from the other three branches and document the return as unmodifiable.

ivosh (dismissed):

src/main/java/com/otilm/core/service/impl/CbomServiceImpl.java
vyskocilm: Fixed in e8cd15e6. Comment now Jackson fact plus per-item-verdict decision. Narration of filter, guard, loop branch gone.

src/main/java/com/otilm/core/aop/AuditLogAspect.java
vyskocilm: Fixed in e8cd15e6. List branch wrapped in `new ArrayList<>(...)`. All four returns one contract — mutable, like siblings.

src/test/java/com/otilm/core/integration/aop/AuditLogAspectITest.java
vyskocilm: Fixed in e8cd15e6. Both test Javadocs cut. `nullUuidElement_isSkippedByTheAudit_andTheCallStillAnswers` now bare. `testBulkDeleteCbom_NullElement` keeps one line for what name does not say — neighbours still deleted. Rationale stays on two production sites.

ivosh (approved):

